PR #268, opened 2019-03-19
BG-11323: Deprecate conflicting params in generateAddress

paddyquinn:
Currently there are 4(!) params in generateAddress that conflict in
specifying what kind of address to create. Chain is the ultimate
source of truth in determining what kind of unspent (and thus what
kind of address) is created by our platform. This commit deprecates
the addressType, segwit, and bech32 params from the generateAddress
function. However, since they are already in the SDK API they are
not removed in this commit, nor is their functionality changed.
There are only two functionality changes to the function. The
first is a check that the addressType derived from addressType,
segwit, and bech32 matches the chain passed. Otherwise an error is
thrown. The second is to check that wrapped segwit is supported
when creating an address for a wrapped segwit unspent. This check
already existed for native segwit and also should have been in
place for wrapped segwit.

We bring in the new @bitgo/unspents library in this commit as well
as it is used to do type checking in generateAddress.
AbstractUtxoCoin.AddressTypes and
AbstractUtxoCoin.AddressTypeChains are elided for canonical enums
defined in this library that will be used across SDK and other
internal bitgo infrastructure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "generate_drone": "drone jsonnet --stream"
   },
   "dependencies": {
+    "@bitgo/unspents": "0.5.0",
     "argparse": "1.0.10",
     "assert": "0.4.9",
     "big.js": "3.1.3",
--- a/src/v2/coins/abstractUtxoCoin.js
+++ b/src/v2/coins/abstractUtxoCoin.js
@@ -11,6 +11,7 @@ const _ = require('lodash');
 const RecoveryTool = require('../recovery');
 const errors = require('../../errors');
 const debug = require('debug')('bitgo:v2:utxo');
+const { Codes } = require('@bitgo/unspents');
 
 class AbstractUtxoCoin extends BaseCoin {
   constructor(network) {
@@ -26,7 +27,17 @@ class AbstractUtxoCoin extends BaseCoin {
   }
 
   static get validAddressTypes() {
-    return _.values(this.AddressTypes);
+    const validAddressTypes = [];
+    _.forEach(Object.keys(Codes.UnspentTypeTcomb.meta.map), function(addressType) {
+      try {
+        Codes.forType(addressType);
+        validAddressTypes.push(addressType);
+      } catch (e) {
+        // Do nothing. Codes.forType will throw if the address type has no chain codes, meaning it is invalid on the
+        // BitGo platform and should not be added to the validAddressTypes array.
+      }
+    });
+    return validAddressTypes;
   }
 
   /**
@@ -133,11 +144,11 @@ class AbstractUtxoCoin extends BaseCoin {
   static inferAddressType(addressDetails) {
     if (_.isObject(addressDetails.coinSpecific)) {
       if (_.isString(addressDetails.coinSpecific.redeemScript) && _.isString(addressDetails.coinSpecific.witnessScript)) {
-        return this.AddressTypes.P2SH_P2WSH;
+        return Codes.UnspentTypeTcomb('p2shP2wsh');
       } else if (_.isString(addressDetails.coinSpecific.redeemScript)) {
-        return this.AddressTypes.P2SH;
+        return Codes.UnspentTypeTcomb('p2sh');
       } else if (_.isString(addressDetails.coinSpecific.witnessScript)) {
-        return this.AddressTypes.P2WSH;
+        return Codes.UnspentTypeTcomb('p2wsh');
       }
     }
     return null;
@@ -516,7 +527,15 @@ class AbstractUtxoCoin extends BaseCoin {
   }
 
   /**
-   * Indicates whether a coin supports pay-to-witness script hash addresses
+   * Indicates whether a coin supports wrapped segwit outputs
+   * @returns {boolean}
+   */
+  supportsP2shP2wsh() {
+    return false;
+  }
+
+  /**
+   * Indicates whether a coin supports native segwit outputs
    * @returns {boolean}
    */
   supportsP2wsh() {
@@ -524,31 +543,60 @@ class AbstractUtxoCoin extends BaseCoin {
   }
 
   /**
+   * TODO(BG-11487): Remove addressType, segwit, and bech32 params in SDKv6
    * Generate an address for a wallet based on a set of configurations
-   * @param addressType
-   * @param keychains Array of objects with xpubs
-   * @param threshold Minimum number of signatures
-   * @param chain Derivation chain
-   * @param index Derivation index
-   * @param segwit
-   * @param bech32
+   * @param addressType {string}   Deprecated
+   * @param keychains   {[object]} Array of objects with xpubs
+   * @param threshold   {number}   Minimum number of signatures
+   * @param chain       {number}   Derivation chain (see https://github.com/BitGo/unspents/blob/master/src/codes.ts for
+   *                                                 the corresponding address type of a given chain code)
+   * @param index       {number}   Derivation index
+   * @param segwit      {boolean}  Deprecated
+   * @param bech32      {boolean}  Deprecated
    * @returns {{chain: number, index: number, coin: number, coinSpecific: {outputScript, redeemScript}}}
    */
   generateAddress({ addressType, keychains, threshold, chain, index, segwit, bech32 }) {
-    if (addressType === this.constructor.AddressTypes.P2WSH && !this.supportsP2wsh()) {
-      throw new errors.P2wshUnsupportedError();
+    let derivationChain = 0;
+    if (_.isInteger(chain) && chain > 0) {
+      derivationChain = chain;
     }
 
-    if (!_.isUndefined(addressType) && !this.constructor.validAddressTypes.includes(addressType)) {
-      throw new errors.UnsupportedAddressTypeError();
-    } else if (_.isUndefined(addressType)) {
-      addressType = this.constructor.AddressTypes.P2SH;
+    if (_.isUndefined(addressType)) {
+      addressType = Codes.UnspentTypeTcomb('p2sh');
       if (_.isBoolean(segwit) && segwit) {
-        addressType = this.constructor.AddressTypes.P2SH_P2WSH;
+        addressType = Codes.UnspentTypeTcomb('p2shP2wsh');
       }
       if (_.isBoolean(bech32) && bech32) {
-        addressType = this.constructor.AddressTypes.P2WSH;
+        addressType = Codes.UnspentTypeTcomb('p2wsh');
       }
+    }
+
+    switch (addressType) {
+      case Codes.UnspentTypeTcomb('p2sh'):
+        if (!Codes.isP2sh(derivationChain)) {
+          throw new errors.AddressTypeChainMismatchError(addressType, derivationChain);
+        }
+        break;
+      case Codes.UnspentTypeTcomb('p2shP2wsh'):
+        if (!this.supportsP2shP2wsh()) {
+          throw new errors.P2shP2wshUnsupportedError();
+        }
+
+        if (!Codes.isP2shP2wsh(derivationChain)) {
+          throw new errors.AddressTypeChainMismatchError(addressType, derivationChain);
+        }
+        break;
+      case Codes.UnspentTypeTcomb('p2wsh'):
+        if (!this.supportsP2wsh()) {
+          throw new errors.P2wshUnsupportedError();
+        }
+
+        if (!Codes.isP2wsh(derivationChain)) {
+          throw new errors.AddressTypeChainMismatchError(addressType, derivationChain);
+        }
+        break;
+      default:
+        throw new errors.UnsupportedAddressTypeError();
     }
 
     let signatureThreshold = 2;
@@ -562,11 +610,6 @@ class AbstractUtxoCoin extends BaseCoin {
       }
     }
 
-    let derivationChain = 0;
-    if (_.isInteger(chain) && chain > 0) {
-      derivationChain = chain;
-    }
-
     let derivationIndex = 0;
     if (_.isInteger(index) && index > 0) {
       derivationIndex = index;
@@ -578,7 +621,6 @@ class AbstractUtxoCoin extends BaseCoin {
 
     const inputScript = bitcoin.script.multisig.output.encode(signatureThreshold, derivedKeys);
     const inputScriptHash = bitcoin.crypto.hash160(inputScript);
-    let outputScript = bitcoin.script.scriptHash.output.encode(inputScriptHash);
 
     const addressDetails = {
       chain: derivationChain,
@@ -588,23 +630,35 @@ class AbstractUtxoCoin extends BaseCoin {
       addressType
     };
 
-    addressDetails.coinSpecific.redeemScript = inputScript.toString('hex');
-
-    if (['p2sh-p2wsh', 'p2wsh'].includes(addressType)) {
+    function createWitnessProgram(inputScript) {
       const witnessScriptHash = bitcoin.crypto.sha256(inputScript);
-      const redeemScript = bitcoin.script.witnessScriptHash.output.encode(witnessScriptHash);
-      const redeemScriptHash = bitcoin.crypto.hash160(redeemScript);
-      outputScript = bitcoin.script.scriptHash.output.encode(redeemScriptHash);
-      addressDetails.coinSpecific.witnessScript = inputScript.toString('hex');
-      addressDetails.coinSpecific.redeemScript = redeemScript.toString('hex');
-      if (addressType === 'p2wsh') {
-        outputScript = redeemScript;
-        delete addressDetails.coinSpecific.redeemScript;
-      }
+      return bitcoin.script.witnessScriptHash.output.encode(witnessScriptHash);
+    }
+
+    let outputScript;
+    switch (addressType) {
+      case Codes.UnspentTypeTcomb('p2sh'):
+        outputScript = bitcoin.script.scriptHash.output.encode(inputScriptHash);
+        addressDetails.coinSpecific.redeemScript = inputScript.toString('hex');
+        break;
+      case Codes.UnspentTypeTcomb('p2shP2wsh'):
+        const redeemScript = createWitnessProgram(inputScript);
+        const redeemScriptHash = bitcoin.crypto.hash160(redeemScript);
+        outputScript = bitcoin.script.scriptHash.output.encode(redeemScriptHash);
+        addressDetails.coinSpecific.redeemScript = redeemScript.toString('hex');
+        addressDetails.coinSpecific.witnessScript = inputScript.toString('hex');
+        break;
+      case Codes.UnspentTypeTcomb('p2wsh'):
+        outputScript = createWitnessProgram(inputScript);
+        addressDetails.coinSpecific.witnessScript = inputScript.toString('hex');
+        break;
+      default:
+        throw new Error(`unexpected addressType ${addressType}`);
     }
 
     addressDetails.coinSpecific.outputScript = outputScript.toString('hex');
-    addressDetails.address = this.getCoinLibrary().address.fromOutputScript(outputScript, this.network);
+    addressDetails.address =
+      this.getCoinLibrary().address.fromOutputScript(outputScript, this.network);
 
     return addressDetails;
   }
@@ -1020,7 +1074,7 @@ class AbstractUtxoCoin extends BaseCoin {
       }
 
       if (hasSigScript) {
-        // p2sh or p2sh-p2wsh
+        // p2sh or p2shP2wsh
         const nonEmptySignatures = sigScript.signatures.filter((sig) => !_.isEmpty(sig));
         const validSignatures = nonEmptySignatures.map((sig, sigIndex) => {
           if (_.isEmpty(sig)) {
@@ -1032,7 +1086,7 @@ class AbstractUtxoCoin extends BaseCoin {
           };
 
           if (hasWitnessScript) {
-            // p2sh-p2wsh
+            // p2shP2wsh
             const parentTxId = Buffer.from(input.hash).reverse().toString('hex');
             const inputId = `${parentTxId}:${input.index}`;
             const amount = unspentValues[inputId];
@@ -1129,8 +1183,8 @@ class AbstractUtxoCoin extends BaseCoin {
    * - krsProvider: necessary if backup key is held by KRS
    * - recoveryDestination: target address to send recovered funds to
    * - scan: the amount of consecutive addresses without unspents to scan through before stopping
-   * - ignoreAddressTypes: (optional) array of AddressTypes to ignore. these are strings defined in AbstractUtxoCoin.AddressTypes
-   *        for example: ['p2sh-p2wsh', 'p2wsh'] will prevent code from checking for wrapped-segwit and native-segwit chains on the public block explorers
+   * - ignoreAddressTypes: (optional) array of AddressTypes to ignore. these are strings defined in Codes.UnspentTypeTcomb
+   *        for example: ['p2shP2wsh', 'p2wsh'] will prevent code from checking for wrapped-segwit and native-segwit chains on the public block explorers
    * @param callback
    */
   recover(params, callback) {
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -41,6 +41,7 @@
     "bignumber.js": "8.0.1",
     "bitcoinjs-message": "2.0.0",
     "bitgo-utxo-lib": "1.3.1",
+    "@bitgo/unspents": "git+https://github.com/BitGo/unspents#e778b07e93d666b8eb62471edff2e066ba508804",
```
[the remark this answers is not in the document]
paddyquinn: Done

test/v2/unit/coins/ltc.js
```diff
@@ -120,19 +119,21 @@ describe('LTC:', function() {
...
 
       [generatedAddress, generatedTestAddress].forEach((currentAddress) => {
-        currentAddress.chain.should.equal(0);
+        currentAddress.chain.should.equal(chain);
         currentAddress.index.should.equal(0);
-        currentAddress.coinSpecific.outputScript.should.equal('a91426e34781478f08fff903cb70ae67311c3f9bc6a987');
-        currentAddress.coinSpecific.redeemScript.should.equal('00209a10eb58331e95333f4a6eafd5f03e442e17e0986c824e392642e872f431b7ef');
-        currentAddress.coinSpecific.witnessScript.should.equal('5221037acffd52bb7c39a4ac3d4c01af33ce0367afec45347e332edca63a38d1fb2e472102658831a87322b3583515ca8725841335505755ada53ee133c70a6b4b8d3978702102641ee6557561c9038242cafa7f538070d7646a969bcf6169f9950abfcfefd6b853ae');
+        currentAddress.coinSpecific.outputScript.should.equal('a9147ff13f3faeba4d439ef40604f7c127951e77eb6a87');
+        currentAddress.coinSpecific.redeemScript.should.equal('00207aad7d57b238a09b5daa10ff47c54483b7f2ad47f3f0c0aa230958b9df334260');
```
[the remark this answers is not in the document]
paddyquinn: These scripts are changing because the chain code is changing. With these changes I can not change the inputs such that the scripts stay the same because the tests generated a p2shP2wsh address on chain 0, which will now throw with these changes due to a mismatch in address type and chain code.

paddyquinn:
@tylerlevine These changes should now be backwards compatible. Originally, I had `addressType` default to its corresponding chain code if it was not present. However, this broke backwards compatibility because prior to this PR `addressType` defaulted to `p2sh` if it was not present. I reverted this change such that this PR does not change the default `addressType` now. The only changed functionality now is that if a mismatched `addressType` and `chain` are passed we will throw instead of generating an address on the wrong chain.

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -517,41 +513,78 @@ class AbstractUtxoCoin extends BaseCoin {
...
-   * @param segwit
-   * @param bech32
+   * @param addressType {string}   Deprecated
+   * @param keychains   {[object]} Array of objects with xpubs
+   * @param threshold   {number}   Minimum number of signatures
+   * @param chain       {number}   Derivation chain (see https://github.com/BitGo/unspents/blob/master/src/codes.ts for
+   *                                                 the corresponding address type of a given chain code)
+   * @param index       {number}   Derivation index
+   * @param segwit      {boolean}  Deprecated
+   * @param bech32      {boolean}  Deprecated
```
Xekyo: Kinda sad that we had the parameter and are deprecating it before it ever got used in production. :hushed:

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -517,41 +513,78 @@ class AbstractUtxoCoin extends BaseCoin {
...
-      throw new errors.UnsupportedAddressTypeError();
-    } else if (_.isUndefined(addressType)) {
-      addressType = this.constructor.AddressTypes.P2SH;
+    if (_.isUndefined(addressType)) {
+      addressType = Codes.typeForCode(derivationChain);
       if (_.isBoolean(segwit) && segwit) {
-        addressType = this.constructor.AddressTypes.P2SH_P2WSH;
+        addressType = Codes.UnspentTypeTcomb('p2shP2wsh');
       }
       if (_.isBoolean(bech32) && bech32) {
```
Xekyo: Since nobody ever actually used `bech32`, could we perhaps replace this with `p2wsh` from the get-go?

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -591,14 +619,14 @@ class AbstractUtxoCoin extends BaseCoin {
...
-    if ([this.constructor.AddressTypes.P2SH_P2WSH, this.constructor.AddressTypes.P2WSH].includes(addressType)) {
+    if ([Codes.UnspentTypeTcomb('p2shP2wsh'), Codes.UnspentTypeTcomb('p2wsh')].includes(addressType)) {
       const witnessScriptHash = bitcoin.crypto.sha256(inputScript);
       const redeemScript = bitcoin.script.witnessScriptHash.output.encode(witnessScriptHash);
       const redeemScriptHash = bitcoin.crypto.hash160(redeemScript);
       outputScript = bitcoin.script.scriptHash.output.encode(redeemScriptHash);
       addressDetails.coinSpecific.witnessScript = inputScript.toString('hex');
       addressDetails.coinSpecific.redeemScript = redeemScript.toString('hex');
-      if (addressType === this.constructor.AddressTypes.P2WSH) {
+      if (addressType === Codes.UnspentTypeTcomb('p2wsh')) {
```
Xekyo: Maybe we should refactor to do the case differentiation before creating the redeemscript on the address in the first place and only create it when it's a wrapped segwit address.

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -1241,7 +1269,7 @@ class AbstractUtxoCoin extends BaseCoin {
 
       // By default, we will ignore P2WSH until we officially support it
       if (_.isUndefined(params.ignoreAddressTypes)) {
-        params.ignoreAddressTypes = [AbstractUtxoCoin.AddressTypes.P2WSH];
+        params.ignoreAddressTypes = [Codes.UnspentTypeTcomb('p2wsh')];
```
Xekyo: Aren't we moving to officially support p2wsh now? :innocent:

Might want to put a fixme here to link the issue with which we'll enable p2wsh.

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -1262,11 +1290,19 @@ class AbstractUtxoCoin extends BaseCoin {
 
       const queries = [];
 
-      _.forEach(AbstractUtxoCoin.AddressTypes, function(addressType) {
+      _.forEach(Object.keys(Codes.UnspentTypeTcomb.meta.map), function(addressType) {
```
Xekyo: Aren't key and value in the UnspentTypes identical anyway?

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -1262,11 +1290,19 @@ class AbstractUtxoCoin extends BaseCoin {
...
         // If we aren't ignoring the address type, we derive the public key and construct the query for the main and change indices
```
Xekyo: Update comment to use "external" and "internal" terminology.

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -1262,11 +1290,19 @@ class AbstractUtxoCoin extends BaseCoin {
...
-          const changeIndex = AbstractUtxoCoin.AddressTypeChains[addressType].internal;
+          let codes;
+          try {
+            codes = Codes.forType(Codes.UnspentTypeTcomb(addressType));
+          } catch (e) {
+            // The unspent type is not supported by bitgo so attempting to get its chain codes throws. Catch that error
+            // and continue.
+            return;
+          }
+          const mainIndex = codes[Codes.PurposeTcomb('external')];
```
Xekyo: `externalIndex` or better yet `externalCode`?

src/v2/coins/btg.js
```diff
@@ -26,6 +26,14 @@ class Btg extends Btc {
     return false;
   }
 
+  supportsP2shP2wsh() {
```
Xekyo: How did we previously determine whether a coin had wrapped segwit support? o.0

test/v2/unit/coins/btc.js
```diff
@@ -81,19 +81,21 @@ describe('BTC:', function() {
     });
 
     it('should generate p2sh-wrapped segwit address', function() {
-      const generatedAddress = coin.generateAddress({ keychains, addressType: AbstractUtxoCoin.AddressTypes.P2SH_P2WSH });
-      const generatedTestAddress = testCoin.generateAddress({ keychains, addressType: AbstractUtxoCoin.AddressTypes.P2SH_P2WSH });
+      const addressType = Codes.UnspentTypeTcomb('p2shP2wsh');
+      const chain = Codes.forType(addressType)[Codes.PurposeTcomb('external')];
+      const generatedAddress = coin.generateAddress({ keychains, addressType, chain });
```
Xekyo: If we're moving to use chain, why would we additionally use `addressType`?

test/v2/unit/coins/btc.js
```diff
@@ -81,19 +81,21 @@ describe('BTC:', function() {
...
+      const generatedTestAddress = testCoin.generateAddress({ keychains, addressType, chain });
 
       [generatedAddress, generatedTestAddress].forEach((currentAddress) => {
-        currentAddress.chain.should.equal(0);
+        currentAddress.chain.should.equal(chain);
```
Xekyo: Oh wow, we tested a wrapped segwit address on chain 0?! Good catch.

test/v2/unit/coins/btc.js
```diff
@@ -144,25 +147,27 @@ describe('BTC:', function() {
     });
 
     it('should generate 3/3 custom chain p2shP2sh address', function() {
-      const generatedAddress = coin.generateAddress({ keychains, threshold: 3, addressType: AbstractUtxoCoin.AddressTypes.P2SH_P2WSH, chain: 20, index: 756 });
+      const addressType = Codes.UnspentTypeTcomb('p2shP2wsh');
```
Xekyo: Not sure why we're testing 3-of-3 addresses here.

test/v2/unit/recovery.js
```diff
@@ -47,7 +47,7 @@ describe('Recovery:', function() {
         walletPassphrase: TestV2BitGo.V2.TEST_WALLET1_PASSCODE,
         recoveryDestination: '2NB5Ynem6iNvA6GBLZwRxwid3Kui33729Nw',
         scan: 5,
-        ignoreAddressTypes: ['p2wsh', 'pwshP2wsh']
+        ignoreAddressTypes: ['p2wsh', 'p2shP2wsh']
```
Xekyo: Why do we ignore segwit addresses in our recovery tests? Do we have tests for this elsewhere?

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -548,7 +548,7 @@ class AbstractUtxoCoin extends BaseCoin {
     }
 
     if (_.isUndefined(addressType)) {
-      addressType = Codes.typeForCode(derivationChain);
+      addressType = Codes.UnspentTypeTcomb('p2sh');
```
Xekyo: This is a change in behavior if `chain` is not zero. This change should rather be applied to line 545.

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -590,14 +591,14 @@ class AbstractUtxoCoin extends BaseCoin {
 
     addressDetails.coinSpecific.redeemScript = inputScript.toString('hex');
 
-    if (['p2sh-p2wsh', 'p2wsh'].includes(addressType)) {
+    if ([this.constructor.AddressTypes.P2SH_P2WSH, this.constructor.AddressTypes.P2WSH].includes(addressType)) {
```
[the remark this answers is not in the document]
paddyquinn: Outdated. I will squash commits before merging to make them more intuitive.

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -517,41 +513,78 @@ class AbstractUtxoCoin extends BaseCoin {
...
-   * @param index Derivation index
-   * @param segwit
-   * @param bech32
+   * @param addressType {string}   Deprecated
+   * @param keychains   {[object]} Array of objects with xpubs
+   * @param threshold   {number}   Minimum number of signatures
+   * @param chain       {number}   Derivation chain (see https://github.com/BitGo/unspents/blob/master/src/codes.ts for
+   *                                                 the corresponding address type of a given chain code)
+   * @param index       {number}   Derivation index
+   * @param segwit      {boolean}  Deprecated
```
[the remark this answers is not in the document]
paddyquinn: This is v2 and this never hits the platform.

src/v2/coins/abstractUtxoCoin.js
paddyquinn: I don't think that gets us anything. If we feel confident that this is not being used by customers then I would suggest removing it altogether as `chain` should be the canonical source. Otherwise we should just leave it as is.

src/v2/coins/abstractUtxoCoin.js
paddyquinn: Not sure your point here. Without `Object.keys` we would be looping over a map.

src/v2/coins/btg.js
paddyquinn: We didn't.

test/v2/unit/coins/btc.js
paddyquinn: @tylerlevine this is why test scripts need to change.

src/v2/coins/abstractUtxoCoin.js
paddyquinn: If you mean from line 545 of the file before these changes then that is where it is applied. I just switched the order of the `if` statement such that we don't need to have an `else if` anymore.

test/v2/unit/coins/btc.js
paddyquinn: Because right now we default to using `p2sh` as the `addressType` so if we pass a `p2shP2wsh` chain with no `addressType` we will get an `AddressTypeChainMismatchError`.

test/v2/unit/coins/btc.js
paddyquinn: Not sure either but not sure it is in scope for this PR or native segwit generally.

test/v2/unit/recovery.js
paddyquinn: I believe it is because we only mock out p2sh txes for recovery tests. Again, I think fixing this is out of scope for this PR and for native segwit generally.

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -517,41 +513,78 @@ class AbstractUtxoCoin extends BaseCoin {
   }
 
   /**
-   * Indicates whether a coin supports pay-to-witness script hash addresses
+   * Indicates whether a coin supports pay-to-witness-script-hash outputs wrapped in pay-to-script-hash outputs
```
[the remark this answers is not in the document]
paddyquinn: Done

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -517,41 +513,78 @@ class AbstractUtxoCoin extends BaseCoin {
...
+   * @returns {boolean}
+   */
+  supportsP2shP2wsh() {
+    return false;
+  }
+
+  /**
+   * Indicates whether a coin supports pay-to-witness-script-hash outputs
```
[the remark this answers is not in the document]
paddyquinn: Done

src/v2/coins/abstractUtxoCoin.js
paddyquinn: Done. Please re-review.

src/v2/coins/abstractUtxoCoin.js
paddyquinn: Done

src/v2/coins/abstractUtxoCoin.js
paddyquinn: Done

src/v2/coins/abstractUtxoCoin.js
paddyquinn: Done

test/v2/unit/coins/btc.js
Xekyo: Okay

test/v2/unit/coins/btc.js
Xekyo: Let's leave it alone at this time, but if it causes work later, we should remove it, because that is not a use case we actually offer.

src/v2/coins/abstractUtxoCoin.js
Xekyo: If i'm missing something, please elaborate, but previously, when `chain` was `> 0`, `addressType` would be set in line with `derivationChain`. Now it is always set to p2sh. This seems to be a breaking change.

src/v2/coins/abstractUtxoCoin.js
Xekyo: How would anyone be using `bech32` when it was always disabled to date?

src/v2/coins/abstractUtxoCoin.js
Xekyo: Nevermind, I was looking at the commits one by one, but this was changed and reverted in different commits.

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -590,16 +619,15 @@ class AbstractUtxoCoin extends BaseCoin {
...
       const witnessScriptHash = bitcoin.crypto.sha256(inputScript);
       const redeemScript = bitcoin.script.witnessScriptHash.output.encode(witnessScriptHash);
-      const redeemScriptHash = bitcoin.crypto.hash160(redeemScript);
-      outputScript = bitcoin.script.scriptHash.output.encode(redeemScriptHash);
-      addressDetails.coinSpecific.witnessScript = inputScript.toString('hex');
-      addressDetails.coinSpecific.redeemScript = redeemScript.toString('hex');
-      if (addressType === 'p2wsh') {
-        outputScript = redeemScript;
-        delete addressDetails.coinSpecific.redeemScript;
+      outputScript = redeemScript;
```
Xekyo: I don't like that we set an incorrect `outputScript` for wrapped segwit first and then overwrite it later. Rather, I would prefer if the `outputScript` would be set only explicitly for the `addressType` that it will finally apply to. Could we please reorder these lines such that there is no overwriting and explicit disambiguation of the two cases?

src/v2/coins/abstractUtxoCoin.js
Xekyo: Just looked complicated and I thought we could just get to a list of the keys or values easier.

src/v2/coins/abstractUtxoCoin.js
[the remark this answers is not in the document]
Xekyo: Excellent.

src/v2/coins/abstractUtxoCoin.js
paddyquinn: Done

src/v2/coins/abstractUtxoCoin.js
Xekyo: Clarified.

package.json
```diff
@@ -42,6 +42,7 @@
     "bignumber.js": "8.0.1",
     "bitcoinjs-message": "2.0.0",
     "bitgo-utxo-lib": "1.3.1",
+    "@bitgo/unspents": "0.5.0",
```
tylerlevine: Should be at the top (these are sorted alphabetically, including special characters)

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -25,10 +26,6 @@ class AbstractUtxoCoin extends BaseCoin {
     return this._network;
   }
 
-  static get validAddressTypes() {
```
tylerlevine: Were there no callers of this function? This is part of the public API and thus would make this pr a major version bump.

test/v2/unit/coins/btc.js
```diff
@@ -144,25 +147,27 @@ describe('BTC:', function() {
...
-        addressType: AbstractUtxoCoin.AddressTypes.P2SH_P2WSH,
-        chain: 20,
+        addressType: Codes.UnspentTypeTcomb('p2shP2wsh'),
+        chain,
         index: 756
       });
 
       [generatedAddress, generatedTestAddress].forEach((currentAddress) => {
-        currentAddress.chain.should.equal(20);
+        currentAddress.chain.should.equal(chain);
```
tylerlevine: How come this test needs changing? Seems like it generated a p2wsh address on chain 20 before. Isn't the updated test doing the same thing?

test/v2/unit/coins/ltc.js
```diff
@@ -154,25 +155,27 @@ describe('LTC:', function() {
...
-        addressType: AbstractUtxoCoin.AddressTypes.P2SH_P2WSH,
-        chain: 20,
+        addressType,
+        chain,
         index: 756
       });
 
       [generatedAddress, generatedTestAddress].forEach((currentAddress) => {
-        currentAddress.chain.should.equal(20);
+        currentAddress.chain.should.equal(chain);
```
tylerlevine: Same here, how come this needs changing?

package.json
paddyquinn: Done

test/v2/unit/coins/btc.js
paddyquinn: Previously it generated a `p2shP2wsh` address on chain 20, not a `p2wsh` address.

test/v2/unit/coins/ltc.js
paddyquinn: Same as above.

src/v2/coins/abstractUtxoCoin.js
paddyquinn: The need for this function was removed when we started using the tcomb validation from the unspents lib. Should I add it back for the sake of backwards compatibility?

paddyquinn:
@tylerlevine once you give your blessing I will merge

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -26,7 +27,17 @@ class AbstractUtxoCoin extends BaseCoin {
   }
 
   static get validAddressTypes() {
-    return _.values(this.AddressTypes);
+    const validAddressTypes = [];
+    _.forEach(Object.keys(Codes.UnspentTypeTcomb.meta.map), function(addressType) {
+      try {
+        Codes.forType(addressType);
```
OttoAllmendinger: we should add a `Codes.hasPlatformSupport(unspentType)` or something

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -1261,15 +1315,23 @@ class AbstractUtxoCoin extends BaseCoin {
...
-          queries.push(queryBlockchainUnspentsPath(changeKey, '/0/0/' + changeIndex));
+          let codes;
+          try {
+            codes = Codes.forType(Codes.UnspentTypeTcomb(addressType));
+          } catch (e) {
+            // The unspent type is not supported by bitgo so attempting to get its chain codes throws. Catch that error
+            // and continue.
+            return;
+          }
+          const externalChainCode = codes[Codes.PurposeTcomb('external')];
```
OttoAllmendinger: why not `codes.external` instead?

src/v2/coins/abstractUtxoCoin.js
```diff
@@ -133,11 +144,11 @@ class AbstractUtxoCoin extends BaseCoin {
   static inferAddressType(addressDetails) {
     if (_.isObject(addressDetails.coinSpecific)) {
       if (_.isString(addressDetails.coinSpecific.redeemScript) && _.isString(addressDetails.coinSpecific.witnessScript)) {
-        return this.AddressTypes.P2SH_P2WSH;
+        return Codes.UnspentTypeTcomb('p2shP2wsh');
```
OttoAllmendinger: the only reason we should use this instead `Codes.UnspentType.p2shP2wsh` is that we get a runtime error on typo (except in `NODE_ENV=production`)

However with TypeScript we should get a compile-time error anyway, no?

src/v2/coins/abstractUtxoCoin.js
paddyquinn: Fixed in #281.

src/v2/coins/abstractUtxoCoin.js
paddyquinn: This is a js file. `Codes.UnspentType.pwshP2wsh` only exists for ts files.

paddyquinn: So actually this is now a ts file. However, it seems that `Codes.UnspentType` is not exported. I tried to hack around for a bit to get it exported but my typescript is poor and I couldn't figure it out. If we can export it then I agree that we should use it here.

paddyquinn: Put in these 2 tickets for this: https://bitgoinc.atlassian.net/browse/BG-11667 & https://bitgoinc.atlassian.net/browse/BG-11668

PR #270, opened 2019-03-19
BRE-380: Configure drone and remove travis

tylerlevine:
Signed-off-by: Tyler Levine <tyler@bitgo.com>

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -1,0 +1,118 @@
+local branches() = {
+  branch: [
+    "master",
+    "rel/*",
+    "prod/production"
+  ],
+};
+
+local BuildInfo(version, limit_branches=false) = {
+  name: "build information",
+  image: "node:" + version,
+  commands: [
+    "node --version",
+    "npm --version",
+  ],
+  [if limit_branches then "when"]: branches(),
+};
+
+local Install(version, limit_branches=false) = {
+  name: "install",
+  image: "node:" + version,
+  commands: [
+    "npm install",
+  ],
+  [if limit_branches then "when"]: branches(),
+};
+
+local UploadCoverage(version, tag="untagged", limit_branches=true) = {
+  name: "upload coverage",
+  image: "node:"  + version,
+  environment: {
+    CODECOV_TOKEN: { from_secret: "codecov" },
+  },
+  commands: [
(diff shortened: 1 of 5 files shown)


.drone.yml
```diff
@@ -0,0 +1,231 @@
...
+  - npm run test-node
+  environment:
+    BITGOJS_TEST_PASSWORD:
+      from_secret: password
+
+- name: upload coverage
+  image: node:11
+  commands:
+  - "npx nyc report --reporter=text-lcov > coverage.lcov"
+  - "npx codecov -f coverage.lcov -B \"${DRONE_BUILD_NUMBER:-0}\" -t \"$CODECOV_TOKEN\""
```
drGrove: Shouldn't codecov just pull this from the environment? Do you have to be explicit?

.drone.jsonnet
```diff
@@ -0,0 +1,106 @@
+local BuildInfo(version) = {
```
drGrove: Can you explain the need for this file? I see you have a .drone.yaml below so i'm not sure what this does

.drone.yml
```diff
@@ -0,0 +1,231 @@
...
+  commands:
+  - npm run test-node
+  environment:
+    BITGOJS_TEST_PASSWORD:
+      from_secret: password
+
+- name: upload coverage
+  image: node:9
+  commands:
+  - "npx nyc report --reporter=text-lcov > coverage.lcov"
```
drGrove: Commands separated by - shouldn't need to be wrapped in strings

.drone.jsonnet
tylerlevine: The `.drone.yml` file that you see in this PR was auto generated from this `.drone.jsonnet` file using `drone jsonnet --stream`.

.drone.yml
tylerlevine: I seem to remember that this only gets set automatically for a few CI providers, and drone was not one of them. However, I'm having trouble finding that page again now. I'll take a look at it, as I'd like to add flags as well.

james8nguyen:
Just for the record, I had reviewed and approved the PR but did not do it fast enough. Got busy on some stuffs. Thanks.

PR #271, opened 2019-03-19
BG-11144 - Overwrite serverXpub in initialization

RobertBaron:
We want to have the possibility to overwrite the serverXpub is using for custom environments.

Also we need to expose the different environments that we have so we can swap keys.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,13 +6,16 @@
 const common = require('./common');
 const bitgo = module.exports;
 bitgo.BitGo = require('./bitgo.js');
-bitgo.Environments = common.Environments;
 
 // Expose bitcoin and sjcl
 bitgo.bitcoin = require('./bitcoin');
 bitgo.sjcl = require('./sjcl.min');
 bitgo.bs58 = require('bs58');
 bitgo.Buffer = Buffer;
+
+// Expose environments
+const _ = require('lodash');
+bitgo.Environments = _.cloneDeep(common.Environments);
 
 /**
  * Set the network, i.e. either "bitcoin" for production with real bitcoin, or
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -6,6 +6,7 @@
 const common = require('./common');
 const bitgo = module.exports;
 bitgo.BitGo = require('./bitgo.js');
+bitgo.Environments = common.Environments;
```
tylerlevine: Why are we exporting the environments?

RobertBaron: In the microservice the platform runs in local mode, so it uses the local key to sign the token.

Since we use a customRootURi in the client, the sdk sets the custom key...
What I did is to get the right config depending on where is running from the sdk

const env = BitGoJS.Environments['local']; 

and then when I am about to initiate the sdk I do something like

```
params =  {
      validate: false,
      microservicesUri: this.URI,
      customRootURI: this.URI,
      serverXpub: env.serverXpub,
    };
```

tylerlevine: What do you think about adding something like this to the bitgo object instead of exporting from the root? This would give us a little more flexibility.
```js
getEnvironmentConfig(env) {
  return common.Environments[env || this.env] || {};
}
```

tylerlevine: Oh, I understand now. You need the xpub as a prerequisite for creating the bitgo object.

src/index.js
```diff
@@ -6,6 +6,7 @@
 const common = require('./common');
 const bitgo = module.exports;
 bitgo.BitGo = require('./bitgo.js');
+bitgo.Environments = JSON.parse(JSON.stringify(common.Environments));
```
tylerlevine:
```js
const _ = require('lodash');
bitgo.Environments = _.cloneDeep(common.Environments);
```

PR #276, opened 2019-03-26
Don't fail wallet creation 1 in 256 times.

dpc:
https://bitgoinc.atlassian.net/browse/BG-11099


dpc:
There are two other places where this potentially might be needed:
```
src/bitgo.js
1599:  const privateKey = this._extensionKey.d.toBuffer();

src/util.js
80:    const ethPrivateKey = hdNode.keyPair.d.toBuffer();
```

PR #280, opened 2019-03-27
BG-11582 slight change to utxo recovery

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-11582

Allow user to recovery with the Backup Key - BitGo combination (previously only possible with key recovery services)


tylerlevine:
What does this have to do with the referenced ticket?

dannybitgo:
> What does this have to do with the referenced ticket?

The customer cant spend with their user key due to Ledger issues, but they have their backup key. We need to give them a tool to sign with their backup key. This allows them to do that. Then they can submit the half-signed tx to bitgo to complete the recovery.

PR #283, opened 2019-03-28
BG-10662 Condition on microservice url on admin access route

anto1990:
This is allow admin access to go to the auth service in the micro-service's.

### DIFF
--- a/src/bitgo.ts
+++ b/src/bitgo.ts
@@ -1364,7 +1364,9 @@ BitGo.prototype.addAccessToken = function(params, callback) {
 
   const bitgo = this;
 
-  const request = this.post(this.url('/user/accesstoken'));
+  const authUrl = this._microservicesUrl ? this.microservicesUrl('/api/v1/auth/accesstoken') : this.url('/user/accesstoken');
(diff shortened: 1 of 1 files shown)


src/bitgo.ts
```diff
@@ -1364,7 +1364,9 @@ BitGo.prototype.addAccessToken = function(params, callback) {
 
   const bitgo = this;
 
-  const request = this.post(this.url('/user/accesstoken'));
+  const authUrl = this._microservicesUrl ? this.microservicesUrl('/api/v1/auth/accesstoken') : this.url('/user/accesstoken');
+  const request = this.post(authUrl);
+  
   if (!bitgo._ecdhXprv) {
     // without a private key, the user cannot decrypt the new access token the server will send
     request.forceV1Auth = true;
```
dannybitgo: 10 lines below @ line 1383 we do:
```
request.verifyResponse(response);
```
will this still work?

tylerlevine: It should - the response should look like we are talking with bitgo's platform directly.

PR #286, opened 2019-03-29
BG-11572: Refactor parseSignatureScript for native segwit

paddyquinn:
JIRA: BG-11572

Technically, prior to this commit, parseSignatureScript will work
for native segwit. However, the code was confusing and purposely
misclassified native segwit inputs as p2sh inputs. This commit
cleans up `parseSignatureScript()` (including adding extensive
explanatory comments) and correctly classifies native segwit
inputs.

### DIFF
--- a/src/v2/coins/abstractUtxoCoin.ts
+++ b/src/v2/coins/abstractUtxoCoin.ts
@@ -686,8 +686,6 @@ class AbstractUtxoCoin extends BaseCoin {
     this.prepareTransactionBuilder(txb);
 
     const signatureIssues = [];
-    const p2wshIndices = [];
-
     for (let index = 0; index < transaction.ins.length; ++index) {
       debug('Signing input %d of %d', index + 1, transaction.ins.length);
       const currentUnspent = txPrebuild.txInfo.unspents[index];
@@ -745,7 +743,6 @@ class AbstractUtxoCoin extends BaseCoin {
       // after signature validation, prepare p2wsh setup
       if (isP2wsh) {
         transaction.setInputScript(index, Buffer.alloc(0));
-        p2wshIndices.push(index);
       }
 
       const isValidSignature = this.verifySignature(transaction, index, currentUnspent.value);
@@ -762,10 +759,6 @@ class AbstractUtxoCoin extends BaseCoin {
       error.code = 'input_signature_failure';
       error.signingErrors = signatureIssues;
       throw error;
-    }
-
-    for (const p2wshIndex of p2wshIndices) {
-      transaction.setInputScript(p2wshIndex, Buffer.alloc(0));
     }
 
     return {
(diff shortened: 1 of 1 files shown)


src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -833,46 +826,56 @@ class AbstractUtxoCoin extends BaseCoin {
...
+      const pubScript = bitcoin.script.pubKeyHash.output.encode(bitcoin.crypto.hash160(publicKey));
+
+      return { signatures, publicKeys, isSegwitInput, inputClassification, pubScript };
+    } else if (inputClassification === bitcoin.script.types.P2SH
+        || inputClassification === bitcoin.script.types.P2WSH) {
+      // Note the assumption here that if we have a p2sh or p2wsh input it will be multisig (appropriate because the
+      // BitGo platform only supports multisig within these types of inputs). Signatures are all but the last entry in
+      // the decompiledSigScript. The redeemScript/witnessScript (depending on which type of input this is) is the last
+      // entry in the decompiledSigScript (denoted here as the pubScript). The public keys are the second through
+      // antepenultimate entries in the decompiledPubScript. See below for a visual representation of the typical 2-of-3
```
OttoAllmendinger: > antepenultimate

beautiful

src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -833,46 +826,56 @@ class AbstractUtxoCoin extends BaseCoin {
...
+      //
+      // decompiledSigScript = 0 <sig1> <sig2> <pubScript>
+      // decompiledPubScript = 2 <pub1> <pub2> <pub3> 3 OP_CHECKMULTISIG
+      const signatures = decompiledSigScript.slice(0, -1);
+      const pubScript = _.last(decompiledSigScript);
+      const decompiledPubScript = bitcoin.script.decompile(pubScript);
+      // the second through antepenultimate entries
+      const publicKeys = decompiledPubScript.slice(1, -2);
+
+      return { signatures, publicKeys, isSegwitInput, inputClassification, pubScript };
```
OttoAllmendinger: please order it so the common fields `isSegwitInput` and `inputClassification` come first

also in the previous branch

src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -833,46 +826,56 @@ class AbstractUtxoCoin extends BaseCoin {
...
+      // Note the assumption here that if we have a p2sh or p2wsh input it will be multisig (appropriate because the
+      // BitGo platform only supports multisig within these types of inputs). Signatures are all but the last entry in
+      // the decompiledSigScript. The redeemScript/witnessScript (depending on which type of input this is) is the last
+      // entry in the decompiledSigScript (denoted here as the pubScript). The public keys are the second through
+      // antepenultimate entries in the decompiledPubScript. See below for a visual representation of the typical 2-of-3
+      // multisig setup:
+      //
+      // decompiledSigScript = 0 <sig1> <sig2> <pubScript>
+      // decompiledPubScript = 2 <pub1> <pub2> <pub3> 3 OP_CHECKMULTISIG
+      const signatures = decompiledSigScript.slice(0, -1);
```
OttoAllmendinger: let's assert the correct signature and pubscript lengths here

src/v2/coins/abstractUtxoCoin.ts
paddyquinn: i cannot take credit for this. It was present before this change. :grimacing:

src/v2/coins/abstractUtxoCoin.ts
paddyquinn: Done

src/v2/coins/abstractUtxoCoin.ts
paddyquinn: I'm not 100% sure we want to do that. Technically, prior to this someone could have been using this function to parse multisig p2sh inputs that had something other than a 2-of-3 configuration. Maybe SDK lord @tylerlevine could wiegh in here?

OttoAllmendinger: at least we should check it is actually a valid multisig though

1. last item should be `OP_CHECKMULTISIG`
2. signature count should match `decompiledSigScript[0]`
3. pubScript count should match `decompileSigScript[-2]`

and then `verifySignature` could make sure it is `2-of-3` or I guess `3-of-3`

tylerlevine: `parseSignatureScript` should (ideally) be agnostic to the type of script it's parsing. @OttoAllmendinger's suggestion is correct though - the script should be validated and it's permissible to throw if the script is invalid.

src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -833,46 +826,56 @@ class AbstractUtxoCoin extends BaseCoin {
    */
```
tylerlevine: I can't leave a comment on line 825, but the @ returns should be updated.

src/v2/coins/abstractUtxoCoin.ts
paddyquinn: Done

src/v2/coins/abstractUtxoCoin.ts
paddyquinn: Done

paddyquinn:
@tylerlevine The only functionality change is that `p2wsh` signature scripts that used to be incorrectly classified as `p2sh` are now correctly classified as `p2wsh`. The one instance where this classification was being used within the SDK (in `verifySignature`) was updated to reflect this change as well.

src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1,5 +1,7 @@
+import { strict as assert } from 'assert';
```
tylerlevine: I'd rather you throw an error instead of asserting (assertion error messages are not very useful).

src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -803,50 +798,74 @@ class AbstractUtxoCoin extends BaseCoin {
...
+      const len = decompiledPubScript.length;
+      const nSignatures = decompiledPubScript[0] - 80;
+      const nPubKeys = decompiledPubScript[len - 2] - 80;
+
+      // Due to a bug in the implementation of multisignature in the bitcoin protocol, a 0 is added to the signature
+      // script, so we add 1 when asserting the number of signatures matches the number of signatures expected by the
+      // pub script. Also, note that we consider a signature script with the the same number of signatures as public
+      // keys (+1 as noted above) valid because we use placeholder signatures when parsing a half-signed signature
+      // script.
+      assert(signatures.length === nSignatures + 1 || signatures.length === nPubKeys + 1);
```
tylerlevine: I'd rather you throw an error instead of asserting (assertion error messages are not very useful).

src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1,5 +1,7 @@
+import { strict as assert } from 'assert';
 import BaseCoin = require('../baseCoin');
 const config = require('../../config');
+const OPS = require('bitcoin-ops');
```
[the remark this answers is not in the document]
paddyquinn: Done

src/v2/coins/abstractUtxoCoin.ts
paddyquinn: Done

src/v2/coins/abstractUtxoCoin.ts
paddyquinn: Done

tylerlevine:
Looks fine to me, but please wait for a review from @OttoAllmendinger. I'll dismiss my review.

src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: actually we should add the typescript signature instead in the future

src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -803,50 +796,83 @@ class AbstractUtxoCoin extends BaseCoin {
...
+      }
+
+      if (publicKeys.length !== nPubKeys) {
+        throw new Error(`expected ${nPubKeys} public keys, got ${publicKeys.length}`);
+      }
 
-        const pubScript = bitcoin.script.pubKeyHash.output.encode(bitcoin.crypto.hash160(publicKey));
-        return { signatures, publicKeys, isSegwitInput, inputClassification, pubScript: pubScript };
+      const lastOpCode = decompiledPubScript[len - 1];
+      if (lastOpCode !== bitcoin.opcodes.OP_CHECKMULTISIG) {
```
OttoAllmendinger: actually I think we can use the bitcoin lib stuff to replace most of this here:

```
          const { redeemScript, redeemScriptSig } = utxolib.script.scriptHash.input.decode(input.script);
          const { m, pubKeys } = utxolib.script.multisig.output.decode(redeemScript);
          const multiSigScript = utxolib.script.multisig.input.decode(redeemScriptSig);
          // throw if `m !== multiSigScript.length`
```
(with `utxolib=bitcoin`)

paddyquinn: While I do prefer this as it would make our code a lot more readable, as it currently stands this would not work. `scriptHash.input.decode()` does not support half-signed multisig scripts. However, `parseSignatureScript` needs to be able to handle half-signed multisig scripts. The change in `bitgo-utxo-lib` needed to make `scriptHash.input.decode()` support half-signed multisig scripts would be small but I am not sure if that function intentionally did not handle those types of scripts. I would suggest leaving this code as is and discussing whether or not `scriptHash.input.decode()` should handle half-signed multisig scripts outside the scope of this PR.

PR #288, opened 2019-03-30
BG-11706: Update @bitgo/unspents to 0.5.1

tylerlevine:


package.json
```diff
@@ -35,7 +35,7 @@
     "generate_drone": "drone jsonnet --stream"
   },
   "dependencies": {
-    "@bitgo/unspents": "0.5.0",
+    "@bitgo/unspents": "0.5.1",
```
james8nguyen: We should use "^0.5.1" to minimize maintenance effort each time a new unspents version is published; otherwise, the release process would be circular (test new unspents version --> publish to npmjs --> upgrade SDK package.json --> test new SDK --> publish SDK) instead of in parallel.

james8nguyen: For future changes, we should use "^0.5.1" to minimize maintenance effort; however, I'm fine with this for now. Thanks you.

PR #292, opened 2019-04-01
BG-11726: Generate js files from ts files on prepare

tylerlevine:
* Exclude more unnecessary files from npm package

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -59,7 +59,6 @@ typings/
 # dotenv environment variables file
 .env
 
-
 # BitGo Repository Files
 lab
 example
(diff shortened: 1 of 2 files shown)


tylerlevine:
## Notice

This will require users of bitgojs who use a github clone url to run the prepublish script upon installation to run tsc. This can be done by running `npm explore bitgo -- npm run prepublish`.

Instead of using the github clone url to depending on bitgojs, we should instead recommend using the latest version which is published on npm.

tylerlevine:
> seems reasonable (though I'm unclear on why we want to exclude the test dir)

The reason is because it's uncommon that an end user of bitgojs (not a developer) would want to run the tests, and thus we should exclude them to lighten the package size.

Developers who want to run the tests are cloning the repo directly, so they will get the tests.

PR #295, opened 2019-04-03
Replace config constants with unspents lib

paddyquinn:
JIRA: BG-6452

Note that the config constants were out of sync with the unspents
lib size estimations. Thus, due to the size changes listed below,
tx hexes in tests changed because fee estimations (and
consequently output amounts) changed.

Input Type               | Old Size | New Size
--------------------|---------|----------------
p2sh input               | 296      | 297
p2shP2wsh input          | 139      | 140
uncompressed p2pkh input | 160      | 180

There are a couple of other things to note in regards to fees.
Output size estimation defaults to a p2pkh output, which is NOT the
worst case output size. Non-segwit tx overhead size is also used,
which is also NOT the worst case overhead size. For transactions
that use segwit, and specifically p2wsh outputs, we will be
underestimating fees.

The final thing to note is that there are several places where
chain codes are compared to chains 10 and 11 to determine if the
output is segwit. `isP2shP2wsh()` from the unspents lib is
intentionally not used because it throws if passed an invalid chain
code, which would change functionality and is apparently something
that may happen for v1 safe (non-hd) wallets.


src/v2/recovery.ts
```diff
@@ -332,13 +333,11 @@ class CrossChainRecoveryTool {
...
 
     // Determine fee with default fee rate
-    const P2SH_INPUT_SIZE = config.tx.P2SH_INPUT_SIZE;
-    const OUTPUT_SIZE = config.tx.OUTPUT_SIZE;
-    const TX_OVERHEAD_SIZE = config.tx.TX_OVERHEAD_SIZE;
     const feeRate = this.feeRates[this.sourceCoin.type];
 
     // Note that we assume one output here (all funds should be recovered to a single address)
-    const txSize = P2SH_INPUT_SIZE * recoveryTx.tx.ins.length + OUTPUT_SIZE + TX_OVERHEAD_SIZE;
+    const txSize = VirtualSizes.txP2shInputSize * recoveryTx.tx.ins.length + VirtualSizes.txP2pkhOutputSize +
```
OttoAllmendinger: please check if this works, it should be more accurate: 

`Dimensions.fromInputs(tx.ins).plus(Dimensions.SingleOutput.p2pkh).getVSize()`

I also think we could be more accurate by using the actual output instead of guessing p2pkh

```
outputDims = Dimensions.fromOutputWithScriptLength(
  utxolib.address.toOutputScript(address, network).length
);
```

src/v2/recovery.ts
```diff
@@ -495,10 +494,8 @@ class CrossChainRecoveryTool {
 
       const feeInfo: any = {};
 
-      const INPUT_SIZE = config.tx.P2SH_INPUT_SIZE;
-      const OUTPUT_SIZE = config.tx.OUTPUT_SIZE;
-      const OVERHEAD_SIZE = config.tx.TX_OVERHEAD_SIZE;
-      feeInfo.size = OVERHEAD_SIZE + (INPUT_SIZE * this.txInfo.inputs.length) + OUTPUT_SIZE;
+      feeInfo.size = VirtualSizes.txOverheadSize + (VirtualSizes.txP2shInputSize * this.txInfo.inputs.length) +
```
OttoAllmendinger: see prev comment

paddyquinn: https://bitgoinc.atlassian.net/browse/BG-11805

src/v2/recovery.ts
paddyquinn: https://bitgoinc.atlassian.net/browse/BG-11805

paddyquinn:
> We should really figure out a better way to assert tx correctness than asserting the tx hex

Couldn't agree more. It's hard to make changes and even harder for a reviewer to grok the changes. Created the following ticket for this issue: https://bitgoinc.atlassian.net/browse/BG-11806

PR #301, opened 2019-04-09
Enable p2wsh in explainTransaction()

paddyquinn:
JIRA: BG-11570, BG-8702

Previously `explainTransaction()` skipped counting signatures for
p2wsh inputs. This must have been an artifact of previous inability
to verify p2wsh sigs because simply removing the check enabled
p2wsh signature counting. No changes to signature validation were
necessary. This commit also refactors `explainTransaction()` and
enables the p2wsh tests for it.

### DIFF
--- a/src/v2/coins/abstractUtxoCoin.ts
+++ b/src/v2/coins/abstractUtxoCoin.ts
@@ -1060,11 +1060,11 @@ class AbstractUtxoCoin extends BaseCoin {
 
       if (!hasSigScript && !hasWitnessScript) {
         // no sig script or witness data for this input
-        debug('no script sig or witness script data for input %s', idx);
+        debug('no signature script or witness script data for input %s', idx);
         return 0;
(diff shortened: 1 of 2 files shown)


src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1081,48 +1081,28 @@ class AbstractUtxoCoin extends BaseCoin {
...
+        }
 
-          // p2sh
-          try {
-            return this.verifySignature(transaction, idx, undefined, verificationParams);
-          } catch (e) {
-            return false;
-          }
-        });
+        const verificationParams = {
```
OttoAllmendinger: nit: unnecessary var

paddyquinn: I like that better. Fixed and amended the commit. Please re-review :)

PR #306, opened 2019-04-11
BG-11931: Add slow-deps CI pipeline to measure install size and timing

tylerlevine:
We should really have a small script instead which does this measurement, but this will get us the metrics we need to move forward on reducing the installed package size.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -80,6 +80,22 @@ local IntegrationTest(version, limit_branches=true) = {
   ],
 };
 
+local MeasureSizeAndTiming(version, limit_branches=false) = {
+  kind: "pipeline",
+  name: "size and timing (node:" + version + ")",
+  steps: [
+    {
+      name: "slow-deps",
+      image: "node:" + version,
+      commands: [
+        "npm install -g slow-deps",
+        "slow-deps"
+      ],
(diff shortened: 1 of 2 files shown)


jaydenseric:
Next.js has pretty comprehensive checks using Github actions:

https://github.com/zeit/next.js/pull/6929#issuecomment-480618515

I wonder if there is a way to see what all the actions are, or how they work.

jaydenseric:
Here's how their actions work:

- https://twitter.com/timneutkens/status/1103759973711003649
- https://github.com/zeit/next-stats-action

tylerlevine:
> Next.js has pretty comprehensive checks using Github actions:
> 
> [zeit/next.js#6929 (comment)](https://github.com/zeit/next.js/pull/6929#issuecomment-480618515)
> 
> I wonder if there is a way to see what all the actions are, or how they work.

That's pretty sweet, thanks for the pointer. I have never heard of github actions before, but I signed up for the waiting list. It seems like the next.js size action is pretty customized for next.js though, so I'm not sure exactly how we can make use of it without forking and customizing.

I'd like to do something better than slow-deps, but I'm more interested in moving forward on fixing the package than on getting super nice size stats.

PR #307, opened 2019-04-12
BG-11787 use updateSingleKeychainPassword in keychain.ts

argjv:
[BG-11787](https://bitgoinc.atlassian.net/browse/BG-11787) Use the right method to update the keychain and update the unit tests.

### DIFF
--- a/test/v2/unit/keychains.ts
+++ b/test/v2/unit/keychains.ts
@@ -8,6 +8,7 @@ const co = Promise.coroutine;
 const nock = require('nock');
 const common = require('../../../src/common');
 import * as _ from 'lodash';
+const sinon = require('sinon');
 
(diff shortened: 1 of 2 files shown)


test/v2/unit/keychains.ts
```diff
@@ -60,6 +60,26 @@ describe('V2 Keychains', function v2keychains() {
       .should.be.rejectedWith('Expecting parameter string: newPassword but found number');
     }));
 
+    it('should fail if at least one of the the private keys is incorrect', co(function *coItFail() {
```
argjv: Is this assumption correct?

argjv:
I see. So it's not possible to detect an `incorrect password` error in `updatePassword()` because those happen as part of the happy path for unrelated wallets.

PR #308, opened 2019-04-12
BG-11671 update maximumSpendable params

argjv:
[BG-11671](https://bitgoinc.atlassian.net/browse/BG-11671) Add new param `recipientAddress` to `maximumSpendable`.

### DIFF
--- a/test/v2/unit/wallet.ts
+++ b/test/v2/unit/wallet.ts
@@ -698,4 +698,43 @@ describe('V2 Wallet:', function() {
     }));
   });
 
+  describe('Maximum Spendable', function maximumSpendable() {
+    let bgUrl;
+
+    before(co(function *() {
+      nock.activeMocks().should.be.empty();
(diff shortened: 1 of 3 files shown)


test/v2/unit/wallet.ts
```diff
@@ -698,4 +698,43 @@ describe('V2 Wallet:', function() {
...
+  describe('Maximum Spendable', function maximumSpendable() {
+    let bgUrl;
+
+    before(co(function *() {
+      nock.activeMocks().should.be.empty();
+      bgUrl = common.Environments[bitgo.getEnv()].uri;
+    }));
+
+    it('arguments', co(function *() {
+      const optionalParams = {
```
tylerlevine: Why do we need to pass all these other options?

I think you can just have `recipientAddress` here.

argjv: We don't have tests for all the other params. If we are gonna test for one, why not the rest?

PR #310, opened 2019-04-15
[BG-6875] Remove extra validation in createAddress

marktoda:
The `count` field was validated twice, remove the less specific one


dannybitgo:
this could really use a test

marktoda:
There already are tests -- test/v2/unit/wallet.ts:479

It checks validation of the count field fairly thoroughly

PR #314, opened 2019-04-16
BG-12002: Wrap parseSignatureScript in a try block

paddyquinn:
JIRA: BG-12002, BG-11994

Validation of number of signatures and pubkeys was added to
parseSignatureScript in #286. However, this validation fails for
our taint inputs, which are not multisig scripts. This commit
wraps `parseSignatureScript` within `explainTransaction` in a try
block to gracefully handle this failed validation.


james8nguyen:
@paddyquinn and @tylerlevine, the package-lock.json was not updated along with package.json in this PR.

PR #315, opened 2019-04-16
Wrap parseSignatureScript in a try block

paddyquinn:
JIRA: BG-12002, BG-11994

Validation of number of signatures and pubkeys was added to
parseSignatureScript in #286. However, this validation fails for
our taint inputs, which are not multisig scripts. This commit
wraps `parseSignatureScript` within `explainTransaction` in a try
block to gracefully handle this failed validation.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,6 +22,7 @@ steps:
   image: node:lts
   commands:
   - npm audit
+  failure: ignore
 
 ---
 kind: pipeline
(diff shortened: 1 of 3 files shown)


paddyquinn:
`handlebars` is the package that is failing `npm audit`. Per @tylerlevine:

```
it's coming in through nyc which does code coverage - it is not a blocker
it's a dev dependency which won't be present in any builds
```

PR #325, opened 2019-04-18
BG-2851: Remove v1 ethereum

tylerlevine:
The server side routes behind these functions were removed some time ago
and thus the functions have been broken for a while.

All eth functionality is now in v2, and is accessible using the 'eth' coin.


tylerlevine:
CC @OttoAllmendinger as reporter of BG-2851

PR #333, opened 2019-04-23
CT-128: Separate signing and verification

paddyquinn:
JIRA: CT-128

Prior to this commit we could not spend native segwit inputs unless
they were the last input in the transaction. This bug was caused by
the fact that we would rebuild the transaction (and subsequently
clear the redeem script in the p2wsh case) after signing each
input. Then we would verify the input and it would pass for p2wsh
because the redeem script had been cleared. However, when we
rebuilt on the next input, the redeem script would be there again.
We don't reverify the previous input so it slips past verification
and we end up with an invalid tx hex. This commit splits up signing
and signature verification such that this cannot happen.

Side note: I am not sure why bitgo-utxo-lib returns a redeem script
for native segwit inputs. This seems like a bug but the SDK was
already handling this (albeit incorrectly) so maybe it was
expected?

### DIFF
--- a/src/v2/coins/abstractUtxoCoin.ts
+++ b/src/v2/coins/abstractUtxoCoin.ts
@@ -685,71 +685,91 @@ class AbstractUtxoCoin extends BaseCoin {
     const txb = bitcoin.TransactionBuilder.fromTransaction(transaction, this.network);
     this.prepareTransactionBuilder(txb);
 
-    const signatureIssues = [];
-    for (let index = 0; index < transaction.ins.length; ++index) {
-      debug('Signing input %d of %d', index + 1, transaction.ins.length);
+    const getSignatureContext = (txPrebuild, index) => {
       const currentUnspent = txPrebuild.txInfo.unspents[index];
-      if (this.isBitGoTaintedUnspent(currentUnspent)) {
-        debug('Skipping input %d of %d (unspent from replay protection address which is platform signed only)', index + 1, transaction.ins.length);
-        continue;
-      }
-      const path = 'm/0/0/' + currentUnspent.chain + '/' + currentUnspent.index;
-      const privKey = hdPath.deriveKey(path);
-      privKey.network = this.network;
-
-      const currentSignatureIssue: any = {
+      return {
         inputIndex: index,
         unspent: currentUnspent,
-        path: path
+        path: 'm/0/0/' + currentUnspent.chain + '/' + currentUnspent.index,
+        isP2wsh: !currentUnspent.redeemScript,
+        isBitGoTaintedUnspent: this.isBitGoTaintedUnspent(currentUnspent),
+        error: undefined
       };
-      debug('Input details: %O', currentSignatureIssue);
+    };
 
+    const signatureIssues = [];
+    // Sign inputs
+    for (let index = 0; index < transaction.ins.length; ++index) {
+      debug('Signing input %d of %d', index + 1, transaction.ins.length);
+      const signatureContext = getSignatureContext(txPrebuild, index);
+      if (signatureContext.isBitGoTaintedUnspent) {
+        debug(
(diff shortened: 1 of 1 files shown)


paddyquinn:
@tylerlevine This should be released as soon as possible.. Should I update the version in this PR as well so we can release a new patch?

tylerlevine:
> @tylerlevine This should be released as soon as possible.. Should I update the version in this PR as well so we can release a new patch?

No, let's do that separately

paddyquinn:
Good deal.

src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -685,71 +685,94 @@ class AbstractUtxoCoin extends BaseCoin {
...
         transaction.setInputScript(index, Buffer.alloc(0));
       }
 
-      const isValidSignature = this.verifySignature(transaction, index, currentUnspent.value);
+      const isValidSignature = this.verifySignature(transaction, index, unspentInfo.unspent.value);
       if (!isValidSignature) {
         debug('Invalid signature');
-        currentSignatureIssue.error = new Error('invalid signature');
-        signatureIssues.push(currentSignatureIssue);
+        unspentInfo.error = new Error('invalid signature');
```
dannybitgo: this error will overwrite the previous error

dannybitgo: nvm

paddyquinn:
Hey @OttoAllmendinger. @tylerlevine rightly wanted to see some tests for this. Since this bug was found by your integration tests can you please run your integration tests against this branch and confirm that this resolves the issue? Also, do we have an ETA on those tests getting into master?

src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -685,71 +685,94 @@ class AbstractUtxoCoin extends BaseCoin {
     const txb = bitcoin.TransactionBuilder.fromTransaction(transaction, this.network);
     this.prepareTransactionBuilder(txb);
 
+    const getUnspentInfo = function(self, txPrebuild, index) {
```
OttoAllmendinger: use arrow syntax (`const getUnspentInfo = (txPrebuild, index) => `) so that you don't have to pass `this` as `self`

In cases where you can't use arrow syntax  (e.g. `co`-wrapped generators)  and have to use `co(function *(){}.bind(this));` instead

OttoAllmendinger: this function here has a less than elegant interface

```
const unspentInfo = getUnspentInfo(...);
unspentInfo.unspentInfo // moar info
unspentInfo.unspentInfo.unspent // this also contains unspent info!
```

It looks like you added this to have some code reuse and also have values ready for `signatureIssues`

I suggest we call this `getVerificationContext()` instead and move `isP2wsh` and `isBitGoTaintedUnspent` next to `inputIndex` and `unspent` to flatten the thing a bit

OttoAllmendinger: since you also use it during signing maybe `getSignatureContext()`

OttoAllmendinger:
> Hey @OttoAllmendinger. @tylerlevine rightly wanted to see some tests for this. Since this bug was found by your integration tests can you please run your integration tests against this branch and confirm that this resolves the issue? 

Merging your change has resolved the issue

https://github.com/BitGo/BitGoJS/commits/BG-10337.utxo-wallet-integration-tests-managedwallets / http://ci-jenkins-01.dev.bitgo:8080/view/E2E/job/test-integ-utxo-wallet/1455/


> Also, do we have an ETA on those tests getting into master?

I think it would take about half a sprint once we decide to do it

src/v2/coins/abstractUtxoCoin.ts
paddyquinn: Done.

PR #335, opened 2019-04-24
BG-11902 / Build and sign a trade payload

dacruz21:
Started the new trading module with a couple of methods to build and sign a DAX payload authorizing a trade. 

buildTradePayload calls the microservices stack to request a DAX payload
signTradePayload first builds the payload then signs it with the user key on the OFC wallet

Added an integration test to validate everything, but it's pretty loose on the validation as the DAX schema is subject to change somewhat often

### DIFF
--- a/src/v2/trading.ts
+++ b/src/v2/trading.ts
@@ -1,0 +1,81 @@
+import * as Bluebird from 'bluebird';
+const co = Bluebird.coroutine;
+
+export class Trading {
+  private bitgo;
+  private basecoin;
+  private wallet;
+
+  constructor(bitgo, basecoin, wallet) {
+    this.bitgo = bitgo;
+    this.basecoin = basecoin;
+    this.wallet = wallet;
+  }
+
+  /**
+   * Builds a payload authorizing a trade from this trading account. The currency and amount must be specified, as well as a list
+   * of trade counterparties.
+   * @param params
+   * @param params.currency the currency this wallet will be sending as part of the trade
+   * @param params.amount the amount of currency (in base units, such as cents, satoshis, or wei) authorized to be spent as part of the trade
+   * @param params.otherParties array of trading account IDs authorized to receive funds as part of the trade
+   * @param callback
+   * @returns unsigned trade payload for the given parameters
+   */
+  public buildTradePayload(params: BuildPayloadParameters, callback?): Bluebird<string> {
+    return co(function *buildTradePayload() {
+      const url = this.bitgo.microservicesUrl('/api/trade/v1/payload');
+
+      const body = {
+        accountId: this.wallet.id(),
+        currency: params.currency,
+        amount: params.amount,
+        otherParties: params.otherParties
+      };
+
+      const response = yield this.bitgo.post(url).send(body).result();
+
+      return response.payload;
+    }).call(this).asCallback(callback);
+  }
+
+  /**
+   * Builds and signs a payload authorizing a trade from this trading account. The currency and amount must be specified, as well
+   * as a list of trade counterparties. Requires the wallet keychain or raw private key to sign the transaction. Both
+   * the payload and signature are returned.
+   * @param params
+   * @param params.currency the currency this wallet will be sending as part of the trade
+   * @param params.amount the amount of currency (in base units, such as cents, satoshis, or wei) authorized to be spent as part of the trade
+   * @param params.otherParties array of trading account IDs authorized to receive funds as part of the trade
+   * @param params.walletPassphrase the wallet password, for decrypting the private key for signing
+   * @param callback
+   * @returns {{payload: string, signature: string}}
+   */
+  public signTradePayload(params: SignPayloadParameters, callback?): Bluebird<SignedPayload> {
+    return co(function *signTradePayload() {
+      const payload = yield this.buildTradePayload(params);
+
+      const key = yield this.basecoin.keychains().get({ id: this.wallet.keyIds()[0] });
+      const prv = this.bitgo.decrypt({ input: key.encryptedPrv, password: params.walletPassphrase });
+
+      const signature = this.basecoin.signMessage({ prv }, payload).toString('hex');
+
+      return { payload, signature };
+    }).call(this).asCallback(callback);
+  }
+}
+
+export interface BuildPayloadParameters {
+  currency: string;
+  amount: string;
+  otherParties: string[]
+}
+
+export interface SignPayloadParameters extends BuildPayloadParameters {
+  walletPassphrase?: string;
+}
+
+export interface SignedPayload {
+  payload: string;
+  signature: string;
+}
--- a/test/lib/test_bitgo.ts
+++ b/test/lib/test_bitgo.ts
@@ -49,6 +49,9 @@ BitGo.TEST_KNOWN_BALANCE = 9999586400;
 BitGo.TEST_ENTERPRISE_CREATION_USER = 'enterprisecreator@bitgo.com';
 BitGo.TEST_ENTERPRISE_CREATION_PASSWORD = BitGo.TEST_PASSWORD;
 
+BitGo.OFC_TEST_USER = 'tester+employee@bitgo.com';
+BitGo.OFC_TEST_PASSWORD = BitGo.TEST_PASSWORD;
+
 BitGo.TEST_CLIENTID = 'test';
 BitGo.TEST_CLIENTSECRET = 'testclientsecret';
 
@@ -93,6 +96,7 @@ BitGo.prototype.initializeTestVars = function() {
     BitGo.V2.TEST_WALLET1_PASSCODE = 'iVWeATjqLS1jJShrPpETti0b';
     BitGo.V2.TEST_WALLET1_XPUB = 'xpub661MyMwAqRbcGicVM5K5UnocWoFt3Yh1RZKzSEHPPARhyMf9w7DVqM3PgBgiVW5NHRp8UteqhMoQb17rCQsLbmGXuPx43MKskyB31R97p3G';
     BitGo.V2.TEST_WALLET1_ID = '585cc5335573b0a8416aadb1fce63ce3';
+    BitGo.V2.OFC_TEST_WALLET_ID = '5cbe3223311315fc7c96ce087f32dbdd';
 
   } else {
     BitGo.TEST_USERID = '543c11ed356d00cb7600000b98794503';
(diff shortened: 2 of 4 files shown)


src/v2/trading.ts
```diff
@@ -0,0 +1,81 @@
...
+   * Builds a payload authorizing a trade from this trading account. The currency and amount must be specified, as well as a list
+   * of trade counterparties.
+   * @param params
+   * @param params.currency the currency this wallet will be sending as part of the trade
+   * @param params.amount the amount of currency (in base units, such as cents, satoshis, or wei) authorized to be spent as part of the trade
+   * @param params.otherParties array of trading account IDs authorized to receive funds as part of the trade
+   * @param callback
+   * @returns unsigned trade payload for the given parameters
+   */
+  buildTradePayload(params: BuildPayloadParameters, callback?): Promise<string> {
```
tylerlevine: The return type of co is actually bluebird\<thing\> (if you have `import * as bluebird from 'bluebird';` at the top). There are a few extra functions on bluebird promises which get lost if you just return a vanilla promise

src/v2/trading.ts
```diff
@@ -0,0 +1,81 @@
...
+   * @param params.otherParties array of trading account IDs authorized to receive funds as part of the trade
+   * @param params.walletPassphrase the wallet password, for decrypting the private key for signing
+   * @param callback
+   * @returns {{payload: string, signature: string}}
+   */
+  signTradePayload(params: SignPayloadParameters, callback?): Promise<SignedPayload> {
+    return co(function *signTradePayload() {
+      const payload = yield this.buildTradePayload(params);
+
+      const key = yield this.basecoin.keychains().get({ id: this.wallet._wallet.keys[0] });
```
tylerlevine: Let's not access private properties of other objects unless we really need to. `this.wallet.keyIds[0]` should have the same effect.

test/v2/integration/trading.ts
```diff
@@ -0,0 +1,45 @@
+import { coroutine as co } from 'bluebird';
+import should = require('should');
```
tylerlevine: `import * as should from 'should';`

`import .. = require()` should only be used for compatibility with default or callable exports.

test/v2/integration/trading.ts
```diff
@@ -0,0 +1,45 @@
...
+    });
+
+    should.exist(payload);
+    // The payload should be a valid JSON object
+    // NOTE: we shouldn't do any more validation than this, as the schema is subject to change often
+    (() => { JSON.parse(payload); }).should.not.throw();
+
+    should.exist(signature);
+    // signature should be a hex string
+    signature.should.match(/[0-9a-f]+/);
```
tylerlevine: needs ^ and $, otherwise this will match any string that contains a single `[0-9a-f]`

test/v2/integration/trading.ts
```diff
@@ -0,0 +1,45 @@
...
+  let bitgo;
+  let basecoin;
+  let wallets;
+  let wallet;
+  let trading;
+
+  before(co(function *() {
+    nock.restore();
+
+    bitgo = new TestV2BitGo({ env: 'dev', microservicesUri: 'https://app.bitgo-dev.com' });
```
tylerlevine: This is now disclosing the microservices uri, are we comfortable with that? (that ship has now sailed however, since this PR is public).

src/v2/trading.ts
```diff
@@ -0,0 +1,81 @@
...
+   * the payload and signature are returned.
+   * @param params
+   * @param params.currency the currency this wallet will be sending as part of the trade
+   * @param params.amount the amount of currency (in base units, such as cents, satoshis, or wei) authorized to be spent as part of the trade
+   * @param params.otherParties array of trading account IDs authorized to receive funds as part of the trade
+   * @param params.walletPassphrase the wallet password, for decrypting the private key for signing
+   * @param callback
+   * @returns {{payload: string, signature: string}}
+   */
+  signTradePayload(params: SignPayloadParameters, callback?): Promise<SignedPayload> {
```
tylerlevine: Same return type and accessibility issues here.

src/v2/trading.ts
tylerlevine: We should be using the `public`/`protected`/`private` accessibility modifiers on new instance methods.

test/v2/integration/trading.ts
```diff
@@ -0,0 +1,45 @@
...
+    basecoin = bitgo.coin('ofc');
+    wallets = basecoin.wallets();
+    basecoin.keychains();
+
+    yield bitgo.authenticateOfcTestUser(bitgo.testUserOTP());
+    wallet = yield wallets.getWallet({ id: TestV2BitGo.V2.OFC_TEST_WALLET_ID });
+    trading = wallet.trading();
+  }));
+
+  it('should create and sign a trade payload', co(function *() {
```
tylerlevine: Seems like this test is simple enough that it could be turned into a unit test with the server nocked.

src/v2/trading.ts
dacruz21: Had to use the legacy `import Bluebird = require('bluebird')` to get this to work, but fixed, thx

test/v2/integration/trading.ts
dacruz21: good catch, fixed

test/lib/test_bitgo.ts
```diff
@@ -93,6 +96,7 @@ BitGo.prototype.initializeTestVars = function() {
     BitGo.V2.TEST_WALLET1_PASSCODE = 'iVWeATjqLS1jJShrPpETti0b';
     BitGo.V2.TEST_WALLET1_XPUB = 'xpub661MyMwAqRbcGicVM5K5UnocWoFt3Yh1RZKzSEHPPARhyMf9w7DVqM3PgBgiVW5NHRp8UteqhMoQb17rCQsLbmGXuPx43MKskyB31R97p3G';
     BitGo.V2.TEST_WALLET1_ID = '585cc5335573b0a8416aadb1fce63ce3';
+    BitGo.V2.OFC_TEST_WALLET_ID = '5cbe3223311315fc7c96ce087f32dbdd';
```
dacruz21: Yeah I'm not actually using these since I converted the test to a unit test, but I'll leave them here for anyone who wants to write integration tests later.

test/v2/unit/trading.ts
```diff
@@ -0,0 +1,74 @@
...
+        amount: '100000000',
+        otherParties: ['test_counter_party_1'],
+        nonce: Date.now()
+      })
+    });
+    const platformScope = nock(bgUrl)
+    .get('/api/v2/ofc/key/keyid')
+    .reply(200, {
+      pub: 'xpub1',
+      encryptedPrv: bitgo.encrypt({ input: 'xprv9s21ZrQH143K2MUz7uPUBVzdmvJQE6fPEQCkR3mypPbZgijPqfmGH7pjijdjeJx3oCoxPWVbjC4VYHzgN6wqEfYnnbNjK7jm2CkrvWrvkbR', password: TestV2BitGo.OFC_TEST_PASSWORD })
```
dacruz21: randomly generated xprv, just needed something to sign with

src/v2/trading.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      const signature = this.basecoin.signMessage({ prv }, payload).toString('hex');
+
+      return { payload, signature };
+    }).call(this).asCallback(callback);
+  }
+}
+
+
+interface BuildPayloadParameters {
```
tylerlevine: I think these interfaces are part of the public API and thus should be exported.

Do you anticipate users of bitgojs would need to know about these types in order to use the feature?

test/lib/test_bitgo.ts
tylerlevine: Sure, that's fine. Hopefully we get some more ofc tests soon.

test/v2/unit/trading.ts
```diff
@@ -0,0 +1,74 @@
+import { coroutine as co } from 'bluebird';
+import * as should from 'should';
+import * as nock from 'nock';
+
+const Wallet = require('../../../src/v2/wallet');
+const TestV2BitGo = require('../../lib/test_bitgo');
+const common = require('../../../src/common');
+
+describe('V2 Trading', function() {
+  const microservicesUri = 'https://bitgo-microservices-endpoint.com';
```
tylerlevine: Please use an invalid tld so we don't inadvertently make requests here. `bitgo-microservices.example` might be a good choice.

test/v2/unit/trading.ts
```diff
@@ -0,0 +1,74 @@
...
+  let bitgo;
+  let basecoin;
+  let wallet;
+  let trading;
+  let bgUrl;
+
+  before(co(function *() {
+    bitgo = new TestV2BitGo({ env: 'dev', microservicesUri });
```
tylerlevine: I think you can use `env: 'mock'`

test/v2/unit/trading.ts
```diff
@@ -0,0 +1,74 @@
...
+      otherParties: ['test_counterparty_1'],
+      walletPassphrase: TestV2BitGo.OFC_TEST_PASSWORD
+    });
+
+    should.exist(payload);
+    // The payload should be a valid JSON object
+    // NOTE: we shouldn't do any more validation than this, as the schema is subject to change often
+    (() => { JSON.parse(payload); }).should.not.throw();
+
+    should.exist(signature);
```
tylerlevine: Should there be a check that the signature is valid over the payload? (up to you)

src/v2/trading.ts
dacruz21: Decent IDEs will still offer type checking even if these interfaces can't be directly imported, but this use case wouldn't work:

```ts
import { BuildPayloadParameters } from 'bitgo';
const params: BuildPayloadParameters {
  ...
}
```

i.e. manually creating an object with that type. We can export that interface if you want, but then we'll be exporting a ton of interfaces which may get confusing. Thoughts?

test/v2/unit/trading.ts
dacruz21: We don't have a method in the SDK to verify arbitrary signatures, only transactions. If we can implement it without bringing in new dependencies we could consider it.

tylerlevine: We have `bitcoinjs-message`, which has a `verify()` function.

See https://github.com/BitGo/BitGoJS/blob/6dead153286316484678de2e777afeec9fa5d5fb/src/v2/coins/abstractUtxoCoin.ts#L408

src/v2/trading.ts
tylerlevine: It won't work because we don't re-export any interfaces out of the top level, so you're correct in saying that they can't be directly imported in the current project setup. However, I still think we should export if only so that future developers are aware that these types are part of the public API.

dacruz21: Ah okay, sounds good

PR #341, opened 2019-05-01
CT-115: Use carat and tilde policy for dependency versions

tylerlevine:
It should not be up to the library to lock their dependency tree, and
instead this should be left to the end application. Pinning single
dependency versions is not helpful in controlling the entire dependency
graph, and still allows in-range updates for packages at a depth
greater than 1.

The correct way to solve this problem is at the end application level,
where the entire dependency tree can be evaluated and locked as a single
unit. For this reason, we are relaxing the version restrictions in place
for BitGoJS.

We strongly recommend that all users of bitgojs who are producing
applications make use of a package-lock.json or yarn.lock file at
install time. Additionally, when upgrading the lock file, we recommend
performing in-depth dependency testing, and upgrading to a version which
has had sufficient time to be vetted by the open source community.


Melvillian:
What happened with the code coverage here why did it go down so much (> 5%)?

tylerlevine:
> What happened with the code coverage here why did it go down so much (> 5%)?

Not entirely clear to me, but my guess is `nyc` changed their coverage analysis and is now not counting some coverage which was counted previously. I will do some more research into this before landing.

tylerlevine:
Closing for now, as I'd like to make this change after bitgo-express is separated into it's own module.

PR #348, opened 2019-05-06
BG-7960: v2 wallet share succeeds without encryptedPrv

LeoKotschenreuther:
We are currently throwing an error when a share is initiated without the `skipKeychains` flag for a cold wallet (no encryptedPrv on any of the wallet keys). This is a problem because there are some wallets that are not marked as cold, i.e. the customer created them via the API by creating their own keys (e.g. having their user key in a customer HSM). This change modifies the wallet share function to ignore this error and to continue with the share.

### DIFF
--- a/src/v2/wallet.ts
+++ b/src/v2/wallet.ts
@@ -979,6 +979,11 @@ Wallet.prototype.shareWallet = function(params, callback) {
             path: sharing.path
           };
         }
+      }).catch(function(e) {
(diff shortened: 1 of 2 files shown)


LeoKotschenreuther:
@tylerlevine yeah, I agree with you. However there are other usages of the `getEncryptedUserKeychain` function where you actually still want the error message.

PR #349, opened 2019-05-06
CT-210: Revert "Revert "BG-11787 use updateSingleKeychainPassword instead of …

tylerlevine:
…changeSingleKeychainPassword and fix unit tests""

This reverts commit 11fddb392a8f41b6c10529117c8eea285009067d.


dannybitgo:
i'm confused, wasn't this already done here: https://github.com/BitGo/BitGoJS/pull/307/files ?
Or was that original one reverted at some point?

tylerlevine:
> i'm confused, wasn't this already done here: https://github.com/BitGo/BitGoJS/pull/307/files ?
> Or was that original one reverted at some point?

The original change was reverted because the fix exposed an incompatibility on the platform side. That has now been resolved, so we are able to re-introduce the SDK fix.

PR #353, opened 2019-05-08
CT-200: Modularization

tylerlevine:


.drone.jsonnet
```diff
@@ -1,42 +1,87 @@
...
-local BuildInfo(version, limit_branches=false) = {
+local BuildInfo(version) = {
   name: "build information",
   image: "node:" + version,
   commands: [
     "node --version",
     "npm --version",
+    "yarn --version",
+    "git --version",
+    "env",
```
tylerlevine: remove just in case we somehow have secret envs defined globally for a pipeline instead of inside a particular pipeline step.

modules/core/RELEASE_NOTES.md
```diff
@@ -1,38 +1,5 @@
 # BitGoJS Release Notes
 
-## 5.2.0
```
tylerlevine: Oops, back this out. We should keep the updated release notes from upstream

modules/core/package.json
```diff
@@ -0,0 +1,134 @@
...
+    "test": "npm run unit-test",
+    "unit-test": "./node_modules/.bin/nyc -- ./node_modules/.bin/_mocha -r ts-node/register --timeout 20000 --reporter list --exit \"test/unit/*.ts\" \"test/unit/**/*.ts\" \"test/v2/unit/*.ts\" \"test/v2/unit/**/*.ts\"",
+    "integration-test": "./node_modules/.bin/nyc -- ./node_modules/.bin/_mocha -r ts-node/register --timeout 20000 --reporter list --exit \"test/v2/integration/**/*.ts\"",
+    "lint": "eslint 'src/**/*.ts' && eslint 'test/**/*.ts' || true",
+    "audit": "if [ \"$(npm --version | cut -d. -f1)\" -ge \"6\" ]; then npm audit; else echo \"npm >= 6 required to perform audit. skipping...\"; fi",
+    "clean": "rm -r dist/*",
+    "build": "./node_modules/.bin/tsc && node scripts/copySjcl.js",
+    "prepublish": "npm run build"
+  },
+  "dependencies": {
```
tylerlevine: Some of these deps can be removed now, since only bitgo-express used them

modules/core/test/integration/bitgoExpress.ts
```diff
@@ -3,7 +3,7 @@ require('should-http');
 const request = require('supertest-as-promised');
 const _ = require('lodash');
 
-const expressApp = require('../../src/expressApp');
+const expressApp = require('../../../express/src/expressApp');
```
tylerlevine: Oops, move to express module

modules/core/test/v2/integration/coins/eth.ts
```diff
@@ -6,7 +6,7 @@ const co = Promise.coroutine;
 import request = require('supertest-as-promised');
 const BN = require('ethereumjs-util').BN;
 
-const expressApp = require('../../../../src/expressApp');
+const expressApp = require('../../../../../express/src/expressApp');
```
tylerlevine: why does this need an express app? This should probably move to express if this is testing express.

modules/express/README.md
```diff
@@ -0,0 +1,106 @@
+# BitGoJS
```
tylerlevine: Rewrite README, don't copy from bitgo

modules/express/package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "@bitgo/express",
+  "version": "6.0.0",
+  "description": "Local signing server and proxy for the BitGo platform",
+  "main": "./dist/src/index.js",
+  "types": "./dist/src/index.d.ts",
+  "bin": {
```
tylerlevine: Test to make sure global binary install is usable on all platforms

modules/express/src/clientRoutes.ts
```diff
@@ -1,16 +1,16 @@
 import bodyParser = require('body-parser');
 
-const BitGoJS = require('./index');
-const TransactionBuilder = require('./transactionBuilder');
-const common = require('./common');
+const BitGoJS = require('../../core/src');
```
tylerlevine: require('bitgo')

modules/express/src/clientRoutes.ts
```diff
@@ -1,16 +1,16 @@
...
+const TransactionBuilder = require('../../core/src/transactionBuilder');
+const common = require('../../core/src/common');
```
tylerlevine: Gross and icky, but fixing may require refactoring these parts out into a common module. I'm open to suggestions here.

modules/express/test/lib/asserts.ts
```diff
@@ -0,0 +1,11 @@
+// Additional convenience assertions for should
```
tylerlevine: Should be in a common lib, or test-common maybe.

dannybitgo:
I'm assuming this isn't ready for review yet because it says "WIP"? Let me know if thats incorrect.

tylerlevine:
> I'm assuming this isn't ready for review yet because it says "WIP"? Let me know if thats incorrect.

It is still somewhat in progress, but there won't be any more huge changes. I think there is enough to start reviewing. Will remove the WIP.

modules/express/test/integration/bitgoExpressV1.ts
```diff
@@ -352,12 +349,12 @@ describe('Bitgo Express', function() {
...
         res.should.have.status(200);
-        res.body.size.should.eql(886);
-        res.body.fee.should.eql(17720);
+        res.body.size.should.eql(920);
+        res.body.fee.should.eql(18400);
       });
     });
 
-    it('create and accept a pending approval', function() {
+    xit('create and accept a pending approval', function() {
```
dannybitgo: why are we canceling all these tests?

tylerlevine: They come from the v1 integration test suite, which is not run during the normal CI process, since many are broken. These tests are of the broken variety, and thus are disabled off the bat until we can fix.

dannybitgo:
is there a way I can test this on my machine? any suggestions?

dacruz21:
I think you need to move the `CODEOWNERS` file back to the root of the repo and update the paths:
https://help.github.com/en/articles/about-code-owners
> To use a CODEOWNERS file, create a new file called CODEOWNERS in the root, docs/, or .github/ directory of the repository, in the branch where you'd like to add the code owners.

tylerlevine:
Thanks, will do that now @dacruz21

PR #359, opened 2019-05-22
[CT-27] Add Hop transactions building and sending

marktoda:
# Overview
Build a hop transaction with flag `hop`
Send a hop transaction with flag `hop`
Verify hop transaction and signature from the HSM
Send a commitment signature to the HSM, so it can verify that the tx was actually intended from the user

### DIFF
--- a/src/v2/baseCoin.ts
+++ b/src/v2/baseCoin.ts
@@ -354,11 +354,29 @@ class BaseCoin {
   }
 
   /**
+   * Get extra parameters for prebuilding a tx. Add things like hop transaction params
+   */
+  getExtraPrebuildParams(buildParams, callback) {
+    return Promise.method(function() {
+      return { };
+    }).call(this).asCallback(callback);
+  }
+
+  /**
    * Modify prebuild after receiving it from the server. Add things like nlocktime
    */
   postProcessPrebuild(prebuildResponse, callback) {
     return Promise.method(function() {
       return prebuildResponse;
+    }).call(this).asCallback(callback);
+  }
+
+  /**
+   * Coin-specific things done before signing a transaction, i.e. verification
+   */
+  presignTransaction(params, callback) {
+    return Promise.method(function() {
+      return params;
     }).call(this).asCallback(callback);
   }
 
--- a/src/v2/coins/eth.ts
+++ b/src/v2/coins/eth.ts
@@ -3,12 +3,13 @@ const Wallet = require('../wallet');
 import common = require('../../common');
 const config = require('../../config');
 const BigNumber = require('bignumber.js');
+const keccak = require('keccak');
 const Util = require('../../util');
 import * as _ from 'lodash';
 import * as Promise from 'bluebird';
 const request = require('superagent');
 const crypto = require('crypto');
-const prova = require('prova-lib');
+const secp256k1 = require('secp256k1');
 const utxoLib = require('bitgo-utxo-lib');
 const co = Promise.coroutine;
 
@@ -51,6 +52,8 @@ class Eth extends BaseCoin {
     return 'Ethereum';
   }
 
+  static hopTransactionSalt = 'bitgoHopAddressRequestSalt';
+
   /**
    * Flag for sending value of 0
    * @returns {boolean} True if okay to send 0 value, false otherwise
@@ -83,7 +86,7 @@ class Eth extends BaseCoin {
    */
   isValidPub(pub) {
     try {
-      prova.HDNode.fromBase58(pub);
+      utxoLib.HDNode.fromBase58(pub);
       return true;
     } catch (e) {
       return false;
@@ -321,7 +324,7 @@ class Eth extends BaseCoin {
       throw new Error('missing prv parameter to sign transaction');
     }
 
-    params.recipients = params.recipients || txPrebuild.recipients;
+    params.recipients = txPrebuild.recipients || params.recipients;
 
     // if no recipients in either params or txPrebuild, then throw an error
     if (!params.recipients || !Array.isArray(params.recipients)) {
@@ -348,7 +351,8 @@ class Eth extends BaseCoin {
       operationHash: operationHash,
       signature: signature,
       gasLimit: params.gasLimit,
-      gasPrice: params.gasPrice
+      gasPrice: params.gasPrice,
+      hopTransaction: txPrebuild.hopTransaction,
     };
     return { halfSigned: txParams };
   }
@@ -506,7 +510,7 @@ class Eth extends BaseCoin {
       let backupSigningKey;
 
       if (isKrsRecovery || isUnsignedSweep) {
-        const backupHDNode = prova.HDNode.fromBase58(backupKey);
+        const backupHDNode = utxoLib.HDNode.fromBase58(backupKey);
         backupSigningKey = backupHDNode.getKey().getPublicKeyBuffer();
         backupKeyAddress = `0x${optionalDeps.ethUtil.publicToAddress(backupSigningKey, true).toString('hex')}`;
       } else {
@@ -522,7 +526,7 @@ class Eth extends BaseCoin {
           throw new Error(`Error decrypting backup keychain: ${e.message}`);
         }
 
-        const backupHDNode = prova.HDNode.fromBase58(backupPrv);
+        const backupHDNode = utxoLib.HDNode.fromBase58(backupPrv);
         backupSigningKey = backupHDNode.getKey().getPrivateKeyBuffer();
         backupKeyAddress = `0x${optionalDeps.ethUtil.privateToAddress(backupSigningKey).toString('hex')}`;
       }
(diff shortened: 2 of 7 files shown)


src/v2/coins/eth.ts
```diff
@@ -321,7 +329,7 @@ class Eth extends BaseCoin {
       throw new Error('missing prv parameter to sign transaction');
     }
 
-    params.recipients = params.recipients || txPrebuild.recipients;
+    params.recipients = txPrebuild.recipients || params.recipients;
```
marktoda: Needed to switch the preference order here because the recipients from the prebuild are more accurate than the originally requested recipients in this case

src/v2/wallet.ts
```diff
@@ -1091,18 +1097,33 @@ Wallet.prototype.prebuildTransaction = function(params, callback) {
...
+Wallet.prototype.getUserPrv = function(params) {
+  const userKeychain = params.keychain || params.key;
   let userPrv = params.prv;
   if (userPrv && typeof userPrv !== 'string') {
     throw new Error('prv must be a string');
   }
   if (userPrv && params.coldDerivationSeed) {
     // the derivation only makes sense when a key already exists
-    const derivation = this.baseCoin.deriveKeyWithSeed({ key: userPrv, seed: params.coldDerivationSeed });
+  const derivation = this.baseCoin.deriveKeyWithSeed({ key: userPrv, seed: params.coldDerivationSeed });
```
dpc: ?

src/v2/wallet.ts
```diff
@@ -1358,7 +1373,7 @@ Wallet.prototype.sendMany = function(params, callback) {
       'message', 'minValue', 'maxValue', 'sequenceId',
       'lastLedgerSequence', 'ledgerSequenceDelta', 'gasPrice',
       'noSplitChange', 'unspents', 'comment', 'otp', 'changeAddress',
-      'instant', 'memo'
+      'instant', 'memo',
```
dpc: Is this legal? Most importantly - is this moral?

test/v2/unit/ethWallet.ts
```diff
@@ -46,6 +51,234 @@ describe('Sign ETH Transaction', co(function *() {
...
+      try {
+        yield ethWallet.baseCoin.validateHopPrebuild(ethWallet, prebuild, badBuildParams);
+      } catch (e) {
+        error = e.message;
+      }
+      should.exist(error);
+      error.should.containEql("does not equal original amount");
+    }));
+
+    it('should fail if the signature is invalid', co(function *() {
```
dpc: Unclear. "should fail if the HSM signature is invalid"?

test/v2/unit/ethWallet.ts
```diff
@@ -46,6 +51,234 @@ describe('Sign ETH Transaction', co(function *() {
...
+      try {
+        yield ethWallet.baseCoin.validateHopPrebuild(ethWallet, prebuild, badBuildParams);
+      } catch (e) {
+        error = e.message;
+      }
+      should.exist(error);
+      error.should.containEql("does not equal original recipient");
+    }));
+
+    it('should fail if the amount is wrong', co(function *() {
```
dpc: unclear. amount where, when?

test/v2/unit/ethWallet.ts
```diff
@@ -46,6 +51,234 @@ describe('Sign ETH Transaction', co(function *() {
...
+      nockBitGoKey();
+      try {
+        yield ethWallet.baseCoin.validateHopPrebuild(ethWallet, prebuild, buildParams);
+      } catch (e) {
+        error = e.message;
+      }
+      should.not.exist(error);
+    }));
+
+    it('should fail if the recipient is wrong', co(function *() {
```
dpc: where, when?

test/v2/unit/ethWallet.ts
```diff
@@ -46,6 +51,234 @@ describe('Sign ETH Transaction', co(function *() {
...
+      try {
+        yield ethWallet.baseCoin.validateHopPrebuild(ethWallet, prebuild, buildParams);
+      } catch (e) {
+        error = e.message;
+      }
+      should.exist(error);
+      error.should.containEql("Hop txid signature invalid");
+    }));
+
+    it('should fail if the signature signed the wrong digest', co(function *() {
```
dpc: who's signature, what digest? :)

dpc:
@tylerlevine - just to be clear about the weight of it: it is vitaly important that the user can't be fooled into signing a TX before validating corresponding HopTX was returned and signed by HSM, not tampered with, and all data on it is just like the user requested.

src/v2/wallet.ts
marktoda: my b

src/v2/wallet.ts
marktoda: legal & normal. I'm not sure the lint rules for BitGoJS, so I'll remove just to avoid the extra diff line

marktoda:
have to fix tests

src/v2/baseCoin.ts
```diff
@@ -386,6 +395,15 @@ class BaseCoin {
       if (params && params.numBlocks) {
         query.numBlocks = params.numBlocks;
       }
+      if (params && params.hop) {
+        query.hop = params.hop;
+      }
+      if (params && params.recipient) {
```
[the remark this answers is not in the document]
marktoda: This param works for non-hop ETH transactions too -- you're right, it should be in a coinspecific function -- will change

src/v2/coins/eth.ts
```diff
@@ -51,6 +53,10 @@ class Eth extends BaseCoin {
     return 'Ethereum';
   }
 
+  static get hopTransactionSalt(): string {
```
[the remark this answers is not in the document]
marktoda: Your example is (unfortunately) not valid in JS: https://stackoverflow.com/questions/32647215/declaring-static-constants-in-es6-classes

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,156 @@ class Eth extends BaseCoin {
...
+      const walletPassphrase = buildParams.walletPassphrase;
+
+      const userKeychain = yield this.keychains().get({ id: wallet._wallet.keys[0] });
+      const userPrv = wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+      const userPrvBuffer = utxoLib.HDNode.fromBase58(userPrv).getKey().getPrivateKeyBuffer();
+      if (!recipients || !Array.isArray(recipients)) {
+        throw new Error('expecting array of recipients');
+      }
+
+      // Right now we only support 1 recipient
```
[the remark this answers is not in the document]
marktoda: Its an array to stay consistent with the API for all other coins. IT will eventually be allowed to have more than 1 when Alex's intern comes :D

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,156 @@ class Eth extends BaseCoin {
...
+        id,
+        signature,
+      } = hopPrebuild;
+
+      const {
+        recipients,
+      } = originalParams;
+
+      // first, validate the HSM signature
+      const bitgoKeychain: { pub: string } = yield this.keychains().get({ id: wallet.keyIds()[2] });
```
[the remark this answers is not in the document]
marktoda: yep, its always [user, backup, bitgo]

src/v2/baseCoin.ts
```diff
@@ -386,6 +395,15 @@ class BaseCoin {
       if (params && params.numBlocks) {
         query.numBlocks = params.numBlocks;
       }
+      if (params && params.hop) {
```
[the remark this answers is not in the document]
marktoda: Moved to coinspecific

src/v2/coins/eth.ts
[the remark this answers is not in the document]
cooncesean: i'm a js n00b. thx for the education.

src/v2/coins/eth.ts
```diff
@@ -3,11 +3,13 @@ const Wallet = require('../wallet');
 import common = require('../../common');
 const config = require('../../config');
 const BigNumber = require('bignumber.js');
+const keccak = require('keccak');
```
tylerlevine: Please add to package.json (this has been a transitive dependency until now).

src/v2/wallet.ts
```diff
@@ -1047,25 +1047,29 @@ Wallet.prototype.removeUser = function(params, callback) {
...
-      'cpfpTxIds', 'cpfpFeeRate', 'maxFee', 'idfVersion', 'idfSignedTimestamp', 'idfUserId', 'strategy'
+      'cpfpTxIds', 'cpfpFeeRate', 'maxFee', 'idfVersion', 'idfSignedTimestamp', 'idfUserId', 'strategy', 'hop'
     ]);
     debug('prebuilding transaction: %O', whitelistedParams);
 
     if (params.reqId) {
       this.bitgo._reqId = params.reqId;
     }
+    const extraParams = yield this.baseCoin.getExtraPrebuildParams(Object.assign(params, { wallet: this }));
+    const buildParams = Object.assign(whitelistedParams, extraParams);
```
tylerlevine: Nit: Don't need this extra var. Since Object.assign mutates it's target, you can continue using `whitelistedParams` instead of aliasing it as `buildParams`.

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,182 @@ class Eth extends BaseCoin {
...
+        signature,
+      } = hopPrebuild;
+
+      const {
+        recipients,
+      } = originalParams;
+
+      // first, validate the HSM signature
+      const bitgoKeychain: { pub: string } = yield this.keychains().get({ id: wallet.keyIds()[2] });
+      const bitgoPubkeyBuffer: Buffer = prova.HDNode.fromBase58(bitgoKeychain.pub).getPublicKeyBuffer();
```
tylerlevine: Use `bitgo-utxo-lib` instead, prova is deprecated

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,182 @@ class Eth extends BaseCoin {
...
+      const feeEstimate = yield this.feeEstimate(feeEstimateParams);
+      const gasLimit = feeEstimate.gasLimitEstimate;
+      // shift by 9 because result is in gwei and we want wei
+      const gasPrice = Math.round((feeEstimate.feeEstimate) / gasLimit);
+      const gasPriceMax = gasPrice * 5;
+      // Payment id as  current datetime so its always unique every time we build one
+      const paymentId = (new Date().toISOString());
+      const hopDigest: Buffer = Eth.getHopDigest([recipientAddress, recipientAmount, gasPriceMax, gasLimit, paymentId]);
+
+      const userReqSig = optionalDeps.ethUtil.addHexPrefix(secp256k1.sign(hopDigest, userPrvBuffer).signature.toString('hex'));
```
tylerlevine: Why are we adding/stripping hex prefixes all over? What's the benefit?

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,182 @@ class Eth extends BaseCoin {
...
+        recipients,
+      } = originalParams;
+
+      // first, validate the HSM signature
+      const bitgoKeychain: { pub: string } = yield this.keychains().get({ id: wallet.keyIds()[2] });
+      const bitgoPubkeyBuffer: Buffer = prova.HDNode.fromBase58(bitgoKeychain.pub).getPublicKeyBuffer();
+      const signatureBuffer: Buffer = Buffer.from(optionalDeps.ethUtil.stripHexPrefix(signature), 'hex');
+      const messageBuffer: Buffer = Buffer.from(optionalDeps.ethUtil.stripHexPrefix(id), 'hex');
+
+      const isValidSignature: boolean = secp256k1.verify(messageBuffer, signatureBuffer.slice(1), bitgoPubkeyBuffer);
```
tylerlevine: Why slice the first byte off?

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,182 @@ class Eth extends BaseCoin {
...
+      if (!isValidSignature) {
+        throw new Error(`Hop txid signature invalid`);
+      }
+
+      // Then validate that the tx params actually equal the requested params
+      const originalAmount: string = recipients[0].amount;
+      const originalDestination: string = recipients[0].address;
+
+      const builtHopTx = new optionalDeps.EthTx(tx);
+      const hopAmount: string = BigNumber(optionalDeps.ethUtil.bufferToHex(builtHopTx.value)).toFixed(0);
```
tylerlevine: Nit: I'd rather you convert `originalAmount` to a `BigNumber` and compare that to the `hopAmount` instead of converting `hopAmount` to a string. We should try to compare numbers instead of strings where possible.

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,182 @@ class Eth extends BaseCoin {
...
+      const hopDestination: string = optionalDeps.ethUtil.bufferToHex(builtHopTx.to);
```
tylerlevine: Can you move this down below the hopAmount if statement?

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,182 @@ class Eth extends BaseCoin {
...
+    }).call(this).asCallback(callback);
+  }
+
+  /**
+   * Gets the hop digest for the user to sign. This is validated in the HSM to prove that the user requested this tx
+   * @param paramsArr The parameters to hash together for the digest
+   */
+  static getHopDigest(paramsArr): Buffer {
+    const hash = new keccak('keccak256');
+    hash.update([Eth.hopTransactionSalt, ...paramsArr].join('$'));
```
tylerlevine: Hm, not sure how useful a public, static salt is here. What's the goal?

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,182 @@ class Eth extends BaseCoin {
...
+      const recipientAddress = recipients[0].address;
+      const recipientAmount = recipients[0].amount;
+      const feeEstimateParams = {
+        recipient: recipientAddress,
+        hop: true,
+      };
+      const feeEstimate = yield this.feeEstimate(feeEstimateParams);
+      const gasLimit = feeEstimate.gasLimitEstimate;
+      // shift by 9 because result is in gwei and we want wei
+      const gasPrice = Math.round((feeEstimate.feeEstimate) / gasLimit);
```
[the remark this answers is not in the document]
marktoda: oh oops, comment -- i thought we wanted wei but we dont

src/v2/coins/eth.ts
```diff
@@ -805,6 +812,182 @@ class Eth extends BaseCoin {
...
+      const gasPriceMax = gasPrice * 5;
+      // Payment id as  current datetime so its always unique every time we build one
+      const paymentId = (new Date().toISOString());
```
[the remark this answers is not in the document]
marktoda: The impact is that we may reuse a hop address which just complicates things. Down for a random number though

src/v2/wallet.ts
marktoda: oof didnt know it mutated

src/v2/coins/eth.ts
marktoda: whenever its serialized as hex string, all the Eth util stuff (and the HSM) expect the 0x prefix. But, Buffer.from() doesnt work with the hex prefix LOL. So my philosophy with ETH stuff is to always have hex prefix when string and only remove it when shoving it into a buffer

src/v2/coins/eth.ts
marktoda: its the DER signature header byte, which secp256k1.verify doesn't know about. I couldnt get bitcoinmessage (which does know about the header byte) to work, so this is how we got interoperability with the signature schemes. HSM does the same thing with my signatures

src/v2/coins/eth.ts
marktoda: done

src/v2/coins/eth.ts
marktoda: @dpc's idea, its mostly to make sure that we dont ever accidentally (though super unlikely) signature collision by accident. Having the salt be super specifically for hop transactions, one has to deliberately be requesting this for the HSM to ever sign it. Signing [to, amount, gasprice, gaslimit, paymentId] isn't entirely implausible under a different circumstance

src/v2/coins/eth.ts
dpc: This is done, right? Double checking.

src/v2/wallet.ts
```diff
@@ -1091,11 +1097,27 @@ Wallet.prototype.prebuildTransaction = function(params, callback) {
  * @return {*}
  */
 Wallet.prototype.signTransaction = function(params, callback) {
-  const userKeychain = params.keychain || params.key;
   const txPrebuild = params.txPrebuild;
   if (!txPrebuild || typeof txPrebuild !== 'object') {
     throw new Error('txPrebuild must be an object');
   }
+  const userPrv = this.getUserPrv(params);
```
tylerlevine: I'd rather decrypt the user prv only after presignTransaction has been called (eg. on line 1109), so we already know the tx is good to sign. We should keep the userPrv around in the clear for only as long as it is absolutely required.

marktoda: Makes sense, done

src/v2/wallet.ts
```diff
@@ -1091,11 +1097,27 @@ Wallet.prototype.prebuildTransaction = function(params, callback) {
...
-  const userKeychain = params.keychain || params.key;
   const txPrebuild = params.txPrebuild;
   if (!txPrebuild || typeof txPrebuild !== 'object') {
     throw new Error('txPrebuild must be an object');
   }
+  const self = this;
+  return Promise.try(function() {
+    return self.baseCoin.presignTransaction(params);
+  }).then(function(params) {
+    const userPrv = this.getUserPrv(params);
```
tylerlevine: Need to use `self` here instead of `this` (that should fix your test failures).

PR #361, opened 2019-05-28
[CT-27] Make payment ID a string rather than a number

marktoda:
# Overview

Payment id is meant to be a string, when changing it from the current ISODate to a random number, we broke this and the platform rejects it as a valid payment id


tylerlevine:
I'm a bit confused why `tsc` didn't catch this, since it looks like the type is declared correctly in `HopParams`.

PR #365, opened 2019-06-05
BG-12941 / Create TradingAccount wrapper class and refactor payloads

dacruz21:
Made all breaking changes in this PR so that we can roll out `bitgo@6.0.0` before other trading features are complete. Updated Mocha tests

**Breaking Changes**
- removed `Wallet.trading()`
- `Wallet.trading().buildTradePayload()` has been moved to `Wallet.toTradingAccount().buildPayload()`
- `Wallet.trading().signTradePayload()` has been moved to `Wallet.toTradingAccount().buildAndSignPayload()`
- Note that `Wallet.toTradingAccount().signPayload()` expects a pre-built trade payload, unlike the removed `Wallet.trading().signTradePayload()`, which built the payload when called. To build and sign a trade payload in a single call, please use `Wallet.toTradingAccount().buildAndSignPayload()`


src/v2/trading/tradingAccount.ts
```diff
@@ -0,0 +1,77 @@
...
+  currency: string;
+  amount: string;
+  nonceHold: string;
+  nonceSettle: string;
+  otherParties: string[];
+}
+
+interface BuildPayloadParameters {
+  currency: string;
+  amount: bigint | string;
```
tylerlevine: let's stick with number, since we are not targeting esnext.

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html (scroll down to BigInt)

May even be worth it to not take a number at all, and only take a string for the amount.

dacruz21: Huh I was wrong. I thought only BigInt *literals* were not supported, and `tsc` only gave me a warning when I tried to use a literal in a unit test, but I'll be damned. Can't imagine why the TS team wouldn't transpile it down to ES5 or just remove the type.

ack

PR #367, opened 2019-06-05
CT-220 - Add ALGO key logic to create a wallet

CapnMigraine:
Adds ALGO methods needed to create a new wallet.

https://bitgoinc.atlassian.net/browse/CT-220

### DIFF
--- a/modules/core/src/v2/coins/algo.ts
+++ b/modules/core/src/v2/coins/algo.ts
@@ -1,35 +1,115 @@
 import BaseCoin = require('../baseCoin');
+const algosdk = require('algosdk');
+
+interface KeyPair {
+  pub: string;
+  prv: string;
+}
 
 class Algo extends BaseCoin {
+  constructor() {
+    super();
+  }
 
-    constructor() {
-      super();
+  getChain(): string {
+    return 'algo';
+  }
+
+  getFamily(): string {
+    return 'algo';
+  }
+
+  getFullName(): string {
+    return 'Algorand';
+  }
+
+  getBaseFactor(): number {
+    return 1e6;
+  }
+
+  /**
+   * Flag for sending value of 0
+   * @returns {boolean} True if okay to send 0 value, false otherwise
+   */
+  valuelessTransferAllowed(): boolean {
+    // TODO: this sounds like its true with the staking txes - confirm before launch
+    return false;
+  }
+
+  /**
+   * Generate ed25519 key pair
+   *
+   * @param seed
+   * @returns {Object} object with generated pub, prv
+   */
+  generateKeyPair(seed?: Buffer): KeyPair {
+    const pair = seed ? algosdk.generateAccountFromSeed(seed) : algosdk.generateAccount();
+    return {
+      pub: pair.addr, // encoded pub
+      prv: algosdk.Seed.encode(pair.sk), // encoded seed
+    };
+  }
+
+  /**
+   * Return boolean indicating whether input is valid public key for the coin.
+   *
+   * @param {String} pub the pub to be checked
+   * @returns {Boolean} is it valid?
+   */
+  isValidPub(pub): boolean {
+    return algosdk.isValidAddress(pub);
+  }
+
+  /**
+   * Return boolean indicating whether input is valid seed for the coin
+   * In Algorand, when the private key is encoded as base32 string only the first 32 bytes are taken,
+   * so the encoded value is actually the seed
+   *
+   * @param {String} prv the prv to be checked
+   * @returns {Boolean} is it valid?
+   */
+  isValidPrv(prv): boolean {
+    return algosdk.isValidSeed(prv);
+  }
+
+  /**
+   * Return boolean indicating whether input is valid public key for the coin
+   *
+   * @param {String} address the pub to be checked
+   * @returns {Boolean} is it valid?
+   */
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -44,7 +44,10 @@
     "@types/mocha": "5.2.6",
     "@types/nock": "9.3.1",
     "@types/node": "11.11.4",
-    "algosdk": "1.2.0",
+    "@babel/cli": "^7.4.4",
```
tylerlevine: Why are we bringing in babel as a dependency?

CapnMigraine: This dependency will likely be removed. Currently it's used by our `algosdk-bitgo` library. We transpile the official `algosdk` so it can be used by the platform.

We'll figure this out before this PR is ready to be merged. I added the DO NOT MERGE label for now.

CapnMigraine: Ok, it's gone now

modules/core/src/v2/coins/algo.ts
```diff
@@ -1,35 +1,115 @@
...
-    getBaseFactor() {
-        return 1e6;
-    }
+  getBaseFactor(): number {
+    return 1e6;
+  }
 
-    /**
+  /**
      * Flag for sending value of 0
```
tylerlevine: Extra whitespace

modules/core/src/v2/coins/algo.ts
```diff
@@ -1,35 +1,115 @@
...
+   * Generate ed25519 key pair
+   *
+   * @param seed
+   * @returns {Object} object with generated pub, prv and secret
+   */
+  generateKeyPair(seed?: Buffer): KeyPair {
+    const pair = seed ? algosdk.generateAccountFromSeed(seed) : algosdk.generateAccount();
+    return {
+      pub: pair.addr, // encoded pub
+      prv: algosdk.Seed.encode(pair.sk), // encoded seed
```
tylerlevine: Why are we returning the encoded seed as part of the keypair? This breaks the basecoin contract by returning a different object than other coins.

Seems like you can go back and forth from the seed to the pub/prv, so why do  we need to keep the seed after we generate the keypair?

modules/core/test/v2/unit/coins/algo.ts
```diff
@@ -0,0 +1,68 @@
...
+import * as algosdk from 'algosdk';
+
+const TestV2BitGo = require('../../../lib/test_bitgo');
+
+describe('ALGO:', function() {
+  let bitgo;
+  let basecoin;
+
+  before(function() {
+    bitgo = new TestV2BitGo({ env: 'test' });
```
tylerlevine: `env: 'mock'` in unit tests, so it's very obvious when a nock is missing

modules/core/src/v2/coins/algo.ts
```diff
@@ -1,35 +1,115 @@
...
+  signMessage(key, message): Buffer {
+    let seed = key.prv;
+    if (!this.isValidPrv(seed)) {
+      throw new Error(`invalid seed: ${seed}`);
     }
+    if (typeof seed === 'string') {
+      try {
+        seed = algosdk.Seed.decode(seed).seed;
+      } catch (e) {
+        throw new Error(`invalid seed: ${seed}`);
```
tylerlevine: Please throw with different message, so we can disambiguate from the first case when we see this in a log from a customer.

modules/core/test/v2/unit/coins/algo.ts
CapnMigraine: done

modules/core/src/v2/coins/algo.ts
tylerlevine: Is there any code which needs to use `secret`? The tx signing code recovers it from `prv` instead of getting it from the keypair: https://github.com/BitGo/BitGoJS/pull/368/files#diff-32749934abcd4fda4b31a0f6496db33aR105

CapnMigraine: yes, `signMessage`: https://github.com/BitGo/BitGoJS/pull/367/files#diff-32749934abcd4fda4b31a0f6496db33aR111

CapnMigraine: my bad, we can regenerate the key with the seed using `algosdk.generateAccountFromSeed` and use `key.pk` for signing. Let me fix that.

modules/core/src/v2/coins/algo.ts
CapnMigraine: done

modules/core/src/v2/coins/algo.ts
CapnMigraine: done

modules/core/src/v2/coins/algo.ts
```diff
@@ -1,35 +1,113 @@
...
+  }
+
+  /**
+   * Sign message with private key
+   *
+   * @param key
+   * @param message
+   */
+  signMessage(key, message): Buffer {
+    let seed = key.prv;
```
tylerlevine: Ok, let's rename this, because I think this is causing the confusion. Seed != prv.

You get a keypair passed to this fn with { pub, prv }. Then you decode the prv to get a seed, then use that seed to re-create the same keypair which was passed into this function, then use the secret key from that new keypair to sign. Seems simpler to just decode key.prv and sign?

PR #369, opened 2019-06-05
CT-334: Move CODEOWNERS back to project root

tylerlevine:


tylerlevine:
> LGTM - but should I be able to approve if I'm not a CODEOWNER?

Technically no, but since we have no CODEOWNERS file at the moment, you are able to. This PR fixes that.

PR #373, opened 2019-06-07
BG-12925 / Manage trading network

dacruz21:
Listing trading partners and balance checking.

### DIFF
--- a/modules/core/src/v2/trading/tradingPartner.ts
+++ b/modules/core/src/v2/trading/tradingPartner.ts
@@ -1,0 +1,43 @@
+/**
+ * @prettier
+ */
+import * as Bluebird from 'bluebird';
+
+import { TradingAccount } from './tradingAccount';
+
+export class TradingPartner {
+  private bitgo: any;
+  private currentAccount: TradingAccount; // account of the user using the SDK, needed to construct balance check URL
+
+  public name: string;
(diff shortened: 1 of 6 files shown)


modules/core/src/v2/trading/tradingPartner.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /**
+   * Check if a trading partner has enough funds to cover the cost of a trade.
+   * @param currency the currency to check
+   * @param amount the amount of currency to check, represented in base units (such as cents, satoshi, or wei)
+   * @param callback
+   */
+  checkBalance(currency: string, amount: string, callback?): Bluebird<boolean> {
+    const url = this.bitgo.microservicesUrl(`/api/trade/v1/account/${this.partneredAccount.id}/tradingPartners/${this.accountId}/balance`);
```
quazeenii: I think this is switched? it should be accountId first, then partnerAccountId

dacruz21: not switched, just bad variable name

dacruz21:
@cooncesean Tyler and I discussed this before my previous PR. The real advantage of modules is reduced dependency size. These functions don't add any additional dependencies so there's no benefit to fragmenting our SDK by introducing another module.

PR #382, opened 2019-06-17
CT-427: Implement `deriveKeyWithSeed` for XLM

tylerlevine:
Note: Only works for deriving new hardened child private keys from a
root private key.


modules/core/src/v2/baseCoin.ts
```diff
@@ -243,9 +243,9 @@ export abstract class BaseCoin {
    * The cold wallet tool uses this function to derive an extended key that is based on the passed key and seed
    * @param key
    * @param seed
-   * @returns {{key: *, derivationPath: string}}
+   * @returns {{key: string, derivationPath: string}}
    */
-  deriveKeyWithSeed({ key, seed }) {
+  deriveKeyWithSeed({ key, seed }: { key: string, seed: string }): { key: string, derivationPath: string } {
```
cooncesean: I'm trying to grok this (excuse my typescript illiteracy) ... am i correct in assuming that we are just typing the input + return values here? ... cleaning up as we go?

modules/core/src/v2/coins/xlm.ts
```diff
@@ -1,15 +1,17 @@
...
-const BigNumber = require('bignumber.js');
+import { BigNumber } from 'bignumber.js';
 import * as querystring from 'querystring';
 import * as url from 'url';
 import * as Bluebird from 'bluebird';
+import * as bitcoin from 'bitgo-utxo-lib';
 const co = Bluebird.coroutine;
-const request = require('superagent');
+import * as request from 'superagent';
+import * as stellar from 'stellar-sdk';
```
cooncesean: Is `import *` fairly common TS pattern? I've always found it more readable + maintainable to be explicit with imports.

modules/core/src/v2/coins/xlm.ts
```diff
@@ -220,13 +223,20 @@ export class Xlm extends BaseCoin {
...
+    if (Array.isArray(queryDetails.memoId) && queryDetails.memoId.length > 1) {
+      // valid addresses can only contain one memo id
+      throw new Error(`invalid address '${address}', can only contain one memoId`);
+    }
+
+    const memoId = queryDetails.memoId as string;
+
     try {
-      new BigNumber(queryDetails.memoId);
+      new BigNumber(memoId);
```
cooncesean: Interesting, so we cast `queryDetails.memoId` to a string, and then pump it into BitNumber ... two thoughts:

1. Why cast it to a string in the first place? Seems like an int would be appropriate.
1. My understanding is that there are other memo types (MemoHash, MemoText, and one other ... i _think_). Are those supported or ignored?

@CapnMigraine can you shed any insight into the second point?

modules/core/src/v2/coins/xlm.ts
```diff
@@ -220,13 +223,20 @@ export class Xlm extends BaseCoin {
...
     } catch (e) {
       throw new Error(`invalid address: ${address}`);
     }
 
-    if (!this.isValidMemoId(queryDetails.memoId)) {
+    if (!this.isValidMemoId(memoId)) {
```
cooncesean: Another thought would be to move all logic from lines 231 up until here into `isValidMemoId` ... would keep this parent fcn cleaner and consolidate memo validation to a single, testable method.

Just a thought - no need to make sweeping changes though.

modules/core/src/v2/coins/xlm.ts
```diff
@@ -759,4 +769,125 @@ export class Xlm extends BaseCoin {
...
+   * Due to technical differences between keypairs on the ed25519 curve and the secp256k1 curve,
+   * only hardened private key derivation is supported.
+   *
+   * @param key seed for the master key. Note: Not the public key or encoded private key. This is the raw seed.
+   * @param entropySeed random seed which is hashed to generate the derivation path
+   */
+  deriveKeyWithSeed({ key, seed }: { key: string; seed: string }): { derivationPath: string; key: any } {
+    const derivationPathInput = bitcoin.crypto.hash256(`${seed}`).toString('hex');
+    const derivationPathParts = [
+      999999,
```
cooncesean: What does this magic number represent?

modules/core/src/v2/coins/xlm.ts
```diff
@@ -759,4 +769,125 @@ export class Xlm extends BaseCoin {
...
+   *
+   * @param key parent key
+   * @param chainCode chain code for parent key
+   * @param index index of child to derive
+   */
+  private static CKDPriv({ key, chainCode }: XlmKeypair, index: number): XlmKeypair {
+    const indexBuffer = Buffer.allocUnsafe(4);
+    indexBuffer.writeUInt32BE(index, 0);
+    const data = Buffer.concat([Buffer.alloc(1, 0), key, indexBuffer]);
+    const I = createHmac('sha512', chainCode)
```
cooncesean: Are `I`, `IL`, and `IR` naming conventions specific to this lib or XLM or this curve that I'm not familiar with? If so, and you feel other contributors will pick this up, pls proceed. Otherwise, we may want to create more meaningful const names for them.

modules/core/src/v2/baseCoin.ts
tylerlevine: Yes, exactly. This defines the types of the inputs, which were just `any` before.

modules/core/src/v2/coins/xlm.ts
tylerlevine: It's the closest analogue to the old pattern of `const lib = require('some-lib')`. I could pull out just the members I need, but that would mean more code changes which are not related to the fix. I'm trying to improve things a little bit at a time instead of all at once - it's easy to get sucked into a rabbit hole if you try to fix too much.

modules/core/src/v2/coins/xlm.ts
tylerlevine: We need to use it as a string later (isValidMemo expects a string), but we don't know if it's valid yet, so we try to create a big number from it. If that fails, we know the address is not valid. We need to cast to string because before that the type of `queryDetails.memoId` is `string | string[]` since a particular key can appear multiple times in a query string. This case was not handled explicitly before this change.

modules/core/src/v2/coins/xlm.ts
tylerlevine: Yeah, you're right, the try/catch is redundant here. Good catch.

modules/core/src/v2/coins/xlm.ts
tylerlevine: No clue, I copied it from the baseCoin implementation.

https://github.com/BitGo/BitGoJS/blob/75977994b406ace9c69634a6e43a6e5d06dac3c7/modules/core/src/v2/baseCoin.ts#L248

modules/core/src/v2/coins/xlm.ts
tylerlevine: They are variable names from the algorithm used for key derivation in BIP-32 and SLIP-0010.

modules/core/src/v2/coins/xlm.ts
CapnMigraine: We don't use numbers for memo ids because they take values up to 2^64 - 1, while JS numbers could lose precision, since they are capped at 2^53 - 1

modules/core/src/v2/coins/xlm.ts
argjv: I believe this number is fixed so we group all our derived keys under the same root tree: https://github.com/bitcoin/bips/blob/master/bip-0032/derivation.png

But I'm not sure of the benefits.

modules/core/src/v2/keyDeriver.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Calculate a child private key given the parent key, the chain code, and the child index.
+   *
+   * @param key parent key
+   * @param chainCode chain code for parent key
+   * @param index index of child to derive
+   */
+  private static CKDPriv({ key, chainCode }: HdKeypair, index: number): HdKeypair {
```
argjv: Just double checking that I understand this right from the paper
> For ed25519, the last step always succeeds since every 256-bit number (even 0) is a valid private key.

Refers to step 5 and means we are never going to get an invalid private key for ed25519 after `CKDPriv()`

tylerlevine: correct

tylerlevine:
> I would add all the test vectors though, just in case they represent some edge cases.

There are only two test vectors for ed25519 and I'm testing both of them. The other lines under each test vector are intermediate keys which must be calculated correctly to arrive at the correct final key. I  think they are included to help debugging.

PR #384, opened 2019-06-18
[CT-442] Check ETH Hop signature from the serverXpub, not the BitGo key

marktoda:
The corresponding change will be made server-side


modules/core/src/common.ts
```diff
@@ -12,6 +12,7 @@ export const Environments = {
     rmgNetwork: 'rmg',
     signingAddress: '1BitGo3gxRZ6mQSEH52dvCKSUgVCAH4Rja',
     serverXpub: 'xpub661MyMwAqRbcEtUgu9HF8ai4ipuVKKHBzUqks4jSFypW8dwwQL1zygLgQx99NmC7zJJznSiwKG6RQfVjAKMtCsx8VjR6kQW8x7HrkXFZdnQ',
+    hsmXpub: 'xpub661MyMwAqRbcGEtTFgMAoxMFoqsRdDaiaR63byNfZdV7cBZFvovQSNJ5bpyeoQtuKVgXBk6sFQ7TtvyWqadt41DnKwveYgM5KoU2EKYjdY2',
```
dpc: Hmmm... I thought this was supposed to be one of the existing keys.

marktoda: This is an existing key, just was not in SDK before

marktoda:
Ah thanks -- fixed. We were nocking bitgokey before which is now unnecessary since we use the hard coded one

dpc:
@tylerlevine On the servier isde I guess we could always keep calling with the old key until we're sure all the previous hoptxes are long pushed (sprint or two), and then retire the old key. 

On the client side though? I don't know. What's the plan rotating other hardcoded keys here?

tylerlevine:
> On the client side though? I don't know. What's the plan rotating other hardcoded keys here?

right now, that is the only option. If we have valid key signatures for the wallet, maybe we have more options. I'll have to think about that a bit more

PR #389, opened 2019-06-20
CT-364: Convert Wallet and Wallets to ES6 class and improve type support

tylerlevine:
* Refactor example keycard code into separate file
* Introduce and begin using standard callback type
* Fix inconsistent error behavior in some methods

### DIFF
--- a/modules/core/src/v2/keycard.ts
+++ b/modules/core/src/v2/keycard.ts
@@ -1,0 +1,381 @@
+/**
+ * Return the list of questions that will appear on the second page of the keycard
+ * @param coin name of the coin
+ */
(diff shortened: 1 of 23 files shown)


modules/core/src/v2/keycard.ts
```diff
@@ -0,0 +1,381 @@
+/**
```
dannybitgo: Is this an entirely new file, or was it copied from somewhere else?

tylerlevine: It's a new file, but the code was extracted out of wallet.ts

dannybitgo:
reviewed in person with tyler, looks good!

PR #390, opened 2019-06-21
CT-403: Add type information to xlm

tylerlevine:


modules/core/src/v2/coins/xlm.ts
```diff
@@ -747,6 +895,7 @@ export class Xlm extends BaseCoin {
           keychains = yield Bluebird.props({
             user: this.keychains().get({ id: wallet._wallet.keys[0] }),
             backup: this.keychains().get({ id: wallet._wallet.keys[1] }),
+            // TODO: why are we getting the bitgo key?
```
dannybitgo: i'm not sure either

tylerlevine: > i'm not sure either

I have a feeling we don't need it, but maybe I'm missing something? @CapnMigraine any ideas? Probably a good idea to have you review this change anyway.

modules/core/src/v2/coins/xlm.ts
```diff
@@ -1,23 +1,148 @@
...
+  prv: string;
+}
+
+interface AddressDetails {
+  address: string;
+  memoId?: string;
+}
+
+interface Memo {
+  type: stellar.MemoType;
```
CapnMigraine: what is this type? I think it should be `string`

modules/core/src/v2/coins/xlm.ts
CapnMigraine: No reason, we don't need it. Please remove it.

modules/core/src/v2/coins/xlm.ts
```diff
@@ -1,23 +1,148 @@
...
+  feeRate: number;
+  size: number;
+}
+
+interface TransactionExplanation {
+  displayOrder: string[];
+  id: string;
+  outputs: TransactionOutput[];
+  outputAmount: string;
+  changeOutputs: [];
```
CapnMigraine: this list doesn't have a type

modules/core/src/v2/coins/xlm.ts
```diff
@@ -259,10 +404,10 @@ export class Xlm extends BaseCoin {
   /**
    * Return boolean indicating whether input is valid public key for the coin
    *
-   * @param {String} address the pub to be checked
-   * @returns {Boolean} is it valid?
+   * @param address the pub to be checked
+   * @returns is it valid?
    */
-  isValidAddress(address) {
+  isValidAddress(address: string) {
```
CapnMigraine: returns `boolean`

modules/core/src/v2/coins/xlm.ts
```diff
@@ -277,19 +422,19 @@ export class Xlm extends BaseCoin {
    * Example of a common stellar username: foo@bar.baz
    * The above example would result in the Stellar address: foo@bar.baz*bitgo.com
    *
-   * @param {String} username - stellar username
-   * @return {boolean} true if stellar username is valid
+   * @param username - stellar username
+   * @return true if stellar username is valid
    */
-  isValidStellarUsername(username) {
+  isValidStellarUsername(username: string) {
```
CapnMigraine: this one too returns `boolean`

modules/core/src/v2/coins/xlm.ts
tylerlevine: It's a type exported by the stellar sdk.

Here's the definition:
```
enum MemoType
{
    MEMO_NONE = 0,
    MEMO_TEXT = 1,
    MEMO_ID = 2,
    MEMO_HASH = 3,
    MEMO_RETURN = 4
};
```

modules/core/src/v2/coins/xlm.ts
tylerlevine: There won't ever be any elements of this array, so the type is actually a correct empty array literal. Here's more info if you are curious, since I haven't seen this sort of feature in other languages: https://basarat.gitbooks.io/typescript/docs/types/literal-types.html

tylerlevine:
> Are we encouraging the use of trailing commas in the SDK? I ask because there are a few missing here, maybe it's a low-hanging fruit for this PR.

Yes, I prefer trailing commas where allowed in es5, but this isn't enforced currently unless prettier is configured for the file in question. I'll add them to this PR as I come across them

modules/core/src/v2/coins/xlm.ts
CapnMigraine: ok, thanks for clarifying

PR #392, opened 2019-06-23
CT-382: Send BitGoJS version in BitGo Express user agent string

tylerlevine:


modules/express/src/expressApp.ts
```diff
@@ -18,10 +18,11 @@ import { ServerResponse } from 'http';
 import { SSL_OP_NO_TLSv1 } from 'constants';
 
 const { Environments } = require('bitgo');
+const { version } = require('bitgo/package.json');
 const pjson = require('../package.json');
 import { TlsConfigurationError, NodeEnvironmentError } from './errors';
 
-const BITGOEXPRESS_USER_AGENT = 'BitGoExpress/' + pjson.version;
+const BITGOEXPRESS_USER_AGENT = `BitGoExpress/${pjson.version} BitGoJS/${version}`;
```
dannybitgo: do you mean to have a space after ```${pjson.version}``` ?

tylerlevine: Yes, this is intentional. It's handled correctly by the user agent parsing code in the backend.

PR #394, opened 2019-06-24
[BG-12973] Change recovery ZEC block explorer

bpagon13:
The previous explorer used for doing non-BitGo recovery for ZEC, zcash.blockexplorer.com, is no longer running, so a new explorer had to be found. The new explorer zcashnetwork.info is now used. This new explorer returns the same JSON information as the previous website, so no code changes were needed.

I explored having two explorers used so that there would be redundancy in case one explorer went out of service like zcash.blockexplorer.com did, however making that change for just ZEC would be inconsistent with how the current recovery code for the other coins works, so I decided not to.

### DIFF
--- a/modules/core/src/common.ts
+++ b/modules/core/src/common.ts
@@ -18,7 +18,7 @@ export const Environments = {
     btgExplorerBaseUrl: 'https://btgexplorer.com/api',
     etherscanBaseUrl: 'https://api.etherscan.io',
     ltcExplorerBaseUrl: 'https://insight.litecore.io/api',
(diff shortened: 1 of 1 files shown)


dannybitgo:
@bpagon13 can you edit your commit message so that it contains the jira ticket number?
```
git commit --amend -m "BG-12973 Change recovery ZEC block explorer"
```

tylerlevine:
> plz update commit message to include jira ticket

^ this. Rest looks good to me. Thanks for the fix!

PR #395, opened 2019-06-25
Add EOS address methods to SDK

CapnMigraine:
Add EOS class with methods for address handling required by the client


modules/core/src/v2/coins/eos.ts
```diff
@@ -0,0 +1,219 @@
...
+  isValidMemo = function(memo: string): boolean {
+    return memo && memo.length <= 256;
+  };
+
+  /**
+   * Return boolean indicating whether a memo id is valid
+   *
+   * @param memoId - the memo id to be checked
+   */
+  isValidMemoId = function(memoId: string): boolean {
```
tylerlevine: I don't think you need the `= function` part there

modules/core/src/v2/coins/eos.ts
```diff
@@ -0,0 +1,219 @@
...
+      return false;
+    }
+  }
+
+  /**
+   * Evaluates whether a memo is valid
+   *
+   * @param memo - the memo to be checked
+   */
+  isValidMemo = function(memo: string): boolean {
```
tylerlevine: I don't think you need the `= function` part there

modules/core/src/v2/coins/eos.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  getAddressDetails(address: string): AddressDetails {
+    const destinationDetails = url.parse(address);
+    const queryDetails = querystring.parse(destinationDetails.query);
+    const destinationAddress = destinationDetails.pathname;
+
+    // EOS addresses have to be "human readable", which means start with a letter, less than 12 characters and only a-z1-5., i.e.mtoda1.bitgo
+    // source: https://developers.eos.io/eosio-cpp/docs/naming-conventions
+    if (!/^[a-z][a-z1-5.]*$/.test(destinationAddress) || destinationAddress.length > EOS_ADDRESS_LENGTH) {
+      throw new Error(`invalid address: ${address}`);
```
tylerlevine: Should throw an `InvalidAddressError` since we already have that error type defined

modules/core/src/v2/coins/eos.ts
```diff
@@ -0,0 +1,219 @@
...
+    if (destinationDetails.pathname === address) {
+      return {
+        address: address,
+        memoId: null,
+      };
+    }
+
+    if (!queryDetails.memoId) {
+      // if there are more properties, the query details need to contain the memoId property
+      throw new Error(`invalid property in address: ${address}`);
```
tylerlevine: I think these Errors can be InvalidAddressErrors also

modules/core/src/v2/coins/eos.ts
```diff
@@ -0,0 +1,219 @@
...
+    }
+  }
+
+  /**
+   * Check if address is a valid EOS address, then verify it matches the root address.
+   *
+   * @param address - the address to verify
+   * @param rootAddress - the wallet's root address
+   */
+  verifyAddress({ address, rootAddress }: { address: string, rootAddress: string }) {
```
tylerlevine: returns void

modules/core/test/v2/unit/coins/eos.ts
```diff
@@ -0,0 +1,100 @@
...
+    it('should generate a keypair from random seed', function() {
+      const keyPair = basecoin.generateKeyPair();
+      keyPair.should.have.property('pub');
+      keyPair.should.have.property('prv');
+
+      basecoin.isValidPub(keyPair.pub).should.equal(true);
+      basecoin.isValidPrv(keyPair.prv).should.equal(true);
+    });
+
+    it('should generate a keypair from seed', function() {
```
tylerlevine: Isn't this testing the same thing as the test above, except we do the `crypto.randomBytes()` before we call the function? If the intention is to make sure we deterministically generate a keypair from a seed, then it might make more sense to use a static seed instead of a random one and make sure you get the key you're expecting.

CapnMigraine: yes, it makes more sense. Done.

CapnMigraine:
> Ah nice! Thanks for doing this. I think we copied a bit of work with CT-393-eos-sdk branch but all good, ill just merge that into this one as this is more fleshed out. Everything looks good to me, will let tyler take over

Yes, I used your branch as a base. I only left out the logic for signing, actually. 
Might be faster to create a new branch when this one is merged and add that code than to fix merge conflicts.

modules/core/src/v2/coins/eos.ts
```diff
@@ -0,0 +1,220 @@
...
+   * @param address - the address
+   */
+  getAddressDetails(address: string): AddressDetails {
+    const destinationDetails = url.parse(address);
+    const queryDetails = querystring.parse(destinationDetails.query);
+    const destinationAddress = destinationDetails.pathname;
+
+    // EOS addresses have to be "human readable", which means start with a letter, less than 12 characters and only a-z1-5., i.e.mtoda1.bitgo
+    // source: https://developers.eos.io/eosio-cpp/docs/naming-conventions
+    if (!/^[a-z][a-z1-5.]*$/.test(destinationAddress) || destinationAddress.length > EOS_ADDRESS_LENGTH) {
```
tylerlevine: if the spec is less than 12, then this condition is wrong and should be >=. What should the behavior be if the address is exactly 12 characters?

CapnMigraine: It's up to 12 characters. Fixed the comment

PR #400, opened 2019-06-28
CT-456 Ensure amount value passes through eth fee estimate

marktoda:
This is required to get an accurate fee estimate -- I added it in the caller but forgot to add it here


marktoda:
Thx tyler, will make that PR today.

PR #402, opened 2019-07-01
CT-558: Run browserify before webpack when producing browser bundle

tylerlevine:


RobertBaron:
I am not going to merge this yet, so I can test this in the legacy client, but so far works well.

PR #409, opened 2019-07-02
CT-256 - algorand ovc signing

twtaylorbitgo:


twtaylorbitgo:
This was based off Tyler's branch, I got a new one based on master now.

PR #411, opened 2019-07-03
[CT-393] Add EOS Signing logic

marktoda:
Add local signing using the eosjs ecc library

### DIFF
--- a/modules/core/src/v2/coins/eos.ts
+++ b/modules/core/src/v2/coins/eos.ts
@@ -8,6 +8,7 @@ import * as utxoLib from 'bitgo-utxo-lib';
 import * as url from 'url';
 import * as querystring from 'querystring';
 import * as _ from 'lodash';
+import * as ecc from 'eosjs-ecc'
 import { InvalidAddressError, UnexpectedAddressError } from '../../errors';
 
 const EOS_ADDRESS_LENGTH = 12;
@@ -20,6 +21,45 @@ interface KeyPair {
 interface AddressDetails {
   address: string;
   memoId: string;
+}
+
+export interface EosTx {
+  signatures: string[];
+  packed_trx: string;
+  compression: string;
+}
+
+export interface Recipient {
+  address: string;
+  amount: string;
+}
+
+interface EosTransactionHeaders {
+  ref_block_prefix: string;
+  ref_block_num: number;
+}
+
+interface EosTransactionPrebuild {
+  rawTx: string;
+  tx: EosTx;
+  headers: EosTransactionHeaders;
+}
+
+export interface EosSignTransactionParams {
+  prv: string;
+  txPrebuild: EosTransactionPrebuild;
+  recipients: Recipient[];
+}
+
+export interface EosHalfSigned {
+  transaction: EosTx;
+  txHex: string;
+  recipients: Recipient[];
+  headers: EosTransactionHeaders;
+}
+
+export interface EosSignedTransaction {
+  halfSigned: EosHalfSigned
 }
 
 export class Eos extends BaseCoin {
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/eos.ts
```diff
@@ -217,4 +218,31 @@ export class Eos extends BaseCoin {
...
+
+    const signBuffer = Buffer.from(txData, 'hex');
+    const privateKeyBuffer = utxoLib.HDNode.fromBase58(prv).getKey().getPrivateKeyBuffer();
+    const signature = ecc.Signature.sign(signBuffer, privateKeyBuffer).toString();
+
+    tx.signatures.push(signature);
+
+    const txParams = {
+      transaction: tx,
+      txHex: txData,
```
dannybitgo: shouldnt the txHex be 
```
tx.buildIncomplete().serialize()
```
or something? looks like we're just returning the unsigned txhex

marktoda: no, this is already serialized signing data. The thing that we send is `transaction`, which is just json lol

dannybitgo:
Being tested in simulator

modules/core/src/v2/coins/eos.ts
```diff
@@ -217,4 +218,31 @@ export class Eos extends BaseCoin {
...
   }
+
+  /**
+   * Assemble keychain and half-sign prebuilt transaction
+   *
+   * @param params
+   * @param params.txPrebuild {Object} prebuild object returned by platform
+   * @param params.prv {String} user prv
+   */
+  signTransaction(params) {
```
tylerlevine: Types?

marktoda: added :)

marktoda: I keep forgetting its typescript now

modules/core/src/v2/coins/eos.ts
```diff
@@ -22,6 +23,12 @@ interface AddressDetails {
   memoId: string;
 }
 
+interface EosTx {
```
tylerlevine: Since you're returning an object with a property of type `EosTx`, this should be exported since it's part of the public API.

modules/core/src/v2/coins/eos.ts
```diff
@@ -217,4 +224,31 @@ export class Eos extends BaseCoin {
...
   }
+
+  /**
+   * Assemble keychain and half-sign prebuilt transaction
+   *
+   * @param params
+   * @param params.txPrebuild {Object} prebuild object returned by platform
+   * @param params.prv {String} user prv
+   */
+  signTransaction(params) {
```
tylerlevine: Please add an interface for the param type and return type (both currently any). Most of the types you added are not needed and are inferred for you automatically.

modules/core/test/v2/unit/coins/eos.ts
```diff
@@ -1,9 +1,13 @@
...
 import * as crypto from 'crypto';
+import * as Promise from 'bluebird';
+const co = Promise.coroutine;
+import * as ecc from 'eosjs-ecc';
+const bitcoin = require('bitgo-utxo-lib');
 
 const TestV2BitGo = require('../../../lib/test_bitgo');
 
-describe('EOS:', function() {
+describe('EOS:', function () {
```
tylerlevine: Extra whitespace

modules/core/test/v2/unit/coins/eos.ts
```diff
@@ -1,9 +1,13 @@
 import * as should from 'should';
 import * as crypto from 'crypto';
+import * as Promise from 'bluebird';
+const co = Promise.coroutine;
+import * as ecc from 'eosjs-ecc';
+const bitcoin = require('bitgo-utxo-lib');
```
tylerlevine: `import * as bitcoin from 'bitgo-utxo-lib'`

modules/core/src/v2/coins/eos.ts
```diff
@@ -8,6 +8,7 @@ import * as utxoLib from 'bitgo-utxo-lib';
 import * as url from 'url';
 import * as querystring from 'querystring';
 import * as _ from 'lodash';
+import * as ecc from'eosjs-ecc'
```
tylerlevine: wait sorry, missing whitespace here

marktoda: whoops not sure why the linter didnt see that

PR #414, opened 2019-07-07
CT-256 - algorand offline signing

twtaylorbitgo:
Algorand offline signing.

Note this changes the Algorand signing flow (instead of the build return json it returns the pre-signed tx encoded as base64) and is tied to a client and platform PR as well.

### DIFF
--- a/modules/core/src/v2/coins/algo.ts
+++ b/modules/core/src/v2/coins/algo.ts
@@ -3,8 +3,7 @@
  */
 import { BaseCoin } from '../baseCoin';
 import * as _ from 'lodash';
-
-const {
+import {
   NaclWrapper,
   Multisig,
   Address,
@@ -13,7 +12,69 @@ const {
   generateAccount,
   isValidAddress,
   isValidSeed,
-} = require('algosdk');
+  Encoding,
+} from 'algosdk';
+
+export interface TransactionExplanation {
+  displayOrder: string[];
+  id: string;
+  outputs: Output[],
+  changeOutputs: Output[],
+  outputAmount: string;
+  changeAmount: number;
+  fee: TransactionFee;
+  memo: string;
+}
+
+export interface SignTransactionOptions {
+  txPrebuild: TransactionPrebuild;
+  prv: string;
+  wallet: {
+    addressVersion: string;
+  }
+  keychain: KeyPair;
+  backupKeychain: KeyPair;
+  bitgoKeychain: KeyPair;
+}
+
+export interface TransactionPrebuild {
+  txBase64: string;
+  txInfo: {
+    from: string;
+    to: string;
+    amount: string;
+    fee: number;
+    firstRound: number;
+    lastRound: number;
+    genesisID: string;
+    genesisHash: string;
+    note?: string;
+  }
+}
+
+export interface HalfSignedTransaction {
+  halfSigned: {
+    txBase64: string,
+  }
+}
+
+export interface Output {
+  address: string;
+  amount: string;
+}
+
+export interface TransactionFee {
+  fee: number;
+  feeRate?: number;
+  size?: number
+}
+
+export interface ExplainTransactionOptions {
+  txBase64: string;
+  wallet: {
+    addressVersion: string;
+  }
+}
 
 interface KeyPair {
   pub: string;
@@ -89,7 +150,7 @@ export class Algo extends BaseCoin {
    * @param {String} prv the prv to be checked
    * @returns {Boolean} is it valid?
    */
-  isValidPrv(prv): boolean {
+  isValidPrv(prv: string): boolean {
     return isValidSeed(prv);
   }
 
@@ -99,7 +160,7 @@ export class Algo extends BaseCoin {
    * @param {String} address the pub to be checked
    * @returns {Boolean} is it valid?
    */
-  isValidAddress(address): boolean {
+  isValidAddress(address: string): boolean {
     return isValidAddress(address);
   }
 
@@ -109,7 +170,7 @@ export class Algo extends BaseCoin {
    * @param key
    * @param message
    */
-  signMessage(key, message): Buffer {
+  signMessage(key: KeyPair, message: string | Buffer): Buffer {
     // key.prv actually holds the encoded seed, but we use the prv name to avoid breaking the keypair schema.
     // See jsdoc comment in isValidPrv
     let seed = key.prv;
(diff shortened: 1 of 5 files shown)


modules/core/src/v2/coins/algo.ts
```diff
@@ -135,28 +198,82 @@ export class Algo extends BaseCoin {
...
+
+      tx = Multisig.MultiSigTransaction.from_obj_for_encoding(txnForDecoding);
+    } catch (ex) {
+      throw new Error('txBase64 needs to be a valid tx encoded as base64 string');
+    }
+
+    const id = tx.txID();
+    const fee = {
+      fee: tx.fee,
+      feeRate: null,
```
argjv: `feeRate` is optional according to `TransactionFee`, we should remove it instead of setting it `null`

modules/core/src/v2/coins/algo.ts
```diff
@@ -135,28 +198,82 @@ export class Algo extends BaseCoin {
...
+      size: null,
```
argjv: Can we make this optional in `TransactionFee` and omit it? Or is expected somewhere else?

modules/core/src/v2/coins/algo.ts
```diff
@@ -135,28 +198,82 @@ export class Algo extends BaseCoin {
...
+    };
+
+    const outputs = [{
+      amount: tx.amount,
+      address: Address.encode(new Uint8Array(tx.to.publicKey)),
+    }];
+
+    const outputAmount = tx.amount;
+
+    // TODO: maybe fix this up
```
argjv: If this comment is for the receiving address feature, add the ticket number CT-480

modules/core/src/v2/coins/algo.ts
```diff
@@ -135,28 +198,82 @@ export class Algo extends BaseCoin {
...
-  keyIdsForSigning() {
+  keyIdsForSigning(): number[] {
     return [ 0, 1, 2 ]; 
   }
 
+  /**
+   * Explain/parse transaction
+   * @param params
+   * - txBase64: transaction encoded as base64 string
+   * @returns {{displayOrder: [string,string,string,string,string], id: *, outputs: Array, changeOutputs: Array}}
```
argjv: This looks odd. Can `displayOrder` type be expressed as `displayOrder: string[]`

modules/core/src/v2/coins/algo.ts
```diff
@@ -205,15 +304,29 @@ export class Algo extends BaseCoin {
     const pair = generateAccountFromSeed(seed);
     const sk = pair.sk;
 
+    // decode our tx
+    let transaction;
+    try {
+      const txToHex = Buffer.from(txBase64, 'base64');
+      const decodedTx = Encoding.decode(txToHex);
```
argjv: In `explainTransaction` you do almost the same, but check for `const txnForDecoding = decodedTx.txn || decodedTx;`

Why not here? And if you do, please move the code to its own method.

modules/core/src/v2/wallet.ts
```diff
@@ -167,6 +167,23 @@ export class Wallet {
     }).call(this).asCallback(callback);
   }
 
+  /**
+   * Gets keys for signing
+   * @param params
```
argjv: This is `reqId`

modules/core/src/v2/wallet.ts
```diff
@@ -167,6 +167,23 @@ export class Wallet {
...
+   * @param callback
+   * @returns {Wallet}
```
argjv: Doesn't look like we return a wallet object.

modules/core/src/v2/coins/algo.ts
twtaylorbitgo: sure

modules/core/src/v2/coins/algo.ts
```diff
@@ -13,8 +13,71 @@ const {
...
+  prv: string;
+  wallet: {
+    addressVersion: string;
+  }
+  keychain: KeyPair;
+  backupKeychain: KeyPair;
+  bitgoKeychain: KeyPair;
+}
+
+interface TransactionPrebuild {
```
tylerlevine: Why not exported?

tylerlevine:
Please have a look at the tests as well

modules/core/src/v2/coins/algo.ts
twtaylorbitgo: I think so

modules/core/src/v2/coins/algo.ts
twtaylorbitgo: ok done

modules/core/src/v2/coins/algo.ts
twtaylorbitgo: hmm sure this is the standard applied in the other docs. I'll put displayOrder[6]

twtaylorbitgo: err string[5]

modules/core/src/v2/coins/algo.ts
twtaylorbitgo: the difference is explainTransaction can take in an unsigned or signed transaction, whereas signTransaction is only primed for an unsigned tx. we would have to call merge on this if we expect more than 1 sig (which we don't so far)

modules/core/src/v2/wallet.ts
twtaylorbitgo: ack

modules/core/src/v2/wallet.ts
twtaylorbitgo: fixed

modules/core/src/v2/coins/algo.ts
```diff
@@ -205,15 +304,29 @@ export class Algo extends BaseCoin {
...
     );
 
-    const txHex = Buffer.from(halfSigned).toString('base64');
+    const signedBase64 = Buffer.from(halfSigned).toString('base64');
 
-    return { txHex };
+    return { 
+      halfSigned: { 
+        txHex: signedBase64,  
+        txBase64: signedBase64,
```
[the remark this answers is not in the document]
twtaylorbitgo: I've removed this, yeah we didn't need to do this

modules/core/src/v2/wallet.ts
```diff
@@ -167,6 +167,23 @@ export class Wallet {
...
+   */
+  keysForSigning(reqId?: string, callback?: NodeCallback<Wallet>):Bluebird<any> {
```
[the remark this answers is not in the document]
twtaylorbitgo: ah good point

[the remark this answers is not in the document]
twtaylorbitgo: sure but are there static methods in Keychains?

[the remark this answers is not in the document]
twtaylorbitgo: I don't see anything that's not just attached to the prototype

modules/core/src/v2/coins/algo.ts
```diff
@@ -13,8 +13,71 @@ const {
...
+  id: string;
+  outputs: Output[],
+  changeOutputs: Output[],
+  outputAmount: string;
+  changeAmount: number;
+  fee: TransactionFee;
+  memo: string;
+}
+
+interface SignTransactionOptions {
```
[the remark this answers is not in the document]
twtaylorbitgo: ack

modules/core/src/v2/coins/algo.ts
twtaylorbitgo: ack

modules/core/src/v2/coins/algo.ts
```diff
@@ -13,8 +13,71 @@ const {
   generateAccount,
   isValidAddress,
   isValidSeed,
+  Encoding,
 } = require('algosdk');
```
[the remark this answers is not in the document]
twtaylorbitgo: hmm maybe

PR #416, opened 2019-07-08
CT-508 - Add explainTransaction method for EOS

CapnMigraine:
Add explainTransaction method used by the client and the ovc.


modules/core/src/v2/coins/eos.ts
```diff
@@ -284,4 +307,81 @@ export class Eos extends BaseCoin {
...
+      return transaction;
+    }).call(this);
+  }
+
+  /**
+   * Explain/parse transaction
+   * @param params - ExplainTransactionOptions
+   * @returns Object containing transaction explanation
+   */
+  explainTransaction(params: ExplainTransactionOptions): Bluebird<TransactionExplanation> {
```
twtaylorbitgo: I don't think this will work in ovc as it doesn't use a promise...

modules/core/src/v2/coins/eos.ts
```diff
@@ -284,4 +307,81 @@ export class Eos extends BaseCoin {
...
+   */
+  deserializeTransaction({ txHex, transactionHeaders }: ExplainTransactionOptions): Bluebird<object> {
+    return co(function *() {
+      const eosClientConfig = { chainId: this.getChainId(), transactionHeaders };
+      const eosClient = new EosJs(eosClientConfig);
+
+      // Get tx base values
+      const eosTxStruct = eosClient.fc.structs.transaction;
+      const serializedTxBuffer = Buffer.from(txHex, 'hex');
+      const transaction = Fcbuffer.fromBuffer(eosTxStruct, serializedTxBuffer);
```
tylerlevine: What is the type of this var? Can we do better than `any`?

modules/core/src/v2/coins/eos.ts
```diff
@@ -5,11 +5,15 @@ import { BaseCoin } from '../baseCoin';
 import { BigNumber } from 'bignumber.js';
 import * as crypto from 'crypto';
 import * as utxoLib from 'bitgo-utxo-lib';
+import * as EosJs from 'eosjs';
+import * as Fcbuffer from 'fcbuffer';
```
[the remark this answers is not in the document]
marktoda: its part of eosjs, maybe we can pull it in separately?

modules/core/src/v2/coins/eos.ts
marktoda: To avoid importing fcbuffer, you should be able to use the one bound to eosclient, i.e. eosClient.fc.fromBuffer({})

CapnMigraine: yes, I added an interface for `DeserializedEosTransaction`

modules/core/src/v2/coins/eos.ts
```diff
@@ -284,4 +325,81 @@ export class Eos extends BaseCoin {
...
+   * @param params - ExplainTransactionOptions
+   * @returns Object containing transaction explanation
+   */
+  explainTransaction(params: ExplainTransactionOptions): Bluebird<TransactionExplanation> {
+    return co(function *() {
+      let transaction;
+      try {
+        transaction = yield this.deserializeTransaction(params);
+      } catch (e) {
+        // throw new Error('txHex needs to be a hex string');
```
tylerlevine: Please clean up leftover comment, and if you really just want to rethrow, then the whole try catch is unnecessary. Maybe consider throwing a custom error type instead?

PR #421, opened 2019-07-10
CT-361: Typescript markets

tylerlevine:
* Also adds 'check-fmt' script to check prettier status

### DIFF
--- a/modules/core/src/v2/markets.ts
+++ b/modules/core/src/v2/markets.ts
@@ -1,78 +1,78 @@
-//
-// Markets Object
-// BitGo accessor to Bitcoin market data.
-//
-// Copyright 2015, BitGo, Inc.  All Rights Reserved.
-//
-
-
-import common = require('../common');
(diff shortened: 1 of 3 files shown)


modules/core/src/v2/markets.ts
```diff
@@ -1,78 +1,78 @@
...
-    if (days && days < 0) {
-      throw new Error('must use a non-negative number of days');
-    }
+      const days = !isNaN(parseInt(params.days, 10)) ? parseInt(params.days, 10) : 90;
 
-    const res = yield this.bitgo.get(this.baseCoin.url('/market/last/' + days + '/' + params.currencyName));
-    return res.body;
-  }).call(this).asCallback(callback);
-};
+      if (days && days < 0) {
```
dannybitgo: days can be 0?

dannybitgo: oh its no change

PR #427, opened 2019-07-11
CT-256 - renaming txbase64 to txhex in algo & trust custodial implementation

twtaylorbitgo:
We need to fixup Algo to be txHex instead of txBase64 (design decision) and then trust custodial implementation items

### DIFF
--- a/modules/core/src/v2/coins/algo.ts
+++ b/modules/core/src/v2/coins/algo.ts
@@ -14,6 +14,7 @@ import {
   isValidSeed,
   Encoding,
 } from 'algosdk';
+import * as stellar from 'stellar-sdk';
 
 export interface TransactionExplanation {
   displayOrder: string[];
@@ -29,12 +30,6 @@ export interface TransactionExplanation {
 export interface SignTransactionOptions {
   txPrebuild: TransactionPrebuild;
   prv: string;
-  wallet: {
-    addressVersion: string;
-  }
-  keychain: KeyPair;
-  backupKeychain: KeyPair;
-  bitgoKeychain: KeyPair;
 }
 
 export interface TransactionPrebuild {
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/algo.ts
```diff
@@ -50,6 +45,8 @@ export interface TransactionPrebuild {
     genesisHash: string;
     note?: string;
   }
+  keys: string[],
```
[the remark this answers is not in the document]
twtaylorbitgo: doneski

modules/core/src/v2/coins/algo.ts
```diff
@@ -244,10 +239,22 @@ export class Algo extends BaseCoin {
...
   }
 
+  isStellarSeed(seed: string): boolean {
+    return stellar.StrKey.isValidEd25519SecretSeed(seed);
+  }
+
+  convertFromStellarSeed(seed: string): string {
+    // assume this is a trust custodial seed if its a valid ed25519 prv
+    if (!this.isStellarSeed(seed)) {
+      return null;
```
argjv: Why not just return the same one you receive in the params? This way, this method will always return a valid seed, and there is no risk of hitting NPE

modules/core/src/v2/coins/algo.ts
```diff
@@ -244,10 +239,22 @@ export class Algo extends BaseCoin {
...
-      memo
+      memo,
     };
   }
 
+  isStellarSeed(seed: string): boolean {
+    return stellar.StrKey.isValidEd25519SecretSeed(seed);
+  }
+
+  convertFromStellarSeed(seed: string): string {
```
argjv: Docs pls

modules/core/src/v2/coins/algo.ts
tylerlevine: Good call, this should really throw an Error instead of returning null.

argjv: https://bitgoinc.atlassian.net/browse/CT-630

PR #431, opened 2019-07-15
CT-656 - Make explainTransaction methods return promise

CapnMigraine:
The explainTransaction method for EOS makes an asynchronous call and returns a promise. In order to be consistent, other coins should return a promise as well.

References:
https://bitgoinc.atlassian.net/browse/CT-656
https://bitgoinc.atlassian.net/browse/CT-637

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1,5 +1,6 @@
 import { hdPath } from '../../bitcoin';
 import { BaseCoin } from '../baseCoin';
+import { NodeCallback } from '../types';
 const config = require('../../config');
 const bitcoin = require('bitgo-utxo-lib');
 const bitcoinMessage = require('bitcoinjs-message');
@@ -12,6 +13,38 @@ const RecoveryTool = require('../recovery');
 const errors = require('../../errors');
 const debug = require('debug')('bitgo:v2:utxo');
 import { Codes, VirtualSizes } from '@bitgo/unspents';
+
+export interface Output {
+  address: string;
+  amount: string;
+}
+
+export interface TransactionFee {
+  fee: number;
+  feeRate?: number;
+  size: number
+}
+
+export interface AbstractUtxoCoinTransactionExplanation {
+  displayOrder: string[];
+  id: string;
+  outputs: Output[],
+  changeOutputs: Output[],
+  outputAmount: string;
+  changeAmount: number;
+  fee: TransactionFee;
+}
+
+export interface Unspent {
+  id: string,
+  value: string,
+}
+
+export interface AbstractUtxoCoinExplainTransactionOptions {
+  txHex: string;
+  txInfo: { changeAddresses: string[], unspents: Unspent[] };
+  feeInfo?: string;
+}
 
 export class AbstractUtxoCoin extends BaseCoin {
 
@@ -191,7 +224,7 @@ export class AbstractUtxoCoin extends BaseCoin {
       const keySignatures = _.get(wallet, '_wallet.keySignatures');
 
       // obtain all outputs
-      const explanation = this.explainTransaction({
+      const explanation = yield this.explainTransaction({
         txHex: txPrebuild.txHex,
         txInfo: txPrebuild.txInfo,
         keychains: keychains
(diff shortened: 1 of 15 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -13,6 +13,38 @@ const errors = require('../../errors');
...
+  amount: string;
+}
+
+export interface TransactionFee {
+  fee: number;
+  feeRate?: number;
+  size: number
+}
+
+export interface TransactionExplanation {
```
marktoda: Is this the same interface across coins? Maybe it could be moved to baseCoin

dannybitgo:
@CapnMigraine This is a breaking change, and would break the OVC: https://github.com/BitGo/offline-signing-tool/blob/f8341f88fba9b2ecb10125909a2520aa563d5e7a/frontend/scripts/signTransaction/signTransactionController.js#L150

dannybitgo:
Does EOS need to return a promise for this? Can we find a way to wrap it so that doesn't happen instead?

CapnMigraine:
> @CapnMigraine This is a breaking change, and would break the OVC: https://github.com/BitGo/offline-signing-tool/blob/f8341f88fba9b2ecb10125909a2520aa563d5e7a/frontend/scripts/signTransaction/signTransactionController.js#L150

Yeah, I'm working on changes to OVC to support this change

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1006,126 +1038,128 @@ export class AbstractUtxoCoin extends BaseCoin {
     return areAllSignaturesValid;
   }
 
-  explainTransaction(params = {} as any) {
-    const { txHex, txInfo } = params;
+  explainTransaction(params: ExplainTransactionOptions): Bluebird<TransactionExplanation> {
```
tylerlevine: This should also take an optional `NodeCallback<TransactionExplanation>`, then pass to `asCallback` on the coroutine. Same for all the other implementations.

modules/core/src/v2/coins/xrp.ts
```diff
@@ -376,40 +376,42 @@ export class Xrp extends BaseCoin {
...
-      txHex = params.txHex;
-    } catch (e) {
+  explainTransaction(params: ExplainTransactionOptions): Bluebird<TransactionExplanation> {
+    return co(function *() {
+      let transaction;
+      let txHex;
       try {
-        transaction = JSON.parse(params.txHex);
-        txHex = rippleBinaryCodec.encode(transaction);
+        transaction = rippleBinaryCodec.decode(params.txHex);
```
tylerlevine: Why is this change here?

modules/core/src/v2/coins/xrp.ts
```diff
@@ -485,32 +487,32 @@ export class Xrp extends BaseCoin {
    * @param callback
    */
   public recover(params: RecoveryOptions, callback): Bluebird<RecoveryInfo | string> {
```
tylerlevine: callback param should be `callback?: NodeCallback<RecoveryInfo | string>`

modules/core/test/v2/unit/coins/btc.ts
```diff
@@ -413,36 +413,48 @@ describe('BTC:', function() {
...
-        it('should fail for invalid transaction hexes', () => {
-          (() => coin.explainTransaction())
-          .should.throw('invalid transaction hex, must be a valid hex string');
-
-          (() => coin.explainTransaction({
-            txHex: ''
-          })).should.throw('invalid transaction hex, must be a valid hex string');
+        it('should fail for invalid transaction hexes', co(function *() {
+          try {
+            yield coin.explainTransaction();
```
tylerlevine: `yield coin.explainTransaction().should.be.rejectedWith('invalid ...');` is a little cleaner

modules/core/src/v2/coins/abstractUtxoCoin.ts
CapnMigraine: It's not the same, but I created one in baseCoin that's extended by each coin that implements explainTransaction.

modules/core/src/v2/coins/xrp.ts
CapnMigraine: git is showing it more complicated than it is... this try/catch matches the one in line 382 of the old version, the next one in line 387 is the same as the one in line 386 of the old version. Only the indentation changed.

modules/core/src/v2/coins/xrp.ts
CapnMigraine: ok, fixed

modules/core/test/v2/unit/coins/btc.ts
CapnMigraine: good one, changed.

modules/core/src/v2/coins/eos.ts
```diff
@@ -82,16 +86,7 @@ interface DeserializedEosTransaction extends EosTransactionHeaders {
...
-
-interface TransactionExplanation {
-  displayOrder: string[];
-  id: string;
-  outputs: Recipient[];
-  changeOutputs: Recipient[];
-  outputAmount: string;
-  changeAmount: number;
-  fee: {};
+  memo?: string;
```
CapnMigraine: small non-related fix, memo was not present in result

modules/core/src/v2/coins/eos.ts
```diff
@@ -365,6 +360,7 @@ export class Eos extends BaseCoin {
       const transferActionData = EosJs.modules.Fcbuffer.fromBuffer(transferStruct, serializedTransferDataBuffer);
       transaction.address = transferActionData.to;
       transaction.amount = this.bigUnitsToBaseUnits(transferActionData.quantity.split(' ')[0]);
+      transaction.memo = transferActionData.memo;
```
CapnMigraine: related to non-related fix up in line 89

modules/core/src/v2/coins/algo.ts
```diff
@@ -49,34 +44,25 @@ export interface TransactionPrebuild {
...
+interface TransactionExplanation extends BaseCoinTransactionExplanation {
+  memo: string;
 }
 
 interface KeyPair {
   pub: string;
   prv: string;
 }
 
-const MAX_ALGORAND_NOTE_LENGTH = 1024;
```
CapnMigraine: removed this because it's not used

PR #432, opened 2019-07-15
BG-13597 / Update payload verification for settlement fees

dacruz21:
Fees are now specified in the trade payload returned from the backend. Update verification logic to match the requested amount against the `subtotal` field, and also verify that the total is calculated correctly

### DIFF
--- a/modules/core/src/v2/trading/tradingAccount.ts
+++ b/modules/core/src/v2/trading/tradingAccount.ts
@@ -1,6 +1,7 @@
 /**
  * @prettier
  */
+import { BigNumber } from 'bignumber.js';
 import * as Bluebird from 'bluebird';
 
 import { Payload } from './payload';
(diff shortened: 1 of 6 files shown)


modules/core/src/v2/trading/tradingAccount.ts
```diff
@@ -2,6 +2,7 @@
  * @prettier
  */
 import * as Bluebird from 'bluebird';
+import BigNumber from 'bignumber.js'
```
tylerlevine: `import { BigNumber }`, it's better to not use the default export.

See https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html for more info

dacruz21:
> Just one nit. Does the `CURRENT_PAYLOAD_VERSION` need to be updated?

Nope, we're not bumping the payload version for fees for some reason

> I think the tests are legit failures too - please take a look

Yep, I need to update some fixtures. Sorry for the delay, got way sidetracked yesterday.

PR #440, opened 2019-07-17
Ct 360 typescript keychains

dannybitgo:
https://bitgoinc.atlassian.net/browse/CT-360

Lets typescriptify this baby

### DIFF
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -14,9 +14,10 @@ import { Wallet } from './wallet';
 import { Wallets } from './wallets';
 import { Markets } from './markets';
 import { Webhooks } from './webhooks';
-const Keychains = require('./keychains');
 import { PendingApprovals } from './pendingApprovals';
+import { Keychains } from './keychains';
 import { Enterprises } from './enterprises';
+import { KeyPair } from './keychains';
 
 export interface BaseCoinTransactionOutput {
   address: string;
@@ -44,7 +45,7 @@ export abstract class BaseCoin {
   protected readonly _url: string;
   protected readonly _enterprises: Enterprises;
   protected readonly _wallets: Wallets;
-  protected readonly _keychains;
+  protected readonly _keychains: Keychains;
   protected readonly _webhooks: Webhooks;
   protected readonly _pendingApprovals;
   protected readonly _markets: Markets;
@@ -399,7 +400,7 @@ export abstract class BaseCoin {
    *
    * @param seed
    */
-  generateKeyPair(seed) {
+  generateKeyPair(seed): KeyPair {
     throw new Error('abstract method');
   }
 
--- a/modules/core/src/v2/coins/algo.ts
+++ b/modules/core/src/v2/coins/algo.ts
@@ -18,6 +18,7 @@ import {
 import * as stellar from 'stellar-sdk';
 import * as Bluebird from 'bluebird';
 const co = Bluebird.coroutine;
+import { KeyPair } from '../keychains';
 
 export interface SignTransactionOptions {
   txPrebuild: TransactionPrebuild;
@@ -53,11 +54,6 @@ interface ExplainTransactionOptions {
 
 interface TransactionExplanation extends BaseCoinTransactionExplanation {
   memo: string;
-}
-
-interface KeyPair {
-  pub: string;
-  prv: string;
 }
 
 export class Algo extends BaseCoin {
--- a/modules/core/src/v2/keychains.ts
+++ b/modules/core/src/v2/keychains.ts
@@ -1,228 +1,308 @@
-import common = require('../common');
+import { validateParams } from '../common';
 import * as _ from 'lodash';
-import * as Promise from 'bluebird';
+import * as Bluebird from 'bluebird';
+import { BaseCoin } from './baseCoin';
+import { NodeCallback } from './types';
+import { Wallet } from './wallet'
 const util = require('../util');
-const co = Promise.coroutine;
+const co = Bluebird.coroutine;
 
-const Keychains = function(bitgo, baseCoin) {
-  this.bitgo = bitgo;
-  this.baseCoin = baseCoin;
-};
+export interface KeyPair {
+  pub?: string;
+  prv: string;
+}
 
-Keychains.prototype.get = function(params, callback) {
-  params = params || {};
-  common.validateParams(params, [], ['xpub', 'ethAddress'], callback);
+export interface GetKeychainOptions {
+  id: string;
+  xpub?: string;
+  ethAddress?: string;
+  reqId?: string;
+}
 
-  if (!params.id) {
-    throw new Error('id must be defined');
+export interface ListKeychainOptions {
+  limit?: number;
+  prevId?: string;
+}
+
+export interface UpdatePasswordOptions {
+  oldPassword: string;
+  newPassword: string;
+}
+
+interface UpdateSingleKeychainPasswordOptions {
+  keychain?: any;
+  oldPassword?: string;
+  newPassword?: string;
+}
+
+interface AddKeychainOptions {
+  pub?: string;
+  encryptedPrv?: string;
+  type?: string;
+  source?: string;
+  originalPasscodeEncryptionCode?: string;
+  enterprise?: string;
+  derivedFromParentWithSeed?: any;
+  disableKRSEmail?: boolean;
+  provider?: string;
+  reqId?: string;
+  krsSpecific?: any
+}
+
+export interface CreateBackupOptions {
+  provider?: string;
+  source?: string;
+}
+
+interface GetKeysForSigningOptions {
+  reqId?: string;
+  wallet?: Wallet;
+}
+
+export class Keychains {
+
+  private readonly bitgo: any;
+  private readonly baseCoin: BaseCoin;
+
+  constructor(bitgo: any, baseCoin: BaseCoin) {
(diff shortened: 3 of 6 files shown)


dannybitgo:
This first commit seeks some feedback. Curious how far down the typescript rabbit hole I should aim to go.

modules/core/src/v2/keychains.ts
```diff
@@ -120,7 +128,7 @@ Keychains.prototype.updatePassword = function(params, callback) {
  * @param callback
  * @returns {object}
  */
-Keychains.prototype.updateSingleKeychainPassword = function({ keychain, oldPassword, newPassword }) {
+Keychains.prototype.updateSingleKeychainPassword = function({ keychain, oldPassword, newPassword }): any {
```
dannybitgo: I wasn't sure what to do here with the params, because they are actually in one object and things I tried weren't working

modules/core/src/v2/keychains.ts
```diff
@@ -120,46 +128,28 @@ Keychains.prototype.updatePassword = function(params, callback) {
...
  */
-Keychains.prototype.updateSingleKeychainPassword = function({ keychain, oldPassword, newPassword }) {
-  if (!_.isString(oldPassword)) {
-    throw new Error('expected old password to be a string');
-  }
-
-  if (!_.isString(newPassword)) {
-    throw new Error('expected new password to be a string');
-  }
+Keychains.prototype.updateSingleKeychainPassword = function( params: { keychain: any, oldPassword: string, newPassword: string }): any {
```
dannybitgo: i dont think this a breaking change is it?

modules/core/src/v2/keychains.ts
```diff
@@ -1,16 +1,24 @@
...
+}
+
+const Keychains = function(bitgo: any, baseCoin: BaseCoin) {
   this.bitgo = bitgo;
   this.baseCoin = baseCoin;
 };
 
-Keychains.prototype.get = function(params, callback) {
-  params = params || {};
+Keychains.prototype.get = function(params: { id: string, xpub?: string, ethAddress?: string, reqId?: string }, callback: NodeCallback<any>) {
```
tylerlevine: Please pull out into interface named `GetKeychainOptions`. The usual naming pattern for options interfaces is <methodname><classname>Options

modules/core/src/v2/keychains.ts
```diff
@@ -1,16 +1,24 @@
...
 const util = require('../util');
 const co = Promise.coroutine;
 
-const Keychains = function(bitgo, baseCoin) {
+export interface KeyPair {
+  pub?: string;
+  prv: string;
+}
+
+const Keychains = function(bitgo: any, baseCoin: BaseCoin) {
```
tylerlevine: This should be refactored into an ES6 class.

modules/core/src/v2/keychains.ts
```diff
@@ -33,7 +41,7 @@ Keychains.prototype.get = function(params, callback) {
  * @param callback
  * @returns {*}
  */
-Keychains.prototype.list = function(params, callback) {
+Keychains.prototype.list = function(params: any, callback: NodeCallback<any>) {
```
tylerlevine: If we don't take any params, and the params object is not used, I think the best way to declare it is `params: {} = {}`. Also callback parameters should always be optional, so please do `callback?:`. Also, missing return type.

modules/core/src/v2/keychains.ts
```diff
@@ -75,7 +83,7 @@ Keychains.prototype.list = function(params, callback) {
  *    ...
  *  }
  */
-Keychains.prototype.updatePassword = function(params, callback) {
+Keychains.prototype.updatePassword = function(params: { oldPassword: string, newPassword: string }, callback: NodeCallback<any>) {
```
tylerlevine: Extract to interface, make callback optional, add return type

modules/core/src/v2/keychains.ts
tylerlevine: Needs return type as well.

modules/core/src/v2/keychains.ts
tylerlevine: No it's not, but please extract into interface. Can we do any better than `any` for the return type?

modules/core/src/v2/keychains.ts
```diff
@@ -120,46 +128,38 @@ Keychains.prototype.updatePassword = function(params, callback) {
...
+    const newEncryptedPrv = this.bitgo.encrypt({ input: decryptedPrv, password: params.newPassword });
+    return _.assign({}, params.keychain, { encryptedPrv: newEncryptedPrv });
   } catch (e) {
     // catching an error here means that the password was incorrect or, less likely, the input to decrypt is corrupted
     throw new Error('password used to decrypt keychain private key is incorrect');
   }
 };
 
-Keychains.prototype.create = function(params) {
+Keychains.prototype.create = function(params: { seed?: string }): KeyPair {
```
tylerlevine: For very small parameter object, defining the type inline like this is good 👍

modules/core/src/v2/keychains.ts
```diff
@@ -120,46 +128,38 @@ Keychains.prototype.updatePassword = function(params, callback) {
...
-Keychains.prototype.add = function(params, callback) {
-  params = params || {};
-  common.validateParams(params, [], ['pub', 'encryptedPrv', 'type', 'source', 'originalPasscodeEncryptionCode', 'enterprise', 'derivedFromParentWithSeed'], callback);
-
-  if (!_.isUndefined(params.disableKRSEmail)) {
-    if (!_.isBoolean(params.disableKRSEmail)) {
-      throw new Error('invalid disableKRSEmail argument, expecting boolean');
-    }
-  }
+Keychains.prototype.add = function(params: { pub?: string, encryptedPrv?: string, type?: string, source?: string, originalPasscodeEncryptionCode?: string, enterprise?: string, derivedFromParentWithSeed?: any, disableKRSEmail?: boolean, provider?: string, reqId?: string, krsSpecific?: any }, callback: NodeCallback<any>): any {
```
tylerlevine: Please extract into interface. Optional callback and return type here as well.

modules/core/src/v2/keychains.ts
```diff
@@ -120,46 +128,38 @@ Keychains.prototype.updatePassword = function(params, callback) {
...
   common.validateParams(params, [], []);
-
   return this.baseCoin.generateKeyPair(params.seed);
 };
 
-Keychains.prototype.add = function(params, callback) {
-  params = params || {};
-  common.validateParams(params, [], ['pub', 'encryptedPrv', 'type', 'source', 'originalPasscodeEncryptionCode', 'enterprise', 'derivedFromParentWithSeed'], callback);
-
-  if (!_.isUndefined(params.disableKRSEmail)) {
```
tylerlevine: We still need this - please don't remove. Typescript does not validate the objects at runtime.

modules/core/src/v2/keychains.ts
```diff
@@ -181,7 +181,7 @@ Keychains.prototype.add = function(params, callback) {
   .nodeify(callback);
 };
 
-Keychains.prototype.createBitGo = function(params, callback) {
+Keychains.prototype.createBitGo = function(params: any, callback: NodeCallback<any>): any {
```
tylerlevine: Doc comment please, also I think we can probably do better than `any` for the types here as well. Just adding any as the type is not really any better than leaving it alone, since it is an any type by default.

dannybitgo: its meant to be empty, but I can't set it to {} because on line 187 it sets ```params.source``` which is not allowed

modules/core/src/v2/baseCoin.ts
```diff
@@ -11,7 +11,7 @@ import { Wallet } from './wallet';
 import { Wallets } from './wallets';
 import { Markets } from './markets';
 import { Webhooks } from './webhooks';
-const Keychains = require('./keychains');
+import { Keychains } from './keychains';
```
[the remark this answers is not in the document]
dannybitgo: not sure which line you are referring to here. maybe the numbers got changed with a recent commit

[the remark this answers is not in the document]
dannybitgo: oh nvm i see

modules/core/src/v2/keychains.ts
```diff
@@ -1,228 +1,316 @@
...
-  common.validateParams(params, [], []);
+  /**
+   * Update the password used to decrypt a single keychain
+   * @param keychain - The keychain whose password should be updated
+   * @param oldPassword - The old password used for encrypting the key
+   * @param newPassword - The new password to be used for encrypting the key
+   * @param callback
+   * @returns {object}
+   */
+  updateSingleKeychainPassword(params: UpdateSingleKeychainPasswordOptions): any {
```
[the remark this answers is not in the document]
dannybitgo: hmm it won't let me assign a default of ```{}``` because the params' properties are not optional

modules/core/src/v2/coins/algo.ts
```diff
@@ -55,10 +56,7 @@ interface TransactionExplanation extends BaseCoinTransactionExplanation {
   memo: string;
 }
 
-interface KeyPair {
-  pub: string;
-  prv: string;
-}
+const MAX_ALGORAND_NOTE_LENGTH = 1024;
```
tylerlevine: I think this was removed in another PR. Please remove if you don't intend on re-adding it.

PR #445, opened 2019-07-19
CT-637 - Add EOS recovery support

CapnMigraine:
https://bitgoinc.atlassian.net/browse/CT-637

### DIFF
--- a/modules/core/src/v2/coins/eos.ts
+++ b/modules/core/src/v2/coins/eos.ts
@@ -5,16 +5,18 @@ import { BaseCoin, BaseCoinTransactionExplanation } from '../baseCoin';
 import { NodeCallback } from '../types';
 import { BigNumber } from 'bignumber.js';
 import * as crypto from 'crypto';
-import * as utxoLib from 'bitgo-utxo-lib';
+import { HDNode } from 'bitgo-utxo-lib';
 import * as EosJs from 'eosjs';
 import * as ecc from 'eosjs-ecc';
 import * as url from 'url';
 import * as querystring from 'querystring';
 import * as _ from 'lodash';
-import { InvalidAddressError, UnexpectedAddressError } from '../../errors';
 import * as Bluebird from 'bluebird';
 const co = Bluebird.coroutine;
 import { KeyPair } from '../keychains';
+import { InvalidAddressError, UnexpectedAddressError } from '../../errors';
+import * as config from '../../config';
+import { Environments } from '../environments';
 
 const EOS_ADDRESS_LENGTH = 12;
 
@@ -44,6 +46,7 @@ interface EosTransactionAction {
   account: string;
   name: string;
   authorization: [{ actor: string; permission: string }];
+  data: { from: string; to: string; quantity: string; memo?: string };
 }
 
 interface EosTransactionPrebuild {
@@ -87,6 +90,36 @@ interface ExplainTransactionOptions {
   headers: EosTransactionHeaders;
 }
 
+interface SendableTransaction {
+  compression: string;
+  packed_trx: string;
+  signatures: string[];
+}
+
+interface RecoveryTransaction {
+  tx: SendableTransaction;
+  txid: string;
+  recoveryAmount: number;
+}
+
+interface RecoveryOptions {
+  userKey: string; // Box A
+  backupKey: string; // Box B
+  bitgoKey?: string; // Box C
+  recoveryDestination: string;
+  krsProvider?: string;
+  walletPassphrase?: string;
+  rootAddress: string;
+}
+
+interface ValidateKeyOptions {
+  key: string;
+  source: string;
+  passphrase?: string;
+  isUnsignedSweep: boolean;
+  isKrsRecovery: boolean;
+}
+
 export class Eos extends BaseCoin {
   static createInstance(bitgo: any): BaseCoin {
     return new Eos(bitgo);
@@ -121,6 +154,12 @@ export class Eos extends BaseCoin {
   }
 
   /**
+   * Get URLs of some active public nodes
+   */
+  getPublicNodeUrls(): string[] {
+    return Environments[this.bitgo.env].eosNodeUrls;
+  }
+  /**
    * Generate secp256k1 key pair
    *
    * @param seed - Seed from which the new keypair should be generated, otherwise a random seed is used
@@ -132,7 +171,7 @@ export class Eos extends BaseCoin {
       // maximum entropy and gives us maximum security against cracking.
       seed = crypto.randomBytes(512 / 8);
     }
-    const extendedKey = utxoLib.HDNode.fromSeedBuffer(seed);
+    const extendedKey = HDNode.fromSeedBuffer(seed);
     const xpub = extendedKey.neutered().toBase58();
     return {
       pub: xpub,
@@ -147,7 +186,7 @@ export class Eos extends BaseCoin {
    */
   isValidPub(pub: string): boolean {
     try {
-      utxoLib.HDNode.fromBase58(pub);
+      HDNode.fromBase58(pub);
       return true;
     } catch (e) {
       return false;
@@ -161,7 +200,7 @@ export class Eos extends BaseCoin {
    */
   isValidPrv(prv: string): boolean {
     try {
-      utxoLib.HDNode.fromBase58(prv);
+      HDNode.fromBase58(prv);
       return true;
     } catch (e) {
       return false;
@@ -302,7 +341,7 @@ export class Eos extends BaseCoin {
     const tx: EosTx = params.txPrebuild.tx;
 
     const signBuffer: Buffer = Buffer.from(txData, 'hex');
-    const privateKeyBuffer: Buffer = utxoLib.HDNode.fromBase58(prv)
+    const privateKeyBuffer: Buffer = HDNode.fromBase58(prv)
       .getKey()
       .getPrivateKeyBuffer();
     const signature: string = ecc.Signature.sign(signBuffer, privateKeyBuffer).toString();
@@ -372,7 +411,6 @@ export class Eos extends BaseCoin {
    * Explain/parse transaction
    * @param params - ExplainTransactionOptions
    * @param callback
-   * @returns Object containing transaction explanation
    */
   explainTransaction(
     params: ExplainTransactionOptions,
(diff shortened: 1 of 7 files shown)


modules/core/src/v2/coins/eos.ts
```diff
@@ -91,6 +93,29 @@ interface ExplainTransactionOptions {
   headers: EosTransactionHeaders;
 }
 
+interface SendableTransaction {
+  compression: string;
+  packed_trx: string;
+  signatures: string[];
+}
```
marktoda: \n

modules/core/src/v2/coins/eos.ts
```diff
@@ -91,6 +93,29 @@ interface ExplainTransactionOptions {
...
+interface RecoveryTransaction extends SendableTransaction {
```
marktoda: Why extend sendabletransaction but also have a sendable transaction inside? Maybe RecoveryTransaction should be a basecoin class, extended here as EosRecoveryTransaction

modules/core/src/v2/coins/eos.ts
```diff
@@ -124,6 +149,12 @@ export class Eos extends BaseCoin {
     return true;
   }
 
+  /**
+   * Get URL of a public node
+   */
+  getPublicNodeUrl(): string {
+    return 'https://bp.cryptolions.io';
```
marktoda: maybe a few different options for each network in case one is down? If that's too annoying we can do it later if it becomes a problem

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
       .call(this)
       .asCallback(callback);
   }
+
+  /**
+   * Validate a public or private key
+   * If passphrase is provided, try to decrypt the key with it
+   * @param key
```
marktoda: types?

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
...
+   * @param source
+   * @param passphrase
+   * @param isUnsignedSweep
+   * @param isKrsRecovery
+   */
+  validateKey = function({ key, source, passphrase, isUnsignedSweep, isKrsRecovery }: ValidateKeyParams): HDNode {
```
marktoda: types for params

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
...
+  getTransactionHeadersFromNode(): Bluebird<any> {
+    return co(function*() {
+      const nodeUrl = this.getPublicNodeUrl();
+      const chainInfoResult = yield this.bitgo.post(nodeUrl + '/v1/chain/get_info').send();
+      if (!chainInfoResult || chainInfoResult.status !== 200) {
+        throw new Error('unable to fetch chain info from node: ' + nodeUrl);
+      }
+      const chainInfo = chainInfoResult.body;
+
+      const headBlockInfoResult = yield this.bitgo.post(nodeUrl + '/v1/chain/get_block').send({
```
marktoda: Since you're already using a public node (i.e. assuming that bitgo is down), can we just use the get_block EOS api to get this info?

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
...
+
+      return {
+        expiration: expirationDate.toISOString(),
+        ref_block_num: chainInfo.head_block_num & 0xffff,
+        ref_block_prefix: headBlockInfoResult.body.ref_block_prefix,
+      };
+    }).call(this);
+  }
+
+  getTransferAction = function({ recipient, sender, amount, memo }) {
```
marktoda: Seeing the amount of stuff that is copied in platform and here makes me wonder how we can refactor for reuse -- it'd be nice if maybe Statics could have utility functions like this for each coin. Just a thought, not relevant to this PR

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
...
+        to: recipient,
+        quantity: `${this.baseUnitsToBigUnits(amount)} EOS`,
+        memo: !_.isNil(memo) ? memo : '', // Memo must be defined, set it to empty string if it is not
+      },
+    };
+  };
+
+  /**
+   * Sign a transaction with a key
+   * @param tx
```
marktoda: There are so many different types of `tx` objects in EOS I think we should be as specific as possible -- like "signableTxString" or something lol. "tx" could be the packed_trx, or this, or the stringified send JSON -- just too confusing

modules/core/src/v2/coins/eos.ts
tylerlevine: Typedoc will take them from the typescript types, so no need.

modules/core/src/v2/coins/eos.ts
tylerlevine: Types are in `ValidateKeyParams`. However, the pattern we have  been following is to use the suffix `Options`,  so this should be `ValidateKeyOptions`.

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
...
+      ];
+      if (!this.isValidAddress(recoveryDestination)) {
+        throw new Error('Invalid destination address!');
+      }
+
+      return keys;
+    }).call(this);
+  }
+
+  getAccountFromNode(address): Bluebird<any> {
```
tylerlevine: Doc comment please, also add types

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
...
+      const nodeUrl = this.getPublicNodeUrl();
+      const result = yield this.bitgo.post(nodeUrl + '/v1/chain/get_account').send({ account_name: address });
+      if (!result || result.status !== 200) {
+        throw new Error('unable to fetch account from node: ' + nodeUrl);
+      }
+      return result.body;
+    }).call(this);
+  }
+
+  getTransactionHeadersFromNode(): Bluebird<any> {
```
tylerlevine: Doc comment

modules/core/src/v2/coins/eos.ts
tylerlevine: No need for `= function`

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
...
+      },
+    };
+  };
+
+  /**
+   * Sign a transaction with a key
+   * @param tx
+   * @param signingKey
+   */
+  signTx = function(tx: string, signingKey: HDNode) {
```
tylerlevine: No need for `= function`, please add return type

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
...
+    return ecc.Signature.sign(signBuffer, privateKeyBuffer).toString();
+  };
+
+  /**
+   * Serialize an EOS transaction, to the format that should be signed
+   * @param eosClient an offline EOSClient that has the transaction structs
+   * @param transaction The EOS transaction returned from `eosClient.transaction` to serialize
+   * @return {String} serialized transaction in hex format
+   */
+  serializeTransaction = function(eosClient, transaction) {
```
tylerlevine: No need for `= function`, please add types

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +438,268 @@ export class Eos extends BaseCoin {
...
+
+    return EosJs.modules.Fcbuffer.toBuffer(eosTxStruct, txObject).toString('hex');
+  };
+
+  /**
+   * Builds a funds recovery transaction without BitGo
+   * @param params
+   * @param callback
+   */
+  recover(params: RecoveryOptions, callback): Bluebird<RecoveryTransaction> {
```
tylerlevine: callback param should be `callback?: NodeCallback<RecoveryTransaction>`

modules/core/src/v2/coins/teos.ts
```diff
@@ -18,4 +18,8 @@ export class Teos extends Eos {
   getFullName() {
     return 'Testnet EOS';
   }
+
+  getPublicNodeUrl(): string {
+    return 'https://jungle2.cryptolions.io';
```
tylerlevine: consider adding to environments.ts instead of adding to the coin class.

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +441,315 @@ export class Eos extends BaseCoin {
...
+   * @param passphrase
+   * @param isUnsignedSweep
+   * @param isKrsRecovery
+   */
+  validateKey({ key, source, passphrase, isUnsignedSweep, isKrsRecovery }: ValidateKeyOptions): HDNode {
+    if (!key.startsWith('xprv') && !isUnsignedSweep) {
+      // Try to decrypt the key
+      try {
+        if (source === 'user') {
+          return HDNode.fromBase58(sjcl.decrypt(passphrase, key));
```
tylerlevine: Please use `this.bitgo.decrypt()` instead of calling sjcl directly

modules/core/src/v2/coins/eos.ts
```diff
@@ -408,4 +441,315 @@ export class Eos extends BaseCoin {
...
+      return keys;
+    }).call(this);
+  }
+
+  /**
+   * Make a request to one of the public EOS nodes available
+   * @param endpoint
+   * @param payload
+   */
+  getDataFromNode({ endpoint, payload }: { endpoint: string; payload?: object }): Bluebird<any> {
```
tylerlevine: Can this be made protected? Or does this need to be part of the public API?

Same for other helper methods below - I'd rather not expose these unless really needed.

modules/core/src/v2/coins/xlm.ts
```diff
@@ -46,15 +42,7 @@ interface VerifyAddressOptions {
...
-interface InitiateRecoveryOptions {
-  userKey: string;
-  backupKey: string;
-  recoveryDestination: string;
-  krsProvider?: string;
-  walletPassphrase?: string;
-}
-
-interface RecoveryOptions extends InitiateRecoveryOptions {
+interface RecoveryOptions extends BaseCoinInitiateRecoveryOptions {
```
tylerlevine: Let's do this in a separate PR please.

modules/core/src/v2/coins/eos.ts
```diff
@@ -1,19 +1,22 @@
...
 import * as url from 'url';
 import * as querystring from 'querystring';
 import * as _ from 'lodash';
-import { InvalidAddressError, UnexpectedAddressError } from '../../errors';
 import * as Bluebird from 'bluebird';
 const co = Bluebird.coroutine;
+import { InvalidAddressError, UnexpectedAddressError } from '../../errors';
+import * as config from '../../config';
+import * as sjcl from '../../vendor/sjcl.min.js';
+import * as common from '../../common';
```
tylerlevine: Please import `Environments` from `v2/environments.ts` instead of importing through common.

modules/core/src/v2/coins/eos.ts
```diff
@@ -124,6 +152,12 @@ export class Eos extends BaseCoin {
     return true;
   }
 
+  /**
+   * Get URLs of some active public nodes
+   */
+  getPublicNodeUrls(): string[] {
+    return common.Environments[this.bitgo.env].eosNodeUrls;
```
tylerlevine: Please use `this.bitgo.getEnv()` instead of accessing a private property of the bitgo object.

modules/core/src/v2/coins/eos.ts
```diff
@@ -1,19 +1,22 @@
...
 import * as ecc from 'eosjs-ecc';
 import * as url from 'url';
 import * as querystring from 'querystring';
 import * as _ from 'lodash';
-import { InvalidAddressError, UnexpectedAddressError } from '../../errors';
 import * as Bluebird from 'bluebird';
 const co = Bluebird.coroutine;
+import { InvalidAddressError, UnexpectedAddressError } from '../../errors';
+import * as config from '../../config';
+import * as sjcl from '../../vendor/sjcl.min.js';
```
tylerlevine: Remove and use `bitgo.encrypt()` or `bitgo.decrypt()` instead.

modules/core/src/v2/coins/eos.ts
marktoda: oh man yep my bad.. thought this was hitting the ims

PR #450, opened 2019-07-24
CT-714: Unify express configuration systems

tylerlevine:
Previously, bitgo express could get configuration options in three
ways: command line arguments, environment variables, and hard coded
defaults. Some options were settable by one method, but not from others,
and the configuration option names were not consistent.

This change unifies these three configuration sources into a generic
configuration object which can decide which source should take
precedence over others.

### DIFF
--- a/modules/express/README.md
+++ b/modules/express/README.md
@@ -10,11 +10,11 @@ This ensures your keys never leave your network, and are not seen by BitGo. BitG
 
 ## Docker
 
-For most users, we recommend running BitGo Express as a docker container.
+For most users, we recommend running BitGo Express as a docker container, since this is the most secure way to run BitGo Express.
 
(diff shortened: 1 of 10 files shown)


modules/express/README.md
```diff
@@ -30,6 +30,11 @@ $ curl localhost:3080/api/v2/ping
 {"status":"service is ok!","environment":"BitGo Testnet","configEnv":"testnet","configVersion":79}
 ```
 
+You can also give command line arguments to BitGo Express at the end of the docker run command:
+```bash
+$ docker run -it --rm -p 3080:3080 bitgosdk/express:latest --env prod
```
dannybitgo: this will fail unless they provide ```--crtpath``` and ```--keypath``` - maybe update to include these?

PR #452, opened 2019-07-24
CT-745 - Resolve responses in explainTransaction coin methods in SDK

CapnMigraine:
Use Bluebird's resolve in responses from explainTransaction functions that don't explicitly return a promise so that they can be used in calls using `then()`.

https://bitgoinc.atlassian.net/browse/CT-745


CapnMigraine:
Closing this because calls work as expected with the latest version of the SDK

PR #455, opened 2019-07-26
CT-774: Copy test_bitgo and testutil subset into express

tylerlevine:
* This should eventually be extracted out into a separate module which
can be used by any other module without doing a cross-module require


tylerlevine:
I understand that this copy and paste solution is less than optimal, but I don't think we're quite ready to extract the test utilities out into a shared module, and we need to be able to run the tests when installing from npm and not yarn for release build verification.

PR #474, opened 2019-08-08
CT-682 - fix a bug with explain transaction for double signed algo txes

twtaylorbitgo:
Unfortunately ovc can't differentiate between these two states easily. This follows the pattern ethereum has setup as well for verifyTransaction. Small bugfix to these.


twtaylorbitgo:
> Looks good, should really have tests for unsigned/half signed/fully signed cases though.
> 
> Is this needed for today's release?

It is, I'll have to talk to @dannybitgo about this one as well

I can add a test

Issue #468, opened 2019-08-07
generateWallet() fails with a CORS error


DCRichards:
### Versions

Version: 6.2.0
BitGoJS.min.js for browser

### Issue

Calling `generateWallet()` fails with the following error in the browser:

> Access to XMLHttpRequest at 'https://test.bitgo.com/api/v2/tbtc/key' from origin 'http://localhost:8085' has been blocked by CORS policy: Request header field request-id is not allowed by Access-Control-Allow-Headers in preflight response.

This appears to be a mismatch between the headers sent by the client and the response from the `Access-Control-Allow-Headers` header of the serve response and could be a backend issue, but I am highlighting here as the point of encounter.

Other calls work as expected, such as `wallets.list({})`

### Example Request

```js
bitgo.coin('tbtc').wallets().generateWallet({
  passphrase: 'secretpassword1133!!',
  label': `My BitGoJS Wallet'
})
.then(res => console.log(res))
.catch(err => console.error(err));
```

PR #475, opened 2019-08-09
Fix errors in generateWallet and downloadKeycard

DCRichards:
## Issue
* `generateWallet()` was failing due to a CORS error.
* `downloadKeycard()` failed due to incorrect scoping.

Fixes https://github.com/BitGo/BitGoJS/issues/468

## Changes

- [X]  Remove `Request-ID` header from `POST /{coin}/key` and `POST /{coin}/wallet` calls as this is not permitted by `Access-Control-Allow-Headers` response from those endpoints and results in a CORS error in browser.
- [X] Pass `encrypt` function to `downloadKeycard()` as `this.bitgo` is not defined in `keycard.ts`.
- [X] Use `this._wallet.label` instead of `this.label()` or `wallet.label()` as the former is not defined (we only have access to the properties, not instance) in `downloadKeycard` in `keycard.ts`.
- [X] Replace concatenation for template string in `keycard.ts` for easier readability.

### DIFF
--- a/modules/core/src/v2/keycard.ts
+++ b/modules/core/src/v2/keycard.ts
@@ -8,7 +8,7 @@ const generateQuestions = (coin: string) => {
       q: 'What is the KeyCard?',
       a:
         [
-          'The KeyCard contains important information which can be used to recover the ' + coin + ' ',
+          `The KeyCard contains important information which can be used to recover the ${coin} `,
           'from your wallet in several situations. Each BitGo wallet' +
           ' has its own, unique KeyCard. ',
           'If you have created multiple wallets, you should retain the KeyCard for each of them.'
@@ -31,7 +31,7 @@ const generateQuestions = (coin: string) => {
       q: 'What should I do if I lose it?',
       a:
         [
-          'If you have lost or damaged all copies of your KeyCard, your ' + coin + ' is still safe, but this ',
+          `If you have lost or damaged all copies of your KeyCard, your ${coin} is still safe, but this `,
           'wallet should be considered at risk for loss. As soon as is convenient, you should use BitGo ',
           'to empty the wallet into a new wallet',
           ', and discontinue use of the old wallet.'
@@ -61,7 +61,7 @@ const generateQuestions = (coin: string) => {
       a:
         [
           'Your KeyCard and wallet passcode can be used together with BitGo’s published open ',
-          'source tools at https://github.com/bitgo to recover your ' + coin + '. Note: You should never enter ',
+          `source tools at https://github.com/bitgo to recover your ${coin}. Note: You should never enter `,
           'information from your KeyCard into tools other than the tools BitGo has published, or your ',
           'funds may be at risk for theft.'
         ]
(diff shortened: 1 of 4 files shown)


modules/core/src/v2/keycard.ts
```diff
@@ -72,14 +72,15 @@ const generateQuestions = (coin: string) => {
...
-          'your KeyCard will be able to steal your ' + coin + '.' + ' We recommend keeping your wallet password ',
+          `your KeyCard will be able to steal your ${coin}. We recommend keeping your wallet password `,
           'safe in a secure password manager such as LastPass, 1Password or KeePass.'
         ]
     }
   ];
 };
 
 interface GetKeyDataOptions {
+  encrypt: (params: {input: string, password: string}) => any,
```
tylerlevine: This should return a string instead of `any`.

modules/core/src/v2/keychains.ts
```diff
@@ -232,9 +232,6 @@ export class Keychains {
       }
     }
 
-    if (params.reqId) {
```
tylerlevine: This should stay in. I think the reason you were getting the CORS errors is because you were using the default environment (test) instead of the localNonSecure environment, which we recommend for doing local testing (although, I'm not sure this is an option for you as you aren't able to run our platform via localhost, but there is probably a workaround for this).

We should be allowing the Request-ID header to be used cross-domain though, so I'll look into updating the CORS allowed headers.

modules/core/src/v2/wallets.ts
```diff
@@ -347,7 +347,6 @@ export class Wallets {
         bitgoKeychain,
       };
       walletParams = yield this.baseCoin.supplementGenerateWallet(walletParams, keychains);
-      this.bitgo._reqId = reqId;
```
tylerlevine: Please leave this in as well.

modules/core/src/v2/keycard.ts
DCRichards: Great point, not sure why I went with `any` there!

modules/core/src/v2/keychains.ts
DCRichards: Thanks for the detailed response, do let us know if there is a workaround. It looks to be returned in the Access-Control-Allow-Headers from other POST requests on the API when I had a look, so makes sense to have it here too.

tylerlevine: > Thanks for the detailed response, do let us know if there is a workaround. It looks to be returned in the Access-Control-Allow-Headers from other POST requests on the API when I had a look, so makes sense to have it here too.

Workaround for localhost testing would be to use socat or something similar to proxy requests. Here's a quick example:

In your terminal, run this command:
```
$ socat TCP-LISTEN:12345,fork,reuseaddr TCP:test.bitgo.com:443
```

This will make a port forward such that it appears that test.bitgo.com is available at localhost:12345 (you might need to change the port).

Then, in another terminal, you can test your connection by running this curl command:
```
$ curl -v --insecure https://localhost:12345/api/v2/ping
```

Note that --insecure is required, since we can't validate the certificate against localhost. Then you should be able to do local testing without running into CORS issues.

DCRichards: That's a great point - noted!

tylerlevine:
Hey @DCRichards, this change looks good for me. I've created #477 for this change and signed and squashed your commits. I'll close this PR now and we can continue on #477. Thanks again for the fixes!

PR #480, opened 2019-08-16
CT-363 Typescriptify recovery

dannybitgo:
https://bitgoinc.atlassian.net/browse/CT-363

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -177,7 +177,7 @@ export interface OfflineVaultTxInfo {
 export interface RecoverFromWrongChainOptions {
   txid: string;
   recoveryAddress: string;
-  wallet: Wallet;
+  wallet: string;
   walletPassphrase: string;
   xprv: string;
   coin?: AbstractUtxoCoin;
--- a/modules/core/src/v2/recovery.ts
+++ b/modules/core/src/v2/recovery.ts
@@ -1,5 +1,9 @@
+/**
+ * @prettier
+ */
 import { VirtualSizes } from '@bitgo/unspents';
 import * as request from 'superagent';
+import { NodeCallback } from './types';
 import * as Bluebird from 'bluebird';
 const co = Bluebird.coroutine;
 import * as _ from 'lodash';
@@ -15,6 +19,34 @@ interface CrossChainRecoveryToolOptions {
   logging: boolean;
 }
 
+export interface SignRecoveryTransactionOptions {
+  prv?: string;
+  passphrase: string;
+}
+
+export interface BuildRecoveryTransactionOptions {
+  wallet: string;
+  faultyTxId: string;
+  recoveryAddress: string;
+}
+
+export interface RecoveryTxInfo {
+  inputAmount: number;
+  outputAmount: number;
+  spendAmount: number;
+  inputs: any[];
+  outputs: any[];
+  externalOutputs: any[];
+  changeOutputs: any[];
+  minerFee: number;
+  payGoFee: number;
+  unspents: any[];
+}
+
+export interface HalfSignedRecoveryTx {
+  txHex: string;
+  tx?: string;
+}
 /**
  * An instance of the recovery tool, which encapsulates the recovery functions
  * Instantiated with parameters:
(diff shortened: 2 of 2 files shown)


modules/core/src/v2/recovery.ts
```diff
@@ -1,11 +1,22 @@
 import { VirtualSizes } from '@bitgo/unspents';
```
tylerlevine: Consider adding @prettier pragma at the top if it doesn't mess up the formatting too much.

modules/core/src/v2/recovery.ts
```diff
@@ -83,7 +94,7 @@ export class CrossChainRecoveryTool {
    * @param walletId {String} wallet ID
    * @param callback
    */
-  setWallet(walletId, callback?) {
```
tylerlevine: This returns void, not any

modules/core/src/v2/recovery.ts
```diff
@@ -350,7 +361,7 @@ export class CrossChainRecoveryTool {
    * @param outputAmount {Number} amount to send to the recovery address
    * @param recoveryFee {Number} miner fee for the transaction
    */
-  buildOutputs(recoveryAddress, outputAmount, recoveryFee) {
+  buildOutputs(recoveryAddress: string, outputAmount: number, recoveryFee: number) {
```
tylerlevine: Returns void (maybe this is inferred correctly though).

modules/core/src/v2/recovery.ts
```diff
@@ -389,7 +400,7 @@ export class CrossChainRecoveryTool {
    * @param callback
    * @returns {Object} half-signed transaction
    */
-  signTransaction({ prv, passphrase }, callback?) {
+  signTransaction(params: SignRecoveryTransactionOptions, callback?: NodeCallback<any>): any {
```
tylerlevine: Bluebird\<any\> as return type

modules/core/src/v2/recovery.ts
```diff
@@ -416,7 +427,7 @@ export class CrossChainRecoveryTool {
    * @param callback
    * @returns {String} decrypted wallet private key
    */
-  getKeys(passphrase, callback?) {
+  getKeys(passphrase: string, callback?: NodeCallback<any>): string {
```
tylerlevine: Return type is `Bluebird<string>`

modules/core/src/v2/recovery.ts
tylerlevine: Also make protected if this is not used outside this class

modules/core/src/v2/recovery.ts
```diff
@@ -138,7 +149,7 @@ export class CrossChainRecoveryTool {
    * @param faultyTxId {String} the txid of the faulty transaction
    * @param callback
    */
-  findUnspents(faultyTxId, callback?) {
+  findUnspents(faultyTxId: string, callback?: NodeCallback<any>): Bluebird<any> {
```
tylerlevine: Consider making protected

modules/core/src/v2/recovery.ts
```diff
@@ -214,7 +225,7 @@ export class CrossChainRecoveryTool {
    * @param callback
    * @returns {Object} partial txInfo object with transaction inputs
    */
-  buildInputs(unspents, callback?) {
+  buildInputs(unspents: any, callback?: NodeCallback<any>): Bluebird<any> {
```
tylerlevine: Consider protected

modules/core/src/v2/recovery.ts
```diff
@@ -329,7 +340,7 @@ export class CrossChainRecoveryTool {
    * @param recoveryTx {Object} recovery transaction containing inputs
    * @returns {Number} recovery fee for the transaction
    */
-  setFees(recoveryTx) {
+  setFees(recoveryTx: any): number {
```
tylerlevine: Consider protected

modules/core/src/v2/recovery.ts
tylerlevine: Consider protected

modules/core/src/v2/recovery.ts
tylerlevine: Consider protected

modules/core/src/v2/recovery.ts
```diff
@@ -15,8 +26,8 @@ const bitcoin = require('bitgo-utxo-lib');
  */
 export class CrossChainRecoveryTool {
   bitgo: any;
-  sourceCoin: any;
-  recoveryCoin: any;
+  sourceCoin: BaseCoin;
+  recoveryCoin: BaseCoin;
   logging: any;
   supportedCoins: string[];
   wallet: any;
```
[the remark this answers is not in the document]
dannybitgo: Can't use ```Wallet``` type for ```wallet``` because it can be a v1 or v2 wallet based on logic in the code later

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -176,7 +176,7 @@ export interface OfflineVaultTxInfo {
 export interface RecoverFromWrongChainOptions {
   txid: string;
   recoveryAddress: string;
-  wallet: Wallet;
+  wallet: string;
```
dannybitgo: this needs to be a string. in WRW they pass in a string, and thats how we do it in tests as well.

modules/core/src/v2/recovery.ts
dannybitgo: just add:
/**
* @prettier
*/
??

PR #481, opened 2019-08-16
CT-878 - tron skeleton

twtaylorbitgo:
Skeleton for Tron in bitgo SDK


modules/core/src/v2/coins/ttrx.ts
```diff
@@ -0,0 +1,30 @@
...
+   */
+  public getChain(): string {
+    return 'ttrx';
+  }
+
+  /**
+   * Complete human-readable name of this coin
+   */
+  public getFullName(): string {
+    return 'Testnet Tron (Shasta)';
```
tylerlevine: What is shasta?

argjv: It's the most popular Tron private test network: https://www.trongrid.io/shasta

twtaylorbitgo: I'll add a comment to this line

tylerlevine: It's a bit strange though. For example, teth is just "Testnet Ethereum", not "Testnet Ethereum (Kovan)"

modules/core/src/v2/coins/trx.ts
```diff
@@ -0,0 +1,49 @@
+const tronweb = require('tronweb');
```
tylerlevine: Is there a types package for this lib?

modules/core/src/v2/coinFactory.ts
```diff
@@ -120,6 +122,8 @@ GlobalCoinFactory.registerCoinConstructor('zec', Zec.createInstance);
 GlobalCoinFactory.registerCoinConstructor('tzec', Tzec.createInstance);
 GlobalCoinFactory.registerCoinConstructor('algo', Algo.createInstance);
 GlobalCoinFactory.registerCoinConstructor('talgo', Talgo.createInstance);
+GlobalCoinFactory.registerCoinConstructor('trx', Trx.createInstance);
```
tylerlevine: Whelp, we now have a problem as the trx ticker is already taken by an erc20 token: https://github.com/BitGo/BitGoJS/blob/master/modules/core/src/config.ts#L46

modules/core/src/v2/coins/trx.ts
twtaylorbitgo: there is not, unfortunately

modules/core/src/v2/coinFactory.ts
twtaylorbitgo: ok... this has become a topic. i've renamed this token for now, until we have a deprecation path

twtaylorbitgo:
This does not need to go out with the release today

modules/core/src/v2/coinFactory.ts
twtaylorbitgo: i've bumped the statics version

modules/core/src/v2/coins/trx.ts
twtaylorbitgo: i lied. there's a d.ts file out there. I don't think it's official but @dpc found something

twtaylorbitgo:
gonna leave this broken until we publish 2.0.0 rc0 tuesday

modules/core/src/v2/coins/ttrx.ts
```diff
@@ -0,0 +1,32 @@
...
+  public getChain(): string {
+    return 'ttrx';
+  }
+
+  /**
+   * Complete human-readable name of this coin
+   */
+  public getFullName(): string {
+    // shasta is the most popular testnet for tron - more info can be found here: https://www.trongrid.io/shasta
+    return 'Testnet Tron';
```
argjv: Shouldn't we get these from `statics` lib?

twtaylorbitgo: adding the statics stuff

modules/core/src/v2/coins/trx.ts
twtaylorbitgo: after talking about it, the typings file is not complete, we'll probably only push typings for the things we use

modules/core/src/v2/coins/ttrx.ts
twtaylorbitgo: I'm pulling from statics now

modules/core/src/v2/coins/trx.ts
```diff
@@ -0,0 +1,54 @@
...
+  getFamily() {
+    return this._staticsCoin.family;
+  }
+
+  getFullName() {
+    return this._staticsCoin.fullName;
+  }
+
+  getBaseFactor() {
+    return new Number('1e' + this._staticsCoin.decimalPlaces);
```
tylerlevine: `Math.pow(10, this._staticsCoin.decimalPlaces)`

modules/core/test/v2/unit/coins/trx.ts
```diff
@@ -0,0 +1,19 @@
...
+
+describe('TRON:', function() {
+  let bitgo;
+
+  before(function() {
+    bitgo = new TestBitGo({ env: 'mock' });
+    bitgo.initializeTestVars();
+  });
+
+  it('should hoist the coin', function() {
```
tylerlevine: hoist seems like a strange word, maybe you mean host?

The test is actually testing if we can get an instance of the coin.

Maybe something like this would be better?:
```
it('should instantiate trx coin', () => {
  const basecoin = bitgo.coin('trx');
  basecoin.should.be.an.instanceof(Trx);
});
```

modules/core/src/v2/coins/trx.ts
twtaylorbitgo: eh sure

modules/core/test/v2/unit/coins/trx.ts
twtaylorbitgo: done

twtaylorbitgo:
image

PR #482, opened 2019-08-19
CT-367 use statics for token config

dannybitgo:
https://bitgoinc.atlassian.net/browse/CT-367


modules/core/src/config.ts
```diff
@@ -290,17 +51,10 @@ export const tokens: Tokens = {
...
-        { type: 'test', coin: 'teth', network: 'Kovan', tokenContractAddress: '0x1fb879581f31687b905653d4bbcbe3af507bed37', decimalPlaces: 18, name: 'Test Mintable ERC20 Token' },
-        { type: 'tbst', coin: 'teth', network: 'Kovan', tokenContractAddress: '0xe5cdf77835ca2095881dd0803a77e844c87483cd', decimalPlaces: 0, name: 'Test BitGo Shield Token' },
-        { type: 'schz', coin: 'teth', network: 'Kovan', tokenContractAddress: '0x050e25a2630b2aee94546589fd39785254de112c', decimalPlaces: 18, name: 'SchnauzerCoin' },
-        { type: 'tcat', coin: 'teth', network: 'Kovan', tokenContractAddress: '0x63137319f3a14a985eb31547370e0e3bd39b03b8', decimalPlaces: 18, name: 'Test CAT-20 Token' },
-        { type: 'tfmf', coin: 'teth', network: 'Kovan', tokenContractAddress: '0xd8463d2f8c5b3be9de95c63b73a0ae4c79423452', decimalPlaces: 18, name: 'Test Formosa Financial Token' }
-      ]
+      tokens: formattedErc20Tokens.filter(token => token.network === 'Testnet'),
     },
     ofc: {
       tokens: [
```
dannybitgo: Not sure what to do about OFC tokens because there seem to be more OFC tokens in statics than in here

dannybitgo:
Looks like the token ```0x5af2be193a6abca9c8817001f45744777db30756``` is defined twice in statics, which is causing the error

dannybitgo:
Found a second duplicate. This is now blocked on the statics PR: https://github.com/BitGo/statics/pull/48

modules/core/src/config.ts
tylerlevine: We should use the ofc tokens defined in statics, those should have also been added to bitgojs

dannybitgo: looks like there are multiple discrepancies between the OFC tokens here and the ones defined in statics. we should probably discuss the differences.

marktoda:
https://github.com/BitGo/statics/pull/49

dannybitgo:
Typescript compile error because statics does not export ```OfcCoin```

tylerlevine:
@marktoda for visibility

dannybitgo:
@tylerlevine @marktoda @thomas4019 The test is failing because statics has slightly different "full names" for the OFC token. 
BitGoJS previous called if ```Offchain Test USD``` but statics calls it ```Test USD```
similarly it expects ```Offchain Bitcoin Test``` but statics has ```Test Bitcoin```

What do you guys think, should we change statics or bitgojs?

Screen Shot 2019-08-23 at 9 21 05 AM

thomas4019:
I think let's standardize on the names in statics.

modules/core/src/config.ts
```diff
@@ -22,296 +23,52 @@ export interface Tokens {
   };
 }
 
+// Get the list of ERC-20 tokens from statics and format it properly
+const formattedErc20Tokens = coins.filter((coin: BaseCoin) => {
+  return coin instanceof Erc20Coin;
+}).map((token: Erc20Coin): TokenConfig => {
+  return {
+    type: token.name,
+    coin: token.network.type === NetworkType.MAINNET ? 'eth' : 'teth',
```
marktoda: Can just do `token.family`

modules/core/src/config.ts
```diff
@@ -22,296 +23,52 @@ export interface Tokens {
...
+    network: token.network.type === NetworkType.MAINNET ? 'Mainnet' : 'Testnet',
+    name: token.fullName,
+    tokenContractAddress: token.contractAddress.toString(),
```
marktoda: Isn't it already a string? Should lowercase it though, as it would break things here if any tokens were accidentally not lowercased in statics

modules/core/test/v2/unit/coins/ofcToken.ts
```diff
@@ -30,7 +30,7 @@ describe('OFC:', function() {
 
     const teth = bitgo.coin('ofcteth');
     teth.getChain().should.equal('ofcteth');
-    teth.getFullName().should.equal('Offchain Ether Testnet');
+    teth.getFullName().should.equal('Test Ether');
```
marktoda: Is it going to hurt anything that these names are different now?

marktoda: oh didnt seee the comments above

modules/core/src/config.ts
dannybitgo: I was getting weird errors, because we use a type: ```ContractAddress``` (even though it extends string)

modules/core/src/config.ts
dannybitgo: it doesn't looks like there is a ```token.family``` property on the statics coin objects

modules/core/src/config.ts
marktoda: Ah ok -- still lowercase it

modules/core/src/config.ts
marktoda: Ah there is but its always 'eth', my bad

dannybitgo: i can do this:
```    
coin: `${token.network.type === NetworkType.MAINNET ? '' : 't'}${token.family}`,
```

dannybitgo: ```token.family``` gives the mainnet chain only

dannybitgo: https://github.com/BitGo/statics/blob/master/src/base.ts#L15

PR #483, opened 2019-08-19
CT-371 typescriptify expressApp

dannybitgo:
https://bitgoinc.atlassian.net/browse/CT-371

### DIFF
--- a/modules/express/src/expressApp.ts
+++ b/modules/express/src/expressApp.ts
@@ -1,19 +1,20 @@
 import * as express from 'express';
 import * as httpProxy from 'http-proxy';
 import * as url from 'url';
-import * as Promise from 'bluebird';
+import * as Bluebird from 'bluebird';
 import * as path from 'path';
 import * as _ from 'lodash';
 import * as debugLib from 'debug';
 import * as https from 'https';
 import * as http from 'http';
+import { Server } from 'net';
 
 const morgan = require('morgan');
-const fs = Promise.promisifyAll(require('fs'));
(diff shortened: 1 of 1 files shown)


dannybitgo:
It didn't seem like there was too much to change here but lmk if you think I should try to do more

dannybitgo:
> Good start, but I think we are missing some types, and it would also be good to move away from the `module.exports.blah` style and start using the ES6 syntax `export function blah`

How you export something that is a Coroutine, like this one:
```
module.exports.init = co(function *() {
```

modules/express/src/expressApp.ts
```diff
@@ -282,13 +283,13 @@ module.exports.app = function(cfg: Config) {
 
 module.exports.init = co(function *() {
```
dannybitgo: not sure how to change this to the Es6 format

modules/express/src/expressApp.ts
```diff
@@ -168,7 +169,7 @@ function createHttpServer(app) {
  * @param baseUri
  * @return {Function}
  */
-module.exports.startup = function(config: Config, baseUri: string) {
+export function startup(config: Config, baseUri: string): any {
```
tylerlevine: return type here is `() => void` (startup returns a function which returns void)

modules/express/src/expressApp.ts
tylerlevine: You can do this:
```
export function init() {
  return co(function*() {
    // you can use yield in here
  }).call(this);
}
```

PR #484, opened 2019-08-19
CT-353: Typescriptify eth and token

tylerlevine:


modules/core/src/v2/coins/eth.ts
```diff
@@ -895,17 +1198,18 @@ export class Eth extends BaseCoin {
...
-  validateHopPrebuild(wallet: any, hopPrebuild: HopPrebuild, originalParams: any, callback: any): void {
-    return co(function *() {
-      const {
-        tx,
-        id,
-        signature,
-      } = hopPrebuild;
-
+  validateHopPrebuild(
+    wallet: Wallet,
```
dannybitgo: hmm it doesn't look like ```wallet``` is used in this function?

tylerlevine: Yep,  you're right, it is definitely unused. I'll clean it up in the next PR which is almost ready.

PR #488, opened 2019-08-21
CT-366: Typescriptify BitGo

tylerlevine:


modules/core/src/bitgo.ts
```diff
@@ -70,59 +83,54 @@ superagent.Request.prototype.end = function(cb) {
...
+function errFromResponse(res: superagent.Response): Error {
   const errString = createResponseErrorString(res);
   const err: any = new Error(errString);
 
   err.status = res.status;
   if (res.body) {
     err.result = res.body;
   }
-  if (_.has(res.headers, 'x-auth-required') && (res.headers['x-auth-required'] === 'true')) {
+  if (_.has(res.header, 'x-auth-required') && (res.header['x-auth-required'] === 'true')) {
```
dannybitgo: why is this singular now?

modules/core/src/bitgo.ts
```diff
@@ -70,59 +83,54 @@ superagent.Request.prototype.end = function(cb) {
...
 /**
  * There are many ways a request can fail, and may ways information on that failure can be
  * communicated to the client. This function tries to handle those cases and create a sane error string
  * @param res Response from an HTTP request
- * @returns {String}
  */
-const createResponseErrorString = function(res) {
-  let errString = res.statusCode.toString(); // at the very least we'll have the status code
+function createResponseErrorString(res: superagent.Response): string {
+  let errString = res.status.toString(); // at the very least we'll have the status code
```
dannybitgo: wondering why ```statusCode``` is now ```status```

modules/core/src/bitgo.ts
```diff
@@ -70,59 +83,54 @@ superagent.Request.prototype.end = function(cb) {
...
   if (res.body.error) {
     // this is the case we hope for, where the server gives us a nice error from the JSON body
     errString = res.body.error;
   } else {
-    // things get messy from here on, we try different parts of the response, salvaging what we can
```
dannybitgo: why is this deleted?

modules/core/src/bitgo.ts
```diff
@@ -141,171 +149,437 @@ const createResponseErrorString = function(res) {
...
+  limit?: number;
+}
+
+interface SimulateWebhookOptions {
+  webhookId: string;
+  blockId: string;
+}
+
+export interface BitGo {
+  get(url: string, callback?: NodeCallback<superagent.Response>): superagent.Request;
```
dannybitgo: shouldn't the return type for```get``` be a ```Response``` rather than a ```Request```?

modules/core/src/bitgo.ts
```diff
@@ -141,171 +149,437 @@ const createResponseErrorString = function(res) {
...
-      if (self.proxy) {
-        req = req.proxy(self.proxy);
+  /**
+   * This is a patching function which can apply our authorization
+   * headers to any outbound request.
+   * @param method
+   */
+  private createPatch(method: string): (url: string, callback?: NodeCallback<superagent.Response>) => superagent.Request {
+    const self = this;
+    return function(...args) {
```
dannybitgo: is ```args``` correct here, or should it be ```arguments``` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

dannybitgo: oh nvm i think thats answered here https://stackoverflow.com/questions/42184674/what-is-the-meaning-of-args-three-dots-in-a-function-definition

modules/core/src/bitgo.ts
tylerlevine: res.headers does not exist, this was always broken as far as I can tell:

From superagent docs:

> Response header fields
> 
> The res.header contains an object of parsed header fields, lowercasing field names much like node does. For example res.header['content-length'].

modules/core/src/bitgo.ts
tylerlevine: Same thing, statusCode does not exist.

modules/core/src/bitgo.ts
tylerlevine: No, you need to resolve the request into a response by yielding/awaiting it (it's like a promise)

modules/core/src/bitgo.ts
tylerlevine: They are equivalent, but ...args plays better with typescript (`arguments` is a very strange beast)

modules/core/src/bitgo.ts
tylerlevine: I don't think `res.res` ever exists, so this will never actually run.

modules/core/src/bitgo.ts
tylerlevine: I should be a bit more explicit: `res.statusCode` does exist (otherwise this would throw an Error): https://github.com/visionmedia/superagent/blob/master/src/response-base.js#L116

However, it's not part of the type definition, so typescript errors out if you try to access `res.statusCode`. I could either add `statusCode` to the `superagent.Response` type augmentation, or change usages of `statusCode` to `status`, which is in the type definition. I opted for the second approach.

modules/core/src/bitgo.ts
```diff
@@ -427,848 +699,920 @@ const BitGo = function(params) {
...
+  }
 
-  // BIP32 derivation path is applied to both client and server master keys
-  const derivationPath = responseBody.derivationPath;
-  const clientDerivedNode = hdPath(clientHDNode).derive(derivationPath);
-  const serverDerivedNode = hdPath(serverHDNode).derive(derivationPath);
+  /**
+   * Calculate the HMAC for an HTTP request
+   */
+  calculateRequestHMAC({ url: urlPath, text, timestamp, token }: CalculateRequestHmacOptions): string {
```
dannybitgo: curious why this is ```url: urlPath``` rather than just ```url```

modules/core/src/bitgo.ts
```diff
@@ -427,848 +699,920 @@ const BitGo = function(params) {
...
-    authParams.extensionAddress = this._extensionKey.getAddress();
+    return authParams;
   }
 
-  return authParams;
-};
+  /**
+   * Login to the bitgo platform.
+   */
+  authenticate(params: AuthenticateOptions, callback?: NodeCallback<any>): Bluebird<any> {
```
dannybitgo: looks like we can ```yield authenticate``` now right? is this a breaking change though?

modules/core/src/bitgo.ts
```diff
@@ -427,848 +699,920 @@ const BitGo = function(params) {
...
-  .nodeify(callback);
-};
+  /**
+   *
+   * @param params
+   * - pushVerificationToken: the token received via push notification to confirm the device's mobility
+   * @param callback
+   * @deprecated
+   */
+  verifyPushToken(params, callback?: NodeCallback<any>): Bluebird<any> {
```
dannybitgo: no type on ```params``` here, is that deliberate?

modules/core/src/bitgo.ts
```diff
@@ -427,848 +699,920 @@ const BitGo = function(params) {
...
+      .send(postParams)
+      .result()
+      .nodeify(callback);
   }
 
-  const postParams = _.pick(params, 'pushVerificationToken');
+  /**
+   * Login to the bitgo system using an authcode generated via Oauth
+   */
+  authenticateWithAuthCode(params, callback) {
```
dannybitgo: no types here

modules/core/src/bitgo.ts
tylerlevine: I don't think it is, as you could do `yield authenticate()` before this change as well. It returned a promise before, but now we use the coroutine syntax instead of .then()

modules/core/src/bitgo.ts
tylerlevine: Yep, missed a couple, good catch. Will update.

modules/core/src/bitgo.ts
```diff
@@ -1282,687 +1626,626 @@ BitGo.prototype.listAccessTokens = function(params, callback) {
...
-  .result()
-  .nodeify(callback);
-};
+  /**
+   * A factory method to create a new Wallet object, initialized with the wallet params
+   * Can be used to reconstitute a wallet from cached data
+   * @param walletParams
+   * @deprecated
+   */
+  newWalletObject(walletParams): any {
```
dannybitgo: no type here, i think thats intended though

modules/core/src/bitgo.ts
```diff
@@ -1282,687 +1626,626 @@ BitGo.prototype.listAccessTokens = function(params, callback) {
...
-//
-BitGo.prototype.instantGuarantee = function(params, callback) {
-  params = params || {};
-  common.validateParams(params, ['id'], [], callback);
+  /**
+   * Create a url for calling BitGo platform APIs
+   * @param path
+   * @param version
+   */
+  url(path: string, version = 1): string {
```
dannybitgo: ```version: number = 1``` ?

modules/core/src/bitgo.ts
```diff
@@ -427,848 +699,920 @@ const BitGo = function(params) {
...
-  this._user = undefined;
-  this._token = undefined;
-  this._refreshToken = undefined;
-  this._ecdhXprv = undefined;
-};
+  /**
+   * Return the current BitGo environment
+   */
+  getEnv(): EnvironmentName {
+    return this._env;
```
dannybitgo: Now that ```env``` has become ```_env```, wont this possibly mess up customers who might be using that?

modules/core/src/v2/coins/xlm.ts
```diff
@@ -398,7 +401,7 @@ export class Xlm extends BaseCoin {
    * @return true if stellar username is valid
    */
   isValidStellarUsername(username: string): boolean {
-    return /^[a-z0-9\-\_\.\+\@]+$/.test(username);
+    return /^[a-z0-9\-_.+@]+$/.test(username);
```
dannybitgo: no need to escape + and @?

modules/core/src/v2/coins/xlm.ts
```diff
@@ -597,11 +606,11 @@ export class Xlm extends BaseCoin {
         throw new Error('Provided wallet has a balance of 0 XLM, recovery aborted');
       }
 
-      const walletBalance = this.bigUnitsToBaseUnits(nativeBalanceInfo.balance);
-      const minimumReserve = yield this.getMinimumReserve();
-      const baseTxFee = yield this.getBaseTransactionFee();
+      const walletBalance = Number(self.bigUnitsToBaseUnits(nativeBalanceInfo.balance));
```
dannybitgo: do we have to worry about overflow here (i.e. use BigNumbers?)

modules/core/test/v2/integration/bitgo.ts
```diff
@@ -71,13 +71,6 @@ describe('BitGo', function() {
     });
 
     describe('Logout API', function() {
-      it('arguments', function(done) {
```
dannybitgo: why delete?

modules/core/package.json
```diff
@@ -85,6 +85,7 @@
     "@types/node": "^11.11.4",
     "@types/sinon": "^7.0.6",
     "@types/stellar-sdk": "^0.11.1",
+    "@types/superagent": "^4.1.3",
```
dannybitgo: seems odd that ```superagent``` wasn't already here?

modules/core/src/bitgo.ts
dannybitgo: woah that's wild

dannybitgo: thats some 2015 Davenport code

modules/core/src/bitgo.ts
dannybitgo: ok nice

modules/core/src/bitgo.ts
dannybitgo: ah ok

modules/core/src/bitgo.ts
tylerlevine: Because it gets passed to `calculateHMACSubject`, which expects a param called `urlPath`, since it uses the `url` module and thus can't use that name. So something has to change the name of the property. Why it was done here instead of inside calculateHMACSubject, I'm not sure.

modules/core/src/bitgo.ts
tylerlevine: Yeah. Unfortunately this needs to stay 1, since a lot of v1 code calls this with the default 2nd arg.

modules/core/src/bitgo.ts
tylerlevine: I thought of that, so `env` still exists, but it's deprecated.

See https://github.com/BitGo/BitGoJS/pull/488/files/8b35edfb3130a0b3df4db84ac359ea195d4abf3c#diff-b2b93516e1aec818bf25c684902edc17R480 and https://github.com/BitGo/BitGoJS/pull/488/files/8b35edfb3130a0b3df4db84ac359ea195d4abf3c#diff-b2b93516e1aec818bf25c684902edc17R383

modules/core/src/v2/coins/xlm.ts
tylerlevine: Nope, since it's inside a character class (brackets).

tylerlevine:
```
$ node
> /[@.+]/.test('@')
true
```

modules/core/src/v2/coins/xlm.ts
tylerlevine: So funny story. `bigUnitsToBaseUnits` returns a string, `getMinimumReserve` returns a number, and so does `getBaseTransactionFee`. So prior to this change the types in the expression which assigns `recoveryAmount` are `string - number - number`. And that works, because javascript.

This change just makes it `number - number - number`, which has all the same overflow problems as the previous implementation, but at least fixes the mixed types.

modules/core/test/v2/integration/bitgo.ts
tylerlevine: It checks to make sure that it throws if a non-function is passed as the callback arg, which causes other problems regarding dual error channels (long story). I'll address that in the PR description as this is a known breaking change (next version will be 8.0.0).

modules/core/package.json
tylerlevine: This is just for the types package - superagent is already declared in the dependencies section

modules/core/src/bitgo.ts
dannybitgo: oh I'm saying there's no type on it though

dannybitgo:
approved pending passing tests

PR #499, opened 2019-09-03
CT-823 - Add support for Stellar tokens

CapnMigraine:
This change allows instantiating a Stellar token defined in the statics library as:
`bitgo.coin('txlm:BST-GBQTIOS3XGHB7LVYGBKQVJGCZ3R4JL5E4CBSWJ5ALIJUHBKS6263644L');`

The token name is composed of `<coin>:<token code>-<token issuer>`

Depends on: https://github.com/BitGo/statics/pull/57

### DIFF
--- a/modules/core/src/config.ts
+++ b/modules/core/src/config.ts
@@ -1,13 +1,17 @@
 import * as _ from 'lodash';
 import { Environments, EnvironmentName } from './v2/environments';
 import { OfcTokenConfig } from './v2/coins/ofcToken';
-import { TokenConfig } from './v2/coins/token';
-import { coins, BaseCoin, Erc20Coin, OfcCoin, CoinKind, NetworkType } from '@bitgo/statics';
+import { Erc20TokenConfig } from './v2/coins/erc20Token';
+import { StellarTokenConfig } from './v2/coins/stellarToken';
+import { coins, BaseCoin, Erc20Coin, StellarCoin, OfcCoin, CoinKind, NetworkType } from '@bitgo/statics';
 
 export interface Tokens {
   bitcoin: {
     eth: {
-      tokens: TokenConfig[];
+      tokens: Erc20TokenConfig[];
+    };
+    xlm: {
+      tokens: StellarTokenConfig[];
     };
     ofc: {
       tokens: OfcTokenConfig[];
@@ -15,7 +19,10 @@ export interface Tokens {
   };
   testnet: {
     eth: {
-      tokens: TokenConfig[];
+      tokens: Erc20TokenConfig[];
+    };
+    xlm: {
+      tokens: StellarTokenConfig[];
     };
     ofc: {
       tokens: OfcTokenConfig[];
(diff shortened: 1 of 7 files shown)


modules/core/src/v2/coinFactory.ts
```diff
@@ -126,11 +127,16 @@ GlobalCoinFactory.registerCoinConstructor('susd', Susd.createInstance);
 GlobalCoinFactory.registerCoinConstructor('tsusd', Tsusd.createInstance);
 
 for (const token of [...tokens.bitcoin.eth.tokens, ...tokens.testnet.eth.tokens]) {
-  const tokenConstructor = Token.createTokenConstructor(token);
+  const tokenConstructor = Erc20Token.createTokenConstructor(token);
   GlobalCoinFactory.registerCoinConstructor(token.type, tokenConstructor);
   GlobalCoinFactory.registerCoinConstructor(token.tokenContractAddress, tokenConstructor);
 }
 
+for (const token of [...tokens.bitcoin.xlm.tokens, ...tokens.testnet.xlm.tokens]) {
```
marktoda: Any idea why we construct both mainnet and testnet tokens in every env? I know you just copied the style above

marktoda:
also fix tests

modules/core/src/config.ts
```diff
@@ -58,44 +78,54 @@ export const tokens: Tokens = {
...
+  return verifiedTokens;
+};
 
-export const testnetTokens = {};
-_.forEach(tokens.testnet.eth.tokens, function(value) {
-  if (testnetTokens[value.type]) {
-    throw new Error('token : ' + value.type + ' duplicated.');
-  }
-  testnetTokens[value.type] = true;
+const mainnetErc20Tokens = verifyTokens(tokens.bitcoin.eth.tokens);
```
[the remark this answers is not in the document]
CapnMigraine: haha yeah, I checked with Tyler. It's the legacy structure where each key is a v1 env name. We don't know who's using those exported fields ¯\_(ツ)_/¯

modules/core/src/v2/coinFactory.ts
CapnMigraine: I think it's because coinFactory doesn't have the concept of environment, it just receives a string and tries to get an instance of the coin.

modules/core/src/config.ts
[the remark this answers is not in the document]
dannybitgo: i think it stems back from the original ```bitcoinjs-lib``` where the network names are ```testnet``` (for testnet) and ```bitcoin``` (for mainnet)

CapnMigraine:
> LGTM As a skeleton -- we have a few other things to add still (recovery, signing) right?

yes, it's missing a `recover()` method, but for signing we'll use the one defined in the parent.

tylerlevine:
Is this blocked on getting a new release of statics out?

CapnMigraine:
> Is this blocked on getting a new release of statics out?

yes, I'm waiting for v2.0.1 of the statics lib

modules/core/test/v2/unit/baseCoin.ts
```diff
@@ -60,21 +76,26 @@ describe('V2 Base Coin:', function() {
   });
 
   describe('Token initialization', function() {
-    it('Tokens initialized with name and contract should be instances of Token', function() {
-      (basecoinTokenWithName instanceof Token).should.equal(true);
-      (basecoinTokenWithContractHash instanceof Token).should.equal(true);
+    it('ERC20 Tokens initialized with name and contract should be instances of Erc20Token', function() {
+      (basecoinErc20TokenWithName instanceof Erc20Token).should.equal(true);
```
tylerlevine: `basecoinErc20TokenWithName.should.be.instanceof(Erc20Token)`

CapnMigraine: it reads better, changed

PR #513, opened 2019-09-10
BLOCK-16 prebuild should accept gasLimit

dannybitgo:
https://bitgoinc.atlassian.net/browse/BLOCK-16

A customer has identified a bug - when inputting a custom gasLimit in the UI, the value does not get recognized by the SDK, and does not get included in the transaction.

This PR fixes the SDK so that ```prebuildTransaction``` correctly handles gasLimit as a parameter

### DIFF
--- a/modules/core/test/v2/unit/ethWallet.ts
+++ b/modules/core/test/v2/unit/ethWallet.ts
@@ -292,3 +292,36 @@ describe('Add final signature to ETH tx from offline vault', function() {
     response.txHex.should.equal(expectedResult.txHex);
   });
 });
+
+describe('prebuildTransaction', function() {
+  let bitgo;
+  let ethWallet;
+  let recipients;
+  let bgUrl;
+  let gasLimit;
+
+  before(function() {
+    bitgo = new TestBitGo({ env: 'test' });
+    bitgo.initializeTestVars();
+    const coin = bitgo.coin('teth');
+    ethWallet = coin.newWalletObject(bitgo, coin, {});
+    gasLimit = 2100000;
+    recipients = [{
+      address: '0xe59dfe5c67114b39a5662cc856be536c614124c0',
+      amount: '100000'
+    }];
+    bgUrl = common.Environments[bitgo.getEnv()].uri;
+  });
+
+  it('should successfully accept gasLimit as a param', co(function *() {
+    const scope = nock(bgUrl)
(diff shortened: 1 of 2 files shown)


modules/core/test/v2/unit/ethWallet.ts
```diff
@@ -292,3 +292,35 @@ describe('Add final signature to ETH tx from offline vault', function() {
...
+    bitgo.initializeTestVars();
+    const coin = bitgo.coin('teth');
+    ethWallet = coin.newWalletObject(bitgo, coin, {});
+    gasLimit = 2100000;
+    recipients = [{
+      address: '0xe59dfe5c67114b39a5662cc856be536c614124c0',
+      amount: '100000'
+    }];
+    bgUrl = common.Environments[bitgo.getEnv()].uri;
+    nock(bgUrl)
```
tylerlevine: Move into test instead of before block

modules/core/test/v2/unit/ethWallet.ts
```diff
@@ -292,3 +292,35 @@ describe('Add final signature to ETH tx from offline vault', function() {
...
+      .post('/api/v2/teth/wallet/' + ethWallet.id() + '/tx/build', {
+        recipients,
+        gasLimit,
+      })
+      .reply(200, { success: true });
+  });
+
+  it('should successfully accept gasLimit as a param', co(function *() {
+    const prebuild = yield ethWallet.prebuildTransaction({ recipients, gasLimit });
+    prebuild.success.should.equal(true);
```
tylerlevine: This should check to make sure the nock was consumed successfully, since that is what you are really testing (param is correctly passed to server).

dannybitgo: the nock will throw an error if its not called with gasLimit

dannybitgo: are you suggesting I uses sinon and stubs?

marktoda:
please put me or @twtaylorbitgo on any eth-y related stuff even with the change because there are a lot of potential side-effects

dannybitgo:
> please put me or @twtaylorbitgo on any eth-y related stuff even with the change because there are a lot of potential side-effects

crap my bad. felt bad I was already pinging you for everything haha

marktoda:
all good, id rather be in the loop than something break

PR #515, opened 2019-09-11
Improve error message for failed eth import

OttoAllmendinger:
When an eth import fails, include missing package name in the error
message.

Issue: BG-14650


modules/core/src/errors.ts
```diff
@@ -132,10 +132,19 @@ export class AddressGenerationError extends BitGoJsError {
...
+
   public constructor(message?: string) {
-    super(message || 'Ethereum library required for operation is not available');
+    super(message || EthereumLibraryUnavailableError.defaultMessage);
     Object.setPrototypeOf(this, EthereumLibraryUnavailableError.prototype);
   }
+
+  static forImport(packageName: string) {
+    return new EthereumLibraryUnavailableError(
+      `${this.defaultMessage}. Please install ${JSON.stringify(packageName)}.`
```
tylerlevine: why the  JSON.stringify?

modules/core/src/errors.ts
```diff
@@ -132,10 +132,19 @@ export class AddressGenerationError extends BitGoJsError {
 }
 
 export class EthereumLibraryUnavailableError extends BitGoJsError {
+  private static readonly defaultMessage: string =
+    'Ethereum library required for operation is not available';
+
   public constructor(message?: string) {
-    super(message || 'Ethereum library required for operation is not available');
+    super(message || EthereumLibraryUnavailableError.defaultMessage);
```
tylerlevine: How about we allow the  message to just be the package name? Then we don't need `forImport`.

Are  there  any callers  using  the free-form message?

modules/core/src/errors.ts
OttoAllmendinger: my lazy way of adding quotes

modules/core/src/errors.ts
OttoAllmendinger: I don't think so. Will change

PR #516, opened 2019-09-12
SDKT-17: Automatic documentation generation and upload

tylerlevine:
* Set up typedoc + required config
* Upgrade `core` and `express` modules to typescript@3.5.3 to match typedoc version
* Add `scripts/upload-docs.js` for uploading documentation files
* Add drone step to generate docs for tag builds
* Add drone step to call `scripts/upload-docs.js`
* Introduce `bitgosdk/upload-tools` container to build pipeline to reduce install times

Example docs build: https://bitgo-sdk-docs.s3.amazonaws.com/core/docs-test-rel-tag/index.html

Note that docs will only get built for tag builds (that is, builds which are the result of pushing a tag). It is expected that this is mostly releases. Each docs run generates ~80 MB of HTML which gets uploaded to S3, so we should only generate docs as needed and not for each build.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -70,16 +70,43 @@ local ExcludeBranches(pipeline, excluded_branches=branches()) = pipeline + {
   },
 };
 
-local UploadReports(version, tag="untagged", only_changed=false) = {
-  name: "upload reports",
-  image: "node:"  + version,
+local GenerateDocs(version) = {
+  name: "generate docs",
+  image: "bitgosdk/upload-tools:latest",
+  commands: [
+    "yarn run gen-docs",
+  ],
+  when: {
+    status: ["success"],
+    event: ["tag"],
+  }
+};
+
+local UploadDocs(version) = {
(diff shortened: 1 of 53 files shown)


modules/core/test/v2/integration/wallets.ts
```diff
@@ -27,7 +27,7 @@ describe('V2 Wallets:', function() {
   describe('Per-coin tests', function() {
     /** enable tbtg once btg is in test **/
     // TODO enable add txlm when it's supported by the platform and IMS in test
-    const coins = ['tbtc', 'tbch', 'txrp', 'teth', 'tltc', 'trmg'];
+    const coins = ['tbtc', 'tbch', 'teth', 'tltc', 'trmg'];
```
dannybitgo: is this because of the txrp node being down?

tylerlevine: yeah, I think this can be re-enabled now.

PR #524, opened 2019-09-19
Fix for bitgo-utxo-lib types

workflow:
* Fix unintended param renaming instead of type specification
* Remove duplicate P2PKH identifier


tylerlevine:
Great catch, thanks for the fix.

Our internal policies require all commits to be signed by a BitGo engineer, so I'll recreate your PR and make your changes there (you'll still be listed as committer and I'll link back to this PR). The fix should go out in the 8.2.0 release next week.

PR #533, opened 2019-09-20
ACID-38: Custom env should use test serverXpub if network is test

thomas4019:
This makes it easier to connect to test microservices and other test environments using a custom URL.


modules/core/src/v2/environments.ts
```diff
@@ -259,6 +259,12 @@ export const Environments: Environments = {
           ? `https://${process.env.BITGO_CUSTOM_ROOT_URI}/api/v2/txlm/federation`
           : `https://${process.env.BITGO_CUSTOM_ROOT_URI}/api/v2/xlm/federation`,
     },
-    mainnetBase
+    mainnetBase,
+    {
+      serverXpub:
+        process.env.BITGO_CUSTOM_BITCOIN_NETWORK === 'testnet'
```
tylerlevine: Please use `!== 'bitcoin'` to match behavior of other config entries.

PR #535, opened 2019-09-23
SERV-32 fix hop on pending approval

dannybitgo:
https://bitgoinc.atlassian.net/browse/SERV-32
Hop TX's didn't work with pending approvals. 
There was a platform change that got us closer.
This should be the final change to enable hop tx's with pending approvals.

### DIFF
--- a/modules/core/src/v2/environments.ts
+++ b/modules/core/src/v2/environments.ts
@@ -97,13 +97,10 @@ const testnetBase: EnvironmentTemplate = {
   ],
 };
 
-const devBase: EnvironmentTemplate = Object.assign(
-  {
-    hsmXpub:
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/wallet.ts
```diff
@@ -1239,7 +1239,7 @@ export class Wallet {
         'targetWalletUnspents', 'message', 'minValue', 'maxValue', 'sequenceId', 'lastLedgerSequence',
         'ledgerSequenceDelta', 'gasPrice', 'gasLimit', 'noSplitChange', 'unspents', 'changeAddress', 'instant', 'memo', 'addressType',
         'cpfpTxIds', 'cpfpFeeRate', 'maxFee', 'idfVersion', 'idfSignedTimestamp', 'idfUserId', 'strategy',
-        'validFromBlock', 'validToBlock',
+        'validFromBlock', 'validToBlock', 'hop',
```
tylerlevine: I think we figured out that this can be removed.

PR #536, opened 2019-09-25
TOKS-31 - Allow sending trustline txs from SDK

CapnMigraine:
Hot wallets need the ability to send trustline txs to manage tokens they support.

### DIFF
--- a/modules/core/src/v2/coins/xlm.ts
+++ b/modules/core/src/v2/coins/xlm.ts
@@ -27,10 +27,12 @@ import {
   TransactionPrebuild as BaseTransactionPrebuild,
   ParseTransactionOptions,
   ParsedTransaction,
-  VerifyTransactionOptions,
+  VerifyTransactionOptions as BaseVerifyTransactionOptions,
   SignTransactionOptions as BaseSignTransactionOptions,
+  TransactionParams as BaseTransactionParams, ExtraPrebuildParamsOptions,
 } from '../baseCoin';
 import { NodeCallback } from '../types';
+import { Wallet } from '../wallet';
 
 const co = Bluebird.coroutine;
 
@@ -61,6 +63,13 @@ interface RecoveryTransaction {
   recoveryAmount: number;
   backupKey?: string;
   coin?: string;
+}
+
+interface BuildOptions {
+  wallet?: Wallet;
+  recipients?: object[];
+  type?: string;
+  walletPassphrase?: string;
 }
 
 interface TransactionPrebuild extends BaseTransactionPrebuild {
@@ -110,10 +119,27 @@ interface VerifyAddressOptions extends BaseVerifyAddressOptions {
   rootAddress: string;
 }
 
+interface TrustlineOptions {
+  token: string;
+  action: string;
+  limit?: string;
+}
+
+interface TransactionParams extends BaseTransactionParams {
+  trustlines?: TrustlineOptions[];
+}
+
+interface VerifyTransactionOptions extends BaseVerifyTransactionOptions {
+  txParams: TransactionParams;
+}
+
 export class Xlm extends BaseCoin {
   public readonly homeDomain: string;
   public static readonly tokenPatternSeparator = '-'; // separator for token code and issuer
   static readonly maxMemoId: string = '0xFFFFFFFFFFFFFFFF'; // max unsigned 64-bit number = 18446744073709551615
+  // max int64 number supported by the network ((2^63)-1)/(10^7)
+  // See: https://www.stellar.org/developers/guides/concepts/assets.html#amount-precision-and-representation
+  static readonly maxTrustlineLimit: string = '922337203685.4775807';
 
   constructor(bitgo: BitGo) {
     super(bitgo);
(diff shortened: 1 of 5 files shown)


modules/core/test/v2/unit/coins/xlm.ts
```diff
@@ -468,26 +468,157 @@ describe('XLM:', function() {
...
+        const validTransaction = yield basecoin.verifyTransaction({ txParams, txPrebuild, wallet, verification });
+        validTransaction.should.equal(true);
+      }));
+    });
+  });
 
-        it('should resolve a stellar address into an account', co(function *() {
-          const stellarAddress = 'test12345*lobstr.co';
-          const accountId = 'GB5MEYCR4V2NRZH4NSI465ZEFRXPBZ74P43BMYN7AWVMAI5NNCKNZSVY';
+  describe('Federation lookups:', function() {
```
CapnMigraine: moved out of the "Transaction Verification" describe block

modules/core/src/v2/coins/xlm.ts
```diff
@@ -503,6 +527,22 @@ export class Xlm extends BaseCoin {
     return true;
   }
 
+  /**
+   * Get extra parameters for prebuilding a tx
+   * Set empty recipients array in trustline txs
+   */
+  getExtraPrebuildParams(buildParams: ExtraPrebuildParamsOptions, callback?: NodeCallback<BuildOptions>): Bluebird<BuildOptions> {
+    return co(function *() {
```
tylerlevine: Nit: Coroutine here is just adding unneeded boilerplate, consider returning `Bluebird.resolve(params).asCallback(callback)` instead;

modules/core/src/v2/coins/xlm.ts
```diff
@@ -896,26 +936,44 @@ export class Xlm extends BaseCoin {
...
-        throw new Error('transaction prebuild does not have any operations');
-      }
-
-      _.forEach(txParams.recipients, (expectedOutput, index) => {
-        const expectedOutputAddress = self.getAddressDetails(expectedOutput.address);
-        const output = outputOperations[index] as (stellar.Operation.Payment | stellar.Operation.CreateAccount);
-        if (output.destination !== expectedOutputAddress.address) {
-          throw new Error('transaction prebuild does not match expected recipient');
+      if (txParams.type === 'trustline') {
+        const trustlineOperations = _.filter(tx.operations, ['type', 'changeTrust']) as [stellar.Operation.ChangeTrust];
```
tylerlevine: Nit: Consider extracting to separate private helper function `verifyTrustlineTransaction` - this is getting a bit dense.

tylerlevine: Incorrect type: Return type should be `stellar.Operation.ChangeTrust[]`, which is not equivalent to `[stellar.Operation.ChangeTrust]`.

The [] around the type means this is a tuple array, which is an array with a length which is known at compile time. I don't think that's the case here, since we check the length dynamically below.

modules/core/src/v2/wallet.ts
```diff
@@ -71,10 +76,7 @@ export interface PrebuildTransactionOptions {
   validFromBlock?: number;
   validToBlock?: number;
   instant?: boolean;
-  memo?: {
-    value: string;
-    type: string;
-  }[];
+  memo?: Memo;
```
tylerlevine: Should this be an array still?

modules/core/test/v2/unit/coins/xlm.ts
```diff
@@ -468,26 +468,157 @@ describe('XLM:', function() {
...
+        const txPrebuild = yield wallet.prebuildTransaction(txParams);
+        const verification = {
+          disableNetworking: true,
+          keychains: {
+            user: { pub: userKeychain.pub },
+            backup: { pub: backupKeychain.pub },
+          },
+        };
+        try {
+          yield basecoin.verifyTransaction({ txParams, txPrebuild, wallet, verification });
```
tylerlevine: Nit: consider 
```javascript
yield basecoin.verifyTransaction({ txParams, txPrebuild, wallet, verification })
  .should.be.rejectedWith('transaction prebuild does not match expected trustline operations')
```

(this pattern is used several times, this comment applies to all usages)

modules/core/src/v2/coins/xlm.ts
```diff
@@ -110,10 +119,25 @@ interface VerifyAddressOptions extends BaseVerifyAddressOptions {
...
+
+interface VerifyTransactionOptions extends BaseVerifyTransactionOptions {
+  txParams: TransactionParams;
+}
+
 export class Xlm extends BaseCoin {
   public readonly homeDomain: string;
   public static readonly tokenPatternSeparator = '-'; // separator for token code and issuer
   static readonly maxMemoId: string = '0xFFFFFFFFFFFFFFFF'; // max unsigned 64-bit number = 18446744073709551615
+  static readonly maxTrustlineLimit: string = '922337203685.4775807'; // max int64 number supported by the network ((2^63)-1)/(10^7)
```
tylerlevine: Nit: consider leaving a link to https://www.stellar.org/developers/guides/concepts/assets.html#amount-precision-and-representation so the next person doesn't find this magic number so magical

CapnMigraine: ok, done

modules/core/src/v2/coins/xlm.ts

modules/core/src/v2/coins/xlm.ts
CapnMigraine: Changed both

modules/core/src/v2/wallet.ts
CapnMigraine: no, it's just an object

modules/core/test/v2/unit/coins/xlm.ts
CapnMigraine: fixed all

PR #540, opened 2019-09-26
SDKT-32: Improve usability for typescript dependents

tylerlevine:
* Add @types/bluebird and @types/superagent to the deps list from dev deps,
  since these types are exported by the BitGoJS public API
* Export BitGo object interfaces
* Return custom `BitGoRequest` instead of raw superagent Request.
  Methods patched onto the superagent Request object will be declared on
  this interface instead of augmenting superagent.Request.
* Separate source files from type declarations in dist/

This commit covers the following tickets:
* SDKT-32
* SDKT-33
* SDKT-31

### DIFF
--- a/modules/core/README.md
+++ b/modules/core/README.md
@@ -71,6 +71,10 @@ console.dir(result);
 ## More examples
 Further demos and examples in both Javascript and Typescript can be found in the [example](example) directory.
 
+# Using with Typescript
+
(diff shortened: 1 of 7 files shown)


tylerlevine:
Needed to rebase to fix merge conflict - @dannybitgo could you re-ack?

PR #542, opened 2019-09-27
Rel/innsbruck

perrybitgo:
Merge release branch to master

### DIFF
--- a/modules/core/package.json
+++ b/modules/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bitgo",
-  "version": "8.2.1",
+  "version": "8.2.2-rc.0",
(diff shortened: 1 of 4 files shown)


marktoda:
- why are we doing this? dont release changes come from master in the first place?
- Do we need to update any deps ie statics?

tylerlevine:
> why are we doing this? dont release changes come from master in the first place?
 
Yes, except version bumps and package-lock.json updates happen on the release branch currently. To get those changes, we need these back merged into master. The other changes are also in master (they were cherry picked to the release branch), so they have no effect in the back merge.

> Do we need to update any deps ie statics?

No, we always take the most recent 2.x.x version currently.

PR #543, opened 2019-10-01
SDKT-24: Add missing options types in wallet and wallets

tylerlevine:


modules/core/src/v2/wallet.ts
```diff
@@ -105,12 +102,311 @@ export interface GetUserPrvOptions {
   walletPassphrase?: string;
 }
 
-export class Wallet {
+export interface WalletCoinSpecific {
```
twtaylorbitgo: what is this interface meant to capture? coin specific properties of all coins?

tylerlevine: There are some wallet properties which only apply to specific coins. These fields are stored in `_wallet.coinSpecific` inside the wallet object.

There are a couple places where we do this (addresses is another), so calling it `WalletCoinSpecific` makes it clear which type of coin specific object we are referring to.

In general, this is an antipattern. If I were to start from scratch, this is not how I would  have designed it, but we're stuck with it for now.

tylerlevine:
@dannybitgo @twtaylorbitgo can one of you please take a look at this?

PR #544, opened 2019-10-01
SDKT-4: Enable more strict tsc options

tylerlevine:


modules/core/src/config.ts
```diff
@@ -31,46 +31,51 @@ export interface Tokens {
...
-  return {
-    type: token.name,
-    coin: token.network.type === NetworkType.MAINNET ? 'eth' : 'teth',
-    network: token.network.type === NetworkType.MAINNET ? 'Mainnet' : 'Testnet',
-    name: token.fullName,
-    tokenContractAddress: token.contractAddress.toString().toLowerCase(),
-    decimalPlaces: token.decimalPlaces,
-  };
-});
+const formattedErc20Tokens = coins.reduce((acc, coin) => {
```
OttoAllmendinger: the `filter`/`map` variant looked much nicer

what do you think about
```
(coins.filter(...) as Erc20Coin[])
.map(...)
```
?

tylerlevine: Well, I suppose that's subjective since I find the reduce version a bit cleaner, but I agree that there is room for improvement here. Let me take another pass at it.

modules/core/src/config.ts
```diff
@@ -31,46 +31,48 @@ export interface Tokens {
...
+      type: coin.name,
+      coin: coin.network.type === NetworkType.MAINNET ? 'eth' : 'teth',
+      network: coin.network.type === NetworkType.MAINNET ? 'Mainnet' : 'Testnet',
+      name: coin.fullName,
+      tokenContractAddress: coin.contractAddress.toString().toLowerCase(),
+      decimalPlaces: coin.decimalPlaces,
+    });
+  }
+  return acc;
+}, []);
```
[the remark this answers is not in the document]
tylerlevine: We discussed this today in person. We concluded that there are three places where the type can be specified:
1) The top level variable type
2) accumulator argument type
3) accumulator initial value type

I opted for 2 as I find that the cleanest.

PR #545, opened 2019-10-01
SERV-58: Fix environment inheritance ordering

tylerlevine:


modules/core/src/v2/environments.ts
```diff
@@ -103,165 +103,115 @@ const devBase: EnvironmentTemplate = Object.assign({}, testnetBase, {
...
-    devBase
-  ),
-  rmglocalNonSecure: Object.assign(
-    {
-      uri: 'http://rmglocalhost:3000',
-      stellarFederationServerUrl: 'http://rmglocalhost:3000/api/v2/txlm/federation',
-    },
-    devBase
-  ),
-  custom: Object.assign(
```
dannybitgo: just curious, what is "custom" used for?

tylerlevine: Custom is used in ms branch deploys, and some other circumstances where the endpoint isn't necessarily known until runtime (eg, dynamic proxy, running SDK against bitgo-express, etc).

PR #546, opened 2019-10-01
SERV-263: Import statics

tylerlevine:


modules/statics/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
```
tylerlevine: extend root tsconfig

modules/statics/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# @bito/statics code owners and first officers will automatically be requested for review whenever a pull request touches the files they own.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners and first officers will be the default owners for everything in the repo.
+*   tyler@bitgo.com elle@bitgo.com
```
tylerlevine: Incorporate into root codeowners

modules/statics/package-lock.json
```diff
@@ -0,0 +1,2411 @@
+{
```
tylerlevine: Remove

modules/statics/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "types": "./dist/src/index.d.ts",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "tsc",
+    "fmt": "prettier --write 'src/**/*.ts'",
+    "check-fmt": "prettier --check 'src/**/*.ts'",
+    "clean": "rm -r ./dist"
+  },
+  "author": "",
+  "license": "Apache-2.0",
```
tylerlevine: Check license

modules/statics/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/BitGo/statics"
```
tylerlevine: Update

modules/statics/scripts/publish.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
```
tylerlevine: Check publish flow

dannybitgo:
Should we change the express dockerfile to use node 10 as well instead of lts?

tylerlevine:
> Should we change the express dockerfile to use node 10 as well instead of lts?

Yep, we should do that, let me update.

tylerlevine:
> > Should we change the express dockerfile to use node 10 as well instead of lts?
> 
> Yep, we should do that, let me update.

Actually, I'm going to break out all the LTS changes into a separate PR.

PR #547, opened 2019-10-03
SERV-269 update express docker production docs

dannybitgo:
https://bitgoinc.atlassian.net/browse/SERV-269

Update the docker production docs

### DIFF
--- a/modules/express/README.md
+++ b/modules/express/README.md
@@ -46,6 +46,35 @@ Environment: test
 Base URI: http://0.0.0.0:4000
 ```
 
+### Docker in production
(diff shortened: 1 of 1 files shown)


modules/express/README.md
```diff
@@ -46,6 +46,35 @@ Environment: test
 Base URI: http://0.0.0.0:4000
 ```
 
+## Docker in Production
```
tylerlevine: Use `###`, I think this should be at the same level as "Building the docker container". Also I think this would make more sense below that section.

PR #548, opened 2019-10-03
TRON-3 - hot wallet sending

twtaylorbitgo:
[TRON-70](https://bitgoinc.atlassian.net/browse/TRON-70) Tron send transaction

### DIFF
--- a/modules/core/src/v2/coins/trx.ts
+++ b/modules/core/src/v2/coins/trx.ts
@@ -1,8 +1,8 @@
 /**
  * @prettier
  */
+import * as _ from 'lodash';
 import * as Bluebird from 'bluebird';
-
 const tronWeb = require('tronweb');
 
 import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
@@ -140,6 +140,9 @@ export class Trx extends BaseCoin {
     const tx = params.txPrebuild.transaction;
     // this prv should be hex-encoded
     const halfSigned = tronWeb.utils.crypto.signTransaction(params.prv, tx);
+    if (_.isEmpty(halfSigned.signature)) {
+      throw new Error('failed to sign transaction');
+    }
     return { halfSigned };
   }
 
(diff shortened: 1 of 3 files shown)


modules/core/src/v2/coins/trx.ts
```diff
@@ -15,11 +16,30 @@ import {
...
 import { NodeCallback } from '../types';
 
+const co = Bluebird.coroutine;
+
+export interface TronSignTransactionOptions extends SignTransactionOptions {
+  txPrebuild: TransactionPrebuild;
+  prv: string;
+}
+
+export interface TransactionPrebuild {
```
marktoda: This should `extend BaseTransactionPrebuild` from baseCoin so we have some common interface between coins

modules/core/src/v2/coins/trx.ts
```diff
@@ -113,8 +133,10 @@ export class Trx extends BaseCoin {
     return Bluebird.resolve(true).asCallback(callback);
   }
 
-  signTransaction(params: SignTransactionOptions = {}): SignedTransaction {
-    throw new MethodNotImplementedError();
+  signTransaction(params: TronSignTransactionOptions): SignedTransaction {
+    const tx = { txID: params.txPrebuild.txInfo.txID };
```
marktoda: re above we will need to change this

modules/core/src/v2/coins/trx.ts
```diff
@@ -20,6 +21,24 @@ import * as utxoLib from 'bitgo-utxo-lib';
...
 
+export interface TronSignTransactionOptions extends SignTransactionOptions {
+  txPrebuild: TransactionPrebuild;
+  prv: string;
+}
+
+export interface TransactionPrebuild {
+  txHex: string;
+  txid: string;
+  transaction: any;
```
twtaylorbitgo: this is actually a string | { txID: string }

twtaylorbitgo: also it can have signatures already in an array but its an optional param

argjv: `transaction` is the full transaction object, not just the `id`
We need id so that `signTransaction` return the full transaction ready to be signed and broadcasted.

PR #549, opened 2019-10-03
TOKS-33 Allow removing Stellar trustlines

CapnMigraine:
The SDK already allows sending a tx to add or update a trustline, with this change it will also allow removing it.

https://bitgoinc.atlassian.net/browse/TOKS-33

### DIFF
--- a/modules/core/src/v2/coins/xlm.ts
+++ b/modules/core/src/v2/coins/xlm.ts
@@ -908,15 +908,22 @@ export class Xlm extends BaseCoin {
     _.forEach(trustlineOperations, op => {
       const opToken = this.getTokenNameFromStellarAsset(op.line);
       const tokenTrustline = _.find(txParams.trustlines, trustline => {
-        return trustline.token === opToken &&
-          ((_.isUndefined(trustline.limit) && op.limit === Xlm.maxTrustlineLimit) || // no limit was specified, max by default
-            trustline.limit === op.limit);
+        // Prepare the conditions to check for
+        // Limit will always be set in the operation, even if it was omitted from txParams in the following cases:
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/xlm.ts
```diff
@@ -908,15 +908,18 @@ export class Xlm extends BaseCoin {
     _.forEach(trustlineOperations, op => {
       const opToken = this.getTokenNameFromStellarAsset(op.line);
       const tokenTrustline = _.find(txParams.trustlines, trustline => {
-        return trustline.token === opToken &&
-          ((_.isUndefined(trustline.limit) && op.limit === Xlm.maxTrustlineLimit) || // no limit was specified, max by default
-            trustline.limit === op.limit);
+        const noLimit = _.isUndefined(trustline.limit); // no limit was specified
+        const addTrustlineWithDefaultLimit = (trustline.action === 'add' && op.limit === Xlm.maxTrustlineLimit);
+        const removeTrustline = (trustline.action === 'remove' && op.limit === '0');
+        return (trustline.token === opToken &&
```
tylerlevine: Condition logic here is making my head hurt a bit, but I think it's correct. Consider bringing some of these expressions out into variables and simplifying the return so it's a bit easier to follow.

CapnMigraine: I thought I was simplifying the condition in the `return` statement by extracting the other 3 into  `noLimit `, `addTrustlineWithDefaultLimit ` and `addTrustlineWithDefaultLimit `. I added more comments.

tylerlevine: Thanks, that helps clear it up a bit.

PR #550, opened 2019-10-04
Remove params.targetWalletUnspents

OttoAllmendinger:
This is a phantom parameter that was described for the `consolidate`
route (but does not exist there either).

Issue: BG-14108


OttoAllmendinger:
superseded by https://github.com/BitGo/BitGoJS/pull/551

PR #551, opened 2019-10-04
Add new method `wallet.manageUnspents`

OttoAllmendinger:
* Combine consolidate/fanout into single method (line on the backend)
* Regroup and fix documentation (remove unimplemented params, add missing ones)

Issue: BG-14108

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -550,7 +550,9 @@ export class Wallet {
    * @param callback
    * @returns txHex {String} the txHex of the incomplete transaction that needs to be signed by the user in the SDK
    */
-  manageUnspents(routeName: ManageUnspents, params: any = {}, callback?: NodeCallback<any>): Bluebird<any> {
+  private manageUnspents(routeName: ManageUnspents, params: any = {}, callback?: NodeCallback<any>): Bluebird<any> {
+    const self = this;
+
     return co(function *() {
       common.validateParams(params, [], ['walletPassphrase', 'xprv'], callback);
 
@@ -570,22 +572,22 @@ export class Wallet {
         routeName === 'consolidate' ? 'limit' : 'maxNumUnspentsToUse',
         'numUnspentsToMake',
       ]);
-      this.bitgo.setRequestTracer(reqId);
-      const response = yield this.bitgo.post(this.url(`/${routeName}Unspents`))
+      self.bitgo.setRequestTracer(reqId);
+      const response = yield self.bitgo.post(self.url(`/${routeName}Unspents`))
         .send(filteredParams)
         .result();
 
-      const keychain = yield this.baseCoin.keychains().get({ id: this._wallet.keys[0], reqId });
+      const keychain = yield self.baseCoin.keychains().get({ id: self._wallet.keys[0], reqId });
       const transactionParams = _.extend({}, params, { txPrebuild: response, keychain });
-      const signedTransaction = yield this.signTransaction(transactionParams);
+      const signedTransaction = yield self.signTransaction(transactionParams);
       const selectParams = _.pick(params, ['comment', 'otp']);
       const finalTxParams = _.extend({}, signedTransaction, selectParams);
 
-      this.bitgo.setRequestTracer(reqId);
-      return this.bitgo.post(this.baseCoin.url('/wallet/' + this._wallet.id + '/tx/send'))
+      self.bitgo.setRequestTracer(reqId);
+      return self.bitgo.post(self.baseCoin.url('/wallet/' + self._wallet.id + '/tx/send'))
         .send(finalTxParams)
         .result();
-    }.bind(this)).call(this).asCallback(callback);
+    }).call(this).asCallback(callback);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


modules/core/src/v2/wallet.ts
```diff
@@ -517,90 +519,87 @@ export class Wallet {
...
+   *                 params.maxNumUnspentsToUse  for routeName === 'fanout'
+   *                  - maximum number of unspents you want to use in the transaction
+   * Output parameters:
+   * @param {Number} params.numUnspentsToMake - the number of new unspents to make
+   *
    * @param callback
    * @returns txHex {String} the txHex of the incomplete transaction that needs to be signed by the user in the SDK
    */
-  consolidateUnspents(params: any = {}, callback?: NodeCallback<any>): Bluebird<any> {
-    const self = this;
```
tylerlevine: Why remove this? Accessing this via outer `self` variable allows tsc to check properties of `this`. Inside coroutines, `this` is inferred to `any` (tsc cannot know about dynamic `this` bindings at compile type).

modules/core/src/v2/wallet.ts
```diff
@@ -105,6 +105,8 @@ export interface GetUserPrvOptions {
   walletPassphrase?: string;
 }
 
+type ManageUnspents = 'consolidate' | 'fanout';
```
tylerlevine: export this type, unless you elect to make manageUnspents private as commented below.

modules/core/src/v2/wallet.ts
```diff
@@ -517,88 +519,87 @@ export class Wallet {
...
+   * @param {Number} params.minValue - the minimum value of unspents to use in satoshis
+   * @param {Number} params.maxValue - the maximum value of unspents to use in satoshis
+   * @param {Number} params.minHeight - the minimum height of unspents on the block chain to use
+   * @param {Number} params.minConfirms - all selected unspents will have at least this many confirmations
+   * @param {Boolean} params.enforceMinConfirmsForChange - if true, minConfirms also applies to change outputs
+   * @param {Number} params.limit                for routeName === 'consolidate'
+   *                 params.maxNumUnspentsToUse  for routeName === 'fanout'
+   *                  - maximum number of unspents you want to use in the transaction
+   * Output parameters:
+   * @param {Number} params.numUnspentsToMake - the number of new unspents to make
```
Xekyo: I believe that this defaults to 1 at least for consolidate. That should perhaps be mentioned here.

modules/core/src/v2/wallet.ts
OttoAllmendinger: will make private

modules/core/src/v2/wallet.ts
OttoAllmendinger: ah I see

`self = this` is a problem because sometimes devs accidentally use `this` anyway (in vanilla js)

modules/core/src/v2/wallet.ts
```diff
@@ -517,90 +519,87 @@ export class Wallet {
...
       const finalTxParams = _.extend({}, signedTransaction, selectParams);
 
-      self.bitgo.setRequestTracer(reqId);
-      return self.bitgo.post(self.baseCoin.url('/wallet/' + self._wallet.id + '/tx/send'))
+      this.bitgo.setRequestTracer(reqId);
+      return this.bitgo.post(this.baseCoin.url('/wallet/' + this._wallet.id + '/tx/send'))
         .send(finalTxParams)
         .result();
-    }).call(this).asCallback(callback);
+    }.bind(this)).call(this).asCallback(callback);
```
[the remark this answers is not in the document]
OttoAllmendinger: good point

but why even `call(this)` then? `call()` is also an option (will use `global` as `this`)

modules/core/src/v2/wallet.ts
```diff
@@ -517,90 +519,87 @@ export class Wallet {
...
+   *                  - maximum number of unspents you want to use in the transaction
+   * Output parameters:
+   * @param {Number} params.numUnspentsToMake - the number of new unspents to make
+   *
    * @param callback
    * @returns txHex {String} the txHex of the incomplete transaction that needs to be signed by the user in the SDK
    */
-  consolidateUnspents(params: any = {}, callback?: NodeCallback<any>): Bluebird<any> {
-    const self = this;
+  manageUnspents(routeName: ManageUnspents, params: any = {}, callback?: NodeCallback<any>): Bluebird<any> {
```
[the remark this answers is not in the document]
OttoAllmendinger: will do

modules/core/src/v2/wallet.ts
OttoAllmendinger: Not sure about that -- all these parameters have defaults that are defined on platform. I think devs should look it up there instead of always repeating it. 

I updated the docs here just for consistency, I actually don't think this is the right place for the full documentation.

modules/core/src/v2/wallet.ts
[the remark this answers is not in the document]
tylerlevine: `call()` would also work, but would break if someone accidentally forgets to use `self`, as you mentioned before.

Longer term, we really just need to move to async/await and drop all the this-rebinding nonsense.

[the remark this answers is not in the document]
OttoAllmendinger: I played around with some AST transforms for that problems, hit me up when you prioritize it

PR #554, opened 2019-10-17
TRON-43: Explain a Tron transaction from txHex

glowang:
* Deserialize a txHex to extract information to display to customer
  before a tx is sent


dannybitgo:
@glowang looks pretty good at first glance, but seems like there are merge conflicts with package.json

README.md
```diff
@@ -58,3 +58,14 @@ manager.
 
 See [DEVELOPERS.md](https://github.com/BitGo/BitGoJS/blob/master/DEVELOPERS.md)
 
+
+# :muscle: Some common tips for developers onboarding to SDK: :muscle:
```
tylerlevine: Nice, this is good info, but I think this would be more appropriate in the DEVELOPERS.md file.

modules/core/package.json
```diff
@@ -47,6 +47,7 @@
     "gen-docs": "typedoc"
   },
   "dependencies": {
+    "@bitgo/tron-lib": "file:../../../bitgo-account-lib",
```
tylerlevine: This needs to point to an actual, published version of tron lib before this PR can be merged.

modules/core/src/v2/coins/trx.ts
```diff
@@ -192,4 +205,47 @@ export class Trx extends BaseCoin {
...
+   * @param params
+   * @param callback
+   */
+  explainTransaction(
+    params: ExplainTransactionOptions,
+    callback?: NodeCallback<TransactionExplanation>
+  ): Bluebird<TransactionExplanation> {
+    return co<TransactionExplanation>(function*() {
+      const txHex = params.txHex || (params.transaction && params.transaction.txHex);
+      if (!txHex || !Tron.safeTronHexString(txHex)) {
```
tylerlevine: Seems a bit silly for this to be in tron lib (I looked at the implementation of this function). This could be replaced by `/^(0x)?([0-9a-f]{2})+$/i.match(txHex)`. Will leave a review on that PR as well.

modules/core/test/v2/unit/coins/trx.ts
```diff
@@ -1,13 +1,36 @@
...
+import Tron from '@bitgo/tron-lib';
 
 const co = Promise.coroutine;
 import { TestBitGo } from '../../../lib/test_bitgo';
 
 describe('TRON:', function() {
   let bitgo;
   let basecoin;
+  // mockTx data taken from testnet while sending a transaction
+  const mockTx = { visible: false,
```
tylerlevine: Consider moving into fixtures directory.

modules/core/test/v2/unit/coins/trx.ts
```diff
@@ -20,6 +43,72 @@ describe('TRON:', function() {
...
+  }));
+
+  it('should throw if txHex does not have the proper format', co(function *() {
+    const explainParams = {
+      fee: 1,
+      txID: mockTx.txID,
+      txHex: 'this_is_an_odd_length_string'
+    };
+    try {
+      yield basecoin.explainTransaction(explainParams);
```
tylerlevine: Could replace whole try/catch with `yield basecoin.explainTransaction(explainParams).should.be.rejectedWith('missing valid txHex string');`

modules/core/test/v2/unit/coins/trx.ts
```diff
@@ -20,6 +43,72 @@ describe('TRON:', function() {
...
+  }));
+
+  it('should throw if the params object is missing parameters', co(function *() {
+    const explainParams = {
+      fee: 1,
+      txID: mockTx.txID,
+      txHex: null,
+    };
+    try {
+      yield basecoin.explainTransaction(explainParams);
```
tylerlevine: Same thing about `yield basecoin.explainTransaction(explainParams).should.be.rejectedWith('missing valid txHex string');`

modules/core/package.json
marktoda: Will be account-lib before merging

modules/core/src/v2/coins/trx.ts
```diff
@@ -39,6 +42,16 @@ export interface TransactionPrebuild {
   };
 }
 
+export interface ExplainTransactionOptions {
+  txHex?: string;
+  transaction?: {
```
marktoda: Half signed and fully signed transactions have different shapes:
```
export interface FullySignedTransaction {
  txHex: string;
}

export interface HalfSignedTransaction {
  halfSigned?: {
    txHex?: string;
    payload?: string;
    txBase64?: string;
  };
}
```

Can this really be both? If so, we can use those base interfaces and say `transaction: HalfSignedTransaction | FullySignedTransaction`

modules/core/src/v2/coins/trx.ts
marktoda: Plus how is that any different than a generic safe hex string? Could use a generic hex sanitizer

modules/core/src/v2/coins/trx.ts
```diff
@@ -192,4 +205,47 @@ export class Trx extends BaseCoin {
...
+  explainTransaction(
+    params: ExplainTransactionOptions,
+    callback?: NodeCallback<TransactionExplanation>
+  ): Bluebird<TransactionExplanation> {
+    return co<TransactionExplanation>(function*() {
+      const txHex = params.txHex || (params.transaction && params.transaction.txHex);
+      if (!txHex || !Tron.safeTronHexString(txHex)) {
+        throw new Error('missing valid txHex string');
+      }
+      if (!params.fee) {
```
marktoda: I think there is a common param validator that we can use to ensure the existence of parameters like this more succinctly

modules/core/src/v2/coins/trx.ts
```diff
@@ -192,4 +205,47 @@ export class Trx extends BaseCoin {
...
+        throw new Error('missing fee');
+      }
+      if (!params.txID) {
+        throw new Error('missing txID');
+      }
+      const rawContract = Tron.getRawTransferContract(txHex);
+      const transfer = Tron.decodeTransferContract(rawContract.contractList[0].parameter.value);
```
marktoda: Change to account lib util decodeTransferContract: https://github.com/BitGo/bitgo-account-lib/pull/5/files#diff-77e5edd96bbcbf039a06a7be97358a2aR149 trx/utils.ts line 149

modules/core/src/v2/coins/trx.ts
```diff
@@ -192,4 +205,47 @@ export class Trx extends BaseCoin {
...
+      };
+      return {
+        displayOrder: ['id', 'outputAmount', 'changeAmount', 'outputs', 'changeOutputs', 'fee', 'memo'],
+        id: params.txID,
+        outputs,
+        outputAmount,
+        changeOutputs: [], // account based does not use change outputs?
+        changeAmount: 0, // account base does not make change
+        fee: params.fee,
+        expiration: rawContract.expiration,
```
marktoda: These are not contained in the base TransactionExplanation -- we can make a TronTransactionExplanation interface that extends the base interface, containing these tron-specific fields

modules/core/src/v2/coins/trx.ts
glowang: It is indeed a generic hex checker. I took it from Platform since we don't have one in SDK AFAIK, but I agree that naming wis misleading. 

What about we take this out of the Tron lib and put it into SDK?

modules/core/src/v2/baseCoin.ts
```diff
@@ -167,6 +167,7 @@ export interface HalfSignedTransaction {
     txHex?: string; // Transaction in any format required by each coin, i.e. in Tron it is a stringifyed JSON
     payload?: string;
     txBase64?: string;
+    rawTransaction?: any;
```
marktoda: Why do we need this? I don't see this field being used anywhere

modules/core/src/v2/coins/trx.ts
marktoda: Since it is a generic account-lib it could be just accountlib.utils.safeHexString(). Though, I personally prefer the `isValidHexString()` style which returns boolean valid/invalid, rather than the `safeHexString()` style which just returns undefined if unsafe

modules/core/src/v2/coins/trx.ts
```diff
@@ -202,11 +220,49 @@ export class Trx extends BaseCoin {
...
+    params: ExplainTransactionOptions,
+    callback?: NodeCallback<TronTransactionExplanation>
+  ): Bluebird<TronTransactionExplanation> {
+    return co<TronTransactionExplanation>(function*() {
+      const txHex = params.txHex || (params.halfSigned && params.halfSigned.txHex);
+      if (!txHex || !bitgoAccountLib.Trx.Utils.isValidHex(txHex)) {
+        throw new Error('missing valid txHex string');
+      }
+      assert(params.fee && params.txID);
+      const rawContract = bitgoAccountLib.Trx.Utils.decodeRawTransaction(txHex);
```
argjv: This is not gonna work. The txHex is a stringfied JSON, like `'{\"some\":\"value\"}'`, you have to call `JSON.parse(txHex)` to get the JSON. Then run `bitgoAccountLib.Trx.Utils.decodeTransaction(tx.raw_data_hex)`

modules/core/src/v2/coins/trx.ts
```diff
@@ -202,11 +220,49 @@ export class Trx extends BaseCoin {
...
+      const transfer = bitgoAccountLib.Trx.Utils.decodeTransferContract(rawContract.contracts[0].parameter.value);
```
argjv: You don't need a method for this. Just do `const transfer = rawContract.contracts[0].parameter.value`

modules/core/src/v2/coins/trx.ts
argjv: This is what the account lib is supposed to abstract from us. I'll add a get field for this.

modules/core/test/v2/unit/coins/trx.ts
```diff
@@ -20,6 +24,62 @@ describe('TRON:', function() {
     basecoin.should.be.an.instanceof(Trx);
   });
 
+  it('explain a txHex', co(function *() {
+    const explainParams = {
+      txHex: mockTx.raw_data_hex,
```
argjv: This should be `txHex: JSON.stringify(mockTx.raw_data_hex)`

modules/core/src/v2/coins/trx.ts
glowang: Note to Gloria:
"{//df}" -> sendTransaction JSON -> use get methods in the library

modules/core/src/v2/coins/trx.ts
glowang: Did that on line 43 :D

modules/core/src/v2/coins/trx.ts
```diff
@@ -35,13 +41,25 @@ export interface TxInfo {
...
   feeInfo: TransactionFee;
 }
 
+export interface ExplainTransactionOptions {
+  txHex?: string; // txHex is poorly named here; it is just a wrapped JSON object
+  halfSigned?: {
+    txHex: string; // txHex is poorly named here; it is just a wrapped JSON object
+  };
+  fee: number;
+}
```
glowang: I deleted the "txID" param that was here before. We no longer need to pass that in since txHex contains it.

.drone.jsonnet
```diff
@@ -168,11 +168,11 @@ local CheckPreconditions(version) = {
 };
 
 local UnitVersions = ["6", "8", "10", "11"];
-local IntegrationVersions = ["lts"];
+local IntegrationVersions = ["10"];
```
twtaylorbitgo: you should rebase, i just merged these fixes

modules/core/src/v2/coins/trx.ts
twtaylorbitgo: good one

modules/core/src/v2/coins/trx.ts
```diff
@@ -202,11 +213,50 @@ export class Trx extends BaseCoin {
...
+    callback?: NodeCallback<TronTransactionExplanation>
+  ): Bluebird<TronTransactionExplanation> {
+    return co<TronTransactionExplanation>(function*() {
+      const txHex = params.txHex || (params.halfSigned && params.halfSigned.txHex);
+      if (!txHex || !params.fee) {
+        throw new Error('missing explain tx parameters');
+      }
+      const coinName = this.getChain();
+      const txBuilder = new TransactionBuilder({ coinName });
+      txBuilder.from(txHex, bitgoAccountLib.BaseCoin.TransactionType.Send);
```
twtaylorbitgo: I think this will be deprecated as of Jorge's PR (transactionType will be removed)

modules/core/test/v2/fixtures/coins/trx.ts
```diff
@@ -0,0 +1,21 @@
...
+              type_url: 'type.googleapis.com/protocol.TransferContract' },
+          type: 'TransferContract' } ],
+      ref_block_bytes: '5123',
+      ref_block_hash: '52a26dea963a47bc',
+      expiration: 1569463320000,
+      timestamp: 1569463261623 },
+  raw_data_hex:
+    '0a025123220852a26dea963a47bc40c0fbb6dad62d5a65080112610a2d747970652e676f6f676c65617069732e636f6d2f70726f746f636f6c2e5472616e73666572436f6e747261637412300a1541e5e00fc1cdb3921b8340c20b2b65b543c84aa1dd1215412c2ba4a9ff6c53207dc5b686bfecf75ea7b80577180a70b7b3b3dad62d',
+  signature:
+    [ ]
```
twtaylorbitgo: note that for a proper mock of what comes out of the build route, there is no signature field

modules/core/src/v2/coins/trx.ts
```diff
@@ -202,11 +213,50 @@ export class Trx extends BaseCoin {
...
+    return co<TronTransactionExplanation>(function*() {
+      const txHex = params.txHex || (params.halfSigned && params.halfSigned.txHex);
+      if (!txHex || !params.feeInfo) {
+        throw new Error('missing explain tx parameters');
+      }
+      const coinName = this.getChain();
+      const txBuilder = new TransactionBuilder({ coinName });
+      txBuilder.from(txHex);
+      const tx = txBuilder.build();
+      //const contract = tx.raw_data.contract;
```
twtaylorbitgo: remove this line please

.drone.yml
```diff
@@ -18,7 +18,7 @@ steps:
 - name: install
   image: node:10
   commands:
-  - "git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true"
+  - git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true
```
twtaylorbitgo: mm we'll want to revert all of these to the base branch

modules/core/package.json
```diff
@@ -48,6 +48,7 @@
   },
   "dependencies": {
     "@bitgo/statics": "^2.3.0-rc",
+    "@bitgo/account-lib": "0.1.0",
```
twtaylorbitgo: nice!

modules/core/src/v2/coins/trx.ts
```diff
@@ -1,16 +1,19 @@
 /**
  * @prettier
  */
+import * as assert from 'assert';
```
twtaylorbitgo: why did we add this?

modules/core/src/v2/coins/trx.ts
```diff
@@ -145,8 +155,9 @@ export class Trx extends BaseCoin {
     if (!_.isEmpty(tx.signature)) {
       expectedSignaturesCount += tx.signature.length;
     }
-    const transaction = tronWeb.utils.crypto.signTransaction(params.prv, tx);
-    if (transaction.signature.length !== expectedSignaturesCount) {
+    //TODO use builder instead of utils to sign tx
```
twtaylorbitgo: we should convert this to the builder once Jorge's change is up

modules/core/package.json
tylerlevine: Please use ^

PR #557, opened 2019-10-18
TOKS-66 - Use base units in Stellar trustlines

CapnMigraine:
Use base units for Stellar token trustline limits in params.

https://bitgoinc.atlassian.net/browse/TOKS-66

Related to platform PR: https://github.com/BitGo/platform/pull/1418


tylerlevine:
I think it will need a rebase to pick up the audit fix in master though

PR #558, opened 2019-10-18
SERV-443 put handlebars^4.3.0 back into yarn.lock

dannybitgo:
https://bitgoinc.atlassian.net/browse/SERV-443

Previous PR removed this for some reason, and James/Perry can't publish because their yarn install puts this back in.


tylerlevine:
This was done in https://github.com/BitGo/BitGoJS/pull/560

PR #564, opened 2019-10-22
TRON-76: Change signTransaction return type

argjv:
[TRON-76](https://bitgoinc.atlassian.net/browse/TRON-76) Fix pending approvals and align with the return types of other coins

### DIFF
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -160,6 +160,7 @@ export interface AddressCoinSpecific {
 
 export interface FullySignedTransaction {
   txHex: string;
+  transaction?: any; // full transaction in JSON format
 }
 
 export interface HalfSignedTransaction {
@@ -167,6 +168,7 @@ export interface HalfSignedTransaction {
     txHex?: string;
     payload?: string;
     txBase64?: string;
+    transaction?: any; // full transaction in JSON format
   };
 }
 
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -159,16 +159,14 @@ export interface AddressCoinSpecific {
 }
 
 export interface FullySignedTransaction {
-  txHex: string;
-  transaction?: any; // full transaction in JSON format
+  txHex: string; // Transaction in any format required by each coin, i.e. in Tron it is a stringifyed JSON
 }
 
 export interface HalfSignedTransaction {
   halfSigned?: {
-    txHex?: string;
+    txHex?: string; // Transaction in any format required by each coin, i.e. in Tron it is a stringifyed JSON
     payload?: string;
     txBase64?: string;
-    transaction?: any; // full transaction in JSON format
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/baseCoin.ts
```diff
@@ -160,13 +160,15 @@ export interface AddressCoinSpecific {
 
 export interface FullySignedTransaction {
   txHex: string;
+  transaction?: any; // full transaction in JSON format
```
tylerlevine: Please use txHex property as we discussed.

modules/core/src/v2/baseCoin.ts
```diff
@@ -160,13 +160,15 @@ export interface AddressCoinSpecific {
...
 }
 
 export interface HalfSignedTransaction {
   halfSigned?: {
     txHex?: string;
     payload?: string;
     txBase64?: string;
+    transaction?: any; // full transaction in JSON format
```
tylerlevine: Please use txHex property as we discussed.

modules/core/src/v2/coins/trx.ts
```diff
@@ -16,6 +17,9 @@ import {
   SignTransactionOptions,
   VerifyAddressOptions,
   VerifyTransactionOptions,
+  TransactionFee,
+  TransactionRecipient as recipient,
```
tylerlevine: Types should begin with a capital letter.

modules/core/src/v2/coins/trx.ts
```diff
@@ -26,17 +30,17 @@ export interface TronSignTransactionOptions extends SignTransactionOptions {
   prv: string;
 }
 
-export interface TransactionPrebuild {
-  txHex: string;
-  txid: string;
+export interface txInfo {
```
tylerlevine: Types should begin with a capital letter

modules/core/src/v2/coins/trx.ts
```diff
@@ -137,13 +141,25 @@ export class Trx extends BaseCoin {
   }
 
   signTransaction(params: TronSignTransactionOptions): SignedTransaction {
-    const tx = params.txPrebuild.transaction;
+    const tx = params.txPrebuild.txInfo.transaction;
     // this prv should be hex-encoded
-    const halfSigned = tronWeb.utils.crypto.signTransaction(params.prv, tx);
-    if (_.isEmpty(halfSigned.signature)) {
+    const transaction = tronWeb.utils.crypto.signTransaction(params.prv, tx);
+    if (_.isEmpty(transaction.signature)) {
```
twtaylorbitgo: it's possible this has a signature on it and is invalid. (so a fully signed tx coming through this flow would have 1 sig). this depends on previous state. we should check if it came in unsigned, does it have 1 sig? and then if it was half, does it have 2?

argjv: I'm not sure how useful this is going to be since I would expect the builder to check for that, but I'll add it.

PR #568, opened 2019-10-25
SERV-460: Add envs for ms domains

tylerlevine:


modules/core/src/v2/environments.ts
```diff
@@ -169,6 +184,18 @@ export const Environments: Environments = {
     uri: 'http://rmglocalhost:3000',
     stellarFederationServerUrl: 'http://rmglocalhost:3000/api/v2/txlm/federation',
   }),
+  msProd: Object.assign({}, mainnetBase, {
+    uri: 'https://app.bitgo.com',
+    stellarFederationServerUrl: 'http://app.bitgo.com/api/v2/xlm/federation',
+  }),
+  msTest: Object.assign({}, testnetBase, {
+    uri: 'https://app.bitgo-test.com',
```
dannybitgo: do we need to add other uri's or nah?

PR #576, opened 2019-10-29
TRON-95: Update Account lib and fix tests

argjv:
[TRON-95](https://bitgoinc.atlassian.net/browse/TRON-95) Account lib output for the `toJSON()` function changed to match the network format, which fixed send transaction flow.

### DIFF
--- a/modules/core/test/v2/unit/coins/trx.ts
+++ b/modules/core/test/v2/unit/coins/trx.ts
@@ -73,20 +73,15 @@ describe('TRON:', function() {
   }));
 
   it('should sign a half signed tx', () => {
-    let tx = basecoin.signTransaction(signTxOptions);
-    let txHex = tx.halfSigned.txHex;
-    txHex = JSON.parse(txHex)
-    txHex.txID.should.equal(signTxOptions.txPrebuild.txInfo.txid);
-    txHex.raw_data.contractType.should.equal(0);
-    const contract = txHex.raw_data.contract;
-    contract.ownerAddress.should.equal(bitgoAccountLib.Trx.Utils.getBase58AddressFromHex(signTxOptions.txPrebuild.txInfo.from))
-    contract.toAddress.should.equal(bitgoAccountLib.Trx.Utils.getBase58AddressFromHex(signTxOptions.txPrebuild.txInfo.recipients[0].address));
-    contract.amount.should.equal(parseInt(signTxOptions.txPrebuild.txInfo.recipients[0].amount,10));
-    txHex.should.have.property('signature');
-    txHex.should.have.property('raw_data_hex');
-    txHex.signature.length.should.equal(1);
-    txHex.raw_data.timestamp.should.equal(mockTx.raw_data.timestamp);
-    txHex.raw_data.expiration.should.equal(mockTx.raw_data.expiration);
+    const tx = basecoin.signTransaction(signTxOptions);
+    const unsignedTxJson = JSON.parse(signTxOptions.txPrebuild.txHex);
+    const signedTxJson = JSON.parse(tx.halfSigned.txHex);
+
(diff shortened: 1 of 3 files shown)


modules/core/test/v2/unit/coins/trx.ts
```diff
@@ -73,20 +73,15 @@ describe('TRON:', function() {
...
-    txHex.raw_data.timestamp.should.equal(mockTx.raw_data.timestamp);
-    txHex.raw_data.expiration.should.equal(mockTx.raw_data.expiration);
+    const tx = basecoin.signTransaction(signTxOptions);
+    const unsignedTxJson = JSON.parse(signTxOptions.txPrebuild.txHex);
+    const signedTxJson = JSON.parse(tx.halfSigned.txHex);
+
+    signedTxJson.txID.should.equal(unsignedTxJson.txID);
+    signedTxJson.raw_data_hex.should.equal(unsignedTxJson.raw_data_hex);
+    (JSON.stringify(signedTxJson.raw_data) === JSON.stringify(unsignedTxJson.raw_data)).should.be.ok;
+    signedTxJson.signature.length.should.equal(1);
```
glowang: are timestamp and expiration also included in this? It would be good to check them just to enforce behaviors.

argjv: They are checked as part of `(JSON.stringify(signedTxJson.raw_data) === JSON.stringify(unsignedTxJson.raw_data)).should.be.ok;`

modules/core/test/v2/unit/coins/trx.ts
```diff
@@ -73,20 +73,15 @@ describe('TRON:', function() {
   }));
 
   it('should sign a half signed tx', () => {
-    let tx = basecoin.signTransaction(signTxOptions);
-    let txHex = tx.halfSigned.txHex;
-    txHex = JSON.parse(txHex)
-    txHex.txID.should.equal(signTxOptions.txPrebuild.txInfo.txid);
-    txHex.raw_data.contractType.should.equal(0);
-    const contract = txHex.raw_data.contract;
-    contract.ownerAddress.should.equal(bitgoAccountLib.Trx.Utils.getBase58AddressFromHex(signTxOptions.txPrebuild.txInfo.from))
```
twtaylorbitgo: why did we remove all these?

argjv: These accessors are tested in another test. Here we care about the JSON format output and verify them with the equalities above.

`(JSON.stringify(signedTxJson.raw_data) === JSON.stringify(unsignedTxJson.raw_data)).should.be.ok` tests that every field in the transaction matches the original one, including the data type.

modules/core/test/v2/unit/coins/trx.ts
```diff
@@ -73,20 +73,15 @@ describe('TRON:', function() {
...
-    txHex.signature.length.should.equal(1);
-    txHex.raw_data.timestamp.should.equal(mockTx.raw_data.timestamp);
-    txHex.raw_data.expiration.should.equal(mockTx.raw_data.expiration);
+    const tx = basecoin.signTransaction(signTxOptions);
+    const unsignedTxJson = JSON.parse(signTxOptions.txPrebuild.txHex);
+    const signedTxJson = JSON.parse(tx.halfSigned.txHex);
+
+    signedTxJson.txID.should.equal(unsignedTxJson.txID);
+    signedTxJson.raw_data_hex.should.equal(unsignedTxJson.raw_data_hex);
+    (JSON.stringify(signedTxJson.raw_data) === JSON.stringify(unsignedTxJson.raw_data)).should.be.ok;
```
tylerlevine: why compare two strings with === instead of `someString.should.eql(otherString)`?

argjv: No particular reason. Updated.

PR #578, opened 2019-10-30
Allow trx to sign with an xprv

argjv:
If an xprv is used to sign a Tron transaction, convert it to the uncompressed version, and then sign.

### DIFF
--- a/modules/core/src/v2/coins/trx.ts
+++ b/modules/core/src/v2/coins/trx.ts
@@ -5,6 +5,7 @@ import * as Bluebird from 'bluebird';
 import { CoinFamily } from '@bitgo/statics';
 const co = Bluebird.coroutine;
 import * as bitgoAccountLib from '@bitgo/account-lib';
+import { HDNode } from 'bitgo-utxo-lib';
 
 import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
 import { MethodNotImplementedError } from '../../errors';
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/trx.ts
```diff
@@ -164,6 +173,20 @@ export class Trx extends BaseCoin {
...
   }
 
+  /**
+   * Return boolean indicating whether input is valid seed for the coin
+   *
+   * @param prv - the prv to be checked
+   */
+  isValidXprv(prv: string): boolean {
+    try {
+      HDNode.fromBase58(prv);
```
twtaylorbitgo: there has to be a better check than this just bombs out if it's not base58

modules/core/test/v2/unit/coins/trx.ts
```diff
@@ -84,6 +84,24 @@ describe('TRON:', function() {
...
+        txHex: signTxOptions.txPrebuild.txHex
+      }
+    };
+    const tx = basecoin.signTransaction(p);
+    const unsignedTxJson = JSON.parse(signTxOptions.txPrebuild.txHex);
+    const signedTxJson = JSON.parse(tx.halfSigned.txHex);
+
+    signedTxJson.txID.should.equal(unsignedTxJson.txID);
+    signedTxJson.raw_data_hex.should.equal(unsignedTxJson.raw_data_hex);
+    JSON.stringify(signedTxJson.raw_data).should.eql(JSON.stringify(unsignedTxJson.raw_data));
```
twtaylorbitgo: hmm i don't think we need this check if we have the check above... should be the same and not mutate the tx

argjv: I think it does make sense. We want to make sure that the structure of the tx doesn't change, but only the signature if it is a different format.
May not be the most important part of the tests for us that we know how it works, but I don't think it is extra.

PR #581, opened 2019-10-30
BLOCK-108 allow unspent reservation param in build

dannybitgo:
https://bitgoinc.atlassian.net/browse/BLOCK-108

We should allow users to take advantage of the unspent reservation system in the build route.

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -79,6 +79,10 @@ export interface PrebuildTransactionOptions {
     addressType?: string;
     hop?: boolean;
     walletPassphrase?: string;
+    reservation?: {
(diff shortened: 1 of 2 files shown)


dannybitgo:
> Side question - how come `reservation` is not an array? Is it possible to only reserve a single unspent at a time?

Well, the idea is that you don't what unspents are going to be used in the build. So you're basically saying "yes, once you build this transaction, reserve all the unspents that were selected, and reserve them until ```expireTime```"

dannybitgo:
If a customer wants to reserve particular unspents, they should use the dedicated endpoint in a separate call:
https://www.bitgo.com/api/v2/#operation/v2.wallet.reservedunspents

PR #587, opened 2019-11-06
BG-14652 change usdx to usdex

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-14652

The official name for USDX has changed to USDEX, so we need to update.


dannybitgo:
do not merge until we coordinate DB migrations

tylerlevine:
please note that this will require a major version bump

tylerlevine:
@dannybitgo is this PR still alive?

PR #591, opened 2019-11-08
DO-554: Update account-lib to 0.1.5 for BitGoJS

perrybitgo:


tylerlevine:
I think this was done in #590

PR #593, opened 2019-11-08
BLOCK-147: Call API route for Federation lookups

CapnMigraine:
The change in https://github.com/BitGo/platform/pull/1508 adds ability to perform Federation lookups through the `v2.federation` platform API
The SDK can now pass the params to look up the account through the BitGo Federation server instead of making the request directly to the external federation server

BLOCK-147

### DIFF
--- a/modules/core/src/errors.ts
+++ b/modules/core/src/errors.ts
@@ -120,7 +120,7 @@ export class KeyRecoveryServiceError extends BitGoJsError {
 export class AddressGenerationError extends BitGoJsError {
   public constructor(message?: string) {
     super(message || 'address generation failed');
-    Object.setPrototypeOf(this, KeyRecoveryServiceError.prototype);
+    Object.setPrototypeOf(this, AddressGenerationError.prototype);
   }
 }
 
@@ -128,5 +128,12 @@ export class EthereumLibraryUnavailableError extends BitGoJsError {
   public constructor(packageName: string) {
     super(`Ethereum library required for operation is not available. Please install "${(packageName)}".`);
     Object.setPrototypeOf(this, EthereumLibraryUnavailableError.prototype);
+  }
+}
+
+export class StellarFederationUserNotFoundError extends BitGoJsError {
+  public constructor(message?: string) {
+    super(message || 'account not found');
+    Object.setPrototypeOf(this, StellarFederationUserNotFoundError.prototype);
   }
 }
--- a/modules/core/src/v2/coins/xlm.ts
+++ b/modules/core/src/v2/coins/xlm.ts
@@ -17,6 +17,7 @@ import {
   InvalidMemoIdError,
   KeyRecoveryServiceError,
   UnexpectedAddressError,
+  StellarFederationUserNotFoundError,
 } from '../../errors';
 import {
   BaseCoin,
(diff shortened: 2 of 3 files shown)


modules/core/src/errors.ts
```diff
@@ -130,3 +130,10 @@ export class EthereumLibraryUnavailableError extends BitGoJsError {
     Object.setPrototypeOf(this, EthereumLibraryUnavailableError.prototype);
   }
 }
+
+export class StellarFederationUserNotFoundError extends BitGoJsError {
```
tylerlevine: Since you're modifying this file anyway, could you fix the incorrect prototype set on line 123? (should be setting to `AddressGenerationError.prototype` instead of `KeyRecoveryServiceError.prototype`).

modules/core/src/v2/coins/xlm.ts
```diff
@@ -459,52 +460,50 @@ export class Xlm extends BaseCoin {
...
     return co<stellar.FederationServer.Record>(function *() {
-      const addressParts = _.split(address, '*');
-      if (addressParts.length !== 2) {
-        throw new InvalidAddressError(`invalid stellar address: ${address}`);
-      }
-      const [, homeDomain] = addressParts;
       try {
-        if (homeDomain === self.homeDomain) {
-          const federationServer = self.getBitGoFederationServer();
+        const federationServer = self.getBitGoFederationServer() as stellar.FederationServer;
```
tylerlevine: Why do we need the `as stellar.FederationServer` cast? Doesn't `getBitGoFederationServer` return this type?

modules/core/src/v2/coins/xlm.ts
```diff
@@ -459,52 +460,50 @@ export class Xlm extends BaseCoin {
...
+        if (address) {
```
tylerlevine: If neither of these are set, is resolving with undefined acceptable behavior? If so, please make return type `Bluebird<stellar.FederationServer.Record | undefined>`.

Or would throwing an Error be more appropriate?

modules/core/src/v2/coins/xlm.ts
```diff
@@ -459,52 +460,50 @@ export class Xlm extends BaseCoin {
...
       }
     }).call(this);
   }
 
+  /**
+   * Attempt to resolve a stellar address into a stellar account
+   *
+   * @param {String} address - stellar address to look for
+   */
+  federationLookupByName(address: string): Bluebird<stellar.FederationServer.Record> {
```
tylerlevine: Should this (and perhaps `federationLookupById`) be made private now that we have `federationLookup`, which can handle both types? Seems like either one or the other should be made private.

tylerlevine:
@CapnMigraine is this PR still alive?

modules/core/src/v2/coins/xlm.ts
CapnMigraine: `federationLookupByName` and `federationLookupById` should be public. I made `federationLookup` private.

CapnMigraine:
> Looks pretty good, just some minor fixes.
> 
> Can we get rid of the CSP rules for stellar federation lookups after this lands?

yes, that's the goal

CapnMigraine:
> @CapnMigraine is this PR still alive?

it is, but it also needs https://github.com/BitGo/platform/pull/1508.

modules/core/src/v2/coins/xlm.ts
```diff
@@ -459,52 +460,63 @@ export class Xlm extends BaseCoin {
...
+    const self = this;
+    if (!address) {
+      throw new Error('invalid Stellar address');
```
tylerlevine: Nit: Move if block inside coroutine so that this function only has a single error channel. (either throw sync Errors or return rejected promises async , but not both. It makes error handling more complex in some cases).

modules/core/src/v2/coins/xlm.ts
```diff
@@ -459,52 +460,63 @@ export class Xlm extends BaseCoin {
...
    * Attempt to resolve an account id into a stellar account
    * Only works for accounts that can be resolved by our federation server
    *
-   * @param accountId - stellar account id
+   * @param {String} accountId - stellar account id
    */
   federationLookupByAccountId(accountId: string): Bluebird<stellar.FederationServer.Record> {
     const self = this;
+    if (!accountId) {
+      throw new Error('invalid Stellar account');
```
tylerlevine: Same nit, move if block inside coroutine.

modules/core/src/v2/coins/xlm.ts

modules/core/src/v2/coins/xlm.ts

PR #598, opened 2019-11-14
[ERC20-157] Add MCDai ERC20 Token

marktoda:
This is an upgraded version of DAI. We are adding it as a separate coin
rather than upgrading the old one because our customers need to manually
upgrade from one to the other and want to be able to swap in their own
time.

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -284,6 +284,7 @@ export const enum UnderlyingAsset {
   LRC = 'lrc',
   MAN = 'man',
   MANA = 'mana',
+  MCDAI = 'mcdai',
   MCO = 'mco',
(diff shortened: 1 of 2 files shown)


tylerlevine:
please correct me if I'm wrong, but I don't believe multicollateral dai (Dai) and single collateral dai (Sai) are interchangable. If this is true, then adding a new coin is really the only option here. Lgtm!

marktoda:
@tylerlevine My understanding is that users can convert from SAI to DAI, but not backwards: https://github.com/makerdao/developerguides/blob/master/mcd/upgrading-to-multi-collateral-dai/upgrading-to-multi-collateral-dai.md#as-a-sai-holder

PR #607, opened 2019-11-21
BG-14806 reject hop param for erc-20 token build

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-14806

Customer was getting a weird error when they tried this and was confused. We should have this error more explicit because we don't support hops for ERC-20 tokens and there is no reason to.

### DIFF
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -22,6 +22,7 @@ import {
   TransactionPrebuild as BaseTransactionPrebuild,
   HalfSignedTransaction as BaseHalfSignedTransaction,
 } from '../baseCoin';
+import { Erc20Token } from './erc20Token';
 import { BitGo } from '../../bitgo';
 import { NodeCallback } from '../types';
 import { Wallet } from '../wallet';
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/eth.ts
```diff
@@ -1268,6 +1269,11 @@ export class Eth extends BaseCoin {
         !_.isUndefined(buildParams.recipients) &&
         !_.isUndefined(buildParams.walletPassphrase)
       ) {
+        if (this instanceof Erc20Token) {
```
marktoda: Isn't there a method like `this.isToken` or something? `instanceof` always feels hacky

dannybitgo: I looked for `this.isToken` but didn't find it in BitGoJS. I believe that's just a platform thing. 
@tylerlevine do you know of a cleaner way of checking if I am a token?

tylerlevine: You could capture the 2nd constructor argument which is the coin from statics, and store as a property on `this`. That would have an `isToken` property you could check.

dannybitgo: Not sure which second constructor argument you're referring to... could you clarify?

tylerlevine: > Not sure which second constructor argument you're referring to... could you clarify?

Sure. When we instantiate a coin, it gets two objects passed to the constructor: the bitgo object and the statics coin (see https://github.com/BitGo/BitGoJS/blob/16427283f54d4183ce2acdda8e908b380fc1f693/modules/core/src/v2/coinFactory.ts#L71)

You're only capturing the first arg, but the second one has the info you need

dannybitgo: instantiating the coin is not really part of this PR, so maybe we leave that for another time and stick with `instanceof` here

PR #613, opened 2019-12-02
SERV-579 filter out duplicate addresses in cross-chain

dannybitgo:
https://bitgoinc.atlassian.net/browse/SERV-579

If a wrong-chain transaction has two outputs to the same wrong-chain address, then the WRW fails. This fixes


dannybitgo:
> looks good, are the test changes related? Is there a test for this case?

The test changes are related. I edited one of the current tests so that it tests the new functionality

PR #618, opened 2019-12-06
BG-14844 Fix accelerateTx circular json bug

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-14844

For some reason a wallet object is getting added to the params before we do submitTransaction, and this is causing a failure as it shouldn't be there. This just deletes that wallet param and fixes the issue.

To test the issue and the fix, use the script linked in the comments below

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -1813,6 +1813,7 @@ export class Wallet {
 
       // We must pass the build params through to submit in case the CPFP tx ever has to be rebuilt.
       const submitParams = Object.assign(params, yield self.prebuildAndSignTransaction(params));
+      delete submitParams.wallet;
       return yield self.submitTransaction(submitParams);
     }).call(this).asCallback(callback);
   }


dannybitgo:
To test, use the testing script below. Once on master to see the error, then once on this branch to see the fix.

```
const BitGoJS = require('bitgo');
const bitgo = new BitGoJS.BitGo({ env: 'test' });
const blue = require('bluebird');

const accessToken = 'your-access-token';

const walletId = 'your-wallet-id';
const walletPassphrase = 'your-wallet-password';
const address = '2MzW9rAYW1m63xKevBg1dJa1QzjtAYU3Ac8'; // the tax address has officially been hijacked by Danny 

const coin = 'tbtc';

const amount = 0.001 * 1e8;
const feeRate = 1000;

function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

blue.coroutine(function *() {
  bitgo.authenticateWithAccessToken({ accessToken });

  const wallet = yield bitgo.coin(coin).wallets().get({ id: walletId });

  const params = {
    recipients: [{
      address,
      amount
    }],
    walletPassphrase,
    feeRate
  };

  const tx = yield wallet.sendMany(params);
  if (tx.txid) {
    console.log(`Successfully sent low-fee tx: ${tx.txid}`);
    console.log(`Now pausing to let platform register the tx`);
    yield sleep(5000);
    console.log(`Attempting to accelerate tx...`);
    const accelerateParams = {
      cpfpTxIds: [tx.txid],
      walletPassphrase,
      cpfpFeeRate: 30000,
      maxFee: 30000,
    };
    const result = yield wallet.accelerateTransaction(accelerateParams);
    console.dir(result);
  }
})();
```

PR #619, opened 2019-12-07
Properly deserialize EOS staking transactions

paddyquinn:
JIRA: ETSTK-11

This commit adds functionality to deserialize EOS staking
transactions. EOS staking transactions consist of two actions,
delegatebw and voteproducer. The delegatebw action specifies the
amount of EOS to be staked and the account to stake it for. In our
case the staker and the receiver of the stake will always be the
same. The voteproducer action specifies the proxy or producers that
the staked funds will vote for.

### DIFF
--- a/modules/core/src/v2/coins/eos.ts
+++ b/modules/core/src/v2/coins/eos.ts
@@ -97,6 +97,19 @@ interface DeserializedEosTransaction extends EosTransactionHeaders {
   amount: string;
   transaction_id: string;
   memo?: string;
+  proxy?: string;
+  producers?: string[];
+}
+
+interface DeserializedStakeAction {
+  address: string;
+  amount: string;
+}
+
+interface DeserializedVoteAction {
+  address: string;
+  proxy: string;
+  producers: string[];
 }
 
 interface ExplainTransactionOptions {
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/eos.ts
```diff
@@ -421,25 +468,80 @@ export class Eos extends BaseCoin {
...
+        if (tx.actions.length === 2) {
+          const txAction2 = tx.actions[1];
+          if (txAction2.name !== 'delegatebw') {
+            throw new Error(`invalid staking transaction action: ${txAction2.name}, expecting: delegatebw`);
+          }
+
+          deserializedStakeAction = self.deserializeStakeAction(eosClient, txAction2.data);
+        }
+
+        const deserializedVoteAction = Eos.deserializeVoteAction(eosClient, txAction.data);
```
tylerlevine: Nit: How come `Eos` instead of `self`?

modules/core/src/v2/coins/eos.ts
```diff
@@ -421,25 +468,80 @@ export class Eos extends BaseCoin {
...
+          );
+        }
+
+        const txAction2 = tx.actions[1];
+        if (txAction2.name !== 'voteproducer') {
+          throw new Error(`invalid staking transaction action: ${txAction2.name}, expecting: voteproducer`);
+        }
+
+        const deserializedStakeAction = self.deserializeStakeAction(eosClient, txAction.data);
+        const deserializedVoteAction = Eos.deserializeVoteAction(eosClient, txAction2.data);
```
tylerlevine: Nit: How come `Eos` instead of `self`?

modules/core/src/v2/coins/eos.ts
```diff
@@ -421,25 +468,80 @@ export class Eos extends BaseCoin {
...
+        }
+
+        const deserializedVoteAction = Eos.deserializeVoteAction(eosClient, txAction.data);
+        if (!!deserializedStakeAction && deserializedStakeAction.address !== deserializedVoteAction.address) {
+          throw new Error(
+            `staker (${deserializedStakeAction.address}) and voter (${deserializedVoteAction.address}) must be the same`
+          );
+        }
+
+        tx.amount = deserializedStakeAction.amount;
```
tylerlevine: Potential issue: If `tx.actions.length` is 1 and `txAction.name` is 'voteproducer', won't this access a property of undefined?

modules/core/src/v2/coins/eos.ts
paddyquinn: This is `Eos` instead of `self` because `deserializeVoteAction` is a static method. I made `deserializeVoteAction` a static method frankly because my IDE was giving me a warning that it could be made static. I can change it such that it is not static and so this would use `self` if you think that it is better for consistency.

modules/core/src/v2/coins/eos.ts
paddyquinn: See above.

modules/core/src/v2/coins/eos.ts
tylerlevine: Ah, I missed that it was static, that's fine.

modules/core/src/v2/coins/eos.ts
paddyquinn: Great catch! Thank you. Fixed.

paddyquinn:
@tylerlevine Don't have examples for all of the different types of staking transactions yet. Created a ticket to add unit tests when we do: https://bitgoinc.atlassian.net/browse/ETSTK-16

tylerlevine:
Please rebase onto latest master for test fix.

PR #622, opened 2019-12-09
AGS-18 / Add by referral code and specify trading partner type

quazeenii:
add trading partner by referral code added to SDK
update trading partner responses to new trading partner response object
add trading partner type ('direct' or 'agency') to response
add requester side ('primary' or 'secondary') and type ('direct', 'agency') to request parameters for referral form and referral code

### DIFF
--- a/modules/core/src/v2/trading/tradingPartners.ts
+++ b/modules/core/src/v2/trading/tradingPartners.ts
@@ -10,15 +10,31 @@ import { BitGo } from '../../bitgo';
 
 import { NodeCallback } from '../types';
 import { TradingAccount } from './tradingAccount';
-import { TradingPartner } from './tradingPartner';
+import { TradingPartner, TradingPartnerType } from './tradingPartner';
 
 const co = Bluebird.coroutine;
 
-interface TradingPartnerReferralParameters {
+// Side of the requester (if they should be considered the primary or the secondary)
+// Only important for agency partnerships
+// the primaryAccount is the agent, settling trades for the secondary account id
+export enum TradingReferralRequesterSide {
+  PRIMARY = 'primary', // if partnership is of type agency, primary is the agent
+  SECONDARY = 'secondary',
+}
+
+export interface TradingPartnerAddByCodeParameters {
+  referralCode: string;
+  type: TradingPartnerType;
+  requesterSide: TradingReferralRequesterSide;
+}
+
+export interface TradingPartnerReferralParameters {
   institutionName: string;
   contactName: string;
   contactEmail: string;
   contactPhoneNumber: string;
+  type: TradingPartnerType;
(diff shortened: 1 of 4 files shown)


modules/core/src/v2/trading/tradingPartners.ts
```diff
@@ -10,15 +10,31 @@ import { BitGo } from '../../bitgo';
...
 
+// Side of the requester (if they should be considered the primary or the secondary)
+// Only important for agency partnerships
+// the primaryAccount is the agent, settling trades for the secondary account id
+export enum TradingReferralRequesterSide {
+  PRIMARY = 'primary', // if partnership is of type agency, primary is the agent
+  SECONDARY = 'secondary',
+}
+
+interface TradingPartnerReferByCodeParameters {
```
tylerlevine: These interfaces should be exported, since they are used in an exported type.

modules/core/src/v2/trading/tradingPartners.ts
```diff
@@ -80,4 +96,31 @@ export class TradingPartners {
...
+   * Refer a new trading partner to join the BitGo network and become your partner.
+   * @param params
+   * @param params.institutionName name of the institution to invite
+   * @param params.contactName full name of a member of the institution to contact
+   * @param params.contactEmail email address of the contact
+   * @param params.contactPhoneNumber phone number of the contact
+   * @param params.memo memo to send to the trading partner when sending the invite
+   * @param callback
+   */
+  referByCode(params: TradingPartnerReferByCodeParameters, callback?: NodeCallback<{}>): Bluebird<{}> {
```
tingwai: Can you add a return type? Same for the method `refer` above

modules/core/test/v2/unit/trading/tradingPartner.ts
```diff
@@ -3,7 +3,8 @@ import * as nock from 'nock';
 import * as should from 'should';
 
 import fixtures from '../../fixtures/trading/tradingPartner';
-import { TradingPartnerStatus } from '../../../../src/v2/trading/tradingPartner';
+import { TradingPartnerStatus, TradingPartnerType } from '../../../../src/v2/trading/tradingPartner';
+import { TradingReferralRequesterSide } from "../../../../src/v2/trading/tradingPartners";
```
tingwai: Interesting there isn't a lint rule for single/double quotes
```suggestion
import { TradingReferralRequesterSide } from '../../../../src/v2/trading/tradingPartners';
```

modules/core/src/v2/trading/tradingPartners.ts
quazeenii: as discussed, fixed for referByCode but not for refer

PR #623, opened 2019-12-10
Update bitgo-utxo-lib to use new Zcash params

argjv:
# Overview
[BLOCK-225](https://bitgoinc.atlassian.net/browse/BLOCK-226) Updating bitgo utxo lib version to the latest one which supports new Blossom network params.

### DIFF
--- a/modules/core/package.json
+++ b/modules/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bitgo",
-  "version": "9.1.0",
+  "version": "9.2.0-rc.0",
   "description": "BitGo Javascript SDK",
   "main": "./dist/src/index.js",
   "types": "./dist/types/src/index.d.ts",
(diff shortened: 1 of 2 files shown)


modules/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bitgo",
-  "version": "9.1.0",
+  "version": "9.2.0-rc.0",
```
james8nguyen: Apparently, the SDK publishing process is quite different 
1) We do not manually update the SDK version and leave it to the "yarn publish" to update during a publish. Please revert this change.

2) Manual changes to yarn.lock are not required as it will get updated automatically by the "yarn update" or "yarn publish" process.

With that said, we should have this PR against the release branch, rel/naples, instead of master.

tylerlevine: > Manual changes to yarn.lock are not required as it will get updated automatically by the "yarn update" or "yarn publish" process.

Actually, if you are making changes to the package.json, please update the yarn.lock file. The package-lock.json in modules/express will  be taken care of for you, do not modify that one.

tylerlevine:
I think this conflicts with #624, which one do we want to go with? They both have minor problems.

argjv:
It does. Closing this in favor of #624

PR #624, opened 2019-12-10
BLOCK-226 Update bitgo-utxo-lib to use new Zcash params.

james8nguyen:
JIRA Ticket: [BLOCK-226](https://bitgoinc.atlassian.net/browse/BLOCK-226)
Updated bitgo-utxo-lib version to 1.7.0-rc (then 1.7.0 later) which supports new Blossom network params.

### DIFF
--- a/modules/core/package.json
+++ b/modules/core/package.json
@@ -57,7 +57,7 @@
     "bigi": "^1.4.0",
     "bignumber.js": "^8.0.1",
     "bitcoinjs-message": "^2.0.0",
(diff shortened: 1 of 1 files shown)


dannybitgo:
tests failing

tylerlevine:
Also this  should have a yarn.lock update with  it I think.

tylerlevine:
I think  this conflicts with  #623, which one do we want to go with. They both have minor problems.

PR #625, opened 2019-12-10
CT-489: Check for wrapped segwit unspents in express v1 integration test

tylerlevine:


dannybitgo:
what commit caused these to start failing?

tylerlevine:
What happened was at some point the change addresses for the test wallet started defaulting to p2sh-p2wsh, so we slowly started converting all our p2sh unspents to p2sh-p2wsh. The tests were written before we supported p2sh-p2wsh, and thus didn't check for those unspents. When it ran  out of p2sh unspents, it thought the wallet was out of unspents, when really it had plenty of p2sh-p2wsh unspents available.

PR #633, opened 2019-12-16
[BLOCK-236] Remove unsupported ERC20 tokens

marktoda:
# Change Overview

There are several ERC20 tokens in statics that are not supported by the
rest of the BitGo platform. In an effort to integrate statics with other
components, we need to remove those extraneous tokens, otherwise we will
accidentally support them.

Tokens removed:
ABT, AGI, AUTO, BLZ, BNB, BST, BTM, CENNZ, CMT, CND, CS, DCN, DEW, DRGN,
DROP, DTR, ENG, ERC, FSN, GNX, GVT, HPB, ICN, ICX, IOST, KCS, MAN, MANA,
NCASH, NULS, OST, PAYX, PLR, POE, PPP, PRL, R, REQ, RHOC, RUFF, SAN,
SMT, SRN, SUB, THETA, TMS, VEE, VEN, VERI, XIN

# How to review
For each token removed, check to verify that it is not supported. One simple way to do so is to check our api docs here: https://www.bitgo.com/api/v2/

# Breaking?
This is an open question to the reviewer -- how badly of a breaking change is this? It won't break any BitGo integrations as none of these tokens are supported by BitGo platform anyway. However, if anyone is using statics for something unrelated to BitGo, it could break that.


tylerlevine:
> how badly of a breaking change is this?

Definitely a major version bump for statics, but it shouldn't affect bitgo components very much.

marktoda:
Removed underlying assets

marktoda:
Thanks @tylerlevine . Merging this, let's remember to do a major version bump next time we update statics, cc @james8nguyen @perrybitgo

PR #634, opened 2019-12-17
Return algo keyreg data from explainTransaction

paddyquinn:
This commit returns type, senderAddress, voteKeyBase64, and
voteLastBlock from explainTransaction. The OVC needs this data to verify
what it is signing.

JIRA: AIR-6


tylerlevine:
approved landing to rel/la branch with failing drone build.

It appears we have at least two problems which make this necessary:
1) yarnpkg.com is having network timeout issues which is causing delay in merging this urgent fix
2) There are some test fixes which are not backported to rel/la, and backporting these to get a green build is of dubious usefulness
3) Dependency upgrades would also need to be backported

PR #639, opened 2020-01-07
SERV-255: Update release notes for versions 8.1.2 through 9.3.0

tylerlevine:


tylerlevine:
Next step: conventional-changelog + commitlint

PR #641, opened 2020-01-10
Fix utxo constants

OttoAllmendinger:
TBH I'm not sure why we even redeclare these constants in statics if they are already declared in bitgo-utxo-lib

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -164,7 +164,6 @@ class DashTestnet extends BitcoinLikeTestnet {
   messagePrefix = '\x19DarkCoin Signed Message:\n';
   pubKeyHash = 0x8c;
   scriptHash = 0x13;
-  wif = 0xef;
   family = CoinFamily.DASH;
   explorerUrl = 'https://tbch.blockdozer.com/tx/';
 }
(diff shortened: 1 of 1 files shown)


tylerlevine:
> TBH I'm not sure why we even redeclare these constants in statics if they are already declared in bitgo-utxo-lib

The reason is that I thought some users of `@bitgo/statics` might just be interested in the value of the constant,  and not the rest of the functionality provided by `bitgo-utxo-lib`. `bitgo-utxo-lib` can get the values from `@bitgo/statics` easily, but the  other way around is not possible (`@bitgo/statics` is dependency free).

The plan was to eventually move all the network constants into `@bitgo/statics`, and have `bitgo-utxo-lib` use the values from statics.

OttoAllmendinger:
> The reason is that I thought some users of @bitgo/statics might just be interested in the value of the constant, and not the rest of the functionality provided by bitgo-utxo-lib.

I don't think the values are useful without bitgo-utxo-lib. You need at least a base58 encode/decoder to do anything with them.

tylerlevine:
> > The reason is that I thought some users of @bitgo/statics might just be interested in the value of the constant, and not the rest of the functionality provided by bitgo-utxo-lib.
> 
> I don't think the values are useful without bitgo-utxo-lib. You need at least a base58 encode/decoder to do anything with them.

That's a good point.

OttoAllmendinger:
How about we replace the network constants with `networkName: string` that references the network name in `bitgo-utxo-lib`? https://github.com/BitGo/bitgo-utxo-lib/blob/c894c71de58c0ab2fdb9dd6bca1ef8ee152548e8/src/networks.js#L45

The format of the constants is very coupled to that lib and I don't think it makes sense to add `statics` as a dependency there.

PR #643, opened 2020-01-15
fix(test): remove illegal use of `bufferutils`

OttoAllmendinger:
The method `bufferutils.reverse` was removed a long time ago:
https://github.com/BitGo/bitgo-utxo-lib/commit/29a865788d30b8b776cc1a1a2fd042d70085ec5f#diff-73e64645f9c04dc17e67b782cb9342fd

It is unclear why this did not cause an error

Issue: BG-16466


OttoAllmendinger:
gotcha

PR #646, opened 2020-01-17
Added redeemScript and witnessScript to SignTransactionOptions

DCRichards:
## Issue

`redeemScript` and `witnessScript` are not documented as being part of the `signTransaction` params, but signing will fail if these are needed for signing the tx.

## Changes

- [X] Added `redeemScript` and `witnessScript` to `SignTransactionOptions`.

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -150,6 +150,8 @@ export interface SignTransactionOptions {
         index?: number;
         value?: number;
         address?: string;
+        redeemScript?: string;
+        witnessScript?: string;
       }[];
     }
   };


DCRichards:
Looks like the integration tests failed, but I don't _think_ this was related to my change. Log is as follows:

```shell
bitgo:     Non-BitGo Token Recovery
--
256 | @bitgo/express:     1) "before all" hook for "can do sendcoins"
257 | @bitgo/express:   39 passing (56s)
258 | @bitgo/express:   5 pending
259 | @bitgo/express:   1 failing
260 | @bitgo/express:   1) Express ETH
261 | @bitgo/express:        "before all" hook for "can do sendcoins":
262 | @bitgo/express:      Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/drone/src/modules/express/test/integration/eth.ts)
```

tylerlevine:
Looks like the build failure is indeed unrelated - I restarted it. Once it's green, I'll resign your commit and merge it. Thanks for the fix!

tylerlevine:
This will get merged shortly in #652. Thanks!

PR #654, opened 2020-01-28
Add `allowPartialSweep` option to sweep

Xekyo:
Express filters body parameters before calling platform, so we have to
whitelist the `allowPartialSweep` parameter for it to be available.

Ticket: BG-17594

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -199,6 +199,7 @@ export interface SweepOptions {
   feeRate?: number;
   maxFeeRate?: number;
   feeTxConfirmTarget?: number;
+  allowPartialSweep?: boolean;
 }
 
 export interface FreezeOptions {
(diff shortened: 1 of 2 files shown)


Xekyo:
I find it extremely surprising that my code would increase the testcoverage by 18%. Do you have an idea what's going on here, @tylerlevine?

tylerlevine:
> I find it extremely surprising that my code would increase the testcoverage by 18%. Do you have an idea what's going on here, @tylerlevine?

Yes, test coverage is not working correctly, please ignore it for now.

dannybitgo:
This will need a corresponding FrontEnd change to work in the UI right?

Xekyo:
@dannybitgo: I don't see how this has anything to do with the frontend. Does the frontend use the `sweep` route anywhere ever?
The frontend is just locked to sending the maximum amount on every transaction in BSV.

dannybitgo:
> @dannybitgo: I don't see how this has anything to do with the frontend. Does the frontend use the `sweep` route anywhere ever?
> The frontend is just locked to sending the maximum amount on every transaction in BSV.

currently if you attempt to send the exact "available balance" from the frontend, the fronted will use the Sweep functioon instead of Build. 

I think its ok though, because available balance should factor in the 200-input max right?

Xekyo:
That would mean that the frontend would have been failing to send the maximum available amount if the wallet has more than 200 unspents. It was probably not noticed because frontend wallets often have fewer unspents.

In that case, I would recommend that the frontend call uses `allowPartialSweep: true`.

PR #657, opened 2020-01-29
fix(express): Deprecate older forms of environment variable config

tylerlevine:
* Emit warning upon usage of a deprecated environment variable
* Update README to clearly explain the preferred environment
variables, and which ones are deprecated but still usable for backwards
compatibility reasons

JIRA: BG-17569

### DIFF
--- a/modules/express/src/config.ts
+++ b/modules/express/src/config.ts
@@ -3,6 +3,19 @@ import { isNil, isNumber } from 'lodash';
 
 import { args } from './args';
 
+function readEnvVar(name, ...deprecatedAliases): string | undefined {
+  if (process.env[name] !== undefined) {
+    return process.env[name];
+  }
+
+  for (const deprecatedAlias of deprecatedAliases) {
+    if (process.env[deprecatedAlias] !== undefined) {
+      console.warn(`warning: using deprecated environment variable '${deprecatedAlias}'. Please use the '${name}' environment variable instead.`);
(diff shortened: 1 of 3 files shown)


modules/express/src/config.ts
```diff
@@ -3,6 +3,19 @@ import { isNil, isNumber } from 'lodash';
 
 import { args } from './args';
 
+function readEnvVar(name, ...deprecatedAliases): string | undefined {
+  if (process.env[name] !== undefined) {
+    return process.env[name];
```
dannybitgo: won't this always return?
when do we ever call `readEnvVar(name)` such that `process.env[name]` in `undefined` ?

tylerlevine: If a user upgrades but does not change the environment variables they are setting, the one referred to by `name` might not be there (since that was not a valid option when they did their initial setup of the env vars). In that case `process.env[name]` would be undefined, and we would try the deprecated aliases for the variable instead. If they were using an older form of the env var, it should get picked up in the loop.

To make it more concrete, imagine a user who has been using express behind a reverse proxy for a while, and therefore sets `export DISABLESSL=true`. When they update, the new version will look at `BITGO_DISABLE_SSL` instead, and will find it to be undefined. In order to keep this user working, we'd need to look at the older forms as well. That's what the 2nd loop does.

dannybitgo: oh i see
makes sense!

PR #659, opened 2020-02-04
Rel/queens

perrybitgo:


tylerlevine:
Needs test fix from #661 to fix build before merge

PR #663, opened 2020-02-10
Add signing policy support to pending approvals

tmpope:
Add the ability to get the payload from a pending approval in state
pendingCryptographicApproval, and to upload a signed payload to approve.
Currently does not verify the validity of the payload before returning
or verify signatures before uploading.

Ticket: BG-17078

Currently we do not want to merge this, as we don't want the existence of this feature to be public, and it is also not yet complete enough that we want public eyes on it. This branch will serve as a develop branch for this feature until we get to that point.

### DIFF
--- a/modules/core/src/v2/pendingApproval.ts
+++ b/modules/core/src/v2/pendingApproval.ts
@@ -20,6 +20,14 @@ export interface PendingApprovalInfo {
     buildParams: any;
     sourceWallet?: string;
   };
+  signingPolicyUpdate?: {
+    //TODO Add this to allow updates via SDK
+  }
+}
+
+export interface SigningPolicyInfo {
+  payload: string;
+  gatheredSignatures: string[];
 }
 
 export interface PendingApprovalData {
@@ -30,6 +38,7 @@ export interface PendingApprovalData {
   creator: string;
   info: PendingApprovalInfo;
   approvalsRequired?: number;
+  signingPolicy?: SigningPolicyInfo;
 }
 
 export const enum OwnerType {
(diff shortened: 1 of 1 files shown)


tmpope:
Despite the do not merge label, review feedback is welcome. Don't go out of your way to provide it, though, this is not high-urgency

modules/core/src/v2/pendingApproval.ts
```diff
@@ -303,6 +318,37 @@ export class PendingApproval {
       .asCallback(callback);
   }
 
+  getCryptographicApprovalPayload(pendingApproval: PendingApprovalData, callback?: NodeCallback<any>): Bluebird<any> {
+    const self = this;
+    return co(function*() {
+      // this function is currently overkill for a getter, but we should in
```
tylerlevine: Good thinking, but you could simplify this to
```
return Bluebird.resolve(pendingApproval.signingPolicy.payload).asCallback(callback);
```
and get rid of the rest of the boilerplate until it's needed.

modules/core/src/v2/pendingApproval.ts
```diff
@@ -303,6 +318,37 @@ export class PendingApproval {
...
+      const approvalParams: any = { 
+        state: 'approved',
+        signatures: [params.signature]
+        otp: params.otp,
+      };
+      return self.bitgo
+        .put(self.url())
+        .send(approvalParams)
+        .result()
+        .nodeify(callback);
```
tylerlevine: Remove the `nodeify` call here, I think this would result in the callback being called twice (since it's also passed to `asCallback` before).

PR #668, opened 2020-02-13
Adding build and send consolidation functions to BitGoJS.

twtaylorbitgo:
This adds two new pieces of functionality to a wallet. Building and sending of a consolidation transaction. This feature is only meant for Algorand for the moment.

Ticket: BG-17572


modules/core/example/js/account-consolidation.js
```diff
@@ -0,0 +1,53 @@
...
+  const unlock = await testBg.unlock({ otp, duration: 3600 });
+  if (!unlock) {
+    console.log('We did not unlock.');
+    throw new Error();
+  }
+
+  // this will take all money off receive addresses in the wallet
+  // you can also specify which receive address by passing fromAddresses here:
+  // e.g. { walletPassphrase, fromAddresses: ['onchainReceiveAddress'] }
+  await wallet.sendAccountConsolidations({ walletPassphrase })
```
tylerlevine: Nit: Use try/catch, we don't need `.then()`, `.catch()`, or `.error()` style promise chaining any more. Also, other js examples use coroutine style, we should stick to that until we can update all examples to use async/await.

modules/core/src/v2/wallet.ts
```diff
@@ -2249,4 +2270,112 @@ export class Wallet {
...
+  /**
+   * Builds and sends a set of account consolidations. This is intended to flush many balances to the root wallet balance.
+   * @param params
+   * @param callback
+   */
+  sendAccountConsolidations(params: BuildConsolidationTransactionOptions = {}, callback?: NodeCallback<any>): Bluebird<any> {
+    const self = this;
+
+    if (!this.baseCoin.allowsAccountConsolidations()) {
+      throw new Error('This coin does not allow account consolidations.');
```
tylerlevine: Move into coroutine - this will throw sync, when what we want is to return a rejected promise.

modules/core/src/v2/wallet.ts
```diff
@@ -2249,4 +2270,112 @@ export class Wallet {
...
+    }
+
+    return co<any>(function *() {
+      // this gives us a set of account consolidation transactions
+      const unsignedBuilds = yield self.buildAccountConsolidations(params);
+      if (unsignedBuilds && unsignedBuilds.length > 0) {
+        const successfulTxes = new Array<any>();
```
tylerlevine: Not wrong, but `new Array<any>` is a bit odd, why not `const successfulTxs: any[] = [];`

modules/core/example/js/account-consolidation.js
twtaylorbitgo: ah yes. I've changed this now.

modules/core/src/v2/wallet.ts
twtaylorbitgo: ah right

modules/core/src/v2/wallet.ts
twtaylorbitgo: I can change this. I just like explicitly calling the constructor for any new object, it's more of a style thing

modules/core/test/v2/fixtures/coins/algo.ts
```diff
@@ -65,9 +115,9 @@ module.exports.prebuild = function() {
...
...2, 112, 107, 196, 32, 92, 254, 60, 121, 55, 231, 149, 166, 80, 75, 74, 78, 74, 98, 202, 168, 189, 178, 220, 138, 19, 149, 73, 60, 213, 152, 204, 125, 140, 33, 86, 171, 129, 162, 112, 107, 196, 32, 217, 138, 204, 201, 225, 190, 239, 5, 235, 72, 8, 40, 93, 82, 5, 6, 68, 226, 246, 5, 45, 99, 21, 144, 229, 121, 202, 182, 245, 184, 253, 160, 163, 116, 104, 114, 2, 161, 118, 1, 163, 116, 120, 110, 137, 163, 97, 109, 116, 205, 3, 232, 163, 102, 101, 101, 206, 0, 3, 177, 80, 162, 102, 118, 205, 3, 232, 163, 103, 101, 110, 173, 116, 101, 115, 116, 110, 101, 116, 45, 118, 51, 56, 46, 48, 162, 103, 104, 196, 32, 224, 121, 14, 64, 66, 246, 163, 102, 213, 135, 99, 245, 193, 42, 100, 203, 123, 58, 115, 7, 4, 131, 240, 0, 119, 154, 165, 122, 188, 189, 227, 104, 162, 108, 118, 205, 7, 207, 163, 114, 99, 118, 196, 32, 187, 14, 182, 52, 21, 74, 24, 11, 106, 39, 77, 215, 117, 41, 92, 54, 211, 186, 122, 174, 107, 93, 176, 204, 16, 197, 70, 45, 176, 243, 48, 223, 163, 115, 110, 100, 196, 32, 5, 164, 47, 197, 123, 139, 63, 163, 202, 150, 236, 41, 203, 206, 251, 3, 143, 143, 192, 123, 178, 233, 145, 104, 64, 61, 159, 224, 205, 29, 135, 208, 164, 116, 121, 112, 101, 163, 112, 97, 121]).toString('base64'),
```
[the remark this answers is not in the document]
twtaylorbitgo: ack

modules/core/src/v2/wallet.ts
```diff
@@ -51,6 +51,13 @@ export interface Memo {
     type: string;
 }
 
+/**
+ * We only let a small subset of these params get through in the route.
```
cooncesean: nit: This doesn't really inform other developers what this interface does or is to be used for.

- why only a small subset of params?
- what are they?
- when are they likely to be passed?

modules/core/src/v2/wallet.ts
```diff
@@ -86,12 +93,13 @@ export interface PrebuildTransactionOptions {
...
 
 export interface PrebuildAndSignTransactionOptions extends PrebuildTransactionOptions {
-    prebuildTx?: string;
+    prebuildTx?: string | PrebuildTransactionResult;
     verification?: VerificationOptions;
 }
 
 export interface PrebuildTransactionResult extends TransactionPrebuild {
     walletId: string;
+    consolidateId?: string;
```
cooncesean: Since this is going in `wallet.js` -- one of the core modules in this library, you may want to leave an inline comment describing what a `consolidateId` is -- when it is used,etc. 

I can see someone looking at this code in 9 months and not completely understanding the nuance of this attr.

modules/core/src/v2/wallet.ts
```diff
@@ -2249,4 +2270,119 @@ export class Wallet {
...
+   * Builds a set of consolidation transactions for a wallet.
+   * @param params
+   *     fromAddresses - these are the on-chain receive addresses we want to pick a consolidation amount from
+   * @param callback
+   */
+  buildAccountConsolidations(params: BuildConsolidationTransactionOptions = {}, callback?: NodeCallback<PrebuildTransactionResult[]>): Bluebird<PrebuildTransactionResult[]> {
+    const self = this;
+    return co<PrebuildTransactionResult[]>(function *() {
+      if (!self.baseCoin.allowsAccountConsolidations()) {
+        throw new Error('This coin does not allow account consolidations.');
```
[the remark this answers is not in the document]
twtaylorbitgo: ok added some more clarity around what coin you're in

modules/core/example/js/account-consolidation.js
```diff
@@ -0,0 +1,52 @@
...
+// https://test.bitgo.com/enterprise/XXXXXXXXX/coin/talgo/YYYYY/transactions
+// YYYYY would be your wallet id in this case minus the last 8 characters
+const walletId = 'your wallet id';
+const user = 'your.user@login.com';
+const password = 'your password';
+// this is your wallet passphrase, which could be different than your login credentials
+const walletPassphrase = password;
+const otp = '000000';
+
+testBg.authenticate({ username: user, password: password, otp: otp }, async function(err, result) {
```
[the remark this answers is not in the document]
twtaylorbitgo: Ok I've changed this so it aligns with the other examples - added access token, removed the user login

Added a typescript example as well

modules/core/src/v2/wallet.ts
twtaylorbitgo: ah yeah that's fair. I can document more clearly what build params can be used here.

modules/core/src/v2/wallet.ts
twtaylorbitgo: that's fair, I've added some color

PR #675, opened 2020-02-21
Update account-lib version to 1.0.0

CapnMigraine:
Update account-lib dependency in core module
Update yarn.lock
Update Express' package-lock.json
Wrap coins' signTransaction methods in coroutines
Yield on (TRON) calls to account-lib's transactionBuilder.build() method

Ticket: BG-18323

### DIFF
--- a/modules/core/package.json
+++ b/modules/core/package.json
@@ -47,7 +47,7 @@
     "gen-docs": "typedoc"
   },
   "dependencies": {
-    "@bitgo/account-lib": "1.0.0",
+    "@bitgo/account-lib": "^1.0.0",
     "@bitgo/statics": "^3.4.1",
     "@bitgo/unspents": "^0.6.0",
     "@types/bluebird": "^3.5.25",
--- a/modules/core/package.json
+++ b/modules/core/package.json
@@ -47,7 +47,7 @@
     "gen-docs": "typedoc"
   },
   "dependencies": {
-    "@bitgo/account-lib": "^0.1.5",
+    "@bitgo/account-lib": "1.0.0",
     "@bitgo/statics": "^3.4.1",
     "@bitgo/unspents": "^0.6.0",
     "@types/bluebird": "^3.5.25",
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -18,7 +18,7 @@ import {
   VerifyRecoveryTransactionOptions,
   VerifyTransactionOptions,
   TransactionParams as BaseTransactionParams,
-  TransactionPrebuild as BaseTransactionPrebuild, VerificationOptions, TransactionRecipient,
+  TransactionPrebuild as BaseTransactionPrebuild, VerificationOptions, TransactionRecipient, SignedTransaction,
 } from '../baseCoin';
 import { Keychain, KeyIndices } from '../keychains';
 import { NodeCallback } from '../types';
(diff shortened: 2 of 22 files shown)


CapnMigraine:
> Change looks ok, but I think there is something a bit wonky with the yarn.lock that's causing test failures. Might be time to regenerate it entirely.
> 
> Do we have any release notes for account lib so we know what's changing?

I have updated the version of account-lib to 1.0.0-alpha.1, in preparation for the Tezos implementation

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -944,145 +944,152 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
    * @param params.isLastSignature True if `TransactionBuilder.build()` should be called and not `TransactionBuilder.buildIncomplete()`
-   * @returns {{txHex}}
+   * @param callback
+   * @returns {Bluebird<SignedTransaction>}
    */
-  signTransaction(params: SignTransactionOptions): { txHex: string } {
-    const txPrebuild = params.txPrebuild;
-    const userPrv = params.prv;
+  signTransaction(params: SignTransactionOptions, callback?: NodeCallback<SignedTransaction>): Bluebird<SignedTransaction> {
+    const self = this;
```
argjv: Is this some kind of Javascript magic?

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -944,145 +944,152 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
    * @param params.prv private key to be used for signing
    * @param params.isLastSignature True if `TransactionBuilder.build()` should be called and not `TransactionBuilder.buildIncomplete()`
-   * @returns {{txHex}}
+   * @param callback
+   * @returns {Bluebird<SignedTransaction>}
    */
-  signTransaction(params: SignTransactionOptions): { txHex: string } {
-    const txPrebuild = params.txPrebuild;
-    const userPrv = params.prv;
+  signTransaction(params: SignTransactionOptions, callback?: NodeCallback<SignedTransaction>): Bluebird<SignedTransaction> {
```
tylerlevine: How come we need to return a Promise now? I don't see any yields in here. Is it already the case that some implementation of `signTransaction` is async (and thus the type has been wrong the whole time)? Or is an async `signTransaction` being introduced in this change?

modules/core/src/v2/coins/ofc.ts
```diff
@@ -102,7 +101,7 @@ export class Ofc extends BaseCoin {
     return Bluebird.resolve(true).asCallback(callback);
   }
 
-  signTransaction(params: SignTransactionOptions): SignedTransaction {
+  signTransaction(params: SignTransactionOptions): Bluebird<SignedTransaction> {
```
tylerlevine: This should take a callback param for consistency.

modules/core/src/v2/coins/ofc.ts
```diff
@@ -102,7 +101,7 @@ export class Ofc extends BaseCoin {
...
     throw new MethodNotImplementedError();
```
tylerlevine: `return Bluebird.reject(new MethodNotImplementedError()).asCallback(callback);`

We should not throw sync errors in async methods, instead we should reject asynchronously.

modules/core/src/v2/coins/susd.ts
```diff
@@ -75,7 +75,7 @@ export class Susd extends BaseCoin {
     return Bluebird.resolve(true).asCallback(callback);
   }
 
-  signTransaction(params: SignTransactionOptions = {}): SignedTransaction {
+  signTransaction(params: SignTransactionOptions = {}): Bluebird<SignedTransaction> {
     throw new MethodNotImplementedError();
```
tylerlevine: Take a callback and return rejected promise here too.

modules/core/src/v2/coins/trx.ts
```diff
@@ -202,30 +202,39 @@ export class Trx extends BaseCoin {
...
+  signTransaction(
+    params: TronSignTransactionOptions,
+    callback?: NodeCallback<SignedTransaction>
+  ): Bluebird<SignedTransaction> {
+    const self = this;
+    return co<SignedTransaction>(function*() {
+      const txBuilder = bitgoAccountLib.getBuilder(self.getChain());
+      txBuilder.from(params.txPrebuild.txHex);
+      txBuilder.sign({ key: params.prv });
+      const transaction = yield txBuilder.build();
```
tylerlevine: Ah, I see it now. All usages of the account-lib `txBuilder.build()` must be async. Is there some deeper reason for this change?

modules/core/src/v2/coins/trx.ts
```diff
@@ -512,19 +521,19 @@ export class Trx extends BaseCoin {
...
       const txHex = params.txHex || (params.halfSigned && params.halfSigned.txHex);
       if (!txHex || !params.feeInfo) {
         throw new Error('missing explain tx parameters');
       }
-      const coinName = this.getChain();
-      const txBuilder = new bitgoAccountLib.TransactionBuilder({ coinName });
+      const txBuilder = bitgoAccountLib.getBuilder(self.getChain());
       txBuilder.from(txHex);
-      const tx = txBuilder.build();
+      const tx = yield txBuilder.build() as Promise<bitgoAccountLib.BaseCoin.BaseTransaction>;
```
tylerlevine: How come the cast is needed here?

CapnMigraine: It's not, removed.

modules/core/package.json
```diff
@@ -47,7 +47,7 @@
     "gen-docs": "typedoc"
   },
   "dependencies": {
-    "@bitgo/account-lib": "^0.1.5",
+    "@bitgo/account-lib": "1.0.0-alpha.1",
```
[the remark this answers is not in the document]
tylerlevine: That's fine with me - seems James has already released 1.0.0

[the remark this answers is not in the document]
CapnMigraine: Done, I changed it to `1.0.0`

modules/core/package.json
```diff
@@ -47,7 +47,7 @@
     "gen-docs": "typedoc"
   },
   "dependencies": {
-    "@bitgo/account-lib": "^0.1.5",
+    "@bitgo/account-lib": "1.0.0",
```
tylerlevine: ^1.0.0 please

CapnMigraine: fixed

Issue #677, opened 2020-02-21
Extend EOS Expiration for Non-BitGo Recoveries


marktoda:
EOS Transactions are built with a 1 hour expiration time by default, as the network generally does not propagate transactions whose expiration is more than 1 hour from the present time. However, when performing a recovery, the offline signing process often takes several hours. This issue is to change the non-BitGo recovery transaction building [logic](https://github.com/BitGo/BitGoJS/blob/29cc698bb3dc3700499cba3d7f69e88f88d0125f/modules/core/src/v2/coins/eos.ts#L823) to use a longer expiration time of 8 hours. You can see [here](https://github.com/BitGo/BitGoJS/blob/29cc698bb3dc3700499cba3d7f69e88f88d0125f/modules/core/src/v2/coins/eos.ts#L762) that the expiration date is always set to 1 hour from the present time. 

## Current Behavior

Transactions expire 1 hour after being built. This is not enough time to perform a multi-party recovery.

## Possible Solution

Extend the expiration farther, to 8 hours beyond the present time. 

## Steps to Reproduce

1. Create a TEOS wallet with Coincover as backup key provider
2. Fund TEOS wallet
3. Use [WRW](https://github.com/bitgo/wallet-recovery-wizard) and do Non-BitGo recovery on the wallet
4. Send half-signed recovery transaction to Coincover
5. They sign and send back
6. Try to broadcast

tylerlevine:
This is being tracked as BG-19622

PR #682, opened 2020-02-29
Extend EOS recovery expiration from 1 to 8 hours

bitgopher:
Resolves https://github.com/BitGo/BitGoJS/issues/677 by extending EOS recovery expiration from 1 to 8 hours.

Adds a unit test to demonstrate that the correct expiration date is encoded, by comparing it to the head block time (which is mocked [here](https://github.com/BitGo/BitGoJS/blob/ade4e9ab1459055bf6805e8ca551e98fb44d9a45/modules/core/test/v2/lib/recovery-nocks.ts#L2594)). Also updates the expected packed transaction used across EOS recovery unit tests. (Note: only the first seven bytes are different from the prior packed transaction, which presumably is where the expiration date is encoded.)

### DIFF
--- a/modules/core/test/v2/unit/recovery.ts
+++ b/modules/core/test/v2/unit/recovery.ts
@@ -9,6 +9,7 @@ import * as nock from 'nock';
 import { TestBitGo } from '../../lib/test_bitgo';
 const recoveryNocks = require('../lib/recovery-nocks');
 import * as config from '../../../src/config';
+import moment = require('moment');
 
 nock.disableNetConnect();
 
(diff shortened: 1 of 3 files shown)


bitgopher:
Note - CI is currently failing on an unrelated jiggler (same error as this PR: https://github.com/BitGo/BitGoJS/pull/675).

tylerlevine:
I have created a ticket for this work - could you also add `Ticket: BG-18622` at the bottom of your commit message? Thanks!

modules/core/test/v2/unit/recovery.ts
```diff
@@ -615,14 +616,16 @@ describe('Recovery:', function() {
 
   describe('Recover EOS', function() {
     let baseCoin;
+    const expectedPackedTrx = '7122315d5e91d408e1b3000000000100a6823403ea3055000000572d3ccdcd0150f3ea2e4cf4bc8300000000a8ed32322150f3ea2e4cf4bc83e0f27c27cc0adf7f40420f000000000004454f53000000000000';
+    const expectedTxId = '99c6a4eedf5cff246314bdc0a053c12d75488df3aa09474bad4ceca88d8b2498';
```
cooncesean: 👍

bitgopher:
thanks @cooncesean! Just force pushed to sign some commits I had missed.

PR #683, opened 2020-03-02
Replace BitcoinAverage with CoinGecko

### LINKED ISSUE #673: Replace BitcoinAverage API url for wallet recoveries
# Overview
BitcoinAverage no longer supports free API access. When you try to do a wallet recovery with a Key-Recovery-Service provider, the SDK currently fails because of the broken BitcoinAverage link (which gets price data). The goal of this challenge is to replace bitcoinaverage with a new provider: CoinGecko

# Definition of Done
Replace all usages of BitcoinAverage with CoinGecko. This will involve some global searches across the repo, and some investigation into the [CoinGecko API ](https://www.coingecko.com/api/documentations/v3#/). BitcoinAverage gets used [here](https://github.com/BitGo/BitGoJS/blob/b3a14c1d52f6bc26d41d2c119b1bb5e03aa49d2d/modules/core/src/v2/coins/abstractUtxoCoin.ts#L1521). And the relevant line in the config file is [here](https://github.com/BitGo/BitGoJS/blob/b3a14c1d52f6bc26d41d2c119b1bb5e03aa49d2d/modules/core/src/config.ts#L173)
Also please update [this test file](https://github.com/BitGo/BitGoJS/blob/b3a14c1d52f6bc26d41d2c119b1bb5e03aa49d2d/modules/core/test/v2/lib/recovery-nocks.ts) properly (used for mocking in the test suite)

antonina-cherednichenko:
Fixes #673

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1518,14 +1518,33 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   getRecoveryMarketPrice(): Bluebird<string> {
     const self = this;
     return co<string>(function *getRecoveryMarketPrice() {
-      const bitcoinAverageUrl = config.bitcoinAverageBaseUrl + self.getFamily().toUpperCase() + 'USD';
-      const response = yield request.get(bitcoinAverageUrl).retry(2).result();
+      const familyNamesToCoinGeckoIds = new Map()
+        .set('BTC', 'bitcoin')
+        .set('LTC', 'litecoin')
+        .set('BCH', 'bitcoin-cash')
+        .set('ZEC', 'zcash')
+        .set('DASH', 'dash');
+      const coinGeckoId = familyNamesToCoinGeckoIds.get(self.getFamily().toUpperCase());
+      if (!coinGeckoId) {
+        throw new Error(`There is no CoinGecko id for family name ${self.getFamily().toUpperCase()}.`);
+      }
+      const coinGeckoUrl = config.coinGeckoBaseUrl + `simple/price?ids=${coinGeckoId}&vs_currencies=USD`;
+      const response = yield request.get(coinGeckoUrl).retry(2).result();
 
-      if (response === null || typeof response.last !== 'number') {
-        throw new Error('unable to reach BitcoinAverage for price data');
+      // An example of response
+      // {
+      //   "ethereum": {
+      //     "usd": 220.64
+      //   }
+      // }
+      if (!response) {
+        throw new Error('Unable to reach Coin Gecko API for price data');
+      }
+      if (!response[coinGeckoId]['usd'] || typeof response[coinGeckoId]['usd'] !== 'number') {
+        throw new Error('Unexpected response from Coin Gecko API for price data');
       }
 
-      return response.last;
+      return response[coinGeckoId]['usd'];
     }).call(this);
   }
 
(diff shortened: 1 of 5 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1515,17 +1515,31 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   /**
    * Get the current market price from a third party to be used for recovery
    */
-  getRecoveryMarketPrice(): Bluebird<string> {
+   getRecoveryMarketPrice(): Bluebird<string> {
```
tylerlevine: Indent is one space off here

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1515,17 +1515,31 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
     const self = this;
     return co<string>(function *getRecoveryMarketPrice() {
-      const bitcoinAverageUrl = config.bitcoinAverageBaseUrl + self.getFamily().toUpperCase() + 'USD';
-      const response = yield request.get(bitcoinAverageUrl).retry(2).result();
-
-      if (response === null || typeof response.last !== 'number') {
-        throw new Error('unable to reach BitcoinAverage for price data');
+      // TODO This map should be moved to the place where the list of families is defined
```
tylerlevine: good thinking, but for simplicity's sake I'm fine with leaving the map close to the place it's being used.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1515,17 +1515,31 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+      const familyNamesToCoinGeckoIds = new Map()
+        .set('BTC', 'bitcoin')
+        .set('LTC', 'litecoin')
+        .set('BCH', 'bitcoin-cash')
+        .set('ZEC', 'zcash')
+        .set('DASH', 'dash');
+      const coinGeckoId = familyNamesToCoinGeckoIds.get(self.getFamily().toUpperCase());
+      const coinGeckoUrl = config.coinGeckoBaseUrl + `simple/price?ids=${coinGeckoId}&vs_currencies=USD`;
```
tylerlevine: What happens if we don't find an entry in the map for the coin family?

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1515,17 +1515,31 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+      const response = yield request.get(coinGeckoUrl).retry(2).result();
+
+      // An example of response
+      // {
+      //   "ethereum": {
+      //     "usd": 220.64
+      //   }
+      // }
+      if (!response || typeof response[coinGeckoId]['usd'] !== 'number') {
```
tylerlevine: Should both these cases be treated the same?

modules/core/src/v2/coins/abstractUtxoCoin.ts
antonina-cherednichenko: I added a check for that + throwing an error if a family is unknown.

modules/core/src/v2/coins/abstractUtxoCoin.ts
antonina-cherednichenko: There are 2 separate checks now, originally I tried to keep logic the same as was written before.

antonina-cherednichenko:
@tylerlevine Thanks for the review, I added changes.

tylerlevine:
I've signed your commit created PR #693 to get this merged. You'll remain as the author of the commit, and I'll close this PR now. Thanks for the fix!

antonina-cherednichenko:
> Looking good.
> 
> @antonina-cherednichenko you're going to need to sign these commits (something we require being a crypto + fintech company focused on security).
> 
> Some notes to help with this:
> 
> * https://help.github.com/en/github/authenticating-to-github/about-commit-signature-verification
> * https://help.github.com/en/github/authenticating-to-github/signing-commits

@cooncesean sorry for the late reply and thank you for the info! I installed gpg locally to use for future commits.

Issue #667, opened 2020-02-12
docker build  -  invalid bin entry for package msgpack-lite


suppix:
Hello. 

I'm trying to build docker image using these commands:
```
$ git clone -b rel/latest https://github.com/bitgo/bitgojs
$ cd bitgojs
$ docker build -t bitgo-express:latest .
```

But I get:
```
> @bitgo/express@9.6.0 build /tmp/bitgo/modules/express
> tsc

added 842 packages in 37.485s
npm ERR! invalid bin entry for package msgpack-lite@git+https://github.com/algorand/msgpack-lite.git#549d9211c47863da5ad1b69276e363617c9a5c78. key=msgpack, value=bin/msgpack

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/node/.npm/_logs/2020-02-12T08_38_32_147Z-debug.log
The command '/bin/sh -c npm ci && npm prune --production' returned a non-zero code: 1
```

sabodash:
It's problem with [npm version](https://github.com/npm/cli/issues/613) in base image. My temporary solution - update npm to latest version during image build.
```
diff --git a/Dockerfile b/Dockerfile
index 027ad81f..914eb1a2 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM node:10 AS builder
 MAINTAINER Tyler Levine <tyler@bitgo.com>
 COPY --chown=node:node . /tmp/bitgo/
+RUN npm install npm@latest -g
 WORKDIR /tmp/bitgo/modules/express
 USER node
 RUN npm ci && npm prune --production
```

PR #700, opened 2020-03-16
Fix docker build "invalid bin entry for package"

jnaulty:
The following error was causing the `docker build -t bitgo-express .` command to fail
`npm ERR! invalid bin entry for package msgpack-lite`

@sabodash proposed the fix related to the npm package upstream issue:
https://github.com/npm/cli/issues/613

Fixes: #667


tylerlevine:
Going to close and recreate, since we don't run integration tests on cross repo PRs. Thanks for the fix @jnaulty

PR #702, opened 2020-03-16
BG-17966 Add XTZ support

argjv:
Update account-lib to 1.0.2 and add Tezos controller.


modules/core/package.json
```diff
@@ -47,7 +47,7 @@
     "gen-docs": "typedoc"
   },
   "dependencies": {
-    "@bitgo/account-lib": "^1.0.0",
+    "@bitgo/account-lib": "https://github.com/BitGo/bitgo-account-lib#9bcafd9f213077de434ca79d2e489880a153b3eb",
```
argjv: Waiting for new `account-lib` release

modules/core/src/v2/coins/xtz.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  /**
+   * Sign message with private key
+   *
+   * @param key
+   * @param message
+   */
+  signMessage(key: KeyPair, message: string | Buffer): Buffer {
+    throw new Error('Method not implemented');
```
tylerlevine: Use existing error type: https://github.com/BitGo/BitGoJS/blob/master/modules/core/src/errors.ts#L101

modules/core/src/v2/coins/xtz.ts
```diff
@@ -0,0 +1,243 @@
...
+   * Builds a funds recovery transaction without BitGo.
+   * We need to do three queries during this:
+   * 1) Node query - how much money is in the account
+   * 2) Build transaction - build our transaction for the amount
+   * 3) Send signed build - send our signed build to a public node
+   * @param params
+   * @param callback
+   */
+  recover(params: any, callback?: NodeCallback<any>): Bluebird<any> {
+    throw new Error('Method not implemented');
```
tylerlevine: use existing error type

modules/core/src/v2/coins/txtz.ts
```diff
@@ -0,0 +1,18 @@
+/**
```
tylerlevine: Add prettier annotation

PR #712, opened 2020-03-25
Add `getFeeAddress()` to Enterprises

tylerlevine:
We've had `getFeeAddressBalance()` on Enterprises for some time now, but
there was never a way to get the actual fee address itself, aside from
manually requesting the enterprise.

This adds a new function to Enterprises for getting the enterprise fee
address.

Ticket: BG-19068


twtaylorbitgo:
This might need to be pluralized - our eth enterprises have two fee addresses

PR #715, opened 2020-04-01
Add Regtest ERC20 Token

marktoda:
In order to run e2e tests on ERC20 tokens, we needed to add one to our
regtest deployments. We added a token which  will be initialized
with the first transaction from the hard-coded
rich address on the regtest chain. This means the token will always
have the same address: 0x731a10897d267e19b34503ad902d0a29173ba4b1.

Ticket: BG-19512

### DIFF
--- a/modules/core/test/unit/bitgo.ts
+++ b/modules/core/test/unit/bitgo.ts
@@ -461,9 +461,9 @@ describe('BitGo Prototype Methods', function() {
       bitgo.initializeTestVars();
       const tokens = bitgo.getConstants().eth.tokens;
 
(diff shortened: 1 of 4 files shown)


marktoda:
Calculation of the contract address:

```
const ethUtil = require('ethereumjs-util');

const creatorAddress = '0x00a329c0648769a73afac7f9381e08fb43dbea72';
const nonce = 0;


const address = ethUtil.bufferToHex(ethUtil.generateAddress(creatorAddress, nonce));

console.log(`address: ${address}`);
```

marktoda:
changed to `terc20` per tyler request

PR #717, opened 2020-04-02
AGS-1 agency feature

quazeenii:
agency was released but the sdk changes were never merged to prod.

This should update all endpoints to match what's expected on the API side

### DIFF
--- a/modules/core/src/v2/trading/tradingAccount.ts
+++ b/modules/core/src/v2/trading/tradingAccount.ts
@@ -1,6 +1,5 @@
 /**
  * @prettier
- * @hidden
  */
 
 /**
--- a/modules/core/src/v2/trading/tradingAccount.ts
+++ b/modules/core/src/v2/trading/tradingAccount.ts
@@ -18,26 +18,32 @@ import { Settlements } from './settlements';
 
 const co = Bluebird.coroutine;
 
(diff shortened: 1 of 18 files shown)


modules/core/src/v2/trading/tradingAccount.ts
```diff
@@ -113,41 +118,47 @@ export class TradingAccount {
...
-    // the amount field will change if fees are present, but subtotal should always equal the requested send amount
-    let expectedAmount: string = params.amount;
-    if (payloadObj.fees) {
-      const totalFees = payloadObj.fees.reduce(
-        (fees: BigNumber, feeObj) => fees.plus(feeObj.feeAmount),
-        new BigNumber(0)
-      );
-      expectedAmount = new BigNumber(payloadObj.subtotal).plus(totalFees).toString();
+      // matching party found, and fee found
+      validAmounts = validAmounts + 1;
```
tylerlevine, line 153: It's a bit unsafe to check for the bad case, and if you don't find it, assume the case is valid. It's better to assume the case is unsafe, and only increment `validAmounts` in the inner most if statement where you know all your conditions have been checked.

I don't think it's an issue here now, but if the logic for determining a valid match ever changes, it could pose a problem.

PR #723, opened 2020-04-09
Send XTZ signatures as buffer

argjv:
Signature verification was failing in the platform because the signature was not a buffer.

TICKET: BG-19801


yarn.lock
```diff
@@ -120,11 +120,6 @@
     protobufjs "^6.8.9"
     tronweb "^2.7.2"
 
-"@bitgo/statics@^3.4.2":
-  version "3.4.4"
-  resolved "https://registry.yarnpkg.com/@bitgo/statics/-/statics-3.4.4.tgz#acc581980d30fa93e38460e2f9ea72beb60cfa91"
-  integrity sha512-rEXKec+K6ZuRCVQju/fH6/+7SOhvVtehyJZavEeytB0y5Ah9urL1lVXaXZGuKwPQaGqo4KGOjmOjGfxwkLRAMw==
-
```
CapnMigraine, line 127: is this an intended change? it doesn't seem to be related

argjv: Good catch. It is probably from the `yarn install` I run. I'll leave it to the BRE team to update it.

PR #727, opened 2020-04-27
BitGoJsError: include stack trace

OttoAllmendinger:
https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt

> The optional constructorOpt argument accepts a function. If given,
> all frames above constructorOpt, including constructorOpt, will be
> omitted from the generated stack trace.

> The constructorOpt argument
> is useful for hiding implementation details of error generation from
> an end user.

Issue: BG-20357

### DIFF
--- a/modules/core/test/v2/unit/errors.ts
+++ b/modules/core/test/v2/unit/errors.ts
@@ -4,17 +4,23 @@ import { BitGoJsError } from '../../../src/errors';
 
 describe('Error handling', () => {
 
-  it('should construct custom errors if Error.captureStackTrace is missing', () => {
+  it('should construct custom errors if Error.captureStackTrace is missing', function () {
     const captureStub = sinon.stub(Error, 'captureStackTrace').value(undefined);
     new BitGoJsError();
     captureStub.callCount.should.equal(0);
     captureStub.restore();
   });
 
-  it('should construct custom errors with Error.captureStackTrace if present', () => {
+  it('should construct custom errors with Error.captureStackTrace if present', function () {
     const captureStub = sinon.stub(Error, 'captureStackTrace').returns();
     new BitGoJsError();
     captureStub.callCount.should.equal(1);
     captureStub.restore();
+  });
+
+  it('should capture stack trace', function namedFunc() {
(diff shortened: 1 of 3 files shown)


modules/core/test/v2/unit/errors.ts
```diff
@@ -4,17 +4,23 @@ import { BitGoJsError } from '../../../src/errors';
 
 describe('Error handling', () => {
 
-  it('should construct custom errors if Error.captureStackTrace is missing', () => {
+  it('should construct custom errors if Error.captureStackTrace is missing', function () {
```
tylerlevine, line 7: Why change these tests to use `function` keyword?

modules/core/test/v2/unit/errors.ts
```diff
@@ -4,17 +4,23 @@ import { BitGoJsError } from '../../../src/errors';
...
     new BitGoJsError();
     captureStub.callCount.should.equal(1);
     captureStub.restore();
   });
+
+  it('should capture stack trace', function namedFunc() {
+    const bitGoJsError = new BitGoJsError();
+    bitGoJsError.stack!.should.match(/at new BitGoJsError/);
+    bitGoJsError.stack!.should.match(/at Context\.namedFunc/);
+  })
```
tylerlevine, line 25: Missing semicolon here.

modules/core/test/v2/unit/errors.ts
OttoAllmendinger: arrow functions are generally discouraged in mocha https://mochajs.org/#arrow-functions

modules/core/test/v2/unit/errors.ts
OttoAllmendinger: fixed

modules/core/test/v2/unit/errors.ts
tylerlevine: I am aware of that, but I'm not referring to the test using `this` (or at all), so it shouldn't matter.

OttoAllmendinger: but you might in the next PR ;)

PR #731, opened 2020-04-30
Add Celo, RSK, and ETC configs

argjv:
Add Celo and RSK configs into statics since they are being developed in `account-lib` already.

There is a token named CGLD already, which is the Celo's underlying asset name.


argjv:
I don't mind. I think whatever is in `coinMarketCap` is good. However, `cgld` has already been taken by an ETH token which generates conflicts in the `UnderlyingAsset` enum. Thoughts? @marktoda @tylerlevine

marktoda:
it's not on coinmarketcap since it's not launched mainnet yet.

barathcj:
I would prefer cgld as thats how its going to be listed elswhere. I looked into current usage of cgld token and its being held by bitnuk (100). Checking Jaoquin if we can deprecate the old one and just use cgld for celo. Will keep you posted. For now lets continue with what we have

modules/statics/src/coins.ts
```diff
@@ -25,10 +25,16 @@ export const coins = CoinMap.fromCoins([
...
+  account('cgld', 'Celo Gold', Networks.main.celo, 18, UnderlyingAsset.CGLD, ETH_FEATURES),
+  account('tcgld', 'Testnet Celo Gold', Networks.test.celo, 18, UnderlyingAsset.CGLD, ETH_FEATURES),
   account('eth', 'Ethereum', Networks.main.ethereum, 18, UnderlyingAsset.ETH, ETH_FEATURES),
   account('teth', 'Testnet Ethereum', Networks.test.kovan, 18, UnderlyingAsset.ETH, ETH_FEATURES),
+  account('etc', 'Ethereum Classic', Networks.main.ethereum, 18, UnderlyingAsset.ETC, ETH_FEATURES),
+  account('tetc', 'Testnet Ethereum Classic', Networks.test.kovan, 18, UnderlyingAsset.ETC, ETH_FEATURES),
   account('eos', 'Eos', Networks.main.eos, 4, UnderlyingAsset.EOS),
   account('teos', 'Testnet Eos', Networks.test.eos, 4, UnderlyingAsset.EOS),
+  account('rbtc', 'Rsk', Networks.main.rbtc, 18, UnderlyingAsset.RBTC, ETH_FEATURES),
+  account('trbtc', 'Testnet Rsk', Networks.test.rbtc, 18, UnderlyingAsset.RBTC, ETH_FEATURES),
```
marktoda, line 37: Should the long name be "RSK Smart Bitcoin"?

modules/statics/src/coins.ts
```diff
@@ -25,10 +25,16 @@ export const coins = CoinMap.fromCoins([
...
   account('talgo', 'Testnet Algorand', Networks.test.algorand, 6, UnderlyingAsset.ALGO),
+  account('cgld', 'Celo Gold', Networks.main.celo, 18, UnderlyingAsset.CGLD, ETH_FEATURES),
+  account('tcgld', 'Testnet Celo Gold', Networks.test.celo, 18, UnderlyingAsset.CGLD, ETH_FEATURES),
   account('eth', 'Ethereum', Networks.main.ethereum, 18, UnderlyingAsset.ETH, ETH_FEATURES),
   account('teth', 'Testnet Ethereum', Networks.test.kovan, 18, UnderlyingAsset.ETH, ETH_FEATURES),
+  account('etc', 'Ethereum Classic', Networks.main.ethereum, 18, UnderlyingAsset.ETC, ETH_FEATURES),
+  account('tetc', 'Testnet Ethereum Classic', Networks.test.kovan, 18, UnderlyingAsset.ETC, ETH_FEATURES),
   account('eos', 'Eos', Networks.main.eos, 4, UnderlyingAsset.EOS),
   account('teos', 'Testnet Eos', Networks.test.eos, 4, UnderlyingAsset.EOS),
+  account('rbtc', 'Rsk', Networks.main.rbtc, 18, UnderlyingAsset.RBTC, ETH_FEATURES),
```
tylerlevine, line 36: I was expecting this to be RSK, how come we are going with RBTC instead?

modules/statics/src/networks.ts
```diff
@@ -226,6 +249,18 @@ class OfcTestnet extends Testnet implements OfcNetwork {
   explorerUrl = undefined;
 }
 
+class Rbtc extends Mainnet implements AccountNetwork {
```
tylerlevine, line 252: Is it appropriate to call it Rootstock and RootstockTestnet, respectively? This would be more aligned with the other network names.

modules/statics/src/coins.ts
marktoda: I think RBTC vs RSK is the same question that I posed above with CELO vs CGLD. I'm curious your thoughts

tylerlevine:
> I would prefer cgld as thats how its going to be listed elswhere. I looked into current usage of cgld ---. Checking Jaoquin if we can deprecate the old one and just use cgld for celo. Will keep you posted. For now lets continue with what we have

It's a bit annoying to deprecate and reuse an existing name - we need to cut a new major version of statics - but it's not a blocker.

Also @barathcj (and I can't really believe I need to say this) this is a public repo, so please don't discuss specific customers, which coins they are holding, and how much.

modules/statics/src/coins.ts
```diff
@@ -25,10 +25,16 @@ export const coins = CoinMap.fromCoins([
   utxo('tzec', 'Testnet ZCash', Networks.test.zCash, UnderlyingAsset.ZEC),
   account('algo', 'Algorand', Networks.main.algorand, 6, UnderlyingAsset.ALGO),
   account('talgo', 'Testnet Algorand', Networks.test.algorand, 6, UnderlyingAsset.ALGO),
+  account('cgld', 'Celo Gold', Networks.main.celo, 18, UnderlyingAsset.CGLD, ETH_FEATURES),
```
tylerlevine, line 28: There is already coin with this name - this is causing the test failures.

modules/statics/src/coins.ts
argjv: Those are the official tickers and how they will be promoted in places like coinmarketcap. Since we support assets, I think it makes sense to use the asset ticker.

modules/statics/src/coins.ts
argjv: We are deprecating CGLD token to use it for Celo. We have a deprecation plan in place (I can add you to it for review if you like)

marktoda:
pushed to restart the test suite

package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "devDependencies": {
     "husky": "^3.0.0",
-    "lerna": "^3.13.4"
+    "lerna": "^3.20.2"
```
twtaylorbitgo, line 6: why did the lerna version change?

argjv: `npm audit` was complaining

modules/statics/src/networks.ts
argjv: I rather not use specific names for the networks since they can change (tTRX recently did).
Even Tezos mainnet name changes per protocol. I'm not sure about this one, but I think it is safer to just use the coin name.

twtaylorbitgo:
yup, approved

PR #743, opened 2020-05-14
Add basic class structure for ETC, RBTC and CGLD

marktoda:
We are adding the 3 new ETH-like coins: ETC, RBTC and CGLD. This commit
adds the basic structure for these 3 coins, throwing not implemented for
most of the actual functions in the classes. This is because the logic
for those functions is still under development in the bitgo-account-lib
library and we will update here to fill in those implementations as
those are finished and published to npm. This commit helps get some of
the boiler plate out of the way when the time comes for that.

It's worth noting that ETH itself does not extend the new abstractEthLikeCoin
yet, because its implementation currently works fine. In the future when
the bitgo-account-lib implementation is finished and stable, we may
change ETH itself to use the new implementation.

Ticket: BG-17980


modules/core/src/v2/coins/abstractEthLikeCoin.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export interface ExplainTransactionOptions {
+  txHex?: string;
+  halfSigned?: {
+    txHex: string;
+  };
+  feeInfo: TransactionFee;
+}
+
+export class AbstractEthLikeCoin extends BaseCoin {
```
tylerlevine, line 58: We already have `AbstractUtxoCoin`, seems like the equivalent would be `AbstractAccountCoin`. Any reasons why you picked `EthLike` for the name?

tylerlevine, line 58: Why call it `Abstract` but not actually make it `abstract`?

modules/core/src/v2/coins/abstractEthLikeCoin.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  getFullName() {
+    return this._staticsCoin.fullName;
+  }
+
+  getBaseFactor() {
+    return Math.pow(10, this._staticsCoin.decimalPlaces);
+  }
+
+  static createInstance(bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>): BaseCoin {
```
tylerlevine, line 87: This should not be defined here if this class is actually abstract.

modules/core/src/v2/coins/cgld.ts
```diff
@@ -0,0 +1,17 @@
...
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import { AbstractEthLikeCoin } from './abstractEthLikeCoin';
+
+export class Cgld extends AbstractEthLikeCoin {
+  protected constructor(bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) {
+    super(bitgo, staticsCoin);
+  }
+
+  static createInstance(bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>): BaseCoin {
+    return new AbstractEthLikeCoin(bitgo, staticsCoin);
```
tylerlevine, line 15: Should return a `Cgld` instance, not an instance of the abstract base. (Same for other subclasses)

modules/core/test/v2/unit/coins/cgld.ts
```diff
@@ -0,0 +1,22 @@
...
+    bitgo = new TestBitGo({ env: 'mock' });
+    bitgo.initializeTestVars();
+    basecoin = bitgo.coin('tcgld');
+  });
+
+  it('should instantiate the coin', function() {
+    let localBasecoin = bitgo.coin('tcgld');
+    localBasecoin.should.be.an.instanceof(Tcgld);
+
+	  localBasecoin = bitgo.coin('cgld');
```
tylerlevine, line 19: Indent (same for next 2 files)

modules/core/test/v2/unit/coins/cgld.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  before(function() {
+    bitgo = new TestBitGo({ env: 'mock' });
+    bitgo.initializeTestVars();
+    basecoin = bitgo.coin('tcgld');
+  });
+
+  it('should instantiate the coin', function() {
+    let localBasecoin = bitgo.coin('tcgld');
+    localBasecoin.should.be.an.instanceof(Tcgld);
```
tylerlevine, line 17: Why does this check pass? Shouldn't the returned instance be an `instanceof AbstractEthLikeCoin` and not `Tcgld`? Am I missing something here?

tylerlevine: Just checked the tests and these `instanceof` checks are indeed failing (glad I'm not going crazy). Please check

modules/core/src/v2/coins/abstractEthLikeCoin.ts
marktoda: Because CELO, RSK and ETC are all either code clones or API-compatible with ETH, making the code for those 3 specifically reusable compared to other account-based coins which have different RPC calls, wallet formats etc. It's not super clear since there's no real implementation here, but once it is -- you'll see that it is logic that would work for ETH-like coins but not account-based coins in general

marktoda: Similar to `abstractUTXOCoin`, it's meant to implement the basic shared logic that all ETH-like coins will need to implement anyways. I can change it to just `baseEthLikeCoin` or something if you like.

modules/core/test/v2/unit/coins/cgld.ts
marktoda: Is it still after the latest change? It works for me:
```
▶ mocha --opts test/mocha.opts "test/v2/unit/coins/cgld.ts"       
  
 
(node:17052) DeprecationWarning: Configuration via mocha.opts is DEPRECATED and will be removed from a future version of Mocha. Use RC files or package.json instead.
(node:17052) [DEP0091] DeprecationWarning: crypto.DEFAULT_ENCODING is deprecated.
(node:17052) [DEP0010] DeprecationWarning: crypto.createCredentials is deprecated. Use tls.createSecureContext instead.
(node:17052) [DEP0011] DeprecationWarning: crypto.Credentials is deprecated. Use tls.SecureContext instead.
  Celo Gold
    ✓ should instantiate the coin
```

modules/core/src/v2/coins/abstractEthLikeCoin.ts
tylerlevine: No, it's fine. It's should probably be marked `abstract` just to prevent inadvertent instance creation, but it's fine to handle in a follow up PR.

modules/core/test/v2/unit/coins/cgld.ts
tylerlevine: Yep, latest changes fixes it. Thanks

PR #750, opened 2020-05-26
Fix fanout parameter mismatch in Express

Xekyo:
Express whitelisted `maxNumUnspentsToUse`, but Platform is expecting
`maxNumInputsToUse`.  `maxNumInputsToUse` matches the documentation.
The issue is fixed by updating Express to whitelist `maxNumInputsToUse`
instead.  Since the currently whitelisted parameter in Express would
never have resulted in the desired outcome, this commit does not break
working functionality of existing integrations.

FIXES TICKET: BG-21100

### DIFF
--- a/modules/core/test/v2/unit/wallet.ts
+++ b/modules/core/test/v2/unit/wallet.ts
@@ -7,8 +7,6 @@ import * as sinon from 'sinon';
 require('should-sinon');
 import '../lib/asserts';
 import * as nock from 'nock';
-import * as Promise from 'bluebird';
-const co = Promise.coroutine;
 import * as _ from 'lodash';
 const bitcoin = require('bitgo-utxo-lib');
 import { Wallet } from '../../../src/v2/wallet';
@@ -24,7 +22,7 @@ describe('V2 Wallet:', function() {
   let bgUrl;
   let basecoin;
 
-  before(co(function *() {
+  before(async function() {
     bitgo = new TestBitGo({ env: 'test' });
     bitgo.initializeTestVars();
     basecoin = bitgo.coin('tbtc');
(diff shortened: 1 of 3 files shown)


modules/express/RELEASE_NOTES.md
```diff
@@ -1,5 +1,8 @@
 # BitGo Express Release Notes
 
+## next
```
tylerlevine, line 3: Could you add this to the RELEASE_NOTES.md for the core module instead?

Xekyo:
Updated the release notes accordingly

modules/core/test/v2/unit/wallet.ts
```diff
@@ -590,6 +590,42 @@ describe('V2 Wallet:', function() {
     }));
   });
 
+  describe('maxNumInputsToUse verification', function() {
+    const address = '5b34252f1bf349930e34020a';
+    const maxNumInputsToUse = 2;
+    let basecoin;
+    let wallet;
+
+    before(co(function* () {
```
OttoAllmendinger, line 599: we can use `async function ()` now can't we?

modules/core/test/v2/unit/wallet.ts
```diff
@@ -590,6 +590,42 @@ describe('V2 Wallet:', function() {
...
+        coin: 'tbtc',
+        keys: [
+          '5b3424f91bf349930e340175'
+        ]
+      };
+      wallet = new Wallet(bitgo, basecoin, walletData);
+    }));
+
+
+    it('should pass maxNumInputsToUse parameter when calling fanout unspents', co(function* () {
```
OttoAllmendinger, line 612: same

modules/core/test/v2/unit/wallet.ts
tylerlevine: yes, we can. I would accept either coroutines or async/await.

Xekyo: I've added a second commit that replaces promises in this test with async-await.

modules/core/test/v2/unit/wallet.ts
Xekyo: I've added a second commit that replaces promises in this test with async-await.

OttoAllmendinger: :+1: 

both the Bluebird coroutine-wrapper and the `async` keyword use Promises, the latter is just nicer

Xekyo:
Sorry, the build was not restartable, so I force pushed the same code to kick off another build. @OttoAllmendinger could you please reack assuming that it'll pass now (since the test failure was unrelated)?

PR #753, opened 2020-05-28
Add Celo USD ERC20 token

marktoda:
The CELO blockchain supports the ERC20 standard and thus has various
ERC20 tokens just like Ethereum. This commit adds a new factory builder
for those tokens, and adds our first one -- the celoUSD token, which
comes by pre-compiled with the blockchain.

Ticket: BG-21618

### DIFF
--- a/modules/statics/src/account.ts
+++ b/modules/statics/src/account.ts
@@ -93,6 +93,26 @@ export class Erc20Coin extends AccountCoinToken {
 }
 
 /**
+ * The CELO blockchain supports tokens of the ERC20 standard similar to ETH ERC20 tokens.
+ */
+export class CeloCoin extends AccountCoinToken {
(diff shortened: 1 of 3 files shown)


marktoda:
One question: Do you think we should continue with the stellar token naming scheme, like `xlm:BST-GADDFE4R72YUP2AOEL67OHZN3GJQYPC3VE734N2XFMEGRR2L32CZ3XYZ` for non-ETH tokens? So these would look like `celo:<token address>`

tylerlevine:
I'd be alright with that, but having the full contract address is a bit ugly. Not sure if there is a better solution though.

marktoda:
Definitely ugly -- I prefer it this way, I think the other is mainly to avoid name collisions, which is definitely a concern

PR #754, opened 2020-05-29
 Default to using app dot URLs

tylerlevine:
We have migrated all primary API traffic to the app.bitgo*.com TLDs, and
it's time to start using those domains by default in the SDK.

Users who continue to pass the deprecated environment names will not
need to upgrade, and these will now just be aliases for the equivalent
supported environment.

Ticket: BG-21412


modules/core/test/v2/unit/environments.ts
```diff
@@ -0,0 +1,12 @@
+import { TestBitGo } from '../../lib/test_bitgo';
+import { AliasEnvironments } from '../../../src/v2/environments';
+
+describe('Environments', () => {
+  it('should swap alias environments for supported environments', () => {
+    const aliasEnvs = ['production', 'msProd', 'msTest', 'msDev'];
```
LeoKotschenreuther, line 6: You forgot `msLatest` here.

modules/core/src/v2/environments.ts
```diff
@@ -193,30 +173,6 @@ export const Environments: Environments = {
...
-    stellarFederationServerUrl: 'https://app.bitgo-test.com/api/v2/txlm/federation',
-  }),
-  msDev: Object.assign({}, devBase, {
-    uri: 'https://app.bitgo-dev.com',
-    stellarFederationServerUrl: 'https://app.bitgo-dev.com/api/v2/txlm/federation',
-  }),
-  msLatest: Object.assign({}, devBase, {
-    uri: 'https://app.bitgo-latest.com',
-    stellarFederationServerUrl: 'https://app.bitgo-latest.com/api/v2/xlm/federation',
-  }),
```
twtaylorbitgo, line 219: Probably needs to keep: https://docs.google.com/document/d/1ef89ZmswlN7rVQMZfvbUhDG6w8XlkmF8PGyrJxzJVIg/edit

in mind

modules/core/test/v2/unit/environments.ts
tylerlevine: Fixed

tylerlevine:
> overall LGTM - how did `package-lock.json` sneak back in?

It was added in https://github.com/BitGo/BitGoJS/commit/b29911a1fe67f1262feaba779acab2cb02ac922c

PR #755, opened 2020-05-30
Allow rebuilding consolidation transactions upon tx approval

tylerlevine:
Ticket: BG-19020

### DIFF
--- a/modules/core/src/v2/pendingApproval.ts
+++ b/modules/core/src/v2/pendingApproval.ts
@@ -5,6 +5,7 @@ import { BitGo } from '../bitgo';
 import { validateParams } from '../common';
 import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
+import { RequestTracer } from './internal/util';
 
 import { NodeCallback } from './types';
 import { Wallet } from './wallet';
@@ -232,6 +233,8 @@ export class PendingApproval {
         canRecreateTransaction = false;
       }
 
+      const reqId = new RequestTracer();
+
       /*
        * Internal helper function to get the serialized transaction which is being approved
        */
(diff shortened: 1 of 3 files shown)


tylerlevine:
I tested this in a branch deploy where I set up a wallet with allTx and whitelist policies, then did a consolidate. It created a pending approval as expected, and was able to be approved successfully by the second admin on the wallet. Before, this was erroring out with `recipients cannot be empty`.

tylerlevine:
Added a proper unit test for this.

modules/core/src/v2/pendingApproval.ts
```diff
@@ -354,6 +360,28 @@ export class PendingApproval {
         prebuildParams.hop = true;
       }
 
+      if (!recipients.length) {
+        // no recipients - this is a consolidation transaction
+        const filteredParams = _.pick(params, [
```
LeoKotschenreuther, line 365: Why do you have to filter the params? I think this is dangerous since we might forget about this when we add more build params in the future.

tylerlevine: you're right, no need for this since the params come straight from the pending approval object.

tylerlevine:
> I think your approach is ok and it looks like it works.
> Another idea albeit more involved would be to consolidate our different build APIs (build, consolidate, fanout, sweep, ...) into one API. Since the build params are stored on the pending approval, the client would simply need to call the build API again and pass the params, no matter what type of build it is.

Not a bad idea, but this is a pretty major refactor. We split out these specialized build functions a long time ago, and I think there was probably a reason we didn't build them into `build` directly. Would be good to get a bit more background before going down that path.

PR #756, opened 2020-06-01
Expose CeloCoin from @bitgo/statics

marktoda:
Other libraries need to be able to compare statics `Coin` objects with
the type of coin that they are an instance of. This commit exposes
`CeloCoin` so that this functionality is available

Ticket: BG-21696


marktoda:
@gonzaalvarez70 FYI

PR #757, opened 2020-06-02
Implement validation functions for ETH-like coins

marktoda:
This commit adds implementations for some key and address validation
functions in abstractEthLikeCoin and child implementations such as etc,
cgld and rbtc. These will be required to continue adding support for
those coins

Ticket: BG-21730

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -18,7 +18,7 @@ steps:
 - name: install
   image: node:10
   commands:
-  - "git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true"
+  - git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true
   - yarn install
 
 - name: audit
@@ -79,7 +79,7 @@ steps:
 trigger:
   branch:
   - master
-  - "rel/*"
+  - rel/*
   - prod/production
 
 ---
@@ -102,8 +102,8 @@ steps:
 - name: install
   image: node:8
   commands:
-  - "git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true"
-  - yarn install
+  - git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true
+  - yarn install --ignore-engines
 
 - name: unit-test-changed
   image: node:8
@@ -135,7 +135,7 @@ trigger:
   branch:
     exclude:
     - master
-    - "rel/*"
+    - rel/*
     - prod/production
 
 ---
@@ -158,7 +158,7 @@ steps:
 - name: install
   image: node:10
   commands:
-  - "git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true"
+  - git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true
   - yarn install
 
 - name: unit-test-changed
@@ -191,7 +191,7 @@ trigger:
   branch:
     exclude:
     - master
-    - "rel/*"
+    - rel/*
     - prod/production
 
 ---
@@ -214,7 +214,7 @@ steps:
 - name: install
   image: node:10
   commands:
-  - "git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true"
+  - git fetch origin +refs/heads/$DRONE_REPO_BRANCH:$DRONE_REPO_BRANCH || true
   - yarn install
 
 - name: integration-test
@@ -246,7 +246,7 @@ steps:
 trigger:
   branch:
   - master
-  - "rel/*"
+  - rel/*
   - prod/production
 
 ...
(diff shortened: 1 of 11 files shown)


marktoda:
Audit problem:

```
Overview
All versions of web3 are vulnerable to Insecure Credential Storage. The package stores encrypted wallets in local storage and requires a password to load the wallet. Once the wallet is loaded, the private key is accessible via LocalStorage. Exploiting this vulnerability likely requires a Cross-Site Scripting vulnerability to access the private key.

Remediation
No fix is currently available. Consider using an alternative module until a fix is made available.
```

This dependency (web3) is a dependency (@celo/contractKit) of `bitgo-account-lib`. That library doesn't use the `wallet` feature of web3, just signing and ser/deser utility functions. Is there any way to quiet this error so CI can pass?

tylerlevine:
I'd be ok with changing the `audit` check to `npm audit --audit-level=moderate` (assuming our version supports it). It looks like there are some hacks to allow ignoring certain vulns while they are being resolved. If you find that works better, feel free to add it to your PR.

modules/core/src/v2/coins/abstractEthLikeCoin.ts
```diff
@@ -1,8 +1,12 @@
 /**
  * @prettier
  */
-import * as Bluebird from 'bluebird';
 import { CoinFamily, BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import { Eth, getBuilder } from '@bitgo/account-lib';
+import * as Bluebird from 'bluebird';
```
tylerlevine, line 6: Not requesting any changes, but we can use native async/await now for new code.

marktoda:
Lets see if `auditLevel=moderate` works here, Thanks

modules/core/test/v2/unit/coins/abstractEthCoin.ts
```diff
@@ -0,0 +1,68 @@
+import * as _ from 'lodash';
```
tylerlevine, line 1: Could you add 

```
/**
 * @prettier
 */
```

or manually run prettier on this? 8 space indents is a bit too much.

modules/statics/src/account.ts
```diff
@@ -33,7 +33,6 @@ export class AccountCoin extends BaseCoin {
 
   constructor(options: AccountConstructorOptions) {
     super({
-      isToken: false,
```
tylerlevine, line 36: What does this imply? We have account coins which are tokens too?

tylerlevine:
> Lets see if `auditLevel=moderate` works here, Thanks

Sorry, for yarn it's `yarn audit --level moderate`. Could you add it to the jsonnet and regen the drone.yml?

.drone.yml
```diff
@@ -24,7 +24,7 @@ steps:
 - name: audit
   image: node:10
   commands:
-  - yarn run audit
+  - yarn run audit --audit-level=moderate
```
tylerlevine, line 27: Please add to `audit` script in root `package.json` instead (so we get same behavior locally).

tylerlevine:
Ok, it's slightly more complicated. The `--level` arg just changes what's printed, but not the exit code, which is what drone looks at.

I think you'll need to check `$? >= 8` and fail the audit step if that is true. See https://classic.yarnpkg.com/en/docs/cli/audit/#toc-yarn-audit for more on the audit exit codes. If this is a PITA, we can talk about alternatives.

modules/statics/src/account.ts
marktoda: It doesn't imply anything -- this field was never actually being set because `AccountConstructorOptions` requires the implementors to set `isToken` themselves. You can see that it is set by default to `false` for a normal `AccountCoin`: https://github.com/BitGo/BitGoJS/blob/f9a3a38617998e66fe62da7be655e97c37cd979b/modules/statics/src/account.ts#L159 -- and overridden to `true` for tokens

marktoda:
:+1: I'm trying the following:
```
    "audit": "yarn audit ; test $? < 8",
```

modules/core/src/v2/coins/abstractEthLikeCoin.ts
marktoda: Oh thats awesome! Good to know

marktoda:
`install` phase for node 8 is failing with:
```
error conventional-changelog-angular@5.0.10: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
480
```

Ever seen that? I'm not sure where we are enforcing version >= 10

modules/core/src/v2/coins/abstractEthLikeCoin.ts
marktoda: oh also just realized i'm not even using `co` here :D

tylerlevine:
> `install` phase for node 8 is failing with:
> 
> ```
> error conventional-changelog-angular@5.0.10: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
> 480
> ```
> 
> Ever seen that? I'm not sure where we are enforcing version >= 10
> 
> Edit: found a bug here: [lerna/lerna#2429](https://github.com/lerna/lerna/issues/2429) -- is this the lerna downgrade issue you found earlier Tyler? Let me try rebasing, maybe that'll fix it..

Odd, that is a new one for me. This is different from the issue I had to downgrade for before (that was some zlib error). We could probably start using `--engine-strict=false` in `node8` env.

modules/statics/src/account.ts
tylerlevine: Gotcha, thanks!

marktoda:
:o That worked! Thanks!

marktoda:
Funding the integration ETH wallet...

PR #761, opened 2020-06-09
Add signMessage to abstractEthLikeCoin

marktoda:
It needs to have an implementation for this functino in order to
properly create wallet creation transactions

Ticket: BG-21971


marktoda:
I thought so, but then went and looked and couldn't find any -- writing a small sanity check now

PR #766, opened 2020-06-16
BG-22155 Add etherscan API Key

argjv:
Etherscan now requires an API key for mainnet requests. This PR adds it as an ENV variable.

Source: https://medium.com/etherscan-blog/psa-for-developers-implementation-of-api-key-requirements-starting-from-february-15th-2020-b616870f3746


tylerlevine:
Would be good to have a test for this which uses an API key - I can configure one in drone for you if you like.

PR #767, opened 2020-06-16
Key generation enforcement and key curve into statics

twtaylorbitgo:
SECP key generation needs to be enforced uniformly for offline signing
and UI applications. This adds a Key curve feature to statics to help
with this aspect of testing.

Ticket: BG-22126

### DIFF
--- a/modules/core/package.json
+++ b/modules/core/package.json
@@ -48,7 +48,7 @@
   },
   "dependencies": {
     "@bitgo/account-lib": "^1.3.0",
-    "@bitgo/statics": "^4.2.0",
+    "@bitgo/statics": "^4.3.0-rc.0",
     "@bitgo/unspents": "^0.6.0",
     "@types/bluebird": "^3.5.25",
     "@types/superagent": "^4.1.3",
--- a/modules/core/src/v2/coins/xtz.ts
+++ b/modules/core/src/v2/coins/xtz.ts
@@ -120,9 +120,14 @@ export class Xtz extends BaseCoin {
   generateKeyPair(seed?: Buffer): KeyPair {
     const keyPair = seed ? new bitgoAccountLib.Xtz.KeyPair({ seed }) : new bitgoAccountLib.Xtz.KeyPair();
     const keys = keyPair.getExtendedKeys();
+
+    if (!keys.xprv) {
+      throw new Error('Missing xprv in key generation.');
+    }
+
     return {
       pub: keys.xpub,
-      prv: keys.xprv!,
+      prv: keys.xprv,
     };
   }
 
--- a/modules/core/src/v2/coins/xtz.ts
+++ b/modules/core/src/v2/coins/xtz.ts
@@ -112,17 +112,17 @@ export class Xtz extends BaseCoin {
   }
 
   /**
-   * Generate Tezos key pair
+   * Generate Tezos key pair - BitGo xpub format
    *
    * @param seed
-   * @returns {Object} object with generated pub, prv
+   * @returns {Object} object with generated xpub, xprv
    */
   generateKeyPair(seed?: Buffer): KeyPair {
     const keyPair = seed ? new bitgoAccountLib.Xtz.KeyPair({ seed }) : new bitgoAccountLib.Xtz.KeyPair();
-    const keys = keyPair.getKeys();
+    const keys = keyPair.getExtendedKeys();
     return {
-      pub: keys.pub,
-      prv: keys.prv!,
+      pub: keys.xpub,
+      prv: keys.xprv!,
     };
   }
 
--- a/modules/core/test/v2/unit/keychains.ts
+++ b/modules/core/test/v2/unit/keychains.ts
@@ -9,7 +9,7 @@ import * as common from '../../../src/common';
 import * as _ from 'lodash';
 import * as sinon from 'sinon';
 import {TestBitGo} from '../../lib/test_bitgo';
-import {CoinKind, coins, SupportedKeyCurve, CoinFamily, UnderlyingAsset} from '@bitgo/statics';
+import {CoinKind, coins, KeyCurve, CoinFamily, UnderlyingAsset} from '@bitgo/statics';
 
 const co = Promise.coroutine;
 
@@ -50,7 +50,7 @@ describe('V2 Keychains', function v2keychains() {
   describe('Key generation enforcement for SECP256K1', function() {
     // iterate over non-fiat crypto secp coins
     const coinFamilyValues = Object.keys(CoinFamily).map(n => n.toLowerCase());
-    const cryptoSecpCoins = coins.filter(n => n.primaryKeyCurve === SupportedKeyCurve.Secp256k1
+    const cryptoSecpCoins = coins.filter(n => n.primaryKeyCurve === KeyCurve.Secp256k1
       && n.kind === CoinKind.CRYPTO
       && n.asset !== UnderlyingAsset.USD
       && coinFamilyValues.includes(n.name));
--- a/modules/core/test/v2/unit/keychains.ts
+++ b/modules/core/test/v2/unit/keychains.ts
@@ -2,14 +2,16 @@
 // Test for Keychains
 //
 
-import 'should';
+import * as should from 'should';
 import * as Promise from 'bluebird';
-const co = Promise.coroutine;
 import * as nock from 'nock';
 import * as common from '../../../src/common';
 import * as _ from 'lodash';
 import * as sinon from 'sinon';
-import { TestBitGo } from '../../lib/test_bitgo';
+import {TestBitGo} from '../../lib/test_bitgo';
+import {CoinKind, coins, SupportedKeyCurve, CoinFamily, UnderlyingAsset} from '@bitgo/statics';
+
+const co = Promise.coroutine;
 
 describe('V2 Keychains', function v2keychains() {
   let bitgo;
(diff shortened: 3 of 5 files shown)


twtaylorbitgo:
There's a package.json change, I'll wait until it gets a review and we can remove as part of the statics=>bitgojs deploy process

modules/core/package.json
```diff
@@ -48,7 +48,7 @@
   },
   "dependencies": {
     "@bitgo/account-lib": "^1.3.0",
-    "@bitgo/statics": "^4.1.0",
+    "@bitgo/statics": "../statics",
```
tylerlevine, line 51: Can't point this to a local statics - this will only work when working inside the monorepo.

modules/core/src/v2/coins/xtz.ts
```diff
@@ -112,17 +112,17 @@ export class Xtz extends BaseCoin {
...
    */
   generateKeyPair(seed?: Buffer): KeyPair {
     const keyPair = seed ? new bitgoAccountLib.Xtz.KeyPair({ seed }) : new bitgoAccountLib.Xtz.KeyPair();
-    const keys = keyPair.getKeys();
+    const keys = keyPair.getExtendedKeys();
     return {
-      pub: keys.pub,
-      prv: keys.prv!,
+      pub: keys.xpub,
+      prv: keys.xprv!,
```
tylerlevine, line 125: Why the `!`?

modules/core/test/v2/unit/keychains.ts
```diff
@@ -40,6 +42,41 @@ describe('V2 Keychains', function v2keychains() {
...
+      && n.asset !== UnderlyingAsset.USD
+      && coinFamilyValues.includes(n.name));
+
+    const expectedXpub = 'xpub661MyMwAqRbcGpZf8mxNWhSPdWaLGvQzzage6vq2oQFzq8toVzmkjygYZ3HcZw6eCzAfn9ZdyGjKoKkcpKwackdgznVbiunpq7rkxDu7quS';
+    const expectedXprv = 'xprv9s21ZrQH143K4LVC2kRN9ZVf5UjqsTh9dMm3JYRRF4j1xLZexTTWCBN4hmdZUHeT3vCJPL181ErVaY489ArBKSWaB7Du7vyVS6XC43WtK7A';
+
+    const seed = Buffer.from('this is some random seed we will use', 'utf-8');
+
+    cryptoSecpCoins.forEach(function(coinName) {
+      before(function() {
```
tylerlevine, line 64: please remove

modules/core/test/v2/unit/keychains.ts
```diff
@@ -40,6 +42,41 @@ describe('V2 Keychains', function v2keychains() {
...
+      });
+
+      it('should create the same key with the same seed for coin ' + coinName.name, function() {
```
tylerlevine, line 67: I'd prefer if you move the loop inside the `it` instead of the other way around. This breaks intellij mocha integration.

modules/statics/src/base.ts
```diff
@@ -456,6 +456,14 @@ export const enum UnderlyingAsset {
   'txlm:TST-GBQTIOS3XGHB7LVYGBKQVJGCZ3R4JL5E4CBSWJ5ALIJUHBKS6263644L' = 'txlm:TST-GBQTIOS3XGHB7LVYGBKQVJGCZ3R4JL5E4CBSWJ5ALIJUHBKS6263644L',
 }
 
+/**
+ * This is the curve BitGo signs against with the user, backup and BitGo key.
+ */
+export const enum SupportedKeyCurve {
```
tylerlevine, line 462: Nit: Maybe rename to just `KeyCurve`? If it's here, it's assumed to be supported.

yarn.lock
```diff
@@ -131,6 +131,9 @@
     protobufjs "^6.8.9"
     tronweb "^2.7.2"
 
+"@bitgo/statics@../statics":
```
tylerlevine, line 134: Please make sure to regenerate `yarn.lock` after fixing the issue I commented on in the package.json.

modules/statics/src/base.ts
```diff
@@ -14,7 +14,7 @@ export const enum CoinKind {
  *
  * For example, the coins `btc` and `tbtc` both belong to the same family, `btc`.
  */
-export const enum CoinFamily {
+export enum CoinFamily {
```
tylerlevine, line 17: Why remove `const`?

modules/statics/src/base.ts
```diff
@@ -456,6 +456,14 @@ export const enum UnderlyingAsset {
   'txlm:TST-GBQTIOS3XGHB7LVYGBKQVJGCZ3R4JL5E4CBSWJ5ALIJUHBKS6263644L' = 'txlm:TST-GBQTIOS3XGHB7LVYGBKQVJGCZ3R4JL5E4CBSWJ5ALIJUHBKS6263644L',
 }
 
+/**
+ * This is the curve BitGo signs against with the user, backup and BitGo key.
+ */
+export const enum SupportedKeyCurve {
+  Secp256k1,
```
tylerlevine, line 463: Please assign to the corresponding string, otherwise debugging const enums after compilation becomes difficult.

modules/statics/src/account.ts
```diff
@@ -154,6 +156,7 @@ export function account(
   decimalPlaces: number,
   asset: UnderlyingAsset,
   features: CoinFeature[] = AccountCoin.DEFAULT_FEATURES,
+  primaryKeyCurve: SupportedKeyCurve = SupportedKeyCurve.Secp256k1,
```
tylerlevine, line 159: Why add in the middle of the args list?

modules/statics/README.md
```diff
@@ -103,4 +103,13 @@ To build the project (from TypeScript to Javascript):
 $ npm run build
 ```
 
-This builds the Javascript and adds it to `dist/src/`. There aren't any tests, but you will receive compilation errors if you have invalid syntax.
+This builds the Javascript and adds it to `dist/src/`. You will receive compilation errors if you have invalid syntax.
+
+
+## Tests
```
tylerlevine, line 109: Nice, thanks 👍

modules/core/test/v2/unit/keychains.ts
```diff
@@ -40,6 +42,41 @@ describe('V2 Keychains', function v2keychains() {
     }));
   });
 
+  /**
+   * This section's intention is to provide some key generation sanity checking. generateKeyPair is a general surface
+   * for key generation but the keys are treated the same by BitGo down the line. Any SECP256K1 based coins key-pairs can
+   * be re-used so need to be the same.
+   **/
+  describe('Key generation enforcement for SECP256K1', function() {
```
tylerlevine, line 50: I don't quite follow this. What are we enforcing exactly? This test just makes sure each coin can take a seed and generate the expected xpub.

modules/core/package.json
twtaylorbitgo: yup, I'll have to remove this for a publish I figure

tylerlevine:
Also, could you please rebase? Looks like you are conflicting on the package.json

modules/core/src/v2/coins/xtz.ts
twtaylorbitgo: yeah, I asked the same question.. this is more of an account-lib => consumers oddity. the prv is optional for this method but in typescript we want to say we always expect it 

https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci

modules/core/test/v2/unit/keychains.ts
twtaylorbitgo: ahhh shucks. done

modules/statics/src/base.ts
twtaylorbitgo: sure

modules/statics/src/base.ts
twtaylorbitgo: so this is basically so we can enumerate in our tests - const enum are not actually iterable, they're not even a thing in JS, they're like preprocessor directives in C I believe

https://stackoverflow.com/questions/28818849/how-do-the-different-enum-variants-work-in-typescript

modules/statics/src/base.ts
twtaylorbitgo: oh, sure

modules/core/test/v2/unit/keychains.ts
twtaylorbitgo: I commented in private, it basically has to do with how we do key gen here

modules/core/test/v2/unit/keychains.ts
twtaylorbitgo: ok done

modules/statics/src/account.ts
twtaylorbitgo: so if you look at how we instantiate coins in `coins.ts`, I'd have to pass undefined like a ton of times. I don't mind moving it or we'd have to destructure these constructor params. I just think it was out of scope for this PR to change all this scaffolding

modules/core/src/v2/coins/xtz.ts
tylerlevine: I'd rather throw in that case - the consumer of this method is expecting that the property is there.

modules/core/package.json
twtaylorbitgo: should be resolved but we still need statics to be bumped as a dep or RC push

modules/core/src/v2/coins/xtz.ts
twtaylorbitgo: done

PR #768, opened 2020-06-16
Add CGLD Celo token

marktoda:
The CGLD coin on CELO chain is both the native asset of the chain AND
a token following the ERC20 interface standard. BitGo intends to *send*
the coin using the ERC20 token interface, but must support receives of
it from both the token interface and the native asset interface. This
commit adds the token version of CELO to @bitgo/statics, including its
token address on celo mainnet chain

Ticket: BG-22179

### DIFF
--- a/modules/statics/src/account.ts
+++ b/modules/statics/src/account.ts
@@ -72,10 +72,10 @@ export class AccountCoinToken extends AccountCoin {
 }
 
 /**
- * ERC 20 is a token standard for the Ethereum blockchain. They are similar to other account coins, but have a
- * contract address property which identifies the smart contract which defines the token.
+ * Some blockchains support tokens which are defined by an address at which they have a smart contract deployed.
+ * Examples are ERC20 tokens, and the equivalent on other chains.
  */
-export class Erc20Coin extends AccountCoinToken {
+export class ContractAddressDefinedToken extends AccountCoinToken {
   public contractAddress: ContractAddress;
 
   constructor(options: Erc20ConstructorOptions) {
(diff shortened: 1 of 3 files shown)


marktoda:
I'm unsure if this is the best way to represent this, and not sure how `CoinMap` will work with two coins w/ the same name. Thoughts?

tylerlevine:
I'm thinking more about how to do this and it seems like having an `Erc20CompatibleAccountToken` subclass might be what we need. Then you could put the contract address in the main coin, and still have everything in a single account coin.

marktoda:
Smart idea -- will work on that

modules/statics/src/account.ts
```diff
@@ -71,10 +71,10 @@ export class AccountCoinToken extends AccountCoin {
 }
 
 /**
- * ERC 20 is a token standard for the Ethereum blockchain. They are similar to other account coins, but have a
- * contract address property which identifies the smart contract which defines the token.
+ * Some blockchains support tokens which are defined by an address at which they have a smart contract deployed.
+ * Examples are ERC20 tokens, and the equivalent on other chains.
  */
-export class Erc20Coin extends AccountCoinToken {
+export class ContractAddressDefinedToken extends AccountCoinToken {
```
marktoda, line 77: Added this for the following reasons:
- deduplicate code for all of these ERC20-esque coins
- allow `instanceof` checks in an intuitive way (i.e. Celo ERC20 coins should *not* be a subclass of ETH ERC20 coins)
- Allow `instanceof ContractAddressDefinedToken` checks if a user of this library wants to see if a given coin has `contractAddress` field since `if (coin.contractAddress)` doesn't work in typescript clients because the output of `CoinMap` is just `BaseCoin`

tylerlevine:
sorry, bit of a traffic jam today with you and @twtaylorbitgo working on the same part of the code. Could you please rebase? Change lgtm.

PR #770, opened 2020-06-17
BG-22155 Add API Key as a parameter

argjv:
It is easier for a user to pass the API key as a method paramter than exposing it as an env variable.
Having it as an ENV variable it's still useful in test environments, so it doesn't hurt to leave it.

TICKET: BG-22155

### DIFF
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -1105,6 +1105,9 @@ export class Eth extends BaseCoin {
   recoveryBlockchainExplorerQuery(query: any, callback?: NodeCallback<any>): Bluebird<any> {
     const self = this;
     return co(function*() {
+      if (common.Environments[self.bitgo.getEnv()].etherscanApiToken) {
+        query.apikey = common.Environments[self.bitgo.getEnv()].etherscanApiToken;
+      }
       const response = yield request
         .get(common.Environments[self.bitgo.getEnv()].etherscanBaseUrl + '/api')
         .query(query);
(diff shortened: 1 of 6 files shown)


modules/core/src/v2/coins/eth.ts
```diff
@@ -331,17 +333,21 @@ export class Eth extends BaseCoin {
   /**
    * Query Etherscan for the balance of an address
    * @param address {String} the ETH address
+   * @param apiKey {String} optional etherscan API key to authenticate the requests with
    * @param callback
    * @returns {BigNumber} address balance
    */
-  queryAddressBalance(address: string, callback?: NodeCallback<any>): Bluebird<any> {
+  queryAddressBalance(address: string, apiKey?: string, callback?: NodeCallback<any>): Bluebird<any> {
```
tylerlevine, line 340: This is a breaking change to the API. I would be surprised if there are customers using this, but is there a way to do this without changing the API?

Could you pass it as part of the options when you create the bitgo object instead? Then you can store it in the environment.

tylerlevine: For example, if someone is doing something like this, they will not be able to upgrade without updating their code.

```javascript
bitgo.coin('eth').queryAddressBalance('0x1234...', (err, res) => doStuffWithBalance(res));
```
I would like to drop callback params, but we should do it all at once in a single major version bump.

argjv: > Could you pass it as part of the options when you create the bitgo object instead? Then you can store it in the environment.

I could do that, where should I plug that in?

PR #772, opened 2020-06-19
BG-21898: Test the theories of transaction errors proposed in investigation

glowang:
Do not merge. This PR creates scenarios for us to test what error messages are thrown in each condition: double spending and duplicate tx.

Test the theories of transaction errors proposed in investigation

### DIFF
--- a/modules/core/example/ts/double-spend-example.ts
+++ b/modules/core/example/ts/double-spend-example.ts
@@ -14,7 +14,7 @@ const walletId = 'hidden'; //TODO
 // TODO: set your passphrase here
 const walletPassphrase = 'hidden';
 
-async function main() { // this is a duplicate transaction example
+async function main() {
   bitgo.authenticateWithAccessToken({ accessToken });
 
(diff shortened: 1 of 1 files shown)


modules/core/example/ts/double-spend-example.ts
```diff
@@ -0,0 +1,51 @@
...
+
+async function main() {
+  bitgo.authenticateWithAccessToken({ accessToken });
+
+  const walletInstance = await basecoin.wallets().get({ id: walletId });
+
+  const newReceiveAddress1 = await walletInstance.createAddress();
+  const newReceiveAddress2 = await walletInstance.createAddress();
+
+  const transaction1 = await walletInstance.send({
```
glowang, line 25: @Xekyo  should we take out the `await` ?

Xekyo: Yes

glowang:
The tests passed and the code change for that has been merged.

PR #777, opened 2020-06-26
Make CELO not a "token"

marktoda:
CELO is a weird edge case in that it is both the native coin on its
chain and follows the ERC20 token interface. Thus we had to make a new
class for it to illustrate that. However, the coin should not be
specified as a "token" as that causes some components to fail, as they
assume that all "token"s as children of another coin and thus should not
be displayed as first class citizens. This commit changes CELO to not be
considered a token in this way

Ticket: BG-22435

### DIFF
--- a/modules/statics/src/account.ts
+++ b/modules/statics/src/account.ts
@@ -66,7 +66,6 @@ export class AccountCoinToken extends AccountCoin {
   constructor(options: AccountConstructorOptions) {
     super({
       ...options,
-      isToken: true,
     });
   }
 }
@@ -101,7 +100,15 @@ export class Erc20Coin extends ContractAddressDefinedToken {}
 /**
  * Some blockchains have native coins which also support the ERC20 interface such as CGLD.
  */
-export class Erc20CompatibleAccountCoin extends ContractAddressDefinedToken {}
+export class Erc20CompatibleAccountCoin extends ContractAddressDefinedToken {
+  constructor(options: Erc20ConstructorOptions) {
+    super({
+      ...options,
+      // These coins should not be classified as tokens as they are not children of other coins
+      isToken: false,
+    });
+  }
+}
 
 /**
  * The CELO blockchain supports tokens of the ERC20 standard similar to ETH ERC20 tokens.
@@ -280,7 +287,7 @@ export function erc20CompatibleAccountCoin(
       features,
       decimalPlaces,
       asset,
-      isToken: true,
+      isToken: false,
       primaryKeyCurve,
     })
   );


marktoda:
cc @bitgoAaron to verify this is gonna be OK

modules/statics/src/account.ts
```diff
@@ -66,7 +66,7 @@ export class AccountCoinToken extends AccountCoin {
   constructor(options: AccountConstructorOptions) {
     super({
       ...options,
-      isToken: true,
+      isToken: options.isToken || true,
```
tylerlevine, line 69: Could just have `isToken: true` above the `...options`.

marktoda: ooo ok fixing

marktoda: actually I get this when I try:
```

src/account.ts:68:7 - error TS2783: 'isToken' is specified more than once, so this usage will be overwritten.

68       isToken: true,
         ~~~~~~~~~~~~~

  src/account.ts:69:7
    69       ...options,
             ~~~~~~~~~~
    This spread always overwrites this property.


Found 1 error.

```

marktoda: OK since `isToken` is a required field for `AccountConstructorOptions` it wouldn't let me do that, but it does give me the simple solution of removing the `true` override in favor of a `false` override in `Erc20CompatibleAccountCoin`. Verified like:
```
> a.coins.get('omg')
Erc20Coin {
  fullName: 'OmiseGO Token',
  name: 'omg',
  prefix: '',
  suffix: 'OMG',
  kind: 'crypto',
  family: 'eth',
  isToken: true,
  features:
   [ 'account-model',
     'requires-big-number',
     'valueless-transfer',
     'transaction-data' ],
  decimalPlaces: 18,
  asset: 'omg',
  network:
   Ethereum {
     type: 'mainnet',
     family: 'eth',
     explorerUrl: 'https://etherscan.io/tx/',
     accountExplorerUrl: 'https://etherscan.io/address/' },
  primaryKeyCurve: 'secp256k1',
  contractAddress: '0xd26114cd6ee289accf82350c8d8487fedb8a0c07' }
> a.coins.get('tcgld')
Erc20CompatibleAccountCoin {
  fullName: 'Testnet Celo Gold',
  name: 'tcgld',
  prefix: '',
  suffix: 'TCGLD',
  kind: 'crypto',
  family: 'cgld',
  isToken: false,
  features:
   [ 'account-model',
     'requires-big-number',
     'valueless-transfer',
     'transaction-data' ],
  decimalPlaces: 18,
  asset: 'cgld',
  network:
   CeloTestnet {
     type: 'testnet',
     family: 'cgld',
     explorerUrl: 'https://alfajores-blockscout.celo-testnet.org/tx/',
     accountExplorerUrl: 'https://alfajores-blockscout.celo-testnet.org/address/' },
  primaryKeyCurve: 'secp256k1',
  contractAddress: '0xf194afdf50b03e69bd7d057c1aa9e10c9954e4c9' }
> a.coins.get('tcusd')
CeloCoin {
  fullName: 'Test Celo USD Token',
  name: 'tcusd',
  prefix: '',
  suffix: 'TCUSD',
  kind: 'crypto',
  family: 'cgld',
  isToken: true,
  features:
   [ 'account-model',
     'requires-big-number',
     'valueless-transfer',
     'transaction-data' ],
  decimalPlaces: 18,
  asset: 'cusd',
  network:
   CeloTestnet {
     type: 'testnet',
     family: 'cgld',
     explorerUrl: 'https://alfajores-blockscout.celo-testnet.org/tx/',
     accountExplorerUrl: 'https://alfajores-blockscout.celo-testnet.org/address/' },
  primaryKeyCurve: 'secp256k1',
  contractAddress: '0xa561131a1c8ac25925fb848bca45a74af61e5a38' }
```

PR #780, opened 2020-07-01
Merge @bitgo/account-lib

tylerlevine:
We've decided that we'd like to merge account-lib into the BitGoJS monorepo, to ease upgrade cycles which touch many packages.

This PR merges the entire history of the account-lib repo into the `modules/account-lib` subtree.

Ticket: BG-22556


modules/account-lib/resources/trx/protobuf/tron.d.ts
```diff
@@ -1,11024 +0,0 @@
-import * as $protobuf from "protobufjs";
```
twtaylorbitgo: removing these is problematic - i'm not so sure we didn't modify these in previous commits

twtaylorbitgo: i think we can disregard this

tylerlevine:
I think we'll need to disable the signed commit requirement, since [this commit](https://github.com/BitGo/bitgo-account-lib/commit/07231694647d0f4a5413575fffa5f371a91d2065) can't be verified.

PR #785, opened 2020-07-07
Change CGLD to CELO

marktoda:
The CELO foundation changed the name of their token from cgld to celo.
This commit updates @bitgo/statics and coreaccordingly.
Another change will be required once @bitgo/account-lib is updated,
but that change requires the @bitgo/statics change, so we finish this
first.

Ticket: BG-22664


modules/core/test/v2/unit/coins/abstractEthCoin.ts
```diff
@@ -57,6 +57,10 @@ describe('ETH-like coins', () => {
         gasPrice = '10000',
         gasLimit = '20000',
       }) {
+        // TODO : Remove this when account lib has changed celo to cgld
```
tylerlevine, line 60: Something isn't quite working right, the monorepo is designed to allow cross-package changes like this. Something I'll have to look into later.

tylerlevine, line 60: Do we need this? Seems this causes unit test failures.

tylerlevine:
oh, I see the problem, this is a breaking change for both packages, and bitgojs can't use the version from the repo tree, it gets statics from npm.

marktoda:
@tylerlevine can we have `core` pull in `statics` with `file:../statics` instead of from npm?

tylerlevine:
Unfortunately no, because this is a library and we can't rely on that path being there all the time (only when installing from the repo). It's actually account-lib that causes the issue. `Error: Coin tcelo not supported` comes from https://github.com/BitGo/bitgo-account-lib/blob/master/src/index.ts#L58

PR #789, opened 2020-07-10
Add single sig transactions for ETH clones

marktoda:
This commit adds the ability for ETH clones transaction builders to send
single-signature transactions. This is useful for BitGo to craft
transactions from the fee address, including recoveries and fixing nonce
holes.

Ticket: BG-22801

### DIFF
--- a/modules/account-lib/src/coin/eth/transactionBuilder.ts
+++ b/modules/account-lib/src/coin/eth/transactionBuilder.ts
@@ -23,12 +23,7 @@ import {
   hasSignature,
   isValidEthAddress,
 } from './utils';
-import {
-  flushForwarderTokensMethodId,
-  flushTokensTypes,
-  walletSimpleByteCode,
-  walletSimpleConstructor,
-} from './walletUtil';
+import { walletSimpleByteCode, walletSimpleConstructor } from './walletUtil';
 
 const DEFAULT_M = 3;
 
@@ -42,6 +37,7 @@ export class TransactionBuilder extends BaseTransactionBuilder {
   private _sourceKeyPair: KeyPair;
   private _counter: number;
   private _fee: Fee;
+  private _value: string;
 
   // the signature on the external ETH transaction
   private _txSignature: SignatureParts;
@@ -68,6 +64,7 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     this._common = getCommon(this._coinConfig.network.type);
     this._type = TransactionType.Send;
     this._counter = 0;
+    this._value = '0';
     this._walletOwnerAddresses = [];
     this.transaction = new Transaction(this._coinConfig, this._common);
   }
@@ -101,6 +98,8 @@ export class TransactionBuilder extends BaseTransactionBuilder {
         return this.buildAddressInitializationTransaction();
       case TransactionType.FlushTokens:
         return this.buildFlushTokensTransaction();
+      case TransactionType.SingleSigSend:
+        return this.buildBase('0x');
       default:
         throw new BuildTransactionError('Unsupported transaction type');
     }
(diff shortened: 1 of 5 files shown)


marktoda:
Doing this specifically to add support for the admin tool to fix nonce holes by sending self-transactions. However it'll be needed for fee address recoveries, and potentially other things in the future

modules/account-lib/src/coin/eth/transactionBuilder.ts
```diff
@@ -358,14 +368,23 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
 
+  /**
+   * The value to send along with this transaction. 0 by default
+   *
+   * @param {string} value The value to send along with this transaction
+   */
+  value(value: string): void {
+    this._value = value;
+  }
+
```
twtaylorbitgo, line 379: why exactly do we need to send from a single sig address and it changes this?

modules/account-lib/src/coin/eth/transactionBuilder.ts
```diff
@@ -415,7 +434,7 @@ export class TransactionBuilder extends BaseTransactionBuilder {
       .replace('0x', '');
     return walletSimpleByteCode + resultEncodedParameters;
   }
-  //endregion
+  // endregion
```
twtaylorbitgo, line 437: these actually shouldn't be indented

https://youtrack.jetbrains.com/issue/IDEA-80636

modules/account-lib/src/coin/eth/utils.ts
```diff
@@ -315,6 +315,12 @@ export function decodeFlushTokensData(data: string): FlushTokensData {
  * @returns {TransactionType} The classified transaction type
  */
 export function classifyTransaction(data: string): TransactionType {
+  if (data.length < 10) {
+    // contract calls must have at least 4 bytes (method id) and '0x'
+    // if it doesn't have enough data to be a contract call it must be a single sig send
+    return TransactionType.SingleSigSend;
```
twtaylorbitgo, line 321: this seems pretty hacky. I think this function is weak overall actually. this is really classifying a transaction based on data but not value as well?

modules/account-lib/src/coin/eth/transactionBuilder.ts
marktoda: my intellij just did it on its own

modules/account-lib/src/coin/eth/utils.ts
marktoda: How do you suggest I do it then? There's no data to classify single sig txs with

marktoda: What do you suggest TT?

modules/account-lib/src/coin/eth/transactionBuilder.ts
marktoda: When sending through the contract, we always send 0 value in the external tx. Otherwise we would be sending from the fee address to the wallet when we are just trying to send from the wallet externally. The actual send value is encoded in `transferBuilder`.

modules/account-lib/src/coin/eth/utils.ts
marktoda: Why would you classify a transaction based on value? The value doesn't provide any information on the type of transactino we're talking about

twtaylorbitgo: ok, this seems fine

twtaylorbitgo: I think my main concern with this function was that it lumps everything that's below a threshold as a single send but this seems to work as it's not a valid contract call period unless it falls below this threshold

PR #794, opened 2020-07-15
Fix eslint issues in account-lib

silalvado:


modules/account-lib/src/coin/celo/transactionBuilder.ts
```diff
@@ -89,7 +89,7 @@ export class TransactionBuilder extends Eth.TransactionBuilder {
     return walletSimpleByteCode + resultEncodedParameters;
   }
 
-  //region Stake methods
+  // region Stake methods
```
marktoda, line 92: @twtaylorbitgo says this isn't supposed to be here but I dunno :shrug:

gonzaalvarez70: Do you mean it should be in another line, or we should not add the region at all? @twtaylorbitgo

modules/account-lib/.eslintrc.js
```diff
@@ -20,6 +20,8 @@ module.exports =  {
     'object-curly-spacing': ['error', 'always'],
     '@typescript-eslint/camelcase': 'warn',
     '@typescript-eslint/no-use-before-define': 'off',
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    "@typescript-eslint/explicit-member-accessibility": "off",
```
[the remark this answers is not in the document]
silalvado: Done. Thanks!

modules/account-lib/src/coin/celo/transactionBuilder.ts
twtaylorbitgo: Resolved this via private conversation

modules/account-lib/.gitignore
```diff
@@ -10,3 +10,4 @@ package/
 coverage.lcov
 resources/trx/protobuf/*.d.ts
 resources/trx/protobuf/*.js
+/package-lock.json
```
twtaylorbitgo, line 13: sorry - why did we remove this?

silalvado: Removed. Maybe we can discuss it later.

PR #795, opened 2020-07-16
BG-22939 Rename bitgo-utxo-lib

argjv:
We are renaming bitgo-utxo-lib to @bitgo/utxo-lib.

TICEKT: BG-22939


modules/account-lib/resources/trx/protobuf/tron.d.ts
```diff
@@ -0,0 +1,11024 @@
+import * as $protobuf from "protobufjs";
```
tylerlevine, line 1: Please do not commit generated files.

modules/account-lib/.gitignore
```diff
@@ -8,3 +8,5 @@ mochawesome-report/
 package/
 .idea/
 coverage.lcov
+modules/account-lib/resources/trx/protobuf/*.d.ts
```
tylerlevine, line 11: Unfortunately this won't actually ignore the intended files, since the paths are evaluated relative to the .gitignore file. Please remove the `modules/account-lib/` prefix to correct.

PR #799, opened 2020-07-17
BG-22526 Update bitcoin explorer

argjv:
Smartbit doesn't work for testnet anymore so I'm updating the testnet explorer. We should probably update the mainnet one too to match.

Thoughts?

TICKET: BG-22526


tylerlevine:
It seems like a good idea, but are these API compatible? Do we need to make more changes in whatever is using these properties?

argjv:
Well, this is just used to form the URL we redirect to. I don't think we consume the raw JSON.

PR #802, opened 2020-07-17
Remove webhook in wallet integration test

marktoda:
There is a wallet integration test that creates a new webhook and checks
that it works. However, it is stateful and if anything happens in the
middle of a test, for example it gets canceled - it's possible that the
webhook will be left in existence and all future tests will fail due to
this. To avoid this issue, this commit adds a before hook which removes
the webhook, hopefully making it self-healing

Ticket: BG-22987

### DIFF
--- a/modules/core/test/v2/integration/wallets.ts
+++ b/modules/core/test/v2/integration/wallets.ts
@@ -328,6 +328,13 @@ describe('V2 Wallets:', function() {
   });
 
   describe('Get Wallet', function() {
+    const webhookUrl = 'https://mockbin.org/bin/dbd0a0cd-060a-4a64-8cd8-f3113b36cb7d';
+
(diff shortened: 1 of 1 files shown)


modules/core/test/v2/integration/wallets.ts
```diff
@@ -328,6 +328,13 @@ describe('V2 Wallets:', function() {
   });
 
   describe('Get Wallet', function() {
+    const webhookUrl = 'https://mockbin.org/bin/dbd0a0cd-060a-4a64-8cd8-f3113b36cb7d';
+
+    before(co(function *() {
```
tylerlevine, line 333: async/await for new code please

marktoda: :balloon:

PR #803, opened 2020-07-17
Add single sig transactions for XTZ

wesgraham:
This commit adds the ability for XTZ transaction builders to send
single-signature transactions. This is useful for BitGo to craft
transactions from the fee address, including recoveries

relates to Ticket: BG-22530

### DIFF
--- a/modules/account-lib/src/coin/xtz/transactionBuilder.ts
+++ b/modules/account-lib/src/coin/xtz/transactionBuilder.ts
@@ -5,11 +5,11 @@ import { BuildTransactionError, SigningError } from '../baseCoin/errors';
 import { BaseKey } from '../baseCoin/iface';
 import { BaseTransactionBuilder, TransactionType } from '../baseCoin';
 import {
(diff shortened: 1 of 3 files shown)


modules/account-lib/src/coin/xtz/transactionBuilder.ts
```diff
@@ -68,7 +71,7 @@ export class TransactionBuilder extends BaseTransactionBuilder {
   /**
    * Public constructor.
    *
-   * @param _coinConfig
+   * @param _coinConfig {CoinConfig }- coin configuration
```
marktoda, line 74: Usually type param goes before the name, and I think it's mostly unnecessary in typescript projects

wesgraham: Sounds good, was coerced into adding this by the linter :) can change the placement though

marktoda:
@CapnMigraine --- can you :eye: this PR for wes? He wants to add ability to recover funds from fee address

PR #806, opened 2020-07-21
Update @celo/contractKit for accountlib

marktoda:
This commit updates the @celo/contractKit library for accountlib to its
newest version, and includes the fixes for API changes that have
happened since then. This is because there was a bug in the previous
version that caused some perfectly valid signatures to be considered
invalid, which threw and caused user transactions to fail and flaky
tests in other systems. This commit also includes one such transaction
here, which previously would have failed due to the above error.

Ticket: BG-23095


modules/account-lib/test/unit/coin/celo/transactionBuilder/staking.ts
```diff
@@ -2,6 +2,7 @@ import should from 'should';
 import { coins } from '@bitgo/statics';
 import { Celo, getBuilder } from '../../../../../src';
 import { StakingOperationTypes, TransactionType } from '../../../../../src/coin/baseCoin';
+import { RLP } from 'ethers/utils';
```
argjv, line 5: Why the extra dependency if we don't reference it?

marktoda: oh, oops - I was using it for debugging. Will remove in a separate cleanup PR

PR #808, opened 2020-07-21
Optimize account-lib imports

marktoda:
This commit optimizes imports for account-lib as a cleanup step for
recent changes

Ticket: BG-23105


marktoda:
@argjv this change is in reference to the comments on https://github.com/BitGo/BitGoJS/pull/806

argjv:
This doesn't work or why you closed it?

PR #810, opened 2020-07-22
BG-23023 Add HBAR Utils

argjv:
Add validation utils methods for Hedera.

TICKET: BG-23023

### DIFF
--- a/modules/account-lib/src/coin/hbar/utils.ts
+++ b/modules/account-lib/src/coin/hbar/utils.ts
@@ -1,0 +1,42 @@
+import * as _ from 'lodash';
+
+import { AccountId } from '@hashgraph/sdk/lib/account/AccountId';
+import { TransactionId } from '@hashgraph/sdk';
+
+/**
+ * Returns whether or not the string is a valid Hedera account.
+ *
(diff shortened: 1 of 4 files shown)


modules/account-lib/src/coin/hbar/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+/**
+ * Returns whether or not the string is a valid Hedera transaction id or not
+ *
+ * @param {string} txId - the transaction id to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidTransactionHash(txId: string): boolean {
```
CapnMigraine, line 32: In Hedera, transactions have an `id` and a `hash`: https://docs.hedera.com/guides/docs/sdks/transactions#transaction-record.

I think we should call this function: `isValidTransactionId()`, to avoid confusion.

argjv: I agree. However, I would like to change it for other coins too to be consistent.
I'll add that as a follow-up task because it will require some updates in other systems consuming it.

PR #804, opened 2020-07-20
Bga 223 wallet initialization (WIP)

FabianSimtlix:
This PR is to have a first validation of the new implementation for hedera coin.
It Includes initial implementation for KeyPars with ed25519 encryption, skeleton for hedera transaction builders factory, WalletInitialization basic logic.

PR #812, opened 2020-07-23
Hedera Wallet initialization

FabianSimtlix:
This is the first implementation for hedera transactions.
It's a continuation of https://github.com/BitGo/BitGoJS/pull/804

### DIFF
--- a/modules/account-lib/src/coin/hbar/transaction.ts
+++ b/modules/account-lib/src/coin/hbar/transaction.ts
@@ -4,7 +4,7 @@ import { proto } from '../../../resources/hbar/protobuf/hedera';
 import { BaseTransaction } from '../baseCoin';
 import { BaseKey } from '../baseCoin/iface';
 import { TxData } from './ifaces';
-import {toHex} from './utils';
+import { toHex } from './utils';
 
 export class Transaction extends BaseTransaction {
   private _hederaTx: proto.Transaction;
--- a/modules/account-lib/src/coin/hbar/transaction.ts
+++ b/modules/account-lib/src/coin/hbar/transaction.ts
@@ -1,0 +1,78 @@
+import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
+import BigNumber from 'bignumber.js';
+import { proto } from '../../../resources/hbar/protobuf/hedera';
+import { BaseTransaction } from '../baseCoin';
+import { BaseKey } from '../baseCoin/iface';
+import { TxData } from './ifaces';
+import {toHex} from './utils';
+
+export class Transaction extends BaseTransaction {
+  private _hederaTx: proto.Transaction;
+  private _txBody: proto.TransactionBody;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  body(tx: proto.Transaction) {
+    this._txBody = proto.TransactionBody.decode(tx.bodyBytes);
+    this._hederaTx = tx;
+  }
+
+  bodyBytes(bytes: Uint8Array) {
+    this.body(proto.Transaction.decode(bytes));
+  }
+
+  /** @inheritdoc */
+  canSign(key: BaseKey): boolean {
+    return true;
+  }
+
+  /**
+   * Return the transaction in a format it can be broadcasted to the blockchain.
+   *
+   *  @returns {Uint8Array} The encoded transaction
+   */
+  toBroadcastFormat(): string {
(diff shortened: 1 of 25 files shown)


modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -0,0 +1,78 @@
...
+    return true;
+  }
+
+  /**
+   * Return the transaction in a format it can be broadcasted to the blockchain.
+   *
+   *  @returns {Uint8Array} The encoded transaction
+   */
+  toBroadcastFormat(): string {
+    return toHex(proto.Transaction.encode(this._hederaTx).finish());
```
twtaylorbitgo, line 37: I think we just need to roll this into another ticket where we convert these surfaces points to use `Buffer`

twtaylorbitgo: This should't be a blocker however

modules/account-lib/src/coin/hbar/walletInitializationBuilder.ts
```diff
@@ -0,0 +1,101 @@
...
+    super.initBuilder(tx);
+    const createAcc = tx.txBody.cryptoCreateAccount;
+    if (createAcc && createAcc.key && createAcc.key.thresholdKey) {
+      this.initOwners(createAcc.key.thresholdKey as proto.ThresholdKey);
+    }
+  }
+
+  private initOwners(keys: proto.ThresholdKey) {
+    if (keys.keys && keys.keys.keys) {
+      keys.keys.keys.forEach(key => {
```
twtaylorbitgo, line 57: This just seems... wrong...

twtaylorbitgo: (not a blocker just funny)

modules/account-lib/src/coin/hbar/walletInitializationBuilder.ts
```diff
@@ -0,0 +1,101 @@
...
+  private _txBodyData: proto.CryptoCreateTransactionBody;
+  private readonly _duration: proto.Duration = new proto.Duration({ seconds: 120 });
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+    this._txBody = new proto.TransactionBody();
+    this._txBody.transactionValidDuration = this._duration;
+    this._txBodyData = new proto.CryptoCreateTransactionBody();
+    this._txBody.cryptoCreateAccount = this._txBodyData;
+    this._txBodyData.autoRenewPeriod = new proto.Duration({ seconds: 7890000 });
```
argjv, line 24: Is this the max? Can you add a link to the source?

modules/account-lib/src/coin/hbar/walletInitializationBuilder.ts
```diff
@@ -0,0 +1,101 @@
...
+        this.owner(hex.encode(key.ed25519!));
+      });
+    }
+  }
+
+  /** @inheritdoc */
+  protected async buildImplementation(): Promise<Transaction> {
+    this._txBodyData.key = { thresholdKey: this.buildOwnersKeys() };
+    this._txBodyData.initialBalance = 0;
+    this._txBody.transactionFee = +new BigNumber(this._fee.fee); // validate
```
argjv, line 67: What is the `= +` for?

twtaylorbitgo: that makes it positive

twtaylorbitgo: https://en.wikibooks.org/wiki/JavaScript/Operators - although not super sure unary operators work on BigNumber - Fabian can you confirm that? There might be like an absolute value or magnitude function helper I would prefer to call here

FabianSimtlix: In this case has the same effect as calling toNumber() funtion on the BigNumber object

`/**
   * Returns the value of this BigNumber as a JavaScript primitive number.
   *
   * Using the unary plus operator gives the same result.
*/
  toNumber(): number;`

twtaylorbitgo: can we use abs instead? I'd rather use an explicit feature rather than implicit casting

twtaylorbitgo:
Merged with the idea the JSDocs and BN casting thing will be resolved

FabianSimtlix:
https://simtlix.atlassian.net/browse/BGA-275 to track pending cleanups

PR #814, opened 2020-07-24
Bg 23027 add hbar skeleton

twtaylorbitgo:


modules/core/package.json
```diff
@@ -45,8 +45,8 @@
     "gen-docs": "typedoc"
   },
   "dependencies": {
-    "@bitgo/account-lib": "^2.0.0-rc.1",
-    "@bitgo/statics": "^4.3.0",
+    "@bitgo/account-lib": "^2.0.1-rc.1",
+    "@bitgo/statics": "^5.0.0-rc.2",
```
argjv, line 49: Hum? Why didn't this get updated automatically with `yarn lerna publish ...`

twtaylorbitgo: I don't know... I bumped statics but I don't think it's necessary. Definitely had to bump account-lib

modules/core/src/v2/coins/hbar.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /**
+   * Checks if this is a valid base58 or hex address
+   * @param address
+   */
+  isValidAddress(address: string): boolean {
+    if (!address) {
+      return false;
+    }
```
argjv, line 102: We should put this in `bitgoAccountLib.Hbar.Utils.isValidAddress()` if it isn't already. I know we check for `_.isEmpty()` but that doesn't include undefined I guess.

modules/core/src/v2/coins/hbar.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  parseTransaction(
+    params: ParseTransactionOptions,
+    callback?: NodeCallback<ParsedTransaction>
+  ): Bluebird<ParsedTransaction> {
+    return Bluebird.resolve({}).asCallback(callback);
+  }
+
+  verifyAddress(params: VerifyAddressOptions): boolean {
+    return true;
```
argjv, line 134: Shouldn't this use `bitgoAccountLib.Hbar.Utils.isValidAddress(address)`? Or how do we use this method?

twtaylorbitgo: the implementation is basically required - it's used for verifying an address is on a wallet - I changed it so that it throws an error now instead of blindly verifying

twtaylorbitgo: eh you know any memoId is basically valid on a wallet

modules/core/src/v2/coins/hbar.ts
twtaylorbitgo: sure, I'll add it to some cleanup we do there

PR #815, opened 2020-07-24
BGA-272 ED25519 cleanup

gonzaalvarez70:
A general cleanup of the ed25519 implementation, using the Hashgraph sdk library as a filter.
Added more tests.


modules/account-lib/test/unit/coin/hbar/keyPair.ts
```diff
@@ -16,12 +13,29 @@ describe('Hedera Key Pair', () => {
       should.equal(Buffer.from(keyPair.getKeys().pub).toString('hex'), pub);
     });
 
+    it('from a public key with pefix', () => {
```
twtaylorbitgo, line 16: sp

PR #821, opened 2020-07-25
BGA-240 Improve Key Pair abstraction

argjv:
I'm making some adjustments to the abstraction model to better accommodate for core logic sharing and new specialized coins Key pairs:
Key Pairs UML class diagram

I'm introducing `@stablelib/hex` dependency again because I noticed hex encoding/decoding with Buffer and Uint8array is not idempotent. An example of a buggy conversion is:

`302e020100300506032b657004220420302e020100300506032b65700422042062b0b669de0ab5e91b4328e1431859a5ca47e7426e701019272f5c2d52825b011` 

This key converted back and forth from hex string to Uint8array did not result in the same string, which is extremely dangerous.

### DIFF
--- a/modules/account-lib/src/coin/hbar/utils.ts
+++ b/modules/account-lib/src/coin/hbar/utils.ts
@@ -1,6 +1,7 @@
 import * as _ from 'lodash';
 import { AccountId } from '@hashgraph/sdk/lib/account/AccountId';
 import { Ed25519PublicKey, TransactionId } from '@hashgraph/sdk';
+import * as hex from '@stablelib/hex';
 
 /**
  * Returns whether or not the string is a valid Hedera account.
@@ -51,7 +52,7 @@ export function isValidPublicKey(key: string): boolean {
     return false;
   }
   try {
-    const pubKey = Ed25519PublicKey.fromString(key);
+    const pubKey = Ed25519PublicKey.fromString(key.toLowerCase());
     return !_.isNaN(pubKey.toString());
   } catch (e) {
     return false;
@@ -65,17 +66,17 @@ export function isValidPublicKey(key: string): boolean {
  * @returns {string} - the hex value
  */
 export function toHex(buffer: Buffer | Uint8Array): string {
-  return Buffer.from(buffer).toString('hex');
+  return hex.encode(buffer, true);
 }
 
 /**
  * Returns a Uint8Array of the given hex string
  *
- * @param {string} hex - the hex string to be converted
+ * @param {string} str - the hex string to be converted
  * @returns {string} - the Uint8Array value
  */
-export function toUint8Array(hex: string): Uint8Array {
-  return Uint8Array.from(Buffer.from(hex, 'hex'));
+export function toUint8Array(str: string): Uint8Array {
+  return hex.decode(str);
 }
 
 /**
--- a/modules/account-lib/src/utils/crypto.ts
+++ b/modules/account-lib/src/utils/crypto.ts
@@ -1,5 +1,7 @@
 import { HDNode, ECPair, networks } from '@bitgo/utxo-lib';
+import * as nacl from 'tweetnacl';
 import { ExtendedKeys } from '../coin/baseCoin/iface';
+import { toUint8Array } from '../coin/hbar/utils';
 
 /**
  * @param {string} xpub - a base-58 encoded extended public key (BIP32)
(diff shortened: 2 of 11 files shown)


modules/account-lib/src/coin/hbar/utils.ts
```diff
@@ -65,17 +66,17 @@ export function isValidPublicKey(key: string): boolean {
...
  * Returns a Uint8Array of the given hex string
  *
- * @param {string} hex - the hex string to be converted
+ * @param {string} str - the hex string to be converted
  * @returns {string} - the Uint8Array value
  */
-export function toUint8Array(hex: string): Uint8Array {
-  return Uint8Array.from(Buffer.from(hex, 'hex'));
+export function toUint8Array(str: string): Uint8Array {
+  return hex.decode(str);
```
twtaylorbitgo, line 79: what happens if we don't pass a hex string to this library? does it throw an error? or does it just run through and give garbage?

modules/account-lib/src/utils/crypto.ts
```diff
@@ -103,3 +105,48 @@ export function isValidPrv(prv: string): boolean {
...
+/**
+ * Whether the input is a valid ed25519 private key
+ *
+ * @param {string} prv A hexadecimal private key to validate
+ * @returns {boolean} Whether the input is a valid public key or not
+ */
+export function isValidEd25519Seed(prv: string): boolean {
+  try {
+    const decodedPrv = toUint8Array(prv);
+    return decodedPrv.length === nacl.sign.seedLength;
```
twtaylorbitgo, line 118: this I think is gonna present a major problem if we ever get a prv that is encoded and the same size as our seed. I think we should use whatever helpers or encoding format to gate this if we can

modules/account-lib/src/utils/crypto.ts
```diff
@@ -103,3 +105,48 @@ export function isValidPrv(prv: string): boolean {
...
+/**
+ * Whether the input is a valid ed25519 private key
+ *
+ * @param {string} prv A hexadecimal private key to validate
+ * @returns {boolean} Whether the input is a valid public key or not
+ */
+export function isValidEd25519SecretKey(prv: string): boolean {
+  try {
+    const decodedPrv = toUint8Array(prv);
+    return decodedPrv.length === nacl.sign.secretKeyLength;
```
twtaylorbitgo, line 133: basically same comment as above

twtaylorbitgo: also is it possible that the seed for any derived coin is much larger than this length?

modules/core/test/v2/unit/coins/hbar.ts
```diff
@@ -42,8 +42,8 @@ describe('Hedera Hashgraph:', function() {
       const seed = Buffer.from(seedText, 'hex');
       const keyPair = basecoin.generateKeyPair(seed);
 
-      keyPair.pub.should.equal('9cc402b5c75214269c2826e3c6119377cab6c367601338661c87a4e07c6e0333');
       keyPair.prv.should.equal('80350b4208d381fbfe2276a326603049fe500731c46d3c9936b5ce036b51377f');
+      keyPair.pub.should.equal('9cc402b5c75214269c2826e3c6119377cab6c367601338661c87a4e07c6e0333');
```
twtaylorbitgo, line 46: why was this moved? weird

argjv: This will change. I needed it like this to make the test pass. I need to release account-lib 1st to use the new getKeys() method.

modules/account-lib/src/utils/crypto.ts
argjv: Size is not variable AFAIK. The standard size for a seed is 32 bytes, the same as public keys. Secret keys are derived from seeds and they are 64 bytes: https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/#:~:text=Ed25519%20keys%20start%20life%20as,and%20a%20%E2%80%9Cright%20half%E2%80%9D

modules/account-lib/src/coin/hbar/utils.ts
argjv: Yup: https://github.com/StableLib/stablelib/blob/master/packages/hex/hex.ts

modules/account-lib/src/utils/crypto.ts
argjv: Also: https://cryptobook.nakov.com/digital-signatures/eddsa-and-ed25519

twtaylorbitgo: oh yeah this looks good. ok my only concern is prvs then so this is dismissed

modules/account-lib/src/utils/crypto.ts
twtaylorbitgo: so maybe this is where it gets weird. i think we should use like an encoded format to ensure this is one of ours. a prefix maybe? 

I'll leave this to chew on later for now

modules/account-lib/src/coin/hbar/utils.ts
twtaylorbitgo: god this is terribly coded. i'll try to figure out this bit shifting later. this looks like perl developers invaded JS

twtaylorbitgo: also we'll need to figure out why `buffer.from(x, 'hex')` doesn't work for this lib

argjv:
I think it happens when the number of characters is not even.

PR #822, opened 2020-07-27
BGA-225 Multisig transfer transaction

gonzaalvarez70:
Multisig transfer funds transaction implementation for account-lib


modules/account-lib/src/coin/hbar/utils.ts
```diff
@@ -112,3 +113,14 @@ export function getCurrentTime(): string {
...
+
+/**
+ * Returns whether or not the string is a valid amount number
+ *
+ * @param {string} amount - the string to validate
+ * @returns {boolean} - the validation result
+ */
+export function isValidAmount(amount: string): boolean {
+  const bigNumberAmount = new BigNumber(amount);
+  return bigNumberAmount.isInteger() && bigNumberAmount.isGreaterThanOrEqualTo(0);
```
argjv, line 125: Is there a max amount of HBARs per transaction?

modules/account-lib/src/coin/hbar/transactionBuilder.ts
```diff
@@ -61,6 +80,43 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
 
   // region Common builder methods
+  /**
+   *  Set the memo
+   *
+   * @param {string} memo A hedera memo
+   * @returns {TransactionBuilder} This transaction builder
+   */
+  memo(memo: string): this {
+    this._memo = memo;
```
argjv, line 90: Add some validations for the memo. Check valid characters and length

gonzaalvarez70: That sounds good, we plan to use the hashgraph sdk validations as a base and then check if we need more validations.

modules/account-lib/src/coin/hbar/utils.ts
gonzaalvarez70: I found this on their SDK `const maxTinybar = new BigNumber(2).pow(63).minus(1);`
We could add that, the idea is to use tinyBars as input right?
(there are 100,000,000 tinyBars in one Hbar)

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,91 @@
...
+import { AccountId } from '@hashgraph/sdk';
+import { proto } from '../../../resources/hbar/protobuf/hedera';
+import { NotImplementedError, BuildTransactionError, InvalidParameterValueError } from '../baseCoin/errors';
+import { BaseKey } from '../baseCoin/iface';
+import { TransactionBuilder } from './transactionBuilder';
+import { Transaction } from './transaction';
+import { isValidAddress, isValidAmount } from './utils';
+
+export class TransferBuilder extends TransactionBuilder {
+  private _txBodyData: proto.CryptoTransferTransactionBody;
```
FabianSimtlix, line 12: same as wallet init builder

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,91 @@
...
+  private _toAddress: string;
+  private _amount: string;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+    this._txBody = new proto.TransactionBody();
+    this._txBody.transactionValidDuration = this._duration;
+    this._txBodyData = new proto.CryptoTransferTransactionBody();
+    this._txBody.cryptoTransfer = this._txBodyData;
+  }
```
FabianSimtlix, line 22: same constructor than wallet init. Perhaps it should be the base class constructor

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,91 @@
...
+    // TODO: Add signature to the tx body
+    return super.buildImplementation();
+  }
+
+  private buildTransferData(): proto.ITransferList {
+    return { accountAmounts: [{ accountID: this.buildRecipientData(), amount: Long.fromString(this._amount) }] };
+  }
+
+  private buildRecipientData(): proto.AccountID {
+    const accountData = new AccountId(this._toAddress);
```
FabianSimtlix, line 36: we should review this approach.
We are coupling an sdk dependency just to build an `proto.AccountId`

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: In WalletInit is CryptoCreateTransactionBody, but all of the other fields can be moved to the base constructor.

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: Yes, you are right most fields are common between the constructors.

modules/account-lib/src/coin/baseCoin/baseTransactionBuilderFactory.ts
```diff
@@ -19,4 +19,9 @@ export abstract class BaseTransactionBuilderFactory {
    * Returns an specific builder to create a wallet initialization transaction
    */
   public abstract getWalletInitializationBuilder();
+
+  /**
+   * Returns a specific builder to create a funds transfer transaction
+   */
+  public abstract getTransferBuilder();
```
twtaylorbitgo, line 26: typically a factory takes like an enum and some params, I'll have to talk internally about this pattern, kinda want to re-factor this already but this looks good for now

modules/account-lib/src/coin/hbar/ifaces.ts
```diff
@@ -6,3 +6,7 @@ export interface TxData {
   fee: number;
   startTime: string;
 }
+
+export interface HederaNode {
+  nodeId: string;
+}
```
twtaylorbitgo, line 12: QUESTION: does this need to be account on node sensitive as well? there's also a realm attribute to the node-account-realm construct I think we might need to understand here

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,91 @@
...
+    this._txBody = new proto.TransactionBody();
+    this._txBody.transactionValidDuration = this._duration;
+    this._txBodyData = new proto.CryptoTransferTransactionBody();
+    this._txBody.cryptoTransfer = this._txBodyData;
+  }
+
+  /** @inheritdoc */
+  protected async buildImplementation(): Promise<Transaction> {
+    this._txBodyData.transfers = this.buildTransferData();
+    // TODO: Add signature to the tx body
```
twtaylorbitgo, line 27: let's put this as TODO with the signing ticket attached

modules/account-lib/src/coin/hbar/walletInitializationBuilder.ts
```diff
@@ -1,6 +1,5 @@
 import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
 import * as hex from '@stablelib/hex';
```
gonzaalvarez70, line 2: I guess we should remove this library right? Since it was removed from the KeyPair PR

modules/account-lib/test/resources/hbar/hbar.ts
```diff
@@ -1,112 +1,46 @@
 export const ACCOUNT_1 = {
   accountId: '0.0.81320',
-  publicKey: '302a300506032b65700321005a9111b5e6881ff20b9243a42ac1a9a67fa16cd4f01e58bab30c1fe611ea8cf9',
   privateKey: '302e020100300506032b65700422042062b0b669de0ab5e91b4328e1431859a5ca47e7426e701019272f5c2d52825b01',
+  publicKey: '302a300506032b65700321005a9111b5e6881ff20b9243a42ac1a9a67fa16cd4f01e58bab30c1fe611ea8cf9',
```
twtaylorbitgo, line 4: mmm jorge just moved this lol - can you check in with @argjv ?

modules/account-lib/src/coin/hbar/ifaces.ts
gonzaalvarez70: Yes, we checked the SDK behaviour to setNodeAccountId and they check this in the same way, by using new AccountID(address). I documented that on the node() method in transaction builder saying that it may take the format `'<shard>.<realm>.<account>'` or `'<account>'`

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: This is a way to take any address since it may take the format `'<shard>.<realm>.<account>'` or `'<account>'`

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: Sure sounds good for next time, this was a work in progress for the same PR.

modules/account-lib/test/resources/hbar/hbar.ts
gonzaalvarez70: I'll merge/rebase from master

gonzaalvarez70: I did a rebase to avoid further conflicts squashing the commits later

gonzaalvarez70:
This PR was rebased in another branch to be compatible with the current signature implementation
New PR: https://github.com/BitGo/BitGoJS/pull/834

PR #825, opened 2020-07-28
Remove smartbit explorer and replace with blockstream.info for V1 CPFP

tylerlevine:
Version 1 CPFP can be easily adapted to use blockstream.info instead of
smartbit, so let's start with converting that logic to unblock some
users of V1 CPFP.

The rest of the recovery code is however fairly tightly coupled to the smartbit API and more work is needed to re-implement this functionality on top of a new explorer API. That work will be done in another PR, and the corresponding tests have been disabled for now.

Ticket: BG-23082


modules/core/src/wallet.ts
```diff
@@ -1363,19 +1364,16 @@ Wallet.prototype.accelerateTransaction = function accelerateTransaction(params,
    * @param parentTxId The ID of the transaction to get the full hex of
    * @returns {Bluebird<any>} The full hex for the specified transaction
    */
```
Xekyo, line 1366: Comment above still says "smartbit.com.au"

modules/core/src/wallet.ts
```diff
@@ -1363,19 +1364,16 @@ Wallet.prototype.accelerateTransaction = function accelerateTransaction(params,
...
-      }
+  async function getParentTxHex({ parentTxId }: { parentTxId: string }): Promise<string> {
+    const explorerBaseUrl = common.Environments[self.bitgo.getEnv()].btcExplorerBaseUrl;
+    const result = await request.get(`${explorerBaseUrl}/tx/${parentTxId}/hex`);
 
-      return result.body.hex[0].hex;
-    }).call(this);
-  };
+    if (!result.text || !/([a-f0-9]{2})+/.test(result.text)) {
+      throw new Error('Did not successfully receive parent tx hex');
```
Xekyo, line 1372: It might be helpful to output what was received instead.

modules/core/src/wallet.ts
Xekyo: Couldn't we just get it from the v2 indexer instead?

tylerlevine: Yeah, we should do this, great idea.

modules/core/src/wallet.ts
tylerlevine: Will do 👍

modules/core/src/wallet.ts
tylerlevine: This turned out to be a bit more difficult than I thought, so I deferred this for later work.

PR #828, opened 2020-07-30
Consolidate ethereumjs-abi version

marktoda:
Consolidate ethereumjs-abi version

This commit makes both core and account-lib module use the same version
of ethereumjs-abi, with the hopes that this will fix an issue where
account-lib's copy does not get browserified when included in frontend
components.

The new ethereumjs-abi version had some breaking changes though, which
we had to work to fix:
- new serialization which i guess requires less padding, so the output
serialized txs were slightly smaller
- rawDecode returns string addresses instead of buffers

Ticket: BG-22775


marktoda:
Any way that folks can think of to statically check for issues like this in the future (assuming that this fixes the actual issue)? I could easily imagine us running into the same issue with another newly added coin. I.e. can we run tests against the browserified versions of the code? Does it work like that?

marktoda:
Yep the new ethereumjs-abi version has a breaking change, I'm working to rix it

marktoda:
Seeing release notes, I think the changes to solidityPack are what caused the different output: https://github.com/ethereumjs/ethereumjs-abi/releases

PR #830, opened 2020-07-30
Bg 23451.whitelist target address

Xekyo:
Allow targetAddress on consolidate and fanout

    The SDK's parameter whitelist actually did not include the
    `targetAddress` parameter, so it was added here now.

    Ticket: BG-23451


Xekyo:
Sorry, I looked it over again and realized that `prebuildConsolidateTransactionParams` was actually only used in the Algo `consolidateAccount` endpoint, so I cleaned that up.

Xekyo:
Rebased as requested.

PR #834, opened 2020-07-31
Added transfer multisig transaction

gonzaalvarez70:
Ticket: BGA-225

### DIFF
--- a/modules/account-lib/src/coin/hbar/transaction.ts
+++ b/modules/account-lib/src/coin/hbar/transaction.ts
@@ -3,12 +3,13 @@ import { hash } from '@stablelib/sha384';
 import BigNumber from 'bignumber.js';
 import { Writer } from 'protobufjs';
 import * as nacl from 'tweetnacl';
+import Long from 'long';
 import { proto } from '../../../resources/hbar/protobuf/hedera';
 import { BaseTransaction } from '../baseCoin';
 import { BaseKey } from '../baseCoin/iface';
 import { SigningError } from '../baseCoin/errors';
 import { TxData } from './ifaces';
-import { toHex, toUint8Array } from './utils';
+import { stringifyAccountId, stringifyTxTime, toHex, toUint8Array } from './utils';
 import { KeyPair } from './';
 
 export class Transaction extends BaseTransaction {
@@ -47,6 +48,7 @@ export class Transaction extends BaseTransaction {
     const sigMap = this._hederaTx.sigMap || new proto.SignatureMap();
     sigMap.sigPair!.push(sigPair);
     this._hederaTx.sigMap = sigMap;
+    this._signatures.push(signature);
   }
 
 
@@ -58,14 +60,42 @@ export class Transaction extends BaseTransaction {
   /** @inheritdoc */
   toJson(): TxData {
     const [acc, time] = this.getTxIdParts();
-    return {
+    const result: TxData = {
       id: acc + '@' + time,
       hash: this.getTxHash(), // TODO: Update once hedera-sdk release this functionality BGA-284
       data: toHex(this._hederaTx.bodyBytes),
       fee: new BigNumber(this._txBody.transactionFee!.toString()).toNumber(),
       from: acc,
       startTime: time,
+      validDuration: this._txBody.transactionValidDuration!.seconds!.toString(),
+      node: stringifyAccountId(this._txBody.nodeAccountID!),
+      memo: this._txBody.memo,
     };
+
+    if (this._txBody.data === 'cryptoTransfer') {
+      const [recipient, amount] = this.getTransferData();
+      result.amount = amount;
+      result.to = recipient;
+    }
+    return result;
+  }
+
+  /**
+   * Get the recipient account and the amount
+   * transferred on this transaction
+   *
+   * @returns {[string, string]} first element is the recipient, second element is the amount
+   */
+  private getTransferData(): [string, string] {
+    let transferData;
+    this._txBody.cryptoTransfer!.transfers!.accountAmounts!.forEach(transfer => {
+      const amount = Long.fromValue(transfer.amount!);
+      if (amount.isPositive()) {
+        transferData = [stringifyAccountId(transfer.accountID!), amount.toString()];
+      }
+    });
+
+    return transferData;
   }
 
   //region getters & setters
@@ -85,6 +115,23 @@ export class Transaction extends BaseTransaction {
   body(tx: proto.Transaction) {
     this._txBody = proto.TransactionBody.decode(tx.bodyBytes);
     this._hederaTx = tx;
+    // this.loadPreviousSignatures();
+  }
+
+  /**
+   * Decode previous signatures from the inner hedera transaction
+   * and save them into the base transaction signature list.
+   */
+  loadPreviousSignatures(): void {
+    if (this._hederaTx.sigMap && this._hederaTx.sigMap.sigPair) {
+      const sigPairs = this._hederaTx.sigMap.sigPair;
+      sigPairs.forEach(sigPair => {
+        const signature = sigPair.ed25519;
+        if (signature) {
+          this._signatures.push(toHex(signature));
+        }
+      });
+    }
   }
 
   /**
@@ -111,31 +158,11 @@ export class Transaction extends BaseTransaction {
       this._txBody.transactionID.transactionValidStart
     ) {
       return [
-        this.stringifyAccountId(this._txBody.transactionID.accountID),
-        this.stringifyTxTime(this._txBody.transactionID.transactionValidStart),
+        stringifyAccountId(this._txBody.transactionID.accountID),
+        stringifyTxTime(this._txBody.transactionID.transactionValidStart),
       ];
     }
     throw new Error('Missing transaction id information');
-  }
-
-  /**
-   * Returns a string representation of an {proto.IAccountID} object
-   *
-   * @param {proto.IAccountID} - account id to be cast to string
-   * @returns {string} - the string representation of the {proto.IAccountID}
-   */
-  private stringifyAccountId({ shardNum, realmNum, accountNum }: proto.IAccountID): string {
-    return `${shardNum || 0}.${realmNum || 0}.${accountNum}`;
-  }
-
-  /**
-   * Returns a string representation of an {proto.ITimestamp} object
-   *
-   * @param {proto.ITimestamp} - timestamp to be cast to string
-   * @returns {string} - the string representation of the {proto.ITimestamp}
-   */
-  private stringifyTxTime({ seconds, nanos }: proto.ITimestamp) {
-    return `${seconds}.${nanos}`;
   }
 
   /**
--- a/modules/account-lib/src/coin/hbar/transactionBuilder.ts
+++ b/modules/account-lib/src/coin/hbar/transactionBuilder.ts
@@ -1,5 +1,6 @@
-import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
 import BigNumber from 'bignumber.js';
+import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
+import { AccountId } from '@hashgraph/sdk';
 import { BaseTransactionBuilder } from '../baseCoin';
 import {
   BuildTransactionError,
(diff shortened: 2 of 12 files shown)


modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -48,14 +50,36 @@ export class Transaction extends BaseTransaction {
...
+
+    if (this._txBody.data === 'cryptoTransfer') {
+      const [recipient, amount] = this.getTransferData();
+      result.amount = amount;
+      result.to = recipient;
+    }
+    return result;
+  }
+
+  private getTransferData(): [string, string] {
```
argjv, line 73: JS Doc

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -75,6 +99,22 @@ export class Transaction extends BaseTransaction {
   body(tx: proto.Transaction) {
     this._txBody = proto.TransactionBody.decode(tx.bodyBytes);
     this._hederaTx = tx;
+    // this.loadPreviousSignatures();
+  }
+
+  /** Decode previous signatures from the inner hedera transaction
+   * and save them into the base transaction signature list.
+   */
```
argjv, line 107: `/** Decode previous signatures from the inner hedera transaction and save them into the base transaction signature list. */`

Or

```
/** 
   * Decode previous signatures from the inner hedera transaction
   * and save them into the base transaction signature list.
   */
```

Choose one or the other but make sure it does not exceed 120 chars per line

modules/account-lib/src/coin/hbar/transactionBuilder.ts
```diff
@@ -75,6 +110,47 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
   // endregion
 
   // region Common builder methods
+  /**
+   *  Set the memo
+   *
+   * @param {string} memo A hedera memo
```
argjv, line 116: Is this a hex? Or can be anything?

modules/account-lib/src/coin/hbar/transactionBuilder.ts
```diff
@@ -75,6 +110,47 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+   * @returns {TransactionBuilder} This transaction builder
+   */
+  memo(memo: string): this {
+    // strings are UTF-16, max 100 bytes
+    if (memo.length * 2 > 100) {
```
argjv, line 121: You should turn this into a buffer/hex 1st and then check the length, no?

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,116 @@
...
+  /** @inheritdoc */
+  protected initBuilder(tx: Transaction): void {
+    super.initBuilder(tx);
+    const transferData = tx.txBody.cryptoTransfer;
+    if (transferData && transferData.transfers && transferData.transfers.accountAmounts) {
+      this.initTransfers(transferData.transfers.accountAmounts);
+    }
+  }
+
+  protected initTransfers(transfers: proto.IAccountAmount[]): void {
```
argjv, line 56: JSDoc

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,116 @@
...
+    transfers.forEach(transferData => {
+      const amount = Long.fromValue(transferData.amount!);
+      if (amount.isPositive()) {
+        this.to(stringifyAccountId(transferData.accountID!));
+        this.amount(amount.toString());
```
argjv, line 61: Are these class variables?

modules/account-lib/src/coin/hbar/transactionBuilder.ts
gonzaalvarez70: it can be any string up to I think 100 bytes. We should add that limit to the JSDoc. Will work on that alongside the other comments.

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,116 @@
...
+  amount(amount: string): this {
+    if (!isValidAmount(amount)) {
+      throw new InvalidParameterValueError('Invalid amount');
+    }
+    this._amount = amount;
+    return this;
+  }
+
+  //endregion
+  //region Validators
```
argjv, line 105: nit: add an empty line before this

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,116 @@
...
+  validateMandatoryFields(): void {
+    if (this._toAddress === undefined) {
```
argjv, line 107: I think we have an address validation method

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,116 @@
...
+      throw new BuildTransactionError('Invalid transaction: missing to');
+    }
+    if (this._amount === undefined) {
```
argjv, line 110: Shouldn't you use isValidAmount? Or smth like that?

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -48,14 +50,36 @@ export class Transaction extends BaseTransaction {
...
     const [acc, time] = this.getTxIdParts();
-    return {
+    const result: TxData = {
       id: acc + '@' + time,
       hash: this.getTxHash(), // TODO: Update once hedera-sdk release this functionality BGA-284
       data: toHex(this._hederaTx.bodyBytes),
       fee: new BigNumber(this._txBody.transactionFee!.toString()).toNumber(),
       from: acc,
       startTime: time,
+      validDuration: new BigNumber(this._txBody.transactionValidDuration!.seconds!.toString()).toNumber(),
```
twtaylorbitgo, line 60: there's a lot of number coercement - can we leave a comment here that this is expressed in seconds and add a check just in case this proto doesn't exist?

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -48,14 +50,36 @@ export class Transaction extends BaseTransaction {
...
+      result.amount = amount;
+      result.to = recipient;
+    }
+    return result;
+  }
+
+  private getTransferData(): [string, string] {
+    let transferData;
+    this._txBody.cryptoTransfer!.transfers!.accountAmounts!.forEach(transfer => {
+      const amount = Long.fromValue(transfer.amount!);
```
twtaylorbitgo, line 76: Why exactly are we using `Long`? Can BN suffice here?

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: Sure!

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: Will do

modules/account-lib/src/coin/hbar/transactionBuilder.ts
gonzaalvarez70: Sure, we will convert it to buffer and check the length. This implementation was based on the SDK current implementation but we will change this and perform some surround testing to check if it keeps working properly.

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: Sounds good

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: The ones that we are getting are decoded from the transaction.
The ones that we are setting are the class variables.

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: 👍

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: Yes, that's validated on the set methods. Here we just check if there's something missing.

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: That validation is done when we set an amount. We just check here if something is missing.

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: Sure we can just return a string

gonzaalvarez70: We did this because it's useful for decoding a transaction. Like we do on transactionBuilder.ts line 91. If we don't do that conversion here we need to add it on that other line.

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: After talking about it with Thomas we agreed that it's ok for now to use Long. It's the same type used by the protobuf generated type for amount. And using BN would involve some sort of casting to the variable.

twtaylorbitgo: Yeah, we'll have to get these things out of our system at some point but I don't think this is the time or place

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,124 @@
...
+   *
+   * @param {string} amount amount to transfer in tinyBars (there are 100,000,000 tinyBars in one Hbar)
+   * @returns {TransferBuilder} the builder with the new parameter set
+   */
+  amount(amount: string): this {
+    if (!isValidAmount(amount)) {
+      throw new InvalidParameterValueError('Invalid amount');
+    }
+    this._amount = amount;
+    return this;
```
twtaylorbitgo, line 108: I don't think we allow valueless transfers, please add a check here so it's > 0

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -0,0 +1,124 @@
...
+    return this.transaction;
+  }
+
+  private buildTransferData(): proto.ITransferList {
+    return {
+      accountAmounts: [
+        { accountID: this.buildAccountData(this._source.address), amount: Long.fromString(this._amount).negate() }, // sender
+        { accountID: this.buildAccountData(this._toAddress), amount: Long.fromString(this._amount) }, // recipient
+      ],
+    };
```
twtaylorbitgo, line 35: let's enforce these amounts to be negative for the sender and positive for the receiver

twtaylorbitgo:
This generally LGTM - if a follow-up ticket can be created, we can land this as-is as long as these two comments above are cleaned up in subsequent PRs.

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: For ethereum Mark confirmed that within account-lib transactions could have an amount of zero. We even added a [ticket ](https://simtlix.atlassian.net/browse/BGA-206) to support that. Could @argjv  and @marktoda  confirm if this is correct? Maybe it was something specific for the smart contract.

marktoda: That is correct for ETH and ETHlike coins, but I do not know about hbar

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: Sure we created this [ticket ](https://simtlix.atlassian.net/browse/BGA-304) to follow up these coments

twtaylorbitgo: thanks

modules/account-lib/src/coin/hbar/transferBuilder.ts
twtaylorbitgo: The only things I could think of would be erc20 txes (which I don't know how they work) or AccountCreateTransaction (which could be 0-valued). I think the risk is minimal for now

PR #837, opened 2020-08-03
Default to native segwit change addresses for testnet

Xekyo:
Default to p2wsh change addresses on tbtc/tltc

    If the `addressType` is not specified explicitly by the user, we will
    default to native segwit (p2wsh) addresses for change outputs on tbtc
    and tltc. After a period of testing, we will enable the same for btc and
    ltc.

Ticket: BG-23508

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -846,11 +846,11 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   }
 
   /**
-   * Indicates whether a coin defaults to native segwit change outputs
-   * @returns {boolean}
+   * Set a default address format for change addresses, which overrides the default address format for that coin
+   * @returns {string}
    */
-  defaultsToP2wshChange() {
-    return false;
+  defaultChangeAddressType(): string {
+    return this.supportsP2shP2wsh() ? `p2shP2wsh` : `p2sh`;
   }
 
   /**
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -846,6 +846,14 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   }
 
   /**
+   * Indicates whether a coin defaults to native segwit change outputs
+   * @returns {boolean}
+   */
+  defaultsToP2wshChange() {
+    return false;
+  }
+
+  /**
    * TODO(BG-11487): Remove addressType, segwit, and bech32 params in SDKv6
    * Generate an address for a wallet based on a set of configurations
    * @param params.addressType {string}   Deprecated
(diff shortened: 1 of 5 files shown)


modules/core/src/v2/wallet.ts
```diff
@@ -1625,9 +1625,16 @@ export class Wallet {
     const self = this;
     return co<PrebuildTransactionResult>(function *() {
       // Whitelist params to build tx
-      const whitelistedParams = _.pick(params, self.prebuildWhitelistedParams());
+      let whitelistedParams = _.pick(params, self.prebuildWhitelistedParams());
```
OttoAllmendinger, line 1628: should the default be set in `prebuildWhitelistedParams()` perhaps? we leak coin-specific code here that could be avoided

tylerlevine: It's a good idea, but I think this just returns the property names to take from the params object and can't actually set any defaults. Would be good to not have coin specific code in here, but it would require a bit of refactoring.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -845,6 +845,14 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     return false;
   }
 
+  /**
+   * Indicates whether a coin defaults to native segwit change outputs
+   * @returns {boolean}
+   */
+  defaultsToP2wshChange() {
```
tylerlevine, line 852: Maybe call this `defaultChangeAddressType()` which returns an element of some `UtxoAddressType` enum? I'm not sure if we already have an enum like that, but if not I'd be fine with defining one.

modules/core/src/v2/wallet.ts
OttoAllmendinger: ah yes of course

weird that we made it this far without having a func providing coin-specific defaults here

let's not refactor in this PR imo

OttoAllmendinger:
> I would tend to use the types already defined in bitgo-utxo-lib to that end, would that be good?

I don't think so, `bitgo-utxo-lib` should move towards `bitcoinjs-lib` not away from it. We already have this enum in `@bitgo/unspents` though: https://github.com/BitGo/unspents/blob/master/src/codes.ts#L11-L19

modules/core/src/v2/coins/tbtc.ts
```diff
@@ -22,4 +22,8 @@ export class Tbtc extends Btc {
   getFullName() {
     return 'Testnet Bitcoin';
   }
+
+  defaultChangeAddressType() {
+    return this.supportsP2wsh() ? `p2wsh` : this.supportsP2shP2wsh() ? `p2shP2wsh` : `p2sh`;
```
OttoAllmendinger, line 27: `return "p2wsh";` would be cleaner here (the commit message suggests that it is equivalent)

Xekyo:
Upon further conferring with Tyler, we are moving the default native segwit change address functionality to platform. We will gate per SDK version there.

PR #838, opened 2020-08-04
Insert external HBAR signatures

marktoda:
Sometimes we need to be able to generate signatures outside of the
account-lib library. In this case we need to be able to insert these
signatures into a transaction and re-serialize it accordingly. This
commit adds this functionality to the hedera transaction builder

Ticket: BG-23598

### DIFF
--- a/modules/account-lib/src/coin/hbar/transactionBuilder.ts
+++ b/modules/account-lib/src/coin/hbar/transactionBuilder.ts
@@ -110,6 +110,15 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
    * @returns This transaction builder
    */
   signature(signature: string, keyPair: KeyPair): this {
+    // if we already have a signature for this key pair, just update it
+    for (const oldSignature of this._signatures) {
+      if (oldSignature.keyPair.getKeys().pub === keyPair.getKeys().pub) {
+        oldSignature.signature = signature;
+        return this;
+      }
+    }
+
+    // otherwise add the new signature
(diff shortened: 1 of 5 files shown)


modules/account-lib/src/coin/hbar/transactionBuilder.ts
```diff
@@ -99,6 +102,18 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
   }
 
+  /**
+   * Set an external transaction signature
+   *
+   * @param signature Hex encoded signature string
+   * @param keyPair The public key keypair that was used to create the signature
+   * @returns This transaction builder
+   */
+  signature(signature: string, keyPair: KeyPair): this {
```
twtaylorbitgo, line 112: could we make this `addSignature`

modules/account-lib/src/coin/hbar/transactionBuilder.ts
```diff
@@ -99,6 +102,18 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+    this._signatures.push({ signature, keyPair });
```
argjv, line 113: We should probably check the signature doesn't exist already for the keyPair pub. If it does, we can just replace it.

modules/account-lib/src/coin/hbar/transactionBuilder.ts
argjv: We are not adding verbs into the builder methods names. `signature` adheres to that convention.

PR #842, opened 2020-08-07
add get builder from implementation

FabianSimtlix:
This PR add a from method in TransactionBuilderFactory that returns the proper builder for that transaction type already initialized.

### DIFF
--- a/modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
+++ b/modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
@@ -1,7 +1,11 @@
 import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
+import { InvalidTransactionError } from '../baseCoin/errors';
 import { BaseTransactionBuilderFactory } from '../baseCoin';
 import { WalletInitializationBuilder } from './walletInitializationBuilder';
 import { TransferBuilder } from './transferBuilder';
+import { TransactionBuilder } from './transactionBuilder';
+import { Transaction } from './transaction';
+import { toUint8Array } from './utils';
 
 export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
   constructor(_coinConfig: Readonly<CoinConfig>) {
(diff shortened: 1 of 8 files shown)


modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
```diff
@@ -1,20 +1,81 @@
...
+  from(raw: Uint8Array | string): TransactionBuilder {
+    this.validateRawTransaction(raw);
+    const tx = this.parseTransaction(raw);
+    switch (tx.txBody.data) {
+      case 'cryptoTransfer':
+        return this.getTransferBuilder(tx);
+      case 'cryptoCreateAccount':
+        return this.getWalletInitializationBuilder(tx);
+      default:
+        throw new InvalidTransactionError('Invalid transaction');
```
twtaylorbitgo, line 35: it's possible we feed this entirely legit transactions and would like to know what type I'd think. maybe let's try to give it the `txBody.data` in the throw?

modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
```diff
@@ -1,20 +1,81 @@
...
+  }
+
+  /**
+   * Initialize the builder with the given transaction
+   *
+   * @param {Transaction | undefined} tx - the transaction used to initialize the builder
+   * @param {TransactionBuilder} builder - the builder to be initialized
+   * @returns {TransactionBuilder} the builder initialized
+   */
+  private initializeBuilder<T extends TransactionBuilder>(tx: Transaction | undefined, builder: T): T {
```
twtaylorbitgo, line 46: yes! generics! awesome

modules/account-lib/src/coin/hbar/utils.ts
```diff
@@ -133,6 +132,22 @@ export function isValidAmount(amount: string): boolean {
   );
 }
 
+/**
+ * Returns whether the provided raw transaction has a valid format and type
+ *
+ * @param {any} rawTransaction - The raw transaction to be checked
+ * @returns {boolean} the validation result
+ */
+export function isValidRawTransactionFormat(rawTransaction: any): boolean {
```
twtaylorbitgo, line 141: this is a good check. it might be good to indicate this is not the native raw format, but instead bitgo's preferred format

modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
FabianSimtlix: Actually it's not posible. `txBody.data` will always be one of Cryptotranfer or CryptoCreate. We need to include this default because protobuff adds an undefined return value to all class variables

PR #845, opened 2020-08-10
Add RSK to the OVC

wesgraham:
Adds required fields for explaining/signing OVC transactions. Specifically, needs expiryTime, signature, and recipients in halfsignedTx.
Ticket: https://bitgoinc.atlassian.net/browse/BG-23403

### DIFF
--- a/modules/core/src/v2/coins/abstractEthLikeCoin.ts
+++ b/modules/core/src/v2/coins/abstractEthLikeCoin.ts
@@ -10,6 +10,7 @@ import { randomBytes } from 'crypto';
 
 import {
   BaseCoin,
+  FullySignedTransaction,
   KeyPair,
   ParsedTransaction,
   ParseTransactionOptions,
@@ -45,6 +46,8 @@ export interface TransactionPrebuild extends BaseTransactionPrebuild {
   feeInfo: EthTransactionFee;
   source: string;
   dataToSign: string;
+  nextContractSequenceId?: string;
+  expireTime?: number;
 }
 
 export interface EthTransactionFee {
@@ -59,6 +62,20 @@ export interface ExplainTransactionOptions {
   };
   feeInfo: TransactionFee;
 }
+
+export interface HalfSignedEthLikeTransaction {
+  halfSigned?: {
+    txHex?: string;
+    payload?: string;
+    txBase64?: string;
+    operationHash?: string;
+    signatures?: string[];
+    recipients?: Recipient[];
+    expiration?: number;
+  };
+}
+
+export type SignedEthLikeTransaction = HalfSignedEthLikeTransaction | FullySignedTransaction;
 
 export abstract class AbstractEthLikeCoin extends BaseCoin {
   protected readonly _staticsCoin: Readonly<StaticsBaseCoin>;
@@ -134,15 +151,35 @@ export abstract class AbstractEthLikeCoin extends BaseCoin {
   async signTransaction(
     params: EthSignTransactionOptions,
     callback?: NodeCallback<SignedTransaction>
-  ): Bluebird<SignedTransaction> {
+  ): Bluebird<SignedEthLikeTransaction> {
     const txBuilder = this.getTransactionBuilder();
     txBuilder.from(params.txPrebuild.txHex);
     txBuilder.transfer().key(new Eth.KeyPair({ prv: params.prv }).getKeys().prv!);
     const transaction = await txBuilder.build();
 
+    if (params.isLastSignature) {
+      // In this case when we're doing the second (final) signature
+      return {
+        txHex: transaction.toBroadcastFormat(),
+      };
+    }
+
+    const outputEntry = transaction.outputs[0];
+    const recipients = [
+      {
+        address: outputEntry.address,
+        amount: outputEntry.value,
+      },
+    ];
+
+    const signatures = transaction.signature;
+
     return {
       halfSigned: {
         txHex: transaction.toBroadcastFormat(),
+        signatures: signatures,
+        recipients: recipients,
+        expiration: params.txPrebuild.expireTime,
       },
     };
   }
--- a/modules/core/src/v2/coins/abstractEthLikeCoin.ts
+++ b/modules/core/src/v2/coins/abstractEthLikeCoin.ts
@@ -11,6 +11,7 @@ import { randomBytes } from 'crypto';
 import {
   BaseCoin,
   FullySignedTransaction,
+  HalfSignedTransaction,
   KeyPair,
   ParsedTransaction,
   ParseTransactionOptions,
@@ -63,14 +64,10 @@ export interface ExplainTransactionOptions {
   feeInfo: TransactionFee;
 }
 
-export interface HalfSignedEthLikeTransaction {
+export interface HalfSignedEthLikeTransaction extends HalfSignedTransaction {
   halfSigned?: {
-    txHex?: string;
-    payload?: string;
-    txBase64?: string;
-    operationHash?: string;
-    signatures?: string[];
-    recipients?: Recipient[];
+    txHex?: never;
+    recipients: Recipient[];
     expiration?: number;
   };
 }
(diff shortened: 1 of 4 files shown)


modules/core/src/v2/coins/abstractEthLikeCoin.ts
```diff
@@ -60,6 +62,24 @@ export interface ExplainTransactionOptions {
   feeInfo: TransactionFee;
 }
 
+export interface FullySignedEthLikeTransaction {
```
CapnMigraine, line 65: This has the same definition as `FullySignedTransaction`, please import that instead

modules/core/src/v2/coins/abstractEthLikeCoin.ts
```diff
@@ -60,6 +62,24 @@ export interface ExplainTransactionOptions {
...
+  txHex: string; // Transaction in any format required by each coin, i.e. in Tron it is a stringifyed JSON
+}
+
+export interface HalfSignedEthLikeTransaction {
+  halfSigned?: {
+    txHex?: string; // Transaction in any format required by each coin, i.e. in Tron it is a stringifyed JSON
```
CapnMigraine, line 71: nit: remove comment or set one specific for eth-like coins

modules/core/src/v2/coins/abstractEthLikeCoin.ts
wesgraham: > Can we add `rsk` to the [eth-like coins tests](https://github.com/BitGo/BitGoJS/blob/master/modules/core/test/v2/unit/coins/abstractEthCoin.ts#L16)?
> Then when the [tx is signed](https://github.com/BitGo/BitGoJS/blob/master/modules/core/test/v2/unit/coins/abstractEthCoin.ts#L303), you could verify that the fields exist in the response

RSK is already in eth-like coins tests I believe? (RSK's ticker == `rbtc`)

wesgraham:
> Can we add `rsk` to the [eth-like coins tests](https://github.com/BitGo/BitGoJS/blob/master/modules/core/test/v2/unit/coins/abstractEthCoin.ts#L16)?
> Then when the [tx is signed](https://github.com/BitGo/BitGoJS/blob/master/modules/core/test/v2/unit/coins/abstractEthCoin.ts#L303), you could verify that the fields exist in the response

RSK is already in eth-like coins tests I believe? (RSK's ticker == `rbtc`)

modules/core/src/v2/coins/abstractEthLikeCoin.ts
```diff
@@ -60,6 +63,20 @@ export interface ExplainTransactionOptions {
   feeInfo: TransactionFee;
 }
 
+export interface HalfSignedEthLikeTransaction {
```
marktoda, line 66: This can `extend` the `HalfSignedTransaction` interface in baseCoin

modules/core/src/v2/coins/abstractEthLikeCoin.ts
```diff
@@ -134,15 +151,36 @@ export abstract class AbstractEthLikeCoin extends BaseCoin {
...
+        amount: output.value,
+      });
+    }
+
+    const signatures = transaction.signature;
+
     return {
       halfSigned: {
         txHex: transaction.toBroadcastFormat(),
+        signatures: signatures,
```
marktoda, line 181: What are these 3 extra fields needed for now that they weren't before?

modules/core/src/v2/coins/abstractEthLikeCoin.ts
```diff
@@ -134,15 +151,36 @@ export abstract class AbstractEthLikeCoin extends BaseCoin {
...
     const transaction = await txBuilder.build();
 
+    if (params.isLastSignature) {
+      // In this case when we're doing the second (final) signature
+      return {
+        txHex: transaction.toBroadcastFormat(),
+      };
+    }
+
+    const recipients: Recipient[] = [];
```
marktoda, line 167: slightly more concise:
```
const recipients = transaction.outputs.map((output) => ({ address: output.address, amount: output.amount }))
```

modules/core/src/v2/wallet.ts
```diff
@@ -483,7 +484,7 @@ export class Wallet {
       'feeRate', 'gasLimit', 'gasPrice', 'idfSignedTimestamp', 'idfUserId', 'idfVersion', 'instant',
       'lastLedgerSequence', 'ledgerSequenceDelta', 'maxFee', 'maxFeeRate', 'maxValue', 'memo', 'message', 'minConfirms',
       'minValue', 'noSplitChange', 'numBlocks', 'recipients', 'reservation', 'sequenceId', 'strategy',
-      'targetWalletUnspents', 'trustlines', 'type', 'unspents', 'validFromBlock', 'validToBlock',
+      'targetWalletUnspents', 'trustlines', 'type', 'unspents', 'validFromBlock', 'validToBlock', 'walletBaseAddress',
```
marktoda, line 487: What is this new param needed for?

modules/core/src/v2/coins/abstractEthLikeCoin.ts
wesgraham: Recipients & Expiration is needed for `explainTransaction`, signatures needed for `signTransaction` (tells us whether isLastSignature should be flagged)

modules/core/src/v2/wallet.ts
wesgraham: WalletBaseAddress will be changed back to walletContractAddress. This field is currently not added to downloaded txns and should be present to specify the "to" field in the final Eth Tx

modules/core/src/v2/wallet.ts
```diff
@@ -483,7 +484,7 @@ export class Wallet {
       'feeRate', 'gasLimit', 'gasPrice', 'idfSignedTimestamp', 'idfUserId', 'idfVersion', 'instant',
       'lastLedgerSequence', 'ledgerSequenceDelta', 'maxFee', 'maxFeeRate', 'maxValue', 'memo', 'message', 'minConfirms',
       'minValue', 'noSplitChange', 'numBlocks', 'recipients', 'reservation', 'sequenceId', 'strategy',
-      'targetWalletUnspents', 'trustlines', 'type', 'unspents', 'validFromBlock', 'validToBlock',
+      'targetWalletUnspents', 'trustlines', 'type', 'unspents', 'validFromBlock', 'validToBlock', 'walletContractAddress',
```
marktoda, line 487: Why does it need to even be a whitelisted prebuild param? This can be found on the wallet object at `wallet.coinSpecific.baseAddress`, so I don't see why we would ever need to explicitly pass it

wesgraham: You're right

wesgraham:
Regarding the three added fields - they were never needed to be able to generate a valid txHex. The issue is that to explain/sign the half-signed transaction in the OVC and the UI you were expected to have these fields in the `halfsigned` of your returned object. Rather than implement the logic in the OVC and the UI to read the recipients, expiry, and signature from the other fields in the tx object it was suggested to have this information returned from BitgoJS directly. Hope this makes sense

wesgraham:
> Regarding the three added fields - they were never needed to be able to generate a valid txHex. The issue is that to explain/sign the half-signed transaction in the OVC and the UI you were expected to have these fields in the `halfsigned` of your returned object. Rather than implement the logic in the OVC and the UI to read the recipients, expiry, and signature from the other fields in the tx object it was suggested to have this information returned from BitgoJS directly. Hope this makes sense

After another review signatures is the only one we can remove. We need recipients and expiryTime

modules/core/src/v2/wallet.ts
```diff
@@ -99,6 +99,7 @@ export interface PrebuildTransactionOptions {
       pendingApprovalId?: string;
     };
     offlineVerification?: boolean;
+    walletContractAddress?: string;
```
marktoda, line 102: can't you remove it from here too for the same reason as the whitelist?

modules/core/src/v2/wallet.ts
```diff
@@ -1623,6 +1624,7 @@ export class Wallet {
    * @param {Boolean} params.hop - Build this as an Ethereum hop transaction
    * @param {Object} params.reservation - Object to reserve the unspents that this tx build uses. Format is reservation = { expireTime: ISODateString, pendingApprovalId: String }
    * @param {String} params.walletPassphrase The passphrase to the wallet user key, to sign commitment data for Ethereum hop transactions
+   * @param {String} params.walletContractAddress - The contract address used as the "to" field of a transaction
```
marktoda, line 1627: and here

modules/core/src/v2/wallet.ts
```diff
@@ -1662,6 +1664,9 @@ export class Wallet {
       delete prebuild.wallet;
       delete prebuild.buildParams;
       prebuild = _.extend({}, prebuild, { walletId: self.id() });
+      if (self._wallet && self._wallet.coinSpecific && !params.walletContractAddress) {
+        prebuild = _.extend({}, prebuild, { walletContractAddress: self._wallet.coinSpecific.baseAddress });
```
marktoda, line 1668: also maybe here?

PR #847, opened 2020-08-10
BG-22317: Implement Block explorer interface and integrate blockchair APIs for BTC Recovery

glowang:
Two things to accomplish:
1. define a common interface that al UTXO block explorers should implement
2. inegrate blockchair's BTC api using the newly defined interface

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -10,6 +10,7 @@ import { UnspentType } from '@bitgo/unspents/dist/codes';
 
 import { hdPath } from '../../bitcoin';
 import { BitGo } from '../../bitgo';
+import { InvalidAddressError } from '../../errors';
 import {
   BaseCoin, AddressCoinSpecific,
   ExtraPrebuildParamsOptions, KeychainsTriplet,
@@ -203,6 +204,9 @@ export interface AddressInfo {
 
 export interface UnspentInfo {
   address: string;
+  amount: number;
+  txid: string;
+  n: number;
 }
 
 export interface RecoverParams {
@@ -214,6 +218,14 @@ export interface RecoverParams {
   ignoreAddressTypes: string[];
   bitgoKey: string;
   walletPassphrase?: string;
+}
+
+export interface UTXOBlockExplorer {
+  baseUrl: string;
+  getAddressUnspent(string): Bluebird<UnspentInfo[]>;
+  //getTxInfo: (string)=>string; // get the full copy of tx with an id
+  // getAddressInfo(addressBase58: string): AddrInfo;
+  //decodeTx: (string)=>string;
 }
 
 export abstract class AbstractUtxoCoin extends BaseCoin {
@@ -233,6 +245,34 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     return this._network;
   }
 
+  static blockExplorer: UTXOBlockExplorer = {
+    baseUrl: "https://api.blockchair.com",
+
+    async getAddressUnspent (addressBase58: string): Bluebird<UnspentInfo[]> {
+      // make a call to blockchair apis
+      const url = `${AbstractUtxoCoin.blockExplorer.baseUrl}/${this.getFullName()}/outputs?addresses=${addressBase58}`;
+      try {
+        const res = await request.get(url);
+        if (res.hasOwnProperty('data')) {
+          const unspentInfos: UnspentInfo[] = res.data.map(function processUnspent(unspent) {
+            return {
+              address: addressBase58,
+              amount: unspent.output_total,
+              txid: unspent.hash,
+              n: unspent.index,
+            }
+          });
+          return unspentInfos;
+        }
+      } catch (e) {
+
+      }
+
+
+      // TODO throw this error after confirming the old error message
+      throw new Error(`something unexpected happened while getting unspents for address ${addressBase58}`);   // TODO make this 400 and an api error
+    },
+  }
   static get validAddressTypes(): UnspentType[] {
     const validAddressTypes: UnspentType[] = [];
     // best way I could find to loop over enum values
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -216,6 +220,14 @@ export interface RecoverParams {
   walletPassphrase?: string;
 }
 
+export interface UTXOBlockExplorer {
+  baseUrl: string;
+  getAddressUnspent(string): Bluebird<UnspentInfo[]>;
```
glowang, line 225: Here we list out required methods that a block explorer must implement.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -233,6 +245,34 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     return this._network;
   }
 
+  static blockExplorer: UTXOBlockExplorer = {
+    baseUrl: "https://api.blockchair.com",
+
+    async getAddressUnspent (addressBase58: string): Bluebird<UnspentInfo[]> {
+      // make a call to blockchair apis
+      const url = `${AbstractUtxoCoin.blockExplorer.baseUrl}/${this.getFullName()}/outputs?addresses=${addressBase58}`;
```
glowang, line 253: Blockchair's api is designed such that all UTXO chains can call this URL, and they just have to substitute their chain names in.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -203,6 +204,9 @@ export interface AddressInfo {
 
 export interface UnspentInfo {
   address: string;
+  amount: number;
+  txid: string;
+  n: number;
```
Xekyo, line 209: The preferred term to refer to the position of an output on a transaction is `vout`. I would suggest that we use that instead of `n` unless we are committed already.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -216,6 +220,14 @@ export interface RecoverParams {
   walletPassphrase?: string;
 }
 
+export interface UTXOBlockExplorer {
+  baseUrl: string;
+  getAddressUnspent(string): Bluebird<UnspentInfo[]>;
+  //getTxInfo: (string)=>string; // get the full copy of tx with an id
+  // getAddressInfo(addressBase58: string): AddrInfo;
+  //decodeTx: (string)=>string;
```
Xekyo, line 228: I think you missed removing dead comments here ^

modules/core/src/v2/coins/abstractUtxoCoin.ts
Xekyo: Note that blockchair does not have BTG. (Yet another reason we should remove BTG. ;))

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -233,6 +245,34 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+              amount: unspent.output_total,
+              txid: unspent.hash,
+              n: unspent.index,
+            }
+          });
+          return unspentInfos;
+        }
+      } catch (e) {
+
+      }
```
Xekyo, line 269: This empty catch block will hide any errors you ever produce here. Either you should specify which error you want to ignore as benign and throw any others, or remove the try-catch altogether, so that errors are visible.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -233,6 +245,34 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
 
+  static blockExplorer: UTXOBlockExplorer = {
+    baseUrl: "https://api.blockchair.com",
+
+    async getAddressUnspent (addressBase58: string): Bluebird<UnspentInfo[]> {
+      // make a call to blockchair apis
+      const url = `${AbstractUtxoCoin.blockExplorer.baseUrl}/${this.getFullName()}/outputs?addresses=${addressBase58}`;
+      try {
+        const res = await request.get(url);
+        if (res.hasOwnProperty('data')) {
```
Xekyo, line 256: Is there an `else` missing here? What should happen if the condition is not met?

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -233,6 +245,34 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+          });
+          return unspentInfos;
+        }
+      } catch (e) {
+
+      }
+
+
+      // TODO throw this error after confirming the old error message
+      throw new Error(`something unexpected happened while getting unspents for address ${addressBase58}`);   // TODO make this 400 and an api error
```
Xekyo, line 273: Oh, I think you just have an extraneous brace above. Because the `catch` is closed already, I would expect this to always throw.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1519,6 +1559,17 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     return Bluebird.reject(new Error('AbtractUtxoCoin method not implemented'));
   }
 
+  /**
+   * The full name of the coin we should use while making quries to blockchair
+   * Blockchair uses slightly different naming conventions for some of the utxo coins
+   * Hence, some utxo coins should override this function and output the accurate
+   * string
+   * @returns stringg
```
Xekyo, line 1567:
```suggestion
   * @returns {string}
```

Jsdoc types should be enclosed in braces: `{string}`

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1519,6 +1559,17 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+   */
+  getBlockchairCoinName(): string {
```
Xekyo, line 1569: I think this method is unused.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -233,6 +245,34 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     return this._network;
   }
 
+  static blockExplorer: UTXOBlockExplorer = {
```
twtaylorbitgo, line 248: static shouldn't be in an abstract class - please add this to a concrete implementation or break it out in its own file. in general all this logic should not live in abstract classes.

modules/core/src/v2/coins/abstractUtxoCoin.ts
argjv: Those are for further reference, but they should have an intro line on top saying that.

glowang:
Latest is in https://github.com/BitGo/BitGoJS/pull/849

PR #848, opened 2020-08-11
Build browser versions with just webpack

bitgoAaron:
BitGoJS & account-lib will no longer use Browserify as a pre-build step before running webpack, instead it will just use webpack to create a viable browser version of BitGoJS. This allows Webpack to properly shim browser compatible modules into the SDK, where before it couldn't when working with an intermediate file (i.e `BitGoJS.browserify.js`) instead of the compiled source itself (`index.js`). 

Browser Unit tests have also been added as a separate build step, verifying that the minified bundle created for browsers works as expected, with one simple smoke test written for each coin.


modules/core/karma.conf.js
```diff
@@ -10,30 +13,44 @@ module.exports = function(config) {
...
+      'test/browser/**/*.ts': ['karma-typescript'] ,
     },
-
+    karmaTypescriptConfig: {
+      bundlerOptions: {
+          entrypoints: /\.spec\.ts$/,
+      },
+      compilerOptions: {
+          sourceMap: true,
+          target: 'ES5',
```
twtaylorbitgo, line 48: question - shouldn't we target what we're supporting? node8+? otherwise we get some weird divergence in JS/browser/node versions in tests

modules/core/test/browser/browser.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+    it("XRP should work", () => {
+        coin = bitgo.coin('xrp');
+        expect(coin).toBeTruthy()
+        expect(coin.type).toEqual('xrp');
+    });
+    it("ZEC should work", () => {
+        coin = bitgo.coin('zec');
+        expect(coin).toBeTruthy()
+        expect(coin.type).toEqual('zec');
+    });
```
twtaylorbitgo, line 221: can we actually add a function to draw from the statics coinmap? and then maybe add a decorator that's like "good for browsering" - this seems like it'll get forgotten about as we onboard new coins

modules/core/karma.conf.js
bitgoAaron: eh, for these browser tests it doesn't really matter since we're only using these compilerOptions to standup the karma test. But sure, i went ahead & changed to use the same target & libs we use in microservices (es6).

modules/core/test/browser/browser.spec.ts
bitgoAaron: ACK

modules/core/karma.conf.js
twtaylorbitgo: awesome

PR #849, opened 2020-08-12
Bg 22317 Remove use of smartbit and use blockchair for recovery

glowang:
Smartbit APIs are unstable, so we would like to switch to an API provider that is stable and extensible. 
Blockchair has the same API contracts for many UTXO coins, which we can take advantage of as backup block explorers. 

In this PR:
- use blockchair for `getUnspentInfoFromExplorer` 
- use blockchair for `getAddressInfoFromExplorer`
- still use smartbit from `decodeTx`, but have a more elegant fallback if smartbit happens to be offline again:
    - blockchair doesn't support txhex decoding
    - we restored the use of smartbit for txhex decoding rather than introducing yet another block explorer
    - some of the coin doesn't have decode tx anyways but they don't throw an error. Hence, we shouldn't throw an error just because smartbit happens to be offline. Instead, we should convey to users that they would have to do the txdecoding and verification by themselves, and we should still produce to them a recovery transaction that they can verify later.

In the future,

We should move away from the 1 block explorer model and have a general fallback. To do this, we would have to refactor more of the codebase and introduce a block explorer adaptor model.


modules/core/src/v2/coins/btc.ts
```diff
@@ -67,17 +67,19 @@ export class Btc extends AbstractUtxoCoin {
...
     return co(function *getAddressInfoFromExplorer() {
-      const addrInfo = yield request.get(self.recoveryBlockchainExplorerUrl(`/address/${addressBase58}`)).result();
-
-      addrInfo.txCount = addrInfo.address.total.transaction_count;
-      addrInfo.totalBalance = addrInfo.address.total.balance_int;
+      const addrInfo = yield request.get(self.recoveryBlockchainExplorerUrl(`/dashboards/address/${addressBase58}`)).result();
+      // addrInfo.txCount = addrInfo.address.total.transaction_count;
+      // addrInfo.totalBalance = addrInfo.address.total.balance_int;
+      addrInfo.txCount = addrInfo.data[addressBase58].address.transaction_count;
+      addrInfo.totalBalance = addrInfo.data[addressBase58].address.balance;
```
OttoAllmendinger, line 82: remove commented code 

add `assert()` that `addressBase58` exists in `addrInfo.data`

modules/core/src/v2/coins/btc.ts
```diff
@@ -67,17 +67,19 @@ export class Btc extends AbstractUtxoCoin {
   }
 
   recoveryBlockchainExplorerUrl(url: string): string {
-    // TODO: BG-23161 - replace smartbit block explorer which is now permanently down
-    throw new Error('btc recovery under maintenance');
+    return common.Environments[this.bitgo.getEnv()].blockchairBaseUrl + url;
   }
 
+  // using blockchair api: https://blockchair.com/api/docs#link_300
+  // https://api.blockchair.com/{:btc_chain}/dashboards/address/{:address}₀
```
OttoAllmendinger, line 74: that's more of an implementation comment than a jsdoc and should therefore live inside the func (where you make the request)

OttoAllmendinger: the commit message could also use a reference to the docs

modules/core/src/v2/coins/btc.ts
```diff
@@ -85,17 +85,23 @@ export class Btc extends AbstractUtxoCoin {
     }).call(this);
   }
 
+  // using blockchair api: https://blockchair.com/api/docs#link_300
+  // https://api.blockchair.com/{:btc_chain}/dashboards/address/{:address}₀
+  // example utxo from response:
+  // {"block_id":-1,"transaction_hash":"cf5bcd42c688cb7c55b5811645e7f0d2a000a85564ca3d6b9fc20f57e14b30bb","index":1,"value":558},
```
OttoAllmendinger, line 91: implementation comment should go inside func

modules/core/src/v2/coins/btc.ts
```diff
@@ -85,17 +85,23 @@ export class Btc extends AbstractUtxoCoin {
...
-      const unspents = unspentInfo.unspent;
+      const unspents = unspentInfo.data[addressBase58].utxo;
 
-      unspents.forEach(function processUnspent(unspent) {
-        unspent.amount = unspent.value_int;
+      unspents.forEach((unspent) => {
+        unspent.amount = unspent.value;
+        unspent.n = unspent.index;
+        unspent.txid = unspent.transaction_hash;
+        unspent.address = addressBase58;
```
OttoAllmendinger, line 103: it would be a bit nicer if we used `.map()` so that we don't have duplicate attributes

also we should probably check the values (it's a shame that we don't have an unspent type)

modules/core/test/v2/lib/recovery-nocks.ts
```diff
@@ -7,7 +7,777 @@ import * as nock from 'nock';
 import { Environment, Environments } from '../../../src/v2/environments';
 
 export function nockBtcRecovery(bitgo, isKrsRecovery) {
-  throw new Error('TODO: BG-23161 - replace smartbit block explorer which is now permanently down');
+  //throw new Error('TODO: BG-23161 - replace smartbit block explorer which is now permanently down');
```
OttoAllmendinger, line 10: remove commented code

modules/core/test/v2/lib/recovery-nocks.ts
```diff
@@ -7,7 +7,777 @@ import * as nock from 'nock';
...
+    : '010000000174eda73749d65473a8197bac5c26660c66d60cc77a751298ef74931a478382e100000000fdfd00004730440220513ff3a0a4d72230a7ca9b1285d5fa19669d7cccef6a9c8408b06da666f4c51f022058e8cc58b9f9ca585c37a8353d87d0ab042ac081ebfcea86fda0da1b33bf474701483045022100e27c00394553513803e56e6623e06614cf053834a27ca925ed9727071d4411380220399ab1a0269e84beb4e8602fea3d617ffb0b649515892d470061a64217bad613014c69522102f5ca5d074093abf996278d1e82b64497333254c786e9a69d34909a785aa9af32210239125d1a21ba8ae375cd37a92e48700cbb3bc1b1268d3c3f7e1d95f42155e1a821031ab00568ea1522a55f277699110649f3b8d08022494af2cc475c09e8a43b3a3a53aeffffffff012c717b000000000017a914c39dcc27823a8bd42cd3318a1dac8c25789b7ac78700000000';
+  // /dashboards/address/${addressBase58}
+  nock(blockchairURL)
+    .get('/dashboards/address/2MztRFcJWkDTYsZmNjLu9pBWWviJmWjJ4hg')
+    .reply(200, {
+      data: {
```
OttoAllmendinger, line 122: let's move this to static json files (https://github.com/BitGo/BitGoJS/tree/master/modules/core/test/v2/fixtures)

modules/core/src/v2/coins/btc.ts
glowang: fixed

modules/core/src/v2/coins/btc.ts
glowang: fixed!

modules/core/src/v2/coins/btc.ts
glowang: thanks, done!

modules/core/test/v2/lib/recovery-nocks.ts
glowang: fixed in the later commit. thanks!

glowang:
Addressed comments.

modules/core/src/v2/coins/btc.ts
glowang: I changed the forEach to map. As for value checking, we have a very loose interface `UnspentInfo` right now, which only requires the property `address` to be present. I think in the future we should just tighten the interface of `UnspentInfo` in general, and we can add type checking for this tighter interface across the board.

modules/core/test/v2/fixtures/coins/recovery.ts
```diff
@@ -0,0 +1,167 @@
+module.exports.recoverBtcSegwitFixtures = function() {
```
OttoAllmendinger, line 1: not quite static json files but neither are the other files in this dir

at least they are in a separate file ;)

OttoAllmendinger (approved):

PR #850, opened 2020-08-12
Bga 316 change hard slice for keys

silalvado:
Add a method to remove prefix in private and public keys.

### DIFF
--- a/modules/account-lib/test/unit/coin/hbar/transaction.ts
+++ b/modules/account-lib/test/unit/coin/hbar/transaction.ts
@@ -7,6 +7,9 @@ import { KeyPair } from '../../../../src/coin/hbar/keyPair';
 describe('Hbar Transaction', () => {
   const coin = coins.get('thbar');
 
+  /**
+   *
+   */
   function getTransaction(): Transaction {
     return new Transaction(coin);
   }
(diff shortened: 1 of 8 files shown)


silalvado:
> This looks good overall - do we need to change the pub in ACCOUNT_1 would be my only question?

We have two accounts, one with prefix and one without prefix. We need both to test different things.

twtaylorbitgo:
That sounds good to me - can you add a comment along those lines to each account or their tests? Just so its explicit

silalvado:
> That sounds good to me - can you add a comment along those lines to each account or their tests? Just so its explicit

I think you're right. I added comments in resources file and I modified const names to identify the keys with prefix in the unit tests.

modules/account-lib/test/unit/coin/hbar/transaction.ts
```diff
@@ -7,6 +7,9 @@ import { KeyPair } from '../../../../src/coin/hbar/keyPair';
 describe('Hbar Transaction', () => {
   const coin = coins.get('thbar');
 
+  /**
+   *
+   */
```
twtaylorbitgo, line 12: NIT: Let's remove this in the next PR @gonzaalvarez70 @silalvado

PR #851, opened 2020-08-13
BGA-297 Compose transaction/transactionBuilder

gonzaalvarez70:
Account-lib refactor to integrate the hedera SDK to use their interfaces in order to build a transaction.
Removed everything from the account-lib proto implementation.


modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
```diff
@@ -26,13 +26,12 @@ export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
...
-      case 'cryptoCreateAccount':
-        return this.getWalletInitializationBuilder(tx);
-      default:
-        throw new InvalidTransactionError('Invalid transaction ' + tx.txBody.data);
+    if (tx.txBody.hasCryptotransfer()) {
+      return this.getTransferBuilder(tx);
+    } else if (tx.txBody.hasCryptocreateaccount()) {
+      return this.getWalletInitializationBuilder(tx);
+    } else {
+      throw new InvalidTransactionError('Invalid transaction');
```
ipelaez-simtlix, line 34: We should complement this message error

modules/account-lib/src/coin/hbar/ifaces.ts
```diff
@@ -22,3 +23,18 @@ export interface SignatureData {
...
+
+export interface SDKKeyPair {
+  pub: Ed25519PublicKey;
+  prv?: Ed25519PrivateKey;
+}
+
+export interface Timestamp {
+  seconds: number;
+  nanos: number;
+}
```
ipelaez-simtlix, line 35: missing space ' ' after line 35

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -26,189 +24,107 @@ export class Transaction extends BaseTransaction {
...
-      }
-    });
-
-    return transferData;
-  }
-
   //region getters & setters
-  get txBody(): proto.TransactionBody {
-    return this._txBody;
+  get txBody(): TransactionBody {
```
gonzaalvarez70, line 104: I think that this method shouldn't be a getter and we should rename it as another function with some jsDoc

modules/account-lib/test/unit/coin/hbar/transaction.ts
```diff
@@ -7,6 +7,9 @@ import { KeyPair } from '../../../../src/coin/hbar/keyPair';
 describe('Hbar Transaction', () => {
   const coin = coins.get('thbar');
 
+  /**
+   *
+   */
```
silalvado, line 12: Remove empty doc

modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
gonzaalvarez70: Sounds good!

modules/account-lib/src/coin/hbar/ifaces.ts
gonzaalvarez70: Yes, solved in next commit

modules/account-lib/test/unit/coin/hbar/transaction.ts
gonzaalvarez70: It looks like the auto-lint is adding these lines to any declared function. We are having the same issue on every transaction test class (if we check celo, etc, or any other we have the same issue.) 
@twtaylorbitgo what do you think? Maybe we can add a cleanup ticket to clean these empty jsdoc and avoid using a function inside of the tests files.

modules/account-lib/src/coin/hbar/keyPair.ts
```diff
@@ -34,6 +35,22 @@ export class KeyPair extends Ed25519KeyPair {
     return result;
   }
 
+  /**
+   * Hedera keypair in the SDK supported format
```
argjv, line 39: s/SDK/protocol/g

modules/account-lib/src/coin/hbar/keyPair.ts
```diff
@@ -34,6 +35,22 @@ export class KeyPair extends Ed25519KeyPair {
     return result;
   }
 
+  /**
+   * Hedera keypair in the SDK supported format
+   *
+   * @returns { SDKKeyPair } the keys in SDK format
+   */
+  getSDKKeys(): SDKKeyPair {
```
argjv, line 43: This and `getKeys` are the same method. I would remove it and make any conversion needed where it is called.

argjv: Also, why do we need the `raw` option in `getKeys`?

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -26,189 +25,117 @@ export class Transaction extends BaseTransaction {
   }
 
   async sign(keyPair: KeyPair): Promise<void> {
-    const keys = keyPair.getKeys(true);
+    const keys = keyPair.getSDKKeys();
```
argjv, line 28: This should be `getKeys()`

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -26,189 +25,117 @@ export class Transaction extends BaseTransaction {
...
+      fee: new BigNumber(txBody.transactionfee).toNumber(),
       from: acc,
       startTime: time,
-      validDuration: this._txBody.transactionValidDuration!.seconds!.toString(),
-      node: stringifyAccountId(this._txBody.nodeAccountID!),
-      memo: this._txBody.memo,
+      validDuration: txBody.transactionvalidduration!.seconds.toString(),
+      node: stringifyAccountId(txBody.nodeaccountid!),
+      memo: txBody.memo,
     };
```
argjv, line 104: Is this a custom format? Don't they have a type or a way to export the tx details in a JSON format?

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -26,189 +25,117 @@ export class Transaction extends BaseTransaction {
...
-    return transferData;
-  }
-
   //region getters & setters
-  get txBody(): proto.TransactionBody {
-    return this._txBody;
-  }
 
-  get hederaTx(): proto.Transaction {
+  get hederaTx(): SDKTransaction {
```
argjv, line 119: Why do we need to expose this? Isn't this the equivalent to `toJSON()`?

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -26,189 +25,117 @@ export class Transaction extends BaseTransaction {
...
-   * Returns a hash of the given proto object.
+   * Set the inner SDK transaction as bytes
    *
-   * @param obj - The object to be hashed, it must be an proto namespace object
-   * @returns {string} - the resulting hash string
+   * @param {Uint8Array} bytes encoded SDK transaction
    */
-  private getHashOf<T>(obj: T): string {
-    return this.sha(this.encode(obj));
+  fromBytes(bytes: Uint8Array) {
```
argjv, line 137: Shouldn't this be a `constructor`? I imagine we use this when we create it a transaction.

modules/account-lib/src/coin/hbar/transactionBuilder.ts
```diff
@@ -9,29 +11,33 @@ import {
...
   protected _fee: BaseFee;
   private _transaction: Transaction;
   protected _source: BaseAddress;
-  protected _startTime: proto.ITimestamp;
+  protected _startTime: Timestamp;
   protected _memo: string;
-  protected _txBody: proto.TransactionBody;
   protected _node: HederaNode = { nodeId: '0.0.4' };
-  protected _duration: proto.Duration = new proto.Duration({ seconds: 120 });
+  protected _duration = 120;
```
argjv, line 34: Move the 120 to DEFAULT_DURATION just like DEFAULT_M

modules/account-lib/src/coin/hbar/transactionBuilder.ts
```diff
@@ -9,29 +11,33 @@ import {
...
 export const DEFAULT_M = 3;
 export abstract class TransactionBuilder extends BaseTransactionBuilder {
   protected _fee: BaseFee;
   private _transaction: Transaction;
   protected _source: BaseAddress;
-  protected _startTime: proto.ITimestamp;
+  protected _startTime: Timestamp;
   protected _memo: string;
-  protected _txBody: proto.TransactionBody;
   protected _node: HederaNode = { nodeId: '0.0.4' };
```
argjv, line 33: Move the '0.0.4' to DEFAULT_NODE_ID just like DEFAULT_M and export it

modules/account-lib/src/coin/hbar/transactionBuilder.ts
```diff
@@ -65,7 +68,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     } else {
       buffer = rawTransaction;
     }
-    tx.bodyBytes(buffer);
+    tx.fromBytes(buffer);
```
argjv, line 71: There should be a constructor so you can do `const tx = new Transaction(this._coinConfig, buffer)`

modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
```diff
@@ -64,7 +63,7 @@ export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
     } else {
       buffer = rawTransaction;
     }
-    tx.bodyBytes(buffer);
+    tx.fromBytes(buffer);
```
argjv, line 66: Ditto. This should be done with a constructor.

modules/account-lib/src/coin/hbar/keyPair.ts
gonzaalvarez70: Sure, that approach sounds good, so we decouple from the raw implementation.

modules/account-lib/src/coin/hbar/keyPair.ts
gonzaalvarez70: Sounds good

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: Yes, that sounds good

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: Yes this is a custom format including the encoded body that isn't included on the SDK JSON.
The toString() method from the SDK returns a JSON with the other fields but includes the signature map and we thought that it could be an issue.

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: It's the inner transaction and we need to expose it to alter some fields inside of the proto. Like the signatures when we add precalculated signatures instead of using the sign method.

gonzaalvarez70: Sorry, we need to expose it for the transaction builder. To retrieve the previous signatures. Maybe we could move that method here and avoid this. I'll try that approach.

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: Yes that actually sounds really good we'll try to add an alternative constructor

modules/account-lib/src/coin/hbar/transactionBuilder.ts
gonzaalvarez70: Sure

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -27,96 +35,119 @@ export class Transaction extends BaseTransaction {
...
+      memo: txBody.memo,
     };
 
-    if (this._txBody.data === 'cryptoTransfer') {
-      const [recipient, amount] = this.getTransferData();
-      result.amount = amount;
-      result.to = recipient;
+    if (txBody.cryptotransfer) {
+      txBody.cryptotransfer!.transfers!.accountamountsList.forEach(transfer => {
+        if (Long.fromValue(transfer.amount).isPositive()) {
```
twtaylorbitgo, line 134: can these be negative? or 0? we should probably throw if they are

modules/account-lib/src/coin/hbar/transferBuilder.ts
```diff
@@ -10,66 +9,49 @@ import { isValidAddress, isValidAmount, stringifyAccountId } from './utils';
...
+    if (transferData && transferData.getTransfers() && transferData.getTransfers()!.getAccountamountsList()) {
+      transferData
+        .getTransfers()!
+        .getAccountamountsList()
+        .forEach(transferData => {
+          const amount = Long.fromValue(transferData.getAmount());
+          if (amount.isPositive()) {
+            this.to(stringifyAccountId(transferData.getAccountid()!.toObject()));
+            this.amount(amount.toString());
+          }
```
twtaylorbitgo, line 52: same as here - if they are non-positive, we should throw or do something

modules/account-lib/src/coin/hbar/walletInitializationBuilder.ts
```diff
@@ -1,63 +1,62 @@
...
 
   // region Base Builder
   /** @inheritdoc */
   protected async buildImplementation(): Promise<Transaction> {
-    this._txBodyData.key = { thresholdKey: this.buildOwnersKeys() };
-    this._txBodyData.initialBalance = 0;
+    this._cryptoBuilder
+      .setInitialBalance(0)
+      .setKey(this.buildOwnersKeys())
+      .setAutoRenewPeriod(7890000);
```
twtaylorbitgo, line 25: hm where is this variable coming from? we could use a constant for this

modules/account-lib/src/coin/hbar/walletInitializationBuilder.ts
```diff
@@ -1,63 +1,62 @@
...
-    return this._owners.reduce((tKeys, key) => {
-      if (tKeys.keys && tKeys.keys.keys) {
-        tKeys.keys.keys.push({
-          ed25519: toUint8Array(new KeyPair({ pub: key }).getKeys(rawKeys).pub),
-        });
-      }
-      return tKeys;
-    }, new proto.ThresholdKey({ threshold: 2, keys: { keys: [] } }));
+  private buildOwnersKeys(): ThresholdKey {
+    const threshold = new ThresholdKey(2);
```
twtaylorbitgo, line 37: let's try to make this 2 part of the m of n in a constant at least

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -27,96 +35,119 @@ export class Transaction extends BaseTransaction {
...
-    sigPair.pubKeyPrefix = toUint8Array(key.getKeys(true).pub);
-    sigPair.ed25519 = toUint8Array(signature);
+    const sigPair = new SignaturePair();
+    sigPair.setPubkeyprefix(toUint8Array(key.getKeys().pub.toString(true)));
+    sigPair.setEd25519(toUint8Array(signature));
+
+    const sigMap = this._hederaTx._toProto().getSigmap() || new SignatureMap();
+
+    const innerTx = this._hederaTx._toProto();
+    sigMap.getSigpairList().push(sigPair);
```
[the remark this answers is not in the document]
gonzaalvarez70: I think it won't we should enforce that, for the other sign method the SDK will check.

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -1,10 +1,10 @@
 import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
-import { hash } from '@stablelib/sha384';
 import BigNumber from 'bignumber.js';
-import { Writer } from 'protobufjs';
-import * as nacl from 'tweetnacl';
+import { Transaction as SDKTransaction } from '@hashgraph/sdk';
+import { TransactionBody } from '@hashgraph/sdk/lib/generated/TransactionBody_pb';
+import { SignatureMap, SignaturePair } from '@hashgraph/sdk/lib/generated/BasicTypes_pb';
```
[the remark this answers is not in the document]
gonzaalvarez70: For what we saw they are not generated every time since they are included in the SDK, probably with a version upgrade for that SDK they could be generated again. But what we saw is that the built version of the SDK includes those generated files.

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: If it's negative that would be the transfer recipient. They could be zero but we haven't tried that. In that case we wouldn't be able to tell which is the sender and which is the recipient.

gonzaalvarez70: If there is no positive amount then the transaction is probably wrong because every transaction should have the same amount of coin input and output (i.e. if there is a 5 negative, there should be a 5 positive)

modules/account-lib/src/coin/hbar/walletInitializationBuilder.ts
gonzaalvarez70: Sure, done

modules/account-lib/src/coin/hbar/walletInitializationBuilder.ts
gonzaalvarez70: This is a constant that we got from the hedera sdk

modules/account-lib/src/coin/hbar/transaction.ts
gonzaalvarez70: After further research we've found that the best alternative for this would be to look into the transfers for an account different from the source account (the account included in the tx ID) and set that as the transfer recipient.

modules/account-lib/src/coin/hbar/transferBuilder.ts
gonzaalvarez70: We are taking the same approach here, checking that the account is different from the source account.

PR #852, opened 2020-08-13
Add gasPrice param to add wallet

wesgraham:
Adds gasPrice param to add wallet
Ticket: [BG-23720](https://bitgoinc.atlassian.net/browse/BG-23720)

### DIFF
--- a/modules/core/test/v2/unit/wallets.ts
+++ b/modules/core/test/v2/unit/wallets.ts
@@ -50,6 +50,10 @@ describe('V2 Wallets:', function() {
       // signingKeyId is invalid
       yield wallets.add({ label: 'label', enterprise: 'enterprise', keys: [], m: 2, n: 3, signingKeyId: 1243 })
         .should.be.rejectedWith('invalid argument for signingKeyId - valid key id string expected');
+
+      // gasPrice is a number
+      yield wallets.add({ label: 'label', enterprise: 'enterprise', keys: [], m: 2, n: 3, gasPrice: '17' })
+        .should.be.rejectedWith('invalid argument for gasPrice - number expected')
     }));
 
     it('creates a paired custodial wallet', co(function *createPairedCustodialWallet() {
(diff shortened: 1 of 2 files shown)


modules/core/test/v2/unit/wallets.ts
```diff
@@ -94,6 +98,22 @@ describe('V2 Wallets:', function() {
...
+      const ethBitGo = new TestBitGo({ env: 'mock' });
+      ethBitGo.initializeTestVars();
+      const ethWallets = ethBitGo.coin('teth').wallets();
+      nock(bgUrl)
+        .post('/api/v2/teth/wallet', function(body) {
+          body.type.should.equal('custodial');
+          body.should.not.have.property('keys');
+          body.should.not.have.property('m');
+          body.should.not.have.property('n');
+          return true;
```
marktoda, line 112: These tests are intended to check that your custom parameters are being passed correctly. To do so, add a check in this callback to ensure that the custom gas price is passed as intended

modules/core/test/v2/unit/wallets.ts
```diff
@@ -94,6 +98,23 @@ describe('V2 Wallets:', function() {
...
     }));
+
+    it('creates a wallet with custom gasPrice', co(function *createWalletCustomGasPrice() {
+      const ethBitGo = new TestBitGo({ env: 'mock' });
+      ethBitGo.initializeTestVars();
+      const ethWallets = ethBitGo.coin('teth').wallets();
+      nock(bgUrl)
+        .post('/api/v2/teth/wallet', function(body) {
+          body.type.should.equal('custodial');
+          body.gasPrice.should.equal(20000000000);
```
twtaylorbitgo, line 109: question: are you sure gasPrice isn't coming back as a string? most of our stuff is BigNumber at the API level

modules/core/test/v2/unit/wallets.ts
```diff
@@ -94,6 +98,23 @@ describe('V2 Wallets:', function() {
...
+          body.should.not.have.property('keys');
+          body.should.not.have.property('m');
+          body.should.not.have.property('n');
+          return true;
+        })
+        .reply(200, {});
+      yield ethWallets.add({ label: 'label', enterprise: 'enterprise', type: 'custodial', gasPrice: 20000000000 });
```
twtaylorbitgo, line 116: this might need to be a string if it's in WEI

modules/core/test/v2/unit/wallets.ts
tylerlevine: There's actually a slightly cleaner way to do this - set up your nock so it only matches if the body params are as you expect.

This isn't incorrect though I'm so not requesting changes for it.

PR #856, opened 2020-08-14
BGA-298 Fix issue by renaming test const

silalvado:
Fix issue in wallet initialization test by renaming constant


twtaylorbitgo:
I think actually we got this in master

gonzaalvarez70:
> I think actually we got this in master

So we should just close this PR right?

twtaylorbitgo:
Yeah, let's just close this

PR #859, opened 2020-08-18
Run unit tests on all events, fix drone config drift

tylerlevine:
Currently, the unit tests are only run on push events and not PR events.
This causes issues with the code coverage reporting.

Also, the drone jsonnet and yaml configurations had some drift introduced by
the browser testing upgrade, and this is fixed.

Ticket: BG-23895


marktoda:
As mentioned in slack, we can remove the `lernapublish` step as well if you care to do that in this change

tylerlevine:
> As mentioned in slack, we can remove the `lernapublish` step as well if you care to do that in this change

Sure, good idea

PR #864, opened 2020-08-20
Improve comments and add logging for recovery

glowang:
Ticket: BG-24071


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1626,6 +1626,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
               numSequentialAddressesWithoutTxs = 0;
 
               if (addrInfo.totalBalance > 0) {
+                console.log(`Found an address with balance: ${address.address} with balance ${addrInfo.totalBalance}`)
```
glowang, line 1629: FYI this will be piped thru to WRW, useful for debugging.

PR #865, opened 2020-08-21
Fix HBAR Sign Transaction

marktoda:
This commit fixes the sign transaction logic for HBAR and adds a test to
ensure that it continues to work in the future as more changes are made.
It also implements signMessage function, which unfortunately had to use
the algosdk library as @bitgo/account-lib does not implement signMessage
functionality

CLOSES TICKET: BG-24080

### DIFF
--- a/modules/core/src/v2/coins/hbar.ts
+++ b/modules/core/src/v2/coins/hbar.ts
@@ -5,6 +5,7 @@ import * as Bluebird from 'bluebird';
 import { CoinFamily, BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
 const co = Bluebird.coroutine;
 import * as bitgoAccountLib from '@bitgo/account-lib';
+import { generateAccountFromSeed, NaclWrapper, Seed } from 'algosdk';
 
 import {
   BaseCoin,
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/hbar.ts
```diff
@@ -179,8 +181,21 @@ export class Hbar extends BaseCoin {
    * @param message
    */
   signMessage(key: KeyPair, message: string | Buffer, callback?: NodeCallback<Buffer>): Bluebird<Buffer> {
```
argjv, line 183: Don't we have this in account-lib?

modules/core/src/v2/coins/hbar.ts
```diff
@@ -179,8 +181,21 @@ export class Hbar extends BaseCoin {
    * @param message
```
argjv, line 181: Add return type doc

modules/core/src/v2/coins/hbar.ts
argjv: If not, I'll add it here https://bitgoinc.atlassian.net/browse/BG-23920

PR #866, opened 2020-08-21
Implement sign and verify message

argjv:
Add a method to produce signatures using the ed25519 public key signature system and also to
verify them. Implemented for HBAR, but can be used for other coins too.

CLOSES TICKET: BG-23920

### DIFF
--- a/modules/account-lib/src/coin/baseCoin/ed25519KeyPair.ts
+++ b/modules/account-lib/src/coin/baseCoin/ed25519KeyPair.ts
@@ -74,4 +74,34 @@ export abstract class Ed25519KeyPair implements BaseKeyPair {
 
   /** @inheritdoc */
   abstract getKeys(): any;
+
+  /**
+   * Generates a signature for an arbitrary string with the current private key using the ed25519 public-key signature
+   * system.
+   *
+   * @param {string} message to produce a signature for
+   * @returns {Uint8Array} The signature produced for the message
+   * @throws if there is no private key
+   */
+  signMessage(message: string): Uint8Array {
+    const messageToSign = toUint8Array(Buffer.from(message).toString('hex'));
+    const { prv } = this.keyPair;
+    if (!prv) {
+      throw new Error('Missing private key');
+    }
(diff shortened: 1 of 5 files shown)


modules/account-lib/src/coin/baseCoin/ed25519KeyPair.ts
```diff
@@ -74,4 +74,34 @@ export abstract class Ed25519KeyPair implements BaseKeyPair {
...
+
+  /**
+   * Generates a signature for an arbitrary string with the current private key using the ed25519 public-key signature
+   * system.
+   *
+   * @param {string} message to produce a signature for
+   * @returns {Uint8Array} The signature produced for the message
+   * @throws if there is no private key
+   */
+  signMessage(message: string): Uint8Array {
```
marktoda, line 86: IMO the interface should either both take in uint8arrays as params and return uint8arrays as results, or take hex strings and return hex strings - its all just bytes so seems they should be represented the same way go avoid confusion IMO

modules/account-lib/src/coin/baseCoin/ed25519KeyPair.ts
```diff
@@ -74,4 +74,34 @@ export abstract class Ed25519KeyPair implements BaseKeyPair {
...
+
+  /**
+   * Verifies a message signature using the current public key.
+   *
+   * @param {string} message signed
+   * @param {Uint8Array} signature to verify
+   * @returns {boolean} True if the message was signed with the current key pair
+   */
+  verifySignature(message: string, signature: Uint8Array): boolean {
+    const messageToVerify = toUint8Array(Buffer.from(message).toString('hex'));
```
marktoda, line 103: Buffer.from(message) is going to read it as utf-8 text - is that what you intended?

marktoda:
```
> a = '1234'
'1234'
> Buffer.from(a).toString('hex')
'31323334'
```

PR #868, opened 2020-08-21
BG-24067 modify recovery function to take blockchair API key

glowang:
Users are likely to get rate limited by blockchair while doing recovery, due to the number of calls they would have to make to blockchair. We add an option for them to add api key to the calls to blockchair. There will be a follow up PR in the WRW side to make the frontend flow with the backend changes done here.

### DIFF
--- a/modules/core/src/v2/coins/btc.ts
+++ b/modules/core/src/v2/coins/btc.ts
@@ -71,33 +71,16 @@ export class Btc extends AbstractUtxoCoin {
     return common.Environments[this.bitgo.getEnv()].blockchairBaseUrl + url;
   }
 
-  /**
-   * Sanitize api key, decorate it so it is ready to be appended to an URL of an explorer
-   * @param {string} apiKey
-   * @returns {string}
-   */
-  transformExplorersApiKey(apiKey?: string): string { // TODO: make this works with multiple explorers
-    if (!apiKey) {
-      return '';
-    }
-    if (!_.isString(apiKey)) {
-      throw new Error('api key must be a valid string')
-    }
-
-    const key = `?key=${apiKey}`;
-    return key;
-  }
-
   getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
     const self = this;
     return co(function *getAddressInfoFromExplorer() {
       // we are using blockchair api: https://blockchair.com/api/docs#link_300
       // https://api.blockchair.com/{:btc_chain}/dashboards/address/{:address}₀
-      const key = self.transformExplorersApiKey(apiKey);
-      const addrInfo = yield request.get(self.recoveryBlockchainExplorerUrl(`/dashboards/address/${addressBase58}`) + key).result();
+      const addrInfo = yield request.get(self.recoveryBlockchainExplorerUrl(`/dashboards/address/${addressBase58}`))
+       .query({ key: apiKey })
+       .result();
       addrInfo.txCount = addrInfo.data[addressBase58].address.transaction_count;
       addrInfo.totalBalance = addrInfo.data[addressBase58].address.balance;
-
       return addrInfo;
     }).call(this);
   }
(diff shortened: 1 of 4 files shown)


modules/core/src/v2/coins/btc.ts
```diff
@@ -71,27 +71,46 @@ export class Btc extends AbstractUtxoCoin {
...
+  getUnspentInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
     const self = this;
     return co(function *getUnspentInfoFromExplorer() {
       // using blockchair api: https://blockchair.com/api/docs#link_300
       // https://api.blockchair.com/{:btc_chain}/dashboards/address/{:address}₀
       // example utxo from response:
       // {"block_id":-1,"transaction_hash":"cf5bcd42c688cb7c55b5811645e7f0d2a000a85564ca3d6b9fc20f57e14b30bb","index":1,"value":558},
-      const unspentInfo = yield request.get(self.recoveryBlockchainExplorerUrl(`/dashboards/address/${addressBase58}`)).result();
+      const key = self.transformExplorersApiKey(apiKey);
+      const unspentInfo = yield request.get(self.recoveryBlockchainExplorerUrl(`/dashboards/address/${addressBase58}`) + key).result();
```
OttoAllmendinger, line 113: just use `.get(url, { qs: { key } })` here 

https://github.com/request/request#requestoptions-callback

modules/core/src/v2/coins/btc.ts
```diff
@@ -71,27 +71,46 @@ export class Btc extends AbstractUtxoCoin {
     return common.Environments[this.bitgo.getEnv()].blockchairBaseUrl + url;
   }
 
-  getAddressInfoFromExplorer(addressBase58: string): Bluebird<any> {
+  /**
+   * Sanitize api key, decorate it so it is ready to be appended to an URL of an explorer
+   * @param {string} apiKey
+   * @returns {string}
+   */
+  transformExplorersApiKey(apiKey?: string): string { // TODO: make this works with multiple explorers
```
OttoAllmendinger, line 79: Some explorers might require passing the api key in the request header. 

If we want to really support multiple explorer backends we need to do this at a higher level and have an interface like `new SomeExplorer(apiKey).getUnspents(address)`

modules/core/test/v2/unit/recovery.ts
```diff
@@ -195,8 +195,23 @@ describe('Recovery:', function() {
...
+        bitgoKey: 'xpub661MyMwAqRbcGsSbYgWmr9G1dFgPE8HEb1ASRShbw9S1Mmu1dTQ7QStNwpaYFESq3MeKivGidN8twMeJzqh1veuSP1t2XLENL3mwpatfTst',
+        walletPassphrase: TestBitGo.V2.TEST_WALLET1_PASSCODE,
+        recoveryDestination: '2NB5Ynem6iNvA6GBLZwRxwid3Kui33729Nw',
+        scan: 5,
+        ignoreAddressTypes: ['p2wsh', 'p2shP2wsh'],
+        apiKey: 'my_______SecretApiKey',
+      });
+      recovery.transactionHex.should.equal('010000000174eda73749d65473a8197bac5c26660c66d60cc77a751298ef74931a478382e100000000fdfd00004730440220513ff3a0a4d72230a7ca9b1285d5fa19669d7cccef6a9c8408b06da666f4c51f022058e8cc58b9f9ca585c37a8353d87d0ab042ac081ebfcea86fda0da1b33bf474701483045022100e27c00394553513803e56e6623e06614cf053834a27ca925ed9727071d4411380220399ab1a0269e84beb4e8602fea3d617ffb0b649515892d470061a64217bad613014c69522102f5ca5d074093abf996278d1e82b64497333254c786e9a69d34909a785aa9af32210239125d1a21ba8ae375cd37a92e48700cbb3bc1b1268d3c3f7e1d95f42155e1a821031ab00568ea1522a55f277699110649f3b8d08022494af2cc475c09e8a43b3a3a53aeffffffff012c717b000000000017a914c39dcc27823a8bd42cd3318a1dac8c25789b7ac78700000000');
+    }));
```
twtaylorbitgo, line 214: what wallet is this matching? is this one you made?

glowang: same test wallet used in the above unit tests

twtaylorbitgo: ok yeah if this is not an integration test lgtm

PR #870, opened 2020-08-21
Bga 320 validate signature helper

ipelaez-simtlix:
We add validate signature helper for BGA-320


modules/account-lib/src/coin/hbar/utils.ts
```diff
@@ -182,3 +183,35 @@ export function removePrefix(prefix: string, key: string): string {
...
+ * @returns {boolean} the result of the validation
+ */
+export function validate(encodedTx: string, key: KeyPair, signature: string): boolean {
+  let wasSigned = false;
+  const signer = key.getKeys().pub.toString(true);
+  const transaction = SDKTransaction.fromBytes(toUint8Array(encodedTx));
+  const transactionData = transaction._toProto();
+
+  if (transactionData.getSigmap()) {
+    transactionData
```
twtaylorbitgo, line 203: should we check if there's 0 sigs? would that break for this?

modules/account-lib/src/coin/hbar/utils.ts
```diff
@@ -182,3 +183,35 @@ export function removePrefix(prefix: string, key: string): string {
...
+/**
+ * Validate if a hedera encoded transaction was signed by a certain
+ * signer and if the signature is the same
+ *
+ * @param {string} encodedTx a hedera encoded transaction
+ * @param {KeyPair} key the signer's keypair
+ * @param {string} signature the signature to be compared
+ * @returns {boolean} the result of the validation
+ */
+export function validate(encodedTx: string, key: KeyPair, signature: string): boolean {
```
argjv, line 196: We should probably put this in the `Transaction` object since we already have that info there

ipelaez-simtlix: That sounds good, w'll go for this approach

modules/account-lib/src/coin/hbar/utils.ts
ipelaez-simtlix: Hmm, that's a good one, i'll check about this

tylerlevine:
@ipelaez-simtlix is this PR still active or can it be closed?

tylerlevine:
Closing, since this is quite stale and now has merge conflicts.

PR #872, opened 2020-08-21
BGA-324 Update toJson method

silalvado:
Update toJson method to return a Hedera transaction object.

### DIFF
--- a/modules/account-lib/src/coin/hbar/ifaces.ts
+++ b/modules/account-lib/src/coin/hbar/ifaces.ts
@@ -6,7 +6,7 @@ import { KeyPair } from '.';
 export interface TxData {
   sigs?: SignatureList.AsObject;
   sigmap?: SignatureMap.AsObject;
-  body?: TransactionBody.AsObject;
+  body: TransactionBody.AsObject;
   bodybytes: Uint8Array | string;
   hash?: string;
 }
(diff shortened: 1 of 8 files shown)


modules/account-lib/src/coin/hbar/ifaces.ts
```diff
@@ -1,18 +1,14 @@
 import { Ed25519PublicKey, Ed25519PrivateKey } from '@hashgraph/sdk';
+import { TransactionBody } from '@hashgraph/sdk/lib/generated/TransactionBody_pb';
+import { SignatureList, SignatureMap } from '@hashgraph/sdk/lib/generated/BasicTypes_pb';
 import { KeyPair } from '.';
 
 export interface TxData {
-  id: string;
+  sigs?: SignatureList.AsObject;
+  sigmap?: SignatureMap.AsObject;
+  body?: TransactionBody.AsObject;
```
FabianSimtlix, line 9: is there a case where this could be undefined? this is forcing you to use `!` when you want to access a value from it

gonzaalvarez70: I think that it shouldn't be undefined, the sigs and sigmap could be undefined so those are ok.

silalvado: I think you're right. I've changed body field to be mandatory.

PR #873, opened 2020-08-23
Create RecoveryProvider

argjv:
The explorer calls used during recoveries are mixed with the recovery code making it hard to change.

Create an interface for 3rd party recovery info providers so that it is easier to add new ones in the future.

CLOSES TICKET: BG-24116


tylerlevine (dismissed):
Nice cleanup. I assume this is covered by the existing recovery tests? Do we need any tests for the RecoveryProvider implementations?

PR #890, opened 2020-09-10
BG-15531 support nonbitgo bsv recovery in SDK

glowang:
This PR:

- Adds a blockexplorer for BSV
- Adds functionality to recover BSV without BitGo

TODO

- Make sure BSV is replay safe

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1610,7 +1610,8 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
         .set('LTC', 'litecoin')
         .set('BCH', 'bitcoin-cash')
         .set('ZEC', 'zcash')
-        .set('DASH', 'dash');
+        .set('DASH', 'dash')
+        .set('BSV', 'bitcoin-sv');
       const coinGeckoId = familyNamesToCoinGeckoIds.get(self.getFamily().toUpperCase());
       if (!coinGeckoId) {
         throw new Error(`There is no CoinGecko id for family name ${self.getFamily().toUpperCase()}.`);
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1699,7 +1699,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
             const address: any = self.createMultiSigAddress(Codes.typeForCode(chain), 2, keys);
 
             const addrInfo: AddressInfo = yield self.getAddressInfoFromExplorer(address.address, params.apiKey);
-
+            // we use txCount here because it implies usage - having tx'es means the addr was generated and used
             if (addrInfo.txCount === 0) {
               numSequentialAddressesWithoutTxs++;
             } else {
--- a/modules/core/src/v2/coins/bsv.ts
+++ b/modules/core/src/v2/coins/bsv.ts
@@ -48,13 +48,13 @@ export class Bsv extends Bch {
     return common.Environments[this.bitgo.getEnv()].bsvExplorerBaseUrl + url;
   }
 
-  getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
-    const explorer = new BlockchairApi(this.bitgo, apiKey, 'bitcoin-sv');
-    return Bluebird.resolve(explorer.getAccountInfo(addressBase58));
+  async getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
+    const explorer = new BlockchairApi(this.bitgo, 'bitcoin-sv', apiKey);
+    return explorer.getAccountInfo(addressBase58);
   }
 
-  getUnspentInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
-    const explorer = new BlockchairApi(this.bitgo, apiKey, 'bitcoin-sv');
-    return Bluebird.resolve(explorer.getUnspents(addressBase58));
+  async getUnspentInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
+    const explorer = new BlockchairApi(this.bitgo, 'bitcoin-sv', apiKey);
+    return explorer.getUnspents(addressBase58);
   }
 }
--- a/modules/core/src/v2/coins/bsv.ts
+++ b/modules/core/src/v2/coins/bsv.ts
@@ -11,6 +11,7 @@ import { UtxoNetwork } from './abstractUtxoCoin';
 const co = Bluebird.coroutine;
 import * as common from '../../common';
 import * as errors from '../../errors';
+import { BlockchairApi } from '../recovery/blockchairApi';
 
 export class Bsv extends Bch {
   constructor(bitgo: BitGo, network?: UtxoNetwork) {
@@ -47,31 +48,13 @@ export class Bsv extends Bch {
     return common.Environments[this.bitgo.getEnv()].bsvExplorerBaseUrl + url;
   }
 
-  getAddressInfoFromExplorer(addressBase58: string): Bluebird<any> {
-    const self = this;
-    return co(function* getAddressInfoFromExplorer() {
-      // TODO BG-9989: Update this method with the correct API route and parsing once we have one
-      const addrInfo = yield request.get(self.recoveryBlockchainExplorerUrl(`/addr/${addressBase58}`)).result();
-
-      addrInfo.txCount = addrInfo.txApperances;
-      addrInfo.totalBalance = addrInfo.balanceSat;
-
-      return addrInfo;
-    }).call(this);
+  getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
+    const explorer = new BlockchairApi(this.bitgo, apiKey, 'bitcoin-sv');
+    return Bluebird.resolve(explorer.getAccountInfo(addressBase58));
   }
 
-  getUnspentInfoFromExplorer(addressBase58: string): Bluebird<any> {
-    const self = this;
-    return co(function* getUnspentInfoFromExplorer() {
-      // TODO BG-9989: Update this method with the correct API route and parsing once we have one
-      const unspents = yield request.get(self.recoveryBlockchainExplorerUrl(`/addr/${addressBase58}/utxo`)).result();
-
-      unspents.forEach(function processUnspent(unspent) {
-        unspent.amount = unspent.satoshis;
-        unspent.n = unspent.vout;
-      });
-
-      return unspents;
-    }).call(this);
+  getUnspentInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
+    const explorer = new BlockchairApi(this.bitgo, apiKey, 'bitcoin-sv');
+    return Bluebird.resolve(explorer.getUnspents(addressBase58));
   }
 }
--- a/modules/core/src/v2/coins/bsv.ts
+++ b/modules/core/src/v2/coins/bsv.ts
@@ -7,7 +7,7 @@ import * as bitcoin from '@bitgo/utxo-lib';
 const request = require('superagent');
 import * as Bluebird from 'bluebird';
 import { BaseCoin } from '../baseCoin';
-import { UtxoNetwork } from './abstractUtxoCoin';
+import { AddressInfo, UnspentInfo, UtxoNetwork } from './abstractUtxoCoin';
 const co = Bluebird.coroutine;
 import * as common from '../../common';
 import * as errors from '../../errors';
@@ -48,13 +48,13 @@ export class Bsv extends Bch {
     return common.Environments[this.bitgo.getEnv()].bsvExplorerBaseUrl + url;
   }
 
-  async getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
+  getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<AddressInfo> {
     const explorer = new BlockchairApi(this.bitgo, 'bitcoin-sv', apiKey);
-    return explorer.getAccountInfo(addressBase58);
+    return Bluebird.resolve(explorer.getAccountInfo(addressBase58));
   }
 
-  async getUnspentInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
+  getUnspentInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<UnspentInfo[]> {
     const explorer = new BlockchairApi(this.bitgo, 'bitcoin-sv', apiKey);
-    return explorer.getUnspents(addressBase58);
+    return Bluebird.resolve(explorer.getUnspents(addressBase58));
   }
 }
--- a/modules/core/src/v2/environments.ts
+++ b/modules/core/src/v2/environments.ts
@@ -15,7 +15,7 @@ interface EnvironmentTemplate {
   signingAddress: string;
   serverXpub: string;
   hsmXpub: string;
-  blockchairBaseUrl: string;
+  blockchairBaseUrl(coin: String): string;
   blockstreamBaseUrl: string;
   smartbitBaseUrl: string;
   btcExplorerBaseUrl: string;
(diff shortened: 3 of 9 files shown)


modules/core/src/v2/coins/bsv.ts
```diff
@@ -47,31 +48,13 @@ export class Bsv extends Bch {
...
-
-      addrInfo.txCount = addrInfo.txApperances;
-      addrInfo.totalBalance = addrInfo.balanceSat;
-
-      return addrInfo;
-    }).call(this);
+  getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
+    const explorer = new BlockchairApi(this.bitgo, apiKey, 'bitcoin-sv');
+    return Bluebird.resolve(explorer.getAccountInfo(addressBase58));
   }
```
OttoAllmendinger, line 54: use `async getAddressInfoFromExplorer(...)` and either `return explorer.getAccountInfo()` or `return await explorer.getAccountInfo()`

modules/core/src/v2/coins/bsv.ts
```diff
@@ -47,31 +48,13 @@ export class Bsv extends Bch {
...
-      const unspents = yield request.get(self.recoveryBlockchainExplorerUrl(`/addr/${addressBase58}/utxo`)).result();
-
-      unspents.forEach(function processUnspent(unspent) {
-        unspent.amount = unspent.satoshis;
-        unspent.n = unspent.vout;
-      });
-
-      return unspents;
-    }).call(this);
+  getUnspentInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
```
OttoAllmendinger, line 56: same

modules/core/src/v2/environments.ts
```diff
@@ -94,7 +94,9 @@ const mainnetBase: EnvironmentTemplate = {
   signingAddress: '1BitGo3gxRZ6mQSEH52dvCKSUgVCAH4Rja',
   serverXpub: hardcodedPublicKeys.serverXpub.prod,
   hsmXpub: hardcodedPublicKeys.hsmXpub.prod,
-  blockchairBaseUrl: 'https://api.blockchair.com/bitcoin',
+  blockchairBaseUrl: (coin: String) => {
```
OttoAllmendinger, line 97: use `blockchairBaseUrl(coin: string) { ... }`

better `getBlockchairBaseUrl(coin)` to highlight that it is a function

modules/core/src/v2/environments.ts
```diff
@@ -177,7 +181,9 @@ export const Environments: Environments = {
   }),
   mock: Object.assign({}, devBase, {
     uri: 'https://bitgo.fakeurl',
-    blockchairBaseUrl: 'https://api.blockchair.fakeurl/bitcoin/testnet',
+    blockchairBaseUrl: (coin: String) => {
```
OttoAllmendinger, line 184: same

modules/core/src/v2/environments.ts
```diff
@@ -219,10 +225,13 @@ export const Environments: Environments = {
...
         : 'https://api.smartbit.com.au/v1',
-    blockchairBaseUrl:
-      process.env.BITGO_CUSTOM_BITCOIN_NETWORK !== 'bitcoin'
-        ? 'https://api.blockchair.com/bitcoin/testnet'
-        : 'https://api.blockchair.com/bitcoin',
+    blockchairBaseUrl: (coin: String) => {
+      if (process.env.BITGO_CUSTOM_BITCOIN_NETWORK !== 'bitcoin') {
+        return `https://api.blockchair.com/${coin}/testnet`;
+      }
+      console.log('!!!!!!!!!!!!!!!!!!!!!', process.env.BITGO_CUSTOM_BITCOIN_NETWORK);
```
OttoAllmendinger, line 232: :scissors:

modules/core/src/v2/environments.ts
OttoAllmendinger: if it is a function it should better move out of `config` or `environments`

this is better suited for  `blockchairApi.ts` with the a signature `getBaseUrl(coin: string, testnet: boolean)` 

the `coin: String` is a bit misleading - typically that indicates our three-character shorthands `btc` `eth` etc, while here we only should allow strings like `bitcoin` `ethereum` `litecoin` etc

We should validate the argument to make sure it is one of those

modules/core/src/v2/recovery/blockchairApi.ts
```diff
@@ -6,24 +6,33 @@ import { BitGo } from '../../bitgo';
 export class BlockchairApi implements RecoveryProvider {
   protected readonly bitgo: BitGo;
   protected readonly apiToken?: string;
+  protected readonly coin: string;
 
-  constructor(bitgo: BitGo, apiToken?: string) {
+  constructor(bitgo: BitGo, apiToken?: string, coin?: string) {
```
OttoAllmendinger, line 11: let's make `coin` mandatory

modules/core/test/v2/unit/recovery/blockchairApi.ts
```diff
@@ -0,0 +1,97 @@
...
+  let apiKey;
+
+  before(() => {
+    apiKey = 'my____ApiKey';
+    bitgo = new TestBitGo({env: 'test'});
+    env = Environments[bitgo.getEnv()] as any;
+    }
+  );
+  for (const coinName of coinNames) {
+    describe('should succeed', function() {
```
OttoAllmendinger, line 64: include `coinName` in the title here

modules/core/test/v2/unit/recovery/blockchairApi.ts
```diff
@@ -0,0 +1,97 @@
...
+      it('should get address information from blockchair', async() => {
```
OttoAllmendinger, line 65: use `async function() ` instead so we can call `this.skip()` etc

modules/core/test/v2/unit/recovery/blockchairApi.ts
```diff
@@ -0,0 +1,97 @@
...
+import * as should from 'should';
+import { Environments } from '../../../../src/v2/environments';
+import { TestBitGo } from '../../../lib/test_bitgo';
+import { BlockchairApi } from '../../../../src/v2/recovery/blockchairApi';
+
+const coinNames = ['bitcoin', 'bitcoin-sv'];
+
+function nockBlockchair(env, coinName) {
+  const baseUrl = env.blockchairBaseUrl(coinName);
+  console.log('baseUrl', baseUrl)
```
OttoAllmendinger, line 11: :scissors:

OttoAllmendinger (changes requested):

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1700,7 +1701,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
 
             const addrInfo: AddressInfo = yield self.getAddressInfoFromExplorer(address.address, params.apiKey);
 
-            if (addrInfo.txCount === 0) {
+            if (addrInfo.txCount === 0) { // we use txCount here because it implies usage - having tx'es means the addr was generated and used
```
argjv, line 1704: nit: put the comment line above or below the code since it is too long

argjv (approved):

modules/core/src/v2/coins/bsv.ts
```diff
@@ -47,31 +48,13 @@ export class Bsv extends Bch {
...
-    return co(function* getAddressInfoFromExplorer() {
-      // TODO BG-9989: Update this method with the correct API route and parsing once we have one
-      const addrInfo = yield request.get(self.recoveryBlockchainExplorerUrl(`/addr/${addressBase58}`)).result();
-
-      addrInfo.txCount = addrInfo.txApperances;
-      addrInfo.totalBalance = addrInfo.balanceSat;
-
-      return addrInfo;
-    }).call(this);
+  async getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
```
tylerlevine, line 51: Return type incorrect (getAccountInfo returns a native Promise). Please wrap in `Bluebird.resolve()` to convert to a Bluebird promise.

modules/core/src/v2/coins/bsv.ts
```diff
@@ -47,31 +48,13 @@ export class Bsv extends Bch {
...
-      const unspents = yield request.get(self.recoveryBlockchainExplorerUrl(`/addr/${addressBase58}/utxo`)).result();
-
-      unspents.forEach(function processUnspent(unspent) {
-        unspent.amount = unspent.satoshis;
-        unspent.n = unspent.vout;
-      });
-
-      return unspents;
-    }).call(this);
+  async getUnspentInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<any> {
```
tylerlevine, line 56: Return type incorrect (getUnspents() returns a native Promise). Please wrap in `Bluebird.resolve()` to convert to a Bluebird promise.

modules/core/test/v2/unit/recovery/blockchairApi.ts
```diff
@@ -0,0 +1,96 @@
...
+      }
+    });
+}
+
+describe('blockchair api', function() {
+  let bitgo;
+  let env;
+  let apiKey;
+
+  before(() => {
```
tylerlevine, line 56: No need to put this in a `before()`, since you're not doing any async work - you can just assign directly in the body of the `describe`

modules/core/test/v2/unit/recovery/blockchairApi.ts
```diff
@@ -0,0 +1,96 @@
...
+        response[0].txid.should.equal('9a57cdf7a8ce94c1cdad90f639fd8dcab8d20f68a117a7c30dbf468652fbf7e0');
+        response[0].address.should.equal('2N7kMMaUjmBYCiZqQV7GDJhBSnJuJoTuBws');
+      });
+    });
+    describe(`${coinName} should fail`,  function() {
+      it('should throw if the address value is an empty string', async function() {
+        const blockchair = new BlockchairApi(bitgo, coinName);
+        let error;
+        try {
+          error = await blockchair.getUnspents('');
```
tylerlevine, line 87: `await blockchair.getUnspents('').should.be.rejectedWith('invalid address')`

tylerlevine (changes requested):
looks pretty good, just some minor comments.

👍

modules/core/src/v2/coins/bsv.ts
tylerlevine: This is bad advice. `getAddressInfoFromExplorer()` and `getUnspentInfoFromExplorer` are both part of the AbstractUtxoCoin API, which is public.

These functions are typed as follows:
```
protected abstract getAddressInfoFromExplorer(address: string, apiKey?: string): Bluebird<AddressInfo>;
protected abstract getUnspentInfoFromExplorer(address: string, apiKey?: string): Bluebird<UnspentInfo[]>;
```

Changing these signatures, or overriding these functions with a non-compatible signature should not be done unless you are intending to make breaking API changes and will increment the major version number accordingly.

Converting coroutines to async functions is a breaking change, since callers may be relying on bluebird specific methods being available on the returned promises. (eg, consider a call like `await getAddressInfoFromExplorer('someAddress').map((address) => /* do something with address */)`)

Coroutines can still be removed and replaced with async functions for code which is not exposed by the public API.

Please remove the `async` keyword from these function signatures.

OttoAllmendinger: ah I see, yes we should stay compatible then return bluebird promises

however I think we can return bluebird promises *without* using coroutines 


are those two not equivalent?


```
f: co(function *() { return y })
```

```
async f() { return Bluebird.resolve(y) }
```

http://bluebirdjs.com/docs/api/promise.resolve.html

modules/core/test/v2/unit/recovery/blockchairApi.ts
OttoAllmendinger: I agree but I need to point out one difference: if any of the calls in the `before()` block throw an exception, the test suite continues

if the it happens in `describe()` context while test suite stops

the advantage is very minor so I also think it should be outside as well

OttoAllmendinger (approved):
sorry about misleading you with the Promise/Bluebird interface

rest is good :+1:

modules/core/src/v2/coins/bsv.ts
tylerlevine: Unfortunately, they are not.
```
$ node
> const bb = require('bluebird')
undefined
> async function promTest() { return bb.resolve('resolved'); }
undefined
> const bbTest = bb.coroutine(function*() { return 'resolved'; });
undefined
> promTest().map
undefined
> bbTest().map
[Function]
```

tylerlevine: The best way I've found to interoperate between different promise types while keeping the return value a bluebird is to use the `try()` method to start the promise chain. (more info: http://cryto.net/~joepie91/blog/2016/05/11/what-is-promise-try-and-why-does-it-matter/)

glowang:
Thanks for reviewing! @tylerlevine  @OttoAllmendinger  one last time...

modules/core/src/v2/coins/bsv.ts
tylerlevine, line 51: I think we can narrow the return type here to `Bluebird<AddressInfo>` and `Bluebird<UnspentInfo[]>`, since those are the return types for the function's you're overriding.

tylerlevine (changes requested):
One more tiny thing on the types and then I think it's good to go.

modules/core/src/v2/coins/bsv.ts
glowang: ah yes, thanks for catching!

tylerlevine (approved):
Great work Gloria! Thanks for your patience and persistence.

modules/core/src/v2/coins/bsv.ts
```diff
@@ -47,31 +48,13 @@ export class Bsv extends Bch {
...
-    return co(function* getAddressInfoFromExplorer() {
-      // TODO BG-9989: Update this method with the correct API route and parsing once we have one
-      const addrInfo = yield request.get(self.recoveryBlockchainExplorerUrl(`/addr/${addressBase58}`)).result();
-
-      addrInfo.txCount = addrInfo.txApperances;
-      addrInfo.totalBalance = addrInfo.balanceSat;
-
-      return addrInfo;
-    }).call(this);
+  async getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<AddressInfo> {
```
tylerlevine, line 51: Actually sorry, one more thing. Please remove the `async` keyword here since they can only return ES6 promises, and wrapping in `Bluebird.resolve()` doesn't actually cause a bluebird promise to be returned.

Since we don't need to use `await` here, we can just get rid of the keyword and we don't need to replace with a coroutine.

tylerlevine (changes requested):
Sorry, one more thing on the async functions. I promise this is the last fix!

modules/core/src/v2/coins/bsv.ts
glowang: Thanks for pointing this out!

The bluebird doc for `resolve()` says "Create a promise that is resolved with the given value. If value is already a trusted Promise, it is returned as is. If value is not a thenable, a fulfilled Promise is returned with value as its fulfillment value. If value is a thenable (Promise-like object, like those returned by jQuery's $.ajax), returns a trusted Promise that assimilates the state of the thenable." 

Why did you say `Bluebird.resolve()` doesnt actually return a promise? Or did you mean that wrapping specifically `ES6` promise in `Bluebird.resolve()` doesn't return promises?

tylerlevine: Right, `Bluebird.resolve()` will always return a bluebird promise. But if you return that bluebird promise from an async function, it gets converted to an ES prom, due to the magic of async/await.

Maybe this code snippet will help clarify:
```
const bb = require('bluebird');

const isBBProm = (prom) => prom._bitField !== undefined;

const bbProm = bb.resolve(1);
const esProm = Promise.resolve(2);

console.log('bbProm is bluebird:', isBBProm(bbProm));
console.log('esProm is bluebird:', isBBProm(esProm));

async function returnsBBProm() {
  return bbProm;
}

async function returnsESProm() {
  return esProm;
}

console.log('returnsBBProm() is bluebird:', isBBProm(returnsBBProm()));
console.log('returnsESProm() is bluebird:', isBBProm(returnsESProm()));
```

This outputs
```
bbProm is bluebird: true
esProm is bluebird: false
returnsBBProm() is bluebird: false
returnsESProm() is bluebird: false
```

tylerlevine (approved):

modules/core/src/v2/coins/bsv.ts
glowang: Thank you so much for this! Got it :)

OttoAllmendinger: however I believe `bb.method(async function () { ... })` wraps it properly

OttoAllmendinger:
```
> (B.method(async (x) => x))(2).then(console.log)
Promise [Object] {
  _bitField: 0,
  _fulfillmentHandler0: undefined,
  _rejectionHandler0: undefined,
  _promise0: undefined,
  _receiver0: undefined }
> 2
```

PR #898, opened 2020-09-16
Use backupkeynonce if signingKeyNonce is missing

glowang:
Background: Client wants to build an unsigned sweep with WRW and double sign the transaction with the SDK `signTransaction` function.

Bug: Client encounter error because their transaction is missing `signingKeyNonce`

Solution: In Unsigned Sweep, WRW builds a Transaction JSON that provides `backupKeyNonce` but not `signingKeyNonce`. However, since the two variables are used for the same purpose, we should use either in the transaction signing process.

From the `generateExplainTransactionParams` function in OVC, we can see that the param signingKeyNonce is essentially the same as backupKeyNonce . The SDK code was built for the sake of OVC, which is why it expected the same param name that the OVC uses.

Ticket: BG-24643


modules/core/src/v2/coins/eth.ts
```diff
@@ -101,6 +101,7 @@ interface SignFinalOptions {
     halfSigned: {
       expireTime: number;
       contractSequenceId: number;
+      backupKeyNonce: number;
```
tylerlevine, line 104: Should this be optional, or is it guaranteed to always be present?

modules/core/test/v2/unit/ethWallet.ts
```diff
@@ -342,3 +342,53 @@ describe('prebuildTransaction', function() {
...
   }));
 });
+
+describe('final-sign transaction from WRW', async function() {
+  const bitgo = new TestBitGo({ env: 'test' });
+  const basecoin = bitgo.coin('teth');
+  const gasPrice = 200000000000;
+  const gasLimit =  500000;
+  const prv = 'xprv9s21ZrQH143K3D8TXfvAJgHVfTEeQNW5Ys9wZtnUZkqPzFzSjbEJrWC1vZ4GnXCvR7rQL2UFX3RSuYeU9MrERm1XBvACow7c36vnz5iYyj2'; // placeholder test prv
+  const wrwUnsignedSweepTx = {
```
tylerlevine, line 352: Maybe put this into `fixtures` instead of inline in the test suite.

modules/core/test/v2/unit/ethWallet.ts
```diff
@@ -342,3 +342,53 @@ describe('prebuildTransaction', function() {
...
+   wrapper.txPrebuild = halfSigned;
+   wrapper.txPrebuild.recipients = halfSigned.halfSigned.recipients;
+   wrapper.txPrebuild.gasPrice = gasPrice.toString();
+   wrapper.txPrebuild.gasLimit = gasLimit.toString();
+   wrapper.isLastSignature = true;
+   wrapper.walletContractAddress = wrwUnsignedSweepTx.walletContractAddress;
+   wrapper.prv = prv;
+  
+  const finalSignedTx = await basecoin.signTransaction(wrapper);
+  finalSignedTx.should.have.property('txHex');
```
tylerlevine, line 393: Are there more checks we should do on the `finalSignedTx` beyond that it has a `txHex` property? Do we need to check the transaction to make sure it's valid? What happens if we break the signing logic? Would this test catch that issue?

tylerlevine (changes requested):
Pretty good, just a few comments from me.

argjv (dismissed):
LGTM after Tyler's comments

tylerlevine (approved):
Nice work 👍

modules/core/src/v2/coins/eth.ts
```diff
@@ -518,7 +522,7 @@ export class Eth extends BaseCoin {
 
     const ethTxParams = {
       to: params.walletContractAddress,
-      nonce: params.signingKeyNonce,
+      nonce: params.signingKeyNonce || params.txPrebuild.halfSigned.backupKeyNonce,
```
marktoda, line 525: What if signingKeyNonce exists and == 0? Then we will use backupKeyNonce here

ericcrosson-bitgo: This is a real bug. If we update to TypeScript 3.7+ (currently 3.5.2, should be no breaking changes) we can use [nullish coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing), which was introduced for exactly this reason

```suggestion
      nonce: params.signingKeyNonce ?? params.txPrebuild.halfSigned.backupKeyNonce,
```

Otherwise it should read

```suggestion
      nonce: params.signingKeyNonce !== undefined ? params.signingKeyNonce : params.txPrebuild.halfSigned.backupKeyNonce,
```

marktoda: Created a ticket for this issue: https://bitgoinc.atlassian.net/browse/BG-25181

PR #904, opened 2020-09-21
Adds seed handling for Stellar to Hbar seeds

twtaylorbitgo:
Ticket: BG-23026

### DIFF
--- a/modules/account-lib/src/coin/hbar/utils.ts
+++ b/modules/account-lib/src/coin/hbar/utils.ts
@@ -193,3 +193,11 @@ export function isValidMemo(memo: string): boolean {
   }
   return true;
 }
+
+/**
+ * @param buffer
+ * @param raw
+ */
+export function fromEd25519PublicKeyBuffer(buffer: Buffer, raw = false): string {
+  return Ed25519PublicKey.fromBytes(buffer).toString(raw);
+}
--- a/modules/account-lib/src/coin/hbar/utils.ts
+++ b/modules/account-lib/src/coin/hbar/utils.ts
@@ -1,5 +1,5 @@
 import * as _ from 'lodash';
-import { Ed25519PublicKey, TransactionId, AccountId } from '@hashgraph/sdk';
+import { TransactionId, AccountId, Ed25519PublicKey, Ed25519PrivateKey } from '@hashgraph/sdk';
 import * as hex from '@stablelib/hex';
 import BigNumber from 'bignumber.js';
 import { proto } from '../../../resources/hbar/protobuf/hedera';
@@ -195,9 +195,10 @@ export function isValidMemo(memo: string): boolean {
 }
 
 /**
- * @param buffer
- * @param raw
+ * Uses the native hashgraph SDK function to get a raw key.
+ *
+ * @param prv
  */
-export function fromEd25519PublicKeyBuffer(buffer: Buffer, raw = false): string {
-  return Ed25519PublicKey.fromBytes(buffer).toString(raw);
+export function createRawKey(prv: string): Ed25519PrivateKey {
+  return Ed25519PrivateKey.fromString(prv);
 }
--- a/modules/core/src/v2/coins/algo.ts
+++ b/modules/core/src/v2/coins/algo.ts
@@ -4,6 +4,7 @@
 import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
 import * as stellar from 'stellar-sdk';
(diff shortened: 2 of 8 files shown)


modules/core/src/v2/coins/algo.ts
```diff
@@ -304,15 +306,20 @@ export class Algo extends BaseCoin {
   }
 
   isStellarSeed(seed: string): boolean {
```
tylerlevine, line 308: Wow, why do we have stellar code in algo? This is going to break when we do the modularization project. Each coin family needs to stand on it's own.

modules/core/src/v2/internal/seedValidator.ts
```diff
@@ -0,0 +1,71 @@
+import { isValidSeed as AlgoIsValidSeed } from 'algosdk';
+import * as stellar from 'stellar-sdk';
+import { Ed25519PrivateKey } from '@hashgraph/sdk';
```
tylerlevine, line 3: Yikes.

modules/account-lib/src/coin/hbar/utils.ts
```diff
@@ -193,3 +193,11 @@ export function isValidMemo(memo: string): boolean {
   }
   return true;
 }
+
+/**
+ * @param buffer
+ * @param raw
+ */
+export function fromEd25519PublicKeyBuffer(buffer: Buffer, raw = false): string {
+  return Ed25519PublicKey.fromBytes(buffer).toString(raw);
```
argjv, line 202: This is done by the `KeyPair` class in HBAR

modules/core/src/v2/coins/algo.ts
argjv: Yeah, the main problem here is that the ed29915 custody keys were stored in Stellar format, not the core one, so every time we want to reuse them, we have to extract the core prv/pub and then we can use it.

We explored the idea of getting a new key for HBAR to prevent this, but apparently it is not that easy to create and store such key.

modules/core/src/v2/internal/seedValidator.ts
```diff
@@ -0,0 +1,71 @@
...
+    return _.sum([isAlgoSeed, isStellarSeed, isHbarSeed]) !== 1;
+  }
+
+  /**
+   * Checks if this is a valid Hbar prv. These can be encoded differently.
+   * @param seed
+   */
+  static isValidHbarSeedFormat(seed: string): boolean {
+    try {
+      Ed25519PrivateKey.fromString(seed);
```
argjv, line 64: I think we have utils to do this in account-lib, don't we? If not, we should add it there

modules/core/src/v2/coins/algo.ts
tylerlevine: Got it - thanks for the background.

modules/account-lib/src/coin/hbar/utils.ts
twtaylorbitgo: Ah yes. We can remove this then

modules/core/src/v2/internal/seedValidator.ts
twtaylorbitgo: ok, removed it and moved this into account-lib

modules/core/src/v2/internal/seedValidator.ts
twtaylorbitgo: Ok, sure, added it there

modules/core/src/v2/internal/seedValidator.ts
```diff
@@ -0,0 +1,71 @@
...
+    return _.sum([isAlgoSeed, isStellarSeed, isHbarSeed]) !== 1;
+  }
+
+  /**
+   * Checks if this is a valid Hbar prv. These can be encoded differently.
+   * @param seed
+   */
+  static isValidHbarSeedFormat(seed: string): boolean {
+    try {
+      Hbar.Utils.createRawKey(seed);
```
argjv, line 64: Don't we have to release account-lib 1st and then point the SDK to that version?

twtaylorbitgo: No I don't think that's the case anymore

tylerlevine: > Don't we have to release account-lib 1st and then point the SDK to that version?

Nope, all the changes in master are released at the same time.

argjv (approved):

PR #910, opened 2020-09-24
add 'branch' environment

bitgoAaron:
Adds a new environment 'branch' which can serve the purposes of both dev & custom environments. Specifically, this branch allows for passing custom parameters when creating BitGo, but will otherwise be treated as a dev environment. 

JIRA: BG-25008


tylerlevine (approved):
Works for me, thanks

PR #913, opened 2020-10-02
Use blockchair as the BCH block explorer

glowang:
Our current blockexplorers for BCH mainnet and testnet recovery are both down. Changing them to blockchair. We will only have mainnet support for BCH recovery as a result. 
https://blockdozer.com/insight-api/addr/3CkKr9gykRKP8V5sE2zZS5L8c32TVX3mvg
https://test-bch-insight.bitpay.com/api/addr/3Fn25AxxbzRu4g4mJH1TdXmo4hTfkxfAz7 
Ticket: BG-24863

### DIFF
--- a/modules/core/src/v2/coins/bch.ts
+++ b/modules/core/src/v2/coins/bch.ts
@@ -9,6 +9,7 @@ import { BaseCoin } from '../baseCoin';
 import { AbstractUtxoCoin, AddressInfo, UnspentInfo } from './abstractUtxoCoin';
 import * as common from '../../common';
 const co = Bluebird.coroutine;
+import { BlockchairApi } from '../recovery/blockchairApi';
 
 const VALID_ADDRESS_VERSIONS = {
   base58: 'base58',
@@ -189,27 +190,13 @@ export class Bch extends AbstractUtxoCoin {
     return common.Environments[this.bitgo.env].bchExplorerBaseUrl + url;
   }
 
-  getAddressInfoFromExplorer(addressBase58) {
-    return co<AddressInfo>(function *getAddressInfoFromExplorer() {
-      const addrInfo = yield request.get(this.recoveryBlockchainExplorerUrl(`/addr/${addressBase58}`)).result();
-
-      addrInfo.txCount = addrInfo.txApperances;
-      addrInfo.totalBalance = addrInfo.balanceSat;
-
-      return addrInfo;
-    }).call(this);
+  getAddressInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<AddressInfo> {
+    const explorer = new BlockchairApi(this.bitgo, 'bitcoin-cash', apiKey);
+    return Bluebird.resolve(explorer.getAccountInfo(addressBase58));
   }
 
-  getUnspentInfoFromExplorer(addressBase58) {
-    return co<UnspentInfo[]>(function *getUnspentInfoFromExplorer() {
-      const unspents = yield request.get(this.recoveryBlockchainExplorerUrl(`/addr/${addressBase58}/utxo`)).result();
-
-      unspents.forEach(function processUnspent(unspent) {
-        unspent.amount = unspent.satoshis;
-        unspent.n = unspent.vout;
-      });
-
-      return unspents;
-    }).call(this);
+  getUnspentInfoFromExplorer(addressBase58: string, apiKey?: string): Bluebird<UnspentInfo[]> {
+    const explorer = new BlockchairApi(this.bitgo, 'bitcoin-cash', apiKey);
+    return Bluebird.resolve(explorer.getUnspents(addressBase58));
   }
 }
--- a/modules/core/test/v2/unit/recovery.ts
+++ b/modules/core/test/v2/unit/recovery.ts
@@ -14,6 +14,7 @@ import moment = require('moment');
 import * as sinon from 'sinon';
 import { Btc } from '../../../src/v2/coins/btc';
 import { Bsv } from '../../../src/v2/coins/bsv';
+import { Bch } from '../../../src/v2/coins/bch';
 import {
   addressInfos,
   addressUnspents,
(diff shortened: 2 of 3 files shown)


modules/core/src/v2/coins/bch.ts
```diff
@@ -189,27 +190,13 @@ export class Bch extends AbstractUtxoCoin {
     return common.Environments[this.bitgo.env].bchExplorerBaseUrl + url;
   }
 
-  getAddressInfoFromExplorer(addressBase58) {
-    return co<AddressInfo>(function *getAddressInfoFromExplorer() {
-      const addrInfo = yield request.get(this.recoveryBlockchainExplorerUrl(`/addr/${addressBase58}`)).result();
```
matthewzipkin-bitgo, line 194: Can we remove `recoveryBlockchainExplorerUrl()` on lines 189-191 now?

modules/core/test/v2/unit/recovery.ts
```diff
@@ -357,8 +374,7 @@ describe('Recovery:', function() {
     }));
 
     it('should generate BCH recovery tx with KRS', co(function *() {
-      recoveryNocks.nockBchRecovery(bitgo, true);
-
+      nockCoingecko(1000,'bitcoin-cash');
```
matthewzipkin-bitgo, line 377: What's `KRS`? (sorry 😄 )

matthewzipkin-bitgo (approved):
LGTM, a few questions here and there.

Manually confirmed the blockchair API (results below).

I suppose it would be to unreliable to actually make HTTP requests out to blockchair from the test suite? (Looks like we use local fake server or data in the tests). That makes sense but that means the tests don't "really" cover the code. Also if blockchair ever breaks their API, the regression tests will inform us of that. I dunno, just food for thought.

https://api.blockchair.com/bitcoin-cash/dashboards/address/2NEXK4AjYnUCkdUDJQgbbEGGks5pjkfhcRN

```
{
  "data": {
    "2NEXK4AjYnUCkdUDJQgbbEGGks5pjkfhcRN": {
      "address": {
        "type": null,
        "script_hex": "",
        "balance": 0,
        "balance_usd": 0,
        "received": 0,
        "received_usd": 0,
        "spent": 0,
        "spent_usd": 0,
        "output_count": 0,
        "unspent_output_count": 0,
        "first_seen_receiving": null,
        "last_seen_receiving": null,
        "first_seen_spending": null,
        "last_seen_spending": null,
        "scripthash_type": null,
        "transaction_count": 0,
        "formats": null
      },
      "transactions": [],
      "utxo": []
    }
  },
  "context": {
    "code": 200,
    "source": "D",
    "limit": "100,100",
    "offset": "0,0",
    "results": 0,
    "state": 655840,
    "cache": {
      "live": true,
      "duration": 30,
      "since": "2020-10-05 20:30:26",
      "until": "2020-10-05 20:30:56",
      "time": null
    },
    "api": {
      "version": "2.0.66",
      "last_major_update": "2020-07-19 00:00:00",
      "next_major_update": null,
      "documentation": "https://blockchair.com/api/docs",
      "notice": "Beginning July 19th, 2020 we start enforcing request cost formulas, see the changelog for details"
    },
    "time": 0.05163884162902832,
    "render_time": 0.031190156936645508,
    "full_time": 0.08282899856567383,
    "request_cost": 1
  }
}
```

modules/core/test/v2/unit/recovery.ts
glowang: Key recovery service, providers would hold the backup keys, instead of the customer themselves.

modules/core/src/v2/coins/bch.ts
glowang: ah thanks, nice catch!

glowang:
> LGTM, a few questions here and there.
> 
> Manually confirmed the blockchair API (results below).
> 
> I suppose it would be to unreliable to actually make HTTP requests out to blockchair from the test suite? (Looks like we use local fake server or data in the tests). That makes sense but that means the tests don't "really" cover the code. Also if blockchair ever breaks their API, the regression tests will inform us of that. I dunno, just food for thought.
> 
> https://api.blockchair.com/bitcoin-cash/dashboards/address/2NEXK4AjYnUCkdUDJQgbbEGGks5pjkfhcRN
> 
> ```
> {
>   "data": {
>     "2NEXK4AjYnUCkdUDJQgbbEGGks5pjkfhcRN": {
>       "address": {
>         "type": null,
>         "script_hex": "",
>         "balance": 0,
>         "balance_usd": 0,
>         "received": 0,
>         "received_usd": 0,
>         "spent": 0,
>         "spent_usd": 0,
>         "output_count": 0,
>         "unspent_output_count": 0,
>         "first_seen_receiving": null,
>         "last_seen_receiving": null,
>         "first_seen_spending": null,
>         "last_seen_spending": null,
>         "scripthash_type": null,
>         "transaction_count": 0,
>         "formats": null
>       },
>       "transactions": [],
>       "utxo": []
>     }
>   },
>   "context": {
>     "code": 200,
>     "source": "D",
>     "limit": "100,100",
>     "offset": "0,0",
>     "results": 0,
>     "state": 655840,
>     "cache": {
>       "live": true,
>       "duration": 30,
>       "since": "2020-10-05 20:30:26",
>       "until": "2020-10-05 20:30:56",
>       "time": null
>     },
>     "api": {
>       "version": "2.0.66",
>       "last_major_update": "2020-07-19 00:00:00",
>       "next_major_update": null,
>       "documentation": "https://blockchair.com/api/docs",
>       "notice": "Beginning July 19th, 2020 we start enforcing request cost formulas, see the changelog for details"
>     },
>     "time": 0.05163884162902832,
>     "render_time": 0.031190156936645508,
>     "full_time": 0.08282899856567383,
>     "request_cost": 1
>   }
> }
> ```

The best reason I can think of is some explorers have rate limits for testnets too, so we wouldn't be able to finish running these tests fast enough without exposing an API key. But otherwise I can't find a good reason for mocking these calls 😅 maybe @tylerlevine has a better explanation ?

modules/core/test/v2/unit/recovery.ts
glowang: no worries :) thank you so much for your thorough review!! @matthewzipkin-bitgo

OttoAllmendinger:
Regarding mocks vs actual requests: from my experience, it is more common that a third-party service has a temporary problem than that it silently breaks the API. For that reason we should use mocked data.

(we could add a separate third-party integration test if we have time, but our build should not depend on the test status)

OttoAllmendinger (approved):

PR #915, opened 2020-10-05
Add ERC20s

wesgraham:
YFI, AMO, CRPT1, UTK1
Ticket: BG-25332


wesgraham:
Good catch - thanks Tyler

PR #917, opened 2020-10-06
Allow param for userkey path for address derivation

glowang:
The current address derivation logic assumes that all keys have paths that start with prefix `m/0/0`, which isn't true for all user keys of v1 wallets, as some might have different path param. 

This PR allows us to take an optional parameter for the user keypath that can override the default

Ticket: BG-24941

### DIFF
--- a/modules/core/src/bitcoin.ts
+++ b/modules/core/src/bitcoin.ts
@@ -43,6 +43,32 @@ function getKey(network?: bitcoin.Network): bitcoin.ECPair {
 bitcoin.HDNode.prototype.getKey = getKey;
 
 /**
+ * Given a key and a path, derive the child key.
+ * @param {bitcoin.HDNode} userKey
+ * @param {string} path
+ * @returns {bitcoin.HDNode}
+ */
+export function deriveKeyByPath(userKey: bitcoin.HDNode, path: string): bitcoin.HDNode {
+  let key = userKey;
+  let splitPath = path.split('/');
+  // if a key path starts with "m", it is the path for a master node. derivePath() is used specifically for
+  // deriving master node and we can call it directly.
+  if (splitPath[0] === 'm') {
+    key = userKey.derivePath(path);
+  } else {
+    // if the path does not start with "m", it typically looks like "/x/y/...", and the splitPath
+    // would look like ['', 'x', 'y',...], and we need to get ride of the empty string at index 0.
+    // Then we continue deriving the child by calling derive() on the new child at each subsequent level.
+    splitPath = splitPath.slice(1);
+    for (const p of splitPath) {
+      const index = parseInt(p, 10);
+      key = key.derive(index);
+    }
+  }
+  return key;
+}
+
+/**
  * Derive a child HDNode from a parent HDNode and index. Uses secp256k1 to speed
  * up public key derivations by 100x vs. bitcoinjs-lib implementation.
  *
--- a/modules/core/src/bitcoin.ts
+++ b/modules/core/src/bitcoin.ts
@@ -7,6 +7,7 @@
 import * as common from './common';
 import * as bitcoin from '@bitgo/utxo-lib';
 import { V1Network } from './v2/types';
+import { InvalidKeyPathError } from './errors';
 const ecurve = require('ecurve');
 const curve = ecurve.getCurveByName('secp256k1');
 const BigInteger = require('bigi');
@@ -62,6 +63,9 @@ export function deriveKeyByPath(userKey: bitcoin.HDNode, path: string): bitcoin.
     splitPath = splitPath.slice(1);
     for (const p of splitPath) {
       const index = parseInt(p, 10);
+      if (isNaN(index) || index.toString() != p) {
+        throw new InvalidKeyPathError(path);
+      }
       key = key.derive(index);
     }
   }
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -232,6 +232,7 @@ export interface RecoverParams {
   bitgoKey: string;
   walletPassphrase?: string;
   apiKey?: string;
+  userKeyPath?: string;
 }
 
 export interface VerifyKeySignaturesOptions {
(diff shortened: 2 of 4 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1664,6 +1665,38 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+      return derivedKeys;
+    }
+    return keyArray.map((k) => k.derivePath(path));
+  }
+
+  /**
+   * Derive child keys at specific index, from provided parent keys
+   * @param {bitcoin.HDNode[]} keyArray
+   * @param {number} index
+   * @returns {Ed25519PrivateKey[]}
```
matthewzipkin-bitgo, line 1694: Is there something generic about the `Ed25519PrivateKey` object? Because I think that's the wrong curve for Bitcoin, but I might be misunderstanding the meaning of "ed25519" or "bitcoin.HDNode" in this context.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1775,9 +1804,11 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
         throw new Error('specified key recovery service does not support recoveries for this coin');
       }
 
+      // check whether key material and password authenticate the users and return parent keys of all three keys of the wallet
       const keys = yield self.initiateRecovery(params);
 
-      const baseKeyPath = deriveKeys(deriveKeys(keys, 0), 0);
+      const defaultPrefix = 'm/0/0';
```
matthewzipkin-bitgo, line 1810: Typically in BIP44 the `purpose` and `coinType` values are hardened (`account` is also, but that would be the *next* value in the path, which I don't see). I need to learn about initial bitgo wallet creation -- do we use un-hardened values like this as the root path?

matthewzipkin-bitgo (dismissed):
I'm not sure about the derivation path used. Also I think this needs a test, I don't think the new functions are covered. There are "official" BIP32 test vectors if you want to test against them: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#Test_Vectors

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: should be `bitcoin.HDNode[]`

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1664,6 +1665,38 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+    if (userKeyPath) {
+      const derivedKeys: bitcoin.HDNode[] = [];
+      const userKey = keyArray.shift();
+      derivedKeys.push(userKey.derivePath(userKeyPath));
+      const keys = keyArray.map((k) => k.derivePath(path));
+      derivedKeys.concat(keys);
+      return derivedKeys;
+    }
+    return keyArray.map((k) => k.derivePath(path));
+  }
```
OttoAllmendinger, line 1688: this is a bit difficult to understand 

how about 

```
const paths = [userKeyPath || path, path, path];
return keyArray.map((k, i) => k.derivePath(paths[i]))
```

modules/core/src/v2/coins/abstractUtxoCoin.ts
glowang: It seems that we do use unhardned values. 
What we used to have here, `deriveKeys(deriveKeys(keys, 0), 0)`, is the equivalance of deriving directly from `m/0/0`, based on the implementation in `bitgo-utxo-lib`.

modules/core/src/v2/coins/abstractUtxoCoin.ts
glowang: Thanks Otto, yeah, looks much more straightforward :)

modules/core/src/v2/coins/abstractUtxoCoin.ts
glowang: Just realized that the `derivePath` function only works for deriving from master node, not child nodes. Then we would still need to use the `derive()` function. I would have to first find out the different possibilities for key paths before proceeding here.

glowang:
Thanks for the review! I added a test for the helper. However, even with my changes here, I was not able to generate the correct address on the WRW side. 

I was wondering, whether the derivation also has to do with the `basePath` param in the `queryBlockchainUnspentsPath` function. It seems that the basePath is only used to provide a "code for chain", so the 0/0/ there shouldn't be the issue that my address generation is incorrect. Any idea on where I can look next? @matthewzipkin-bitgo  @OttoAllmendinger

glowang:
Nvm, I just tested this using a user's prod key pair that we used to struggle with (via the WRW), and now it worked! I was able to derive the v1 key address there :) https://github.com/BitGo/wallet-recovery-wizard/pull/96

modules/core/src/bitcoin.ts
```diff
@@ -42,6 +42,32 @@ function getKey(network?: bitcoin.Network): bitcoin.ECPair {
...
+  // deriving master node and we can call it directly.
+  if (splitPath[0] === 'm') {
+    key = userKey.derivePath(path);
+  } else {
+    // if the path does not start with "m", it typically looks like "/x/y/...", and the splitPath
+    // would look like ['', 'x', 'y',...], and we need to get ride of the empty string at index 0.
+    // Then we continue deriving the child by calling derive() on the new child at each subsequent level.
+    splitPath = splitPath.slice(1);
+    for (const p of splitPath) {
+      const index = parseInt(p, 10);
```
OttoAllmendinger, line 64: we should usually check `parseInt` result with `isNaN` but in this case we should go even further

sadly `parseInt` is very lenient 

```
parseInt(`1'`) === 1
parseInt(`1x`) === 1
```

Let's be extra strict here and throw and throw an error if `index.toString() !== p`

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #918, opened 2020-10-07
BGA-435 Add Tron WBTC and WETH configuration to Statics

silalvado:
Add configurations for WBTC and WETH tokens deployed on Tron network.

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -502,8 +502,8 @@ export const coins = CoinMap.fromCoins([
     'WXT'
   ),
   // TODO Set correct contract addresses for WBTC and WETH on Tron main network
-  tronToken('trx:wbtc', 'Tron Wrapped Bitcoin', 8, 'TPamvhorQcFpvtMtYb6qDDJNMsUM1doPty', UnderlyingAsset.WBTC),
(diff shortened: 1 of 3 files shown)


modules/statics/src/coins.ts
```diff
@@ -499,6 +501,9 @@ export const coins = CoinMap.fromCoins([
     '',
     'WXT'
   ),
+  // TODO Set correct contract addresses for WBTC and WETH on Tron main network
+  tronToken('trx:wbtc', 'Tron Wrapped Bitcoin', 8, 'TPamvhorQcFpvtMtYb6qDDJNMsUM1doPty', UnderlyingAsset.WBTC),
```
twtaylorbitgo, line 505: I'm not sure these UnderlyingASsets are right, I'll have to get back to this

twtaylorbitgo (changes requested):
Overall LGTM - I just have to check out how underlyingasset is used in our arch.

modules/statics/src/coins.ts
silalvado: Ok. I set two example addresses for WBTC and WETH in mainnet.

tylerlevine:
Hi @silalvado, is this PR still active or can it be closed?

FabianSimtlix:
This was deprecated. Closing the PR

PR #919, opened 2020-10-07
Add support for ETH Goerli testnet

marktoda:
This commit adds configurations to support the ETH Goerli testnet

CLOSES TICKET: BG-25415

### DIFF
--- a/modules/core/src/v2/coinFactory.ts
+++ b/modules/core/src/v2/coinFactory.ts
@@ -31,6 +31,7 @@ import {
   Teos,
   Tetc,
   Teth,
+  Gteth,
   Thbar,
   Tltc,
   Trbtc,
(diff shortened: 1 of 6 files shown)


modules/core/src/v2/coinFactory.ts
```diff
@@ -31,6 +31,7 @@ import {
   Teos,
   Tetc,
   Teth,
+  Gteth,
```
wesgraham, line 34: Have you double checked that all the `isFamily()` functions/checks don't look for the 't' prefix? Know i've seen something somewhere that evaluated`isFamilyEth()` by looking at whether the coin was `eth` or 't' + `eth`

wesgraham (approved):

modules/core/src/v2/coinFactory.ts
marktoda: I've looked a bit, didn't find any. I've tested integration w/ platform and it's working well there - we have been trying to remove all such checks over time, as it's not very robust

PR #922, opened 2020-10-13
remove _microservicesUrl and always use auth route

bitgoAaron:
BitGoJS will no longer use `/user/login` to login in any environment and will instead always use the `/api/auth` routes for authenticating.

### DIFF
--- a/modules/core/src/bitgo.ts
+++ b/modules/core/src/bitgo.ts
@@ -152,7 +152,6 @@ export interface BitGoOptions {
   serverXpub?: string;
   stellarFederationServerUrl?: string;
   useProduction?: boolean;
-  microservicesUri?: string;
   refreshToken?: string;
   validate?: boolean;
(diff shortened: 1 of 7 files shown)


modules/core/src/bitgo.ts
```diff
@@ -2001,7 +1994,7 @@ export class BitGo {
    * Create a url for calling BitGo microservice APIs
    */
```
tylerlevine, line 1995: This should be marked as `@deprecated` now, but we can deal with that in a follow up

tylerlevine (approved):
lgtm but looks like tests need some fixing

tylerlevine (approved):

modules/core/src/bitgo.ts
bitgoAaron: actually we still need to use this to wrap the url with baseUrl prefix.

PR #924, opened 2020-10-15
Use right semantics to check for undefined var

glowang:
- We currently use the wrong comparator to check whether
   a keynonce is undefined. This could lead to the wrong
   keynonces to beUse right semantics to check for undefined var
 - This was introduced in BG-24643

Ticket: BG-25181


ericcrosson-bitgo (approved):
Nice :) LGTM 🚀

PR #927, opened 2020-10-20
Turn on logging of the crosschain recovery tool

glowang:
- Our current instantiation of the cross chain recovery tool diabled
    logging of recovery steps, which makes it hard to get error traces
    when debugging.
  - This PR enable logging by setting "logging" to true, logging info
    to console.log (default logger when none provided)

Ticket: BG-22261


matthewzipkin-bitgo (approved):
LGTM, pretty simple - but is there a test I can run to see it in action?

glowang:
> LGTM, pretty simple - but is there a test I can run to see it in action?
This function is used in WrongChainRecovery of recovery.ts, and you can see that the console logs are printed.

PR #937, opened 2020-10-29
fix: hard-code current ZEC consensus branch ID using updated utxo-lib

matthewzipkin-bitgo:
Update consensus branch IDs for tzec (Canopy hard fork activated) and zec (still on old Heartwood fork).
Pulls in updated utxo-lib from https://github.com/BitGo/bitgo-utxo-lib/pull/91

TICKET BG-24753


matthewzipkin-bitgo:
@argjv added comments in https://github.com/BitGo/BitGoJS/pull/937/commits/c3e30cda7be955c464be8977bd20f3edc6c01c31 documenting the zcash-cli commands I used to generate test vectors. I modified the scripts you sent me and added stuff to decrypt the private keys, etc but I don't think that needs to be included in the repo.

modules/core/package.json
```diff
@@ -47,7 +47,7 @@
     "@bitgo/account-lib": "^2.3.0",
     "@bitgo/statics": "^5.3.0",
     "@bitgo/unspents": "^0.6.0",
-    "@bitgo/utxo-lib": "^1.7.1",
+    "@bitgo/utxo-lib": "git+https://github.com/BitGo/bitgo-utxo-lib.git#a08adf4478d8bace8a0e932158674e76ae463fca",
```
marktoda, line 50: Should we wait for utxo-lib to get a release?

matthewzipkin-bitgo: it is actually: https://github.com/BitGo/bitgo-utxo-lib/tree/v1.7.x but we decided to require it by hash for security reasons rather than a tag name.

tylerlevine: Could we release bitgo-utxo-lib rather than refer to a hash? This precludes dep hoisting with other semver-compatible utxo-lib consumers.

tylerlevine (changes requested):
Please use a semver instead of a git hash in the package.json

modules/core/package.json
matthewzipkin-bitgo: @argjv could you weigh in on this? We kinda went back and forth...

argjv: Oh, here we are using the new repo name. I just published 1.7.3 for `@bitgo/utxo-lib`.
I thought we were referencing the `bitgo-utxo-lib` package.

matthewzipkin-bitgo: fixed in 1798510690766438e4faa30bbb0c3f4188d99e91

tylerlevine (approved):

PR #939, opened 2020-11-06
BG-26001 support Bitcoin-ABC (BCHA) recovery in SDK

glowang:
This PR adds BCHA as a new coin in the SDK, so we could recover it with non bitgo recovery functions after the fork.

Please note that the chain params of BCHA and BCHN are identical, so we are reusing the same set of params defined on the UTXO lib. (Since UTXO Lib is a bit messy anyhow.) Therefore, BCHA does not need to override the constructor of BCHN.

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1602,6 +1602,14 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
 
   /**
    * Get the current market price from a third party to be used for recovery
+   * This function is only intended for non-bitgo recovery transactions, when it is necessary
+   * to calculate the rough fee needed to pay to Keyternal. We are okay with approximating,
+   * because the resulting price of this function only has less than 1 dollar influence on the
+   * fee that needs to be paid to Keyternal.
+   *
+   * See calculateFeeAmount function:  return Math.round(feeAmountUsd / currentPrice * self.getBaseFactor());
+   *
+   * This end function should not be used as an accurate endpoint, since some coins' prices are missing from the provider
    */
   getRecoveryMarketPrice(): Bluebird<string> {
     const self = this;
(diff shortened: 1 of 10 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1612,7 +1612,8 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
         .set('BCH', 'bitcoin-cash')
         .set('ZEC', 'zcash')
         .set('DASH', 'dash')
-        .set('BSV', 'bitcoin-sv');
+        .set('BSV', 'bitcoin-sv')
+        .set('BCHA', 'bitcoin-cash-abc');
```
argjv, line 1616: I think the official name is bitcoin-abc

modules/statics/src/networks.ts
```diff
@@ -131,6 +131,16 @@ class BitcoinCashTestnet extends BitcoinLikeTestnet {
   explorerUrl = 'https://tbch.blockdozer.com/tx/';
 }
 
+class BitcoinABC extends BitcoinLikeMainnet {
+  family = CoinFamily.BCHA;
+  explorerUrl = 'https://api.cryptoapis.io/v1/bc/bch/mainnet';
```
argjv, line 136: Shouldn't this be blockchair?

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1612,7 +1620,11 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
         .set('BCH', 'bitcoin-cash')
         .set('ZEC', 'zcash')
         .set('DASH', 'dash')
-        .set('BSV', 'bitcoin-sv');
+        // note: we don't have a source for price data of BCHA and BSV, but we will use BCH as a proxy. We will substitute
+        // it out for a better source when it becomes available.  TODO BG-26359.
+        .set('BCHA', 'bitcoin-cash')
```
argjv, line 1625: bitcoin-abc

argjv: My bad, this depends on the API

argjv (approved):

PR #940, opened 2020-11-09
fix(config): add BSV and BCHA as recoverable coin with coincover

glowang:
We need to add bsv and bcha under the supportedCoins param of SDK, so that the recovery transaction can be initiated.

### DIFF
--- a/modules/core/src/config.ts
+++ b/modules/core/src/config.ts
@@ -193,7 +193,7 @@ export const krsProviders = {
   dai: {
     feeType: 'flatUsd',
     feeAmount: 0, // dai will receive payments off-chain
-    supportedCoins: ['btc', 'eth', 'xlm', 'xrp', 'dash', 'zec', 'ltc', 'bch']
+    supportedCoins: ['btc', 'eth', 'xlm', 'xrp', 'dash', 'zec', 'ltc', 'bch', 'bsv', 'bcha']
   }
 };
 


modules/core/src/config.ts
```diff
@@ -193,7 +193,7 @@ export const krsProviders = {
   dai: {
     feeType: 'flatUsd',
     feeAmount: 0, // dai will receive payments off-chain
-    supportedCoins: ['btc', 'eth', 'xlm', 'xrp', 'dash', 'zec', 'ltc', 'bch']
+    supportedCoins: ['btc', 'eth', 'xlm', 'xrp', 'dash', 'zec', 'ltc', 'bch', 'bsv', 'bcha']
```
matthewzipkin-bitgo, line 196: So the two bitcoin cash forks are going to be `bch` and `bcha`? For somer reason I thought `bcha` meant Bitcoin ABC, and the other fork will be called BCHN ? And then, is there a way to verify that "dai" (which I figured out is now called "Coincover") operates on these two "new" coins?

matthewzipkin-bitgo (approved):
LGTM, just a few questions

modules/core/src/config.ts
glowang: BCHN will inherent the existing BCH's nomenclature. We are talking to coincover to help us test this!

PR #941, opened 2020-11-10
Add BLS key generation for eth2

wesgraham:
Adds ability to create user and backup keys for ETH2 wallets client-side.
Ticket: BG-26379

### DIFF
--- a/modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
+++ b/modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
@@ -1,0 +1,66 @@
+import * as BLS from '@chainsafe/bls';
+import { KeyPairOptions, isPrivateKey } from './iface';
+import { BaseKeyPair } from './baseKeyPair';
+import { AddressFormat } from './enum';
+import { NotImplementedError } from './errors';
+
+/**
+ * Base class for BLS keypairs.
+ */
+export abstract class BlsKeyPair implements BaseKeyPair {
+  protected keyPair: BLS.Keypair;
+
+  /**
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   */
+  protected constructor() {
+    this.keyPair = BLS.generateKeyPair();
+  }
+
+  /**
+   * Build a keyPair from private key.
+   *
+   * @param {string} prv a hexadecimal private key
+   */
+  recordKeysFromPrivateKey(prv: string) {
+    if (this.isValidBLSPrv(prv)) {
+      const privateKey = BLS.PrivateKey.fromHexString(prv);
+      this.keyPair = new BLS.Keypair(privateKey);
+    } else {
+      throw new Error('Invalid private key');
+    }
+  }
+
+  /**
+   * Note - this is not possible using BLS. BLS does not support pubkey derived key gen
+   *
+   * @param {string} pub - An extended, compressed, or uncompressed public key
+   */
+  recordKeysFromPublicKey(pub: string): void {
+    throw new NotImplementedError('Public key derivation is not supported in bls');
+  }
+
+  getAddress(format?: AddressFormat): string {
+    throw new NotImplementedError('getAddress not implemented');
+  }
+
+  getKeys(): any {
+    throw new NotImplementedError('getKeys not implemented');
+  }
+
+  /**
+   * Whether the input is a valid BLS private key
+   *
+   * @param {string} prv A hexadecimal public key to validate
+   * @returns {boolean} Whether the input is a valid private key or not
+   */
+  isValidBLSPrv(prv: string): boolean {
+    try {
+      BLS.PrivateKey.fromHexString(prv);
+      return true;
+    } catch (e) {
+      return false;
+    }
+  }
+}
--- a/modules/account-lib/src/coin/eth2/keyPair.ts
+++ b/modules/account-lib/src/coin/eth2/keyPair.ts
@@ -1,0 +1,37 @@
+import * as BLS from '@chainsafe/bls';
+import { DefaultKeys } from '../baseCoin/iface';
+import { BlsKeyPair } from '../baseCoin/blsKeyPair';
+
+/**
+ * Ethereum keys and address management.
+ */
+export class KeyPair extends BlsKeyPair {
+  /**
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   */
+  constructor() {
+    super();
+  }
+
+  /**
+   * ETH2 default keys format is a pair of Uint8Array keys
+   *
+   * @returns { DefaultKeys } The keys in the defined format
+   */
+  getKeys(): BLS.Keypair {
+    if (this.keyPair) {
+      return this.keyPair;
+    }
+    throw new Error('KeyPair has not been specified');
+  }
+
+  /**
+   * Get an Ethereum public address
+   *
+   * @returns {string} The address derived from the public key
+   */
+  getAddress(): string {
+    return this.getKeys().publicKey.toHexString();
+  }
+}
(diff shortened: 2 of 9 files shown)


modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   * @param {KeyPairOptions} source Either a master seed, a private key (extended or raw), or a public key
+   *     (extended, compressed, or uncompressed)
+   */
+  protected constructor(source?: KeyPairOptions) {
+    if (!source) {
+      const blsKeypair = BLS.generateKeyPair();
+      this.keyPair = { prv: blsKeypair.privateKey().toBytes(), pub: blsKeypair.publicKey().toBytesCompressed() };
```
marktoda, line 24: Let's just store the full `BLS` object and do format conversions in getter functions. Best to keep full context in state for flexibility and to get some validation for free. For example current impl of `recordKeysFromPrivateKey` doesn't validate the prv, but if we create and store a bls key object using the library we will likely get validation from the library for free

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -0,0 +1,62 @@
...
+    // TODO: FIX THIS CHECK
```
marktoda, line 25: What is wrong with the check? Also add a ticket for future cleanup if required

modules/account-lib/src/coin/eth2/keyPair.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (source) {
+      // TODO: FIX THIS CHECK
+      if (isPrivateKey(source)) {
+        super.recordKeysFromPrivateKey(source.prv);
+      } else {
+        throw new Error('Invalid key pair options');
+      }
+    } else {
+      throw new Error('Invalid key pair options');
+    }
```
marktoda, line 27: Doesn't the super constructor do all of this anyways? Why do you need it duplicated here?

modules/account-lib/src/coin/eth2/keyPair.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  /**
+   * ETH2 default keys format is a pair of Uint8Array keys
+   *
+   * @returns { DefaultKeys } The keys in the defined format
+   */
+  getKeys(): DefaultKeys {
+    const result: DefaultKeys = {
+      // Need some way to get pubkey from keyPair item
```
marktoda, line 37: See above, by just storing the BLS object you can avoid having to re-animate the object every time we need to serialize the keys

modules/account-lib/src/coin/eth2/keyPair.ts
```diff
@@ -0,0 +1,57 @@
...
+      pub: BLS.PublicKey.fromBytes(this.keyPair.pub).toHexString(),
```
marktoda, line 38: Just to be sure: does this add a 0x prefix?

modules/account-lib/src/coin/eth2/keyPair.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  /**
+   * Get an Ethereum public address
+   *
+   * @returns {string} The address derived from the public key
+   */
+  getAddress(): string {
+    const publicKey = Buffer.from(this.getKeys().pub, 'hex'); // first two characters identify a public key
+    return addHexPrefix(pubToAddress(publicKey, true).toString('hex'));
```
marktoda, line 55: This is an ETH1 library and not how addresses are generated in eth2. In fact I'm surprised this works at all since the pubkey format is different as well

modules/account-lib/src/utils/crypto.ts
```diff
@@ -150,3 +151,18 @@ export function isValidEd25519PublicKey(pub: string): boolean {
...
   }
 }
+
+/**
+ * Whether the input is a valid BLS private key
+ *
+ * @param {string} prv A hexadecimal public key to validate
+ * @returns {boolean} Whether the input is a valid public key or not
+ */
+export function isValidBLSPrv(prv: Uint8Array): boolean {
```
marktoda, line 161: I know you're just following the pattern but seems that the `keyPair` classes should have these verification functions IMO

marktoda (changes requested):

modules/account-lib/src/utils/crypto.ts
marktoda: No action required here, just a comment

modules/account-lib/src/coin/eth2/keyPair.ts
wesgraham: The latest commit does!

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -0,0 +1,75 @@
...
+import { BaseKeyPair } from './baseKeyPair';
+import { AddressFormat } from './enum';
+import { NotImplementedError } from './errors';
+
+/**
+ * Base class for BLS keypairs.
+ */
+export abstract class BlsKeyPair implements BaseKeyPair {
+  protected keyPair: BLS.Keypair;
+  protected source?: KeyPairOptions;
```
marktoda, line 12: Do you need this field? I don't see you ever set or read from it

marktoda (changes requested):
one small nit but LGTM after

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
wesgraham: I was assuming some upstream tools may use it (thought it was a standard object for doing things like getting a keypair from privkey, etc.). It's included in the unit tests as well - can remove though if desired?

marktoda: It's not even set to a value anywhere currently, though

modules/account-lib/test/unit/coin/eth2/keyPair.ts
```diff
@@ -0,0 +1,74 @@
...
+      should.throws(
+        () => new KeyPair(longerPrv),
+        e => e.message === testData.errorMessageInvalidPrivateKey,
+      );
+      should.throws(
+        () => new KeyPair({ prv: prv + pub }),
+        e => e.message === testData.errorMessageInvalidPrivateKey,
+      );
+    });
+  });
```
marktoda, line 73: Add some tests for `recordFromPrivateKey` as well as it's part of the public interface for KeyPairs

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
marktoda: I don't see it in the unit tests either. Fields shouldn't be part of the interface - callers can access everything they need via the interface functions. I think you can just remove it no problem

marktoda (approved):
LGTM nice work

modules/account-lib/test/unit/coin/eth2/keyPair.ts
```diff
@@ -0,0 +1,81 @@
+import should from 'should';
+import * as BLS from '@chainsafe/bls';
+import { KeyPair } from '../../../../src/coin/eth2';
+import * as testData from '../../../resources/eth2/eth2';
+
+const pub = testData.ACCOUNT_1.publicKey;
+const prv = testData.ACCOUNT_1.privateKey;
+
+before(async function f() {
+  await BLS.initBLS();
```
marktoda, line 10: oh, you should do this inside account-lib - we don't want to make users always do this before using account-lib

PR #942, opened 2020-11-12
Add eth2 coin controller

wesgraham:
Adds ETH2 coin controller with methods for recovery, validating addresses, retrieving address/balance/nonce info.

### DIFF
--- a/modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
+++ b/modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
@@ -37,7 +37,7 @@ export abstract class BlsKeyPair implements BaseKeyPair {
    * @param {string} prv a hexadecimal private key
    */
   recordKeysFromPrivateKey(prv: string) {
-    if (this.isValidBLSPrv(prv)) {
+    if (BlsKeyPair.isValidBLSPrv(prv)) {
       const privateKey = BLS.PrivateKey.fromHexString(prv);
       this.keyPair = new BLS.Keypair(privateKey);
     } else {
@@ -68,9 +68,24 @@ export abstract class BlsKeyPair implements BaseKeyPair {
    * @param {string} prv A hexadecimal public key to validate
    * @returns {boolean} Whether the input is a valid private key or not
    */
-  isValidBLSPrv(prv: string): boolean {
+  public static isValidBLSPrv(prv: string): boolean {
     try {
       BLS.PrivateKey.fromHexString(prv);
+      return true;
+    } catch (e) {
+      return false;
+    }
+  }
+
+  /**
+   * Return boolean indicating whether input is valid public key for the coin.
+   *
+   * @param {string} pub the pub to be checked
+   * @returns {boolean} is it valid?
+   */
+  public static isValidBLSPub(pub: string): boolean {
+    try {
+      BLS.PublicKey.fromHex(pub);
       return true;
     } catch (e) {
       return false;
--- a/modules/account-lib/src/coin/eth2/keyPair.ts
+++ b/modules/account-lib/src/coin/eth2/keyPair.ts
@@ -1,6 +1,6 @@
-import * as BLS from '@bitgo/bls';
-import { DefaultKeys } from '../baseCoin/iface';
 import { BlsKeyPair } from '../baseCoin/blsKeyPair';
+import { DefaultKeys } from '../baseCoin/iface';
+import * as testData from '../../../test/resources/eth2/eth2';
 
 /**
  * Ethereum keys and address management.
@@ -19,11 +19,11 @@ export class KeyPair extends BlsKeyPair {
    *
    * @returns { DefaultKeys } The keys in the defined format
    */
-  getKeys(): BLS.Keypair {
+  getKeys(): DefaultKeys {
     if (this.keyPair) {
-      return this.keyPair;
+      return { prv: this.keyPair.privateKey.toHexString(), pub: this.keyPair.publicKey.toHexString() };
     }
-    throw new Error('KeyPair has not been specified');
+    throw new Error('Error geting keys. Check keyPair has been specified & privae key is valid');
   }
 
   /**
@@ -32,6 +32,22 @@ export class KeyPair extends BlsKeyPair {
    * @returns {string} The address derived from the public key
    */
   getAddress(): string {
-    return this.getKeys().publicKey.toHexString();
+    return this.getKeys().pub;
+  }
+
+  static isValidPub(pub: string): boolean {
+    return BlsKeyPair.isValidBLSPub(pub);
+  }
+
+  static isValidPrv(prv: string | Buffer) {
+    if (typeof prv === 'string') {
+      return BlsKeyPair.isValidBLSPrv(prv);
+    }
+    try {
+      prv = '0x' + Buffer.from(prv).toString('hex');
+      return BlsKeyPair.isValidBLSPrv(prv);
+    } catch (e) {
+      return false;
+    }
   }
 }
--- a/modules/core/src/v2/coins/eth2.ts
+++ b/modules/core/src/v2/coins/eth2.ts
@@ -1,0 +1,347 @@
+/**
+ * @prettier
+ */
+import * as Bluebird from 'bluebird';
+import * as _ from 'lodash';
+import * as request from 'superagent';
+import { Eth2 as Eth2AccountLib } from '@bitgo/account-lib';
+import BigNumber from 'bignumber.js';
+
+import {
+  BaseCoin,
+  FeeEstimateOptions,
+  KeyPair,
+  ParsedTransaction,
+  ParseTransactionOptions,
+  VerifyAddressOptions,
+  VerifyTransactionOptions,
+  HalfSignedTransaction as BaseHalfSignedTransaction,
+} from '../baseCoin';
+import { BitGo } from '../../bitgo';
+import { NodeCallback } from '../types';
+import * as common from '../../common';
+
+const co = Bluebird.coroutine;
+
+interface Recipient {
+  address: string;
+  amount: string;
+  data?: string;
+}
+
+interface SignFinalOptions {
+  txPrebuild: {
+    gasPrice: string;
+    gasLimit: string;
+    recipients: Recipient[];
+    halfSigned: {
+      expireTime: number;
+      contractSequenceId: number;
+      backupKeyNonce?: number;
+      signature: string;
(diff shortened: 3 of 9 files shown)


modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,767 @@
...
+   */
+  getRecoveryGasLimit(): any {
+    return new optionalDeps.ethUtil.BN('500000');
+  }
+
+  /**
+   * Default expire time for a contract call (1 week)
+   * @returns {number} Time in seconds
+   */
+  getDefaultExpireTime(): number {
```
marktoda, line 235: not needed as eth2 won't use the wallet contract

modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,767 @@
...
+    return this.isValidPub(address);
+  }
+
+  /**
+   * Return boolean indicating whether input is valid public key for the coin.
+   *
+   * @param {String} pub the pub to be checked
+   * @returns {Boolean} is it valid?
+   */
+  isValidPub(pub: string): boolean {
```
marktoda, line 206: didn't you impl this in account-lib? can just reuse

modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,767 @@
...
+        new optionalDeps.ethUtil.BN(optionalDeps.ethUtil.stripHexPrefix(recipient.address), 16),
+        recipient.amount,
+        Buffer.from(optionalDeps.ethUtil.stripHexPrefix(recipient.data) || '', 'hex'),
+        expireTime,
+        contractSequenceId,
+      ],
+    ];
+  }
+
+  getOperationSha3ForExecuteAndConfirm(
```
marktoda, line 280: don't need this as we aren't using contract for eth

modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,767 @@
...
+  }
+
+  /**
+   * Get transfer operation for coin
+   * @param recipient recipient info
+   * @param expireTime expiry time
+   * @param contractSequenceId sequence id
+   * @returns {Array} operation array
+   */
+  getOperation(recipient: Recipient, expireTime: number, contractSequenceId: number): (string | Buffer)[][] {
```
marktoda, line 266: don't need this as we aren't using contract for eth

modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,767 @@
...
+  /**
+   * Query Etherscan for the balance of an address
+   * @param address {String} the ETH address
+   * @param callback
+   * @returns {BigNumber} address balance
+   */
+  queryAddressBalance(address: string, callback?: NodeCallback<any>): Bluebird<any> {
+    const self = this;
+    return co(function*() {
+      const result = yield self.recoveryBlockchainExplorerQuery({
```
marktoda, line 248: Does this work for ETH2 chain?

modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,767 @@
...
+    );
+  }
+
+  /**
+   * Queries the contract (via BeaconScan) for the next sequence ID
+   * @param address {String} address of the contract
+   * @param callback
+   * @returns {Number} sequence ID
+   */
+  querySequenceId(address: string, callback?: NodeCallback<number>): Bluebird<number> {
```
marktoda, line 338: Don't need this as not using contracts in eth2

modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,767 @@
...
+   * @param params.enterprise {String} the enterprise id to associate with this key
+   * @param params.newFeeAddress {Boolean} create a new fee address (enterprise not needed in this case)
+   */
+  preCreateBitGo(params: PrecreateBitGoOptions): void {
+    // We always need params object, since either enterprise or newFeeAddress is required
+    if (!_.isObject(params)) {
+      throw new Error(`preCreateBitGo must be passed a params object. Got ${params} (type ${typeof params})`);
+    }
+
+    if (_.isUndefined(params.enterprise) && _.isUndefined(params.newFeeAddress)) {
```
marktoda, line 384: no fee address in eth2

marktoda (commented):
Lots of this stuff doesn't quite work for ETH2. Defer to out-of-band conversation about how to handle this

modules/core/src/v2/coins/eth2.ts
wesgraham: The check was for `isValidPriv` --> looking at it it's simpler to keep it in here I think, allows us to avoid initializing a keypair() object before checking if the pubkey is valid

modules/core/src/v2/coins/eth2.ts
wesgraham: This should - for ETH1 we query etherscan - for ETH2 I set it up to query beaconscan (same API as Etherscan, made by same creators). If there is a better API certainly let me know (this might fail until transferring is enabled in ETH2 though)

modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,399 @@
...
+    const keyPair = new Eth2AccountLib.KeyPair();
+    if (seed) {
+      const prv = BLS.PrivateKey.fromBytes(seed).toHexString();
+      // @ts-ignore
+      keyPair.recordKeysFromPrivateKey(prv);
+    }
+
+    return {
+      // @ts-ignore
+      pub: keyPair.getKeys().publicKey.toHexString(),
```
marktoda, line 375: I guess I should have mentioned this on your account-lib PR, but personally I think that `keyPair.getKeys()` should already be returning the hex string serializations. Similar to reasoning above and below about abstracting the BLS library away. A user of the accountLib keyPair library shouldn't also have to concern themselves with this third party BLS library - we should just provide a generic interface to them

modules/core/test/v2/unit/coins/eth2.ts
```diff
@@ -0,0 +1,55 @@
...
+import * as BLS from '@chainsafe/bls';
+
+describe('Ethereum 2.0', function() {
+  let bitgo;
+  let basecoin;
+
+  before(async function() {
+    bitgo = new TestBitGo({ env: 'mock' });
+    bitgo.initializeTestVars();
+    await BLS.initBLS();
```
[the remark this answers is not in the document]
wesgraham: agree - waiting on a comment from tyler on how to initialize then will fix

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -68,12 +82,42 @@ export abstract class BlsKeyPair implements BaseKeyPair {
...
     }
   }
+
+  /**
+   * Whether the input is a valid BLS private key
+   *
+   * @param {string} prv A hexadecimal public key to validate
+   * @returns {boolean} Whether the input is a valid private key or not
+   */
+  public static isValidBLSPrvBytes(prv: Buffer): boolean {
```
marktoda, line 100: Why do we need a bytes and hex string version of the same function? Seems one or the other should suffice and callers can serialize into the input format the function takes.

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -37,14 +37,28 @@ export abstract class BlsKeyPair implements BaseKeyPair {
...
       throw new Error('Invalid private key');
     }
   }
 
+  /**
+   * Build a keyPair from private key.
+   *
+   * @param {Buffer} prv a UInt8Array private key
+   */
+  recordKeysFromPrivateKeyBytes(prv: Buffer) {
```
marktoda, line 53: Same as below. I don't see the need for this duplicity

marktoda: At the very least, if you really want this interface, have one function use the other function to dedupe code

modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,392 @@
...
+   * Fetch fee estimate information from the server
+   * @param {Object} params The params passed into the function
+   * @param {String} [params.recipient] The recipient of the transaction to estimate a send to
+   * @param {String} [params.data] The ETH tx data to estimate a send for
+   * @param callback
+   * @returns {Object} The fee info returned from the server
+   */
+  feeEstimate(params: FeeEstimateOptions, callback?: NodeCallback<FeeEstimate>): Bluebird<FeeEstimate> {
+    const self = this;
+    return co<FeeEstimate>(function* coFeeEstimate() {
```
marktoda, line 336: nit: IMO we shouldn't implement this because we really don't even know how the fees will work yet and definitely don't have an estimator implementation in the server yet

modules/core/src/v2/coins/eth2.ts
```diff
@@ -0,0 +1,392 @@
...
+import { BitGo } from '../../bitgo';
+import { NodeCallback } from '../types';
+import * as common from '../../common';
+import { EthereumLibraryUnavailableError } from '../../errors';
+
+const co = Bluebird.coroutine;
+const debug = debugLib('bitgo:v2:eth');
+
+export const optionalDeps = {
+  get ethUtil() {
```
marktoda, line 29: It looks like your only usage of `ethUtil` is for the `BigNumber` package inside... Let's just remove ethUtil and optionalDeps and just use bignumber on its own

marktoda (changes requested):

modules/account-lib/src/coin/eth2/keyPair.ts
```diff
@@ -32,6 +32,22 @@ export class KeyPair extends BlsKeyPair {
...
+    return BlsKeyPair.isValidBLSPub(pub);
+  }
+
+  static isValidPrv(prv: string | Buffer) {
+    if (typeof prv === 'string') {
+      return BlsKeyPair.isValidBLSPrv(prv);
+    }
+    try {
+      prv = '0x' + Buffer.from(prv).toString('hex');
+      return BlsKeyPair.isValidBLSPrv(prv);
```
marktoda, line 48: Nit: generally bad form to modify parameters. I would just directly pass `'0x' + Buffer...` into `BlsKeyPair.isValid..`, or use a new local variable

marktoda (approved):

PR #943, opened 2020-11-12
Update BLS initialization

wesgraham:
Updates BLS initialization so client does not need to worry about dependency
Ticket: BG-26379

### DIFF
--- a/modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
+++ b/modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
@@ -1,8 +1,17 @@
-import * as BLS from '@chainsafe/bls';
-import { KeyPairOptions, isPrivateKey } from './iface';
+import * as BLS from '@bitgo/bls';
 import { BaseKeyPair } from './baseKeyPair';
 import { AddressFormat } from './enum';
 import { NotImplementedError } from './errors';
+
+let initialized = false;
+const initialize = async () => {
+  await BLS.initBLS();
+  initialized = true;
+};
+
+setImmediate(async () => {
+  await initialize();
+});
 
 /**
  * Base class for BLS keypairs.
@@ -15,7 +24,11 @@ export abstract class BlsKeyPair implements BaseKeyPair {
    *
    */
   protected constructor() {
-    this.keyPair = BLS.generateKeyPair();
+    if (initialized) {
+      this.keyPair = BLS.generateKeyPair();
+    } else {
+      throw new Error('BLS has not completed initialization, initializing now');
+    }
   }
 
   /**
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -32,10 +32,10 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@bitgo/bls": "^0.1.0",
     "@bitgo/statics": "^5.3.0",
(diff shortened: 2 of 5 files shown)


marktoda (commented):
Yay nice. Want to run by @tylerlevine though - Tyler, this library requires this weird async initializer function to be run globally before the lib can be used. Is this the best way to handle this?

tylerlevine:
> Yay nice. Want to run by @tylerlevine though - Tyler, this library requires this weird async initializer function to be run globally before the lib can be used. Is this the best way to handle this?

The current implementation is going to kick off a race condition which could cause usage of the library before it's initialized.

Where are we going to use this? Can we lazy initialize before use? (eg, generating a key)

wesgraham:
@tylerlevine dumb question here but how would lazy initialization look? Can't stick the initialization into the constructor, and am trying to avoid i) creating the object then ii) calling `initialization()` separately

tylerlevine:
> @tylerlevine dumb question here but how would lazy initialization look? Can't stick the initialization into the constructor, and am trying to avoid i) creating the object then ii) calling `initialization()` separately

What do the calls to get to here look like? It might be the best you can do is to kick off the init in a `setImmediate` callback and then set a boolean once it's resolved, and then check to make sure the boolean is set before using the lib (or throwing if it's not set). At least you avoid weird race conditions this way.

Even better would be to find the closest async context which would use the library, and initialize it there before use. Unfortunately it doesn't look like account-lib is designed to do any async work on initialization, so callers might need to do this before using certain functions in account-lib.

marktoda (changes requested):
I don't think this will work. The `BaseKeyPair` interface doesn't define those functions as async, so callers won't know to `await` them and they will end up with race conditions where sometimes their keypair will not be initialized by the time they try to use it. 

I think instead you should use tyler's `setImmediate` idea and throw if uninitialized when you need it

wesgraham:
@tylerlevine Was this the approach as intended? Still not certain how to get around the failing browser test (no module declaration recognized)

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -24,9 +37,11 @@ export abstract class BlsKeyPair implements BaseKeyPair {
    * @param {string} prv a hexadecimal private key
    */
   recordKeysFromPrivateKey(prv: string) {
-    if (this.isValidBLSPrv(prv)) {
+    if (this.isValidBLSPrv(prv) && initialized) {
```
marktoda, line 40: Why do you need this here if it is in the constructor?

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -56,11 +71,15 @@ export abstract class BlsKeyPair implements BaseKeyPair {
    * @returns {boolean} Whether the input is a valid private key or not
    */
   isValidBLSPrv(prv: string): boolean {
-    try {
-      BLS.PrivateKey.fromHexString(prv);
-      return true;
-    } catch (e) {
-      return false;
+    if (initialized) {
```
marktoda, line 74: ditto

marktoda (changes requested):

marktoda (approved):
LGTM

PR #951, opened 2020-11-20
Add BLS pubkey aggregation

wesgraham:
Adds pubkey aggregation to support wallet creation in platform
Ticket: BG-26472

### DIFF
--- a/modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
+++ b/modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
@@ -13,6 +13,12 @@ setImmediate(async () => {
   await initialize();
 });
 
+function ensureInitialized() {
+  if (!initialized) {
+    throw new Error('BLS lib not yet initialized, please retry');
+  }
+}
+
 /**
  * Base class for BLS keypairs.
  */
@@ -24,11 +30,8 @@ export abstract class BlsKeyPair implements BaseKeyPair {
    *
    */
   protected constructor() {
-    if (initialized) {
-      this.keyPair = BLS.generateKeyPair();
-    } else {
-      throw new Error('BLS has not completed initialization, initializing now');
-    }
+    ensureInitialized();
+    this.keyPair = BLS.generateKeyPair();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -84,11 +87,25 @@ export abstract class BlsKeyPair implements BaseKeyPair {
...
       return true;
     } catch (e) {
       return false;
     }
   }
+
+  public static aggregatePubkeys(pubKeys: Uint8Array[]): Buffer {
+    if (!initialized) {
+      throw new Error('BLS lib not yet initialized, please retry');
+    }
```
marktoda, line 104: might as well move this into a helper function `ensureInitialized` or something, just so its a 1-liner in all these fns instead of a 3-liner, and to ensure error msg consistency

marktoda (changes requested):

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -69,6 +69,7 @@ export abstract class BlsKeyPair implements BaseKeyPair {
    * @returns {boolean} Whether the input is a valid private key or not
    */
   public static isValidBLSPrv(prv: string): boolean {
+    this.ensureInitialized();
```
marktoda, line 72: nit: could do it in the constructor too

modules/account-lib/src/coin/baseCoin/blsKeyPair.ts
```diff
@@ -84,11 +85,27 @@ export abstract class BlsKeyPair implements BaseKeyPair {
...
+  public static aggregatePubkeys(pubKeys: Uint8Array[]): Buffer {
+    this.ensureInitialized();
+    try {
+      return BLS.aggregatePubkeys(pubKeys);
+    } catch (e) {
+      throw new Error('Error aggregating pubkeys: ' + e);
+    }
+  }
+
+  private static ensureInitialized() {
```
marktoda, line 106: nit: doesn't need to be a static member, probably better as just a function

marktoda (approved):

marktoda (approved):

PR #962, opened 2020-12-09
bug: fix signing for Tezos

CapnMigraine:
Remove derivation for signing with the wallet base address that uses
the path /0/0 in order to prevent signature validation errors when
verifying against the root public key

Ticket: BG-27210


argjv (dismissed):

argjv (approved):
LGTM

PR #964, opened 2020-12-10
feat(eth2): add signMessage for ETH2

marktoda:
This commit implements the signMessage functionality for ETH2 so that we can perform key signatures
while generating wallets

### DIFF
--- a/modules/account-lib/test/unit/coin/eth2/keyPair.ts
+++ b/modules/account-lib/test/unit/coin/eth2/keyPair.ts
@@ -128,17 +128,15 @@ describe('Eth2 Key Pair', () => {
   });
 
   describe('Signatures', () => {
-    it('should accept 2-of-2 aggregated signature', () => {
+    it('should accept 2-of-2 aggregated signature', async () => {
       const msg = Buffer.from('test message');
       const keyPair1 = new KeyPair();
       const pub1 = Uint8Array.from(Buffer.from(keyPair1.getKeys().pub.slice(2), 'hex'));
-      const sec1 = Uint8Array.from(Buffer.from(keyPair1.getKeys().prv!.slice(2), 'hex'));
-      const sig1 = BLS.sign(sec1, msg);
+      const sig1 = keyPair1.sign(msg);
 
       const keyPair2 = new KeyPair();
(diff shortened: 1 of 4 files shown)


modules/account-lib/test/unit/coin/eth2/keyPair.ts
```diff
@@ -128,39 +128,36 @@ describe('Eth2 Key Pair', () => {
...
       const pub1 = Uint8Array.from(Buffer.from(keyPair1.getKeys().pub.slice(2), 'hex'));
-      const sec1 = Uint8Array.from(Buffer.from(keyPair1.getKeys().prv!.slice(2), 'hex'));
-      const sig1 = BLS.sign(sec1, msg);
+      const sig1 = await keyPair1.sign(msg);
 
       const keyPair2 = new KeyPair();
       const pub2 = Uint8Array.from(Buffer.from(keyPair2.getKeys().pub.slice(2), 'hex'));
-      const sec2 = Uint8Array.from(Buffer.from(keyPair2.getKeys().prv!.slice(2), 'hex'));
-      const sig2 = BLS.sign(sec2, msg);
+      const sig2 = keyPair2.sign(msg);
```
wesgraham, line 139: do we need to await sig2 as well?

marktoda: ah good catch actually we don't need to await either

wesgraham (approved):

modules/core/test/v2/unit/coins/eth2.ts
```diff
@@ -57,4 +59,24 @@ describe('Ethereum 2.0', function() {
...
+      const message = 'hello world';
+      const signature = await basecoin.signMessage(keyPair, message);
+
+      Eth2AccountLib.KeyPair.verifySignature(keyPair.pub, Buffer.from(message), signature).should.be.true();
+    });
+
+    it('should fail to validate a string message with wrong public key', async function() {
+      const keyPair = basecoin.generateKeyPair();
+      const message = 'hello world';
+      const signature = await basecoin.signMessage(keyPair, message);
```
wesgraham, line 75: await needed here?

PR #965, opened 2020-12-10
fix: replace sed with node script for replacing unsafe eval

tylerlevine:
The sed -i"" syntax for in-place replacement of unsafe eval code in the
webpack pipeline is not portable. This caused the unsafe eval code to
remain in the browser bundle since the release process was recently
switched onto an ubuntu machine.

Instead, replace invocation of sed -i"" with a simple node script that
is more portable.

Ticket: BG-27261

### DIFF
--- a/modules/core/webpack.config.js
+++ b/modules/core/webpack.config.js
@@ -1,9 +1,36 @@
 const path = require('path');
+const fs = require('fs');
 const webpack = require('webpack');
 const TerserPlugin = require('terser-webpack-plugin');
 const HTMLWebpackPlugin = require('html-webpack-plugin');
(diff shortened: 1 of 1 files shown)


tylerlevine:
Tested locally by commenting out lines 132-134.

If those lines are commented out, I get this:
```
$ grep -o "Function(\"return this\")()" ./dist/browser/BitGoJS.min.js | wc -l
79
```

If they are not commented out, I get this:
```
$ grep -o "Function(\"return this\")()" ./dist/browser/BitGoJS.min.js | wc -l
0
```

CapnMigraine (approved):
lgtm

PR #968, opened 2020-12-12
[DontMergeYet - BitGoExpress] Fixes generateWallet endpoint - Breaking change?

andreujuanc:
Hello,

Every single endpoint in the express app returns the data directly from the SDK; well, all except `/wallet/generate` which calls `generateWallet` that returns a `WalletWithKeychains` but only the `wallet` field is returned without the keys.

The problem is that critical data is lost, ie, the **keychains** that according to the JS SDK: https://github.com/BitGo/BitGoJS/blob/master/modules/core/src/v2/wallets.ts#L471 we should:

> Be sure to backup the backup keychain -- it is not stored anywhere else! 

The fix is just to simply return all the data from the SDK without taking only the "wallet._wallet" object (which is BTW supposed to be private https://github.com/BitGo/BitGoJS/blob/master/modules/core/src/v2/wallet.ts#L444 ).

### Possible breaking change
Current users are expecting a wallet object to be returned, so changing this will break integrations. 
Solution: We can create a separate endpoint `/wallet/generateWithKeyChains` but to be honest, people should be getting their private keys on this call.

### Blockers:
https://github.com/BitGo/BitGoJS/pull/963 has a commit that is not merged, so it also appears here because I'm fixing from rel/latest If needed I can just redo it from master.

PS: We are planning to go live soon, and this will delay our plans.

Thanks ❤️

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
-FROM node:10 AS builder
+FROM node:10-alpine AS builder
 MAINTAINER Tyler Levine <tyler@bitgo.com>
+RUN apk add --no-cache git python make g++
 COPY --chown=node:node . /tmp/bitgo/
 WORKDIR /tmp/bitgo/modules/express
 RUN npm install npm@latest -g
 USER node
 RUN npm ci && npm prune --production
 FROM node:10-alpine
-RUN apk add --no-cache tini
+RUN apk add --no-cache tini gcompat
 COPY --from=builder /tmp/bitgo/modules/express /var/bitgo-express
 ENV NODE_ENV production
 ENV BITGO_BIND 0.0.0.0
--- a/modules/express/package-lock.json
+++ b/modules/express/package-lock.json
@@ -61,9 +61,9 @@
       }
     },
     "@azure/core-http": {
-      "version": "1.1.9",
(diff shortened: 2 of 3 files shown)


tylerlevine:
Hey @andreujuanc, thanks for the PR and apologies for the slow response. Are you still interested in seeing this feature merged, or were you able to find an acceptable workaround? You're right that this is a breaking change for the express API so it's a little bit more involved than a backwards compatible change. We'll also need to do some documentation updates and get a test added for this (I can handle that part).

andreujuanc:
Hi  @tylerlevine , 
Not a problem, I have my own docker image using the code from my fork already. So personally I have no rush, but I would love to contribute to start a nice relationship since I'll be using a lot of Bitgo's services :)

How would you handle this? Maybe a minor version?

tylerlevine:
> Hi @tylerlevine ,
> Not a problem, I have my own docker image using the code from my fork already. So personally I have no rush, but I would love to contribute to start a nice relationship since I'll be using a lot of Bitgo's services :)
> 
> How would you handle this? Maybe a minor version?

I think we can avoid doing this as a breaking change by adding a `?includeKeychains=true` query param, would that work for your use case? If so, this can be a minor version.

andreujuanc:
I will close this in favor of your solution on https://github.com/BitGo/BitGoJS/pull/995

PR #969, opened 2020-12-14
Fix ETH2 lib initialization and key signatures

marktoda:
This commit fixes the key signatures by parsing the keys as hex rather
than utf-8, and fixes the BLS lib initialization by removing
setImmediate

CLOSES TICKET: BG-27361


marktoda:
Note that the lib initialization was previously not working when calling BitGoJS ETH2 functions from a script. However, when removing `setImmediate`, it worked. Strange :shrug:

wesgraham (approved):
Ah good catch

PR #974, opened 2020-12-21
feat(accountlib): BGA-656 Implement KeyPair Generation and Unit Tests

smaria95:
BGA-656 KeyPair Generation Implemented along with Unit Tests.

**Notes**:
- @bitgo/statics version has been updated to 6.1.0-rc.1 since it's required for unit tests in this task. Previous version was v6.0.0.
- Build is failing in drone due to a version issue with secp256k1 lib between BitGoJs and casper-client.

### DIFF
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -45,6 +45,7 @@
     "bignumber.js": "^9.0.0",
     "blake2b": "git+https://github.com/BitGo/blake2b.git#6268e6dd678661e0acc4359e9171b97eb1ebf8ac",
     "bs58check": "^2.1.2",
+    "casper-client-sdk": "^1.0.7",
     "elliptic": "^6.5.2",
     "ethereumjs-abi": "^0.6.5",
     "ethereumjs-tx": "2.1.2",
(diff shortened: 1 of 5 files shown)


modules/account-lib/package.json
```diff
@@ -33,7 +33,7 @@
   "license": "ISC",
   "dependencies": {
     "@bitgo/bls": "^0.1.0",
-    "@bitgo/statics": "^6.0.0",
+    "@bitgo/statics": "^6.1.0-rc.1",
```
tylerlevine, line 36: Please don't modify the versions of packages which are inside this monorepo (that's `@bitgo/statics`, `bitgo` or `@bitgo/account-lib` for now). These versions are managed by the monorepo management tool lerna during release, and modifying it outside a release branch will create merge conflicts when I release new versions.

tylerlevine (changes requested):

modules/account-lib/package.json
smaria95: Hi  @tylerlevine.

We need statics version 6.1.0 not only to run the tests but also to being able to import casper-client as dependency.
How should we proceed with this if we cannot upgrade this lib until release?

tylerlevine: When you are developing in this monorepo, the dependencies for packages within the monorepo are not downloaded and installed from npm. Instead they are symlinked to the version in the monorepo. If you have already made the changes you need in `statics`, then there is no need to update this version and you will get the version that currently exists in `master`.

CapnMigraine (changes requested):
Depends on https://github.com/BitGo/BitGoJS/pull/975

tylerlevine:
#975 is merged, please rebase on top of latest master

tylerlevine (approved):
lgtm, should be good once @CapnMigraine approves

CapnMigraine (approved):
lgtm

PR #975, opened 2020-12-21
feat: update secp256k1 in core to ^4

tylerlevine:
The library needed to support casper in account-lib requires
@types/secp256k1@4, but this causes type errors when compiling core,
since it uses a much older version.

This PR updates secp256k1 to 4.0.2, and fixes all issues caused by
breaking changes in the new major version.

Ticket: BG-27646

### DIFF
--- a/modules/core/src/bitcoin.ts
+++ b/modules/core/src/bitcoin.ts
@@ -13,7 +13,7 @@ const curve = ecurve.getCurveByName('secp256k1');
 const BigInteger = require('bigi');
 const createHmac = require('create-hmac');
 
-let secp256k1;
+let secp256k1: typeof import('secp256k1') | undefined;
 
 try {
   secp256k1 = require('secp256k1');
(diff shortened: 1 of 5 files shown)


marktoda (approved):

modules/core/src/bitcoin.ts
```diff
@@ -49,7 +49,7 @@ bitcoin.HDNode.prototype.getKey = getKey;
  * @param {string} path
  * @returns {bitcoin.HDNode}
  */
-export function deriveKeyByPath(userKey: bitcoin.HDNode, path: string): bitcoin.HDNode {
+export function deriveKeyByPath(userKey: bitcoin.HDNode, path: string): bitcoin.HdNode {
```
CapnMigraine, line 52: are both of these 2 references to hdnode correct? they have different spelling

tylerlevine: Good catch, will fix

tylerlevine:
I backed out a couple of unneeded changes and re-requested review.

CapnMigraine (approved):
lgtm, thanks!

PR #978, opened 2020-12-22
Enable batching eth transfers

karankurbur:
Wallet platform was denying sendMany calls because the isBatch field was not included in the output of signTransaction. The isBatch param is returns when calling /tx/build in wallet platform and is set automatically if the length of recipients is more than 1.

This commit fixed this issue by adding isBatch to the output of signTransaction and would pass the param value from tx/build.

https://bitgoinc.atlassian.net/browse/BG-27047

Ticket: BG-27047

### DIFF
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -107,6 +107,7 @@ interface SignFinalOptions {
     nextContractSequenceId?: number;
     hopTransaction?: string;
     backupKeyNonce?: number;
+    isBatch?: boolean;
   };
   signingKeyNonce: number;
   walletContractAddress: string;
@@ -593,6 +594,7 @@ export class Eth extends BaseCoin {
       const signature = Util.ethSignMsgHash(operationHash, Util.xprvToEthPrivateKey(userPrv));
 
       const txParams = {
+        isBatch: params.txPrebuild.isBatch,
         recipients: params.recipients,
         expireTime: expireTime,
         contractSequenceId: sequenceId,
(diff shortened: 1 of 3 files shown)


modules/core/src/v2/coins/eth.ts
```diff
@@ -593,6 +594,7 @@ export class Eth extends BaseCoin {
       const signature = Util.ethSignMsgHash(operationHash, Util.xprvToEthPrivateKey(userPrv));
 
       const txParams = {
+        isBatch: params.txPrebuild.isBatch === true,
```
marktoda, line 597: Why `=== true` instead of just the value?

modules/core/test/v2/integration/coins/eth.ts
```diff
@@ -17,6 +19,34 @@ describe('ETH:', function() {
...
+          {
+            amount: '5',
+            address: '0x431745b89e73230b3bc8a19e019194efb4b99efd',
+          },
+        ],
+        walletPassphrase: TestBitGo.V2.TEST_ETH_WALLET_PASSPHRASE,
+      });
+
+      should.exist(transaction);
+      transaction.should.have.property('transfer');
```
marktoda, line 41: Worth checking to make sure the entries are properly populated?

marktoda (changes requested):

modules/core/src/v2/coins/eth.ts
karankurbur: Wasn't sure if it was populated in build, made the change

modules/core/test/v2/integration/coins/eth.ts
karankurbur: Added check inside entries

marktoda (approved):

PR #981, opened 2021-01-04
Pass forwarderVersion flag

marktoda:
When creating ETH wallets, there is a forwarderVersion flag that can be
used to toggle between the old forwarders and the fee-improved
forwarders. This commit passes that flag when given to the createAddress
function

CLOSES TICKET: BG-28044


marktoda:
Tested change locally and it properly passed the flag

wesgraham (dismissed):
Commit message

wesgraham (approved):

PR #985, opened 2021-01-11
feat(support flush coins): support flushing native ETH coin

glowang:
Sometimes, clients might send ETH to an address before the wallet deployment tx has been confirmed due to fee spike, which results in ETH stuck in forwarders that need to be flushed. 

This PR adds a transaction type to the ETH transaction builder. This allows us to build flush tx in Wallet Platform later.

### DIFF
--- a/modules/account-lib/src/coin/eth/utils.ts
+++ b/modules/account-lib/src/coin/eth/utils.ts
@@ -30,7 +30,9 @@ import { KeyPair } from './keyPair';
 import {
   createForwarderMethodId,
   flushForwarderTokensMethodId,
+  flushCoinsMethodId,
   flushTokensTypes,
+  flushCoinsTypes,
   sendMultisigMethodId,
   sendMultisigTokenMethodId,
   sendMultiSigTokenTypes,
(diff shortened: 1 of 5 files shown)


marktoda (approved):
Awesome!! LGTM.

modules/account-lib/src/coin/eth/utils.ts
```diff
@@ -154,6 +156,16 @@ export function flushTokensData(forwarderAddress, tokenAddress): string {
   return addHexPrefix(Buffer.concat([method, args]).toString('hex'));
 }
 
+/**
+ * Get the data required to make a flush native coins contract call
+ */
+export function flushCoinsData(): string {
+  const params = [];
+  const method = EthereumAbi.methodID('flush', flushCoinsTypes);
+  const args = EthereumAbi.rawEncode(flushCoinsTypes, params);
```
marktoda, line 165: We could probably just make this an empty array and skip the types / params right?

marktoda: I.e. I think this function can just return `addHexPrefix(method).toString('hex')` for the same result

glowang: Yeah, just wanted to make it a bit more formalized :P

marktoda: Totally fine then!

tylerlevine:
next time, please avoid adding `console.log` s to the tests - these make the tests very noisy

PR #987, opened 2021-01-12
Adding new erc ofc tokens to statics

anthonylynch90:
Adding new erc20 tokens for ofc so that we can start using for WD. These
currently display on test as empty icons so this will allow the UI to
pick these up and display the ofc token images


tylerlevine (approved):
lgtm

PR #989, opened 2021-01-14
fix: adds wallet version support in core

wesgraham:
Adds support for the `walletVersion` flag in `generateWallet` requests
Ticket: BG-28508

### DIFF
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -96,6 +96,7 @@ export interface SupplementGenerateWalletOptions {
   enterprise?: string;
   disableTransactionNotifications?: boolean;
   gasPrice?: number;
+  walletVersion?: number;
   keys: string[];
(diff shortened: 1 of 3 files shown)


marktoda (approved):

tylerlevine (commented):
So now we are going to have v2 eth v1 and v2 eth v2 wallets? Yikes. Is there anything more descriptive we can use here? Maybe `walletContractVersion`? Also does this apply to any other coins? If not, maybe we should introduce a `coinSpecific` property, similar to the existing `krsSpecific` property in `GenerateWalletOptions`.

I won't block the PR but please consider a more generic approach here.

wesgraham:
Good point - this is something I'll make a ticket for @tylerlevine. `walletContractVersion` is not a bad idea, will require platform & api-doc changes as well (all will be included in he ticket). ForwarderVersion might also need to be renamed `forwarderContractVersion` as well.

PR #991, opened 2021-01-17
feat: add XRP skeleton

argjv:
Add XRP skeleton in account-lib.

CLOSES TICKET: BG-28586


FabianSimtlix (commented):
LGTM!

tylerlevine (approved):
Looks pretty good to me. I think I saw this yesterday when the tests were failing due to `ripple-lib` having trouble - did that get resolved?

argjv:
Yes, I removed the dependency since this is just a template. We will deal with the lib issue when we start implementation.
I'm gonna be adding a few more templates in the next couple of days.

PR #992, opened 2021-01-22
add support for IPC socket (unix socket) in Express

spoonincode:
Sometimes it can be desirable to further restrict access to the BitGo Express API endpoint beyond that of a localhost TCP connection. An effective low-effort way of accomplishing further restriction is by using a Unix domain socket instead. For example, the Unix domain socket can be placed in a part of the file hierarchy where only the Unix users/groups that require access to BitGo Express have access. Other users/groups will be unable to access the Unix socket.

This PR adds a new command line argument and environment variable to configure BitGo Express to listen on a Unix socket. Node.js refers to its API here as "IPC" due to the API working for both Unix sockets and named pipes on Win32. Thus, I've used this _IPC_ term for command line argument etc

### DIFF
--- a/modules/express/src/expressApp.ts
+++ b/modules/express/src/expressApp.ts
@@ -180,10 +180,14 @@ function createHttpServer(app: express.Application): http.Server {
  */
 export function startup(config: Config, baseUri: string): () => void {
   return function() {
-    const { env, customRootUri, customBitcoinNetwork } = config;
+    const { env, ipc, customRootUri, customBitcoinNetwork } = config;
     console.log('BitGo-Express running');
     console.log(`Environment: ${env}`);
-    console.log(`Base URI: ${baseUri}`);
+    if (ipc) {
+      console.log(`IPC path: ${ipc}`);
+    } else {
+      console.log(`Base URI: ${baseUri}`);
+    }
     if (customRootUri) {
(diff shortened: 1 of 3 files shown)


tylerlevine:
Hi @spoonincode, thanks for the PR :+1:. This feature is a nice improvement. I'll recreate this PR according to the [contribution guidelines](https://github.com/BitGo/BitGoJS/blob/master/DEVELOPERS.md#pull-requests) and add some docs and a test.

modules/express/src/expressApp.ts
```diff
@@ -299,9 +303,18 @@ export async function init(): Bluebird<void> {
...
   const server = await createServer(cfg, expressApp);
 
-  const { port, bind } = cfg;
+  const { port, bind, ipc } = cfg;
   const baseUri = createBaseUri(cfg);
 
-  server.listen(port, bind, startup(cfg, baseUri));
+  if (ipc) {
+    if (process.platform !== "win32" && fs.existsSync(ipc) && fs.statSync(ipc).isSocket()) {
+      fs.unlinkSync(ipc);
```
tylerlevine, line 311: I'm curious why we need to call `unlink` here before binding to the IPC socket - is there some magic here?

tylerlevine (approved):

tylerlevine:
#994 has been merged and will be in the next release

PR #993, opened 2021-01-25
Allow user defined gas price and limit

glowang:
Cumberland wants to customize their own gas price and limit in the WRW. This PR introduces the necessary backend changes

Ticket: BG-24734

### DIFF
--- a/modules/core/src/config.ts
+++ b/modules/core/src/config.ts
@@ -52,6 +52,15 @@ const formattedErc20Tokens = coins.reduce((acc: Erc20TokenConfig[], coin) => {
   return acc;
 }, []);
 
+export const ethGasConfigs = {
+  minimumGasPrice: 1000000000, // minimum gas price a user can provide (1 Gwei)
+  defaultGasPrice: 20000000000, // default gas price if estimation fails (20 Gwei)
+  maximumGasPrice: 2500000000000, // minimum gas price a user can provide (2500 Gwei)
+  defaultGasLimit: 500000, // Default gas limit we set for contract send
+  defaultGasLimitTokenSend: 1000000, // Default gas limit we set for token send
+  minimumGasLimit: 30000,    // minimum gas limit a user can set for a send
+  maximumGasLimit: 20000000, // Customers cannot set gas limits beyond this amount
+};
 // Get the list of Stellar tokens from statics and format it properly
 const formattedStellarTokens = coins.reduce((acc: StellarTokenConfig[], coin) => {
   if (coin instanceof StellarCoin) {
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -170,6 +170,8 @@ export interface RecoverOptions {
   walletContractAddress: string;
   recoveryDestination: string;
   krsProvider?: string;
+  gasPrice?: number;
+  gasLimit?: number;
 }
 
 export interface RecoveryInfo {
@@ -723,6 +725,42 @@ export class Eth extends BaseCoin {
   }
 
   /**
+   * Check whether the gas price passed in by user are within our max and min bounds
+   * If they are not set, set them to the defaults
+   * @param userGasPrice user defined gas price
+   * @returns the gas price to use for this transaction
+   */
+  setGasPrice(userGasPrice?: number): number {
+    if (!userGasPrice) {
+      return config.ethGasConfigs.defaultGasPrice;
+    }
+
+    const gasPriceMax = config.ethGasConfigs.maximumGasPrice;
+    const gasPriceMin = config.ethGasConfigs.minimumGasPrice;
+    if (userGasPrice < gasPriceMin || userGasPrice > gasPriceMax) {
+      throw new Error(`Gas price must be between ${gasPriceMin} and ${gasPriceMax}`);
+    }
+    return userGasPrice;
+  }
+  /**
+   * Check whether gas limit passed in by user are within our max and min bounds
+   * If they are not set, set them to the defaults
+   * @param userGasLimit user defined gas limit
+   * @returns the gas limit to use for this transaction
+   */
+  setGasLimit(userGasLimit?: number): number {
+    if (!userGasLimit) {
+      return config.ethGasConfigs.defaultGasLimit;
+    }
+    const gasLimitMax = config.ethGasConfigs.maximumGasLimit;
+    const gasLimitMin = config.ethGasConfigs.minimumGasLimit;
+    if (userGasLimit < gasLimitMin || userGasLimit > gasLimitMax) {
+      throw new Error(`Gas limit must be between ${gasLimitMin} and ${gasLimitMax}`);
+    }
+    return userGasLimit;
+  }
+
+  /**
    * Builds a funds recovery transaction without BitGo
    * @param params
    * @param params.userKey {String} [encrypted] xprv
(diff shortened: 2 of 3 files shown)


modules/core/src/v2/coins/eth.ts
```diff
@@ -170,8 +170,14 @@ export interface RecoverOptions {
   walletContractAddress: string;
   recoveryDestination: string;
   krsProvider?: string;
+  gasPrice?: number;
+  gasLimit?: number;
 }
 
+export interface GasInfo {
+  gasPrice: any;
+  gasLimit: any;
```
glowang, line 179: I would have made the types "BN", which is a variation of BigNumber.  However, `BN` is a type in the `ethereum-util` library, which is only imported dynamically upon use. Therefore, I can't really get Typescript type to work with it.

modules/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bitgo",
-  "version": "11.9.1",
+  "version": "11.10.1",
```
tylerlevine, line 3: Please do not modify the versions - this is done as part of the release process.

modules/core/src/v2/coins/eth.ts
tylerlevine: If the types are part of our API, we should depend on the library (or at least the types package) ourselves so we can safely export the type.

tylerlevine: How come these are `any` here when they are `number` in other places? Also, is there any place where these need to be numeric strings?

modules/core/src/v2/coins/eth.ts
```diff
@@ -722,6 +728,44 @@ export class Eth extends BaseCoin {
...
+   * @param userGasLimit user defined gas limit 
+   */
+  checkGasPriceAndLimit(userGasPrice?: number, userGasLimit?: number): GasInfo {
+    let gasLimit;
+    let gasPrice;
+
+    if(userGasLimit) {
+      // check whether the number makes sense
+      const gasLimitMax = config.ethGasConfigs.maximumGasLimit;
+      const gasLimitMin = config.ethGasConfigs.minimumGasLimit;
```
tylerlevine, line 744: Can you lift these out of the `if` statements since they are used in both of them?

modules/core/src/config.ts
```diff
@@ -52,6 +52,15 @@ const formattedErc20Tokens = coins.reduce((acc: Erc20TokenConfig[], coin) => {
   return acc;
 }, []);
 
+export const ethGasConfigs = {
```
tylerlevine, line 55: Shouldn't it be up to the platform to enforce the min/max/default gas price/limits? How can we hard code these values in the SDK in any reasonable way, given that we can't predict future changes in the gas market?

modules/core/test/v2/unit/recovery.ts
```diff
@@ -1024,24 +1023,71 @@ describe('Recovery:', function() {
...
-        '8GG0JE3DhXW3AUmz2G5RqMejdz1+t4/vovIP7lleegI7VYyWiiLvlM0OCFf3EVvV/RyXr8+2vsn\n' +
-        'Q0Vn8c2CV5FRZ80OjGYrW3A/6T/zpOz6E8CMvnD++iIpeO4r2eZJavejZxdzlxF0BRz7VI="}',
-        backupKey: '{"iv":"asB356ofC7nZtg4NBvQkiQ==","v":1,"iter":10000,"ks":256,"ts":64,"mode"\n' +
-        ':"ccm","adata":"","cipher":"aes","salt":"1hr2HhBbBIk=","ct":"8CZc6upt+XNOto\n' +
-        'KDD38TUg3ZUjzW+DraZlkcku2bNp0JS2s1g/iC6YTGUGtPoxDxumDlXwlWQx+5WPjZu79M8DCrI\n' +
-        't9aZaOvHkGH9aFtMbavFX419TcrwDmpUeQFN0hRkfrIHXyHNbTpGSVAjHvHMtzDMaw+ACg="}',
-        walletContractAddress: '0x5df5a96b478bb1808140d87072143e60262e8670',
-        walletPassphrase: TestBitGo.V2.TEST_RECOVERY_PASSCODE,
-        recoveryDestination: '0xac05da78464520aa7c9d4c19bd7a440b111b3054'
+      const error = await bitgo.getAsyncError(basecoin.recover({
```
tylerlevine, line 1044: Please avoid using `getAsyncError` here - this should really be deprecated and removed.

tylerlevine: `await basecoin.recover(...).should.be.rejectedWith('Gas limit must be between 30000 and 20000000')` is the preferred way of doing this.

tylerlevine (changes requested):
good start, but there are some conceptual problems here - SDK cannot hard code gas prices or limits in any reasonable way. I think the best we can do is pass along whatever the user gave us and enforcing limits/defaults should be done at the application or platform level where we can use a gas oracle.

modules/core/src/v2/coins/eth.ts
glowang: Not sure what you meant by used in both of them? These two are the gas_limit_ max and min, whereas the two below are gas_price_ max and min. Is this what you are referring to?

modules/core/package.json
glowang: Ah sorry, will revert this!

modules/core/src/config.ts
glowang: These values are hardcoded in platform which I copied over. However, I agree that it would be ideal to have only one source of truth and have SDK to call platform for the value. But there aren't exposed endpoints from Platform at this point and we still want to make SDK and platform's gas config consistent. 

We can certainly improve this gas configs once the routes are added to platform, but IMO this is an improvement , since we only had one hardcoded, default value in the SDK before. 

We have to get this change out for cumberland asap, what do you think about us moving forward here first and then working on the platform routes later?

modules/core/src/v2/coins/eth.ts
glowang: > How come these are any here when they are number in other places? Also, is there any place where these need to be numeric strings?

What you see on line 173 and 174 are the input formats, which are numbers. But `GasInfo` is the output type of `checkGasPriceAndLimit`, which were expected to be BN in the original implementation. So I made them BN to work with the old code.

modules/core/test/v2/unit/recovery.ts
glowang: Thanks!

modules/core/src/config.ts
glowang: Actually, I should be able to use the getFeeEstimate route!

modules/core/src/v2/coins/eth.ts
tylerlevine: Ah my mistake, you are right.

modules/core/src/config.ts
tylerlevine: SDK's fee estimation functions would use WP, so that might not be an option (unless I'm mistaken and we can source this info from a third party, I haven't verified this). If that's true, WRW would need some non-bitgo way of determining reasonable gas price/limit. I'm ok if we need to merge this with the hardcoded values you provided, but this is going to cause problems down the road if gas prices ever go real crazy again.

tylerlevine (approved):

PR #994, opened 2021-01-26
feat(express): add support for binding to an IPC socket (unix socket)

tylerlevine:
Sometimes it can be desirable to further restrict access to the BitGo
Express API endpoint beyond that of a localhost TCP connection. An
effective low-effort way of accomplishing further restriction is by
using a Unix domain socket instead. For example, the Unix domain socket
can be placed in a part of the file hierarchy where only the Unix
users/groups that require access to BitGo Express have access. Other
users/groups will be unable to access the Unix socket.

Ticket: BG-28829

### DIFF
--- a/modules/express/src/expressApp.ts
+++ b/modules/express/src/expressApp.ts
@@ -11,9 +11,8 @@ import * as debugLib from 'debug';
 import * as https from 'https';
 import * as http from 'http';
 import * as morgan from 'morgan';
+import * as fs from 'fs';
 import { Request as StaticRequest } from 'express-serve-static-core';
-
-const fs = Bluebird.promisifyAll(require('fs'));
 
 import { Config, config } from './config';
 
@@ -21,7 +20,7 @@ const debug = debugLib('bitgo:express');
 
 // eslint-disable-next-line @typescript-eslint/camelcase
 import { SSL_OP_NO_TLSv1 } from 'constants';
-import { NodeEnvironmentError, TlsConfigurationError } from './errors';
+import { IpcError, NodeEnvironmentError, TlsConfigurationError } from './errors';
 
 import { Environments } from 'bitgo';
 import { setupRoutes } from './clientRoutes';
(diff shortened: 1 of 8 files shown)


tylerlevine:
Recreated from #992 - still need to add docs and a test.

twtaylorbitgo (dismissed):

modules/express/src/expressApp.ts
```diff
@@ -150,10 +149,13 @@ function configureProxy(app: express.Application, config: Config): void {
  * @param app
  * @return {Server}
  */
-async function createHttpsServer(app: express.Application, config: Config): Promise<https.Server> {
+async function createHttpsServer(
+  app: express.Application,
+  config: Config & { keyPath: string; crtPath: string }
```
LeoKotschenreuther, line 154: What does the `&` and the object afterwards do? Is that a way to immediately have access to properties of the config object? In that case, wouldn't you no more need the `const { keyPath, crtPath } = config;` line anymore?

modules/express/src/expressApp.ts
```diff
@@ -196,7 +202,7 @@ export function startup(config: Config, baseUri: string): () => void {
 /**
  * helper function to determine whether we should run the server over TLS or not
  */
-function isTLS(config: Config): boolean {
+function isTLS(config: Config): config is Config & { keyPath: string; crtPath: string } {
```
LeoKotschenreuther, line 205: I think this function should still return a boolean and not a config object.

LeoKotschenreuther: This might just be really advanced TypeScript stuff, would you mind to explain?

tylerlevine: Sure, this is called a "User Defined Type Guard": https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

This function still returns a boolean, but this is a way to tell the compiler "if this function returns true, then we can have some more guarantees about the parameter type". More info here: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

tylerlevine: In this case, if `isTLS` returns true, we know the config must have the `keyPath` and `crtPath` properties, so we can make tell the compiler that these properties are now non-optional.

modules/express/src/expressApp.ts
tylerlevine: This is an [intersection type](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#intersection-types) which defines the type of the `config` parameter. It's saying that the type is `Config` (which has optional `keyPath` and `crtPath` properties) and'ed with the second type, which has non-optional `keyPath` and `crtPath` properties. This allows us to use these properties without checking them first in `createHttpsServer`.

We already check these properties in `isTLS` so there's no need to check them again here.

tylerlevine: This change was required since I removed the `Bluebird.promisifyAll` and just used the new `fs.promises` equivalents which are available starting in node 10. `Bluebird.promisifyAll` throws away all types and replaces them with `any` so we weren't able to do this at the type level before.

LeoKotschenreuther (approved):

PR #995, opened 2021-01-26
feat(express): add support for returning keychains with generated wallet

tylerlevine:
It's convenient for callers of `wallet/generate` to be able to get the
keychains for their newly generated wallet in the response, instead of
needing to make additional API calls.

To do this in a backwards-compatible way, this change adds a
`includeKeychains` query parameter which, if set, will return the return
of the SDK `generateWallet()` call directly, which includs the wallet
keychains.

Ticket: BG-28844

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -366,7 +366,7 @@ export interface SendManyOptions {
   memo?: Memo;
 }
 
-interface WalletData {
+export interface WalletData {
   id: string;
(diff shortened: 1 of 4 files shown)


tylerlevine:
recreated from #968 - still needs docs and a test.

andreujuanc:
This is good enough.  As long as it is clarified in the docs. 👍

LeoKotschenreuther (approved):

PR #996, opened 2021-01-26
Remove non existing ofcerc token and fix asset name for TERC

FloBitGo:
There is no `ERC` token. Therefore there can't be an `ofcerc` token and the underlying asset for `ofcterc` and `terc` should be `terc`, like all other ERC20 tokens.

Ticket: BG-28775


modules/statics/src/coins.ts
```diff
@@ -588,10 +588,9 @@ export const coins = CoinMap.fromCoins([
     '',
     'TST'
   ),
-  ofcerc20('ofcerc', 'ERC Token', 0, UnderlyingAsset.ERC),
   ofcerc20('ofcusdt', 'Tether', 6, UnderlyingAsset.USDT),
   ofcerc20('ofcusdc', 'USD Coin', 6, UnderlyingAsset.USDC),
-  tofcerc20('ofcterc', 'ERC Test Token', 0, UnderlyingAsset.ERC),
+  tofcerc20('ofcterc', 'Test ERC Token', 0, UnderlyingAsset.TERC),
   tofcerc20('ofctusdt', 'Test Tether', 6, UnderlyingAsset.USDT),
   tofcerc20('ofctusdc', 'Test USD Coin', 6, UnderlyingAsset.USDC),
```
jjn2009, line 595: as per our conversation in slack these basically dont exist I think? I've been directed to use TERC instead which suggests this actually doesnt exist on the ropstine network

 @marktoda

tylerlevine (dismissed):

modules/statics/src/coins.ts
marktoda: > ropstine network.


We use kovan tesnet not ropsten

marktoda (dismissed):

modules/statics/src/coins.ts
```diff
@@ -588,10 +591,4 @@ export const coins = CoinMap.fromCoins([
     '',
     'TST'
   ),
-  ofcerc20('ofcerc', 'ERC Token', 0, UnderlyingAsset.ERC),
-  ofcerc20('ofcusdt', 'Tether', 6, UnderlyingAsset.USDT),
-  ofcerc20('ofcusdc', 'USD Coin', 6, UnderlyingAsset.USDC),
```
jjn2009, line 593: these actually should exist in the future

jjn2009: just need to remove the testnet ones

FloBitGo:
Sorry folks, I had to rewrite history to fix the commit message, I would appreciate another review.

jjn2009 (approved):

marktoda (approved):

PR #1001, opened 2021-01-28
add TokenCreate transaction support

Dodecane:
Added support for the TokenCreate transaction (part of Hedera Token Service) along with test cases.


argjv:
Hello @Dodecane, thank you for your submission. We will be reviewing the code as part of the Hedera hackathon and get back to you soon.
Thanks!

PR #1002, opened 2021-01-29
build: update yarn.lock file

CapnMigraine:
casper-client-sdk was updated in account-lib, to v1.0.13 but yarn.lock wasn't

Ticket: BG-28987


tylerlevine (approved):
thanks!

PR #1005, opened 2021-02-01
feat(core): add compatibility for @bitgo/utxo-lib 1.9.x

OttoAllmendinger:
This change removes some uses of `@bitgo/utxo-lib` that are not
supported any longer in newer versions.

BG-22918


OttoAllmendinger:
also removed `tbtg` (we don't support it on platform anyway)

tylerlevine (approved):

PR #1007, opened 2021-02-01
fix(express): always prefer command line arguments to env var args

tylerlevine:
When the "IPC" argument was added, all subsequent configs had their
index increased by one. This revealed a bug where odd-numbered config
item indexes incorrectly preferred the env var arguments to the command
line arguments. Because the env vars defaulted to testnet, this broke
the ability to start bitgo express in the production configuration with
the `-e` flag on the command line.

Instead of reversing the configs in the `get()` call, just provide the
configs in the correct order to start with.

Ticket: BG-29084

### DIFF
--- a/modules/express/src/config.ts
+++ b/modules/express/src/config.ts
@@ -52,9 +52,9 @@ export const ArgConfig = (args): Partial<Config> => ({
 
 export const EnvConfig = (): Partial<Config> => ({
   port: Number(readEnvVar('BITGO_PORT')),
-  bind: readEnvVar('BITGO_BIND') || DefaultConfig.bind,
+  bind: readEnvVar('BITGO_BIND'),
   ipc: readEnvVar('BITGO_IPC'),
-  env: (readEnvVar('BITGO_ENV') as EnvironmentName) || DefaultConfig.env,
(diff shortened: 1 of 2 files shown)


modules/express/src/config.ts
```diff
@@ -83,18 +83,20 @@ export const DefaultConfig: Config = {
...
- * Earlier configs have higher precedence over subsequent configs.
+ * Later configs have higher precedence over earlier configs.
  */
 function mergeConfigs(...configs: Partial<Config>[]): Config {
   function isNilOrNaN(val: unknown): val is null | undefined | number {
     return isNil(val) || (isNumber(val) && isNaN(val));
   }
+
   // helper to get the first defined value for a given config key
   // from the config sources in a type safe manner
```
ericcrosson-bitgo, line 94: I _think_ this comment is accurate, since we reversed the args and nix'd the `.reverse`, but wanted to call your attention to it in case it needs to be updated

ericcrosson-bitgo (approved):

modules/express/src/config.ts
tylerlevine: Good catch, this actually produces the last defined value.

ericcrosson-bitgo (approved):
beautiful 🚀

PR #1010, opened 2021-02-02
chore: enable github action to close stale issues and PRs

tylerlevine:
Ticket: BG-28830

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


ericcrosson-bitgo:
Have we done a security audit of this GitHub Action / GH Actions in general?

I am wary of the security implications without an internal protocol

tylerlevine:
> Have we done a security audit of this GitHub Action / GH Actions in general?
> 
> I am wary of the security implications without an internal protocol

I don't believe this has undergone a security audit, but it should have no ability to impact the source code and should only be involved in automating the management of issues/PRs. If we think this should be audited, all the source code is available here: https://github.com/actions/stale

LeoKotschenreuther (commented):
How does this work? I feel like I'm missing a link here. I understand how to define the yml file but where does this run and how does this magic process know to look at the yml file?

tylerlevine:
It gets run through the magic of github actions: https://docs.github.com/en/actions/quickstart

Github will look for these yamls, and use an action runner to execute them when specified. It's a bit like drone in that it looks at certain "magic" files to determine what to do, but it's a bit more flexible because it's natively integrated with github.

LeoKotschenreuther (approved):

PR #1012, opened 2021-02-03
fix: wait a second between 2 subsequent API calls

glowang:
Recovery of an ETH transaction takes four calls to etherscan, but it bans the 4th call since it exceeds their rate limit of 3 calls per second. We wait one second between the 3rd and 4th call to avoid hitting this limit. 

Ticket: BG-29196


marktoda (approved):

tylerlevine (approved):
We should probably be agnostic to the underlying provider of the required on-chain data, but this is outside the scope of this work.

Nice work, thanks for the fix!

PR #1014, opened 2021-02-04
fix(core): removes automati…

twtaylorbitgo:
…c rebuilding of txes

Ticket: bg-29242

### DIFF
--- a/modules/core/src/v2/pendingApproval.ts
+++ b/modules/core/src/v2/pendingApproval.ts
@@ -360,15 +360,6 @@ export class PendingApproval {
         prebuildParams.hop = true;
       }
 
-      if (!recipients.length) {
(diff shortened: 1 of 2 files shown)


tylerlevine (changes requested):
I can't accept this PR as it would mean re-introducing the bug identified in BG-19020.

I think we know this isn't a great fix, but we have workarounds that we can use until we properly fix this with the unified build work.

twtaylorbitgo:
> I can't accept this PR as it would mean re-introducing the bug identified in BG-19020.
> 
> I think we know this isn't a great fix, but we have workarounds that we can use until we properly fix this with the unified build work.

ok yeah I remember us talking about this but just wanted to clearly state what's messing us up now

tylerlevine:
This is fixed in #1067

PR #1027, opened 2021-02-18
Bg 29498 fix eth recovery

glowang:
Fix testnet URL for etherscan and improve error catching for ETH recovery.

### DIFF
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -347,6 +347,10 @@ export class Eth extends BaseCoin {
         action: 'balance',
         address: address,
       });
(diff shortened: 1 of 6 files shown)


modules/core/src/v2/coins/eth.ts
```diff
@@ -347,6 +347,10 @@ export class Eth extends BaseCoin {
         action: 'balance',
         address: address,
       });
+      // throw if the result does not exist or the result is not a valid number
+      if (!result || !result.result || isNaN(result.result)) {
```
argjv, line 351: nit: I have a feeling there is a `lodash` command for this.

argjv (approved):

PR #1029, opened 2021-02-20
fix(core): AccountSet txn support

barathbitgo:
Add AccountSet txn support for explain transaction in SDK

TICKET:BG-29731


modules/core/src/v2/coins/xrp.ts
```diff
@@ -301,6 +301,26 @@ export class Xrp extends BaseCoin {
...
+          changeOutputs: [],
+          outputAmount: 0,
+          changeAmount: 0,
+          outputs: [],
+          fee: {
+            fee: transaction.Fee,
+            feeRate: null,
+            size: txHex.length / 2,
+          },
+          accountSet: {
```
tylerlevine, line 318: Does this property need to be added to the return type?

tylerlevine: I think callers will need to cast to `any` to use this property, maybe that's ok for now instead of doing some more refactoring.

modules/core/test/v2/unit/coins/xrp.ts
```diff
@@ -90,6 +90,18 @@ describe('XRP:', function() {
     signedExplanation.outputAmount.should.equal('253481');
   }));
 
+  it('Should be able to explain an XRP AccountSet transaction', co(function *() {
+    const signedExplanation = yield basecoin.explainTransaction({ txHex: '1200032400E5F4AA201B00E9C54768400000000000002D722102000000000000000000000000415F8315C9948AD91E2CCE5B8583A36DA431FB61730081145FB0771C7BCA6BBB7B2DAF362B7FEFC35AC5DF00F3E01073210228085BA918B150F05B34CE4613AC4A031A816866E143AA7470FB2044D79EAA1474473045022100A8D2B720EFA46A88B4267EB3EFBBA0A6F9432884BC7F8DBF0E962B76E95DE0DE022004430D10DC7B4D1B2D0555EA22FF73FEA2E91636B5715F8909A6D9BC7689A4AC8114E9B5B8F9EC3ACFFB31958A3C1CFBB3CE41CB0725E1F1' });
```
tylerlevine, line 94: Why have two cases here (it looks like the logic isn't different for signed vs unsigned)? If we do want to keep both of these, can they be split into two separate tests?

tylerlevine (changes requested):
a couple small nits, but overall looks fine

modules/core/test/v2/unit/coins/xrp.ts
barathbitgo: I kind of followed the test pattern in other tests: https://github.com/BitGo/BitGoJS/blob/4dae4d1d7f6d9fc560e1925f96d6c22b88cde585/modules/core/test/v2/unit/coins/xrp.ts#L84

Only difference is the way `explaintransaction` parses JSON and or decodes binary codec

Is this really need to be in two tests? my preference would be to keep it in one test like in other tests

modules/core/src/v2/coins/xrp.ts
barathbitgo: I don't think we actually define the return type anywhere. It seems like the callers know what to expect. One example is the explainTransaction for xlm. 
But we could make a case for defining the type within each coin in account-lib. Couple of things are missing in account-lib like explainTransaction and recover. I have couple of ideas and we could discuss it offline

argjv (approved):

CapnMigraine (approved):
lgtm

modules/core/src/v2/coins/xrp.ts
tylerlevine: Sounds good. I think callers will need to downcast but until we have a more generic way of extending the return types, this will have to do for now.

modules/core/test/v2/unit/coins/xrp.ts
tylerlevine: It doesn't need to be but if you're testing two separate things, it should really be two separate tests. I wouldn't use the existing tests as an argument for continuing the pattern - those tests should be split as well if they are testing two separate cases.

tylerlevine (approved):
approving despite comments, since they aren't feature critical

modules/core/test/v2/unit/coins/xrp.ts
barathbitgo: This is not separating two different things, It is kind of testing the same if then else block

PR #1031, opened 2021-02-22
feat(account-lib): STLX-1774 Update Casper SDK to version 1.0.20

smaria95:
Update Casper Client SDK from v1.0.19 to v1.0.20

[STLX-1774](https://bitgoinc.atlassian.net/browse/STLX-1774)

### DIFF
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -45,7 +45,7 @@
     "bignumber.js": "^9.0.0",
     "blake2b": "git+https://github.com/BitGo/blake2b.git#6268e6dd678661e0acc4359e9171b97eb1ebf8ac",
     "bs58check": "^2.1.2",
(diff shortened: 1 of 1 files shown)


CapnMigraine (commented):
Currently the latest version is `1.0.23`. Can we try and update to `v.1.0.22`?

smaria95:
> Can we try and update to the latest version `v.1.0.23`?

@CapnMigraine, I think we can and should do it on the next sprint.
I think Wallet-Platform is priority 1 right now and updating the SDK takes a lot of time to check if this new versions is working correctly not only on Account-Lib but also on all the existing scripts on coin-sandbox.

@rliberal-simtlix, please correct me if I'm wrong.

CapnMigraine (approved):

PR #1044, opened 2021-03-04
feat(core): sign functions for casper

smaria95:
Implement  SignTransaction and SignMessage for Casper and unit tests

This depends on #1046 (Merged)

[STLX-676](https://bitgoinc.atlassian.net/browse/STLX-676)

### DIFF
--- a/modules/core/src/v2/coins/cspr.ts
+++ b/modules/core/src/v2/coins/cspr.ts
@@ -4,12 +4,32 @@
 import * as Bluebird from 'bluebird';
 import * as accountLib from '@bitgo/account-lib';
 import { ECPair } from '@bitgo/utxo-lib';
-import { BaseCoin, KeyPair, SignedTransaction, VerifyAddressOptions, VerifyTransactionOptions } from '../baseCoin';
+
+import {
+  BaseCoin,
+  KeyPair,
+  SignedTransaction,
+  VerifyAddressOptions,
+  VerifyTransactionOptions,
+  SignTransactionOptions as BaseSignTransactionOptions,
+  TransactionPrebuild as BaseTransactionPrebuild,
+} from '../baseCoin';
+
 import { NodeCallback } from '../types';
 import { BitGo } from '../../bitgo';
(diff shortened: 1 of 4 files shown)


modules/core/src/v2/coins/cspr.ts
```diff
@@ -105,9 +125,39 @@ export class Cspr extends BaseCoin {
...
+  ): Bluebird<SignedTransaction> {
+    const self = this;
+    return co<SignedTransaction>(function*() {
+      const txBuilder = accountLib.getBuilder(self.getChain()).from(params.txPrebuild.txJson);
+      const key = params.prv;
+      txBuilder.sign({ key });
+
+      const transaction: any = yield txBuilder.build();
+      if (!transaction) {
+        throw new InvalidTransactionError('Error while trying to build transaction');
```
tylerlevine, line 150: Please create a new error type for this in `src/errors.ts` instead of throwing an error type from account-lib

tylerlevine (changes requested):

modules/core/test/v2/unit/coins/cspr.ts
```diff
@@ -1,20 +1,24 @@
-import * as Promise from 'bluebird';
-const co = Promise.coroutine;
-
+import { Promise as BluebirdPromise } from 'bluebird'; 
+import { Cspr as CsprAccountLib, register } from '@bitgo/account-lib';
 import { TestBitGo } from '../../../lib/test_bitgo';
 import { Tcspr } from '../../../../src/v2/coins';
 import { Cspr } from '../../../../src/v2/coins';
+import { Transaction } from '@bitgo/account-lib/dist/src/coin/cspr/transaction';
```
tylerlevine, line 6: How come this needs a path import? In general, this is something that should be avoided - if the type isn't exported from the package, it probably shouldn't be used by consumers of the package.

Since it's a test I won't block the PR, but please keep this in mind in the future.

tylerlevine (approved):

CapnMigraine (approved):
looks good

PR #1047, opened 2021-03-07
fix(account-lib): fix addSignature method of cspr transaction

CapnMigraine:
Set signature on transaction deploy

Ticket: STLX-2076

PR #1048, opened 2021-03-08
fix(account-lib): fix processSigning method of cspr transactionBuilder

CapnMigraine:
Set signatures on txBuilder deploy only once per call to the build() method

Depends on: #1047 

Ticket: STLX-2077

### DIFF
--- a/modules/account-lib/src/coin/cspr/transactionBuilder.ts
+++ b/modules/account-lib/src/coin/cspr/transactionBuilder.ts
@@ -16,6 +16,7 @@ import { KeyPair } from './keyPair';
 import { Fee, CasperModuleBytesTransaction, CasperTransferTransaction, SignatureData } from './ifaces';
 import { isValidAddress } from './utils';
 import { SECP256K1_PREFIX, CHAIN_NAME, TRANSACTION_EXPIRATION } from './constants';
+import * as _ from "lodash";
 
 export const DEFAULT_M = 3;
 export const DEFAULT_N = 2;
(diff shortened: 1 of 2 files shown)


rliberal-simtlix (approved):

modules/account-lib/src/coin/cspr/transactionBuilder.ts
```diff
@@ -305,12 +306,35 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
   }
 
+  /**
+   * Checks whether the transaction has the owner signature
+   * @param {string} pub - public key of the signer
+   * @returns {boolean} true if the pub key already signed th transaction
+   * @private
+   */
+  private isTransactionSignedByPub(pub: string): boolean {
+    return _.findIndex(this.transaction.casperTx.approvals, approval => {
```
rliberal-simtlix, line 316: Function looks good, but difficult to read.

rliberal-simtlix (dismissed):

modules/account-lib/src/coin/cspr/transactionBuilder.ts
```diff
@@ -305,12 +306,35 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+      const approvalSigner = approval.signer.slice(2).toUpperCase();
```
rliberal-simtlix, line 317: Better if you can use something like this.
https://github.com/BitGo/BitGoJS/blob/STLX-2076-cspr-fix-addSignature/modules/account-lib/src/coin/cspr/constants.ts#L1

modules/account-lib/src/coin/cspr/transactionBuilder.ts
```diff
@@ -305,12 +306,35 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+      return approvalSigner === pub;
+    }) === -1;
```
smaria95, line 319: Shouldn't this be `!== -1`?
For what I can see, findIndex would return a number >= 0 iff `pub` has signed the transaction. So I think it should return true in that case.

smaria95 (changes requested):
Just one correction.

modules/account-lib/src/coin/cspr/transactionBuilder.ts
CapnMigraine: You're right, I had the logic inverted and the checks in transactionBuilder.ts too.
Fixed.

modules/account-lib/src/coin/cspr/transactionBuilder.ts
CapnMigraine: Fixed with a new utils method called `removeAlgoPrefixFromHexValue`

rliberal-simtlix (approved):

smaria95 (approved):
LGTM!

PR #1050, opened 2021-03-08
feat(account-lib): add stx coin (blockstack) and supporting utils

fariedt:
This PR adds support for the Stacks keypair implementation, along with tests.

The PR is extracted from the exploratory code in [feat/stacks-tx-builer](https://github.com/fariedt/BitGoJS/tree/feat/stacks-tx-builder).

Cc: @wesgraham

### DIFF
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -39,6 +39,7 @@
     "@hashgraph/sdk": "1.4.6",
     "@stablelib/hex": "^1.0.0",
     "@stablelib/sha384": "^1.0.0",
+    "@stacks/transactions": "1.3.0",
     "@taquito/local-forging": "6.3.5-beta.0",
     "@taquito/signer": "6.3.5-beta.0",
     "@types/lodash": "^4.14.151",
@@ -53,10 +54,10 @@
     "libsodium-wrappers": "^0.7.6",
     "lodash": "^4.17.15",
     "protobufjs": "^6.8.9",
+    "secp256k1": "4.0.2",
     "stellar-sdk": "^0.11.0",
     "tronweb": "^2.7.2",
-    "tweetnacl": "^1.0.3",
-    "secp256k1": "4.0.2"
+    "tweetnacl": "^1.0.3"
   },
   "devDependencies": {
     "@types/elliptic": "^6.4.12",
--- a/modules/account-lib/src/coin/stx/utils.ts
+++ b/modules/account-lib/src/coin/stx/utils.ts
@@ -1,0 +1,167 @@
+import BigNumber from 'bignumber.js';
+import { bufferToHex, stripHexPrefix } from 'ethereumjs-util';
+import {
+  AddressHashMode,
+  StacksTransaction,
+  TransactionVersion,
+  addressFromVersionHash,
+  addressHashModeToVersion,
+  addressToString,
+  validateStacksAddress,
+} from '@stacks/transactions';
+import { ec } from 'elliptic';
+
+/**
+ * Encodes a buffer as a "0x" prefixed lower-case hex string.
+ *
+ * @param {Buffer} buff - a buffer with a hexadecimal string
+ * @returns {string} - the hexadecimal string prefixed with "0x"
+ */
+export function bufferToHexPrefixString(buff: Buffer): string {
+  return bufferToHex(buff);
+}
+
+/**
+ * Remove the "0x" prefix from the given string, if present.
+ *
+ * @param {string} hex - a hexadecimal string
+ * @returns {string} - the hexadecimal string without a leading "0x"
+ */
+export function removeHexPrefix(hex: string): string {
+  return stripHexPrefix(hex);
+}
+
+/**
+ * @param publicKeyHash
+ * @param hashMode
+ * @param transactionVersion
+ */
+function getAddressFromPublicKeyHash(
+  publicKeyHash: Buffer,
+  hashMode: AddressHashMode,
+  transactionVersion: TransactionVersion,
+): string {
+  if (publicKeyHash.length !== 20) {
+    throw new Error('expected 20-byte pubkeyhash');
+  }
+
+  const addrVer = addressHashModeToVersion(hashMode, transactionVersion);
+  const addr = addressFromVersionHash(addrVer, publicKeyHash.toString('hex'));
+  const addrString = addressToString(addr);
+  return addrString;
+}
+
+/**
+ * @param tx
+ */
+export function getTxSenderAddress(tx: StacksTransaction): string {
+  if (tx.auth.spendingCondition !== null && tx.auth.spendingCondition !== undefined) {
+    const spendingCondition = tx.auth.spendingCondition;
+    const txSender = getAddressFromPublicKeyHash(
+      Buffer.from(spendingCondition.signer, 'hex'),
+      spendingCondition.hashMode as number,
+      tx.version,
+    );
+    return txSender;
+  } else throw new Error('spendingCondition should not be null');
+}
+
+/**
+ * Returns whether or not the string is a valid amount number
+ *
+ * @param {string} amount - the string to validate
+ * @returns {boolean} - the validation result
+ */
+export function isValidAmount(amount: string): boolean {
+  const bigNumberAmount = new BigNumber(amount);
+  return bigNumberAmount.isInteger() && bigNumberAmount.isGreaterThanOrEqualTo(0);
+}
+
+/**
+ * Returns whether or not the string is a valid protocol address
+ *
+ * @param {string} address - the address to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidAddress(address: string): boolean {
+  return validateStacksAddress(address);
+}
+
+/**
+ * Returns whether or not the string is a valid protocol transaction id or not.
+ *
+ * A valid transaction id is a SHA-512/256 hash of a serialized transaction; see
+ * the txidFromData function in @stacks/transaction:
+ * https://github.com/blockstack/stacks.js/blob/master/packages/transactions/src/utils.ts#L97
+ *
+ * @param {string} txId - the transaction id to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidTransactionId(txId: string): boolean {
+  if (txId.length !== 64 && txId.length !== 66) return false;
+  const noPrefix = removeHexPrefix(txId);
+  if (noPrefix.length !== 64) return false;
+
+  return allHexChars(noPrefix);
+}
+
+/**
+ * Returns whether or not the string is a valid protocol public key
+ *
+ * The key format is documented at
+ * https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md#transaction-authorization
+ *
+ * @param {string} pub - the  public key to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidPublicKey(pub: string): boolean {
+  if (pub.length !== 66 && pub.length !== 130) return false;
+
+  const firstByte = pub.slice(0, 2);
+
+  // uncompressed public key
+  if (pub.length === 130 && firstByte !== '04') return false;
+
+  // compressed public key
+  if (pub.length === 66 && firstByte !== '02' && firstByte !== '03') return false;
+
+  if (!allHexChars(pub)) return false;
+
+  // validate the public key
+  const secp256k1 = new ec('secp256k1');
+  try {
+    const keyPair = secp256k1.keyFromPublic(Buffer.from(pub, 'hex'));
+    const { result } = keyPair.validate();
+    return result;
+  } catch (e) {
+    return false;
+  }
+}
+
+/**
+ * Returns whether or not the string is a valid protocol private key
+ *
+ * The key format is described in the @stacks/transactions npm package, in the
+ * createStacksPrivateKey function:
+ * https://github.com/blockstack/stacks.js/blob/master/packages/transactions/src/keys.ts#L125
+ *
+ * @param {string} prv - the  private key to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidPrivateKey(prv: string): boolean {
+  if (prv.length !== 64 && prv.length !== 66) return false;
+
+  if (prv.length === 66 && prv.slice(64) !== '01') return false;
+
+  return allHexChars(prv);
+}
+
+/**
+ * Returns whether or not the string is a composed of hex chars only
+ *
+ * @param {string} maybe - the  string to be validated
+ * @returns {boolean} - the validation result
+ */
+function allHexChars(maybe: string): boolean {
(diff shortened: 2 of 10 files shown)


modules/statics/src/base.ts
```diff
@@ -33,6 +33,7 @@ export enum CoinFamily {
   OFC = 'ofc',
   RMG = 'rmg',
   RBTC = 'rbtc',
+  STX = 'stx',
```
wesgraham, line 36: Can we pull the statics addition out into a new PR please? Will need to merge it individually, then cut a release and update the package.json of account-lib to support the latest statics version (with stacks)

modules/account-lib/package.json
```diff
@@ -39,6 +39,7 @@
     "@hashgraph/sdk": "1.4.6",
     "@stablelib/hex": "^1.0.0",
     "@stablelib/sha384": "^1.0.0",
+    "@stacks/transactions": "1.2.4",
```
wesgraham, line 42: Can we use stacks/transactions 1.3.0 - re: this message from the blockstacks team: 
```
Hey @Wesley Graham! wanted to let you know that we just released version 1.3.0 of @stacks/transactions which includes support for uncompressed keys. Please let me know if you have any further issues with this!
```

modules/account-lib/src/coin/stacks/utils.ts
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * @param {string} pub - the  public key to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidPublicKey(pub: string): boolean {
+  if (pub.length !== 66 && pub.length !== 130) return false;
+
+  const firstByte = pub.slice(0, 2);
+
+  if (pub.length === 130) {
```
wesgraham, line 110: We try to avoid nested ifs for readability - can we reorg to 
`
if (pub.length !== 130) { return false }
if (pub.length === 130 && firstbyte == '04' return allHexChars(pub)
`

modules/account-lib/src/coin/stacks/keyPair.ts
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @param {string} prv A raw private key
+   */
+  recordKeysFromPrivateKey(prv: string): void {
+    if (!isValidPrivateKey(prv)) {
+      throw new Error('Unsupported private key');
+    }
+
+    const compressed = prv.length === 66;
+    this.keyPair = ECPair.fromPrivateKeyBuffer(Buffer.from(prv.slice(0, 64), 'hex'));
```
wesgraham, line 51: Do we ignore the final byte in the case the key is compressed?

modules/account-lib/src/coin/stacks/keyPair.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.keyPair.compressed = compressed;
+  }
+
+  /**
+   * Build an ECPair from a public key.
+   *
+   * The public key is either 33 or 65 bytes long (66 or 130 characters in hex),
```
wesgraham, line 58: nit: can you clarify comment? Public key is 33 or 65 bytes? Below you also mention that if it is 32 bytes it's uncompressed, so can public key be 32 as well then?

modules/account-lib/src/coin/stacks/keyPair.ts
```diff
@@ -0,0 +1,95 @@
...
+      prv: this.keyPair.d ? this.keyPair.d.toBuffer(32).toString('hex') : undefined,
+    };
+  }
+
+  /**
+   * Get a public address of an uncompressed public key.
+   *
+   * @returns {string} The public address
+   */
+  getAddress(): string {
```
wesgraham, line 92: merits in allowing `compressed` to be an optional variable here?

modules/account-lib/src/coin/stacks/utils.ts
```diff
@@ -0,0 +1,142 @@
...
+ * @returns {string} - the hexadecimal string prefixed with "0x"
+ */
+export function bufferToHexPrefixString(buff: Buffer): string {
+  return '0x' + buff.toString('hex');
+}
+
+/**
+ * @param hex
+ */
+export function removeHexPrefix(hex: string): string {
```
wesgraham, line 25: merits in using `stripHexPrefix` from `ethereumjs-util`? Trying to avoid reinventing the wheel

modules/account-lib/src/coin/stacks/utils.ts
```diff
@@ -0,0 +1,142 @@
...
+ * @param hashMode
+ * @param transactionVersion
+ */
+function getAddressFromPublicKeyHash(
+  publicKeyHash: Buffer,
+  hashMode: AddressHashMode,
+  transactionVersion: TransactionVersion,
+): string {
+  const addrVer = addressHashModeToVersion(hashMode, transactionVersion);
+  if (publicKeyHash.length !== 20) {
```
wesgraham, line 41: nit: can we move this above `addressHashModeToVersion` line?

modules/account-lib/src/coin/stacks/utils.ts
```diff
@@ -0,0 +1,142 @@
...
+  const addr = addressFromVersionHash(addrVer, publicKeyHash.toString('hex'));
+  const addrString = addressToString(addr);
+  return addrString;
+}
+
+/**
+ * @param tx
+ */
+export function getTxSenderAddress(tx: StacksTransaction): string {
+  if (tx.auth.spendingCondition !== null && tx.auth.spendingCondition !== undefined) {
```
wesgraham, line 53: https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md#transaction-authorization

want to make sure we cover for sponsored transactions as well - do we clean/verify for both spending conditions?

modules/account-lib/src/coin/stacks/utils.ts
```diff
@@ -0,0 +1,142 @@
...
+  return validateStacksAddress(address);
+}
+
+/**
+ * Returns whether or not the string is a valid protocol transaction id or not
+ *
+ * @param {string} txId - the transaction id to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidTransactionId(txId: string): boolean {
```
wesgraham, line 91: source

wesgraham:
Nicely done on the whole! Just left a few nits + a few structural changes. Notably, can we pull the statics change into its own PR & add sources for some of the sanitizer methods? Also, would love if it would be possible to have some tests around creating a multisig account from the generate keys methods (https://stacks-js-git-master-blockstack.vercel.app/globals.html#addressfrompublickeys)

modules/account-lib/src/coin/stacks/keyPair.ts
fariedt: `getAddress` in the parent class takes an optional `AddressFormat` -- a string that's either `hex` or `base58`; adding support for a boolean for `compressed` breaks that signature.

Perhaps this function should be something like

```
import { getAddressFromPublicKey, TransactionVersion } from '@stacks/transactions';

  /**
   * Get a public address of an uncompressed public key.
   *
   * @param {boolean} compressed - Compress public key (defaults to false)
   * @param {TransactionVersion} network - select Mainnet or Testnet for the address
   * @returns {string} The public address
   */
  getAddress(compressed = false, network = TransactionVersion.Mainnet): string {
    return getAddressFromPublicKey(this.getKeys(compressed).pub, network);
  }
```

This lets us generate addresses for both the Stacks mainnet and testnet.  The address encodes the network [internally](https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md#accounts) and in string format using the [account generation rules](https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md#transaction-authorization): `SP...` for single-signature transactions on the mainnet and `ST...` on the testnet.

modules/account-lib/src/coin/stacks/utils.ts
fariedt: This does not appear to be documented in SIP-005.  However, the code to generate a transaction id is in the `txid()` method in the [StacksTransaction](https://github.com/blockstack/stacks.js/blob/master/packages/transactions/src/transaction.ts#L176) class.  It performs a SHA-512/256 [hash](https://github.com/blockstack/stacks.js/blob/3f889aab44fa62c480258213a920e30f29833c2a/packages/transactions/src/utils.ts#L97) of the serialized transaction.

modules/statics/src/base.ts
fariedt: I've removed the statics-related commits for now.

modules/account-lib/package.json
fariedt: I've updated the dependency to 1.3.0.

modules/account-lib/src/coin/stacks/utils.ts
fariedt: This is done.

modules/account-lib/src/coin/stacks/keyPair.ts
wesgraham: Yes please, let's go with this approach

modules/account-lib/src/coin/stacks/utils.ts
```diff
@@ -0,0 +1,142 @@
...
+  return allHexChars(pub);
+}
+
+/**
+ * Returns whether or not the string is a valid protocol private key
+ *
+ * @param {string} prv - the  private key to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidPrivateKey(prv: string): boolean {
```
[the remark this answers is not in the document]
wesgraham: Sounds good, can you add https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md as a source in the comments?

modules/account-lib/src/coin/stacks/keyPair.ts
```diff
@@ -0,0 +1,93 @@
...
+   * compressed public key.  A prefix of 0x04 denotes an uncompressed public key.
+   *
+   * @param {string} pub A raw public key
+   */
+  recordKeysFromPublicKey(pub: string): void {
+    if (!isValidPublicKey(pub)) {
+      throw new Error('Unsupported public key');
+    }
+
+    this.keyPair = ECPair.fromPublicKeyBuffer(Buffer.from(pub, 'hex'));
```
wesgraham, line 69: Want to confirm this works for both compressed and uncompressed?

wesgraham: can we add unit tests for both as a sanity check?

wesgraham:
Very close, just a few comments left :)

modules/account-lib/src/coin/stacks/keyPair.ts
fariedt: The tests in `test/unit/coin/stacks/keyPair.ts` cover both cases --
```
    it('from an uncompressed public key', () => {
      const keyPair = new Stacks.KeyPair({ pub: testData.pubKey2 });
      should.equal(keyPair.getKeys(false).pub, testData.pubKey2);
      should.equal(keyPair.getKeys(true).pub, testData.pubKey2Compressed);
    });

    it('from a compressed public key', () => {
      const keyPair = new Stacks.KeyPair({ pub: testData.pubKey2Compressed });
      should.equal(keyPair.getKeys(false).pub, testData.pubKey2);
      should.equal(keyPair.getKeys(true).pub, testData.pubKey2Compressed);
    });
```

modules/account-lib/src/coin/stacks/keyPair.ts
fariedt: I could not figure out how to reuse the `getAddress` type signature in the base class, so I created a separate function for this.

modules/account-lib/src/coin/stacks/utils.ts
fariedt: I've added a comment for this.

modules/account-lib/src/coin/stacks/utils.ts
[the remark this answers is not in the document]
fariedt: I've added a comment for this.

wesgraham (approved):
LGTM

wesgraham (approved):

modules/account-lib/src/coin/stx/utils.ts
```diff
@@ -0,0 +1,156 @@
...
+ * Returns whether or not the string is a valid protocol public key
+ *
+ * The key format is documented at
+ * https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md#transaction-authorization
+ *
+ * @param {string} pub - the  public key to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidPublicKey(pub: string): boolean {
+  if (pub.length !== 66 && pub.length !== 130) return false;
```
tylerlevine, line 117: There's no check for actual key validity here (only key format), so this will accept invalid (off curve) public keys. Not sure how much of an issue that is here, but it's worth calling out.

modules/account-lib/src/coin/stx/utils.ts
```diff
@@ -0,0 +1,156 @@
...
+ * Returns whether or not the string is a valid protocol private key
+ *
+ * The key format is described in the @stacks/transactions npm package, in the
+ * createStacksPrivateKey function:
+ * https://github.com/blockstack/stacks.js/blob/master/packages/transactions/src/keys.ts#L125
+ *
+ * @param {string} prv - the  private key to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidPrivateKey(prv: string): boolean {
```
tylerlevine, line 140: I read the code linked above, but it kind of makes no sense to have a compressed private key - I'm not even sure what that means in the context of the secp256k1 curve (usually public keys are compressed, but not private keys). I'll assume that this is some strange implementation detail

modules/account-lib/src/coin/stx/utils.ts
```diff
@@ -0,0 +1,156 @@
...
+export function removeHexPrefix(hex: string): string {
+  return stripHexPrefix(hex);
+}
+
+/**
+ * @param publicKeyHash
+ * @param hashMode
+ * @param transactionVersion
+ */
+function getAddressFromPublicKeyHash(
```
tylerlevine, line 38: How do we get an address from just one public key? Is this capable of creating a multisig address? Are we doing single sig for STX? Feels like I'm missing something here.

tylerlevine (commented):
I am definitely missing some context here - would be helpful to fill in some of the missing pieces so we can be confident that the mechanics here are correct before I accept. Alternatively, getting a review from someone with a bit more background may be good.

modules/account-lib/src/coin/stx/utils.ts
fariedt: This code is part of the transactions PR.  I can remove it for now if needed.

wesgraham (approved):

PR #1052, opened 2021-03-08
fix(account-lib): fix getTransferId method of cspr utils

CapnMigraine:
Return `undefined` if there's no transferId set on the deploy instead of throwing
The transferId value is optional

Depends on: #1047

Ticket: STLX-2082

### DIFF
--- a/modules/account-lib/src/coin/cspr/transferBuilder.ts
+++ b/modules/account-lib/src/coin/cspr/transferBuilder.ts
@@ -48,7 +48,10 @@ export class TransferBuilder extends TransactionBuilder {
     this.transaction.setTransactionType(TransactionType.Send);
     this.to(getTransferDestinationAddress(tx.casperTx.session));
     this.amount(getTransferAmount(tx.casperTx.session));
-    this.transferId(getTransferId(tx.casperTx.session));
+    const transferId = getTransferId(tx.casperTx.session);
+    if (transferId !== undefined) {
+      this.transferId(transferId);
+    }
   }
 
   /** @inheritdoc */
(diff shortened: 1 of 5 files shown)


smaria95 (approved):
LGTM

smaria95:
@CapnMigraine , Unit Tests are failing. 
Most probably due to the change to getTransferId method. There must be some `should.throw` tests

modules/account-lib/src/coin/cspr/utils.ts
```diff
@@ -117,17 +117,17 @@ export function getTransferAmount(transferTx: ExecutableDeployItem): string {
  * Get transfer id from deploy transfer session
  *
  * @param {ExecutableDeployItem} transferTx transfer session
- * @returns {number} the transfer id
```
tylerlevine, line 120: Seems like returning `undefined` and making the return type `number | undefined` would be a bit more natural here? Is that going to cause issues somewhere else?

CapnMigraine: I agree. Fixed.

modules/account-lib/src/coin/cspr/transferBuilder.ts
```diff
@@ -48,7 +48,10 @@ export class TransferBuilder extends TransactionBuilder {
     this.transaction.setTransactionType(TransactionType.Send);
     this.to(getTransferDestinationAddress(tx.casperTx.session));
     this.amount(getTransferAmount(tx.casperTx.session));
-    this.transferId(getTransferId(tx.casperTx.session));
+    const transferId = getTransferId(tx.casperTx.session);
+    if (transferId) {
```
tylerlevine, line 52: is `0` a valid transfer id? If so, an explicit for `transferId === undefined` would be preferred.

CapnMigraine: good catch! fixed

tylerlevine (approved):

PR #1054, opened 2021-03-09
feat: add log for xpub during tx signing

glowang:
Currently we don't have a way to see which exact xprv user uses when signing a tx, which makes it hard to debug. This PR 1) sanity checks that the user is not providing an xpub and 2) logs out the corresponding xpub of the xprv that they are using.

BG-24535

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1064,6 +1064,12 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       }
 
       const keychain = bitcoin.HDNode.fromBase58(userPrv);
+
+      if (keychain.toBase58() === keychain.neutered().toBase58()) {
+        throw new Error('expected user private key but received public key');
+      }
+      debug(`Here is the public key of the xprv you used to sign: ${keychain.neutered().toBase58()}`);
+
       const keychainHdPath = hdPath(keychain);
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1064,6 +1064,12 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       }
 
       const keychain = bitcoin.HDNode.fromBase58(userPrv);
+
+      if (keychain.toBase58() === keychain.neutered().toBase58()) {
+        throw new Error('user private key is only public');
```
argjv, line 1069: not: `user private key is only public`/`invalid user private key, a public key was provided` or something along those lines

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1064,6 +1064,12 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       }
 
       const keychain = bitcoin.HDNode.fromBase58(userPrv);
```
tylerlevine, line 1066: Is there some case where we are providing an `xpub` as the `prv` param?

tylerlevine (approved):

modules/core/src/v2/coins/abstractUtxoCoin.ts
glowang: yes I added a test case for this.

PR #1055, opened 2021-03-10
stacks: implementation of transaction and transfer builders

fdmirza:
Branch based on `feat/stacks-keypair`
Should be merged after  https://github.com/BitGo/BitGoJS/pull/1050

cc: @wesgraham


modules/account-lib/src/coin/stx/keyPair.ts
```diff
@@ -76,9 +76,18 @@ export class KeyPair extends Secp256k1ExtendedKeyPair {
    * @returns {DefaultKeys} The keys in the protocol default key format
    */
   getKeys(compressed = false): DefaultKeys {
+    let prv;
```
wesgraham, line 79: nit: can we reformat to 

`let prv;`
`prv = this.keyPair.d ? this.keyPair.d.toBuffer(32).toString('hex') : undefined`
`prv = prv && compressed ? prv + '01' : prv`

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,171 @@
...
+  private _stxTransaction: StacksTransaction;
+  protected _type: TransactionType;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  /** @inheritdoc */
+  canSign(key: BaseKey): boolean {
+    // TODO: implement this validation method
```
wesgraham, line 35: What is the blocker here?

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,171 @@
...
+    return true;
+  }
+
+  async sign(keyPair: KeyPair[] | KeyPair): Promise<void> {
+    const keyPairs = keyPair instanceof Array ? keyPair : [keyPair];
+    const signer = new TransactionSigner(this._stxTransaction);
+    for (const kp of keyPairs) {
+      const keys = kp.getKeys(true);
```
wesgraham, line 43: Is compressed always true here?

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,171 @@
...
+      if (!keys.prv) {
+        throw new SigningError('Missing private key');
+      }
+      const privKey = createStacksPrivateKey(keys.prv);
+      signer.signOrigin(privKey);
```
wesgraham, line 48: Do we not save any state in this method? We don't set _stxTransaction or any signature data or anything like that?

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+  }
+
+  async appendOrigin(keyPair: KeyPair): Promise<void> {
+    const keys = keyPair.getKeys(true);
```
wesgraham, line 53: Again is compressed always true?

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (!keys.pub) {
+      throw new SigningError('Missing public key');
+    }
+    const pubKey = createStacksPublicKey(keys.pub);
+    const signer = new TransactionSigner(this._stxTransaction);
+    signer.appendOrigin(pubKey);
```
wesgraham, line 59: Again do we save no state?

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,171 @@
...
+      case StacksMessageType.MessageSignature:
+        return field.contents.data
+    }
+  }
+
+  /** @inheritdoc */
+  toJson() {
+    const result: TxData = {
+      id: this._stxTransaction.txid(),
+      fee: new BigNumber(this._stxTransaction.auth.getFee().toString()).toNumber(),
```
wesgraham, line 96: What's the point of going from string to bigNumber to number? Think we can just getFee here in Numerical value (can't imagine us getting overprecise on the fees)

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * Sets this transaction payload
+   *
+   * @param rawTransaction
+   * @param {Payload} payload transaction payload
+   */
+  fromRawTransaction(rawTransaction: string) {
+    const raw = removeHexPrefix(rawTransaction);
+    this._stxTransaction = deserializeTransaction(BufferReader.fromBuffer(Buffer.from(raw, 'hex')));
```
wesgraham, line 133: can we log a clean error if this deserialization fails?

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,171 @@
...
+   *
+   * @param {TransactionType} transactionType The transaction type to be set
+   */
+  setTransactionType(transactionType: TransactionType): void {
+    this._type = transactionType;
+  }
+
+  /**
+   * Load the input and output data on this transaction using the transaction json
+   * if there are outputs. For transactions without outputs (e.g. wallet initializations),
```
wesgraham, line 148: nit: We don't have wallet initializations in stacks - can remove from comments

modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -0,0 +1,267 @@
...
+    const txData = tx.toJson();
+    this.fee({ fee: txData.fee.toString() });
+    this.source({ address: txData.from });
+    if (txData.payload.memo) {
+      this.memo(txData.payload.memo);
+    }
+    // check if it is signed or unsigned tx
+    if (isSingleSig(tx.stxTransaction.auth.spendingCondition!)) {
+      if (tx.stxTransaction.auth.spendingCondition.signature.data === emptyMessageSignature().data) {
+        // unsigned transaction
```
wesgraham, line 62: can we log/throw an error here?

wesgraham, line 62: can we remove this dead space? Instead the `else` case should be the default case here

wesgraham: something like `if isSingleSig() && data !== emptyMessageSig() {
 ...
}`

modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -0,0 +1,267 @@
...
+  protected _numberSignatures: number;
+  protected _multiSignerKeyPairs: KeyPair[];
+  protected _signatures: SignatureData[];
+  protected _network: StacksNetwork;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+    this._multiSignerKeyPairs = [];
+    this._signatures = [];
+    this._numberSignatures = 1;
```
wesgraham, line 41: Why is this initialized to 1? Seems like 0 would be a better default candidate

wesgraham: would have to adjust `buildImplementation` too

modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -0,0 +1,267 @@
...
+  /** @inheritdoc */
+  validateKey(key: BaseKey): void {
+    const keyPair = new KeyPair({ prv: key.key });
+    if (!keyPair.getKeys().prv) {
+      throw new BuildTransactionError('Invalid key');
+    }
+  }
+
+  /** @inheritdoc */
+  validateRawTransaction(rawTransaction: string): void {
```
wesgraham, line 217: we should be validating before all deserialization calls (I think I pointed out a case of this not happening somewhere above)

modules/account-lib/src/coin/stx/transactionBuilderFactory.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  /** @inheritdoc */
+  from(raw: string): TransactionBuilder {
+    this.validateRawTransaction(raw);
+    try {
+      const tx = this.parseTransaction(raw);
+      switch (tx.stxTransaction.payload.payloadType) {
+        case PayloadType.TokenTransfer:
+          return this.getTransferBuilder(tx);
+        // TODO: add case of wallet
```
wesgraham, line 23: Ah one big thing here is that we haven't handled any ContractCall types. We need to be able to handle interactions with the stacks staking contract. This will be fairly important to add

modules/account-lib/src/coin/stx/walletInitializationBuilder.ts
```diff
@@ -0,0 +1,26 @@
+import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import { BaseKey } from '../baseCoin/iface';
+import { NotImplementedError } from '../baseCoin/errors';
+import { Transaction } from './transaction';
+import { TransactionBuilder } from './transactionBuilder';
+
+export class WalletInitializationBuilder extends TransactionBuilder {
```
wesgraham, line 7: There are no wallet initializations in Stacks - this should be removable

modules/core/test/v2/unit/coins/stx.ts
```diff
@@ -0,0 +1,58 @@
...
+      'ST11NJTTKGVT6D1HY4NJRVQWMQM7TVAR091EJ8P2Y',
+      'SP2T758K6T2YRKG9Q0TJ16B6FP5QQREWZSESRS0PY',
+      'SP2T758K6T2YRKG9Q0TJ16B6FP5QQREWZSESRS0PY',
+    ];
+
+    badAddresses.map(addr => { basecoin.isValidAddress(addr).should.equal(false); });
+    goodAddresses.map(addr => { basecoin.isValidAddress(addr).should.equal(true); });
+  }));
+
+  describe('Keypairs:', () => {
```
wesgraham, line 42: Is this left over from the keypair generation task?

wesgraham (changes requested):
Nice work so far! Realizing we will also need to support the `contract_call` transaction type in some of these switches - we can discuss leaving this for another PR, or if it's an easy fix we can include in this one. Otherwise, just a bunch of nits

modules/account-lib/src/coin/stx/keyPair.ts
fariedt: I've pushed a change for this, commit 387f2c3ed87789b8256ba802b15c56e5e9b947c7.

modules/core/test/v2/unit/coins/stx.ts
fariedt: I looked at a few other coins' test code, and copied several of the methods from there.

While the test here might not be necessary, it did expose a flaw -- `modules/core/test/v2/unit/keychains.ts` expects to check secp256k1 coins for *extended* key support, using `getExtendedKeys()` in account-lib's `baseCoin/secp256k1ExtendedKeyPair.ts`.  I modified `core/src/v2/coins/stx.ts` to return extended keys in `generateKeyPair()`.  That led to another problem -- `isValidPub` doesn't support extended keys.  The right place to make a change there is in account-lib.

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: stx transactions do not have source address to validate with the key provided.

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: Added a `getCompressed()` method which returns `this.keyPair.compressed`.

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: Used `getCompressed()` method from keypair

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: Updated.

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: Added validation

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: removed

modules/account-lib/src/coin/stx/transactionBuilder.ts
fdmirza: updated

modules/account-lib/src/coin/stx/transactionBuilder.ts
fdmirza: I used 1, as minimum 1 sign should be required for signed tx. Updated to 0. so now `_numberSignatures` gets updated value in `signImplementation` method

modules/account-lib/src/coin/stx/transactionBuilder.ts
fdmirza: done

modules/account-lib/src/coin/stx/walletInitializationBuilder.ts
fdmirza: Removed

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: The signer signs the transaction (updates spendingCondition in transaction authorization)

modules/account-lib/src/coin/stx/transaction.ts
wesgraham: Ah ok - can we remove the TODO in that case?

modules/account-lib/src/coin/stx/transaction.ts
wesgraham: got it, thanks

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+   * Sets this transaction payload
+   *
+   * @param rawTransaction
+   * @param {Payload} payload transaction payload
+   */
+  fromRawTransaction(rawTransaction: string) {
+    const raw = removeHexPrefix(rawTransaction);
+    try {
```
wesgraham, line 133: Can we use validateRawTransaction here first rather than duplicating logic?

modules/account-lib/src/coin/stx/transactionBuilderFactory.ts
```diff
@@ -0,0 +1,71 @@
...
+          throw new InvalidTransactionError('Invalid transaction');
+      }
+    } catch (e) {
+      console.error(e);
+      throw new ParseTransactionError('There was an error parsing the raw transaction');
+    }
+  }
+
+  private parseTransaction(rawTransaction: string): Transaction {
+    const tx = new Transaction(this._coinConfig);
```
wesgraham, line 33: can we validate the raw transaction here as well? Or is it assumed it will always be validated by this point?

modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -0,0 +1,241 @@
...
+    try {
+      deserializeTransaction(BufferReader.fromBuffer(Buffer.from(removeHexPrefix(rawTransaction), 'hex')));
+    } catch (e) {
+      throw new ParseTransactionError('There was an error parsing the raw transaction');
+    }
+  }
+
+  /** @inheritdoc */
+  validateTransaction(transaction?: Transaction): void {
+    this.validateFee();
```
wesgraham, line 214: are there no other checks in validateTransaction? Just validate fee?

modules/account-lib/src/coin/stx/transactionBuilderFactory.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    return builder;
+  }
+
+  /** @inheritdoc */
+  validateRawTransaction(rawTransaction: any): void {
+    if (!rawTransaction) {
+      throw new InvalidTransactionError('Raw transaction is empty');
+    }
+  }
```
wesgraham, line 70: do we not try to deserialize on this validateRawTransaction?

modules/account-lib/src/coin/stx/transactionBuilderFactory.ts
fdmirza: Yes, it will always be validated by this point

modules/account-lib/src/coin/stx/transactionBuilder.ts
fdmirza: yes, validateNonce should be there. Adding it. Could not think of any other validations at this point.

modules/account-lib/src/coin/stx/transactionBuilderFactory.ts
fdmirza: updated

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: validateRawTransaction is not available here. In transaction, we just have one place for deserialization.

modules/account-lib/package.json
```diff
@@ -33,7 +33,7 @@
   "license": "ISC",
   "dependencies": {
     "@bitgo/bls": "^0.1.0",
-    "@bitgo/statics": "^6.4.0",
+    "@bitgo/statics": "^6.4.1-rc.0",
```
wesgraham, line 36: Can we remove this line please?

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,179 @@
...
+  private _stxTransaction: StacksTransaction;
+  protected _type: TransactionType;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  /** @inheritdoc */
+  canSign(key: BaseKey): boolean {
+    // TODO: implement this validation method
```
wesgraham, line 32: nit: can we remove this todo?

modules/account-lib/src/coin/stx/transactionBuilderFactory.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  /** @inheritdoc */
+  from(raw: string): TransactionBuilder {
+    this.validateRawTransaction(raw);
+    try {
+      const tx = this.parseTransaction(raw);
+      switch (tx.stxTransaction.payload.payloadType) {
+        case PayloadType.TokenTransfer:
+          return this.getTransferBuilder(tx);
+        // TODO: add case of wallet
```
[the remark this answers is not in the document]
wesgraham: or modify it to "TODO: Add case of contract_call"

wesgraham (dismissed):

wesgraham (approved):

PR #1057, opened 2021-03-11
statics: add STX coin config

fariedt:
- add STX to `CoinFamily` and `UnderlyingAsset` enums
- add stx and tstx to `CoinMap`
- add stx to main and test `Networks` maps.


wesgraham (approved):

modules/core/test/v2/unit/keychains.ts
```diff
@@ -55,6 +55,7 @@ describe('V2 Keychains', function v2keychains() {
       && n.kind === CoinKind.CRYPTO
       && n.asset !== UnderlyingAsset.USD
       && n.asset !== UnderlyingAsset.CSPR
+      && n.asset !== UnderlyingAsset.STX    // Remove after merging STX
```
wesgraham, line 58: can we remove this now?

fariedt: I've removed this line, but to work around a [browser test error](https://cloud.drone.io/BitGo/BitGoJS/7042/4/4) I needed to add a similar temporary exclusion to `modules/core/test/browser/browser.spec.ts`.

wesgraham (approved):
Excellent work!

PR #1059, opened 2021-03-12
Add signed drone manifest

drGrove:


tylerlevine (approved):

tylerlevine:
Actually, since I have to make a few changes to the .drone.yml anyway which will require a re-sign, I'll just add this as part of that PR.

Thanks for your help @drGrove!

PR #1064, opened 2021-03-15
feat(account-lib): add Delegate and Undelegate Builders

JulioAchaval-SimTLiX:
Add delegate and undelegate builder


modules/account-lib/src/coin/cspr/delegateBuilder.ts
```diff
@@ -0,0 +1,135 @@
...
+      throw new InvalidParameterValueError('Invalid amount');
+    }
+    this._amount = amount;
+    return this;
+  }
+
+  //endregion
+
+  //region Validators
+  validateMandatoryFields(): void {
```
rliberal-simtlix, line 113: Add js doc

modules/account-lib/src/coin/cspr/ifaces.ts
```diff
@@ -38,6 +38,14 @@ export interface CasperTransferTransaction {
   extraArguments: Map<string, CLValue>;
 }
 
+export interface CasperDelegateTransaction {
```
rliberal-simtlix, line 41: Add js doc

rliberal-simtlix (changes requested):

modules/account-lib/src/coin/cspr/delegateBuilder.ts
```diff
@@ -0,0 +1,135 @@
...
+} from './constants';
+import { isValidAmount, isValidAddress, getTransferAmount, getDelegatorAddress, getValidatorAddress } from './utils';
+
+export class DelegateBuilder extends TransactionBuilder {
+  private _validator: string;
+  private _delegator: string;
+  private _action: string;
+  private _amount: string;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
```
rliberal-simtlix, line 23: add js doc

rliberal-simtlix (dismissed):

modules/account-lib/src/coin/cspr/delegateBuilder.ts
```diff
@@ -0,0 +1,146 @@
...
+      extraArguments.set(DELEGATE_FROM_ADDRESS, CLValue.string(this._delegator));
+      extraArguments.set(DELEGATE_VALIDATOR, CLValue.string(this._validator));
+    }
+    this._session = {
+      action: this._action,
+      delegator: PublicKey.fromHex(SECP256K1_PREFIX + this._delegator),
+      validator: PublicKey.fromHex(SECP256K1_PREFIX + this._validator),
+      amount: this._amount,
+      extraArguments: extraArguments,
+    };
```
smaria95, line 50: I think this should be done like in WalletInitializationBuilder.
Since delgate operation will be done using a contract, we should use ModuleBytes to send the contract and pass contract parameters as RuntimeArgs.

https://github.com/BitGo/BitGoJS/blob/master/modules/account-lib/src/coin/cspr/walletInitializationBuilder.ts#L24-L36

smaria95 (commented):
Just a comment about delegation contract

modules/account-lib/src/coin/cspr/transaction.ts
```diff
@@ -183,6 +228,36 @@ export class Transaction extends BaseTransaction {
...
+      const delegateValues = new Map();
+      delegateValues.set(
+        TRANSACTION_TYPE,
+        (this.casperTx.session.getArgByName(TRANSACTION_TYPE) as CLValue).asString(),
+      );
+      delegateValues.set('amount', getDelegateAmount(this.casperTx.session));
+      delegateValues.set(DELEGATE_FROM_ADDRESS, getDelegatorAddress(this.casperTx.session));
+      delegateValues.set(DELEGATE_VALIDATOR, getValidatorAddress(this.casperTx.session));
+
+      txJson['deploy']!['session']['ModuleBytes']['args'].forEach(arg => {
```
CapnMigraine, line 253: This looks a bit fragile. Is this different from using `txJson.deploy.session.ModuleBytes.args`?

modules/account-lib/src/coin/cspr/utils.ts
```diff
@@ -81,6 +101,23 @@ export function isValidTransferId(id: number): boolean {
...
+ * @returns {string} the transfer amount
+ */
+function getAmountArgValue(transferTx: ExecutableDeployItem): string {
+  const amount = transferTx.getArgByName('amount');
+
+  if (!amount || !amount.isBigNumber()) {
+    throw new InvalidTransactionError('Transfer does not have an amount defined');
+  }
+
+  return amount.asBigNumber().toString();
```
CapnMigraine, line 118: Can we convert using `toFixed(0)` instead?

modules/account-lib/test/unit/coin/cspr/transactionBuilder/delegateBuilder.ts
```diff
@@ -0,0 +1,182 @@
+import should from 'should';
```
CapnMigraine, line 1: nit:
```suggestion
import * as should from 'should';
```

modules/account-lib/test/unit/coin/cspr/transactionBuilder/undelegateBuilder.ts
```diff
@@ -0,0 +1,182 @@
+import should from 'should';
```
CapnMigraine, line 1:
```suggestion
import * as should from 'should';
```

modules/account-lib/src/coin/cspr/delegateBuilder.ts
```diff
@@ -0,0 +1,150 @@
...
+   * @returns {DelegateUndelegateContractArgs} contracts args to create a session
+   */
+  private buildDelegateParameters(): DelegateUndelegateContractArgs {
+    const delegator = PublicKey.fromHex(SECP256K1_PREFIX + this._source.address);
+    const validator = PublicKey.fromHex(this._validator);
+
+    return {
+      action: CLValue.string(this._action),
+      delegator: CLValue.publicKey(delegator),
+      validator: CLValue.publicKey(validator),
```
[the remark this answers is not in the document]
smaria95: This is valid on lastest versions of Casper SDK. 
The function signature has changed and now it takes a PublicKey instance as argument.

https://github.com/casper-ecosystem/casper-client-sdk/blob/master/src/lib/CLValue.ts#L1640

modules/account-lib/src/coin/cspr/transaction.ts
smaria95: It's the same. However, it uses strings.
I'll change this line to use your suggestion in exchange.

modules/account-lib/src/coin/cspr/utils.ts
smaria95: `asBigNumber` does not return an instance of `ethersproject` lib.
This implementation does not have a toFixed() function.

Should I use something like `new BigNumber(amount.asBigNumber().toString()).toFixed(0)`?

CapnMigraine: hmm... I see. Your suggestion wouldn't work if `toString()` causes the number to be represented using exponential notation. Can we confirm that's not the case with this library?

CapnMigraine (approved):
lgtm

modules/account-lib/src/coin/cspr/utils.ts
smaria95: I checked and `ethersproject.BigNumber.toString()` returns the same result as `BigNumber.toFixed(0)`
So there is no issue on leaving the current implementation.

PR #1068, opened 2021-03-17
feat(account-lib): add unit tests related to extended keys support

AgusBerteaSimtlix:
Add unit tests related to extended keys support

https://bitgoinc.atlassian.net/browse/STLX-2053 - stlx-2053


modules/account-lib/src/coin/cspr/transactionBuilder.ts
```diff
@@ -248,6 +248,16 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
       if (_sourceKeyPair.getKeys().prv === key.key.toUpperCase()) {
         throw new SigningError('Repeated sign: ' + key.key);
       }
+      // Try to get extended keys in order to validate them
+      let xprv;
+      try {
+        xprv = _sourceKeyPair.getExtendedKeys().xprv;
+      } catch (err) {
+        return;
```
rliberal-simtlix, line 256: Why we don't log the error?

AgusBerteaSimtlix: Here the return is to break the for each in the line 248, we don't want exception thrown. Whit this we want to avoid an unexpected behavior when the key pair doesn't have extended keys, if the extended keys exists, they will be validate correctly

rliberal-simtlix: :rocket: gotcha! thanks.

rliberal-simtlix (approved):

modules/account-lib/test/unit/coin/cspr/transaction.ts
```diff
@@ -170,6 +284,36 @@ describe('Cspr Transaction', () => {
...
+    it('wallet initialization using extended key', async function() {
+      const walletInitTx = await getWalletInitTransactionUsignExtendedKey();
+      const encodedTx = walletInitTx.toBroadcastFormat();
+      const walletInitJsonTx = JSON.parse(encodedTx);
+
+      const argName = 0;
+      const argValue = 1;
+      const owner0 = 0;
+      const owner1 = 1;
+      const owner2 = 2;
```
CapnMigraine, line 296: nit: You could use a regular for loop instead of creating 3 consts to use in `forEach`

CapnMigraine (commented):
lgtm

PR #1070, opened 2021-03-18
fix(hbar): send the message signatures in hex instead of buffer

argjv:
TICKET: BG-29353

### DIFF
--- a/modules/core/test/v2/unit/coins/hbar.ts
+++ b/modules/core/test/v2/unit/coins/hbar.ts
@@ -3,6 +3,7 @@ import * as accountLib from '@bitgo/account-lib';
 
 import { TestBitGo } from '../../../lib/test_bitgo';
 import { rawTransactionForExplain } from '../../fixtures/coins/hbar';
+import { randomBytes } from 'crypto';
 
 describe('Hedera Hashgraph:', function() {
   let bitgo;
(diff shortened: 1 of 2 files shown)


marktoda (dismissed):

tylerlevine (changes requested):
Test?

modules/core/test/v2/unit/coins/hbar.ts
```diff
@@ -98,6 +99,21 @@ describe('Hedera Hashgraph:', function() {
     });
   });
 
+  describe('Sign Message', () => {
+    it('should be performed', async () => {
+      const keyPair = new accountLib.Hbar.KeyPair();
+      const messageToSign = Buffer.from(randomBytes(32)).toString('hex');
+      const signature = await basecoin.signMessage(keyPair.getKeys(), messageToSign);
+      keyPair.verifySignature(messageToSign, Uint8Array.from(Buffer.from(signature, 'hex'))).should.equals(true);
```
tylerlevine, line 107: It's not incorrect, but the extra `Uint8Array.from()` wrapper is unnecessary because `Buffer` is a subtype of `Uint8Array`:

```
$ node
> Buffer.from([]) instanceof Uint8Array
true
```

tylerlevine (approved):
Nice work :+1:

PR #1072, opened 2021-03-19
Revert stx change and improve unit test

tylerlevine:
It appears that STX doesn't produce xpubs from it's `generateKeyPair` function, and instead produces derived keys directly.

This is a behavior break from other secp256k1 based coins, and causes a unit test to fail.

While we investigate the proper behavior for key creation for STX, and how this change was allowed to merge with a failing test at all, let's revert to unblock further pull requests.

Ticket: BG-30487


tylerlevine:
CC @fariedt

fariedt:
I discovered this problem [yesterday](https://github.com/BitGo/BitGoJS/pull/1055#pullrequestreview-615715546), but my fix likely won't be merged this week.

fariedt:
I've pushed several commits related to this: 

* commit a6cb9cfd37fdc6ee14d30f23d6671f28a408d624 for core's stx's `generateKeyPair()` to return extended keys
* commit 37c8ed5219255312a9735118be55af46480d8319 for account-lib's stx: generate stx keypairs using extended public/private keys
* commit 6f5b63449146f29efc9f1fd0b7ae3e97cc7ec0c7 to test account-lib's implementation

wesgraham (approved):

PR #1074, opened 2021-03-19
feat(statics): add WETH and WBTC on tron

silalvado:
Add WETH and WBTC on Tron network

[stlx-2263](https://bitgoinc.atlassian.net/browse/STLX-2263)

### DIFF
--- a/modules/statics/src/account.ts
+++ b/modules/statics/src/account.ts
@@ -96,10 +96,34 @@ export class ContractAddressDefinedToken extends AccountCoinToken {
 }
 
 /**
+ * ERC20 token addresses are Base58 formatted on some blockchains.
+ */
+export class Base58ContractAddressDefinedToken extends AccountCoinToken {
+  public contractAddress: ContractAddress;
+
+  constructor(options: Erc20ConstructorOptions) {
+    super({
+      ...options,
+    });
+
+    if (!/^[1-9A-HJ-NP-Za-km-z]{34}$/.test(options.contractAddress)) {
+      throw new InvalidContractAddressError(options.name, options.contractAddress);
+    }
+
+    this.contractAddress = (options.contractAddress as unknown) as ContractAddress;
+  }
+}
+
+/**
  * ERC 20 is a token standard for the Ethereum blockchain. They are similar to other account coins, but have a
  * contract address property which identifies the smart contract which defines the token.
  */
 export class Erc20Coin extends ContractAddressDefinedToken {}
+
+/**
+ * The TRON blockchain supports tokens of the ERC20 standard similar to ETH ERC20 tokens.
+ */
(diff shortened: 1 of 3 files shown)


modules/statics/src/account.ts
```diff
@@ -95,12 +95,36 @@ export class ContractAddressDefinedToken extends AccountCoinToken {
...
+ */
+export class Base58ContractAddressDefinedToken extends AccountCoinToken {
+  public contractAddress: ContractAddress;
+
+  constructor(options: Erc20ConstructorOptions) {
+    super({
+      ...options,
+    });
+
+    if (!options.contractAddress.match(/[1-9A-HJ-NP-Za-km-z]{34}$/)) {
```
smaria95, line 109: I think you're trying to check if the full string match with the regex argument.
If that's the case, I'd use `/^[1-9A-HJ-NP-Za-km-z]{7}$/` instead of `/[1-9A-HJ-NP-Za-km-z]{7}$/` (see ^ at the beginning means the start of the string)

Example:
`'1AJNPa'.match(/[1-9A-HJ-NP-Za-km-z]{7}$/) // Returns true`
`'prefix1AJNPasufix'.match(/[1-9A-HJ-NP-Za-km-z]{7}$/) // Returns true too`
while
`'1AJNPa'.match(/^[1-9A-HJ-NP-Za-km-z]{7}$/) // Returns true`
`'prefix1AJNPasufix'.match(/^[1-9A-HJ-NP-Za-km-z]{7}$/) // Returns false`

Also, I think `regex.test` is more performant than `string.match` (https://stackoverflow.com/questions/10940137/regex-test-v-s-string-match-to-know-if-a-string-matches-a-regular-expression)

silalvado: Thank you Santi! I've modified it.

modules/statics/src/account.ts
```diff
@@ -95,12 +95,36 @@ export class ContractAddressDefinedToken extends AccountCoinToken {
...
+      ...options,
+    });
+
+    if (!/^[1-9A-HJ-NP-Za-km-z]{34}$/.test(options.contractAddress)) {
+      throw new InvalidContractAddressError(options.name, options.contractAddress);
+    }
+
+    this.contractAddress = (options.contractAddress as unknown) as ContractAddress;
+  }
+}
```
FabianSimtlix, line 115: Which are the use cases for this? why are we creating a new class where the only difference is the regex used to validate the address?
Maybe this could be model different, depending on the usage for each one

twtaylorbitgo: yeah this doesn't seem great. i think this anti-pattern was setup for other things since there's a weird hierarchy to these. i'm not terribly against it since it's already kind of in there as an anti pattern

twtaylorbitgo (approved):

PR #1078, opened 2021-03-22
feat(account-lib): add verifySignature Method

JulioAchaval-SimTLiX:
add method to verify msg and tx signatures

Jira ticket: https://bitgoinc.atlassian.net/browse/STLX-2048


rliberal-simtlix (approved):
LGTM

tylerlevine (changes requested):
I don't see a ticket for this in the PR or commit message - could you please add one?

Also, CSPR is using secp-based signatures right? Those don't have a fixed length (assuming they are DER encoded), so I'm not sure the length assertion is correct.

CapnMigraine (approved):
nice work!

tylerlevine (approved):

PR #1080, opened 2021-03-24
stacks: add contract builder for contract calls

fdmirza:
This PR adds support for contract calls builder.
branched off of feat/stx-builders
cc @wesgraham


modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -0,0 +1,251 @@
+import BigNumber from 'bignumber.js';
+import BigNum from 'bn.js';
+import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
```
tylerlevine, line 3:
```suggestion
import { BaseCoin as CoinConfig } from '@bitgo/statics';
```

I think this is exported from the top level.

modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -0,0 +1,251 @@
+import BigNumber from 'bignumber.js';
+import BigNum from 'bn.js';
```
tylerlevine, line 2: Why do we need two bignumber libs?

modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -0,0 +1,251 @@
...
+  protected _multiSignerKeyPairs: KeyPair[];
+  protected _signatures: SignatureData[];
+  protected _network: StacksNetwork;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+    this._multiSignerKeyPairs = [];
+    this._signatures = [];
+    this._numberSignatures = 0;
+    this._network = new StacksTestnet();
```
tylerlevine, line 41: Is mainnet coming later? Or is it expected that the user calls the `network()` setter after construction?

modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  /**
+   *  Set the number of signatures for multi-sig
+   *
+   * @param {number} numSigns
+   * @returns {TransactionBuilder} This transaction builder
+   */
+  numberSignatures(numSigns: number): this {
+    this.validateValue(new BigNumber(numSigns));
```
tylerlevine, line 177: Seems odd that this works fine with `0`. Is this the right check here?

modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -0,0 +1,251 @@
...
+   * Validates that nonce is defined
+   */
+  private validateNonce(): void {
+    if (this._nonce === undefined) {
+      throw new BuildTransactionError('Invalid transaction: missing nonce');
+    }
+    try {
+      this.validateValue(new BigNumber(this._nonce));
+    } catch (e) {
+      throw new BuildTransactionError('Invalid nonce');
```
tylerlevine, line 247: It may be useful to include the nonce value here so we have some more info if we need to debug bad nonces.

modules/account-lib/src/coin/stx/transactionBuilderFactory.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /**
+   * Initialize the builder with the given transaction
+   *
+   * @param {Transaction | undefined} tx - the transaction used to initialize the builder
+   * @param {TransactionBuilder} builder - the builder to be initialized
+   * @returns {TransactionBuilder} the builder initialized
+   */
+  private static initializeBuilder<T extends TransactionBuilder>(tx: Transaction | undefined, builder: T): T {
```
tylerlevine, line 65: optional params typically come last in the parameter list - do we need it like this to fulfill some abstract class constraints?

modules/account-lib/src/coin/stx/transactionBuilderFactory.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (tx) {
+      builder.initBuilder(tx);
+    }
+    return builder;
+  }
+
+  /** @inheritdoc */
+  validateRawTransaction(rawTransaction: any): void {
```
tylerlevine, line 73:
```suggestion
  validateRawTransaction(rawTransaction: unknown): void {
```

modules/account-lib/src/coin/stx/transferBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  initBuilder(tx: Transaction): void {
+    const txData = tx.toJson();
+    if (txData.payload === undefined) {
+      throw new InvalidTransactionError('payload must not be undefined');
+    }
+    if (txData.payload.payloadType == PayloadType.TokenTransfer) {
```
tylerlevine, line 30:
```suggestion
    if (txData.payload.payloadType === PayloadType.TokenTransfer) {
```

modules/account-lib/src/coin/stx/transferBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+      nonce: new BigNum(this._nonce),
+    };
+    if (this._multiSignerKeyPairs.length === 0) {
+      throw new InvalidParameterValueError('supply at least 1 public key');
+    }
+    if (this._multiSignerKeyPairs.length === 1) {
+      return {
+        ...defaultOpts,
+        publicKey: this._multiSignerKeyPairs[0].getKeys(this._multiSignerKeyPairs[0].getCompressed()).pub,
+      };
```
tylerlevine, line 64: How come we need this special path for single keypairs?

modules/account-lib/test/unit/coin/stx/keyPair.ts
```diff
@@ -4,7 +4,7 @@ import * as testData from '../../../resources/stx/stx';
 import { Stx } from '../../../../src';
 import { TransactionVersion } from '@stacks/transactions';
 
-describe('Stx KeyPair', function() {
+describe('Stx KeyPair', function () {
```
tylerlevine, line 7: No need for the whitespace before the parens

```suggestion
describe('Stx KeyPair', function() {
```

modules/account-lib/test/unit/coin/stx/transaction.ts
```diff
@@ -0,0 +1,49 @@
...
+import { KeyPair } from '../../../../src/coin/stx/keyPair';
+
+describe('Stx Transaction', () => {
+  const coin = coins.get('stx');
+
+  function getTransaction(): Transaction {
+    return new Transaction(coin);
+  }
+
+  it('should throw empty transaction', () => {
```
tylerlevine, line 14: It would be better to test that a specific error type/message is thrown, instead of any error.

modules/account-lib/test/unit/coin/stx/transaction.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Transaction } from '../../../../src/coin/stx/transaction';
+import * as testData from '../../../resources/stx/stx';
+import { KeyPair } from '../../../../src/coin/stx/keyPair';
+
+describe('Stx Transaction', () => {
+  const coin = coins.get('stx');
+
+  function getTransaction(): Transaction {
+    return new Transaction(coin);
+  }
```
tylerlevine, line 12: Seems like this logic should just be inline - is there more that needs to happen in here?

modules/account-lib/test/unit/coin/stx/transactionBuilder/contractBuilder.ts
```diff
@@ -0,0 +1,86 @@
...
+      //   );
+      // });
+
+      // it('a transfer transaction with an invalid amount: negative value', () => {
+      //   const txBuilder = factory.getTransferBuilder();
+      //   should.throws(
+      //     () => txBuilder.amount('-5'),
+      //     e => e.message === 'Invalid amount',
+      //   );
+      // });
```
tylerlevine, line 83: Please remove if not needed
```suggestion
```

modules/core/package.json
```diff
@@ -44,8 +44,8 @@
     "gen-docs": "typedoc"
   },
   "dependencies": {
-    "@bitgo/account-lib": "^2.8.0-rc.2",
-    "@bitgo/statics": "^6.3.0-rc.1",
+    "@bitgo/account-lib": "^2.8.0-rc.3",
+    "@bitgo/statics": "^6.3.0-rc.2",
```
tylerlevine, line 48: Please do not modify the package.json versions for packages within this repository - these are managed by lerna as part of the release process.

tylerlevine (changes requested):
Nice work, this is off to a great start. Do we still need #1076? It seems like a lot of the logic was brought into this PR.

tylerlevine:
Also please rebase to fix merge conflicts (and take conflict resolutions from `master` for `yarn.lock` and `modules/core/package.json`)

modules/account-lib/src/coin/stx/transactionBuilder.ts
fdmirza: I am expecting explicity setting network to `StacksMainnet` or `StacksTestnet`.

modules/account-lib/src/coin/stx/transactionBuilder.ts
fdmirza: Validating numSigns to be positive value.

modules/account-lib/src/coin/stx/transactionBuilder.ts
fdmirza: Stacks lib uses `BigNum` from `bn.js` while signature of `abstract validateValue(value: BigNumber): void;` in BaseTransactionBuilder has `BigNumber`

modules/account-lib/src/coin/stx/contractBuilder.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Set the contract name
+   *
+   * @param {string} name name of contract
+   * @returns {TransferBuilder} the builder with the new parameter set
+   */
+  contractName(name: string): this {
+    if (name.length == 0) {
```
wesgraham, line 106: Maybe also a question for Tyler: can we restrict the contract name to just https://explorer.stacks.co/txid/SP000000000000000000002Q6VF78.pox (pox) for mainnet and ST000000000000000000002AMW42H.pox (pox) for testnet here or should we do it in platform? 

Purpose is that we will only support interactions with the pox contract for blockstack for this project

modules/account-lib/src/coin/stx/contractBuilder.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+  * Set the function name in contract
+  *
+  * @param {string} name name of function
+  * @returns {TransferBuilder} the builder with the new parameter set
+  */
+  functionName(name: string): this {
+    if (name.length == 0) {
```
wesgraham, line 120: Similarly @tylerlevine should we restrict the function names to `stack-stx`, `delegate-stx`, `delegate-stack-stx`, `stack-aggregation-commit`, and `revoke-delegate-stx` here as per https://bitgoinc.atlassian.net/browse/STLX-1898?

modules/account-lib/src/coin/stx/keyPair.ts
```diff
@@ -34,10 +37,11 @@ export class KeyPair extends Secp256k1ExtendedKeyPair {
   }
 
   /**
-   * Build an ECPair from a private key.
+   * Build a keypair from a protocol private key or extended private key.
```
wesgraham, line 40: Can we pull the extended keypair logic out of this PR? The extended keypair logic was merged separately

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -0,0 +1,219 @@
+import BigNumber from 'bignumber.js';
```
wesgraham, line 1: This also seems to overlap with https://github.com/BitGo/BitGoJS/pull/1055/files - can you pull this out of this PR as well? You can rebase on the other PR if that makes life easier

wesgraham: only want to be focused on the contract builder stuff here

modules/account-lib/src/coin/stx/transferBuilder.ts
```diff
@@ -0,0 +1,103 @@
+import { BaseCoin as CoinConfig } from '@bitgo/statics';
```
wesgraham, line 1: All of this needs to be separated back into the transactins & transfer builder PR's. Need to clean this up

modules/account-lib/test/resources/stacks/stacks.ts
```diff
@@ -0,0 +1,85 @@
+import { Stacks } from '../../../src/';
```
wesgraham, line 1: Separate out

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: i created this branch off the feat/stx-builders and this PR is to merge in master, so as long as feat/stx-builders in not merged in master, these changes would appear in this branch.

modules/account-lib/src/coin/stx/contractBuilder.ts
tylerlevine: Sure, I don't see a problem with that if we really ever only want to interact with these contracts - better to fail as early as possible. It does keep the library from working for other use cases, but we can always remove/adjust this check if needed later.

modules/account-lib/src/coin/stx/contractBuilder.ts
tylerlevine: Same as above - if that's all we need to support then restricting to those values is fine with me.

modules/account-lib/src/coin/stx/transaction.ts
tylerlevine: What's the status on that PR? Is there a lot of work remaining to get it merged. I agree with @wesgraham, it would be good to keep the contract builder and keychain changes separated.

modules/account-lib/test/resources/stacks/stacks.ts
```diff
@@ -0,0 +1,85 @@
...
+
+// multi sig
+export const prv1 = '21d43d2ae0da1d9d04cfcaac7d397a33733881081f0b2cd038062cf0ccbb752601'
+export const prv2 = 'c71700b07d520a8c9731e4d0f095aa6efb91e16e25fb27ce2b72e7b698f8127a01'
+export const prv3 = 'e75dcb66f84287eaf347955e94fa04337298dbd95aa0dbb985771104ef1913db01'
+export const MULTI_SIG_SINGED_TRANSACTION = '0x808000000004012fe507c09dbb23c3b7e5d166c81fc4b87692510b000000000000000000000000000000b400000003020123d6d3d22d2ca1b7310f5c45314d1e783c03193e1bb5e08685fc1a48f8bdac7c0d93a463efee18b9b41fffe38d4b4c7475809a37c0fd53f52cef456a8d9280db020010b311f82b1b111eb5efa16c4790687b16dc92b7b4d07a1f0f1fe65baee43eb02ede226f807a097237e7d0ee3aa4fdf9b6287dd23d819ec560df180875f4019a00038e3c4529395611be9abf6fa3b6987e81d402385e3d605a073f42f407565a4a3d000203020000000000051a1ae3f911d8f1d46d7416bfbe4b593fd41eac19cb00000000000003e874657374206d656d6f00000000000000000000000000000000000000000000000000'
```
tylerlevine, line 85:
```suggestion
export const MULTI_SIG_SIGNED_TRANSACTION = '0x808000000004012fe507c09dbb23c3b7e5d166c81fc4b87692510b000000000000000000000000000000b400000003020123d6d3d22d2ca1b7310f5c45314d1e783c03193e1bb5e08685fc1a48f8bdac7c0d93a463efee18b9b41fffe38d4b4c7475809a37c0fd53f52cef456a8d9280db020010b311f82b1b111eb5efa16c4790687b16dc92b7b4d07a1f0f1fe65baee43eb02ede226f807a097237e7d0ee3aa4fdf9b6287dd23d819ec560df180875f4019a00038e3c4529395611be9abf6fa3b6987e81d402385e3d605a073f42f407565a4a3d000203020000000000051a1ae3f911d8f1d46d7416bfbe4b593fd41eac19cb00000000000003e874657374206d656d6f00000000000000000000000000000000000000000000000000'
```

tylerlevine (changes requested):
Looking pretty good, just a couple things remaining. I think a few of my comments from the previous review were not addressed btw, please make sure you click the "Show Hidden" button to show all the review comments (these shouldn't be hidden any more since I resolved the fixed ones)

modules/account-lib/src/coin/stx/transferBuilder.ts
fdmirza: There is difference in keys for single and multi sig tx. (`publicKey` vs `publicKeys`)

modules/account-lib/src/coin/stx/contractBuilder.ts
```diff
@@ -0,0 +1,139 @@
...
+  private _functionName: string;
+  private _functionArgs: ClarityValue[];
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  initBuilder(tx: Transaction): void {
+    const txData = tx.toJson();
+    if (txData.payload === undefined) {
```
wesgraham, line 34: Is there any other verification we can do on the presence/correctness of contractName, functionName, functionArgs?

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -157,22 +189,24 @@ export class Transaction extends BaseTransaction {
...
-      ];
-
-      this._inputs = [
-        {
-          address: txJson.from,
-          value: txJson.payload.amount,
-          coin: this._coinConfig.name,
-        },
-      ];
+    if (txJson.payload.payloadType === PayloadType.TokenTransfer) {
```
wesgraham, line 192: Are there no inputs/outputs for contract calls?

modules/account-lib/src/coin/stx/utils.ts
```diff
@@ -172,3 +185,56 @@ export function isValidPrivateKey(prv: string): boolean {
...
+
+/**
+ * Checks for valid contract address
+ *
+ * @param {string} addr - contract deployer address
+ * @param {StacksNetwork} network - network object
+ * @returns {boolean} - the validation result
+ */
+export function isValidContractAddress(addr: string, network: StacksNetwork): boolean {
+  if (network.isMainnet()) {
```
wesgraham, line 216: Thinking this through we might want to `else if network.isTestnet())` for the else case. Otherwise we won't be able to have the contract call testing in develop.

If network is develop (local), I think we can just say all contract addresses are valid. Just because I dont think we'll know the staking contract address until we deploy it. @marktoda do you see any issues with this?

modules/account-lib/test/resources/stacks/stacks.ts
```diff
@@ -1,85 +0,0 @@
-import { Stacks } from '../../../src/';
```
wesgraham, line 1: Why are we deleting all of this?

marktoda:
I wonder if this logic wouldn't be better suited in https://github.com/bitgo/smart-contracts

modules/account-lib/src/coin/stx/utils.ts
marktoda: That sounds reasonable

modules/account-lib/test/resources/stacks/stacks.ts
fariedt: This deletes a file left over from when the coin was called "stacks" in the repo.  The real fixtures are in `test/resources/stx/stx.ts`.

modules/account-lib/src/coin/stx/transaction.ts
fdmirza: Not that I can think of. Please suggest if you think there should be inputs/outputs/ for contract call.

modules/account-lib/test/resources/stacks/stacks.ts
fdmirza: In the beginning we called it as `stacks` and now its moved to `stx.ts`

wesgraham (approved):

tylerlevine (approved):
Nice work

Issue #1079, opened 2021-03-23
Disable SSL does not handled correct


nikita-gorodeckij:
## Expected Behavior
I play with BitGo api  and found that flag `BITGO_DISABLE_SSL` is not working

I run:

```
docker run \
        -e BITGO_ENV="prod" \
        -e BITGO_DISABLE_SSL="true" \
        -it -p 4445:4445 bitgosdk/express:latest  --port 4445
```

Also I tried

```
docker run \
        -e BITGO_ENV="prod" \
        -e BITGO_DISABLE_SSL="1" \
        -it -p 4445:4445 bitgosdk/express:latest  --port 4445
```

But I always catch error: `Error: Must enable TLS when running against prod and listening on external interfaces!`

However, this code works fine:

```

docker run \
        -e BITGO_ENV="prod" \
        -it -p 4445:4445 bitgosdk/express:latest  --port 4445 --disablessl
```

Is it my mistake or it is docker image's bug?

tylerlevine:
Thanks for opening the issue, I'll look into this.

rakveli:
Hello my exchange wallets are showing un able to connect to blockchain in the production mode but everything is working fine on the test mode. please kindly look into this ?

tylerlevine:
This should be fixed by #1094, could you confirm that this issue has been resolved for you with the latest bitgo express image @nikita-gorodeckij?

tylerlevine:
> Hello my exchange wallets are showing un able to connect to blockchain in the production mode but everything is working fine on the test mode. please kindly look into this ?

@rakveli, could you please open a separate issue for this and provide some more details?

PR #1094, opened 2021-03-30
fix(express): don't store false when boolean flags are not given

tylerlevine:
The config precedence system expects that command line options which
aren't present to not exist (ie, set to `null` or `undefined`), but when
we use the `storeTrue` action for argparse arguments, omitted options
get set to `false`, which causes them to override any values passed via
environment variables.

Instead, use `storeConst` and store `true` explicitly if the option is
given. This will cause argparse to set the argument value to `null` if
it is omitted.

Ticket: BG-30729

### DIFF
--- a/modules/express/src/args.ts
+++ b/modules/express/src/args.ts
@@ -59,18 +59,20 @@ parser.addArgument(['-l', '--logfile'], {
 });
 
 parser.addArgument(['--disablessl'], {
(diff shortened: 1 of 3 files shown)


tylerlevine:
This should fix #1079

wesgraham (approved):

PR #1097, opened 2021-04-01
fix(core): set minimal required node version to 10.22.0

OttoAllmendinger:
The project dependency `ripple-binary-codec` has node `10.22.0` as the
lower bound:

https://github.com/ripple/ripple-binary-codec/blob/0.2.4/package.json

Therefore BitGoJS has the same minimal lower bound.

This is not a breaking change since Node 8 was not really supported
anyway.

Issue: BG-30763


README.md
```diff
@@ -25,34 +25,19 @@ The release notes for the `core` module are [here](https://github.com/BitGo/BitG
...
-
-* `10`
-
-**Note:** We intend to support the current `lts` (node 12), but it is not currently running due to compability issues with dependencies)
-
-JavaScript package managers should allow you to install this package with any
+Due to constraints from library dependencies, we currenty only support node
+versions `>=10.22.0 <12.0.0`.
+
+**Note:** We intend to support the current `lts` (node 12), in the future.
```
OttoAllmendinger, line 31: I tried `yarn install` node 12 LTS locally: it only failed for `ed25519` which is an optional dependency, so the final state was `success`

tylerlevine (approved):

PR #1100, opened 2021-04-03
feat(hbar): update hbar sdk

argjv:
BG-30823


modules/account-lib/src/coin/hbar/keyPair.ts
```diff
@@ -53,7 +55,10 @@ export class KeyPair extends Ed25519KeyPair {
   /** @inheritdoc */
   recordKeysFromPrivateKeyInProtocolFormat(prv: string): DefaultKeys {
     try {
-      const hederaPrv = Ed25519PrivateKey.fromString(prv);
+      if (prv.length % 2 !== 0) {
```
argjv, line 58: Without this check, this step does not fail now. We have a unit test that was making sure that by adding an extra number at the end of the key, we failed.

tylerlevine, line 58:
```suggestion
      if (/^([a-f0-9]{2})+$/i.test(prv)) {
```

This might be a bit more what you're after if you want to make sure it's hex (remove the `i` at the end of the regex if you want this to be case sensitive)

tylerlevine: Also, pull this check above the try/catch so you don't catch this locally thrown exception.

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -237,7 +241,7 @@ export class Transaction extends BaseTransaction {
   /**
    * Returns a hash of the given proto object.
    *
-   * @param obj - The object to be hashed, it must be an proto namespace object
```
tylerlevine, line 240: What's this `proto` type? Can we use it to constrain the generic type `T`?

modules/account-lib/package.json
```diff
@@ -53,6 +53,7 @@
     "ethereumjs-util": "5.2.0",
     "libsodium-wrappers": "^0.7.6",
     "lodash": "^4.17.15",
+    "long": "^4.0.0",
```
tylerlevine, line 56: Was this just totally undeclared here and brought in via a transitive dependency? :fearful:

tylerlevine (changes requested):
Needs a rebase, but otherwise lgtm aside from some small things. Nice work :+1:

modules/account-lib/package.json
barathbitgo: looks like brought by changes made in `transaction.ts`

argjv: I think so. I saw we were using it transactions.ts file that Barath mentioned, but when I tried to run the test locally they were failing due to that missing dependency.

barathbitgo: fixed

modules/account-lib/src/coin/hbar/transaction.ts
barathbitgo: I think it may require quite a bit of code changes unfortunately :(

modules/account-lib/src/coin/hbar/keyPair.ts
```diff
@@ -52,8 +54,11 @@ export class KeyPair extends Ed25519KeyPair {
 
   /** @inheritdoc */
   recordKeysFromPrivateKeyInProtocolFormat(prv: string): DefaultKeys {
+    if (prv.length % 2 !== 0 && /^([a-f0-9]{2})+$/i.test(prv)) {
+      throw new InvalidKey('Invalid private key length. Must be a hex and multiple of 2');
+    }
```
tylerlevine, line 59: There's some lack of understanding here - the length check is now redundant, since the regex only matches on pairs of hex digits. We're also missing a negation on the regex test so throwing the `InvalidKey` error isn't logically possible.
```suggestion
    if (!/^([a-f0-9]{2})+$/i.test(prv)) {
      throw new InvalidKey('Invalid private key. Must be a hex string with even length');
    }

```

Please write a unit test for these cases, since this is apparently uncovered.

tylerlevine (changes requested):

tylerlevine (approved):

PR #1107, opened 2021-04-06
feat(statics): add requires reserve

JulioAchaval-SimTLiX:
add new param "requiesReserve" to all coins (only true for Casper)

[STLX-2337](https://bitgoinc.atlassian.net/browse/STLX-2337)

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -124,6 +124,10 @@ export enum CoinFeature {
    * Are fees for transactions of this coin paid for by the Enterprise (eg, Enterprise gas tank)?
    */
   ENTERPRISE_PAYS_FEES = 'enterprise-pays-fees',
+  /*
+   * This coin requires that accounts keep a minimum balance as reserve
+   */
+  REQUIRES_RESERVE = 'requires-reserve',
 }
 
 /**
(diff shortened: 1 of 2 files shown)


tylerlevine (changes requested):
This is probably more appropriate as a `CoinFeature`: https://github.com/BitGo/BitGoJS/blob/ba2d8707fff934fe5124163bd78e52bf9a1730da/modules/statics/src/base.ts#L53

tylerlevine (approved):

modules/statics/src/base.ts
```diff
@@ -124,6 +124,10 @@ export enum CoinFeature {
    * Are fees for transactions of this coin paid for by the Enterprise (eg, Enterprise gas tank)?
    */
   ENTERPRISE_PAYS_FEES = 'enterprise-pays-fees',
+  /*
+   * This coin requires a reserve amount to fund a wallet
```
CapnMigraine, line 128: nit: The reserve is not just for funding
```suggestion
   * This coin requires that accounts keep a minimum balance as reserve
```

CapnMigraine (dismissed):
LGTM with a nit comment

tylerlevine (approved):

CapnMigraine (approved):

PR #1113, opened 2021-04-10
fix(core): fix txPrebuild param in CSPR signTransaction method

CapnMigraine:
Use params.txHex instead of params.txJson in CSPR signTransaction method

Ticket: STLX-2555


CapnMigraine:
Merged in #1115

PR #1117, opened 2021-04-12
chore: use new blake2b packages

tylerlevine:
Ticket: BG-30979


ericcrosson-bitgo (approved):
Just the fix I need, LGTM

PR #1120, opened 2021-04-13
fix(ethlike): add chainid to statics

marktoda:
This commit adds the Ethereum chainid to bitgo statics so we can include
it in our built transactions. Account-lib already does this, but
Ethereum does not use account-lib in platform currently. Adding the
field to statics will give us a single source of truth for these values
and allow us to do it in platform. This commit also updates
account-lib's store of chainids to use statics instead of its own
hard-coded values

Ticket: BG-30890


tylerlevine (approved):
Cool, good idea

PR #1121, opened 2021-04-14
feat(account-lib): update casper sdk version

smaria95:
Update casper sdk version from v1.0.29 to v1.0.39

[STLX-2653](https://bitgoinc.atlassian.net/browse/STLX-2653)

### DIFF
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -46,7 +46,7 @@
     "@types/lodash": "^4.14.151",
     "bignumber.js": "^9.0.0",
     "bs58check": "^2.1.2",
-    "casper-client-sdk": "1.0.29",
+    "casper-client-sdk": "1.0.39",
     "elliptic": "^6.5.2",
     "ethereumjs-abi": "^0.6.5",
     "ethereumjs-tx": "2.1.2",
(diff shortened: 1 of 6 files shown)


modules/account-lib/package.json
```diff
@@ -46,7 +46,7 @@
     "@types/lodash": "^4.14.151",
     "bignumber.js": "^9.0.0",
     "bs58check": "^2.1.2",
-    "casper-client-sdk": "1.0.29",
+    "casper-client-sdk": "1.0.39",
```
tylerlevine, line 49: Please use `^1.0.39`

tylerlevine: Also please commit updated `yarn.lock`

smaria95: Hi @tylerlevine , we decided stop using `^` some time ago, at least for `casper-client-sdk`. Casperabs release new versions twice a week and we had a lot of issues because of bugs on these updates. Also, each release require some sort of integration testing to make sure everything was still working as expected; which made as lose a lot of time too.
@CapnMigraine , please correct me if I'm missing something.

Regarding the `yarn.lock`, I'll update this PR with the updated one.

CapnMigraine (approved):
lgtm

tylerlevine (approved):

PR #1123, opened 2021-04-15
stx: explainTransaction and signTransaction

fariedt:
Two methods for the `modules/core` implementation.


modules/core/src/v2/coins/stx.ts
```diff
@@ -15,6 +24,35 @@ interface SupplementGenerateWalletOptions {
...
+  contractAddress?: string;
+  contractName?: string;
+  contractFunction?: string;
+  contractFunctionArgs?: { type: string; value: string }[];
+}
+
+export interface ExplainTransactionOptions {
+  txHex?: string; // txHex is poorly named here; it is just a wrapped JSON object
+  halfSigned?: {
+    txHex: string; // txHex is poorly named here; it is just a wrapped JSON object
```
wesgraham, line 42: rename: txJson?

modules/core/src/v2/coins/stx.ts
```diff
@@ -110,10 +148,121 @@ export class Stx extends BaseCoin {
...
+   * @param params
+   * @param callback
+   */
+  signTransaction(
+    params: StxSignTransactionOptions,
+    callback?: NodeCallback<SignedTransaction>
+  ): Bluebird<SignedTransaction> {
+    const self = this;
+
+    return co<SignedTransaction>(function*() {
```
wesgraham, line 162: can we not use async/await here?

modules/core/src/v2/coins/stx.ts
```diff
@@ -110,10 +148,121 @@ export class Stx extends BaseCoin {
...
+   * Explain a Stacks transaction from txHex
+   * @param params
+   * @param callback
+   */
+  explainTransaction(
+    params: ExplainTransactionOptions,
+    callback?: NodeCallback<StxTransactionExplanation>
+  ): Bluebird<StxTransactionExplanation> {
+    const self = this;
+    return co<TransactionExplanation>(function*() {
```
wesgraham, line 196: same async/await

modules/core/src/v2/coins/stx.ts
```diff
@@ -110,10 +148,121 @@ export class Stx extends BaseCoin {
...
+      if (tx.type === accountLib.BaseCoin.TransactionType.Send) {
+        const outputs: TransactionRecipient[] = [
+          {
+            address: txJson.payload.to,
+            amount: txJson.payload.amount,
+            memo: txJson.payload.memo,
+          },
+        ];
+
+        const displayOrder = ['id', 'outputAmount', 'changeAmount', 'outputs', 'changeOutputs', 'fee', 'type'];
```
wesgraham, line 221: can you add memo

modules/core/src/v2/coins/stx.ts
```diff
@@ -110,10 +148,121 @@ export class Stx extends BaseCoin {
...
+
+        return explanationResult;
+      }
+
+      if (tx.type === accountLib.BaseCoin.TransactionType.ContractCall) {
+        const displayOrder = [
+          'id',
+          'fee',
+          'type',
+          'contractAddress',
```
wesgraham, line 242: can you add memo as well

modules/core/src/v2/coins/stx.ts
tylerlevine: No async await in implementation code yet please - we need to return a consistent promise type.

modules/core/src/v2/coins/stx.ts
fariedt: The comments are from the code copied from the `trx.ts` coin.  It's not accurate -- the `txHex` field really is a hex string, not JSON.  I'll remove the comments.

wesgraham (approved):

PR #1124, opened 2021-04-15
stacks: fix multisig signing issue

fdmirza:
For fixing signing issues of multisig transaction.
cc @wesgraham


fariedt:
The fix for the failing test is to use version `2.9.0-rc.4` of `@bitgo/account-lib` instead of `2.8.2`.  Skipping the memo check in the tests will also work (for now).

tylerlevine:
I don't quite follow here - why are we commenting this out pending a new version being released? Because this repo is managed with `lerna`, the version of `account-lib` should be exactly the one that exists in this repo (`node_modules/@bitgo/account-lib` should be a symlink to `modules/account-lib`) and not the npm version when installing using `yarn`.

modules/core/test/v2/unit/coins/stx.ts
```diff
@@ -53,7 +53,8 @@ describe('STX:', function() {
     explain.outputAmount.should.equal(testData.txExplainedTransfer.outputAmount);
     explain.outputs[0].amount.should.equal(testData.txExplainedTransfer.outputAmount);
     explain.outputs[0].address.should.equal(testData.txExplainedTransfer.recipient);
-    explain.outputs[0].memo.should.equal(testData.txExplainedTransfer.memo);
+    // TODO: uncomment the following line after account-lib 2.9.0 is released
```
tylerlevine, line 56: I don't quite follow here - why are we commenting this out pending a new version being released? Because this repo is managed with `lerna`, the version of `account-lib` should be exactly the one that exists in this repo (`node_modules/@bitgo/account-lib` should be a symlink to `modules/account-lib`) and not the npm version when installing using `yarn`.

tylerlevine (changes requested):

modules/core/test/v2/unit/coins/stx.ts
tylerlevine: What is the fix in account-lib 2.9.0 that should make this test pass?

wesgraham (dismissed):

modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -56,18 +60,41 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
-      const sigHashPreSign = makeSigHashPreSign(
-        tx.stxTransaction.verifyBegin(),
-        tx.stxTransaction.auth.authType!,
-        new BigNum(this._fee.fee),
-        new BigNum(this._nonce),
-      );
-      this._signatures.push(tx.stxTransaction.auth.spendingCondition.signature);
-      this._fromPubKeys = [publicKeyFromSignature(sigHashPreSign, this._signatures[0])];
+    if (tx.stxTransaction.auth.spendingCondition === undefined)
+      throw new InvalidTransactionError('spending condition cannot be undefined');
```
tylerlevine, line 64: Braces please

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -57,16 +60,19 @@ export class Transaction extends BaseTransaction {
     signer.appendOrigin(pubKey);
   }
 
-  async signWithSignatures(signature: SignatureData): Promise<void> {
+  async signWithSignatures(signature: SignatureData[], publicKey: string[]): Promise<void> {
     if (!signature) {
       throw new SigningError('Missing signatures');
     }
-    this._stxTransaction = this._stxTransaction.createTxWithSignature(signature.data);
+    if (signature.length === 1) this._stxTransaction = this._stxTransaction.createTxWithSignature(signature[0].data);
```
tylerlevine, line 67: braces please

tylerlevine (changes requested):
some style nits but other than that it's good

tylerlevine (approved):
Thanks!

PR #1125, opened 2021-04-19
core: implement verifyAddress for stx

fariedt:
This checks if the given address is valid, and if it is one of the supported address versions (SingleSig or MultiSig on Mainnet or Testnet).

### DIFF
--- a/modules/core/src/v2/coins/stx.ts
+++ b/modules/core/src/v2/coins/stx.ts
@@ -3,6 +3,7 @@
  */
 import * as Bluebird from 'bluebird';
 import * as accountLib from '@bitgo/account-lib';
+import { c32addressDecode } from 'c32check';
 import {
   BaseCoin,
   KeyPair,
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/stx.ts
```diff
@@ -87,9 +89,20 @@ export class Stx extends BaseCoin {
...
   verifyAddress(params: VerifyAddressOptions): boolean {
-    // TODO: Implement when available on the SDK.
-    throw true;
+    const { address } = params;
+    if (!this.isValidAddress(address)) throw new InvalidAddressError(`invalid address: ${address}`);
+
+    try {
+      const [version] = c32addressDecode(address);
+      const versionString = accountLib.Stx.AddressVersion[version];
+      if (versionString === undefined) throw new InvalidAddressError(`invalid address version: ${address}`);
```
wesgraham, line 100: Can we create an enum of valid address versions and ask if versionString is in the enum?

wesgraham (approved):

Issue #1126, opened 2021-04-20
bitcoinfees.earn.com is publishing inaccurate data, please use another bitcoin fees API


bretton:
Hi, I'm not familiar with your project. I did a search for `https://bitcoinfees.earn.com/api/v1/fees/recommended` and am alerting projects using this API to inaccurate & old data.

An up-to-date API with accurate data would be `https://mempool.space/api/v1/fees/recommended` or some alternative.

https://github.com/BitGo/BitGoJS/blob/6b2f193042064330c1f988c959ed195b9bdc29ad/modules/core/src/v2/coins/btc.ts

## Environment Details
n/a

## Expected Behavior
Get accurate fees for current time

## Current Behavior
Inaccurate fees from `https://bitcoinfees.earn.com/api/v1/fees/recommended`

## Possible Solution
Use an up-to-date API such as `https://mempool.space/api/v1/fees/recommended` or some alternative.

## Steps to Reproduce
n/a

tylerlevine:
Thanks for the heads up! We'll switch to the mempool.space fees endpoint.

PR #1127, opened 2021-04-20
fix(core): use mempool.space instead of earn.com for recovery fee

tylerlevine:
The earn.com endpoint is provided inaccurate data and doesn't appear to
be maintained any longer. The mempool.space endpoint is API-compatible
with our use case and provides a more accurate fee estimation.

Closes: #1126
Ticket: BG-31192


ericcrosson-bitgo (approved):
LGTM

PR #1128, opened 2021-04-21
fix: update version support policy

tylerlevine:
New node versions were recently added to the supported versions list,
but the README was not updated with this information.

Ticket: BG-30998


ericcrosson-bitgo (dismissed):
LGTM

ericcrosson-bitgo (approved):

PR #1131, opened 2021-04-22
Backmerge rel/latest for 11.13.0 release

tylerlevine:


modules/express/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bitgo/express",
-  "version": "9.9.2",
+  "version": "9.9.4",
```
tylerlevine, line 3: I'm not entirely sure why lerna decided to bump the express version when I published the stable versions of @bitgo/bls-dkg and @bitgo/blake2b, but it did, so even though it looks like we skipped a version here we actually didn't and there are two express versions being merged in here.

ericcrosson-bitgo (approved):
:ship:

PR #1132, opened 2021-04-22
chore: add step to drone to check managed package version numbers 

tylerlevine:
Check versions of in-repo packages to make sure they match with the package.json version.
This is to prevent inadvertent updates of dependency versions for packages which are symlinked
by lerna. The dependency versions of these packages should only ever be updated in the release branch.

Ticket: BG-31227

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -221,6 +221,7 @@ local CheckPreconditions(version) = {
     BuildInfo(version),
     Install(version),
     Command("check-commits", version),
+    Command("check-versions", version),
     Command("audit", version),
     Command("lint", version),
     Command("check-fmt", version),
--- a/.drone.yml
+++ b/.drone.yml
@@ -26,6 +26,11 @@ steps:
   commands:
   - yarn run check-commits
 
+- name: check-versions
+  image: node:10
+  commands:
+  - yarn run check-versions
+
 - name: audit
   image: node:10
   commands:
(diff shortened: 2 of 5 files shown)


ericcrosson-bitgo (dismissed):
automation station :bullettrain_side:

tylerlevine:
Going to replace this with a node script so I don't have to install jq in the build container.

ericcrosson-bitgo:
> Going to replace this with a node script so I don't have to install jq in the build container.

Are you interested in using [@typescript-tools/pin-lerna-package-versions](https://github.com/typescript-tools/typescript-tools/tree/master/packages/pin-lerna-package-versions) or contributing something like a `--check` flag? 

The function `pin-lerna-package-versions` uses to find dependencies, `@typescript-tools/lerna-packages`, is [open-source](https://www.npmjs.com/package/@typescript-tools/lerna-packages). Documentation is shoddy but [types are strong](https://github.com/typescript-tools/typescript-tools/blob/7939e6eb8ed5b05f49529f9e16cd5242f80c5d58/packages/lerna-packages/src/index.ts#L41) and [here](https://github.com/typescript-tools/typescript-tools/blob/7939e6eb8ed5b05f49529f9e16cd5242f80c5d58/packages/pin-lerna-package-versions/src/index.ts#L184) is an example of use.

tylerlevine:
Seems to be doing the right thing: https://cloud.drone.io/BitGo/BitGoJS/7524/1/5

tylerlevine:
Need to wait for the next stable release to go out, then I think we can merge this and start enforcing.

twtaylorbitgo (approved):

PR #1138, opened 2021-04-28
stx: explain unsigned transactions

fariedt:
This sends the public key(s) along with an unsigned transaction to `explainTransaction`.


modules/core/src/v2/coins/stx.ts
```diff
@@ -212,11 +213,11 @@ export class Stx extends BaseCoin {
         throw new Error('missing explain tx parameters');
       }
 
-      const factory = accountLib.getBuilder(self.getChain());
+      const factory = accountLib.getBuilder(self.getChain()) as accountLib.Stx.TransactionBuilderFactory;
       const txBuilder = factory.from(txHex);
 
-      if (!(txBuilder instanceof accountLib.BaseCoin.BaseTransactionBuilder)) {
```
wesgraham, line 218: Why are we removing this check?

modules/account-lib/test/resources/stx/stx.ts
```diff
@@ -90,7 +90,7 @@ export const RAW_TX_UNSIGNED =
   '0x80800000000400164247d6f2b425ac5771423ae6c80c754f7172b0000000000000000000000000000000b400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003020000000000051a1ae3f911d8f1d46d7416bfbe4b593fd41eac19cb00000000000003e800000000000000000000000000000000000000000000000000000000000000000000';
 
 export const SIGNED_TRANSACTION =
-  '0x80800000000400164247d6f2b425ac5771423ae6c80c754f7172b0000000000000000000000000000000b400011ae06c14c967f999184ea8a7913125f09ab64004446fca89940f092509124b9e773aef483e925476c78ec58166dcecab3875b8fab8e9aa4213179d164463962803020000000000051a1ae3f911d8f1d46d7416bfbe4b593fd41eac19cb00000000000003e800000000000000000000000000000000000000000000000000000000000000000000';
+  '80800000000400164247d6f2b425ac5771423ae6c80c754f7172b0000000000000000000000000000000b400011ae06c14c967f999184ea8a7913125f09ab64004446fca89940f092509124b9e773aef483e925476c78ec58166dcecab3875b8fab8e9aa4213179d164463962803020000000000051a1ae3f911d8f1d46d7416bfbe4b593fd41eac19cb00000000000003e800000000000000000000000000000000000000000000000000000000000000000000';
```
wesgraham, line 93: Why do we remove the 0x here but not in the other unit test file below?

fariedt: I didn't run the full tests and missed this change.  Fixed now.

modules/core/src/v2/coins/stx.ts
fariedt: The factory's type is set to `accountLib.Stx.TransactionBuilderFactory` because it has an additional method to add public keys to the transaction builder (needed for `explainTransaction`), and the base class doesn't have that.

wesgraham (approved):

PR #1133, opened 2021-04-23
chore: update dependency versions

tylerlevine:
Update superagent-proxy to ^2.1.0
Update stellar-sdk to ^8.1.1
Update tronweb to ^3.2.6
Update lerna to ^4.0.0

This should resolve (some of) the dependency audit failures.

Ticket: BG-27462

PR #1139, opened 2021-04-28
chore(account-lib): upgrade stellar-sdk dependency version

LeoKotschenreuther:
We have been on a really outdated version and I have a hunch that it might be causing some issues when upgrading other services to Node 14.


tylerlevine:
Already in progress in #1133

PR #1152, opened 2021-05-04
STLX-3031: Check transfer amount limits

JulioAchaval-SimTLiX:
Jira ticket: [STLX-3031](https://bitgoinc.atlassian.net/browse/STLX-3031)

- Max motes amount updated.
- Get transfer amount limits function added.
- Unit test added.


JulioAchaval-SimTLiX:
DEPRECATED

PR #1165, opened 2021-05-11
feat(account-lib): add latest version of algo-sdk

ryan-bitgo:
We want to replace all usages of the algosdk fork that we have with account-lib and have account-lib
consume the latest version of algosdk directly. This is because we don't want to maintain this fork
anymore and we want to standardize the implementation for all our coins to use account-lib.

This change adds the latest version of algo-sdk to account-lib. It also augments the algosdk module in the core package because the types from the latest version of the sdk are overriding the untyped fork.

ticket: [BG-31597](https://bitgoinc.atlassian.net/browse/BG-31597)

### DIFF
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -44,6 +44,7 @@
     "@taquito/local-forging": "6.3.5-beta.0",
     "@taquito/signer": "6.3.5-beta.0",
     "@types/lodash": "^4.14.151",
+    "algosdk": "^1.9.1",
     "bignumber.js": "^9.0.0",
     "bs58check": "^2.1.2",
     "casper-client-sdk": "1.0.39",
--- a/modules/core/src/v2/coins/algo.ts
+++ b/modules/core/src/v2/coins/algo.ts
@@ -193,7 +193,7 @@ export class Algo extends BaseCoin {
     return co<Buffer>(function* cosignMessage() {
       // key.prv actually holds the encoded seed, but we use the prv name to avoid breaking the keypair schema.
       // See jsdoc comment in isValidPrv
-      let seed = key.prv;
+      let seed: string | Uint8Array = key.prv;
       if (!self.isValidPrv(seed)) {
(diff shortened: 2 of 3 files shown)


modules/account-lib/package.json
```diff
@@ -44,6 +44,7 @@
     "@taquito/local-forging": "6.3.5-beta.0",
     "@taquito/signer": "6.3.5-beta.0",
     "@types/lodash": "^4.14.151",
+    "algosdk": "^1.9.1",
```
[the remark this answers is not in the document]
ryan-bitgo: We don't have an algo implementation in account-lib yet. In the other modules were using our fork of algosdk which is not the official one.

https://github.com/BitGo/BitGoJS/blob/master/modules/core/package.json#L53

argjv (approved):
LGTM but would like @twtaylorbitgo to review it too

modules/core/src/v2/coins/algo.ts
```diff
@@ -443,3 +443,62 @@ export class Algo extends BaseCoin {
...
+// -------------------------------------------------------------------------------------
+
+// Augments the algosdk module to include types from the BitGo algosdk fork.
+//
+// This is required because latest version of algosdk that is installed in
+// account-lib cause compilation to fail here without augmentation.
+//
+// TODO: Remove augmentation after algosdk fork is completely replaced by
+// account lib.
+declare module 'algosdk' {
```
tylerlevine, line 458: Module augmentations should really go in https://github.com/BitGo/BitGoJS/tree/master/types so we can use them across all packages in the monorepo

ryan-bitgo: I don't think we would want this because in account-lib where we have the latest version of the SDK installed, these augmentations don't exist because they only exist in the fork.

tylerlevine: Fair enough, this is fine with me as it is then

tylerlevine (approved):

PR #1166, opened 2021-05-12
feat(account-lib) add skeleton implementation for Algorand

ryan-bitgo:
Adds the skeleton implementation for onboarding Algorand to account lib.

Epic: [BG-31382](https://bitgoinc.atlassian.net/browse/BG-31382)


tylerlevine (approved):
so far so good. I'm not super familiar with account-lib, so I'm not sure if there's anything missing or not, but I think we can come back later to add any missing items while we implement these functions.

PR #1171, opened 2021-05-17
feat(account-lib): add setters for algo txn builder

ryan-bitgo:
Sets base attributes required for algo transaction and add validations checks

ticket: [BG-31556](https://bitgoinc.atlassian.net/browse/BG-31677)

PR #1177, opened 2021-05-19
feat(account-lib): add asset transfer builder implementation

ryan-bitgo:
Adds implementations for asset transfer builder. Also includes whitelisting token support.

Depends on #1171 

ticket: [BG-31565](https://bitgoinc.atlassian.net/browse/BG-31565)


modules/account-lib/src/coin/algo/assetTransferBuilder.ts
```diff
@@ -1,26 +1,171 @@
...
+    this._transaction.setAlgoTransaction(
+      algosdk.makeAssetTransferTxnWithSuggestedParams(
+        this._sender,
+        this._receiver,
+        this._closeTo,
+        undefined,
+        this._assetAmount,
+        this._note,
+        this._assetIndex,
+        {
```
fdmirza, line 124: can we suggested params from TransactionBuilder?

ryan-bitgo: ah yeah forgot about that

fdmirza (approved):

modules/account-lib/src/coin/algo/assetTransferBuilder.ts
```diff
@@ -1,26 +1,188 @@
...
+import BigNumber from 'bignumber.js';
+import algosdk from 'algosdk';
+import { BaseAddress } from '../baseCoin/iface';
+import { InvalidTransactionError } from '../baseCoin/errors';
+import { TransactionType } from '../baseCoin';
 import { TransactionBuilder } from './transactionBuilder';
 import { Transaction } from './transaction';
+import { AssetTransferTxnSchema } from './txnSchema';
 
 export class AssetTransferBuilder extends TransactionBuilder {
```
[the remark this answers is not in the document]
ryan-bitgo: hmm do we still keep "assetId" as a field then? versus tokenId? might be confusing to the client.

modules/account-lib/src/coin/algo/assetTransferBuilder.ts
```diff
@@ -1,26 +1,188 @@
...
+  private _assetIndex: number;
+  private _assetAmount: number | bigint;
+  private _receiver: string;
+  private _closeTo?: string;
```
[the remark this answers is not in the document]
ryan-bitgo: Subclassing won't work due to the nature of how `buildImplementation` works. 

https://github.com/BitGo/BitGoJS/blob/BG-31382-feat/algo-tokens/modules/account-lib/src/coin/algo/transferBuilder.ts#L34

Build implementaiton needs to be called last. If we subclass asset transfer builder to transfer builder, then transfer builder will overwrite the txn.

modules/account-lib/src/coin/algo/assetTransferBuilder.ts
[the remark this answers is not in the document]
pancham88: No need to keep assetId. However, this class shall be derived from TransferBuilder.

pancham88 (changes requested):

modules/account-lib/src/coin/algo/assetTransferBuilder.ts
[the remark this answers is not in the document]
pancham88: The reason I suggest to derive from TransferBuilder is because asset transfer is specialization of regular transfer, with addition assetId and a different transaction type.
Can we have a method `internalBuildImplementation()` which is called from `TransferBuilder.buildImplementation()`? This method shall be overridden in `assetTransfer`.

Alternatively, if the approach cannot be implemented for some reason - let us have the property and method names in sync between AssetTransferBuilder and TransferBuilder.

pancham88 (changes requested):

modules/account-lib/src/coin/algo/assetTransferBuilder.ts
```diff
@@ -1,26 +1,120 @@
...
+import algosdk from 'algosdk';
+import { BaseAddress } from '../baseCoin/iface';
+import { InvalidTransactionError } from '../baseCoin/errors';
+import { TransactionType } from '../baseCoin';
+import { TransferBuilder } from './transferBuilder';
 import { Transaction } from './transaction';
+import { AssetTransferTxnSchema } from './txnSchema';
+
+export class AssetTransferBuilder extends TransferBuilder {
+  private _assetId: number;
```
pancham88, line 11: Since BitGo uses the term token, let us use `tokenId` instead of `assetId`

modules/account-lib/src/coin/algo/assetTransferBuilder.ts
```diff
@@ -1,26 +1,120 @@
...
-export class AssetTransferBuilder extends TransactionBuilder {
   constructor(coinConfig: Readonly<CoinConfig>) {
     super(coinConfig);
   }
 
-  /** @inheritdoc */
-  protected buildImplementation(): Promise<Transaction> {
-    throw new NotImplementedError('buildImplementation not implemented');
+  /**
+   * Sets the asset index.
```
pancham88, line 18: (nit) Use tokenId in docs as well

modules/account-lib/src/coin/algo/assetTransferBuilder.ts
```diff
@@ -1,26 +1,120 @@
...
+   * To allow list an asset, you send 0 units of the asset to yourself.
+   *
+   * This method sets the assetIndex, sender, receiver, asset amount, and
+   * fee parameters to their respective values to allowlist and asset.
+   *
+   * @param {number} assetIndex The unique identifier of the asset.
+   * @param {BaseAddress} userAddress The address of the user.
+   * @returns {TransactionBuilder} This transaction builder.
+   */
+  allowListAsset(assetIndex: number, userAddress: BaseAddress): this {
```
pancham88, line 48: Similar to above, please use token instead of asset

pancham88 (changes requested):

pancham88 (approved):

PR #1183, opened 2021-05-21
feat(utxo-lib): use new package name and new external links

peng-huang-cc:


tylerlevine (approved):
Thanks for the PR! We require commits to be signed for this repo, so I'll sign your commits and push as a separate branch.

PR #1189, opened 2021-05-24
fix(core): allow paygo outputs for empty verification options object

tylerlevine:
The UI will pass an empty object for the verifications options, so this
caused a breaking change in behavior for PayGo users who use the UI.

Instead, consider paygo outputs to be allowed unless the verification
option to allow them is both set and false.

Ticket: CR-296


ericcrosson-bitgo (approved):
Beautiful as usual

PR #1191, opened 2021-05-25
fix: Update freeze request to include sending params

BRosentreter:
Freeze wasn't sending the params with the post-call, so it always defaulted to 1 hour.


tylerlevine (changes requested):
good catch, but please add a test

tylerlevine (approved):

PR #1192, opened 2021-05-27
feat(core): export tx Enum Type from core

alebusse:
Re-export accountLib imported transaction types

[stlx-3452](https://bitgoinc.atlassian.net/browse/STLX-3452)

### DIFF
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -20,6 +20,10 @@ import { PendingApprovals } from './pendingApprovals';
 import { Keychain, Keychains, KeyIndices } from './keychains';
 import { Enterprises } from './enterprises';
 
+// re-export account lib transaction types
+import { BaseCoin as AccountLibBasecoin } from '@bitgo/account-lib';
+export type TransactionType = AccountLibBasecoin.TransactionType;
+
 export interface TransactionRecipient {
   address: string;
   amount: string | number;


rliberal-simtlix (approved):

CapnMigraine (dismissed):
lgtm, but I'd like to have @tylerlevine's review as well.

tylerlevine (changes requested):
Why do we need to re-export account lib types?

modules/core/src/v2/types.ts
```diff
@@ -14,3 +16,13 @@ export interface RequestTracer {
   inc(): void;
   toString(): string;
 }
+
+
+export class Types {
```
rliberal-simtlix, line 21: Talked with Tyler, better if we move this to basecoin
https://github.com/BitGo/BitGoJS/blob/db79e9500298563586a1febe1820ed245dda8fd7/modules/core/src/v2/baseCoin.ts#L4

tylerlevine:
Yeah, probably this is all we need to do in `core`:
```
diff --git a/modules/core/src/v2/baseCoin.ts b/modules/core/src/v2/baseCoin.ts
index 2a25d9f4..e58855a2 100644
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -12,6 +12,10 @@ import { NodeCallback } from './types';
 import { RequestTracer } from './internal/util';
 
+// re-export account lib transaction types
+import { BaseCoin as AccountLibBasecoin } from '@bitgo/account-lib';
+export type TransactionType = AccountLibBasecoin.TransactionType;
+
 import { Wallets } from './wallets';
 import { Markets } from './markets';
```

Then we can use it in the client by doing this:
```
import { BaseCoin } from 'bitgo';
// use BaseCoin.TransactionType
```

tylerlevine (approved):

rliberal-simtlix (approved):

CapnMigraine (approved):

PR #1208, opened 2021-06-07
feat:(account-lib): record key pair in protocol format

nimraarif21:
record public and secret keys in protocol format and their test cases

### DIFF
--- a/modules/account-lib/src/coin/algo/errors.ts
+++ b/modules/account-lib/src/coin/algo/errors.ts
@@ -1,4 +1,4 @@
-import { BuildTransactionError } from '../baseCoin/errors';
+import { BuildTransactionError, InvalidKey } from '../baseCoin/errors';
 
 export class InsufficientFeeError extends BuildTransactionError {
   constructor(providedFee: number, minimum: number) {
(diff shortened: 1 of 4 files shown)


nimraarif21:
@fdmirza @ryan-bitgo @gibrangul95

modules/account-lib/src/coin/algo/keyPair.ts
```diff
@@ -1,7 +1,8 @@
 /* eslint-disable no-unused-vars */
+import { PrivateKey, PublicKey } from '@hashgraph/sdk';
```
ryan-bitgo, line 2: We shouldn't be using the hedera hashgraph library for algorand. Algorand doesnt have a special format so this can be a simple passthrough implementation. 

i.e. `return { pub }`

modules/account-lib/src/coin/algo/errors.ts
```diff
@@ -16,9 +16,31 @@ export class AddressValidationError extends BuildTransactionError {
...
 export class KeyDilutionError extends BuildTransactionError {
   constructor(incorrectValue: number) {
     super(`The keyDilution value '${incorrectValue}' must be less than or equal to the square root of the voteKey validity range.`);
     this.name = KeyDilutionError.name;
   }
-}
\ No newline at end of file
+}
+
+export class PublicKeyValidationError extends InvalidKey {
```
ryan-bitgo, line 27: we don't need these. see: https://github.com/BitGo/BitGoJS/blob/master/modules/account-lib/src/coin/baseCoin/errors.ts#L86

ryan-bitgo (changes requested):

PR #1209, opened 2021-06-07
feat(algo): BG-31598 Add ALGO token support

pancham88:
[JIRA BG-31598](https://bitgoinc.atlassian.net/browse/BG-31598)

Add support for ALGo transactions in BitGoJS statics


modules/statics/src/account.ts
```diff
@@ -183,6 +187,28 @@ export class HederaCoin extends AccountCoinToken {
...
+ * The Algo network supports tokens (assets)
+ * Algo tokens work similar to native ALGO coin, but the token name is determined by the chain,
+ * and the unique asset id on the chain: (t)algo:<assetId>
+ */
+export class AlgoCoin extends AccountCoinToken {
+  public assetURL: string;
+
+  constructor(options: AlgoCoinConstructorOptions) {
+    super({
+      ...options,
```
ryan-bitgo, line 200: why this instead of just `super(options)`?

modules/statics/src/account.ts
```diff
@@ -183,6 +187,28 @@ export class HederaCoin extends AccountCoinToken {
   }
 }
 
+/**
+ * The Algo network supports tokens (assets)
+ * Algo tokens work similar to native ALGO coin, but the token name is determined by the chain,
+ * and the unique asset id on the chain: (t)algo:<assetId>
```
ryan-bitgo, line 193: [np] This doesn't really describe what the class does and it worded confusingly.

modules/statics/src/account.ts
```diff
@@ -183,6 +187,28 @@ export class HederaCoin extends AccountCoinToken {
...
+ */
+export class AlgoCoin extends AccountCoinToken {
+  public assetURL: string;
+
+  constructor(options: AlgoCoinConstructorOptions) {
+    super({
+      ...options,
+    });
+
+    const domainPattern = /^(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]$/;
```
ryan-bitgo, line 203: could we add a comment for a domain that would match this regex? Or the source if you pulled it from somewhere.

modules/statics/src/account.ts
```diff
@@ -183,6 +187,28 @@ export class HederaCoin extends AccountCoinToken {
...
+    if (options.assetURL !== '' && !options.assetURL.match(domainPattern)) {
+      throw new InvalidDomainError(options.name, options.assetURL);
+    }
+
+    this.assetURL = options.assetURL as string;
```
ryan-bitgo, line 208: why do we need `as string` here?

modules/statics/src/account.ts
```diff
@@ -592,3 +618,75 @@ export function hederaCoin(
...
+  fullName: string,
+  decimalPlaces: number,
+  asset: UnderlyingAsset,
+  assetURL: string = '',
+  features: CoinFeature[] = AccountCoin.DEFAULT_FEATURES,
+  prefix: string = '',
+  suffix: string = name.toUpperCase(),
+  network: AccountNetwork = Networks.test.algorand
+) {
+  return stellarToken(name, fullName, decimalPlaces, asset, assetURL, features, prefix, suffix, network);
```
ryan-bitgo, line 691: would be good to document why we use stellarToken here for newcomers that aren't familar.

ryan-bitgo (changes requested):

modules/statics/src/account.ts
pancham88: It can be `super(options)` as well, I just kept it consistent with all other constructors in the same file

modules/statics/src/account.ts
pancham88: Have updated it ti be talgoToken - had missed to add it after my change in earlier push.

modules/statics/src/account.ts
pancham88: I have updated the implementation to use a URL class instead of regex, which was borrowed from steller

ryan-bitgo (approved):

modules/statics/src/account.ts
```diff
@@ -183,6 +187,32 @@ export class HederaCoin extends AccountCoinToken {
...
+    super({
+      ...options,
+    });
+
+    if (options.assetURL) {
+      try {
+        new URL(options.assetURL);
+      } catch (ex) {
+        throw new InvalidDomainError(options.name, options.assetURL);
+      }
```
OttoAllmendinger, line 209: this an overly broad catch

for instance, on node 8 there is no global `URL` so this will incorrectly throw `InvalidDomainError`

please guard this with `e.code == 'ERR_INVALID_URL'`

PR #1217, opened 2021-06-11
Fix/algo keyreg lint fix

gibrangul95:
cc @ryan-bitgo  @fdmirza


gibrangul95:
Solved in the Explain transaction PR

PR #1221, opened 2021-06-14
Feat/algo transaction builder factory

nimraarif21:
@ryan-bitgo @fdmirza @gibrangul95


modules/account-lib/src/coin/algo/transactionBuilder.ts
```diff
@@ -309,6 +312,13 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     return this;
   }
 
+  setSigners(addrs: string | string[]): this {
+    const signers = addrs instanceof Array ? addrs : [addrs];
```
ryan-bitgo, line 316: prefer `Array.isArray`

ryan-bitgo (changes requested):
does this PR need to include the half signed txns changes? the diff is hard to read

modules/account-lib/src/coin/algo/transactionBuilder.ts
nimraarif21: I have changed this.

nimraarif21:
> does this PR need to include the half signed txns changes? the diff is hard to read

My branch depends on the half signed txns changes for it to work correctly.

PR #1224, opened 2021-06-15
feat(algo): BG-31598-Add-ALGO-Token-Support

pancham88:
[JIRA BG-31598](https://bitgoinc.atlassian.net/browse/BG-31598)

Added support for ALGO tokens and updated previous commit to correct ALGO token configurations and use the term token instead of asset since BitGo uses the term token.


ryan-bitgo:
seems like audit is failing. lgtm otherwise.

PR #1228, opened 2021-06-15
fix: change the token address for cqt

barathcj:


tylerlevine:
https://bitgoinc.atlassian.net/browse/BG-32553

tylerlevine (approved):

PR #1236, opened 2021-06-18
feat(eth): verify pre-built eth txns

benjamin-wu-bitgo:
### Feature Description 

Client-side verification of transaction pre-builds received from BitGo servers should be verified to match transaction parameters specified by the client before the client signs the transaction.

This is an extra safeguard to protect the client from blindly signing malicious transaction pre-builds from the BitGo servers that do not match the transaction parameters specified by the client. An example of when this can be an issue is if BItGo servers were to be compromised and pre-builds are getting sent to the client that try to send funds to an attacker instead of the recipient address provided by the client.

### Implementation

This PR implements the `verifyTransaction` function for Eth which accepts `txParams` supplied by the client, `txPrebuild` supplied by the BitGo server, and the `wallet` for the transaction as parameters. The function returns true if the transaction is verified, and fails to verify throwing an error with the reason for failing otherwise.

It first checks that the parameters required exist, that the `txPrebuild` only has one recipient (note even multi-recipient transactions specified by the client `txParams` get squashed into a single recipient send to a batcher contract address with a data payload), and that the `txPrebuild` is not both a hop transaction and a batcher transaction.

Then, depending on the type of `txPrebuild` (hop, batcher, or normal), we perform different verification logic.

For a hop transaction, we check that there is only one recipient in the `txParams` and that the recipient amounts match those in `txParams`.  Then, we verify that the `txPrebuild` actually sends to the hop contract address corresponding to the `hopTransaction` in the `txPrebuild`. Finally, we convert the `txParams` `TransactionRecipient` array into a `Recipient` array by retyping the `amount` field from `string | number` to `string` to leverage the existing `validateHopPrebuild` function which checks the `hopTransaction` sends the correct amount to the recipient specified in `txParams`.

For a batch transaction, we verify that the `txPrebuild` is sending the correct total amount to a batcher contract address.

For a normal transaction (that is neither of the above), we check that there is only one recipient in the `txParams` and that the recipient amounts and address match those in `txParams`.

Lastly, we always check the coin of the `txPrebuild` matches the coin we are sending from the wallet.

### Testing

This PR includes unit tests covering all lines in the `verifyTransaction` function.

### Future Considerations

We do not yet fully verify batch transactions as we do not verify the data payload sent to the batcher contract address dictating how the total funds are re-distributed to the multiple recipients. Batch contracts are not yet in prod and parsing the data payload to verify this would be challenging and not worth doing at the moment. We do verify we are sending to a batcher contract address with the correct total amount though.

### DIFF
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -23,6 +23,8 @@ import {
   TransactionPrebuild as BaseTransactionPrebuild,
   VerifyAddressOptions,
   VerifyTransactionOptions,
+  TransactionParams,
+  TransactionRecipient,
 } from '../baseCoin';
 import { Erc20Token } from './erc20Token';
 import { BitGo } from '../../bitgo';
@@ -32,6 +34,7 @@ import * as common from '../../common';
 import * as config from '../../config';
 import { Util } from '../internal/util';
 import { EthereumLibraryUnavailableError } from '../../errors';
+import { BaseCoin as StaticsBaseCoin, EthereumNetwork } from '@bitgo/statics';
 
 const co = Bluebird.coroutine;
 const debug = debugLib('bitgo:v2:eth');
@@ -87,6 +90,14 @@ interface HopPrebuild {
   tx: string;
   id: string;
   signature: string;
+  paymentId: string;
+  gasPrice: number;
+  gasLimit: number;
+  amount: number;
+  recipient: string;
+  nonce: number;
+  userReqSig: string;
+  gasPriceMax: number;
 }
 
 interface Recipient {
@@ -224,11 +235,30 @@ interface FeeEstimate {
   feeEstimate: number;
 }
 
-interface TransactionPrebuild extends BaseTransactionPrebuild {
+export interface TransactionPrebuild extends BaseTransactionPrebuild {
   hopTransaction?: HopPrebuild;
   buildParams: {
     recipients: Recipient[];
   };
+  recipients: TransactionRecipient[];
+  nextContractSequenceId: string;
+  gasPrice: number;
+  gasLimit: number;
+  isBatch: boolean;
+  coin: string;
+  token?: string;
+}
+
+// TODO: This interface will need to be updated for the new fee model introduced in the London Hard Fork
+interface EthTransactionParams extends TransactionParams {
+  gasPrice?: number;
+  gasLimit?: number;
+  hopParams?: HopParams;
+}
+
+interface VerifyEthTransactionOptions extends VerifyTransactionOptions {
+  txPrebuild: TransactionPrebuild;
+  txParams: EthTransactionParams;
 }
 
 interface PresignTransactionOptions extends TransactionPrebuild, BasePresignTransactionOptions {
@@ -252,8 +282,15 @@ interface RecoverTokenTransaction {
 export class Eth extends BaseCoin {
   static hopTransactionSalt = 'bitgoHopAddressRequestSalt';
 
-  static createInstance(bitgo: BitGo): BaseCoin {
-    return new Eth(bitgo);
+  readonly staticsCoin?: Readonly<StaticsBaseCoin>;
+
+  protected constructor(bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) {
+    super(bitgo);
+    this.staticsCoin = staticsCoin;
+  }
+
+  static createInstance(bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>): BaseCoin {
+    return new Eth(bitgo, staticsCoin);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


modules/core/src/v2/coins/eth.ts
```diff
@@ -1483,7 +1513,106 @@ export class Eth extends BaseCoin {
...
+      if (!txParams?.recipients || !txPrebuild?.recipients || !wallet) {
+        throw new Error(`missing params`);
+      }
+      if (txPrebuild.hopTransaction && txPrebuild.isBatch) {
+        throw new Error(`tx cannot be both a batch and hop transaction`);
+      }
+      if (txPrebuild.recipients.length !== 1) {
+        throw new Error(`txPrebuild should only have 1 recipient but ${txPrebuild.recipients.length} found`);
+      }
+      if (txPrebuild.hopTransaction) {
```
tylerlevine, line 1543: Could these individual validation cases be broken out into separate `protected` functions (eg, `verifyHopTransaction()`, `verifyBatchTransaction()`, etc)? That should help improve readability.

modules/statics/src/coins.ts
```diff
@@ -27,6 +27,12 @@ const XTZ_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.ENTERPRISE_PA
 const CSPR_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.REQUIRES_RESERVE];
 const ALGO_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.SUPPORTS_TOKENS];
 
+export const batcherContractAddresses = {
```
tylerlevine, line 30: I'm not sure this is a great place for this, but putting it on the coin might be more trouble than it's worth. If you do want to do this though, here's what I'd do:

* Add a `batcherContractAddress?: string` property to the [`EthereumNetwork` interface](https://github.com/BitGo/BitGoJS/blob/db7d1c1819d31632c8d2a89387003f944443362a/modules/statics/src/networks.ts#L47)
* Add the `batcherContractAddress: '0x_the_address'` property to the implementation for Ethereum, Kovan, and Goerli ([here](https://github.com/BitGo/BitGoJS/blob/db7d1c1819d31632c8d2a89387003f944443362a/modules/statics/src/networks.ts#L218), [here](https://github.com/BitGo/BitGoJS/blob/db7d1c1819d31632c8d2a89387003f944443362a/modules/statics/src/networks.ts#L238), and [here](https://github.com/BitGo/BitGoJS/blob/db7d1c1819d31632c8d2a89387003f944443362a/modules/statics/src/networks.ts#L246))

Then you can get the correct batcher contract address for the current coin by capturing the 2nd param to `createInstance` (the statics coin, see [here](https://github.com/BitGo/BitGoJS/blob/5ad86846805b94eec3f125f33a8579286c3fc7d8/modules/core/src/v2/coinFactory.ts#L90). It's currently not being captured in the `createInstance` factory method). If you capture it as `this.staticsCoin`, then you can get the actual batcher address using `this.staticsCoin.network.batcherAddress`.

tylerlevine (approved):
Nice work, no major problems for me but I have a couple ideas on how to clean things up a bit.

Please wait for a review from someone on the eth team however so they can check that the verification logic is correct and comprehensive.

modules/core/src/v2/coins/eth.ts
```diff
@@ -224,11 +235,30 @@ interface FeeEstimate {
...
 
-interface TransactionPrebuild extends BaseTransactionPrebuild {
+export interface TransactionPrebuild extends BaseTransactionPrebuild {
   hopTransaction?: HopPrebuild;
   buildParams: {
     recipients: Recipient[];
   };
+  recipients?: TransactionRecipient[];
+  nextContractSequenceId?: string;
+  gasPrice?: number;
```
wesgraham, line 245: will this field ever be empty? Unsure, if it can be ignore

modules/core/src/v2/coins/eth.ts
```diff
@@ -224,11 +235,30 @@ interface FeeEstimate {
...
+  gasLimit?: number;
+  isBatch?: boolean;
+  coin?: string;
+  token?: string;
+  // eth2DepositInfo?: any;
+}
+
+interface EthTransactionParams extends TransactionParams {
```
wesgraham, line 253: Also might be worth leaving a TODO for the london hard fork task that will likely need to update this

modules/core/src/v2/coins/eth.ts
benjamin-wu-bitgo: I considered splitting out the validation cases into separate functions but it seems that doing so would require me to add a bunch of extra null/undefined checks (by pushing the check that the required parameters are provided down into the validation function for each case) or else the linter would complain.

Alternatively, I would need to create new interfaces that mark parameters as required rather than optional (e.g. create a new txPrebuild interface for hop transaction validation that has the hopTransaction parameter as required instead of optional in the txPrebuild).

Either way, it seems like it trying to split out functions for each case will create a lot of extra lines of code.

Unless you have an idea for splitting out functions without having to add extra interfaces or null/undefined checks, this doesn't seem worth it.

modules/core/src/v2/coins/eth.ts
benjamin-wu-bitgo: Some of these fields won't ever be empty so I updated the interface to no longer mark them as optional.

modules/core/src/v2/coins/eth.ts
benjamin-wu-bitgo: Added TODO comment

modules/statics/src/coins.ts
benjamin-wu-bitgo: Moved `batcherContractAddress` to the `EthereumNetwork` interface

tylerlevine (approved):
lgtm

:1st_place_medal:

wesgraham (approved):

PR #1242, opened 2021-06-25
Adding comment to SubmitTransactionOptions

dmu1313:
Adding a comment field to SubmitTransactionOptions so that a comment
can be uploaded along with transactions.

Ticket: BG-32873


modules/core/src/v2/wallet.ts
```diff
@@ -332,6 +332,7 @@ export interface SubmitTransactionOptions {
   otp?: string;
   txHex?: string;
   halfSigned?: string;
+  comment ?: string;
```
tylerlevine, line 335: Extra space here before the `?`

tylerlevine (approved):
tiny styling nit but not a blocker

tylerlevine (approved):

PR #1245, opened 2021-06-30
BG 33901: Moved block exploring into coin controllers

raovinnakota:
Card: BG 33901: https://bitgoinc.atlassian.net/browse/BG-33091?filter=-1
Abstracting the tx explorer from recovery to UTXO coin controller

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1700,6 +1700,14 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   protected abstract getAddressInfoFromExplorer(address: string, apiKey?: string): Bluebird<AddressInfo>;
   protected abstract getUnspentInfoFromExplorer(address: string, apiKey?: string): Bluebird<UnspentInfo[]>;
 
+  getTxInfoFromExplorer(faultyTxId: string): any {
+    return co(function *getUnspentFromWrongChain() {
+      const TX_INFO_URL = this.url(`/public/tx/${faultyTxId}`);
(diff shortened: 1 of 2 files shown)


raovinnakota:
Moved the transaction explorer into the coin controller so that specific coins can override with 3rd party explorers when ready to implement

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1700,6 +1700,14 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   protected abstract getAddressInfoFromExplorer(address: string, apiKey?: string): Bluebird<AddressInfo>;
   protected abstract getUnspentInfoFromExplorer(address: string, apiKey?: string): Bluebird<UnspentInfo[]>;
 
+  getUnspentFromWrongChain(faultyTxId: string): any {
```
tylerlevine, line 1703: Name here seems a bit strange - maybe something like `getTxInfoFromExplorer` would be a bit more descriptive? As it is, this doesn't have much to do with unspents or wrong chains.

tylerlevine (changes requested):
comment on method name, but other than that lgtm

modules/core/src/v2/coins/abstractUtxoCoin.ts
raovinnakota: Yeah that sounds good. The tests used to the "Wrong Chain" as the descriptor, which led me to name it that. Going to change now.

raovinnakota: Change pushed @tylerlevine

tylerlevine (approved):

PR #1246, opened 2021-07-01
fix(eos): can accept addresses with memoId when making recovery

bjon-bitgo:
transactions.

Ticket: [BG-30012](https://bitgoinc.atlassian.net/browse/BG-30012)

Originally, is someone called the recover method with a root address or destination address having a `?memoId=0`, it will try complain that the address is more than 12 characters. This is because we did not strip the `?memoId=0` before making a http request to the chain. Now we do.

The memoId can now be passed in to the destination address. Also, it will not fail if you add it to the root address. Modified core/v2/coins/eos.ts. Also added a basic testcase for the recovery method, as it never existed before. If I knew more about the EOS coin, I would be more elaborate with the test cases. 

Side Note: In the unit tests portion, I had to add a new resources folder to mock the http requests the recover function makes. The public keys and addresses in those files are not linked to any wallet. From my understanding, all other things should be safe to just be kept in the json file.


modules/core/test/v2/unit/coins/eos.ts
```diff
@@ -103,6 +106,67 @@ describe('EOS:', function() {
...
   });
 
+  it('should create unsigned recovery transaction without Bitgo', async function() {
+    const userKey = 'xpub661MyMwAqRbcH1oUADxatLuKkVjaDB2zTNJoZQsGVQEvoogpbXJw24QMokNwFKj9Qhci6KWaCcQKrzpL4LCQXXX3YpTQxgD9KLBjhDrUWo4';
+    const backupKey = 'xpub661MyMwAqRbcH1n6sgY29G7dAxL7twS8rt1jyuuQb1kfnA7s3FJPGoVqb9JenXkeJmC4jZ8iVscn3AH6MkYAVc61FTYCHpxv5cxWar5Jw3C';
+    const rootAddress = 'i1skda3kso43';
+    const destinationAddress = 'ks13kdh245ls';
+
+    // load mock responses from json files
+    const rootResponse = JSON.parse(
```
tylerlevine, line 116: You don't need to use `fs.readFileSync` to load json files, you can just `import` them if they are set up as the default export - they are valid js.

modules/core/test/v2/unit/coins/eos.ts
```diff
@@ -103,6 +106,67 @@ describe('EOS:', function() {
...
+      fs.readFileSync(`${__dirname}/resources/eos/getAccountResponseSuccess1.json`).toString());
+    const desResponse = JSON.parse(
+      fs.readFileSync(`${__dirname}/resources/eos/getAccountResponseSuccess2.json`).toString());
+    const getInfoResponse = JSON.parse(
+      fs.readFileSync(`${__dirname}/resources/eos/getInfoResponseSuccess1.json`).toString());
+    const getBlockResponse = JSON.parse(
+      fs.readFileSync(`${__dirname}/resources/eos/getBlockResponseSuccess1.json`).toString());
+
+    // mock responses to the block chain
+    const callBack = sinon.createSandbox().stub(Eos.prototype, <any>'getDataFromNode');
```
tylerlevine, line 126: do you ever call `restore()` on the stub or the sandbox? If not, you're leaving the mock in place for future callers

modules/core/test/v2/unit/coins/eos.ts
```diff
@@ -103,6 +106,67 @@ describe('EOS:', function() {
...
+
+    // can create unsigned recovery transaction
+    const unsignedRecoveryTransaction = await basecoin.recover({
+      userKey,
+      backupKey,
+      bitgoKey: 'key',
+      recoveryDestination: destinationAddress,
+      rootAddress,
+    });
+    unsignedRecoveryTransaction.txid.should
```
tylerlevine, line 151: Why assert the txid? Is that an important part of the test? It's better form to just assert on the properties that you want to hold true (destination and amount are the most important things to verify).

modules/core/test/v2/unit/coins/resources/eos/getAccountResponseSuccess1.json
```diff
@@ -0,0 +1,80 @@
+{
```
tylerlevine, line 1: `export = { ...`

tylerlevine (changes requested):

modules/core/test/v2/unit/coins/resources/eos/getAccountResponseSuccess1.json
tylerlevine: and rename to `ts` if you want to do this

modules/core/test/v2/unit/coins/eos.ts
bjon-bitgo: txid is not really important, yea.

modules/core/test/v2/unit/coins/resources/eosResponses.ts
```diff
@@ -0,0 +1,212 @@
+const getAccountResponseSuccess1 = {
```
tylerlevine, line 1: These should go in `test/v2/fixtures/coins/eos.ts`

tylerlevine (changes requested):

tylerlevine (approved):

PR #1247, opened 2021-07-01
feat(account-lib): algo removed from account-lib and reverted in core

gibrangul95:
cc @fdmirza @pancham88 @ryan-bitgo 

Reverted algo in account-lib to NotImplemented and reverted core to previous version


ryan-bitgo:
I think we are deciding to not revert

ryan-bitgo (approved):

ryan-bitgo:
We have discussed that we should revert

PR #1253, opened 2021-07-02
fix(account-lib): fix amount in cspr delegate & undelegate builders

CapnMigraine:
Use string values to handle large amounts for delegate & undelegate builders

CLOSES TICKET: STLX-4299


alebusse (approved):
LGTM

PR #1256, opened 2021-07-05
fix(core): fix cspr address validation to account for transferId

CapnMigraine:
Change isValidAddress method for cspr in core to validate address with transferId
Add isValidAddressWithPaymentId method to account-lib to validate address with transferId


rliberal-simtlix (approved):
LGTM

PR #1260, opened 2021-07-07
additional memoId field for STX addresses

fariedt:
feat(account-lib): add support for "memoId" field for STX addresses

cc @wesgraham


wesgraham (dismissed):
LGTM thanks Faried!

wesgraham:
https://github.com/BitGo/BitGoJS/blob/f73c5c5fac9a0fe87e1ce518e9150c729714e2b7/modules/core/src/v2/coins/stx.ts#L98-L110 thinking about it now we might also need to amend this method as the `c32addressDecode` might fail right?

fariedt:
I think it's safe.  The `isValidAddress` function in core calls `isValidAddress` in account-lib, which calls [`validateStacksAddress`](https://github.com/blockstack/stacks.js/blob/master/packages/transactions/src/utils.ts#L201), and that uses `c32addressDecode` internally.

wesgraham (approved):

PR #1272, opened 2021-07-09
(BG-33901) Replace IMS Services with 3rd Party Litecoin explorer for Wrong Chain Recoveries

raovinnakota:
Link to card: https://bitgoinc.atlassian.net/browse/BG-33091

Changes in this PR:
- Added LTC block explorer to getTxInfo in ltc.ts
- Added nock response for LTC block explorer
- Modified LTC Wrong Chain Recovery Unit Tests to account for new nock response


tylerlevine (approved):
good enough for me. The recovery code is in a terrible state but this is good enough to unblock the LTC recovery case. Nice work

PR #1283, opened 2021-07-15
fix(eth): restore fixed hop transaction verification

benjamin-wu-bitgo:
### Background

A bug in the new Eth verification logic for hop transactions caused valid Eth hop transactions to be rejected. As a result, the verification logic was temporarily disabled for hop transactions.

The bug was caused by the verification logic requiring the `txPrebuild` from the sender to the hop address to have an amount exactly equal to the amount the client wanted to send to the final recipient. This was incorrect as the sender actually needs to send an extra amount to the hop address for the hop address to pay the gas fees and still get the desired amount to the final recipient.

### Fix

This PR adds the non-buggy hop transaction verification logic back to the BitGoJS client. 

It omits the buggy check for the amount sent to the hop address as adding such a check in a robust way does not seem feasible. The client has no trust-less way for verifying what a reasonable additional amount for hop transaction gas fees are (the gas price estimate is currently given to the client by BitGo servers). The best alternative at the moment is hard-coding a maximum additional gas fee that is acceptable. However, this may result in transactions getting rejected if gas prices inflate too much and can still allow a malicious attacker of BitGo servers to inflate gas fees which is still quite bad. 

Moreover, if an attacker compromises BitGo servers, inflating amounts sent to hop addresses is unlikely to be the most appealing attack vector as an attacker does not benefit from extra funds in a hop address unless they also compromise the hop address by compromising the HSM.

Verification of hop transactions also was not specifically requested by Block.one who originally asked for Eth verification logic to be added client-side.

### Testing

In addition to unit test coverage, manual tests using `sendMany` from the SDK was used to verify that correct normal, hop, batch, and token Eth transactions are verified and signed as expected.

Each mock hop `txPreBuild` in the unit tests were also updated to have an amount greater than what is specified by the client sent to the hop address to more accurately reflect the values of a real transaction. The extra amount is to ensure the hop address can pay the gas fees and have enough to send the amount specified by the client to the final recipient.

There is a unit test case that ensures sending amounts above what is specified by the client to the hop address is correctly accepted, verified, and signed.

### Ticket: BG-33834

### DIFF
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -1548,8 +1548,31 @@ export class Eth extends BaseCoin {
         throw new Error(`txPrebuild should only have 1 recipient but ${txPrebuild.recipients.length} found`);
       }
       if (txPrebuild.hopTransaction) {
-        // Temporarily do not verify hop transactions
-        return true;
+        // Check recipient amount for hop transaction
+        if (txParams.recipients.length !== 1) {
+          throw new Error(`hop transaction only supports 1 recipient but ${txParams.recipients.length} found`);
+        }
+
+        // Check tx sends to hop address
+        const decodedHopTx = new optionalDeps.EthTx(txPrebuild.hopTransaction.tx);
+        const expectedHopAddress = decodedHopTx.getSenderAddress().toString('hex');
+        if (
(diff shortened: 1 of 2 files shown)


barathcj (approved):
lgtm

PR #1284, opened 2021-07-15
feat(core): [STLX-4305] add feeLimit parameter to Send options

matilareo:
stlx-4309

## Changes 
- add feeLimit parameter to Send Options on core

### DIFF
--- a/modules/core/src/v2/coins/trx.ts
+++ b/modules/core/src/v2/coins/trx.ts
@@ -312,6 +312,26 @@ export class Trx extends BaseCoin {
     return hdNode.keyPair.__Q.getEncoded(false).toString('hex');
   }
 
+  /**
+   * Modify prebuild before sending it to the server.
+   * @param buildParams The whitelisted parameters for this prebuild
+   * @param callback
+   */
+  getExtraPrebuildParams(buildParams: any, callback?: NodeCallback<any>): Bluebird<any> {
+    const recipients: any[] = [{}];
+    return co<any>(function* () {
+      if (buildParams.recipients[0].data && buildParams.feeLimit) {
+        recipients[0].feeLimit = buildParams.feeLimit;
+        return {
+          recipients,
+        };
+      }
+      return {};
+    })
+      .call(this)
+      .asCallback(callback);
+  }
+
   compressedPubToHexAddress(pub: string): string {
     const byteArrayAddr = bitgoAccountLib.Trx.Utils.getByteArrayFromHexAddress(pub);
     const rawAddress = bitgoAccountLib.Trx.Utils.getRawAddressFromPubKey(byteArrayAddr);
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -339,6 +339,7 @@ export interface SendOptions {
   address?: string;
   amount?: number | string;
   data?: string;
+  feeLimit?: string;
   message?: string;
   walletPassphrase?: string;
   prv?: string;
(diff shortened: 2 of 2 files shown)


fmlenarduzzi (approved):

modules/core/src/v2/wallet.ts
```diff
@@ -1981,7 +1983,9 @@ export class Wallet {
     if (params.data && coin.transactionDataAllowed()) {
       recipients[0].data = params.data;
     }
-
+    if (coin.getFamily()==='trx' && params.feeLimit) {
```
twtaylorbitgo, line 1986: spacing

twtaylorbitgo (changes requested):

fmlenarduzzi (dismissed):

mcasares-simtlix (dismissed):

twtaylorbitgo (approved):

twtaylorbitgo (changes requested):
This needs to be on build and send transaction in BitGoJS

Also this should be called `maxFee` - this is a parameter we should re-use in WP

karankurbur (dismissed):

matilareo:
> This needs to be on build and send transaction in BitGoJS
> 
> Also this should be called `maxFee` - this is a parameter we should re-use in WP

The method that i've modified is doing both, build and send. 
look at these lines:

- https://github.com/BitGo/BitGoJS/blob/7f6ac94eab0d62805a5265bfba6f582a67339593/modules/core/src/v2/wallet.ts#L1990
- https://github.com/BitGo/BitGoJS/blob/7f6ac94eab0d62805a5265bfba6f582a67339593/modules/core/src/v2/wallet.ts#L2047
- https://github.com/BitGo/BitGoJS/blob/7f6ac94eab0d62805a5265bfba6f582a67339593/modules/core/src/v2/wallet.ts#L1807
- https://github.com/BitGo/BitGoJS/blob/7f6ac94eab0d62805a5265bfba6f582a67339593/modules/core/src/v2/wallet.ts#L1682

CapnMigraine (approved):

tylerlevine (changes requested):

modules/core/src/v2/wallet.ts
```diff
@@ -352,6 +353,7 @@ export interface SendManyOptions {
   recipients?: {
     address: string;
     amount: string | number;
+    feeLimit?: string;
```
tylerlevine, line 356: Why is this option specified per-recipient?

modules/core/src/v2/wallet.ts
```diff
@@ -1981,7 +1983,9 @@ export class Wallet {
     if (params.data && coin.transactionDataAllowed()) {
       recipients[0].data = params.data;
     }
-
+    if (coin.getFamily() === 'trx' && params.feeLimit) {
```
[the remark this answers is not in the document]
matilareo: you are right, thank you

modules/core/src/v2/wallet.ts
matilareo: this is because each contract call could have different fee limit

fmlenarduzzi (approved):

karankurbur (approved):

modules/core/src/v2/coins/trx.ts
```diff
@@ -312,6 +312,26 @@ export class Trx extends BaseCoin {
...
 
+  /**
+   * Modify prebuild before sending it to the server.
+   * @param buildParams The whitelisted parameters for this prebuild
+   * @param callback
+   */
+  getExtraPrebuildParams(buildParams: any, callback?: NodeCallback<any>): Bluebird<any> {
+    const recipients: any[] = [{}];
+    return co<any>(function* () {
+      if (buildParams.recipients[0].data && buildParams.feeLimit) {
```
tylerlevine, line 323: is only one recipient supported?

tylerlevine (approved):

PR #1286, opened 2021-07-16
Feat/eos core

gibrangul95:
cc @pancham88 @fdmirza @fariedt 
The complete implementation for Eos covering both account-lib and core


modules/account-lib/src/coin/eos/BuyRamBytesActionBuilder.ts
```diff
@@ -0,0 +1,99 @@
...
+   * @returns {string} The name of the action e.g. transfer, buyrambytes, delegatebw etc
+   */
+  actionName(): string {
+    return 'buyrambytes';
+  }
+
+  /** @inheritdoc */
+  build(builder: EosTxBuilder): EosJs.Serialize.Action {
+    const data = this.action.data;
+    if (typeof data === 'string') {
```
pancham88, line 66: Why the check for data type as string vs undefined (or not present)?

modules/account-lib/src/coin/eos/DeleteAuthActionBuilder.ts
```diff
@@ -0,0 +1,89 @@
+import * as EosJs from 'eosjs';
```
pancham88, line 1: Can you please link the documentation for these actions from EOS site? These can be included in each action documentation. It is not easy to find the documentation on EOS site?

modules/account-lib/src/coin/eos/BuyRamBytesActionBuilder.ts
gibrangul95: Data is returned as a string when we deserialize the transaction without actions. Deserializing with actions returns the data as an object. Planning to move this check into the parent class in the next PR/Commit

gibrangul95:
@pancham88 reminder for this repo

PR #1292, opened 2021-07-20
chore!(core): remove bluebird from bitgo object http methods

tylerlevine:
BREAKING CHANGE: bluebird-specific promise methods are no longer present
on the `get`/`post`/`put`/`del`/`patch` request helper methods on BitGo
objects.

BREAKING CHANGE: remove support for callbacks to async fns

Native promises don't support the `asCallback` or `nodeify` helpers that
are on Bluebird promises. We will need to add a compatiblity layer for
these since we don't (yet) want to entirely deprecate support for
callback style usage of Bitgo object methods, but for now let's remove
all the callback params and make everything work correctly as regular
async functions.

We still want the current API surface to remain callback compatible for
the moment, so convert from raw promises to bluebird promises where
callbacks are needed.

Ticket; BG-31214

### DIFF
--- a/modules/core/src/api.ts
+++ b/modules/core/src/api.ts
@@ -1,0 +1,127 @@
+/**
+ * @prettier
+ */
+import * as superagent from 'superagent';
+import * as urlLib from 'url';
+import * as querystring from 'querystring';
+import Debug from 'debug';
+
+import { ApiResponseError } from './errors';
+import { BitGo, VerifyResponseOptions } from './bitgo';
+
+const debug = Debug('bitgo:api');
+
+/**
+ * Serialize request data based on the request content type
+ * Note: Not sure this is still needed or even useful. Consider removing.
+ * @param req
+ */
+export function serializeRequestData(req: superagent.Request): string | undefined {
+  let data: string | Record<string, unknown> = (req as any)._data;
+  if (typeof data !== 'string') {
+    let contentType = req.get('Content-Type');
+    // Parse out just the content type from the header (ignore the charset)
+    if (contentType) {
+      contentType = contentType.split(';')[0];
+    }
+    let serialize = superagent.serialize[contentType];
+    if (!serialize && /[\/+]json\b/.test(contentType)) {
+      serialize = superagent.serialize['application/json'];
+    }
+    if (serialize) {
+      data = serialize(data);
+      (req as any)._data = data;
+      return data;
+    }
+  }
+}
+
+/**
+ * Set the superagent query string correctly for browsers or node.
+ * @param req
+ */
+export function setRequestQueryString(req: superagent.SuperAgentRequest): void {
+  const urlDetails = urlLib.parse(req.url);
+
+  let queryString: string | undefined;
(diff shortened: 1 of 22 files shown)


twtaylorbitgo (approved):

modules/core/src/api.ts
```diff
@@ -0,0 +1,127 @@
...
+    throw new ApiResponseError('invalid response HMAC, possible man-in-the-middle-attack', 511, errorDetails);
+  }
+
+  if (bitgo.getAuthVersion() === 3 && !verificationResponse.isInResponseValidityWindow) {
+    const errorDetails = {
+      timestamp: response.header.timestamp,
+      verificationTime: verificationResponse.verificationTime,
+    };
+    debug('Server response outside response validity time window: %O', errorDetails);
+    const error: any = new Error(
```
sangaman, line 119: Why are you using a regular `Error` object here cast to `any` instead of the `ApiResponseError` like a few lines above? It looks like that class has the `status` and `result` fields built in.

Edit: I realize now this is old code that's being moved from one file to another, but I guess the question still stands if only for curiosity's sake.

modules/core/src/bitgo.ts
```diff
@@ -48,63 +49,42 @@ if (!(process as any).browser) {
...
-      error = _error;
-      return reject(error);
-    }
-  });
-};
-
 // Handle HTTP errors appropriately, returning the result body, or a named
 // field from the body, if the optionalField parameter is provided.
-(superagent as any).Request.prototype.result = function (optionalField?: string) {
+function toBitgoRequest<ResponseResultType = any>(req: superagent.SuperAgentRequest): BitGoRequest<ResponseResultType> {
```
sangaman, line 54: This could maybe have some more comments about what it's doing and why it's needed, even just taking a line or two from the PR description and adding it here.

modules/core/src/bitgo.ts
```diff
@@ -1442,159 +1298,139 @@ export class BitGo {
...
-   * @param callback
    * @deprecated
    */
-  verifyPushToken(params: VerifyPushTokenOptions, callback?: NodeCallback<any>): Bluebird<any> {
-    const self = this;
-    return co(function *() {
-      if (!_.isObject(params)) {
-        throw new Error('required object params');
-      }
+  verifyPushToken(params: VerifyPushTokenOptions): any {
```
sangaman, line 1346: This method will return a promise, right?

sangaman (commented):
Looks good as far as I can tell, just a couple of questions.

modules/core/src/api.ts
tylerlevine: that's a good point, I will do this in a follow up

modules/core/src/bitgo.ts
tylerlevine: yep, that's worthwhile

modules/core/src/bitgo.ts
tylerlevine: Yes it will, good catch. I'll fix this in the follow up PR

PR #1296, opened 2021-07-21
STLX-4754 build fee info migration

nicolasbellanich-simtlix:
Move buildFeeInfo logic to account-lib.

[STLX-4754](https://bitgoinc.atlassian.net/browse/STLX-4754)

### DIFF
--- a/modules/account-lib/src/coin/algo/utils.ts
+++ b/modules/account-lib/src/coin/algo/utils.ts
@@ -12,6 +12,7 @@ import { KeyPair } from './keyPair';
 
 const ALGORAND_CHECKSUM_BYTE_LENGTH = 4;
 const ALGORAND_ADDRESS_LENGTH = 58;
+const ALGORAND_MINIMUM_FEE = 1000;
 
 /**
  * Determines whether the string is a composed of hex chars only.
@@ -172,7 +173,7 @@ export class Utils implements BaseUtils {
       return txInfo.estimateSize();
     }
 
-    const { fee = 1000 } = txInfo;
+    const { fee = 1 } = txInfo;
     const transaction = this.createMultisigTransaction(Object.assign({}, txInfo, { fee }));
     return transaction.estimateSize();
   }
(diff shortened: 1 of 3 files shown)


modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -198,6 +199,65 @@ export class Utils implements BaseUtils {
...
+
+  /**
+   * Check if passed number is positive and integer
+   *
+   * @param number {number} Number to validate.
+   * @returns {number} Same passed number.
+   */
+  protected positiveInt(number: number): any {
+    if (!Number.isInteger(number) || number < 0) {
+      throw new Error('Must be positive number');
```
rliberal-simtlix, line 257: nit: Must be integer positive number

modules/account-lib/test/unit/coin/algo/utils.ts
```diff
@@ -21,4 +22,25 @@ describe('utils', () => {
     const objEncodedByAlgo = algosdk.encodeObj(note);
     should.deepEqual(objEncoded, objEncodedByAlgo);
   });
+
+  it('should build fee data on Pay Tx and Non Participation Tx', () => {
```
rliberal-simtlix, line 26: This is only available for PayTx and NonParticipationTx? What about KeyRegTx? Do we need extra tests?

modules/account-lib/test/unit/coin/algo/utils.ts
```diff
@@ -21,4 +22,25 @@ describe('utils', () => {
...
+    should.equal(feeDataKeyRegTxn.feeInfo.size, 240);
+
+    const feeDataNonParticipationTxn = utils.getFeeData(nonParticipationTxn, feeRate);
+    should.equal(feeDataNonParticipationTxn.feeInfo.feeRate, 5);
+    should.equal(feeDataNonParticipationTxn.feeRate, 5);
+    should.equal(feeDataNonParticipationTxn.feeInfo.fee, 1000);
+    should.equal(feeDataNonParticipationTxn.feeInfo.size, 207);
+  });
+
+  it('should fail building fee info because fee is negative', () => {
```
rliberal-simtlix, line 42: Add extra tests when fee or feerate are missed values, should throw an excepction.

rliberal-simtlix (changes requested):

modules/account-lib/test/unit/coin/algo/utils.ts
nicolasbellanich-simtlix: About that. I`ll make the test about feeRate missed and I'll commit it. But I cant handle missed fee here. Because the fee is not on the input of getFeeData. It's made inside using the size calculated.

modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -198,6 +199,65 @@ export class Utils implements BaseUtils {
   encodeObj(obj: Record<string | number | symbol, any>): Uint8Array {
     return algosdk.encodeObj(obj);
   }
+  /**
```
JulioAchaval-SimTLiX, line 202: Add new line between both methods

modules/account-lib/test/unit/coin/algo/utils.ts
nicolasbellanich-simtlix: Ok about that. To make this tests over here I've run MS tests and copy the exactly returned Tx here. So nowdays we only have this two builders on MS which use buildFeeInfo. Keyreg will use it when my other PR is done, so now I cant figure out the spefic output to compare on mocha test.

modules/account-lib/src/coin/algo/utils.ts
JulioAchaval-SimTLiX: (Update unit test associated too)

modules/account-lib/test/unit/coin/algo/utils.ts
rliberal-simtlix: Sounds good let's have a meeting to check that. Thank you. Good Job. :+1:

nicolasbellanich-simtlix: About the feeRate missed I've just checked out that typescript handle this. Because getFeeData() require a number on it's sign. So if you send null or undefined typescript catches and throws the error.

modules/account-lib/test/unit/coin/algo/utils.ts
nicolasbellanich-simtlix: Let me add this

rliberal-simtlix (approved):

modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -258,6 +259,66 @@ export class Utils implements BaseUtils {
...
+
+  /**
+   * Check values for feeInfo building
+   *
+   * @param size {number} required size to check.
+   * @param feeData {Object} required fees to check.
+   * @returns {Object} with fee data validated.
+   */
+  protected validateFeeInfo(size: number, { feeRate, fee, baseFactor = 1000 }): any {
+    const fee_size = this.positiveInt(size);
```
pancham88, line 289: It will be preferred to have explicit one time validation vs accessing the value everytime via a wrapper that has validation as side-effect.

My suggestion is to not have positiveInt method and have explicit one time validation, such as:
if (!Number.isSafeNumber(fee) && fee < 0) {
    // throw error
}

pancham88 (changes requested):

modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -258,6 +259,74 @@ export class Utils implements BaseUtils {
...
+    return feeObj;
+  }
+
+  /**
+   * Check if passed number is positive and integer
+   *
+   * @param number {number} Number to validate.
+   * @returns {number} Same passed number.
+   */
+  protected positiveInt(number: number): any {
```
pancham88, line 324: Please remove this method since it is no longer used.

nicolasbellanich-simtlix: Sorry about that... done!

rliberal-simtlix (approved):

pancham88 (approved):

PR #1302, opened 2021-07-22
fix(statics): update goerli testnet tokens to use terc20

ryan-bitgo:
Fixes newly added test erc20 tokens to use the `terc20` method and goerli network

ticket: BG-34227


modules/statics/src/base.ts
```diff
@@ -319,7 +319,6 @@ export enum UnderlyingAsset {
   FET1 = 'fet1',
   FF1 = 'ff1',
   FFT = 'fft',
-  FIXED = 'fixed',
```
tylerlevine, line 322: probably we still want these underlying assets since 1 FIXED != 1 TERC20 != 1 GUSDT but I don't think it's a huge problem since we don't really care about pricing data for these. Up to you if you want to fix

tylerlevine (approved):

tylerlevine (approved):

PR #1307, opened 2021-07-23
feat(core): added algo token config on core

ofornaroli:
added algo token config on core

stlx-4986

### DIFF
--- a/modules/core/src/v2/coins/algoToken.ts
+++ b/modules/core/src/v2/coins/algoToken.ts
@@ -1,0 +1,86 @@
+/**
+ * @prettier
+ */
+import * as _ from 'lodash';
+import { BitGo } from '../../bitgo';
+import { Algo } from './algo';
+import { CoinConstructor } from '../coinFactory';
+import { BitGoJsError } from '../../errors';
+import { coins } from '@bitgo/statics';
+
+export interface AlgoTokenConfig {
+  name: string;
+  type: string;
+  coin: string;
+  network: string;
+  decimalPlaces: number;
+}
+
+export class AlgoToken extends Algo {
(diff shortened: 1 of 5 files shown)


rliberal-simtlix (approved):
LGTM

modules/core/src/v2/coins/algoToken.ts
```diff
@@ -0,0 +1,90 @@
...
+  get issuer() {
+    return this._issuer;
+  }
+
+  get decimalPlaces() {
+    return this.tokenConfig.decimalPlaces;
+  }
+
+  protected getAlgoNetwork() {
+    return this.tokenConfig.network === 'Testnet' ? coins.get('algo') : coins.get('talgo');
```
tylerlevine, line 72: this is going to return the `algo` or `talgo` coins, not the network. Are you sure that's what you want?

tylerlevine (approved):
looks good to me

modules/core/src/v2/coins/algoToken.ts
ofornaroli: I'm not sure if this is ok, because I don't know where will it is used, it's based on stellar.

pancham88: What does this method retiurn? Is it all the coin and tokens defined for algo/talgo?
It  will be appropriate to rename it as getAlgoTokens - unless there is a polymorphic code that calls getAlgoNetwork. However, since the name has `Algo` in it - this cannot be polymorphic. 
We can improve upon stellar implementation if there is opportunity.

pancham88 (changes requested):

pancham88 (approved):

PR #1308, opened 2021-07-26
fix(core): replace bitcoin-abc with ecash in blockchair apis

gianchandania:
Changes in this PR:

- Bitcoin Cash ABC has been rebranded to eCash. Blockchair has changed the routes related to Bitcoin Cash ABC. Replaced "bitcoin-abc" with "ecash" in blockchair related code and tests

- Fixed a typo in JSDoc comment for the function getUnspentInfoForCrossChainRecovery

- Clean up formatting errors in touched files

Ticket: BG-32764

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1719,7 +1719,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
 
   /**
    * Fetch unspent transaction outputs using IMS unspents API
-   * @params addresses
+   * @param addresses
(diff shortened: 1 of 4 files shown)


tylerlevine (commented):
change looks good to me, but are there any other issues that are going to require more changes (eg, `ecash` has 2 decimal places for the default unit). Has this been tested for compatibility?

gianchandania:
I checked for the changes that are done in eCash. Changing decimals from 8 to 2 is the only change that could have impacted our code. I tried to do a couple of recoveries and they are completed successfully, so seems like no other change would be required

tylerlevine (approved):

PR #1309, opened 2021-07-26
feat(core): allowed amount 0 on recipients for enable token

ofornaroli:
allowed amount 0 on recipients for enable token

stlx-5021

### DIFF
--- a/modules/core/src/v2/coins/algo.ts
+++ b/modules/core/src/v2/coins/algo.ts
@@ -91,6 +91,9 @@ export interface VerifiedTransactionParameters {
 }
 
 export class Algo extends BaseCoin {
+  readonly ENABLE_TOKEN = 'enabletoken';
+  readonly DISABLE_TOKEN = 'disabletoken';
+
   constructor(bitgo: BitGo) {
     super(bitgo);
   }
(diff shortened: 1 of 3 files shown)


modules/core/src/v2/wallet.ts
```diff
@@ -2030,12 +2034,12 @@ export class Wallet {
...
       const coin = self.baseCoin;
       if (_.isObject(params.recipients)) {
-        params.recipients.map(function(recipient) {
+        params.recipients.map(function (recipient) {
           const amount = new BigNumber(recipient.amount);
           if (amount.isNegative()) {
             throw new Error('invalid argument for amount - positive number greater than zero or numeric string expected');
           }
-          if (!coin.valuelessTransferAllowed() && amount.isZero()) {
+          if (!coin.valuelessTransferAllowed() && amount.isZero() && this.typeKeys.includes(params.type)) {
```
pancham88, line 2042: A better way to implement this could be to pass in an option `type` param in valuelessTransferAllowed and implement it in `algo` so that it checks the param and return true or false.
With this, the logic will be encapsulated in coin implementations.

pancham88 (changes requested):

rliberal-simtlix (approved):

modules/core/src/v2/coins/algo.ts
```diff
@@ -121,7 +121,7 @@ export class Algo extends BaseCoin {
    */
   valuelessTransferAllowed(): boolean {
     // TODO: this sounds like its true with the staking txes - confirm before launch
-    return false;
+    return true;
```
pancham88, line 124: The valueless transfer is not always allowed, hence returning it as true is not correct.

It is allowed for certain transaction types (enableToken) and that is what we shall implement. Enhance this method to accept optional transaction type param and based on it return `true` or `false`

pancham88 (approved):

pancham88 (approved):

PR #1311, opened 2021-07-26
Add eip1559 params in send/sendMany

karankurbur:
Adds eip1559 params to send/sendMany routes. If we pass in the new params it will pass it /build and insert the new params as part of the halfSignedTx body to /send route.

Throws an error if a user passes both `gasPrice` and `eip1559` params

### DIFF
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -108,7 +108,8 @@ interface Recipient {
 
 interface SignFinalOptions {
   txPrebuild: {
-    gasPrice: string;
+    eip1559?: { maxPriorityFeePerGas: number; maxFeePerGas: number };
+    gasPrice?: string;
     gasLimit: string;
     recipients: Recipient[];
     halfSigned: {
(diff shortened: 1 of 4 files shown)


modules/core/src/v2/coins/eth.ts
```diff
@@ -97,6 +97,7 @@ interface Recipient {
 
 interface SignFinalOptions {
   txPrebuild: {
+    eip1559: { maxPriorityFeePerGas: number; maxFeePerGas: number };
```
ryan-bitgo, line 100: add question mark

ryan-bitgo: ?

modules/core/src/v2/wallet.ts
```diff
@@ -2049,9 +2053,15 @@ export class Wallet {
...
         'instant', 'memo', 'type', 'trustlines', 'transferId',
-        'stakingOptions'
+        'stakingOptions',
       ]);
       const finalTxParams = _.extend({}, halfSignedTransaction, selectParams);
+
+      console.log('/send params');
+      console.log(finalTxParams);
+
+      throw new Error('BAD');
```
ryan-bitgo, line 2063: :)

karankurbur:
modules/core/test/v2/unit/wallet.ts was not linted - look for `should error eip1559 and gasPrice are passed`

tylerlevine (approved):

PR #1313, opened 2021-07-28
fix(core): fix verify sign parameters for Algorand

fmeichtri-simtlix:
Wallet platform is returning addresses within the pubs array and this causes an error when we try to sign a tx. This fix should allow us to sign and send enable token transactions.

[STLX-5074](https://bitgoinc.atlassian.net/browse/STLX-5074)

### DIFF
--- a/modules/core/src/v2/coins/algo.ts
+++ b/modules/core/src/v2/coins/algo.ts
@@ -328,9 +328,13 @@ export class Algo extends BaseCoin {
       throw new Error('missing addressVersion parameter to sign transaction');
     }
 
-    const signers = params.txPrebuild.keys.map((key) =>
-      accountLib.Algo.algoUtils.publicKeyToAlgoAddress(accountLib.Algo.algoUtils.toUint8Array(key))
-    );
+    const signers = params.txPrebuild.keys.map((key) => {
+      // if we are receiving addresses do not try to convert them
+      if (!accountLib.Algo.algoUtils.isValidAddress(key)) {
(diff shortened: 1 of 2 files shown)


modules/core/test/v2/unit/coins/algo.ts
```diff
@@ -600,5 +600,41 @@ describe('ALGO:', function () {
       });
       signed.txHex.should.equal(AlgoResources.rawTx.transfer.multisig);
     });
+
+    it('should verify sign params if the key array contains addresses', function () {
```
rliberal-simtlix, line 604: Please add failure cases if you can.

modules/core/src/v2/coins/algo.ts
```diff
@@ -328,9 +328,13 @@ export class Algo extends BaseCoin {
       throw new Error('missing addressVersion parameter to sign transaction');
     }
 
-    const signers = params.txPrebuild.keys.map((key) =>
-      accountLib.Algo.algoUtils.publicKeyToAlgoAddress(accountLib.Algo.algoUtils.toUint8Array(key))
-    );
+    const signers = params.txPrebuild.keys.map((key) => {
+      // if we are receiving addresses do not try to convert them
+      if (!accountLib.Algo.algoUtils.isValidAddress(key)) {
+        return accountLib.Algo.algoUtils.publicKeyToAlgoAddress(accountLib.Algo.algoUtils.toUint8Array(key));
```
rliberal-simtlix, line 334: Add a comment here explaining why you use this method

pancham88 (approved):

rliberal-simtlix (approved):

PR #1314, opened 2021-07-28
feat(account-lib): Replaced the use of Fee to use FeeRate instead

smaria95:
Removed Flat Fee public setter and replaced Fee with Fee Rate

**Note**: Drone has some errors because the changes were done on Account-Lib but Core was not yet updated.

[STLX-5062](https://bitgoinc.atlassian.net/browse/STLX-5062)


rliberal-simtlix (approved):

pancham88 (approved):

rliberal-simtlix:
Won't fix

PR #1316, opened 2021-07-28
fix(core): request body not being included in HMAC

ryan-bitgo:
The request body for an authenticated request was always `undefined`because it wasn't wrapped in the `then` callback of the request. This fixes that.

ticket: BG-34460

### DIFF
--- a/modules/core/test/unit/bitgo.ts
+++ b/modules/core/test/unit/bitgo.ts
@@ -452,6 +452,29 @@ describe('BitGo Prototype Methods', function() {
       responseData.isValid.should.equal(true);
     }));
 
+    it('should include request body as part of the hmac', co(function* () {
+      const url = 'https://bitgo.fakeurl';
+      const body = { test: 'test' };
+
+      const fixedUnixTime = 1627374646;
+      const originalDateNow = Date.now;
+      Date.now = () => fixedUnixTime;
+
+      try {
+        nock(url)
+          .post('/', body)
+          .reply(201, undefined, {
+            hmac: '677e0c9a65ca384415945cb19b40ae38eaadfbce3ccce8c5d7bf37c1973b2553',
+            timestamp: String(fixedUnixTime),
+          });
+
+        const resp = (yield bitgo.post(url).send(body)) as any;
+        resp.req.headers['hmac'].should.equal('4425a4004ef2724add25b4dd019d21c66394653a049d82e37df3a2c356b5706d');
+      } finally {
+        Date.now = originalDateNow;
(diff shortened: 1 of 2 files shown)


modules/core/test/unit/bitgo.ts
```diff
@@ -452,6 +452,29 @@ describe('BitGo Prototype Methods', function() {
       responseData.isValid.should.equal(true);
     }));
 
+    it('should include request body as part of the hmac', co(function* () {
```
tylerlevine, line 455:
```suggestion
    it('should include request body as part of the hmac', async function () {
```

modules/core/test/unit/bitgo.ts
```diff
@@ -452,6 +452,29 @@ describe('BitGo Prototype Methods', function() {
...
+      const url = 'https://bitgo.fakeurl';
+      const body = { test: 'test' };
+
+      const fixedUnixTime = 1627374646;
+      const originalDateNow = Date.now;
```
tylerlevine, line 460: instead of overwriting properties of the `Date` global, `useFakeTimers` from `sinon` is preferred.

modules/core/test/unit/bitgo.ts
```diff
@@ -452,6 +452,29 @@ describe('BitGo Prototype Methods', function() {
...
+      Date.now = () => fixedUnixTime;
+
+      try {
+        nock(url)
```
tylerlevine, line 464: Instead of using `nock` to do this test, it would be better to use sinon to mock the `calculateRequestHeaders` method on the bitgo object and check the `text` property on the params object for the correct contents (namely, it should match the string representation of the `body` variable on 457 above)

modules/core/test/unit/bitgo.ts
```diff
@@ -452,6 +452,29 @@ describe('BitGo Prototype Methods', function() {
...
+          .post('/', body)
+          .reply(201, undefined, {
+            hmac: '677e0c9a65ca384415945cb19b40ae38eaadfbce3ccce8c5d7bf37c1973b2553',
+            timestamp: String(fixedUnixTime),
+          });
+
+        const resp = (yield bitgo.post(url).send(body)) as any;
```
tylerlevine, line 471: I think the cast to `any` here is unnecessary

tylerlevine (approved):
good catch on this.

Test is a bit old-school but the tests need a pass soon anyway (and I think we need more tests around the HMAC functionality in general), so I'm good with just merging this as is and we can come back and improve it later.

PR #1318, opened 2021-07-29
build(utxo-lib): allow typescript

OttoAllmendinger:
Add tsconfig.json. Does not actually contain any typescript code yet but
sets up the build process.

BG-7478


modules/utxo-lib/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "rootDir": ".",
+    "typeRoots": [
+      "./node_modules/@types",
+      "../../node_modules/@types"
+    ],
+    "declaration": false,
+    "declarationMap": false,
```
OttoAllmendinger, line 11: sadly at this stage there are build errors when these stay `true`

modules/utxo-lib/package.json
```diff
@@ -17,9 +18,10 @@
     "javascript"
   ],
   "scripts": {
+    "build": "tsc --project tsconfig.build.json",
     "coverage-report": "nyc report --reporter=lcov",
     "coverage-html": "nyc report --reporter=html",
-    "coverage": "BITGO_UTXO_LIB_TEST_EXPECTED_COUNT=3539 nyc --check-coverage --branches 90 --functions 90 mocha --recursive",
+    "coverage": "npm run build && BITGO_UTXO_LIB_TEST_EXPECTED_COUNT=3539 nyc --check-coverage --branches 89 --functions 90 mocha",
```
OttoAllmendinger, line 24: the coverage was reduced for spurious reasons (typescript boilerplate?)

modules/utxo-lib/test/integration_local_rpc/generate/fixtures.ts
```diff
@@ -1,7 +1,8 @@
 /**
  * @prettier
  */
-import * as fs from 'fs/promises';
+import * as fs from 'fs-extra';
+import { promisify } from 'util';
```
tylerlevine, line 5: nit: is this used?

modules/utxo-lib/package.json
tylerlevine: Do we need to build first? I think mocha is configured to run the tests with `ts-node`, so isn't this just extra work?

tylerlevine (changes requested):
great step in the right direction here, just a couple comments to resolve before I can approve

modules/utxo-lib/package.json
OttoAllmendinger: we need it because `nyc` uses the `main` entry point from `package.json` (and somehow maps it back even when run with `ts-node`)

modules/utxo-lib/test/integration_local_rpc/generate/fixtures.ts
OttoAllmendinger: nope

tylerlevine (approved):
build is flaky, I restarted it

PR #1319, opened 2021-07-29
WIP(account-lib):  upgrade celo to 1.2.4

fdmirza:
The upgraded celo library pulls in a newer version of the ethereum js library, which has deprecated/moved some utility functions.  To work around that, I have added the old ethereum library as [ethereumjs-utils-old](https://github.com/BitGo/BitGoJS/pull/1319/files#diff-bb31881d2a7899dc1d14e23f052c93e20dbe09b4408bde5a98b1549a55a9dc8aR55), and updated all calls to its functions (`stripHexPrefix`, etc.) in the code.  The alternative was to make a much larger change to all the other coins that depended on functions imported from it.


modules/account-lib/package.json
```diff
@@ -52,6 +52,7 @@
     "elliptic": "^6.5.2",
     "ethereumjs-abi": "^0.6.5",
     "ethereumjs-util": "5.2.0",
+    "ethereumjs-utils-old": "npm:ethereumjs-utils@5.2.0",
```
tylerlevine, line 55: why is one package `ethereumjs-utils` (plural) and the original one is `ethereumjs-util` (singular)?

Seems to me like these are totally separate packages:

* https://www.npmjs.com/package/ethereumjs-util
* https://www.npmjs.com/package/ethereumjs-utils

tylerlevine (changes requested):
neat hack, but please make sure we aren't using some random unsupported fork of ethereumjs-util

modules/account-lib/package.json
fdmirza: Nice catch. I will update

tylerlevine:
done in #1323

PR #1322, opened 2021-07-29
feat(account-lib): add method generateAccount() in account-lib

JoaquinTissera-Simtlix:
add method generateAccount() in account-lib

STLX-4522


modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -396,6 +396,16 @@ export class Utils implements BaseUtils {
       addrs,
     });
   }
+
+  /**
+   * generateAccount returns a new Algorand address and its corresponding secret key
+   *
+   * @param null
```
rliberal-simtlix, line 403: Please remove param and null here. Function has not params

rliberal-simtlix (changes requested):

rliberal-simtlix (approved):

pancham88 (approved):

PR #1323, opened 2021-07-29
feat(account-lib): upgrade celo to 1.2.4

tylerlevine:
Ticket: BG-27462


tylerlevine:
we need this fix urgently so this will replace #1319

ryan-bitgo (approved):
LGTM

PR #1327, opened 2021-07-30
fix(core): ignore typescript errors from incompatible `@types/ethereumjs-util`

tylerlevine:
After upgrading the `@celo/contractkit` package, a new version of the
`@types/ethereumjs-util` package was also installed. However, since we're using
an old version of `ethereumjs-util` from before it had built-in types support,
we were forced to start using the incompatible `@types` package.

To work around this, we added an alias for the old version of `ethereumjs-util`
in `account-lib` called `ethereumjs-utils-old`. This stopped typescript from
using the incompatible `@types/ethereumjs-util` types since the package names
no longer match. However, since that change was done only in `account-lib`,
that alias package is only resolvable within the `account-lib` source and
tests.

This caused an issue when code in the `core` library was modified to use the
`ethereumjs-utils-old` package. Since this is resolvable when running tests,
the tests passed, but when actually using the package in downstream software
the `ethereumjs-utils-old` package is not installed and therefore can't be
resolved.

Ticket: BG-27462

### DIFF
--- a/modules/core/src/v2/coins/eth2.ts
+++ b/modules/core/src/v2/coins/eth2.ts
@@ -3,7 +3,7 @@
  */
 import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
-import { stripHexPrefix } from 'ethereumjs-utils-old';
+import * as ethUtil from 'ethereumjs-util';
 import * as request from 'superagent';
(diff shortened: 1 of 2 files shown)


sangaman (approved):
Nice explanation.

PR #1331, opened 2021-08-02
refactor(core): use secp256k1 for ECDH

OttoAllmendinger:
We have two separate places where we use ECDH: token issuance and wallet sharing.

They are mutually incompatible but internally compatible, which is what counts.

Issue:  BG-34381

### DIFF
--- a/modules/core/test/unit/bitgo.ts
+++ b/modules/core/test/unit/bitgo.ts
@@ -2,6 +2,7 @@
 // Tests for BitGo Object
 //
 
+import * as crypto from 'crypto';
 import * as should from 'should';
 import * as nock from 'nock';
 import * as Bluebird from 'bluebird';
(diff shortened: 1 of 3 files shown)


modules/core/test/unit/bitgo.ts
```diff
@@ -279,13 +280,30 @@ describe('BitGo Prototype Methods', function() {
...
+
     it('should calculate a new ECDH sharing secret correctly', () => {
-      const bitgo = new TestBitGo();
-      const eckey1 = bitcoin.ECPair.makeRandom({ network: bitcoin.networks[common.getNetwork()] });
-      const eckey2 = bitcoin.ECPair.makeRandom({ network: bitcoin.networks[common.getNetwork()] });
-      const sharingKey1 = bitgo.getECDHSecret({ eckey: eckey1, otherPubKeyHex: eckey2.getPublicKeyBuffer().toString('hex') });
-      const sharingKey2 = bitgo.getECDHSecret({ eckey: eckey2, otherPubKeyHex: eckey1.getPublicKeyBuffer().toString('hex') });
-      sharingKey1.should.equal(sharingKey2);
+      for (let i=0; i<256; i++) {
+        const bitgo = new TestBitGo();
```
tylerlevine, line 291: do we need to instantiate 256 `TestBitGo` objects?

modules/core/test/unit/bitgo.ts
```diff
@@ -279,13 +280,30 @@ describe('BitGo Prototype Methods', function() {
...
+        const eckey1 = getKey(`${i}.a`);
+        const eckey2 = getKey(`${i}.b`);
+        const sharingKey1 = bitgo.getECDHSecret({ eckey: eckey1, otherPubKeyHex: eckey2.getPublicKeyBuffer().toString('hex') });
+        const sharingKey2 = bitgo.getECDHSecret({ eckey: eckey2, otherPubKeyHex: eckey1.getPublicKeyBuffer().toString('hex') });
+        sharingKey1.should.equal(sharingKey2);
+
+        switch (i) {
```
tylerlevine, line 298: Why assert the first two keys? Do these match the prior implementation (if not, pending shares would need to be invalidated)

tylerlevine: Ok, I reviewed the ordering of the commits and since the test was modified first, I'm assuming the new ECDH implementation is entirely compatible with the old one. Approving based on that assumption

tylerlevine (approved):

modules/core/test/unit/bitgo.ts
OttoAllmendinger: I was just making sure here that the result is actually different every round. 

(It would be a nice feature if CI actually for every commit, not just the last one)

modules/core/test/unit/bitgo.ts
OttoAllmendinger: we don't - actually `bitgo.getECDHSecret` should at least be a static class method, probably just a regular method instead

PR #1332, opened 2021-08-02
feat(algo): misc updates for platform migration to use account lib

pancham88:
[BG-34606](https://bitgoinc.atlassian.net/browse/BG-34606)

Updates to support migration of following transaction types:

1. Native transfers
2. Enable/Disable token
3. Token transfers
4. Keyreg transactions
5. Non-participating keyreg transactiosn

### DIFF
--- a/modules/account-lib/src/coin/algo/utils.ts
+++ b/modules/account-lib/src/coin/algo/utils.ts
@@ -137,7 +137,17 @@ export class Utils implements BaseUtils {
    * @returns {EncodedTx} The decoded transaction.
    */
   decodeAlgoTxn(txnBytes: Uint8Array | string): EncodedTx {
-    const buffer = typeof txnBytes === 'string' ? Buffer.from(txnBytes, 'hex') : txnBytes;
+    let buffer;
+    if (typeof txnBytes === 'string') {
+      if (allHexChars(txnBytes)) {
+        buffer = Buffer.from(txnBytes, 'hex');
+      } else {
+        buffer = Buffer.from(txnBytes, 'base64');
+      }
+    } else {
+      buffer = txnBytes;
+    }
+
     if (this.isDecodableUnsignedAlgoTxn(buffer)) {
       return {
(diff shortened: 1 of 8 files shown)


modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -137,7 +137,17 @@ export class Utils implements BaseUtils {
    * @returns {EncodedTx} The decoded transaction.
    */
   decodeAlgoTxn(txnBytes: Uint8Array | string): EncodedTx {
-    const buffer = typeof txnBytes === 'string' ? Buffer.from(txnBytes, 'hex') : txnBytes;
+    let buffer;
+    if (typeof txnBytes === 'string') {
+      if (allHexChars(txnBytes)) {
+        buffer = Buffer.from(txnBytes, 'hex');
```
rliberal-simtlix, line 143: In which cases tx should be hex from platform?

pancham88: From platform, it is always 'base64'. 
However, this method can support both 'hex' and 'base64'. The current tests in BitGoJS use 'hex' format and without support for both - BitGoJS unit tests fail.
We can update the data in resources to use base64 instead of hex and that can be a separate task. In either case, it is good to have this method support both formats

modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -137,7 +137,17 @@ export class Utils implements BaseUtils {
...
+      } else {
+        buffer = Buffer.from(txnBytes, 'base64')
```
rliberal-simtlix, line 145: Add semicolon here.

rliberal-simtlix (approved):

modules/account-lib/src/coin/algo/utils.ts
pancham88: Thanks for pointing, added it.

PR #1341, opened 2021-08-03
feat(account-lib): AVAX KeyPair Support

smaria95:
Implemented KeyPair class and unit tests for AVAX.

[STLX-5078](https://bitgoinc.atlassian.net/browse/STLX-5078).


jrafael-simtlix (approved):

modules/account-lib/test/unit/coin/avax/keyPair.ts
```diff
@@ -0,0 +1,167 @@
...
+
+describe('Avax Key Pair', () => {
+  describe('should create a valid KeyPair', () => {
+    it('from an empty value', () => {
+      const keyPairObj = new KeyPair();
+      const keys = keyPairObj.getKeys();
+      should.exists(keys.prv);
+      should.exists(keys.pub);
+      should.equal(keys.prv!.length, 64);
+      should.equal(keys.pub.length, 130);
```
CapnMigraine, line 13: I'm a little confused here, I used avalanche-sdk's `getPublicKeyString()` on a key pair, and I got a public key with 66 chars (`8jAVW8gn6WaARg4PUPdbpnenqNYBMtUWSzvoh2uVgU91T1LfdL`)

CapnMigraine (changes requested):
I think it would be safer to call the coin `avaxc` even though there's only one account-based Avalanche chain.
We'll use `avaxc` in wallet-platform and other parts of the system, so it makes sense to keep consistency to avoid the need for special conditions for AVAX (e.g. [creating instance of  txBuilder](https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/v2/coins/abstractEthLikeCoin.js#L268))

@BitGo/eth-team please help review.

jrafael-simtlix:
The coin name is AVAXC. Rename the folder, please.

modules/account-lib/test/unit/coin/avax/keyPair.ts
smaria95: The reason for that is that `avalanche-sdk ` generates keys in Avax Native Format. i.e, P|X|C - <66 base58 encoded string>
However, that C-<key> is only valid when importing and exporting balance between chains. 
When we use the C-Chain it works as Ethereum EVM so the accepted format for keys is the same as ethereum (130 uncompressed public key)

I originally had include methods to convert from EVM to AVM keys formats, but avalanche-sdk requires Node >= v14.17
So I removed that since we still need to keep giving support to Node 12 the next 7 months I think.

smaria95:
@jrafael-simtlix , @CapnMigraine 
I've renamed the coin folder to `avaxc`

tylerlevine:
> @jrafael-simtlix , @CapnMigraine
> I've renamed the coin folder to `avaxc`

please rename the test folder as well

modules/account-lib/test/unit/coin/avax/keyPair.ts
CapnMigraine: I see. `getKeys()` returns pub in compressed format when the keypair is created from an `xprv`, but it returns an uncompressed pub if a `prv` is used.
@BitGo/eth-team is that behavior intentional? @smaria95's change makes it consistently return pubs in uncompressed format. Would that cause an issue when we create a controller in wp that extends from `AbstractEthLikeCoin`?

jrafael-simtlix (dismissed):

wesgraham (approved):
LGTM

modules/account-lib/test/unit/coin/avax/keyPair.ts
CapnMigraine: @smaria95 and I looked for call sites to this logic from SDK, Wallet-platform and OVC. Nothing appears to be affected by this change. 
At some point we could change the ETH KeyPaid implementation to match this one for consistency.

CapnMigraine (approved):
lgtm

PR #1342, opened 2021-08-03
chore(core): deprecate hdPath in transactionBuilder

OttoAllmendinger:
The `hdPath()` wrapper is not needed any longer.

Issue: BG-34381

99ed890a (Otto Allmendinger, 6 minutes ago)
fix(core): transactionBuilder: ignore `walletSubPath === 'm'`

The `m` component is only allowed at the very beginning, but some of our
tests have it in the middle of a path.

Sadly our buggy `deriveKey` method allows it anywhere (and simply ignores
it).

Issue: BG-34381

### DIFF
--- a/modules/core/src/transactionBuilder.ts
+++ b/modules/core/src/transactionBuilder.ts
@@ -921,8 +921,11 @@ exports.signTransaction = function (params) {
 
     const chainPath = currentUnspent.chainPath;
     if (rootExtKeyPath) {
-      const subPath = keychain.walletSubPath || '/0/0';
-      const path = keychain.path + subPath + chainPath;
+      let { walletSubPath = '/0/0' } = keychain;
+      if (walletSubPath === 'm') {
+        walletSubPath = ''
+      }
+      const path = keychain.path + walletSubPath + chainPath;
       privKey = rootExtKeyPath.deriveKey(path);
     }
 
--- a/modules/core/src/transactionBuilder.ts
+++ b/modules/core/src/transactionBuilder.ts
@@ -898,7 +898,6 @@ exports.signTransaction = function (params) {
   let rootExtKey;
   if (keychain) {
     rootExtKey = bitcoin.HDNode.fromBase58(keychain.xprv);
-    rootExtKeyPath = hdPath(rootExtKey);
   }
 
   const txb = bitcoin.TransactionBuilder.fromTransaction(transaction, network);
@@ -920,13 +919,13 @@ exports.signTransaction = function (params) {
     }
 
     const chainPath = currentUnspent.chainPath;
-    if (rootExtKeyPath) {
+    if (rootExtKey) {
       let { walletSubPath = '/0/0' } = keychain;
       if (walletSubPath === 'm') {
         walletSubPath = ''
       }
       const path = keychain.path + walletSubPath + chainPath;
-      privKey = rootExtKeyPath.deriveKey(path);
+      privKey = rootExtKey.derivePath(path).keyPair;
     }
 
     privKey.network = network;
--- a/modules/core/src/transactionBuilder.ts
+++ b/modules/core/src/transactionBuilder.ts
@@ -19,6 +19,7 @@ import { getNetwork, hdPath } from './bitcoin';
 import debugLib = require('debug');
 const debug = debugLib('bitgo:v1:txb');
 import * as common from './common';
+import { sanitizeLegacyPath } from "./bip32path";
 
 interface BaseOutput {
   amount: number;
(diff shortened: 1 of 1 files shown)


modules/core/src/transactionBuilder.ts
```diff
@@ -898,7 +898,6 @@ exports.signTransaction = function(params) {
   let rootExtKey;
   if (keychain) {
     rootExtKey = bitcoin.HDNode.fromBase58(keychain.xprv);
-    rootExtKeyPath = hdPath(rootExtKey);
```
sangaman, line 901: You can remove the `let rootExtKeyPath;` declaration a couple of lines up because it's no longer being used.

modules/core/src/transactionBuilder.ts
```diff
@@ -920,10 +919,13 @@ exports.signTransaction = function(params) {
     }
 
     const chainPath = currentUnspent.chainPath;
-    if (rootExtKeyPath) {
-      const subPath = keychain.walletSubPath || '/0/0';
-      const path = keychain.path + subPath + chainPath;
-      privKey = rootExtKeyPath.deriveKey(path);
+    if (rootExtKey) {
+      let { walletSubPath = '/0/0' } = keychain;
+      if (walletSubPath === 'm') {
```
sangaman, line 924: Just to be clear, `m` will only be here as the only character and not with anything following it, right?

OttoAllmendinger: yes, although on second though we should probably be a bit more flexible

OttoAllmendinger: let's do https://github.com/BitGo/BitGoJS/pull/1347 first

tylerlevine (dismissed):

modules/core/package.json
```diff
@@ -75,7 +75,7 @@
     "sanitize-html": "^1.27.5",
     "secp256k1": "^4.0.2",
     "secrets.js-grempe": "^1.1.0",
-    "stellar-sdk": "^8.1.0",
+    "stellar-sdk": "^8.2.5",
```
OttoAllmendinger, line 78: @CapnMigraine @argjv

modules/core/src/transactionBuilder.ts
sangaman: Looks like this unused var is still there.

modules/core/src/transactionBuilder.ts
```diff
@@ -920,10 +920,10 @@ exports.signTransaction = function(params) {
     }
 
     const chainPath = currentUnspent.chainPath;
-    if (rootExtKeyPath) {
-      const subPath = keychain.walletSubPath || '/0/0';
-      const path = keychain.path + subPath + chainPath;
-      privKey = rootExtKeyPath.deriveKey(path);
+    if (rootExtKey) {
+      let { walletSubPath = '/0/0' } = keychain;
```
sangaman, line 924: This can go back to a `const` now I believe.

sangaman (approved):
Much nicer with the `sanitizeLegacyPath` function, I see just a couple of cleanup items.

OttoAllmendinger:
it seems like we are missing a eslint step

modules/core/src/transactionBuilder.ts
OttoAllmendinger: yes

modules/core/src/transactionBuilder.ts
OttoAllmendinger: fixed

sangaman (dismissed):
LGTM. Not sure why the build is failing though.

modules/core/src/transactionBuilder.ts
```diff
@@ -19,6 +19,7 @@ import { getNetwork, hdPath } from './bitcoin';
 import debugLib = require('debug');
 const debug = debugLib('bitgo:v1:txb');
 import * as common from './common';
+import { sanitizeLegacyPath } from "./bip32path";
```
tylerlevine, line 22: single quotes

modules/core/package.json
tylerlevine: is there a fix for the drone issue in the newer version, or is this purely speculative?

tylerlevine (approved):

modules/core/src/transactionBuilder.ts
OttoAllmendinger: I'll add prettier in a follow-up

modules/core/package.json
OttoAllmendinger: this is me just guessing

tylerlevine: for posterity, the flaky CI failures should indeed be fixed by upgrading: https://github.com/stellar/js-stellar-sdk/issues/645

sangaman (approved):

PR #1343, opened 2021-08-03
Stlx 4593 update verification methods again

mam1982:


JulioAchaval-SimTLiX (changes requested):
Needs to rebase again.

PR #1346, opened 2021-08-04
feat(account-lib): skeleton code for avalanche c-chain in account-lib

shiksb:
Ticket: STLX-3250

### DIFF
--- a/modules/account-lib/src/coin/avaxc/keyPair.ts
+++ b/modules/account-lib/src/coin/avaxc/keyPair.ts
@@ -1,0 +1,41 @@
+import { KeyPairOptions, DefaultKeys, isPrivateKey, isPublicKey } from '../baseCoin/iface';
+import { NotImplementedError } from '../baseCoin/errors';
+import { Secp256k1ExtendedKeyPair } from '../baseCoin/secp256k1ExtendedKeyPair';
+
+export class KeyPair extends Secp256k1ExtendedKeyPair {
+  /**
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   * @param { KeyPairOptions } source Either a master seed, a private key, or a public key
+   */
+  constructor(source?: KeyPairOptions) {
+    super(source);
+    if (source) {
+      if (isPrivateKey(source)) {
+        super.recordKeysFromPrivateKey(source.prv);
+      } else if (isPublicKey(source)) {
+        super.recordKeysFromPublicKey(source.pub);
+      } else {
+        throw new Error('Invalid key pair options');
+      }
+    } else {
+      throw new Error('Invalid key pair options');
+    }
+  }
+
+  /**
+   * Avalanche default keys format is CB58 strings with prefixes
+   *
+   * @returns { DefaultKeys } The keys in the protocol default key format
+   */
+  getKeys(): DefaultKeys {
+    throw new NotImplementedError('getKeys not implemented');
+  }
+
+  /**
+   * Avalanche C-chain address is an Eth-like hex value starting with 0x prefix
+   */
+  getAddress(): string {
+    throw new NotImplementedError('getAddress not implemented');
+  }
+}
--- a/modules/account-lib/src/coin/avaxc/transaction.ts
+++ b/modules/account-lib/src/coin/avaxc/transaction.ts
@@ -1,0 +1,69 @@
+import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
+import { BaseTransaction, TransactionType } from '../baseCoin';
+import { BaseKey } from '../baseCoin/iface';
+import { KeyPair } from './keyPair';
+import { NotImplementedError } from '../baseCoin/errors';
+
+export class Transaction extends BaseTransaction {
+  protected _type: TransactionType;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  /** @inheritdoc */
+  canSign(key: BaseKey): boolean {
+    return true;
+  }
+
+  async sign(keyPair: KeyPair): Promise<void> {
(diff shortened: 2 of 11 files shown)


OttoAllmendinger:
please use `feat(account-lib)` so that we can easier identify the component

modules/account-lib/src/coin/avaxc/keyPair.ts
```diff
@@ -0,0 +1,41 @@
...
+      if (isPrivateKey(source)) {
+        super.recordKeysFromPrivateKey(source.prv);
+      } else if (isPublicKey(source)) {
+        super.recordKeysFromPublicKey(source.pub);
+      } else {
+        throw new Error('Invalid key pair options');
+      }
+    } else {
+      throw new Error('Invalid key pair options');
+    }
```
smaria95, line 23: I think the constructor `implementation` and the `jsdoc` are not consistent within each other.

The doc says: `By default, creates a key pair with a random master seed`.
While the constructor throws an error if it's called without an argument.

I think the correct behavior is the one detailed on the jsdoc, but since this class will be implemented as part of [STLX-5078](https://bitgoinc.atlassian.net/browse/STLX-5078) maybe we can just add a comment like: `TODO(STLX-5078): Implement random seed and key pair generation`.

modules/account-lib/src/coin/avaxc/transaction.ts
```diff
@@ -0,0 +1,69 @@
...
+    throw new NotImplementedError('toJson not implemented');
+  }
+
+  /**
+   * Set the transaction type
+   *
+   * @param {TransactionType} transactionType The transaction type to be set
+   */
+  setTransactionType(transactionType: TransactionType): void {
+    throw new NotImplementedError('getTransferData not implemented');
```
smaria95, line 48:
```suggestion
    throw new NotImplementedError('setTransactionType not implemented');
```

modules/account-lib/src/coin/avaxc/transaction.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Decode previous signatures from the inner transaction
+   * and save them into the base transaction signature list.
+   */
+  loadPreviousSignatures(): void {
+    throw new NotImplementedError('getTransferData not implemented');
```
smaria95, line 56:
```suggestion
    throw new NotImplementedError('loadPreviousSignatures not implemented');
```

modules/account-lib/src/coin/avaxc/transactionBuilder.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Set an external transaction signature
+   *
+   * @param { string }signature Hex encoded signature string
+   * @param { KeyPair }keyPair The public key keypair that was used to create the signature
+   * @returns { TransactionBuilder }This transaction builder
+   */
+  signature(signature: string, keyPair: KeyPair): TransactionBuilder {
+    throw new NotImplementedError('initializeBuilder not implemented');
```
smaria95, line 83:
```suggestion
    throw new NotImplementedError('signature not implemented');
```

modules/account-lib/src/coin/avaxc/transactionBuilderFactory.ts
```diff
@@ -0,0 +1,59 @@
...
+    throw new NotImplementedError('parseTransaction not implemented');
+  }
+
+  /**
+   * Check the raw transaction has a valid format in the blockchain context, throw otherwise.
+   *
+   * @param {any} rawTransaction - Transaction in any format
+   */
+  private validateRawTransaction(rawTransaction: any) {
+    throw new NotImplementedError('parseTransaction not implemented');
```
smaria95, line 57:
```suggestion
    throw new NotImplementedError('validateRawTransaction not implemented');
```

modules/account-lib/src/coin/avaxc/walletInitializationBuilder.ts
```diff
@@ -0,0 +1,60 @@
...
+    throw new NotImplementedError('initBuilder not implemented');
+  }
+
+  // endregion
+
+  // region Common builder methods
+  /**
+   * Set one of the owners of the multisig wallet.
+   *
+   * @param {string} address The public key of the owner's account
```
smaria95, line 30: I think the owners will be set as addresses and not as public keys, because that is what the contract expects to.
```suggestion
   * @param {string} address The address of the owner's account
```

smaria95 (changes requested):

modules/account-lib/src/coin/avaxc/transaction.ts
```diff
@@ -0,0 +1,69 @@
+import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
```
CapnMigraine, line 1: I think we could re-use `eth`'s implementation of `Transaction`, like we did with [etc](https://github.com/BitGo/BitGoJS/blob/master/modules/account-lib/src/coin/etc/index.ts#L2)

CapnMigraine (changes requested):
Please update the `@bitgo/statics` dependency in account-lib to `6.12.0-rc.2` to access the avaxc definition.
We need to add the coin to the exports and the `coinBuilderMap` in `index.ts`
We also need an index.ts in the [coin folder](https://github.com/BitGo/BitGoJS/blob/master/modules/account-lib/src/coin/etc/index.ts) for the exports.
The classes defined for Avaxc will likely be used by the eth-like coin controller in wallet-platform, we should try to re-use as much as the `eth` implementation in account-lib as possible, so we don't have to create a different path for avaxc in wallet-platform unless we have to.

shiksb:
I forked this repo instead of cloning. Addressed all your comments and created a new PR for the skeleton code: https://github.com/BitGo/BitGoJS/pull/1375

PR #1349, opened 2021-08-04
fix(core): address verification should fail for uppercase bech32 addresses

0xJacobV:
1. If an upper case bech32 address is provided, the isValidAddress function call should return false.
2. Apply es lint fixes to some files
3. Replaced mocha.opts with .mocharc.yml to remove deprecation warning

Issue: [BG-21499](https://bitgoinc.atlassian.net/browse/BG-21499)


modules/core/test/v2/unit/coins/ltc.ts
```diff
@@ -63,21 +63,35 @@ describe('LTC:', function() {
...
       // but mixed case is always invalid. Canonical bech32 addresses should be all lower case however
       const bech32Address = 'LTC1QGRL8ZPNDSKLAA9SWGD5VEVYXMX5X63VCRL7DK4';
-      const newAddress = ltc.canonicalAddress(bech32Address);
-      newAddress.should.equal(bech32Address.toLowerCase());
+      try {
+        ltc.canonicalAddress(bech32Address);
+        should.fail(null, null, 'Upper case bech32 addresses are not valid');
+      } catch (e) {
+        e.message.should.equal('invalid address');
+      }
```
tylerlevine, line 75: This is a more idiomatic way to make sure a function throws a given error string.

```suggestion
      (() => ltc.canonicalAddress(bech32Address)).should.throw('invalid address');
```

modules/core/test/v2/unit/coins/ltc.ts
```diff
@@ -96,9 +110,9 @@ describe('LTC:', function() {
       ltc.isValidAddress('ltc1qq7fzt3ek5ege3v92wh0q6wzcjr39pqswlpe36mu28f6yufark3wspfryg7').should.be.true();
       tltc.isValidAddress('tltc1qq7fzt3ek5ege3v92wh0q6wzcjr39pqswlpe36mu28f6yufark3ws2x86ht').should.be.true();
 
-      // all upper case is also valid
-      ltc.isValidAddress('LTC1QQ7FZT3EK5EGE3V92WH0Q6WZCJR39PQSWLPE36MU28F6YUFARK3WSPFRYG7').should.be.true();
-      tltc.isValidAddress('TLTC1QQ7FZT3EK5EGE3V92WH0Q6WZCJR39PQSWLPE36MU28F6YUFARK3WS2X86HT').should.be.true();
+      // all upper case is invalid
+      ltc.isValidAddress('LTC1QQ7FZT3EK5EGE3V92WH0Q6WZCJR39PQSWLPE36MU28F6YUFARK3WSPFRYG7').should.be.false();
+      tltc.isValidAddress('TLTC1QQ7FZT3EK5EGE3V92WH0Q6WZCJR39PQSWLPE36MU28F6YUFARK3WS2X86HT').should.be.false();
```
tylerlevine, line 115: Let's flip these back to not break existing behavior (see top level review comment)

tylerlevine (changes requested):
I checked [BIP 173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki) which actually does cover the reasoning behind lowercase/uppercase bech32 addresses:

> The lowercase form is used when determining a character's value for checksum purposes.
\
Encoders MUST always output an all lowercase Bech32 string. If an uppercase version of the encoding result is desired, (e.g.- for presentation purposes, or QR code use), then an uppercasing procedure can be performed external to the encoding process.
\
Decoders MUST NOT accept strings where some characters are uppercase and some are lowercase (such strings are referred to as mixed case strings).
\
For presentation, lowercase is usually preferable, but inside QR codes uppercase SHOULD be used, as those permit the use of alphanumeric mode, which is 45% more compact than the normal byte mode. 

I think we should stick to the spec and continue to accept uppercase bech32 addresses as valid.

The issue that's causing problems for customers is that when we go to validate a transaction prebuild, we deserialize the transaction hex and look at the addresses corresponding to each output to see if they match the ones provided in the transaction build request.

That verification comparison happens inside a function called `verifyAddress` [here](https://github.com/BitGo/BitGoJS/blob/c8e9c566310b9f31cc43380a42b283d801d15b3f/modules/core/src/v2/coins/abstractUtxoCoin.ts#L924).

What I think we really ought to do is modify `verifyAddress` to check to see if the `addressType` is `p2wsh`, and if so, convert the address to lowercase before comparing to the expected address. Remember though, mixed case addresses should still be invalid according to the spec, so please make sure you include a test case for that.

The rest of the cleanup looks pretty good to me. Next time though, a separate PR for something tangental like the `mocha.opts` cleanup would be appreciated. Whitespace/formatting/comma/import fixes are acceptable as drive-by fixes, but please try to limit it to nothing more involved than that.

tylerlevine:
Please see updated comment - your direction is correct but please see other review comments. You'll also need to rebase on https://github.com/BitGo/BitGoJS/pull/1353 once that gets merged to fix the build issues

0xJacobV:
@tylerlevine I made updates to the test description to reflect non-lowercase bech32 addresses. I don't think the implementation needs to change to account for it.

modules/core/test/v2/unit/coins/ltc.ts
0xJacobV: Suggestion applied

modules/core/test/v2/unit/coins/ltc.ts
```diff
@@ -63,21 +63,24 @@ describe('LTC:', function() {
...
     });
 
-    it('upper case bech32 mainnet address', function() {
-      // bech32 addresses which are all upper case or all lower case are potentially valid,
-      // but mixed case is always invalid. Canonical bech32 addresses should be all lower case however
-      const bech32Address = 'LTC1QGRL8ZPNDSKLAA9SWGD5VEVYXMX5X63VCRL7DK4';
-      const newAddress = ltc.canonicalAddress(bech32Address);
-      newAddress.should.equal(bech32Address.toLowerCase());
+    it('non-lowercase bech32 mainnet address should fail', function () {
+      // bech32 addresses which are non-lowercase are invalid,
```
sangaman, line 67: Might make sense to put some detail about how we only accept "canonical" lower case bech32 addresses, even though they are "valid" according to the spec, akin to the old comment. I know that if I came across this line without any context I'd be confused since I know upper case bech32 addresses to be valid according to the spec.

modules/core/test/v2/unit/coins/ltc.ts
```diff
@@ -63,21 +63,24 @@ describe('LTC:', function() {
...
-    it('bech32 testnet address', function() {
+    it('lower case bech32 testnet address', function () {
       const bech32Address = 'tltc1qu78xur5xnq6fjy83amy0qcjfau8m367defyhms';
       const version1Address = tltc.canonicalAddress(bech32Address, 1);
       version1Address.should.equal(bech32Address);
       const version2Address = tltc.canonicalAddress(bech32Address, 2);
       version2Address.should.equal(bech32Address);
     });
+
+    it('non-lowercase bech32 testnet address should fail', function () {
```
sangaman, line 80: "upper case" rather than "non-lowercase"? Since we're explicitly testing upper case here, whereas non-lower could be mixed case as well (which is always invalid, maybe should be its own separate test case even).

sangaman (dismissed):
LGTM, just a couple thoughts on comments/test naming. And I agree that it'd be helpful to put the style changes in a separate PR (tho it's good they're at least separate commits for history purposes) just to make reviewing easier.

modules/core/test/v2/unit/coins/ltc.ts
0xJacobV: @sangaman good idea. test description has been updated

modules/core/test/v2/unit/coins/ltc.ts
0xJacobV: added more test cases to cover lowercase, uppercase and mixed case scenarios

OttoAllmendinger (approved):
please get a review from @CapnMigraine or @argjv for the stellar bump

tylerlevine (dismissed):
nice work

OttoAllmendinger (approved):

tylerlevine (approved):

PR #1351, opened 2021-08-04
fix(eth): goerli coins now set to gteth in core/src/config.ts

bjon-bitgo:
Before, we assumed erc20 tokens are either mainnet or kovan (If it is not eth, then it is teth).
erc20 tokens now can be eth, teth, and gteth

More information can be found on this related PR on the WP: https://github.com/BitGo/BitGoJS/pull/1351

Ticket: BG-34419


tylerlevine (changes requested):
test please

wesgraham (dismissed):

tylerlevine (dismissed):

tylerlevine (approved):

PR #1356, opened 2021-08-05
fix(core): fix method name to TRX.xpubToUncompressedPub

OttoAllmendinger:
Make method name reflect the actual implementation

Issue: BG-34381

### DIFF
--- a/modules/core/src/v2/coins/trx.ts
+++ b/modules/core/src/v2/coins/trx.ts
@@ -300,16 +300,16 @@ export class Trx extends BaseCoin {
   }
 
   /**
-   * Converts an xpub to a compressed pub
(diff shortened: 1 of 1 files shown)


tylerlevine (approved):
[`xprvToCompressedPrv`](https://github.com/BitGo/BitGoJS/pull/1356/files#diff-5152d2c80f4a8b7a20bd9d0f74cf4245a321a4225d0c4dd68e31c5cfc3ab1175R341): what's a compressed prv 🤔. I've never seen one of these. Have we been wasting bytes on all these uncompressed scalar values the whole time?!?

/s

good clean up

OttoAllmendinger:
must be some very advanced technology

PR #1357, opened 2021-08-05
refactor(core): move utxo recover method to separate file

OttoAllmendinger:
The class method implementation was using deprecated coroutine wrappers and
was way too long.

Issue: BG-34381

5e89226e (Otto Allmendinger, 36 minutes ago)
chore(core): apply prettier to abstractUtxoCoin

Issue: BG-34381

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1,3 +1,6 @@
+/**
+ * @prettier
+ */
 import { Codes, VirtualSizes } from '@bitgo/unspents';
 import { UnspentType } from '@bitgo/unspents/dist/codes';
 import * as bitcoin from '@bitgo/utxo-lib';
@@ -60,27 +63,27 @@ export interface Output {
 export interface TransactionFee {
   fee: number;
   feeRate?: number;
-  size: number
+  size: number;
 }
 
 export interface TransactionExplanation {
   displayOrder: string[];
   id: string;
-  outputs: Output[],
-  changeOutputs: Output[],
+  outputs: Output[];
+  changeOutputs: Output[];
   outputAmount: string;
   changeAmount: number;
   fee: TransactionFee;
 }
 
 export interface Unspent {
-  id: string,
-  value: string,
+  id: string;
+  value: string;
 }
 
 export interface ExplainTransactionOptions {
   txHex: string;
-  txInfo?: { changeAddresses: string[], unspents: Unspent[] };
+  txInfo?: { changeAddresses: string[]; unspents: Unspent[] };
   feeInfo?: string;
 }
 
@@ -108,7 +111,6 @@ export interface TransactionParams extends BaseTransactionParams {
   walletPassphrase?: string;
   changeAddress?: string;
 }
-
 
 export interface ParseTransactionOptions extends BaseParseTransactionOptions {
   txParams: TransactionParams;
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * @prettier
  */
-import { Codes, VirtualSizes } from '@bitgo/unspents';
+import { Codes } from '@bitgo/unspents';
 import { UnspentType } from '@bitgo/unspents/dist/codes';
```
tylerlevine, line 5: Hm, do we need to pull this out of `dist/codes`? Is it not exported? (Optional nit since this code is unchanged)

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1788,8 +1778,8 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     return response;
   }
 
-  protected abstract getAddressInfoFromExplorer(address: string, apiKey?: string): Bluebird<AddressInfo>;
-  protected abstract getUnspentInfoFromExplorer(address: string, apiKey?: string): Bluebird<UnspentInfo[]>;
+  public abstract getAddressInfoFromExplorer(address: string, apiKey?: string): Bluebird<AddressInfo>;
+  public abstract getUnspentInfoFromExplorer(address: string, apiKey?: string): Bluebird<UnspentInfo[]>;
```
tylerlevine, line 1782: I'm not sure we want to add this to the public api, since it's really supposed to only be used for recoveries. Instead, maybe keep these `protected` and pass the functions into `recover`?

modules/core/src/v2/coins/utxo/recover.ts
```diff
@@ -0,0 +1,284 @@
...
+  coin: AbstractUtxoCoin,
+  params: RecoverParams,
+  keyArray: bitcoin.HDNode[],
+  basePath: string,
+  addressesById
+) {
+  const MAX_SEQUENTIAL_ADDRESSES_WITHOUT_TXS = params.scan || 20;
+  let numSequentialAddressesWithoutTxs = 0;
+
+  async function gatherUnspents(addrIndex: number) {
```
tylerlevine, line 37: I'd probably refactor this to be iterative instead of recursive but let's leave that for another PR

modules/core/src/v2/coins/utxo/recover.ts
```diff
@@ -0,0 +1,284 @@
...
+      const externalKey = coin.deriveKeys(baseKeyPath, externalChainCode);
+      const internalKey = coin.deriveKeys(baseKeyPath, internalChainCode);
+      queries.push(queryBlockchainUnspentsPath(coin, params, externalKey, '/0/0/' + externalChainCode, addressesById));
+      queries.push(queryBlockchainUnspentsPath(coin, params, internalKey, '/0/0/' + internalChainCode, addressesById));
+    }
+  });
+
+  // Execute the queries and gather the unspents
+  const queryResponses = await Promise.all(queries);
+  const unspents: any[] = _.flatten(queryResponses); // this flattens the array (turns an array of arrays into just one array)
```
tylerlevine, line 188: `// this comment repeats something from the documentation`

😂

tylerlevine (approved):
couple small comments, I'm mostly interested in your thoughts on the addition to the public API for the explorer functions

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: if it really is only used for recoveries IMO we should move it to `/recover.ts` and use `switch` instead of subclassing

OttoAllmendinger:
let's do further cleanups in separate PRs

PR #1362, opened 2021-08-05
fix(core): defer application of authorization headers

tylerlevine:
Prior to v12, the `Authorization` header was applied just before sending a
request, but in v12 it was applied when the request was instantiated. It's
possible for tokens to be swapped out in time between the request getting
instantiated and getting sent, so we need to continue to lazily apply the
`Authorization` header.

Ticket: BG-34772

### DIFF
--- a/modules/core/src/bitgo.ts
+++ b/modules/core/src/bitgo.ts
@@ -505,49 +505,50 @@ export class BitGo {
         req = req.proxy(self._proxy);
       }
 
-      // intercept a request before it's submitted to the server for v2 authentication (based on token)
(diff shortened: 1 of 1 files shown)


sangaman (approved):
Looks good, to be clear this only affects the "old" method of authentication right? The newer method was already set to apply the `Authorization` header right before the request is sent?

tylerlevine:
> Looks good, to be clear this only affects the "old" method of authentication right? The newer method was already set to apply the `Authorization` header right before the request is sent?

yep, that's correct :+1:

PR #1371, opened 2021-08-06
refactor(core): deprecate coin method `initiateRecovery()`

OttoAllmendinger:
This method has an awkward type signature that does not really fit well with
non-bip32 coins (XLM). It is also needlessly async.

Instead of messing around with complex generics we can simply define methods
`getBip32Keys()` and `getStellarKeys()` with good type signatures.

Outside of BitGoJS this method is only used in the Wallet-Recovery-Wizard, which can simply
use `getBip32Keys()` instead.

See individual commits for details.

Issue: BG-34381

### DIFF
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -469,28 +469,9 @@ export abstract class BaseCoin {
 
   /**
    * @deprecated - use getBip32Keys() in conjunction with isValidAddress instead
-   * @param params
    */
-  initiateRecovery(params: InitiateRecoveryOptions): Bluebird<any> {
-    const self = this;
-    return co(function* initiateRecovery() {
-      // Validate the destination address
-      try {
-        if (!self.isValidAddress(params.recoveryDestination)) {
-          throw new Error('Invalid destination address!');
-        }
-      } catch (e) {
-        // if isValidAddress is not implemented, assume the destination
-        // address is valid and let the tx go through. If the destination
-        // is actually invalid (`isValidAddress` returns false and does
-        // not throw), this method will still throw
-        if (!(e instanceof errors.MethodNotImplementedError)) {
-          throw e;
-        }
-      }
-
-      return getBip32Keys(self.bitgo, params, { requireBitGoXpub: false });
-    }).call(this);
+  initiateRecovery(params: InitiateRecoveryOptions): never {
+    throw new Error('deprecated method');
   }
 
   // Some coins can have their tx info verified, if a public tx decoder is available
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -23,6 +23,7 @@ import { Enterprises } from './enterprises';
 import { BaseCoin as AccountLibBasecoin } from '@bitgo/account-lib';
 import { signMessage } from '../bip32util';
 import * as crypto from 'crypto';
+import { InitiateRecoveryOptions } from './recovery/initiate';
 export type TransactionType = AccountLibBasecoin.TransactionType;
 
 export interface TransactionRecipient {
@@ -155,14 +156,6 @@ export interface ParsedTransaction {
 // TODO (SDKT-9): reverse engineer and add options
 export interface SignTransactionOptions {
   [index: string]: unknown;
-}
-
-export interface InitiateRecoveryOptions {
-  userKey: string;
-  backupKey: string;
-  bitgoKey?: string; // optional for xrp recoveries
-  recoveryDestination: string;
-  walletPassphrase?: string;
 }
 
 export interface KeychainsTriplet {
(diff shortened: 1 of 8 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1821,7 +1821,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
    * @param callback
    */
   recover(params: RecoverParams, callback?: NodeCallback<any>): Bluebird<any> {
-    return Bluebird.resolve(recover(this, params)).asCallback(callback);
+    return Bluebird.resolve(recover(this, this.bitgo, params)).asCallback(callback);
```
OttoAllmendinger, line 1824: `bitgo` is protected :shield:

modules/core/src/v2/coins/eos.ts
```diff
@@ -831,10 +759,26 @@ export class Eos extends BaseCoin {
...
+
+      const { krsProvider } = params;
+      if (getIsKrsRecovery(params)) {
+        if (!krsProvider || _.isUndefined(config.krsProviders[krsProvider])) {
+          throw new Error('unknown key recovery service provider');
+        }
+        const krsProviderConfig = config.krsProviders[krsProvider];
+        if (!krsProviderConfig.supportedCoins.includes(self.getFamily())) {
+          throw new Error('specified key recovery service does not support recoveries for this coin');
+        }
```
OttoAllmendinger, line 774: I have a separate PR queued for cleaning up this copypasta

modules/core/src/v2/recovery/initiate.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export function getBip32Keys(
+  bitgo: BitGo,
+  params: InitiateRecoveryOptions,
+  { requireBitGoXpub }: { requireBitGoXpub: boolean }
+): utxolib.HDNode[] {
+  const isKrsRecovery = getIsKrsRecovery(params);
+  const isUnsignedSweep = getIsUnsignedSweep(params);
+  const keys = [
+    // Box A
```
sangaman, line 62: Out of curiosity what is Box A/B/C in this context?

modules/core/src/v2/baseCoin.ts
```diff
@@ -474,75 +467,11 @@ export abstract class BaseCoin {
...
-        }
-      }
-
-      return keys;
-    }).call(this);
+  /**
+   * @deprecated - use getBip32Keys() in conjunction with isValidAddress instead
+   */
+  initiateRecovery(params: InitiateRecoveryOptions): never {
+    throw new Error('deprecated method');
```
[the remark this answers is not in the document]
OttoAllmendinger: It is only used here: https://github.com/BitGo/wallet-recovery-wizard/blob/397266def593dd71821fca6b8d4703aeac381858/src/utils.js - which is a noncritical codepath that can be adapted pretty easily.

Still, we probably should bump the minor version.

modules/core/src/v2/recovery/initiate.ts
OttoAllmendinger: this refers to the keycard PDF that is generated when you create a wallet

modules/core/src/v2/recovery/initiate.ts
```diff
@@ -0,0 +1,133 @@
...
+  source: string;
+  passphrase?: string;
+  isUnsignedSweep: boolean;
+  isKrsRecovery: boolean;
+}
+
+export interface InitiateRecoveryOptions {
+  userKey: string;
+  backupKey: string;
+  bitgoKey?: string; // optional for xrp recoveries
```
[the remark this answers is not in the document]
OttoAllmendinger: I considered that as well but decided the PR was already large enough. We can do it in a follow-up.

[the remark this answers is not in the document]
OttoAllmendinger: Ideally all of this code should move to the WRW anyway, which is the only place that needs it.

modules/core/src/v2/recovery/initiate.ts
```diff
@@ -0,0 +1,133 @@
...
+        return utxolib.HDNode.fromBase58(bitgo.decrypt({ password: passphrase, input: key }));
+      }
+    } catch (e) {
+      throw new Error(`Failed to decrypt ${source} key with passcode - try again!`);
+    }
+  }
+  try {
+    return utxolib.HDNode.fromBase58(key);
+  } catch (e) {
+    throw new Error(`Failed to validate ${source} key - try again!`);
```
argjv, line 50: Any concrete suggestions? `try again!` is too vague. Would be nice if we can suggest what else they can try to tell them what is wrong. In this case, looks like the key may be invalid?

OttoAllmendinger: They should probably check if they entered/scanned the correct box. 

Since I only moved this code around it's out of scope for this PR for a fix but I can add a JIRA.

modules/core/src/v2/recovery/initiate.ts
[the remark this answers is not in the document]
tylerlevine: > Ideally all of this code should move to the WRW anyway, which is the only place that needs it.

I was thinking a new `@bitgo/recovery` package might be a good place for this actually, because I think some similar code may be needed in https://github.com/BitGo/key-recovery-service-v2 or bgadmin.

modules/core/src/v2/recovery/initiate.ts
tylerlevine: This is probably a "try one more time, then contact support" failure scenario

modules/core/src/v2/recovery/initiate.ts
```diff
@@ -0,0 +1,133 @@
...
+  } catch (e) {
+    if (requireBitGoXpub) {
+      throw new Error('Failed to parse bitgo xpub!');
+    }
+  }
+
+  return keys;
+}
+
+export function getStellarKeys(bitgo: BitGo, params: InitiateRecoveryOptions): stellar.Keypair[] {
```
tylerlevine, line 92: Shouldn't this live with the rest of the stellar code? How come it was moved out?

tylerlevine (approved):

modules/core/src/v2/recovery/initiate.ts
OttoAllmendinger: It could live in `xlm.js` but it should not be a class method there (the type signature would be awkward).

This method is only used for recovery code paths so from that perspective it makes sense to put it under `recovery/`.

PR #1372, opened 2021-08-06
refactor(core): deprecate hdPath

OttoAllmendinger:
Instead of `hdPath(v).derive(p)` use `v.derivePath(p)` or
`v.derivePath(sanitizeLegacyPath(v))` depending on context.

Instead of `deriveKey(p)` use `derivePath(p).keyPair`.

Issue: BG-34381

### DIFF
--- a/modules/core/src/bitgo.ts
+++ b/modules/core/src/bitgo.ts
@@ -6,7 +6,7 @@
 
 import * as superagent from 'superagent';
 import * as bitcoin from '@bitgo/utxo-lib';
-import { makeRandomKey, hdPath } from './bitcoin';
+import { makeRandomKey } from './bitcoin';
 import * as secp256k1 from 'secp256k1';
 import bitcoinMessage = require('bitcoinjs-message');
 import { BaseCoin } from './v2/baseCoin';
@@ -45,6 +45,7 @@ import {
   toBitgoRequest,
   verifyResponse,
 } from './api';
+import { sanitizeLegacyPath } from './bip32path';
 
 const debug = debugLib('bitgo:index');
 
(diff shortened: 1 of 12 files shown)


modules/core/src/travelRule.ts
```diff
@@ -139,11 +140,9 @@ TravelRule.prototype.decryptReceivedTravelInfo = function (params: DecryptReceiv
     hdNode = bitcoin.HDNode.fromBase58(keychain.xprv);
   }
 
-  const self = this;
-  const hdPathNode = hdPath(hdNode);
-  tx.receivedTravelInfo.forEach(function (info) {
-    const key = hdPathNode.deriveKey(info.toPubKeyPath);
-    const secret = self.bitgo.getECDHSecret({
+  tx.receivedTravelInfo.forEach((info) => {
```
OttoAllmendinger, line 143: we call `this.bitgo.decrypt` further down so I think this is dead code

modules/core/src/v2/baseCoin.ts
```diff
@@ -440,14 +440,17 @@ export abstract class BaseCoin {
    * @returns {{key: string, derivationPath: string}}
    */
   deriveKeyWithSeed({ key, seed }: { key: string; seed: string }): { key: string; derivationPath: string } {
-    const derivationPathInput = bitcoin.crypto.hash256(`${seed}`).toString('hex');
```
OttoAllmendinger, line 443: `hash256` is double-sha256

modules/core/src/bitgo.ts
```diff
@@ -1045,8 +1046,8 @@ export class BitGo {
 
     // BIP32 derivation path is applied to both client and server master keys
     const derivationPath = responseBody.derivationPath;
-    const clientDerivedNode = hdPath(clientHDNode).derive(derivationPath);
-    const serverDerivedNode = hdPath(serverHDNode).derive(derivationPath);
+    const clientDerivedNode = clientHDNode.derivePath(sanitizeLegacyPath(derivationPath));
+    const serverDerivedNode = serverHDNode.derivePath(sanitizeLegacyPath(derivationPath));
```
sangaman, line 1050: You can call `sanitizeLegacyPath` once and make a `sanitizedDerivationPath` variable right above, saves a call and will make these lines a bit more readable.

sangaman (approved):

modules/core/src/bitgo.ts
OttoAllmendinger: true, I will revisit this anyway

PR #1375, opened 2021-08-06
feat(account-lib): skeleton code for avalanche c-chain in account-lib

shiksb:
Ticket: STLX-3250

### DIFF
--- a/modules/account-lib/src/coin/avaxc/transferBuilder.ts
+++ b/modules/account-lib/src/coin/avaxc/transferBuilder.ts
@@ -1,0 +1,22 @@
+import { TransferBuilder as EthTransferBuilder } from '../eth';
+
+/** AVAXC transfer builder */
+export class TransferBuilder extends EthTransferBuilder {
+  /**
+   * Get the prefix used in generating an operation hash for sending tokens
+   *
+   * @returns the string prefix
+   */
+  protected getTokenOperationHashPrefix(): string {
+    return 'AVAX-ERC20';
+  }
+
+  /**
+   * Get the prefix used in generating an operation hash for sending native coins
+   *
+   * @returns the string prefix
+   */
+  protected getNativeOperationHashPrefix(): string {
+    return 'AVAX';
+  }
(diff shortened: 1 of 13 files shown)


modules/account-lib/src/coin/avaxc/transferBuilder.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/** AVAXC transfer builder */
+export class TransferBuilder extends EthTransferBuilder {
+  /**
+   * Get the prefix used in generating an operation hash for sending tokens
+   *
+   * @returns the string prefix
+   */
+  protected getTokenOperationHashPrefix(): string {
+    return 'AVAXC-ERC20';
```
smaria95, line 11: @CapnMigraine 
This string should match the one on `SendMultisigToken` method on the contract (which is `AVAX-ERC20` at the moment).
Should we update the contract PR or this PR instead?

modules/account-lib/src/coin/avaxc/transferBuilder.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  /**
+   * Get the prefix used in generating an operation hash for sending native coins
+   *
+   * @returns the string prefix
+   */
+  protected getNativeOperationHashPrefix(): string {
+    return 'AVAXC';
```
smaria95, line 20: @CapnMigraine 
The same here but for contract method `SendMultisig`

modules/account-lib/src/coin/avaxc/transferBuilder.ts
CapnMigraine: @shiksb I think we can assume that if we were to support an asset like `AVAX`, but with multiple eth-like chains, the asset should be the same across them —i.e. `AVAX` on the X/P chains—, and the contract would be the same too. We can use `AVAX-ERC20` and `AVAX`.

CapnMigraine (approved):
lgtm

PR #1376, opened 2021-08-06
fix(core): fix regression in `addAccessToken` when using v1 auth

tylerlevine:
Prior to v12, we always called `handleResponseResult()` on the value returned
from the result of the call to platform, whether it was a v1 or v2 request.

After a bad refactor in v12, `handleResponseResult()` was no longer called when
using v1 auth. This broke callers who relied on this behavior for v1.

Instead, call `handleResponseResult()` on both return paths.

Additionally, add in some debug lines which were helpful in diagnosing this
issue.

Ticket: BG-34772

### DIFF
--- a/modules/core/src/bitgo.ts
+++ b/modules/core/src/bitgo.ts
@@ -545,6 +545,7 @@ export class BitGo {
           req.isV2Authenticated = false;
 
           req.set('Authorization', 'Bearer ' + self._token);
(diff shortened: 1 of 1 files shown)


bendapkiewicz:
Remove debug lines?

tylerlevine:
> Remove debug lines?

why? They are disabled by default, unlike `console.log`

bendapkiewicz:
Ah, I didn't know they had that purpose. Disregard.

argjv (approved):

PR #1377, opened 2021-08-09
refactor(core): use bip32 in ecdh operations

OttoAllmendinger:
* Move from class method to regular function (we don't use the bitgo
  object)
* Move tests
* Add support for bip32 arguments
* Return buffer, convert to `hex` at callsite

Issue: BG-34381

### DIFF
--- a/modules/core/src/bitgo.ts
+++ b/modules/core/src/bitgo.ts
@@ -138,11 +138,6 @@ export interface VerifyShardsOptions {
   xpub: string;
 }
 
-export interface GetEcdhSecretOptions {
-  otherPubKeyHex: string;
-  eckey: bitcoin.ECPair;
-}
-
 export interface AccessTokenOptions {
   accessToken: string;
 }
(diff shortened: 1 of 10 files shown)


modules/core/src/bitgo.ts
```diff
@@ -942,29 +937,6 @@ export class BitGo {
     return true;
   }
 
-  /**
-   * Construct an ECDH secret from a private key and other user's public key
-   */
-  getECDHSecret({ otherPubKeyHex, eckey }: GetEcdhSecretOptions): string {
```
tylerlevine, line 948: Could you please keep this function on the `bitgo` object as a pass through to the new one in `ecdh.ts` for the moment (and mark as `@deprecated`)? There are at least two places in the UI where this is called directly.

We can remove in a future major version.

tylerlevine (changes requested):

modules/core/src/bitgo.ts
OttoAllmendinger: ouch, forgot to check there

modules/core/src/ecdh.ts
```diff
@@ -0,0 +1,57 @@
...
+  } else if (privateKey instanceof utxolib.ECPair) {
+    privateKey = privateKey.d.toBuffer(32);
+  }
+
+  if (!Buffer.isBuffer(publicKey)) {
+    publicKey = publicKey.publicKey;
+  }
+
+  if (!Buffer.isBuffer(privateKey) || !Buffer.isBuffer(publicKey)) {
+    throw new Error(`invalid state`);
```
sangaman, line 44: Maybe add some more detail to this message?

sangaman (commented):
Changes look good overall.

modules/core/src/ecdh.ts
OttoAllmendinger: This is more to satisfy `tsc` than anything else. You'd really have put in some effort to get here. I would have used `assert` but I didn't want the extra import.

OttoAllmendinger: Wait a second, I actually do have `assert` already imported. I think for some reason the type assertion signature was not recognized. In any case, the user will almost certainly never see this and us devs have the stack trace.

tylerlevine (approved):

PR #1380, opened 2021-08-09
refactor(core): use bip32 in more places

OttoAllmendinger:
Issue: BG-34381


modules/core/src/v2/coins/eth.ts
```diff
@@ -347,8 +347,7 @@ export class Eth extends BaseCoin {
    */
   isValidPub(pub: string): boolean {
     try {
-      utxoLib.HDNode.fromBase58(pub);
```
tylerlevine, line 350: yikes, this was a bug. Thanks for fixing.

tylerlevine (approved):

PR #1381, opened 2021-08-09
fix(core):[STLX-4989] fix prebuild transaction for tron contractCalls

matilareo:
fix prebuild transaction for tron contractCalls. Pending approvals weren't saving fee limit.

stlx-4989

### DIFF
--- a/modules/core/src/v2/coins/trx.ts
+++ b/modules/core/src/v2/coins/trx.ts
@@ -325,13 +325,8 @@ export class Trx extends BaseCoin {
   getExtraPrebuildParams(buildParams: any, callback?: NodeCallback<any>): Bluebird<any> {
     return co<any>(function* () {
       if (buildParams.recipients[0].data && buildParams.feeLimit) {
-        const recipients: any[] = [{ ...buildParams.recipients[0] }];
-        recipients[0].feeLimit = buildParams.feeLimit;
-        return {
-          recipients,
-        };
+        buildParams.recipients[0].feeLimit = buildParams.feeLimit;
       }
-      return {};
     })
       .call(this)
       .asCallback(callback);
--- a/modules/core/test/v2/unit/coins/trx.ts
+++ b/modules/core/test/v2/unit/coins/trx.ts
@@ -1,7 +1,7 @@
 import { Trx } from '../../../../src/v2/coins/';
 import * as bitgoAccountLib from '@bitgo/account-lib';
 import { signTxOptions, mockTx } from '../../fixtures/coins/trx';
-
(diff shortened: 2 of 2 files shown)


modules/core/src/v2/coins/trx.ts
```diff
@@ -320,13 +320,8 @@ export class Trx extends BaseCoin {
   getExtraPrebuildParams(buildParams: any, callback?: NodeCallback<any>): Bluebird<any> {
     return co<any>(function* () {
       if (buildParams.recipients[0].data && buildParams.feeLimit) {
```
fmlenarduzzi, line 322: Why you are asking about `.data` if it is not used into `then` piece of code?

matilareo: Yes, i used that condition to check if the transaction is of type ContractCall

fmlenarduzzi: @matilareo  confirmed that .data is used to be sure that this is a contract call.

Done.

fmlenarduzzi (dismissed):

fmlenarduzzi (approved):
LGTM

modules/core/src/v2/coins/trx.ts
```diff
@@ -325,13 +325,8 @@ export class Trx extends BaseCoin {
...
-        const recipients: any[] = [{ ...buildParams.recipients[0] }];
-        recipients[0].feeLimit = buildParams.feeLimit;
-        return {
-          recipients,
-        };
+        buildParams.recipients[0].feeLimit = buildParams.feeLimit;
       }
-      return {};
```
tylerlevine, line 334: Why is the return type changing? I think we need this to return the new build params: https://github.com/BitGo/BitGoJS/blob/cc358bde41e46ade21e744ad710ef4610652cb78/modules/core/src/v2/wallet.ts#L1687

tylerlevine (changes requested):

modules/core/src/v2/coins/trx.ts
matilareo: i ve removed those returns and change the buildParams itself because the data was lost on the way to the send().

tylerlevine: Does this actually work? Have you tested the complete prebuild flow?

matilareo: Yes, we have tested the complete flow of contract calls on wallet-platform and trust admin tool with @pviolasimtlix

tylerlevine (approved):

PR #1385, opened 2021-08-10
feat(account-lib): Add Avax C-Chain support as Eth Like Coin

smaria95:


modules/account-lib/package.json
```diff
@@ -35,7 +35,7 @@
   "dependencies": {
     "@bitgo/blake2b": "^3.0.0",
     "@bitgo/bls": "^0.1.0",
-    "@bitgo/statics": "^6.12.0-rc.2",
+    "@bitgo/statics": "^6.12.0-rc.3",
```
tylerlevine, line 38: please do not modify the statics version - this should only be changed during the release process.

tylerlevine (changes requested):
1) how come a `tgz` file is being added?
2) please do not modify the version of statics in the package.json
3) please rebase due to linter changes made since this PR was created

PR #1386, opened 2021-08-10
revert: return master branch package versions to non-rc versions

tylerlevine:
Version numbers for packages inside the bitgojs monorepo are managed by lerna
and should not be updated except on the rel/rc branch during the release
process.

Ticket: NO-000


bitgoAaron (approved):

shiksb:
thank you - will keep this in mind next time

PR #1392, opened 2021-08-11
chore(utxo-lib): port networks to typescript

OttoAllmendinger:
Issue: BG-7478

f8eb4df0 (Otto Allmendinger, 28 minutes ago)
refactor(utxo-lib): deprecate use of `network.coin`

We already have `getNetworkName()` and don't need three-letter codes.

Issue: BG-7478


modules/utxo-lib/test/integration_local_rpc/generate/main.ts
```diff
@@ -36,7 +36,7 @@ async function initBlockchain(rpc: RpcClient, network: Network): Promise<void> {
   }
 }
 
-function toRegtestAddress(network: Network, scriptType: ScriptType, script: Buffer): string {
+function toRegtestAddress(network: { bech32?: string }, scriptType: ScriptType, script: Buffer): string {
```
sangaman, line 39: Is the `bech32` variable being used anywhere?

modules/utxo-lib/src/networkTypes.ts
```diff
@@ -0,0 +1,58 @@
...
+export type Network = {
+  messagePrefix: string;
+  pubKeyHash: number;
+  scriptHash: number;
+  wif: number;
+  bip32: {
+    public: number;
+    private: number;
+  };
+  bech32?: string;
```
sangaman, line 48: Is this supposed to represent the bech32 prefix? Just not clear what it's being used for from the changes afaict.

sangaman (approved):
Looks good, `networks.ts` and `networkTypes.ts` would ideally be kept around even after wrapping bitcoinjs-lib right?

modules/utxo-lib/test/integration_local_rpc/generate/main.ts
OttoAllmendinger: yep it is passed to `fromOutputScript` later

OttoAllmendinger:
yes this code will stick around after we convert the lib to a wrapper

PR #1397, opened 2021-08-11
fix(core): change type of `sequenceId` to string

tylerlevine:
This parameter is a free-form string and not a number.

Ticket: BG-35021


zhongtan912 (approved):
Looks good

sangaman (approved):

PR #1398, opened 2021-08-12
feat(utxo-lib): extend signature tests

OttoAllmendinger:
Move some code from BitGoJS/core to utxolib, add more tests

Issue: BG-34381


modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -0,0 +1,267 @@
...
+/**
+ * Parse a transaction's signature script to obtain public keys, signatures, the sig script,
+ * and other properties.
+ *
+ * Only supports script types used in BitGo transactions.
+ *
+ * @param input
+ * @returns ParsedSignatureScript
+ */
+export function parseSignatureScript(input: Input): ParsedSignatureScript {
```
OttoAllmendinger, line 63: see https://github.com/BitGo/BitGoJS/blob/693b02dab445623848c301cce7a0c6dcfc0bf003/modules/core/src/v2/coins/abstractUtxoCoin.ts#L1288

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -0,0 +1,267 @@
...
+ * Verify the signature on a (half-signed) transaction
+ * @param transaction bitcoinjs-lib tx object
+ * @param inputIndex The input whererfore to check the signature
+ * @param amount For segwit and BCH, the input amount needs to be known for signature verification
+ * @param verificationSettings
+ * @param verificationSettings.signatureIndex The index of the signature to verify (only iterates over non-empty signatures)
+ * @param verificationSettings.publicKey The hex of the public key to verify (will verify all signatures)
+ * @returns {boolean}
+ */
+export function verifySignature(
```
OttoAllmendinger, line 153: see https://github.com/BitGo/BitGoJS/blob/693b02dab445623848c301cce7a0c6dcfc0bf003/modules/core/src/v2/coins/abstractUtxoCoin.ts#L1399

modules/utxo-lib/test/integration_local_rpc/parse.ts
```diff
@@ -27,26 +33,90 @@ function runTestParse(network: Network, txType: FixtureTxType, scriptType: Scrip
...
+            assert.strictEqual(result.inputClassification, 'scripthash');
+            break;
+          case 'p2wsh':
+            assert.strictEqual(result.inputClassification, 'witnessscripthash');
+            break;
+        }
+      });
+    });
+
+    if (txType === 'deposit') {
```
sangaman, line 73: Is this `if` statement supposed to be here or in the `it` block below?

modules/utxo-lib/test/integration_local_rpc/parse.ts
```diff
@@ -27,26 +33,90 @@ function runTestParse(network: Network, txType: FixtureTxType, scriptType: Scrip
...
+    it(`verifySignatures`, function () {
+      parsedTx.ins.forEach((input, i) => {
+        const inputTxid = Buffer.from(input.hash).reverse().toString('hex');
+        const inputTx = fixture.inputs.find((tx) => tx.txid === inputTxid);
+        if (!inputTx) {
+          throw new Error(`could not find inputTx`);
+        }
+        const prevOutput = inputTx.vout[input.index];
+        if (!prevOutput) {
+          throw new Error(`could not prevOutput`);
```
sangaman, line 86: could not "find"

modules/utxo-lib/test/integration_local_rpc/generate/regtestNode.ts
```diff
@@ -79,6 +79,7 @@ export async function getRegtestNode(network: Network): Promise<Node> {
     dockerParams.image,
     ...(dockerParams.binary ? [dockerParams.binary] : []),
     '-regtest',
+    '-txindex',
```
sangaman, line 82: Why is `txindex` necessary with this PR?

modules/utxo-lib/test/integration_local_rpc/generate/main.ts
```diff
@@ -113,19 +113,32 @@ async function run(network: Network) {
...
 }
 
-async function main() {
+async function main(args: string[]) {
+  const allowedNetworks = args.map((name) => {
+    const network = utxolib.networks[name];
+    if (!network) {
+      throw new Error(`invalid network ${name}`);
+    }
+    return getMainnet(network);
```
sangaman, line 122: Also curious here as to why we want the mainnet network.

modules/utxo-lib/src/transaction.js
```diff
@@ -564,6 +564,41 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
...
+      return this.hashForZcashSignature(inIndex, prevoutScript, value, hashType)
+    default:
+      if (isSegwit) {
+        return this.hashForWitnessV0(inIndex, prevoutScript, value, hashType)
+      } else {
+        return this.hashForSignature(inIndex, prevoutScript, hashType)
+      }
+  }
+
+  throw new Error('invalid state') // eslint-disable-line
```
sangaman, line 599: This looks like unreachable code, I see you're disabling the eslint warning but why have it at all? Nice to have this logic in its own reusable method though.

modules/utxo-lib/src/bitgo/signature.ts
sangaman: I haven't checked (and `verifySignature`) closely but it looks like straight copies or close to it from the methods you've referenced. Would the next step then be to remove these referenced methods from `core` and use the ones you're adding here instead?

sangaman (commented):
Looks like good test cases and refactoring.

modules/utxo-lib/src/transaction.js
sangaman: Also, should this be within the `bitgo` subfolder since it appears to be bitgo specific logic? This file could be typescript too.

modules/utxo-lib/src/bitgo/signature.ts
OttoAllmendinger: yes

modules/utxo-lib/test/integration_local_rpc/generate/regtestNode.ts
OttoAllmendinger: because I am looking up confirmed input transactions now

modules/utxo-lib/src/transaction.js
OttoAllmendinger: While it is bitgo-specific, it would still be a overridden class method when the lib becomes a wrapper.

The `src/bitgo` should be for higher-level wrappers. I want to avoid dependencies from this code back.

modules/utxo-lib/test/integration_local_rpc/generate/main.ts
OttoAllmendinger: we normalize to mainnet so that it works regardless if you use `bitcoingoldTestnet` or just `bitcoingold` as the argument

modules/utxo-lib/src/transaction.js
OttoAllmendinger: Re: unreachable code - this is just a habit. If the code changes and we don't hit a `return` we are still safe.

modules/utxo-lib/test/integration_local_rpc/parse.ts
OttoAllmendinger: will fix

modules/utxo-lib/test/integration_local_rpc/parse.ts
OttoAllmendinger: For deposits we don't even want to set up the `verifySignatures` test.

If I return inside `it` it might look like the test actually executed successfully.

sangaman (approved):
Thanks for the clarifications/answers - all fine with me.

PR #1403, opened 2021-08-13
feat(root): set tsconfig target to `es6`

OttoAllmendinger:
This allows typescript to compile with native `async`/`await` syntax,
greatly improving stack traces.

Issue: BG-34381

### DIFF
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -518,5 +518,5 @@ export abstract class BaseCoin {
   /**
    * Sign a transaction
    */
-  abstract signTransaction(params: SignTransactionOptions): Bluebird<SignedTransaction>;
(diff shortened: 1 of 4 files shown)


sangaman (approved):
Was the build failure without the second commit due to eslint or just typescript compilatiion issues? Just looks like the explicit `Bluebird` return type declaration was incompatible with returning a non-Bluebird object in an `async` function. I'm thinking it'd be nice to transition more from Bluebird to Promise.

The `es6` target change would break compatibility with some (very old) clients I believe but I figure that's not an issue.

OttoAllmendinger:
The earlier build failure (as well as the new one) are actual typescript compiler errors, oddly enough.

Since we use node10 as the lower end in our build matrix I think we cannot expect earlier versions to work correctly anyway.

tylerlevine (approved):

tylerlevine:
build is failing on express typescript errors, please take a look

sangaman (approved):

PR #1404, opened 2021-08-13
BG-30057 fix(core): fix issue of erc20 token recovery using unsigned sweep

gianchandania:
**Background**

There was an issue in the recovery of ERC20 tokens after generating a transaction using an unsigned sweep. The transaction was getting generated and signed successfully using OVC. But on broadcasting the txHex on the network, the tokens were not transferred to the recovery destination address and stayed in its original address. The issue was coming because `signFinal()` function which is used to do the 2nd signature to the transaction was having the 'sendMultiSig' function name hardcoded while forming the sendData for the transaction. This same function is getting used for signing transactions for Eth and ERC20 tokens. So when it was being invoked for signing the transaction for ERC20 tokens, the method name 'sendMultiSigToken' should have been used. But currently method name 'sendMultiSig' was being used.

**Changes**

- Declared the property `sendMethodName` in the class Eth. have initialized it as 'sendMultiSig' in the constructor of Eth, Teth, and Gteth classes, and initialized it as 'sendMultiSigToken' in the constructor of Erc20Token class. Then instead of using the method name hardcoded in the `signFinal()` and `recover()` method, have used the property `sendMethodName`.

- Added a constant `TEST_TDAI_TOKEN_ADDRESS` in the test_bitgo.ts file for the unit test of recovering ERC20 tokens

- Added a sample transaction for the unsigned recovery transaction for ERC20 tokens in eth.js fixtures file

- Added nocks for unit tests of generation of Eth and ERC20 recovery transactions using unsigned sweep
 
- Added unit test for recovery of Eth and ERC20 using unsigned sweep

- Added unit test for adding a second signature to unsigned sweep for erc20 token

Ticket: 30057


barathcj (approved):
lgtm

karankurbur (approved):

modules/core/src/v2/coins/eth.ts
```diff
@@ -283,12 +283,14 @@ interface RecoverTokenTransaction {
 
 export class Eth extends BaseCoin {
   static hopTransactionSalt = 'bitgoHopAddressRequestSalt';
+  protected sendMethodName: string;
```
tylerlevine, line 286: Could this be typed a bit tighter?
```suggestion
  protected readonly sendMethodName: 'sendMultiSig' | 'sendMultiSigToken';
```

gianchandania: I think this will have to be typed like this because if I will assign it as readonly, then I will not be able to assign its value in the constructors, right?
`protected sendMethodName: 'sendMultiSig' | 'sendMultiSigToken';`

gianchandania: I have updated the type. Can you please once review?

twtaylorbitgo (approved):
This looks good. Let's try to follow-up with unsupported tokens. That's a more common use-case IMO and I don't believe this PR encompasses that

tylerlevine (approved):

gianchandania:
> This looks good. Let's try to follow-up with unsupported tokens. That's a more common use-case IMO and I don't believe this PR encompasses that

Sure, will look into if unsupported tokens are working fine or not as a part of other ticket

PR #1405, opened 2021-08-13
refactor(utxo-lib): fix, simplify hashForSignatureByNetwork

OttoAllmendinger:
Like Bitcoin Cash, Bitcoin Gold also uses `SIGHASH_FORKID`.

Both code paths can fit nicely in hashForSignatureByNetwork

Issue: BG-34381

3624b698 (Otto Allmendinger, 13 hours ago)
fix(utxo-lib): fix fixture generation for bitcoingold

The REAL regtest chain which mirrors mainnet only starts at block height
2000.

From that point it is also required to set the correct SIGHASH flag
(BIP143).

Issue: BG-34381


tylerlevine (approved):

modules/utxo-lib/src/transaction.js
```diff
@@ -581,22 +583,48 @@ Transaction.prototype.hashForSignatureByNetwork = function (
...
       return this.hashForZcashSignature(inIndex, prevoutScript, value, hashType)
-    default:
-      if (isSegwit) {
-        return this.hashForWitnessV0(inIndex, prevoutScript, value, hashType)
-      } else {
-        return this.hashForSignature(inIndex, prevoutScript, hashType)
+    case networks.bitcoingold:
+    case networks.bitcoinsv:
+    case networks.bitcoincash:
+      if (hashType & Transaction.SIGHASH_FORKID) {
```
sangaman, line 591: This block of code is some what difficult to follow, as is often the case when using bitwise operators. It would be nice to have more comments here. It looks this PR is removing some comment blocks like what was in `hashForCashSignature` which was helpful for me to understand this PR, so even just restoring those or re-summarizing the intended logic would be good I think. Maybe also putting a description for the `hashType` param.

modules/utxo-lib/test/integration_local_rpc/generate/main.ts
```diff
@@ -21,12 +21,19 @@ import { RpcClient } from './RpcClient';
...
   switch (network) {
     case utxolib.networks.testnet:
       await rpc.createWallet('utxolibtest');
+      break;
+    case utxolib.networks.bitcoingoldTestnet:
+      // The actual BTC/BTG fork flag only gets activated at this height.
+      // On mainnet the height was at 491407 (Around 10/25/2017 12:00 UTC)
+      // Prior to that, signatures that use the BIP143 sighash flag are invalid.
+      // https://github.com/BTCGPU/BTCGPU/blob/71894be9/src/chainparams.cpp#L371
+      minBlocks = 2001;
```
sangaman, line 34: Nice explanation here.

sangaman (dismissed):
Looks like good refactoring, just want to make sure we keep as good or better commenting and clarity around some of this logic since it's not very intuitive on its own.

modules/utxo-lib/src/transaction.js
OttoAllmendinger: improved comments

tylerlevine (approved):

PR #1407, opened 2021-08-16
chore(core): upgrade ethereumjs tx lib

ryan-bitgo:
Upgrades the ethereumjs tx library in order to support handling EIP1559 hop transactions.

Minior changes to the "should successfully fully sign a half-signed transaction from the offline vault" test because the signature changes due to the inclusion of the chain id in the new library.

ticket: BG-35058


modules/core/src/v2/coins/erc20Token.ts
```diff
@@ -244,16 +244,18 @@ export class Erc20Token extends Eth {
       const encodedArgs = optionalDeps.ethAbi.rawEncode(_.map(sendMethodArgs, 'type'), _.map(sendMethodArgs, 'value'));
       const sendData = Buffer.concat([methodSignature, encodedArgs]);
 
-      // Build contract call and sign it
-      const tx = new optionalDeps.EthTx({
+      const txParams = {
```
ryan-bitgo, line 247: needs to be extracted because spendAmount isnt an actual parameter for `fromTxData`. Not sure if this parameter is actually needed...

modules/core/test/v2/fixtures/coins/eth.ts
```diff
@@ -46,7 +46,7 @@ module.exports.getHalfSignedTethFromVault = function () {
...
+      txHex: 'f901cb018504a817c8008307a1209427121c36c854c8775166f01ee9b2a5cc012718af80b9016439125215000000000000000000000000ebd0d0c1f101ab5a27ef1c54430e0c4b1166548b0000000000000000000000000000000000000000000000000de0b6b3a764000000000000000000000000000000000000000000000000000000000000000000c0000000000000000000000000000000000000000000000000000000005c589d35000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000e00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004145864316d0c882ace5a3d5a000db8b9047835281c7dc9759441424f3f470d48821f9baebc35e4d1955af1cb072d4e9b1fa4d4dddcb7a859ba6d3bbe33f6efdfb1c0000000000000000000000000000000000000000000000000000000000000026a0382bdabfd4a58c043f16748f10d2700458aa549911a8c4d3418f77999d6a2be2a022f01307dc56baefee69d0d39454eba04e8cd6637ff33b42c34649004aed8c0b',
```
ryan-bitgo, line 49: changed due to changes in signature from inclusion of chain id.

modules/core/src/v2/coins/eth.ts
```diff
@@ -36,6 +36,7 @@ import { Util } from '../internal/util';
 import { EthereumLibraryUnavailableError } from '../../errors';
 import { BaseCoin as StaticsBaseCoin, EthereumNetwork } from '@bitgo/statics';
 import { checkKrsProvider, getIsKrsRecovery, getIsUnsignedSweep } from '../recovery/initiate';
+import type * as EthTxLib from '@ethereumjs/tx';
```
ryan-bitgo, line 39: this should get erased at compile time, so even if this is an optional dependency should be fine.

modules/core/src/v2/coins/erc20Token.ts
sangaman: Looks like it's not used anywhere else here either, why not just remove it if it's not being used?

wesgraham (approved):
nicely done

modules/core/src/v2/coins/erc20Token.ts
ryan-bitgo: https://github.com/BitGo/BitGoJS/pull/1408

PR #1408, opened 2021-08-16
refactor(core): remove extraneous spendAmount field

ryan-bitgo:
Removes the extraneous spendAmount field which is not used anywhere for constructing an eth transaction.

Not even used in the constructor: https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/baseTransaction.ts#L71

ticket: BG-35058


tylerlevine (changes requested):
Please check build failure

ryan-bitgo:
Fixed :)

tylerlevine (approved):

PR #1415, opened 2021-08-17
chore(core): use bip32 in abstractUtxoCoin

OttoAllmendinger:
Issue: BG-34381

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -47,6 +47,7 @@ import { CrossChainRecoveryTool } from '../recovery';
 import { NodeCallback } from '../types';
 import { Wallet } from '../wallet';
 import { toBitgoRequest } from '../../api';
+import { sanitizeLegacyPath } from '../../bip32path';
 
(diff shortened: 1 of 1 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1172,8 +1179,8 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
           );
           continue;
         }
-        const privKey = keychain.derivePath(signatureContext.path).keyPair;
-        privKey.network = self.network;
+
+        const keyPair = keychain.derivePath(sanitizeLegacyPath(signatureContext.path));
```
tylerlevine, line 1183: How come we're assigning `network` to the `keychain` now instead of the `keyPair` (what used to be called `privKey`)?

Do we need the network property on the `keychain` at all?

OttoAllmendinger: I am now assigning `network` in L1147 already

Currently `txb.network` needs to match `keychain.network` which is `networks.bitcoin` by default.

tylerlevine (approved):

PR #1418, opened 2021-08-17
feat(utxo-lib): add createSpendTransaction match

OttoAllmendinger:
Assert that re-signing the transaction with the same inputs and keys
results in the exact same hex output.

Issue: BG-34381


OttoAllmendinger:
superseded by https://github.com/BitGo/BitGoJS/pull/1418

PR #1420, opened 2021-08-17
Bg 34791 teos explorer url

lcovar:
Addressing issue BG-34791. 

Explorer url for teos should have been local.bloks.io and new param optional field added since teos explorer url needs params appended after the txid.

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -294,7 +294,7 @@ class Eos extends Mainnet implements AccountNetwork {
 
 class EosTestnet extends Testnet implements AccountNetwork {
   family = CoinFamily.EOS;
-  explorerUrl = 'https://jungle.bloks.io/transaction/';
+  explorerUrl = 'https://local.bloks.io/transaction/';
 }
 
 class Hedera extends Mainnet implements AccountNetwork {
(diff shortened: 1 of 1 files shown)


modules/statics/src/networks.ts
```diff
@@ -9,6 +9,7 @@ export abstract class BaseNetwork {
   public abstract readonly type: NetworkType;
   public abstract readonly family: CoinFamily;
   public abstract readonly explorerUrl: string | undefined;
+  public readonly params?: string;
 }
```
shihao-guo, line 13: let's call it urlParams to be more explicit

modules/statics/src/networks.ts
```diff
@@ -294,7 +295,9 @@ class Eos extends Mainnet implements AccountNetwork {
 
 class EosTestnet extends Testnet implements AccountNetwork {
   family = CoinFamily.EOS;
-  explorerUrl = 'https://jungle.bloks.io/transaction/';
+  explorerUrl = 'https://local.bloks.io/transaction/';
+  urlParams =
```
shihao-guo, line 299: do you still need this given https://github.com/BitGo/bitgo-microservices/pull/10947/files#diff-f6421b8a50f8aff713d166802c0070055b0f48f6deaecbbc50603d8440638ebbR441

PR #1421, opened 2021-08-18
fix(accountlib): improve multisig in order to user any order or combi…

jrafael-simtlix:
…nation of keys

Updated @stacks/transaction to 2.0.1 to support uncompressed keys in multisig   Added anchorMode
needed for the updated stacks lib   TransactionBuilder and buildImplementation has been refactored
to allow to use any order or combination of keys   Added Unit tests

stlx-5665


modules/account-lib/src/coin/stx/transactionBuilder.ts
```diff
@@ -110,39 +106,41 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
-      }
-    }
     this.initBuilder(tx);
     return this.transaction;
   }
 
   // region Base Builder
   /** @inheritdoc */
   protected async buildImplementation(): Promise<Transaction> {
+    const isMultiSig: boolean = this._fromPubKeys.length > 1;
```
wesgraham, line 116: why are we continually redefining what multisig criteria are? can we use `!isSingleSig()`here?

modules/account-lib/test/unit/coin/stx/transactionBuilder/transferBuilder.ts
```diff
@@ -202,6 +203,62 @@ describe('Stx Transfer Builder', () => {
...
+      builder.fromPubKey([testData.pub1, testData.pub2, testData.pub3]);
+      builder.numberSignatures(2);
+      const tx = await builder.build();
+      const txBuilder2 = factory.getTransferBuilder();
+      txBuilder2.from(tx.toBroadcastFormat());
+      txBuilder2.sign({ key: testData.prv3 });
+      const signedTx = await txBuilder2.build();
+
+      const txJson = signedTx.toJson();
+      should.deepEqual(signedTx.toBroadcastFormat(), testData.SIGNED_TRANSACTION_PK_2_3);
```
wesgraham, line 253: can we also verify that these broadcastablt tx's are not going to be rejected by the node? In short, can you make sure the only issue is `insufficient balance` from the sender when you broadcast it to the node, or better yet, can you fund the sender's address on testnet then try to broadcast this signed tx to make sure it's valid?

wesgraham: Same goes for these other tx's

modules/account-lib/src/coin/stx/contractBuilder.ts
```diff
@@ -27,9 +27,11 @@ export class ContractBuilder extends TransactionBuilder {
   private _contractName: string;
   private _functionName: string;
   private _functionArgs: ClarityValue[];
+  private _anchorMode: number;
 
   constructor(_coinConfig: Readonly<CoinConfig>) {
     super(_coinConfig);
+    this._anchorMode = 3;
```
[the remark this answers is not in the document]
alebusse:
```
enum AnchorMode {
  /** The transaction MUST be included in an anchored block */
  OnChainOnly = 0x01,
  /** The transaction MUST be included in a microblock */
  OffChainOnly = 0x02,
  /** The leader can choose where to include the transaction (anchored block or microblock)*/
  Any = 0x03,
}
```
when we updated the stacks.js these params new params are required.

modules/account-lib/src/coin/stx/transaction.ts
```diff
@@ -161,7 +169,7 @@ export class Transaction extends BaseTransaction {
 
   private getNonce(): number {
     if (this._stxTransaction.auth.spendingCondition) {
-      return this._stxTransaction.auth.spendingCondition.nonce.toNumber();
+      return Number(this._stxTransaction.auth.spendingCondition.nonce);
```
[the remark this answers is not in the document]
alebusse: when we updated the stacks.js the type changed and toNumber is not longer working

modules/account-lib/src/coin/stx/transactionBuilder.ts
jrafael-simtlix: That condition is relocated from the following reference. It means the opposite of isSingleSig. They are two different things. The plurality of the semantics was changed to avoid confusion. isSingleSig is an internal function of stack.js. On the first try I used isSingleSig. You can go back for that truck if you think it will be clearer.

https://github.com/BitGo/BitGoJS/pull/1421/files#diff-7e28454346122befb99605f628617c5b26224561ccb8411bd2971193c24c91bdL60

modules/account-lib/test/unit/coin/stx/transactionBuilder/transferBuilder.ts
jrafael-simtlix: The raw tx was built in coin-sandbox. Actually, this Tx does not have enough fee to work. But you can try creating a similar one that works. Look at the following reference.

https://github.com/BitGo/coins-sandbox/pull/85/files#diff-cd398d1a67e4543c01e6ec623432f224e55834573abf74cbc11eda29a17cc50bR18

modules/account-lib/src/coin/stx/transactionBuilder.ts
jrafael-simtlix: https://github.com/BitGo/coins-sandbox/blob/8cd53d9c0d69d53b72434a86019ea86a1cae34a1/stack/sendMultiSigMultiStep.js#L18

modules/account-lib/test/unit/coin/stx/transactionBuilder/transferBuilder.ts
jrafael-simtlix: https://github.com/BitGo/coins-sandbox/blob/8cd53d9c0d69d53b72434a86019ea86a1cae34a1/stack/sendMultiSigMultiStep.js#L18

wesgraham (approved):

PR #1428, opened 2021-08-18
feat(core): Add support for Avalanche C-Chain

smaria95:
[STLX-5265](https://bitgoinc.atlassian.net/browse/STLX-5265).

### DIFF
--- a/modules/core/src/v2/coins/avaxc.ts
+++ b/modules/core/src/v2/coins/avaxc.ts
@@ -1,0 +1,62 @@
+/**
+ * @prettier
+ */
+import { BaseCoin, KeyPair, SignedTransaction } from '../baseCoin';
+import { BitGo } from '../../bitgo';
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import { AbstractEthLikeCoin, EthSignTransactionOptions, SignedEthLikeTransaction } from './abstractEthLikeCoin';
+import { AvaxC as AvaxCAccountLib } from '@bitgo/account-lib';
+import { NodeCallback } from '../types';
+
+export class AvaxC extends AbstractEthLikeCoin {
+  protected constructor(bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) {
+    super(bitgo, staticsCoin);
+  }
+
+  static createInstance(bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>): BaseCoin {
+    return new AvaxC(bitgo, staticsCoin);
+  }
+
+  generateKeyPair(seed?: Buffer): KeyPair {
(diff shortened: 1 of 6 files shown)


jrafael-simtlix (approved):

modules/core/src/v2/coins/tavaxc.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @prettier
+ */
+import { BaseCoin, KeyPair, SignedTransaction } from '../baseCoin';
+import { BitGo } from '../../bitgo';
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import { AbstractEthLikeCoin, EthSignTransactionOptions, SignedEthLikeTransaction } from './abstractEthLikeCoin';
+import { AvaxC as AvaxCAccountLib } from '@bitgo/account-lib';
+import { NodeCallback } from '../types';
+
+export class TavaxC extends AbstractEthLikeCoin {
```
CapnMigraine, line 11: why don't we extend from `AvaxC` instead? 
The methods `generateKeyPair`, `isValidPub`, `isValidAddress` and `signTransaction` can be inherited

modules/core/test/v2/unit/coins/avaxc.ts
```diff
@@ -0,0 +1,243 @@
...
+    bitgo = new TestBitGo({ env: 'mock' });
+    bitgo.initializeTestVars();
+  });
+
+  beforeEach(() => {
+    tavaxCoin = bitgo.coin('tavaxc') as TavaxC;
+    avaxCoin = bitgo.coin('avaxc') as AvaxC;
+  });
+
+  describe('Instanciate', () => {
```
CapnMigraine, line 20: nit: typo
```suggestion
  describe('Instantiate', () => {
```

CapnMigraine (changes requested):

modules/core/src/v2/coins/tavaxc.ts
smaria95: I thought about this when doing these changes and I still think is the best approach too. 
The only reason I did not do that is because it's not done for the other coins and I wanted to keep it consistent with that.

I'll do it and update this PR

modules/core/src/v2/coins/avaxc.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  isValidPub(pub: string): boolean {
+    let valid = true;
+    try {
+      new AvaxCAccountLib.KeyPair({ pub });
+    } catch (e) {
+      valid = false;
+    }
+    return valid;
```
shiksb, line 36:
```
try {
      new AvaxCAccountLib.KeyPair({ pub });
} catch (e) {
      return false;
}
return true;
```

CapnMigraine (approved):
lgtm

PR #1432, opened 2021-08-20
feat(accountlib): STX ContractBuilder functionArg missing options arg…

jrafael-simtlix:
…uments

functionArgs don't work with optional arguments. Also add support tu buffer in string or int.

[STLX-6034](https://bitgoinc.atlassian.net/browse/STLX-6034)

### DIFF
--- a/modules/account-lib/src/coin/stx/contractBuilder.ts
+++ b/modules/account-lib/src/coin/stx/contractBuilder.ts
@@ -1,20 +1,24 @@
 import { BaseCoin as CoinConfig } from '@bitgo/statics';
 import BigNum from 'bn.js';
 import {
-  makeUnsignedContractCall,
-  UnsignedContractCallOptions,
-  PayloadType,
-  ContractCallOptions,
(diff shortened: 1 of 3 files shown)


alebusse (approved):
LGTM!

wesgraham (approved):

modules/account-lib/src/coin/stx/contractBuilder.ts
```diff
@@ -149,25 +153,37 @@ export class ContractBuilder extends TransactionBuilder {
...
-      if (arg.val instanceof Array) {
-        const data = {};
-        arg.val.forEach((a) => {
-          data[a.key] = this.parseCv({ type: a.type, val: a.val });
-        });
-        return tupleCV(data);
-      } else {
-        return noneCV();
-      }
-    } else if (typeof arg.val === 'string') {
```
jrafael-simtlix, line 167: @wesgraham I think that if val was number, it didn't encoded and it was also working.

PR #1433, opened 2021-08-20
fix(eth): make replay protection optional

ryan-bitgo:
follow up to: https://github.com/BitGo/BitGoJS/pull/1407

To preserve backwards compatibility, we need to make replay protection optional. Otherwise, we cannot broadcast transactions signed on other chains using OVC because it will default to mainnet.

See the ticket for the error in etherscan.

ticket: BG-35058


modules/core/test/v2/unit/ethWallet.ts
```diff
@@ -314,21 +314,21 @@ describe('Add final signature to ETH tx from offline vault', function () {
...
     const actualTx = EthTx.fromSerializedTx(Buffer.from(response.txHex, 'hex'));
-    actualTx.nonce.should.deepEqual(expectedTx.nonce);
+    actualTx.nonce.toString().should.deepEqual(expectedTx.nonce.toString());
     should.exist(actualTx.to);
     actualTx.to?.should.deepEqual(expectedTx.to);
     actualTx.value.should.deepEqual(expectedTx.value);
     actualTx.data.should.deepEqual(expectedTx.data);
     actualTx.isSigned().should.equal(true);
-    actualTx.supports(Capability.EIP155ReplayProtection).should.equal(true);
+    actualTx.supports(Capability.EIP155ReplayProtection).should.equal(false);
```
ryan-bitgo, line 323: this is the key change in this PR to ensure this works.

wesgraham (approved):

PR #1439, opened 2021-08-24
refactor(core): use utxolib.bitgo.verifySignature()

OttoAllmendinger:
Use functionally identical `verifySignature()` method from utxolib.

Issue: BG-34381

### DIFF
--- a/modules/core/test/v2/unit/coins/btc.ts
+++ b/modules/core/test/v2/unit/coins/btc.ts
@@ -3,6 +3,7 @@ const { Codes } = require('@bitgo/unspents');
 
 import { TestBitGo } from '../../../lib/test_bitgo';
 import { Wallet } from '../../../../src/v2/wallet';
+import { AbstractUtxoCoin } from '../../../../src/v2/coins';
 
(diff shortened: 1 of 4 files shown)


modules/core/test/v2/unit/coins/btc.ts
```diff
@@ -320,20 +321,20 @@ describe('BTC:', function () {
         },
       };
 
-      let coin;
+      let coin: AbstractUtxoCoin;
```
sangaman, line 324: I'm curious why you specified a type here if you have to cast to `any` below?

sangaman (approved):
Nice deduplification.

modules/core/test/v2/unit/coins/btc.ts
OttoAllmendinger: to get IntelliJ support :)

OttoAllmendinger: the tests where I have to cast to `any` are simulating use from untyped JS

PR #1440, opened 2021-08-24
feat(utxo-lib): add thirdparty fixtures

OttoAllmendinger:
All bitcoin core forks (so far) have static test vectors, which include
tests for valid transactions and signature hashes.

Add script to download and extract these vectors as well as some tests to
verify those vectors.

Issue: BG-34381

057cae5a2
fix(utxo-lib): write `version` as `Int32`.

It gets parsed as Int32 so we should write it as Int32 as well.

See also https://github.com/bitcoinjs/bitcoinjs-lib/pull/1717

Issue: BG-34381

2d5f7e6bf
fix(utxo-lib): fix `addForkId` evaluation

Turns out `>` has higher precedence (12) than `&`!

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

Issue: BG-34381


modules/utxo-lib/test/fixtures_thirdparty/parse.ts
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        function runCheckHashForSignature([
+          rawTransaction,
+          script,
+          inputIndex,
+          hashType,
+          signatureHash,
+        ]: SigHashTestVector) {
+          if (isBitcoinGold(network) && hashType & Transaction.SIGHASH_FORKID) {
```
tylerlevine, line 29: Nit: add parens to make precedence clear here

modules/utxo-lib/test/fixtures_thirdparty/parse.ts
```diff
@@ -0,0 +1,68 @@
...
+          assert.strict(refSignatureHash.equals(hash));
+        }
+
+        testFixtureArray(network, sigHashTestFile, function (vectors: SigHashTestVector[]) {
+          if (isZcash(network)) {
+            return this.skip();
+          }
+          vectors.forEach((v) => {
+            try {
+              runCheckHashForSignature(v);
```
tylerlevine, line 53: Why wrap in `try`/`catch` just to rethrow?

tylerlevine (dismissed):

modules/utxo-lib/test/fixtures_thirdparty/parse.ts
OttoAllmendinger: oops

(there was a console.log with the vector index)

modules/utxo-lib/test/fixtures_thirdparty/parse.ts
OttoAllmendinger: `prettier` strips them off again :(

tylerlevine (approved):

OttoAllmendinger:
lint failures for `account-lib` :(

PR #1446, opened 2021-08-25
feat(utxo-lib): increase test coverage

OttoAllmendinger:
* increase overall test coverage by comparing RPC transaction data to parsed data
* increase dash test coverage by comparing "special" transactions to parsed data

Issue: BG-34381


modules/utxo-lib/test/integration_local_rpc/compare.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  return normalizedTx;
+}
+
+export function normalizeRpcTransaction(tx: RpcTransaction, network: Network): NormalizedObject {
+  const normalizedTx: NormalizedObject = {
+    ...tx,
+    vin: tx.vin.map((v: any) => {
+      delete v.scriptSig.asm;
```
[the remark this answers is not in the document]
sangaman: In this case why not just `return { ....v, scriptSig: { hex: v.scriptSig.hex }};` ? In other words just returning only the keys you want, rather than deleting the keys you don't want.

modules/utxo-lib/test/integration_local_rpc/compare.ts
```diff
@@ -0,0 +1,118 @@
...
+      delete v.scriptPubKey.asm;
+      delete v.scriptPubKey.addresses;
+      delete v.scriptPubKey.reqSigs;
+      delete v.scriptPubKey.type;
+      delete v.valueSat;
+      if (isZcash(network)) {
+        delete v.valueZat;
+      }
+      v.value = v.value * 1e8;
+      return v;
```
sangaman, line 90: Here too could be `return {n: v.n, value: v.value * 1e8, scriptPubKey: { hex: v.scriptPubKey.hex }};` I think?

I might be missing something but if we know the specific fields we want, just returning those vs `delete`ing the ones we don't is less verbose and a little clearer I think.

modules/utxo-lib/test/integration_local_rpc/compare.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * @prettier
+ */
+import { RpcTransaction } from './generate/RpcTypes';
+import * as networks from '../../src/networks';
+import { Network } from '../../src/networkTypes';
+import { getMainnet, isZcash } from '../../src/coins';
+
+type NormalizedObject = Record<string, unknown>;
```
sangaman, line 9: Is there a reason we can't make this more specific? It seems like we should be able to define the fields we should include, and I think it'd make the `normalize` functions below easier to follow.

modules/utxo-lib/test/transaction_util.ts
```diff
@@ -0,0 +1,15 @@
...
+import * as assert from 'assert';
+import { Network } from '../src/networkTypes';
+
+const Transaction = require('../src/transaction');
+
+export function parseTransactionRoundTrip(buf: Buffer, network: Network) {
+  const tx = Transaction.fromBuffer(buf, network);
+  assert.strictEqual(tx.byteLength(), buf.length);
+  assert.strictEqual(tx.toBuffer().toString('hex'), buf.toString('hex'));
+  assert.strictEqual(tx.clone().toBuffer().toString('hex'), buf.toString('hex'));
```
sangaman, line 13: What's the point of cloning the tx and checking it again?

sangaman (approved):
Love the added test coverage and cases. I left a few optional suggestions/questions.

modules/utxo-lib/test/integration_local_rpc/compare.ts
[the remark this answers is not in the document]
OttoAllmendinger: If there is some new protocol update or new coin where inputs have a new property `foo`, I want to deal with it explicitly. 

I probably want to delete/ignore it but I might also want to compare it against the parsed object.

modules/utxo-lib/test/transaction_util.ts
OttoAllmendinger: making sure that `clone()` is implemented correctly

modules/utxo-lib/test/integration_local_rpc/compare.ts
OttoAllmendinger: The point of the tests is actually to discover new properties that might pop up in either the RPC response or the parsed object and deal with them as they appear.

If I exhaustively enumerated the tx properties, it would suggest that we know them upfront, when often we do not.   

Let's assume one of our coins makes a protocol upgrade and a new field `foo` appears: these tests will fail unless we _decide_ that either `foo` is not important (we delete it) or it needs to be parsed and exposed. 

If we were only comparing know properties that would not work.

sangaman: Thanks, I thought that might be the case but wasn't clear on it. Then this approach all makes sense.

PR #1451, opened 2021-08-26
refactor(core): use utxo.bitgo.createTransaction* wrappers

OttoAllmendinger:
Do not create Transactions(Builders) directly, use wrapper methods instead.

Issue: BG-34381

19b484171
feat(utxo-lib): add createTransactionFromHex()

Some code paths love to use hex strings so let's add a convenience method

Issue: BG-34381


sangaman (approved):
Good refactoring, I searched this branch for any lingering references to the non-wrapped transaction calls to `utxolib` and looks like they're all cleaned up.

PR #1459, opened 2021-08-30
fix(core): fix type of TransferOptions.allTokens

takuto-h:
`allTokens` must be a boolean as validated below.
https://github.com/BitGo/BitGoJS/blob/44b9ee9063f9ca71972aad3b4875fa61a6892ad8/modules/core/src/v2/wallet.ts#L731-L736


takuto-h:
Fixed `valueGte` & `valueLt` as well.

PR #1463, opened 2021-08-30
refactor(account-lib): add `getPublicKey()`/`getPrivateKey()`

OttoAllmendinger:
Reduce copypasta and low-level utxolib access:

* add method `getPublicKey({ compressed: boolean })`
* add method `getPrivateKey()` methods
* assume `keyPair` property is always set (like the type implies), and
  that it always set to `this.hdNode.keyPair`.

Issue: BG-34381


sangaman (dismissed):
Changes look good. It's not clear to me whether `getPublicKey`/`getPrivateKey` will give the same values as the old `HDNode.fromBase58...` code. But it looks like there's already quite a bit of existing testing around `getKeys` that would catch any deviations.

modules/account-lib/src/coin/stx/keyPair.ts
```diff
@@ -72,11 +73,9 @@ export class KeyPair extends Secp256k1ExtendedKeyPair {
     }
 
     if (isValidXpub(pub)) {
-      this.hdNode = HDNode.fromBase58(pub);
-      this.hdNode.keyPair.compressed = false;
```
OttoAllmendinger, line 80: the parameter `compressed` in `getKeys` has precedence

modules/account-lib/src/coin/baseCoin/secp256k1ExtendedKeyPair.ts
```diff
@@ -24,17 +33,25 @@ export abstract class Secp256k1ExtendedKeyPair implements BaseKeyPair {
     this.source = source;
   }
 
+  getPublicKey({ compressed }: { compressed: boolean }): Buffer {
+    return ECPair.fromPublicKey(this.keyPair.publicKey, { compressed }).publicKey;
+  }
+
+  getPrivateKey(): Buffer | undefined {
+    return this.keyPair.privateKey;
+  }
```
argjv, line 42: Add some JS docs explaining the format these keys are returned in.

OttoAllmendinger: done

OttoAllmendinger:
> 774	ERROR in bitgo-account-lib.min.js from Terser
> 775	Unexpected token: punc ()) [bitgo-account-lib.min.js:11377,2]

I can't reproduce this locally :(

OttoAllmendinger:
resolved compilation error

modules/account-lib/src/coin/avaxc/keyPair.ts
```diff
@@ -20,9 +20,9 @@ export class KeyPair extends Secp256k1ExtendedKeyPair {
     super(source);
```
argjv, line 20: Shouldn't this `if` go into the base class now that it looks the same for almost all coins?

sangaman: Looks like xrp doesn't have the `if` statement and `eth` has one that's a bit different. Still maybe this logic could be moved into a member function called in the constructor which the two coins above override.

sangaman (approved):
Looks good but there might be even more copypasta to remove per jorge's comment if you think it's worthwhile.

modules/account-lib/src/coin/avaxc/keyPair.ts
OttoAllmendinger: there is room for a lot more cleanup, but this PR was more about getting rid of utxolib

PR #1464, opened 2021-08-31
refactor(core): use createOutputScript2of3 in v1 wallet

OttoAllmendinger:
Issue: BG-34381


OttoAllmendinger:
some early plans maybe?

PR #1466, opened 2021-08-31
refactor(utxo-lib): extend verifySignature test

OttoAllmendinger:
We have an implicit test in `integration_local_rpc` but it is worth having
a more thorough dedicated one.

Issue: BG-34381

1e0b6bfe6
refactor(utxo-lib): improve integration_local_rpc signature test

More exhaustive `verifySignature(..., { publicKey })` test.

Issue: BG-34381

### DIFF
--- a/modules/utxo-lib/test/bitgo/signature.ts
+++ b/modules/utxo-lib/test/bitgo/signature.ts
@@ -1,0 +1,130 @@
+/**
+ * @prettier
+ */
+import * as assert from 'assert';
+import * as bip32 from 'bip32';
+
+import networks = require('../../src/networks');
+import { createOutputScript2of3, ScriptType2Of3, scriptTypes2Of3 } from '../../src/bitgo/outputScripts';
+import { getNetworkList, getNetworkName, isBitcoin, isMainnet } from '../../src/coins';
+import { Network } from '../../src/networkTypes';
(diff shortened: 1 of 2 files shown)


modules/utxo-lib/test/bitgo/signature.ts
```diff
@@ -0,0 +1,130 @@
...
+    });
+  });
+}
+
+describe('signature', function () {
+  getNetworkList()
+    .filter(isMainnet)
+    // The signing and verification methods are largely network-independent so let's focus on a
+    // single network to reduce test time.
+    // During development it might make sense to test all networks.
```
sangaman, line 125: I'm a bit confused by this comment since as I see it development is ongoing and the tests are here to support development. When would you envision it being worth testing every network? And would we do that by just editing this test to drop the `filter` statements above?

sangaman (approved):

modules/utxo-lib/test/bitgo/signature.ts
OttoAllmendinger: My idea is to uncomment the `isBitcoin` filter below when touching code that could alter the behavior for other coins. 

When I run this for all coins locally, it takes ~20secs which more then doubles the overall test time!

PR #1467, opened 2021-08-31
feat(core): add enable and disable token txs to explain transaction method

shihao-guo:
explain enable and disable token transactions

stlx-6324

https://bitgoinc.atlassian.net/browse/STLX-6324


CapnMigraine (approved):
lgtm

PR #1476, opened 2021-09-01
refactor(utxo-lib): simplify hashForZcashSignature

OttoAllmendinger:
Reduce indentation

Issue: BG-34381

ce24cc8ef
refactor(utxo-lib): only support transparent inputs

Issue: BG-34381


sangaman (approved):
Is this also related to BG-35119 for dropping support for shielded transactions?

OttoAllmendinger:
yup this removes a remnant of unshielded transaction "support"

PR #1477, opened 2021-09-01
feat(statics): add new tokens

facundoCarreSimtlix:
add new tokens

stlx-5659


argjv (approved):
Contracts look good in etherscan

PR #1478, opened 2021-09-01
feat(account-lib): determine how to use contract method IDs

nachofm:
All avax-c coin-sandbox scripts passing on AvaxWalletSimple.sol contract with the latest account-lib
version "2.16.0-rc.1.1.0". 
Method IDs hashes are equal between eth-multisig-v2/contracts/WalletSimple.sol and eth-multisig-v2/contracts/coins/AvaxWalletSimple.sol, the constants are deleted.  Method IDs hashes are different between
eth-multisig-v4/contracts/WalletSimple.sol and eth-multisig-v2/contracts/WalletSimple.sol since one
uses uint256 ad the other uses uint but since they are aliases they work the same way. 
More info on the comparisons here:
https://docs.google.com/spreadsheets/d/1VcM5UVwfmjxnGHy-hfhoBNguO8vIaVWq2cjXSwoMKGA/edit?usp=sharing


smaria95 (approved):
LGTM

CapnMigraine (approved):
lgtm

PR #1483, opened 2021-09-02
fix(core): fix xpubToEthAddress

OttoAllmendinger:
Issue: BG-34381

c8713ab1a
chore(test): add test xpubToEthAddress

Issue: BG-35804

### DIFF
--- a/modules/core/src/v2/internal/util.ts
+++ b/modules/core/src/v2/internal/util.ts
@@ -118,7 +118,7 @@ export class Util {
     if (!isEthAvailable) {
       throw new EthereumLibraryUnavailableError(ethImport);
     }
-    return ethUtil.bufferToHex(ethUtil.publicToAddress(bip32.fromBase58(xpub).publicKey, false));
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/internal/util.ts
```diff
@@ -118,7 +118,7 @@ export class Util {
     if (!isEthAvailable) {
       throw new EthereumLibraryUnavailableError(ethImport);
     }
-    return ethUtil.bufferToHex(ethUtil.publicToAddress(bip32.fromBase58(xpub).publicKey, false));
+    return ethUtil.bufferToHex(ethUtil.publicToAddress(bip32.fromBase58(xpub).publicKey, true /* sanitize */));
```
sangaman, line 121: What's the "sanitize" comment mean?

sangaman (approved):
Confirmed that the new test fails without the fix and passes with it.

modules/core/src/v2/internal/util.ts
OttoAllmendinger: that's the name of the function argument

PR #1484, opened 2021-09-02
refactor(utxo-lib): add address tests, fixtures

OttoAllmendinger:
We will have to change some address formatting/parsing code and should have
some tests ready for that.

Issue: BG-32852

204e4044b
fix(utxo-lib): fix imports in test

Concurrent merge oopsie

Issue: BG-32852


modules/utxo-lib/test/address/address.ts
```diff
@@ -0,0 +1,67 @@
...
+          .reduce((all: AddressTestVector[], v) => [...all, ...v], []);
+
+        refVectors = await readFixture(network, vectors);
+      });
+
+      it('matches test vectors, parses to scriptPubKeyHex', function () {
+        assert.strictEqual(vectors.length, refVectors.length);
+        vectors.forEach((v, i) => {
+          assert.deepStrictEqual(v, refVectors[i]);
+          const [scriptType, scriptPubKeyHex, address] = v;
```
sangaman, line 61: Is `scriptType` deliberately unused here?

sangaman (approved):
Very nice test suite, will be useful in ensuring the functionality isn't impacted by future refactoring.

modules/utxo-lib/test/address/address.ts
OttoAllmendinger: yes

I could also write this as `[, scriptPubKeyHex, address] = v;` if you think it is cleaner

sangaman: It's not a big deal if you're not going to change the PR for other reasons. Mainly wanted to make sure that something wasn't being overlooked.

PR #1489, opened 2021-09-06
chore(core): move utxo recovery fixtures to json files

OttoAllmendinger:
Reduce inlined txhex which are a pain to update

Issue: BG-34381

----

Also some general code cleanup


modules/core/test/v2/unit/recovery/utxo.ts
```diff
@@ -0,0 +1,587 @@
...
+    afterEach(() => {
+      sandbox.restore();
+    });
+
+    it('should generate BTC recovery tx', async function () {
+      sandbox.stub(Btc.prototype, 'verifyRecoveryTransaction').resolves(btcNonKrsRecoveryDecodedTx.transaction);
+      const basecoin = bitgo.coin('tbtc');
+      const recovery = await basecoin.recover({
+        userKey:
+          '{"iv":"fTcRIg7nlCf9fPSR4ID8XQ==","v":1,"iter":10000,"ks":256,"ts":64,"mode":"ccm","adata":"","cipher":"aes","salt":"pkIS5jVDi0Y=","ct":"SJQgP+ZzfOMf2fWxyQ2jpoWYioq6Tqfcw1xiKS1WpWAxLvXfH059sZvPrrYMdijJEbqA8EEaYXWmdgYSkMXdwckRMyvM3uWl9H8iKw1ZJmHyy2eDSy5r/pCtWICkcO3oi2I492I/3Op2YLfIX6XqKWs2mztu/OY="}',
```
sangaman, line 73: This `userKey` gets used several times, maybe declare it once as a variable and reuse it?

Looks like it's the same case for `backupKey`, `bitgoKey`, and `recoveryDestination`.

sangaman (approved):
Just made one suggestion about putting string literals into variables for reuse, but also ok to leave it as is if it makes fixture regeneration simpler.

modules/core/test/v2/unit/recovery/utxo.ts
OttoAllmendinger: yup there is a bunch more we could and should do (loop over coins actually)

I didn't want to go down that rabbithole all the way

PR #1491, opened 2021-09-07
fix(core): Implement verifyTransaction method for Eos

shiksb:
Ticket: [STLX-5967](https://bitgoinc.atlassian.net/browse/STLX-5967)


modules/core/src/v2/coins/eos.ts
```diff
@@ -871,8 +882,52 @@ export class Eos extends BaseCoin {
...
-    return Bluebird.resolve(true).asCallback(callback);
+  /**
+   * Verify that a transaction prebuild complies with the original intention
+   *
+   * @param params
+   * @param params.txParams params used to build the transaction
+   * @param params.txPrebuild the prebuilt transaction
+   * @param callback
+   */
+  verifyTransaction(params: EosVerifyTransactionOptions, callback?: NodeCallback<boolean>): Bluebird<boolean> {
```
pancham88, line 893: This shall also include validation for amount and memo fields, as here:
https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/v2/coins/eos.js#L679

pancham88 (changes requested):

shihao-guo:
Posting Matt's response


```
1.  The eos.verifyTransaction validates txPrebuild.transaction.packed_trx, while eos.signTransaction signs txPrebuild.txHex. There is nothing stopping an attacker from specifying a legitimate txPrebuild.transaction.packed_trx, but a malicious txPrebuild.txHex. This would bypass the protections, as the legitimate packed_trx would pass the verification, but then the malicious txHex would be signed and returned to the BitGo server (and the attacker).
In order for the verification to prevent malicious transactions from being signed, it must deserialize the content that will be signed and validate it matches the requested parameters.
2.  The verification in eos.verifyTransaction is not complete:
It does not validate that the transaction’s action account/name and data match the expected values. Thus, the transaction could be for any smart contract.
txParams.txPrebuild isn’t always set, and can’t be set if txParams.recepients is set (which is the case for wallet.send). Thus, the header check seems like it will always fail when called from wallet.send. The unit test for EOS transaction verification sets txParam.txPrebuild to the txPrebuild being verified. In that scenario, the user will have passed in the txPrebuild parameter, and thus it would not need to be verified. Instead, transaction verification should focus on validating transactions returned from the server (via wallet.prebuildTransaction) match the requested transaction instead.
```

modules/core/package.json
```diff
@@ -68,6 +68,7 @@
     "ecurve": "^1.0.6",
     "eol": "^0.5.0",
     "eosjs": "^16.0.8",
+    "eosjs-latest": "npm:eosjs@^21.0.2",
```
pancham88, line 71: Why are both 21.02 and 16.x are used? I will be very skeptical of releasing this code in production which uses 2 different versions.

modules/core/src/v2/coins/eos.ts
```diff
@@ -9,25 +9,31 @@ import {
...
   VerifyAddressOptions as BaseVerifyAddressOptions,
   HalfSignedAccountTransaction as BaseHalfSignedTransaction,
   SignTransactionOptions as BaseSignTransactionOptions,
+  VerificationOptions,
+  VerifyTransactionOptions as BaseVerifyTransactionOptions,
 } from '../baseCoin';
 import { NodeCallback } from '../types';
 import { BigNumber } from 'bignumber.js';
 import { randomBytes } from 'crypto';
+// TODO BG-36375: remove dependency on eosjs v16
```
pancham88, line 21: Please see comment above. I think this shall be resolved before releasing into production.

pancham88 (changes requested):

modules/core/package.json
shiksb: eosjs v21 [doesn't give us a transaction_id](https://github.com/BitGo/BitGoJS/pull/1491/files?authenticity_token=nwuvBW71B%2FnfL%2FHVWhhS4hVYcAjuaS5otOZS6PpB9AxM0XV2vX%2Bl2ILMffSVQJ9ZrymGM5mMpneQ7FXNzMNvjg%3D%3D&file-filters%5B%5D=.ts#diff-6efbc825de1f8d6f3fae2bfff41cf42cc40beae8d0af3309e52e11582ab83d9fR579-R583) when we create a transaction since it's not supported anymore. I couldn't find any way to create transaction_id in v21. 

v16 is only being used in 2 places, I was able to refactor all other usages to v21.

modules/core/src/v2/coins/eos.ts
```diff
@@ -458,17 +484,20 @@ export class Eos extends BaseCoin {
...
+        textDecoder: new TextDecoder(),
+        textEncoder: new TextEncoder(),
+      });
 
-      // Get tx base values
-      const eosTxStruct = eosClient.fc.structs.transaction;
       const serializedTxBuffer = Buffer.from(transaction.packed_trx, 'hex');
-      const tx = EosJs.modules.Fcbuffer.fromBuffer(eosTxStruct, serializedTxBuffer);
-
+      const deserializedTxJsonFromPackedTrx = yield api.deserializeTransactionWithActions(serializedTxBuffer);
```
joey-bai, line 496: `deserializeTransactionWithActions` expects a String or Uint8Array as input. Passing in a buffer will throw an error.

confirmed this will work by passing in `packed_trx` directly.
`const deserializedTxJsonFromPackedTrx = yield api.deserializeTransactionWithActions(transaction.packed_trx);`

modules/core/src/v2/coins/eosutil/eosabiprovider.ts
```diff
@@ -0,0 +1,32 @@
...
+    };
+    if (account === 'eosio') {
+      ret.abi = eosioAbi;
+    } else if (account === 'eosio.token') {
+      ret.abi = eosioTokenAbi;
+    } else if (account === 'eosio.null') {
+      ret.abi = eosioNullAbi;
+    } else {
+      throw new Error(`${account} is not supported by OfflineAbiProvider`);
+    }
```
joey-bai, line 27: let's define a dictionary for supported accounts.
something like:
```
SUPPORTED_ABIS = {
    'eosio': ...,
    'eosio.token': ...,
    ...
}
```

error is thrown when `account` is not present in the dict.
this will help us trim down the if-else checks as we support new ABIs (which we will need for EOS tokens)

modules/core/package.json
joey-bai: @shiksb I ran into this problem for tokens and saw that transactionId is a derived value:
From EOS [docs](https://developers.eos.io/welcome/v2.0/protocol/transactions_protocol/#21-transaction-id):
> Consequently, a transaction ID consists of a cryptographic hash of the basic fields included in a transaction instance. Therefore, the transaction ID is solely determined by the list of actions encapsulated within the transaction, the transaction header, and any embedded transaction extensions, which are optional.

We can see how transaction_id is generated by looking at v16's source code for the `transaction` func: https://github.com/EOSIO/eosjs/blob/v16.0.8/src/write-api.js#L496

I verified with the following in `deserializeTransaction`:
```
      if (headers) {
        const rebuiltTransaction = yield api.transact({ ...tx, ...headers }, { sign: false, broadcast: false });
        const transactionId = crypto
          .createHash('sha256')
          .update((rebuiltTransaction as any).serializedTransaction)
          .digest()
          .toString('hex');
        tx.transaction_id = transactionId;
      }
```

modules/core/src/v2/coins/eos.ts
```diff
@@ -9,26 +9,29 @@ import {
   KeyPair,
   ParseTransactionOptions,
   ParsedTransaction,
-  VerifyTransactionOptions,
   VerifyAddressOptions as BaseVerifyAddressOptions,
   HalfSignedAccountTransaction as BaseHalfSignedTransaction,
   SignTransactionOptions as BaseSignTransactionOptions,
+  VerificationOptions,
```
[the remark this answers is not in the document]
pancham88: No, just keep `VerifyAddressOptions`, `VerificationOptions` and `VerifyTransactionOptions` together. This is just a nit comment

modules/core/src/v2/coins/eos.ts
```diff
@@ -822,23 +886,32 @@ export class Eos extends BaseCoin {
...
-        txid: (transaction as any).transaction_id,
+        txid: transactionId,
         recoveryAmount: accountBalance,
         coin: self.getChain(),
         txHex: '',
       };
       const signableTx = Buffer.concat([
         Buffer.from(self.getChainId(), 'hex'), // The ChainID representing the chain that we are on
-        Buffer.from(serializedTransaction, 'hex'), // The serialized unsigned tx
+        Buffer.from(serializedTransactionHex, 'hex'), // The serialized unsigned tx
```
tylerlevine, line 914: seems a bit strange to convert to a buffer, then to a hex string, then use the hex string, then later convert that hex string back to a buffer 🤔

modules/core/src/v2/coins/eos.ts
```diff
@@ -813,7 +858,26 @@ export class Eos extends BaseCoin {
...
 
       const transactionHeaders = yield self.getTransactionHeadersFromNode();
-      const eosClient = new EosJs({ chainId: self.getChainId(), transactionHeaders });
+      if (!transactionHeaders) {
+        throw new Error('Could not get transaction headers from node');
+      }
+      const headers: EosTransactionHeaders = transactionHeaders;
+
+      // drop milliseconds and trailing Z from expiration
+      const date = new Date(headers.expiration as string);
```
tylerlevine, line 867: please be very defensive when it comes to date manipulations and I'd recommend even using the `moment` library for this.

The behavior of `Date.prototype.toString` differs between nodejs and browsers:

browser:
```
new Date()
Thu Oct 14 2021 14:08:10 GMT-0700 (Pacific Daylight Time)
```

node:
```
> new Date()
2021-10-14T21:10:09.938Z
```

modules/core/src/v2/coins/eosutil/eosabiprovider.ts
```diff
@@ -0,0 +1,36 @@
+const { Numeric } = require('eosjs');
```
tylerlevine, line 1: use `import` here if possible

modules/core/src/v2/coins/eosutil/eosabiprovider.ts
```diff
@@ -0,0 +1,36 @@
...
+  kvszn1xyz1bu: 'DmVvc2lvOjphYmkvMS4yAAgHYWNjb3VudAABB2JhbGFuY2UFYXNzZXQFY2xvc2UAAgVvd25lcgRuYW1lBnN5bWJvbAZzeW1ib2wGY3JlYXRlAAIGaXNzdWVyBG5hbWUObWF4aW11bV9zdXBwbHkFYXNzZXQOY3VycmVuY3lfc3RhdHMAAwZzdXBwbHkFYXNzZXQKbWF4X3N1cHBseQVhc3NldAZpc3N1ZXIEbmFtZQVpc3N1ZQADAnRvBG5hbWUIcXVhbnRpdHkFYXNzZXQEbWVtbwZzdHJpbmcEb3BlbgADBW93bmVyBG5hbWUGc3ltYm9sBnN5bWJvbAlyYW1fcGF5ZXIEbmFtZQZyZXRpcmUAAghxdWFudGl0eQVhc3NldARtZW1vBnN0cmluZwh0cmFuc2ZlcgAEBGZyb20EbmFtZQJ0bwRuYW1lCHF1YW50aXR5BWFzc2V0BG1lbW8Gc3RyaW5nBgAAAAAAhWlEBWNsb3NlAAAAAACobNRFBmNyZWF0ZQAAAAAAAKUxdgVpc3N1ZQAAAAAAADBVpQRvcGVuAAAAAACo67K6BnJldGlyZQAAAABXLTzNzQh0cmFuc2ZlcgACAAAAOE9NETIDaTY0AAAHYWNjb3VudAAAAAAAkE3GA2k2NAAADmN1cnJlbmN5X3N0YXRzAAAAAA==',
+};
+
+interface EosRawAbi {
+  accountName: string;
+  abi: Uint8Array
+}
+
+export class OfflineAbiProvider {
+  getRawAbi(account: string): EosRawAbi {
```
tylerlevine, line 23: would be good to make this `async` if possible - will make removing the ABI data from the source code in the future much easier (we should really have these defined in files which get read in at runtime)

modules/core/test/v2/unit/coins/eos.ts
```diff
@@ -212,4 +215,133 @@ describe('EOS:', function () {
       explainedTx.memo.should.equal('1');
     });
   });
+
+  describe('Transaction Verification', function () {
+    let wallet;
+    let basecoin;
+
+    before(co(function *() {
```
tylerlevine, line 223: `async` instead of coroutines for new code please

modules/core/test/v2/unit/coins/eos.ts
```diff
@@ -212,4 +215,133 @@ describe('EOS:', function () {
       explainedTx.memo.should.equal('1');
     });
   });
+
+  describe('Transaction Verification', function () {
```
tylerlevine, line 219: Where are the tests for the failure cases? This seems like only a happy path test.

modules/core/src/v2/coins/eosutil/eosabiprovider.ts
tylerlevine: Also, please add `@prettier` annotation to new files.

modules/core/src/v2/coins/eosutil/eosabiprovider.ts
```diff
@@ -0,0 +1,36 @@
...
+  kvszn1xyz1bu: 'DmVvc2lvOjphYmkvMS4yAAgHYWNjb3VudAABB2JhbGFuY2UFYXNzZXQFY2xvc2UAAgVvd25lcgRuYW1lBnN5bWJvbAZzeW1ib2wGY3JlYXRlAAIGaXNzdWVyBG5hbWUObWF4aW11bV9zdXBwbHkFYXNzZXQOY3VycmVuY3lfc3RhdHMAAwZzdXBwbHkFYXNzZXQKbWF4X3N1cHBseQVhc3NldAZpc3N1ZXIEbmFtZQVpc3N1ZQADAnRvBG5hbWUIcXVhbnRpdHkFYXNzZXQEbWVtbwZzdHJpbmcEb3BlbgADBW93bmVyBG5hbWUGc3ltYm9sBnN5bWJvbAlyYW1fcGF5ZXIEbmFtZQZyZXRpcmUAAghxdWFudGl0eQVhc3NldARtZW1vBnN0cmluZwh0cmFuc2ZlcgAEBGZyb20EbmFtZQJ0bwRuYW1lCHF1YW50aXR5BWFzc2V0BG1lbW8Gc3RyaW5nBgAAAAAAhWlEBWNsb3NlAAAAAACobNRFBmNyZWF0ZQAAAAAAAKUxdgVpc3N1ZQAAAAAAADBVpQRvcGVuAAAAAACo67K6BnJldGlyZQAAAABXLTzNzQh0cmFuc2ZlcgACAAAAOE9NETIDaTY0AAAHYWNjb3VudAAAAAAAkE3GA2k2NAAADmN1cnJlbmN5X3N0YXRzAAAAAA==',
+};
+
+interface EosRawAbi {
+  accountName: string;
+  abi: Uint8Array
+}
+
+export class OfflineAbiProvider {
```
tylerlevine, line 22: Can you have this implement `ApiInterfaces.AbiProvider`?

tylerlevine (changes requested):
good improvement but still needs some additional work + failure case tests

modules/core/test/v2/unit/coins/eos.ts
shiksb: added failure cases

modules/core/src/v2/coins/eos.ts
tylerlevine: is still an issue?

PR #1493, opened 2021-09-07
fix(core): use `buildIncomplete()` in utxo recovery

OttoAllmendinger:
Since we are only adding the first of two signatures, we need to use
`buildIncomplete()`.

Tested with https://github.com/BitGo/key-recovery-service-v2

Issue: BG-34381


OttoAllmendinger:
the txhex changes are the `OP_0` placeholder signatures

sangaman (approved):

PR #1496, opened 2021-09-08
feat(wp) : added support of cashaddr for create address using SDK

mukeshsp:
Jira Ticket : https://bitgoinc.atlassian.net/browse/BG-35555

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -264,6 +264,7 @@ export interface CreateAddressOptions {
   label?: string;
   lowPriority?: boolean;
   forwarderVersion?: number;
+  format?: 'base58' | 'cashaddr';
 }
 
 export interface UpdateAddressOptions {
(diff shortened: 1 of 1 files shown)


modules/core/src/v2/wallet.ts
```diff
@@ -264,6 +264,7 @@ export interface CreateAddressOptions {
   label?: string;
   lowPriority?: boolean;
   forwarderVersion?: number;
+  format?: string;
```
tylerlevine, line 267:
```suggestion
  format?: 'base58' | 'cashaddr';
```

tylerlevine (commented):
looks good to me, just a slightly tighter type on the new `format` property would be nice

tylerlevine (changes requested):
forgot to request changes, see inline comment

modules/core/src/v2/wallet.ts
mukeshsp: updated the changes as suggested.

tylerlevine (approved):

PR #1498, opened 2021-09-08
Merge tokens master

JulioAchaval-SimTLiX:


tylerlevine (changes requested):
why does this include commits already merged into master?

JulioAchaval-SimTLiX:
> why does this include commits already merged into master?

When a "revert" is performed on master, the branch cannot be overridden normally because it removes the changes from the "feature branch". That is why you have to take the missing master changes and integrate them into the "feature branch" regardless of the commit that removes these changes. To do this, we must create an intermediate branch (This is the branch) and we made a PR to make sure that it passes all the necessary controls before incorporating said code to the "feature branch".
This branch will never be merged to master.

PR #1499, opened 2021-09-09
Include test case and base chain for algo tokens.

0xabrarhussain:
Algo tokens don't seem to be working correctly on bitgo-js. Inside `algo.ts` we call `explainTransaction` which uses the following code:

```
      const factory = accountLib.getBuilder(this.getChain()) as unknown as accountLib.Algo.TransactionBuilderFactory;

      const tx = (yield txBuilder.build()) as any;
      const txJson = tx.toJson();
```

The builder doesn't exist for the given chain.

image

Solution I imagine would be to update the value returned from `getChain` from algo tokens to reference coin instead, but not sure if there's a convention for how these are handled.

### DIFF
--- a/modules/core/src/v2/coins/algoToken.ts
+++ b/modules/core/src/v2/coins/algoToken.ts
@@ -71,6 +71,10 @@ export class AlgoToken extends Algo {
     return this.tokenConfig.type;
   }
 
+  getBaseChain() {
+    return this.coin;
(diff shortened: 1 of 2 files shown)


0xabrarhussain:
closing as duplicate of https://github.com/BitGo/BitGoJS/pull/1502

PR #1500, opened 2021-09-09
feat(core): expose compatibility layer at `require('bitgo').bitcoin`

OttoAllmendinger:
For `client-wallet-platform`, we previously exposed the full utxolib
API.

However a small subset is enough:

* `makeRandomKey()`
* `HDNode`
* `hdPath()`
* `address.fromBase58Check()`

Since some of these methods are being deprecated, expose them as
compatibility wrappers around `bip32`.

Issue: BG-34381


modules/core/src/legacyBitcoin.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  get privateKey(): Buffer | undefined {
+    return this.bip32.privateKey;
+  }
+
+  public static fromSeedBuffer(buf: Buffer): HDNode {
+    return new HDNode(bip32.fromSeed(buf));
+  }
+
+  public static formBase58(str: string): HDNode {
```
tylerlevine, line 34:
```suggestion
  public static fromBase58(str: string): HDNode {
```

tylerlevine (changes requested):
looks good, small change needed. Do we have any tests to ensure compatibility before removing the old code?

sangaman (approved):

tylerlevine (approved):

PR #1502, opened 2021-09-09
Fix algo tokens builder step by using base chain.

0xabrarhussain:
Algo tokens don't seem to be working correctly on bitgo-js. Inside algo.ts we call explainTransaction which uses the following code:

      const factory = accountLib.getBuilder(this.getChain()) as unknown as accountLib.Algo.TransactionBuilderFactory;

      const tx = (yield txBuilder.build()) as any;
      const txJson = tx.toJson();

image

Updating this code so we use the base chain instead of the given chain.


rliberal-simtlix (approved):
LGTM

rliberal-simtlix (approved):

JulioAchaval-SimTLiX (approved):

pancham88 (approved):

PR #1513, opened 2021-09-14
refactor(statics): move trx fee limit to network object

silalvado:
Move trx fee limit to network object

Jira ticket
https://bitgoinc.atlassian.net/browse/STLX-6770


ipelaez-simtlix (approved):
lgtm

modules/statics/src/account.ts
```diff
@@ -558,15 +553,16 @@ export function tronToken(
   features: CoinFeature[] = AccountCoin.DEFAULT_FEATURES,
   prefix: string = '',
   suffix: string = name.toUpperCase(),
-  network: AccountNetwork = Networks.main.trx,
-  primaryKeyCurve: KeyCurve = KeyCurve.Secp256k1,
-  feeLimit: string
+  network: TronNetwork = Networks.main.trx,
+  primaryKeyCurve: KeyCurve = KeyCurve.Secp256k1
 ) {
+  const tokenNetwork = { ...network };
```
tylerlevine, line 559: We can't do this - we rely on the coin network being the same object instance as those exported by the library

tylerlevine (changes requested):

tylerlevine:
ok, let's get to the bottom of this. Is the `feeLimit` a property of the Tron network, or is it a property of coins which exist on the Tron network and could be different for different asset types? I talked with @twtaylorbitgo about this and got the impression that it's a network limit and not a limit that changes based on the asset type.

modules/statics/src/account.ts
silalvado: We have added another field to the network object for the contract call default fee limit. Both values are required to be on Statics.

tylerlevine (approved):

PR #1514, opened 2021-09-14
feat(account lib): Get trx fee limit from statics

silalvado:
get trx fee limit from statics

Jira Ticket:
https://bitgoinc.atlassian.net/browse/STLX-6526

### DIFF
--- a/modules/account-lib/src/coin/trx/contractCallBuilder.ts
+++ b/modules/account-lib/src/coin/trx/contractCallBuilder.ts
@@ -1,5 +1,5 @@
 import { createHash } from 'crypto';
-import { BaseCoin as CoinConfig } from '@bitgo/statics/';
+import { BaseCoin as CoinConfig, TronNetwork } from '@bitgo/statics/';
 import BigNumber from 'bignumber.js';
 import { TransactionType } from '../baseCoin';
 import { protocol } from '../../../resources/trx/protobuf/tron';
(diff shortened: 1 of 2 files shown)


ipelaez-simtlix (dismissed):

tylerlevine (changes requested):
please rebase

ipelaez-simtlix (dismissed):

modules/statics/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bitgo/statics",
-  "version": "6.12.0",
+  "version": "6.13.0-rc.9",
```
silalvado, line 3: I've updated the version here because of the check-versions script running on the drone

modules/core/package.json
```diff
@@ -47,7 +47,7 @@
   },
   "dependencies": {
     "@bitgo/account-lib": "^2.15.0",
-    "@bitgo/statics": "^6.12.0",
+    "@bitgo/statics": "^6.13.0-rc.9",
```
silalvado, line 50: I've updated the version here because of the check-versions script running on the drone

ipelaez-simtlix (commented):
lgtm

ipelaez-simtlix (approved):

emiacc (approved):

tylerlevine (changes requested):
do not update the versions of internal packages please - this should only be done during the release process

emiacc:
> do not update the versions of internal packages please - this should only be done during the release process

@tylerlevine 
we need to use the new version of `statics` in `account-lib`
it's because we changed the dependency version in the package.json of `account-lib`
and then realized that the script of check-version failed. 

What is the correct process to do that? should we reference to the old version until a release?

emiacc (approved):

tylerlevine (approved):

PR #1517, opened 2021-09-15
chore: remove node 10 from list of supported versions

ericcrosson-bitgo:
Closes Ticket: BG-36346

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -230,13 +230,13 @@ local CheckPreconditions(version) = {
   ]
 };
 
-local UnitVersions = ["10", "12", "14"];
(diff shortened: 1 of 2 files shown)


.drone.jsonnet
```diff
@@ -230,13 +230,13 @@ local CheckPreconditions(version) = {
   ]
 };
 
-local UnitVersions = ["10", "12", "14"];
+local UnitVersions = ["12", "14"];
```
tylerlevine, line 233: We should probably add 16 soon, but let's save that for another PR

tylerlevine (approved):

PR #1520, opened 2021-09-15
feat(core) : added support for send on cashaddr

mukeshsp:
Jira Ticket : https://bitgoinc.atlassian.net/browse/BG-35557

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -528,7 +528,10 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       const allOutputs = [...explanation.outputs, ...explanation.changeOutputs];
 
       // verify that each recipient from txParams has their own output
-      const expectedOutputs = _.get(txParams, 'recipients', [] as TransactionRecipient[]);
+      const expectedOutputs = _.get(txParams, 'recipients', [] as TransactionRecipient[]).map((output) => {
+        return { ...output, address: this.canonicalAddress(output.address) };
+      });
+
       const missingOutputs = AbstractUtxoCoin.findMissingOutputs(expectedOutputs, allOutputs);
 
       // get the keychains from the custom change wallet if needed
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -354,6 +354,14 @@ export abstract class BaseCoin {
   abstract verifyAddress(params: VerifyAddressOptions): boolean;
 
   /**
+   * convert address into desired address format.
+   * @param address
+   */
+  canonicalAddress(address: string): string {
+    return address;
+  }
+
+  /**
    * Check whether a coin supports blockTarget for transactions to be included in
    * @returns {boolean}
    */


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -434,14 +435,26 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+  protected static findMissingOutputs(wallet: Wallet, expectedOutputs: Output[], actualOutputs: Output[]): Output[] {
     const keyFunc = ({ address, amount }: Output): string => `${address}:${Number(amount)}`;
     const groupedOutputs = _.groupBy(expectedOutputs, keyFunc);
 
     actualOutputs.forEach((output) => {
       const group = groupedOutputs[keyFunc(output)];
       if (group) {
         group.pop();
+      } else if (wallet.baseCoin.getFamily() === 'bch') {
+        const cashAddress = (wallet.baseCoin as Bch).canonicalAddress(output.address, 'cashaddr');
```
OttoAllmendinger, line 447: are the addresses for `actualOutputs` always supposed to be in canonical form?

if so, please call `findMissingOutputs()` with `actualOutputs` mapped to the canonical address format

if you do that, the code changes for this method are not required

OttoAllmendinger (changes requested):
please change the PR to use `feat(core)` and remove the use of the word "token" as it has a specific meaning that's unrelated to this change

modules/core/src/v2/coins/abstractUtxoCoin.ts
mukeshsp: it is not always be in cashaddr format, if user can pass base58 or cashaddr. So if user pass the base58 that will be picked up by the first if condition or for cashaddr we need to convert base58 to cashaddr and then check the condition.
Also we are updating expectedOutputs address with cashaddr so we can reuse for other function call.

OttoAllmendinger: It is much simpler if we convert the addresses on both sides of the equation to the canonical format first.

OttoAllmendinger: As far as I understand, `expectedOutputs` will always have `base58` addresses and `actualOutputs` will have either cashaddr or base58. If you pass `expectedOutputs` with addresses in canonical form this method is much simpler.

mukeshsp: got your point.  I updated the code as expectedOutputs can have cashaddr and base58 so we are converting into base58 for handling the flow.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -529,6 +530,12 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
 
       // verify that each recipient from txParams has their own output
       const expectedOutputs = _.get(txParams, 'recipients', [] as TransactionRecipient[]);
+      if (wallet.baseCoin.getFamily() === 'bch') {
+        expectedOutputs.forEach((output) => {
+          const baseAddress = (wallet.baseCoin as Bch).canonicalAddress(output.address);
+          output.address = baseAddress;
+        });
+      }
```
OttoAllmendinger, line 538: 1. You can use `this.getFamily()` and `this.canonicalAddress()` instead
2. You should not mutate the object. Instead do this

```
const expectedOutputs = _.get(...).map((output) => {
  if (!isBch) {
    return output;
  }
  return { ...output, address: (this as Bch).canonicalAddress(output.address) };
});
```

OttoAllmendinger: we should actually add this method to AbstractBaseCoin
```
canonicalAddress(address: string): string {
  return address;
}
```

then you can skip the `as BCH` casts and the conditional

mukeshsp: yes, Thanks for suggestion. I updated the all the changes as suggested.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -528,7 +528,13 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       const allOutputs = [...explanation.outputs, ...explanation.changeOutputs];
 
       // verify that each recipient from txParams has their own output
-      const expectedOutputs = _.get(txParams, 'recipients', [] as TransactionRecipient[]);
+      const expectedOutputs = _.get(txParams, 'recipients', [] as TransactionRecipient[]).map((output) => {
+        if (!output.address.startsWith('bitcoincash')) {
+          return output;
+        }
```
OttoAllmendinger, line 534: You can remove this. The `canonicalAddress` will work on any bch address.

OttoAllmendinger (changes requested):

modules/core/src/v2/coins/abstractUtxoCoin.ts
mukeshsp: yes, updated this one too.

OttoAllmendinger (approved):
please wait for review by @tylerlevine

tylerlevine (approved):

PR #1521, opened 2021-09-16
feat(utxolib): implement padInputScript for p2wsh transactions

OttoAllmendinger:
Turns out we need this after all.

Issue: BG-36314


modules/utxo-lib/src/bitgo/nonStandardHalfSigned.ts
```diff
@@ -37,9 +29,12 @@ export function padInputScript(input: Input, signatureIndex: number): void {
   }
 
   const [op0, signatureBuffer, sigScript] = decompiledSigScript;
-
   if (op0 !== opcodes.OP_0 && !(Buffer.isBuffer(op0) && op0.length === 0)) {
-    throw new Error(`unexpected instruction`);
+    return;
```
sangaman, line 33: It's unclear to me when this method is supposed to return or throw an exception. Maybe you could add comments here on in the method comment block with the expected behavior.

sangaman (approved):
The p2wsh change looks good, I'm a bit confused about when this should be `return`ing without padding anything vs throwing an exception, tho.

modules/utxo-lib/src/bitgo/nonStandardHalfSigned.ts
OttoAllmendinger: I changed my mind here and decided that it should not throw an exception when it hits a weird input that it doesn't recognize.

PR #1524, opened 2021-09-17
Stlx 6839 fix decode implenetation in sdk multisigtransactions

Krin2:
https://bitgoinc.atlassian.net/browse/STLX-6839


modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -580,6 +583,18 @@ export class Utils implements BaseUtils {
       sk: keys.secretKey,
     };
   }
+
+  getTxID(txBase64: string): string {
```
rliberal-simtlix, line 587: In this case I think getTxId is not the best name because we will only use this util for doubleSpend @pancham88 may be here we need another proper name for this method. Also please add js doc.

modules/account-lib/test/unit/coin/algo/utils.ts
```diff
@@ -335,4 +336,19 @@ describe('utils', () => {
     const size = Algo.algoUtils.getTransactionByteSize(keyregTxn);
     should.equal(size, 320);
   });
+
+  it('Should be able to get a txID from a multising Tx', () => {
```
rliberal-simtlix, line 340: Please try to do a failure case like an a not multisig tx, the enconding object for multisig should be fail, just to be clear that this txId is only for multisig tx to bring support for double spend.

rliberal-simtlix (changes requested):

modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -580,6 +583,18 @@ export class Utils implements BaseUtils {
       sk: keys.secretKey,
     };
   }
+
```
rliberal-simtlix, line 586: Add js doc please with a further explanation why we use this method

rliberal-simtlix (changes requested):

smaria95 (approved):
lgtm

rliberal-simtlix (approved):

JulioAchaval-SimTLiX (approved):

pancham88 (approved):

PR #1530, opened 2021-09-22
chore(core): remove insecure modules from webpack

bitgoAaron:
libsodium-wrappers & @bitgo/bls both rely on WebAssembly, but wasm code generation is only supported in Chrome if 'unsafe-eval' is added to CSP headers, which is a security risk.
BG-36664

## Changes

- Browser Tests updated to include CSP headers that will fail next time WebAssembly dependencies are added
- generated `browser.html` updated to include CSP headers that will also break when WebAssembly dependencies are added.


tylerlevine (approved):
nice!

PR #1532, opened 2021-09-22
refactor(account-lib): add calculateForwarderV1Address method

gianchandania:
**Background**

Methods to calculate ethereum forwarder version 1 addresses are being used in bitgo-admin and BitGoJS core module as well. These methods have been added in account-lib, so that they can be used from account lib in both places.

**Changes**

- Added methods `calculateForwarderV1Address()` and `getProxyInitcode()` to calculate forwarder v1 addresses in utils.ts file under eth folder.

- Updated the ethereumjs-util version in order to use `generateAddress2()` method

- Added test for `calculateForwarderV1Address()` method

- Updated the version of @bitgo/statics in order to use constants forwarderImplementationAddress and forwarderFactoryAddress for the test of the above method

Ticket: BG-35735


twtaylorbitgo (changes requested):
why do we have two versions of ethereumjs utils? this seems kinda nutty to me

gianchandania:
> why do we have two versions of ethereumjs utils? this seems kinda nutty to me

By two versions, you are referring to 6.2.1 and 6.0.0, right? 6.0.0 is a dependency of ethereumjs-abi and I have updated ethereumjs-util version 5.2.0 to 6.2.1 in order to use generateAddress2 method. I chose 6.2.1 because it is the same version used in microservices for ethereumjs-util. So I thought it would be better to keep the versions in sync in both the repos

modules/account-lib/src/coin/eth/utils.ts
```diff
@@ -389,6 +390,36 @@ export function calculateForwarderAddress(contractAddress: string, contractCount
...
 
+/**
+ * Calculate the forwarder v1 address that will be generated if `creatorAddress` creates it with salt `salt`
+ * and initcode `inicode using the create2 opcode
+ * @param {string} creatorAddress The address that is sending the tx to create a new address, hex string
+ * @param {string} salt The salt to create the address with using create2, hex string
+ * @param {string} initcode The initcode that will be deployed to the address, hex string
+ * @return {string} The calculated address
+ */
+export function calculateForwarderV1Address(creatorAddress: string, salt: string, initcode: string): string {
```
ryan-bitgo, line 401: i rather we start using buffers directly for parameters. then we don't have to cause from an implementation POV whether it is hex, uft-8, base64, etc... 

but not a blocker

modules/account-lib/src/coin/eth/utils.ts
```diff
@@ -389,6 +390,36 @@ export function calculateForwarderAddress(contractAddress: string, contractCount
...
+  );
+  return addHexPrefix(forwarderV1Address.toString('hex'));
+}
+
+/**
+ * Take the implementation address for the proxy contract, and get the binary initcode for the associated proxy
+ * @param {string} implementationAddress The address of the implementation contract for the proxy
+ * @return {string} Binary hex string of the proxy
+ */
+export function getProxyInitcode(implementationAddress: string): string {
```
ryan-bitgo, line 415: same with return values, buffers preferred

ryan-bitgo (approved):
looks fine, added some comments but not blockers.

twtaylorbitgo (approved):

PR #1533, opened 2021-09-22
feat(account-lib): support new fee model in EthTransactionData

ryan-bitgo:
Adds support for new transaction types in EthTransactionData. Update the tests and transaction builder to only use legacy txn to get everything to compile.

Update to transaction builder to support EIP1559 txns will come in a followup PR.

ticket: STLX-6731


modules/account-lib/src/coin/eth/iface.ts
```diff
@@ -3,15 +3,20 @@ import { KeyPair } from './keyPair';
 
 export interface Fee extends BaseFee {
   gasLimit: string;
+  eip1559?: EIP1559Fee;
```
wesgraham, line 6: do we want gasPrice to also be an optionalParam here? not sure why eip1559 needs to be here if gasPrice wasnt before

wesgraham: nvm

modules/account-lib/src/coin/eth/types.ts
```diff
@@ -25,22 +35,41 @@ export class EthTransactionData implements EthLikeTransactionData {
...
+    const maxPriorityFeePerGas = isEIP1559Txn(tx)
+      ? addHexPrefix(new BigNumber(tx.maxPriorityFeePerGas).toString(16))
+      : undefined;
+
     return new EthTransactionData(
-      new EthereumTx(
+      TransactionFactory.fromTxData(
         {
-          nonce: addHexPrefix(new BigNumber(tx.nonce).toString(16)),
+          type: isLegacyTx(tx) ? LEGACY_TX_TYPE : EIP1559_TX_TYPE,
```
wesgraham, line 53: will we always have either legacy or eip1559? want to make sure we dont allow type 1 to be misclassified if other methods won't first error with tx_type 1 for example

modules/account-lib/src/coin/eth/types.ts
```diff
@@ -61,9 +90,9 @@ export class EthTransactionData implements EthLikeTransactionData {
 
   /** @inheritdoc */
   toJson(): TxData {
-    const result: TxData = {
+    const result: BaseTxData = {
       nonce: bufferToInt(this.tx.nonce),
-      gasPrice: new BigNumber(bufferToHex(this.tx.gasPrice), 16).toString(10),
+      // gasPrice: this.tx instanceof LegacyTransaction ? gasPrice : undefined,
```
wesgraham, line 95: delete?

wesgraham (commented):
Changes look good but can you add even 1 unit test on a lhf tx to ensure that the fee object is specified correctly? or just try initializing a lhf fee object and verify it looks ok?

modules/account-lib/test/unit/coin/eth/transactionBuilder/walletInitialization.ts
```diff
@@ -90,7 +90,7 @@ describe('Eth Transaction builder wallet initialization', function () {
       tx.type.should.equal(TransactionType.WalletInitialization);
       const txJson = tx.toJson();
       txJson.gasLimit.should.equal('1000');
-      txJson.gasPrice.should.equal('10');
+      txJson.gasPrice?.should.equal('10');
```
wesgraham, line 93: also what is the reasoning behind all of these specifications?

modules/account-lib/src/coin/eth/iface.ts
ryan-bitgo: can add

modules/account-lib/src/coin/eth/types.ts
ryan-bitgo: yes

modules/account-lib/test/unit/coin/eth/transactionBuilder/walletInitialization.ts
ryan-bitgo: gasPrice is now optional but all the tests still assume legacy transaction, so i can either use the not null assertion `!` here or the `?`. I think i will go with the `!` and also add a check that the txn type is legacy.

modules/account-lib/src/coin/eth/types.ts
ryan-bitgo: Yes because TxData is a type we defined where the type is either legacy or eip1559. it would only be an issue when deserializing a tx. which i handled by throwing an error

modules/account-lib/src/coin/eth/iface.ts
```diff
@@ -1,17 +1,31 @@
 import { BaseFee } from '../baseCoin/iface';
 import { KeyPair } from './keyPair';
 
-export interface Fee extends BaseFee {
+export interface EthFee extends BaseFee {
```
wesgraham, line 4: nit: Why is this called EthFee now? nothing has changed, it still lives in the eth controller

wesgraham (approved):

PR #1534, opened 2021-09-23
chore(utxo-lib): depend on @bitgo/bitcoinjs-lib@6.0.0

OttoAllmendinger:
Fix nonstandard `Signer` import

Issue: BG-36608

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,360 +1,248 @@
----
-kind: pipeline
-name: check preconditions (node:10)
-
-platform:
-  os: linux
-  arch: amd64
-
-steps:
-- name: build information
-  image: node:10
-  commands:
-  - node --version
-  - npm --version
-  - yarn --version
-  - git --version
-
-- name: install
-  image: node:10
-  commands:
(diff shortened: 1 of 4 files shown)


sangaman (dismissed):
6.0 change looks good but I forgot where we ended up with discussing the `.drone.yml` file. I thought we want to keep it after all?

OttoAllmendinger:
we do, I'm just looking for a good way to generate it and check the correct compilation via CI

sangaman (dismissed):
I get the same output from running `drone jsonnet --format`, weird that it's rewriting the whole file but I guess that's how it goes.

tylerlevine:
I think this broke drone, when it hangs like this usually that means it can't parse the .drone.yml successfully

tylerlevine (changes requested):
please take a look at the drone build not getting triggered - usually this means there's a problem with the .drone.yml file

tylerlevine:
looks like using the provided `yarn gen-drone` script produced a valid .drone.yml with the latest changes.

I'm running 
```
$ drone --version
drone version 1.2.4
```

I wonder if a newer version of drone-cli is producing the different yaml?

tylerlevine (approved):

PR #1541, opened 2021-09-29
feat(statics): add eos mainnet token config and prefix token 

lcovar:
TICKET: STLX-6299


lcovar:
@pancham88 Build is failing since teos token and eos token name are the same. Can we remove teos token statics config?

pancham88:
> @pancham88 Build is failing since teos token and eos token name are the same. Can we remove teos token statics config?

@lcovar  Yes, the prod tokens (which are implicitly available in testnet) shall not be defined in teosToken

CapnMigraine (approved):
looks good 👍

PR #1543, opened 2021-10-01
fix(teth): add terc tokens with 2,6,18 decimals

wesgraham:
Ticket: BG-35974


samwise2 (approved):

samwise2:
Kovan addresses are correct. Goerli addresses are incorrect, but are fixed by #1548

PR #1547, opened 2021-10-04
refactor(core): use eth utils method from acctlib

gianchandania:
**Background**

In order to calculate forwarder version 1 address for Eth, we had to add calculateEthAddress2, getProxyInitcode methods and forwarderFactoryAddress, forwarderImplementationAddress constants in BitGoJS core module. But the same methods and constants were required in bitgo-admin as well. So we decided to add the methods in account lib and constants in @bitgo/statics

**Changes**

- Used calculateEthAddress2, getProxyInitcode  methods from account lib
- Used forwarderFactoryAddress, forwarderImplementationAddress from @bitgo/statics
- Modified the condition to invoke `verifyAddress()` method in `createAddress()` method. In case of Eth forwarder v1, if the address is pending onchain initialization, then need to verify the address
- Added unit test for this change in condition

Ticket: BG-35735

### DIFF
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -44,6 +44,7 @@ import { BaseCoin as StaticsBaseCoin, EthereumNetwork } from '@bitgo/statics';
 import { checkKrsProvider, getIsKrsRecovery, getIsUnsignedSweep } from '../recovery/initiate';
 import type * as EthTxLib from '@ethereumjs/tx';
 import type * as EthCommon from '@ethereumjs/common';
+import * as accountLib from '@bitgo/account-lib';
 
 const co = Bluebird.coroutine;
 const debug = debugLib('bitgo:v2:eth');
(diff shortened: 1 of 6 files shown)


modules/core/src/v2/coins/eth.ts
```diff
@@ -1674,16 +1675,23 @@ export class Eth extends BaseCoin {
     if (coinSpecific.forwarderVersion === 0) {
       return true;
     } else {
-      const forwarderFactoryAddress = config.defaults[self.getChain()].forwarderFactoryAddress;
-      const initcode = Util.getProxyInitcode(config.defaults[self.getChain()].forwarderImplementationAddress);
+      const forwarderFactoryAddress = (self.staticsCoin?.network as EthereumNetwork).forwarderFactoryAddress!;
```
ryan-bitgo, line 1678: Can we turn this into a method so we don't have to cast every time

`(self.staticsCoin?.network as EthereumNetwork)`

--->

```
getNetwork(): EthereumNetwork | undefined {
   ...
}
```

ryan-bitgo (approved):

barathcj (dismissed):

twtaylorbitgo (dismissed):
Thank you for the follow through!

modules/core/src/v2/coins/eth.ts
gianchandania: Thanks for the suggestion. I have updated the PR

ryan-bitgo (approved):

PR #1548, opened 2021-10-04
fix(gterc-tokens): add missing gterc tokens

wesgraham:
Adds missing gterc tokens to statics configs
Ticket: BG-35974


twtaylorbitgo (approved):

samwise2:
The addresses are correct

PR #1551, opened 2021-10-05
feat(utxo-lib): add p2tr output scripts support

sangaman:
This adds support for creation of p2tr output scripts using recent changes to the BitGo bitcoinjs-lib fork.

Ticket: BG-35573


modules/utxo-lib/src/bitgo/outputScripts.ts
```diff
@@ -63,3 +65,39 @@ export function createOutputScript2of3(pubkeys: Buffer[], scriptType: ScriptType
...
+ * @returns {{scriptPubKey}}
+ */
+export function createTaprootScript2of3([userKey, backupKey, bitGoKey]: [Buffer, Buffer, Buffer]): SpendableScript {
+  [userKey, backupKey, bitGoKey].forEach((key) => {
+    if (key.length !== 33 && key.length !== 32) {
+      throw new Error(`Unexpected key length ${key.length}. Must use compressed keys.`);
+    }
+  });
+
+  const userBitGoScript = bitcoinjs.payments.p2ms({ m: 2, pubkeys: [userKey, bitGoKey]}).output;
```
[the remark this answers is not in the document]
sangaman: Ah yes thank you that's right. I'll straighten this out.

sangaman:
I updated the PR to use the correct scripts and fix some linting errors.

brandonblack (approved):

modules/utxo-lib/package.json
```diff
@@ -46,7 +46,7 @@
   "dependencies": {
     "@bitgo/blake2b": "^3.0.1",
     "bitcoin-ops": "^1.3.0",
-    "bitcoinjs-lib": "npm:@bitgo/bitcoinjs-lib@6.0.0",
+    "bitcoinjs-lib": "git+https://github.com/BitGo/bitcoinjs-lib.git#e445ccbd706fc0bc00646e6942fdc5177bbf1061",
```
tylerlevine, line 49: let's make sure to change this back to an npm version once we have something stable - this is fine for now though

modules/utxo-lib/src/bitgo/outputScripts.ts
```diff
@@ -63,3 +66,39 @@ export function createOutputScript2of3(pubkeys: Buffer[], scriptType: ScriptType
...
 }
+
+/**
+ * Creates and returns a taproot output script using the user and bitgo keys for the aggregate
+ * public key and a taptree containing a user+bitgo 2-of-2 script at the first depth level of the
+ * tree and user+backup and bitgo+backup 2-of-2 scripts one level deeper.
+ * @param pubkeys - a pubkey array containing the user key, backup key, and bitgo key in that order
+ * @returns {{scriptPubKey}}
+ */
+export function createTaprootScript2of3([userKey, backupKey, bitGoKey]: [Buffer, Buffer, Buffer]): SpendableScript {
```
tylerlevine, line 77: I wonder if we can unify this with `createOutputScript2Of3` - seems like we just have another case here that could be handled as part of that function

sangaman: I did originally try it that way but it felt a bit cluttered, I couldn't think of a better way of going about it besides just an`if (p2tr) then... else...` approach, at which point a separate function made sense. If I just added it to the `switch` statement then we'd be creating`script2of3` for nothing and creating a separate set of scripts used only by p2tr. I also thought it would be nice to be able to use this `[Buffer, Buffer, Buffer]` type with named array elements since the order of the keys matters and makes the code easier to read I think. Technically `createOutputScript2of3` could also use that type for its `pubkeys` parameter but it would've required a decent bit of refactoring to do.

However it might make sense to have `createTaprootscript2of3` called from within `createOutputScript2of3`, I'll explore that in a separate PR.

OttoAllmendinger: for clarity I would prefer we used the named keys in `createOutputScript2of3` and supported `case 'p2tr'` there

PR #1553, opened 2021-10-05
chore(algo): allow valuesless transfer BG-37109

aaldemir:
`allowValuelessTransfer` for algo to be set to `true`. Adds supportsValuelessTransfer method to algo coin object. This will allow a (t)ALGO transaction with zero value to be built. Zero value transaction are needed for ALGO for various reasons including participation in staking and governance.

Related change in WP: https://github.com/BitGo/bitgo-microservices/pull/12207

JIRA ticket: https://bitgoinc.atlassian.net/browse/BG-37109


twtaylorbitgo (dismissed):

modules/core/src/v2/coins/algo.ts
```diff
@@ -135,12 +135,7 @@ export class Algo extends BaseCoin {
    * @returns {boolean} True if okay to send 0 value, false otherwise
    */
   valuelessTransferAllowed(type?: unknown): boolean {
```
tylerlevine, line 137: please remove the `type` param from here and from the base class - I think it was only added for this use case which we don't need any more.

tylerlevine (changes requested):

tylerlevine (changes requested):
Please check the build, looks like you have a type error

tylerlevine (approved):

PR #1560, opened 2021-10-07
test(core): ticket : BG-35558 test cases of cashaddr support for send…

mukeshsp:
ticket : https://bitgoinc.atlassian.net/browse/BG-35558


modules/core/test/v2/integration/coins/bch.ts
```diff
@@ -0,0 +1,44 @@
...
+import { TestBitGo } from '../../../lib/test_bitgo';
+import * as nock from 'nock';
+
+nock.restore();
+
+describe('BCH:', function () {
+  let bitgo;
+  let wallet;
+
+  before(co(function *() {
```
tylerlevine, line 16: please use async functions instead of coroutines for new code

modules/core/test/v2/integration/coins/bch.ts
```diff
@@ -0,0 +1,44 @@
+import * as should from 'should';
```
tylerlevine, line 1: Please add `@prettier` annotation for new files

modules/core/test/v2/integration/coins/bch.ts
```diff
@@ -0,0 +1,44 @@
...
+
+      const transaction = await wallet.send({
+        address: TestBitGo.V2.TEST_BCH_WALLET_CASH_ADDRESS,
+        amount: '5000',
+        walletPassphrase: TestBitGo.V2.TEST_BCH_WALLET_PASSPHRASE,
+        otp: bitgo.testUserOTP(),
+      });
+
+      should.exist(transaction);
+      transaction.should.have.property('transfer');
```
tylerlevine, line 36: it would probably be good to check the response has the correctly formatted cashaddr address too, but you can do that later if you want to come back to this.

PR #1561, opened 2021-10-07
feat(express): add route for create address

gianchandania:
This PR adds a route and unit test for wallet.createAddress in express

Ticket: BG-33701

### DIFF
--- a/modules/express/test/unit/clientRoutes/createAddress.ts
+++ b/modules/express/test/unit/clientRoutes/createAddress.ts
@@ -1,0 +1,45 @@
+import * as sinon from 'sinon';
+
+import 'should-http';
+import 'should-sinon';
+import '../../lib/asserts';
+
+import * as express from 'express';
+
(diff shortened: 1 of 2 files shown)


modules/express/test/unit/clientRoutes/createAddress.ts
```diff
@@ -0,0 +1,64 @@
...
+      body: {
+        chain: 0
+      },
+    } as unknown as express.Request;
+
+    await handleV2CreateAddress(req)
+      .should.be.resolvedWith(res);
+  });
+
+  it('should return the address object', async () => {
```
tylerlevine, line 45: nit: seems like this is the same test, but for eth. Do we need both? If so, could you add that to the test name so we know these are actually doing different things?

gianchandania: I have removed one test as there was no other difference apart from the coin and in both cases, the behaviour is going to be the same only

tylerlevine (approved):

PR #1562, opened 2021-10-07
started sign and send defi transaction function

0xJacobV:


modules/core/src/v2/wallet.ts
```diff
@@ -1894,6 +1894,71 @@ export class Wallet {
     }).call(this).asCallback(callback);
   }
 
+  /**
+   * Sign a metamask transaction and send to wallet platform for pending approval.
+   *
+   * @param params
+   * @param callback
+   */
+  signAndSendMMITransaction(params: any/* TODO: add type */, callback?: NodeCallback<any>): Bluebird<any> {
```
karankurbur, line 1903: Can we convert this to async/await syntax inside this file or are there complications? Otherwise lets move to a new file

modules/core/src/v2/wallet.ts
```diff
@@ -1894,6 +1894,62 @@ export class Wallet {
...
+            confirmedBalance: self.confirmedBalance(),
+            spendableBalance: self.spendableBalance(),
+          };
+          error.txParams = _.omit(params, ['keychain', 'prv', 'passphrase', 'walletPassphrase', 'key']);
+        }
+        throw error;
+      }
+
+      // Add MMI flag to half signed tx for tx state updated webhooks
+      halfSignedTx.halfSigned.isMMITransaction = true;
```
0xJacobV, line 1944: remove

modules/core/src/v2/environments.ts
```diff
@@ -57,6 +57,7 @@ export type EnvironmentName =
   | 'staging'
   | 'test'
   | 'dev'
+  | 'testnet'
```
karankurbur, line 60: maybe something a bit more verbose, devtestnet? testnet feels like its referring to test

modules/core/src/v2/wallet.ts
```diff
@@ -1898,6 +1903,65 @@ export class Wallet {
...
+    const self = this;
+    return co(function *() {
+      // retrieve our keychains needed to run the prebuild - some coins use all pubs
+      const keychains = (yield self.baseCoin.keychains().getKeysForSigning({ wallet: self, reqId: params.reqId })) as any;
+      const txPrebuild = {
+        eip1559: false,
+        isBatch: false,
+        recipients: [
+          {
+            address: params.to,
```
karankurbur, line 1922: does params automatically give us info from `recipients[0]`?

tsconfig.json
```diff
@@ -17,6 +17,7 @@
     "strictPropertyInitialization": true,
     "strictBindCallApply": true,
     "strictFunctionTypes": true,
+    "skipDefaultLibCheck": true,
```
karankurbur, line 20: what does this do?

PR #1563, opened 2021-10-07
fix(utxo-lib): pass 0 offset to `readUInt16BE` for zcash `fromBase58C…

tylerlevine:
…heck`

When polyfilled to use `UInt8Array` instead of node's `Buffer`, the
`readUInt16BE` function requires an offset to be given instead of using `0` as
a default.

Ticket: CR-400

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,9 +53,10 @@
     }
   },
   "resolutions": {
+    "axios": "^0.21.2",
     "elliptic": "^6.5.4",
-    "underscore": "^1.12.1",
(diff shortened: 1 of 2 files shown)


sangaman (dismissed):

package.json
```diff
@@ -53,9 +53,10 @@
     }
   },
   "resolutions": {
+    "axios": "^0.21.2",
```
tylerlevine, line 56: added axios here and alphabetized

PR #1567, opened 2021-10-13
refactor(utxo-lib): use Buffer[] in createTaprootScript2Of3

OttoAllmendinger:
see commit msg

### DIFF
--- a/modules/utxo-lib/src/bitgo/outputScripts.ts
+++ b/modules/utxo-lib/src/bitgo/outputScripts.ts
@@ -1,6 +1,7 @@
 import * as assert from 'assert';
 import * as bitcoinjs from 'bitcoinjs-lib';
 import { OP_CHECKSIG, OP_CHECKSIGVERIFY } from 'bitcoin-ops';
+import { isTriple, Triple } from './types';
 
 export const scriptTypes2Of3 = ['p2sh', 'p2shP2wsh', 'p2wsh', 'p2tr'] as const;
 export type ScriptType2Of3 = typeof scriptTypes2Of3[number];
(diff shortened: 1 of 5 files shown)


OttoAllmendinger:
After some thinking I actually decided to go in the other direction. It's not really worth cluttering all the callsites.

modules/utxo-lib/src/bitgo/outputScripts.ts
```diff
@@ -94,10 +96,14 @@ export function createOutputScript2of3(pubkeys: Buffer[], scriptType: ScriptType
  * Creates and returns a taproot output script using the user and bitgo keys for the aggregate
  * public key and a taptree containing a user+bitgo 2-of-2 script at the first depth level of the
  * tree and user+backup and bitgo+backup 2-of-2 scripts one level deeper.
- * @param pubkeys - a pubkey array containing the user key, backup key, and bitgo key in that order
+ * @param pubkeys - a pubkey triple containing the user key, backup key, and bitgo key in that order
  * @returns {{scriptPubKey}}
  */
-function createTaprootScript2of3([userKey, backupKey, bitGoKey]: [Buffer, Buffer, Buffer]): SpendableScript {
+function createTaprootScript2of3(pubkeys: Buffer[]): SpendableScript {
```
sangaman, line 102: This can still use the `Triple<Buffer>` type no? SInce it's only called internally and not exported and we're already calling `isTriple` in `createOutputScript2of3` with the proper type guards, we then wouldn't need to call it again below.

sangaman (dismissed):
This is also fine with me, see one note below to consider.

tylerlevine (approved):

PR #1569, opened 2021-10-13
test: use TxbSignArg in txb.sign()

OttoAllmendinger:
While the other sign method is not directly tested anymore, it is easy to
check out the previous commit and verify that the old style works as well.

We will completely deprecate the other `sign()` call style soon enough.

Issue: BG-35578

f15fb36e6 (Otto Allmendinger, 6 minutes ago)
feat: add support for sign(signParams: TxbSignArg)

The other `sign()` calling style is being deprecated and results in 
deprecation warning spam in the console.

Issue: BG-35578

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoTransactionBuilder.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoTransactionBuilder.ts
@@ -54,14 +54,28 @@ export class UtxoTransactionBuilder<T extends UtxoTransaction = UtxoTransaction>
     witnessValue?: number,
     witnessScript?: Buffer
   ): void {
-    if (typeof signParams !== 'number') {
-      throw new Error(`TxbSignArg not supported yet`);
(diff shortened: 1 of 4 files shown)


modules/utxo-lib/src/bitgo/UtxoTransactionBuilder.ts
```diff
@@ -54,14 +54,28 @@ export class UtxoTransactionBuilder<T extends UtxoTransaction = UtxoTransaction>
     witnessValue?: number,
     witnessScript?: Buffer
   ): void {
-    if (typeof signParams !== 'number') {
-      throw new Error(`TxbSignArg not supported yet`);
-    }
+    // Regular bitcoin p2sh-p2ms inputs do not include the input amount (value) in the signature and
+    // thus do not require the parameter `value` to be set.
+    // For bitcoincash and bitcoinsv p2sh-p2ms inputs, the value parameter *is* required however.
```
sangaman, line 59: I guess bitcoinsv still allows for spending p2sh inputs?

PR #1570, opened 2021-10-13
fix(account-lib): add parsing for optional type in stringifyCV

alebusse:
Improved stringifyCV method to parse optionalSome and Tuple params to avoid leaving BigInts unparsed
Added and Improved Unit tests

[stlx-7912](https://bitgoinc.atlassian.net/browse/STLX-7912)


alebusse:
Tested in testnet-06 using custom release version https://github.com/BitGo/bitgo-microservices/commit/ddd4dcea6ebecf7b4b5a4bd2a2fc0a3417f6195f

modules/account-lib/src/coin/stx/utils.ts
```diff
@@ -435,6 +435,18 @@ export function stringifyCv(cv: ClarityValue): any {
     case ClarityType.Int:
     case ClarityType.UInt:
       return { type: cv.type, value: cv.value.toString() };
+    case ClarityType.OptionalSome:
+      return { type: cv.type, value: stringifyCv(cv.value) };
+    case ClarityType.Tuple:
+      return {
+        type: cv.type,
+        data: Object.assign({}, ...Object.entries(cv.data).map(([key, value]) => ({ [key]: stringifyCv(value) }))),
```
tylerlevine, line 443: This is a complicated way to do `_.mapValues()`, I would try to use that too instead (since you already have it in scope anyway)
```suggestion
        data: _.mapValues(cv.data, (value) => stringifyCV(value)),
```

alebusse:
> looks reasonable to me, I just have one suggestion for improvement.
> 
> Also I wonder if we should `throw new Error('unknown clarity value type')` in the default case instead of just returning `cv` - would that have helped find errors like this sooner?

We are not trying to parse or validate every type, just the ones that can contain more inside and these may contain Ints or Uints

tylerlevine (approved):

PR #1573, opened 2021-10-15
Backmerge for SDK release 13.0.1

tylerlevine:


modules/core/src/v2/wallet.ts
```diff
@@ -2075,7 +2075,7 @@ export class Wallet {
           if (amount.isNegative()) {
             throw new Error('invalid argument for amount - positive number greater than zero or numeric string expected');
           }
-          if (!coin.valuelessTransferAllowed() && amount.isZero()) {
+          if (!coin.valuelessTransferAllowed(params.type) && amount.isZero()) {
```
tylerlevine, line 2078: remove

modules/core/src/v2/baseCoin.ts
```diff
@@ -276,7 +276,7 @@ export abstract class BaseCoin {
    * Flag for sending value of 0.
    * @returns {boolean} True if okay to send 0 value, false otherwise
    */
-  valuelessTransferAllowed(): boolean {
+  valuelessTransferAllowed(params?: unknown): boolean {
```
tylerlevine, line 279: remove

PR #1575, opened 2021-10-15
fix(utxo-lib): fix sighash for dash transactions

OttoAllmendinger:
Previously, we were not including `extraPayload` in the hash preimage for
transaction signatures. This was because the `hashForSignature` method
calls the `__toBuffer()` method which was not overridden in the 
DashTransaction subclass (only the public `toBuffer()` method was).

This change now overrides `__toBuffer` with a not so pretty hack. We can
fix that later.

Test fixtures were created using utxolib `v1.9.6`.

Issue: BG-37672

### DIFF
--- a/modules/utxo-lib/test/integration_local_rpc/parseDashEvo.ts
+++ b/modules/utxo-lib/test/integration_local_rpc/parseDashEvo.ts
@@ -11,10 +11,15 @@ However the generated fixtures have the same format as the RPC responses so we w
 
 import * as fs from 'fs-extra';
 import * as assert from 'assert';
-import { parseTransactionRoundTrip } from '../transaction_util';
-import networks = require('../../src/networks');
+
 import { RpcTransaction } from './generate/RpcTypes';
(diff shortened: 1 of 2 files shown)


modules/utxo-lib/test/integration_local_rpc/parseDashEvo.ts
```diff
@@ -26,8 +31,13 @@ export async function readDashEvoTransactions(): Promise<RpcTransaction[]> {
...
   const network = networks.dash;
-  it(`parses Evolution (EVO) special transactions`, async function () {
-    const txs = await readDashEvoTransactions();
+  let txs: RpcTransaction[];
+
+  before('read fixtures', async function () {
+    txs = await readDashEvoTransactions();
+  });
+
+  it(`parses Evolution (EVO) special transactions`, function () {
```
brandonblack, line 40: I wonder what the `EVO` terminology used here is compared to the `proReg` terminology used in HSM. Don't understand Dash. Should we be helping future engineers by using both terms in both places or only one in both?

PR #1576, opened 2021-10-15
Move CI to GitHub actions

tylerlevine:
cloud.drone.io has had persistent issues with build backlogs, and currently we haven't had any builds run for 22 hours.

This PR sets up a largely equivalent workflow for Github Actions.

Additionally, node 16 is added to the build matrix and the `engines` property of the `core` packages was extended to allow node 16.

Ticket: BG-37746


OttoAllmendinger:
```
bitgo: Error: Need to set BITGOJS_TEST_PASSWORD env variable - please see the developer setup docs.
```

maybe we should skip these sorts of tests for now until we have set it up properly

PR #1582, opened 2021-10-19
fix(algo): support for signing unsigned keyreg transaction (bg-37892)

pancham88:
[JIRA](https://bitgoinc.atlassian.net/browse/BG-37892)

Add support for signing unsigned `keyreg` algo transactions.


modules/core/src/v2/wallet.ts
```diff
@@ -513,7 +513,7 @@ export class Wallet {
       'lastLedgerSequence', 'ledgerSequenceDelta', 'maxFee', 'maxFeeRate', 'maxValue', 'memo', 'transferId', 'message', 'minConfirms',
       'minValue', 'noSplitChange', 'numBlocks', 'recipients', 'reservation', 'sequenceId', 'strategy',
       'targetWalletUnspents', 'trustlines', 'type', 'unspents', 'nonParticipation', 'validFromBlock', 'validToBlock', 'messageKey',
-      'stakingOptions', 'eip1559',
+      'stakingOptions', 'eip1559', 'eip1559', 'keyregTxBase64',
```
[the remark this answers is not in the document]
pancham88: Thanks for pointing it out, resolved

modules/core/src/v2/wallet.ts
```diff
@@ -513,7 +513,7 @@ export class Wallet {
       'lastLedgerSequence', 'ledgerSequenceDelta', 'maxFee', 'maxFeeRate', 'maxValue', 'memo', 'transferId', 'message', 'minConfirms',
       'minValue', 'noSplitChange', 'numBlocks', 'recipients', 'reservation', 'sequenceId', 'strategy',
       'targetWalletUnspents', 'trustlines', 'type', 'unspents', 'nonParticipation', 'validFromBlock', 'validToBlock', 'messageKey',
-      'stakingOptions', 'eip1559',
+      'stakingOptions', 'eip1559', 'keyregTxBase64',
```
tylerlevine, line 516: why are we whitelisting this param? I don't see this anywhere else in the SDK code base. Where is this created/used?

argjv: That is a valid parameter in the wallet platform: https://github.com/BitGo/bitgo-microservices/blob/e201919bccd3852f9e1888c780dfcbafc4e677e4/packages/wallet-platform/app/controllers/api/v2/transactions.js#L111

I'm guessing this was passed as part in bitgo-admin, and we are now passing it in the SDK.

PR #1585, opened 2021-10-20
feat(utxo-lib): support p2shP2pk inputs

OttoAllmendinger:
* Add method `createOutputScriptP2shP2pk(Buffer): SpendableScript`
* Add new optional property `p2shOutputClassification`, which will be
  set for all `p2sh` scripts
* Inputs that are of type `p2shP2pk` will have
  `p2shOutputClassification` set to `pubkey`.

Issue: BG-37916

### DIFF
--- a/modules/utxo-lib/src/bitgo/outputScripts.ts
+++ b/modules/utxo-lib/src/bitgo/outputScripts.ts
@@ -36,6 +36,22 @@ export type SpendableScript = {
 };
 
 /**
+ * Return scripts for p2sh-p2pk (used for BCH/BSV replay protection)
+ * @param pubkey
+ */
+export function createOutputScriptP2shP2pk(pubkey: Buffer): SpendableScript {
+  const p2pk = bitcoinjs.payments.p2pk({ pubkey });
+  const p2sh = bitcoinjs.payments.p2sh({ redeem: p2pk });
+  if (!p2sh.output || !p2pk.output) {
+    throw new Error(`invalid state`);
+  }
+  return {
+    scriptPubKey: p2sh.output,
+    redeemScript: p2pk.output,
+  };
+}
+
+/**
  * Return scripts for 2-of-3 multisig output
  * @param pubkeys - the key triple for multisig
  * @param scriptType
--- a/modules/utxo-lib/src/bitgo/signature.ts
+++ b/modules/utxo-lib/src/bitgo/signature.ts
@@ -94,10 +94,12 @@ export function parseSignatureScript(
   }
 
   if (inputClassification === classify.types.P2PKH) {
+    /* istanbul ignore next */
     if (!decompiledSigScript || decompiledSigScript.length !== 2) {
       throw new Error('unexpected signature for p2pkh');
     }
     const [signature, publicKey] = decompiledSigScript;
+    /* istanbul ignore next */
     if (!Buffer.isBuffer(signature) || !Buffer.isBuffer(publicKey)) {
       throw new Error('unexpected signature for p2pkh');
     }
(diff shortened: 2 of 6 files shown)


modules/utxo-lib/test/integration_local_rpc/parse.ts
```diff
@@ -87,13 +87,13 @@ function runTestParse(network: Network, txType: FixtureTxType, scriptType: Scrip
...
+        return;
+      }
+
       parsedTx.ins.forEach((input, i) => {
         const result = parseSignatureScript(input) as ParsedSignatureScript2Of3;
 
-        if (txType === 'deposit') {
-          return;
-        }
-
```
OttoAllmendinger, line 96: For unclear reasons, `parseSignatureScript` throws an error when parsing unknown inputs for certain altcoins (bitcoin gold and litecoin)

However since we should not be calling `parseSignatureScript` on external inputs it does not matter, so I removed this test for deposit transactions.

modules/utxo-lib/src/bitgo/outputScripts.ts
```diff
@@ -35,6 +35,22 @@ export type SpendableScript = {
   controlBlocks?: [userBitGoScript: Buffer, userBackupScript: Buffer, backupBitGoScript: Buffer];
 };
 
+/**
+ * Return scripts for p2sh-p2pkh (used for BCH/BSV replay protection)
+ * @param pubkey
+ */
+export function createOutputScriptP2shP2pkh(pubkey: Buffer): SpendableScript {
+  const p2pkh = bitcoinjs.payments.p2pk({ pubkey });
```
sangaman, line 43: I'm guessing this should use `payments.p2pkh`?

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -70,15 +71,16 @@ export function parseSignatureScript(
 ): ParsedSignatureScript | ParsedSignatureP2PKH | ParsedSignatureScript2Of3 {
   const isSegwitInput = input.witness.length > 0;
   const isNativeSegwitInput = input.script.length === 0;
```
sangaman, line 73: Drive-by observation here and I don't think this makes a difference since it's only used when `isSegwitInput` is true, but for the value to match the variable name this should be `isSegwitInput && input.script.length === 0;` Or can be renamed to `isNotScriptHashInput` but I like the other way better. Either way it makes the code a bit cleaner and might prevent a future mistake.

modules/utxo-lib/test/bitgo/signature.ts
```diff
@@ -80,3 +81,22 @@ describe('signature', function () {
...
+      fixtureKeys.slice(0, 1),
+      'p2shP2pkh',
+      networks.bitcoin
+    ).build();
+
+    signedTransaction.ins.forEach((input) => {
+      assert.deepStrictEqual(parseSignatureScript(input), {
+        inputClassification: 'scripthash',
+        isSegwitInput: false,
+        p2shOutputClassification: 'pubkey',
```
sangaman, line 98: I'm also guessing this should be `pubkeyhash`, unless in fact the RP inputs are p2sh-p2pk and not p2sh-p2pkh.

OttoAllmendinger: actually the RP inputs ARE `p2sh-p2pk`

modules/utxo-lib/src/bitgo/outputScripts.ts
OttoAllmendinger: nope, I misnamed the function

modules/utxo-lib/src/bitgo/signature.ts
OttoAllmendinger: agreed but will fix in follow-up

PR #1590, opened 2021-10-20
fix(sdk): Support Duplicate Token Contract Addresses

samwise2:
Ticket: BG-37099

This PR refactors statics + coinFactory to support duplicate token contract addresses (as long as the tokens are on different networks). This PR should be followed up with another PR to support duplicate token names in `statics/map.ts`, which I can create using a similar technique to that used in this PR. However, since this PR is already quite large, I've omitted it for now to keep the changes tighter.


modules/core/src/bitgo.ts
```diff
@@ -607,10 +607,11 @@ export class BitGo {
 
   /**
    * Create a basecoin object
-   * @param coinName
+   * @param coinIdentifier
+   * @param networkName
    */
-  coin(coinName: string): BaseCoin {
-    return GlobalCoinFactory.getInstance(this, coinName);
+  coin(coinIdentifier: string, networkName: string | null | undefined = undefined): BaseCoin {
```
wesgraham, line 613: why are we renaming to identifier? isnt it still a name (i.e usdt)

wesgraham: also like name bc it lets us distinguish whether the "identifier" is a name or an address

modules/core/src/v2/coinFactory.ts
```diff
@@ -68,137 +68,228 @@ import { Tbcha } from './coins/tbcha';
...
-          type: 'unknown',
-          coin: 'eth',
-          network: 'Mainnet',
-          name: 'Unknown',
-          tokenContractAddress: name,
-          decimalPlaces: 0,
-        });
-        return unknownTokenConstructor(bitgo);
+    const coinsOnEthereum = this.coinConstructors.get('Ethereum');
+    if (coinsOnEthereum) {
```
wesgraham, line 119: will this ever not be satisfied?

modules/core/src/v2/coinFactory.ts
```diff
@@ -68,137 +68,228 @@ import { Tbcha } from './coins/tbcha';
...
-GlobalCoinFactory.registerCoinConstructor('hbar', Hbar.createInstance);
-GlobalCoinFactory.registerCoinConstructor('thbar', Thbar.createInstance);
-GlobalCoinFactory.registerCoinConstructor('ofc', Ofc.createInstance);
-GlobalCoinFactory.registerCoinConstructor('susd', Susd.createInstance);
-GlobalCoinFactory.registerCoinConstructor('tsusd', Tsusd.createInstance);
-GlobalCoinFactory.registerCoinConstructor('cspr', Cspr.createInstance);
-GlobalCoinFactory.registerCoinConstructor('tcspr', Tcspr.createInstance);
-GlobalCoinFactory.registerCoinConstructor('stx', Stx.createInstance);
-GlobalCoinFactory.registerCoinConstructor('tstx', Tstx.createInstance);
+// TODO: leverage statics instead of making 30 manual calls
```
wesgraham, line 186: is this still a todo? if coming back to later can you add a ticket number?

modules/statics/src/map.ts
```diff
@@ -2,10 +2,13 @@ import { BaseCoin } from './base';
...
 
 export class CoinMap {
+  // A mapping form coin name -> BaseCoin object
   private readonly _map = new Map<string, Readonly<BaseCoin>>();
 
   private constructor() {}
 
+  // TODO: update formCoins to allow for duplicate coin names on different networks
+  // using a similar strategy to that used for coinFactory
   static fromCoins(coins: Readonly<BaseCoin>[]): CoinMap {
```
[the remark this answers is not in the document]
wesgraham: yea if you could add a ticket number here that would be great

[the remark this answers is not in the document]
samwise2: Totally, this is it: https://bitgoinc.atlassian.net/browse/BG-37944

modules/core/src/config.ts
```diff
@@ -111,6 +115,8 @@ const formattedAlgoTokens = coins.reduce((acc: AlgoTokenConfig[], coin) => {
       type: coin.name,
       coin: coin.network.type === NetworkType.MAINNET ? 'algo' : 'talgo',
       network: coin.network.type === NetworkType.MAINNET ? 'Mainnet' : 'Testnet',
+      networkName: coin.network.name,
```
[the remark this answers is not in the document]
samwise2: Cool, this should just be adding a field to the token object, but I'll add him to the PR and send him a quick message

modules/core/src/v2/coinFactory.ts
samwise2: Totally, created a ticket here: https://bitgoinc.atlassian.net/browse/BG-37946

tylerlevine: Could you add the ticket number to the comment here?

tylerlevine (changes requested):
At a high level,  this feels backwards to me. Having a token with the same ticker that exists on multiple networks is the exception, not the rule, so it seems like we should really not require both the ticker and the network for new coins unless it's otherwise ambiguous which coin that means.

At a high level, I would rather do this:
1) Loop over the `coin` export in statics as I commented in the last PR. This way we have the statics coin at registration time and thus the network
2) If we go to register a coin and see that it's already in the `coinConstructors` map, remove the existing coin and then add both to the `explicitNetworkCoinConstructors` map. Remove all references to "legacy" since there's nothing being deprecated about the existing coins or how they're used
3) Remove `requiresExplicitNetwork` from all coin definitions in both core and statics - we'll know they require an explicit network because they conflict on the ticker.

Sorry it's taken me so long to get to this - it's been a crazy week. But I'm happy to chat more on this async.

modules/statics/src/map.ts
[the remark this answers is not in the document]
tylerlevine: Also add to source code comment please

samwise2:
First: I agree that everything related to “legacy” should be removed, and I’ll take that out in my next commit. I’ll also add the ticket into that comment.

samwise2:
The design that you’re suggesting makes sense, but I think there are some other considerations that are relevant.

If we consider ETH, for example, it may not be the exception to the rule that coin identifiers are duplicate. Considering that we also use this map to store token contract addresses, a very normal thing to do would be to deploy an ERC-20 token to the same addresses in testnet and mainnet. 

A primary tennant of the design choices in this PR, were aligned with the idea that at some point we would likely move away from “bitgo.coin(0x…)” in favor of “bitgo.coin(0x.., Ethereum)”. Because the latter will not run into problems no matter how many new tokens etc. are added to BitGo, whereas the former could develop a gradually larger and larger mapping of coins which need special handling.

And this is why I specified `requiresExplicitNetwork` in statics. While it may be a bit “uglier” code-wise, it shows, in a declarative way, (for the SDK user) which tokens they can instantiate with `bitgo.coin(name)` and which they must insatiate with `bitgo.coin(name, network)`. If we employ this solution of only storing exceptions in a separate mapping at runtime, then the SDK consumer has no way of knowing which coins they can instantiate with which method. They could only learn via trial and error, and if they don’t run their tests against ALL ERC-20s, for example, they may not catch these errors in advance of prod/whatever their final use case is. By using statics, the behavior is predictable for the end user.

tylerlevine:
> If we consider ETH, for example, it may not be the exception to the rule that coin identifiers are duplicate.

That's a very ETH-centric view on it which I don't feel is appropriate for a library which supports many different blockchains. I'd even argue that it's still the exception to the rule for ETH, and I would provide as evidence the fact that we've already gotten to where we are without needing to support duplicate coin identifiers.

> Because the latter will not run into problems no matter how many new tokens etc. are added to BitGo, whereas the former could develop a gradually larger and larger mapping of coins which need special handling.

That feels like PMO to me - I don't think this mapping will ever grow so large as to pose a real problem. It would only ever store the tokens which have a name conflict, which is currently zero.

> And this is why I specified requiresExplicitNetwork in statics. While it may be a bit “uglier” code-wise, it shows, in a declarative way, (for the SDK user) which tokens they can instantiate with bitgo.coin(name) and which they must instantiate with bitgo.coin(name, network).

That's just pushing the problem down a level without really solving it. In order to know whether I need to pass the second param to the coin() function, I need to look it up in the `coins` map in statics, where the same problem exists.

samwise2:
I agree, important to consider blockchains more broadly. But this paradigm for ETH isn’t limited to just ETH, also ETH2 (I think), general EVM chains like Polygon, BSC, others. It also likely exists on some non-EVM chains, but I’m not sure.

In general, I would say that relying on hashing (by hashing, I mean using data as a key in a hashmap) values that we know can be non-unique i.e. token contract address, token symbols etc. is something that could cause scalability problems. 

Especially so if we use a data structure of exceptions that is produced at runtime. I think the user of the SDK should know, at compile time, which coins they can instantiate via identifier only, and which they must specify both identifier and network. 

I suppose that the user could put all of their `bitgo.coin` instantiations in a try..catch block, but personally, I think that giving the user all the info they need to preform correct instantiation at compile time, without the need for exception handling, would be nice.

samwise2:
Closing this for now. There's some good design discussion and approaches in here that can be re-visited later when this issue has higher priority.

PR #1601, opened 2021-10-22
feat(express): add external API signing to v2 API

0xabrarhussain:
ticket: BG-38025

```

      ✓ signs p2sh using external api (40ms)

  V2 Wallet External API:
    Transaction Signature Verification
      ✓ should sign a prebuild with external api


  2 passing (95ms)


```

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -1765,9 +1765,24 @@ export class Wallet {
         throw new Error('txPrebuild must be an object');
       }
       const presign = yield self.baseCoin.presignTransaction(params);
-      const userPrv = self.getUserPrv(presign);
-      const signingParams = _.extend({}, presign, { txPrebuild: txPrebuild, prv: userPrv });
-      return self.baseCoin.signTransaction(signingParams);
+      const externalSignerUrl = this.bitgo._externalSignerUrl;
+      if (externalSignerUrl) {
+        return fetch(externalSignerUrl, {
+          method: 'POST',
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify({
+            txPrebuild: txPrebuild,
+            presign: presign,
+          }),
+        }).then(response => response.json() as SignedTransaction);
+      } else {
+        const userPrv = self.getUserPrv(presign);
+        const signingParams = _.extend({}, presign, { txPrebuild: txPrebuild, prv: userPrv });
+        return self.baseCoin.signTransaction(signingParams);
+      }
     }).call(this).asCallback(callback);
   }
 
--- a/modules/core/src/wallet.ts
+++ b/modules/core/src/wallet.ts
@@ -929,6 +929,31 @@ Wallet.prototype.signTransaction = function (params, callback) {
     throw error;
   }
 
+  const externalSignerUrl = this.bitgo._externalSignerUrl;
+  if (externalSignerUrl) {
+    const transactionHex = params.transactionHex;
+    const unspents = params.unspents;
+    const keychain = params.keychain;
+    if (!_.isString(params.transactionHex)) {
+      throw new Error('expecting the transaction hex as a string');
+    }
+    return Bluebird.resolve(fetch(externalSignerUrl, {
+      method: 'POST',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        transactionHex: transactionHex,
+        unspents: unspents,
+        keychain: keychain,
+      }),
+    })).then(function (result) {
+      return result.json();
+    })
+      .nodeify(callback);
+  }
+
   params.validate = params.validate !== undefined ? params.validate : this.bitgo.getValidate();
   params.bitgo = this.bitgo;
   return TransactionBuilder.signTransaction(params)
--- a/modules/core/test/unit/externalApi.ts
+++ b/modules/core/test/unit/externalApi.ts
@@ -1,0 +1,112 @@
+const Wallet = require('../../src/wallet');
+import { TestBitGo } from '../lib/test_bitgo';
+import * as _ from 'lodash';
+import * as Bluebird from 'bluebird';
+const co = Bluebird.coroutine;
+import * as nock from 'nock';
+import * as common from '../../src/common';
+
+nock.disableNetConnect();
+const externalHost = 'https://externalsign.com';
+const externalApiPath = '/api/sign';
+
+
+describe('Wallet External API:', function () {
+
+  const userKeypair = {
+    xprv: 'xprv9s21ZrQH143K2fJ91S4BRsupcYrE6mmY96fcX5HkhoTrrwmwjd16Cn87cWinJjByrfpojjx7ezsJLx7TAKLT8m8hM5Kax9YcoxnBeJZ3t2k',
+    xpub: 'xpub661MyMwAqRbcF9Nc7TbBo1rZAagiWEVPWKbDKThNG8zqjk76HAKLkaSbTn6dK2dQPfuD7xjicxCZVWvj67fP5nQ9W7QURmoMVAX8m6jZsGp',
+    rawPub: '02c103ac74481874b5ef0f385d12725e4f14aedc9e00bc814ce96f47f62ce7adf2',
+    rawPrv: '936c5af3f8af81f75cdad1b08f29e7d9c01e598e2db2d7be18b9e5a8646e87c6',
+    path: 'm',
+    walletSubPath: '/0/0',
+  };
+  const backupKeypair = {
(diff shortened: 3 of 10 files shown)


modules/core/test/v2/unit/externalApi.ts
```diff
@@ -0,0 +1,158 @@
...
+            user: '543c11ed356d00cb7600000b98794503',
+            permissions: [
+              'admin',
+              'view',
+              'spend',
+            ],
+          },
+        ],
+        coin: 'tbch',
+        label: 'Signature Verification Wallet',
```
0xabrarhussain, line 82: copied over majority of this from an existing test, but assuming a lot of it isn't required solely for signing logic

modules/core/src/v2/wallet.ts
```diff
@@ -1765,9 +1765,24 @@ export class Wallet {
         throw new Error('txPrebuild must be an object');
       }
       const presign = yield self.baseCoin.presignTransaction(params);
-      const userPrv = self.getUserPrv(presign);
-      const signingParams = _.extend({}, presign, { txPrebuild: txPrebuild, prv: userPrv });
-      return self.baseCoin.signTransaction(signingParams);
+      const externalSignerUrl = this.bitgo._externalSignerUrl;
+      if (externalSignerUrl) {
+        return fetch(externalSignerUrl, {
```
tylerlevine, line 1770: please don't mix promise syntax, let's stick with `yield` for now instead of using `.then()`

modules/core/src/wallet.ts
```diff
@@ -929,6 +929,31 @@ Wallet.prototype.signTransaction = function (params, callback) {
...
 
+  const externalSignerUrl = this.bitgo._externalSignerUrl;
+  if (externalSignerUrl) {
+    const transactionHex = params.transactionHex;
+    const unspents = params.unspents;
+    const keychain = params.keychain;
+    if (!_.isString(params.transactionHex)) {
+      throw new Error('expecting the transaction hex as a string');
+    }
+    return Bluebird.resolve(fetch(externalSignerUrl, {
```
tylerlevine, line 940: Why use `fetch` instead of `superagent`, which is what we use for making HTTP requests elsewhere in the package?

modules/core/src/v2/wallet.ts
```diff
@@ -1765,9 +1765,24 @@ export class Wallet {
...
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify({
+            txPrebuild: txPrebuild,
+            presign: presign,
+          }),
+        }).then(response => response.json() as SignedTransaction);
+      } else {
```
tylerlevine, line 1781: No need for the `else` here, since the `if` case returns.

modules/core/src/v2/wallet.ts
```diff
@@ -1765,9 +1765,24 @@ export class Wallet {
...
+          method: 'POST',
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify({
+            txPrebuild: txPrebuild,
+            presign: presign,
+          }),
+        }).then(response => response.json() as SignedTransaction);
```
tylerlevine, line 1780: How do we know that this cast is safe? Should we do any validation on the returned object to make sure it looks correct before returning it?

modules/core/test/unit/externalApi.ts
```diff
@@ -0,0 +1,112 @@
...
+import * as common from '../../src/common';
+
+nock.disableNetConnect();
+const externalHost = 'https://externalsign.com';
+const externalApiPath = '/api/sign';
+
+
+describe('Wallet External API:', function () {
+
+  const userKeypair = {
```
tylerlevine, line 16: Please move all this hard coded data into a fixture file

modules/core/test/unit/externalApi.ts
```diff
@@ -0,0 +1,112 @@
+const Wallet = require('../../src/wallet');
+import { TestBitGo } from '../lib/test_bitgo';
+import * as _ from 'lodash';
+import * as Bluebird from 'bluebird';
+const co = Bluebird.coroutine;
+import * as nock from 'nock';
+import * as common from '../../src/common';
+
+nock.disableNetConnect();
+const externalHost = 'https://externalsign.com';
```
tylerlevine, line 10: please use an invalid TLD for testing, as per https://datatracker.ietf.org/doc/html/rfc2606#page-2

I prefer `.invalid`

modules/core/test/unit/externalApi.ts
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+  describe('Transaction Signature Verification', function () {
+    it('signs p2sh using external api', co(function *() {
+      const prodBitgo = new TestBitGo({ env: 'prod', externalSignerUrl: `${externalHost}${externalApiPath}` });
+      prodBitgo.initializeTestVars();
+      const bgUrl = common.Environments[prodBitgo.getEnv()].uri;
+      const fakeProdWallet = new Wallet(prodBitgo, { id: '2NCoSfHH6Ls4CdTS5QahgC9k7x9RfXeSwY4', private: { keychains: [userKeypair, backupKeypair, bitgoKey] } });
+
+      const firstSignature = '010000000144dea5cb05425f94976e887ccba5686a9a12a3f49710b021508d3d9cd8de16b801000000b600473044022021fa73d5fe61ac8942cd70ff4507c574677ce747de5bc46c3dd2e38ec2448fce022047906d2c0154337ab96041e8fb58c243b9bce5f8818fa991643c1260a1859ad80100004c695221031cd227e40ad61b4e137109cb2845eb6f5a584ed5c67d9d3135cdaa5045a842ea2103a2e7b54c7b2da0992555353b8e26c6acff4248f4351f08787bf3e2efc94b658321025c2a6cde33c2d73ccf12eecf64c54f08f722c2f073824498950695e9883b141253aeffffffff02e803000000000000116a0f426974476f2070327368207465737422a107000000000017a914d039cb3344294a5a384a5508a006444c420cbc118700000000';
```
tylerlevine, line 45: Please move hard coded data to test fixture. Tests should contain only the test logic, and the test data is just noise that isn't relevant to that.

modules/core/test/unit/externalApi.ts
```diff
@@ -0,0 +1,112 @@
...
+      transaction.transactionHex.should.equal('010000000144dea5cb05425f94976e887ccba5686a9a12a3f49710b021508d3d9cd8de16b80100000000ffffffff02e803000000000000116a0f426974476f2070327368207465737422a107000000000017a914d039cb3344294a5a384a5508a006444c420cbc118700000000');
+
+      // add first signature
+      const scope = nock(externalHost)
+        .post(externalApiPath)
+        .reply(200, { tx: firstSignature });
+      transaction.keychain = userKeypair;
+      const signature1 = (yield fakeProdWallet.signTransaction(transaction)) as any;
+      signature1.tx.should.equal(firstSignature);
+      scope.isDone().should.be.True();
```
tylerlevine, line 98:
```suggestion
      scope.done();
```

modules/core/test/unit/externalApi.ts
```diff
@@ -0,0 +1,112 @@
...
+      // add second signature
+      const scope2 = nock(externalHost)
+        .post(externalApiPath)
+        .reply(200, { tx: secondSignature });
+      transaction.transactionHex = signature1.tx;
+      transaction.keychain = backupKeypair;
+      transaction.fullLocalSigning = true;
+      const signature2 = (yield fakeProdWallet.signTransaction(transaction)) as any;
+      signature2.tx.should.equal(secondSignature);
+      scope2.isDone().should.be.True();
```
tylerlevine, line 109:
```suggestion
      scope2.done();
```

modules/core/test/unit/externalApi.ts
tylerlevine: Also, these are transactions and not signatures, so this naming is misleading. Also how come we have two separate transactions but only one test?

modules/core/test/v2/unit/externalApi.ts
```diff
@@ -0,0 +1,111 @@
...
+  before(async function () {
+    bitgo = new TestBitGo({ env: 'test', externalSignerUrl: `${externalHost}${externalApiPath}` });
+    bitgo.initializeTestVars();
+  });
+
+  describe('Transaction Signature Verification', function () {
+    let wallet: Wallet;
+    let basecoin;
+
+    const userKeychain = {
```
tylerlevine, line 24: Please move hard coded data into fixtures.

modules/core/test/v2/unit/externalApi.ts
```diff
@@ -0,0 +1,111 @@
+
+import * as assert from 'assert';
+
+import * as nock from 'nock';
+import { HalfSignedUtxoTransaction, Wallet } from '../../../src';
+import { TestBitGo } from '../../lib/test_bitgo';
+
+nock.disableNetConnect();
+const externalHost = 'https://externalsign.com';
```
tylerlevine, line 9: use invalid TLD here as well.

tylerlevine (changes requested):
nice work, this is quite close!

modules/core/src/v2/wallet.ts
tylerlevine: also same comment as later on about using fetch vs superagent.

PR #1614, opened 2021-10-27
feat(account-lib): addition txType on account-lib

Krin2:
addition txType on account-lib

STLX-7611

### DIFF
--- a/modules/account-lib/src/coin/algo/transaction.ts
+++ b/modules/account-lib/src/coin/algo/transaction.ts
@@ -229,6 +229,10 @@ export class Transaction extends BaseTransaction {
         result.nonParticipation = this._algoTransaction.nonParticipation;
       }
     }
+    if (result.type === 'axfer' && result.to && result.amount) {
+      result.txType = utils.getTokenTxType(result.amount, result.from, result.to, result.closeRemainderTo);
+      result.tokenName = this._coinConfig.suffix;
+    }
     return result;
   }
 
--- a/modules/account-lib/test/unit/coin/algo/transaction.ts
+++ b/modules/account-lib/test/unit/coin/algo/transaction.ts
@@ -2,14 +2,34 @@ import should from 'should';
 import { coins } from '@bitgo/statics';
 import * as AlgoResources from '../../../resources/algo';
 import { Transaction } from '../../../../src/coin/algo/transaction';
-import { KeyPair } from '../../../../src/coin/algo';
+import { KeyPair, AssetTransferBuilder, TransactionBuilderFactory } from '../../../../src/coin/algo';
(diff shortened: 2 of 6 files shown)


modules/account-lib/src/coin/algo/transaction.ts
```diff
@@ -229,6 +229,10 @@ export class Transaction extends BaseTransaction {
         result.nonParticipation = this._algoTransaction.nonParticipation;
       }
     }
+    if (result.type === 'axfer' && result.to && result.amount) {
```
JulioAchaval-SimTLiX, line 232:
```suggestion
    if (result.type === 'axfer' && result.to && result.amount && result.from) {
```

JulioAchaval-SimTLiX: I know this is a mandatory param, but an extra check could be great.

JulioAchaval-SimTLiX (approved):

smaria95 (approved):
LGTM

modules/account-lib/src/coin/algo/transaction.ts
Krin2: I only include parameters that are optional on TxData. Parameter "from" is required on TxData so it must be there in order to build this interfase or will break before this condition

modules/core/src/v2/coins/algo.ts
```diff
@@ -257,8 +257,14 @@ export class Algo extends BaseCoin {
 
         const isTokenTx = this.isTokenTx(txJson.type);
         if (isTokenTx) {
+          const type = accountLib.Algo.algoUtils.getTokenTxType(
```
rliberal-simtlix, line 260: Changes looks good, please create a generic version and test with the platform integration.

rliberal-simtlix (dismissed):

rliberal-simtlix (dismissed):

smaria95 (dismissed):

modules/account-lib/test/unit/coin/algo/transaction.ts
```diff
@@ -48,4 +51,128 @@ describe('Algo Transaction', () => {
       should.deepEqual(tx.canSign({ key: AlgoResources.accounts.account2.secretKey.toString('hex') }), true);
     });
   });
+
+  describe('Transaction type', () => {
+    it('field txType and tokenName are added on enableToken transaction', async () => {
```
pancham88, line 56: Please build the common part of tx in `beforeEach` so that there is less repetition of code.

pancham88 (changes requested):

pancham88 (approved):

PR #1615, opened 2021-10-28
fix(eos): fix issue verifying EOS transactions

joey-bai:
**fixes the following issues**:
- we would always set expiration header's milliseconds to 0 when we
  should be rounding (this will fail verification if milliseconds is > 500)
- we set memoId to undefined when absent instead of empty string (leads to `undefined !== ''` check)
- we incorrectly parse for token symbol (`coin` is always teos or eos; need to check `token` property instead)

Ticket: BG-38309

additional manual verification:
EOS
```
async function test() {
  const wallet = await bitgo.coin('teos').wallets().get({ id: '61521886121c030006d523b8553d972a' });
  const res = await wallet.prebuildAndSignTransaction({
    recipients: [
      {
        amount: '1000',
        address: 'lionteste212',
      },
    ],
    walletPassphrase: passphrase,
  });
}
```

Token
```
async function test() {
  const wallet = await bitgo.coin('teos:CHEX').wallets().get({ id: '61521886121c030006d523b8553d972a' });
  const res = await wallet.prebuildAndSignTransaction({
    recipients: [
      {
        amount: '1000',
        address: 'lionteste212',
      },
    ],
    walletPassphrase: passphrase,
  });
}
```

### DIFF
--- a/modules/core/src/v2/coins/eos.ts
+++ b/modules/core/src/v2/coins/eos.ts
@@ -70,6 +70,8 @@ interface EosTransactionPrebuild {
   transaction: EosTx;
   txid: string;
   coin: string;
+  // full token name with the format, [t]eos:SYMBOL. This will only be present for token transactions. e.g. teos:CHEX.
+  token?: string;
 }
 
 export interface EosSignTransactionParams extends BaseSignTransactionOptions {
@@ -319,7 +321,7 @@ export class Eos extends BaseCoin {
     if (destinationDetails.pathname === address) {
       return {
         address: address,
-        memoId: undefined,
+        memoId: '',
       };
     }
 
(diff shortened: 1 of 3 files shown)


modules/core/src/v2/coins/eos.ts
```diff
@@ -319,7 +321,7 @@ export class Eos extends BaseCoin {
     if (destinationDetails.pathname === address) {
       return {
         address: address,
-        memoId: undefined,
+        memoId: '',
```
pancham88, line 324: On the platform side, we set it to `null`.

https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/v2/coins/eos.js#L360

Should it be same here in sdk as well?

modules/core/src/v2/coins/eos.ts
```diff
@@ -1067,10 +1069,14 @@ export class Eos extends BaseCoin {
       const txJsonFromHexHeaders = _.pick(txJsonFromHex, eosTransactionHeaderFields);
       const txJsonFromPackedTrxHeaders = _.pick(txJsonFromPackedTrx, eosTransactionHeaderFields);
 
-      // milliseconds are dropped in packed_trx and txHex
+      // dates are rounded to the nearest second in packed_trx and txHex
       _.map([txJsonFromPackedTrxHeaders, txJsonFromHexHeaders, txPrebuild.headers], (headers) => {
         const date = moment(headers.expiration as string);
-        headers.expiration = date.milliseconds(0).toISOString();
+
+        headers.expiration = date
```
pancham88, line 1076: Please refer to the logic of setting the expiration header in platform:

https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/v2/coins/eos.js#L1748

The two shall be similar.

modules/core/src/v2/coins/eos.ts
joey-bai: this might be an issue - wasn't aware that this code was duplicated here as well. will need to update this to match the SDK.

the problem with `verifyTransaction` is that when we deserialize a transaction without a memoId, it comes back as empty string instead of undefined.

PR #1616, opened 2021-10-28
feat(account-lib): initial setup

gibrangul95:
@argjv @fdmirza

### DIFF
--- a/modules/account-lib/src/coin/dot/keyPair.ts
+++ b/modules/account-lib/src/coin/dot/keyPair.ts
@@ -1,0 +1,29 @@
+import { DefaultKeys, KeyPairOptions } from '../baseCoin/iface';
+import { Ed25519KeyPair } from '../baseCoin';
+import { AddressFormat } from '../baseCoin/enum';
+
+export class KeyPair extends Ed25519KeyPair {
+  /**
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   * @param { KeyPairOptions } source Either a master seed, a private key, or a public key
+   */
+  constructor(source?: KeyPairOptions) {
+    super(source);
+  }
+
+  getAddress(format?: AddressFormat): string {
+    throw new Error('Method not implemented.');
+  }
+
+  getKeys(): DefaultKeys {
+    throw new Error('Method not implemented.');
+  }
+
+  recordKeysFromPrivateKeyInProtocolFormat(prv: string): DefaultKeys {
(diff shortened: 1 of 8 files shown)


jrafael-simtlix (dismissed):

modules/account-lib/src/coin/dot/keyPair.ts
```diff
@@ -0,0 +1,13 @@
+import { Secp256k1ExtendedKeyPair } from '../baseCoin/secp256k1ExtendedKeyPair';
+import { KeyPairOptions } from '../baseCoin/iface';
+
+export class KeyPair extends Secp256k1ExtendedKeyPair {
```
argjv, line 4: Isn't DOT ed25519?

argjv (approved):

tylerlevine (dismissed):
1) we can't currently accept PRs from forks directly. Please commit to a branch inside this repository if possible
2) The commit message is lacking in information - *what* is getting the initial setup? Please leave more info for posterity along with a ticket number

modules/account-lib/src/coin/dot/utils.ts
```diff
@@ -0,0 +1,34 @@
+import { BaseUtils } from '../baseCoin';
+import { NotImplementedError } from '../baseCoin/errors';
```
rliberal-simtlix, line 2: If we are going to encode and decode address with keyring, please add unimplemented methods for encode and decode.
https://polkadot.js.org/docs/keyring/start/ss58#how-is-the-encoding-done

modules/account-lib/src/coin/dot/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+
```
rliberal-simtlix, line 3: Should we derive accounts with the platform? @argjv 
https://polkadot.js.org/docs/ui-keyring/start/accounts/#non-mnemonic-seeds-and-derivation

rliberal-simtlix (commented):
Added my review, but many funcionality on my comment was implemented on:
https://github.com/BitGo/BitGoJS/pull/1632/files
So we need here to validate what funcionality is left.
Please mark comment as resolved for which funcionality is implemented.

modules/account-lib/src/coin/dot/utils.ts
argjv: Derivation should be done in the KeyPair, but we don't really need it for DOT

PR #1617, opened 2021-10-28
feat(statics): change name to gHDO and gHCN

facundoCarreSimtlix:
change name to gHDO and gHCN

stlx-6767


tylerlevine (commented):
Did we verify that there are no existing transfers for this coin?

tylerlevine (approved):

PR #1622, opened 2021-10-29
feat(utxo-lib): use p2tr-p2ns redeems for outputs

sangaman:
This uses the recently implemented `p2tr-p2ns` redeem scripts for p2tr output script creation.

Ticket: BG-35573

### DIFF
--- a/modules/utxo-lib/src/bitgo/outputScripts.ts
+++ b/modules/utxo-lib/src/bitgo/outputScripts.ts
@@ -70,7 +70,9 @@ export function createOutputScript2of3(pubkeys: Buffer[], scriptType: ScriptType
   if (scriptType === 'p2tr') {
     // p2tr addresses use a combination of 2 of 2 multisig scripts distinct from
     // the 2 of 3 multisig used for other script types
-    return createTaprootScript2of3(pubkeys);
+    const { output } = createTaprootScript2of3(pubkeys);
+    assert(output);
+    return { scriptPubKey: output };
   }
 
   const script2of3 = bitcoinjs.payments.p2ms({ m: 2, pubkeys });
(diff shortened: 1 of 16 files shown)


modules/utxo-lib/src/bitgo/outputScripts.ts
```diff
@@ -115,25 +116,47 @@ export function createOutputScript2of3(pubkeys: Buffer[], scriptType: ScriptType
...
+ * path spend using one of the three n of n scripts in each taproot's taptree.
+ * @param pubkeys - a pubkey array containing the user key, backup key, and bitgo key in that order
+ * @param redeemIndex 
+ * @returns 
+ */
+export function prepareP2trP2nsWitness(pubkeys: Triple<Buffer>, redeemIndex = 0) {
+  const {
+    controlBlock,
+    redeem,
+    output,
```
sangaman, line 150: `output` won't be necessary if/when `TransactionBuilder` derives it on its own from the control block and tapscript.

sangaman:
This branch was used to create a bona fide, legit, fully signed taproot script path 2 of 2 multi sig spend transaction: https://blockstream.info/testnet/tx/62cd8a14a20a18f8e7fff9dcaaa37f73e9feba3c4b16c20e90c8828ccbcb2fff?expand - The tx hex for that is used in the test case added in this PR. It's a userKey + backupKey key combination and both signatures were created in this branch and added to a prebuilt, unsigned tx hex.

I will clean up the commits and try to improve test coverage to get this out of draft state tomorrow. The last key remaining piece of functionality is actually verifying the schnorr signatures after we create them, for that we need to export the `schnorrBip340` module from bitcoinjs-lib which I also aim to do tomorrow.

sangaman:
Since a good chunk of the existing commits/code in this branch have been used or modified in  #1626 and #1637, and since the remaining changes have drifted from the purpose of the original ticket number and commit, I'm going to close this PR and open a different one with the latest code from here rebased on master.

PR #1624, opened 2021-11-01
feat(statics): add testnet tokens

facundoCarreSimtlix:
add testnet tokens

stlx-8714


modules/statics/src/base.ts
```diff
@@ -361,6 +361,17 @@ export enum UnderlyingAsset {
   GOLD = 'gold',
   GOT = 'got',
   GRT = 'grt',
+  GTAAVE18DP = 'GTAAVE18DP',
```
doodles92, line 364: i think to keep it consistent, we need to lowercase the quotes. e.g. `GTAAVE18DP = 'gtaave18dp'

modules/statics/src/coins.ts
```diff
@@ -871,6 +871,127 @@ export const coins = CoinMap.fromCoins([
...
   ),
+  terc20(
+    'gtaave18dp',
+    'Goerli Test AAVE Token 18 Decimals',
+    18,
+    '0x631d5e3c45a459e8f98b9d6a2734fce7b051f845',
+    UnderlyingAsset.GTAAVE18DP,
+    undefined,
+    undefined,
+    undefined,
```
doodles92, line 882: why is this undefined?

facundoCarreSimtlix: image
because I use this function and I need to pass the last parameter (network) and the other 3 (features, prefix, suffix) I don't use them

samwise2 (approved):

samwise2:
Maybe wait on merge though for someone who has context on your naming etc.

tylerlevine (approved):

tylerlevine (approved):

PR #1626, opened 2021-11-02
feat: add signInput2Of3()

OttoAllmendinger:
Simplifies p2tr signing.

The `signInput2Of3()` wraps `txBuilder.sign()` and selects the correct 
parameters depending on scriptType, signer and cosigner.

Issue: BG-35578

9ad8349ec (Otto Allmendinger, 5 minutes ago)
feat(utxo): update createTaprootScript2of3

Use updated bitcoinjs-lib interface and update test vectors.

Issue: BG-35578


modules/utxo-lib/src/bitgo/outputScripts.ts
```diff
@@ -29,10 +28,10 @@ export function scriptType2Of3AsPrevOutType(t: ScriptType2Of3): string {
 
 export type SpendableScript = {
   scriptPubKey: Buffer;
+  /** @deprecated - use createSpendScripts2of3 */
```
OttoAllmendinger, line 31: that func is TODO

OttoAllmendinger:
p2tr support preview: https://github.com/BitGo/BitGoJS/commit/5a4230795c58c8acebf0e636bc46124973c7ea4e

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -360,3 +370,41 @@ export function verifySignature(
...
+  amount: number
+): void {
+  let prevOutScriptType;
+  let redeemScript;
+  let witnessScript;
+  let witnessValue: number | undefined = amount;
+
+  if (isScriptType2Of3(scriptType)) {
+    prevOutScriptType = scriptType2Of3AsPrevOutType(scriptType);
+    ({ redeemScript, witnessScript } = createOutputScript2of3(pubkeys, scriptType));
```
sangaman, line 390: For taproot I think we'll want a redeem/tapscript value chosen by redeem index. These would both be undefined I think and tx builder wouldn't be able to sign without the tapscript if I'm following this correctly.

sangaman (commented):
The changes to outputscripts.ts and the test fixtures json looks like overlap with #1622 (although I like the added `toXOnlyPublicKey` helper function). I like the `signInput2Of3` pattern but see my comment about signing for p2tr script path.

modules/utxo-lib/src/bitgo/signature.ts
OttoAllmendinger: instead of selecting by redeem index (which is a little opaque) my suggestion is to select the controlBlock/witnessScript by the cosigner argument: https://github.com/BitGo/BitGoJS/commit/5a4230795c58c8acebf0e636bc46124973c7ea4e

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -360,3 +370,41 @@ export function verifySignature(
...
+  } else if (scriptType === 'p2shP2pk') {
+    prevOutScriptType = 'p2sh-p2pk';
+    ({ redeemScript, witnessScript } = createOutputScriptP2shP2pk(keyPair.publicKey));
+    witnessValue = undefined;
```
OttoAllmendinger, line 394: the support for `p2shP2pk` makes the name of the function slightly inaccurate

maybe I should rename it to just `signInput`

modules/utxo-lib/src/bitgo/outputScripts.ts
```diff
@@ -108,31 +107,51 @@ export function createOutputScript2of3(pubkeys: Buffer[], scriptType: ScriptType
...
+function getTaptreeKeyCombinations(keys: Triple<Buffer>): Tuple<Buffer>[] {
+  const [userKey, backupKey, bitGoKey] = keys.map((k) => toXOnlyPublicKey(k));
+  return [
+    [userKey, bitGoKey],
+    [userKey, backupKey],
+    [backupKey, bitGoKey],
+  ];
+}
+
+export function createPaymentP2tr(pubkeys: Triple<Buffer>, redeemIndex?: number): bitcoinjs.Payment {
```
OttoAllmendinger, line 129: add comment about usage (unit tests)

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -360,3 +370,44 @@ export function verifySignature(
 
   return signatureVerifications.length > 0 && signatureVerifications.every((v) => v.signedBy !== undefined);
 }
+
+export function signInputP2shP2pk(txBuilder: UtxoTransactionBuilder, vin: number, keyPair: bip32.BIP32Interface): void {
```
OttoAllmendinger, line 374: TODO: add comment about usage

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -360,3 +370,44 @@ export function verifySignature(
...
+  const prevOutScriptType = scriptType2Of3AsPrevOutType(scriptType);
+  const { redeemScript, witnessScript } = createOutputScript2of3(pubkeys, scriptType);
+
+  keyPair.network = txBuilder.network;
+
+  txBuilder.sign({
+    vin,
+    prevOutScriptType,
+    keyPair,
+    hashType: getDefaultSigHash(txBuilder.network as Network),
```
OttoAllmendinger, line 408: TODO: add second parameter `scriptType` and return `SIGHASH_DEFAULT` for `p2tr`

brandonblack (approved):

PR #1627, opened 2021-11-02
test(utxo-lib): add support for bitcoin core 22

OttoAllmendinger:
More recent bitcoin core versions do not have a default wallet set and
require `loadwallet` and `createwallet` calls.

The current bitcoind git master version also requires a scoped RPC
path for wallet operations.

This change supports Bitcoin 22.x.x in a backwards-compatible manner.

Issue: BG-38416

### DIFF
--- a/modules/utxo-lib/test/integration_local_rpc/generate/RpcClient.ts
+++ b/modules/utxo-lib/test/integration_local_rpc/generate/RpcClient.ts
@@ -16,40 +16,98 @@ function sleep(millis: number): Promise<void> {
   });
 }
 
+export class RpcError extends Error {
+  constructor(public rpcError: { code: number; message: string }) {
+    super(`RPC error: ${rpcError.message} (code=${rpcError.code})`);
+  }
+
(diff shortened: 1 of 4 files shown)


tylerlevine (approved):

modules/utxo-lib/test/integration_local_rpc/generate/RpcClient.ts
```diff
@@ -16,40 +16,98 @@ function sleep(millis: number): Promise<void> {
...
 
       throw e;
     }
   }
 
+  requiresWalletPath(): boolean {
+    if (!this.networkInfo) {
+      throw new Error(`networkInfo must be set`);
+    }
+    return this.networkInfo.subversion === BITCOIN_CORE_22_99;
```
sangaman, line 95: Why the 99 version here and not 22.0.0 as well?

OttoAllmendinger: the `22.0.0` version does not require the wallet path in the RPC URL

PR #1637, opened 2021-11-03
feat(utxo-lib): add support for p2tr in bitgo.signInput2Of3()

OttoAllmendinger:
Issue: BG-35578


modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -55,14 +56,21 @@ export interface ParsedSignatureScript2Of3 extends ParsedSignatureScript {
   pubScript: Buffer;
 }
 
-export function getDefaultSigHash(network: Network): number {
+export interface ParsedSignatureScriptTaproot extends ParsedSignatureScript {
+  // P2TR tapscript spends are for 2-of-2 multisig scripts
+  signatures: [Buffer | 0, Buffer | 0]; // missing signature is encoded as OP_0
```
brandonblack, line 61: @sangaman ran into a situation where missing signature was encoded as `Buffer.of()` instead of `0`.

brandonblack: Don't think the difference matters here, come to think of it.

OttoAllmendinger: I think `Buffer.of()` is the correct encoding

brandonblack (dismissed):

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -316,29 +378,80 @@ export function getSignatureVerifications(
...
   const publicKeys = parsedScript.publicKeys.filter(
-    (buf) => verificationSettings.publicKey === undefined || verificationSettings.publicKey.equals(buf)
+    (buf) =>
+      verificationSettings.publicKey === undefined ||
+      verificationSettings.publicKey.equals(buf) ||
+      verificationSettings.publicKey.slice(1).equals(buf)
   );
 
+  if (verificationSettings.publicKey) {
+    debugger;
```
tylerlevine, line 388: should be removed

tylerlevine (changes requested):
looks really good, just remove the `debugger` statement and I'm good to go

tylerlevine (dismissed):

tylerlevine (approved):

PR #1639, opened 2021-11-03
feat(core): create wallet with eip1559

micaela-ledesma:
Create wallet with eip1559 fee params

bg-38441

### DIFF
--- a/modules/core/src/v2/wallets.ts
+++ b/modules/core/src/v2/wallets.ts
@@ -12,6 +12,7 @@ import * as _ from 'lodash';
 import { RequestTracer } from './internal/util';
 import { sanitizeLegacyPath } from '../bip32path';
 import { getSharedSecret } from '../ecdh';
+import { BigNumber } from 'bignumber.js';
 
 const co = Bluebird.coroutine;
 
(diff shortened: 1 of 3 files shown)


modules/core/src/v2/wallets.ts
```diff
@@ -336,10 +341,26 @@ export class Wallets {
       }
 
       if (!_.isUndefined(params.gasPrice)) {
-        if (!_.isNumber(params.gasPrice)) {
+        const gasPriceBN = new BigNumber(params.gasPrice);
```
wesgraham, line 344: can we also throw if gasPrice is not a number or a string? 
want to avoid weird BigNum output if we pass gasPrice as a boolean or something

modules/core/src/v2/wallets.ts
```diff
@@ -336,10 +341,26 @@ export class Wallets {
...
+        if (gasPriceBN.isNaN()) {
           throw new Error('invalid gas price argument, expecting number');
         }
-        walletParams.gasPrice = params.gasPrice;
+        walletParams.gasPrice = gasPriceBN.toString();
+      }
+
+      if (!_.isUndefined(params.eip1559) && !_.isEmpty(params.eip1559)) {
```
wesgraham, line 351: if gasPrice and eip1559 are passed what happens?

micaela-ledesma: Adding validation in case of sending both fees

modules/core/src/v2/wallets.ts
micaela-ledesma: I modified the message, and it appeared correctly in the cases of boolean and string with letters.

wesgraham (approved):

PR #1640, opened 2021-11-03
fix(eos): fix deserialize transaction with OVC

joey-bai:
OVC uses Electron to package their application. There is a known issue
with Electron apps crashing when using TextDecoder with large inputs.
Workaround is to implement our own decoder using StringDecoder.

Ticket: BG-38309

Verified OVC works with locally built BitGoJS
Screen Shot 2021-11-03 at 1 44 47 PM
:

### DIFF
--- a/modules/core/src/stringTextDecoder.ts
+++ b/modules/core/src/stringTextDecoder.ts
@@ -1,0 +1,22 @@
+/**
+ * @prettier
+ */
+import { StringDecoder } from 'string_decoder';
+
+/**
+ * Custom utf8 TextDecoder that uses StringDecoder under the hood.
+ * This should only be used with utf8 and does NOT support TextDecoder options, like streaming.
+ */
+export class StringTextDecoder extends TextDecoder {
+  public decode(input?: BufferSource, options?: TextDecodeOptions): string {
+    // Note: streaming is not necessary for deserializing EOS transactions.
+    const decoder = new StringDecoder('utf8');
+
+    if (input) {
+      const decoded = decoder.end(Buffer.from(input));
+      return decoded;
+    }
+
+    return '';
+  }
+}
--- a/modules/core/src/v2/coins/eos.ts
+++ b/modules/core/src/v2/coins/eos.ts
@@ -26,6 +26,7 @@ import * as querystring from 'querystring';
 import * as _ from 'lodash';
 import * as Bluebird from 'bluebird';
 import { OfflineAbiProvider } from './eosutil/eosabiprovider';
+import { StringTextDecoder } from '../../stringTextDecoder';
 const co = Bluebird.coroutine;
 import { InvalidAddressError, UnexpectedAddressError } from '../../errors';
 import { Environments } from '../environments';
@@ -541,7 +542,8 @@ export class Eos extends BaseCoin {
         rpc: new NoopJsonRpc(),
         signatureProvider: new NoopSignatureProvider(),
         chainId: self.getChainId(),
-        textDecoder: new TextDecoder(),
+        // Use a custom TextDecoder as the global TextDecoder leads to crashes in OVC / Electron.
+        textDecoder: new StringTextDecoder(),
         textEncoder: new TextEncoder(),
       });
 
@@ -579,6 +581,7 @@ export class Eos extends BaseCoin {
       // deserializeTransaction
       const serializedTxBuffer = Buffer.from(transaction.packed_trx, 'hex');
       const deserializedTxJsonFromPackedTrx = yield api.deserializeTransactionWithActions(serializedTxBuffer);
+
       if (!deserializedTxJsonFromPackedTrx) {
         throw new Error('could not process transaction from txHex');
       }
--- a/modules/core/test/v2/fixtures/coins/eos.ts
+++ b/modules/core/test/v2/fixtures/coins/eos.ts
@@ -374,6 +374,50 @@ const explainRefundOutput = {
   producers: undefined,
 };
 
+const unsignedTransaction = {
+  recipients: [
+    {
+      amount: '560000',
+      address: 'jzjkpn1bjnti',
+      recipientLabel: null,
+      memo: {
+        type: 'id',
+        value: 'Testing EOS Memo',
+      },
+    },
+  ],
+  headers: {
+    expiration: '2019-07-27T02:47:20.500Z',
+    ref_block_num: 21495,
+    ref_block_prefix: 3447713214,
+  },
+  txHex: 'e70aaab8997e1dfce58fbfac80cbbb8fecec7b99cf982a9444273cbc64c4147339bb3b5df753beed7fcd000000000100a6823403ea3055000000572d3ccdcd012012a6f68a5ed4bf00000000a8ed3232212012a6f68a5ed4bfe0f27c27cc0adf7f808b08000000000004454f530000000000000000000000000000000000000000000000000000000000000000000000000000',
+  transaction: {
+    compression: 'none',
+    packed_trx: '39bb3b5df753beed7fcd000000000100a6823403ea3055000000572d3ccdcd012012a6f68a5ed4bf00000000a8ed3232212012a6f68a5ed4bfe0f27c27cc0adf7f808b08000000000004454f53000000000000',
+    signatures: [],
+  },
+  txid: '8bb4dc888bad8108a6c2f1bb8a2b74fbd8f68fa7216f86545786850517754e25',
+  walletId: '5d39c7da66ce12715470c9b2fd6a639d',
+  amount: '560000',
+  address: 'jzjkpn1bjnti',
+  receivedCoin: {
+    name: 'teos',
+    svg: 'eos',
+    modifier: 10000,
+    modifierExp: 4,
+    fullDisplay: 'EOS',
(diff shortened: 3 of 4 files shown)


modules/core/src/v2/coins/eos.ts
```diff
@@ -541,7 +542,7 @@ export class Eos extends BaseCoin {
         rpc: new NoopJsonRpc(),
         signatureProvider: new NoopSignatureProvider(),
         chainId: self.getChainId(),
-        textDecoder: new TextDecoder(),
+        textDecoder: new EosTextDecoder(),
         textEncoder: new TextEncoder(),
```
0xabrarhussain, line 546: seems pretty likely here that the textEncoder is also going to have the same issues? I'm assuming we're not running into an error because the encoder here isn't being used? but for consistency sake, we should should try to also make this a `stringEncoder` under the hood so that the encoder/decoder work as proper inverses of each other

shihao-guo: we can have a test covering the encoding and decoding phase with various tools

modules/core/src/v2/coins/eosutil/eosTextDecoder.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Custom utf8 TextDecoder that uses StringDecoder. This is used to deserialize EOS transactions.
+ *
+ * OVC uses Electron to package their apps. Electron crashes with the global TextDecoder when decoding large inputs.
+ * Our work around is to implement our own TextDecoder using StringDecoder instead.
+ *
+ * Electron issue: https://github.com/electron/electron/issues/18733
+ */
+export class EosTextDecoder extends TextDecoder {
```
shihao-guo, line 14: why don't we use call it stringDecoder and allow it to be used in a generic way? Not sure why we need to tie this function to EOS.

modules/core/src/v2/coins/eos.ts
joey-bai: @0xabrarhussain `StringEncoder` doesn't actually exist.

Looking at [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder), its primarily used to transform utf8 strings into uint8arrays.

Our `EosTextDecoder` is implemented to do the opposite - data back to utf8 strings.

modules/core/test/v2/unit/coins/eosutil/eosTextDecoder.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * @prettier
+ */
+
+import { EosTextDecoder } from '../../../../../src/v2/coins/eosutil/eosTextDecoder';
+import 'should';
+
+describe('Eos Text Decoder', function () {
```
shihao-guo, line 8: i think we also need to test if both string & test decoder can encode and decode the same message with the same output each time. Ideally we use a sample EOS transaction for this test.

modules/core/src/v2/coins/eos.ts
```diff
@@ -541,7 +542,7 @@ export class Eos extends BaseCoin {
         rpc: new NoopJsonRpc(),
         signatureProvider: new NoopSignatureProvider(),
         chainId: self.getChainId(),
-        textDecoder: new TextDecoder(),
+        textDecoder: new EosTextDecoder(),
```
shihao-guo, line 545: is this function also used by other services like TAT or platform-api?

modules/core/src/v2/coins/eos.ts
shihao-guo: oh wow this behavior is pretty confusing, can we have some comments on it to make sure people won't misuse it?

shihao-guo: also what is TestDecoder doing? cause we are swapping it with EosTextDecoder, 

I'm guessing 
TestDecoder - uint8arrays into utf8 strings 
EosTextDecoder - uint8arrays into utf8 strings too?

modules/core/test/v2/fixtures/coins/eos.ts
```diff
@@ -208,7 +208,7 @@ const explainTransactionInputNative = {
   headers: {
     ref_block_prefix: 686520069,
     ref_block_num: 3530,
-    expiration: '2021-10-04T16:41:53'
+    expiration: '2021-10-04T16:41:53',
```
shihao-guo, line 211: oh wow I like how we are modifying test. At the same time, was this file breaking test before?

shihao-guo (approved):

modules/core/src/v2/coins/eos.ts
joey-bai: @shihao-guo In short, yes, we want `TextDecoder` and `EosTextDecoder` to produce identical results.

`TextDecoder` is a standard class provided in javascript. However, its implementation leads to crashes in OVC (underlying problem is in Electron, the framework used to package OVC into an executable).

To get around this issue, we'll be switching to Node's `StringDecoder` instead. But since eosjs expects a TextDecoder, we have to subclass it.

modules/core/test/v2/fixtures/coins/eos.ts
shihao-guo: or does it mean the encoding and decoding is breaking our previous format?

modules/core/test/v2/unit/coins/eosutil/eosTextDecoder.ts
joey-bai: Updated - added a test that encodes a packed_trx hex string.
then decodes it back using both StringTextDecoder and TextDecoder

modules/core/src/v2/coins/eosutil/eosTextDecoder.ts
joey-bai: Updated - there are some limitations - this only supports utf8 and does not support streaming. but shouldn't be a problem.

added comment to make this clear.

modules/core/test/v2/fixtures/coins/eos.ts
joey-bai: @shihao-guo no - tests were always passing here (i think most of these changes were formatting from prettier lol). for most use cases, using the standard TextDecoder is fine.

there are only problems when using it within an Electron app.

modules/core/src/v2/coins/eos.ts
joey-bai: @shihao-guo did a code search in bitgo-ms and it's used primarily for `send` and `stake` transactions.
but staking is only supported for cspr and stx.

modules/core/test/v2/fixtures/coins/eos.ts
shihao-guo: i would recommend revert the change to the test file here if we can't explain why it needs to change

joey-bai: looked it over again - all the changes are adding a trailing comma to JSON dicts.

the only intentional change is adding `unsignedTransaction`

shihao-guo: got it. I was referring to those commas added by prettier

modules/core/src/v2/coins/eos.ts
0xabrarhussain: can we try to retrieve encoder as `util.TextEncoder` ? or does that also result in type errors

joey-bai: @0xabrarhussain this will lead to type errors. We would have to do something similar - subclass TextEncoder and override `encode` to use `util.TextEncoder`.

modules/core/test/v2/fixtures/coins/eos.ts
joey-bai: reverted commas

modules/core/src/stringTextDecoder.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * @prettier
+ */
+import { StringDecoder } from 'string_decoder';
+
+/**
+ * Custom utf8 TextDecoder that uses StringDecoder under the hood.
+ * This should only be used with utf8 nad does NOT support TextDecoder options, like streaming.
```
shiksb, line 8: nit: typo `nad`

modules/core/src/stringTextDecoder.ts
```diff
@@ -0,0 +1,22 @@
...
+  public decode(input?: BufferSource, options?: TextDecodeOptions): string {
+    // Note: streaming is not necessary for deserializing EOS transactions.
+    const decoder = new StringDecoder('utf8');
+
+    if (input) {
+      const decoded = decoder.end(Buffer.from(input));
+      return decoded;
+    } else {
+      return '';
+    }
```
shiksb, line 20:
```suggestion
    }
    return '';
```

shihao-guo (dismissed):

shiksb (approved):
I have a couple of non-blocking nits, but everything else LGTM

modules/core/src/stringTextDecoder.ts
joey-bai: good catch, updated

shihao-guo (approved):

shiksb (approved):

PR #1641, opened 2021-11-04
feat(core): add support for p2tr script path sign 

sangaman:
feat(core): add support for p2tr script path sign 

This adds support for p2tr script path signing for btc via the `prebuildAndSignTransaction` interactive flow with the BitGo API. A test case is added for p2tr transaction signing starting with a prebuilt, unsigned transaction using a single p2tr input and then signing it twice with the user and backup keys via a tapscript embedded in the taptree. The resulting signed txHex represents a valid bitcoin transaction that has been broadcast and confirmed on the live testnet btc chain.

build(core): bump @bitgo/unspents dependency 

This upgrades to the latest @bitgo/unspents dependency which introduces support for p2tr key path and script path input types. It also has subtle changes to the way fees/dimensions are calculated for previously supported input types, which required a number of changes to existing test cases and fixtures to match the latest logic.

Ticket: BG-35578

Note: This is a continuation of #1622 which has been rebased.


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1120,6 +1133,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     callback?: NodeCallback<SignedTransaction | HalfSignedUtxoTransaction>
   ): Bluebird<SignedTransaction | HalfSignedUtxoTransaction> {
     const self = this;
+    params;
```
OttoAllmendinger, line 1136: :scissors:

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1192,22 +1207,72 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+            .derivePath(sanitizeLegacyPath(signatureContext.path)).publicKey;
+
+          const { controlBlock, redeem, output } = utxolib.bitgo.outputScripts.createPaymentP2tr(
+            [userPubkey, backupPubkey, bitgoPubkey],
+            params.taprootRedeemIndex ?? 0
+          );
+          prevOutputs.push({
+            prevOutScript: output!,
+            value: signatureContext.unspent.value,
+          });
```
OttoAllmendinger, line 1237: we won't reach this for replay protection puts

let's use `tx.ins.map()` instead

OttoAllmendinger (changes requested):

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1192,22 +1207,72 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+          const scriptType = witnessScript ? (redeemScript ? 'p2shP2wsh' : 'p2wsh') : 'p2sh';
+          debug(`Signing ${scriptType} input`);
+          try {
+            txb.sign(index, keyPair, redeemScript, sigHashType, signatureContext.unspent.value, witnessScript);
+          } catch (e) {
+            debug('Failed to sign input:', e);
+            signatureContext.error = e;
+            signatureIssues.push(signatureContext);
+            continue;
+          }
```
OttoAllmendinger, line 1273: I'm working on a change that uses `utxolib.bitgo.signInput2Of3()` instead

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: for now let's go with a simpler change though

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -183,6 +184,10 @@ export interface SignTransactionOptions extends BaseSignTransactionOptions {
     };
   };
   prv: string;
+  userKeychain?: Keychain;
+  backupKeychain?: Keychain;
+  bitgoKeychain?: Keychain;
+  taprootRedeemIndex?: number;
```
OttoAllmendinger, line 190: I propose we add an argument `cosigner: Keychain` instead.

Then we can abstract away from the taproot tree structure and don't have to memorize the correct index. I will propose that change in my PR.

OttoAllmendinger, line 190: `Keychain` is a bit too constraining here, for instance in tests we sometimes only have string xprv/xpubs 

`keychains: [string, string, string]` (extended pubkey format) is more convenient and matches the `prv` argument

OttoAllmendinger:
we are not passing `prevOutputs` in `AbstractUtxoCoin.prototype.explainTransaction()` yet, but we can fix that later

OttoAllmendinger (approved):
I would like to make a few changes but I will submit a separate PR for that

brandonblack (dismissed):

modules/core/src/v2/coins/abstractUtxoCoin.ts
sangaman: Yeah that's fine, I was thinking of refactoring to use that but ran out of time/focus.

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: hmm sadly we are bound to the baseCoin signature

brandonblack (dismissed):

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -49,6 +49,7 @@ import { NodeCallback } from '../types';
 import { Wallet } from '../wallet';
 import { toBitgoRequest } from '../../api';
 import { sanitizeLegacyPath } from '../../bip32path';
+import { PrevOutput } from '../../../../utxo-lib/dist/src/bitgo';
```
tylerlevine, line 52: We can't use relative imports in the SDK like this, since we don't have guarantees that this path is valid at runtime. Please use 
```
import { bitgo } from '@bitgo/utxo-lib';
import PrevOutput = bitgo.PrevOutput;
```

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1192,22 +1212,74 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+          const signParams = {
+            controlBlock,
+            vin: index,
+            prevOutScriptType: 'p2tr-p2ns',
+            keyPair: keyPair,
+            hashType: sigHashType,
+            witnessScript: redeem?.output,
+            witnessValue: signatureContext.unspent.value,
+          };
+
```
tylerlevine, line 1255: nit: maybe add a `debug('Signing p2tr input');` line here so we're consistent with the other input type cases

modules/core/src/v2/coins/utxo/recover.ts
```diff
@@ -241,6 +241,11 @@ export async function recover(coin: AbstractUtxoCoin, bitgo: BitGo, params: Reco
         return;
       }
 
+      if (addressType === Codes.UnspentTypeTcomb('p2tr') && !coin.supportsP2tr()) {
+        // P2tr is not supported. Skip.
```
tylerlevine, line 245: Could we add a `TODO: BG-<ticket>` reference here so we know to come back to this?

tylerlevine (changes requested):
some small changes needed, but all in all excellent work 👍

modules/core/src/v2/coins/abstractUtxoCoin.ts
sangaman: Thanks, this was an auto-import but I'll keep this distinction in mind going forward.

modules/core/src/v2/coins/abstractUtxoCoin.ts
sangaman: We're logging `debug('Signing input %d of %d', index + 1, transaction.ins.length);` for all input types currently at the start of this for loop, so that should cover p2tr at least as far as consistency goes. I could maybe add the input type to the log statement if you think it's worthwhile?

modules/core/src/v2/coins/utxo/recover.ts
sangaman: I think this is just to exclude the coins that don't support `p2tr` so I don't think there's a relevant TODO here. This comment is probably just not worded great, I'll make it a bit clearer and also for the other if blocks I modeled this one off of.

tylerlevine: Got it - so just to be clear we don't need to do anything else to support recovery (signing with backup key) for txs spending taproot inputs?

modules/core/src/v2/coins/abstractUtxoCoin.ts
tylerlevine: Let's leave it alone for now, we can figure out what the input type is in other ways if needed.

tylerlevine (approved):

brandonblack (approved):

modules/core/src/v2/coins/utxo/recover.ts
sangaman: Not that I'm aware of although it might be worth taking a closer look at. The test case in this PR signs the user + backup script which is what we'd do for recovery, so functionally speaking it is definitely supported. But this if block was necessary because otherwise we'd test it for coins that don't support p2tr.

PR #1643, opened 2021-11-04
feat(account-lib): implement keypair, transaction, builders and builder factory for solana

alebusse:
Work in Progress 

Implemented keypair, transaction, builder and builder factory for solana
implemented transfer and wallet init builders

[stlx-8108](https://bitgoinc.atlassian.net/browse/STLX-8108)


modules/account-lib/package.json
```diff
@@ -36,11 +36,12 @@
   "dependencies": {
     "@bitgo/blake2b": "^3.0.1",
     "@bitgo/bls": "^0.1.0",
-    "@bitgo/statics": "^6.14.0",
+    "@bitgo/statics": "^6.15.0-rc.4",
```
alebusse, line 39: This is needed since contains the new statics config for Sol (the one it was reverted)

modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -0,0 +1,122 @@
...
+   *
+   * @param {TransactionType} transactionType The transaction type to be set.
+   */
+  setTransactionType(transactionType: TransactionType): void {
+    this._type = transactionType;
+  }
+
+  /** @inheritdoc */
+  canSign(): boolean {
+    if (!this._solTransaction || (!this._solTransaction.nonceInfo && !this._solTransaction.recentBlockhash)) {
```
rliberal-simtlix, line 38: We need check here in detail the nonce and the recent blockhash on the tx

modules/account-lib/src/coin/baseCoin/baseKeyPair.ts
```diff
@@ -21,7 +21,7 @@ export interface BaseKeyPair {
   /**
    * Returns the keys in the protocol default key format
    */
-  getKeys(): any;
+  getKeys(): unknown;
```
rliberal-simtlix, line 24: Why this is unknown?

modules/account-lib/src/coin/sol/keyPair.ts
```diff
@@ -0,0 +1,83 @@
...
+  protected keyPair: DefaultKeys;
+  protected source?: KeyPairOptions;
+  /**
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   * @param { KeyPairOptions } source Either a master seed, a private key, or a public key
+   */
+  constructor(source?: KeyPairOptions) {
+    let kp: Keypair;
+    if (!source) {
```
rliberal-simtlix, line 18: May be here we need to think to apply a command pattern.
Not mandatory change.

modules/account-lib/src/coin/sol/keyPair.ts
```diff
@@ -0,0 +1,83 @@
...
+      kp = Keypair.generate();
+      this.setKP(kp);
+    } else if (isSeed(source)) {
+      kp = Keypair.fromSeed(source.seed.slice(0, 32));
```
rliberal-simtlix, line 22: Create a const for seedLength

modules/account-lib/src/coin/sol/iface.ts
```diff
@@ -0,0 +1,39 @@
...
+  type: InstructionBuilderTypes.CreateNonceAccount;
+  params: { fromAddress: string; nonceAddress: string; authAddress: string; amount: string };
+}
+
+export interface Transfer {
+  type: InstructionBuilderTypes.Transfer;
+  params: { fromAddress: string; toAddress: string; amount: string };
+}
+
+export type ValidInstructionTypes = SystemInstructionType | 'Memo';
```
rliberal-simtlix, line 39: According to the BitGo standard we should add the interfaces for validityWindow, feeOptions and SequenceId.

https://docs.google.com/document/d/1cMGyXa8GwgwDMcZ2Uv7dMDsqRWlIEp1MRdVgpPW8p1o/edit#

We can add a TODO and resolve this in a separate PR.

modules/account-lib/src/coin/sol/instructionBuilders.ts
```diff
@@ -0,0 +1,78 @@
...
+    case InstructionBuilderTypes.Transfer:
+      return transferBuilder(instructionToBuild);
+    case InstructionBuilderTypes.CreateNonceAccount:
+      return createNonceAccountBuilder(instructionToBuild);
+    default:
+      throw new Error(`Invalid instruction type or not supported`);
+  }
+}
+
+function advanceNonceBuilder(data: Nonce) {
```
rliberal-simtlix, line 22: I don't know if the proper name for this function is good to call it builder.
I think just change for advanceNonceInstruction, because is not a builder pattern implemented here.

rliberal-simtlix: Same for the other instructions.

modules/account-lib/src/coin/sol/instructionParser.ts
```diff
@@ -0,0 +1,79 @@
...
+import { NotSupported } from '../baseCoin/errors';
+import { InstructionBuilderTypes, ValidInstructionTypesEnum } from './constants';
+import { InstructionParams, WalletInit, Transfer, Nonce, Memo } from './iface';
+import { getInstructionType } from './utils';
+
+export function instructionParser(type: TransactionType, instructions: TransactionInstruction[]): InstructionParams[] {
+  const instructionData: InstructionParams[] = [];
+  if (type === TransactionType.WalletInitialization) {
+    const walletInit = parseWalletInitTransaction(instructions);
+    instructionData.push(walletInit);
```
rliberal-simtlix, line 13: Here we are have two responsabilities: one for parsing and other to pushing into instructionData. I think this should be on separate methods.

modules/account-lib/src/coin/sol/instructionParser.ts
```diff
@@ -0,0 +1,79 @@
...
+    instructionData.push(...send);
+  } else {
+    throw new NotSupported('Invalid transaction, transaction type not supported: ' + type);
+  }
+
+  return instructionData;
+}
+
+export function parseWalletInitTransaction(instructions: TransactionInstruction[]): WalletInit {
+  const createInstruction = SystemInstruction.decodeCreateAccount(instructions[0]);
```
rliberal-simtlix, line 25: Create const here for these indexes

modules/account-lib/src/coin/sol/instructionParser.ts
alebusse: Renamed

modules/account-lib/src/coin/sol/instructionBuilders.ts
alebusse: Renamed

joey-bai (commented):
General comment on architecture.

I feel like we have our TransactionBuilder pattern flipped.
With our current direction, we have a builder for a specific type of action:
- transfer (sending assets)
- creating accounts (wallet init, nonce, ATAs, etc.)
- staking (maybe in the future?)

Since Solana supports multiple instructions in a single Transaction, our SDK should follow suit. I'm imagine something like this:
```
// transaction to create a native asset and nonce account
const transaction = new SolanaTransactionBuilder()
    .createAccount(AccountType.Native)
    .createAccount(AccountType.Nonce)
    .feePayer(feePayer)
    .recentBlockHash(...)
    .build()
```

modules/account-lib/src/coin/sol/InstructionParamsFactory.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+/**
+ * Parses Solana instructions to Wallet initialization tx instructions params
+ *
+ * @param {TransactionInstruction[]} instructions - containing create and initialize nonce solana instructions
+ * @returns {InstructionParams[]} An array containing instruction params for Wallet initialization tx
+ */
+function parseWalletInitTransaction(instructions: TransactionInstruction[]): Array<WalletInit> {
```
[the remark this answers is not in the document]
alebusse: Renaming it to parseWalletInitInstructions

alebusse:
> General comment on architecture.
> 
> I feel like we have our TransactionBuilder pattern flipped. With our current direction, we'll need a builder for a specific type of action:
> 
> * transfer (sending assets)
> * creating accounts (wallet init, nonce, ATAs, etc.)
> * staking (maybe in the future?)
> 
> Users need to know which builder to use based on their intent. This also limits them to a specific instruction per transaction, when Solana supports multiple instructions in a single Transaction.
> 
> I'm imagining something like this:
> 
> ```
> // transaction to create a native asset and nonce account
> const transaction = new SolanaTransactionBuilder()
>     .createAccount(AccountType.Native)
>     .createAccount(AccountType.Nonce)
>     .feePayer(feePayer)
>     .recentBlockHash(...)
>     .build()
> ```
> 
> this way, users only need to interact with a single builder class to perform their desired operation.

Idk how to identify the type of the tx if we let any mix of instruction inside, i actually though about this approach before but after discussing with Rafa and Rodri we decided to keep the current architecture where you get a builder for each type of tx that bitgo supports (WalletInit, Send, StakingLock, etc), where WalletInit type will have only the required instructions for the nonce account creation and Send type will support multiple transfers, advanceNonce and memo instructions. 

The most common use of these builders is the wallet platform which usually have a switch or has to build based on a type of tx.
Maybe @CapnMigraine or @pancham88 can give us a review about whats the best path

modules/account-lib/src/coin/sol/InstructionParamsFactory.ts
```diff
@@ -0,0 +1,100 @@
...
+  type: TransactionType,
+  instructions: TransactionInstruction[],
+): InstructionParams[] {
+  switch (type) {
+    case TransactionType.WalletInitialization:
+      return parseWalletInitTransaction(instructions);
+    case TransactionType.Send:
+      return parseSendTransaction(instructions);
+    default:
+      throw new NotSupported('Invalid transaction, transaction type not supported: ' + type);
```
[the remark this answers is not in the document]
joey-bai: not required/

modules/account-lib/src/coin/sol/transferBuilder.ts
```diff
@@ -1,37 +1,67 @@
...
+   */
+  memo(memo: string): this {
+    this.validateMemo(memo);
+    const memoData: Memo = {
+      type: InstructionBuilderTypes.Memo,
+      params: { memo },
+    };
+    this._instructionsData.push(memoData);
+    return this;
   }
```
joey-bai, line 33: this should be moved to the base transaction builder - memo's are supported for all transactions (and not just transfers)

modules/account-lib/src/coin/sol/transferBuilder.ts
```diff
@@ -1,37 +1,67 @@
...
+    };
+    this._instructionsData.push(transferData);
+    return this;
   }
 
   /** @inheritdoc */
-  protected signImplementation(key: BaseKey): Transaction {
-    throw new NotImplementedError('signImplementation not implemented');
+  protected async buildImplementation(): Promise<Transaction> {
+    this.transaction.setTransactionType(TransactionType.Send);
```
joey-bai, line 64: this info is provided by overriding `transactionType()`. with this pattern, this should be set in parent implementation.

modules/account-lib/src/coin/sol/transferBuilder.ts
```diff
@@ -1,37 +1,67 @@
...
-  }
+  /**
+   *  Set a transfer
+   *
+   * @param {string} fromAddress - the sender address
+   * @param {string} toAddress - the receiver address
+   * @param {string} amount - the amount sent
+   * @returns {TransactionBuilder} This transaction builder
+   */
+  transfer(fromAddress: string, toAddress: string, amount: string): this {
```
joey-bai, line 43: traditional builder pattern is to split this up into multiple functions to set each param. each func validates its own inputs.

e.g.
```
const solTransfer = new TransferBuilder()
    .from('fromAddress')
    .to('destAddress')
    .amount('100 SOL')
    .memo('hi there')
    .build();
```

modules/account-lib/src/coin/sol/transferBuilder.ts
```diff
@@ -1,37 +1,67 @@
...
+    }
 
-  /** @inheritdoc */
-  protected fromImplementation(rawTransaction: any): Transaction {
-    throw new NotImplementedError('fromImplementation not implemented');
+    const transferData: Transfer = {
+      type: InstructionBuilderTypes.Transfer,
+      params: { fromAddress, toAddress, amount },
+    };
+    this._instructionsData.push(transferData);
```
joey-bai, line 58: feels like this should be done in a `build()` function.

also, if someone calls `.transfer()` multiple times, this is going to add multiple transfer instructions into this builder. is that intentional?

modules/account-lib/src/coin/sol/utils.ts
```diff
@@ -73,10 +92,163 @@ export function isValidAmount(amount: string): boolean {
...
+}
+
+/**
+ * Returns the transaction Type based on the  transaction instructions.
+ * Only Wallet initialization and Transfer transactions are supported.
+ *
+ * @param {SolTransaction} transaction - the solana transaction
+ * @returns {TransactionType} - the type of transaction
+ */
+export function getTransactionType(transaction: SolTransaction): TransactionType {
```
joey-bai, line 192: need a better understanding of how this works. maybe @pancham88 or @CapnMigraine can help here.

why do we limit ourselves to 1 type per transaction?

modules/account-lib/src/coin/sol/utils.ts
```diff
@@ -73,10 +92,163 @@ export function isValidAmount(amount: string): boolean {
...
+  const { instructions } = transaction;
+  validateIntructionTypes(instructions);
+  if (
+    instructions.length === 2 &&
```
joey-bai, line 196: if we use a durable nonce account, we'll have 2 instructions as well:
1. advance nonce
2. transfer

modules/account-lib/src/coin/sol/walletInitializationBuilder.ts
```diff
@@ -1,37 +1,65 @@
...
-  /** @inheritdoc */
-  protected buildSolTransaction(): SolTransaction {
-    throw new NotImplementedError('buildImplementation not implemented');
-  }
+  /**
+   *  Creates a nonce account for an address
+   *
+   * @param {string} fromAddress - the address whos gonna own the nonce
+   * @param {string} nonceAddress - the nonce address to be created
+   * @param {string} authAddress - the address whos gonna own the nonce
```
joey-bai, line 23: doc for `fromAddress` and `authAddress` are the same.

consider renaming `fromAddress` to `feePayer`

modules/account-lib/src/coin/sol/walletInitializationBuilder.ts
```diff
@@ -1,37 +1,65 @@
...
+   * @param {string} amount - the amount sent to cover the rent of the nonce address
+   * @returns {TransactionBuilder} This transaction builder
+   */
+  walletInit(fromAddress: string, nonceAddress: string, authAddress: string, amount: string): this {
```
joey-bai, line 27: traditional builder pattern is to split this up into multiple functions, 1 for each input.
`build()` then returns the resulting, immutable transaction

modules/account-lib/src/coin/sol/walletInitializationBuilder.ts
```diff
@@ -1,37 +1,65 @@
...
   /** @inheritdoc */
-  protected fromImplementation(rawTransaction: any): Transaction {
-    throw new NotImplementedError('fromImplementation not implemented');
-  }
-
-  /** @inheritdoc */
-  protected signImplementation(key: BaseKey): Transaction {
-    throw new NotImplementedError('signImplementation not implemented');
+  protected async buildImplementation(): Promise<Transaction> {
+    this.transaction.setTransactionType(TransactionType.WalletInitialization);
```
joey-bai, line 62: similar - setting type should be done in super class

modules/account-lib/src/coin/sol/walletInitializationBuilder.ts
```diff
@@ -1,37 +1,65 @@
...
+    const walletInitData: WalletInit = {
+      type: InstructionBuilderTypes.CreateNonceAccount,
+      params: {
+        fromAddress,
+        nonceAddress,
+        authAddress,
+        amount,
+      },
+    };
+    this._instructionsData.push(walletInitData);
```
joey-bai, line 56: similar points as transfer builder:
- maybe move this to build
- if this function is called multiple times, will the transaction create multiple nonces?

modules/account-lib/src/coin/sol/transferBuilder.ts
alebusse: `instructionsData` is just where i put the data used to build the instructions in the build step, and yes, multiple transfer are allowed

modules/account-lib/src/coin/sol/transferBuilder.ts
alebusse: The thing is, do we want to support memos everywhere or just in transfers?

modules/account-lib/src/coin/sol/transferBuilder.ts
alebusse: In order to be able to add multiple transfers i though it may be  better to set the all 3 transfer params (from, to, amount) in the same method.

modules/account-lib/src/coin/sol/utils.ts
alebusse: Wallet init should only have the create and initializeNonce, in the specific order mentioned.
Transfers can have advanceNonce, transfer and memo (for now)

Im limiting the intructions types to the type of tx what we want to support.

modules/account-lib/src/coin/sol/walletInitializationBuilder.ts
alebusse: The question here is, do we want to be able to set a fromAddress(the one who will send the lamports to cover the nonce account rent) that can be different than authAddress (the owner of the nonce account)?

modules/account-lib/src/coin/sol/walletInitializationBuilder.ts
alebusse: Adding a check to prevent this.

modules/account-lib/src/coin/sol/constants.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const VALID_SYSTEM_INSTRUCTION_TYPES: ValidInstructionTypes[] = [
+  ValidInstructionTypesEnum.AdvanceNonceAccount,
+  ValidInstructionTypesEnum.Create,
+  ValidInstructionTypesEnum.Transfer,
+  ValidInstructionTypesEnum.InitializeNonceAccount,
+  ValidInstructionTypesEnum.Memo,
+];
+
+export enum WALLET_INIT_INSTRUCTION_INDEXES {
```
rliberal-simtlix, line 32: nit: this enum should be in cammel case not snake case.

modules/account-lib/src/coin/sol/iface.ts
rliberal-simtlix: Please add the mentioned TODO.

modules/account-lib/src/coin/sol/solInstructionFactory.ts
```diff
@@ -0,0 +1,108 @@
...
+  const {
+    params: { fromAddress, toAddress, amount },
+  } = data;
+  assert(fromAddress, 'Missing fromAddress param');
+  assert(toAddress, 'Missing toAddress param');
+  assert(amount, 'Missing toAddress param');
+  const transferInstruction = SystemProgram.transfer({
+    fromPubkey: new PublicKey(fromAddress),
+    toPubkey: new PublicKey(toAddress),
+    lamports: new BigNumber(amount).toNumber(),
```
rliberal-simtlix, line 82: Seems to be ok:
https://github.com/solana-labs/solana-web3.js/blob/master/src/system-program.ts#L36

modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -41,21 +44,145 @@ export class Transaction extends BaseTransaction {
...
+    }
   }
 
   /** @inheritdoc */
   toBroadcastFormat(): string {
-    throw new NotImplementedError('toBroadcastFormat not implemented');
+    if (!this._solTransaction) {
+      throw new ParseTransactionError('Empty transaction');
+    }
+    // The signatures can have null signatures (which means they are required but yet unsigned)
```
rliberal-simtlix, line 84: nit: Add here extra line

modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -41,21 +44,145 @@ export class Transaction extends BaseTransaction {
...
+    // In order to be able to serializer the txs, we have to change the requireAllSignatures based
+    // on if the TX is fully signed or not
+    const requireAllSignatures = requiresAllSignatures(this._solTransaction.signatures);
```
rliberal-simtlix, line 87: nit: Please here add a comment in order to specify why when we got the broadcast format is on base64 and not base58 just only to clarify and inform

modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -41,21 +44,145 @@ export class Transaction extends BaseTransaction {
...
+        this._id = base58.encode(this._solTransaction.signature);
+      }
+      const transactionType = getTransactionType(this._solTransaction);
+      switch (transactionType) {
+        case TransactionType.WalletInitialization:
+          this.setTransactionType(TransactionType.WalletInitialization);
+          break;
+        case TransactionType.Send:
+          this.setTransactionType(TransactionType.Send);
+          break;
```
rliberal-simtlix, line 115: No default cases for error or notify ? Why if we have a new type of tx?

modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -41,21 +44,145 @@ export class Transaction extends BaseTransaction {
...
+            value: instruction.params.amount,
+            coin: this._coinConfig.name,
+          });
+          outputs.push({
+            address: instruction.params.toAddress,
+            value: instruction.params.amount,
+            coin: this._coinConfig.name,
+          });
+          break;
+        default:
```
rliberal-simtlix, line 181: Is default needed here?

modules/account-lib/src/coin/sol/transactionBuilder.ts
```diff
@@ -21,55 +38,210 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+   * Validates that the feePayer field is defined
+   */
+  private validateFeePayer(): void {
+    if (this._feePayer === undefined) {
+      throw new BuildTransactionError('Invalid transaction: missing feePayer');
+    }
+  }
+
+  /**
+   * Validates that the feePayer field is defined
```
rliberal-simtlix, line 239: change feePayer for nonce here

modules/account-lib/src/coin/sol/transactionBuilder.ts
```diff
@@ -21,55 +38,210 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+   * Set the transaction nonce
+   * Requires both optional params in order to use the durable nonce
+   *
+   * @param {Blockhash} blockHash The latest blockHash
+   * @param {boolean} useDurableNonce Set to true to use Durable Nonce, default is false
+   * @param {string} [walletNoncePubKey] the pub key of the nonce Wallet (required for durable nonce)
+   * @param {string} [authWalletPubKey] the pub key of the owner of the nonce Wallet (required for durable nonce)
+   * @returns {TransactionBuilder} This transaction builder
+   */
+  nonce(blockHash: Blockhash, useDurableNonce = false, walletNoncePubKey?: string, authWalletPubKey?: string): this {
```
joey-bai, line 145: looks like `walletNoncePubKey` and `authWalletPubKey` need to be provided together.
lets create a new interface for it, maybe something like:
```
interface DurableNonceParams {
  walletNonceAddress: string,
  authWalletAddress: string
}
```

then you can update function signature to something like:
```
nonce(blockHash: Blockhash, durableNonceParams?: DurableNonceParams)
```

note users should not need to set `useDurableNonce` explicitly. its derived from the existence of `durableNonceParams`

joey-bai: also, if we use the concept where `feePayer === sender`, we don't need a separate `authWalletAddress`.

for the initial impl, let's assume that we will only support use cases where there is 1 signer.

modules/account-lib/src/coin/sol/transactionBuilder.ts
```diff
@@ -21,55 +38,210 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+    }
+    this._recentBlockhash = blockHash;
+    return this;
+  }
+
   // endregion
 
   // region Validators
   /** @inheritdoc */
   validateAddress(address: BaseAddress, addressFormat?: string): void {
```
joey-bai, line 172: let's be consistent with type of `address` in builder funcs. the ones we added are all strings. we should probably update them to `BaseAddress`

modules/account-lib/src/coin/sol/keyPair.ts
```diff
@@ -1,34 +1,120 @@
...
+import nacl from 'tweetnacl';
+import { BaseKeyPair } from '../baseCoin';
+import { InvalidKey } from '../baseCoin/errors';
+import { DefaultKeys, isPrivateKey, isPublicKey, isSeed, KeyPairOptions } from '../baseCoin/iface';
+import { SEED_LENGTH } from './constants';
+import { SolanaKeys } from './iface';
+import { base58ToUint8Array, isValidPrivateKey, isValidPublicKey, Uint8ArrayTobase58 } from './utils';
 
-export class KeyPair extends Ed25519KeyPair {
+export class KeyPair implements BaseKeyPair {
```
joey-bai, line 10: why did we need to change this? Solana key pairs are ed25519.

modules/account-lib/src/coin/sol/iface.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface WalletInit {
+  type: InstructionBuilderTypes.CreateNonceAccount;
+  params: { fromAddress: string; nonceAddress: string; authAddress: string; amount: string };
+}
+
+export interface Transfer {
+  type: InstructionBuilderTypes.Transfer;
+  params: { fromAddress: string; toAddress: string; amount: string };
+}
```
joey-bai, line 38: for these interfaces, why do we need to include a type? can they be derived?

modules/account-lib/src/coin/sol/keyPair.ts
alebusse: Mostly b/c the Ed25519KeyPair implementation converts keys to hex instead of base58 and i liked to use the Solana lib instead of nacl for address generation.

modules/account-lib/src/coin/sol/transactionBuilder.ts
```diff
@@ -21,55 +38,210 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+    return this._transaction;
+  }
+
+  /**
+   * Set the transaction fee payer
+   *
+   * @param {string} feePayerAddress the fee payer account
+   * @returns {TransactionBuilder} This transaction builder
+   */
+  feePayer(feePayerAddress: string | undefined): this {
```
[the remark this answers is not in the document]
alebusse: During the nonce account creation , we actually need 2 signers, the auth account and the nonce account. The sign method can be used multiple times.

modules/account-lib/src/coin/sol/iface.ts
```diff
@@ -0,0 +1,40 @@
+import { Blockhash, SystemInstructionType, TransactionSignature } from '@solana/web3.js';
+import { InstructionBuilderTypes } from './constants';
+
+// TODO(STLX-9890): Add the interfaces for validityWindow, feeOptions and SequenceId
+export interface SolanaKeys {
```
joey-bai, line 5: should be singular: `SolanaKey` since this represents a single key.

alebusse: Is similar to DefaultKeys, it contains a key pair, priv and pub

modules/account-lib/src/coin/sol/iface.ts
```diff
@@ -0,0 +1,40 @@
...
+  prv?: Uint8Array | string;
+  pub: string;
+}
+
+export interface TxData {
+  id?: TransactionSignature;
+  feePayer?: string;
+  feeMultiplier: number;
```
joey-bai, line 13: `numSignatures`

modules/account-lib/src/coin/sol/keyPair.ts
joey-bai: add comment explaining this is still a ed25519 key pair for documentation.
also add a TODO to refactor `Ed25519KeyPair` to support encoding changes (e.g. hex or base58)

modules/account-lib/src/coin/sol/walletInitializationBuilder.ts
joey-bai: initial impl doesn't need to have a different `authAddress` from `fromAddress`. remove for now, we will add back if needed

modules/account-lib/src/coin/sol/walletInitializationBuilder.ts
alebusse: Removed fromAddress

joey-bai:
Updated with notable changes:
- refactor transfer and wallet init builder to become more conventional
- allow setting memo at any point (would previously fail if no instructions were added yet)

modules/account-lib/src/coin/sol/transaction.ts
joey-bai: capturing this comment in https://bitgoinc.atlassian.net/browse/STLX-10065
let's work on it as a follow up - there's other improvements we can make

joey-bai (approved):
@alebusse @rliberal-simtlix please review changes to the builders (since i made changes). approval is for everything else.

modules/account-lib/src/coin/sol/transferBuilder.ts
joey-bai: memos should be supported everywhere.

PR #1644, opened 2021-11-05
feat: add support for generating p2tr addresses

tylerlevine:
Ticket: BG-35567

### DIFF
--- a/modules/core/src/errors.ts
+++ b/modules/core/src/errors.ts
@@ -56,6 +56,13 @@ export class P2wshUnsupportedError extends BitGoJsError {
   }
 }
 
+export class P2trUnsupportedError extends BitGoJsError {
+  public constructor(message?: string) {
+    super(message || 'p2tr not supported by this coin');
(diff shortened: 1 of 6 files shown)


modules/core/src/errors.ts
```diff
@@ -56,6 +56,13 @@ export class P2wshUnsupportedError extends BitGoJsError {
   }
 }
 
+export class P2trUnsupportedError extends BitGoJsError {
+  public constructor(message?: string) {
+    super(message || 'p2tr not supported by this coin');
+    Object.setPrototypeOf(this, P2trUnsupportedError.prototype);
```
OttoAllmendinger, line 62: FYI we don't need these any longer

this was for `target: es5` times https://www.dannyguo.com/blog/how-to-fix-instanceof-not-working-for-custom-errors-in-typescript/#upgrade-the-compilation-target

OttoAllmendinger (approved):

PR #1645, opened 2021-11-05
feat: add unspents module from BitGo/unspents

OttoAllmendinger:
https://github.com/BitGo/unspents

Plain copy without history

Commit e1818b4638e5c76cde7625a19667e42e65fde487

Issue: BG-38767


OttoAllmendinger:
companion PR: https://github.com/BitGo/unspents/pull/53

modules/unspents/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "scripts": {
+    "build": "tsc --project tsconfig.build.json",
+    "test": "mocha --require ts-node/register \"test/**/*.ts\"",
+    "lint": "eslint --quiet .",
+    "prepare": "npm run build",
+    "unit-test": "npm run test"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/BitGo/BitGoJS.git"
```
sangaman, line 19: I think we should add `"directory": "modules/unspents"` here per the recommendation for monorepos here https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository

modules/unspents/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "types": "dist/index.d.ts",
+  "directories": {
+    "test": "test"
+  },
+  "scripts": {
+    "build": "tsc --project tsconfig.build.json",
+    "test": "mocha --require ts-node/register \"test/**/*.ts\"",
+    "lint": "eslint --quiet .",
+    "prepare": "npm run build",
+    "unit-test": "npm run test"
```
sangaman, line 15: Should these use `yarn` now?

modules/unspents/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "keywords": [
+    "bitcoin",
+    "utxo"
+  ],
+  "author": "BitGo Inc.",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/BitGo/BitGoJS/issues"
+  },
+  "homepage": "https://github.com/BitGo/BitGoJS#readme",
```
sangaman, line 30: Should this point to the `unspents` specific readme? That would make sense to me since it's a module-specific package.json.

sangaman (changes requested):
Concept ACK. One suggestion and a couple questions below.

I also get a diff on `yarn.lock` if I run `yarn install` on this branch from the root folder.

All the scripts/tests work for me. I haven't inspected most of the PR though since I'm assuming it's a straight copy from the latest in the unspents repo.

modules/unspents/package.json
OttoAllmendinger: thanks, fixed

modules/unspents/package.json
sangaman: OK if that's the case then that's fine. I confirmed they work for me.

modules/utxo-lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bitgo/utxo-lib",
-  "version": "2.1.0",
+  "version": "2.3.0",
```
sangaman, line 3: This version bump is just for the dependency change? I notice this skips over the most recently published 2.2.0x version.

sangaman (approved):
Still getting a diff from `yarn install` but maybe that's due to some subtle difference in my environment, anyway all looks good now.

modules/utxo-lib/package.json
OttoAllmendinger: I was getting strange stale dependency errors prior to this change

PR #1648, opened 2021-11-05
feat(utxo-lib): add support for `prevOutputs: TxOutput[]` in TransactionBuilder

OttoAllmendinger:
Add PrevOutput[] argument to createTransactionBuilderFromTransaction that
sets input data which allows `signInput2Of3()` without hacks.

Issue: BG-38773


modules/utxo-lib/src/bitgo/UtxoTransactionBuilder.ts
```diff
@@ -25,14 +26,29 @@ export class UtxoTransactionBuilder<T extends UtxoTransaction = UtxoTransaction>
...
   }
 
-  static fromTransaction(tx: UtxoTransaction): UtxoTransactionBuilder {
-    return new UtxoTransactionBuilder(tx.network, TransactionBuilder.fromTransaction(tx));
+  static fromTransaction(
+    tx: UtxoTransaction,
+    network?: bitcoinjs.Network,
+    prevOutputs?: PrevOutput[]
+  ): UtxoTransactionBuilder {
+    return new UtxoTransactionBuilder(tx.network, TransactionBuilder.fromTransaction(tx), prevOutputs);
```
OttoAllmendinger, line 51: I needed to add `network?: bitcoinjs.Network` here otherwise TS yells at me for not extending `bitcoinjs.TransactionBuilder` correctly

modules/utxo-lib/test/integration_local_rpc/generate/main.ts
```diff
@@ -69,15 +69,17 @@ async function createTransactionsForScriptType(
...
+  await writeTransactionFixtureWithInputs(rpc, network, `deposit_${scriptType}.json`, deposit1Txid);
   if (!isScriptType2Of3(scriptType) || !isSupportedSpendType(network, scriptType)) {
     console.log(logTag + ': spend not supported, skipping spend');
     return;
   }
 
-  const spendTx = createSpendTransaction(fixtureKeys, scriptType, depositTxid, depositTx, script, network);
+  const deposit2Txid = await sendFromFaucet(rpc, address, 1);
+  const deposit2Tx = await rpc.getRawTransaction(deposit2Txid);
+  const spendTx = createSpendTransaction(fixtureKeys, scriptType, [deposit1Tx, deposit2Tx], script, network);
```
OttoAllmendinger, line 82: here we now use two inputs (with the same scriptPubKey) from two different transactions

modules/utxo-lib/test/integration_local_rpc/generate/outputScripts.util.ts
```diff
@@ -93,10 +97,15 @@ export function createScriptPubKey(keys: KeyTriple, scriptType: ScriptType, netw
   }
 }
 
+export type TxbInput = PrevOutput & {
```
brandonblack, line 100: Kinda confusing naming here when `TxbInput` has a different type inside `TransactionBuilder`. Maybe `TxInput` would be more appropriate actually?

modules/utxo-lib/test/integration_local_rpc/parse.ts
```diff
@@ -81,14 +89,52 @@ function runTestParse(network: Network, txType: FixtureTxType, scriptType: Scrip
       return Buffer.from(getPrevOutput(input).scriptPubKey.hex, 'hex');
     }
 
-    function getPrevOutputs(): [txid: string, index: number, value: number, script: Buffer][] {
-      return parsedTx.ins.map((i) => [getTxidFromHash(i.hash), i.index, getPrevOutputValue(i), getPrevOutputScript(i)]);
+    function getPrevOutputs(): TxbInput[] {
```
brandonblack, line 92: Maybe rename to `getInputs`?

brandonblack (changes requested):

modules/utxo-lib/test/integration_local_rpc/generate/outputScripts.util.ts
sangaman: I didn't realize these were different but I agree we should name them differently if they're not referring to the same type.

modules/utxo-lib/src/bitgo/UtxoTransactionBuilder.ts
```diff
@@ -25,14 +26,29 @@ export class UtxoTransactionBuilder<T extends UtxoTransaction = UtxoTransaction>
     if (txb) {
       (this as any).__INPUTS = (txb as any).__INPUTS;
     }
+
+    if (prevOutputs) {
+      const txbInputs = (this as any).__INPUTS;
```
sangaman, line 31: This a bit hacky too but I will go away when we implement psbt. I guess this is the least hacky workaround for having the prevout scripts available for tx builder.

sangaman (dismissed):
Changes look good but I'm a bit confused as to why there was so much change to the test fixtures, particularly for non-p2tr spends?

I'm also curious if you've tried this for signing within `core` without the current workaround of attempting to sign twice? If not I might give that a shot to make sure this works before we merge.

modules/utxo-lib/test/integration_local_rpc/generate/outputScripts.util.ts
OttoAllmendinger: well `TxInput` also exists and is even exported

modules/utxo-lib/test/integration_local_rpc/parse.ts
OttoAllmendinger: that would suggest `TxInput` which requires some props that we don't have

OttoAllmendinger:
> Changes look good but I'm a bit confused as to why there was so much change to the test fixtures, particularly for non-p2tr spends?
> 
> I'm also curious if you've tried this for signing within `core` without the current workaround of attempting to sign twice? If not I might give that a shot to make sure this works before we merge.

The fixtures now spend from two inputs instead of just one

sangaman (approved):

sangaman (dismissed):

brandonblack (dismissed):

OttoAllmendinger:
sigh unsigned commits

brandonblack (approved):

PR #1659, opened 2021-11-08
test(core): add transaction test suite for all utxo coins

OttoAllmendinger:
Create prebuild / half-signed / full-signed transactions and compare them
with reference vectors.

Issue: BG-38773


modules/core/test/v2/unit/coins/utxo/fixtures/btc/transactions-p2tr.json
```diff
@@ -0,0 +1,162 @@
...
+    "ins": [
+      {
+        "txid": "9cf295ddfa67cfcc33ab4321f1c4f0c31a343069c841e41d352c7053536049ea",
+        "index": 1,
+        "script": "",
+        "witness": [
+          "",
+          "a489417f71cae151810d2d55b1d97d6949ad0f3a57f87420983a3ee7c69f35f57ad103cb7c6a0b1fad8abd86e84cf619270ff998c499d2d881982137c3d3041c01",
```
brandonblack, line 36: I see we're using SIGHASH_ALL instead of SIGHASH_DEFAULT here. Might want to test both, and prefer SIGHASH_DEFAULT to save the bytes.

brandonblack: Actually, looking at the code, I don't see where the SIGHASH_ALL is coming from. Looks like we're passing `undefined` for `hashType` which should give `SIGHASH_DEFAULT`. Hmm...

brandonblack: Found it -- not a critical bug, but it's coming from https://github.com/BitGo/BitGoJS/blob/master/modules/core/src/v2/coins/abstractUtxoCoin.ts#L1234

OttoAllmendinger: it's not a parameter in `wallet.signTransaction` or `coin.signTransaction` though, and I don't think it is worthwhile adding it

BitGoJS should pick the appropriate default

brandonblack: Looks like the default is only per network not per input type. We can definitely launch as is, but maybe make a ticket to fix the default for p2tr inputs.

brandonblack: (or we could change the default for bitcoin to `undefined` and see what that does, since bitcoinjs-lib and HSM will both handle that fine afaik)

OttoAllmendinger: https://github.com/BitGo/BitGoJS/commit/d2a8c7b1aadd77e3da08981fa462d8cf403c661b

brandonblack: Nice!

brandonblack (approved):

PR #1661, opened 2021-11-08
fix(core): fix memoid check for eos txn

lcovar:
We need to validate the the memoId in the buildParams and address are equal ONLY when memoId is present in the address as a query param. If the memoId is not present in the address then do not check for equality.

TICKET: BG-38801

### DIFF
--- a/modules/core/src/v2/coins/eos.ts
+++ b/modules/core/src/v2/coins/eos.ts
@@ -336,7 +336,7 @@ export class Eos extends BaseCoin {
     if (destinationDetails.pathname === address) {
       return {
         address: address,
-        memoId: '',
+        memoId: undefined,
       };
     }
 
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/eos.ts
```diff
@@ -336,7 +336,7 @@ export class Eos extends BaseCoin {
     if (destinationDetails.pathname === address) {
       return {
         address: address,
-        memoId: '',
+        memoId: undefined,
```
shihao-guo, line 339: i saw this with the BitGoJS fix as well where the OVC broken because of something like this. @joey-bai is this still going to be an issue for OVC?

shihao-guo (approved):

modules/core/src/v2/coins/eos.ts
lcovar: OVC actually does not use verifyTransaction, that was a wrong assumption on my end. OVC was broken due to TextDecoder like which @joey-bai fixed.

shihao-guo: There's a memoID treatment change in BitGoJS https://github.com/BitGo/BitGoJS/pull/1615/files#diff-6efbc825de1f8d6f3fae2bfff41cf42cc40beae8d0af3309e52e11582ab83d9fR324

joey-bai: @lcovar could we triple check this? i remember this led to problems when we used `bitgo.preBuildAndSign` without a memo id.

we ran into an issue with comparing empty string to `undefined.

lcovar: That's right, before we were always checking that memoId in the address was = to memoId in the buildParams. Now I have updated the check to be more similar to the check we have in bitgo-microservices where we only check for equality if it's present in the as a query param. https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/v2/coins/eos.js#L736-L745

`        if (!_.isUndefined(expectedOutputAddressAndMemoId.memoId)) {
          if (txHexTransferAction.memo !== expectedOutputAddressAndMemoId.memoId) {
            throw new Error('txHex receive memoId does not match expected recipient memoId');
          }
        }`

joey-bai (approved):

PR #1663, opened 2021-11-09
fix(core): repair replay protection input signing

OttoAllmendinger:
It broke with the double-signing workaround for p2tr

Issue: BG-38773


sangaman (approved):
Thanks for catching this.

PR #1664, opened 2021-11-09
test(core): add tests for signing with replay protection inputs

OttoAllmendinger:
* add reusable `getReplayProtectionAddresses()`

Issue: BG-38773


brandonblack (approved):
Looks good to me, although I hate the naming on the fixtures now. could it be `normal-*` vs. `replayProtection-*`

OttoAllmendinger:
maybe it makes more sense if you read it like this:

* `p2sh-p2sh` means that there are two p2sh inputs

* `replayProtection-p2sh` means that there is a `replayProtection` input and a subsequent `p2sh` input

PR #1666, opened 2021-11-09
test(core): add tests for verifySignature/explainTransaction

OttoAllmendinger:
Issue: BG-38773


modules/core/test/v2/unit/coins/utxo/transaction.ts
```diff
@@ -43,14 +48,18 @@ function run(coin: AbstractUtxoCoin, inputScripts: InputScriptType[]) {
       }
     }
 
-    const unspents = inputScripts.map((type, i) => getUnspent(type, i));
+    function getUnspents(): Unspent[] {
+      return inputScripts.map((type, i) => getUnspent(type, i));
+    }
```
OttoAllmendinger, line 53: this way the error is more readable if it fails

(errors during `before()` show up differently)

brandonblack (approved):

PR #1670, opened 2021-11-09
feat(statics): define coin sol in statics

joey-bai:
- comment out adding sol/tsol until it is initializable.
- added a test to ensure all coins that get added to `statics` is initializable.

This partially reverts commit 62df94d898621c71a77d54856ac9c45b4d0746ed.

Ticket: STLX-8229

### DIFF
--- a/modules/statics/test/unit/coins.ts
+++ b/modules/statics/test/unit/coins.ts
@@ -1,10 +1,13 @@
 import 'should';
 import { CoinFamily, coins, Erc20Coin, EthereumNetwork, BaseNetwork } from '../../src';
+import { BitGo } from '../../../core';
 
 interface DuplicateCoinObject {
   name: string;
   network: BaseNetwork;
 }
+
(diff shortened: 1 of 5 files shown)


modules/statics/test/unit/coins.ts
```diff
@@ -35,3 +38,10 @@ describe('ERC20 Coins', () => {
       }, {});
   });
 });
+
+describe('All Coins', () => {
```
shihao-guo, line 42: nice!

shihao-guo (approved):

CapnMigraine (approved):
lgtm

modules/statics/test/unit/coins.ts
OttoAllmendinger: `statics` should not have any outside dependencies

when we add a coin, we should add it in `statics` first and later use these definitions in downstream projects

this test creates a circular dependency

PR #1674, opened 2021-11-10
feat(unspents): classify p2tr script path sigs

sangaman:
This adds support for classifying a partially signed p2tr input after parsing the signature by checking for an `inputClassification` of `taproot` and checking the size of the control block to distinguish between a level 1 and level 2 depth script path spend.

Ticket: BG-37329

Note: This fixes the bug mentioned here: https://github.com/BitGo/bitgo-microservices/pull/13243/files#diff-2cc59901c8cb5f7eacecfab40a529dc3c0ac12387846622d9a6ab7423a9a8f46R1129

### DIFF
--- a/modules/unspents/src/dimensions.ts
+++ b/modules/unspents/src/dimensions.ts
@@ -399,6 +399,17 @@ Dimensions.fromInput = function (input: utxolib.TxInput, params = {}) {
       case 'witnessscripthash':
         return p2wshInput;
     }
+    if (utxolib.bitgo.isParsedSignatureScriptTaproot(parsed)) {
+      if (parsed.controlBlock.length === 65) {
+        // 33 bytes + 32 bytes for depth 1
+        return p2trScriptPathLevel1Input;
+      } else if (parsed.controlBlock.length === 97) {
+        // 33 bytes + 64 bytes for depth 2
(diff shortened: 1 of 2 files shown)


modules/unspents/src/dimensions.ts
```diff
@@ -399,6 +399,17 @@ Dimensions.fromInput = function (input: utxolib.TxInput, params = {}) {
       case 'witnessscripthash':
         return p2wshInput;
     }
+    if (utxolib.bitgo.isParsedSignatureScriptTaproot(parsed)) {
```
OttoAllmendinger, line 402: why not `case 'taproot' in L41?

OttoAllmendinger (changes requested):

modules/unspents/src/dimensions.ts
sangaman: It came down to playing nice with the types. If I just add another `case` then I can't access `parsed.controlBlock` without manually casting `parsed` to `ParsedSIgnatureScriptTaproot`.

OttoAllmendinger: I see

OttoAllmendinger (approved):

tylerlevine (approved):

PR #1675, opened 2021-11-10
chore(utxo-lib): update to `@bitgo/bitcoinjs-lib@6.1.0-rc.2`

tylerlevine:
Ticket: BG-39039

### DIFF
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -50,7 +50,7 @@
     "algosdk": "^1.12.0",
     "bignumber.js": "^9.0.0",
     "bip32": "^2.0.6",
-    "bitcoinjs-lib": "^5.2.0",
+    "bitcoinjs-lib": "npm:@bitgo/bitcoinjs-lib@6.1.0-rc.2",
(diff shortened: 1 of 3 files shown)


brandonblack (changes requested):
I _think_ we should update this in `account-lib` too to prevent issues in WP later.

tylerlevine:
> I _think_ we should update this in `account-lib` too to prevent issues in WP later.

We only use `ECPair` and `networks` in account-lib, I don't think these have really changed.

brandonblack:
> > I _think_ we should update this in `account-lib` too to prevent issues in WP later.
> 
> We only use `ECPair` and `networks` in account-lib, I don't think these have really changed.

The problem is that account-lib's version clobbers the one pulled by utxo-lib when WP pulls both :(. That's what Luis' was working around by putting bitcoinjs-lib directly in WP's package.json

sangaman (approved):
Changes look good, possible account-lib concern notwithstanding

brandonblack (approved):

PR #1681, opened 2021-11-12
fix(statics): ensure UnderlyingAssets values are unique

joey-bai:
each value in UnderlyingAsset represents an asset and clients use the value to determine which icon to show. additionally, the same underlying asset may be used by multiple chains (e.g. USDT exists as an erc20 and eos token).

previously, we had two enum entries with the same value 'box'. one was meant for the erc20 token and the other for eos. this led to us showing the wrong icon for the eos box token. fix is to make sure BOX and EOS_BOX have different values.

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -627,6 +627,7 @@ export enum UnderlyingAsset {
   UQC = 'uqc',
   URHD = 'urhd',
   USDC = 'usdc',
+  // Also available on EOS
   USDT = 'usdt',
   USDX = 'usdx',
   USG = 'usg',
@@ -711,10 +712,9 @@ export enum UnderlyingAsset {
   'talgo:180447' = 'talgo:180447',
 
   // EOS tokens
-  CHEX = 'CHEX',
-  IQ = 'IQ',
-  EOS_BOX = 'BOX',
-  EOS_USDT = 'USDT',
+  CHEX = 'chex',
+  IQ = 'iq',
+  EOS_BOX = 'eos:box',
 }
 
 /**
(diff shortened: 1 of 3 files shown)


modules/statics/src/base.ts
```diff
@@ -711,10 +712,9 @@ export enum UnderlyingAsset {
   'talgo:180447' = 'talgo:180447',
 
   // EOS tokens
-  CHEX = 'CHEX',
-  IQ = 'IQ',
-  EOS_BOX = 'BOX',
-  EOS_USDT = 'USDT',
+  CHEX = 'chex',
```
shihao-guo, line 715: seems like a non-trivial change, are we safe in downstream?

modules/statics/src/base.ts
```diff
@@ -711,10 +712,9 @@ export enum UnderlyingAsset {
...
+  IQ = 'iq',
+  EOS_BOX = 'eos:box',
```
shihao-guo, line 717: seems like a non-trivial change, are we safe in downstream? Also this seems to introduce a new pattern of asset naming. How do we know this is safe and is generic enough for other assets to use as a pattern?

shihao-guo: i see we do it for algo, so should be ok. Just need to make sure we are setting this as a pattern.

lcovar (approved):

modules/statics/src/base.ts
joey-bai: great call out - i am considering to use a chain-agnostic name.
i.e. change `EOS_BOX` to `TOKEN_DEFI_BOX` (https://bloks.io/tokens/BOX-eos-token.defi).

Another example is USDT (Tether), which exists on a bunch of chains. Luckily, they seem to have done well with branding/integrating so that USDT is synonymous with Tether.

shihao-guo: i think chain specific is fine ... but do we then have teos:box and eos:box?

joey-bai: we do not, we use the same underlying asset for eos and teos. we use contract name to differentiate.
Using CHEX as an example:
prod contract name = `chexchexchex`
test contract name = `testtoken111`

```
eosToken('eos:CHEX', 'Chintai', 8, 'chexchexchex', UnderlyingAsset.CHEX, AccountCoin.DEFAULT_FEATURES, '', 'CHEX'),
teosToken('teos:CHEX', 'Chintai', 8, 'testtoken111', UnderlyingAsset.CHEX, AccountCoin.DEFAULT_FEATURES, '', 'CHEX'),
```

PR #1683, opened 2021-11-12
feat(core): add amount to refund eos txn

lcovar:
OVC requires an amount for any transaction therefore
defaulting the amount to 0 will work for this scenario
since refund txns dont have an amount to be refunded as
part of the transaction

TICKET: STLX-9486

Successful signing

Screen Shot 2021-11-12 at 4 02 06 PM


modules/core/src/v2/coins/eos.ts
```diff
@@ -715,6 +715,7 @@ export class Eos extends BaseCoin {
 
         const refundActionData = txAction.data;
         tx.address = refundActionData.owner;
+        tx.amount = '0';
```
joey-bai, line 718: pre-existing issue, but if you look above, the error messages are for unstake.
line 709: `new Error(`unstake should only have 1 action: ${tx.actions.length} given`);`
line 713: `throw new Error('Invalid or incomplete unstake action data');`

since we're making a change here, could you update them to `refund`?

lcovar: Good catch, updated the error messages.

joey-bai (approved):
lgtm

PR #1685, opened 2021-11-16
chore(core/statics): change eth2 urls

micaela-ledesma:
Change eth2 tx, address and explorer urls

bg-27116


modules/statics/src/networks.ts
```diff
@@ -288,8 +288,8 @@ class Ethereum extends Mainnet implements EthereumNetwork {
 class Ethereum2 extends Mainnet implements AccountNetwork {
   name = 'Ethereum2';
   family = CoinFamily.ETH2;
-  explorerUrl = 'https://beaconscan.com/tx';
-  accountExplorerUrl = 'https://beaconscan.com/address';
+  explorerUrl = 'https://beaconcha.in/tx';
```
wesgraham, line 291: these url's give 404's

micaela-ledesma: The kovan ones take you to an error screen so I figured it was okay that without extra data it returned 404. Shall we leave them as they are for the time being then?

PR #1687, opened 2021-11-16
feat(statics): update Token Gog

Joelsx:
Updating GOG contract. We confirmed that there are no transfers existing for the old GOG and updating the contract accordingly. Ticket: STLX-9469
Jira ticket: https://bitgoinc.atlassian.net/browse/STLX-9469
STLX-9469


argjv (approved):
Seems to match: https://etherscan.io/address/0x9AB7bb7FdC60f4357ECFef43986818A2A3569c62

PR #1689, opened 2021-11-17
docs: capitalize JavaScript and TypeScript properly

ericcrosson-bitgo:
As seen here[^1] and here[^2].

[^1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript
[^2]: https://www.typescriptlang.org/

### DIFF
--- a/modules/statics/README.md
+++ b/modules/statics/README.md
@@ -15,7 +15,7 @@
 
 ### Get the number of decimal places in a Bitcoin
 
-#### Javascript
+#### JavaScript
 ```js
 const { coins } = require('@bitgo/statics');
 
@@ -23,7 +23,7 @@ const btc = coins.get('btc');
 console.log(btc.decimalPlaces);
 ```
 
-#### Typescript
+####typescript
 ```typescript
 import { coins } from '@bitgo/statics';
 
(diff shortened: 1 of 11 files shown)


modules/statics/README.md
```diff
@@ -15,15 +15,15 @@
 
 ### Get the number of decimal places in a Bitcoin
 
-#### Javascript
+####javascript
```
tylerlevine, line 18: These look like broken section headers to me now when I preview this file.

tylerlevine: image

tylerlevine (changes requested):
looks good, just need to fix the section headers. Thanks!

modules/statics/README.md
ericcrosson-bitgo: Whoopsie

My brain's markdown parser malfunctioned, I thought that was a code block. Un momento

ericcrosson-bitgo: Fixed 3 occurrences

ericcrosson-bitgo:
image

0xabrarhussain (approved):

modules/statics/README.md
```diff
@@ -33,15 +33,15 @@ console.log(btc.decimalPlaces);
...
-#### Javascript
+#### JavaScript
 ```js
 const { coins } = require('@bitgo/statics');
 
 const omg = coins.get('omg');
 console.log(omg.contractAddress);
 ```
 
-#### Typescript
```
0xabrarhussain, line 44: actually missed this. was merged but think this is a typo. same with other places it got changed back to `typescript`

PR #1693, opened 2021-11-18
fix(express): enable tezos consolidations route in express

shiksb:
Ticket: BG-378
Screen Shot 2021-11-22 at 10 07 21 AM
30


shihao-guo:
were you able to test if consolidation is working through express SDK given this fix?

argjv (approved):

PR #1705, opened 2021-11-22
feat(core): add send support for XLM muxed addresses

CapnMigraine:
Add validation for xlm muxed addresses

TICKET: BG-39117

### DIFF
--- a/modules/core/src/v2/coins/xlm.ts
+++ b/modules/core/src/v2/coins/xlm.ts
@@ -41,9 +41,17 @@ import { checkKrsProvider, getStellarKeys } from '../recovery/initiate';
 
 const co = Bluebird.coroutine;
 
+/**
+ * XLM accounts support virtual (muxed) addresses
+ * A base address starts with "G" and is tied to the underlying "real" account
+ * A muxed address starts with "M" and combines the base address with a 64-bit integer ID in order to provide
+ * an alternative to memo ids.
+ */
 interface AddressDetails {
+  baseAddress: string;
   address: string;
-  memoId?: string;
+  id?: string;
+  memoId?: string | undefined;
 }
 
 interface Memo {
@@ -296,6 +304,38 @@ export class Xlm extends BaseCoin {
   }
 
   /**
+   * Create instance of stellar.MuxedAccount from M address
+   * See: https://developers.stellar.org/docs/glossary/muxed-accounts
+   */
+  getMuxedAccount(address: string): stellar.MuxedAccount {
+    try {
+      return stellar.MuxedAccount.fromAddress(address, '0');
+    } catch (e) {
+      throw new Error(`invalid muxed address: ${address}`);
+    }
+  }
+
+  /**
+   * Return boolean indicating whether a muxed address is valid
+   * See: https://developers.stellar.org/docs/glossary/muxed-accounts
+   *
+   * @param address
+   * @returns {boolean}
+   */
+  isValidMuxedAddress(address: string): boolean {
+    if (!_.isString(address) || !address.startsWith('M')) {
+      return false;
+    }
+
+    try {
+      // return true if muxed account is valid or throw
+      return !!stellar.MuxedAccount.fromAddress(address, '0');
+    } catch (e) {
+      return false;
+    }
+  }
+
+  /**
    * Minimum balance of a 2-of-3 multisig wallet
    * @returns minimum balance in stroops
    */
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/xlm.ts
```diff
@@ -42,8 +42,10 @@ import { checkKrsProvider, getStellarKeys } from '../recovery/initiate';
 const co = Bluebird.coroutine;
 
 interface AddressDetails {
+  baseAddress?: string;
   address: string;
```
joey-bai, line 46: could you document the difference between `address` and `baseAddress`?

modules/core/src/v2/coins/xlm.ts
```diff
@@ -295,6 +297,40 @@ export class Xlm extends BaseCoin {
...
 
+  /**
+   * Create instance of stellar.MuxedAccount from M address
+   * See: https://developers.stellar.org/docs/glossary/muxed-accounts
+   *
+   * @param address
+   */
+  getMuxedAccount(address): stellar.MuxedAccount {
+    try {
+      return stellar.MuxedAccount.fromAddress(address, '0');
```
joey-bai, line 308: do we need to check that `address` starts with "M"?

joey-bai (dismissed):

modules/core/src/v2/coins/xlm.ts
```diff
@@ -295,6 +297,40 @@ export class Xlm extends BaseCoin {
     return true;
   }
 
+  /**
+   * Create instance of stellar.MuxedAccount from M address
+   * See: https://developers.stellar.org/docs/glossary/muxed-accounts
+   *
+   * @param address
+   */
+  getMuxedAccount(address): stellar.MuxedAccount {
```
tylerlevine, line 306: type on `address` param please.

modules/core/src/v2/coins/xlm.ts
```diff
@@ -295,6 +297,40 @@ export class Xlm extends BaseCoin {
...
+  }
+
+  /**
+   * Return boolean indicating whether a muxed address is valid
+   * See: https://developers.stellar.org/docs/glossary/muxed-accounts
+   *
+   * @param address
+   * @returns {boolean}
+   */
+  isValidMuxedAddress(address): boolean {
```
tylerlevine, line 321: type on `address` param please.

modules/core/src/v2/coins/xlm.ts
```diff
@@ -541,7 +594,9 @@ export class Xlm extends BaseCoin {
     const addressDetails = this.getAddressDetails(address);
     const rootAddressDetails = this.getAddressDetails(rootAddress);
 
-    if (addressDetails.address !== rootAddressDetails.address) {
+    const baseAddress = this.isValidMuxedAddress(address) ? addressDetails.baseAddress : addressDetails.address;
```
joey-bai, line 597: mentioned this in other PR, but why do we need to keep track of both `address` and `baseAddress`?
it seems like they should be mutually exclusive where only 1 address field will be used at any given time.

tylerlevine (changes requested):
looks good, just a couple small nits on missing types.

modules/core/src/v2/coins/xlm.ts

modules/core/src/v2/coins/xlm.ts

modules/core/src/v2/coins/xlm.ts
CapnMigraine: the call will throw if the address is for any reason the input is not a valid muxed address

modules/core/src/v2/coins/xlm.ts

modules/core/src/v2/coins/xlm.ts
CapnMigraine: it helps to know the base address when we're sending to a muxed one, because we need to be able to check the existence of the underlying account in order to decide whether to add a `createAccount` operation or a `payment` operation

CapnMigraine: I simplified the checks by setting always `baseAddress` and `address` to AddressDetails. Now we can just check for `baseAddress` whenever we need it.

tylerlevine (approved):

modules/core/src/v2/coins/xlm.ts
```diff
@@ -540,10 +602,9 @@ export class Xlm extends BaseCoin {
 
     const addressDetails = this.getAddressDetails(address);
     const rootAddressDetails = this.getAddressDetails(rootAddress);
-
-    if (addressDetails.address !== rootAddressDetails.address) {
+    if (addressDetails.baseAddress !== rootAddressDetails.address) {
```
joey-bai, line 605: should this be `addressDetails.baseAddress !== rootAddressDetails.baseAddress` ?

joey-bai (dismissed):

modules/core/src/v2/coins/xlm.ts
CapnMigraine: we don't need to, since `rootAddress` is passed as param and it's always the root (or base) address of the wallet

joey-bai (approved):

PR #1706, opened 2021-11-22
test(core): add parametric crossChainRecovery tests

OttoAllmendinger:
Issue: BG-38773

----

Please see individual commits


modules/core/test/v2/unit/coins/utxo/crossChainRecovery.ts
```diff
@@ -0,0 +1,212 @@
...
+    before('setup nocks', function () {
+      nocks.push(nockWallet(recoveryCoin, recoveryWalletId, keychainsBase58));
+      nocks.push(nockWalletAddress(recoveryCoin, recoveryWalletId, depositAddressRecoveryCoin));
+      nocks.push(
+        ...nockThirdpartyExplorerTxInfo(sourceCoin, depositTx, getDepositUnspents(), depositAddressSourceCoin.address)
+      );
+    });
+
+    after(function () {
+      nocks.forEach((n) => n.isDone());
```
tylerlevine, line 156: `isDone()` won't throw and will instead return false if the nock isn't consumed, which isn't the behavior you want here I think.

```suggestion
      nocks.forEach((n) => n.done());
```

tylerlevine (approved):
excellent improvement, just a minor issue nock consumption test. Otherwise lgtm

modules/core/test/v2/unit/coins/utxo/crossChainRecovery.ts
OttoAllmendinger: Ah yes of course

Sadly the nocks are persistent and not "done", so the this would throw. I can't find `n.used()` or equivalent.

I'll fix it in a follow-up.

PR #1707, opened 2021-11-23
test(unspents): validate taproot level-2 script path spend weights

OttoAllmendinger:
Issue: BG-38773

----

See commits for details


modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -80,21 +79,25 @@ export interface ParsedSignatureScriptP2PKH extends ParsedSignatureScript {
...
-  publicKeys: [Buffer, Buffer];
+  // P2TR tapscript spends are for keypath spends or 2-of-2 multisig scripts
+  // A single signature indicates a keypath spend.
+  // Two signatures indicate a scriptPath spend.
+  publicKeys: [Buffer] | [Buffer, Buffer];
+  signatures: [Buffer] | [Buffer, Buffer];
+  // For scriptpath signatures, this indicates level. For keypath signatures this is undefined.
+  scriptPathLevel: number | undefined;
   pubScript: Buffer;
   controlBlock: Buffer;
```
sangaman, line 102: Thinking about this again, `controlBlock` ought to be as optional as it will be undefined for keypath spends.

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -191,20 +194,25 @@ export function parseSignatureScript(
...
     }
 
+    const signatures = [sig1, sig2].map((b) => {
+      if (Buffer.isBuffer(b)) {
+        return b;
+      }
+      throw new Error(`unexpected signature element ${b}`);
+    }) as [Buffer, Buffer];
+
+    const scriptPathLevel = controlBlock.length === 65 ? 1 : controlBlock.length === 97 ? 2 : undefined;
```
sangaman, line 204: Should we throw in the `undefined` case here? If we have a control block, it ought to be either 65 or 97 bytes or else something is very wrong.

sangaman (dismissed):
Changes look good and added test coverage is great - how exactly did you generate the new test fixtures?

OttoAllmendinger:
> Changes look good and added test coverage is great - how exactly did you generate the new test fixtures?

I am using the following pattern everywhere I have test fixtures:

```
async function getFixture<T>(path: string, defaultValue: T): Promise<T> {
  try {
    return JSON.parse(await fs.readFile(path));
  } catch (e) {
    if (isFileNotFound(e)) {
      await fs.writeFile(path, JSON.stringify(defaultValue));
      throw new Error(`please check ${path} and run test again`);
    }
    throw e;
  }
}

/// test
it('matches fixture', async function () {
  foo.should.eql(await getFixture('foo.json', foo));
});
```
So fixtures are written once if they don't exist. If you want to update a fixture because it has changed, just delete the old one.

modules/utxo-lib/src/bitgo/signature.ts
OttoAllmendinger: let's do that

modules/utxo-lib/src/bitgo/signature.ts
OttoAllmendinger: will change

sangaman (approved):
Changes to my feedback look good, and thanks for the explanation about the test fixture generation that's great and I will keep it in mind.

PR #1708, opened 2021-11-23
refactor(core): add, use signAndVerifyWalletTransaction for utxo

OttoAllmendinger:
* Change AbstractUtxoCoin SignTransactionOptions to accept a `pubs`
string triple and cosignerPub string.
This makes it more consistent with the `prv` parameter which is also
a string.
* Add, use `signAndVerifyWalletTransaction()` method which wraps
`signWalletTransactionWithUnspent`

Issue: BG-38773

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -1021,6 +1021,10 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     return this.supportsAddressType(utxolib.bitgo.outputScripts.scriptTypeForChain(chain));
   }
 
+  keyIdsForSigning(): number[] {
+    return [KeyIndices.USER, KeyIndices.BACKUP, KeyIndices.BITGO];
+  }
+
   /**
    * TODO(BG-11487): Remove addressType, segwit, and bech32 params in SDKv6
    * Generate an address for a wallet based on a set of configurations
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -44,7 +44,7 @@ import {
 } from '../baseCoin';
 import { CustomChangeOptions, parseOutput } from '../internal/parseOutput';
 import { RequestTracer } from '../internal/util';
-import { Keychain, KeyIndices } from '../keychains';
+import { Keychain, KeyIndices, Triple } from '../keychains';
 import { promiseProps } from '../promise-utils';
 import { NodeCallback } from '../types';
 import { Wallet } from '../wallet';
@@ -54,8 +54,9 @@ const debug = debugLib('bitgo:v2:utxo');
 const co = Bluebird.coroutine;
 
 import ScriptType2Of3 = utxolib.bitgo.outputScripts.ScriptType2Of3;
+import { ReplayProtectionUnspent, Unspent } from './utxo/unspent';
 import { getReplayProtectionAddresses } from './utxo/replayProtection';
-import { Unspent } from './utxo/unspent';
+import { signAndVerifyWalletTransaction, WalletUnspentSigner } from './utxo/sign';
 
 export interface VerifyAddressOptions extends BaseVerifyAddressOptions {
   chain: number;
@@ -168,6 +169,7 @@ export interface AddressDetails {
 }
 
 export interface SignTransactionOptions extends BaseSignTransactionOptions {
+  /** Transaction prebuild from bitgo server */
   txPrebuild: {
     txHex: string;
     txInfo: {
@@ -182,11 +184,16 @@ export interface SignTransactionOptions extends BaseSignTransactionOptions {
       }[];
     };
   };
+  /** xprv of user key or backup key */
   prv: string;
-  userKeychain?: Keychain;
-  backupKeychain?: Keychain;
-  bitgoKeychain?: Keychain;
-  taprootRedeemIndex?: number;
+  /** xpubs triple for wallet (user, backup, bitgo) */
+  pubs: Triple<string>;
+  /** xpub for cosigner (defaults to bitgo) */
+  cosignerPub?: string;
+  /**
+   * When true, creates full-signed transaction without placeholder signatures.
+   * When false, creates half-signed transaction with placeholder signatures.
+   */
   isLastSignature?: boolean;
 }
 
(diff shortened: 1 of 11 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -182,11 +183,12 @@ export interface SignTransactionOptions extends BaseSignTransactionOptions {
...
     };
   };
+  // xprv
   prv: string;
-  userKeychain?: Keychain;
-  backupKeychain?: Keychain;
-  bitgoKeychain?: Keychain;
-  taprootRedeemIndex?: number;
+  // xpubs (user, backup, bitgo)
+  pubs: Triple<string>;
```
sangaman, line 189: Should this be called `xpubs` to clearly identify it for what it is? Without context I'd expect these to be regular pubkeys. I guess on the other hand we have `prv` which refers to an xprv, but I wonder if we should just rename that to `xprv` as well and eliminate any possibility for confusion.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -182,11 +183,12 @@ export interface SignTransactionOptions extends BaseSignTransactionOptions {
...
+  // cosigner xpub
+  cosignerPub?: string;
```
sangaman, line 191: Same question here for xpub naming.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1045,6 +1047,10 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     return this.supportsAddressType(utxolib.bitgo.outputScripts.scriptTypeForChain(chain));
   }
 
+  keyIdsForSigning(): number[] {
```
sangaman, line 1050: Is this moved here for forward compatibility purposes when non-btc coins might use taproot?

modules/core/src/v2/coins/utxo/sign.ts
```diff
@@ -78,3 +90,87 @@ export function signWalletTransactionWithUnspent(
...
+}
+
+/**
+ * Sign all inputs of a wallet transaction and verify signatures after signing.
+ * Collects and logs signing errors and verification errors, throws error in the end if any of them
+ * failed.
+ *
+ * @param transaction - wallet transactions to be signed
+ * @param unspents - transaction unspents
+ * @param walletSigner - singing parameters
```
sangaman, line 116: "signing" although "singing parameters" has a nice ring to it.

modules/core/test/v2/unit/wallet.ts
```diff
@@ -410,75 +251,9 @@ describe('V2 Wallet:', function () {
         address: TestBitGo.V2.TEST_WALLET1_ADDRESS,
         walletPassphrase: TestBitGo.V2.TEST_WALLET1_PASSCODE,
       };
-      const error = await bitgo.getAsyncError(wallet.send(params));
-      should.exist(error);
-    });
-
-
-    it('should error when amount is zero', async function () {
```
sangaman, line 418: Are all these deleted test cases being covered elsewhere?

sangaman (commented):
Very nice refactoring, removes the no longer needed double-sign workaround. Just a few comments/questions.

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: The problem is that `pub` is part of the generic interface, so I can't just rename it.

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: No, it is just easier if we always get all three pubkeys in general case. We should not make exceptions for coins just because they don't support taproot ;)

One side benefit is that having the pubkeys at hand allows us to derive all the required scripts before signing, so we do not have to use `redeemScript`/`witnessScript` on the unspent anymore. Plus we can sanity check the pubScript.

modules/core/src/v2/coins/utxo/sign.ts
OttoAllmendinger: :musical_note:

modules/core/test/v2/unit/wallet.ts
OttoAllmendinger: see commit message https://github.com/BitGo/BitGoJS/pull/1708/commits/067873364bfb89850730442dea91fb4198a092fc

modules/core/src/v2/coins/utxo/sign.ts
OttoAllmendinger: fixed

modules/core/src/v2/wallet.ts
```diff
@@ -122,9 +122,8 @@ export interface PrebuildTransactionResult extends TransactionPrebuild {
 export interface WalletSignTransactionOptions {
     txPrebuild?: TransactionPrebuild;
     prv?: string;
-    userKeychain?: Keychain,
-    backupKeychain?: Keychain,
-    bitgoKeychain?: Keychain,
+    pubs?: string[];
```
tylerlevine, line 125: This is a non-backwards change to the API. Could you please properly indicate that in the commit message according to the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer) rules? This is important when I go to publish since we need to make this a major version bump (which we're planning on doing shortly anyway, since we need to drop node 12 support soon).

Even better would be to avoid a breaking change here, but I think it's a good change and the timing is not bad either.

tylerlevine (changes requested):
great clean up, minor nit and a question about the breaking change.

modules/core/src/v2/wallet.ts
OttoAllmendinger: I actually don't think this is a breaking change here: if `pubs` is empty, the keys are fetched in `signTransaction`

OttoAllmendinger: However `AbstractUtxoCoin.prototype.signTransaction` now requires `pubs` so I guess that is a breaking change. I will explain this in the commit message.

OttoAllmendinger: the removal of `userKeychain`/`backupKeychain`/`bitgoKeychain` from the interface is a breaking change I suppose

modules/core/src/v2/coins/abstractUtxoCoin.ts
sangaman: Hm OK. In that case it might be nice to make these comments use the `/** */` style, that way many IDEs will associate that comment with the property so if anyone ever mouses over it they'll get a tooltip with this comment. I know that kind of thing is helpful for me.

modules/core/src/v2/wallet.ts
sangaman: Timing is good because these are relatively new - I believe I added these within the past few weeks for taproot purposes.

sangaman (approved):
I added a comment above about possibly using `/** */` comments to better indicate that `prv`/`pubs` are extended keys but that's a very minor consideration so proceed however you'd like.

sangaman (approved):
:+1:  to new comments, thank you.

sangaman (approved):

tylerlevine (approved):

PR #1716, opened 2021-11-24
feat(core): fix version of core dependecies

Joelsx:
fix version of core dependecies for continue using actual version of npm

STLX-9878


modules/core/package.json
```diff
@@ -72,9 +72,9 @@
     "lodash": "^4.17.14",
     "moment": "^2.20.1",
     "ripple-address-codec": "~4.1.3",
-    "ripple-binary-codec": "^0.2.4",
+    "ripple-binary-codec": "~1.1.3",
```
argjv, line 75: Can you try `"ripple-binary-codec": "~0.2.4",` ?

Joelsx: okay let me try

modules/core/package.json
```diff
@@ -72,9 +72,9 @@
     "lodash": "^4.17.14",
     "moment": "^2.20.1",
     "ripple-address-codec": "~4.1.3",
-    "ripple-binary-codec": "^0.2.4",
+    "ripple-binary-codec": "~0.2.4",
```
FabianSimtlix, line 75: should be ~1.1.3

modules/core/package.json
FabianSimtlix: Why `0.2.4`? On MS the version currently installed is 1.1.3

Joelsx: @argjv Why you want to use the version 0.2.4 since its updating at the moment to the 1.1.3 wich its the stable one?

tylerlevine (approved):

PR #1717, opened 2021-11-24
feat: add TSS key generation and signing functions

0xabrarhussain:
adds key generation/combine along with message signing/combine/verify

Closes: BG-39242, BG-38662, BG-39243, BG-39712

### DIFF
--- a/modules/account-lib/src/mpc/shamir.ts
+++ b/modules/account-lib/src/mpc/shamir.ts
@@ -1,0 +1,99 @@
+const assert = require('assert');
+
+const Shamir = (curve: any) => {
+  /**
+   * Perform Shamir sharing on the secret `secret` to the degree `threshold - 1` split `numShares`
+   * ways. The split secret requires `threshold` shares to be reconstructed.
+   *
+   * @param secret secret to split
+   * @param threshold share threshold required to reconstruct secret
+   * @param numShares total number of shares to split to split secret into
+   * @param indices
+   * @returns Dictionary of shares. Each key is an int in the range 1<=x<=numShares
+   * representing that share's free term.
+   */
+  const split = (
+    secret: Buffer,
+    threshold: number,
+    numShares: number,
+    indices?: Array<number>,
+  ): Record<number, Buffer> => {
+    if (indices === undefined) {
+      // make range(1, n + 1)
+      indices = Array.from({ length: numShares }, (_, i) => i + 1);
+    }
+    assert(threshold > 1);
+    assert(threshold <= numShares);
+    const coefs: Buffer[] = [];
+    for (let ind = 0; ind < threshold - 1; ind++) {
+      const random_value = curve.scalarRandom();
+      coefs.push(random_value);
+    }
+    coefs.push(secret);
+
+    const shares: Record<number, Buffer> = {};
+    for (let ind = 0; ind < indices.length; ind++) {
+      const x = indices[ind];
+      const x_buffer = Buffer.alloc(32);
+      // TODO BG-40908 : converting internal representation to buffers
+      x_buffer.writeUInt32LE(x, 0);
+      let partial = coefs[0];
+      for (let other = 1; other < coefs.length; other++) {
+        const scalarMult = curve.scalarMult(partial, x_buffer);
+        const newAdd = curve.scalarAdd(coefs[other], scalarMult);
+        partial = newAdd;
+      }
+      shares[x] = Buffer.from(partial);
+    }
+    return shares;
+  };
+
+  /**
+   * Reconstitute a secret from a dictionary of shares. The number of shares must
+   * be equal to `t` to reconstitute the original secret.
+   *
+   * @param shares dictionary of shares. each key is the free term of the share
+   * @returns secret
+   */
+  const combine = (shares: Record<number, Buffer>): Buffer => {
+    let s = Buffer.alloc(32);
+    for (const xi in shares) {
+      const yi = shares[xi];
+      const xi_buffer = Buffer.alloc(32);
+      let num_buffer = Buffer.alloc(32);
+      let denum_buffer = Buffer.alloc(32);
+      xi_buffer.writeUInt32LE(parseInt(xi, 10), 0);
+      num_buffer.writeUInt32LE(1, 0);
+      denum_buffer.writeUInt32LE(1, 0);
+
+      for (const xj in shares) {
+        const xj_buffer = Buffer.alloc(32);
+        xj_buffer.writeUInt32LE(parseInt(xj, 10), 0);
+        if (xi !== xj) {
+          num_buffer = curve.scalarMult(num_buffer, xj_buffer);
+        }
+      }
+      num_buffer = Buffer.from(num_buffer);
+      for (const xj in shares) {
+        const xj_buffer = Buffer.alloc(32);
+        xj_buffer.writeUInt32LE(parseInt(xj, 10), 0);
+        if (xi !== xj) {
+          denum_buffer = curve.scalarMult(denum_buffer, curve.scalarSub(xj_buffer, xi_buffer));
+        }
+      }
+      denum_buffer = Buffer.from(denum_buffer);
+      const inverted = curve.scalarInvert(denum_buffer);
+      const innerMultiplied = curve.scalarMult(num_buffer, inverted);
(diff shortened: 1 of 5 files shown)


modules/account-lib/src/mpc/tss.ts
```diff
@@ -0,0 +1,304 @@
...
+ * functions for key generation and message signing with EdDSA.
+ *
+ *
+ * ======================
+ * EdDSA Key Generation
+ * ======================
+ * 1. Each signer generates their own key share, which involves a private u-share and a public y-share.
+ * 2. Signers distribute their y-share to other signers.
+ * 3. After exhanging y-shares the next phase is to combine key shares. Each signer combines their u-share
+ *    with the y-shares received from other signers in order to generate a p-share for themselves. We
```
0xabrarhussain, line 12: @johnoliverdriscoll Q for here, is there a specific format that's expected when we save data into the hsm? even though here I'm saying that I'm generating a u-share, I'm not really generating that as a single string value. instead, it's the whole json from

```
interface PlayerKeyShare {
  i: string;
  y: string;
  u: string;
  prefix: string;
}
```

while an x-share is the whole collection from: 

```
interface PlayerSignShare {
  i: string;
  y: string;
  x: string;
  r: string;
  R: string;
}
```

modules/account-lib/test/unit/mpc/tss.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * @prettier
+ */
+import 'should';
+
+import Eddsa from '../../../src/mpc/tss';
+
+describe('TSS EDDSA key generation and signing', () => {
```
shihao-guo, line 8: any unhappy path test you can add here?

modules/account-lib/src/mpc/tss.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    let uBuffer = combinedBuffer.slice(0, 32);
+    uBuffer[0] &= 248;
+    uBuffer[31] &= 63;
+    uBuffer[31] |= 64;
+
+    // using big endian
+    const zeroBuffer32 = Buffer.alloc(32);
+    uBuffer = Buffer.concat([zeroBuffer32, uBuffer]);
+    uBuffer.reverse();
```
johnoliverdriscoll, line 134: I believe this based on an incorrect implementation of private key generation from ggmpc. It should be:

```
uBuffer = Buffer.concat([uBuffer, zeroBuffer32])
```

Note that we don't reverse the u buffer after concating the zeros *after* the u value.

You can test this against ggmpc 0.0.7 and it should work

johnoliverdriscoll (changes requested):

modules/account-lib/src/mpc/tss.ts
0xabrarhussain: good catch, it passes tests with either version. updated so that uBuffer and prefix aren't using big endian

modules/account-lib/src/mpc/shamir.ts
```diff
@@ -0,0 +1,98 @@
...
+      xi_buffer.writeUInt32LE(parseInt(xi, 10), 0);
+      num_buffer.writeUInt32LE(1, 0);
+      denum_buffer.writeUInt32LE(1, 0);
+
+      for (const xj in shares) {
+        const xj_buffer = Buffer.alloc(32);
+        xj_buffer.writeUInt32LE(parseInt(xj, 10), 0);
+        if (xi !== xj) {
+          num_buffer = curve.scalarMult(num_buffer, xj_buffer);
+        }
```
johnoliverdriscoll, line 73: This seems a little precarious. Since you're using buffers as your internal value type, you need to assume endianess, and here you're assuming little endian by writing `xj` to the zeroth index. This will work for Ed25519 since you'll be using sodium for the math, and sodium expects little endian buffers. However, this presents an opportunity to shoot yourself in the foot when dealing with secp256k1 which expects buffers to be big endian. You could solve for this by converting buffers from little to big endian in the secp256k1 curve implementation, but that seems a little hacky. The better solution would be to use native BigInt as the internal type and have the curve implementation convert to buffers themselves.

modules/account-lib/test/unit/mpc/tss.ts
```diff
@@ -0,0 +1,31 @@
...
+    const A_sign_share = MPC.signShare(message_buffer, A_combine.pShare, [A_combine.jShares[2]]);
+    const B_sign_share = MPC.signShare(message_buffer, B_combine.pShare, [B_combine.jShares[1]]);
+
+    const A_sign = MPC.sign(message_buffer, A_sign_share.xShare, [B_sign_share.rShares[1]]);
+    const B_sign = MPC.sign(message_buffer, B_sign_share.xShare, [A_sign_share.rShares[2]]);
+
+    const signature = MPC.signCombine([A_sign, B_sign]);
+    const result = MPC.verify(message_buffer, signature).toString();
+    result.should.equal(message);
+  });
```
johnoliverdriscoll, line 30: Please add negative test. Test that number of signers less than threshold produces an invalid signature at the very least.

johnoliverdriscoll (changes requested):

modules/account-lib/src/mpc/shamir.ts
0xabrarhussain: converted to numbers as internal buffers and ran into some errors. added a task and todo to shamir for conversion to internal number type representation for when we add `secp256k1`. for now, figured to unblock other people to use this thought should be OK to still have the buffer representation

modules/account-lib/test/unit/mpc/tss.ts
```diff
@@ -0,0 +1,64 @@
...
+    const A_sign_share = MPC.signShare(message_buffer, A_combine.pShare, [A_combine.jShares[2]]);
+    const B_sign_share = MPC.signShare(message_buffer, B_combine.pShare, [B_combine.jShares[1]]);
+
+    const A_sign = MPC.sign(message_buffer, A_sign_share.xShare, [B_sign_share.rShares[1]]);
+    const signature = MPC.signCombine([A_sign]);
+
+    try {
+      const result = MPC.verify(message_buffer, signature).toString();
+      result.should.throw();
+    } catch (e) {}
```
johnoliverdriscoll, line 62: This test doesn't look like it's doing what it's supposed to be doing. First, you're catching any thrown errors, so even if `result.should.throw()` results in an error, the test will still pass successfully. Second, you should assign a function to result in order to use `should.throw`:
```
const result = () => MPC.verify(...);
result.should.throw();
```

johnoliverdriscoll (changes requested):

modules/account-lib/test/unit/mpc/tss.ts
```diff
@@ -0,0 +1,65 @@
...
+    B_sign_share = MPC.signShare(message_buffer, B_combine.pShare, [B_combine.jShares[1], B_combine.jShares[3]]);
+    const C_sign_share = MPC.signShare(message_buffer, C_combine.pShare, [C_combine.jShares[1], C_combine.jShares[2]]);
+
+    A_sign = MPC.sign(message_buffer, A_sign_share.xShare, [B_sign_share.rShares[1], C_sign_share.rShares[1]]);
+    B_sign = MPC.sign(message_buffer, B_sign_share.xShare, [A_sign_share.rShares[2], C_sign_share.rShares[2]]);
+    const C_sign = MPC.sign(message_buffer, C_sign_share.xShare, [A_sign_share.rShares[3], B_sign_share.rShares[3]]);
+    signature = MPC.signCombine([A_sign, B_sign, C_sign]);
+    result = MPC.verify(message_buffer, signature).toString();
+    result.should.equal(message);
+  });
```
johnoliverdriscoll, line 44: Add one more test where you sign with signers at indices 1 and 3 (signers A and C). This is something we need to be careful with as it will be the most commonly used signing pair, and I even made a mistake in the HSM implementation that threw an error when signing with those indices.

johnoliverdriscoll (dismissed):
One more change, sorry

johnoliverdriscoll (dismissed):

0xabrarhussain:
@johnoliverdriscoll sorry re-format test and pushed to branch again, could you re-approve this?

johnoliverdriscoll (approved):

PR #1718, opened 2021-11-25
fix(core): break cyclical dependency

OttoAllmendinger:
We had a cyclical dependency:

`v2/baseCoin -> v2/wallet -> v2/abstractUtxoCoin -> v2/baseCoin`

Remove dependency from v2/wallet to v2/abstractUtxoCoin.

Issue: BG-38773

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -12,7 +12,6 @@ import {
   SignedTransaction, TransactionPrebuild,
   VerificationOptions, VerifyAddressOptions,
 } from './baseCoin';
-import { AbstractUtxoCoin } from './coins/abstractUtxoCoin';
 import { Eth } from './coins';
 import * as internal from './internal/internal';
 import { drawKeycard } from './internal/keycard';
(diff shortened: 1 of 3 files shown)


modules/core/src/v2/wallet.ts
```diff
@@ -1740,9 +1739,8 @@ export class Wallet {
         .send(whitelistedParams)
         .result();
 
-      const utxoCoin = self.baseCoin as AbstractUtxoCoin;
-      const blockHeightQuery = _.isFunction(utxoCoin.getLatestBlockHeight) ?
-        utxoCoin.getLatestBlockHeight(params.reqId) :
+      const blockHeightQuery = _.isFunction((self.baseCoin as any).getLatestBlockHeight) ?
+        (self.baseCoin as any).getLatestBlockHeight(params.reqId) :
```
sangaman, line 1743: Refactoring for another day I guess?

sangaman (approved):

modules/core/src/v2/wallet.ts
OttoAllmendinger: indeed

tylerlevine (approved):

PR #1722, opened 2021-11-26
fix(statics): use `utxolibName` instead of redefining constants

OttoAllmendinger:
The network constants are already defined in utxolib, there is no need
to repeat them in `statics`.

Also add `BitcoinGoldTestnet`.

Sanity check:

```
» yarn ts-node -e '
import * as utxolib from "@bitgo/utxo-lib";
import { coins, UtxoCoin } from "./src";

for (const name in utxolib.networks) {
  console.log(
    name,
    coins
      .filter((coin => (coin as UtxoCoin).network.utxolibName === name))
      .map(coin => coin.name)
  );
}
'

bitcoin [ 'btc' ]
testnet [ 'tbtc' ]
bitcoincash [ 'bch', 'bcha' ]
bitcoincashTestnet [ 'tbch', 'tbcha' ]
bitcoingold [ 'btg' ]
bitcoingoldTestnet [ 'tbtg' ]
bitcoinsv [ 'bsv' ]
bitcoinsvTestnet [ 'tbsv' ]
dash [ 'dash' ]
dashTest [ 'tdash' ]
litecoin [ 'ltc' ]
litecoinTest [ 'tltc' ]
zcash [ 'zec' ]
zcashTest [ 'tzec' ]
```

Issue: BG-39753

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -157,11 +157,11 @@ class BitcoinGold extends Mainnet implements UtxoNetwork {
   explorerUrl = 'https://btgexplorer.com/tx/';
 }
 
-class BitcoinGoldTestnet extends Mainnet implements UtxoNetwork {
+class BitcoinGoldTestnet extends Testnet implements UtxoNetwork {
   name = 'BitcoinGoldTestnet';
   family = CoinFamily.BTG;
   utxolibName = 'bitcoingoldTestnet';
-  explorerUrl = 'https://btgexplorer.com/tx/';
+  explorerUrl = undefined;
 }
 
 class Dash extends Mainnet implements UtxoNetwork {
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -13,29 +13,9 @@ export abstract class BaseNetwork {
   public abstract readonly explorerUrl: string | undefined;
 }
 
-/*
-The values for the various fork coins can be found in these files:
-
-property       filename             varname                           notes
-------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


modules/statics/src/networks.ts
```diff
@@ -158,90 +94,87 @@ class AvalancheCTestnet extends Testnet implements AccountNetwork {
...
-  messagePrefix = '\x18Bitcoin Gold Signed Message:\n';
-  bech32 = 'btg';
-  pubKeyHash = 0x26;
-  scriptHash = 0x17;
   family = CoinFamily.BTG;
+  utxolibName = 'bitcoingold';
+  explorerUrl = 'https://btgexplorer.com/tx/';
+}
+
+class BitcoinGoldTestnet extends Mainnet implements UtxoNetwork {
```
sangaman, line 160: Should this extend `Testnet`?

sangaman (changes requested):
Is this a `fix` or a `refactor`? Curious if this is causing any problems currently. Either way it's good to remove redundant code. I just have one note below.

modules/statics/src/networks.ts
OttoAllmendinger: yes!

OttoAllmendinger:
> Is this a `fix` or a `refactor`? Curious if this is causing any problems currently. Either way it's good to remove redundant code. I just have one note below.

Kind of: I can't really use `@bitgo/statics` in BitGoJS at the moment because we need the objects defined by `utxolib.networks` there (not structural equality but referential equality)

sangaman (approved):

PR #1723, opened 2021-11-29
feat(account-lib): dot explain transaction

gibrangul95:
@argjv @jrafael-simtlix @rliberal-simtlix @nimraarif21 @fdmirza

### DIFF
--- a/modules/account-lib/src/coin/dot/iface.ts
+++ b/modules/account-lib/src/coin/dot/iface.ts
@@ -1,4 +1,6 @@
 import { BaseTxInfo, TypeRegistry, DecodedUnsignedTx } from '@substrate/txwrapper-core/lib/types';
+import { TransactionType } from '../baseCoin';
+import { TransactionExplanation as BaseTransactionExplanation } from '../baseCoin/iface';
 
 /**
  * Method names for the transaction method. Names change based on the type of transaction e.g 'bond' for the staking transaction
@@ -15,6 +17,7 @@ export enum MethodNames {
  * The transaction data returned from the toJson() function of a transaction
  */
 export interface TxData {
+  id: string;
   sender: string;
   referenceBlock: string;
   blockNumber: number;
@@ -33,9 +36,7 @@ export interface TxData {
   owner?: string;
   proxyType?: string;
   delay?: string;
-  real?: string;
   forceProxyType?: proxyType;
-  call?: string;
 }
 
 /**
@@ -169,3 +170,13 @@ export type FeeOptions = {
   formula?: 'fixed' | 'feeRate' | 'perKB' | 'custom';
   type?: 'base' | 'max' | 'tip';
 };
+
+export interface TransactionExplanation extends BaseTransactionExplanation {
+  type: TransactionType;
+  forceProxyType?: proxyType;
+  controller?: string;
+  payee?: string;
+  owner?: string;
+  proxyType?: string;
+  delay?: string;
+}
--- a/modules/account-lib/src/coin/dot/transaction.ts
+++ b/modules/account-lib/src/coin/dot/transaction.ts
@@ -1,13 +1,21 @@
 import { BaseCoin as CoinConfig } from '@bitgo/statics';
 import { BaseTransaction, TransactionType } from '../baseCoin';
-import { BaseKey } from '../baseCoin/iface';
+import { BaseKey, TransactionRecipient } from '../baseCoin/iface';
 import { InvalidTransactionError, SigningError } from '../baseCoin/errors';
 import { construct, decode } from '@substrate/txwrapper-polkadot';
 import { UnsignedTransaction } from '@substrate/txwrapper-core';
 import { TypeRegistry } from '@substrate/txwrapper-core/lib/types';
 import Keyring, { decodeAddress } from '@polkadot/keyring';
 import { KeyPair } from './keyPair';
-import { TxData, DecodedTx, StakeArgs, StakeArgsPayeeRaw, AddProxyArgs, UnstakeArgs } from './iface';
+import {
+  TxData,
+  DecodedTx,
+  StakeArgs,
+  StakeArgsPayeeRaw,
+  AddProxyArgs,
+  UnstakeArgs,
+  TransactionExplanation,
+} from './iface';
 import utils from './utils';
 
 export class Transaction extends BaseTransaction {
@@ -97,6 +105,7 @@ export class Transaction extends BaseTransaction {
     }) as unknown as DecodedTx;
 
     const result: TxData = {
+      id: construct.txHash(this.toBroadcastFormat()),
       sender: decodedTx.address,
       referenceBlock: decodedTx.blockHash,
       blockNumber: decodedTx.blockNumber,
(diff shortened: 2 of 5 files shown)


modules/account-lib/src/coin/dot/transaction.ts
```diff
@@ -175,6 +182,61 @@ export class Transaction extends BaseTransaction {
     return result;
   }
 
+  explainTransaction(): TransactionExplanation {
```
argjv, line 185: Add this method to the `BaseTransaction`. Also, add JS docs

modules/account-lib/src/coin/dot/transaction.ts
```diff
@@ -175,6 +182,61 @@ export class Transaction extends BaseTransaction {
...
+      explanationResult.displayOrder.push('owner', 'proxyType', 'delay');
+    }
+    if (this.type === TransactionType.StakingUnlock) {
+      explanationResult.outputs = [
+        {
+          address: result.sender.toString(),
+          amount: result.amount || '',
+        },
+      ];
+    }
```
argjv, line 236: This should be a `switch/case`. Also, each `case` body should be its own method

rliberal-simtlix: Also on the switch the default value should have an error message, mentioning that we not support this type of tx. Because we have many type of tx types in polka with all the existent extrinsic.

modules/account-lib/test/unit/coin/dot/transaction.ts
```diff
@@ -51,4 +60,110 @@ describe('Dot Transaction', () => {
       should.deepEqual(tx.transactionSize(), DotResources.rawTx.transfer.unsigned.length / 2);
     });
   });
+
+  describe('Dot Explain Transaction', () => {
```
rliberal-simtlix, line 64: Add failure cases.

gibrangul95: We can never reach to a failure with this as the transaction building would fail for any sort of invalid transaction because of the other internal checks.

rliberal-simtlix (approved):

modules/account-lib/src/coin/baseCoin/iface.ts
```diff
@@ -87,3 +87,28 @@ export interface Entry extends BaseAddress {
...
+
+export interface TransactionExplanation {
+  displayOrder: string[];
+  id: string;
+  outputs: TransactionRecipient[];
+  outputAmount: string;
+  changeOutputs: TransactionRecipient[];
+  changeAmount: string;
+  fee: TransactionFee;
+  proxy?: string;
```
argjv, line 112: This is probably coin-specific

modules/account-lib/src/coin/baseCoin/iface.ts
```diff
@@ -87,3 +87,28 @@ export interface Entry extends BaseAddress {
...
+  producers?: string[];
```
argjv, line 113: This is probably coin (dot) specific

modules/account-lib/src/coin/dot/iface.ts
```diff
@@ -170,3 +169,14 @@ export type FeeOptions = {
   formula?: 'fixed' | 'feeRate' | 'perKB' | 'custom';
   type?: 'base' | 'max' | 'tip';
 };
+
+export interface TransactionExplanation extends Omit<BaseTransactionExplanation, 'id'> {
```
argjv, line 173: Does `Omit` mean we exclude the `id` field? That is the tx id, right?

modules/account-lib/src/coin/dot/iface.ts
```diff
@@ -170,3 +169,14 @@ export type FeeOptions = {
...
+  type: TransactionType;
+  forceProxyType?: proxyType;
+  to?: string;
```
argjv, line 176: This is actually a list because you can have multiple outputs. If so, we should probably put it in the `BaseTransactionExplanation` as an array of strings

modules/account-lib/src/coin/dot/iface.ts
gibrangul95: Yep

argjv: But we should provide that, right? Or why do we omit it? Or is it optional?

gibrangul95: Dot doesn’t have ids. Some coins do, hence its in the base class.

joey-bai: curious about lack of ID and how it'll impact our workflows.

In Solana, transaction id === transaction's first signature.
Which is only available for signed transactions.

I wonder how our system will handle transactions without IDs.

argjv (approved):

argjv (dismissed):

modules/account-lib/test/unit/coin/dot/transaction.ts
```diff
@@ -1,8 +1,17 @@
...
+import { buildTestConfig, TestDotNetwork, TEST_NETWORK_DATA } from './transactionBuilder/base';
+
+class StubTransactionBuilderFactory extends TransactionBuilderFactory {
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    const testNetwork = { ..._coinConfig.network, ...TEST_NETWORK_DATA } as TestDotNetwork;
+    super({ ..._coinConfig, network: testNetwork });
+  }
+}
 
 describe('Dot Transaction', () => {
```
argjv, line 16: If these tests are commented, do the unit tests pass? Are we able to single point if one of these tests is the culprit?
There are only 7 tests initializing a tx here.

gibrangul95: Yes, the test suite passes if we comment out the explain transaction tests.

modules/account-lib/test/unit/coin/dot/transaction.ts
```diff
@@ -51,4 +60,125 @@ describe('Dot Transaction', () => {
...
+    const { rawTx } = DotResources;
+    const sender = DotResources.accounts.account1;
+    const receiver = DotResources.accounts.account3;
+    const proxySender = DotResources.accounts.account3;
+
+    beforeEach(() => {
+      factory = register('dot', StubTransactionBuilderFactory);
+    });
+
+    it('should explain a transfer transaction', async () => {
```
alebusse, line 75: This test  throwing an unhandled exception and failing if i run it alone. But pass if i run the whole Dot Explain Transaction.

gibrangul95: Strange, working on my system

alebusse: Ignore my comment, is probably webstorm not loading the config correctly.

modules/account-lib/src/coin/dot/transaction.ts
```diff
@@ -175,6 +182,94 @@ export class Transaction extends BaseTransaction {
...
+      changeAmount: '0',
+      changeOutputs: [],
+      outputs,
+      fee: {
+        fee: result.tip?.toString() || '',
+        type: 'tip',
+      },
+      type: this.type,
+    };
+    switch (this.type) {
```
aaldemir, line 259: is addressInitialization a TransactionType for DOT? If so, why is it not included here?
If a new proxy account is set onto a primary account for example, is this an addressInitialization type?

gibrangul95: WalletIntialization is a type. Used for addProxy transactions to setup proxy accounts.

byhow:
sanity check on if increasing the memory will get a green build. feel free to squash that commit

aaldemir (approved):
LGTM

argjv (approved):

PR #1725, opened 2021-11-29
feat(core): dot sign transaction

gibrangul95:
@argjv @rliberal-simtlix @jrafael-simtlix @nimraarif21 @fdmirza

### DIFF
--- a/modules/core/src/v2/coins/dot.ts
+++ b/modules/core/src/v2/coins/dot.ts
@@ -15,6 +15,8 @@ import {
 } from '../baseCoin';
 import { NodeCallback } from '../types';
 
+const co = Bluebird.coroutine;
+
 export interface SignTransactionOptions extends BaseSignTransactionOptions {
   txPrebuild: TransactionPrebuild;
   prv: string;
@@ -27,7 +29,7 @@ export interface TransactionPrebuild {
   validity: {
     firstValid: number;
   };
-  blockHash: string;
+  referenceBlock: string;
   version: number;
 }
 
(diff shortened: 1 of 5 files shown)


modules/core/test/v2/unit/coins/dot.ts
```diff
@@ -0,0 +1,68 @@
...
+  let bitgo;
+  let basecoin;
+
+  before(function () {
+    bitgo = new TestBitGo({ env: 'mock' });
+    bitgo.initializeTestVars();
+    basecoin = bitgo.coin('tdot');
+  });
+
+  describe('Sign transaction', () => {
```
argjv, line 14: Can you add some error tests cases?

modules/core/src/v2/coins/dot.ts
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    if (!_.isNumber(addressVersion)) {
+      throw new Error('missing addressVersion parameter to sign transaction');
+    }
+
+    // if we are receiving addresses do not try to convert them
+    const signer = !accountLib.Dot.Utils.default.isValidAddress(params.txPrebuild.key)
+      ? new accountLib.Dot.KeyPair({ pub: params.txPrebuild.key }).getAddress()
+      : params.txPrebuild.key;
```
argjv, line 183: You can switch the conditional validation to be non-negative (remove the `!`) and easier to read by swapping the if/else content

modules/core/src/v2/coins/dot.ts
```diff
@@ -0,0 +1,249 @@
...
+    const self = this;
+    return co<SignedTransaction>(function* () {
+      const { txHex, signer, prv } = self.verifySignTransactionParams(params);
+      const factory = accountLib.register(self.getChain(), accountLib.Dot.TransactionBuilderFactory);
+      const txBuilder = factory.from(txHex);
+      txBuilder
+        .validity(params.txPrebuild.validity)
+        .referenceBlock(params.txPrebuild.referenceBlock)
+        .version(params.txPrebuild.version)
+        .sender({ address: signer })
```
argjv, line 209: Are these parameters always provided? Or are they optional? If they are optional, we should validate they are present (and valid) before assigning them to the builder

modules/core/src/v2/coins/dot.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  parseTransaction(
+    params: ParseTransactionOptions,
+    callback?: NodeCallback<ParsedTransaction>
+  ): Bluebird<ParsedTransaction> {
+    return Bluebird.resolve({}).asCallback(callback);
+  }
+
+  verifyAddress(params: VerifyAddressOptions): boolean {
+    return true;
```
argjv, line 239: Why not `accountLib.Dot.Utils.default.isValidAddress()`?

modules/core/src/v2/coins/dot.ts
gibrangul95: They're always mandatory. Wrote a test to ensure that.

modules/core/src/v2/coins/dot.ts
argjv: Or call `isValidAddress` from this class, no?

argjv (approved):

PR #1727, opened 2021-11-29
feat(account-lib): dot optimization

gibrangul95:
@argjv @rliberal-simtlix @jrafael-simtlix @fdmirza


argjv (approved):

rliberal-simtlix (approved):

argjv (changes requested):
Are we now making a call to the node?

gibrangul95:
> Are we now making a call to the node?

@argjv no, no calls to the node. Just an internal clean up.

argjv (approved):

PR #1731, opened 2021-11-30
verify address - rootAddress parameter

lcovar:
Request from ticket https://bitgoinc.atlassian.net/browse/BG-39092

> I meant that the root address (normally retrieved via the Get Wallet BitGo Server API as _wallet.receiveAddress.address), should instead be taken from the request parameters. This could obviously be an optional parameter, so that it does not affect other customers who do not which to specify it.For example, you could add a new option to the CreateAddressOptions struct that was the wallet root_address. Then the customer could fill it in when sending the Create Address request, and this line could be modified to only use the address from the server-retrieved wallet if the parameter is empty.

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -268,6 +268,7 @@ export interface CreateAddressOptions {
   lowPriority?: boolean;
   forwarderVersion?: number;
   format?: 'base58' | 'cashaddr';
+  rootAddress?: string;
 }
 
 export interface UpdateAddressOptions {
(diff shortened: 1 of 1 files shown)


lcovar:
@tylerlevine can you review when you get a chance

tylerlevine (changes requested):
As I've discussed many times before, this is not a good approach. Any information that we need to get from the server that's critical for address verification should be signed by the user private key at the time of wallet creation, and those signatures should be verified before the information is used. This just pushes that requirement onto the caller, which doesn't actually solve the issue.

PR #1732, opened 2021-11-30
feat(core): implement transaction signing methods

Krin2:
implement signTransaction, isValidPub, isValidPrv, verifyAddress and generateKeyPair methods and
test

STLX-8786

### DIFF
--- a/modules/core/src/v2/coins/sol.ts
+++ b/modules/core/src/v2/coins/sol.ts
@@ -2,8 +2,10 @@
  * @prettier
  */
 import * as Bluebird from 'bluebird';
-import { BaseCoin as StaticsBaseCoin, CoinFamily } from '@bitgo/statics';
+const co = Bluebird.coroutine;
 
+import { BaseCoin as StaticsBaseCoin, CoinFamily } from '@bitgo/statics';
+import * as accountLib from '@bitgo/account-lib';
 import {
   BaseCoin,
   KeyPair,
@@ -27,10 +29,7 @@ export interface TransactionFee {
 export type SolTransactionExplanation = TransactionExplanation;
 
 export interface ExplainTransactionOptions {
-  txHex?: string;
-  halfSigned?: {
-    txHex: string;
-  };
+  txBase64: string;
   publicKeys?: string[];
   feeInfo: TransactionFee;
 }
@@ -47,7 +46,7 @@ export interface SolSignTransactionOptions extends SignTransactionOptions {
   pubKeys?: string[];
 }
 export interface TransactionPrebuild extends BaseTransactionPrebuild {
-  txHex: string;
+  txBase64: string;
   txInfo: TxInfo;
   source: string;
 }
@@ -82,12 +81,13 @@ export class Sol extends BaseCoin {
     return Math.pow(10, this._staticsCoin.decimalPlaces);
   }
 
+  // TODO (https://bitgoinc.atlassian.net/browse/STLX-10202)
   verifyTransaction(params: VerifyTransactionOptions, callback?: NodeCallback<boolean>): Bluebird<boolean> {
     throw new MethodNotImplementedError('verifyTransaction method not implemented');
   }
 
   verifyAddress(params: VerifyAddressOptions): boolean {
-    throw new MethodNotImplementedError('verifyAddress method not implemented');
+    return this.isValidAddress(params.address);
   }
 
   /**
@@ -96,8 +96,9 @@ export class Sol extends BaseCoin {
    * @param {Buffer} seed - Seed from which the new keypair should be generated, otherwise a random seed is used
    * @returns {Object} object with generated pub and prv
    */
-  generateKeyPair(seed?: Buffer): KeyPair {
-    throw new MethodNotImplementedError('generateKeyPair method not implemented');
+  generateKeyPair(seed?: Buffer | undefined): KeyPair {
+    const result = seed ? new accountLib.Sol.KeyPair({ seed }).getKeys() : new accountLib.Sol.KeyPair().getKeys();
+    return result as KeyPair;
   }
 
   /**
@@ -107,7 +108,7 @@ export class Sol extends BaseCoin {
    * @returns is it valid?
    */
   isValidPub(pub: string): boolean {
-    throw new MethodNotImplementedError('isValidPub method not implemented');
+    return accountLib.Sol.Utils.isValidPublicKey(pub);
   }
 
   /**
@@ -117,11 +118,11 @@ export class Sol extends BaseCoin {
    * @returns is it valid?
    */
   isValidPrv(prv: string): boolean {
-    throw new MethodNotImplementedError('isValidPrv method not implemented');
+    return accountLib.Sol.Utils.isValidPrivateKey(prv);
   }
 
   isValidAddress(address: string): boolean {
-    throw new MethodNotImplementedError('isValidAddress method not implemented');
+    return accountLib.Sol.Utils.isValidAddress(address);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/coins/sol.ts
```diff
@@ -133,15 +157,32 @@ export class Sol extends BaseCoin {
...
+      const transaction: any = yield txBuilder.build();
+
+      if (!transaction) {
+        throw new Error('Invalid transaction');
+      }
+
+      const response = {
+        txBase64: transaction.toBroadcastFormat(),
+      };
+      return transaction.signature.length >= 2 ? response : { halfSigned: response };
```
alebusse, line 174: @rliberal-simtlix @joey-bai 
Should we just return the response without this check? Since there is no multi sign for Solana, at least for most operations that are not secondary wallet creation

joey-bai: to start, lets assume this is only for single sig.
we will need to update this impl when TSS is ready.

modules/core/src/v2/coins/sol.ts
```diff
@@ -27,9 +29,9 @@ export interface TransactionFee {
 export type SolTransactionExplanation = TransactionExplanation;
 
 export interface ExplainTransactionOptions {
-  txHex?: string;
+  txBase64?: string;
   halfSigned?: {
-    txHex: string;
+    txBase64: string;
   };
```
joey-bai, line 35: lets make `txBase64` a required string.
also, remove `halfSigned` for now. we'll add when needed

modules/core/src/v2/coins/sol.ts
```diff
@@ -83,11 +85,32 @@ export class Sol extends BaseCoin {
   }
 
   verifyTransaction(params: VerifyTransactionOptions, callback?: NodeCallback<boolean>): Bluebird<boolean> {
-    throw new MethodNotImplementedError('verifyTransaction method not implemented');
+    return co<boolean>(function* () {
```
joey-bai, line 88: lets implement `verifyTransaction` in a separate PR to keep this change more focused.

there are a lot more checks that are needed.
namely, we need to verify that the instructions when deserializing `txPrebuild.txBase64` matches the info from `params.txInfo`

modules/core/src/v2/coins/sol.ts
```diff
@@ -133,15 +157,32 @@ export class Sol extends BaseCoin {
...
+      const transaction: any = yield txBuilder.build();
+
+      if (!transaction) {
+        throw new Error('Invalid transaction');
+      }
+
+      const response = {
+        txBase64: transaction.toBroadcastFormat(),
+      };
+      return response;
```
joey-bai, line 174: response doesn't match any of the allowed types for `SignedTransaction`. since we're focusing on single sig, we can assume that this will always produce a fully signed transaction.

i.e.
```
return {
  txHex: transaction.toBroadcastFormat()
}
```

modules/core/src/v2/coins/sol.ts
```diff
@@ -133,15 +157,32 @@ export class Sol extends BaseCoin {
     params: SolSignTransactionOptions,
     callback?: NodeCallback<SignedTransaction>
   ): Bluebird<SignedTransaction> {
-    throw new MethodNotImplementedError('signTransaction method not implemented');
+    const self = this;
+    return co<SignedTransaction>(function* () {
+      const factory = accountLib.register(self.getChain(), accountLib.Sol.TransactionBuilderFactory);
+      const txBuilder = factory.from(params.txPrebuild.txBase64);
+      txBuilder.sign({ key: params.prv });
+      const transaction: any = yield txBuilder.build();
```
joey-bai, line 165: can be more strongly typed. we can use `BaseTransaction` from `account-lib`

modules/core/src/v2/coins/sol.ts
Krin2: I create [this story](https://bitgoinc.atlassian.net/browse/STLX-10202) to implement verifyTransaction.
Could you please add a description and a DoD? (I think it will be better if you add it)

modules/core/src/v2/coins/sol.ts
```diff
@@ -27,10 +29,7 @@ export interface TransactionFee {
 export type SolTransactionExplanation = TransactionExplanation;
 
 export interface ExplainTransactionOptions {
-  txHex?: string;
-  halfSigned?: {
-    txHex: string;
-  };
```
0xabrarhussain, line 33: seems like we'd want this? see that this is included for some points, and not then not included for others. where it is, have the following code under `explainTransaction`

```
      const txHex = params.txHex || (params.halfSigned && params.halfSigned.txHex);
      if (!txHex || !params.feeInfo) {
      }
```

joey-bai: @0xabrarhussain imo i don't think this is necessary. i'm not sure why we need to have clients differentiate between an unsigned/halfsigned tx when calling this as we won't do anything differently.

some examples where we don't differentiate:
```
// XLM
interface ExplainTransactionOptions {
    txBase64: string;
}

// EOS
interface ExplainTransactionOptions {
    transaction: {
        packed_trx: string;
    };
    headers: EosTransactionHeaders;
}
```

modules/core/test/v2/unit/coins/sol.ts
```diff
@@ -75,7 +93,26 @@ describe('SOL:', function () {
 
   describe('Sign transaction:', () => {
     it('should sign transaction', async function () {
-      await should.throws(() => basecoin.signTransaction('placeholder'), 'signTransaction method not implemented');
+      const signed = await basecoin.signTransaction({
+        txPrebuild: {
+          txBase64: resources.RAW_TX_UNSIGNED,
+          keys: [resources.accountWithSeed.publicKey.toString()],
+          addressVersion: 1,
```
joey-bai, line 100: what's `addressVersion` used for? is it required?
if not, let's remove it.

joey-bai (dismissed):
LGTM, comment is not a blocker.

modules/core/src/v2/coins/sol.ts
0xabrarhussain: looking at it, I think it's coin specific. I see that some coins have a `txhex` in the half-signed object while others don't

joey-bai: @alebusse good to know, but from the client's perspective, it should matter if the call `explainTransaction` using an unsigned, half signed, or fully signed hex.

looking at the param, the client already knows if the transaction is half-signed or not. otherwise, they wouldn't know whether to put their tx string in `halfSigned` or not.

e.g.
```
// we expect the two to be equivalent
const halfSigned = sol.explainTransaction({ halfSigned: { txBase64: halfSignedBase64 }});
const alsoHalfSigned = sol.explainTransaction({ txBase64: halfSignedBase64 });
```

modules/core/test/v2/unit/coins/sol.ts
Krin2: removed

modules/core/src/v2/coins/sol.ts
rliberal-simtlix: If tx initially is single sig. Why we should have halfsigned tx? For this will be only signed or unsigned for this initial case. We should discard halfsigned tx for Solana.
OVC will be used for initial case that we are supporting only single sig?
This discussion should be taken when we encourage the TSS approach and Cold Wallet Integration.

joey-bai: agreed - we can add it if/when we have a need to differentiate between a half-signed tx or not.

joey-bai (approved):

PR #1734, opened 2021-11-30
feat(core): implement message signing

facundoCarreSimtlix:
implement message signing

stlx_8108

### DIFF
--- a/modules/core/test/v2/unit/coins/sol.ts
+++ b/modules/core/test/v2/unit/coins/sol.ts
@@ -10,6 +10,11 @@ describe('SOL:', function () {
   const badAddresses = resources.addresses.invalidAddresses;
   const goodAddresses = resources.addresses.validAddresses;
   
+  const keypair = {
+    pub: resources.accountWithSeed.publicKey,
+    prv: resources.accountWithSeed.privateKey.base58,
+  };
+
   before(function () {
     bitgo = new TestBitGo({ env: 'mock' });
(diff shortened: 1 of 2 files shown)


modules/core/test/v2/unit/coins/sol.ts
```diff
@@ -78,4 +83,20 @@ describe('SOL:', function () {
...
   });
+
+  describe('Sign message', () => {
+    it('should sign message', async function () {
+      const signed = await basecoin.signMessage(
+        keypair,
+        'signed message',
+      );
+      signed.toString('base64').should.equal('s+7d/8aW/twfM/0wLSKOGxd9+LhDIiz/g0FfJ39ylJhQIkjK0RYPm/Y+gdeJ5DIy6K6h6gCXXESDomlv12DBBQ==');
+    });
```
joey-bai, line 94: let's use a real solana transaction as the message. then verify the signed message can be deserialized using our transaction builder w/ expected payload.

facundoCarreSimtlix: i dont undestand why i need to use a transaction, maybe change String message ? , this function is for sign a message is not for signed transaction

joey-bai: sorry, there was some confusion on my end. this is okay

joey-bai (dismissed):

alebusse (approved):

PR #1735, opened 2021-12-01
fix(ltc): update block explorer link for ltc

zhongtan912:
Ticket: BG-39883

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -328,7 +328,7 @@ class Litecoin extends Mainnet implements UtxoNetwork {
   name = 'Litecoin';
   family = CoinFamily.LTC;
   utxolibName = 'litecoin';
-  explorerUrl = 'https://live.blockcypher.com/ltc/tx/';
+  explorerUrl = 'https://blockchair.com/litecoin/transaction/';
 }
 
 class LitecoinTestnet extends Testnet implements UtxoNetwork {


YPAzevedo (approved):

tylerlevine (changes requested):
Ticket says the link in the UI goes to `http://explorer.litecoin.net/` which does indeed appear to be down.

But that's not what we have in `statics` here. Are you sure this is the right place to change the explorer URL? Where is the litecoin.net URL coming from?

zhongtan912:
> Ticket says the link in the UI goes to `http://explorer.litecoin.net/` which does indeed appear to be down.
> 
> But that's not what we have in `statics` here. Are you sure this is the right place to change the explorer URL? Where is the litecoin.net URL coming from?

Thanks @tylerlevine for highlighting this. Let me investigate further the where the incorrect data is coming from. Will put this in draft for now.

zhongtan912:
> > Ticket says the link in the UI goes to `http://explorer.litecoin.net/` which does indeed appear to be down.
> > But that's not what we have in `statics` here. Are you sure this is the right place to change the explorer URL? Where is the litecoin.net URL coming from?
> 
> Thanks @tylerlevine for highlighting this. Let me investigate further the where the incorrect data is coming from. Will put this in draft for now.

Have confirmed that bad data was indeed coming from metadata within lumina submodule. Nonetheless, it would be good to update the explorer link to the one the product team is using (as mentioned in the ticket) to stay in sync

tylerlevine (approved):

PR #1736, opened 2021-12-01
fix(hbar-validateaddress): add validation for hedera addresses fix ca…

FabianSimtlix:
…se where hex address were valid

stlx-10067


shihao-guo (approved):
Thanks for taking care of this!

PR #1737, opened 2021-12-01
feat(core): fIx version ripple-keypairs

Joelsx:
FIx version ripple-keypairs

STLX-9667

### DIFF
--- a/modules/core/package.json
+++ b/modules/core/package.json
@@ -73,7 +73,7 @@
     "moment": "^2.20.1",
     "ripple-address-codec": "~4.1.3",
     "ripple-binary-codec": "~0.2.4",
-    "ripple-keypairs": "^0.11.0",
+    "ripple-keypairs": "~0.11.0",
     "ripple-lib": "~1.4.1",
     "sanitize-html": "^1.27.5",
     "secp256k1": "^4.0.2",


tylerlevine (changes requested):
This makes no difference.  Please review the semver behavior using this tool: https://semver.npmjs.com/

Joelsx:
> This makes no difference. Please review the semver behavior using this tool: https://semver.npmjs.com/

We need to set fix the version since rigth now we are having problems with newer versions were they need new npm and node. @FabianSimtlix

FabianSimtlix:
> This makes no difference. Please review the semver behavior using this tool: https://semver.npmjs.com/

Basically with the last version (1.1.0) of this lib this change was added and that's why we need to prevent npm from getting the latest version and continue using the previous one.
image
https://github.com/XRPLF/xrpl.js/blob/ripple-keypairs%401.1.0/packages/ripple-keypairs/package.json

tylerlevine:
> > This makes no difference. Please review the semver behavior using this tool: https://semver.npmjs.com/
> 
> We need to set fix the version since rigth now we are having problems with newer versions were they need new npm and node. @FabianSimtlix

Yes, I understand that. This change will not fix that issue.

> > This makes no difference. Please review the semver behavior using this tool: https://semver.npmjs.com/
> 
> Basically with the last version (1.1.0) of this lib this change was added and that's why we need to prevent npm from getting the latest version and continue using the previous one. image https://github.com/XRPLF/xrpl.js/blob/ripple-keypairs%401.1.0/packages/ripple-keypairs/package.json

Yes, I understand this too. The current semver string precludes version 1.1.0, so it's probably coming in from some transitive dependency.

PR #1738, opened 2021-12-01
refactor(core): add BaseApi for recovery

OttoAllmendinger:
Unify request handling and error wrapping.

BaseApi returns a `Response<T>` object which provides a method
`map<T, U>(f: (body: T) => U)`.

Any errors thrown inside `map` will automatically wrapped with
`ApiRequestError`.

Issue: BG-38773


modules/core/src/v2/coins/utxo/recovery/blockchairApi.ts
```diff
@@ -67,11 +60,13 @@ export class BlockchairApi implements RecoveryProvider {
     if (!address || address.length === 0) {
       throw new Error('invalid address');
     }
-    const response = await request.get(this.getExplorerUrl(`/dashboards/address/${address}`));
-    return {
-      txCount: response.body.data[address].address.transaction_count,
-      totalBalance: response.body.data[address].address.balance,
-    };
+    const res = await this.get<any>(`/dashboards/address/${address}`);
```
OttoAllmendinger, line 63: I will introduce better types in a subsequent PR

modules/core/src/v2/coins/utxo/recovery/baseApi.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (!path.startsWith(path)) {
+      throw new Error(`path must start with '/' character`);
+    }
+    const url = this.baseUrl + path;
+    let response;
+    try {
+      response = await superagent(method, url)
+        .retry(params.retry)
+        // eslint-disable-next-line @typescript-eslint/ban-types
+        .send(requestBody as object);
```
tylerlevine, line 60: can you use `Record` here to avoid the `ban-types` violation?

tylerlevine (approved):

modules/core/src/v2/coins/utxo/recovery/baseApi.ts
OttoAllmendinger: indeed I can, will add it to the follow-up PR

PR #1739, opened 2021-12-01
fix(sol): get signature data from a Sol transaction

joey-bai:
previously, this would always return empty.

Ticket: STLX-10065


alebusse (dismissed):

CapnMigraine (approved):
lgtm

PR #1743, opened 2021-12-02
test(core): add parametric backupKeyRecovery tests

OttoAllmendinger:
Covered usages:

* "unsigned sweep" mode (no signatures)
* "Key Recovery Service (KRS)" mode (only user signature)
* Full-Signed recovery (user signature and backup signature)
* Full-Signed recovery with custom user key prefix

Issue: BG-38773


modules/core/src/v2/coins/utxo/WalletKeys.ts
```diff
@@ -0,0 +1,96 @@
+/**
+ * Classes for deriving key triples for wallet addresses.
+ *
+ * By default, BitGo wallets consist of a triple of bip32 extend keypairs.
```
sangaman, line 4: nit: "extended" vs "extend"

modules/core/src/v2/coins/utxo/WalletKeys.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+import * as bip32 from 'bip32';
+import { Triple } from '../../triple';
+
+export function eqPublicKey(a: bip32.BIP32Interface, b: bip32.BIP32Interface): boolean {
+  return a.publicKey.equals(b.publicKey);
+}
+
+/**
+ * Base method for
```
sangaman, line 24: Looks like this comment is incomplete or unintended?

modules/core/src/v2/coins/utxo/WalletKeys.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+export class WalletKeys {
+  public readonly publicKeys: Triple<Buffer>;
+
+  constructor(public readonly triple: Triple<bip32.BIP32Interface>) {
+    triple.forEach((a, i) => {
+      triple.forEach((b, j) => {
+        if (eqPublicKey(a, b) && i !== j) {
```
sangaman, line 32: Just a suggestion but since we always have exactly 3 keys and there are only 3 comparisons necessary I find it a bit more succinct/readable to do
`if(eqPublicKey(this,user, this.backup) || eqPublicKey(this,user, this.bitgo) || eqPublicKey(this,backup this.bitgo))` - but that's just me so feel free to keep as is.

modules/core/src/v2/coins/utxo/WalletKeys.ts
```diff
@@ -0,0 +1,96 @@
...
+  constructor(public parent: RootWalletKeys, public paths: Triple<string>) {
+    super(parent.triple.map((k, i) => k.derivePath(paths[i])) as Triple<bip32.BIP32Interface>);
+  }
+}
+
+export class RootWalletKeys extends WalletKeys {
+  static readonly defaultPrefix = '0/0';
+  constructor(
+    triple: Triple<bip32.BIP32Interface>,
+    public readonly derivationPrefixes: Triple<string> = [
```
sangaman, line 64: Is the only time we'd be overriding this default prefix for "exotic" derivation like mentioned in the tests? I was under the impression that all 3 keys ought to be using the same derivation path but I figure that's not always the case.

sangaman (approved):
Just a few questions and nits. I like the new wallet extended keys classes a lot.

modules/core/src/v2/coins/utxo/WalletKeys.ts
OttoAllmendinger: it can be different for wallets migrated from v1 to v2

we will use it here: https://github.com/BitGo/BitGoJS/blob/master/modules/core/src/v2/coins/utxo/recovery/backupKeyRecovery.ts#L357-L366

modules/core/src/v2/coins/utxo/WalletKeys.ts
OttoAllmendinger: oops

will fix in follow-up

modules/core/src/v2/coins/utxo/WalletKeys.ts
OttoAllmendinger: will fix in follow-up

modules/core/src/v2/coins/utxo/WalletKeys.ts
OttoAllmendinger: 3 is right on the threshold ;)

I will leave it as is

PR #1744, opened 2021-12-02
test(core): move bespoke btc tests to utxo/custom

OttoAllmendinger:
There are now no custom one-off utxo tests in `unit/coins`.

Issue: BG-35154


sangaman (approved):
Eventually the hope here is to remove the "custom" tests altogether and only have generic ones? I suppose with the exception of testing logic that truly is specific to a particular coin.

OttoAllmendinger:
Yes, that is the goal. From what I can see, nonstandard address formats are the only custom tests at this point.

PR #1745, opened 2021-12-02
fix(core): implement explainTransaction for p2tr

OttoAllmendinger:
Issue: BG-35154

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -79,11 +79,12 @@ export interface TransactionFee {
 export interface TransactionExplanation {
   displayOrder: string[];
   id: string;
+  locktime: number;
   outputs: Output[];
   changeOutputs: Output[];
-  outputAmount: string;
+  outputAmount: number;
   changeAmount: number;
-  fee: TransactionFee;
+  fee: TransactionFee | string;
 
   inputSignatures: number[];
   signatures: number | false;
@@ -1258,12 +1259,12 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     if (txInfo && txInfo.changeAddresses) {
       changeAddresses = txInfo.changeAddresses;
     }
-    const explanation: any = {
+    const explanation = {
       displayOrder: ['id', 'outputAmount', 'changeAmount', 'outputs', 'changeOutputs'],
       id: id,
-      outputs: [],
-      changeOutputs: [],
-    };
+      outputs: [] as Output[],
+      changeOutputs: [] as Output[],
+    } as TransactionExplanation;
 
     transaction.outs.forEach((currentOutput) => {
       const currentAddress = utxolib.address.fromOutputScript(currentOutput.script, this.network);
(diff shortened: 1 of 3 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1236,136 +1237,93 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
-          if (_.isEmpty(sig)) {
-            return false;
-          }
+    const prevOutputs = params.txInfo?.unspents.map((u) => toOutput(u, this.network));
 
-          const parentTxId = (Buffer.from(input.hash).reverse() as Buffer).toString('hex');
-          const inputId = `${parentTxId}:${input.index}`;
-          const amount = unspentValues[inputId];
+    // get information on tx inputs
+    const inputSignatures = transaction.ins.map((input, idx): number => {
```
sangaman, line 1306: Input signatures represents the number of valid signatures for each input in the transaction right? I think that could be better explained in comments, either here or on the `TransactionExplanation` type. This had me confused for a while because I thought this was an array of the actual signatures, that's what the current name makes it seem like.

(Again I realize it's mostly old code, but might as well improve while we're here)

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1236,136 +1237,93 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
-        });
+      if (!prevOutputs) {
+        throw new Error(`invalid state`);
+      }
 
-        return validSignatures.reduce((validCount, isValid) => (isValid ? validCount + 1 : validCount), 0);
-      });
+      try {
+        return utxolib.bitgo
+          .getSignatureVerifications(transaction, idx, txInfo.unspents[idx].value, {}, prevOutputs)
```
sangaman, line 1317: This `getSignatureVerifications` function is what makes this all work for p2tr, right?

sangaman (approved):
Good fix and cleanup. I had a few comments but I don't consider them blockers since it relates to existing code.

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: yep (the prevOut parameter in particular)

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1236,136 +1237,93 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
 
-        spendAmount += currentAmount;
-        explanation.outputs.push({
+    const id = transaction.getId();
+    let changeAddresses: string[] = [];
+    let spendAmount = 0;
+    let changeAmount = 0;
+    const txInfo = _.get(params, 'txInfo');
+    if (txInfo && txInfo.changeAddresses) {
+      changeAddresses = txInfo.changeAddresses;
```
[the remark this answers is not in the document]
OttoAllmendinger: will do in follow-up

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: will do in follow-up

PR #1749, opened 2021-12-03
fix(sol): fix deserializing signed sol transaction

joey-bai:
when deserializing a sol transaction, we would lose out on durable nonce
info. this fixes it.

Ticket: STLX-10065


modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -140,10 +140,21 @@ export class Transaction extends BaseTransaction {
...
+        walletNonceAddress: nonceInstruction.noncePubkey.toString(),
+        authWalletAddress: nonceInstruction.authorizedPubkey.toString(),
+      };
+    }
+
     const result: TxData = {
       id: this.id,
       feePayer: this._solTransaction.feePayer?.toString(),
       nonce: this.getNonce(),
+      durableNonce: durableNonce,
```
rliberal-simtlix, line 157: So here in TxData durable nonce could be undefined?
We should build a TxData with and without this field?
@CapnMigraine , @joey-bai what you recommend here?

alebusse: Undefined should be fine. Idk whats gonna be this field data be used for

alebusse (approved):

modules/account-lib/src/coin/sol/transaction.ts
joey-bai: @rliberal-simtlix yes, the blockhash/nonce can come from two sources:
- recent value from full node
- value from a durable nonce account

`durableNonce` is used to indicate that the nonce value comes from the nonce account.
if this field is empty, then the nonce is a recent value from the full node.

rliberal-simtlix: Sounds good. My question was related in order to add increase the payload size of tx, which base on the fee the tx per byte could increase cost. But in that case that you mention i think is ok.
Thanks for the clarification.

rliberal-simtlix (approved):

PR #1751, opened 2021-12-03
fix(core): use wallet keys in explainTransaction

OttoAllmendinger:
We want to know that a transaction was signed by the wallet keys.

Issue: BG-38773

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -87,8 +87,16 @@ export interface TransactionExplanation {
   changeAmount: number;
   fee: TransactionFee | string;
 
+  /**
+   * Number of input signatures per input.
+   * NOTE: only counts valid signatures, not the public keys they belong to
+   */
   inputSignatures: number[];
-  signatures: number | false;
+
+  /**
+   * Highest input signature count for the transaction
+   */
+  signatures: number;
 }
 
 export interface ExplainTransactionOptions {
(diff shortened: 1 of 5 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -1311,9 +1322,9 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
 
     const prevOutputs = params.txInfo?.unspents.map((u) => toOutput(u, this.network));
 
-    // get information on tx inputs
-    const inputSignatures = transaction.ins.map((input, idx): number => {
-      if (!txInfo || txInfo.unspents.length !== transaction.ins.length) {
+    // get the number of signatures per input
+    const inputSignatureCounts = transaction.ins.map((input, idx): number => {
```
sangaman, line 1326: :+1:

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -560,6 +562,64 @@ export function verifySignature(
...
+      (publicKey.length === 33 && isSignatureByPublicKey(v, publicKey.slice(1))))
+  );
+}
+
+/**
+ * @param transaction
+ * @param inputIndex
+ * @param prevOutputs - transaction outputs for inputs
+ * @param publicKeys - public keys to check signatures for
+ * @return array with signature verification result for each public key in publicKeys
```
sangaman, line 584: Minor note but I think still helpful in making this all easy to follow, I'd say something more like "boolean array indicating whether a valid signature exists for each public key in publicKeys" - this as opposed to "signature verification result" since we have a `SignatureVerification` type in use that's distinct from what this method is returning.

modules/core/src/v2/coins/utxo/sign.ts
```diff
@@ -91,6 +91,34 @@ export function signWalletTransactionWithUnspent(
   );
 }
 
+/**
+ * @param tx
+ * @param inputIndex
+ * @param unspents
+ * @param walletKeys
+ * @return signature verification result for each key
```
sangaman, line 99: Same here as above comment.

sangaman (dismissed):
Very nice.

modules/core/src/v2/coins/utxo/sign.ts
OttoAllmendinger: improved docstring

modules/utxo-lib/src/bitgo/signature.ts
OttoAllmendinger: improved docstring

tylerlevine (approved):

PR #1755, opened 2021-12-03
feat(sol): address fee and id in explain transaction

joey-bai:
show 'UNAVAILABLE' for unsigned transactions and fees.
in solana, transaction id is the first signature, thus unavailable
for unsigned transactions.

add fee to transaction builder, which is set when calling
explainTransaction in core.

Ticket: STLX-8791


modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -27,12 +28,22 @@ export class Transaction extends BaseTransaction {
...
 
-  set id(id: string) {
-    this._id = id;
+  /** @inheritDoc **/
+  get id(): string {
+    // Solana transaction ID === first signature: https://docs.solana.com/terminology#transaction-id
+    if (this._solTransaction.signature) {
+      return base58.encode(this._solTransaction.signature);
+    } else {
+      return 'UNAVAILABLE';
```
rliberal-simtlix, line 37: make UNAVAILABLE as a cons

modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -266,16 +282,16 @@ export class Transaction extends BaseTransaction {
...
       changeOutputs: [],
       changeAmount: '0',
       outputAmount: outputAmount.toFixed(0),
       outputs: outputs,
       fee: {
-        // TODO(STLX-8791): need to figure out how to get fee information.
-        fee: '0',
+        fee: feeString,
+        feeRate: this.lamportsPerSignature,
+        type: 'perSignature',
```
rliberal-simtlix, line 294: Where is this type field?

modules/account-lib/src/coin/sol/transactionBuilder.ts
```diff
@@ -82,11 +80,14 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
    * Builds the solana transaction.
    */
   protected buildSolTransaction(): SolTransaction {
+    assert(this._sender, new BuildTransactionError('sender is required before building'));
```
rliberal-simtlix, line 83: Good check :+1:

modules/core/src/v2/coins/sol.ts
```diff
@@ -23,7 +23,7 @@ import { MethodNotImplementedError } from '../../errors';
 import base58 = require('bs58');
 
 export interface TransactionFee {
-  fee: string;
+  lamportsPerSignature: string;
```
rliberal-simtlix, line 26: Here on transactionBuilder from account-lib you are setting the fee as lamportPerSignature, so if this is gonna be exposed by the core I think fee name must not changed. In order to have a common interface like in the txBuilder.

modules/core/src/v2/coins/sol.ts
```diff
@@ -178,6 +178,28 @@ export class Sol extends BaseCoin {
...
     callback?: NodeCallback<SolTransactionExplanation>
   ): Bluebird<SolTransactionExplanation> {
-    throw new MethodNotImplementedError('explainTransaction method not implemented');
+    const self = this;
+    return co<SolTransactionExplanation>(function* () {
+      const factory = accountLib.register(self.getChain(), accountLib.Sol.TransactionBuilderFactory);
+      let rebuiltTransaction;
+
+      try {
+        rebuiltTransaction = yield factory
```
rliberal-simtlix, line 187: You hace here try and then an if to comparing the rebuiltTransaction.
If the try is success, could be the rebuilt transaction undefined for the next if?

modules/account-lib/src/coin/baseCoin/iface.ts
```diff
@@ -120,3 +120,13 @@ export interface TransactionExplanation {
   changeAmount: string;
   fee: TransactionFee;
 }
+
+/**
+ * Fee options.
+ */
+export type FeeOptions = {
```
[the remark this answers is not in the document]
joey-bai: these fee options are meant to be standardized.

it was previously defined in Dot, but i moved it out here so it can be re-used for multiple coins.
original comment was: `Fee options based on the BitgoJS standard doc.`

modules/account-lib/src/coin/sol/transaction.ts
joey-bai: it comes from the base fee: `baseCoin/iface.ts`:
```
export interface TransactionFee {
  fee: string;
  feeRate?: number;
  size?: number;
  type?: string;
}
```

joey-bai: now that you pointed this out, type should probably be 'base'.
going to remove this for now since this is an optional field.

modules/core/src/v2/coins/sol.ts
joey-bai: good to know, renamed back to fee

rliberal-simtlix (approved):
Approved, but we should need discuss a little bit further why we add the fee into the builder and then is not used when is broadcasted. As you mention there is a compatibility issue and now explain tx belongs also to account-lib. But we need review why a builder param is finally not stored in tx that we are building.

When we have a .from() for a raw tx we need to pass the fee param again to the builder if we want to use the explain tx method.
This mean that we have to get this fee from another source and we just can pass the calculated fee without setting the fee as build param.

If we don't have another approach let's pass the fee as param, but i think we need to analyze how to improve this.

cc: @alebusse , @CapnMigraine.

joey-bai:
> Approved, but we should need discuss a little bit further why we add the fee into the builder and then is not used when is broadcasted. As you mention there is a compatibility issue and now explain tx belongs also to account-lib. But we need review why a builder param is finally not stored in tx that we are building.
> 
> When we have a .from() for a raw tx we need to pass the fee param again to the builder if we want to use the explain tx method. This mean that we have to get this fee from another source and we just can pass the calculated fee without setting the fee as build param.
> 
> If we don't have another approach let's pass the fee as param, but i think we need to analyze how to improve this.
> 
> cc: @alebusse , @CapnMigraine.

Agree that we should improve the pattern if possible, but merging as is first.

PR #1761, opened 2021-12-07
feat(sol): implement parse transaction

joey-bai:
implements parse transaction for Solana

also sneaks in a fix to calculate `fee` correctly by using the transaction's `numSignaturesRequired` instead of current number of signatures.

Ticket: STLX-10400

### DIFF
--- a/modules/account-lib/src/coin/sol/transaction.ts
+++ b/modules/account-lib/src/coin/sol/transaction.ts
@@ -30,6 +30,10 @@ export class Transaction extends BaseTransaction {
     this._solTransaction = tx;
   }
 
+  private get numberOfRequiredSignatures(): number {
+    return this._solTransaction.compileMessage().header.numRequiredSignatures;
+  }
+
   /** @inheritDoc **/
   get id(): string {
     // Solana transaction ID === first signature: https://docs.solana.com/terminology#transaction-id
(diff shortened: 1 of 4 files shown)


modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -30,6 +30,10 @@ export class Transaction extends BaseTransaction {
     this._solTransaction = tx;
   }
 
+  private get numSignaturesRequired(): number {
```
rliberal-simtlix, line 33: I think here we need to follow the bitgo standard for this. Here is related to number of signatures.
https://docs.google.com/document/d/1cMGyXa8GwgwDMcZ2Uv7dMDsqRWlIEp1MRdVgpPW8p1o/edit#bookmark=id.u64ygc8x58e5

modules/core/test/v2/unit/coins/sol.ts
```diff
@@ -72,8 +72,92 @@ describe('SOL:', function () {
...
+        txBase64: testData.rawTransactions.walletInit.unsigned,
+        feeInfo: {
+          fee: '5000',
+        },
+      });
+
+      parsedTransaction.should.deepEqual({
+        inputs: {
+          address: '8Y7RM6JfcX4ASSNBkrkrmSbRu431YVi9Y3oLFnzC2dCh',
+          amount: 310000,
```
rliberal-simtlix, line 130: I'm not following why here the fee is charged on the input as 10000 and the feeInfo is 5000 lamports.

alebusse: The input fee is the cost per signature, and since its a walletInit it has 2 signatures

rliberal-simtlix: Excellent.

modules/account-lib/src/coin/sol/transaction.ts
joey-bai: good catch, renamed

PR #1764, opened 2021-12-07
feat(utxo-lib): add `bitgo/wallet` package

OttoAllmendinger:
Allows easy reuse of `WalletUnspent`, `WalletKeys` and related methods in
different projects.

Most of these classes are slightly adjusted copies of corresponding methods
in `modules/core/.../coins/utxo/`.

Since `@bitgo/utxo-lib` is a much smaller dependency than `core`, this 
makes the classes available to a broader range of projects.

Issue: BG-38773


modules/utxo-lib/src/bitgo/Unspent.ts
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Public unspent data in BitGo-specific representation.
+ */
+export interface Unspent {
+  /**
+   * Format: ${txid}:${vout}
+   */
+  id: string;
+  /**
+   * In network-specific representation
```
sangaman, line 15: This feels a bit terse - maybe "The network-specific encoded address"

modules/utxo-lib/src/bitgo/Unspent.ts
```diff
@@ -0,0 +1,106 @@
...
+   */
+  address: string;
+  /**
+   * In satoshi
```
sangaman, line 19: "The amount in satoshis"

modules/utxo-lib/src/bitgo/Unspent.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @return TxOutPoint
+ */
+export function parseOutputId(outputId: string): TxOutPoint {
+  const parts = outputId.split(':');
+  if (parts.length !== 2) {
+    throw new Error(`invalid outputId, must have format txid:vout`);
+  }
+  const [txid, voutStr] = parts;
+  const vout = Number(voutStr);
+  if (txid.length !== 64) {
```
sangaman, line 45: Is it worthwhile to check here that the txid is valid hex?

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -0,0 +1,70 @@
...
+import { UtxoTransaction } from '../UtxoTransaction';
+import { Triple } from '../types';
+import { toOutput, Unspent } from '../Unspent';
+
+export interface WalletUnspent {
+  id: string;
+  address: string;
+  value: number;
+  index: number;
+  chain: number;
```
sangaman, line 17: Would it make to sense to use a type from `Codes` in `unspents` so we're restricting this to one of the known/valid chains?

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -0,0 +1,70 @@
...
+import { createOutputScript2of3, scriptTypeForChain } from '../outputScripts';
+import { toOutputScript } from '../../address';
+import { signInput2Of3, verifySignatureWithPublicKeys } from '../signature';
+import { WalletUnspentSigner } from './WalletUnspentSigner';
+import { RootWalletKeys } from './WalletKeys';
+import { UtxoTransaction } from '../UtxoTransaction';
+import { Triple } from '../types';
+import { toOutput, Unspent } from '../Unspent';
+
+export interface WalletUnspent {
```
sangaman, line 12: I figure this should extend/implement the `Unspent` interface and inherit the `id`/`address`/`value` fields.

modules/utxo-lib/src/bitgo/wallet/chains.ts
```diff
@@ -0,0 +1,100 @@
...
+ * The address script type (ScriptType2Of3) is defined by the `chain` parameter.
+ *
+ * This file defines the mapping between chain parameter and address type.
+ */
+import { ScriptType2Of3 } from '../outputScripts';
+
+/**
+ * All valid chain codes
+ */
+export const chainCodes = [0, 1, 10, 11, 20, 21, 30, 31];
```
sangaman, line 17: This has me again wondering whether we should be inheriting this (as well as other types/constants below) from `Codes` in `unspents` - and I'm guessing this may be due to concerns over circular dependencies since `unspents` uses `utxo-lib`? If so the best solution might be to have the `Codes` types in a separate module. However that'd be out of scope for this PR. I'm just concerned that it's a bit confusing/harder to maintain with the codes defined in multiple places.

sangaman (commented):
Some questions/notes below. Conceptually I'm a fan of the changes.

modules/utxo-lib/src/bitgo/Unspent.ts
OttoAllmendinger: nah I don't think it is - this is just to catch the worst offenders

it gets checked further down the line in `addInput`

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
OttoAllmendinger: good idea

modules/utxo-lib/src/bitgo/wallet/chains.ts
OttoAllmendinger: I actually want this file to supersede `Codes` (I sort of mention it in the commit message). Let me add a deprecation notice to `@bitgo/unspents`.

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
OttoAllmendinger: Sadly we are not really restricting it anyway, since we are not parsing the WP result (yet) (see https://github.com/gcanti/io-ts/blob/master/index.md). We basically just cast the WP response without actually validating it. 

Let's do it and see how it goes.

OttoAllmendinger: done

modules/utxo-lib/src/bitgo/Unspent.ts
OttoAllmendinger: improved docstring

modules/utxo-lib/src/bitgo/Unspent.ts
OttoAllmendinger: improved docstring

modules/utxo-lib/src/bitgo/wallet/chains.ts
OttoAllmendinger: actually let's do this in a separate PR, ideally after https://github.com/BitGo/BitGoJS/pull/1765

modules/core/src/v2/coins/utxo/recovery/backupKeyRecovery.ts
```diff
@@ -253,43 +258,24 @@ export async function backupKeyRecovery(
...
+          (addressType) => coin.supportsAddressType(addressType) && !params.ignoreAddressTypes?.includes(addressType)
+        )
+        .reduce(
+          (queries, addressType) => [
+            ...queries,
+            queryBlockchainUnspentsPath(coin, params, walletKeys, getExternalChainCode(addressType)),
+            queryBlockchainUnspentsPath(coin, params, walletKeys, getInternalChainCode(addressType)),
+          ],
+          [] as Promise<WalletUnspent[]>[]
+        )
```
brandonblack, line 274:
```suggestion
        .map((addressType) => ([
          queryBlockchainUnspentsPath(coin, params, walletKeys, getExternalChainCode(addressType)),
          queryBlockchainUnspentsPath(coin, params, walletKeys, getInternalChainCode(addressType)),
        ]))
        .flat()
```

brandonblack (dismissed):
One suggestion for code clarity that was hard for me to parse while reviewing. Otherwise seems like a good direction to go.

modules/utxo-lib/src/bitgo/wallet/chains.ts
sangaman: Alright that's fine with me, I'm not too concerned where it ends up, just that it only lives in one place. (and I'll try to read commit messages more closely in the future)

modules/core/src/v2/coins/utxo/recovery/backupKeyRecovery.ts
OttoAllmendinger: sadly not available with the current compilation target set in `tsconfig.json` (we need es2019 and have es2017)

I didn't want to blow up this PR even more

tylerlevine (approved):

modules/core/src/v2/coins/utxo/recovery/backupKeyRecovery.ts
OttoAllmendinger: https://github.com/BitGo/BitGoJS/pull/1774

PR #1765, opened 2021-12-08
Remove coroutines

tylerlevine:
Remove Bluebird coroutines from all v2 code in favor of async/await.

Due to pattern inconsistencies, this was not done with an automated source code refactor so there may be some parts that were not yet converted.

There are some minor behavior changes in order to align with inherited types. See last commit for more details. We are publishing the next version of the `core` package as a major version so breaking changes are acceptable, but we should be updating any calling code that relies on the old behavior when we update downstream packages to the new version.

Ticket: BG-31214


OttoAllmendinger (approved):
:crown:

PR #1771, opened 2021-12-09
chore(account-lib): export polkadot interface

byhow:
In order to use the type def outside of account-lib, we will need to export the type like eth or celo do.

Ticket: STLX-8335


byhow:
Will rebase master once the fix on bluebird removal is merged #1772

byhow:
rebased

tylerlevine (approved):

PR #1774, opened 2021-12-09
chore(root): target ES2019

OttoAllmendinger:
For our node deployments this should not be an issue, since require Node 12 anyway.

The remaining question is browser support, which looks good for the 
features we care about:

Support: https://caniuse.com/?search=es2019

Issue: BG-40313

### DIFF
--- a/modules/core/package.json
+++ b/modules/core/package.json
@@ -16,7 +16,7 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=10.22.0 <17",
+    "node": ">=12 <17",
     "npm": ">=3.10.10"
   },
   "scripts": {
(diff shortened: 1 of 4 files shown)


brandonblack (approved):
LGTM, but I don't know much about what browsers we need to target.

modules/core/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=10.22.0 <17",
+    "node": ">=12 <17",
```
tylerlevine, line 19: we can go to 14 here

tylerlevine (changes requested):
minor fix to engines

modules/core/package.json
OttoAllmendinger: coming up

tylerlevine:
also could you please add `engines` here and set to 14? https://github.com/BitGo/BitGoJS/blob/master/modules/account-lib/package.json

tylerlevine (approved):

PR #1775, opened 2021-12-09
feat(core): dot core explain tx

nimraarif21:


modules/core/src/v2/coins/dot.ts
```diff
@@ -221,6 +210,27 @@ export class Dot extends BaseCoin {
...
   }
 
+  /**
+   * Explain/parse transaction
+   * @param params
+   * @param callback
+  */
+  async explainTransaction(params: ExplainTransactionOptions): Promise<accountLib.Dot.TransactionExplanation> {
+    const txHex = params.txPrebuild.txHex;
+    if (!txHex || !params.txPrebuild.validity || !params.txPrebuild.key || !params.txPrebuild.referenceBlock || !params.txPrebuild.version) {
```
aaldemir, line 220: `!params.txPrebuild.version` is it possible for version to ever be `0`?

modules/core/src/v2/coins/dot.ts
```diff
@@ -221,6 +210,27 @@ export class Dot extends BaseCoin {
     return { txHex: signedTxHex };
   }
 
+  /**
+   * Explain/parse transaction
+   * @param params
+   * @param callback
+  */
+  async explainTransaction(params: ExplainTransactionOptions): Promise<accountLib.Dot.TransactionExplanation> {
+    const txHex = params.txPrebuild.txHex;
```
aaldemir, line 219: there might be an issue here if `params.txPrebuild` is undefined - maybe move this below the invalid params check?

modules/core/src/v2/coins/dot.ts
```diff
@@ -221,6 +210,27 @@ export class Dot extends BaseCoin {
...
+      throw new Error('missing explain tx parameters');
+    }
+    const sender = new accountLib.Dot.KeyPair({ pub: params.txPrebuild.key }).getAddress();
+    const factory = accountLib.register(this.getChain(), accountLib.Dot.TransactionBuilderFactory);
+    const txBuilder = factory.from(txHex);
+    txBuilder
+      .validity(params.txPrebuild.validity)
+      .referenceBlock(params.txPrebuild.referenceBlock)
+      .sender({ address: sender })
+      .version(params.txPrebuild.version);
```
argjv, line 230: Aren't these in the txHex already?

modules/core/src/v2/coins/dot.ts
nimraarif21: No it cannot be. Version is always 1 or greater.

modules/core/src/v2/coins/dot.ts
nimraarif21: No they have to be set beforehand. They cannot be decoded from the hex.

aaldemir (approved):
LGTM

modules/core/src/v2/coins/dot.ts
```diff
@@ -221,6 +210,27 @@ export class Dot extends BaseCoin {
...
+   * @param params
+   * @param callback
+  */
+  async explainTransaction(params: ExplainTransactionOptions): Promise<accountLib.Dot.TransactionExplanation> {
+    if (!params.txPrebuild.txHex || !params.txPrebuild.validity || !params.txPrebuild.key || !params.txPrebuild.referenceBlock || !params.txPrebuild.version) {
+      throw new Error('missing explain tx parameters');
+    }
+    const txHex = params.txPrebuild.txHex;
+    const sender = new accountLib.Dot.KeyPair({ pub: params.txPrebuild.key }).getAddress();
+    const factory = accountLib.register(this.getChain(), accountLib.Dot.TransactionBuilderFactory);
```
0xabrarhussain, line 224: I believe DOT also has tokens right? might need a call for `getBaseChain` when `dotTokens` is added in

PR #1776, opened 2021-12-09
feat(utxo-lib): add isSupportedScriptType(network, scriptType)

OttoAllmendinger:
Provide canonical place for this information.

BREAKING CHANGE: Removes these methods from AbstractUtxoCoin:

* `supportsP2sh()`
* `supportsP2shP2wsh()`
* `supportsP2wsh()`
* `supportsP2tr()`

Use `supportsAddressType(ScriptType2Of3)` instead.

Issue: BG-38773

### DIFF
--- a/modules/core/src/v2/coins/abstractUtxoCoin.ts
+++ b/modules/core/src/v2/coins/abstractUtxoCoin.ts
@@ -321,7 +321,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       return version;
     } catch (e) {
       // if that fails, and we aren't supporting p2wsh, then we are done and did not find a version
-      if (!this.supportsP2wsh()) {
+      if (!this.supportsAddressType('p2wsh') && !this.supportsAddressType('p2tr')) {
         return;
       }
     }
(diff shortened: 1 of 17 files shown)


modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -321,7 +321,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       return version;
     } catch (e) {
       // if that fails, and we aren't supporting p2wsh, then we are done and did not find a version
```
brandonblack, line 323: Update comment to `and we aren't supporting SegWit`?

brandonblack (changes requested):

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: Actually it's about `bech32` - there may be a coin that has segwit support but not bech32 support, in which case we should forgo the parsing attempt

Will fix. Also the address parsing should fully live in utxolib probably.

brandonblack (approved):

PR #1784, opened 2021-12-13
feat(core): add AvaxToken coins in sdk

JoaquinTissera-Simtlix:
Add AvaxcToken coins in sdk
Improve coinFactory to support repeated token contract addresses between Mainnet and Testnet.

STLX-10425

### DIFF
--- a/modules/core/src/config.ts
+++ b/modules/core/src/config.ts
@@ -6,7 +6,8 @@ import { StellarTokenConfig } from './v2/coins/stellarToken';
 import { CeloTokenConfig } from './v2/coins/celoToken';
 import { EosTokenConfig } from './v2/coins/eosToken';
 import { AlgoTokenConfig } from './v2/coins/algoToken';
-import { coins, Erc20Coin, StellarCoin, OfcCoin, CeloCoin, CoinKind, NetworkType, EosCoin, Networks, AlgoCoin } from '@bitgo/statics';
+import { AvaxcTokenConfig } from './v2/coins/avaxcToken';
+import { coins, Erc20Coin, StellarCoin, OfcCoin, CeloCoin, CoinKind, NetworkType, EosCoin, Networks, AlgoCoin, AvaxERC20Token } from '@bitgo/statics';
 
 export interface Tokens {
   bitcoin: {
@@ -28,6 +29,9 @@ export interface Tokens {
     eos: {
       tokens: EosTokenConfig[];
     };
+    avaxc: {
+      tokens: AvaxcTokenConfig[];
+    };
   };
   testnet: {
     eth: {
@@ -48,6 +52,9 @@ export interface Tokens {
     eos: {
       tokens: EosTokenConfig[];
     }
+    avaxc: {
+      tokens: AvaxcTokenConfig[];
+    };
   };
 }
 
@@ -161,6 +168,20 @@ const formattedEosTokens = coins.reduce((acc: EosTokenConfig[], coin) => {
   return acc;
 }, []);
 
+const formattedAvaxCTokens = coins.reduce((acc: AvaxcTokenConfig[], coin) => {
+  if (coin instanceof AvaxERC20Token) {
+    acc.push({
+      type: coin.name,
+      coin: coin.network.type === NetworkType.MAINNET ? 'avaxc' : 'tavaxc',
+      network: coin.network.type === NetworkType.MAINNET ? 'Mainnet' : 'Testnet',
+      name: coin.fullName,
+      tokenContractAddress: coin.contractAddress.toString().toLowerCase(),
+      decimalPlaces: coin.decimalPlaces,
+    });
+  }
+  return acc;
+}, []);
+
 export const tokens: Tokens = {
   // network name for production environments
   bitcoin: {
@@ -182,6 +203,9 @@ export const tokens: Tokens = {
     eos: {
       tokens: formattedEosTokens.filter(token => token.network === 'Mainnet'),
     },
+    avaxc: {
+      tokens: formattedAvaxCTokens.filter(token => token.network === 'Mainnet'),
+    },
   },
   // network name for test environments
   testnet: {
@@ -202,6 +226,9 @@ export const tokens: Tokens = {
     },
     eos: {
       tokens: formattedEosTokens.filter(token => token.network === 'Testnet'),
+    },
+    avaxc: {
+      tokens: formattedAvaxCTokens.filter(token => token.network === 'Testnet'),
     },
   },
 };
--- a/modules/core/src/v2/coinFactory.ts
+++ b/modules/core/src/v2/coinFactory.ts
@@ -63,30 +63,14 @@ import {
   AvaxCToken,
 } from './coins';
 import { tokens } from '../config';
-
 import * as errors from '../errors';
 import { Bcha } from './coins/bcha';
 import { Tbcha } from './coins/tbcha';
 import { Dot } from './coins/dot';
 import { Tdot } from './coins/tdot';
-import { Environments } from './environments';
+import { AvaxcTokenConfigEnvDependent } from './coins/avaxcToken';
 
 export type CoinConstructor = (bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) => BaseCoin;
-export type OptionsConstructor = {
-  isTestnetEnvDependant?: boolean;
-};
-
-/**
- * It creates a constructor according to the network env.
- * Avalanche network has the same contract address for both Mainnet and Testnet.
- * The goal is to select the right one when a particular network is used.
- */
-const envDependentCoinConstructor =
-  (coinConstructorMainNet: CoinConstructor, coinConstructorTestNet: CoinConstructor) =>
-  (bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) =>
-    Environments[bitgo.getEnv()].network !== 'testnet'
-      ? coinConstructorMainNet(bitgo, staticsCoin)
-      : coinConstructorTestNet(bitgo, staticsCoin);
 
 export class CoinFactory {
   private coinConstructors = new Map<string, CoinConstructor>();
@@ -132,15 +116,11 @@ export class CoinFactory {
     throw new errors.UnsupportedCoinError(name);
   }
 
-  public registerCoinConstructor(name: string, constructor: CoinConstructor, options: OptionsConstructor = {}): void {
+  public registerCoinConstructor(name: string, constructor: CoinConstructor): void {
     if (this.coinConstructors.has(name)) {
-      if (!options.isTestnetEnvDependant) {
-        throw new Error(`coin '${name}' is already defined`);
-      }
-      this.coinConstructors.set(name, envDependentCoinConstructor(this.coinConstructors.get(name)!, constructor));
-    } else {
-      this.coinConstructors.set(name, constructor);
+      throw new Error(`coin '${name}' is already defined`);
     }
+    this.coinConstructors.set(name, constructor);
   }
 }
 
@@ -232,17 +212,21 @@ for (const token of [...tokens.bitcoin.algo.tokens, ...tokens.testnet.algo.token
   GlobalCoinFactory.registerCoinConstructor(token.type, tokenConstructor);
 }
 
+const zipAvaxToken: Record<string, AvaxcTokenConfigEnvDependent> = {};
+
 for (const token of [...tokens.bitcoin.avaxc.tokens]) {
   const tokenConstructor = AvaxCToken.createTokenConstructor(token);
   GlobalCoinFactory.registerCoinConstructor(token.type, tokenConstructor);
-  GlobalCoinFactory.registerCoinConstructor(token.tokenContractAddress, tokenConstructor);
+  zipAvaxToken[token.tokenContractAddress] = { mainnet: token };
 }
 
 for (const token of [...tokens.testnet.avaxc.tokens]) {
   const tokenConstructor = AvaxCToken.createTokenConstructor(token);
   GlobalCoinFactory.registerCoinConstructor(token.type, tokenConstructor);
-  // this tokenContractAddress could be added as mainnet token.
-  GlobalCoinFactory.registerCoinConstructor(token.tokenContractAddress, tokenConstructor, {
-    isTestnetEnvDependant: true,
-  });
+  zipAvaxToken[token.tokenContractAddress].testnet = token;
+}
+
+for (const [tokenContractAddress, tokenConfig] of Object.entries(zipAvaxToken)) {
+  const tokenConstructor = AvaxCToken.createTokenConstructorEnvDependent(tokenConfig);
+  GlobalCoinFactory.registerCoinConstructor(tokenContractAddress, tokenConstructor);
 }
--- a/modules/core/src/v2/coinFactory.ts
+++ b/modules/core/src/v2/coinFactory.ts
@@ -60,6 +60,7 @@ import {
   OfcToken,
   Zec,
   EosToken,
+  AvaxCToken,
 } from './coins';
 import { tokens } from '../config';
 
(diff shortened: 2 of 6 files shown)


modules/core/src/v2/coinFactory.ts
```diff
@@ -68,8 +69,15 @@ import { Bcha } from './coins/bcha';
...
 import { Tdot } from './coins/tdot';
+import { Environments } from './environments';
 
 export type CoinConstructor = (bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) => BaseCoin;
+const coinConstructorDual =
+  (coinConstructorMainNet: CoinConstructor, coinConstructorTestNet: CoinConstructor) =>
+  (bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) =>
+    Environments[bitgo.getEnv()].network !== 'testnet'
+      ? coinConstructorMainNet(bitgo, staticsCoin)
+      : coinConstructorTestNet(bitgo, staticsCoin);
```
jrafael-simtlix, line 80: @shiksb @CapnMigraine We find this solution. But I don't like it. 
Could we talk about it?

modules/core/src/config.ts
```diff
@@ -161,6 +168,20 @@ const formattedEosTokens = coins.reduce((acc: EosTokenConfig[], coin) => {
   return acc;
 }, []);
 
+const formattedAvaxCTokens = coins.reduce((acc: AvaxcTokenConfig[], coin) => {
```
charly-simtlix, line 171: This construct is equivalent to `coins.filter().map()` which is perhaps more explicit
```javascript
const formattedAvaxCTokens = coins.filter(coin) => coin instanceof AvaxERC20Token).map((coin) => ({
      type: coin.name,
      coin: coin.network.type === NetworkType.MAINNET ? 'avaxc' : 'tavaxc',
      network: coin.network.type === NetworkType.MAINNET ? 'Mainnet' : 'Testnet',
      name: coin.fullName,
      tokenContractAddress: coin.contractAddress.toString().toLowerCase(),
      decimalPlaces: coin.decimalPlaces,
    });
```

modules/core/src/v2/coinFactory.ts
```diff
@@ -115,11 +123,15 @@ export class CoinFactory {
     throw new errors.UnsupportedCoinError(name);
   }
 
-  public registerCoinConstructor(name: string, constructor: CoinConstructor) {
+  public registerCoinConstructor(name: string, constructor: CoinConstructor, isDualEnable = false): void {
```
charly-simtlix, line 126: I'd rather add another method (like `registerDualCoinConstuctor`) than introducing an additional parameter

modules/core/src/v2/coinFactory.ts
```diff
@@ -68,8 +69,15 @@ import { Bcha } from './coins/bcha';
 import { Tbcha } from './coins/tbcha';
 import { Dot } from './coins/dot';
 import { Tdot } from './coins/tdot';
+import { Environments } from './environments';
 
 export type CoinConstructor = (bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) => BaseCoin;
+const coinConstructorDual =
```
charly-simtlix, line 75: The name could not help in understand the use, it is a "Environment Dependent Coin Constructor", at least a comment can help readers understand "why" it is needed

modules/core/src/v2/coinFactory.ts
```diff
@@ -210,3 +222,15 @@ for (const token of [...tokens.bitcoin.algo.tokens, ...tokens.testnet.algo.token
...
   GlobalCoinFactory.registerCoinConstructor(token.type, tokenConstructor);
 }
+
+for (const token of [...tokens.bitcoin.avaxc.tokens]) {
+  const tokenConstructor = AvaxCToken.createTokenConstructor(token);
+  GlobalCoinFactory.registerCoinConstructor(token.type, tokenConstructor);
+  GlobalCoinFactory.registerCoinConstructor(token.tokenContractAddress, tokenConstructor);
+}
+
+for (const token of [...tokens.testnet.avaxc.tokens]) {
```
charly-simtlix, line 232: duplicated code, you can refactor extract a function with common code

modules/core/src/v2/coins/avaxcToken.ts
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Flag for sending value of 0
+   * @returns {boolean} True if okay to send 0 value, false otherwise
+   */
+  valuelessTransferAllowed(): boolean {
+    return false;
+  }
+
+  /**
+   * Flag for sending data along with transactions
```
charly-simtlix, line 77: You can add in comment: "Avax does not allow to send additional data".

modules/core/src/v2/coinFactory.ts
```diff
@@ -68,8 +69,15 @@ import { Bcha } from './coins/bcha';
 import { Tbcha } from './coins/tbcha';
 import { Dot } from './coins/dot';
 import { Tdot } from './coins/tdot';
+import { Environments } from './environments';
 
 export type CoinConstructor = (bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) => BaseCoin;
+const coinConstructorDual =
+  (coinConstructorMainNet: CoinConstructor, coinConstructorTestNet: CoinConstructor) =>
+  (bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) =>
+    Environments[bitgo.getEnv()].network !== 'testnet'
```
charly-simtlix, line 78: is there any way to test this?

modules/core/src/config.ts
jrafael-simtlix: It's a good propose. but  it's copy-paste code. Do we change all of them?

modules/core/src/v2/coinFactory.ts
jrafael-simtlix: done

modules/core/src/v2/coinFactory.ts
jrafael-simtlix: done

modules/core/src/v2/coinFactory.ts
jrafael-simtlix: it's not exactly duplicated.

modules/core/src/v2/coinFactory.ts
jrafael-simtlix: unit test is added.

modules/core/src/config.ts
charly-simtlix: small improvements

modules/core/src/v2/coinFactory.ts
charly-simtlix: Yes, the list and the "accept duplicates" boolean parameter

charly-simtlix (approved):
Some nit comments that improves code readability.

modules/core/src/v2/coins/avaxcToken.ts
```diff
@@ -0,0 +1,83 @@
...
+  get decimalPlaces(): number {
+    return this.tokenConfig.decimalPlaces;
+  }
+
+  getChain(): string {
+    return this.tokenConfig.type;
+  }
+
+  getFullName(): string {
+    return 'Avaxc Token';
```
jicli, line 61: should it be `return 'Avaxc Token' + getChain();` to provide more info on what the token is?

modules/core/src/v2/coins/avaxcToken.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  get tokenContractAddress(): string {
+    return this.tokenConfig.tokenContractAddress;
+  }
+
+  get decimalPlaces(): number {
+    return this.tokenConfig.decimalPlaces;
+  }
+
+  getChain(): string {
```
jicli, line 56: the same as `type()`, why do we need two?

jrafael-simtlix: This is how it is designed.

https://github.com/BitGo/BitGoJS/blob/STLX-10425-add-avaxcToken-in-SDK/modules/core/src/v2/baseCoin.ts#L258-L265

modules/core/src/v2/coins/avaxcToken.ts
jrafael-simtlix: It could be a good idea. I didn't  go deep in reason of this static label. We only follow pattern like [stellarToken](https://github.com/BitGo/BitGoJS/blob/STLX-10425-add-avaxcToken-in-SDK/modules/core/src/v2/coins/stellarToken.ts#L80), [erc20Token](https://github.com/BitGo/BitGoJS/blob/STLX-10425-add-avaxcToken-in-SDK/modules/core/src/v2/coins/erc20Token.ts#L65), [celoToke](https://github.com/BitGo/BitGoJS/blob/STLX-10425-add-avaxcToken-in-SDK/modules/core/src/v2/coins/celoToken.ts#L65) and [algoToken](https://github.com/BitGo/BitGoJS/blob/STLX-10425-add-avaxcToken-in-SDK/modules/core/src/v2/coins/algoToken.ts#L79)

Do we analyze impact?

modules/core/src/v2/coinFactory.ts
```diff
@@ -210,3 +228,16 @@ for (const token of [...tokens.bitcoin.algo.tokens, ...tokens.testnet.algo.token
   const tokenConstructor = AlgoToken.createTokenConstructor(token);
   GlobalCoinFactory.registerCoinConstructor(token.type, tokenConstructor);
 }
+
+for (const token of [...tokens.bitcoin.avaxc.tokens]) {
```
twtaylorbitgo, line 232: what is a `bitcoin.avaxc` ? that doesn't seem great

modules/core/src/v2/coinFactory.ts
```diff
@@ -68,9 +69,22 @@ import { Bcha } from './coins/bcha';
...
 import { Tdot } from './coins/tdot';
+import { Environments } from './environments';
 
 export type CoinConstructor = (bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) => BaseCoin;
 
+/**
+ * It creates a constructor that checks network env to choose.
+ * Avalanche network has the same contract address for Mainnet and Testnet.
+ * The issue is select the right one see with network is used.
+ */
```
tylerlevine, line 80: Please take another pass at this comment. I have no idea what this is trying to say.

modules/core/src/v2/coinFactory.ts
```diff
@@ -68,9 +69,22 @@ import { Bcha } from './coins/bcha';
...
+const envDependantCoinConstructor =
```
tylerlevine, line 81: Dependent is spelled with an `e`. There is no `a`.

modules/core/src/v2/coinFactory.ts
```diff
@@ -68,9 +69,22 @@ import { Bcha } from './coins/bcha';
...
+  (coinConstructorMainNet: CoinConstructor, coinConstructorTestNet: CoinConstructor) =>
+  (bitgo: BitGo, staticsCoin?: Readonly<StaticsBaseCoin>) =>
+    Environments[bitgo.getEnv()].network !== 'testnet'
```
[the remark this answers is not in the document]
tylerlevine: This is kind of a hack but it should be ok. I think we'll run into issues if we have the same token contract on multiple different testnets though.

modules/core/src/v2/coinFactory.ts
```diff
@@ -115,11 +129,15 @@ export class CoinFactory {
     throw new errors.UnsupportedCoinError(name);
   }
 
-  public registerCoinConstructor(name: string, constructor: CoinConstructor) {
+  public registerCoinConstructor(name: string, constructor: CoinConstructor, isTestnetEnvDependant = false): void {
```
tylerlevine, line 132: spelling error here on `dependent` and also drop the `Testnet` from the option name. Also let's use an options object instead of a trailing boolean param.

modules/core/src/v2/coins/avaxcToken.ts
```diff
@@ -0,0 +1,83 @@
...
+  public readonly tokenConfig: AvaxcTokenConfig;
+
+  constructor(bitgo: BitGo, tokenConfig: AvaxcTokenConfig) {
+    const staticsCoin = tokenConfig.network === 'Mainnet' ? coins.get('avaxc') : coins.get('tavaxc');
+    super(bitgo, staticsCoin);
+    this.tokenConfig = tokenConfig;
+  }
+
+  static createTokenConstructor(config: AvaxcTokenConfig): CoinConstructor {
+    return (bitgo: BitGo) => new AvaxCToken(bitgo, config);
```
tylerlevine, line 29: Can you push the env dependent constructor logic down to here? This is really what it's for (custom instance construction dependent on bitgo object).

modules/core/src/config.ts
tylerlevine: I'm fine with this but if you do it please do it as a separate commit or separate PR entirely.

modules/core/test/v2/unit/coins/avaxcToken.ts
```diff
@@ -0,0 +1,68 @@
+import 'should';
```
tylerlevine, line 1: please add `@prettier` annotation and format

modules/core/test/browser/browser.spec.ts
```diff
@@ -18,6 +18,8 @@ describe('Coins', () => {
             'CeloToken': 1,
             'EosToken': 1,
             'AlgoToken': 1,
+            'AvaxCToken': 1,
+
```
tylerlevine, line 22: remove extra newline

tylerlevine (changes requested):

modules/core/src/v2/coinFactory.ts
jrafael-simtlix: I think that it is part of legacy code where bitcoin is naming mainnet. If you see others coins, all them has bitcoin and testnet. Here is the definition. 

https://github.com/BitGo/BitGoJS/blob/e3f3f67784f392f23cead05b62c91a11eee5709a/modules/core/src/v2/types.ts#L1

modules/core/test/browser/browser.spec.ts
jrafael-simtlix: done

modules/core/test/v2/unit/coins/avaxcToken.ts
jrafael-simtlix: done

modules/core/src/v2/coinFactory.ts
jrafael-simtlix: done

modules/core/src/v2/coinFactory.ts
jrafael-simtlix: done, now is it  understand?

tylerlevine (changes requested):
To summarize the current approach:

First, we add all the mainnet AvaxCTokens by `token.type` and `token.tokenContractAddress` to the `GlobalCoinFactory`.

Then we add the testnet AvaxCTokens by `token.type` and `token.tokenContractAddress` to the `GlobalCoinFactory` too, with the new `isEnvDependent` flag when adding by contract address.

If there's a conflict on the contract address, we replace the entry in the `GlobalCoinFactory` with the special `envDependentCoinConstructor` which takes the prior entry for the contract address in the `GlobalCoinFactory` and combines that with the entry being added.

Seems like a bit of a roundabout way to do essentially this:

zip the main net and testnet coins by contract address. (this means reduce by creating an object where the key is the contract address, and the value is an array of the token configs which have that contract address).

iterate through the zipped object, and pass the whole array of token configs to the `AvaxCToken.createTokenConstructor` function. Let it choose which of the token configs to use in the inner function which it returns.

Then you can remove all the changes to the `GlobalCoinFactory` along with the new `isEnvDependent` flag, the whole options object, and the `envDependentCoinConstructor`. We'll have a special `createTokenConstructor`implementation for `AvaxCToken`s but it lets us push the complexity into there instead of in the `GlobalCoinFactory`. I hope that makes sense, please let me know if you have any questions.

modules/core/src/v2/coins/avaxcToken.ts
jrafael-simtlix: I can find a way. I asume that this idea is to avoid changes in registerCoinConstructor.  
The doubt is how match the tokens before registering. Do I create a temporary object for it?

tylerlevine (approved):
works for me, nice job

PR #1785, opened 2021-12-13
refactor(core): use utxolib for cashaddr

OttoAllmendinger:
* Remove `bech32` format for bch (synonymous with `cashaddr`)
* Fix test vector that used mixed case instead of uppercase

BREAKING CHANGES:
* remove support for `bech32` parameter in (T)BCH `canonicalAddress()`

Issue: BG-40432


modules/utxo-lib/src/addressFormat.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  throw new Error(`could not parse outputScript [formats=${formats}]`);
+}
+
+/**
+ * Same as `toOutputScriptAndFormat`, only returning script
+ * @param address
+ * @param network
+ * @param formats
```
sangaman, line 100: I think it's worth repeating the `defaults to all supported address formats for network` comment from above.

modules/utxo-lib/src/addressFormat.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/**
+ * Attempts to parse address with different address formats, returns first hit.
+ * @param address
+ * @param network
+ * @param formats - defaults to all supported address formats for network
+ * @return tuple with [AddressFormat, Buffer] containing format and parsed output script
+ */
+export function toOutputScriptAndFormat(
```
sangaman, line 76: It seems this function is only used in tests, are we ever going to need the format with the output script or can this be dropped?

sangaman (approved):

modules/utxo-lib/src/addressFormat.ts
OttoAllmendinger: I think we might: we do have some code with format-specific handling. Not in utxolib directly though.

modules/utxo-lib/src/addressFormat.ts
OttoAllmendinger: I'll submit a follow-up

PR #1789, opened 2021-12-13
fix(core): don't require `pubs` param to `explainTransaction`

tylerlevine:
The public keys passed to `AbstractUtxoCoin`'s `explainTransaction`
function are used to count the number of valid signatures on a
transaction. However, this new param was added as a required param which
causes some existing usages of `explainTransaction` to break. Instead
this new param should be optional and if it's not provided, we'll just
be unable to count the number of signatures. The rest of the behavior of
`explainTransaction` should remain as it was prior to the breaking
change.`

Ticket: BG-40467


argjv (commented):
LGTM but could use a second pair of eyes.

argjv (approved):

PR #1791, opened 2021-12-14
fix(core): add base `explainTransaction` method

OttoAllmendinger:
The method, while optional, is part of the generic coin interface.

Unfortunately it is very difficult to have very strong types here so
we have to add a bunch of type parameters.

The `options` argument in fact is so heterogeneous between coin
implementations that it cannot be narrowed to anything meaningful
besides `any` (see implementation for EOS).

Issue: BG-40467

### DIFF
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -1,12 +1,14 @@
 /**
  * @prettier
  */
-import { BigNumber } from 'bignumber.js';
-import * as utxolib from '@bitgo/utxo-lib';
+import * as crypto from 'crypto';
 import * as bip32 from 'bip32';
+import { BigNumber } from 'bignumber.js';
+import { BaseCoin as AccountLibBasecoin } from '@bitgo/account-lib';
+import * as utxolib from '@bitgo/utxo-lib';
+
 import { BitGo } from '../bitgo';
 import { RequestTracer } from './internal/util';
-
 import { Wallet } from './wallet';
 import { Wallets } from './wallets';
 import { Markets } from './markets';
@@ -15,11 +17,10 @@ import { PendingApprovals } from './pendingApprovals';
 import { Keychain, Keychains, KeyIndices } from './keychains';
 import { Enterprises } from './enterprises';
 
-// re-export account lib transaction types
-import { BaseCoin as AccountLibBasecoin } from '@bitgo/account-lib';
-import { signMessage } from '../bip32util';
-import * as crypto from 'crypto';
 import { InitiateRecoveryOptions } from './recovery/initiate';
+import { signMessage } from '../bip32util';
+
+// re-export account lib transaction types
 export type TransactionType = AccountLibBasecoin.TransactionType;
 
 export interface TransactionRecipient {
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/baseCoin.ts
```diff
@@ -347,6 +348,14 @@ export abstract class BaseCoin {
     return signMessage(message, bip32.fromBase58(key.prv), utxolib.networks.bitcoin);
   }
 
+  /**
+   * Decompose a raw transaction into useful information.
+   * @param options - coin-specific
+   */
+  explainTransaction(options: any): Promise<TransactionExplanation<any, string | number> | undefined> {
```
sangaman, line 355: Would `Record<string, any>` make sense for `options`? It looks like that should be compatible with eos.

sangaman (approved):
The types restructuring makes sense, does this fix something that's currently broken?

OttoAllmendinger:
@sangaman this type signature would have prevented me from removing `async` from the AbstractUtxoCoin implementation. 

Because there was declaration in AbstractBaseCoin I assumed this method was not part of the public interface.

brandonblack (approved):

modules/core/src/v2/baseCoin.ts
```diff
@@ -347,6 +348,14 @@ export abstract class BaseCoin {
     return signMessage(message, bip32.fromBase58(key.prv), utxolib.networks.bitcoin);
   }
 
+  /**
+   * Decompose a raw transaction into useful information.
+   * @param options - coin-specific
+   */
+  explainTransaction(options: Record<string, any>): Promise<TransactionExplanation<any, string | number> | undefined> {
```
tylerlevine, line 355: Is it possible to make this `abstract` to force inheritors to implement? Or is that going to drag in a lot more work?

I'd rather have the concrete implementations throw `not implemented` if they so choose instead of making it the default behavior in the base class.

tylerlevine (approved):
One comment but it's not a blocker. Good clean up

modules/core/src/v2/baseCoin.ts
OttoAllmendinger: interestingly, a lot of coins do not implement this method at all

```
src/v2/coins/erc20Token.ts(22,14): error TS2515: Non-abstract class 'Erc20Token' does not implement inherited abstract member 'explainTransaction' from class 'Eth'.
src/v2/coins/eth.ts(334,14): error TS2515: Non-abstract class 'Eth' does not implement inherited abstract member 'explainTransaction' from class 'BaseCoin'.
src/v2/coins/eth2.ts(107,14): error TS2515: Non-abstract class 'Eth2' does not implement inherited abstract member 'explainTransaction' from class 'BaseCoin'.
src/v2/coins/gteth.ts(6,14): error TS2515: Non-abstract class 'Gteth' does not implement inherited abstract member 'explainTransaction' from class 'Eth'.
src/v2/coins/ofc.ts(19,14): error TS2515: Non-abstract class 'Ofc' does not implement inherited abstract member 'explainTransaction' from class 'BaseCoin'.
src/v2/coins/ofcToken.ts(29,14): error TS2515: Non-abstract class 'OfcToken' does not implement inherited abstract member 'explainTransaction' from class 'Ofc'.
src/v2/coins/susd.ts(17,14): error TS2515: Non-abstract class 'Susd' does not implement inherited abstract member 'explainTransaction' from class 'BaseCoin'.
src/v2/coins/teth.ts(6,14): error TS2515: Non-abstract class 'Teth' does not implement inherited abstract member 'explainTransaction' from class 'Eth'.
src/v2/coins/teth2.ts(5,14): error TS2515: Non-abstract class 'Teth2' does not implement inherited abstract member 'explainTransaction' from class 'Eth2'.
src/v2/coins/tsusd.ts(8,14): error TS2515: Non-abstract class 'Tsusd' does not implement inherited abstract member 'explainTransaction' from class 'Susd'.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

PR #1796, opened 2021-12-14
feat(account-lib): add utils to validate tx and block hash

nachofm:
STLX-10625


modules/account-lib/test/unit/coin/dot/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+describe('utils', () => {
+  it('should validate addresses correctly', () => {
+    should.equal(utils.isValidAddress(accounts.account1.address), true);
+    should.equal(utils.isValidAddress(accounts.account2.address), true);
+    should.equal(utils.isValidAddress(accounts.account3.address), true);
+    should.equal(utils.isValidAddress(accounts.account4.address), true);
+  });
+
+  it('should validate block hash correctly', () => {
+    should.equal(utils.isValidBlockId(blockHash.block.hash), true);
```
argjv, line 14: Can you add a couple more block hashes? Also, need error cases tests

modules/account-lib/test/unit/coin/dot/utils.ts
```diff
@@ -0,0 +1,42 @@
+import should from 'should';
+import utils from '../../../../src/coin/dot/utils';
+import { accounts, blockHash } from '../../../resources/dot';
+
+describe('utils', () => {
```
argjv, line 5: Please add a few error scenarios

modules/account-lib/src/coin/dot/utils.ts
```diff
@@ -52,7 +51,8 @@ export class Utils implements BaseUtils {
 
   /** @inheritdoc */
   isValidSignature(signature: string): boolean {
-    throw new NotImplementedError('method not implemented');
+    const signatureU8a = u8aToU8a(signature);
+    return [64, 65, 66].includes(signatureU8a.length);
```
argjv, line 55: This is pretty much what `signatureVerify()` does in the polkadot lib

raovinnakota: Yeah - does it make sense to just have it use it like this:
`const { isValid } = signatureVerify(message, signature, alice.address);`

source: https://polkadot.js.org/docs/keyring/start/sign-verify/

argjv: Oh, we have that method right below here `verifySignature(signedMessage: string, signature: string, address: string): boolean`

raovinnakota (approved):
LGTM - verifySignature nit addressed

PR #1802, opened 2021-12-16
tests(core): add tests for unsigned crossChainRecovery

OttoAllmendinger:
Add test for unsigned prebuild with `sign` call in second step

Issue: BG-39265


brandonblack:
Could the new tests be made by loading the half-signed test vectors and stripping the signatures rather than copy-pasta the files?

tylerlevine (approved):

OttoAllmendinger:
> Could the new tests be made by loading the half-signed test vectors and stripping the signatures rather than copy-pasta the files?

The test fixtures are there to verify the test output. They are not used as inputs.

PR #1804, opened 2021-12-17
feat(account-lib): add getTxHash to dot utils

lcovar:
Add a method that retrieves the txHash (id) from a signed dot transaction hex

TICKET: STLX-10869


modules/account-lib/test/unit/coin/dot/transaction.ts
```diff
@@ -54,6 +54,14 @@ describe('Dot Transaction', () => {
       tx.sender(DotResources.accounts.account2.address);
       should.deepEqual(tx.canSign({ key: DotResources.accounts.account2.secretKey }), true);
     });
+
+    it('can generate valid txHash from signed transaction', () => {
+      const signedTx = DotResources.rawTx.transfer.westendSigned2;
+      const txHash = Utils.default.getTxHash(signedTx);
```
byhow, line 60: will it throw if it cannot generate the valid txHash? what is the expected behavior here

lcovar: It accepts any string and will return a txHash. We just have to trust that as long as we pass the same txHex we broadcast on chain then we will have a matching transaction id.

byhow (approved):

modules/account-lib/test/unit/coin/dot/transaction.ts
```diff
@@ -54,6 +54,14 @@ describe('Dot Transaction', () => {
...
+      const expectedHash = '0x252e9b53c1d068c275ef4c9b5afcffb2df42859203be1305d148c0c1441a5b20';
+
+      expectedHash.should.equal(txHash);
```
tylerlevine, line 63: nit: the assertion should be the other way around (otherwise the assertion error message doesn't make much sense)
```suggestion
      txHash.should.equal(expectedHash);
```

tylerlevine (approved):

byhow (approved):

PR #1805, opened 2021-12-17
feat(account-lib): remove builder conditions

argjv:
Block number, reference block and transaction version are not required to decode the transaction.

TICKET: STLX-10869

### DIFF
--- a/modules/account-lib/src/coin/dot/transactionBuilder.ts
+++ b/modules/account-lib/src/coin/dot/transactionBuilder.ts
@@ -361,7 +361,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     });
 
     if (validationResult.error) {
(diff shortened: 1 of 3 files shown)


modules/account-lib/src/coin/dot/transactionBuilder.ts
```diff
@@ -361,7 +361,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     });
 
     if (validationResult.error) {
-      throw new InvalidTransactionError(`Transaction validation failed: ${validationResult.error.message}`);
+      // throw new InvalidTransactionError(`Transaction validation failed: ${validationResult.error.message}`);
```
tylerlevine, line 364: please remove `if` statement if there are no statements inside it

tylerlevine (changes requested):

0xabrarhussain (approved):
existing comment from tyler, otherwise lgtm

PR #1808, opened 2021-12-17
refactor(unspents): convert Dimensions to standard Typescript

OttoAllmendinger:
Replace odd mix of tcomb and Typescript with standard typescript.

BREAKING CHANGES:
* Instantiation without `new` is not supported anymore: use
  `new Dimensions(v)` instead of just `Dimensions(v)`

Issue: BG-40758


modules/unspents/src/dimensions.ts
```diff
@@ -1,287 +1,151 @@
...
-  size: number;
+/**
+ * Apply `f` to all properties of `d`
+ */
+function mapDimensions(
+  d: Partial<Dimensions>,
+  f: <T extends keyof Dimensions>(key: T, v: Dimensions[T] | undefined) => unknown
+): Dimensions {
+  return new Dimensions(
+    Object.fromEntries(Object.entries(d).map(([key, value]) => [key, f(key as keyof Dimensions, value)]))
```
sangaman, line 18: I'm a bit surprised the `as keyof Dimensions` is necessary here, I would've thought that'd be implied from the `Parital<Dimensions>` type.

Edit: Just checked this and `key` is `string` type so the cast is necessary.

modules/unspents/src/dimensions.ts
```diff
@@ -1,287 +1,151 @@
...
+
+  /* Input count for single-signature inputs (Replay Protection inputs) */
+  public nP2shP2pkInputs = 0;
+
+  public outputs: OutputDimensions = new OutputDimensions();
+
+  constructor(d: Partial<Dimensions> = {}) {
+    Object.entries(d).forEach(([key, value]) => this.setProperty(key, value));
+
+    Object.freeze(this);
```
sangaman, line 86: Does it make sense to set all the props above to `readonly` instead of use `freeze`? That way we ensure no props are changed at compile time rather than runtime I believe.

If that's done you might need to move `setProperty` inside the constructor but that's fine since we don't want to call it anywhere besides the constructor anyway.

modules/unspents/src/dimensions.ts
```diff
@@ -1,287 +1,151 @@
...
+        if (typeof v !== 'number') {
+          throw new Error(`property ${k} must be number`);
+        }
+        if (!Number.isSafeInteger(v) || v < 0) {
+          throw new Error(`property ${k} must be zero or positive integer`);
+        }
+        break;
+      case 'outputs':
+        if (!(v instanceof OutputDimensions)) {
+          v = new OutputDimensions(v as OutputDimensions);
```
sangaman, line 107: Is this reachable code? Seems like it shouldn't be possible if types are properly checked. This jumped out at me because if you change `v: unknown` to `v: number | OutputDimensions` this line gives an error.

modules/unspents/src/dimensions.ts
```diff
@@ -290,434 +154,392 @@ Object.defineProperty(Dimensions.prototype, 'nInputs', {
...
+   *        [param.assumeUnsigned] - default type for unsigned input
+   */
+  static fromInput(input: utxolib.TxInput, params: FromInputParams = {}): Dimensions {
+    const p2shInput = Dimensions.sum({ nP2shInputs: 1 });
+    const p2shP2wshInput = Dimensions.sum({ nP2shP2wshInputs: 1 });
+    const p2wshInput = Dimensions.sum({ nP2wshInputs: 1 });
+    const p2trKeypathInput = Dimensions.sum({ nP2trKeypathInputs: 1 });
+    const p2trScriptPathLevel1Input = Dimensions.sum({ nP2trScriptPathLevel1Inputs: 1 });
+    const p2trScriptPathLevel2Input = Dimensions.sum({ nP2trScriptPathLevel2Inputs: 1 });
+    const p2shP2pkInput = Dimensions.sum({ nP2shP2pkInputs: 1 });
```
sangaman, line 219: Should these be statics or constants defined outside this function? It seems a bit unnecessary to redeclare these (and go through the steps of constructing new objects in `sum`) each time `fromInput` is called.

modules/unspents/src/dimensions.ts
```diff
@@ -290,434 +154,392 @@ Object.defineProperty(Dimensions.prototype, 'nInputs', {
...
+      case 'p2wsh':
+        return Dimensions.sum({ nP2wshInputs: 1 });
+      case 'p2tr':
+        switch (params.p2trSpendType) {
+          case 'keypath':
+            return Dimensions.sum({ nP2trKeypathInputs: 1 });
+          case 'scriptpath-level1':
+            return Dimensions.sum({ nP2trScriptPathLevel1Inputs: 1 });
+          case 'scriptpath-level2':
+            return Dimensions.sum({ nP2trScriptPathLevel2Inputs: 1 });
```
sangaman, line 360: This and the `Dimensions.sum` above can also use the same constants/statics per my comment above, rather than recreating them each time `fromUnspent` is called.

modules/unspents/src/dimensions.ts
```diff
@@ -290,434 +154,392 @@ Object.defineProperty(Dimensions.prototype, 'nInputs', {
...
 
-    if (outputs.count !== nOutputs) {
-      throw new Error('deprecated partial addition: inconsistent values for "nOutputs" and "outputs.count"');
+  /**
+   * @param dimensions (can be partially defined)
+   * @return new dimensions with argument added
+   */
+  plus(dimensions: Partial<Dimensions>): Dimensions {
+    if (typeof dimensions !== 'object') {
+      throw new TypeError(`expected argument to be object`);
```
sangaman, line 405: This also seems like unreachable code assuming the types are respected.

sangaman (commented):
Very good refactoring/cleanup overall, I made a few comments/suggestions and will review again since there's a lot that's changing.

modules/unspents/src/dimensions.ts
OttoAllmendinger: It is not reachable from typescript but we still have some JS callsites. I would not add these types of guards in new codes for for now I want to retain them here.

modules/unspents/src/dimensions.ts
OttoAllmendinger: here is what happened: just changing 

```
  public nP2shInputs = 0;
```

to

```
  public readonly nP2shInputs = 0;
```

gives a build error (that I didn't look at too closely). TS infers that the type is `0` (not `number`!)

with
```

```
  public readonly nP2shInputs: number = 0;
```
it works fine

Comment (OttoAllmendinger) [reply modules/unspents/src/dimensions.ts:86]: I will keep `freeze()` for runtime safety in potential JS callsites

File: modules/unspents/src/dimensions.ts (hunk 3 above)
Comment (OttoAllmendinger) [reply modules/unspents/src/dimensions.ts:107]: this is to allow `new Dimensions({ outputs: { count: 1, size: 10 } })` etc

File: modules/unspents/src/dimensions.ts (hunk 4 above)
Comment (OttoAllmendinger) [reply modules/unspents/src/dimensions.ts:219]: yes let's clean it up while we are here

File: modules/unspents/src/dimensions.ts (hunk 5 above)
Comment (OttoAllmendinger) [reply modules/unspents/src/dimensions.ts:360]: done

File: modules/unspents/src/dimensions.ts (hunk 2 above)
Comment (sangaman) [reply modules/unspents/src/dimensions.ts:86]: Interesting. I think the approach of `:number = 0` is fine, another option is to deconstruct the parameters to the constructor and set defaults there, then assign line by line in the constructor rather than via the `setParameter` loop, as in:

```typescript
public readonly nP2shInputs: number;
constructor({ nP2shInputs = 0... }: { nP2shInputs?: number...}) {
  this.nP2shInputs = nP2shInputs;
  ..
```

However this would be more verbose so fine to leave as is, just something to consider.

modules/unspents/src/dimensions.ts
```diff
@@ -1,723 +1,531 @@
...
+          v = new OutputDimensions(v as OutputDimensions);
+        }
+        break;
+      default:
+        throw new Error(`unknown property ${k}`);
+    }
 
-  getOutputScriptLengthForChain(chain: ChainCode): number;
-  getVSizeForOutputWithScriptLength(length: number): number;
+    (this as any)[k] = v;
```
sangaman, line 114: Ah interesting so this cast to any gets around the `readonly` type restriction internally without having to put `setProperty` in the constructor.

modules/unspents/src/dimensions.ts
```diff
@@ -1,723 +1,531 @@
...
-
-  if (input.script?.length || input.witness?.length) {
-    const parsed = utxolib.bitgo.parseSignatureScript(input);
-    switch (parsed.scriptType) {
-      case undefined:
-        // unknown script type, continue with `assumeUnsigned`
-        break;
-      case 'p2shP2pk':
-        return p2shP2pkInput;
+  static readonly SingleInput = Object.freeze({
```
sangaman, line 208: :+1:

sangaman (approved):

PR #1810, opened 2021-12-17
refactor(core): rewrite CrossChainRecovery

OttoAllmendinger:
Original code was very stateful, had weak types and was unnecessarily 
verbose.

Issue: BG-39265


modules/core/src/v2/coins/utxo/recovery/crossChainRecovery.ts
```diff
@@ -41,607 +52,354 @@ class BitgoPublicApi {
...
+    outputAmount,
+    minerFee: inputAmount - outputAmount,
+    spendAmount: outputAmount,
+    inputs,
+    unspents: inputs,
+    outputs,
+    externalOutputs: outputs,
+    changeOutputs: [],
+    payGoFee: 0,
+  } as TransactionInfo;
```
sangaman, line 347: Why is this cast necessary? I'm also a bit confused as to why there are so many of these fields aren't defined within the `TransactionInfo` interface.

modules/core/src/v2/coins/abstractUtxoCoin.ts
```diff
@@ -93,17 +93,23 @@ export interface TransactionExplanation extends BaseTransactionExplanation<strin
   signatures: number;
 }
 
+export interface TransactionInfo {
+  txHexes?: Record<string, string>;
```
sangaman, line 97: A comment here might be good to explain what the tx hexes are being mapped to. It looks like that's txIds based on the code below altho I'm also a bit confused why there multiple txIds for one transaction - these would be transactions from other chains potentially I'm guessing?

modules/core/src/v2/coins/utxo/recovery/crossChainRecovery.ts
```diff
@@ -41,607 +52,354 @@ class BitgoPublicApi {
...
+  const fee = inputAmount - outputAmount;
+  return {
+    size: vsize,
+    fee,
+    feeRate: fee / vsize,
+    payGoFee: 0,
+  };
+}
 
+export async function recoverCrossChain(
```
sangaman, line 363: A comment block would be good here too explaining what this function does since it's exported.

sangaman (approved):
I'm not so familiar with this code/process so I can't really compare it to the old logic but the new code makes sense and I have no objections. I just left a few comments for clarification.

modules/core/src/v2/coins/utxo/recovery/crossChainRecovery.ts
OttoAllmendinger: I kept these fields to stay backwards compatible. I don't think they are required by any methods these days but they may be needed by legacy consumers of this. Will add a comment.

modules/core/src/v2/coins/utxo/recovery/crossChainRecovery.ts
OttoAllmendinger: will do

modules/core/src/v2/coins/abstractUtxoCoin.ts
OttoAllmendinger: will add comment

sangaman (approved):

PR #1811, opened 2021-12-17
feat(statics): CoinFeature Custody

bitgoAaron:
- all coins except XTZ support custody

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -118,6 +118,10 @@ export enum CoinFeature {
    * This coin requires that accounts keep a minimum balance as reserve
    */
   REQUIRES_RESERVE = 'requires-reserve',
+  /**
+   * This coin supports custodial wallet types
+   */
+  CUSTODY = 'custody',
 }
 
 /**
--- a/modules/statics/test/unit/coins.ts
+++ b/modules/statics/test/unit/coins.ts
@@ -1,5 +1,5 @@
 import 'should';
-import { BaseNetwork, CoinFamily, coins, Erc20Coin, EthereumNetwork, NetworkType } from '../../src';
+import { BaseNetwork, CoinFamily, CoinFeature, coins, Erc20Coin, EthereumNetwork, NetworkType } from '../../src';
 
 interface DuplicateCoinObject {
   name: string;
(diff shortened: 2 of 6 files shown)


modules/statics/src/base.ts
```diff
@@ -118,6 +118,10 @@ export enum CoinFeature {
    * This coin requires that accounts keep a minimum balance as reserve
    */
   REQUIRES_RESERVE = 'requires-reserve',
+  /**
+   * This coin supports custodial wallet types
+   */
+  CUSTODY = 'custody',
```
tylerlevine, line 124: this doesn't really fit here as I discussed earlier, but it doesn't fit in the same way that the `PAYGO` entry doesn't fit, so I guess I'll let it slide if it makes your life easier.

We should be wary of adding platform feature flags here - that's not the intended purpose.

tylerlevine (approved):

modules/statics/src/base.ts
bitgoAaron: @FloBitGo do you still feel we should put this in statics instead of hardcoded in the UI(s) ?

modules/statics/test/unit/coins.ts
```diff
@@ -1,29 +1,39 @@
...
+
+    if (coin.family === CoinFamily.XTZ) {
+      it('does not support custody', () => {
+        coin.features.includes(CoinFeature.CUSTODY).should.eql(false);
+      });
+    } else {
+      it('does support custody', () => {
+        coin.features.includes(CoinFeature.CUSTODY).should.eql(true);
+      });
+    }
```
OttoAllmendinger, line 29:
```
it('has expected custody support', function () {

coin.features.includes(CoinFeature.CUSTODY).should.eql(coin.family !== CoinFamily.XTZ)

});
```

OttoAllmendinger (approved):
nit

modules/statics/src/base.ts
FloBitGo: I thought the idea behind statics was that it would be our central coin configuration. I wasn't aware that it should be agnostic to platform. If that is the case, then we should probably discuss different solutions.

OttoAllmendinger: The idea of `statics` is to avoid unnecessary duplication between BitGoJS and `platform`. So if both `platform` and `BitGoJS` need to know about this flag it belongs in here.

bitgoAaron: yes, both platform & all the UIs will need to know about this flag so this makes the most sense then to define it here.

PR #1814, opened 2021-12-17
chore(account-lib,core): refactor keyDeriver to account-lib

joey-bai:
moves keyDeriver from core to account-lib.
plan is to leverage this code to implement address derivation for
Ed25519KeyPair

Ticket: STLX-10000


rliberal-simtlix (approved):

tylerlevine (changes requested):
Looks fine but I think it's missing a lockfile update

joey-bai:
> Looks fine but I think it's missing a lockfile update

@tylerlevine doesn't look like package-lock is being tracked for core and account lib.
https://github.com/BitGo/BitGoJS/tree/master/modules/account-lib
https://github.com/BitGo/BitGoJS/tree/master/modules/core

tylerlevine (approved):
No package lock updates needed

PR #1816, opened 2021-12-20
fix(core): correctly pass `pubs`

OttoAllmendinger:
The `params` are not actually passed to `signTransaction()`

Issue: BG-39265

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -1768,20 +1768,20 @@ export class Wallet {
    * @return {*}
    */
   async signTransaction(params: WalletSignTransactionOptions = {}): Promise<SignedTransaction> {
-    const txPrebuild = params.txPrebuild;
+    const { txPrebuild } = params;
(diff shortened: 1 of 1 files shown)


sangaman (approved):
Fix makes sense, before `pubs` was just getting assigned to `params` and then ignored.

modules/core/src/v2/wallet.ts
```diff
@@ -1768,20 +1768,20 @@ export class Wallet {
    * @return {*}
    */
   async signTransaction(params: WalletSignTransactionOptions = {}): Promise<SignedTransaction> {
-    const txPrebuild = params.txPrebuild;
+    const { txPrebuild } = params;
     if (!txPrebuild || typeof txPrebuild !== 'object') {
       throw new Error('txPrebuild must be an object');
     }
     const presign = await this.baseCoin.presignTransaction(params);
```
OttoAllmendinger, line 1775: @sangaman actually, due to a "lucky accident", `presign === params` here and the `params.pubs` assignment later on also modifies `presign`

still we should prefer the new variant :)

PR #1819, opened 2021-12-21
fix(account-lib): add token sub tx types

charly-simtlix:
Added EnableAsset and DisableAsset to TransactionType

Issues: STLX-9929

### DIFF
--- a/modules/account-lib/src/coin/algo/assetTransferBuilder.ts
+++ b/modules/account-lib/src/coin/algo/assetTransferBuilder.ts
@@ -5,7 +5,7 @@ import { InvalidTransactionError } from '../baseCoin/errors';
 import { TransactionType } from '../baseCoin';
 import { TransferBuilder } from './transferBuilder';
 import { Transaction } from './transaction';
-import { AssetTransferTxnSchema, AssetToggleTxnSchema } from './txnSchema';
+import { AssetToggleTxnSchema, AssetTransferTxnSchema } from './txnSchema';
 import Utils from './utils';
 
 export class AssetTransferBuilder extends TransferBuilder {
@@ -71,8 +71,39 @@ export class AssetTransferBuilder extends TransferBuilder {
     );
   }
 
+  /**
+   * Computes the specific (sub) type for this Asset Transaction in algorand
+   * https://developer.algorand.org/docs/get-details/transactions/transactions/#asset-transfer-transaction
+   * @protected
+   */
   protected get transactionType(): TransactionType {
+    if (this.isOptOutAssetTxn()) {
+      return TransactionType.DisableAsset;
+    }
+
+    if (this.isOptInAssetTrx()) {
+      return TransactionType.EnableAsset;
+    }
+
     return TransactionType.Send;
+  }
+
+  /**
+   * Check if this transaction to opt-in to an asset (allowlist)
+   * @see https://developer.algorand.org/articles/algos-asas/
+   * @private
+   */
+  private isOptInAssetTrx() {
+    return this._tokenId && !this.isOptOutAssetTxn() && this._amount === 0 && this._to === this._sender;
+  }
+
+  /**
+   * Check if this transaction to opt-out to an asset (do not allowlist)
+   * @see https://developer.algorand.org/articles/algos-asas/
+   * @private
+   */
+  private isOptOutAssetTxn() {
+    return this._tokenId && this._closeRemainderTo;
   }
 
   /** @inheritdoc */
--- a/modules/account-lib/src/coin/baseCoin/enum.ts
+++ b/modules/account-lib/src/coin/baseCoin/enum.ts
@@ -29,6 +29,10 @@ export enum TransactionType {
   StakingWithdraw,
   // Handle smart contract calls
   ContractCall,
+  // Wallet owner opt-in to receive this asset in his wallet
+  EnableAsset,
+  // Wallet owner opt-out to receive this asset in his wallet
+  DisableAsset,
 }
 
 /**


JulioAchaval-SimTLiX (commented):
Do we need to difference between native send and tokens send?

JulioAchaval-SimTLiX (dismissed):

modules/core/src/v2/coins/algo.ts
```diff
@@ -228,104 +236,113 @@ export class Algo extends BaseCoin {
...
-        explanationResult.tokenId = txJson.tokenId;
-      }
+    switch (tx.type) {
+      case TransactionType.Send:
+      case TransactionType.EnableAsset:
+      case TransactionType.DisableAsset:
+        return this.explainSendTransaction(txJson);
+      case TransactionType.WalletInitialization:
+        return this.explainWalletInitializationException(txJson);
+    }
```
ofornaroli, line 249: you could avoid this switch with a simple if statement

```
if (tx.type === TransactionType.WalletInitialization) {
    return this.explainWalletInitializationException(txJson);
} else {
    return this.explainSendTransaction(txJson);
}
```

modules/account-lib/src/coin/baseCoin/enum.ts
```diff
@@ -29,6 +29,10 @@ export enum TransactionType {
   StakingWithdraw,
   // Handle smart contract calls
   ContractCall,
+  // Wallet owner opt-in to receive this asset in his wallet
+  EnableAsset,
```
ofornaroli, line 33: you can't disable an asset, you can disable a token
```suggestion
  EnableToken,
```

modules/account-lib/src/coin/baseCoin/enum.ts
```diff
@@ -29,6 +29,10 @@ export enum TransactionType {
...
+  // Wallet owner opt-out to receive this asset in his wallet
+  DisableAsset,
```
ofornaroli, line 35: you can't disable an asset, you can disable a token
```suggestion
  DisableToken,
```

modules/account-lib/src/coin/algo/transaction.ts
```diff
@@ -210,14 +206,22 @@ export class Transaction extends BaseTransaction {
...
     }
-    if (this.type === TransactionType.Send) {
+
+    if (this._algoTransaction.to) {
       result.to = algosdk.encodeAddress(this._algoTransaction.to.publicKey);
+    }
+
+    if (this._algoTransaction.amount) {
       result.amount = this._algoTransaction.amount.toString();
     }
```
ofornaroli, line 223: you could do something like this instead the if statement, if you like.
result.amount = this._algoTransaction.amount?.toString();

modules/account-lib/src/coin/algo/transaction.ts
```diff
@@ -210,14 +206,22 @@ export class Transaction extends BaseTransaction {
...
     };
+
     if (this._algoTransaction.closeRemainderTo) {
       result.closeRemainderTo = algosdk.encodeAddress(this._algoTransaction.closeRemainderTo.publicKey);
     }
-    if (this.type === TransactionType.Send) {
+
+    if (this._algoTransaction.to) {
       result.to = algosdk.encodeAddress(this._algoTransaction.to.publicKey);
+    }
```
ofornaroli, line 219: you could do something like this instead the if statement, if you like.
result.to = algosdk.encodeAddress(this._algoTransaction.to?.publicKey);

modules/account-lib/src/coin/baseCoin/enum.ts
```diff
@@ -29,6 +29,10 @@ export enum TransactionType {
   StakingWithdraw,
   // Handle smart contract calls
   ContractCall,
+  // Wallet owner opt-in to receive this asset in his wallet
+  EnableAsset,
+  // Wallet owner opt-out to receive this asset in his wallet
```
ofornaroli, line 34:
```suggestion
  // Wallet owner opt-out to stop receiving this asset in his wallet
```

modules/account-lib/test/unit/coin/algo/transaction.ts
```diff
@@ -88,7 +88,7 @@ describe('Algo Transaction', () => {
       txInfo.genesisID.should.equal('testnet-v1.0');
       txInfo.genesisHash.should.equal('SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=');
       txInfo.should.have.property('txType');
-      txInfo.txType.should.equal('enableToken');
+      txInfo.txType.should.equal('EnableAsset');
```
ofornaroli, line 91:
```suggestion
      txInfo.txType.should.equal('EnableToken');
```

modules/account-lib/test/unit/coin/algo/transaction.ts
```diff
@@ -114,7 +114,7 @@ describe('Algo Transaction', () => {
       txInfo.genesisID.should.equal('testnet-v1.0');
       txInfo.genesisHash.should.equal('SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=');
       txInfo.should.have.property('txType');
-      txInfo.txType.should.equal('disableToken');
+      txInfo.txType.should.equal('DisableAsset');
```
ofornaroli, line 117:
```suggestion
      txInfo.txType.should.equal('DisableToken');
```

modules/account-lib/test/unit/coin/algo/transactionBuilder/assetTransferBuilder.ts
```diff
@@ -93,6 +95,8 @@ describe('Algo Asset Transfer Transaction Builder', () => {
 
       should.exists(txHex);
 
+      txInfo.type.should.equal('axfer');
+      txInfo.txType.should.equal('DisableAsset');
```
ofornaroli, line 99:
```suggestion
      txInfo.txType.should.equal('DisableToken);
```

modules/account-lib/test/unit/coin/algo/transactionBuilder/assetTransferBuilder.ts
```diff
@@ -158,6 +162,8 @@ describe('Algo Asset Transfer Transaction Builder', () => {
 
       should.exists(txHex);
 
+      txInfo.type.should.equal('axfer');
+      txInfo.txType.should.equal('EnableAsset');
```
ofornaroli, line 166:
```suggestion
      txInfo.txType.should.equal('EnableToken);
```

modules/account-lib/test/unit/coin/algo/transactionBuilder/assetTransferBuilder.ts
```diff
@@ -194,6 +200,8 @@ describe('Algo Asset Transfer Transaction Builder', () => {
...
```
ofornaroli, line 204:
```suggestion
      txInfo.txType.should.equal('EnableToken');
```

modules/account-lib/test/unit/coin/algo/transactionBuilder/assetTransferBuilder.ts
```diff
@@ -232,6 +240,8 @@ describe('Algo Asset Transfer Transaction Builder', () => {
 
       should.exists(txHex);
 
+      txInfo.type.should.equal('axfer');
+      txInfo.txType.should.equal('DisableAsset');
```
ofornaroli, line 244:
```suggestion
      txInfo.txType.should.equal('DisableToken');
```

ofornaroli (changes requested):
In general I see a good implementation, however I found some words that could land in a bad interpretation and I made some recomendations.

modules/account-lib/src/coin/baseCoin/enum.ts
JulioAchaval-SimTLiX: We choose that word (Asset) because this type will be used by other coins. And it maybe use other types.

ofornaroli: Isn't `token` a generic name for non native assets? if not, you should find other name wich represents non native assets, because each coin is an asset and you can't enable or disable native assets

charly-simtlix: In Algorand the token operations are called "Assets Transactions". They use the term ASA (Algorand Standard Asset). See https://developer.algorand.org/docs/get-details/asa/

I choose the "Asset" instead of "Token" because other networks can require to  opt-in to receive other types of assets (including tokens, native coins, or NFTs).

What do you think?

modules/account-lib/src/coin/algo/transaction.ts
charly-simtlix: Your expression will set the undefined value to "result.amount". The 'if' will not add the field to 'result' structure at all. They similar but not totally equivalent. I'm unsure on the better way in this case.

modules/account-lib/src/coin/baseCoin/enum.ts
charly-simtlix: "opt-out to receive" is equivalent to "choose to stop receiving".

modules/core/src/v2/coins/algo.ts
charly-simtlix: They differ on the "default" case. The if creates a Send transaction while the switch is explicit on the handled cases. For a not handled type, it returns nothing (that was the original behavior).
But it should be better to be explicit on this behaviour

modules/account-lib/src/coin/baseCoin/enum.ts
charly-simtlix: I was looking on the blockchain terminology and it is something ambiguous the use. But in general terms they use the concept "Asset" in a broader sense than blockchain, and the concept "token" as a representation of an (external) asset inside the blockchain.
I used "Asset" because Algorand use the term "Asset" (instead of token) in the context of ASAs (Algorand Native Assets). 
But if BitGo terminology is more inclined to "Token", if perhaps is a better name.

modules/core/src/v2/coins/algo.ts
charly-simtlix: improved

modules/account-lib/src/coin/baseCoin/enum.ts
charly-simtlix: changed

modules/account-lib/test/unit/coin/algo/transaction.ts
charly-simtlix: changed

modules/account-lib/test/unit/coin/algo/transactionBuilder/assetTransferBuilder.ts
charly-simtlix: changed

modules/account-lib/test/unit/coin/algo/transactionBuilder/assetTransferBuilder.ts
charly-simtlix: changed

modules/account-lib/test/unit/coin/algo/transactionBuilder/assetTransferBuilder.ts
charly-simtlix: changed

modules/account-lib/test/unit/coin/algo/transactionBuilder/assetTransferBuilder.ts
charly-simtlix: changed

jrafael-simtlix (dismissed):
LGFM. I gave this PR a cursory check, I little know of Algo.

ofornaroli (dismissed):
LGTM

modules/account-lib/src/coin/algo/assetTransferBuilder.ts
```diff
@@ -71,10 +71,41 @@ export class AssetTransferBuilder extends TransferBuilder {
...
+    return this._tokenId && !this.isOptOutAssetTxn() && BigInt(this._amount) === BigInt(0) && this._to === this._sender;
+  }
+
+  /**
+   * Check if this transaction to opt-out to an asset (do not allowlist)
+   * @see https://developer.algorand.org/articles/algos-asas/
+   * @private
+   */
+  private isOptOutAssetTxn() {
+    return this._tokenId && this._closeRemainderTo;
```
rliberal-simtlix, line 106: Here also amount should be zero?

modules/account-lib/src/coin/baseCoin/enum.ts
```diff
@@ -29,6 +29,10 @@ export enum TransactionType {
   StakingWithdraw,
   // Handle smart contract calls
   ContractCall,
+  // Wallet owner opt-in to receive this token in his wallet
```
rliberal-simtlix, line 32: I think here will be good to provide a more generic documentation not only the valid case for algo.

modules/core/src/v2/coins/algo.ts
```diff
@@ -453,7 +472,32 @@ export class Algo extends BaseCoin {
...
+    return accountLib.Algo.algoUtils.publicKeyToAlgoAddress(pubkey);
+  }
+
+  /**
+   * Return (sub) type of transaction handling backward compatibility with
+   * transactions created before new TransactionTypes were added
+   * @param tx the transaction
+   * @private
+   */
+  private static getCompatibleTxType(tx: TxData) {
```
rliberal-simtlix, line 485: This is great :+1:

pancham88 (commented):
Please test this brach with OVC and wallet platform, there is a fair chance that the changes may be breaking something in OVC and/or wallet platform.

modules/account-lib/src/coin/algo/assetTransferBuilder.ts
charly-simtlix: it is not required to be zero, it seems you can send some amount to a destination account, send the remaining to the closeRemainderTo, and "close" this asset in the sender address.
https://developer.algorand.org/docs/get-details/transactions/transactions/#asset-transfer-transaction

modules/account-lib/src/coin/algo/transaction.ts
```diff
@@ -26,17 +26,13 @@ export class Transaction extends BaseTransaction {
...
-    if (this._numberOfRequiredSigners === 1) {
-      const kp = new KeyPair({ prv: key });
-      const addr = kp.getAddress();
-      if (addr === this._sender) {
-        return true;
-      } else {
-        return false;
-      }
-    } else {
+    if (this._numberOfRequiredSigners !== 1) {
```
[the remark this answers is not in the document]
charly-simtlix: It reduce the nesting of the lines with computation. I replaced the IFs with a switch to be more explicit on intention.

modules/account-lib/src/coin/baseCoin/enum.ts
charly-simtlix: hope, it is better now

PR #1822, opened 2021-12-22
feat(sol): implement derive address function

joey-bai:
performs SLIP-0010 hardened derivation using an existing key's private key as the
seed.

Ticket: STLX-11014


modules/account-lib/src/coin/sol/keyPair.ts
```diff
@@ -67,4 +68,19 @@ export class KeyPair extends Ed25519KeyPair {
...
     return keys.pub;
   }
+
+  /**
+   * Derives a hardened child key pair using this key pair's secret key
+   * as the seed.
+   *
+   * @param path derivation path
+   */
+  deriveHardened(path: string): KeyPair {
```
argjv, line 78: If this is using the raw seed then we can put it in the ed25519 base class, no?

joey-bai: good point, you're right. i moved to ed25519 key pair and return DefaultKeys instead of the actual key pair.

rliberal-simtlix (approved):

PR #1826, opened 2021-12-23
feat(account-lib): add key factory for coins

byhow:
Ticket: STLX-10821

### DIFF
--- a/modules/account-lib/src/keyPair/index.ts
+++ b/modules/account-lib/src/keyPair/index.ts
@@ -1,0 +1,15 @@
+import * as coinModules from '..';
+import { BaseKeyPair } from '../coin/baseCoin';
+import { KeyPairOptions } from '../coin/baseCoin/iface';
+
+export function register(coinName: string, source?: KeyPairOptions): BaseKeyPair {
+  const sanitizedCoinName = coinName.trim().toLowerCase();
+  const key = Object.keys(coinModules)
+    .filter((k) => coinModules[k].KeyPair)
+    // TODO(BG-40990): eth2 BLS keypair init error
+    .find((k) => k.trim().toLowerCase() !== 'eth2' && k.trim().toLowerCase() === sanitizedCoinName);
+  if (key) {
+    return new coinModules[key].KeyPair(source);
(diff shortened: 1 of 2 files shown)


modules/account-lib/src/keyPair/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export function register(coinName: string, source?: KeyPairOptions): BaseKeyPair {
+  const sanitizedCoinName = coinName.trim().toLowerCase();
+  const key = Object.keys(coinModules)
+    .filter((k) => coinModules[k].KeyPair)
+    // TODO(BG-40990): eth2 BLS keypair init error
+    .find((k) => k.trim().toLowerCase() !== 'eth2' && k.trim().toLowerCase() === sanitizedCoinName);
+  if (key) {
+    return new coinModules[key].KeyPair(source);
+  }
+  return {} as BaseKeyPair;
```
jrafael-simtlix, line 14: If a coin is not find, do we return empty object?
How does user handle this case?
Also there are no test case about this behavior.

0xabrarhussain (changes requested):
this change seems pretty wonky. this makes the convention assumption that everything in the folder above only has coin modules with the same structure

which sure you can have that convention, but it wont' be enforced

byhow:
@0xabrarhussain The structure is not enforced in that we will implement the interface gradually if such structure is needed. For an initial factory method that gets the corresponding `key` implementation of each coin, we are not looking to refactor all of the previous coins' export structure.

For the exports module, it is checking if the structure exported the baseKey class implementation, so it does not need to follow the same structure, but only need to export that specific `key` module.

That being said, we are open to better ways to group the `key` class implementations without refactoring the old codebase for sure.

byhow:
ref on the additional unit tests audits: https://github.com/BitGo/BitGoJS/pull/1832

byhow:
@0xabrarhussain Do you mind we move this conversation over to #1832? And then we can close this in favor of the other PR

byhow:
closing in favor of the other PR

PR #1828, opened 2021-12-23
feat(utxo-lib): add support for Zcash version 5 "NU5"

OttoAllmendinger:
Add optional support for new transaction format.

While it is possible to continue using the v4 transaction format, having
full support for v5 is useful for our integration test suite.

Support is limited to transactions that only have transparent inputs and
outputs.

References:

* https://zips.z.cash/zip-0252
* https://github.com/zcash-hackworks/zcash-test-vectors/blob/dd8fdb/zip_0244.py

Issue: BG-37908


modules/utxo-lib/src/bitgo/transaction.ts
```diff
@@ -32,23 +32,47 @@ export function createTransactionFromHex(hex: string, network: Network): UtxoTra
...
-export function setTransactionBuilderDefaults(txb: UtxoTransactionBuilder, network: Network): void {
+export function getDefaultTransactionVersion(network: Network): number {
   switch (getMainnet(network)) {
     case networks.bitcoincash:
     case networks.bitcoinsv:
     case networks.bitcoingold:
-      txb.setVersion(2);
+      return 2;
+    case networks.zcash:
+      return 4;
```
OttoAllmendinger, line 42: we will change this to `return isTestnet(network) ? 5 : 4` in a follow-up

tylerlevine (approved):
minor nits, feel free to handle in follow up

modules/utxo-lib/src/bitgo/transaction.ts
OttoAllmendinger: make that `return isTestnet(network) ? 450 : 400` with  https://github.com/BitGo/BitGoJS/pull/1843

PR #1836, opened 2022-01-03
feat(account-lib): add anonymous proxy txn builder STLX-11137

aaldemir:
This PR is meant to add a transaction builder for Polkadot to account-lib. The type of transaction builder is to create anonymous proxy addresses (which internally will be used for receive addresses). This builder is similar to `addProxy` except a `owner` is not defined for anonymous proxies.


modules/account-lib/src/coin/dot/anonymousAddressInitializationBuilder.ts
```diff
@@ -0,0 +1,136 @@
...
+import BigNumber from 'bignumber.js';
+import { InvalidTransactionError } from '../baseCoin/errors';
+import { TransactionBuilder } from './transactionBuilder';
+import { Transaction } from './transaction';
+import { DecodedSignedTx, DecodedSigningPayload, UnsignedTransaction } from '@substrate/txwrapper-core';
+import { TransactionType } from '../baseCoin';
+import { AddAnonymousProxyArgs, MethodNames, ProxyType } from './iface';
+import { AnonymousAddressInitializationSchema } from './txnSchema';
+
+export class AnonymousAddressInitializationBuilder extends TransactionBuilder {
```
byhow, line 12: Is this an extension on the `AddressInitializationBuilder` ? If so, we can extend on that builder as well as the data interfaces

byhow (commented):
Do we need to add this new builder to the factory?

modules/account-lib/src/coin/baseCoin/enum.ts
```diff
@@ -29,6 +29,8 @@ export enum TransactionType {
   StakingWithdraw,
   // Handle smart contract calls
   ContractCall,
+  // Spawn address that does not have own key pair
+  AnonymousAddressInitialization,
```
argjv, line 33: I think this is what `AddressInitialization` represents.

modules/account-lib/src/coin/dot/anonymousAddressInitializationBuilder.ts
argjv: We can probably modify `AddressInitializationBuilder` to do this by default. We don't really use the proxy addresses it builds as it is implemented right now.

aaldemir: I added the anonymous proxy fields as a part of the origin AddressInitialization builder. This means that both addProxy and anonymous proxy transactions would use this builder - the only difference is for `addProxy`, the `owner()` function needs to be called to pass the proxy address in, and for `anonymous` the `index()` can be called to set the index (but will be defaulted to 0). Both transactions will need `delay()` and `type()`

lcovar (approved):

PR #1837, opened 2022-01-04
fix(statics): update avaxc tokens names to not use dot

shiksb:
Since we store token names as mongo keys, and
mongo keys cannot contain . and $ characters,
use & instead of . in usdt.e and usdc.e tokens

Ticket: STLX-11681

### DIFF
--- a/modules/account-lib/test/unit/coin/avaxc/transferBuilder.ts
+++ b/modules/account-lib/test/unit/coin/avaxc/transferBuilder.ts
@@ -15,15 +15,15 @@ const tokensNames = [
   'avaxc:KLO',
   'avaxc:JOE',
   'avaxc:QI',
-  'avaxc:USDT.E',
-  'avaxc:USDC.E',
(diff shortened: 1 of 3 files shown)


tylerlevine (commented):
will these be properly escaped when using the token name in a url?

shiksb:
> will these be properly escaped when using the token name in a url?

@tylerlevine I believe so. We will use the same mechanisms we do for other symbols inside token names such as `-` and `:`

tylerlevine:
> > will these be properly escaped when using the token name in a url?
> 
> @tylerlevine I believe so. We will use the same mechanisms we do for other symbols inside token names such as `-` and `:`

Those tokens don't have a special meaning in urls, but ampersands do. Please verify that these new coin names are properly url encoded

shiksb:
Avalanche foundation has backtracked from `USDC.e` and `USDT.e` and will now use `USDC` and `USDT`. We don't need this change any more.

PR #1838, opened 2022-01-04
fix(eth): allow for ENS resolution in WP to change recipient address for

rlutzbitgo:
eth

Ticket: BG-41173


rlutzbitgo:
Some more details here, the Eth team is adding support for ENS resolution for sends. In WP, if the recipient address is not an eth address, we attempt to resolve it to an address via ENS and change the recipient address accordingly. Doing so causes this validation check to fail, so added something to say, "only do this check if the txPrebuild recipient address is an eth address"

wesgraham (approved):

PR #1849, opened 2022-01-10
fix(statics): update zcash explorer url

lzjluzijie:
Ticket: BG-41455


sangaman (approved):
So Chain. Much wow.

New URLs work for me.

tylerlevine (approved):
looks good to me, nice work 🥇

PR #1850, opened 2022-01-11
fix(account-lib): add a check for undefined in isValidPublicKey for sol

alebusse:
Addded a check to validate that the incoming address is not undefined in the isValidPublicKey method
of solana utils     
Added unit tests

STLX-11994

### DIFF
--- a/modules/account-lib/src/coin/sol/utils.ts
+++ b/modules/account-lib/src/coin/sol/utils.ts
@@ -54,6 +54,7 @@ export function isValidPrivateKey(prvKey: string | Uint8Array): boolean {
 
 /** @inheritdoc */
 export function isValidPublicKey(pubKey: string): boolean {
+  if (!pubKey) return false;
   if (Crypto.isValidXpub(pubKey)) return true;
   try {
     new PublicKey(pubKey);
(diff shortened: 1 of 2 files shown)


rliberal-simtlix (approved):

ipelaez-simtlix (approved):

joey-bai (approved):

modules/account-lib/src/coin/sol/utils.ts
```diff
@@ -54,6 +54,7 @@ export function isValidPrivateKey(prvKey: string | Uint8Array): boolean {
 
 /** @inheritdoc */
 export function isValidPublicKey(pubKey: string): boolean {
```
tylerlevine, line 56: Param should be optional if it can be undefined. Otherwise callers should be fixed as they should get a compile error if they attempt to pass a value which could be undefined

tylerlevine (changes requested):

modules/account-lib/src/coin/sol/utils.ts
alebusse: The caller (in this case is the IMS and is in JS) 
PR for reference https://github.com/BitGo/indexerdb-microservice/pull/510
@noel-bitgo

alebusse: I makes no sense to use optional when its the value we want to validate in an isValid method

alebusse:
Closing this as we gonna fix it in the IMS. Thanks Tyler

PR #1852, opened 2022-01-12
STLX-11856 - implement a field for transaction material

aseoane-simtlix:
dot transactionBuilder had static materials. This feature will allow to dynamicaly pass new materials to the builder when new releases are made (preventing having to constantly update the static files). 
Several constructors were changed and methods were removed.

Several tests had hardcoded materials in them, which were replaced a variable filled with the static materials.
transactionBuilderFactory had several look-alike tests that where collapsed into a foreach run

JIRA ticket: [STLX-11856](https://bitgoinc.atlassian.net/browse/STLX-11856)


jrafael-simtlix (dismissed):

modules/account-lib/test/unit/coin/dot/transactionBuilder/addressInitializationBuilder.ts
```diff
@@ -3,17 +3,20 @@ import sinon, { assert } from 'sinon';
...
+  let materialData: Material;
 
   const sender = DotResources.accounts.account1;
   const receiver = DotResources.accounts.account3;
 
   beforeEach(() => {
     const config = buildTestConfig();
-    builder = new AddressInitializationBuilder(config);
+    materialData = utils.getMaterial(config);
+    builder = new AddressInitializationBuilder(config, materialData);
```
argjv, line 19: Instead of passing it in the constructor, we should pass it with a method to stick with the builder pattern. If it is mandatory, we should validate it was set before we call build() and sign().

Also, do we have default values out these change too often?

aseoane-simtlix: Method was added in the factory class, then sent in the constructor because it's going through its super() method.

Regarding validation, it should not be needed as it is first set to the static values, and then overwritten in case the dynamic values are used.

Didn't quite understand the default values question. If you're asking if there are default values, the answer would be yes, inside statics. This PR is one half of a change supposed to prevent often mandatory updates of said static values, making it dynamic. The other half is being worked on in wallet-platform, which would use an already existing rpc call to fill the data. Does this make sense? Did this answer your question?

lcovar: Yea,  im not too fond of passing it in the constructor either. 

What if instead we open up these fields for modification and use the default static values otherwise. For example something like

```
builder = new AddressInitializationBuilder(config);
builder.specVersion(<specVersion>).genesisHash(<genesisHash)...build()
```
?

argjv: What Luis said, or even a `.metadata(responseFromNode)` would work better than passing it in the constructor

lcovar (changes requested):
hold off on merge until we decide how we want to consume the metadata

aseoane-simtlix:
Closed and redeployed on PR [1890](https://github.com/BitGo/BitGoJS/pull/1890)

PR #1859, opened 2022-01-13
fix(account-lib): polkadot/api dependency is removed

jrafael-simtlix:
@polkadot/api has multiple versions, ensure that there is only one installed.
Either remove and explicitly install matching versions or dedupe using your package manager.
The following conflicting packages were found:
	cjs 7.3.1 	/home/node/srv/node_modules/@substrate/txwrapper-core/node_modules/@polkadot/api
	    6.12.1	/home/node/srv/node_modules/@polkadot/api
	    
[STLX-12053](https://bitgoinc.atlassian.net/browse/STLX-12053)


jrafael-simtlix:
Fix in https://github.com/BitGo/BitGoJS/pull/1865

PR #1862, opened 2022-01-13
chore(utxo-lib): default to VERSION4_BRANCH_NU5 for zcashTest

OttoAllmendinger:
According to https://zips.z.cash/zip-0252 the testnet activation was
at block https://sochain.com/block/ZECTEST/1599200 (Oct 6, 2021)

Issue: BG-37908


sangaman (dismissed):
Is this expected to fix our testnet zcash woes?

modules/utxo-lib/src/bitgo/transaction.ts
```diff
@@ -1,6 +1,6 @@
 import { TxOutput } from 'bitcoinjs-lib';
 
-import { networks, Network, getMainnet } from '../networks';
+import {networks, Network, getMainnet, isMainnet} from '../networks';
```
tylerlevine, line 3: Formatting

tylerlevine (approved):
Looks fine but needs a prettier run and a rebase

sangaman (approved):

PR #1863, opened 2022-01-13
fix(account-lib): remove proxy type from constants STLX-12064

aaldemir:
the proxy type of “Governance” is no longer supported by the substrate package used for transaction building - removing the type from account-lib will help remove testing for the type which could cause failures.

JIRA ticket: https://bitgoinc.atlassian.net/browse/STLX-12064


raovinnakota (approved):
LGTM - All tests pass + governance removed

PR #1867, opened 2022-01-14
feat(core): support signing single sig dot transactions

lcovar:
TICKET: STLX-11874

PR #1874, opened 2022-01-19
chore: update polkadot spec version to 9150

lcovar:
CLOSES TICKET: STLX-12242


modules/account-lib/test/unit/coin/dot/transactionBuilder/transactionBuilderFactory.ts
```diff
@@ -88,10 +88,10 @@ describe('dot Transaction Builder Factory', () => {
     const builder = factory.from(rawTx.proxy.signed);
     should(builder).instanceOf(TransferBuilder);
     builder
-      .validity({ firstValid: 3933 })
+      .validity({ firstValid: 3933, maxDuration: 64 })
```
byhow, line 91: What's the reason behind adding the `maxDuration` param here? It is because of the updates in #1867?

byhow (approved):
Do you mind quickly summarizing the intuition behind the static updates and possibly include some reference of how the test data got updated in the description? Other than that, code changes look good

aaldemir (approved):

lcovar:
@byhow 
Sure. The reason we needed to update statics was because the spec version changed during a runtime upgrade on the node and that change is causing all transactions to fail since they're built with the version hardcoded in statics. 

The way we run these tests needs to be refactored since it's a PITA to have to update the version and then update every single hardcoded txHex. The reason we need to update is because the old txHex are generated with the old specVersion and when that's updated the txHex hash changes so tests are trying to generate a new transaction from the old txHex with the same values and checking if the txHex stays the same after it's rebuilt. Since the specVersion chnges the txHex wont match anymore so new ones have to be generated. The way I do this is for example if I want to update the txHex of the unstake hardcoded value i run a test that builds a signed unstake transaction like this one https://github.com/BitGo/BitGoJS/pull/1874/files#diff-49fb65582995d698456f0525af14d4c9b7a9c502d2e0c5e4cf6b59e11c2d5c66R40 and print out the `tx.toBroadcastFormat()` so I can grab the new txHex that has the right specVersion. I then replace it the one found here https://github.com/BitGo/BitGoJS/pull/1874/files#diff-cb20bd78216ba1baaa599c15536cc6d3dd765e1c5e92a2fff0f9a7d6331dfd8bR73.

We mostly have to do this for the test found under `transactionBuilderFactory.ts` since these test are the ones comparing that the txHex stays the same. Im not sure if this test is needed in the first place, it should probably just test that it returns an instance of the expected builder.

If you have any suggestions on how to improve these tests, do let us know so we can avoid this time consuming works wheneve we need to udpate statics.

modules/account-lib/test/unit/coin/dot/transactionBuilder/transactionBuilderFactory.ts
lcovar: I generated the new proxy txHex using this test here https://github.com/BitGo/BitGoJS/blob/77d800ca23e845fdc481e35790524c958c9f52a9/modules/account-lib/test/unit/coin/dot/transactionBuilder/transferBuilder.ts#L223

This test was failing since the txHex wasnt matching and it was because I had to add `maxDuration: 64`  since the other transaction was being built like this.

PR #1880, opened 2022-01-20
fix(core): add flush threshold example

jamalchowdhury-bitgo:
Added JavaScript and TypeScript example to set flush threshold for a token.
Ticket: [BG-41448](https://bitgoinc.atlassian.net/browse/BG-41448)


wesgraham (dismissed):
Super clean - thank you!

wesgraham (approved):

PR #1882, opened 2022-01-20
chore(core): upgrade typescript 4.5.5

bitgoAaron:
- upgrades to same version as bgms
bg-41926

### DIFF
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -39,7 +39,7 @@
     "@bitgo/blake2b": "^3.0.2",
     "@bitgo/bls": "^0.1.0",
     "@bitgo/bls-dkg": "^1.0.1",
-    "@bitgo/statics": "^6.16.0",
+    "@bitgo/statics": "6.16.0",
     "@celo/contractkit": "^1.2.4",
     "@ethereumjs/common": "^2.4.0",
     "@ethereumjs/tx": "^3.3.0",
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -39,7 +39,7 @@
     "@bitgo/blake2b": "^3.0.2",
     "@bitgo/bls": "^0.1.0",
     "@bitgo/bls-dkg": "^1.0.1",
-    "@bitgo/statics": "6.16.0",
+    "@bitgo/statics": "^6.16.0",
     "@celo/contractkit": "^1.2.4",
     "@ethereumjs/common": "^2.4.0",
     "@ethereumjs/tx": "^3.3.0",
@@ -81,9 +81,6 @@
     "stellar-sdk": "^10.0.1",
     "tronweb": "^3.2.6",
     "tweetnacl": "^1.0.3"
-  },
-  "resolutions": {
-    "typescript": "4.5.5"
   },
   "devDependencies": {
     "@types/bs58": "^4.0.1",
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -20,7 +20,7 @@
     "lint": "eslint --quiet 'src/**/*.ts' 'test/**/*.ts'",
     "lint-fix": "eslint --fix 'src/**/*.ts' 'test/**/*.ts'",
     "prepublishOnly": "npm run compile",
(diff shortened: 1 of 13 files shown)


modules/account-lib/package.json
```diff
@@ -39,7 +39,7 @@
     "@bitgo/blake2b": "^3.0.2",
     "@bitgo/bls": "^0.1.0",
     "@bitgo/bls-dkg": "^1.0.1",
-    "@bitgo/statics": "^6.16.0",
+    "@bitgo/statics": "6.16.0",
```
tylerlevine, line 42: Is there a reason the carat was removed here? This will get re-added when doing a release

modules/account-lib/package.json
```diff
@@ -82,6 +82,9 @@
     "tronweb": "^3.2.6",
     "tweetnacl": "^1.0.3"
   },
+  "resolutions": {
+    "typescript": "4.5.5"
+  },
```
tylerlevine, line 87: how come we need a special resolution here?

yarn.lock
```diff
@@ -14382,20 +14382,20 @@ typeforce@^1.11.3, typeforce@^1.11.5:
...
 typescript@^4.5.4:
-  version "4.5.5"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.5.5.tgz#d8c953832d28924a9e3d37c73d729c846c5896f3"
-  integrity sha512-TCTIul70LyWe6IJWT8QSYeA54WQe8EjQFU4wY52Fasj5UKx88LNYKCgBEHcOMOrFF1rKGbD8v/xcNWVUq9SymA==
-
-typescript@~4.2.4:
-  version "4.2.4"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.2.4.tgz#8610b59747de028fda898a8aef0e103f156d0961"
-  integrity sha512-V+evlYHZnQkaz8TRBuxTA92yZBPotr5H+WhQ7bD3hZUndx5tGOa1fuCgeSjxAzM1RiN5IzvadIXTVefuuwZCRg==
+  version "4.6.2"
```
tylerlevine, line 14396: can we remove the explicit `typescript` dep from our internal packages? It should only exist in the root package.json:

```
$ yarn why typescript
yarn why v1.22.17
[1/4] 🤔  Why do we have the module "typescript"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "typescript@4.5.5"
info Has been hoisted to "typescript"
info Reasons this module exists
   - "workspace-aggregator-846fbfd9-bfaa-4869-ba14-497b73770ea7" depends on it
   - Specified in "devDependencies"
   - Hoisted from "_project_#typescript"
info Disk size without dependencies: "61.75MB"
info Disk size with unique dependencies: "61.75MB"
info Disk size with transitive dependencies: "61.75MB"
info Number of shared dependencies: 0
=> Found "@bitgo/blockapis#typescript@4.6.2"
info This module exists because "_project_#@bitgo#blockapis" depends on it.
info Disk size without dependencies: "62.45MB"
info Disk size with unique dependencies: "62.45MB"
info Disk size with transitive dependencies: "62.45MB"
info Number of shared dependencies: 0
=> Found "@bitgo/utxo-bin#typescript@4.6.2"
info This module exists because "_project_#@bitgo#utxo-bin" depends on it.
info Disk size without dependencies: "62.45MB"
info Disk size with unique dependencies: "62.45MB"
info Disk size with transitive dependencies: "62.45MB"
info Number of shared dependencies: 0
=> Found "bls12377js#typescript@3.9.10"
info This module exists because "_project_#@bitgo#account-lib#@celo#contractkit#@celo#utils#bls12377js" depends on it.
info Disk size without dependencies: "52.33MB"
info Disk size with unique dependencies: "52.33MB"
info Disk size with transitive dependencies: "52.33MB"
info Number of shared dependencies: 0
✨  Done in 1.06s.
```

We'll have to do something else about the really old version coming in from `@celo/contractkit` separately

tylerlevine (changes requested):
nice work, just a few comments

modules/account-lib/package.json
bitgoAaron: there's a lint step in bgms drone that requires all dependencies within the build system use the same exact version. I thought the `^` here was a typo and the exact version was preferred, but we can make the umbrella packages in bgms use the carat instead if the release is going to put it in there every time anyways.

PR #1887, opened 2022-01-20
feat: include feature flag for external signing API

DariusParvin:
includes #BG-38025

continuing efforts from #1857

## Changes:

- added three flags to bitgo express: 
    - `--externalMode` - Flag setting 'generator' or 'signer' value, which enables Express to run a mode that either only generates or signs transactions, for use with an external API.
    - `--externalSignerUrl`  - URL specifying the external API to call for signing. Required when externalMode is set to 'signer.'
    - `--signerFileSystemPath` - Local path specifying where an Express signer machine keeps encrypted user private keys. Required when externalMode is set to 'signer.'
- `setupRoutes` in clientRoutes.ts will set up `setupGeneratorRoutes` or `setupSigningRoutes` if running in external signing mode. 

## Test:

- unit/config.ts  updated with the new flags to ensure it correctly handles config precedence for a complete config
- unit/bitgoExpress.ts updated to check that the 'ExternalMode' information is logged when BitGo Express starts up


modules/express/src/clientRoutes.ts
```diff
@@ -352,6 +352,15 @@ function handleCanonicalAddress(req: express.Request) {
   return (coin as Coin.Bch | Coin.Bsv | Coin.Ltc).canonicalAddress(address, version || fallbackVersion);
 }
 
+function handleV1Sign(req: express.Request) {
+  console.log('v1 sign was called');
+  return 'v1_sign_transaction';
+}
```
tylerlevine, line 358: let's change these to just throw an error for now instead
```suggestion
  throw new Error('not yet implemented');
}
```

modules/express/src/clientRoutes.ts
```diff
@@ -352,6 +352,15 @@ function handleCanonicalAddress(req: express.Request) {
...
+function handleV2Sign(req: express.Request) {
+  console.log('v2 sign was called');
```
tylerlevine, line 360: same here, let's throw an error for now

modules/express/test/unit/bitgoExpress.ts
```diff
@@ -203,6 +203,25 @@ describe('Bitgo Express', function () {
       logStub.restore();
     });
 
+    it('should output external mode upon server startup', () => {
```
tylerlevine, line 206: could use a test to make sure we only call `setupAPIRoutes()` when running in regular mode (and that `setupSigningRoutes()` is not called), and another test for the inverse case

tylerlevine (changes requested):
nice work so far - just a few small things to add before merging this. Let's chat on next steps

tylerlevine:
Also please don't forget to mark the pull request as "ready for review" using the button at the bottom once you're ready to get some 👀  on your work. 👍

DariusParvin:
Thanks @tylerlevine! I just saw your review now (I need to change my notifications so I see these earlier). I added a commit which is a WIP of what should probably be another PR. I'll separate the latest commit and address your feedback here first.

modules/express/src/clientRoutes.ts
DariusParvin: done

modules/express/src/clientRoutes.ts
DariusParvin: done

tylerlevine (approved):

PR #1889, opened 2022-01-21
feat(account-lib): add withdraw staking builder

ipelaez-simtlix:
add withdraw staking builder in order to be able to get rewards after deactivate, add from() method

[stlx-12191](https://bitgoinc.atlassian.net/browse/STLX-12191)


modules/account-lib/src/coin/sol/utils.ts
```diff
@@ -204,26 +231,14 @@ export function requiresAllSignatures(signatures: SignaturePubkeyPair[]): boolea
...
+  if (matchTransactionTypeByInstructionsOrder(instructions, walletInitInstructionIndexes)) {
     return TransactionType.WalletInitialization;
-  } else if (
-    getInstructionType(instructions[stakingActivateInstructionsIndexes.Create]) === ValidInstructionTypesEnum.Create &&
-    getInstructionType(instructions[stakingActivateInstructionsIndexes.Initialize]) ===
-      ValidInstructionTypesEnum.StakingInitialize &&
-    getInstructionType(instructions[stakingActivateInstructionsIndexes.Delegate]) ===
-      ValidInstructionTypesEnum.StakingDelegate
-  ) {
+  } else if (matchTransactionTypeByInstructionsOrder(instructions, stakingActivateInstructionsIndexes)) {
```
pcardo-simtlix, line 236: As `if` content ends in a return, `else` is not required. The same logic applies to l238 and l240

ipelaez-simtlix: nit, I take it into account since I probably change this fn after talk with Rodrigo

joey-bai (approved):

PR #1890, opened 2022-01-21
feat(account-lib): implement a field for transaction material

aseoane-simtlix:
Redeploy from [PR 1852](https://github.com/BitGo/BitGoJS/pull/1852)

dot transactionBuilder had static materials. This feature will allow to dynamicaly pass new materials to the builder when new releases are made (preventing having to constantly update the static files). 
Several constructors were changed and methods were removed.

Several tests had hardcoded materials in them, which were replaced a variable filled with the static materials.
transactionBuilderFactory had several look-alike tests that where collapsed into a foreach run

JIRA ticket: [STLX-11856](https://bitgoinc.atlassian.net/browse/STLX-11856)


lcovar:
Thinking about this a bit more and the fields that will change the most are fields such as txVersion and specVersion.

I should be able to do something like:
`builder.material({specVersion: 9151, txVersion: 10})......builder()` and have those fields updated only while keeping the existing values for the other fields rather than always having to provide every single field in the Material interface. 

This is the response from /transactoin/metadata
```
{
  "at": {
    "hash": "string",
    "height": "string"
  },
  "genesisHash": "string",
  "chainName": "string",
  "specName": "string",
  "specVersion": "string",
  "txVersion": "string",
  "metadata": "string"
}
```

but the fields we extract are only these
```
{
  "at": {
    "hash": "string",
    "height": "string"
  },
 "specVersion" : "string"
  "txVersion": "string",
}
```


  material(material: Material): this {
    this._material = material;
    return this;
  }

modules/core/test/v2/fixtures/coins/dot.ts
```diff
@@ -28,8 +28,7 @@ export const accounts = {
 
 export const rawTx = {
   transfer: {
-    unsigned: '0xa40403009f7b0675db59d19b4bd9c8c72eaabba75a9863d02b30115b8b3c3ca5c20f02540700e40b5402d501210300a223000007000000e143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e149799bc9602cb5cf201f3425fb8d253b2d4e61fc119dcab3249f307f594754d',
-    signed:
-      '0x4502840061b18c6dc02ddcabdeac56cb4f21a971cc41cc97640f6f85b073480008c53a0d0045351a73a7aa6ef4f3618078897c17858fbee42f7e5a18849e87d62a3eeb913e1b0ea184bc6159a37fd2aa233993661e04be20845de9a00d2f7f343dd4183601dbf52103000403009f7b0675db59d19b4bd9c8c72eaabba75a9863d02b30115b8b3c3ca5c20f02540700e40b5402',
+    unsigned: '0xa80403009f7b0675db59d19b4bd9c8c72eaabba75a9863d02b30115b8b3c3ca5c20f02540bfadb9bbae251d501210300aa23000007000000e143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e149799bc9602cb5cf201f3425fb8d253b2d4e61fc119dcab3249f307f594754d',
+    signed: '0x4902840061b18c6dc02ddcabdeac56cb4f21a971cc41cc97640f6f85b073480008c53a0d00b6d868a11d202b56df1959f5d5f81f44ce1f95c8e70424b17080ea869d1c39d453f16c38fbef600a636c9a62a49ede5ee695a1822faf2f94fcfbb184a4254009d5012103000403009f7b0675db59d19b4bd9c8c72eaabba75a9863d02b30115b8b3c3ca5c20f02540bfadb9bbae251',
```
byhow, line 32: Will need to update this signed TX hash with the new default Validity window: 5d7239ca82e492d5a7bd6a85a35fa3394e5a7d8b

lcovar:
Concerns have been addressed with @aseoane-simtlix offline. There are still WP changes to be made.

We made some recent changes to validity window that was merged, please rebase and pull those changes in so that unit tests dont fail after this merge.

lcovar (approved):

PR #1896, opened 2022-01-23
feat(core): generate receive addresses on client

joey-bai:
during wallet creation, we'll generate a secret that's used solely for
address derivation. secret will be encrypted with wallet's passphrase.

currently, only supports hardened derivation, so will need to provide
passphrase when deriving new addresses.

Ticket: BG-00000

### DIFF
--- a/modules/core/src/v2/wallets.ts
+++ b/modules/core/src/v2/wallets.ts
@@ -46,6 +46,10 @@ export interface GenerateWalletOptions {
   };
   coldDerivationSeed?: string;
   rootPrivateKey?: string;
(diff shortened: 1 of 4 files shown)


modules/core/src/v2/wallets.ts
```diff
@@ -455,6 +459,20 @@ export class Wallets {
       walletParams.rootPrivateKey = params.rootPrivateKey;
     }
 
+    // TODO: only generate if coin supports?
+    if (canEncrypt) {
```
joey-bai, line 463: `canEncrpyt && !cold`

PR #1902, opened 2022-01-24
feat(account-lib): add unnominate for dot staking

yangran0:
Ticket: STLX-12072


modules/account-lib/test/unit/coin/dot/transactionBuilder/unnominateBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+        .fee({ amount: 0, type: 'tip' })
+        .version(8);
+      builder.sign({ key: sender.secretKey });
+      const tx = await builder.build();
+      const txJson = tx.toJson();
+      should.deepEqual(txJson.sender, sender.address);
+      should.deepEqual(txJson.blockNumber, 3933);
+      should.deepEqual(txJson.referenceBlock, '0x149799bc9602cb5cf201f3425fb8d253b2d4e61fc119dcab3249f307f594754d');
+      should.deepEqual(txJson.genesisHash, '0xe143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e');
+      should.deepEqual(txJson.specVersion, 9150);
```
aaldemir, line 32: I think we should start initializing a variable for `specVersion` at the start of the unit tests for each builder. We've ha to go back and manually change each specVersion every time there is a node update, so it would make things easier just to change 1 value per builder's test set. Same goes for genesisHash and referenceBlock I think - maybe we should use variables instead of hard coding them into the unit tests.

yangran0: I set variables for those 3 values like you suggested.

modules/account-lib/src/coin/dot/iface.ts
```diff
@@ -12,6 +12,7 @@ export enum MethodNames {
   TransferKeepAlive = 'transferKeepAlive',
   Unbond = 'unbond',
   Anonymous = 'anonymous',
+  Chill = 'chill',
```
aaldemir, line 15: what do you think about using `unnominate` as the MethodName so that it's easy to understand it's used to select the `UnnominateBuilder`? If you think chill is appropriate then thats fine with me as well.

modules/account-lib/test/unit/coin/dot/transactionBuilder/unnominateBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+        .sender({ address: sender.address })
+        .validity({ firstValid: 3933, maxDuration: 64 })
+        .referenceBlock(refBlock)
+        .sequenceId({ name: 'Nonce', keyword: 'nonce', value: 200 })
+        .fee({ amount: 0, type: 'tip' })
+        .version(8);
+      builder.sign({ key: sender.secretKey });
+      const tx = await builder.build();
+      const txJson = tx.toJson();
+      should.deepEqual(txJson.sender, sender.address);
```
aaldemir, line 31: are there any properties on the tx or txJson that can verify that a `chill` txn was built?
you may need to add properties from the tx to txJson by adding a check here https://github.com/BitGo/BitGoJS/blob/c259b9d815da67c7e21cda59b53412417d27e3ee/modules/account-lib/src/coin/dot/transaction.ts#L100

aaldemir (commented):
thanks for addressing previous comment. a couple more ideas - apologies for not catching it earlier.

modules/account-lib/src/coin/dot/iface.ts
yangran0: I wasn't too sure about the naming conventions so I followed how stake/unstake did it. For stakingBuilder the method was just bond, and for unstakeBuilder the method is unbond. In this case I used chill to follow the previous pattern.

modules/account-lib/test/unit/coin/dot/transactionBuilder/unnominateBuilder.ts
yangran0: For chill, there are no arguments that need to be passed in so there really isn't any chill specific properties. I'm not quite sure how else to verify it is a chill transaction.

aaldemir: Have you checked what the `tx` object looks like? But yeah I see what you're saying, if there are no args then I doubt they'd show up in the `tx` object. It might be worth double checking tho. It may hold a method name or something `chill` specific.

yangran0: I double checked, the only thing unique is the method name, and this should be tested in 'should build from raw unsigned/signed tx' tests because it calls validateDecodedTransaction which checks method name.

aaldemir: thats true. alrighty - looks like this needs a rebase to resolve conflicts. everything else looks good to me. nice work!

aaldemir (approved):
LGTM + all comments/questions addressed

PR #1903, opened 2022-01-24
feat(account-lib): add batch txn builder for DOT

aaldemir:
Enable batching of multiple DOT transaction calls into a single transaction with the primary intent of allowing multiple receive address creation in a single txn. This will help clients save on fees and avoid making 1 request per receive address creation (each of which would require a signed transaction broadcast).

JIRA ticket: https://bitgoinc.atlassian.net/browse/STLX-12465


lcovar (approved):
nice work

PR #1905, opened 2022-01-25
feat: add module `@bitgo/blockapis`

OttoAllmendinger:
The package contains lightweight abstractions around the Blockchair API and
the Blockstream API.

The code is largely based on the implementation that previously lived in
`core/.../utxo/recovery/`.

Issue: BG-41897


modules/core/src/v2/coins/utxo/recovery/insightApi.ts
```diff
@@ -1,90 +0,0 @@
...
-      case 'tltc':
-        // FIXME: defunct
-        return 'https://insight.litecore-test.io/api';
-      case 'zec':
-        return 'https://zcashnetwork.info/api';
-      case 'tzec':
-        // FIXME: defunct
-        return 'https://explorer.testnet.z.cash/api';
-    }
-  }
```
OttoAllmendinger, line 53: mainnet networks are covered by blockchair API

modules/blockapis/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@bitgo/blockapis",
+  "description": "Access public block explorer APIs for a variety of coins",
```
OttoAllmendinger, line 3: This is not quite true yet: the package defines the interfaces `AddressApi`, `TransactionApi` and `UtxoApi`. 

* The `BlockchairApi` class implements all three interfaces for all utxo mainnet networks plus bitcoin testnet.
* The `BlockstreamApi` class implements all three interfaces for btc mainnet and testnet

However it should be quite simple to implement the `TransactionApi` and `AddressApi` for `BlockchairApi` for other coins as well.

tylerlevine (approved):
awesome, this will be useful in a lot of places 👍

PR #1909, opened 2022-01-25
test(account-lib): add specific explainTransatction

facundoCarreSimtlix:
Add specific explainTransatction() tests related to Activate, Deactivate and Withdraw staking
builders

STLX-12452


modules/account-lib/test/unit/coin/sol/transaction.ts
```diff
@@ -542,5 +545,138 @@ describe('Sol Transaction', () => {
...
+        outputs: [
+          {
+            address: '7dRuGFbU2y2kijP6o1LYNzVyz4yf13MooqoionCzv5Za',
+            amount: '10000',
+          },
+        ],
+        fee: {
+          fee: '10000',
+          feeRate: 5000,
+        },
```
joey-bai, line 588: @alebusse do you think this could be related to the bug we're seeing with transfer amounts?
our `explainTransaction` impl doesn't add `fee` to `outputAmount`.

joey-bai (approved):

PR #1912, opened 2022-01-26
chore(account-lib): add defined types for dot

byhow:
Some of the types in dot is still `any`. Adding defined types to avoid loose types.

Ticket: BG-00000


raovinnakota (approved):
LGTM - all tests pass

lcovar (approved):

PR #1914, opened 2022-01-26
fix(statics): apply prettier to full project

OttoAllmendinger:
Bring IDE config in sync with `yarn` target

Issue: BG-41897


tylerlevine (approved):
excellent clean up, thanks

PR #1916, opened 2022-01-26
fix(core): fix verityTransaction for sol

alebusse:
Fixed verityTransaction for sol so it only compares amount and address from recipients and output
objects    
Added unit tests

STLX-12684

### DIFF
--- a/modules/core/test/v2/unit/coins/sol.ts
+++ b/modules/core/test/v2/unit/coins/sol.ts
@@ -12,6 +12,7 @@ describe('SOL:', function () {
   let newTxPrebuild;
   let newTxParams;
   let newTxParamsWithError;
+  let newTxParamsWithExtraData;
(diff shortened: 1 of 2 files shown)


rliberal-simtlix (approved):

modules/core/test/v2/unit/coins/sol.ts
```diff
@@ -63,6 +64,16 @@ describe('SOL:', function () {
       },
     ],
   };
+  const txParamsWithExtraData = {
+    txPrebuild,
+    recipients: [
+      {
+        address: 'CP5Dpaa42RtJmMuKqCQsLwma5Yh3knuvKsYDFX85F41S',
+        amount: '300000',
+        data: undefined,
```
joey-bai, line 73: should we have this be a random string?

joey-bai (approved):

PR #1918, opened 2022-01-26
feat(core): update createAddress to perform hardened derivation

alebusse:
updated createAddress to perform hardened derivation and added unit test
updated generateWallets and added unit test

[STLX-12543](https://bitgoinc.atlassian.net/browse/STLX-12543)

### DIFF
--- a/modules/core/src/v2/baseCoin.ts
+++ b/modules/core/src/v2/baseCoin.ts
@@ -117,6 +117,12 @@ export interface SupplementGenerateWalletOptions {
   disableKRSEmail?: boolean;
 }
 
+export interface DeriveKeypairOptions {
+  addressDerivationPrv?: string;
+  addressDerivationPub: string;
+  index: number;
+}
+
 export interface FeeEstimateOptions {
   numBlocks?: number;
   hop?: boolean;
@@ -524,4 +530,26 @@ export abstract class BaseCoin {
    * Sign a transaction
    */
   abstract signTransaction(params: SignTransactionOptions): Promise<SignedTransaction>;
+
+  /**
+   * Determines if this coin supports address derivation from a single sig key pair.
+   * If true, a separate key pair will be generated and encrypted in the User's key chain during wallet creation.
+   * This key pair will later be used to create new addresses.
+   * @returns {boolean}
+   */
+  supportsDerivationKeypair(): boolean {
+    return false;
+  }
+
+  /**
+   * Derives a new KeyPair from the user's derivation key pair
+   *
+   * @param {DeriveKeypairOptions} params
+   * @param {String} params.addressDerivationPrv - private key
+   * @param {Number} params.index - index
+   * @returns {KeyPair|undefined} - a derived Keypair or undefined
+   */
+  deriveKeypair(params: DeriveKeypairOptions): KeyPair | undefined {
+    return undefined;
+  }
 }
--- a/modules/core/src/v2/coins/sol.ts
+++ b/modules/core/src/v2/coins/sol.ts
@@ -23,6 +23,7 @@ import {
 import { BitGo } from '../../bitgo';
 import { Memo } from '../wallet';
 import * as _ from 'lodash';
+import { DeriveKeypairOptions } from '..';
 
 export interface TransactionFee {
   fee: string;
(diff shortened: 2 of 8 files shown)


modules/core/src/v2/wallet.ts
```diff
@@ -1359,8 +1366,27 @@ export class Wallet {
     // get keychains for address verification
     const keychains = await Promise.all(this._wallet.keys.map((k) => this.baseCoin.keychains().get({ id: k, reqId })));
     const rootAddress = _.get(this._wallet, 'receiveAddress.address');
+    const addressDerivationKeypair = _.get(keychains[KeyIndices.USER], 'addressDerivationKeypair');
```
alebusse, line 1369: @joey-bai  this will always exist, right?

joey-bai: right

modules/core/src/v2/wallet.ts
```diff
@@ -1359,8 +1366,27 @@ export class Wallet {
...
+    const lastChainIndex = _.get(this.coinSpecific(), `lastChainIndex.${chain || 0}`, -1);
```
joey-bai, line 1370: lets default this to `0` instead of `-1` since Address created with Wallet will have index 0.

modules/core/src/v2/wallet.ts
joey-bai: this may not exist since we only create it if `this.baseCoin.supportsDerivationKeypair() === true`.

modules/core/src/v2/baseCoin.ts
```diff
@@ -115,6 +115,15 @@ export interface SupplementGenerateWalletOptions {
...
+  addressDerivationKeypair?: {
+    encryptedPrv: string;
+    pub: string;
+  };
+}
+
+export interface DeriveKeypairOptions {
+  addressDerivationPrv: string;
+  index: number;
 }
```
joey-bai, line 127: lets make `prv` an optional string and add required string `pub`.

for SOL, we can only do hardened derivation right now, but can change in the future. (Sol's impl can throw error if prv is not present)
other coins may not need to rely on hardened derivation.

modules/core/src/v2/baseCoin.ts
```diff
@@ -115,6 +115,16 @@ export interface SupplementGenerateWalletOptions {
   };
   rootPrivateKey?: string;
   disableKRSEmail?: boolean;
+  addressDerivationKeypair?: {
+    encryptedPrv: string;
```
joey-bai, line 119: maybe make this optional too so we're ready to switch to unhardened derivation

joey-bai (approved):

modules/core/src/v2/baseCoin.ts
alebusse: I just noticed this is not used at all here, since we needed this to in the create user key part, it has nothing to do with the wallet. https://github.com/BitGo/BitGoJS/pull/1918/files#diff-5ac4cbb306f6d9e468e73cde27b685ff575ee08f06fd884ed2463ca58b7cb622R408

modules/core/src/v2/baseCoin.ts
```diff
@@ -524,4 +530,22 @@ export abstract class BaseCoin {
    * Sign a transaction
    */
   abstract signTransaction(params: SignTransactionOptions): Promise<SignedTransaction>;
+
+  /** Flag for support of creation of a derivation Keypair
```
joey-bai, line 534: update for more clarity:
```
Determines if this coin supports address derivation from a single sig key pair.

If true, a separate key pair will be generated and encrypted in the User's key chain during wallet creation.
This key pair will later be used to create new addresses.
```

modules/core/src/v2/baseCoin.ts
```diff
@@ -524,4 +530,22 @@ export abstract class BaseCoin {
...
+   * @returns {boolean}
+   */
+  supportsDerivationKeypair(): boolean {
+    return false;
+  }
+
+  /** Derives a new KeyPair from the passed addressDerivationPrv and the index
+   *
```
joey-bai, line 542:
```
Derives a new KeyPair from the user's derivation key pair
```

modules/core/src/v2/coins/sol.ts
```diff
@@ -288,4 +289,35 @@ export class Sol extends BaseCoin {
...
+        throw new Error('Key derivation failed - missing derived base58 prv key');
+      }
+
+      return {
+        prv: derivedBase58KeyPair.prv,
+        pub: derivedBase58KeyPair.pub,
+      };
+    } catch (e) {
+      return undefined;
+    }
```
joey-bai, line 321: no need to catch imo, let's surface the error back to the user

joey-bai (approved):

joey-bai:
merging as is
@tylerlevine don't hesitate to reach out if you have any comments.

PR #1919, opened 2022-01-26
feat(account-lib): migrate BLS key pair from @bitgo/bls to @bitgo/bls-dkg lib

fperalta-simtlix:
With bls-dkg lib it is possible to verify a signature of a meessage signed by 2 of the 3 key pairs

BREAKING CHANGE: BlsKeyPair is not just a default prv and pub object, instead it has an array of
secretShares and a publicShare which should be merged with the other BLS key pairs to get the common
public key and the private key of each keypair.

BG-35989

### DIFF
--- a/modules/account-lib/src/coin/baseCoin/iface.ts
+++ b/modules/account-lib/src/coin/baseCoin/iface.ts
@@ -26,7 +26,17 @@ export type Seed = {
   seed: Buffer;
 };
 
-export type KeyPairOptions = Seed | PrivateKey | PublicKey;
+/**
+ * The number of participants for which to generate shares and a threshold of those that would be required when signing
+ */
+export type DkgOptions = {
+  threshold: number;
+  participants: number;
+};
+
+export type BlsKeyPairOptions = DkgOptions | BlsKeys;
+
+export type KeyPairOptions = Seed | PrivateKey | PublicKey | BlsKeyPairOptions;
 
 export type BaseBuilder = BaseTransactionBuilder | BaseTransactionBuilderFactory;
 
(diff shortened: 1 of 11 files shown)


modules/account-lib/src/coin/baseCoin/iface.ts
```diff
@@ -40,17 +50,30 @@ export function isSeed(source: KeyPairOptions): source is Seed {
 /**
  * @param source
  */
-export function isPrivateKey(source: KeyPairOptions): source is PrivateKey {
+export function isPrivateKey(source: KeyPairOptions | BlsKeyPairOptions): source is PrivateKey {
```
wesgraham, line 53: can we merge blsKeyPairOptions into KeyPairOptions? in order to minimize the modification to the typing here?

modules/account-lib/test/unit/coin/eth2/keyPair.ts
```diff
@@ -89,75 +115,185 @@ describe('Eth2 Key Pair', () => {
...
       should.throws(() => KeyPair.aggregatePubkeys([pub]));
     });
 
     it('from a set of invalid pubkeys', () => {
-      const pub1 = Uint8Array.from(Buffer.from('abd', 'hex'));
+      const pub1 = '0x' + Buffer.from('abc').toString('hex');
 
       const keyPair2 = new KeyPair();
-      const pub2 = Uint8Array.from(Buffer.from(keyPair2.getKeys().pub.slice(2), 'hex'));
+      const pub2 = keyPair2.getKeys().publicShare;
```
wesgraham, line 141: out of curiousity why did we modify pub to be publicShare now? is there something wrong with using pub as a field?

wesgraham (approved):

modules/account-lib/test/unit/coin/eth2/keyPair.ts
fperalta-simtlix: it is because we are using the BlsKeys type we defined in iface. There, we declared the property as publicShare instead of pub. We can rename it if that is confusing.

wesgraham: That's ok thanks for clarification

wesgraham (approved):

wesgraham (approved):

PR #1923, opened 2022-01-27
feat(utxo-bin): use prevOutputs, spend status

OttoAllmendinger:
* allow lookup via txid
* allow transaction status lookup (found/confirmed)
* allow transaction input lookup (signature verification, fee info)
* allow transaction output spend status lookup

Issue: BG-41897


modules/statics/src/map.ts
```diff
@@ -59,4 +59,8 @@ export class CoinMap {
   public forEach(callback: (coin: Readonly<BaseCoin>, coinName: string) => void): void {
     this._map.forEach(callback);
   }
+
+  public [Symbol.iterator](): IterableIterator<[string, Readonly<BaseCoin>]> {
```
sangaman, line 63: What is this needed/used for?

modules/utxo-bin/src/commands.ts
```diff
@@ -7,25 +7,44 @@ import * as utxolib from '@bitgo/utxo-lib';
...
 
 export const cmdParse = {
-  command: 'parse <path>',
+  command: 'parse [path]',
   describe: 'display transaction components in human-readable form',
 
   builder(b: yargs.Argv<unknown>): yargs.Argv<Args> {
     return b
-      .option('path', { type: 'string', default: '-' })
+      .option('path', { type: 'string', nargs: 1, default: '' })
```
sangaman, line 41: Why is `nargs` and `default` set here? Would this option just default to `undefined` without it?

sangaman (approved):
Nice work, this is a great tool.

modules/statics/src/map.ts
OttoAllmendinger: https://github.com/BitGo/BitGoJS/pull/1923/commits/a4a2f4b830084a136840abbaf3b30fe5852a60e1

modules/utxo-bin/src/commands.ts
OttoAllmendinger: This is to make `yargs parse -` actually parse stdin.  Without this the, `-` argument is ignored: https://github.com/yargs/yargs/issues/1312#issuecomment-940633564

PR #1940, opened 2022-02-01
fix(account-lib): add check for anonymous transaction to build getter

aaldemir:
add check for anonymous transaction when creating anonymous proxy transactions so that `getBuilder` returns the proper builder instead of throwing

JIRA ticket: https://bitgoinc.atlassian.net/browse/STLX-12770


lcovar (approved):

modules/account-lib/package.json
```diff
@@ -20,7 +20,7 @@
     "lint": "eslint --quiet 'src/**/*.ts' 'test/**/*.ts'",
     "lint-fix": "eslint --fix 'src/**/*.ts' 'test/**/*.ts'",
     "prepublishOnly": "npm run compile",
-    "prepare": "npm run gen-protobuf && npm run gen-protobufts &&  npm run hedera-gen-protobuf && npm run hedera-gen-protobufts && tsc && cp -r ./resources ./dist",
+    "prepare": "npm run gen-protobuf && npm run gen-protobufts &&  npm run hedera-gen-protobuf && npm run hedera-gen-protobufts && NODE_OPTIONS=\"--max-old-space-size=4096\" tsc && cp -r ./resources ./dist",
```
tylerlevine, line 23: `tsc` is not the thing OOM'ing, it's `mocha`. Try adding this to the front of the `unit-test` script instead.

tylerlevine (changes requested):
I'd rather get to the bottom of _why_ it's using so much memory, but if you just want a quick fix please follow the comment I left in the PR

modules/account-lib/package.json
aaldemir: I think it is using all the memory to store the string `metadataRpc` here https://github.com/BitGo/BitGoJS/blob/42fd74c709e0e726cfc75b38707a08a5483532af/modules/account-lib/src/coin/dot/transactionBuilder.ts#L180

which is a part of the flow triggered in the beginning of the unit tests this PR adds. I am not sure how I can stub this part of the flow - maybe use `spy` and pull in the values from a file?

PR #1943, opened 2022-02-02
fix(express): correctly handle failed proxy calls 

tylerlevine:
When bitgo-express is unable to proxy a request, an error is thrown
which isn't handled properly by the error response code, which causes
the whole express application to exit ungracefully.

Instead, handle this proxy failure error properly, along with also
handling timeout conditions in the same way.

First two commits are just some general clean up.

All the interesting changes are in the third commit.

Ticket: BG-40697


byhow (approved):
LGTM

PR #1944, opened 2022-02-02
feat: add external signing functionality for v2 wallet

DariusParvin:
- Add signing functionality for v2 wallet in `handleV2Sign`
- Check config arguments to prevent a user from running as a normal BG Express instance and an external signer instance at the same time
- Check the `signerFileSystemPath` argument on startup to make sure it contains a JSON file with a "prv" field
- Only allow external signer mode to run in test mode

### DIFF
--- a/modules/express/src/errors.ts
+++ b/modules/express/src/errors.ts
@@ -40,3 +40,10 @@ export class IpcError extends Errors.BitGoJsError {
     Object.setPrototypeOf(this, IpcError.prototype);
   }
 }
+
+export class ExternalSignerConfigError extends Errors.BitGoJsError {
+  public constructor(message?: string) {
+    super(message || 'External signer configuration is invalid');
+    Object.setPrototypeOf(this, ExternalSignerConfigError.prototype);
+  }
+}
--- a/modules/express/src/expressApp.ts
+++ b/modules/express/src/expressApp.ts
@@ -220,6 +220,10 @@ function checkPreconditions(config: Config) {
     config.env = 'custom';
   }
 
+  if (env !== 'test' && (externalSignerUrl !== undefined || signerMode !== undefined)) {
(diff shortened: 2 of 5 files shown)


tylerlevine (dismissed):
awesome work 👍

modules/express/src/errors.ts
```diff
@@ -40,3 +40,10 @@ export class IpcError extends Errors.BitGoJsError {
     Object.setPrototypeOf(this, IpcError.prototype);
   }
 }
+
+export class ExternalSignerConfigError extends Errors.BitGoJsError {
+  public constructor(message?: string) {
+    super(message || 'External signer configuration is invalid');
+    Object.setPrototypeOf(this, ExternalSignerConfigError.prototype);
```
tylerlevine, line 47: nit: this is no longer necessary and can be removed from all other errors defined here. Please remove in a follow up PR though, no need to do this clean up item here

DariusParvin:
Thanks for the feedback! I added changes (and a test) that require express to be in test mode when using the external signer feature.

modules/express/src/expressApp.ts
```diff
@@ -190,6 +219,30 @@ function checkPreconditions(config: Config) {
...
+      'signerMode and signerFileSystemPath must both be set in order to run in external signing mode.'
+    );
+  }
+
+  if (signerFileSystemPath !== undefined) {
+    try {
+      checkSignerPrvPath(signerFileSystemPath);
+    } catch (e) {
+      throw e;
+    }
```
tylerlevine, line 244: No need to catch if you're just going to immediately rethrow - feel free to handle in a follow up PR to clean up some other nits as well.
```suggestion
    checkSignerPrvPath(signerFileSystemPath);
```

tylerlevine (approved):

PR #1950, opened 2022-02-03
chore(core): add deriveKeypair method to DOT

aaldemir:
When DOT wallet object is used in core for address creation, use hardened key derivation to generate an address to POST to WP for receive addresses. 

Follows workflow implemented by Solana

JIRA ticket: https://bitgoinc.atlassian.net/browse/STLX-13001

### DIFF
--- a/modules/core/src/v2/coins/dot.ts
+++ b/modules/core/src/v2/coins/dot.ts
@@ -4,6 +4,7 @@ import { BitGo } from '../../bitgo';
 import { MethodNotImplementedError } from '../../errors';
 import {
   BaseCoin,
+  DerivedKeyPair,
   DeriveKeypairOptions,
   KeyPair,
   ParsedTransaction,
@@ -103,7 +104,7 @@ export class Dot extends BaseCoin {
   }
 
   /** @inheritDoc */
-  deriveKeypair(params: DeriveKeypairOptions): KeyPair | undefined {
+  deriveKeypair(params: DeriveKeypairOptions): DerivedKeyPair | undefined {
     try {
       if (_.isNil(params.addressDerivationPrv)) {
         throw new Error('addressDerivationPrv is missing');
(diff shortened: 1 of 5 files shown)


modules/core/src/v2/coins/dot.ts
```diff
@@ -96,6 +102,40 @@ export class Dot extends BaseCoin {
...
+        throw new Error('failed to create valid derived base58 prv key');
+      }
+      if (!_.isString(pub) || !accountLib.Dot.Utils.default.isValidPublicKey(pub)) {
+        throw new Error('failed to create valid derived base58 pub key');
+      }
+      if (!accountLib.Dot.Utils.default.isValidAddress(derivedAddress)) {
+        throw new Error('failed to create valid DOT address from derived base58 keys');
+      }
+      return {
+        prv: '',
```
alebusse, line 130: I think you gonna need to return this prv since is needed to sign consolidation transfer for example.

aaldemir: ah okay - I was wondering where it was getting used. thanks.

modules/core/src/v2/wallet.ts
```diff
@@ -1414,7 +1414,7 @@ export class Wallet {
         });
 
         if (derivedKeypair) {
-          addressParams.derivedAddress = derivedKeypair.pub;
+          addressParams.derivedAddress = derivedKeypair.address ?? derivedKeypair.pub;
```
alebusse, line 1417: I created this PR https://github.com/BitGo/BitGoJS/pull/1952, it should help to make the code more coin agnostic

modules/core/test/v2/unit/wallet.ts
```diff
@@ -792,6 +792,141 @@ describe('V2 Wallet:', function () {
...
 
+  describe('Create Address for Polkadot', () => {
+    let dotWallet, coinNocks;
+    const passphrase = 'Passphrase1234';
+    const coinBitgo = new TestBitGo({ env: 'mock' });
+    coinBitgo.initializeTestVars();
+    const tdot = coinBitgo.coin('tdot');
+    const walletData = {
+      id: '598f606cd8fc24710d2ebadb1d9459bb',
+      coinSpecific: {
```
alebusse, line 803: Check the coinSpecific data here to see if they are ok

alebusse (approved):

PR #1955, opened 2022-02-04
feat(core): add function in SDK and write examples for deploying forwarder

mgarofalo-simtlix:
To enable users to  perform these operations from SDK as well. Ticket: STLX-12550

### DIFF
--- a/modules/core/example/ts/eth/deployForwarder.ts
+++ b/modules/core/example/ts/eth/deployForwarder.ts
@@ -1,0 +1,37 @@
+/**
+ * * To manually deploy an ETH address
+ *
+ * Copyright 2022 BitGo, Inc.  All Rights Reserved.
+ */
+const BitGoJS = require('bitgo');
+
+// change this to env: 'production' when you are ready for production
+const bitgo = new BitGoJS.BitGo({ env: 'test' });
+ 
+// Change coin to 'eth' when working with production
+const coin = '';
+ 
+// TODO: set your access token here
+const accessToken = '';
+ 
+// set your wallet from the YYYYY parameter here in the URL on app.bitgo-test.com
+// https://test.bitgo.com/enterprise/XXXXXXXXX/coin/teth/YYYYY/transactions
+const walletId = '';
+// set your address from wallet dashboard on app.bitgo-test.com or using the get-wallet example from this repo
+const addressId = '';
+ 
+async function deployForwarders() {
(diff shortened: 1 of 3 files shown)


modules/core/example/ts/eth/deployForwarder.ts
```diff
@@ -0,0 +1,42 @@
...
+// set your address from wallet dashboard on app.bitgo-test.com or using the get-wallet example from this repo
+const addressId = '';
+ 
+async function setForwarderFlags() {
+  bitgo.authenticateWithAccessToken({ accessToken });
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+ 
+  const deployForwarderOptions = {
+    address: addressId,
+    gasPrice: 0,
```
wesgraham, line 29: dont need gasprice and eip1559 - in fact by default you dont need either of them (it will auto specify)

modules/core/src/v2/wallet.ts
```diff
@@ -1097,6 +1104,33 @@ export class Wallet {
     }
     this._wallet = await this.bitgo.put(this.url()).send(forwarderFlags).result();
   }
+
+  /**
+     * To manually deploy an ETH address
+     * @param DeployForwardersOptions {Object} - {
+        "address": {String},
```
fperalta-simtlix, line 1111: `id` is missing

modules/core/src/v2/wallet.ts
```diff
@@ -273,6 +273,13 @@ export interface GetAddressOptions {
   reqId?: RequestTracer;
 }
 
+export interface DeployForwardersOptions {
+  address?: string;
```
wesgraham, line 277: every field is optional?

wesgraham (approved):

modules/core/src/v2/wallet.ts
```diff
@@ -1097,6 +1102,32 @@ export class Wallet {
...
+      throw new Error('address or id of address required');
+    }
+    let query;
+    if (params.address) {
+      query = params.address;
+    } else {
+      query = params.id;
+    }
+    this._wallet = await this.bitgo
+      .post(this.url(`/address/${encodeURIComponent(query)}/deployment`))
```
wesgraham, line 1125: you can maybe set this.url() outside of this call? nested function calls always harder to read

wesgraham (dismissed):
One nit otherwise LGTM

wesgraham (approved):

PR #1956, opened 2022-02-04
chore: dot TSS unit test

lcovar:
TICKET: BG-0000

### DIFF
--- a/modules/account-lib/test/unit/mpc/dotTss.ts
+++ b/modules/account-lib/test/unit/mpc/dotTss.ts
@@ -1,0 +1,61 @@
+/**
+ * @prettier
+ */
+import 'should';
+import { KeyPair, TransferBuilder } from '../../../src/coin/dot';
+
+import Eddsa from '../../../src/mpc/tss';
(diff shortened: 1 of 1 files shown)


modules/account-lib/test/unit/mpc/dotTss.ts
```diff
@@ -0,0 +1,61 @@
...
+      .fee({ amount: 0, type: 'tip' });
+    const tx = await builder.build();
+
+    // tx.toBroadcastFormat() generates a signing payload to be signed
+    // construct.signingPayload(this._dotTransaction, {
+    //   registry: this._registry,
+    // });
+
+    // create a buffer out of the txHex
+    const message_buffer = Buffer.from(tx.toBroadcastFormat(), 'utf-8');
```
byhow, line 44:
```suggestion
    const message_buffer = Buffer.from(tx.toBroadcastFormat(), 'hex');
```
from https://github.com/paritytech/txwrapper-core/blob/main/packages/txwrapper-core/src/core/construct/createSigningPayload.ts#L63 looks like they returned a hex

PR #1793, opened 2021-12-14
feat(account-lib): implement validityWindow and sequenceId for Sol

facundoCarreSimtlix:
implement validityWindow and sequenceId for Sol

stlx-9890

PR #1963, opened 2022-02-08
feat(account-lib): validate ValidityWindows in baseBuildTransaction

Krin2:
Implement getValidityWindows as an optional method to validate and return a validated ValidityWindow
or default params

[STLX-11627](https://bitgoinc.atlassian.net/browse/STLX-11627)

### DIFF
--- a/modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
+++ b/modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
@@ -1,6 +1,6 @@
 import BigNumber from 'bignumber.js';
 import { BaseCoin as CoinConfig } from '@bitgo/statics';
-import { BaseAddress, BaseKey, PublicKey } from './iface';
+import { BaseAddress, BaseKey, PublicKey, ValidityWindow } from './iface';
 import { BaseTransaction } from './baseTransaction';
 import { SigningError } from './errors';
 
@@ -138,6 +138,75 @@ export abstract class BaseTransactionBuilder {
    */
   coinName(): string {
     return this._coinConfig.name;
+  }
+
+  /**
+   * Verified validity windows params if them exist and return a valid validity windows.
+   * Unit param must be specified
+   * If params are not consistent, default params will be return based on firstValid and minDuration
+   * @param {ValidityWindow} params validity windows parameters to validate.
+   * @param {String} params.unit Parameter that could be 'blockheight' or 'timestamp'
+   * @param {Number} [params.minDuration] Optional - Minimum duration of the window
+   * @param {Number} [params.maxDuration] Optional - Maximum duration of the window
+   * @param {Number} [params.firstValid] Optional - First valid value
+   * @param {Number} [params.lastValid] Optional - Last valid value
+   * @returns {ValidityWindow} verified validity windows or default values
+   */
+  getValidityWindow(params: ValidityWindow): ValidityWindow {
+    if (!params.unit || (params.unit !== 'timestamp' && params.unit !== 'blockheight')) {
+      throw new Error('Unit parameter must be specified as blockheight or timestamp');
+    }
+    const unit = params.unit;
+    let defaultMinDuration: number;
+    let defaultMaxDuration: number;
+    let defaultFirstValid: number;
+    let defaultLastValid: number;
+
+    /* Set Default Params
+      minimum duration is set as 1 hr (3600000 msec) if unit is timestamp or 20 blocks if it is blockheight
+      maximum duration is set as 1 year (31536000000 msec) if unit is timestamp or 1000000 blocks if it is blockheight.
+     */
+    if (unit === 'timestamp') {
+      defaultMinDuration = 0;
+      defaultMaxDuration = 31536000000;
+      defaultFirstValid = Date.now();
+      defaultLastValid = defaultFirstValid + defaultMaxDuration;
+    } else {
+      defaultMinDuration = 0;
+      defaultMaxDuration = 1000000;
+      defaultFirstValid = 0;
+      defaultLastValid = defaultFirstValid + defaultMaxDuration;
+    }
+
+    // If any params exist, they will be used, otherwise it will be used default params.
+    let firstValid: number = params.firstValid || defaultFirstValid;
+    let lastValid: number = params.lastValid || defaultLastValid;
+    let minDuration: number = params.minDuration || defaultMinDuration;
+    let maxDuration: number = params.maxDuration || defaultMaxDuration;
+
+    /* Validate Params:
+      minDuration < maxDuration
+      firstValid < lastValid
+      firstValid + minDuration <= lastValid <= firstValid + maxDuration
+     */
+    if (minDuration >= maxDuration) {
+      throw new Error(`Expected maxDuration (${maxDuration}) to be grather than minDuration (${minDuration})`);
+    }
+    firstValid = firstValid >= 0 ? firstValid : defaultFirstValid;
+    minDuration = minDuration >= 0 ? minDuration : defaultMinDuration;
+    maxDuration = maxDuration > minDuration ? maxDuration : defaultMaxDuration;
+    lastValid =
+      lastValid >= firstValid + minDuration && lastValid <= firstValid + maxDuration
+        ? lastValid
+        : firstValid + maxDuration;
+
+    return {
+      firstValid,
+      lastValid,
+      minDuration,
+      maxDuration,
+      unit,
+    };
   }
 
   /**
--- a/modules/account-lib/test/unit/coin/baseCoin/transactionBuilder.ts
+++ b/modules/account-lib/test/unit/coin/baseCoin/transactionBuilder.ts
@@ -62,4 +62,98 @@ describe('Transaction builder', () => {
     sandbox.assert.calledOnce(validateTransaction);
     sandbox.assert.calledOnce(buildImplementation);
   });
+
+  it('should verify validity window params', () => {
+    const testTx = sinon.createStubInstance(TestTransaction);
+    testTx.canSign.returns(true);
+    txBuilder.from(testTx);
+    txBuilder.build();
+    let validityWindow;
+
+    let params = {};
+    should(() => txBuilder.getValidityWindow(params)).throwError(
+      'Unit parameter must be specified as blockheight or timestamp',
+    );
+
+    params = {
+      unit: 'wrongUnit',
+    };
+    should(() => txBuilder.getValidityWindow(params)).throwError(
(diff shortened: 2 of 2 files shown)


Krin2:
Default values need to be defined.
Params validation came from [here](https://github.com/BitGo/BitGoJS/pull/1793/files#r771034682)

rliberal-simtlix (approved):

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -140,6 +140,34 @@ export abstract class BaseTransactionBuilder {
...
+   * @returns {ValidityWindow} verified validity windows or default values
+   */
+  getValidityWindow?(params: ValidityWindow): ValidityWindow {
+    const size = 100000;
+    const window: ValidityWindow = params;
+
+    window.firstValid = window.firstValid ? window.firstValid : 0;
+    window.lastValid = window.lastValid ? window.lastValid : 0;
+    window.minDuration = window.minDuration ? window.minDuration : size;
+    window.maxDuration = window.maxDuration ? window.maxDuration : size;
```
CapnMigraine, line 155: I find the prop names `minDuration` and `maxDuration` confusing. If they're always timestamps, doesn't it make more sense to use `firstValid` and `lastValid` when the window is determined by either timestamps or block heights?

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -140,6 +140,34 @@ export abstract class BaseTransactionBuilder {
...
+    window.unit = window.unit ? window.unit : 'blockheight';
+    if (window.minDuration > window.maxDuration) {
+      window.maxDuration = window.minDuration;
+    }
```
CapnMigraine, line 159: Shouldn't we throw an exception instead? Otherwise we're closing the window here.

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -140,6 +140,34 @@ export abstract class BaseTransactionBuilder {
...
+    if (
+      window.lastValid <= window.firstValid ||
+      window.lastValid < window.firstValid + window.minDuration ||
+      window.lastValid >= window.firstValid + window.maxDuration ||
+      window.firstValid + window.minDuration > window.firstValid + window.maxDuration
```
CapnMigraine, line 164: Are `minDuration` and `maxDuration` the number of blocks included in the validity window?

CapnMigraine: Please add comments to explain what's happening here.

CapnMigraine (commented):
I'm a bit confused with the way we calculate the window using first/last valid and min/max duration. Maybe we should create the window based on absolute or relative values, bot not a mix.

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -140,6 +140,34 @@ export abstract class BaseTransactionBuilder {
     return this._coinConfig.name;
   }
 
+  /**
+   * Verified validity windows params if them exist and return a valid validity windows.
+   * @param {ValidityWindow} [params] optional validity windows parameter to validate.
+   * @returns {ValidityWindow} verified validity windows or default values
+   */
+  getValidityWindow?(params: ValidityWindow): ValidityWindow {
```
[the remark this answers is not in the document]
Krin2: when you put that mark in an abstract method, it said that method is optional. I made this method in several ways and I forgot it.

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
Krin2: The reason I'm not throwing an exception is that I am assigning maxDuration to be equal to minDuration by default because maxDuration shouldn't be less than minDuration. I based on [this comment](https://github.com/BitGo/BitGoJS/pull/1793/files#r771034682) for validations.

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -140,6 +140,34 @@ export abstract class BaseTransactionBuilder {
...
+    const size = 100000;
+    const window: ValidityWindow = params;
+
+    window.firstValid = window.firstValid ? window.firstValid : 0;
+    window.lastValid = window.lastValid ? window.lastValid : 0;
+    window.minDuration = window.minDuration ? window.minDuration : size;
+    window.maxDuration = window.maxDuration ? window.maxDuration : size;
+    window.unit = window.unit ? window.unit : 'blockheight';
```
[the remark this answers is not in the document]
Krin2: ValidityWindows interface was implemented by DOT and they use this parameter as optional. I agreed that the unit should be required, but I will break DOT everywhere. I'm seeing at least 43 test implementing this interface just in account-lib. I think this change about unit should be managed by DOT in order to see that this change does not break WP too or something else.

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -140,6 +140,51 @@ export abstract class BaseTransactionBuilder {
...
+   * If params are not consistent, default params will be return based on firstValid and minDuration
+   * @param {ValidityWindow} params validity windows parameters to validate.
+   * @param {String} params.unit Parameter that could be 'blockheght' or 'timestamp'
+   * @param {Number} [params.minDuration] Optional - Minimum duration of the window
+   * @param {Number} [params.maxDuration] Optional - Maximum duration of the window
+   * @param {Number} [params.firstValid] Optional - First valid value
+   * @param {Number} [params.lastValid] Optional - Last valid value
+   * @returns {ValidityWindow} verified validity windows or default values
+   */
+  getValidityWindow(params: ValidityWindow): ValidityWindow {
```
joey-bai, line 155: not 100% sure how this is going to be used.

one thing that sticks out is that this function is in the builder, but doesn't actually mutate the underlying transaction in some way.

as an example, EOS supports setting a transaction expiration time (e.g. `lastValid = now + '1 hour'`). how can i use this function to set that expiration time?

Krin2: It is not clear to me either, but base on comments, this method:
- should return some default values if params are empty
- should verify params if some or all of them are present
- should build new default params based on the existent ones if they are not consisten.

In the example you mentioned, **_now_** could be the `firstValid` and **_1 hour_** could be the `maxDuration`.
Anyway, `defaultDuration` could be differenciate for `blockheight` and `timestamp` and it could be differenciated an `initialValue` to be **_now_** for `timestamp` an **_0_** for `blockheight`

shiksb:
Diagram to help reviewers 

validity (1)

shiksb (changes requested):

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -140,6 +140,75 @@ export abstract class BaseTransactionBuilder {
...
+      throw new Error(`Expected maxDuration (${maxDuration}) to be grather than minDuration (${minDuration})`);
+    }
+    firstValid = firstValid >= 0 ? firstValid : defaultFirstValid;
+    minDuration = minDuration >= 0 ? minDuration : defaultMinDuration;
+    lastValid =
+      lastValid >= firstValid + minDuration && lastValid <= firstValid + maxDuration
+        ? lastValid
+        : firstValid + maxDuration;
+    maxDuration = maxDuration > minDuration && firstValid + maxDuration > lastValid ? maxDuration : lastValid;
+
```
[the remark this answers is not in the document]
Krin2: done

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -140,6 +140,51 @@ export abstract class BaseTransactionBuilder {
...
+    }
+    /* Validate that lastValid is grather than firstValid + minDuration and lower than firstValid + maxDuration
+      firstValid + minDuration < lastValid < first valid + maxDuration
+      If not, it is assigned as firstValid + minDuration by  default
+    */
+    if (
+      window.lastValid <= window.firstValid ||
+      window.lastValid < window.firstValid + window.minDuration ||
+      window.lastValid >= window.firstValid + window.maxDuration ||
+      window.firstValid + window.minDuration > window.firstValid + window.maxDuration
```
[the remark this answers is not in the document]
Krin2: done

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
Krin2: done

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
Krin2: done

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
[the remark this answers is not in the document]
Krin2: I already remove it

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -140,6 +140,34 @@ export abstract class BaseTransactionBuilder {
     return this._coinConfig.name;
   }
 
+  /**
+   * Verified validity windows params if them exist and return a valid validity windows.
+   * @param {ValidityWindow} [params] optional validity windows parameter to validate.
+   * @returns {ValidityWindow} verified validity windows or default values
+   */
+  getValidityWindow?(params: ValidityWindow): ValidityWindow {
+    const size = 100000;
```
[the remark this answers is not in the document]
Krin2: What was told to me was to return always default param. If you don't pass some param, then you put one bigger or smaller enough to consider "from the beginning" or "for ever"

modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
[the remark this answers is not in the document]
Krin2: I just throw an exception if no unit is passed

shiksb (approved):

rliberal-simtlix (approved):

alebusse (approved):

joey-bai (approved):

PR #1966, opened 2022-02-09
feat(root): update SDK sendMany to use TSS

alebusse:
Update this this commit msg later

STLX-13002


modules/core/src/v2/wallet.ts
```diff
@@ -484,6 +487,53 @@ export interface DownloadKeycardOptions {
...
 
+// #region TSS interfaces
+
+export interface PrebuildTransactionWithIntentOptions {
+  reqId?: RequestTracer;
+  intentType?: string;
+  sequenceId?: string;
+  recipients?: {
+    address: string;
+    amount: string | number;
```
rliberal-simtlix, line 498: I think here we can unify in just one type. This is just a comment and may need more analysis.

modules/core/src/v2/wallet.ts
```diff
@@ -484,6 +487,53 @@ export interface DownloadKeycardOptions {
...
+  memo?: Memo;
+  verification?: VerificationOptions;
+}
+
+export interface TSSSendManyOptions extends SendManyOptions {
+  reqId: RequestTracer;
+  intentType: string;
+}
+
+// find a better name
```
rliberal-simtlix, line 510: Please add TODOS for this.

modules/core/src/v2/wallet.ts
```diff
@@ -484,6 +487,53 @@ export interface DownloadKeycardOptions {
...
+  BITGO = '3',
+}
+
+// complete with more props
+export interface TransactionRequestJSON {
+  txRequestId: string;
+  unsignedTxs: string[];
+}
+
+export enum SignatureShareType {
```
rliberal-simtlix, line 523: I was thinking here a kind of map structure with key value {[user,1], [backup,2], [bitgo, 3]} but the two enums is also valid. The only bad thing is your repeating names.

modules/core/src/v2/wallet.ts
alebusse: Most comments are just placeholders so we can discuss around. This is not final at all.

modules/core/src/v2/wallet.ts
```diff
@@ -2587,4 +2643,146 @@ export class Wallet {
...
     }
   }
+
+  // #region TSS region
+
+  async sendManyTSS(params: TSSSendManyOptions): Promise<any> {
+    // await this.verifyParams(params); improve validations
+
+    if (_.isNil(params.walletPassphrase)) {
+      throw new Error('something');
```
rliberal-simtlix, line 2653: Add proper error here

modules/core/src/v2/wallet.ts
```diff
@@ -2587,4 +2643,146 @@ export class Wallet {
       };
     }
   }
+
+  // #region TSS region
+
+  async sendManyTSS(params: TSSSendManyOptions): Promise<any> {
```
rliberal-simtlix, line 2649: Function is too long. We need verify here if we can reduce it. But on the other hand I know you need to do execute several calls for the shares to final execute the send many for tss.

modules/core/src/v2/wallet.ts
```diff
@@ -2587,4 +2643,146 @@ export class Wallet {
...
+    // await this.verifyParams(params); improve validations
+
+    if (_.isNil(params.walletPassphrase)) {
+      throw new Error('something');
+    }
+
+    const unsignedTx = await this.prebuildTxWithIntent(params);
+    const { txRequestId } = unsignedTx;
+
+    const bufferUnsignedTx = Buffer.from(unsignedTx.unsignedTxs[0]);
```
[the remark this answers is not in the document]
alebusse: Same as in WP, https://github.com/BitGo/bitgo-microservices/blob/1754e88cba74f2ceb6483fdf3fec6ea0b4406e00/packages/wallet-platform/app/base/transactionRequests.ts#L53

Its an array but i didnt see anywhere we they actually put more than 1 tx inside.

modules/core/src/v2/wallet.ts
alebusse: Its the pipeline design we talked about, its better than the current sendMany i guess.

modules/core/src/v2/wallet.ts
alebusse: This is the WP implementation of this interface, i didnt like it much neither
https://github.com/BitGo/bitgo-microservices/blob/e9e69bce28695c8ad545b61c6d93c5460ebd5ad6/packages/wallet-platform/app/models/v2/TransactionRequest/TransactionRequest.types.ts#L37-L39

modules/core/src/v2/wallet.ts
```diff
@@ -484,6 +487,53 @@ export interface DownloadKeycardOptions {
   backupKeyID?: string;
 }
 
+// #region TSS interfaces
+
+export interface PrebuildTransactionWithIntentOptions {
+  reqId?: RequestTracer;
+  intentType?: string;
```
joey-bai, line 494: this should be required since this interface is meant to be used with intents

modules/core/src/v2/wallet.ts
joey-bai: i think string is better than number here.

clients have been experience a couple of issues with numbers being too large (i think AVAX), which produced some unexpected results due to rounding.

safer to have this be a string and use `BigNumber` to perform operations on it

modules/core/src/v2/wallet.ts
```diff
@@ -484,6 +487,53 @@ export interface DownloadKeycardOptions {
...
+  verification?: VerificationOptions;
+}
+
+export interface TSSSendManyOptions extends SendManyOptions {
+  reqId: RequestTracer;
+  intentType: string;
+}
+
+// find a better name
+export enum ShareKeyPosition {
```
joey-bai, line 511: do we need to export this? looks like its only used in this file.
plz double check your other exports to make sure we're not over exposing info

modules/core/src/v2/wallet.ts
```diff
@@ -2587,4 +2643,144 @@ export class Wallet {
       };
     }
   }
+
+  // #region TSS region
```
joey-bai, line 2647: lets make the below functions private.

alternatively, you can move them to `TssUtils` and make them public since that class will never be exported publicly. this will help with unit testing

modules/core/src/v2/wallet.ts
alebusse: I will have to convert it to string, since im just passing what i got from the params of https://github.com/BitGo/BitGoJS/blob/master/modules/core/src/v2/wallet.ts#L386

Also i may have to convert it to this 'Amount' type https://github.com/BitGo/bitgo-microservices/blob/210059943d402e8abdc88dca08202768b462f4ae/packages/common-interface/src/interface/api.ts#L305

modules/core/src/v2/wallet.ts
alebusse: Removing every not needed export

modules/core/src/v2/wallet.ts
alebusse: @zahin-mohammad Seems that old `recipients:{address:string, amount:string|number}[]` are not compatible with the txRequest expected types `recipients: { address:Address, amount: Amount}[]`

modules/core/src/v2/wallet.ts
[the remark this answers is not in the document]
joey-bai: maybe we're meant to do this in a loop?
either way, its okay to assume that there will always be 1 unsigned tx to start.

```
for (let unsginedTx of unsignedTx.unsignedTxs) {
   // create sign share
   // send sendShare
}

return this.sendTxRequest();
```

modules/core/src/v2/wallet.ts
zahin-mohammad: We don't necessarily need to do it here, but as long as we have a way of translating things to the intent when it comes time to interact with the WP api.

joey-bai (commented):
was able to get send many 'working' with the commented fixes.

there are a couple of server side bugs though:
- send fails due to a signing error from hsm
- can't create more than 1 tx request per wallet (violates mongo db unique constraint)

modules/core/src/v2/wallet.ts
```diff
@@ -2654,4 +2712,133 @@ export class Wallet {
...
+      .result()) as TransactionRequestJSON;
+
+    try {
+      await this.baseCoin.verifyTransaction({
+        txParams: params,
+        txPrebuild: { txHex: unsignedTx.unsignedTxs[0] },
+        wallet: this,
+        verification: params.verification ?? {},
+        reqId: params.reqId,
+      });
```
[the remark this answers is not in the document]
alebusse: Adding a TODO

modules/core/src/v2/wallet.ts
```diff
@@ -2654,4 +2712,133 @@ export class Wallet {
...
+
+  async sendSignatureShare(txRequestId: string, signatureShare: SignatureShareRecord): Promise<SignatureShareRecord> {
+    return this.bitgo
+      .post(this.baseCoin.url('/wallet/' + this.id() + '/txrequests/' + txRequestId + '/signatureshares'))
+      .send(signatureShare)
+      .result();
+  }
+
+  async getBitgoToUserRShare(txRequestId: string, userSignShare: SignShare): Promise<SignatureShareRecord> {
+    const rShare: RShare = userSignShare.rShares[SignatureShareType.BITGO];
```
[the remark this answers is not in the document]
alebusse: Ya, sorry,  wrong interface. I had to use this one `ShareKeyPosition`

modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -181,7 +236,7 @@ export class TssUtils {
     const bitgoKey = await openpgp.readKey({ armoredKey: bitgoPublicKeyStr });
 
     const userToBitGoMessage = await openpgp.createMessage({
-      text: Buffer.from(userKeyShare.yShares[3].u).toString('hex'),
+      text: Buffer.concat([Buffer.from(userKeyShare.yShares[3].u, 'hex'), Buffer.alloc(32)]).toString('hex'),
```
rliberal-simtlix, line 239: Why here buffer.alloc(32)

modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -1,22 +1,77 @@
...
+}
+
+enum ShareKeyPosition {
+  USER = '1',
+  BACKUP = '2',
+  BITGO = '3',
+}
+
+// complete with more props if neccesary
+interface TransactionRequestJSON {
```
joey-bai, line 34: thoughts on renaming this `TxRequest` to match the resource name from the url?
makes it clear that this model is what we get back from calling the `.../txRequests` endpoint

modules/core/src/v2/internal/tssUtils.ts
joey-bai: HSM expects this value to have the trailing empty 32 bits (bytes?).
without it, the keyshares do not get combined properly during wallet creation, and then signing won't work.

i merged a change recently to fix this independently.

modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -285,4 +340,153 @@ export class TssUtils {
 
     return keychains;
   }
+
+  async signTSSTxRequest(
```
joey-bai, line 344: add documentation:
```
Signs and sends the input Transaction request.
```

modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -285,4 +340,153 @@ export class TssUtils {
...
+    unsignedTx: TransactionRequestJSON,
```
joey-bai, line 345: `unsignedTx` -> `txRequest`

modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -285,4 +340,153 @@ export class TssUtils {
...
+    // TODO(STLX-13411): Implement verify transaction
+    return unsignedTx;
+  }
+
+  async getSignablePayload(unsignedTx: TransactionRequestJSON): Promise<Buffer> {
+    const signablePayload = Buffer.from(unsignedTx.unsignedTxs[0].signableHex, 'hex');
+    const signablePayloadfromPayload = await this.baseCoin.getSignablePayload(unsignedTx.unsignedTxs[0].serializedTx);
+    if (signablePayloadfromPayload.toString('hex') !== signablePayload.toString('hex')) {
+      throw new Error('Missmatched signable payloads');
+    }
```
joey-bai, line 399: let's leave this part out. i added it while i was debugging tss signing.
ideally, this would be verified in the `verifyTransaction` function _before_ `.send()` is called.

modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -285,4 +340,153 @@ export class TssUtils {
...
   }
+
+  async signTSSTxRequest(
+    unsignedTx: TransactionRequestJSON,
+    walletPassphrase: string,
+    reqId: RequestTracer
+  ): Promise<any> {
+    const { txRequestId } = unsignedTx;
+
+    const signablePayload = await this.getSignablePayload(unsignedTx);
```
joey-bai, line 351: change to `const signablePayload = Buffer.from(unsignedTx.signableHex, 'hex');`

we can expect validation to be done _before_ this function is called.

modules/core/src/v2/internal/tssUtils.ts
joey-bai, line 344: maybe rename to `signAndSendTxRequest`
can drop TSS since its implied - function's in `tssUtils`

modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -285,4 +340,153 @@ export class TssUtils {
...
+      share: userToBitgoGShare.R + userToBitgoGShare.gamma,
+    };
+
+    await this.sendSignatureShare(txRequestId, signatureShare);
+  }
+
+  async sendTxRequest(txRequestId: string): Promise<any> {
+    return this.bitgo
+      .post(this.baseCoin.url('/wallet/' + this.wallet.id() + '/tx/send'))
+      .send(txRequestId)
```
joey-bai, line 489: i _think_ this needs to be an object:
`{ txRequestId }`

modules/core/src/v2/internal/tssUtils.ts
alebusse: Working on it, all js doc is missing yet

modules/account-lib/src/mpc/tss.ts
```diff
@@ -36,7 +36,7 @@ import Shamir from './shamir';
 import { randomBytes as cryptoRandomBytes } from 'crypto';
 
 const convertObjectHexToBuffer = (object: Record<string, any>, keys: string[]) => {
-  const result: any = object;
+  const result: any = { ...object };
```
alebusse, line 39: Change needed to make a deep copy of the object since it was just doing a reference and overwriting the object itself.

modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -1,22 +1,77 @@
...
+  signatureShares?: SignatureShareRecord[];
+}
+
+enum SignatureShareType {
+  USER = 'user',
+  BACKUP = 'backup',
+  BITGO = 'bitgo',
+}
+
+export interface SignatureShareRecord {
```
joey-bai, line 49: do we need to export this?

modules/core/src/v2/wallet.ts
```diff
@@ -451,6 +454,7 @@ export interface WalletData {
     backup?: string;
     bitgo?: string;
   };
+  multisigType: string;
```
joey-bai, line 457: lets change to `multisigType: 'onchain' | 'tss'`

modules/core/src/v2/wallet.ts
```diff
@@ -424,6 +426,7 @@ export interface SendManyOptions extends PrebuildAndSignTransactionOptions {
   instant?: boolean;
   memo?: Memo;
   transferId?: number;
+  intentType?: string;
```
joey-bai, line 429: don't think we should expose this field yet. im not 100% convinced we want backwards compatibility with `send` and `sendMany` now that we're moving towards TxRequests / Intents.

to start, let's assume that all TSS sends are payments until we figure out how to proceed

modules/core/src/v2/wallet.ts
```diff
@@ -2254,38 +2274,41 @@ export class Wallet {
...
+        'type',
+        'trustlines',
+        'transferId',
+        'stakingOptions',
+      ]);
+      const finalTxParams = _.extend({}, halfSignedTransaction, selectParams);
+
+      return this.bitgo.post(this.url('/tx/send')).send(finalTxParams).result();
+    } else {
+      return this.sendManyTSS(params as TSSSendManyOptions);
```
joey-bai, line 2310: lets do something like:
`const tssParams = _.assign(params, { intentType: 'payment' }) as TSSSendManyOptions`

joey-bai (approved):
LGTM.

This is a good starting point. I think its best to not make any interface changes to `send / sendMany` if we can avoid it. I get the feeling we may want to branch off and create a different set of functions used only with intents. but im not sure yet.

modules/core/src/v2/internal/tssUtils.ts
alebusse: Im using it in the unit tests https://github.com/BitGo/BitGoJS/pull/1966/files#diff-2cef92cfdc88be0400ccdbd5d0b082064f0d66e8bbd01aa9ac4efa43ee2d9415R244

modules/account-lib/src/mpc/tss.ts
```diff
@@ -34,9 +34,10 @@ import Ed25519Curve from './curves';
 import { sha512 } from 'js-sha512';
 import Shamir from './shamir';
 import { randomBytes as cryptoRandomBytes } from 'crypto';
+import _ from 'lodash';
 
 const convertObjectHexToBuffer = (object: Record<string, any>, keys: string[]) => {
-  const result: any = object;
+  const result: any = _.cloneDeep(object);
```
zahin-mohammad, line 40: Nice debugging 🙏🏾

joey-bai (approved):

PR #1975, opened 2022-02-09
feat(account-lib): implement add signature for sol

joey-bai:
implements add signature func for sol.
needed for TSS

Ticket: BG-42635

### DIFF
--- a/modules/account-lib/src/coin/sol/transactionBuilder.ts
+++ b/modules/account-lib/src/coin/sol/transactionBuilder.ts
@@ -2,7 +2,7 @@ import BigNumber from 'bignumber.js';
 import { BaseCoin as CoinConfig } from '@bitgo/statics';
 import { BaseTransactionBuilder, TransactionType } from '../baseCoin';
 import { BuildTransactionError, SigningError } from '../baseCoin/errors';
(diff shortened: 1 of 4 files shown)


modules/account-lib/src/coin/sol/transactionBuilder.ts
```diff
@@ -13,8 +13,14 @@ import assert from 'assert';
 import { DurableNonceParams, InstructionParams, Memo, Nonce } from './iface';
 import { instructionParamsFactory } from './instructionParamsFactory';
 
+interface Signature {
```
byhow, line 16: We can add this to `modules/account-lib/src/coin/baseCoin/iface.ts` since we use this in the `baseTransaction.ts` maybe?

byhow (approved):
Just a nit. Thanks for the reference!

modules/account-lib/src/coin/sol/transactionBuilder.ts
joey-bai: good call out. moving.

joey-bai: Refactoring in subsequent PR.

PR #1984, opened 2022-02-10
Adding Sol Tokens

bitgoandrew:


modules/statics/src/account.ts
```diff
@@ -73,6 +73,7 @@ export interface EosCoinConstructorOptions extends AccountConstructorOptions {
 
 export interface SolCoinConstructorOptions extends AccountConstructorOptions {
   tokenAddress: string;
+  mintAddress: string;
```
joey-bai, line 76: i think when we first added this, `tokenAddress` was meant to represent the mint pk.
don't need both properties, lets just keep 1 (no strong preference which).

modules/statics/src/coins.ts
```diff
@@ -1252,4 +1254,136 @@ export const coins = CoinMap.fromCoins([
...
+    AccountCoin.DEFAULT_FEATURES,
+    '',
+    'SRM'
+  ),
+  solToken(
+    'sol:ORCA',
+    'Orca',
+    6,
+    'orcaEKTdK7LKz57vaAYr9QeNsVEPfiu6QeMU1kektZE', // https://explorer.solana.com/address/orcaEKTdK7LKz57vaAYr9QeNsVEPfiu6QeMU1kektZE
+    '8DzsCSvbvBDYxGB4ytNF698zi6Dyo9dUBVRNjZQFHSUt',
```
joey-bai, line 1273: this is the mint authority, which shouldn't be needed in token metadata.

also, this authority can potentially change (i.e. owner of the mint decides to change authority to a different account for whatever reason).

modules/statics/src/coins.ts
```diff
@@ -1252,4 +1254,136 @@ export const coins = CoinMap.fromCoins([
...
+    AccountCoin.DEFAULT_FEATURES,
+    '',
+    'SLND'
+  ),
+  tsolToken(
+    'tsol:SRM',
+    'Serum',
+    6,
+    'Test SRM',
+    'Test Mint',
```
joey-bai, line 1328: we should create our own mints for these tokens in devnet (in sandbox) and refer to them here.

this way, we will be able to do internal testing in dev, test, testnet environments

modules/statics/src/coins.ts
```diff
@@ -1252,4 +1254,136 @@ export const coins = CoinMap.fromCoins([
...
+  ),
+  solToken(
+    'sol:ORCA',
+    'Orca',
+    6,
+    'orcaEKTdK7LKz57vaAYr9QeNsVEPfiu6QeMU1kektZE', // https://explorer.solana.com/address/orcaEKTdK7LKz57vaAYr9QeNsVEPfiu6QeMU1kektZE
+    '8DzsCSvbvBDYxGB4ytNF698zi6Dyo9dUBVRNjZQFHSUt',
+    UnderlyingAsset.ORCA,
+    AccountCoin.DEFAULT_FEATURES,
+    '',
```
joey-bai, line 1276: what are these emtpy strings for?

bitgoandrew: removed bc theyre initialized in token ctor. the prefix/suffix fields seem to be used for formatting in ms

joey-bai (approved):
lgtm

PR #1987, opened 2022-02-10
docs: add documentation for external signer mode

DariusParvin:
Ticket: BG-39865

Add external signer mode description and diagram to express readme.


modules/express/README.md
```diff
@@ -163,31 +163,48 @@ BitGo Express currently supports the following proxy protocols:
 * SOCKSv5
 * PAC
 
+### External Signing Mode
```
tylerlevine, line 166: Let's add some text here to say that this mode is currently available only in testnet and is under active development. Breaking changes may be made to it without incrementing the major version until the feature is stabilized and made available in production.

modules/express/README.md
```diff
@@ -163,31 +163,48 @@ BitGo Express currently supports the following proxy protocols:
...
+BitGo Express can be run in an external signing mode, where the signing of transactions is performed in a separate instance of BitGo Express that has access to the private keys. This may be preferable for users who would like to apply their signature to their transactions using a more secure environment than BitGo SDK or BitGo Express, such as an offline signer.
```
tylerlevine, line 167: Does this allow for the signer to be offline?

It might be more accurate to say a signer with restricted access or network connectivity.

modules/express/README.md
```diff
@@ -163,31 +163,48 @@ BitGo Express currently supports the following proxy protocols:
...
+
+To set up BitGo Express with an external signer, a url to the external signer instance of BitGo Express must be provided using the `externalSignerUrl` configuration option. The corresponding external signer instance of BitGo Express must have `signerMode` set, and `signerFileSystemPath` set to the path of a json containing the private key.
+
+The JSON file containing the unencrypted private key(s) must be in the format `"<walletId>": "<privateKey>"`wallet ID and private key. Here is an example json file containing two wallet IDs and their corresponding private keys.
```
tylerlevine, line 171:
```suggestion
The JSON file containing the unencrypted private key(s) must be in the format `"<walletId>": "<privateKey>"` wallet ID and private key. Here is an example json file containing two wallet IDs and their corresponding private keys.
```

modules/express/README.md
```diff
@@ -163,31 +163,48 @@ BitGo Express currently supports the following proxy protocols:
...
-| -d              | --debug                | N/A, use `BITGO_DEBUG_NAMESPACE` instead | N/A           | Enable debug output for bitgo-express. This is equivalent to passing `--debugnamespace bitgo:express`.                  |
+| Flag Short Name | Flag Long Name         | Environment Variable                     | Default Value | Description                                                                                                                                                                                                                                                                                                                                         |
+| --------------- | ---------------------- | ---------------------------------------- | ------------- |-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
tylerlevine, line 188: No need to add extra spaces to align these, they will automatically be spaced correctly in the rendered markdown.

tylerlevine (changes requested):
nice work. One other feedback that I can't easily leave on the PR is to remove the "Sign Tx API" from the arrow in the middle of the diagram - that's an internal implementation detail which isn't needed in order to understand how to run express in external signer mode.

modules/express/README.md
DariusParvin: Ah okay, somehow my IDE did this automatically. I'll try to undo it.

DariusParvin:
Thanks for the feedback @tylerlevine! I made the changes you suggested and edited the diagram. Let me know what you think.

PR #1988, opened 2022-02-10
feat: allow external signer to sign for multiple wallet ids

DariusParvin:
Previously the external signer would produce signatures using the single private key provided in the json file (specified by the config arg `signerFileSystemPath`). Now, it'll find the appropriate private key corresponding to the `walletId` for the transaction being signed and use that private key. 

Note: Only the last commit corresponds to this PR. The previous commits are from #1944 waiting to be merged


tylerlevine (approved):
🚀

PR #1989, opened 2022-02-10
refactor: address nits

DariusParvin:
addressing nits from #1944 and #2004

- remove external signing routes and comments for v1 wallet
- remove unecessary prototpe set in ExternalSignerConfigError
- add more descriptive debug msg for retries to external signer
- remove unnecessary `catch` block 

Ticket: BG-42470

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -115,7 +115,6 @@ function handleSendCoins(req: express.Request) {
     .wallets()
     .get({ id: req.params.id })
     .then(function (wallet) {
-      // signs a tx, needs custom signing function passed in
       return wallet.sendCoins(req.body);
     })
     .catch(function (err) {
@@ -139,7 +138,6 @@ function handleSendMany(req: express.Request) {
     .wallets()
     .get({ id: req.params.id })
     .then(function (wallet) {
-      // signs a tx, needs custom signing function passed in
       return wallet.sendMany(req.body);
     })
     .catch(function (err) {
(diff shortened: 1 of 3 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -114,7 +114,6 @@ function handleSendCoins(req: express.Request) {
     .wallets()
     .get({ id: req.params.id })
     .then(function (wallet) {
-      // signs a tx, needs custom signing function passed in
```
tylerlevine, line 117: Are we not supporting external signer mode for these functions where the comments are getting removed?

modules/express/src/errors.ts
```diff
@@ -40,10 +40,3 @@ export class IpcError extends Errors.BitGoJsError {
     Object.setPrototypeOf(this, IpcError.prototype);
   }
 }
-
-export class ExternalSignerConfigError extends Errors.BitGoJsError {
```
tylerlevine, line 44: I'm curious why we're getting rid of this error in favor of throwing the top level `Error` type instead?

tylerlevine (changes requested):
Good clean up - could you add a fix for the nit I mentioned in #2004 as well?

modules/express/src/clientRoutes.ts
DariusParvin: Yeah I believe that's the idea since they are all v1 wallet functions. Currently if signer mode is enabled, the v1 wallet functions would just operate the same as before.

modules/express/src/errors.ts
DariusParvin: Ah okay I realize now that I misunderstood this original comment. https://github.com/BitGo/BitGoJS/pull/1944#discussion_r798135161 

I believe you meant that just `Object.setPrototypeOf(this, ExternalSignerConfigError.prototype);` was unnecessary. I'll edit this commit.

DariusParvin:
rebased and addressed the comments including the nit from https://github.com/BitGo/BitGoJS/pull/2004

tylerlevine (approved):

PR #1992, opened 2022-02-11
chore(root): remove package level mocha

byhow:
The package-level mocha versions are not in sync, and a lot of them use the deprecated opts config file `mocha.opts`. In this PR, the package-level mocha will be replaced with a root-level mocha so that we can have an easier maintenance experience.

Ticket: BG-43009


modules/bls-dkg/.mocharc.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
```
byhow, line 1: For some reasons, the `verifies backup-wallet sig` suite keeps timing out on ci while running in <200ms on local. CI-wise, the tests that are not timed out often run ~1800ms, which is a big difference. Adding a timeout of 5000ms here

modules/blockapis/test/UtxoApi.ts
```diff
@@ -128,14 +128,13 @@ function runTestFetch(api: UtxoApi, coinName: string) {
...
       it('fetches resource', async function () {
-        this.timeout(10_000);
         if (testCase.expectError(api)) {
           assert.rejects(() => testCase.func(api));
           return;
         }
         const resource = await testCase.func(api);
         deepStrictEqualJSON(resource, await testCase.getFixture(api, resource));
-      });
+      }, 10_000);
```
byhow, line 137: Had to refactor the timeout here since `this` was not defined anymore under the package scope

modules/blockapis/test/UtxoApi.ts
```diff
@@ -1,3 +1,4 @@
+import 'mocha';
```
OttoAllmendinger, line 1: @byhow used an alternative solution here

it seems like typescript picks up the jasmine definitions from `describe`/`it` sometimes, unless we explicitly import mocha.

tylerlevine (changes requested):
Don't delete the express package-lock.json - this is needed when building for docker

modules/blockapis/test/UtxoApi.ts
byhow: Thanks for the pointer!

tylerlevine:
> @tylerlevine Thanks for pointing it out. I regenerated the lock file with `npm i` under the `modules/express` dir. Hopefully updating the lock file is desired. Otherwise, I can just add back the original lock

I don't think there's any need to modify the express package-lock at all, since mocha is not involved when creating the container image which uses that lock file. Can you please add back the original lock file from `master`?

byhow:
@tylerlevine Got it, makes sense! The original file was added back

tylerlevine (approved):

PR #1995, opened 2022-02-11
feat(static): dot metadata loaded as resource

jrafael-simtlix:
Metadata resource must not consume compiler memory.

STLX-13359


modules/statics/package.json
```diff
@@ -4,6 +4,9 @@
   "description": "dependency-free static configuration for the bitgo platform",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
+  "files": [
+    "/resources/**/*"
+  ],
```
jrafael-simtlix, line 9: @yangran0 Do you know why it doesn't add resource in dist?

PR #2004, opened 2022-02-15
feat: add retry logic to external signer

DariusParvin:
if the request to the external signer fails with `ECONNREFUSED`, retry the
request three times with exponentially increasing timeout up to 3 times.

Ticket: BG-43083

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -28,6 +28,7 @@ import { Config } from './config';
 import { ApiResponseError } from './errors';
 import { promises as fs } from 'fs';
 import * as assert from 'assert';
+import { retryPromise } from './retryPromise';
 
 const { version } = require('bitgo/package.json');
 const pjson = require('../package.json');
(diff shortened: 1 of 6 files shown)


modules/express/src/retryPromise.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Thrown in `retryPromise()`
+ */
+export class ErrorMaxTries extends Error {
+  constructor(tries: number) {
+    super(`giving up after ${tries} tries`);
```
lifeiscontent, line 6: when extending `Error` you must set the prototype otherwise it will break the prototype chain
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#support-for-newtarget

DariusParvin: Thanks for the feedback @lifeiscontent! Is this just a matter of adding the following line under `super()`?
```
Object.setPrototypeOf(this, new.target.prototype);
```

tylerlevine: @lifeiscontent that fix is no longer needed since we switched to targeting ES2019

modules/express/src/clientRoutes.ts
```diff
@@ -1025,11 +1026,16 @@ export function createCustomSigningFunction(externalSignerUrl: string): CustomSi
...
-      .send({ txPrebuild: params.txPrebuild, pubs: params.pubs });
-
+    const { body: signedTx } = await retryPromise(
+      () =>
+        superagent
+          .post(`${externalSignerUrl}/api/v2/${params.coin.getChain()}/sign`)
+          .type('json')
+          .send({ txPrebuild: params.txPrebuild, pubs: params.pubs }),
+      (err, tryCount) => {
+        console.error(`attempt number ${tryCount}`);
```
tylerlevine, line 1036: let's make this a debug instead and make it a little more informative
```suggestion
        debug(`failed to connect to external signer (attempt ${tryCount}, error: ${err.message})`);
```

tylerlevine (approved):
small nit, feel free to handle it in a follow up PR

modules/express/src/clientRoutes.ts
DariusParvin: added to https://github.com/BitGo/BitGoJS/pull/1989

PR #2009, opened 2022-02-15
feat(account-lib): token transfer support

bitgoandrew:


modules/account-lib/src/coin/sol/solInstructionFactory.ts
```diff
@@ -98,16 +99,40 @@ function memoInstruction(data: Memo): TransactionInstruction[] {
...
+    transferInstruction = splToken.Token.createTransferCheckedInstruction(
+      splToken.TOKEN_PROGRAM_ID,
+      new PublicKey(source),
+      new PublicKey(mint),
+      new PublicKey(toAddress),
+      new PublicKey(fromAddress),
+      [],
+      new BigNumber(amount).toNumber(),
+      decs,
+    );
```
joey-bai, line 134: what's the difference between `source` and `fromAddress`?

modules/account-lib/src/coin/sol/solInstructionFactory.ts
```diff
@@ -98,16 +99,40 @@ function memoInstruction(data: Memo): TransactionInstruction[] {
...
+    transferInstruction = SystemProgram.transfer({
+      fromPubkey: new PublicKey(fromAddress),
+      toPubkey: new PublicKey(toAddress),
+      lamports: new BigNumber(amount).toNumber(),
+    });
+  } else {
+    // token transfer
+    assert(mint, 'Missing mint');
+    assert(source, 'Missing source');
+    let decs;
```
joey-bai, line 119: type this out: `decimalPlaces`

modules/account-lib/src/coin/sol/transferBuilder.ts
```diff
@@ -12,6 +12,8 @@ import assert from 'assert';
 export interface SendParams {
   address: string;
   amount: string;
+  mint?: string;
+  source?: string;
```
joey-bai, line 16: not obvious what `source` is for - could you add a comment to explain what this is used for? and why it needs to be different per recipient address?

modules/account-lib/src/coin/sol/solInstructionFactory.ts
bitgoandrew: my mistake, fromAddress IS the source address - fixed this by keeping the sender (ata addr for tokens) called fromAddress, to be compatable with native transfers, but using an optional owner field in Transfer (instead of fromAddress & source, as proper use of this would be duplicate fields)

modules/account-lib/src/coin/sol/iface.ts
```diff
@@ -51,7 +51,13 @@ export interface WalletInit {
 
 export interface Transfer {
   type: InstructionBuilderTypes.Transfer;
-  params: { fromAddress: string; toAddress: string; amount: string };
+  params: {
+    fromAddress: string;
+    toAddress: string;
+    amount: string;
+    mint?: string;
+    owner?: string;
```
jimmyqi-bitgo, line 59: consider extending Transfer to something like TokenTransfer to have required fields mint and owner

modules/account-lib/src/coin/sol/instructionParamsFactory.ts
```diff
@@ -23,6 +23,8 @@ import {
   StakingWithdraw,
 } from './iface';
 import { getInstructionType } from './utils';
+import { decodeTransferCheckedInstruction } from './tokenEncodeDecode';
+const splToken = require('@solana/spl-token');
```
jimmyqi-bitgo, line 27: same thing, explore typescript imports

modules/account-lib/src/coin/sol/solInstructionFactory.ts
```diff
@@ -98,16 +99,40 @@ function memoInstruction(data: Memo): TransactionInstruction[] {
...
+      toPubkey: new PublicKey(toAddress),
+      lamports: new BigNumber(amount).toNumber(),
+    });
+  } else {
+    // token transfer
+    assert(mint, 'Missing mint');
+    assert(owner, 'Missing owner');
+    let decimalPlaces;
+    coins.forEach((value, key) => {
+      if (value['tokenAddress'] === mint) {
```
jimmyqi-bitgo, line 121: add a check here for whether the coin is solana, on the unlikely event that the same mint address is valid on multiple chains

modules/account-lib/src/coin/sol/transaction.ts
```diff
@@ -241,16 +242,32 @@ export class Transaction extends BaseTransaction {
...
+              address: instruction.params.fromAddress,
+              value: instruction.params.amount,
+              coin: this._coinConfig.name,
+            });
+            outputs.push({
+              address: instruction.params.toAddress,
+              value: instruction.params.amount,
+              coin: this._coinConfig.name,
+            });
+          }
```
jimmyqi-bitgo, line 269: duplicate code, refactor out coin name, eg.

```suggestion
            coinName = coin.name;
          } else {
            coinName = this._coinConfig.name;
          }
            inputs.push({
              address: instruction.params.fromAddress,
              value: instruction.params.amount,
              coin: coinName,
            });
            outputs.push({
              address: instruction.params.toAddress,
              value: instruction.params.amount,
              coin: coinName,
            });
          }
```

modules/account-lib/src/coin/sol/transferBuilder.ts
```diff
@@ -12,6 +12,8 @@ import assert from 'assert';
 export interface SendParams {
   address: string;
   amount: string;
+  mint?: string;
+  owner?: string;
```
jimmyqi-bitgo, line 16: same here, perhaps investigate subclassing

modules/account-lib/test/unit/coin/near/transaction.ts
```diff
@@ -53,7 +53,7 @@ describe('Near Transaction', () => {
     it('a signed transfer transaction', async () => {
       tx.fromRawTransaction(NearResources.rawTx.transfer.signed);
       const explain = tx.explainTransaction();
-      explain.id.should.equal('CEpsBC4fA64phQMfDnNzJqLFiDmgMhyAeu9vqN18EExH');
```
jimmyqi-bitgo, line 56: why was near impacted?

modules/account-lib/src/coin/sol/instructionParamsFactory.ts
```diff
@@ -113,15 +115,36 @@ function parseSendInstructions(instructions: TransactionInstruction[]): Array<No
...
+        let transfer: Transfer;
+        if (isToken) {
+          transfer = {
+            type: InstructionBuilderTypes.Transfer,
+            params: {
+              fromAddress: transferInstruction.keys.source.pubkey.toString(),
+              toAddress: transferInstruction.keys.destination.pubkey.toString(),
+              amount: transferInstruction.data.amount.toString(),
+              mint: transferInstruction.keys.mint.pubkey.toString(),
+              owner: transferInstruction.keys.owner.pubkey.toString(),
```
jimmyqi-bitgo, line 135: maybe call those mintAddress and ownerAddress for consistency

jimmyqi-bitgo (commented):
Nicely done!

modules/account-lib/test/unit/coin/near/transaction.ts
bitgoandrew: investigating and will revert this change when solved

modules/account-lib/src/coin/sol/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const transferData = this._sendParams.map((sendParams: SendParams): TokenTransfer => {
+      return {
+        type: InstructionBuilderTypes.TokenTransfer,
+        params: {
+          fromAddress: this._sender,
+          toAddress: sendParams.address,
+          amount: sendParams.amount,
+          mintAddress: sendParams.mintAddress,
+          ownerAddress: sendParams.ownerAddress,
```
joey-bai, line 78: We shouldn't need to maintain a separate `ownerAddress` property. this should be the same as `fromAddress`.

Also, let me know how you feel about this approach to support multi token transfers:
1. `this._sender` should always be the wallet's root address (also `ownerAddress`).
2. An ATA's address is deterministically derived from `(ownerAddress, mintAddress)` - this should give us the actual `fromAddress`.

joey-bai: Also double check this, but make sure the `feeAddress` is the native sol address (which i believe is taken care of automatically if you set it as `sender`)

bitgoandrew: yep, this is inside buildSolTransaction: `tx.feePayer = new PublicKey(this._sender);`!

modules/account-lib/src/coin/sol/solInstructionFactory.ts
```diff
@@ -111,6 +116,46 @@ function transferInstruction(data: Transfer): TransactionInstruction[] {
...
+    params: { fromAddress, toAddress, amount, mintAddress, sourceAddress },
+  } = data;
+  assert(fromAddress, 'Missing fromAddress (owner) param');
+  assert(toAddress, 'Missing toAddress param');
+  assert(amount, 'Missing amount param');
+  assert(mintAddress, 'Missing mint');
+  assert(sourceAddress, 'Missing ata address');
+  let decimalPlaces;
+  coins.forEach((value, key) => {
+    if (value['tokenAddress'] === mintAddress && value.family === 'sol') {
```
joey-bai, line 136: prefer `value.tokenAddress` if possible

modules/account-lib/src/coin/sol/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import assert from 'assert';
+import { TransactionBuilder } from './transactionBuilder';
+import { PublicKey } from '@solana/web3.js';
+import { ASSOCIATED_TOKEN_PROGRAM_ID, Token, TOKEN_PROGRAM_ID } from '@solana/spl-token';
+
+export interface SendParams {
+  address: string;
+  amount: string;
+  mintAddress: string;
```
joey-bai, line 17: I feel like we might have gotten this backwards - maybe we should have this be `tokenName` instead of mint address.
then we do something like `coins.from(tokenName)` to check if the token is supported or not.

then get token specific details (`mintAddress`, `decimals`, etc.)

OK to leave this as is since we're doing the same for creating ATAs. will think about this some more

modules/account-lib/test/unit/coin/sol/transactionBuilder/transferBuilder.ts
```diff
@@ -18,8 +18,8 @@ describe('Sol Transfer Builder', () => {
   const otherAccount = new KeyPair({ prv: testData.prvKeys.prvKey1.base58 }).getKeys();
   const invalidPubKey = testData.pubKeys.invalidPubKeys[0];
   const recentBlockHash = 'GHtXQBsoZHVnNFa9YevAzFr17DJjgHXk3ycTKD5xD3Zi';
-  const amount = '300000';
-  const memo = 'test memo';
+  const amount = testData.tokenTransfers.amount.toString();
+  const memo = testData.tokenTransfers.memo;
```
joey-bai, line 22: should we revert this? this test shouldn't be testing token transfers right?

modules/account-lib/test/unit/coin/sol/instructionParamsFactory.ts
```diff
@@ -76,6 +76,66 @@ describe('Instruction Parser Tests: ', function () {
...
+        },
+      };
+      const transferInstruction = Token.createTransferCheckedInstruction(
+        TOKEN_PROGRAM_ID,
+        new PublicKey(sourceORCA),
+        new PublicKey(mintORCA),
+        new PublicKey(nonceAccount),
+        new PublicKey(ownerORCA),
+        [],
+        +amount,
```
joey-bai, line 117: `+amount` - is the `+` intentional?

joey-bai (approved):

modules/account-lib/src/coin/sol/tokenTransferBuilder.ts
bitgoandrew: Good one I like that - either store fields we need in the Transfer object or store whole coin on the transaction object

PR #2010, opened 2022-02-15
fix(statics): Fix Solana transactions explorers

silalvado:
Fix Solana's explorers to access transactions' details.

Jira ticket: https://bitgoinc.atlassian.net/browse/STLX-13338

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -389,13 +389,13 @@ class StellarTestnet extends Testnet implements AccountNetwork {
 class Sol extends Mainnet implements AccountNetwork {
   name = 'Sol';
   family = CoinFamily.SOL;
-  explorerUrl = 'https://explorer.solana.com/';
+  explorerUrl = 'https://explorer.solana.com/tx/';
 }
 
 class SolTestnet extends Testnet implements AccountNetwork {
   name = 'SolTestnet';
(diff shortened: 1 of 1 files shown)


rliberal-simtlix (approved):

alebusse (dismissed):

modules/statics/src/networks.ts
```diff
@@ -389,13 +389,13 @@ class StellarTestnet extends Testnet implements AccountNetwork {
...
   family = CoinFamily.SOL;
-  explorerUrl = 'https://explorer.solana.com/';
+  explorerUrl = 'https://explorer.solana.com/tx/';
 }
 
 class SolTestnet extends Testnet implements AccountNetwork {
   name = 'SolTestnet';
   family = CoinFamily.SOL;
-  explorerUrl = 'https://explorer.solana.com/?cluster=devnet';
+  explorerUrl = 'https://explorer.solana.com/tx/?cluster=devnet';
```
joey-bai, line 398: how will this eventually be used in Lumina?
do we have enough info to preserve the query parameter (`?cluster=devnet`) but also append the txid ?

silalvado: In Lumina we are splitting the URL when it has a query param and adding the transaction id in the middle.

PR #2015, opened 2022-02-15
fix: fix build

joey-bai:
fixes build by pinning solana dependency version and disabling dot tests

Ticket: BG-00000


alebusse:
@joey-bai Updated the versions to the latest working i found.

lcovar (approved):

PR #2017, opened 2022-02-16
Luis/dot hex sign

lcovar:
The purpose of this PR is to add support to add a signature to a dot transaction so it can support TSS signatures. 

The signature is expected to be in a buffer format and will return a hex string signature after the transaction is built since the polkadot library returns a hex signature in their signing implementation. TSS signatures will not have `0x` prepended therefore the returned signatures returned from the builder will not have `0x` prepended either.

### DIFF
--- a/modules/account-lib/src/coin/dot/transaction.ts
+++ b/modules/account-lib/src/coin/dot/transaction.ts
@@ -18,8 +18,10 @@ import {
   AddAnonymousProxyArgs,
   BatchArgs,
   WithdrawUnstakedArgs,
+  HexString,
 } from './iface';
 import utils from './utils';
+import { u8aToBuffer } from '@polkadot/util';
 
 export class Transaction extends BaseTransaction {
   protected _dotTransaction: UnsignedTransaction;
@@ -36,14 +38,15 @@ export class Transaction extends BaseTransaction {
   canSign({ key }: BaseKey): boolean {
     const kp = new KeyPair({ prv: key });
     const addr = kp.getAddress();
-    if (addr === this._sender) {
-      return true;
-    }
-    return false;
+    return addr === this._sender;
   }
 
-  /** @inheritdoc */
-  async sign(keyPair: KeyPair): Promise<void> {
+  /**
+   * Sign a polkadot transaction and update the transaction hex
+   *
+   * @param {KeyPair} keyPair - ed signature
+   */
+  sign(keyPair: KeyPair): void {
     if (!this._dotTransaction) {
       throw new InvalidTransactionError('No transaction data to sign');
     }
@@ -64,6 +67,8 @@ export class Transaction extends BaseTransaction {
       registry: this._registry,
     });
 
+    // get signature from signed txHex generated above
+    this._signatures = [utils.recoverSignatureFromRawTx(txHex, { registry: this._registry })];
     this._signedTransaction = txHex;
   }
 
@@ -352,8 +357,40 @@ export class Transaction extends BaseTransaction {
     }
   }
 
+  /**
+   * Constructs a signed payload using construct.signTx
+   * This method will be called during the build step if a TSS signature
+   * is added and will set the signTransaction which is the txHex that will be broadcasted
+   * As well as add the signature used to sign to the signature array in hex format
+   *
+   * @param {Buffer} signature The signature to be added to a dot transaction
+   */
+  constructSignedPayload(signature: Buffer): void {
+    // 0x00 means its an ED25519 signature
+    const edSignature = `0x00${signature.toString('hex')}` as HexString;
+
+    try {
+      this._signedTransaction = construct.signedTx(this._dotTransaction, edSignature, {
+        registry: this._registry,
+        metadataRpc: this._dotTransaction.metadataRpc,
+      });
+    } catch (e) {
+      throw new SigningError(`Unable to sign dot transaction with signature ${edSignature} ` + e);
+    }
+
+    this._signatures = [signature.toString('hex')];
+  }
+
   setTransaction(tx: UnsignedTransaction): void {
     this._dotTransaction = tx;
+  }
+
+  /** @inheritdoc **/
+  get signablePayload(): Buffer {
+    const extrinsicPayload = this._registry.createType('ExtrinsicPayload', this._dotTransaction, {
+      version: this._dotTransaction.version,
+    });
+    return u8aToBuffer(extrinsicPayload.toU8a({ method: true }));
   }
 
   /**
--- a/modules/account-lib/src/coin/dot/transactionBuilder.ts
+++ b/modules/account-lib/src/coin/dot/transactionBuilder.ts
@@ -2,12 +2,19 @@ import { BaseCoin as CoinConfig, PolkadotSpecNameType } from '@bitgo/statics';
 import { UnsignedTransaction } from '@substrate/txwrapper-core';
 import { DecodedSignedTx, DecodedSigningPayload, TypeRegistry } from '@substrate/txwrapper-core/lib/types';
 import { decode, getRegistry } from '@substrate/txwrapper-polkadot';
-import * as _ from 'lodash';
+import _ from 'lodash';
 import BigNumber from 'bignumber.js';
 import { isValidEd25519Seed } from '../../utils/crypto';
-import { BaseTransactionBuilder, TransactionType } from '../baseCoin';
+import { BaseTransactionBuilder, TransactionType, Interface } from '../baseCoin';
 import { BuildTransactionError, InvalidTransactionError } from '../baseCoin/errors';
-import { BaseAddress, BaseKey, FeeOptions, SequenceId, ValidityWindow } from '../baseCoin/iface';
+import {
+  BaseAddress,
+  BaseKey,
+  FeeOptions,
+  SequenceId,
+  ValidityWindow,
+  PublicKey as BasePublicKey,
+} from '../baseCoin/iface';
 import { AddressValidationError, InvalidFeeError } from './errors';
 import { CreateBaseTxInfo, Material, TxMethod } from './iface';
 import { KeyPair } from './keyPair';
@@ -28,6 +35,10 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
   protected _registry: TypeRegistry;
   protected _method?: TxMethod;
   protected __material?: Material;
+  // signatures that will be used to sign a transaction when building
+  // not the same as the _signatures in transaction which is the signature in
+  // string hex format used for validation after we call .build()
+  protected _signatures: Interface.Signature[] = []; // only support single sig for now
 
   constructor(_coinConfig: Readonly<CoinConfig>) {
     super(_coinConfig);
@@ -184,6 +195,8 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
       this.referenceBlock(decodedTxn.blockHash);
     } else {
       this.sender({ address: utils.decodeDotAddress(decodedTxn.address) });
+      const edSignature = utils.recoverSignatureFromRawTx(rawTransaction, { registry: this._registry });
+      this._transaction.signature.push(edSignature);
     }
     this.validity({ maxDuration: decodedTxn.eraPeriod });
     this.sequenceId({
@@ -207,7 +220,12 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     if (this._keyPair) {
       this.transaction.sign(this._keyPair);
     }
+    if (this._signatures?.length > 0) {
+      // if we have a signature, apply that and update this._signedTransaction
+      this.transaction.constructSignedPayload(this._signatures[0].signature);
+    }
     this._transaction.loadInputsAndOutputs();
+
     return this._transaction;
   }
 
@@ -371,6 +389,10 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
   }
 
   // endregion
+  /** @inheritdoc */
+  addSignature(publicKey: BasePublicKey, signature: Buffer): void {
+    this._signatures.push({ publicKey, signature });
+  }
 
   /** @inheritdoc */
   protected signImplementation({ key }: BaseKey): Transaction {
--- a/modules/account-lib/src/coin/dot/utils.ts
+++ b/modules/account-lib/src/coin/dot/utils.ts
@@ -12,10 +12,10 @@ import base32 from 'hi-base32';
 import { KeyPair } from '.';
 import { BaseUtils } from '../baseCoin';
 import { Seed } from '../baseCoin/iface';
-import { ProxyCallArgs, TransferArgs, ProxyArgs, TxMethod, Material } from './iface';
+import { HexString, Material, ProxyArgs, ProxyCallArgs, TransferArgs, TxMethod } from './iface';
 import nacl from 'tweetnacl';
 import { BaseCoin as CoinConfig, DotNetwork } from '@bitgo/statics';
-import { createTypeUnsafe, GenericExtrinsicPayload, GenericCall, GenericExtrinsic } from '@polkadot/types';
+import { createTypeUnsafe, GenericCall, GenericExtrinsic, GenericExtrinsicPayload } from '@polkadot/types';
 
 const PROXY_METHOD_ARG = 2;
 export class Utils implements BaseUtils {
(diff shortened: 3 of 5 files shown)


modules/account-lib/src/coin/dot/transaction.ts
```diff
@@ -346,10 +348,30 @@ export class Transaction extends BaseTransaction {
...
 
+  constructSignedPayload(signature: Buffer): void {
+    const edSignature = `0x00${signature.toString('hex')}` as HexString;
+
+    this._signedTransaction = construct.signedTx(this._dotTransaction, edSignature, {
+      registry: this._registry,
+      metadataRpc: this._dotTransaction.metadataRpc,
+    });
+
+    const recoveredSignature = utils.recoverSignatureFromRawTx(this._signedTransaction, { registry: this._registry });
```
byhow, line 359: Is this the same as `${signature.toString('hex')}`?

lcovar: It should be, but I figured that the best way to extract the signature is from the signed transaction itself rather than simply adding the signature provided from addSignature() to the list of signatures. Ideally it should  always be the same but though i should get it from the best source of truth and figured it would be from the signed transaction itself.

let me know what you think.

byhow: Hmm if we are confident that these two will always have the same value, it feels to me like duplicated computation here (since looks like`recoveredSignature` is expected to be the same as `edSignature`) But if we are unsure, maybe we can run an assert and see if we hit situations that they they are different, and deprecate the recoverySignature in the future once we are certain. Like:
```ts
assert(recoveredSignature === edSignature, new Error("derived signature is different from provided signature");
this._signature = signature.toString('hex');
```
or something

byhow: or we can run this case in unit tests

lcovar: Actually it's not expected to be the same as edSignature since edSignature prepends 0x00. 

The output of recoveredSignature = signature.toString('hex'). Test pass if I do 

this._signature = signature.toString('hex');

Im going to remove the double computation and just do the above instead.

modules/account-lib/src/coin/dot/transaction.ts
```diff
@@ -352,10 +353,29 @@ export class Transaction extends BaseTransaction {
     }
   }
 
+  constructSignedPayload(signature: Buffer): void {
```
argjv, line 356: nit: JS docs

modules/account-lib/src/coin/dot/transaction.ts
```diff
@@ -352,10 +353,29 @@ export class Transaction extends BaseTransaction {
...
+    const edSignature = `0x00${signature.toString('hex')}` as HexString;
+
+    this._signedTransaction = construct.signedTx(this._dotTransaction, edSignature, {
```
argjv, line 359: Can we validate `edSignature` is valid before using it? Or will this throw if it isn't? If it throws, is the error descriptive enough to guide the user on what failed? As in, maybe we can capture the error and print the tx and signature if it is not printed already

modules/account-lib/src/coin/dot/transaction.ts
```diff
@@ -352,10 +353,29 @@ export class Transaction extends BaseTransaction {
...
+
+    this._signatures = [signature.toString('hex')];
+  }
+
   setTransaction(tx: UnsignedTransaction): void {
     this._dotTransaction = tx;
   }
 
+  /** @inheritdoc **/
+  get signablePayload(): Buffer {
```
argjv, line 372: This is what you should get from `toBroadcastFormat()` but in string hex. How is this different?

modules/account-lib/src/coin/dot/utils.ts
```diff
@@ -231,6 +236,26 @@ export class Utils implements BaseUtils {
...
     return (arg as TransferArgs).dest?.id !== undefined && (arg as TransferArgs).value !== undefined;
   }
+
+  checkValidHex(str: string): asserts str is HexString {
+    if (!str.startsWith(`0x`)) {
+      throw new Error(`This is not in correct format of hex string: ${str}`);
+    }
+  }
+
+  recoverSignatureFromRawTx(rawTx: string, options: { registry: TypeRegistry }): string {
```
byhow, line 246: ditto on jsDoc

modules/account-lib/src/coin/dot/transactionBuilder.ts
```diff
@@ -205,9 +215,15 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     this.transaction.registry(this._registry);
     this.transaction.chainName(this._material.chainName);
     if (this._keyPair) {
-      this.transaction.sign(this._keyPair);
+      await this.transaction.sign(this._keyPair);
+    }
+    // TSS signature added
```
argjv, line 220: This can be any signature created offline, not just from TSS

modules/account-lib/src/coin/dot/transactionBuilder.ts
```diff
@@ -28,6 +35,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
   protected _registry: TypeRegistry;
   protected _method?: TxMethod;
   protected __material?: Material;
+  protected _signatures: Interface.Signature[] = []; // only support single sig for now
```
byhow, line 38: We might want to distinguish this `signatures` from the one we had in `transaction.ts`, one is for builder's functionality and the other is something else

modules/account-lib/src/coin/dot/transactionBuilder.ts
```diff
@@ -372,6 +388,20 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
 
   // endregion
 
+  /**
+   * Adds a signature to the transaction.
+   *
+   * @param publicKey public key that produced the signature
+   * @param signature raw signature as a hex encoded Buffer
+   */
+  addSignature(publicKey: BasePublicKey, signature: Buffer): void {
```
argjv, line 397: This builder step should be in the base interface

modules/account-lib/src/coin/dot/transactionBuilder.ts
```diff
@@ -372,6 +388,20 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+    this._signatures.push({ publicKey, signature });
+  }
+
+  get signatures(): Interface.Signature[] {
```
argjv, line 401: ditto, if it isn't in the interface already

modules/account-lib/src/coin/dot/transaction.ts
lcovar: toBroadcastFormat is actually a serialized signature or txHex.

Signable payload is portion that actually needs to be signed. This is what the HSM will sign.

modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts
```diff
@@ -178,4 +181,183 @@ xdescribe('Dot Transfer Builder', () => {
...
+      const transferBuilder2 = factory
+        .getTransferBuilder()
+        .amount('90034235235322')
+        .sender({ address: sender.address })
+        .to({ address: receiver.address })
+        .validity({ firstValid: 3933, maxDuration: 64 })
+        .referenceBlock('0x149799bc9602cb5cf201f3425fb8d253b2d4e61fc119dcab3249f307f594754d')
+        .sequenceId({ name: 'Nonce', keyword: 'nonce', value: 200 })
+        .fee({ amount: 0, type: 'tip' });
+      transferBuilder2.addSignature({ pub: accounts.account1.publicKey }, Buffer.from(signature, 'hex'));
```
argjv, line 221: Since the signature is an attribute used to construct the transaction, there is no need for the `add` prefix, you can simply to `builder.signature(...)`

argjv, line 221: Why not concatenate the signature addition to the previous builder step (`fee()`)?

modules/account-lib/src/coin/dot/transactionBuilder.ts
byhow: This method was implemented as an async method so `await` was added, but looking back at that `transaction.sign`, there's nothing really async going on there. Maybe let's just remove the `async` modifier from `transaction.ts` and also the `@inheritDoc` too, doesn't seem to inherit any methods from the BaseTransaction

modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts
lcovar: I can do this but sol is using `addSignature` and would prefer to stay consistent. 

I can update this for sol/dot in a follow up PR.

modules/account-lib/src/coin/dot/transactionBuilder.ts
lcovar: Do you mean this?

https://github.com/BitGo/BitGoJS/blob/00cd5662bf9f89c9c4bdab948f6548107c9ef696/modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts#L61

argjv, line 397: I think the signature should be: `signature(publicKey: BasePublicKey, signature: string)` and then you should validate both parameters are valid pubs and hex strings.

Also, `publicKey: BasePublicKey` makes it unnatural to pass the pub since you have to craft an object like `{ pub: ... }` instead of passing the pub. How can this be fixed?

lcovar: Honestly, we have no use for the public key in our addSignature implementation. It's needed for sol when adding a signature but not dot. 

Not sure what the best approach here is.

modules/account-lib/src/coin/dot/transactionBuilder.ts
lcovar: added a comment. lmk if you think that's enough

modules/account-lib/src/coin/dot/transactionBuilder.ts
lcovar: Removed the await but it does inherit this method

[src/coin/dot/transactionBuilder.ts](https://github.com/BitGo/BitGoJS/blob/00cd5662bf9f89c9c4bdab948f6548107c9ef696/modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts#L84-L90)

byhow: Ah my bad, let me clarify, it is the `sign` method, doesnt see it in the `BaseTransaction` class: 

https://github.com/BitGo/BitGoJS/blob/master/modules/account-lib/src/coin/dot/transaction.ts#L46

modules/account-lib/src/coin/dot/transactionBuilder.ts
byhow: yeah it looks good!

modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts
lcovar: It looks like addSignature is consistent with the behavior of `sign()` where it returns void so it has it can't be concatenated like so because it will throw an error. Has to be standalone because of the return type.

```
.
.
.fee({ amount: 0, type: 'tip' })
.addSignature({ pub: accounts.account1.publicKey }, Buffer.from(signature, 'hex')).build;
```

lcovar: @joey-bai - heads up, im gonna be making this change in a follow up

modules/account-lib/src/coin/dot/transactionBuilder.ts
lcovar: realized this was not needed. we were already using `this._signatures` to access this like the other properties in this file.

Removed the method

modules/account-lib/src/coin/sol/transactionBuilder.ts
```diff
@@ -15,7 +15,6 @@ import { instructionParamsFactory } from './instructionParamsFactory';
 
 export abstract class TransactionBuilder extends BaseTransactionBuilder {
   private _transaction: Transaction;
-  private _signatures: Interface.Signature[] = [];
```
byhow, line 18: Updated the access modifier, might need Joey for the solana review

byhow: Actually we can't really make changes to move this field to the base transaction builder. Some other coins already have different implementations of how they treat their signatures:

image

Will need to remove this

lcovar: got it. dropped the commit. We can access the field using `this._signatures`

modules/account-lib/src/coin/dot/transaction.ts
lcovar: Good idea.
Not sure if we can validate it before we actually use it to sign. but instead I added a try catch to throw a more descriptive error message since the error it displayed wasn't descriptive enough. 

`ex: Unable to create Enum via index 124, in Ed25519, Sr25519, Ecdsa
`
Updated to this:

 ```
  try {
      this._signedTransaction = construct.signedTx(this._dotTransaction, edSignature, {
        registry: this._registry,
        metadataRpc: this._dotTransaction.metadataRpc,
      });
    } catch (e) {
      throw new SigningError(`Unable to sign dot transaction with signature ${edSignature} ` + e);
    }
```

File: modules/account-lib/src/coin/dot/transactionBuilder.ts (hunk 8 above)
Comment (lcovar) [reply modules/account-lib/src/coin/dot/transactionBuilder.ts:393]: I think it's fine leaving at as a buffer and we can convert it to hex when we do the actual signing. We know that if we do `signature.toString('hex')` that it will convert to a valid hex. 

We will sign with this hex and I now updated the code where we sign to throw a more descriptive message if the signing fails.

I think we need to give this addSignature method some more though. I can revisit this when I make the changes to sol/dot and to simply use `signature` instead of `addSignature`. Im thinking we can overload this method as well and have one with just a signature since we dont need the public key.

File: modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts (hunk 10 above)
Comment (joey-bai) [reply modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts:221]: @argjv @lcovar i wanted this to be `addSignature` because Solana supports multiple signatures. so rather than setting a single `signature` field, we're _adding_ signatures with each call to `addSignature`.

this is needed when creating staking accounts and durable nonce accounts.

File: modules/account-lib/src/coin/dot/transactionBuilder.ts (hunk 8 above)
Comment (joey-bai) [reply modules/account-lib/src/coin/dot/transactionBuilder.ts:393]: @argjv @lcovar i don't feel too strongly about this one way or the other.

i'm open to changing `publicKey` to type `string` since it makes the code a little easier.

also open to changing `signature: Buffer` to `signatureHex: string`. each coin and their SDK will likely expect a slightly different format. the overhead of re-encoding isn't a concern. just need to be extremely clear what the input is.

File: modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts (hunk 10 above)
Comment (joey-bai) [reply modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts:221]: but i suppose, Solana's implementation can treat this as appending signatures.
other coins can treat this as a single signature.

Comment (argjv) [reply modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts:221]: Sounds like we are talking about the same. Instead of:

```
transferBuilder2.addSignature({ pub: accounts.account1.publicKey }, Buffer.from(signature1, 'hex'));
transferBuilder2.addSignature({ pub: accounts.account2.publicKey }, Buffer.from(signature2, 'hex'));
transferBuilder2.addSignature({ pub: accounts.account3.publicKey }, Buffer.from(signature3, 'hex'));
```

You would have:

```
transferBuilder2.signature({ pub: accounts.account1.publicKey }, Buffer.from(signature1, 'hex'));
transferBuilder2.signature({ pub: accounts.account2.publicKey }, Buffer.from(signature2, 'hex'));
transferBuilder2.signature({ pub: accounts.account3.publicKey }, Buffer.from(signature3, 'hex'));
```

My change request for the name is because `addSignature` is a verb but what you are doing is setting a field just like `amount`, `sender`, `fee`, etc, so `signature` would be more proper naming. Otherwise, we would also do `addAmount`, `addSender`, `addFee`

argjv: Sounds like a bug in `sign()` :/  anyway, outside this PR scope

modules/account-lib/src/coin/dot/transactionBuilder.ts
argjv: You should always be able to verify a signature if you have the pub key. We can add a `TODO` and figure it out in another PR.
Another option is to make it optional (preferred).

Regarding the buffer vs string hex, we've preferred strings in the past because they are easier to read and pass around as data. It saves you from having to deal with extra libs too (in your tests you are doing `Buffer.from(signature, 'hex')` to be able to use this method when the string hex is what you are storing because it is easier)

argjv (approved):

modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts
lcovar: ^ going to have a refactor PR open by EOW.

PR #2020, opened 2022-02-16
feat(sdk): dynamic load dot material

aseoane-simtlix:
After [STLX-11856](https://bitgoinc.atlassian.net/browse/STLX-11856), material can be loaded dynamically.

In [STLX-12105](https://bitgoinc.atlassian.net/browse/STLX-12105) we added a new endpoint that exposes the material data.

Add that endpoint to the coin and load it into the factory method.

Since  transaction material only changes with  a new release, the first time that factory is called it should load the data. 
To refresh client need to be reload.

Jira ticket: [STLX-12110](https://bitgoinc.atlassian.net/browse/STLX-12110)


modules/core/src/v2/coins/dot.ts
```diff
@@ -238,7 +252,10 @@ export class Dot extends BaseCoin {
    */
   async signTransaction(params: SignTransactionOptions): Promise<SignedTransaction> {
     const { txHex, prv } = this.verifySignTransactionParams(params);
-    const factory = accountLib.register(this.getChain(), accountLib.Dot.TransactionBuilderFactory);
+    
+    const material = await this.materialDataLookup();
```
lcovar, line 256: The only thing Im not 100% sure on here is the fact that  we're calling an endpoint when signing a transaction. I believe OVC will error out here when it tries to sign.

Do I have that right @Logicwax ?

Logicwax: OVC is offline.  And so cannot do this.

Logicwax: OVC is offline.  And so cannot do this.

aseoane-simtlix:
closed in favor of [2052](https://github.com/BitGo/BitGoJS/pull/2052)

PR #2021, opened 2022-02-16
fix(core): rename verifyAddress and remove invalid implementations

OttoAllmendinger:
Rename to `isWalletAddress()` and throw exceptions where it is not
implemented correctly.

Issue: BG-43225

### DIFF
--- a/modules/core/src/v2/coins/stx.ts
+++ b/modules/core/src/v2/coins/stx.ts
@@ -3,7 +3,6 @@
  */
 import * as accountLib from '@bitgo/account-lib';
 import { BaseCoin as StaticsBaseCoin, CoinFamily } from '@bitgo/statics';
-import { c32addressDecode } from 'c32check';
 
 import {
   BaseCoin,
@@ -17,7 +16,7 @@ import {
   TransactionPrebuild as BaseTransactionPrebuild,
 } from '../baseCoin';
 import { BitGo } from '../../bitgo';
-import { InvalidAddressError } from '../../errors';
+import { MethodNotImplementedError } from '../../errors';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 interface SupplementGenerateWalletOptions {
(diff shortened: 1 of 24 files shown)


modules/core/src/v2/coins/stx.ts
```diff
@@ -91,20 +90,8 @@ export class Stx extends BaseCoin {
...
-      const versionString = accountLib.Stx.AddressVersion[version];
-      if (versionString === undefined) throw new InvalidAddressError(`invalid address version: ${address}`);
-    } catch (e) {
-      throw new InvalidAddressError(`invalid address: ${address}`);
-    }
-
-    return true;
+  isWalletAddress(params: VerifyAddressOptions): boolean {
+    throw new MethodNotImplementedError();
   }
```
OttoAllmendinger, line 95: not sure what this was doing but it doesn't look wallet-related

modules/core/test/v2/unit/coins/sol.ts
```diff
@@ -254,9 +253,9 @@ describe('SOL:', function () {
   });
 
 
-  it('should verify valid address', (function () {
-    const params = { address: goodAddresses[0] };
-    basecoin.verifyAddress(params).should.equal(true);
```
sangaman, line 259: Do we want to have this test use `isValidAddress` instead of just scrapping it?

modules/core/test/v2/unit/coins/stx.ts
```diff
@@ -102,10 +94,10 @@ describe('STX:', function () {
     goodAddresses.map(addr => { basecoin.isValidAddress(addr).should.equal(true); });
   }));
 
-  it('should verify addresses', (function () {
-    goodAddresses.map(address => { basecoin.verifyAddress({ address }).should.equal(true); });
```
sangaman, line 106: Same question here.

sangaman (approved):
Good catch, the old naming was confusing and didn't line up with what the comment block said the function actually did so it's not surprising it was implemented incorrectly in several places. My only concern would be whether anyone is currently using `verifyAddress` for the purpose of validating addresses (for which they ought to be using `isValidAddress`) I take it. But I see this is appropriately marked as a `BREAKING CHANGE`.

modules/core/test/v2/unit/coins/stx.ts
OttoAllmendinger: in this case here we actually have tests for `isValidAddress` a few lines above

sangaman (approved):
LGTM

tylerlevine (approved):
excellent work - will push to get these methods implemented

PR #2029, opened 2022-02-17
feat(root): add unit-test-all to ci

byhow:
## Motivation
On package updates, the current CI won't prevent breaking changes with running all unit tests but instead relies on running unit test diff. This is what we encountered in the Solana and Dot tests updates, where it broke some tests. With package updates, most likely there will not be new test cases introduced and it may risk breaking the code. 

## What is included
- A new workflow that runs all unit tests if there are changes involved in any of the `package.json`

## Testing
An additional commit was introduced to purposedly trigger the workflow in the PR, and will be borked before merging

### Why not run integration tests too?
I was going to add `integration` tests into the run options too but seems like they are mostly obsolete.

### Why not just add a new job?
There isn't really a quick native way to run jobs on file changes without relying on 3rd party actions. Reference: https://github.community/t/is-it-possible-to-run-the-job-only-when-a-specific-file-changes/115484/10

## Misc
The test drive `package.json` update commit will need to be borked before merging

NEW TICKET: BG-43414

### DIFF
--- a/.github/workflows/test_package_updates.yml
+++ b/.github/workflows/test_package_updates.yml
@@ -1,0 +1,52 @@
+name: Package updates
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "*/**/package.json"
+      - "!*/**/node_modules/package.json"
+      - "package.json"
+      - "!node_modules/package.json"
+  pull_request:
+    branches:
+      - master
+      - rel/**
+    paths:
+      - "*/**/package.json"
+      - "!*/**/node_modules/package.json"
+      - "package.json"
+      - "!node_modules/package.json"
+
+jobs:
+  unit-test-all:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x, 16.x]
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Setup node:${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Fetch Base Ref
+        run: git fetch origin $GITHUB_BASE_REF
+
+      - name: Remove yarn.lock to simulate downstream installs
+        run: rm yarn.lock
+
+      - name: Install Packages
+        run: yarn install
+
+      - name: Unit test all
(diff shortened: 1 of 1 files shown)


tylerlevine (dismissed):
seems pretty reasonable to me as long as we're not running this all the time.

It might also be good to run this without any lockfiles present - that would replicate the versions that a client would get if they installed the package from npm. That's not in scope here though but could be done as a separate change

byhow:
@tylerlevine Thanks for the review! I need to revert the intentional package update to show/test the triggered ci, so I applied the changes you suggested to remove the lock file as well. Hopefully it looks good now!

tylerlevine (changes requested):
sorry I might be blind but where are the lock files getting removed as you said?

byhow:
@tylerlevine Ah sorry I thought it was to remove `yarn.lock` from the checklist to still skip the unit tests when changes apply to that file. If you meant to remove the `--with-frozen-lockfile` flag, I can remove that and put back the yarn lock file.

tylerlevine:
> @tylerlevine Ah sorry I thought it was to remove `yarn.lock` from the checklist to still skip the unit tests when changes apply to that file. If you meant to remove the `--with-frozen-lockfile` flag, I can remove that.

no, what I meant was a CI run that does `rm yarn.lock` before installing and running the tests. This would create an install which contains the same set of packages that a downstream user of the packages would get (since they install the latest allowable version of each package, as determined by their semver ranges).

byhow:
Thanks @tylerlevine for the further clarification. I added the `rm yarn.lock` step before the install, hopefully that addresses your comment. I also test triggered the ci by modifying some arbitrary `package.json` in `modules` as well.

.github/workflows/test_package_updates.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Setup node:${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Fetch Base Ref
+        run: git fetch origin $GITHUB_BASE_REF
+
+      - name: Remove yarn.lock to simulate downstream installs
+        run: rm yarn.lock
```
byhow, line 44: the remove lock file step here ^

tylerlevine (approved):
thanks for doing this @byhow!

PR #2034, opened 2022-02-21
refactor(account-lib): implement registry singleton

nachofm:
STLX-13615
Although this is a good practice and should be implemented, we have only seen a small decrease in memory usage. All the tests have been run on my local machine and the results are:

- Memory usage with the implementation of the Singleton Pattern for registry: 1392 MB
- Memory usage without the implementation of the Singleton Pattern for registry: 1457 MB


modules/account-lib/test/unit/coin/dot/utils.ts
```diff
@@ -75,4 +78,25 @@ describe('utils', () => {
...
+      rawTx.transfer.unsigned,
+      signature.replace(/^/, '0x00'),
+      material.metadata as `0x${string}`,
+      registry,
+    );
+    should.equal(
+      txHex,
+      '0xb9018400000000000000000000000000000000000000000000000000000000000000000000b6d868a11d202b56df1959f5d5f81f44ce1f95c8e70424b17080ea869d1c39d453f16c38fbef600a636c9a62a49ede5ee695a1822faf2f94fcfbb184a4254009d501210300000000000000',
+    );
+  });
```
jrafael-simtlix, line 101: This is for other PR!

modules/account-lib/test/unit/coin/dot/transactionBuilder/transactionBuilderFactory.ts
```diff
@@ -25,7 +25,6 @@ xdescribe('dot Transaction Builder Factory', () => {
   const factory = register('tdot', StubTransactionBuilderFactory).material(materialData as Material);
   const sender = accounts.account1;
   const sender2 = accounts.account3;
-
```
jrafael-simtlix, line 28: useless changes!

modules/account-lib/src/coin/dot/utils.ts
```diff
@@ -162,16 +161,26 @@ export class Utils implements BaseUtils {
...
+  /**
+   * Serializes the signed transaction
+   *
+   * @param transaction Transaction to serialize
+   * @param signature Signature of the message
+   * @param metadataRpc Network metadata
+   * @param registry Transaction registry
+   * @returns string Serialized transaction
+   */
+  serializeSignedTransaction(transaction, signature, metadataRpc, registry) {
```
jrafael-simtlix, line 183: this is for other PR

modules/account-lib/src/coin/dot/transactionBuilder.ts
```diff
@@ -157,12 +158,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
    */
   material(material: Material): this {
     this.__material = material;
-    this._registry = getRegistry({
-      chainName: material.chainName,
-      specName: material.specName,
-      specVersion: material.specVersion,
-      metadataRpc: material.metadata,
-    });
+    this._registry = SingletonRegistry.getInstance(material);
```
byhow, line 161: How does this reduce the memory usage for the transaction builder method? Correct me if I am wrong, but aren't we still storing this in an instance the same way we do here ^? Still having 2 materials stored in this builder

modules/account-lib/src/coin/dot/utils.ts
```diff
@@ -109,8 +109,7 @@ export class Utils implements BaseUtils {
     tx: string | UnsignedTransaction,
     options: { metadataRpc: string; registry: TypeRegistry },
   ): TransferArgs {
-    const { metadataRpc, registry } = options;
-    registry.setMetadata(createMetadata(registry, metadataRpc));
```
byhow, line 113: Why have we changed the setMetadata here? Based on the comments, it seems crucial that we set metadata before signing

modules/account-lib/src/coin/dot/singletonRegistry.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Material } from './iface';
+
+export class SingletonRegistry {
+  private static instance: TypeRegistry;
+  private static material: Material;
+  private static metadata: Metadata;
+
+  static getInstance(material): TypeRegistry {
+    if (material !== SingletonRegistry.material) {
+      SingletonRegistry.material = material;
```
byhow, line 14: ditto on the reducing memory here: Since we are storing the material and the instance at the same time, it seems like we are duplicating and persisting the 2 big strings in memory again. This seems like a caching effort to reduce computation?

modules/account-lib/src/coin/dot/transactionBuilder.ts
nachofm: Material is a string and since JS has string interning (https://en.wikipedia.org/wiki/String_interning) there are no two instances of it. Memory is consumed only once, parsing the metadata is what consumes memory so we should be careful about that.

modules/account-lib/src/coin/dot/utils.ts
nachofm: Metadata is set when the singleton is created as in:  
metadataRpc: material.metadata as `0x${string}`,
That is why all tests are passing even if we are not setting metadata explicitly, we need to set metadata  again whenever we make a change in the material.

modules/account-lib/src/coin/dot/singletonRegistry.ts
```diff
@@ -0,0 +1,24 @@
+import { TypeRegistry } from '@substrate/txwrapper-core/lib/types';
+import { getRegistry } from '@substrate/txwrapper-polkadot';
+import { PolkadotSpecNameType } from '@bitgo/statics';
+import { Metadata } from '@polkadot/types/metadata/Metadata';
+import { Material } from './iface';
+
+export class SingletonRegistry {
+  private static instance: TypeRegistry;
+  private static material: Material;
+  private static metadata: Metadata;
```
byhow, line 10: `metadata` doesn't seem to be read? Or did I miss something. Where do we need this for?:
```ts
'metadata' is declared but its value is never read
```

modules/account-lib/src/coin/dot/utils.ts
byhow: ah ok, good to know!

modules/account-lib/src/coin/dot/singletonRegistry.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  static getInstance(material: Material): TypeRegistry {
```
byhow, line 12: For test coverage purposes, let's add a quick unit test check for this new method:
if === material and if !== material

modules/account-lib/src/coin/dot/singletonRegistry.ts
nachofm: You are right, we tried a different approach that did not work and forgot to delete that variable.

modules/account-lib/src/coin/dot/singletonRegistry.ts
nachofm: Done

byhow (approved):
LGTM. Good to have the memory usage difference in the description! Very helpful metrics on reducing the memory.

PR #2036, opened 2022-02-22
feat(account-lib): include rent exempt amount in solana ata init transaction

jimmyqi-bitgo:
https://bitgoinc.atlassian.net/browse/STLX-11958

### DIFF
--- a/modules/account-lib/src/coin/sol/ataInitializationBuilder.ts
+++ b/modules/account-lib/src/coin/sol/ataInitializationBuilder.ts
@@ -10,6 +10,7 @@ import { BuildTransactionError } from '../baseCoin/errors';
 import { ASSOCIATED_TOKEN_PROGRAM_ID, Token, TOKEN_PROGRAM_ID } from '@solana/spl-token';
 import assert from 'assert';
 import { SolCoin } from '@bitgo/statics/dist/src/account';
+import { AtaInitializationTransaction } from './ataInitializationTransaction';
 
 export class AtaInitializationBuilder extends TransactionBuilder {
   private _mint: string;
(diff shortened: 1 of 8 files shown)


modules/account-lib/src/coin/sol/ataInitializationBuilder.ts
```diff
@@ -10,13 +10,15 @@ import { BuildTransactionError } from '../baseCoin/errors';
...
 import { SolCoin } from '@bitgo/statics/dist/src/account';
+import { AtaInitializationTransaction } from './ataInitializationTransaction';
 
 export class AtaInitializationBuilder extends TransactionBuilder {
   private _mint: string;
   private _rentExemptAmount: string;
 
   constructor(_coinConfig: Readonly<CoinConfig>) {
     super(_coinConfig);
+    this._transaction = new AtaInitializationTransaction(_coinConfig);
```
joey-bai, line 21: is it possible to re-declare the `_transaction` property as `AtaInitializationTransaction`?
e.g.

`private _transaction: AtaInitializationTransaction`

modules/account-lib/src/coin/sol/ataInitializationTransaction.ts
```diff
@@ -0,0 +1,86 @@
...
+      return;
+    }
+    const outputs: Entry[] = [];
+    const inputs: Entry[] = [];
+    const instructionParams = instructionParamsFactory(this.type, this._solTransaction.instructions);
+
+    for (const instruction of instructionParams) {
+      if (instruction.type === InstructionBuilderTypes.CreateAssociatedTokenAccount) {
+        const token = getSolTokenFromAddress(instruction.params.mintAddress, this._coinConfig.network);
+        const tokenName = token ? token.name : UNSUPPORTED_TOKEN_NAME;
```
joey-bai, line 40: ok to leave as is for now, but im wondering if we should throw an error if the token isn't supported.

joey-bai (approved):

PR #2042, opened 2022-02-23
fix(core): change etherscan url to default goerli

sachushajiabraham:
The recover function calls etherscan internally to get some details regarding the address but seems like it defaults to kovan even if we initialise the sdk with goerli. Since kovan is already retiring, reverting the default to goerli instead of kovan.

Ticket: [BG-42778](https://bitgoinc.atlassian.net/browse/BG-42778)


barathcj (approved):

wesgraham (approved):
Approving - but want to make sure this wont cause test failures when we bump the sdk version in platform -> can you check if UT's in Bitgo-MS depend on `etherscanBaseUrl`, and maybe update platform when we bump sdk version as well?

sachushajiabraham:
I saw some unit tests which is expecting kovan, I will update the UTs when we bump @wesgraham

PR #2044, opened 2022-02-23
fix(root): fix isWalletAddress for STX

alebusse:
Fixed isWalletAddress for STX
Added unit tests

[BG-43225](https://bitgoinc.atlassian.net/browse/BG-43225)
Subtasks: 
STX -> [BG-43277](https://bitgoinc.atlassian.net/browse/BG-43277)

### DIFF
--- a/modules/core/src/v2/coins/stx.ts
+++ b/modules/core/src/v2/coins/stx.ts
@@ -16,7 +16,6 @@ import {
   TransactionPrebuild as BaseTransactionPrebuild,
 } from '../baseCoin';
 import { BitGo } from '../../bitgo';
-import { MethodNotImplementedError } from '../../errors';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 interface SupplementGenerateWalletOptions {
@@ -90,8 +89,22 @@ export class Stx extends BaseCoin {
     return true;
   }
 
+  /**
+   * Check if address is valid, then make sure it matches the base address.
+   *
+   * @param {VerifyAddressOptions} params
+   * @param {String} params.address - the address to verify
+   * @param {String} params.baseAddress - the base address from the wallet
+   */
   isWalletAddress(params: VerifyAddressOptions): boolean {
-    throw new MethodNotImplementedError();
+    const { address, keychains } = params;
+    if (!keychains || keychains.length !== 3) {
+      throw new Error('Invalid keychains');
+    }
+    const pubs = keychains.map((keychain) => accountLib.Stx.Utils.xpubToSTXPubkey(keychain.pub));
+    const addressVersion = accountLib.Stx.Utils.getAddressVersion(address);
+    const baseAddress = accountLib.Stx.Utils.getSTXAddressFromPubKeys(pubs, addressVersion).address;
+    return accountLib.Stx.Utils.isSameBaseAddress(address, baseAddress);
   }
 
   /**
--- a/modules/core/test/v2/unit/baseCoin.ts
+++ b/modules/core/test/v2/unit/baseCoin.ts
@@ -219,5 +219,4 @@ describe('V2 Base Coin:', function () {
 
     });
   });
-
 });
--- a/modules/core/test/v2/unit/coins/stx.ts
+++ b/modules/core/test/v2/unit/coins/stx.ts
@@ -1,7 +1,8 @@
-import { Stx, Tstx } from '../../../../src/v2/coins/';
-import * as accountLib from '@bitgo/account-lib';
 import { TestBitGo } from '../../../lib/test_bitgo';
 import * as testData from '../../fixtures/coins/stx';
+import * as accountLib from '@bitgo/account-lib';
+import * as should from 'should';
+import { Stx, Tstx } from '../../../../src/v2/coins/';
 
 describe('STX:', function () {
   let bitgo;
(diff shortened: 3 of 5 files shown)


modules/core/src/v2/baseCoin.ts
```diff
@@ -402,7 +402,7 @@ export abstract class BaseCoin {
 
   /**
    * @param params
-   * @return true iff address is a wallet address. Must return false if address is outside wallet.
+   * @return true if address is a wallet address. Must return false if address is outside wallet.
```
tylerlevine, line 405: iff is short for "if and only if": https://en.m.wikipedia.org/wiki/If_and_only_if

tylerlevine (changes requested):
I don't think this is correct. The idea behind isWalletAddress is to prove that the address being checked belongs to the wallet in a probable, cryptographic sense by rederiving it from the wallet keys.

modules/core/src/v2/baseCoin.ts
alebusse: Oh my bad, sorry.

alebusse:
> I don't think this is correct. The idea behind isWalletAddress is to prove that the address being checked belongs to the wallet in a probable, cryptographic sense by rederiving it from the wallet keys.

Ok, i though was like the current implementations in CSPR and XRP for example.
I think there is a way for STX and probably leave SOL for later.

alebusse:
@tylerlevine For STX and HBAR there is not cryptographic derivation, we just increase the memoId and use the same base/root address.

modules/core/src/v2/coins/stx.ts
```diff
@@ -91,7 +95,18 @@ export class Stx extends BaseCoin {
...
+      throw new InvalidAddressError(`invalid address: ${address}`);
+    }
+    const addressDetails = accountLib.Stx.Utils.getAddressDetails(address);
+    const rootAddressDetails = accountLib.Stx.Utils.getAddressDetails(baseAddress);
+    if (addressDetails.address !== rootAddressDetails.address) {
+      throw new UnexpectedAddressError(
+        `address validation failure: ${addressDetails.address} vs ${rootAddressDetails.address}`
+      );
+    }
+    return true;
```
OttoAllmendinger, line 109:
```suggestion
    return addressDetails.address === rootAddressDetails.address;
```

modules/core/src/v2/coins/hbar.ts
```diff
@@ -146,7 +155,18 @@ export class Hbar extends BaseCoin {
...
+      throw new InvalidAddressError(`invalid address: ${address}`);
+    }
+    const addressDetails = this.getAddressDetails(address);
+    const rootAddressDetails = this.getAddressDetails(baseAddress);
+    if (addressDetails.address !== rootAddressDetails.address) {
+      throw new UnexpectedAddressError(
+        `address validation failure: ${addressDetails.address} vs ${rootAddressDetails.address}`
+      );
+    }
+    return true;
```
OttoAllmendinger, line 169: please deduplicate this code (maybe `hasSameBaseAddress()` in a new file `coins/address/index.ts`)

alebusse:
@tylerlevine @OttoAllmendinger I removed CSPR and HBAR part, left only the STX implementation using the `xpubs` from the `keychains` to regenerate the `baseAddress` in order to do the verification

modules/core/test/v2/unit/coins/stx.ts
```diff
@@ -94,9 +95,19 @@ describe('STX:', function () {
...
+    const bitgoKey = { pub: 'xpub661MyMwAqRbcGP1adk34VzRQJEMX25rCxjEyU9YFFWNhWNzwPoqgjLoKfnqotLwrz7kBevWbRZnqTSQrQDuJuYUQaDQ5DDPEzEXMwPS9PEf' };
+    const keychains = [userKey, backupKey, bitgoKey];
+    const validAddress1 = 'SNAYQFZ6EF54D5XWJP3GAE1Y8DPYXKFC7TTMYXFV';
+    const validAddress2 = 'SNAYQFZ6EF54D5XWJP3GAE1Y8DPYXKFC7TTMYXFV?memoId=2';
+    const unrelatedValidAddress = 'ST11NJTTKGVT6D1HY4NJRVQWMQM7TVAR091EJ8P2Y?memoId=1';
+    const invalidAddress = 'ST1T758K6T2YRKG9Q0TJ16B6FP5QQREWZSESRS0PY';
+    basecoin.verifyAddress({ address: validAddress1, keychains }).should.true();
+    basecoin.verifyAddress({ address: validAddress2, keychains }).should.true();
+    basecoin.verifyAddress({ address: unrelatedValidAddress, keychains }).should.false();
+    should.throws(() => basecoin.verifyAddress({ address: invalidAddress, keychains }), `invalid address: ${invalidAddress}`);
```
OttoAllmendinger, line 110: please use `isWalletAddress`

OttoAllmendinger (dismissed):
I am not the right person to review this

modules/core/src/v2/baseCoin.ts
```diff
@@ -604,4 +604,14 @@ export abstract class BaseCoin {
...
     return Buffer.from(serializedTx);
   }
+
+  /**
+   * Returns the base address portion of an address
+   *
+   * @param {String} address - an address
+   * @returns {String} - the base address
+   */
+  getBaseAddress(address: string): string {
```
tylerlevine, line 614: Not all coins have a concept of a base address, so this doesn't belong in `baseCoin`

modules/core/src/v2/coins/stx.ts
```diff
@@ -90,8 +91,25 @@ export class Stx extends BaseCoin {
...
+   * Check if address is valid, then make sure it matches the base address.
+   *
+   * @param {VerifyAddressOptions} params
+   * @param {String} params.address - the address to verify
+   * @param {String} params.baseAddress - the base address from the wallet
+   */
   isWalletAddress(params: VerifyAddressOptions): boolean {
-    throw new MethodNotImplementedError();
+    const { address, keychains } = params;
+    const pubs = _.map(keychains, (keychain) => accountLib.Stx.Utils.xpubToSTXPubkey(keychain.pub));
```
tylerlevine, line 103: can this be removed in favor of the built-in .map() function instead?

modules/core/src/v2/coins/stx.ts
```diff
@@ -90,8 +91,25 @@ export class Stx extends BaseCoin {
...
+    const addressVersion = accountLib.Stx.Utils.getAddressVersion(address);
+    const baseAddress = accountLib.Stx.Utils.getSTXAddressFromPubKeys(pubs, addressVersion).address;
+    return isSameBaseAddress(this, address, baseAddress);
```
tylerlevine, line 106: Can the code of the `isSameBaseAddress` be pulled inline here? I don't see any other calls to it

modules/core/test/v2/unit/baseCoin.ts
```diff
@@ -220,4 +220,12 @@ describe('V2 Base Coin:', function () {
     });
   });
 
+  describe('getBaseAddress', function () {
```
tylerlevine, line 223: please move to `stx` test file since we shouldn't have this function on `baseCoin`

tylerlevine (changes requested):

tylerlevine (approved):

PR #2045, opened 2022-02-23
chore(core): improve support for tss wallets

joey-bai:
adds functionality to build, sign, and send transactions from a
tss wallets. previously, sendMany would handle this, but we want to be
able to build, sign, and send independently.

Ticket: STLX-13780

this change is needed to support UI clients, where the build, sign, and send operations are performed separately. this is typically:
1. User builds transaction
2. Client displays details for user to review
3. User inputs password
4. Client signs and sends transaction

sample script that mimics the above:
```
async function prebuildTransaction() {
  const wallet = await bitgo.coin('tsol').wallets().get({ id: walletId });
  console.log(wallet);
  const builtTx = await wallet.prebuildTransaction({
    recipients: [{
      address: '6DadkZcx9JZgeQUDbHh12cmqCpaqehmVxv6sGy49jrah',
      amount: '1000'
    }],
    type: 'transfer',
  });

  return builtTx;
}

async function prebuildAndSignTx(prebuiltTransaction) {
  const wallet = await bitgo.coin('tsol').wallets().get({ id: walletId });
  const prebuildAndSignTx = await wallet.prebuildAndSignTransaction({ prebuildTx: prebuiltTransaction, walletPassphrase: 'Ghghjkg!455544llll' });

  return prebuildAndSignTx;
}

async function submitTx(signedTx) {
  const wallet = await bitgo.coin('tsol').wallets().get({ id: walletId });
  const sentTx = await wallet.submitTransaction(signedTx);

  console.log('sentTx', JSON.stringify(sentTx, undefined, 2));
}

prebuildTransaction()
  .then(prebuildAndSignTx)
  .then(submitTx)
  .catch(console.error);
```

### DIFF
--- a/modules/core/src/v2/wallet.ts
+++ b/modules/core/src/v2/wallet.ts
@@ -13,6 +13,7 @@ import { AddressGenerationError } from '../errors';
 import {
   BaseCoin,
   SignedTransaction,
+  SignedTransactionRequest,
   TransactionPrebuild,
   VerificationOptions,
   VerifyAddressOptions,
@@ -20,12 +21,11 @@ import {
 import { Eth } from './coins';
 import * as internal from './internal/internal';
 import { drawKeycard } from './internal/keycard';
-import { Keychain } from './keychains';
+import { Keychain, KeyIndices } from './keychains';
 import { TradingAccount } from './trading/tradingAccount';
 import { PendingApproval, PendingApprovalData } from './pendingApproval';
 import { RequestTracer } from './internal/util';
 import { getSharedSecret } from '../ecdh';
-import { KeyIndices } from '.';
 import { TssUtils } from './internal/tssUtils';
 
 const debug = debugLib('bitgo:v2:wallet');
@@ -134,6 +134,7 @@ export interface CustomSigningFunction {
 }
 
 export interface WalletSignTransactionOptions {
+  reqId?: RequestTracer;
   txPrebuild?: TransactionPrebuild;
   prv?: string;
   pubs?: string[];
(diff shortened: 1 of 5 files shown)


modules/core/src/v2/wallet.ts
```diff
@@ -2680,23 +2690,91 @@ export class Wallet {
...
+      const signedTxRequest = await this.tssUtils.signTxRequest(
+        params.txPrebuild.txRequestId,
+        params.prv,
+        params.reqId || new RequestTracer()
+      );
+      return {
+        txRequestId: signedTxRequest.txRequestId,
+      };
+    } catch (e) {
+      console.error(e);
```
lcovar, line 2762:
```suggestion
```

modules/core/src/v2/wallet.ts
```diff
@@ -2030,14 +2031,19 @@ export class Wallet {
...
-        _.omit(params, ['keychain', 'prv', 'passphrase', 'walletPassphrase', 'key', 'wallet'])
-      );
-      console.error('transaction prebuild:', txPrebuild);
-      console.trace(e);
-      throw e;
+      if (this._wallet.multisigType === 'tss') {
+        // TODO: STLX-13411 - implement verify tx for tss wallets
+        console.warn('transaction prebuild failed local validation:', e.message);
+      } else {
+        console.error('transaction prebuild failed local validation:', e.message);
```
lcovar, line 2038: Is pushing these console.error/warn intentional or just for local testing?

joey-bai: this is intentional. right now we don't have verify transaction working with TxRequests.
so i decided to catch the error and output a warning so we can continue developing with TSS wallets.

after we get verify tx working, we will log the error and re-throw.

lcovar (approved):
Nice work

PR #2051, opened 2022-02-25
feat(account-lib): spl-token encode/decode rework STLX-11959

bitgoandrew:


modules/account-lib/src/coin/sol/tokenEncodeDecode.ts
```diff
@@ -0,0 +1,216 @@
+import { AccountMeta, PublicKey, Signer, TransactionInstruction } from '@solana/web3.js';
```
joey-bai, line 1: add prettier:
```
/**
 * @prettier
 */
```

modules/account-lib/src/coin/sol/tokenEncodeDecode.ts
```diff
@@ -0,0 +1,216 @@
...
+const splToken = require('@solana/spl-token');
+
```
joey-bai, line 3: let's add a comment on why we even need to do this - mention that decoding is in alpha, but its not functional at the moment, so we're cherrypicking what we need until we have a stable release

joey-bai (commented):
need unit tests

modules/account-lib/src/coin/sol/tokenEncodeDecode.ts
```diff
@@ -0,0 +1,216 @@
...
+ * @param destination  Destination account
+ * @param owner        Owner of the source account
+ * @param amount       Number of tokens to transfer
+ * @param decimals     Number of decimals in transfer amount
+ * @param multiSigners Signing accounts if `owner` is a multisig
+ * @param programId    SPL Token program account
+ *
+ * @return Instruction to add to a transaction
+ */
+export function createTransferCheckedInstruction(
```
joey-bai, line 82: double check if we need this. if we can create a transfer checked instruction in the current, stable version of the token library, we should use it.

bitgoandrew: The existing createTransferCheckedInstruction relies on a structure that causes the errors in alpha (https://github.com/solana-labs/solana-program-library/blob/87fe3793ea87aac5630564c86c1b7792bb3e8ace/token/ts/src/instructions/transferChecked.ts#L22)

Creating a new one allows us to work around this struct error by encoding/decoding through JSON

modules/account-lib/test/unit/coin/sol/transactionBuilder/transactionBuilder.ts
```diff
@@ -310,7 +310,7 @@ describe('Sol Transaction Builder', async () => {
         .nonce(validBlockhash)
         .fee({ amount: 5000 })
         .send({ address: nonceAccount.pub, amount: '1000' });
-      transferBuilder2.addSignature({ pub: authAccount.pub }, bs58.decode(signature));
+      transferBuilder2.addSignature({ pub: authAccount.pub }, Buffer.from(bs58.decode(signature)));
```
joey-bai, line 313: is this needed?

modules/account-lib/src/coin/sol/tokenEncodeDecode.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+
+/**
+ * Decode a TransferChecked instruction without validating it
+ *
+ * @param instruction Transaction instruction to decode
+ *
+ * @return Decoded, non-validated instruction
+ */
+export function decodeTransferCheckedInstructionUnchecked({
```
joey-bai, line 150: let's not export this. callers should always be using the `checked` function

modules/account-lib/src/coin/sol/tokenEncodeDecode.ts
```diff
@@ -0,0 +1,173 @@
...
+      destination,
+      owner,
+      multiSigners,
+    },
+    data,
+  };
+}
+
+/** A decoded, non-validated TransferChecked instruction */
+export interface DecodedTransferCheckedInstructionUnchecked {
```
joey-bai, line 114: no export

modules/account-lib/src/coin/sol/tokenEncodeDecode.ts
```diff
@@ -0,0 +1,173 @@
...
+
+/*
+  IMPORTANT:
+
+  Since token transfer encoding and decoding is in alpha (v 0.2.0-alpha.2), a version that is currently not functional due to typing bugs,
+  this file is the cherry-picked methods that are required to complete the token transfer feature. When the Solana foundation releases a stable
+  version, this file should be deleted and the spl-token methods should be used.
+ */
+
+export enum TokenInstruction {
```
joey-bai, line 16: let's minimize the models/functions we're exporting.
i feel like we really only need to export `decodeTransferCheckedInstruction` and `DecodedTransferCheckedInstruction`

joey-bai (commented):
overall lgtm

joey-bai (approved):

PR #2052, opened 2022-02-25
feat(sdk): dynamic load dot material

aseoane-simtlix:
Add material endpoint call for client

We want SDK to solve the material data issue, which needs to be up-to-date for the tx to be signed.Idea would be to fetch and load materials dynamically, hence removing some load from the client.

We found the [getExtraPrebuildParams()](https://github.com/BitGo/BitGoJS/blob/39aa238720ebb2de91ca46246acc5b04b94302f3/modules/core/src/v2/baseCoin.ts#L424) to get extra parameters for prebuilding a tx.

STLX-13766


aseoane-simtlix:
ticket closed as `won't fix`

PR #2054, opened 2022-02-26
Sync up with bitcoinjs-lib 6

brandonblack:
This is a pretty major change:

* TransactionBuilder, classify, and related are no longer part of bitcoinjs-lib, so they're moved into utxo-lib.
* tiny-secp256k1 no longer works in browser w/o WASM support, so switched crypto implementation to noble-secp256k1
  * Schnorr signatures changed because lack of extra nonce data is now empty buffer instead of skipped
  * I made some moderately extensive changes to noble-secp256k1 https://github.com/paulmillr/noble-secp256k1/pull/50


modules/utxo-lib/src/networks.ts
```diff
@@ -58,7 +58,7 @@ export type Network = {
     pubKeyHash: number;
     scriptHash: number;
   };
-  bech32?: string;
+  bech32: string;
```
OttoAllmendinger, line 61: why do we need this?

modules/utxo-lib/test/bitgo/fixtures/signature/bitcoin/p2tr/parsed-backup-bitgo-fullSigned.json
```diff
@@ -7,8 +7,8 @@
     "e21c29b4a7eeace9c7a8cefb568ca00c86ff9bf5e79e07e5442c29d4a0950d04"
   ],
   "signatures": [
-    "9f74390b6c7f5988772a909fc689a3e2dd680ceb8d54d4caffdd5907132bc415d7d964f4b25ce02690938cd4b62057f80b929aca4155c64db205a336b2e6ce79",
-    "69a969c8df93bc7894d045bc5e5d73f50d815fed7f703f11080f6b46561220156a0eed3382ec0eddbfa2737a29a7b07ba572880ae26f05f56ebca40bd7a5a987"
+    "a2074b68b115e970c8c4bee9cf75c03b82bd3e3086d58801a4adb7b31b9b5f2cea4a5615a4feef74b61db6d765bb814feaf27331fad128dda55cfe106d25d44e",
+    "6bace9d49a0a7617e8ab40cd5361257829c0f09809d6f70ce130a70d7af3de1bf7409cbe379481e0494b32531b955fff67825cd969a3e95fc853c048692feda1"
```
OttoAllmendinger, line 11: why did this change?

modules/utxo-lib/test/networks.ts
```diff
@@ -133,7 +133,10 @@ describe('networks', function () {
       it('has expected properties', function () {
         assert.strictEqual(typeof network, 'object');
         assert.strictEqual(typeof network.messagePrefix, 'string');
-        assert.strictEqual(typeof (network as any).bech32, bech32Coins(network) ? 'string' : 'undefined');
+        assert.strictEqual(typeof (network as any).bech32,  'string');
+        if (!bech32Coins(network)) {
+          assert.strictEqual((network as any).bech32.length, 0);
+        }
```
OttoAllmendinger, line 139: IMO `undefined` is better than zero-length string here

modules/utxo-lib/src/networks.ts
brandonblack: To match the Network type from `bitocinjs-lib` I'd love a better idea if you have one. I considered changing the upstream type in my "BitGo specific" commit, but it looked like it would become a maintenance pain.

modules/utxo-lib/test/bitgo/fixtures/signature/bitcoin/p2tr/parsed-backup-bitgo-fullSigned.json
brandonblack: See the PR description:

> Schnorr signatures changed because lack of extra nonce data is now empty buffer instead of skipped`

modules/utxo-lib/test/networks.ts
brandonblack: No disagreement, but that's not what the upstream type has.

brandonblack:
Improved commit history, formatted code, and tidied up `noble_ecc.ts`

modules/utxo-lib/test/networks.ts
OttoAllmendinger: changed it so we do not use the upstream type anymore: https://github.com/BitGo/BitGoJS/pull/2054/commits/6f553c087ff01d433b1cd8498dde583dc47960a6

OttoAllmendinger (dismissed):

PR #2062, opened 2022-03-02
feat(statics): create FIAT currency in Testnet

emiacc:
STLX-13388 Create FIAT currency in Testent
STLX-13387 Create USD token to represent the FIAT US Dollar currency

FIAT Improvements
Our Prime micro services today all use SUSD to represent FIAT. 
SUSD stands for Silvergate USD, which up to this point has served as well since USD has been our exclusive fiat currency and Silvergate has been our exclusive banking partner. However, this is extremely restrictive, and as we near adding additional currencies and bank support, we need to rework how we represent different types of fiats in our services. 

Proposal: 
FIAT will be a new coin and each currency, represented by the ISO codes, would be a token


modules/core/src/config.ts
```diff
@@ -168,6 +188,20 @@ const formattedEosTokens = coins.reduce((acc: EosTokenConfig[], coin) => {
...
 }, []);
 
+const formattedFiatTokens = coins.reduce((acc: FiatTokenConfig[], coin) => {
+  if (coin instanceof FiatCoin) {
+    acc.push({
+      name: coin.fullName,
+      type: coin.name,
+      coin: coin.network.type === NetworkType.MAINNET ? 'fiat' : 'tfiat',
+      network: coin.network.type === NetworkType.MAINNET ? 'Mainnet' : 'Testnet',
+      tokenContractAddress: '',
```
tylerlevine, line 198: why do we need a `tokenContractAddress` and why is it always just an empty string?

modules/statics/src/coins.ts
```diff
@@ -108,6 +109,7 @@ export const coins = CoinMap.fromCoins([
   account('txtz', 'Testnet Tezos', Networks.test.xtz, 6, UnderlyingAsset.XTZ, XTZ_FEATURES),
   account('susd', 'Silvergate USD', Networks.main.susd, 2, UnderlyingAsset.USD),
   account('tsusd', 'Testnet Silvergate USD', Networks.test.susd, 2, UnderlyingAsset.USD),
+  account('tfiat', 'Testnet FIAT', Networks.test.fiat, 2, UnderlyingAsset.FIAT),
```
tylerlevine, line 112: does this make sense as an account coin? What's the distinction between a fiat coin that uses the `account()` function vs one that uses the `tfiatToken()` function?

tylerlevine, line 112: does this make sense as an account coin? What's the distinction between a fiat coin that uses the `account()` function vs one that uses the `tfiatToken()` function?

modules/statics/src/account.ts
```diff
@@ -982,3 +1014,63 @@ export function tavaxErc20(
...
+
+/**
+ * Factory function for testnet FIAT currency instances.
+ *
+ * @param name unique identifier of the currency
+ * @param fullName Complete human-readable name of the currency
+ * @param asset Asset which this coin represents. This is the same for both mainnet and testnet variants of a currency.
+ * @param decimalPlaces Number of decimal places this currency supports (divisibility exponent)
+ */
+export function tfiatToken(
```
tylerlevine, line 1068: Are we missing a `fiatToken()` function that is the main net equivalent? Or is that to come later?

modules/statics/src/account.ts
```diff
@@ -982,3 +1014,63 @@ export function tavaxErc20(
...
+ * @param asset Asset which this coin represents. This is the same for both mainnet and testnet variants of a coin.
+ * @param decimalPlaces Number of decimal places this coin supports (divisibility exponent)
+ * @param network Network object for this coin
+ * @param prefix? Optional coin prefix. Defaults to empty string
+ * @param suffix? Optional coin suffix. Defaults to coin name.
+ * @param features? Features of this coin. Defaults to the DEFAULT_FEATURES defined in `AccountCoin`
+ * @param isToken? Whether or not this account coin is a token of another coin
+ * @param primaryKeyCurve The elliptic curve for this chain/token
+ */
+export function fiatToken(
```
tylerlevine, line 1032: do we need `fiat()` / `tfiat()` factory functions? What's the difference between a fiat asset and a fiat token asset?

tylerlevine, line 1032: do we need `fiat()` / `tfiat()` factory functions? What's the difference between a fiat asset and a fiat token asset?

tylerlevine (changes requested):
no conceptual issues from me, just some questions about implementation

modules/core/src/config.ts
emiacc: Removed. We don't need the `tokenContractAddress` for Fiat. Thanks!

modules/statics/src/coins.ts
emiacc: Right, we aren't going to use Fiat or TFiat as coin, we will be creating just Fiat Tokens, so this call was removed as well.

emiacc: duplicated

modules/statics/src/account.ts
emiacc: That comes later, addressed with: https://bitgoinc.atlassian.net/browse/STLX-13365

modules/statics/src/account.ts
emiacc: `fiat` and `tfiat` will be our base currency. `USD`, `EUR`, `YEN` will be our fiat "tokens".
Similar to how ERC20 tokens are an extension of ETH, and use the same backing wallet. 
But in this case we wont create Fiat assets. Some unnecessary code was removed. Thanks!

emiacc: Duplicated

modules/core/src/v2/coins/fiat.ts
```diff
@@ -0,0 +1,76 @@
...
+  ParsedTransaction,
+  ParseTransactionOptions,
+  SignedTransaction,
+  SignTransactionOptions,
+  VerifyAddressOptions,
+  VerifyTransactionOptions,
+} from '../baseCoin';
+import { MethodNotImplementedError } from '../../errors';
+
+export class Fiat extends BaseCoin {
```
tylerlevine, line 17: if we're not creating instances of this, can we make it `abstract` and remove the `createInstance()` method?

modules/statics/src/account.ts
```diff
@@ -265,6 +265,38 @@ export class AvaxERC20Token extends ContractAddressDefinedToken {
   }
 }
 
+/**
+ * FIAT based coins, such as USD, EUR, or YEN.
+ */
+export class FiatCoin extends BaseCoin {
```
tylerlevine, line 271: Can we rename to `FiatToken` so we don't have a mismatch with the name of the factory function?

tylerlevine (changes requested):

modules/core/src/v2/coins/fiat.ts
emiacc: Done!

modules/statics/src/account.ts
emiacc: Done!

modules/core/src/v2/coins/fiat.ts
```diff
@@ -0,0 +1,71 @@
...
+  generateKeyPair(seed?: Buffer): KeyPair {
+    throw new MethodNotImplementedError();
+  }
+
+  isValidPub(pub: string): boolean {
+    throw new MethodNotImplementedError();
+  }
+
+  async parseTransaction(params: ParseTransactionOptions): Promise<ParsedTransaction> {
+    return {};
```
tylerlevine, line 57: please throw if you're not ready to implement these yet, or you can also leave them unimplemented in the abstract base class and throw in the derived class instead.
```suggestion
    throw new MethodNotImplementedError();
```

modules/core/src/v2/coins/fiat.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  isWalletAddress(params: VerifyAddressOptions): boolean {
+    throw new MethodNotImplementedError();
+  }
+
+  async verifyTransaction(params: VerifyTransactionOptions): Promise<boolean> {
+    return true;
```
tylerlevine, line 65:
```suggestion
    throw new MethodNotImplementedError();
```

tylerlevine (changes requested):

tylerlevine (approved):

PR #2069, opened 2022-03-03
feat: external signer to read encrypted privkeys

DariusParvin:
- When running in external signer mode, express will expect the private
key json file to be in the encrypted format (same as on the keycard but
with escape characters).
- Express expects the wallet passphrase to be set in the environment
  with the name `WALLET_<walletId>_PASSPHRASE`
- Update documentation and move to new file EXTERNAL_SIGNER.md

Ticket: BG-43925

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -384,28 +384,25 @@ async function getEncryptedPrivKey(path: string, walletId: string): Promise<stri
   return encryptedPrivKey[walletId];
 }
 
-function decryptPrivKey(encryptedPrivKey: string, walletPw: string): string {
-  const bg = new BitGo();
+function decryptPrivKey(bg: BitGo, encryptedPrivKey: string, walletPw: string): string {
   try {
-    const decrypted = bg.decrypt({ password: walletPw, input: encryptedPrivKey });
-    return decrypted;
+    return bg.decrypt({ password: walletPw, input: encryptedPrivKey });
   } catch (e) {
     throw new Error(`Error when trying to decrypt private key: ${e}`);
   }
 }
 
 export async function handleV2Sign(req: express.Request) {
-  express;
   const walletId = req.body.txPrebuild.walletId;
   const walletPw = getWalletPwFromEnv(walletId);
   const privKeyPath = req.config.signerFileSystemPath;
   assert(typeof privKeyPath === 'string');
   const encryptedPrivKey = await getEncryptedPrivKey(privKeyPath, walletId);
-  const privKey = decryptPrivKey(encryptedPrivKey, walletPw);
   const bitgo = req.bitgo;
+  const privKey = decryptPrivKey(bitgo, encryptedPrivKey, walletPw);
   const coin = bitgo.coin(req.params.coin);
   try {
-    return await coin.signTransaction({ ...req.body, ...{ prv: privKey } });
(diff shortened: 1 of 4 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -366,19 +366,46 @@ function handleCanonicalAddress(req: express.Request) {
...
+  try {
+    const decrypted = bg.decrypt({ password: walletPw, input: encryptedPrivKey });
+    return decrypted;
+  } catch (e) {
+    throw new Error(`Error when trying to decrypt private key: ${e}`);
   }
+}
+
+export async function handleV2Sign(req: express.Request) {
+  express;
```
tylerlevine, line 398: what's this?

modules/express/src/clientRoutes.ts
```diff
@@ -366,19 +366,46 @@ function handleCanonicalAddress(req: express.Request) {
...
-    throw new Error(`Could not find a field for walletId: ${walletId} in ${req.config.signerFileSystemPath}`);
+  const encryptedPrivKey = JSON.parse(privKeyFile);
+  if (encryptedPrivKey[walletId] === undefined) {
+    throw new Error(`Could not find a field for walletId: ${walletId} in ${path}`);
+  }
+  return encryptedPrivKey[walletId];
+}
+
+function decryptPrivKey(encryptedPrivKey: string, walletPw: string): string {
+  const bg = new BitGo();
```
tylerlevine, line 388: pass bitgo object in as param - it's available in the calling scope as `req.bitgo` (see line 405)

modules/express/src/clientRoutes.ts
```diff
@@ -366,19 +366,46 @@ function handleCanonicalAddress(req: express.Request) {
...
+  try {
+    const decrypted = bg.decrypt({ password: walletPw, input: encryptedPrivKey });
+    return decrypted;
```
tylerlevine, line 391:
```suggestion
    return bg.decrypt({ password: walletPw, input: encryptedPrivKey });
```

modules/express/src/clientRoutes.ts
```diff
@@ -366,19 +366,46 @@ function handleCanonicalAddress(req: express.Request) {
...
+  const walletPw = getWalletPwFromEnv(walletId);
+  const privKeyPath = req.config.signerFileSystemPath;
+  assert(typeof privKeyPath === 'string');
+  const encryptedPrivKey = await getEncryptedPrivKey(privKeyPath, walletId);
+  const privKey = decryptPrivKey(encryptedPrivKey, walletPw);
   const bitgo = req.bitgo;
   const coin = bitgo.coin(req.params.coin);
   try {
-    return await coin.signTransaction({ ...req.body, ...{ prv: privKey[walletId] } });
+    return await coin.signTransaction({ ...req.body, ...{ prv: privKey } });
```
tylerlevine, line 408: literal destruction is redundant here
```suggestion
    return await coin.signTransaction({ ...req.body, prv: privKey });
```

tylerlevine (changes requested):
good work, just a few comments

modules/express/src/clientRoutes.ts
DariusParvin: that was unintentional. removed

DariusParvin:
Good catches, thanks for the feedback. I've addressed them now

tylerlevine (approved):

PR #2076, opened 2022-03-04
fix(core): support password reset and enterprise with MPC

joey-bai:
Create Keychains with originalPasscodeEncryptionCode and enterprise
populated if client provides as input.

`originalPasscodeEncryptionCode` is sent when creating the User key.
`enterprise` is sent when creating  the BitGo Key and Wallet

Ticket: STLX-14231


lcovar (approved):
nice work 💯

PR #2079, opened 2022-03-07
fix(statics): fix stx explorer url

alebusse:
Fixed STX explorer url

STLX-14219

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -401,7 +401,7 @@ class SolTestnet extends Testnet implements AccountNetwork {
 class Stx extends Mainnet implements StacksNetwork {
   name = 'Stx';
   family = CoinFamily.STX;
-  explorerUrl = 'https://explorer.stacks.co/';
+  explorerUrl = 'https://explorer.stacks.co/txid/';
   sendmanymemoContractAddress = 'SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE';
   stakingContractAddress = 'SP000000000000000000002Q6VF78';
 }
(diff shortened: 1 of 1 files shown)


modules/statics/src/networks.ts
```diff
@@ -401,15 +401,15 @@ class SolTestnet extends Testnet implements AccountNetwork {
 class Stx extends Mainnet implements StacksNetwork {
   name = 'Stx';
   family = CoinFamily.STX;
-  explorerUrl = 'https://explorer.stacks.co/';
+  explorerUrl = 'https://explorer.stacks.co/txid/';
```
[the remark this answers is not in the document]
alebusse: @mmcshinsky-bitgo I havent found any in our side besides Lumina which is not working b/c of this and old UI is not using statics to create the url for STX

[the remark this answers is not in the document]
tylerlevine: @mmcshinsky-bitgo downstream consumers should be using the url defined here so they would just need to upgrade to the new version of statics to get the fixed url.

modules/statics/src/networks.ts
```diff
@@ -401,15 +401,15 @@ class SolTestnet extends Testnet implements AccountNetwork {
...
   sendmanymemoContractAddress = 'SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE';
   stakingContractAddress = 'SP000000000000000000002Q6VF78';
 }
 
 class StxTestnet extends Testnet implements StacksNetwork {
   name = 'StxTestnet';
   family = CoinFamily.STX;
-  explorerUrl = 'https://explorer.stacks.co/?chain=testnet';
+  explorerUrl = 'https://explorer.stacks.co/txid/?chain=testnet';
```
tylerlevine, line 412: I don't think this will work - the txid gets appended to the end of the URL so the query param would end up in the middle.

tylerlevine: Is there any other way to specify the chain other than a query param? Otherwise we might need to change how the `explorerUrl` works so that it's a format string instead of a string to append to.

tylerlevine (changes requested):
See comment on testnet url

modules/statics/src/networks.ts
alebusse: We had to deal with the same issue in Solana https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/networks.ts#L398
But its the UI doing the split at `?`
https://github.com/BitGo/lumina/blob/master/packages/shared/src/api/utils.ts#L307-L311

tylerlevine: Got it, thanks

tylerlevine (approved):

PR #2080, opened 2022-03-07
feat: add fetchEncryptedPrivKeys.ts

DariusParvin:
This file will help create a json with the users wallet ids and encrypted
private keys to be used by the external signer.

I created this file based off the files in `modules/core/example`. 

Ticket: BG-43925

### DIFF
--- a/modules/express/src/fetchEncryptedPrivKeys.ts
+++ b/modules/express/src/fetchEncryptedPrivKeys.ts
@@ -1,0 +1,60 @@
+/**
+ * This tool will help you creates a private key JSON file to be used with the external singing mode feature.
+ * It creates a JSON file containing a list of wallet IDs and their corresponding encrypted user private keys.
+ *
+ * To run this file, use the command:
+ * `yarn ts-node <path/to/fetchEncryptedPrivKeys.ts>`
+ *
+ * Copyright 2022, BitGo, Inc.  All Rights Reserved.
+ */
+import { writeFile } from 'fs';
+import { BitGo } from 'bitgo';
+
+type Output = {
+  [key: string]: string
+}
+
+type WalletIds = {
+  [key: string]: string[]
+}
+
+// TODO: set env to 'test' or 'prod'
+const bg = new BitGo({ env: 'test' });
+// TODO: set your access token here
+const accessToken = '';
+// // TODO: set your coin type and wallet ids here e.g.
+const walletIds: WalletIds = {
+  // tbtc: ['61f039aad587c2000745c687373e0111', '6225b081cd291300071fed36b1362222'],
+  // gteth: ['61fb21819c54dd000755f8de3a18e333'],
+};
+
+async function main() {
+  bg.authenticateWithAccessToken({ accessToken });
+
+  // get the encrypted user privKey for each walletId and store in the JSON output
+  const output: Output = {};
+  for (const [coinName, ids] of Object.entries(walletIds)) {
+    const coin = bg.coin(coinName);
+    for (const id of ids) {
+      const wallet = await coin.wallets().get({ id });
+      const userKeyId = wallet.keyIds()[0];
+      const keychain = await coin.keychains().get({ id: userKeyId });
+      if (keychain.encryptedPrv === undefined) {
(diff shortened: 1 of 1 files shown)


modules/express/src/fetchEncryptedPrivKeys.ts
```diff
@@ -0,0 +1,46 @@
+import { writeFile } from 'fs';
+import { BitGo } from 'bitgo';
+
+// TODO: set env to 'test' or 'prod'
```
tylerlevine, line 4: Could you add a doc comment to explain how to run this? I think the user needs to do something like `yarn ts-node <file>`

modules/express/src/fetchEncryptedPrivKeys.ts
```diff
@@ -0,0 +1,46 @@
...
+const bg = new BitGo({ env: 'test' });
+// TODO: set your access token here
+const accessToken = '';
+// TODO: set your coin type and wallet ids here e.g.
+// const walletIds = [
+//   ['tbtc', '61f039aad587c2000145c687373e0fa1'],
+//   ['tgeth', '61fb21819c14dd000755f8de3a18e461']
+// ];
+const walletIds = [];
```
tylerlevine, line 13: Maybe make this an object with each coin name as a key, and an array of wallet ids as the value
```suggestion
const walletIds = {
//   tbtc: ['61f039aad587c2000145c687373e0fa1'],
//   tgeth: ['61fb21819c14dd000755f8de3a18e461'],
};
```

modules/express/src/fetchEncryptedPrivKeys.ts
```diff
@@ -0,0 +1,46 @@
...
+    const coin = walletIds[i][0];
+    const id = walletIds[i][1];
+    const wallet = await bg.coin(coin).wallets().get({ id });
+    const userKeyId = wallet.keyIds()[0];
+    const keychain = await bg.coin(coin).keychains().get({ id: userKeyId });
+    if (keychain.encryptedPrv === undefined) {
+      throw new Error(`could not find a ${coin} private key for wallet id ${id}`);
+    }
+    output[id] = keychain.encryptedPrv;
+  }
```
tylerlevine, line 34:
```suggestion
  for (const [coin, walletIds] of Object.entries(walletIds)) {
    const coin = bg.coin(coin);
    for (const id of walletIds) {
      const wallet = await coin.wallets().get({ id });
      const userKeyId = wallet.keyIds()[0];
      const keychain = await coin.keychains().get({ id: userKeyId });
      if (keychain.encryptedPrv === undefined) {
        console.warn(`could not find a ${coin} encrypted user private key for wallet id ${id}, skipping`);
        continue;
      }
      output[id] = keychain.encryptedPrv;
    }
  }
```

tylerlevine (changes requested):
more good work here - just some suggestions for a little bit of a cleaner implementation and doc improvements.

modules/express/src/fetchEncryptedPrivKeys.ts
```diff
@@ -0,0 +1,60 @@
...
+import { writeFile } from 'fs';
+import { BitGo } from 'bitgo';
+
+type Output = {
+  [key: string]: string
+}
+
+type WalletIds = {
+  [key: string]: string[]
+}
```
DariusParvin, line 19: added this `WalletIds` type here so that the compiler wouldn't complain about not knowing the type of `id` in L38

modules/express/src/fetchEncryptedPrivKeys.ts
DariusParvin: added a doc comment

modules/express/src/fetchEncryptedPrivKeys.ts
DariusParvin: thanks, this is cleaner. I implemented this but used new names `coin`/`coinName` and `walletIds`/`ids` to avoid the names conflicting.

DariusParvin:
ready for re-review

tylerlevine (approved):
nice job

PR #2085, opened 2022-03-08
STLX-13577-feat(wallet-platform): add emergency param to whitelist

nicolasbellanich-simtlix:
[stlx-13577](https://bitgoinc.atlassian.net/browse/STLX-13577)

### DIFF
--- a/modules/bitgo/src/v2/wallet.ts
+++ b/modules/bitgo/src/v2/wallet.ts
@@ -605,6 +605,9 @@ export class Wallet {
       'keyregTxBase64',
       'closeRemainderTo',
       'tokenName',
+      // param to set emergency flag on a custodial transaction.
+      // This transaction should be performed in less than 1 hour or it will fail.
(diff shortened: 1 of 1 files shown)


modules/core/src/v2/wallet.ts
```diff
@@ -598,6 +598,7 @@ export class Wallet {
       'eip1559',
       'keyregTxBase64',
       'closeRemainderTo',
+      'emergency',
```
charly-simtlix, line 601: Please document field use

charly-simtlix (changes requested):

modules/core/src/v2/wallet.ts
JulioAchaval-SimTLiX: Adding

charly-simtlix (approved):

mmcshinsky-bitgo (approved):

PR #2089, opened 2022-03-08
refactor(core): update BLS-DKG utils for keygen to match the WP changes based on the HSM changes

fperalta-simtlix:
HSM has adopted the same implementation as TSS on BLS-DKG.

BG-43029

### DIFF
--- a/modules/core/src/v2/internal/mpcUtils.ts
+++ b/modules/core/src/v2/internal/mpcUtils.ts
@@ -2,8 +2,15 @@
  * @prettier
  */
 
+import { Key, SerializedKeyPair, readKey, createMessage, readPrivateKey, encrypt, decrypt, readMessage } from 'openpgp';
 import { BitGo } from '../../bitgo';
 import { BaseCoin, KeychainsTriplet } from '../baseCoin';
+import { Keychain } from '../keychains';
+
+export interface MpcKeyShare {
+  publicShare: string;
+  privateShare: string;
+}
 
 export abstract class MpcUtils {
   protected bitgo: BitGo;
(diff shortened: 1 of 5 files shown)


modules/core/src/v2/internal/mpcUtils.ts
```diff
@@ -2,8 +2,15 @@
  * @prettier
  */
 
+import * as openpgp from 'openpgp';
```
wesgraham, line 5: same thing here

wesgraham (dismissed):

wesgraham (approved):

modules/core/src/v2/internal/mpcUtils.ts
```diff
@@ -14,6 +21,85 @@ export abstract class MpcUtils {
     this.baseCoin = baseCoin;
   }
 
+  protected async getBitgoGpgPubKey(): Promise<Key> {
+    const constants = await this.bitgo.fetchConstants();
+    if (!constants.mpc || !constants.mpc.bitgoPublicKey) {
```
lcovar, line 26: https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/client.js#L45

Wallet platform returns this as constants.tss not constants.mpc

fperalta-simtlix: Right, we are changing it in this PR https://github.com/BitGo/bitgo-microservices/pull/15616/files#diff-35b5736f90ff9fc9818876d37c5df36f7ff8a0589d7a87ec96bac15192e349c8R44

PR #2099, opened 2022-03-10
feat(core): Reencrypt derivation keypair when pwd changed

Krin2:
When a user changes their password, we need to update their keychains with
the new encrypted prvs. this changes ensures we update the key's
address derivation encrypted prv as well.

[STLX-13582](https://bitgoinc.atlassian.net/browse/STLX-13582)

### DIFF
--- a/modules/core/src/v2/keychains.ts
+++ b/modules/core/src/v2/keychains.ts
@@ -218,6 +218,9 @@ export class Keychains {
           if (updatedKeychain.encryptedPrv) {
             changedKeys[updatedKeychain.pub] = updatedKeychain.encryptedPrv;
           }
+          if (updatedKeychain.addressDerivationKeypair) {
+            changedKeys[updatedKeychain.addressDerivationKeypair.pub] = updatedKeychain.addressDerivationKeypair.encryptedPrv;
+          }
         } catch (e) {
           // if the password was incorrect, silence the error, throw otherwise
           if (!e.message.includes('private key is incorrect')) {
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/keychains.ts
```diff
@@ -255,10 +258,31 @@ export class Keychains {
...
     }
 
+    if (_.isObject(params.keychain) && _.isString(params.keychain.encryptedPrv) && params.keychain.addressDerivationKeypair) {
+      if (!params.keychain.addressDerivationKeypair.pub) {
+        throw new Error('expected keychain to have addressDerivationKeypair.pub property');
+      }
+      if (!params.keychain.addressDerivationKeypair.encryptedPrv) {
+        throw new Error('expected keychain to have addressDerivationKeypair.encryptedPrv property');
+      }
+    }
```
alebusse, line 268:
```suggestion
```
remove this, no all keychains have the addressDerivationKeypair so only its necesary to check when they do.

modules/core/src/v2/keychains.ts
```diff
@@ -255,10 +258,31 @@ export class Keychains {
...
     const oldEncryptedPrv = params.keychain.encryptedPrv;
     try {
       const decryptedPrv = this.bitgo.decrypt({ input: oldEncryptedPrv, password: params.oldPassword });
       const newEncryptedPrv = this.bitgo.encrypt({ input: decryptedPrv, password: params.newPassword });
+      if (params.keychain.addressDerivationKeypair) {
```
alebusse, line 273: Do the all verifications here instead.

alebusse (changes requested):

alebusse (approved):

alebusse (approved):

tylerlevine (changes requested):
could you please provide some more background on what this extra keychain is used for, especially how it's used in recovery scenarios?

Does it ever appear on the keycard?

PR #2105, opened 2022-03-14
feat(core): support new Algo token name format

charly-simtlix:
Add functions to implemented compatibility
between formats.

STLX-14551


modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -556,12 +571,70 @@ export class Utils implements BaseUtils {
...
+      this._algoAssets = this._coins.filter((coin) => coin.network.family === CoinFamily.ALGO);
+    }
+
+    return this._algoAssets;
+  }
+
+  /**
+   * Convert a tokenName to the old format. If the name is already is in the old format,
+   * the same name is returned.
+   * @param tokenName a token name in the old or new format
```
JulioAchaval-SimTLiX, line 613:
```suggestion
   * Convert a tokenName to the new format. If the name is already is in the new format,
   * the same name is returned.
   * @param tokenName a token name in the old or new format
```

JulioAchaval-SimTLiX: Add a comment to explain what if statics has old tokens. Because in that case you will not be able to get the new format.

modules/account-lib/src/coin/algo/utils.ts
```diff
@@ -556,12 +571,70 @@ export class Utils implements BaseUtils {
...
+    return this.getAlgoAssets().reduce(
+      (newTokenName, _, coinName) => (this.isSameAssetName(tokenName, coinName) ? coinName : newTokenName),
+      tokenName,
+    );
+  }
+
+  /**
+   * Compares two Algo token names for equality. The comparison ignores if they are new or old format.
+   * For example: "algo:USDt-312769" is equal to "algo:312769"
+   * Note: It does not detect false new names, so "algo:USDt-312769" will be equal to "algo:KAL-312769".
```
JulioAchaval-SimTLiX, line 630: Can we get a Suffix from the static coin list? We may compare if the suffix and the Id are compatible...

modules/statics/src/map.ts
```diff
@@ -22,14 +20,18 @@ export class CoinMap {
...
       return this._map.get(key)!;
     }
 
     throw new CoinNotDefinedError(key);
   }
 
+  public has(key: string): boolean {
+    return this._map.has(key);
+  }
+
```
JulioAchaval-SimTLiX, line 34: This change could affect several things on WP. We should try a custom version on T02 and check if other coins have any issue...

modules/account-lib/src/coin/algo/utils.ts
charly-simtlix: We don't know if that condition is enforced.

modules/statics/src/map.ts
charly-simtlix: This is a new method, because it is new, nobody is using it.

nicolasbellanich-simtlix (dismissed):

JulioAchaval-SimTLiX (dismissed):

nicolasbellanich-simtlix (dismissed):

JulioAchaval-SimTLiX (approved):

ofornaroli (dismissed):

nicolasbellanich-simtlix (approved):

JulioAchaval-SimTLiX (approved):

modules/account-lib/src/coin/algo/utils.ts
charly-simtlix: code was removed

modules/statics/test/unit/coins.ts
```diff
@@ -10,6 +10,16 @@ describe('CoinMap', function () {
   it('should have iterator', function () {
     [...coins].length.should.be.greaterThan(100);
   });
+
+  it('should report if it contains coin', () => {
```
tylerlevine, line 14: What do these tests have to do with the change being made?

modules/statics/src/coins.ts
```diff
@@ -40,7 +40,19 @@ const EOS_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.SUPPORTS_TOKE
 const SOL_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.REQUIRES_RESERVE, CoinFeature.SUPPORTS_TOKENS];
 const NEAR_FEATURES = [...AccountCoin.DEFAULT_FEATURES];
 
-export const coins = CoinMap.fromCoins([
+// Alorand is migration the name for the tokens, this are the equivalences to support new names during transition
+// STLX-10016 will revert this to support old names when new names were released
+const algorandKeyEquivalences = new Map([
```
tylerlevine, line 45: Can we add a new coin type that's an `aliasCoin` or something like that, which just redirects to another coin name? This approach feels quite hacky.

tylerlevine (changes requested):

modules/statics/test/unit/coins.ts
charly-simtlix: method `has` was added to CoinMap to check for the existence of some name in the map.

modules/statics/src/coins.ts
charly-simtlix: Added field alias to coin definition so it can be also be mapped by its alias. I prefer this change than add a `aliasCoin` to avoid duplication in case of listings.

modules/statics/src/base.ts
```diff
@@ -800,6 +801,11 @@ export abstract class BaseCoin {
   public readonly name: string;
   public readonly prefix?: string;
   public readonly suffix?: string;
+  /*
+    Property to help during migration of token names.
+    Helps to find a coin/token with a name different from current one
```
ofornaroli, line 806:
```suggestion
    Helps to find a coin/token with a different name from current one
```

ofornaroli (changes requested):

JulioAchaval-SimTLiX (approved):

ofornaroli (dismissed):
LGTM

modules/statics/src/base.ts
charly-simtlix: done

modules/statics/src/map.ts
```diff
@@ -3,33 +3,46 @@ import { DuplicateCoinDefinitionError, CoinNotDefinedError } from './errors';
 
 export class CoinMap {
   private readonly _map = new Map<string, Readonly<BaseCoin>>();
-
-  private constructor() {}
```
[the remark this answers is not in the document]
charly-simtlix: good catch, It was not the intention

modules/statics/src/map.ts
```diff
@@ -3,33 +3,46 @@ import { DuplicateCoinDefinitionError, CoinNotDefinedError } from './errors';
...
+  // Holds key equivalences used during an asset name migration
+  private readonly _coinByAliases = new Map<string, Readonly<BaseCoin>>();
```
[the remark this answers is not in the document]
charly-simtlix: Using same map for keys for names and aliases has the problem that if the map is used to list all assets, those assets with an alias will appear twice and that could impact several places. Separating the maps will keep the original semantics in the original map.

modules/statics/src/map.ts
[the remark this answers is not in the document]
charly-simtlix: fixed

tylerlevine (approved):

PR #2107, opened 2022-03-15
fix(account-lib): default goerli for terc20 token

sachushajiabraham:
make goerli as a default testnet coin for terc20 instead
of kovan as it is deprecating

BG-42778


tylerlevine (changes requested):
please check tests

PR #2108, opened 2022-03-15
fix(core): support eip-1559 and eip-155 in wrw

sachushajiabraham:
add eip-1559 and replay protection options for unsigned json
which enables to properly construct replay protected eth tx from
ovc

Ticket: BG-44045

### DIFF
--- a/modules/core/src/v2/coins/eth.ts
+++ b/modules/core/src/v2/coins/eth.ts
@@ -123,13 +123,20 @@ interface Recipient {
   data?: string;
 }
 
+interface EIP1559 {
+  maxPriorityFeePerGas: number;
+  maxFeePerGas: number;
+}
+
+interface ReplayProtectionOptions {
+  chain: string | number;
+  hardfork: string;
+}
+
 interface SignFinalOptions {
   txPrebuild: {
-    eip1559?: { maxPriorityFeePerGas: number; maxFeePerGas: number };
-    replayProtectionOptions?: {
-      chain: string | number;
-      hardfork: string;
-    };
+    eip1559?: EIP1559;
+    replayProtectionOptions?: ReplayProtectionOptions;
     gasPrice?: string;
     gasLimit: string;
     recipients: Recipient[];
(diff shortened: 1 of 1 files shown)


barathcj (dismissed):

modules/core/src/v2/coins/eth.ts
```diff
@@ -207,11 +219,8 @@ export interface RecoverOptions {
   krsProvider?: string;
   gasPrice?: number;
   gasLimit?: number;
-  eip1559?: { maxPriorityFeePerGas: number; maxFeePerGas: number };
-  replayProtectionOptions?: {
-    chain: string | number;
-    hardfork: string;
-  };
+  eip1559?: EIP1599;
+  replayProtectionOptions?: ReplayProtectionOptions;
```
mmcshinsky-bitgo, line 223: Can we update BuildTransactionParams on line 226 to reflect and standardize the use of the newly added interfaces?

sachushajiabraham: Good catch @mmcshinsky-bitgo , Updated it.

modules/core/src/v2/coins/eth.ts
```diff
@@ -123,13 +123,20 @@ interface Recipient {
   data?: string;
 }
 
+interface EIP1599 {
```
tylerlevine, line 126: EIP number is 1559, not 1599

tylerlevine (changes requested):
just one minor fix, otherwise lgtm

modules/core/src/v2/coins/eth.ts
```diff
@@ -822,16 +826,19 @@ export class Eth extends BaseCoin {
...
+   * @param eip1559
+   * @param replayProtectionOptions
+   * @returns {Promise<OfflineVaultTxInfo>}
    */
   async formatForOfflineVault(
     txInfo: UnformattedTxInfo,
     ethTx: EthTxLib.Transaction | EthTxLib.FeeMarketEIP1559Transaction,
     userKey: string,
     backupKey: string,
     gasPrice: Buffer,
```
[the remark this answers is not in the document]
sachushajiabraham: Left it as is since I am not sure whether any eth like chains which we support is not implementing eip 1559

wesgraham (approved):

PR #2110, opened 2022-03-16
fix(utxo-lib): always verify ECDSA in strict mode

OttoAllmendinger:
Per consensus rules, signatures must have a lower-S value

See https://github.com/bitcoin/bips/blob/master/bip-0146.mediawiki for 
details.

This change replaces the indirect call to `ecc.verify()` inside
`ECPair.verify()`[^1] with a direct call that sets the last argument
`strict` to `true`.

Note that this strict (low-S) verification is the default in the
`noble-secp256k1` library[^2], which is the prospective secp256k1 library going
forward.

Issue: BG-44787

----

adba1c287

test(utxo-lib): add test  for high-S ECDSA signatures

Add `verifySignature()` test where we take an existing signature and tweak
the signature `S` value to `n - S` (high-S).

The resulting signature is valid according to old consensus rules, but 
invalid according BIP-0146 which requires a low-S value.

https://github.com/bitcoin/bips/blob/master/bip-0146.mediawiki

Note that the test will fail in this commit, showing that we currently do
not reject high-S signatures. The fix is included in the next commit.

Issue: BG-44787

----

[^1]: https://github.com/bitcoinjs/ecpair/blob/d35a64c/ts_src/ecpair.ts#L215
[^2]: https://github.com/paulmillr/noble-secp256k1/blob/97aa518/index.ts#L1212

### DIFF
--- a/modules/utxo-lib/src/bitgo/signature.ts
+++ b/modules/utxo-lib/src/bitgo/signature.ts
@@ -1,8 +1,9 @@
 import * as opcodes from 'bitcoin-ops';
 import * as bip32 from 'bip32';
 
+const ecc = require('tiny-secp256k1');
+
 import {
-  ECPair,
   payments,
   script,
   Transaction,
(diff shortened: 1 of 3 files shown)


brandonblack (approved):
Nice.

PR #2111, opened 2022-03-16
feat: add nft tokens to statics

ryan-bitgo:
Add supports for nft tokens to statics.

We add two testnet tokens: a test ERC721 and a test ERC1155 and we add a ERC721 token in mainnet: Crypto Covens which will be using for testing.

https://opensea.io/collection/cryptocoven

ticket: BG-44467

### DIFF
--- a/modules/statics/src/account.ts
+++ b/modules/statics/src/account.ts
@@ -134,6 +134,20 @@ export class Base58ContractAddressDefinedToken extends AccountCoinToken {
 export class Erc20Coin extends ContractAddressDefinedToken {}
 
 /**
+ * ERC 721 is the non fungible token standard for the Ethereum blockchain.
+ *
+ * {@link https://eips.ethereum.org/EIPS/eip-721 EIP721}
+ */
+export class Erc721Coin extends ContractAddressDefinedToken {}
+
+/**
+ * ERC 1155 is the multi token standard for the Ethereum blockchain.
+ *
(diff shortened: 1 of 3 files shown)


modules/statics/src/account.ts
```diff
@@ -409,6 +411,49 @@ export function terc20(
   return erc20(name, fullName, decimalPlaces, contractAddress, asset, features, prefix, suffix, network);
 }
 
+export function nft(
```
mmcshinsky-bitgo, line 414: Would it make sense to follow the naming convention of the other functions that are similar to somthing like `erc721()` and `terc721()`? Or is this more agnostic?

mmcshinsky-bitgo (changes requested):

modules/statics/src/base.ts
```diff
@@ -764,6 +764,13 @@ export enum UnderlyingAsset {
   'avaxc:usdt' = 'avaxc:usdt',
   'avaxc:usdc' = 'avaxc:usdc',
   'avaxc:link' = 'avaxc:link',
+
+  // ERC721
+  ERC721 = 'erc721',
+  TERC721 = 'terc721',
```
tylerlevine, line 770: remove testnet versions - they are not separate underlying assets

ryan-bitgo:
addressed feedback

tylerlevine (approved):

mmcshinsky-bitgo (approved):

PR #2112, opened 2022-03-16
feat: update params to post /signatureshares

zahin-mohammad:
In preparation for TSS HD Wallets, we now need an optional `path` parameter when signing transactions with signature shares.

### DIFF
--- a/modules/core/src/v2/internal/tssUtils.ts
+++ b/modules/core/src/v2/internal/tssUtils.ts
@@ -396,12 +396,21 @@ export class TssUtils extends MpcUtils {
    *
    * @param {String} txRequestId - the txRequest Id
    * @param {SignatureShareRecord} signatureShare - a Signature Share
+   * @returns {string} path - a bip32 path, required when signing for a derived address.
+   * When the path is not provided, signing will happen with the root wallet address.
    * @returns {Promise<SignatureShareRecord>} - a Signature Share
    */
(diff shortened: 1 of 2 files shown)


modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -398,10 +398,17 @@ export class TssUtils extends MpcUtils {
    * @param {SignatureShareRecord} signatureShare - a Signature Share
    * @returns {Promise<SignatureShareRecord>} - a Signature Share
    */
-  async sendSignatureShare(txRequestId: string, signatureShare: SignatureShareRecord): Promise<SignatureShareRecord> {
+  async sendSignatureShare(
+    txRequestId: string,
+    signatureShare: SignatureShareRecord,
+    path?: string
```
joey-bai, line 404: add jsdoc for this param.

if this is undefined, do we assume that this is for the root wallet?

joey-bai (approved):

modules/core/src/v2/internal/tssUtils.ts
zahin-mohammad: Yes, if it's undefined, it's the root wallet. Will add some comments my morning tomorrow.

zahin-mohammad: Fixed

joey-bai (approved):

PR #2113, opened 2022-03-16
fix: update outdated package-lock.json version and .env updates

mmcshinsky-bitgo:
* Update package-lock.json to be properly formatted to required npm version.
* Fix incoming variable checks to both fail the API and succeed the linting rules.
* Allow and ignore empty key/value pairs for .env file.

[BG-44852](https://bitgoinc.atlassian.net/browse/BG-44852)

### DIFF
--- a/modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
+++ b/modules/account-lib/src/coin/baseCoin/baseTransactionBuilder.ts
@@ -1,6 +1,6 @@
 import BigNumber from 'bignumber.js';
 import { BaseCoin as CoinConfig } from '@bitgo/statics';
-import { BaseAddress, BaseKey, PublicKey } from './iface';
+import { BaseAddress, BaseKey, PublicKey, ValidityWindow } from './iface';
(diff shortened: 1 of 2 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -393,16 +392,25 @@ function decryptPrivKey(bg: BitGo, encryptedPrivKey: string, walletPw: string):
...
+    throw new Error('Missing required configuration: signerFileSystemPath');
+  }
+
+  const encryptedPrivKey = await getEncryptedPrivKey(signerFileSystemPath, walletId);
   const bitgo = req.bitgo;
   const privKey = decryptPrivKey(bitgo, encryptedPrivKey, walletPw);
   const coin = bitgo.coin(req.params.coin);
   try {
-    return await coin.signTransaction({ ...req.body, prv: privKey });
+    await coin.signTransaction({ ...req.body, prv: privKey });
```
tylerlevine, line 413: why delete the `return` here?

tylerlevine (changes requested):
looks good to me, just one question

PR #2114, opened 2022-03-16
fix(core): rename token to tokenName, possible clash with auth token for algo

jimmyqi-bitgo:


modules/core/src/v2/internal/tssUtils.ts
```diff
@@ -362,7 +362,7 @@ export class TssUtils extends MpcUtils {
         comment: params.comment,
         recipients: intentRecipients,
         memo: params.memo?.value,
-        token: params.token,
+        token: params.tokenName,
```
jimmyqi-bitgo, line 365: this is wrapped in intent, don't need change

ofornaroli (commented):
I agree with the changes. but did you test it into your local lumina?. Ping me if you need some help.

modules/core/test/v2/unit/wallet.ts
```diff
@@ -1599,7 +1599,7 @@ describe('V2 Wallet:', function () {
 
       it('should build an enable token transaction', async function () {
         const recipients = [];
-        const token = 'tokenName';
+        const tokenName = 'tokenName';
```
charly-simtlix, line 1602: avoid using value as same variable name, this can hide potential error. You can use anything different like: 'tokenName-123'

jimmyqi-bitgo: changed

ofornaroli (approved):
LGTM.

this were the steps that I did to test it:
I checked out in your branch on BitgoJS project.
I added a console.log('jimmy qi') inside prebuildTransaction method
I run build and compile commands
I run npm run proxy test on lumina and it works.

those images show a token enabled with the SDK pulled from this branch 
image (1)
image

PR #2126, opened 2022-03-22
chore(root): resolve high vulnerability

bitgoAaron:
- adds resolution for minimist

Ticket: BG-45291

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "glob": "^7.1.6",
     "husky": "^4.3.0",
     "lerna": "^4.0.0",
+    "minimist": "1.2.6",
     "make-dir": "^3.1.0",
     "mocha": "^9.2.0",
     "mochawesome": "^7.0.1",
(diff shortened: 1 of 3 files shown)


mmcshinsky-bitgo:
@bitgoAaron  New web-stream-tools types in openpgp@5.2.1 is breaking our implementation of encrypt and the yarn install when the unit-test removes the yarn.lock file. Possibly lock the library from ^5.1.0 to 5.1.0?

...or we could update the functions if we want the newer version.

package.json
```diff
@@ -16,6 +16,7 @@
     "glob": "^7.1.6",
     "husky": "^4.3.0",
     "lerna": "^4.0.0",
+    "minimist": "1.2.6",
```
tylerlevine, line 19: do we need this in dev deps?

tylerlevine (changes requested):
one small comment

package.json
bitgoAaron: well it's not used by the app directly, and only exists to satisfy yarn audit...so yes?

tylerlevine (approved):

tylerlevine:
Ref: https://github.com/substack/minimist/issues/164

PR #2135, opened 2022-03-23
chore(statics): change Solana token names to lower case

jimmyqi-bitgo:
STLX-14417

Similar to https://bitgoinc.atlassian.net/browse/STLX-13674, the avax rename.
* Indexer returns token names in lower case, change this for consistency
* Avax has seen many errors which are fixed by renaming tokens to lower case, this is also a preemptive measure for solana


mmcshinsky-bitgo (approved):
@jimmyqi-bitgo @tylerlevine With this move forward, can we consider this a breaking change as we have set names that are now changing and we don't control how third parties are handling case-insensitive name checking?

joey-bai (approved):

PR #2138, opened 2022-03-24
feat(account-lib): add solana util functions for use in wp, refactor

jimmyqi-bitgo:
STLX-14417

Add `getAssociatedTokenAccountAddress` to be used in wp, this avoids adding a dependency on spl-token in wp. 

Also refactor token builders to use new util methods


joey-bai (approved):
nice

PR #2141, opened 2022-03-25
feat(account-lib): add staking deactivate builder

ipelaez-simtlix:
add staking deactivate builder and unit test.

[stlx-14826](https://bitgoinc.atlassian.net/browse/STLX-14826)


modules/account-lib/src/coin/near/transaction.ts
```diff
@@ -101,16 +101,19 @@ export class Transaction extends BaseTransaction {
...
       try {
-        this._nearTransaction = nearAPI.utils.serialize.deserialize(
+        const unsignedTx = nearAPI.utils.serialize.deserialize(
           nearAPI.transactions.SCHEMA,
           nearAPI.transactions.Transaction,
           nearAPI.utils.serialize.base_decode(rawTransaction),
         );
+        unsignedTx.nonce = parseInt(unsignedTx.nonce.toString(), 10);
+        this._nearTransaction = unsignedTx;
         this._id = utils.base58Encode(this.getTransactionHash());
```
ipelaez-simtlix, line 117: Please pay attention to this part, there's something weird here, we have found that toJson() return nonce wrong, because when you use from() it returns nonce as BN, it happen because deserialize util method from api-near-js return BN instead of number as mentioned before, which is inconsistent because currently near treat nonce as a number.

silalvado (approved):

ewangbitgo (approved):

PR #2145, opened 2022-03-25
fix: update package-lock.json and clientRoutes

mmcshinsky-bitgo:
- Update package-lock.json to be properly formatted to required npm version.
- Fix incoming variable checks to both fail the API and succeed the linting rules.
- Allow and ignore empty key/value pairs for .env file.
- Update tests to pass linting rules.
- Update docker file node requirements.

BG-44852


modules/express/src/clientRoutes.ts
```diff
@@ -53,11 +52,12 @@ function handlePingExpress(req: express.Request) {
   };
 }
 
-function handleLogin(req: express.Request) {
+async function handleLogin(req: express.Request) {
   const username = req.body.username || req.body.email;
   const body = req.body;
   body.username = username;
-  return req.bitgo.authenticate(body);
+  const results = await req.bitgo.authenticate(body);
```
bitgoAaron, line 59: does this need to be async if it's just awaiting & immediately returning?

bitgoAaron (dismissed):
just nit about changing the method signature to async if it's not actually doing anything async

bitgoAaron (approved):

PR #2149, opened 2022-03-29
STLX-13365-feat(statics): Create FIAT-USD on SDK

dkotlirevsky-simtlix:
statics:
		* Add fiat coin to coins
		* Create Class to represent fiat network
		* Add tests for fiat
		* Add tests for tsusd

STLX-13365

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -22,6 +22,7 @@ import {
   tfiatToken,
   terc721,
   erc721,
+  fiatToken,
 } from './account';
 import { CoinFeature, CoinKind, KeyCurve, UnderlyingAsset } from './base';
 import { CoinMap } from './map';
@@ -1382,4 +1383,5 @@ export const coins = CoinMap.fromCoins([
     AccountCoin.DEFAULT_FEATURES
   ),
   tfiatToken('tfiat:USD', 'FIAT US Dollar', UnderlyingAsset.USD, 2),
+  fiatToken('fiat:USD', 'FIAT US Dollar', UnderlyingAsset.USD, 2),
 ]);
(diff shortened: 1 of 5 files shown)


modules/statics/src/coins.ts
```diff
@@ -1382,4 +1383,5 @@ export const coins = CoinMap.fromCoins([
     AccountCoin.DEFAULT_FEATURES
   ),
   tfiatToken('tfiat:USD', 'FIAT US Dollar', UnderlyingAsset.USD, 2),
+  fiatToken('fiat:USD', 'FIAT US Dollar', UnderlyingAsset.USD, 2, Networks.main.fiat),
```
drew-bitgo, line 1386: Should network not be implied for `fiatToken`, since we differentiate between then with a `tfiatToken` and `fiatToken`? Would expect a default value here. Looking at all the other definitions, there is only one case where a non-default network is used, and thats georili ETH, which makes sense. We wont have a Fiat thats not on the fiat network. And if we do, it will be an exception, not the standard.

drew-bitgo (changes requested):

modules/statics/src/coins.ts
dkotlirevsky-simtlix: Ok, but as it's not defined into the fiatToken method ; Which should be the default value in this case ? maybe @emiacc  can help here too.

drew-bitgo: I would expect `Networks.main.fiat` would be the default network for fiat tokens

drew-bitgo (approved):

tylerlevine (approved):

mmcshinsky-bitgo (approved):

PR #2153, opened 2022-03-29
chore: update actions to v4

mmcshinsky-bitgo:
Update to v4. Documentation states to put permissions in .yml file. Might be different if the settings for the repo need to be changes as well for workflows.

BG-000000


ducnguyen-bitgo (approved):
Lgtm. Just a [reference](https://github.com/actions/stale#recommended-permissions) to the recommended permissions from actions stale.

PR #2160, opened 2022-03-31
fix(core): fix avaxctoken cannot withdraw

jicli:
Becasuse missing this method, it uses AVAXC verifyCoin
Ticket: BG-0


tylerlevine (changes requested):
test?

PR #2163, opened 2022-03-31
chore: update dot tests to use tss way of signing

lcovar:
Updates dot tss tests to use `addSignature` instead of `.sign` since we're now using TSS to add a signature to a dot transaction.

TICKET: BG-45942

### DIFF
--- a/modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts
+++ b/modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts
@@ -176,56 +176,6 @@ describe('Dot Transfer Builder', () => {
     });
   });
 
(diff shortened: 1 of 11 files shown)


modules/account-lib/test/unit/coin/dot/transactionBuilder/base.ts
```diff
@@ -176,56 +176,6 @@ describe('Dot Transfer Builder', () => {
     });
   });
 
-  describe('add signature', () => {
```
lcovar, line 179: removing this test as we other tests are implicitly testing this when using `.addSignature` to sign a transaction.

joey-bai (approved):

PR #2166, opened 2022-03-31
test(bitgo): integration test for ETH2 BLS keychains message signing

fperalta-simtlix:
Integration test for ETH2 BLS keychains signing methods. We should be able to sign a message and
verify its signature.

BG-36125


wesgraham (approved):
This is awesome! Would we also be able to test with user<>bitgo key (using array indices 0 and 2) - just as a sanity check!

fperalta-simtlix:
> 

That was my first thought, but of course by design bitgo private key should not be exposed.

PR #2174, opened 2022-04-01
feat(bitgo): update consolidation flow

Krin2:
update consolidation flow

STLX-14666


rliberal-simtlix (commented):
Add a consolidation test with TSS

PR #2180, opened 2022-04-04
fix: v1 wallet cross chain recovery

DariusParvin:
Currently, bitgo throws a 400 error for v1 wallet Ids that are not found in the `api/v2/:coin/wallets/:walletId` route, causing cross chain recovery for v1 wallets to fail. This commit expands the range of error responses to prevent similar errors occurring in the future.

A test for this will be created in another ticket (BG-46179)

Ticket: BG-44440


jimmyqi-bitgo (approved):
lgtm

DariusParvin:
Thanks @jimmyqi-bitgo !

PR #2186, opened 2022-04-06
chore(express): update bitgo version

jicli:
Ticket: BG-0


mmcshinsky-bitgo:
I don't believe we do any `-rc` dependencies directly on the main branch except for a couple one-off exceptions. Those are usually in the `rel/rc` branch `rc` tag environment until the next stable release.

PR #2188, opened 2022-04-07
chore(lumina): improve error logging for hop tx

zhongtan912:
Ticket: BG-43405


wesgraham (commented):
Can we log it as follows: Hop txid signature invalid, pub: ${serverXpub}, msg: ${messageBuffer}, sig: ${sig}

wesgraham (approved):

PR #2190, opened 2022-04-07
refactor(core): refactor isWalletAddress for CSPR

alebusse:
refactored isWalletAddress for CSPR    
Fixed unit tests

[BG-43225](https://bitgoinc.atlassian.net/browse/BG-43225)

### DIFF
--- a/modules/bitgo/src/v2/coins/cspr.ts
+++ b/modules/bitgo/src/v2/coins/cspr.ts
@@ -15,12 +15,13 @@ import {
   TransactionExplanation,
   ParseTransactionOptions,
   ParsedTransaction,
-  VerifyAddressOptions as BaseVerifyAddressOptions,
+  VerifyAddressOptions,
 } from '../baseCoin';
 
 import { BitGo } from '../../bitgo';
 import { BaseCoin as StaticsBaseCoin, CoinFamily } from '@bitgo/statics';
 import { InvalidAddressError, InvalidTransactionError, UnexpectedAddressError } from '../../errors';
+import { KeyIndices } from '..';
 
 interface SignTransactionOptions extends BaseSignTransactionOptions {
   txPrebuild: TransactionPrebuild;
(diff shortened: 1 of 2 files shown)


modules/bitgo/src/v2/coins/cspr.ts
```diff
@@ -106,17 +107,20 @@ export class Cspr extends BaseCoin {
    * @param {VerifyAddressOptions} params address and rootAddress to verify
    */
   isWalletAddress(params: VerifyAddressOptions): boolean {
-    const { address, rootAddress } = params;
+    const { address, keychains } = params;
```
mmcshinsky-bitgo, line 110: Looking at release 14.0.0 of `bitgo`, this looks like a breaking change in past to future usage. Think we should label this with the `BREAKING CHANGE` label?

alebusse: Based on the idea that this method should validate  that an `address` is from a wallet using the keys or in a cryptographic manner (instead of getting the `rootAddress` from the `walletDoc`), and that the keys are in the interface of the method. 
Its something being work for many coins, but im not sure if its a breaking change or its something unused yet until every coin gets implemented or refactored.

mmcshinsky-bitgo: Since it is open source, even if it isn't a breaking change because internally we update everything together, it would be a breaking change requiring external companies to update their code upon a stable release.

alebusse: Check this https://github.com/BitGo/BitGoJS/blob/master/modules/bitgo/src/v2/wallet.ts#L1497-L1523, it has a TODO, i already made changes to 2 coins before this one and there are other coins that need implementation or refactor.
I think its a method only used internally, but its not private atm.

CapnMigraine (approved):

rliberal-simtlix (dismissed):

mmcshinsky-bitgo (dismissed):

PR #2192, opened 2022-04-07
fix(bitgo): send passcodeEncryptionCode to fix mpc wallet pw reset

pranavjain97:
Ticket: BG-00000

### DIFF
--- a/modules/bitgo/src/v2/internal/tssUtils.ts
+++ b/modules/bitgo/src/v2/internal/tssUtils.ts
@@ -116,6 +116,7 @@ export class TssUtils extends MpcUtils {
    * @param backupKeyShare - backup's TSS key share
    * @param bitgoKeychain - previously created BitGo keychain; must be compatible with user and backup key shares
    * @param passphrase - wallet passphrase used to encrypt user's signing materials
+   * @param [originalPasscodeEncryptionCode] - optional encryption code needed for wallet password reset for hot wallets
    */
   async createUserKeychain(
     userGpgKey: openpgp.SerializedKeyPair<string>,
--- a/modules/bitgo/src/v2/keychains.ts
+++ b/modules/bitgo/src/v2/keychains.ts
@@ -380,6 +380,10 @@ export class Keychains {
     }
 
     const mpcUtils = new MpcUtils(this.bitgo, this.baseCoin);
-    return await mpcUtils.createKeychains({ passphrase: params.passphrase });
+    return await mpcUtils.createKeychains({
+      passphrase: params.passphrase,
+      enterprise: params.enterprise,
+      originalPasscodeEncryptionCode: params.originalPasscodeEncryptionCode,
+    });
   }
 }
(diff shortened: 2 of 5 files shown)


modules/bitgo/src/v2/internal/tssUtils.ts
```diff
@@ -116,14 +116,15 @@ export class TssUtils extends MpcUtils {
...
    * @param passphrase - wallet passphrase used to encrypt user's signing materials
+   * @param [originalPasscodeEncryptionCode]
    */
   async createUserKeychain(
     userGpgKey: openpgp.SerializedKeyPair<string>,
     userKeyShare: KeyShare,
     backupKeyShare: KeyShare,
     bitgoKeychain: Keychain,
     passphrase: string,
-    originalPasscodeEncryptionCode?: string
```
pranavjain97, line 126: Not sure if I should make this mandatory right away? Incase it's not passed in for some reason, wouldn't want it to block wallet creation. Thought we can do it after April 11.

pranavjain97:
Note: Haven't updated the BLS tests to pass this in for now.

joey-bai (approved):

modules/bitgo/src/v2/keychains.ts
```diff
@@ -380,6 +380,9 @@ export class Keychains {
     }
 
     const mpcUtils = new MpcUtils(this.bitgo, this.baseCoin);
-    return await mpcUtils.createKeychains({ passphrase: params.passphrase });
+    return await mpcUtils.createKeychains({
+      passphrase: params.passphrase,
+      originalPasscodeEncryptionCode: params.originalPasscodeEncryptionCode,
+    });
```
joey-bai, line 386: while you're here, could you pass in `params.enterprise` too?

pranavjain97: done

joey-bai (dismissed):

modules/bitgo/src/v2/internal/tssUtils.ts
pranavjain97: Making this mandatory to be consistent with api's requirement for TSS user key.

modules/bitgo/src/v2/internal/tssUtils.ts
```diff
@@ -314,7 +315,7 @@ export class TssUtils extends MpcUtils {
       backupKeyShare,
       bitgoKeychain,
       params.passphrase,
-      params.originalPasscodeEncryptionCode
+      params.originalPasscodeEncryptionCode!
```
joey-bai, line 318: we should update `originalPasscodeEncryptionCode?: string;` in input to be non-optional

joey-bai (approved):

modules/bitgo/src/v2/internal/tssUtils.ts
pranavjain97: oh tru. i did this cus its not needed for backup/bitgo, but since all of them are created under this fn its fine.

joey-bai (approved):

PR #2195, opened 2022-04-11
update moment and polkadot

lcovar:
TICKET: BG-0000

### DIFF
--- a/modules/account-lib/test/unit/coin/dot/transactionBuilder/transferBuilder.ts
+++ b/modules/account-lib/test/unit/coin/dot/transactionBuilder/transferBuilder.ts
@@ -114,7 +114,7 @@ describe('Dot Transfer Builder', () => {
       should.deepEqual(txJson.tip, 0);
       should.deepEqual(txJson.transactionVersion, txVersion);
       should.deepEqual(txJson.chainName, chainName);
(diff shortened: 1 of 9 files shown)


modules/account-lib/test/unit/coin/dot/transactionBuilder/transferBuilder.ts
```diff
@@ -114,7 +114,7 @@ describe('Dot Transfer Builder', () => {
       should.deepEqual(txJson.tip, 0);
       should.deepEqual(txJson.transactionVersion, txVersion);
       should.deepEqual(txJson.chainName, chainName);
-      should.deepEqual(txJson.eraPeriod, 4);
+      should.deepEqual(txJson.eraPeriod, 0);
```
lcovar, line 117: When maxDuration value of `0` is passed when building a transaction, we know we want an immortal transaction. polkadot `17.15.1` added a new flag `isImmortalEra` to fix previous version where immortal transactions were being build incorrectly. When that flag is passed the `eraPeriod` is `0` which is the correct behavior now.

mmcshinsky-bitgo (dismissed):

mmcshinsky-bitgo (approved):

PR #2198, opened 2022-04-11
dot fix address format

lcovar:


byhow (approved):

modules/account-lib/src/coin/baseCoin/enum.ts
```diff
@@ -49,6 +49,10 @@ export enum TransactionType {
 export enum AddressFormat {
   hex = 'hex',
   base58 = 'base58',
+  // format for westend addresses
```
aaldemir, line 52: If this `enum` is supposed to be coin agnostic, what do you think about creating a DOT specific `AddressFormat` and [creating a union](https://stackoverflow.com/questions/52857698/extending-enum-in-typescript) of the two `enums` (since you can not extend `enums` in ts)

modules/account-lib/src/coin/dot/keyPair.ts
```diff
@@ -34,9 +34,19 @@ export class KeyPair extends Ed25519KeyPair {
...
+  /**
+   // https://wiki.polkadot.network/docs/learn-accounts#address-format
+   * Returns the address in either mainnet polkadot format (starts with 1)
+   * or substrate format used for westend (starts with 5)
+   */
   getAddress(format?: AddressFormat): string {
-    return this.createPolkadotPair().address;
+    const substrateAddress = this.createPolkadotPair().address;
+    // generate polkadot (mainnet) address format
+    if (format && format === AddressFormat.polkadot) {
```
aaldemir, line 45: Is there a way to use a similar pattern to the one in WP where we have a `tdot` controller and a `dot` controller? If so, the `tdot KeyPair` can override the `dot KeyPair` class method `getAddress`

modules/account-lib/src/coin/baseCoin/enum.ts
lcovar: Good call and ran through my mind as well and agree it should be done the way you suggested. I can do this on a follow up. I would create a ticket but you know.

modules/account-lib/src/coin/dot/keyPair.ts
lcovar: I see where you're going with this. Need to let this one simmer and see if there is a way to do it this way instead.

PR #2199, opened 2022-04-11
feat(statics): add avaxP

pvalles-simtlix:
create statics for avaxp

STLX-15063

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -100,6 +100,23 @@ class AvalancheCTestnet extends Testnet implements AccountNetwork {
   chainId = 43113;
 }
 
+class AvalancheP extends Mainnet implements UtxoNetwork {
+  name = 'AvalancheP';
+  family = CoinFamily.AVAXP;
+  utxolibName = 'avax';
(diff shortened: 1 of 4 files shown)


modules/statics/src/networks.ts
```diff
@@ -100,6 +100,23 @@ class AvalancheCTestnet extends Testnet implements AccountNetwork {
   chainId = 43113;
 }
 
+class AvalancheP extends Mainnet implements UtxoNetwork {
+  name = 'AvalancheP';
+  family = CoinFamily.AVAXP;
+  utxolibName = 'avax';
```
jrafael-simtlix, line 106: @shiksb I don't know the use of this field.

modules/statics/src/utxo.ts
```diff
@@ -61,6 +61,7 @@ export function utxo(
   fullName: string,
   network: UtxoNetwork,
   asset: UnderlyingAsset,
+  decimalPlaces: number = 8,
```
mmcshinsky-bitgo, line 64: This looks to be taken care of already on the constructor of UtxoCoin?

PR #2202, opened 2022-04-12
fix(statics): update base factor for dot and tdot

lcovar:
The base factor defaults to 1e12 which is only valid for
Westend transaction since 12 decimal places is used vs 10
decimal places in mainnet.

TICKET: BG-0000

### DIFF
--- a/modules/bitgo/src/v2/coins/tdot.ts
+++ b/modules/bitgo/src/v2/coins/tdot.ts
@@ -1,14 +1,22 @@
 import { BitGo } from '../../bitgo';
 import { BaseCoin } from '../baseCoin';
 import { Dot } from './dot';
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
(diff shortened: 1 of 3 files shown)


YPAzevedo (approved):

modules/bitgo/src/v2/coins/tdot.ts
```diff
@@ -18,4 +26,8 @@ export class Tdot extends Dot {
   getFullName(): string {
     return 'Testnet Polkadot';
   }
+
+  getBaseFactor(): number {
+    return Math.pow(10, this._staticsCoin.decimalPlaces);
```
dszheng92, line 31: Is this suppose to be 12 like in static?
https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/coins.ts#L78

dszheng92 (commented):
LGTM

PR #2206, opened 2022-04-13
chore: update support and owner references

mmcshinsky-bitgo:
Update support and codeowner references for SDK code ownership and contact points.

Looks like @BitGo/sdk-team needs to be fixed for assigned repositories. Same with eth-team? @BitGo/git-admins 

BG-000000

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,12 +2,12 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners and first officers will be the default owners for everything in the repo.
-*   tyler@bitgo.com
+*   @BitGo/sdk-team
 
 # Order is important. The last matching pattern has the most precedence.
 
 # Recovery
-modules/bitgo/src/v2/recovery.ts tyler@bitgo.com
+modules/bitgo/src/v2/recovery.ts @BitGo/sdk-team
 
 # Coin Support
 modules/bitgo/src/v2/coins/*btc.ts @BitGo/btc-team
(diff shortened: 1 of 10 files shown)


CODEOWNERS
```diff
@@ -2,12 +2,12 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners and first officers will be the default owners for everything in the repo.
-*   tyler@bitgo.com
+*   @BitGo/sdk-team
```
mmcshinsky-bitgo, line 5: Don't know if you want to keep this one as tyler@bitgo.com? @tylerlevine

modules/statics/CODEOWNERS
```diff
@@ -2,4 +2,4 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners and first officers will be the default owners for everything in the repo.
-*   tyler@bitgo.com
+*   @BitGo/sdk-team
```
tylerlevine, line 5: I think we can remove this file - only the root codeowners file is actually used

modules/bls-dkg/package.json
```diff
@@ -7,7 +7,7 @@
     "unit-test": "mocha",
     "test": "npm run unit-test"
   },
-  "author": "John Driscoll <john@bitgo.com>",
+  "author": "BitGo Inc.",
```
tylerlevine, line 10: How about this?
```suggestion
  "author": "BitGo SDK Team <sdkteam@bitgo.com>",
```

tylerlevine (changes requested):
looks pretty good, still need to sort out the team access issues.

A couple minor comments from me.

modules/bls-dkg/package.json
mmcshinsky-bitgo: Looks good to me. I'll make the changes.

tylerlevine (approved):

PR #2207, opened 2022-04-13
fix: whitelist nonce as an intent param

lcovar:
If nonce is passed as part of a transfer request then send that nonce value to wallet-platform to build the transaction accordingly. Tested in testnet-05 for dot successfully.


aseoane-simtlix (approved):
LGTM

PR #2220, opened 2022-04-15
fix(express): Fix docker build for BitGoJS Express

bitgoAaron:
- opts out of TS Build system for express package
- adds a CI pipeline step to ensure docker build will not regress


mmcshinsky-bitgo (approved):

tylerlevine (approved):
tested container build locally, everything looks good to me

nice work

PR #2221, opened 2022-04-15
fix: v1 get wallet

DariusParvin:
A bad fix was introduced in #2180, this is now the correct version. 

When `getWallet` is called, if `coin.wallets().get({ id: walletId })` returns a 4xx error, the correct behavior is to search for the wallet using the v1 endpoint.

Originally the condition was just `if (e.status !== 404) {` so that if the wallet wasn't found, it would search v1 wallets. The problem was the v1 endpoint was returning a 400 error. It's not necessarily incorrect because if v1 wallets are a different format to v2 wallets, that would be an appropriate response. So that's why the range of errors was expanded to any 4xx error

Ticket: BG-44440

### DIFF
--- a/modules/bitgo/src/v2/coins/utxo/recovery/crossChainRecovery.ts
+++ b/modules/bitgo/src/v2/coins/utxo/recovery/crossChainRecovery.ts
@@ -91,7 +91,7 @@ type WalletV1 = {
   getEncryptedUserKeychain(): Promise<{ encryptedXprv: string }>;
 };
 
-async function getWallet(bitgo: BitGo, coin: AbstractUtxoCoin, walletId: string): Promise<Wallet | WalletV1> {
+export async function getWallet(bitgo: BitGo, coin: AbstractUtxoCoin, walletId: string): Promise<Wallet | WalletV1> {
   try {
     return await coin.wallets().get({ id: walletId });
   } catch (e) {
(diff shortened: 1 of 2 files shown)


modules/bitgo/src/v2/coins/utxo/recovery/crossChainRecovery.ts
```diff
@@ -95,7 +95,7 @@ async function getWallet(bitgo: BitGo, coin: AbstractUtxoCoin, walletId: string)
   try {
     return await coin.wallets().get({ id: walletId });
   } catch (e) {
-    if (e.status >= 400 && e.status < 500) {
+    if (e.status < 400 || e.status >= 500) {
```
mmcshinsky-bitgo, line 98: I wouldn't expect a status < 400 to be an error. Those should be successful responses from the api, yes?

DariusParvin: It should only enter the catch block if the response is an error though, no?

mmcshinsky-bitgo: Status' below 400 aren't error reponses.

mmcshinsky-bitgo: If we are seeing an error below 400 in the catch block, we're dealing with a logic error that needs to be fixed.

DariusParvin: oh right, yes I get what you mean now. So perhaps instead it should be:

```suggestion
    if (e.status >= 500) {
```

DariusParvin: After discussing with @mmcshinsky-bitgo in slack, the plan is to:
- go with `e.status >= 500`, 
- add a comment to explain it in the code,
- create a ticket for the backlog to better gracefully fail different errors for different wallet versions

DariusParvin:
I added a test for `getWallet` in the latest commit.

tylerlevine (approved):

PR #2224, opened 2022-04-19
feat(account-lib): support creating TSS keyshares with seed

joey-bai:
This will be used by OVC, which generates some random entropy used
during key creation

Ticket: BG-43873

### DIFF
--- a/modules/account-lib/src/mpc/tss.ts
+++ b/modules/account-lib/src/mpc/tss.ts
@@ -138,12 +138,16 @@ export default class Eddsa {
     this.hdTree = hdTree;
   }
 
-  keyShare(index: number, threshold: number, numShares: number): KeyShare {
+  keyShare(index: number, threshold: number, numShares: number, seed?: Buffer): KeyShare {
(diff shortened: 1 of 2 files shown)


johnoliverdriscoll (approved):

modules/account-lib/src/mpc/tss.ts
```diff
@@ -138,12 +138,16 @@ export default class Eddsa {
...
+    }
+
+    const seedchain = seed ?? randomBytes(64);
+    const actualSeed = seedchain.slice(0, 32);
     const chaincode = seedchain.slice(32);
-    const h = createHash('sha512').update(seed).digest();
+    const h = createHash('sha512').update(actualSeed).digest();
     const u = clamp(bigIntFromBufferLE(h.slice(0, 32)));
     const y = Eddsa.curve.basePointMult(u);
     const split_u = Eddsa.shamir.split(u, threshold, numShares);
```
Logicwax, line 153: we should also support passing in this entropy into the shamir split function as well

PR #2227, opened 2022-04-21
fix(account-lib): test tss hsm signex tx

alebusse:
test tss hsm signex tx

STLX-0000


modules/account-lib/src/coin/near/utils.ts
```diff
@@ -73,6 +73,10 @@ export class Utils implements BaseUtils {
     return bs58.encode(value);
   }
 
+  base58Decode(value: string): Uint8Array {
```
mmcshinsky-bitgo, line 76: Will this be used for more than a test? If not, can we import bs58 into the test?

PR #2233, opened 2022-04-26
feat(statics): create Fiat and Tfiat coins

emiacc:
create Fiat and Tfiat coins to be used as backingCoin for `ofctusd` and `ofcusd`

We had added 'fiat tokens' (fiat:USD) but decided that we should also add the 'fiat coins' as they will support the creation of wallets.

STLX-13382

### DIFF
--- a/modules/bitgo/src/v2/coins/fiat.ts
+++ b/modules/bitgo/src/v2/coins/fiat.ts
@@ -1,6 +1,7 @@
 /**
  * @prettier
  */
+import { BitGo } from '../../bitgo';
 import {
   BaseCoin,
   KeyPair,
@@ -13,7 +14,14 @@ import {
 } from '../baseCoin';
 import { MethodNotImplementedError } from '../../errors';
 
-export abstract class Fiat extends BaseCoin {
+export class Fiat extends BaseCoin {
+  constructor(bitgo: BitGo) {
+    super(bitgo);
+  }
+
+  static createInstance(bitgo: BitGo): BaseCoin {
+    return new Fiat(bitgo);
+  }
   /**
    * Returns the factor between the base unit and its smallest subdivison
    * @return {number}
--- a/modules/bitgo/src/v2/coins/tfiat.ts
+++ b/modules/bitgo/src/v2/coins/tfiat.ts
@@ -1,0 +1,24 @@
+/**
+ * @prettier
+ */
+import { BitGo } from '../../bitgo';
+import { BaseCoin } from '../baseCoin';
+import { Fiat } from './fiat';
+
+export class Tfiat extends Fiat {
+  constructor(bitgo: BitGo) {
+    super(bitgo);
+  }
+
+  static createInstance(bitgo: BitGo): BaseCoin {
+    return new Tfiat(bitgo);
+  }
+
+  getChain() {
+    return 'tfiat';
+  }
+
+  getFullName() {
+    return 'Test Fiat';
+  }
+}
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -22,6 +22,7 @@ import {
   tfiatToken,
   terc721,
   erc721,
+  fiat,
   fiatToken,
 } from './account';
 import { CoinFeature, CoinKind, KeyCurve, UnderlyingAsset } from './base';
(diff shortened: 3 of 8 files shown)


modules/statics/src/coins.ts
```diff
@@ -1355,6 +1356,8 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.SLND,
     AccountCoin.DEFAULT_FEATURES
   ),
+  fiat('fiat', 'Fiat', UnderlyingAsset.USD, 2, Networks.main.fiat),
```
drew-bitgo, line 1359: Should the `UnderlyingAsset` here be USD? I imagine there should be a generic `UnderlyingAsset.FIAT` for this case, but maybe not? Not sure where that is used, so maybe its irrelevant for FIAT

tylerlevine (commented):
it's not really clear to me why we need both fiat tokens and fiat coins - could you expand on this a bit more. What's the difference between these?

emiacc:
> it's not really clear to me why we need both fiat tokens and fiat coins - could you expand on this a bit more. What's the difference between these?

@tylerlevine  
We discussed it with the Prime team, and we think that FIAT implementation should be like ETH, where ETH tokens use the ETH backing wallet, so FIAT tokens should use a FIAT wallet. 
We can't create FIAT wallets since there is no coin created, I'd only created tokens.
That is the propose of this PR

emiacc:
> it's not really clear to me why we need both fiat tokens and fiat coins - could you expand on this a bit more. What's the difference between these?

Comment from @drew-bitgo in the [ticket](https://bitgoinc.atlassian.net/browse/STLX-16000?focusedCommentId=264916):

The backing wallet for FIAT is currently susd. Since we only have one FIAT IMS (Fiat-Indexer), we would like to support all FIATs (currently only usd, but working on adding euro now), we need to update the backing wallet to be generic. The closest example to this is ETH and ETH-tokens. Using this example, we have decided to create a generic FIAT “coin” in the sdk, so that the backing wallet can utilize it, and create “Fiat Tokens” for our trading needs (so ofcusd for now (FIAT:USD), ofceuro (FIAT:EURO) in the future)

modules/bitgo/src/v2/coins/fiat.ts
```diff
@@ -13,7 +14,10 @@ import {
 } from '../baseCoin';
 import { MethodNotImplementedError } from '../../errors';
 
-export abstract class Fiat extends BaseCoin {
+export class Fiat extends BaseCoin {
+  static createInstance(bitgo: BitGo): BaseCoin {
```
mmcshinsky-bitgo, line 18: Missing constructor:

```
  constructor(bitgo: BitGo) {
    super(bitgo);
  }
```

modules/bitgo/src/v2/coins/tfiat.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * @prettier
+ */
+import { BitGo } from '../../bitgo';
+import { BaseCoin } from '../baseCoin';
+import { Fiat } from './fiat';
+
+export class Tfiat extends Fiat {
+  static createInstance(bitgo: BitGo): BaseCoin {
```
mmcshinsky-bitgo, line 9: Same as above.

mmcshinsky-bitgo (changes requested):

mmcshinsky-bitgo (approved):

mmcshinsky-bitgo:
Approved pending Tyler's comments.

tylerlevine:
Mechanically the change looks fine but I have a few questions on the approach here.

> we need to update the backing wallet to be generic
Why does it need to be generic? In other words, why create a FIAT backing wallet that can hold several fiat "tokens" which represent the actual currencies (eg, USD, EUR, GBP, JPY, etc) instead of a separate backing wallet for each of these currencies?

> The closest example to this is ETH and ETH-tokens

I'm not sure this analogy really works. One key feature of an asset (eg, something in `UnderlyingAsset`) is that it has a well defined price and that price can be fetched from a market data source. What's the price of the FIAT asset? It's unlike ETH and ERC20 tokens in this way (each of those has a well defined price).

I don't feel strong enough about this to block the PR - you likely have better insight on the usage of these new coins/tokens than I do. But please keep in mind that you're proposing a fairly significant break from the current modeling for fiat assets which may prove problematic later on down the road. If you still feel this is the right path forward, feel free to merge.

emiacc:
Closing this PR since we will implement a different solution

PR #2235, opened 2022-04-26
feat(bls-dkg): bLS-DKG HD support library implementation

fperalta-simtlix:
Implements HD in BLS-DKG lib so it can be used for the different SDK functionalities (e.g. ETH2
wallets creation)

BG-46338

### DIFF
--- a/modules/bls-dkg/index.js
+++ b/modules/bls-dkg/index.js
@@ -2,6 +2,9 @@
 'use strict'
 const bls = require('noble-bls12-381')
 const math = require('noble-bls12-381/math')
+const { createHmac } = require('crypto');
+
+const chaincodeBase = 1n << 256n;
 
 /** Convert a string to a buffer. */
 function stringToBuffer(string) {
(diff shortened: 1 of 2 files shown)


modules/bls-dkg/index.js
```diff
@@ -43,6 +44,75 @@ function bigIntToBuffer(bigint) {
...
+    const offsetMod = new math.Fq(offset);
+    const pkPoint = bls.PointG1.fromCompressedHex(pkBuf);
+    const publicChild = pkPoint.add(bls.PointG1.BASE.multiply(offsetMod));
+    return bufferToBigInt(publicChild.toCompressedHex());
+  }
+  return indices.reduce(deriveIndex, pk);
+}
+
+/** Derive a private key */
+function privateDerive(parentSk, path) {
```
wesgraham, line 105: will we have the parentSk ever aggregated in one place?

johnoliverdriscoll:
If I'm reading this correctly, the subkey offset is `sha256(pubkey || seri(index))`?

johnoliverdriscoll:
Is privacy a concern at all? A third party can easily brute force every address a customer generated only by knowing their public key. If customer privacy is a concern, we should also implement a chaincode that is only shared between the customer and bitgo. Something like `hmac_sha512(chaincode, pk || seri(i)) = zl || zr` where zl is interpreted as a 32 byte private key and zr is interpreted as a 32 byte chaincode.

modules/bls-dkg/index.js
```diff
@@ -43,6 +44,96 @@ function bigIntToBuffer(bigint) {
...
+  return path.replace(/^m\//, '')
+    .split('/')
+    .map((index) => parseInt(index, 10));
+}
+
+/** Calculate offset from a public key */
+function calculateOffset(pk, index, chaincode) {
+  const salt = i2osp(index, 4);
+  const buffer = concatBytes(pk, salt);
+  return createHash('sha512', bigIntToBufferBE(chaincode, 32)).update(buffer).digest();
```
johnoliverdriscoll, line 96: This should be `createHmac`

modules/bls-dkg/index.js
```diff
@@ -106,6 +197,10 @@ function mergePublicShares(shares) {
   return bufferToBigInt(sum.toCompressedHex())
 }
 
+function mergeChaincodes(chaincodes) {
+  return chaincodes.reduce((sum, chaincode) => sum + chaincode, 0n);
```
johnoliverdriscoll, line 201: The merged chaincode should be modulo `2^256` (`1n << 256n`).

johnoliverdriscoll (changes requested):

modules/bls-dkg/index.js
fperalta-simtlix: No, it is distributed through the secret shares to the participants.

modules/bls-dkg/index.js
```diff
@@ -43,6 +46,96 @@ function bigIntToBuffer(bigint) {
...
+
+/** Calculate offset from a public key */
+function calculateOffset(pk, index, chaincode) {
+  const salt = i2osp(index, 4);
+  const buffer = concatBytes(pk, salt);
+  return createHmac('sha512', bigIntToBufferBE(chaincode, 32)).update(buffer).digest();
+}
+
+/** Derive a public key */
+function publicDerive(pk, path, chaincode) {
```
johnoliverdriscoll, line 102: More of a nit, but the signature of `publicDerive` and `privateDerive` are a bit unconventional. Would be nicer if the key and the chain parameters were together (ie a "keychain"): `function publicDerive(pk, chaincode, path)`

modules/bls-dkg/test/bls-dkg.js
```diff
@@ -8,113 +8,139 @@ const {
...
+    });
     if (!await verify(backupWalletSig, message, commonPub)) {
-      throw new Error('Could not verify backup + wallet signature')
+      throw new Error('Could not verify backup + wallet signature');
+    }
+  });
+
+  it('verifies user-backup subkey sig with public derived', async function() {
+    const userPublicShareDKG = publicShare(userPolynomial);
+    const derivedUserPublicShareDKG = publicDerive(userPublicShareDKG, subKeyPath, userPolynomial[1]);
```
johnoliverdriscoll, line 128: Using `userPolynomial[1]` as the chaincode here is a bit misleading. Please allocate some new entropy to emulate a common chaincode.

modules/bls-dkg/test/bls-dkg.js
```diff
@@ -8,113 +8,139 @@ const {
...
+    const userBackupSig = mergeSignatures({
+      1: userSig,
+      2: backupSig,
+    });
+
+    if (!await verify(userBackupSig, message, pub)) {
+      throw new Error('Could not verify user + backup signature');
     }
-  })
+  });
```
johnoliverdriscoll, line 145: Please also add a test that verifies `mergePublicShares([publicDerive(pk_A, chaincode_Common, path), pk_B, pk_C]) == publicDerive(mergePublicShares([pk_A, pk_B, pk_C]), chaincode_Common, path)`

johnoliverdriscoll (changes requested):

modules/bls-dkg/index.js
fperalta-simtlix: You are right, thanks!

modules/bls-dkg/test/bls-dkg.js
fperalta-simtlix: Sure, forgot to change it. Thanks!

modules/bls-dkg/test/bls-dkg.js
fperalta-simtlix: This is interesting, I can't make it work. Doesn't match between them.

modules/bls-dkg/index.js
```diff
@@ -43,6 +46,96 @@ function bigIntToBuffer(bigint) {
...
+    return {
+      pk: bufferToBigInt(publicChild.toCompressedHex()),
+      chaincode: bufferToBigInt(zr),
+    };
+  }
+  return indices.reduce(deriveIndex, { pk, chaincode });
+}
+
+/** Derive a private key */
+function privateDerive(parentSk, chaincode, path) {
```
johnoliverdriscoll, line 122: In order for this to be useful in TSS, you need to accept a separate public key parameter and use that as the input to the index derivation. That way you can calculate the offset computed for the derivation of the common public key and apply it to a single party's contribution to the key. As far as I know, that is the only way it can work. See modules/account-lib/mpc/hdTree.ts for reference.

modules/bls-dkg/test/bls-dkg.js
johnoliverdriscoll: My mistake, this wouldn't work because the offset for each subkey is going to be different between the user key and the common key in the derivation path. Instead you need to rewrite the `verifies user-backup subkey sig with public derived` test;
1) The user derives a new secret share `derivedUserSk`.
2) Multiply `derivedUserSk` by the curve's base to get `derivedUserPublicShare`.
3) When you merge the derived user public share with the other public shares, you get a new `derivedCommonPub`.
4) `derivedCommonPub1` should be equal to `derivedCommonPub2 = publicDerive(commonPub, commonChaincode, path)`
5) Signature sholud verify against the newly derived common public key.

johnoliverdriscoll (changes requested):

modules/bls-dkg/test/bls-dkg.js
fperalta-simtlix: Makes sense, will do that UT.

modules/bls-dkg/index.js
fperalta-simtlix: Oh, got it, I'll do it in that way.

johnoliverdriscoll (approved):

PR #2237, opened 2022-04-28
STLX-15822_add_explainTransaction_implemented_in_account_lib_call

fmartinAltamirano:
Add explainTransaction implemented in account lib call in core.

STLX-15822


modules/bitgo/src/v2/coins/near.ts
```diff
@@ -155,10 +158,26 @@ export class Near extends BaseCoin {
...
-    throw new MethodNotImplementedError('Near recovery not implemented');
+  ): Promise<NearTransactionExplanation> {
+    const factory = accountLib.register(this.getChain(), accountLib.Near.TransactionBuilderFactory);
+    let rebuiltTransaction;
+
+    try {
+      let rawTxBase64 = params.txPrebuild.txHex;
+      if (HEX_REGEX.test(params.txPrebuild.txHex)) {
+        rawTxBase64 = Buffer.from(params.txPrebuild.txHex, 'hex').toString('base64');
+      }
```
ipelaez-simtlix, line 171: Maybe be can do something like
const txHex = params.txPrebuild.txHex;
const rawTxBase64 = HEX_REGEX.test(txHex) ? Buffer.from(txHex, 'hex').toString('base64') : txHex; 

I propose this because because Im kind of confused, params.txPrebuild.txHex is always a hex? or it could come as base64? if we always expect a hex, lets just the try catch do his job, delete the if.

modules/bitgo/test/v2/unit/coins/near.ts
```diff
@@ -125,4 +126,199 @@ describe('NEAR:', function () {
       isTransactionVerified.should.equal(true);
     });
   });
+
+  describe('Explain Transactions:', () => {
```
rliberal-simtlix, line 130: Get Invalid cases.

modules/bitgo/src/v2/coins/near.ts
```diff
@@ -155,10 +158,26 @@ export class Near extends BaseCoin {
...
+      const transactionBuilder = factory.from(rawTxBase64);
+      rebuiltTransaction = await transactionBuilder.build();
+    } catch {
+      throw new Error('Invalid transaction');
+    }
+
+    const explainedTransaction = (rebuiltTransaction as accountLib.BaseCoin.BaseTransaction).explainTransaction();
+
+    return explainedTransaction as NearTransactionExplanation;
```
ipelaez-simtlix, line 180: I think we can just add the type to the statement instead of use 'as'
`let rebuiltTransaction : accountLib.BaseCoin.BaseTransaction;`

and then 
    `const explainedTransaction: NearTransactionExplanation = rebuiltTransaction.explainTransaction();
return explainedTransaction;`

modules/bitgo/test/v2/unit/coins/near.ts
```diff
@@ -125,4 +126,199 @@ describe('NEAR:', function () {
...
+    const amount = '1000000';
+    const gas = '125000000000000';
+
+    it('should explain an unsigned transfer transaction', async function () {
+      const explainedTransaction = await basecoin.explainTransaction({
+        txPrebuild: {
+          txHex: rawTx.transfer.signed,
+        },
+      });
```
ipelaez-simtlix, line 139: What happen if we add
feeInfo: {
          fee: '5000',
        },

to the parameters? it is allowed by the function, can we add a test with that?

ipelaez-simtlix (changes requested):

modules/bitgo/src/v2/coins/near.ts
```diff
@@ -155,10 +158,26 @@ export class Near extends BaseCoin {
...
    * @param params
    */
-  explainTransaction(
+  async explainTransaction(
     params: ExplainTransactionOptions,
-  ): Promise<any> {
-    throw new MethodNotImplementedError('Near recovery not implemented');
+  ): Promise<NearTransactionExplanation> {
+    const factory = accountLib.register(this.getChain(), accountLib.Near.TransactionBuilderFactory);
+    let rebuiltTransaction;
```
rliberal-simtlix, line 165: reorder code here

modules/bitgo/src/v2/coins/near.ts
fmartinAltamirano: Well, I'm going to try this refactoring

modules/bitgo/src/v2/coins/near.ts
fmartinAltamirano: If I declare `rebuiltTransaction` as a constant inside the try block, I cannot use it later in line 175 because it is outside the scope of the try, that is why it was defined outside.

rliberal-simtlix (dismissed):

ewangbitgo (dismissed):

ipelaez-simtlix (approved):
LGTM

modules/bitgo/src/v2/coins/near.ts
```diff
@@ -164,10 +167,23 @@ export class Near extends BaseCoin {
...
+
+    try {
+      const transactionBuilder = factory.from(rawTxBase64);
+      rebuiltTransaction = await transactionBuilder.build();
+    } catch {
+      throw new Error('Invalid transaction');
+    }
+
+    const explainedTransaction: NearTransactionExplanation = rebuiltTransaction.explainTransaction();
+    return explainedTransaction;
```
CapnMigraine, line 186: nit: redundant variable

CapnMigraine (approved):

silalvado (approved):

alebusse (approved):

CapnMigraine (approved):

ipelaez-simtlix (approved):

CapnMigraine (approved):

rliberal-simtlix (approved):

PR #2240, opened 2022-04-29
chore(examples): add token transaction example

shiksb:
Add an example of how to build, sign, and send a token transaction.

### DIFF
--- a/modules/bitgo/example/ts/send-token-transaction.ts
+++ b/modules/bitgo/example/ts/send-token-transaction.ts
@@ -1,0 +1,57 @@
+/**
+ * Build, sign, and submit a token transaction to the BitGo network.
+ *
+ * Copyright 2022, BitGo, Inc.  All Rights Reserved.
+ */
+
+import { BitGo } from 'bitgo';
+
+// authentication fixtures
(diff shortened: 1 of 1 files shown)


jicli (approved):

modules/bitgo/example/ts/send-token-transaction.ts
```diff
@@ -0,0 +1,57 @@
...
+    recipients: [
+      {
+        amount: baseUnitAmount,
+        address: recipientAddress,
+      },
+    ],
+    walletPassphrase,
+  });
+
+  const rawHalfSignedTransaction: string = (halfSignedTransaction as any).halfSigned.txHex;
```
mmcshinsky-bitgo, line 45: Lgtm :) Can we update the `any` to the appropriate type?

mmcshinsky-bitgo (changes requested):

modules/bitgo/example/ts/send-token-transaction.ts
mmcshinsky-bitgo: Also needs JavaScript equivalent example.

PR #2241, opened 2022-04-29
chore: refactor tss signing

ewangbitgo:


modules/bitgo/src/tss.ts
```diff
@@ -1,5 +1,17 @@
-import Eddsa, { KeyShare, UShare, YShare } from '@bitgo/account-lib/dist/src/mpc/tss';
+import Eddsa, {
```
joey-bai, line 1: all the functions that make API calls won't be re-usable by HSM

joey-bai (approved):
LGTM

PR #2243, opened 2022-04-29
chore: refactor TSS lib

joey-bai:
- deriveUnhardened returns full keychain; clients will need to take the
first 32 bytes if they wish to use it as a public key
- initialize functions will also return a newly instantiated object

Ticket: BG-00000

### DIFF
--- a/modules/account-lib/test/unit/mpc/tss.ts
+++ b/modules/account-lib/test/unit/mpc/tss.ts
@@ -9,18 +9,20 @@ import * as sol from '@solana/web3.js';
 import { Dot, Sol } from '../../../src';
 
 import Eddsa from '../../../src/mpc/tss';
-import { Ed25519BIP32 } from '../../../src/mpc/hdTree';
+import HDTree, { Ed25519BIP32 } from '../../../src/mpc/hdTree';
 
 import { bigIntFromBufferLE, bigIntToBufferLE, bigIntFromBufferBE, bigIntToBufferBE } from '../../../src/mpc/util';
 
 describe('TSS EDDSA key generation and signing', function () {
+  let MPC: Eddsa;
+  let hdTree: HDTree;
+
   before('initialize modules', async function () {
-    await Eddsa.initialize();
-    await Ed25519BIP32.initialize();
+    hdTree = await Ed25519BIP32.initialize();
+    MPC = await Eddsa.initialize(hdTree);
   });
 
   it('should generate keys and sign message', function () {
-    const MPC = new Eddsa();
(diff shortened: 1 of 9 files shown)


modules/account-lib/test/unit/mpc/tss.ts
```diff
@@ -172,21 +168,22 @@ describe('TSS EDDSA key generation and signing', function () {
       const path = `m/0/0/${index}`;
       const derive1 = MPC.deriveUnhardened(commonKeychain, path);
       const derive2 = MPC.deriveUnhardened(commonKeychain, path);
+      const derivedPk = Buffer.from(derive1, 'hex').slice(0, 32).toString('hex');
```
joey-bai, line 171: @pranavjain97 when updating SDK in platform, we'll also need to extract the public key like this.

pranavjain97: Yeah makes sense. I think extracting the public key on platform is only done here-
https://github.com/BitGo/bitgo-microservices/blob/824d140eee92168e8c47eeab12f6800d2bb9af18/packages/wallet-platform/app/controllers/api/v2/coins/features/thresholdSignatureScheme.ts#L83

modules/account-lib/test/unit/mpc/tss.ts
```diff
@@ -139,27 +136,26 @@ describe('TSS EDDSA key generation and signing', function () {
       const derive1 = MPC.deriveUnhardened(commonKeychain, path);
       const subkey = MPC.keyDerive(A.uShare, [B.yShares[1], C.yShares[1]], path);
       const derive2 = MPC.deriveUnhardened(commonKeychain, path);
+      const derivedPk = Buffer.from(derive1, 'hex').slice(0, 32).toString('hex');
```
pranavjain97, line 139: Dumb question, but why do we need to convert to buffer and then back to hex? 
Instead of something like derivedPk = derive2.slice(0, 64) since its already in hex

joey-bai: mistake on my side - i thought i observed some flakiness in my test when using `string.slice(0, 64)`
updated one usage to use `string.slice(0, 64)` and left others as Buffer. (i guess to demonstrate that they're equivalent).

pranavjain97: Alright, thanks for checking. I guess on the platform side i'll just use `string.slice(0, 64)` then

pranavjain97 (approved):

PR #2245, opened 2022-05-02
chore: refactor tss key creation

joey-bai:
externalizes a few helper functions for reusability in OVC
- encrypt, sign, and format y shares to be sent to BitGo
- combine tss keys and format signing material

Ticket: BG-45374


ewangbitgo (approved):
lgtm. I did the same to signing flow as export a few function in tss.ts

PR #2246, opened 2022-05-02
chore: use keyType instead of type for tss + bls keys

joey-bai:
type was originally used to differentiate different types of backup keys
(i.e. using an external key provider or holding on to your own).

this was overloaded with the introduction of TSS and BLS.

this change refactors to use keyType to determine multisig type (tss,
bls, or independent/onchain).

Ticket: BG-43890

### DIFF
--- a/modules/bitgo/src/v2/keychains.ts
+++ b/modules/bitgo/src/v2/keychains.ts
@@ -9,6 +9,9 @@ import { TssUtils } from './internal/tssUtils';
 import { BlsUtils } from './internal/blsUtils';
 
 const { validateParams } = common;
+
+export type KeyType = 'tss' | 'independent' | 'blsdkg';
+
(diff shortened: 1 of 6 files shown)


modules/bitgo/src/v2/keychains.ts
```diff
@@ -9,6 +9,9 @@ import { TssUtils } from './internal/tssUtils';
 import { BlsUtils } from './internal/blsUtils';
 
 const { validateParams } = common;
+
+export type KeyType = 'tss' | 'independent' | 'blsdkg';
```
joey-bai, line 13: Note to reviewers: this comes from https://github.com/BitGo/bitgo-microservices/blob/develop/packages/api-docs/shared/KeyType.yaml

fperalta-simtlix (approved):

PR #2252, opened 2022-05-04
feat: adding TRS balance property to wallet

MaxVTurcan:
This property is required for reporting features

CLOSES Ticket: BG-45329

### DIFF
--- a/modules/bitgo/src/v2/wallet.ts
+++ b/modules/bitgo/src/v2/wallet.ts
@@ -445,9 +445,11 @@ export interface WalletData {
   balance: number;
   confirmedBalance: number;
   spendableBalance: number;
+  trsBalance: number;
   balanceString: string;
   confirmedBalanceString: string;
   spendableBalanceString: string;
+  trsBalanceString: string;
   coin: string;
   label: string;
   keys: string[];
@@ -661,6 +663,15 @@ export class Wallet {
    */
   spendableBalanceString(): string {
     return this._wallet.spendableBalanceString;
+  }
+
+  /**
+   * Get a string representation of the TRS balance of this wallet
+   *
+   * This is useful when balances have the potential to overflow standard javascript numbers
+   */
+  trsBalanceString(): string {
+    return this._wallet.trsBalanceString;
   }
 
   /**


modules/bitgo/src/v2/wallet.ts
```diff
@@ -663,6 +665,15 @@ export class Wallet {
     return this._wallet.spendableBalanceString;
   }
 
+  /**
+   * Get a string representation of the TRS balance of this wallet
+   *
+   * This is useful when balances have the potential to overflow standard javascript numbers
+   */
+  trsBalanceString(): string {
```
mmcshinsky-bitgo, line 673: I'm not sure how this solves the problem in the ticket. Probably missing some context. Can you expand on how these new fields are utilized with an example?

mmcshinsky-bitgo (changes requested):

modules/bitgo/src/v2/wallet.ts
MaxVTurcan: Sure, these wallet objects are passed back to bga when executing the wallet command and then used to display the result:
https://github.com/BitGo/bitgo-admin/blob/master/src/handlers/wallet.js#L151

I wanted to add the property to the object instead of doing a direct property reference on a undeclared field. The TRS balance field will go right below the other 3 balances that get printed.

This PR will work in tandem with 2 other PRs to achieve this. Here is the PR that will populate that new field:
https://github.com/BitGo/bitgo-microservices/pull/18262/files#diff-008ef6124f852724918acc1cb7c461bd3409c7afae62c2740b9952e00f81fd94R417

mmcshinsky-bitgo: To make sure, these are fields that only for bga?

tylerlevine (changes requested):
This should not be part of the public facing API - if we want to get a balance from TRS, it should be optional (and opted into via a query param or request body option) and should replace the `balanceString`.

We can't expect customers to know what `trs` is, and we should not be exposing internal service names through the public API in this way. I'll leave more comments on the wallet platform change.

MaxVTurcan:
Works in tandem with the following changes:
https://github.com/BitGo/bitgo-microservices/pull/19046
https://github.com/BitGo/bitgo-admin/pull/639

MaxVTurcan:
> This should not be part of the public facing API - if we want to get a balance from TRS, it should be optional (and opted into via a query param or request body option) and should replace the `balanceString`.
> 
> We can't expect customers to know what `trs` is, and we should not be exposing internal service names through the public API in this way. I'll leave more comments on the wallet platform change.

Got it Tyler, what you are saying makes sense. I will work on making updates and see what I come up with. I need to clarify if it is ok to replace balanceString completely in the bga output.

MaxVTurcan:
No longer needed

Issue #14, opened 2015-11-12
Cannot get example working


JustinDrake:
Followed the tutorial step-by-step (in particular, I did `npm install bitgo`). Below is the error I get:

```
node BitGoJS/example/auth.js
module.js:339
    throw err;
    ^

Error: Cannot find module 'superagent'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/justin/Desktop/Duo-Money/backend/BitGoJS/src/bitgo.js:7:18)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Module.require (module.js:366:17)
```

AlanVerbner:
Hello @JustinDrake, we have installed bitgo without any problem. 

I'm not sure if this is going to help you but it might give you an idea:
- We are using Node v4.2.3 and Npm 2.14.7. What versions are you using?
- If you try `npm view bitgo` what do you get? for example I can see `superagent: '^1.4.0'` in dependencies
- Did you try to clear npm cache?

bencxr:
Closing due to inactivity

PR #2253, opened 2022-05-04
chore: github action master to rel/rc

mmcshinsky-bitgo:
Create automatic pull request from `master` to `rel/rc` whenever `master` is updated.

BG-000000

### DIFF
--- a/.github/workflows/master-to-relrc.yml
+++ b/.github/workflows/master-to-relrc.yml
@@ -1,0 +1,25 @@
+name: Create rel/rc pull request
+on:
+  push:
+    branches:
+      - master
+jobs:
+  relrc-promotion:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: rel/rc
+      - name: Merge latest master to rel/rc
+        run: |
+          git fetch origin master:master
+          git reset --hard master
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: format('chore', ':', 'merge latest to rel/rc')
+          title: Merge latest to rel/rc
+          body: |
+            chore: merge latest master to rel/rc
+            BG-000000
+          branch: master-to-rel/rc


ducnguyen-bitgo (changes requested):

.github/workflows/master-to-relrc.yml
```diff
@@ -0,0 +1,25 @@
...
+          git reset --hard master
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: format('chore', ':', 'merge latest to rel/rc')
+          title: Merge latest to rel/rc
+          body: |
+            chore: merge latest master to rel/rc
+            BG-000000
+          branch: master-to-rel/rc
```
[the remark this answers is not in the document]
mmcshinsky-bitgo: Depends on preference from the team. If there are merge conflicts from master to rel/rc with master being the source of truth, we can make the merge conflicts visible and resolve them via a `master-to-rel/rc` branch since master doesn't need to be updated. Also allows us to go in and fix `rc` tags if needed since the merge doesn't know that it isn't supposed to override `rc` tags in package.json files.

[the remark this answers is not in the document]
ducnguyen-bitgo: That makes sense and looks good to me!

ducnguyen-bitgo (approved):
I tested the action in my own private repo and it seemed to work correctly. Everything looks good to me!

.github/workflows/master-to-relrc.yml
```diff
@@ -0,0 +1,25 @@
...
+          git fetch origin master:master
+          git reset --hard master
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: format('chore', ':', 'merge latest to rel/rc')
+          title: Merge latest to rel/rc
+          body: |
+            chore: merge latest master to rel/rc
+            BG-000000
```
tylerlevine, line 24:
```suggestion
```

.github/workflows/master-to-relrc.yml
```diff
@@ -0,0 +1,25 @@
...
+          ref: rel/rc
+      - name: Merge latest master to rel/rc
+        run: |
+          git fetch origin master:master
+          git reset --hard master
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: format('chore', ':', 'merge latest to rel/rc')
+          title: Merge latest to rel/rc
```
tylerlevine, line 21:
```suggestion
          title: Merge master to rel/rc
```

.github/workflows/master-to-relrc.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          ref: rel/rc
+      - name: Merge latest master to rel/rc
+        run: |
+          git fetch origin master:master
+          git reset --hard master
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: format('chore', ':', 'merge latest to rel/rc')
```
tylerlevine, line 20: How come this needs `format()`? Also let's not confuse `master` with `latest`, which might refer to the `rel/latest` branch, which is something else.
```suggestion
          commit-message: 'chore: merge master to rel/rc'
```

.github/workflows/master-to-relrc.yml
```diff
@@ -0,0 +1,25 @@
...
+  push:
+    branches:
+      - master
+jobs:
+  relrc-promotion:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: rel/rc
```
tylerlevine, line 12: I'm confused why we check out `rel/rc`, then fetch and hard reset to `master` - why not just check out master directly?

tylerlevine (changes requested):
I like the idea, but please create a jira for this instead of using BG-00000 so we can track it as part of the sprint. That fake ticket number shouldn't really ever be used and if it is, it should be for quick fixes and not net new features.

Also, is this going to create signed commits? Note that the repo requires all commits to be signed (but a signature from Github is acceptable).

Other than that, just some minor improvement suggestions.

.github/workflows/master-to-relrc.yml
mmcshinsky-bitgo: Can do!

.github/workflows/master-to-relrc.yml
mmcshinsky-bitgo: format is needed because the `:` breaks the yml file.

.github/workflows/master-to-relrc.yml
mmcshinsky-bitgo: It came from this example here: https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md#keep-a-branch-up-to-date-with-another

.github/workflows/master-to-relrc.yml
mmcshinsky-bitgo: I wonder if we can set something up like this? https://github.com/marketplace/actions/jira-create-issue

.github/workflows/master-to-relrc.yml
mmcshinsky-bitgo: Done

.github/workflows/master-to-relrc.yml
ducnguyen-bitgo: image
We could either checkout `rel/rc` or specify it as `base` in `Create Pull Request` (CPR) according to the CPR docs.

.github/workflows/master-to-relrc.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Merge master to rel/rc
+        run: |
+          git fetch origin master:master
+          git reset --hard master
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: format('chore', ':', 'merge master to rel/rc')
```
ducnguyen-bitgo, line 20: Since there's no changes made on line #14 (run), `commit-message` won't take any effect in this case. That's why they don't include `commit-message` in the example of CPR.

.github/workflows/master-to-relrc.yml
tylerlevine: even inside quotes? TIL

ducnguyen-bitgo (dismissed):
Just left a couple of comments on how the create pull request action plugin should work. Otherwise, everything looks good to me! Btw, I haven't looked at why one of the checks failed.

.github/workflows/master-to-relrc.yml
ducnguyen-bitgo: Just a couple of screenshots on how it looks like in my private test repo. The `commit-message` doesn't have any effect even though we specify it as `format('chore', ':', 'merge master to rel/rc')`.

image
image

.github/workflows/master-to-relrc.yml
mmcshinsky-bitgo: Removed per comments below.

.github/workflows/master-to-relrc.yml
mmcshinsky-bitgo: Thanks, going to remove this line.

.github/workflows/master-to-relrc.yml
mmcshinsky-bitgo: I've played around with a few different configurations on a private repo and the current one in the PR is the only one that properly triggers the pull request for me.

tylerlevine (approved):

PR #2254, opened 2022-05-04
fix(bitgo): attempt to sign using fallback derivation for v1 wallet

tylerlevine:
We have seen an issue for at least one production v1 wallet where an
input is unable to be signed for due to using an incorrect derivation
path for the user keychain. We believe this wallet has a non-standard
derivation scheme may have been broken in a change to the v1
transaction builder.

In the event of a signing failure, try to use a fallback path derivation
scheme which we believe will work for the wallet in question. If the
fallback path derivation fails, log the error which caused the fallback
derivation signing to fail, but throw the original error to maintain
previous error behavior.

Ticket: BG-46497


OttoAllmendinger (approved):

sangaman (approved):
LGTM, interested to see if this resolves the underlying issue.

PR #2256, opened 2022-05-05
feat(account-lib): add support for secp256k1 curve TSS

sachushajiabraham:
this change adds support for secp256k1 curve for MPC / TSS and also changes verify function in ed25519 to return boolean

TICKET:BG-47402

### DIFF
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -43,6 +43,7 @@
     "@bitgo/bls-dkg": "^1.0.2",
     "@bitgo/statics": "^6.17.0",
     "@bitgo/sdk-core": "^1.0.1",
+    "@noble/secp256k1": "git+https://github.com/brandonblack/noble-secp256k1.git#856129c06436fb696bab7d827edad5effdffaee2",
     "@celo/contractkit": "^1.2.4",
     "@ethereumjs/common": "^2.4.0",
     "@ethereumjs/tx": "^3.3.0",
--- a/modules/account-lib/src/mpc/curves/ed25519.ts
+++ b/modules/account-lib/src/mpc/curves/ed25519.ts
@@ -61,7 +61,16 @@ export class Ed25519Curve implements BaseCurve {
     );
   }
 
-  verify(y: bigint, signedMessage: Buffer): Buffer {
-    return Buffer.from(sodium.crypto_sign_open(signedMessage, bigIntToBufferLE(y, 32)));
+  verify(message: Buffer, signature: Buffer, publicKey: bigint): boolean {
+    const signedMessage = Buffer.concat([signature, message]);
+    try {
+      // Returns the message which was signed if the signature is valid
+      const result = Buffer.from(sodium.crypto_sign_open(signedMessage, bigIntToBufferLE(publicKey)));
+      const isValid = !!!Buffer.compare(message, result); // Buffer compare gives 0 if equal and a non zero integer if not
+      return isValid;
+    } catch (error) {
+      // Invalid signature causes an exception
+      return false;
+    }
   }
 }
--- a/modules/account-lib/src/mpc/curves/secp256k1.ts
+++ b/modules/account-lib/src/mpc/curves/secp256k1.ts
@@ -1,0 +1,55 @@
+import { bigIntFromU8ABE, bigIntToBufferBE } from '../util';
+import BaseCurve from './baseCurve';
+import * as secp from '@noble/secp256k1';
+
+const order = secp.CURVE.n;
+
+export class Secp256k1Curve implements BaseCurve {
+  scalarRandom(): bigint {
+    return bigIntFromU8ABE(secp.utils.randomPrivateKey());
+  }
+
+  scalarAdd(x: bigint, y: bigint): bigint {
+    return bigIntFromU8ABE(secp.utils.privateAdd(x, bigIntToBufferBE(y)));
+  }
+
+  scalarSub(x: bigint, y: bigint): bigint {
+    const negatedY = secp.utils.privateNegate(y);
+    return bigIntFromU8ABE(secp.utils.privateAdd(x, negatedY));
+  }
+
+  scalarMult(x: bigint, y: bigint): bigint {
+    return secp.utils.mod(x * y, order);
+  }
+
+  scalarReduce(s: bigint): bigint {
+    return secp.utils.mod(s, order);
+  }
+
+  scalarNegate(s: bigint): bigint {
+    return bigIntFromU8ABE(secp.utils.privateNegate(s));
+  }
+
+  scalarInvert(s: bigint): bigint {
+    return secp.utils.invert(s, order);
+  }
+
+  pointAdd(a: bigint, b: bigint): bigint {
+    const pointA = secp.Point.fromHex(bigIntToBufferBE(a));
+    const pointB = secp.Point.fromHex(bigIntToBufferBE(b));
+    return bigIntFromU8ABE(pointA.add(pointB).toRawBytes(true));
+  }
+
+  basePointMult(n: bigint): bigint {
+    const point = bigIntToBufferBE(n);
+    return bigIntFromU8ABE(secp.getPublicKey(point, true));
+  }
+
+  verify(message: Buffer, signature: Buffer, publicKey: bigint): boolean {
+    return secp.verify(signature, message, bigIntToBufferBE(publicKey));
+  }
+
+  order(): bigint {
+    return order;
+  }
+}
--- a/modules/account-lib/src/mpc/tss.ts
+++ b/modules/account-lib/src/mpc/tss.ts
@@ -386,13 +386,9 @@ export default class Eddsa {
     return result;
   }
 
-  verify(message: Buffer, signature: Signature): Buffer {
+  verify(message: Buffer, signature: Signature): boolean {
     const publicKey = bigIntFromBufferLE(Buffer.from(signature.y, 'hex'));
-    const signedMessage = Buffer.concat([
-      Buffer.from(signature.R, 'hex'),
-      Buffer.from(signature.sigma, 'hex'),
-      message,
-    ]);
-    return Eddsa.curve.verify(publicKey, signedMessage);
+    const signedMessage = Buffer.concat([Buffer.from(signature.R, 'hex'), Buffer.from(signature.sigma, 'hex')]);
+    return Eddsa.curve.verify(message, signedMessage, publicKey);
   }
 }
--- a/modules/account-lib/test/unit/mpc/curves/secp256k1.ts
+++ b/modules/account-lib/test/unit/mpc/curves/secp256k1.ts
@@ -1,0 +1,108 @@
+/**
+ * @prettier
+ */
+import { Secp256k1Curve } from '../../../../src/mpc/curves';
+import { bigIntToBufferBE } from '../../../../src/mpc/util';
+
+describe('secp256k1 curve implementation', function () {
+  const sec256k1 = new Secp256k1Curve();
+  const testValue = BigInt(12000);
+
+  it('should properly generate a random number in the specified range', function () {
+    const start = BigInt(1);
+    const stop = sec256k1.order();
+    const test = sec256k1.scalarRandom();
+    const isOkay = test >= start && test <= stop;
+    isOkay.should.equal(true);
+  });
+
+  it('should correctly perform scalar add', function () {
+    const valueOne = sec256k1.scalarAdd(sec256k1.order() - BigInt(1), testValue);
+    valueOne.should.equal(testValue - BigInt(1));
+    const valueTwo = sec256k1.scalarAdd(BigInt(5), testValue);
+    valueTwo.should.equal(BigInt(5) + testValue);
+  });
+
+  it('should correctly perform scalar sub', function () {
+    const valueOne = sec256k1.scalarSub(BigInt(2) * testValue, testValue);
+    valueOne.should.equal(testValue);
+    const valueTwo = sec256k1.scalarSub(BigInt(5), testValue);
+    valueTwo.should.equal(sec256k1.order() + (BigInt(5) - testValue));
+  });
+
+  it('should correctly perform scalar multiplication', function () {
+    const valueOne = sec256k1.scalarMult(sec256k1.order(), testValue);
+    valueOne.should.equal(BigInt(0));
+    const valueTwo = sec256k1.scalarMult(BigInt(5), testValue);
+    valueTwo.should.equal(BigInt(5) * testValue);
+  });
+
+  it('should correctly perform scalar reduce', function () {
+    const valueOne = sec256k1.scalarReduce(sec256k1.order() + testValue);
+    valueOne.should.equal(testValue);
+    const valueTwo = sec256k1.scalarReduce(sec256k1.order() - testValue);
+    valueTwo.should.equal(sec256k1.order() - testValue);
+  });
+
+  it('should correctly perform scalar negate', function () {
+    const valueOne = sec256k1.scalarNegate(testValue);
+    valueOne.should.equal(sec256k1.order() - testValue);
+    const valueTwo = sec256k1.scalarNegate(sec256k1.order() - testValue);
+    valueTwo.should.equal(testValue);
+  });
+
+  it('should correctly perform inverts', function () {
+    // Had to convert bigint to string as comparison was not properly working in mocha
+    const testOne = sec256k1.scalarInvert(BigInt(3)).toString();
+    testOne.should.equal('77194726158210796949047323339125271901891709519383269588403442094345440996225');
+
+    const testTwo = sec256k1.scalarInvert(sec256k1.order() + BigInt(1)).toString();
+    testTwo.should.equal('1');
+
+    const testThree = sec256k1
+      .scalarInvert(BigInt('57657865876576467547584635432343132435146576543543134312435443514313234131324551034534'))
+      .toString();
+    testThree.should.equal('45199239020920791752653273479365580486664737121914751844510223387235717825519');
+
+    const testFour = sec256k1.scalarInvert(BigInt('-23232424342224')).toString();
+    testFour.should.equal('109007816100548644677753129517258866095932985532097924870123596204155595187610');
+
+    const testFive = sec256k1.scalarInvert(testValue);
+    testFive.toString().should.equal('16664411509403755791375590925833668071820872792496863322396593062116822075060');
+    // Checking invert(invert(x)) === x
+    sec256k1.scalarInvert(testFive).should.equal(testValue);
+  });
+
+  it('should correctly perform point add', function () {
+    const A = BigInt('0x0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798');
+    const B = BigInt('0x0252972572d465d016d4c501887b8df303eee3ed602c056b1eb09260dfa0da0ab2');
+    const publicKey = '0x0' + sec256k1.pointAdd(A, B).toString(16);
+    publicKey.should.equal('0x03e2d3ec45106a89f82e7d46400faefc191ed72344caa781bfa97677a999d0828e');
+  });
+
+  it('should correctly perform base point multiplication', function () {
+    const privKey = BigInt('0x79FE45D61339181238E49424E905446A35497A8ADEA8B7D5241A1E7F2C95A04D');
+    const publicKey = '0x0' + sec256k1.basePointMult(privKey).toString(16);
+    publicKey.should.equal('0x032a574ea59cae80b09d6ba415746e9b031abfbe83f149b43b37be035b87164872');
+  });
+
+  it('should correctly verify signature', function () {
+    const publicKey = BigInt('0x33ccfc7998b1ce2fad8b3f287d873f07ca8b3037897c2136c11100c02213c028e');
+    const signature = bigIntToBufferBE(
+      BigInt(
+        '0x6a43d5da66ec4b4c6eb307d18791b9744cc59dd2402ad395bb9efb513898c07f767139118b76e387564bf572b2cc481dbf3068b7ac27620f83a2819d735b7e5f',
+      ),
+    );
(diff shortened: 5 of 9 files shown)


modules/account-lib/src/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  scalarInvert(s: bigint): bigint {
+    return BigInt(0); // TODO
+  }
+
+  scalarAdd(x: bigint, y: bigint): bigint {
+    return (x + y) % order;
+  }
+
+  scalarSub(x: bigint, y: bigint): bigint {
```
sachushajiabraham, line 28: modulus operation for negative numbers in python strangely behaves differently from what Js does, So need to recheck this.

```
-5 % 15 ==> 10 // python
-5 % 15 ==>-5 // js
```

johnoliverdriscoll: The result of a mod operation should always produce a positive number. You don't want your private key to be a negative number. If the result `(x - y)` is less than zero, return `(x - y) + order`. Please also ensure that `scalarNegate` only returns positive numbers.

modules/account-lib/src/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,51 @@
...
+  scalarMult(x: bigint, y: bigint): bigint {
+    return (x * y) % order;
+  }
+
+  basePointMult(n: bigint): bigint {
+    return BigInt(0); // TODO
+  }
+
+  pointAdd(p: bigint, q: bigint): bigint {
+    return p + q;
```
johnoliverdriscoll, line 41: `// TODO`

johnoliverdriscoll (changes requested):

tylerlevine (commented):
How come we're implementing these primitive curve operations in account-lib instead of using a dependency for these that's already in the utxo-lib dependency tree (eg, noble-secp256k1 or a pre-wasm version of tiny-secp256k1)?

sachushajiabraham:
Hi @tylerlevine , the implementations are basically for the scalar operations over that finite field  , for point operations we will be making use of noble-secp256k1 or tiny-secp256k1 or any suitable library. Still finalising on it

modules/account-lib/src/mpc/tss.ts
```diff
@@ -387,12 +387,8 @@ export default class Eddsa {
...
 
   verify(message: Buffer, signature: Signature): Buffer {
-    const publicKey = bigIntFromBufferLE(Buffer.from(signature.y, 'hex'));
-    const signedMessage = Buffer.concat([
-      Buffer.from(signature.R, 'hex'),
-      Buffer.from(signature.sigma, 'hex'),
-      message,
-    ]);
-    return Eddsa.curve.verify(publicKey, signedMessage);
+    const publicKey = Buffer.from(signature.y, 'hex');
```
barathcj, line 390: This does not have bigIntFromBufferLE

modules/account-lib/src/mpc/tss.ts
```diff
@@ -387,12 +387,8 @@ export default class Eddsa {
...
+    const signedMessage = Buffer.concat([Buffer.from(signature.R, 'hex'), Buffer.from(signature.sigma, 'hex')]);
+    return Eddsa.curve.verify(message, signedMessage, publicKey);
```
barathcj, line 392: This seems specific to eddsa curve

modules/account-lib/src/utils/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+const _1n = BigInt(1);
+/**
+ * Modulo operator
+ * @description In javascript % gives you the reminder. This functions
+ * computes actual modulus of two numbers
+ * @param {bigint} num
+ * @param {bigint} mod
+ * @returns {bigint}
+ */
+export function modulo(num: bigint, mod: bigint): bigint {
```
barathcj, line 13: I would just reject unexpected negative values here

modules/account-lib/src/mpc/tss.ts
sachushajiabraham: Will be refactoring TSS file in another PR to address this

modules/account-lib/src/utils/helpers.ts
sachushajiabraham: 👍🏻

barathcj (dismissed):

johnoliverdriscoll (approved):

tylerlevine (changes requested):
How come we're reimplementing these primitive curve operations instead of using something like `tiny-secp256k1` or another already-used library instead?

tylerlevine:
> Hi @tylerlevine , the implementations are for the basic scalar operations over that finite field , for point operations we will be making use of noble-secp256k1 or tiny-secp256k1 or any suitable library. Still finalising on it
cc: @johnoliverdriscoll

Is this still in progress?

sachushajiabraham:
> > Hi @tylerlevine , the implementations are for the basic scalar operations over that finite field , for point operations we will be making use of noble-secp256k1 or tiny-secp256k1 or any suitable library. Still finalising on it
> > cc: @johnoliverdriscoll
> 
> Is this still in progress?

Its done, we have used the `secp256k1` dependency for point operations, I tried using `tiny-secp256k1` first, but was giving some wasm related errors.

modules/account-lib/src/utils/helpers.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Secure random number generator
+ * generates number in the given range
+ * @param {bigint} start
+ * @param {bigint} end
+ * @param {number} size
+ * @returns {bigint}
+ */
+export function randomNumberGenerator(start: bigint, end: bigint, size = 32): bigint {
+  const value = modulo(BigInt('0x' + randomBytes(size).toString('hex')), end);
```
brandonblack, line 56: Modulo generates random numbers with bias. This is not cryptographically sound.

modules/account-lib/src/utils/helpers.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @param {bigint} mod
+ * @returns {bigint}
+ */
+export function modulo(num: bigint, mod: bigint): bigint {
+  const result = num % mod;
+  // modulo operation should always return positive values for curve operations
+  return result >= _0n ? result : mod + result;
+}
+
+/**
```
brandonblack, line 18: We already pull noble-secp as a dependency. In noble-secp master, I exposed this function. I have a [branch](https://github.com/brandonblack/noble-secp256k1/tree/bitgo) that includes build products in git so you can depend on a git version until Paul cuts a new release.

brandonblack (changes requested):
I'd suggest pulling noble-secp from here https://github.com/brandonblack/noble-secp256k1/tree/bitgo and using that in preference to `secp256k1` . `secp256k1` uses elliptic under the hood for browser operation, where noble-secp256k1 uses the same code path in both browser and server for safer testing.

modules/account-lib/src/utils/helpers.ts
sachushajiabraham: I have updated it

modules/account-lib/src/utils/helpers.ts
sachushajiabraham: Updated to use the random number generator from noble which addresses this issue

sachushajiabraham:
I tried to replace the operations with ones which are exposed via node. The ones which we were not available, I left the implementation as is. For point addition I used publickeyCombine from secp256k1 as it was not available in noble

modules/account-lib/package.json
```diff
@@ -45,6 +45,7 @@
     "@ethereumjs/common": "^2.4.0",
     "@ethereumjs/tx": "^3.3.0",
     "@hashgraph/sdk": "~2.3.0",
+    "@noble/secp256k1": "git+https://github.com/brandonblack/noble-secp256k1.git#bitgo",
```
sachushajiabraham, line 48: Should be have this hosted under bitgo org or probably this would be temporary until Paul does the actual release

modules/account-lib/src/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  scalarInvert(s: bigint): bigint {
+    return bigIntFromBufferBE(Buffer.from(secp.utils.privateInvert(bigIntToBufferBE(modulo(s), 32))));
+  }
+
+  pointAdd(a: bigint, b: bigint): bigint {
+    const pointA = bigIntToBufferBE(a);
+    const pointB = bigIntToBufferBE(b);
+    return bigIntFromBufferBE(Buffer.from(publicKeyCombine([pointA, pointB], true)));
```
brandonblack, line 41: Let's use 1 crypto library consistently:

```suggestion
    const pointA = secp.Point.fromHex(bigIntToBufferBE(a));
    const pointB = secp.Point.fromHex(bigIntToBufferBE(b));
    return bigIntFromBufferBE(Buffer.from(pointA.add(pointB).toRawBytes(true)));
```

modules/account-lib/src/utils/helpers.ts
```diff
@@ -0,0 +1,12 @@
+import * as secp from '@noble/secp256k1';
```
brandonblack, line 1: With the changes suggested in secp256k1, this helper is no longer needed.

modules/account-lib/test/unit/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    const testFive = sec256k1.scalarInvert(sec256k1.order() + testValue).toString();
+    testFive.should.equal('16664411509403755791375590925833668071820872792496863322396593062116822075060');
+  });
+
+  it('should correctly perform point add', function () {
+    const A = BigInt('0x0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798');
+    const B = BigInt('0x0252972572d465d016d4c501887b8df303eee3ed602c056b1eb09260dfa0da0ab2');
+    const publicKey = sec256k1.pointAdd(A, B).toString();
+    publicKey.should.equal('449973406896538015988913407864183977074571439549869550234638933112191976505998');
```
brandonblack, line 78: Let's specify this result as hex, since it's really not a number, but a compact byte-encoding of an x-y point.

modules/account-lib/test/unit/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  it('should correctly perform base point multiplication', function () {
+    const privKey = BigInt('0x79FE45D61339181238E49424E905446A35497A8ADEA8B7D5241A1E7F2C95A04D');
+    const publicKey = sec256k1.basePointMult(privKey).toString(16);
+    publicKey.should.equal('32a574ea59cae80b09d6ba415746e9b031abfbe83f149b43b37be035b87164872');
```
brandonblack, line 84: Hmm, this is the problem with using numbers to represent byte-encodings. We should pad this encoding with a leading 0 so that it represents a number of bytes, not nibbles.

modules/account-lib/test/unit/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,106 @@
...
+    let verify = sec256k1.verify(message, signature, publicKey);
+    verify.should.equal(true);
+    const messageWrong = bigIntToBufferBE(BigInt('0xd3b91b70417e766b738288cb1c5f6ef13a607011134544d4bce2ebf6ed5944ef'));
+    verify = sec256k1.verify(messageWrong, signature, publicKey);
+    verify.should.equal(false);
+  });
+
+  it('should have the correct curve order', function () {
+    const curveOrder = sec256k1.order().toString();
+    curveOrder.should.equal('115792089237316195423570985008687907852837564279074904382605163141518161494337');
```
brandonblack, line 104: Let's either specify this as a hex value, as it is defined in [documentation](https://en.bitcoin.it/wiki/Secp256k1) of secp256k1:
`n = FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFE BAAEDCE6 AF48A03B BFD25E8C D0364141`

modules/account-lib/test/unit/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,106 @@
...
+    const testThree = sec256k1
+      .scalarInvert(BigInt('57657865876576467547584635432343132435146576543543134312435443514313234131324551034534'))
+      .toString();
+    testThree.should.equal('45199239020920791752653273479365580486664737121914751844510223387235717825519');
+
+    const testFour = sec256k1.scalarInvert(BigInt('-23232424342224')).toString();
+    testFour.should.equal('109007816100548644677753129517258866095932985532097924870123596204155595187610');
+
+    const testFive = sec256k1.scalarInvert(sec256k1.order() + testValue).toString();
+    testFive.should.equal('16664411509403755791375590925833668071820872792496863322396593062116822075060');
```
brandonblack, line 71: We should also verify that `invert(invert(x)) === x`

brandonblack (changes requested):

modules/account-lib/package.json
brandonblack: Yeah, if it ends up being much longer before he cuts a release, we should at least cut a release, and probably host under bitgo.

modules/account-lib/src/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,56 @@
...
+  scalarAdd(x: bigint, y: bigint): bigint {
+    return bigIntFromBufferBE(Buffer.from(secp.utils.privateAdd(modulo(x), modulo(y).toString(16))));
+  }
+
+  scalarSub(x: bigint, y: bigint): bigint {
+    return modulo(x - y);
+  }
+
+  scalarMult(x: bigint, y: bigint): bigint {
+    return modulo(x * y);
```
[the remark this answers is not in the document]
brandonblack: noble-secp rejected my change to add private multiply, so nvm this suggestion.

modules/account-lib/src/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,56 @@
...
+  scalarReduce(s: bigint): bigint {
+    return modulo(s);
+  }
+
+  scalarNegate(s: bigint): bigint {
+    return bigIntFromBufferBE(Buffer.from(secp.utils.privateNegate(modulo(s))));
+  }
+
+  scalarInvert(s: bigint): bigint {
+    return bigIntFromBufferBE(Buffer.from(secp.utils.privateInvert(bigIntToBufferBE(modulo(s), 32))));
```
brandonblack, line 35: https://github.com/paulmillr/noble-secp256k1/commit/450ebc76ef73ff375c8ce8252468d3a04e87c7f5

noble-secp actually ended up prefering to export `invert` directly as you had originally proposed. Sorry about that! I've updated the `bitgo` branch against his latest `main` so you can sync. up.

modules/account-lib/src/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const order = secp.CURVE.n;
+
+export class Secp256k1Curve implements BaseCurve {
+  scalarRandom(): bigint {
+    return bigIntFromBufferBE(Buffer.from(secp.utils.randomPrivateKey()));
+  }
+
+  scalarAdd(x: bigint, y: bigint): bigint {
+    return bigIntFromBufferBE(Buffer.from(secp.utils.privateAdd(x, y.toString(16))));
```
brandonblack, line 13: I just noticed that this needs to be `bigIntToBufferBE` not `.toString(16)`, as you rightly corrected on other functions.

modules/account-lib/src/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,55 @@
+import { bigIntFromBufferBE, bigIntToBufferBE } from '../util';
```
brandonblack, line 1: I wonder if we should make a `bigIntFromU8ABE` utility function instead of having to do `Buffer.from()` all over the place.

brandonblack (changes requested):
Couple of minor things left. Looking very clean.

modules/account-lib/src/mpc/curves/secp256k1.ts
sachushajiabraham: image
Seems like tweek is expecting a hex value ? @brandonblack

brandonblack:
The `Hex` type in noble-secp is `Uint8Array | string`, Buffer satisfies the
interface, IIUC.

modules/account-lib/src/mpc/curves/secp256k1.ts
```diff
@@ -0,0 +1,55 @@
...
+  scalarRandom(): bigint {
+    return bigIntFromBufferBE(Buffer.from(secp.utils.randomPrivateKey()));
+  }
+
+  scalarAdd(x: bigint, y: bigint): bigint {
+    return bigIntFromBufferBE(Buffer.from(secp.utils.privateAdd(x, y.toString(16))));
+  }
+
+  scalarSub(x: bigint, y: bigint): bigint {
+    const negatedY = Buffer.from(secp.utils.privateNegate(y)).toString('hex');
```
brandonblack, line 17: Should be able to pass `secp.utils.privateNegate(y)` into the next line without converting, I think (because Buffer satisfies Uint8Array).

modules/account-lib/src/mpc/curves/secp256k1.ts
sachushajiabraham: Good Suggestion, Updated

brandonblack (dismissed):

brandonblack:
I believe that Paul on noble-secp is working on adding the build products to his tree similar to how I have them in my fork, so we should be able to move to the mainline soon. Not sure on his plans for a release. Until then, I think it is OK to depend on one commit off mainline with the build products checked in. We should probably update to use the commit hash that we're happy with rather than the branch though.

brandonblack (dismissed):

modules/account-lib/package.json
```diff
@@ -43,6 +43,7 @@
     "@bitgo/bls-dkg": "^1.0.2",
     "@bitgo/statics": "^6.17.0",
     "@bitgo/sdk-core": "^1.0.1",
+    "@noble/secp256k1": "git+https://github.com/brandonblack/noble-secp256k1.git#856129c06436fb696bab7d827edad5effdffaee2",
```
tylerlevine, line 46: let's fork this into the bitgo org instead of using @brandonblack's fork. We can do this in a follow up.

modules/account-lib/src/mpc/curves/ed25519.ts
```diff
@@ -61,7 +61,8 @@ export class Ed25519Curve implements BaseCurve {
     );
   }
 
-  verify(y: bigint, signedMessage: Buffer): Buffer {
-    return Buffer.from(sodium.crypto_sign_open(signedMessage, bigIntToBufferLE(y, 32)));
+  verify(message: Buffer, signature: Buffer, publicKey: bigint): Buffer {
```
tylerlevine, line 64: Why return a buffer? What do the return values mean? Please add a doc comment if it's not going to return a boolean.

tylerlevine (changes requested):
almost there. Thanks for providing in-depth review on this @brandonblack

modules/account-lib/src/mpc/curves/ed25519.ts
sachushajiabraham: From the docs I looked up and reading through testcases, `sodium.crypto_sign_open` returns the original message if the signature is valid. Boolean would have been ideal but I have added a comment to give more clarity on what is being returned. Seems like this is already being utilised and changing to boolean would be a breaking change now. @tylerlevine

sachushajiabraham: @joey-bai just want to bring your attention to this change to the verify function to have it follow same base parameters as ECDSA to have the message also passed to verify after discussing with @johnoliverdriscoll , I have a feeling that this could break some places, double checking with you. I think it would make sense to have it return as a boolean value if it verifies the signature correctly if there is no specific reason to return the actual message if signature is valid.

modules/account-lib/package.json
sachushajiabraham: I have created a ticket BG-48681 for this.

joey-bai (approved):
overall LGTM.

you may also want to get input from @johnoliverdriscoll, who implemented ed25519 + tss

modules/account-lib/src/mpc/curves/ed25519.ts
joey-bai: Makes sense - i agree that boolean makes it more more clear how to use the output of this func

sachushajiabraham: updated to have it as boolean

modules/account-lib/src/mpc/curves/ed25519.ts
```diff
@@ -61,7 +61,16 @@ export class Ed25519Curve implements BaseCurve {
...
 
-  verify(y: bigint, signedMessage: Buffer): Buffer {
-    return Buffer.from(sodium.crypto_sign_open(signedMessage, bigIntToBufferLE(y, 32)));
+  verify(message: Buffer, signature: Buffer, publicKey: bigint): boolean {
+    const signedMessage = Buffer.concat([signature, message]);
+    try {
+      // Returns the message which was signed if the signature is valid
+      const result = Buffer.from(sodium.crypto_sign_open(signedMessage, bigIntToBufferLE(publicKey)));
+      const isValid = !!!Buffer.compare(message, result); // Buffer compare gives 0 if equal and a non zero integer if not
+      return isValid;
```
tylerlevine, line 70: nit: three `!` is two too many - it would be much more readable to check for what you're looking for directly too.

```suggestion
      // Buffer compare gives 0 if equal and a non zero integer if not
      return Buffer.compare(message, result) === 0;
```

tylerlevine (approved):
one nit but feel free to handle in follow up. Nice work 👍

PR #2257, opened 2022-05-05
refactor(root): move encoding convertion to fromRawTx for near

alebusse:
Moved the logic to convert from hex to base64 encoding from bitgo(core) to accountLib   
Added unit tests

[STLX-16118](https://bitgoinc.atlassian.net/browse/STLX-16118)


modules/account-lib/test/unit/coin/near/transaction.ts
```diff
@@ -32,13 +32,25 @@ describe('Near Transaction', () => {
   });
 
   describe('from raw transaction', () => {
-    it('build a transfer from raw', async () => {
+    it('build a transfer from raw hex', async () => {
```
rliberal-simtlix, line 35: Just a comment:
@alebusse why you are no covering signed tx?

alebusse: Its just the title, thats actually the signed one. I will fix it

rliberal-simtlix (approved):

PR #2259, opened 2022-05-06
fix(utxo-lib): update zcash NU5_BRANCH_ID

OttoAllmendinger:
See https://github.com/zcash/zcash/commit/77a971f

Use v4.7.0 as zcash reference node.

Issue: BG-47967 BREAKING CHANGE: signature validity of certain zcash
transaction is changed

6a161f277
chore(utxo-lib): add prettierignore, apply prettier

Issue: BG-47967


brandonblack (dismissed):
Code changes look good to me, but I admit I don't fully understand the sequence of changes needed to stay in sync with ZCash.

modules/utxo-lib/src/bitgo/zcash/ZcashTransaction.ts
```diff
@@ -14,14 +14,15 @@ const ZERO = Buffer.from('000000000000000000000000000000000000000000000000000000
...
 const SAPLING_VERSION_GROUP_ID = 0x892f2085;
+// https://github.com/zcash/zcash/blob/v4.7.0/src/primitives/transaction.h#L52
 const ZIP225_VERSION_GROUP_ID = 0x26a7270a;
 
-// https://github.com/zcash/zcash/blob/v4.5.1/src/consensus/upgrades.cpp#L11
+// https://github.com/zcash/zcash/blob/v4.7.0/src/consensus/upgrades.cpp#L11
 const OVERWINTER_BRANCH_ID = 0x5ba81b19;
 const CANOPY_BRANCH_ID = 0xe9ff75a6;
-const NU5_BRANCH_ID = 0x37519621;
+const NU5_BRANCH_ID = 0xc2d6d0b4;
```
dpc, line 25: So this changed between releases? Kind of weird to change constants like this.

dpc (approved):

modules/utxo-lib/test/integration_local_rpc/generate/fixtures.ts
```diff
@@ -16,7 +16,9 @@ export type Protocol = {
 export function getProtocolVersions(network: Network): number[] {
   switch (getMainnet(network)) {
     case networks.zcash:
-      return [400, 450, 500];
+      // FIXME: re-enable protocol version 500
```
mmcshinsky-bitgo, line 19: What's the impact if this doesn't get fixed by the next stable release?

OttoAllmendinger: from what I can tell: lower test coverage for transactions types we probably don't need yet

modules/utxo-lib/src/bitgo/zcash/ZcashTransaction.ts
OttoAllmendinger: agreed

str4d: The reason this constant changed has been documented in [ZIP 200](https://zips.z.cash/zip-0200#specification) for over four years:

> The relationship between `CONSENSUS_BRANCH_ID` and `ACTIVATION_HEIGHT` is many-to-one: it is possible for many distinct consensus branches to descend from the same parent block (and thus have the same `ACTIVATION_HEIGHT`), but a specific consensus branch can only have one parent block. Concretely, this means that if the `ACTIVATION_HEIGHT` of a network upgrade is changed for any reason (e.g. security vulnerabilities or consensus bugs are discovered), the `CONSENSUS_BRANCH_ID` **MUST** also be changed.

The consensus rules were [altered between 4.5.1 and 4.7.0](https://github.com/zcash/zcash/releases/tag/v4.7.0), which required changing the activation height of NU5 on testnet, which therefore required changing the consensus branch ID. More generally, a consensus branch ID refers to a specific set of consensus rules; any change to that set requires a change to the consensus branch ID.

PR #2261, opened 2022-05-06
feat(statics): add fiatusd and tfiatusd coins

emiacc:
Add `fiatusd` and `tfiatusd` coins to support fiat backing wallets creation.

STLX-13382


modules/statics/test/unit/coins.ts
```diff
@@ -30,7 +30,17 @@ coins.forEach((coin, coinName) => {
...
-        coin.network.type.should.eql(coin.name === coin.family ? NetworkType.MAINNET : NetworkType.TESTNET);
+        if (coin.family !== CoinFamily.FIAT) {
+          coin.network.type.should.eql(coin.name === coin.family ? NetworkType.MAINNET : NetworkType.TESTNET);
+        } else {
+          if (coin.name.startsWith('fiat')) {
+            coin.network.type.should.eql(NetworkType.MAINNET);
+          } else if (coin.name.startsWith('tfiat')) {
+            coin.network.type.should.eql(NetworkType.TESTNET);
+          } else {
+            throw new Error(`Ivalid coin name: ${coin.name} should start with fiat or with tfiat`);
```
drew-bitgo, line 41: I am indifferent to this being an Error over an assert that always fails, but will let SDK team make final decision.

modules/statics/test/unit/coins.ts
```diff
@@ -30,7 +30,17 @@ coins.forEach((coin, coinName) => {
 
     if (!coin.isToken) {
       it(`has expected network type`, function () {
-        coin.network.type.should.eql(coin.name === coin.family ? NetworkType.MAINNET : NetworkType.TESTNET);
+        if (coin.family !== CoinFamily.FIAT) {
```
drew-bitgo, line 33: I think this should be the other way around, so `if (coin.family === CoinFamily.FIAT) {` with our fiat specific tests in that if, instead of if not else. Better prepares for future edge cases, if they arise

modules/statics/src/coins.ts
```diff
@@ -115,6 +115,32 @@ export const coins = CoinMap.fromCoins([
   account('txtz', 'Testnet Tezos', Networks.test.xtz, 6, UnderlyingAsset.XTZ, XTZ_FEATURES),
   account('susd', 'Silvergate USD', Networks.main.susd, 2, UnderlyingAsset.USD),
   account('tsusd', 'Testnet Silvergate USD', Networks.test.susd, 2, UnderlyingAsset.USD),
+  account(
```
drew-bitgo, line 118: There are existing examples of both formats (all in one line vs all spread out), is there one in particular the SDK team prefers?

modules/statics/src/account.ts
```diff
@@ -36,8 +37,8 @@ export class AccountCoin extends BaseCoin {
 
   constructor(options: AccountConstructorOptions) {
     super({
-      ...options,
       kind: CoinKind.CRYPTO,
```
drew-bitgo, line 40: since `kind` is now supplied in options I do not think the spread (`...`) is needed (or `kind` here)

drew-bitgo (commented):
All of my comments are strictly formatting, no functional changes. Otherwise I Approve. A few questions for the SDK team, as well.

modules/statics/test/unit/coins.ts
emiacc: thanks! I did some clean-up in the tests. I moved the tests related to fiat tokens to the `fiatTokens.ts` file and use `fiat.ts` file to check fiat coins.

modules/statics/src/coins.ts
emiacc: The format is forced by prettier (all spread out when the line contains certain amount of characters)

modules/statics/src/account.ts
emiacc: We still need the spread and the `kind` parameter since in the "AccountConstructorOptions" interface `kind` is optional and in the base class "BaseCoin" it's required.

modules/statics/src/coins.ts
drew-bitgo: ahh, good to know

drew-bitgo (dismissed):

tylerlevine (commented):
seems pretty reasonable. Debatably we should not reuse the `account` convenience function for creating fiat coin instances and should have a separate one. And maybe even a separate child class from `BaseCoin` altogether. What do you think?

emiacc:
> seems pretty reasonable. Debatably we should not reuse the `account` convenience function for creating fiat coin instances and should have a separate one. And maybe even a separate child class from `BaseCoin` altogether. What do you think?

Thanks @tylerlevine for your suggestions. I added the new child class and the new factory function to create the Fiat coins.

tylerlevine (approved):
lgtm

PR #2262, opened 2022-05-06
feat(statics): remove fiat tokens

emiacc:
Remove unnecessary fiat token implementation.
Tokens will be replaced for coins in a new PR

STLX-13382


drew-bitgo:
Through testing and discovery, we found that this approach is no longer desired.

This PR will have some minor conflicts with another PR once merged, so not giving it the green check quite yet.

emiacc:
> This PR will have some minor conflicts with another PR once merged, so not giving it the green check quite yet.

I branched out the other PR from this one. 
We should merge this one first and then change the source branch on the other and there wont be conflicts.

tylerlevine (approved):

PR #2273, opened 2022-05-10
feat(bitgo): support chaincodes on BLS-DKG keychains creation

fperalta-simtlix:
As part of the BLS-DKG HS implementation, chaincodes needs to be supported at the time of keychain
creation so then it is possible to derive the subkeys when signing tx. Each participant contributes
with its own chaincode and those are added up to the common keychain.

Also fixes BG-47812. Add padding to the secret shares hex so it always
is 64. It sometimes was 62.

BG-46184

### DIFF
--- a/modules/sdk-coin-eth2/test/unit/keyPair.ts
+++ b/modules/sdk-coin-eth2/test/unit/keyPair.ts
@@ -1,4 +1,5 @@
 import assert from 'assert';
+import { randomBytes } from 'crypto';
 import should from 'should';
 import { KeyPair } from '../../src';
 import * as testData from '../resources/eth2';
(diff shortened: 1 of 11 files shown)


wesgraham (approved):
lgtm - will defer to john for final approval

modules/sdk-coin-eth2/test/unit/keyPair.ts
```diff
@@ -296,5 +333,38 @@ describe('Eth2 Key Pair', () => {
...
+      const prv1 = KeyPair.aggregatePrvkeys([secrets1[0], secrets2[0], secrets3[0]]);
+      const signKeyPair1 = new KeyPair({ prv: prv1 });
+      const sig1 = await signKeyPair1.sign(msg);
+      const prv3 = KeyPair.aggregatePrvkeys([secrets1[2], secrets2[2], secrets3[2]]);
+      const signKeyPair3 = new KeyPair({ prv: prv3 });
+      const sig3 = await signKeyPair3.sign(msg);
+      const aggregatedSig = KeyPair.aggregateSignatures({ 1: BigInt(sig1), 3: BigInt(sig3) });
+      const derivedCommonPub = KeyPair.aggregatePubkeys([childKey.publicShare, pub2, pub3]);
+
+      (await KeyPair.verifySignature(derivedCommonPub, msg, aggregatedSig)).should.be.true();
```
wesgraham, line 367: awesome!

wesgraham (approved):

mmcshinsky-bitgo (approved):

PR #2274, opened 2022-05-10
STLX-15823 - Add parse transaction in core

fmartinAltamirano:
Implement parse transaction in bitgo core for near adding method and interfaces

STLX-15823


modules/bitgo/test/v2/unit/coins/near.ts
```diff
@@ -554,4 +553,165 @@ describe('NEAR:', function () {
       }
     });
   });
+
+  describe('Parse Transactions:', () => {
```
ipelaez-simtlix, line 557: I'll elaborate my concern from test file even if it its actually related to code.

Can you add a test that has a huge amount as a number? because you allow from the TransactionOutput/Input numbers
`amount: number | string;` however Near has 24 decimal places, I think the posibility to broke this code is open.

Shouldnt be only string allowed? And why on L:306 we are parsing Input amount to string 
https://github.dev/BitGo/BitGoJS/blob/45c263271e8ff683dba4a8b3e8258fcb9b976414/modules/bitgo/src/v2/coins/near.ts#L306

However output amount could be both types, number or string?
https://github.dev/BitGo/BitGoJS/blob/45c263271e8ff683dba4a8b3e8258fcb9b976414/modules/bitgo/src/v2/coins/near.ts#L313

Another question, where basecoin.parseTransaction() is being used, it is a problem if we only return a string if this is the approach we are taking?

ipelaez-simtlix (changes requested):

modules/bitgo/test/v2/unit/coins/near.ts
rliberal-simtlix: We need only I think to consider string and not number.
Number of zeros is risky.

modules/bitgo/src/v2/coins/near.ts
```diff
@@ -47,6 +47,33 @@ export interface VerifiedTransactionParameters {
   signer: string;
 }
 
+export interface TransactionFee {
+  fee: string;
```
rliberal-simtlix, line 51: Why this interface?

fmartinAltamirano: I checked, and it was not in use

modules/bitgo/test/v2/unit/coins/near.ts
fmartinAltamirano: @ipelaez-simtlix  as we discussed, I changed the tests to verify that they work with a large amount and also converted them to string.

modules/bitgo/test/v2/unit/coins/near.ts
```diff
@@ -554,4 +554,163 @@ describe('NEAR:', function () {
       }
     });
   });
+
+  describe.only('Parse Transactions:', () => {
```
ipelaez-simtlix, line 558: remember to delete .only and imports

modules/bitgo/test/v2/unit/coins/near.ts
```diff
@@ -554,4 +554,163 @@ describe('NEAR:', function () {
...
+    const amount = '10000000000000000000000000000000';
```
ipelaez-simtlix, line 559: did you declare amount but didnt use any place?

ipelaez-simtlix (changes requested):
lets change this to PR after address my change requests if you think its done

modules/bitgo/test/v2/unit/coins/near.ts
fmartinAltamirano: Thanks, I forgot to remove them

modules/bitgo/test/v2/unit/coins/near.ts
```diff
@@ -554,4 +553,163 @@ describe('NEAR:', function () {
...
+        },
+        feeInfo: {
+          fee: '',
+        },
+      });
+
+      parsedTransaction.should.deepEqual({
+        inputs: [{
+          address: '9f7b0675db59d19b4bd9c8c72eaabba75a9863d02b30115b8b3c3ca5c20f0254',
+          amount: '1000000000000000000000000',
```
ipelaez-simtlix, line 598: then create a constant for these variables and lets add a coment about how much near is

modules/bitgo/test/v2/unit/coins/near.ts
ipelaez-simtlix: answer on another comment

modules/bitgo/test/v2/unit/coins/near.ts
fmartinAltamirano: Done!

modules/bitgo/src/v2/coins/near.ts
```diff
@@ -47,6 +47,29 @@ export interface VerifiedTransactionParameters {
...
+  txPrebuild: TransactionPrebuild;
+  publicKey: string;
+  feeInfo: {
+    fee: string;
+  };
+}
+
+interface TransactionOutput {
+  address: string;
+  amount: number | string;
```
ipelaez-simtlix: remove number and use only string to avoid scientific notation.

ipelaez-simtlix (dismissed):
lgtm

ipelaez-simtlix (approved):

PR #2276, opened 2022-05-11
feat(statics): create statics for dogecoin

jonathancn-bitgo:
add statics for dogecoin with generic utxo.

Ticket: STLX-16142


shiksb (approved):
Nice work!

PR #2277, opened 2022-05-11
chore(account-lib): upgrade to es2019

bitgoAaron:
- sets account-lib's build target to same as root (es2019)
- replaces should.throws with assert.throws


ducnguyen-bitgo (approved):
The PR looks good to me. Also, I just pulled it to my local machine and ran all the unit tests. Everything was passed.
image

tylerlevine (approved):
nice work

PR #2278, opened 2022-05-13
feat(sdk-core): select hsmpub key based on node env

shiksb:
Select the HSM pub key based on the node environment.

STLX-16438

### DIFF
--- a/modules/sdk-core/src/v2/environments.ts
+++ b/modules/sdk-core/src/v2/environments.ts
@@ -172,7 +172,8 @@ export const Environments: Environments = {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     uri: process.env.BITGO_CUSTOM_ROOT_URI!,
     network: process.env.BITGO_CUSTOM_BITCOIN_NETWORK as V1Network,
-    hsmXpub: hardcodedPublicKeys.hsmXpub.dev,
+    hsmXpub:
+      process.env.NODE_ENV === 'production' ? hardcodedPublicKeys.hsmXpub.prod : hardcodedPublicKeys.hsmXpub.test,
     btcExplorerBaseUrl:
       process.env.BITGO_CUSTOM_BITCOIN_NETWORK !== 'bitcoin'
         ? 'https://blockstream.info/testnet/api'


mmcshinsky-bitgo (dismissed):

modules/sdk-core/src/v2/environments.ts
```diff
@@ -172,7 +172,12 @@ export const Environments: Environments = {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     uri: process.env.BITGO_CUSTOM_ROOT_URI!,
     network: process.env.BITGO_CUSTOM_BITCOIN_NETWORK as V1Network,
-    hsmXpub: hardcodedPublicKeys.hsmXpub.dev,
+    hsmXpub:
+      process.env.NODE_ENV === 'production'
+        ? hardcodedPublicKeys.hsmXpub.prod
+        : process.env.NODE_ENV === 'test'
```
tylerlevine, line 178: This doesn't get to `test` set for testnet. We either need another way of detecting we're running in testnet mode or we need to add a new env var to represent the HSM key.

`NODE_ENV` is an env var used in many packages in the nodejs ecosystem to do things like hide debug logs, etc

tylerlevine (changes requested):

modules/sdk-core/src/v2/environments.ts
tylerlevine: See https://nodejs.dev/learn/nodejs-the-difference-between-development-and-production for a bit more info on this.

shiksb: ahh I see this is the nodejs env

modules/sdk-core/src/v2/environments.ts
```diff
@@ -172,7 +172,8 @@ export const Environments: Environments = {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     uri: process.env.BITGO_CUSTOM_ROOT_URI!,
     network: process.env.BITGO_CUSTOM_BITCOIN_NETWORK as V1Network,
-    hsmXpub: hardcodedPublicKeys.hsmXpub.dev,
+    hsmXpub:
+      process.env.NODE_ENV === 'production' ? hardcodedPublicKeys.hsmXpub.prod : hardcodedPublicKeys.hsmXpub.test,
```
mmcshinsky-bitgo, line 176: Do you want `dev` instead of `test`?

mmcshinsky-bitgo (changes requested):

modules/sdk-core/src/v2/environments.ts
shiksb: I think it should be `test` since clients can use NODE_ENV=development for testing out features in bitgo-test

mmcshinsky-bitgo (approved):

mmcshinsky-bitgo (approved):

PR #2283, opened 2022-05-16
feat(SDK-avax-p-lib): add avax-p-lib module

sskim16:
add avax-p-lib module decoupled from account lib.  Avaxp is a utxo based
coin but uses features of account lib.

Ticket: STLX-15980

### DIFF
--- a/modules/bitgo/src/v2/coins/tavaxp.ts
+++ b/modules/bitgo/src/v2/coins/tavaxp.ts
@@ -1,0 +1,31 @@
+import { BitGo } from '../../bitgo';
+import { BaseCoin } from '../baseCoin';
+import { AvaxP } from './avaxp';
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+
+export class TavaxP extends AvaxP {
+
+  protected readonly _staticsCoin: Readonly<StaticsBaseCoin>;
+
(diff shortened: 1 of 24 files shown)


modules/avax-p-lib/package.json
```diff
@@ -0,0 +1,110 @@
+{
+  "name": "@bitgo/avax-p-lib",
```
mmcshinsky-bitgo, line 2: You'll need to add this package to the root `tsconfig.packages.json` file as well.

sskim16: updated!

modules/bitgo/src/v2/coins/tavaxp.ts
```diff
@@ -0,0 +1,31 @@
+import { BitGo } from '../../bitgo';
+import { BaseCoin } from '../baseCoin';
+import { AvaxP } from './avaxp';
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+
+export class TavaxP extends AvaxP {
```
shiksb, line 6:
```suggestion
export class TAvaxP extends AvaxP {
```

sskim16: I think we should follow the syntax of avaxc and keep things consistent.  TavaxC is how its written.

mmcshinsky-bitgo (approved):

PR #2288, opened 2022-05-17
Ducnguyen/bitgobase interface

ducnguyen-bitgo:
1. Defined a BitGoBase interface in `sdk-core` used by baseCoin, enterprises, keychains, markets, pendingApprovals, wallets, and etc.
2. Moved api interface from `sdk-api` to `sdk-core`.
3. Extract keychain, webhooks, wallet, wallets, enterprise, enterprises, pendingApproval, pendingApprovals, and tradingAccount interfaces into `sdk-core`.
4. Added baseCoin interface into `sdk-core`.
5. Updated all coins needed imports to new locations in `sdk-core`.
6. Changed baseCoin and its dependent classes in `bitgo` to implement corresponding interfaces in `sdk-core`.
7. Updated unit test imports.
8. Resolved naming collisions for the types `BlsKeyPair` and `TransactionExplanation` which were defined in both `account-lib` and `bitgo`.
9. `TransactionExplanation` previously defined in `bitgo` is now `ITransactionExplanation`. `TransactionExplanation` previously defined in account-lib extends this generic type.
10. `BlsKeyPair` previously defined in `bitgo` is now `IBlsKeyPair`. `BlsKeyPair` previously defined in `account-lib` remains unchanged.


modules/sdk-core/src/bitgo/bitgoBase.ts
```diff
@@ -0,0 +1,43 @@
+import * as superagent from 'superagent';
+import { EnvironmentName } from './environments';
+
+export interface BitGoRequest<ResultType = any> extends superagent.SuperAgentRequest {
```
bitgoAaron, line 4: most of these interfaces already exist in `@bitgo/sdk-api`. Those should be moved here, not duplicated.

modules/sdk-core/src/bitgo/bitgoBase.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface GetSharingKeyOptions {
+  email: string;
+}
+
+export interface IRequestTracer {
+  inc(): void;
+  toString(): string;
+}
+
+export interface BitGoBase {
```
bitgoAaron, line 27: `BitGoAPI` in `@bitgo/sdk-api` needs to implement this interface.

ducnguyen-bitgo: Should it be part of ticket BG-48170 or should we use a different ticket for this?

modules/sdk-core/src/bitgo/bitgoBase.ts
```diff
@@ -0,0 +1,43 @@
...
+  coin(coinName: string): any; // need to change it to BaseCoin once it's moved to @bitgo/sdk-core
```
bitgoAaron, line 28: we shouldn't use `any` here. We need to define the `BaseCoin` interface, which requires defining the other interfaces. 🐔  <-> 🥚

modules/sdk-core/src/bitgo/bitgoBase.ts
ducnguyen-bitgo: Do you mean we should create a folder called `sdk-api` in `sdk-core` and create a file i.e `api.ts` to define the BitGoRequest interface?

bitgoAaron: yes

modules/sdk-core/src/bitgo/bitgoBase.ts
bitgoAaron: it should be part of this

modules/sdk-core/src/bitgo/bitgoBase.ts
ducnguyen-bitgo: I agree with you on that. The reason I used `any` there was to resolve an issue in one of the classes in `bitgo` module. Also, `BaseCoin` is still in `bitgo` module so I did not want to import `BaseCoin` from `bitgo` since it would create a circular dependency between `bitgo` and `sdk-core`.

bitgoAaron (approved):

OttoAllmendinger (approved):

PR #2289, opened 2022-05-17
fix: v1 getWallet call

DariusParvin:
`bitgo.wallets().get()` mistakenly had a `this.` in front of it, causing the function to be `undefined`.  

This type of bug isn't caught automatically by typescript (surprisingly), but we can add `no-invalid-this` to the lint rules to prevent it. A ticket for that has been created: https://bitgoinc.atlassian.net/browse/BG-48397

Ticket: BG-48417


tylerlevine (approved):
nice fix

PR #2290, opened 2022-05-17
fix(bitgo): use dynamic config for eos decimal places

byhow:
CHEX is having issue displaying bignumber decimals where we
output 4 digits of decimals while they output 8 decimals.

Rounding up the number to reflect the changes.

Ticket: BG-47340

### DIFF
--- a/modules/bitgo/src/v2/coins/eos.ts
+++ b/modules/bitgo/src/v2/coins/eos.ts
@@ -223,6 +223,10 @@ export class Eos extends BaseCoin {
     return 1e4;
   }
 
+  get decimalPlaces(): number {
+    return 4;
+  }
+
   /**
    * Flag for sending value of 0
    * @returns {boolean} True if okay to send 0 value, false otherwise
@@ -361,7 +365,7 @@ export class Eos extends BaseCoin {
     const dividend = this.getBaseFactor();
     const bigNumber = new BigNumber(baseUnits).dividedBy(dividend);
     // set the format so commas aren't added to large coin amounts
-    return bigNumber.toFormat(4, null as any, { groupSeparator: '', decimalSeparator: '.' });
+    return bigNumber.toFormat(this.decimalPlaces, null as any, { groupSeparator: '', decimalSeparator: '.' });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


joey-bai (approved):

modules/bitgo/src/v2/coins/eos.ts
```diff
@@ -223,6 +223,10 @@ export class Eos extends BaseCoin {
     return 1e4;
   }
 
+  get decimalPlaces(): number {
+    return 4;
```
mmcshinsky-bitgo, line 227: I'm not sure how this fixes the ticket? It will alway still be 4 since we `eos: Object.freeze(new Eos()).` The ticket seems to refer to using the token which is set to `8` as opposed to the baseCoin which is `4`. Am In understanding this correctly? @byhow

mmcshinsky-bitgo (changes requested):

modules/bitgo/src/v2/coins/eos.ts
byhow: since `CHEX` is a token, it is using the `get decimalPlaces()` method that is defined in `eosToken.ts` , which it is able to get the correct amount of configurated decimal places of `8`. 

However when it tries to call `baseUnitsToBigUnits` from token, since it is hardcoded to `4`, even the decimals are set correctly, the unit transform will always be `4` regardless of the config. 

By making `this.decimalPlaces` in the base class, I was expecting it to be overrode in the child token class. is there anything we can do to override it to use the token configured decimals?

mmcshinsky-bitgo: `eos:CHEX` implements `EosCoin` which is different than `eos` which implements `Eos`. `EosCoin` doesn't have a `get decimalPlaces()` which derives all the way back to `BaseCoin`.

mmcshinsky-bitgo: `eos` is also derived from `BaseCoin` but is used for the coin, while `eos:CHEX` is used for account tokens.

byhow: quick question: how do you find out whether `eos:CHEX` extends on `EosCoin` since the eos tokens `EosToken` do extend on `Eos`

mmcshinsky-bitgo: @byhow Do we have an update for this one, or are we good to close this out?

byhow: @mmcshinsky-bitgo Hi, let's close this out as this ticket is transferred to other teams now

PR #2291, opened 2022-05-18
chore(express): bump bitgo version in express

shiksb:
This version corrects the environment to get the right hsmpub key
https://github.com/BitGo/BitGoJS/pull/2278

STLX-16438


tylerlevine (changes requested):
This version should only be updated as part of the release process

shiksb:
will request a release on sdk channel

PR #2292, opened 2022-05-18
fix(bitgo): getUnspentInfo to handle missing unspents

DariusParvin:
Ticket: BG-48421


modules/bitgo/src/v2/coins/utxo/recovery/crossChainRecovery.ts
```diff
@@ -47,8 +46,24 @@ class BitgoPublicApi {
    * @returns {*}
    */
   async getUnspentInfo(addresses: string[]): Promise<Unspent[]> {
-    const url = this.coin.url(`/public/addressUnspents/${_.uniq(addresses).join(',')}`);
-    return (await request.get(url)).body as Unspent[];
+    const res = await Bluebird.map(
```
DariusParvin, line 49: I just realized that at this point it would make sense to get rid of any duplicate addresses in `addresses`. I'm pushing a commit for it.

modules/bitgo/src/v2/coins/utxo/recovery/crossChainRecovery.ts
```diff
@@ -47,8 +46,25 @@ class BitgoPublicApi {
    * @returns {*}
    */
   async getUnspentInfo(addresses: string[]): Promise<Unspent[]> {
-    const url = this.coin.url(`/public/addressUnspents/${_.uniq(addresses).join(',')}`);
-    return (await request.get(url)).body as Unspent[];
+    const uniqueAddresses = new Set(addresses);
```
mmcshinsky-bitgo, line 49: Is there any merit to doing the one call first to the API and then attempting this secondary approach only if it is 404?

DariusParvin: yeah that's a good idea, i'll add that now. Thanks!

DariusParvin:
@mmcshinsky-bitgo ready for re-review. Sorry the force push on [a3cc42a](https://github.com/BitGo/BitGoJS/commit/a3cc42a4305ca55b83500739c3d0b499b23a7a21) was premature. I added the change you suggested and updated the tests (with some minor improvements). Here is the diff since your previous review:

https://github.com/BitGo/BitGoJS/compare/cef988f..8fe1ae9094c728f25f0cc3a53ec170c1c348b49d

mmcshinsky-bitgo (approved):

PR #2299, opened 2022-05-23
chore(sdk-core): refactor bitgo BaseCoin to sdk-core

ducnguyen-bitgo:
- Refactored bitgo BaseCoin to sdk-core.

Ticket: BG-48167

### DIFF
--- a/modules/sdk-core/src/bitgo/index.ts
+++ b/modules/sdk-core/src/bitgo/index.ts
@@ -8,7 +8,6 @@ export * from './market';
 export * from './pendingApproval';
 export * from './recovery';
 export * from './trading';
-export * from './trading';
 export * from './types';
 export * from './utils';
 export * from './wallet';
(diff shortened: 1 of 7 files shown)


modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
```diff
@@ -0,0 +1,413 @@
...
+  SignTransactionOptions,
+  SupplementGenerateWalletOptions,
+  TransactionPrebuild,
+  TransactionType,
+  VerifyAddressOptions,
+  VerifyTransactionOptions,
+  Wallet,
+  Wallets,
+  Webhooks,
+} from '../../index';
```
bitgoAaron, line 44: we should avoid importing types from the barrel, as this is a good way to accidentally introduce circular references. These should all import directly from the source file, or at least the index of the type they're importing from. For example:

`import { Enterprise } from '../../enterprise';`

ducnguyen-bitgo:
Closing it as it doesn't preserve the history of files moving from bitgo to sdk-core.

PR #2302, opened 2022-05-25
chore(sdk-core): refactor BaseCoin from bitgo to sdk-core.

ducnguyen-bitgo:
- Moved baseCoin to sdk-core.
- Moved enterprise and enterprises to sdk-core.
- Moved internal and keycard to sdk-core.
- Moved keychains to sdk-core.
- Moved markets to sdk-core.
- Moved pendingApproval and pendingApprovals to sdk-core.
- Moved affirmation and affirmations to sdk-core.
- Moved settlement and settlements to sdk-core.
- Moved tradingAccount to sdk-core.
- Moved tradingPartner, and tradingPartners to sdk-core.
- Moved blsUtils, mpcUtils, opengpgUtils, promise-utils to sdk-core.
- Moved tssUtils, util to sdk-core.
- Moved wallet, and wallets to sdk-core.
- Moved webhooks to sdk-core.
- Moved bip32util, bitcoin, ecdh, and tss to sdk-core.

Ticket: BG-48167


modules/sdk-core/src/bitgo/utils/abstractUtxoCoinUtil.ts
```diff
@@ -0,0 +1,7 @@
+import * as utxolib from '@bitgo/utxo-lib';
+import ScriptType2Of3 = utxolib.bitgo.outputScripts.ScriptType2Of3;
+import { isChainCode, scriptTypeForChain } from '@bitgo/utxo-lib/dist/src/bitgo';
```
bitgoAaron, line 3: let's avoid importing from `/dist` folders. Can we instead make sure utxo-lib exports these types correctly?

modules/sdk-core/package.json
```diff
@@ -37,15 +37,21 @@
...
     "@stablelib/hex": "^1.0.0",
+    "big.js": "^3.1.3",
     "bignumber.js": "^9.0.0",
     "bip32": "^2.0.6",
     "bitcoinjs-lib": "npm:@bitgo/bitcoinjs-lib@6.1.0-rc.3",
+    "bitcoinjs-message": "^2.0.0",
     "bs58": "^4.0.1",
     "create-hmac": "^1.1.7",
+    "debug": "^3.1.0",
+    "libsodium-wrappers-sumo": "^0.7.9",
```
bitgoAaron, line 50: same here. Remove this from bitgo's package.json

bitgoAaron:
looks like all of the files under `sdk-core/src/account-lib/mpc` were manually copied instead of `git mv`, thus losing the history. Can you redo these as well? We need to keep track of as much as possible

bitgoAaron (changes requested):
see comments

modules/sdk-core/package.json
ducnguyen-bitgo: I think `libsodium-wrappers-sumo` is from account-lib's package.json. I'm removing it now.

modules/sdk-core/src/bitgo/utils/abstractUtxoCoinUtil.ts
ducnguyen-bitgo: Those functions are exported under `utxo-lib/src/bitgo` so I guess we could just do `import { bitgo } from '@bitgo/utxo-lib'` and then destructure those functions from `bitgo`?

modules/sdk-core/src/api/bip32path.ts
```diff
@@ -0,0 +1,42 @@
+/**
```
bitgoAaron, line 1: the file should be `git mv`'d, not copy/pasted.

bitgoAaron (changes requested):
2 more copy/pasted files need to be git mv'd to preserve history.

modules/sdk-core/src/bitgo/utils/util.ts
```diff
@@ -0,0 +1,199 @@
+/**
```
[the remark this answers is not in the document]
ducnguyen-bitgo: @bitgoAaron The reason we didn't git-mv it was to keep it backward compatible for bitgo-express until the next release of sdk-core.

bitgoAaron (approved):

PR #2305, opened 2022-05-26
fix(bitgo): fix v1 wallet get address

DariusParvin:
Add missing 'Bluebird.resolve()' to get address call

Ticket: BG-44440

### DIFF
--- a/modules/bitgo/src/wallet.ts
+++ b/modules/bitgo/src/wallet.ts
@@ -449,9 +449,10 @@ Wallet.prototype.address = function (params, callback) {
 
   const url = this.url('/addresses/' + params.address);
 
-  return this.bitgo.get(url)
-    .result()
-    .nodeify(callback);
+  return Bluebird.resolve(
+    this.bitgo.get(url)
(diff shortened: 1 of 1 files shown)


modules/bitgo/src/wallet.ts
```diff
@@ -449,9 +449,10 @@ Wallet.prototype.address = function (params, callback) {
 
   const url = this.url('/addresses/' + params.address);
 
-  return this.bitgo.get(url)
-    .result()
-    .nodeify(callback);
+  return Bluebird.resolve(
```
mmcshinsky-bitgo, line 452: What is the needed response that is being returned that is currently being returned with the previous code?

DariusParvin: previously it would throw an error saying `this.bitgo.get(...).result(...).nodeify is not a function`. Based on other places where `.nodify()` was used, it seemed like something specific to Bluebird requests.

mmcshinsky-bitgo (approved):

PR #2306, opened 2022-05-26
feat(account-lib): fix is valid account id

facundoCarreSimtlix:
fix isValidAccountId method

What happened?
If an empty string is sent to isValidAccountId it returns true.

Requeriments:
Should not pass for empty strings ' '.
STLX-16458

### DIFF
--- a/modules/account-lib/src/coin/near/utils.ts
+++ b/modules/account-lib/src/coin/near/utils.ts
@@ -46,16 +46,12 @@ export class Utils implements BaseUtils {
    * or can create a new keyPair
    */
   isValidAccountId(accountId: string): boolean {
-    try {
-      // this regex is from near doc, https://docs.near.org/docs/concepts/account
+    return (
       (/^(([a-z\d]+[\-_])*[a-z\d]+\.)*([a-z\d]+[\-_])*[a-z\d]+$/.test(accountId) &&
-        accountId.length > 2 &&
-        accountId.length < 64) ||
-        new KeyPair({ pub: accountId });
-      return true;
-    } catch {
-      return false;
-    }
+        accountId.length >= 2 &&
+        accountId.length <= 64) ||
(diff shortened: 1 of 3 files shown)


ipelaez-simtlix (approved):

modules/account-lib/src/coin/near/utils.ts
```diff
@@ -46,14 +46,15 @@ export class Utils implements BaseUtils {
    * or can create a new keyPair
    */
   isValidAccountId(accountId: string): boolean {
-    try {
+    if (
```
silalvado, line 49: You can avoid this "if" and directly return the logical expression

silalvado (approved):

ipelaez-simtlix (dismissed):

rliberal-simtlix (dismissed):

modules/account-lib/src/coin/near/utils.ts
```diff
@@ -46,16 +46,12 @@ export class Utils implements BaseUtils {
    * or can create a new keyPair
    */
   isValidAccountId(accountId: string): boolean {
-    try {
-      // this regex is from near doc, https://docs.near.org/docs/concepts/account
+    return (
       (/^(([a-z\d]+[\-_])*[a-z\d]+\.)*([a-z\d]+[\-_])*[a-z\d]+$/.test(accountId) &&
         accountId.length > 2 &&
```
ducnguyen-bitgo, line 51: @facundoCarreSimtlix According to the near [doc](https://docs.near.org/docs/concepts/account), 2 is a valid account length so shouldn't it be `accountId.length >= 2`?

facundoCarreSimtlix: yes, You are right, i fix that

ducnguyen-bitgo (approved):
Lgtm.

PR #2307, opened 2022-05-26
chore(sdk): add Euro support

drew-bitgo:
TICKET: BG-47335


ducnguyen-bitgo (dismissed):
Lgtm. It is similar to the fiatusd's approved and merged [PR](https://github.com/BitGo/BitGoJS/pull/2261).

modules/bitgo/test/v2/unit/coins/ofcToken.ts
```diff
@@ -23,6 +23,13 @@ describe('OFC:', function () {
     otestusdCoin.getBaseFactor().should.equal(PRECISION_2);
   });
 
+  it('functions that return constants', function () {
```
ducnguyen-bitgo, line 26: Should we change this to something different than `functions that return constants` as it was used for the ofctusd test above?

PR #2311, opened 2022-05-27
chore(root): add noImplicitThis to tsconfig.json

DariusParvin:
Without fixing errors associated with v1 code or tests.

Ticket: BG-48796


mmcshinsky-bitgo:
If we're going to introduce this rule, we should fix all the issues rather than bypass them.

DariusParvin:
@tylerlevine recommended not spending time fixing the v1 wallet issues. My guess was that it's not worth fixing them for the tests either.

PR #2319, opened 2022-05-31
fix(utxolib): always use VERSION4_BRANCH_NU5 for zcash

OttoAllmendinger:
Issue: BG-47967


mmcshinsky-bitgo:
BG-47967 looks to be a non-related ticket?

tylerlevine (approved):
We should be using ticket [CR-604](https://bitgoinc.atlassian.net/browse/CR-604) for this.

PR #2330, opened 2022-06-02
fix(account-lib): fix sdk avax build issues

sachushajiabraham:
this change fixes build issues caused by avax files by correctly importing basecoin from sdk core

TICKET: BG-49438


ducnguyen-bitgo (approved):
Lgtm.

PR #2333, opened 2022-06-02
feat: move coin factory to sdk-core and add coin registration to sdk-api 

mmcshinsky-bitgo:
- BG-49173: Refactor CoinFactory to sdk-core
- BG-49173: Register coins from sdk-api

- Backwards compatibility allows all current coins to be registered in `bitgo`.
- `@bitgo/sdk-api` can now register coins independent from all other coins with all the same functionalities of `bitgo`.
- In `@bitgo/sdk-api`, `BitGoAPI` has been refactored to `AbstractBitGoAPI` with a consumer implementation of `BitGoAPI` now available.
  - Updated `web-demo` and `bitgo` to reflect the change.

## Specific PR Queries
- Is the `AbstractBitGoAPI` an acceptable change. I would have extended `class BitGo extends BitGoAPI` in `@bitgo/sdk-api`, except that would have created a duplicate export in `bitgo` since `bitgo` exports everything from `@bitgo/sdk-api`.
- erc20Token is a specific dependency of how `coinFactory` operates, but that would be counter to having `@bitgo/sdk-api` rely on this class.
  - All current coins properly register with the removal of the old implementation of the removed lines 223-237 of the ethConstructor/token check. Would like insight into if there is an instance where this is used that I may be missing.

## Example - Coin Registration

```javascript
import { Txrp } from '@bitgo/sdk-coin-xrp';

const sdk = new BitGoAPI();

sdk.register('txrp', Txrp.createInstance);
```

BG-49173

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -20,11 +20,13 @@ import {
   AliasEnvironments,
   BitGoBase,
   BitGoRequest,
+  CoinConstructor,
   common,
   DecryptOptions,
   EncryptOptions,
   EnvironmentName,
   GetSharingKeyOptions,
+  GlobalCoinFactory,
   IBaseCoin,
   IRequestTracer,
   sanitizeLegacyPath,
@@ -33,26 +35,26 @@ import { getAddressP2PKH, makeRandomKey } from './util';
 import * as sjcl from '@bitgo/sjcl';
 import {
   AccessTokenOptions,
-  PingOptions,
+  AddAccessTokenOptions,
   AuthenticateOptions,
+  BitGoAPIOptions,
+  BitGoJson,
   CalculateHmacSubjectOptions,
   CalculateRequestHeadersOptions,
   CalculateRequestHmacOptions,
+  ExtendTokenOptions,
+  GetUserOptions,
+  PingOptions,
   ProcessedAuthenticationOptions,
+  RemoveAccessTokenOptions,
   RequestHeaders,
+  TokenIssuance,
+  TokenIssuanceResponse,
+  UnlockOptions,
+  User,
+  VerifyPasswordOptions,
   VerifyResponseInfo,
   VerifyResponseOptions,
-  BitGoAPIOptions,
-  User,
-  BitGoJson,
-  VerifyPasswordOptions,
-  TokenIssuanceResponse,
-  TokenIssuance,
-  RemoveAccessTokenOptions,
-  AddAccessTokenOptions,
-  GetUserOptions,
-  UnlockOptions,
-  ExtendTokenOptions,
 } from './types';
 import pjson = require('../package.json');
 import { decrypt, encrypt } from './encrypt';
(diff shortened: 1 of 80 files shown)


modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -66,7 +66,7 @@ if (!(process as any)?.browser) {
 
 const patchedRequestMethods = ['get', 'post', 'put', 'del', 'patch'] as const;
 
-export abstract class BitGoAPI implements BitGoBase {
+export abstract class AbstractBitGoAPI implements BitGoBase {
```
bitgoAaron, line 69: can we just make this `export class BitGoAPI implements BitGoBase ` ? I don't think this class needs to be `abstract` after-all & would make things a bit simpler.

modules/sdk-core/src/bitgo/errors.ts
```diff
@@ -20,7 +20,7 @@ export class NodeEnvironmentError extends BitGoJsError {
 export class UnsupportedCoinError extends BitGoJsError {
   public constructor(coin: string) {
     super(
-      `Coin or token type ${coin} not supported or not compiled. Please be sure that you are using the latest version of BitGoJS.`
+      `Coin or token type ${coin} not supported or not compiled. Please be sure that you are using the latest version of BitGoJS. If using @bitgo/sdk-api, please confirm you have registered your coin.`
```
bitgoAaron, line 23: `please confirm you have registered ${coin} first.`

modules/sdk-api/src/bitgoAPI.ts
mmcshinsky-bitgo: Yeah, that works and compiles correctly.

modules/account-lib/package.json
```diff
@@ -79,7 +79,7 @@
     "near-api-js": "^0.44.2",
     "noble-bls12-381": "0.7.2",
     "paillier-bigint": "3.3.0",
-    "protobufjs": "^6.8.9",
+    "protobufjs": "^6.11.3",
```
bitgoAaron, line 82: why the changes to dependencies here? Don't see any changes to code in account-lib

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -66,7 +66,7 @@ if (!(process as any)?.browser) {
 
 const patchedRequestMethods = ['get', 'post', 'put', 'del', 'patch'] as const;
 
-export abstract class BitGoAPI implements BitGoBase {
+export class AbstractBitGoAPI implements BitGoBase {
```
bitgoAaron, line 69: I think i may have been a bit unclear, I think we should just make this class `BitGoAPI` and take all the implementations you've added in `./bitgo.ts` and do them here. There shouldn't be a reason to have both an `AbstractBitGoAPI` and `BitGoAPI` class, is there?

bitgoAaron (changes requested):
let's combine `AbstractBitGoAPI` and `BitGoAPI` into a single class & file.

modules/account-lib/package.json
mmcshinsky-bitgo: This change is already merged in master with other branches for a change. I'll remove this.

modules/sdk-api/src/bitgoAPI.ts
mmcshinsky-bitgo: Makes sense, I agree with implementing it as one. I'll make the change right now.

bitgoAaron (approved):
LGTM

PR #2342, opened 2022-06-06
fix(core): sol send token

jonathancn-bitgo:
Mostly adjustment to support solana token transfer
BG-49568


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1756,37 +1756,39 @@ export class Wallet implements IWallet {
    * @returns {*}
    */
   async send(params: SendOptions = {}): Promise<any> {
-    common.validateParams(params, ['address'], ['message', 'data']);
-
-    if (_.isUndefined(params.amount)) {
-      throw new Error('missing required parameter amount');
-    }
+    const coin = this.baseCoin;
+    if (params.type == 'transfer') {
```
jonathancn-bitgo, line 1760: check if address and amount exist just for `transfer`

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2280,6 +2282,7 @@ export class Wallet implements IWallet {
         });
         break;
       case 'enabletoken':
+        params.recipients = [];
```
jonathancn-bitgo, line 2285: recipients must be empty to do `enabletoken`

modules/bitgo/src/v2/coins/sol.ts
```diff
@@ -160,11 +165,34 @@ export class Sol extends BaseCoin {
...
       for (const recipients of txParams.recipients) {
-        totalAmount = totalAmount.plus(recipients.amount);
+        // totalAmount based on each token
+        if (recipients.tokenName != undefined) {
+          if (!(recipients.tokenName in totalAmount)) {
+            totalAmount[recipients.tokenName] = new BigNumber(0);
+          }
+          totalAmount[recipients.tokenName] = totalAmount[recipients.tokenName].plus(recipients.amount);
+        } else {
+          totalAmount['tsol'] = totalAmount['tsol'].plus(recipients.amount);
```
joey-bai, line 177: can't hard code `'tsol'` because it will be different for production, which uses `'sol'`.

use `this.getChain()` instead.

modules/bitgo/src/v2/coins/sol.ts
```diff
@@ -160,11 +165,34 @@ export class Sol extends BaseCoin {
...
+        }
+      }
+
+      // total output amount from explainedTx
+      const explainedTxTotal = {};
+      explainedTxTotal['tsol'] = new BigNumber(explainedTx.outputAmount);
```
joey-bai, line 183: same, use `this.getChain()`

modules/bitgo/src/v2/coins/sol.ts
```diff
@@ -145,9 +145,14 @@ export class Sol extends BaseCoin {
...
     // users do not input recipients for consolidation requests as they are generated by the server
     if (txParams.recipients !== undefined) {
-      const filteredRecipients = txParams.recipients?.map((recipient) =>
-        _.pick(recipient, ['address', 'amount', 'tokenName'])
-      );
+      const filteredRecipients = txParams.recipients?.map((recipient) => {
+        const filtered = _.pick(recipient, ['address', 'amount']);
+        if (recipient['tokenName'] != undefined) {
+          filtered['tokenName'] = recipient['tokenName'];
+        }
```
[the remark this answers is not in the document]
joey-bai: if its to deal with one map having undefined values and the other doesn't, we can compare `JSON.stringify` results instead

modules/bitgo/src/v2/coins/sol.ts
```diff
@@ -160,11 +165,34 @@ export class Sol extends BaseCoin {
...
-        totalAmount = totalAmount.plus(recipients.amount);
+        // totalAmount based on each token
+        if (recipients.tokenName != undefined) {
+          if (!(recipients.tokenName in totalAmount)) {
+            totalAmount[recipients.tokenName] = new BigNumber(0);
+          }
+          totalAmount[recipients.tokenName] = totalAmount[recipients.tokenName].plus(recipients.amount);
+        } else {
+          totalAmount['tsol'] = totalAmount['tsol'].plus(recipients.amount);
+        }
```
joey-bai, line 178: can make this more concise and reduce amount of if-else checks.
```
const assetName = recipients.tokenName || this.getChain();
const amount = totalAmount[assetName] || new BigNumber(0);
totalAmount[assetName] = amount.plus(recipients.amount);
```

modules/bitgo/src/v2/coins/sol.ts
```diff
@@ -160,11 +165,34 @@ export class Sol extends BaseCoin {
...
+      }
+
+      // total output amount from explainedTx
+      const explainedTxTotal = {};
```
joey-bai, line 182: add type: `const explainedTxTotal: Record<string, BigNumber>`

modules/bitgo/src/v2/coins/sol.ts
joey-bai: can also leverage this pattern for `explainedTx` below

modules/sdk-core/src/bitgo/wallet/wallet.ts
joey-bai: This seems like a pretty risky change since other types of transactions may have address amount, like staking.

Maybe we should support zero amounts for create token account transactions

modules/bitgo/src/v2/coins/sol.ts
```diff
@@ -123,7 +123,7 @@ export class Sol extends BaseCoin {
   }
 
   async verifyTransaction(params: SolVerifyTransactionOptions): Promise<any> {
```
joey-bai, line 125: please update unit tests that cover token transfer + creation transactions.
see `modules/bitgo/test/v2/unit/coins/sol.ts`

jonathancn-bitgo: Updated the test, will cover the `verifyTransaction` for send token once the sendMany fixed, so the test can support multiple token as well (kinda line expensive to cover one by one since need to specify new `TxParam` and `TxPrebuild`)

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1756,18 +1757,21 @@ export class Wallet implements IWallet {
    * @returns {*}
    */
   async send(params: SendOptions = {}): Promise<any> {
+    const coin = this.baseCoin;
     common.validateParams(params, ['address'], ['message', 'data']);
+    if (params.type != 'enableToken') {
```
jonathancn-bitgo, line 1762: only skip params check for enableToken now

modules/sdk-core/src/bitgo/wallet/wallet.ts
joey-bai: i don't think we should force it - this is coin agnostic. instead, we should leave it so it is `params.recipients`.

However, i do see the problem you're running into...
in the `send` function, we always construct 1 recipient:
```
    const recipients: SendManyOptions['recipients'] = [
      {
        address: params.address,
        amount: params.amount,
        tokenName: params.tokenName,
      },
    ];
```

in this case, maybe we filter out recipients that have address === undefined?

e.g.
```
const recipients = params.recipients.filter((recipient) => recipient.address !== undefined))
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1757,6 +1758,11 @@ export class Wallet implements IWallet {
    */
   async send(params: SendOptions = {}): Promise<any> {
     common.validateParams(params, ['address'], ['message', 'data']);
+    // Note: address and amount not required for enable token
+    if (params.type == 'enabletoken') {
+      params.address = params.address || 'not required';
+      params.amount = params.amount || 1;
+    }
```
joey-bai, line 1765: sorry to flip flop, but lets switch back and NOT set default values for `address` and `amount`.

```
if (params.type !== 'enabletoken') {
   // do validation
}
```

jonathancn-bitgo: Ohh okk suree, yes default value + set to empty is kinda ugly in terms of design so yeahh will change to that

joey-bai (approved):

modules/bitgo/src/v2/coins/sol.ts
```diff
@@ -161,10 +162,23 @@ export class Sol extends BaseCoin {
...
+      const explainedTxTotal: Record<string, BigNumber> = {};
+
+      for (const output of explainedTx.outputs) {
+        // total output amount based on each token
+        const assetName = output.tokenName || this.getChain();
+        const amount = explainedTxTotal[assetName] || new BigNumber(0);
+        explainedTxTotal[assetName] = amount.plus(output.amount);
+      }
+
+      if (JSON.stringify(explainedTxTotal) !== JSON.stringify(totalAmount)) {
```
bitgoAaron, line 181: same here

modules/bitgo/src/v2/coins/sol.ts
```diff
@@ -283,10 +297,13 @@ export class Sol extends BaseCoin {
     ];
 
     const outputs: TransactionOutput[] = solTransaction.outputs.map((output) => {
-      return {
-        address: output.address,
-        amount: output.amount,
-      };
+      return JSON.parse(
```
bitgoAaron, line 300: why?

modules/bitgo/src/v2/coins/sol.ts
```diff
@@ -150,7 +151,7 @@ export class Sol extends BaseCoin {
       );
       const filteredOutputs = explainedTx.outputs.map((output) => _.pick(output, ['address', 'amount', 'tokenName']));
 
-      if (!_.isEqual(filteredOutputs, filteredRecipients)) {
+      if (!_.isEqual(JSON.stringify(filteredOutputs), JSON.stringify(filteredRecipients))) {
```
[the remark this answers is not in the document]
jonathancn-bitgo: No, I just want to do check deep equality. Is it using `should.deepEqual`?

modules/bitgo/src/v2/coins/sol.ts
jonathancn-bitgo: this one is because tokenName can be undefined so I want a way to not include the tokenName field when it's undefined.

bitgoAaron: Object de-referencing is your friend.

```
const outputs: TransactionOutput[] = solTransaction.outputs.map(( { address, amount, tokenName }) => ({ address, amount, tokenName });

Comment (jonathancn-bitgo) [reply modules/bitgo/src/v2/coins/sol.ts:300]: Just tested and for this one, `tokenName` in `solTransaction.outputs` can exist with value undefined, so when its de-referenced, it stays while keeping `tokenName` undefined instead of removing it
(it cause failing in unit-test for native coin)

File: modules/bitgo/src/v2/coins/sol.ts (hunk 11 above)
Comment (jonathancn-bitgo) [reply modules/bitgo/src/v2/coins/sol.ts:181]: Same issue with above, because of `explainTransaction`, the output can have `tokenName` field undefined where the cause is from bitgo-microservice somewhere during prebuildTransaction

File: modules/bitgo/src/v2/coins/sol.ts (hunk 12 above)
Comment (bitgoAaron) [reply modules/bitgo/src/v2/coins/sol.ts:300]: ok. It could be that the unit test can just be updated, but if you really dont' want to return an object with `tokenName: undefined` then you probably just want to conditionally assign in a separate step.

```
solTransaction.outputs.map(( { address, amount, tokenName }) => {
  const output = { address, amount };
  if (tokenName) {
    output.tokenName = tokenName;
  }
  return output;
});

jonathancn-bitgo: Ohh okk, I'll just keep tokenName undefined to not showed then

modules/bitgo/src/v2/coins/sol.ts
bitgoAaron: is this still the case?

jonathancn-bitgo: yes, since below is updating `parseTransaction` while here is using `explainTransaction`
can't use `parseTransaction` since it only return inputs and outputs value while `verifyTransaction` still need to check Nonce etc

bitgoAaron: I'm not seeing the tests fail here on this one. Seems to pass when I use 
```
if (!_.isEqual(explainedTxTotal, totalAmount))
```

We should avoid unnecessarily serializing objects just to check for equality, there's usually a better way to accomplish that.

jonathancn-bitgo: Ohh got it, my bad

joey-bai (approved):

PR #2351, opened 2022-06-08
refactor(sdk-core): tss refactor moving eddsa into a sub folder

olibeyene:
This pr is moving eddsa into a sub folder of tss so we can ecdsa to tss. 

Ticket: BG-49287

### DIFF
--- a/modules/sdk-core/src/bitgo/tss/eddsa/index.ts
+++ b/modules/sdk-core/src/bitgo/tss/eddsa/index.ts
@@ -1,3 +1,9 @@
+import * as EDDSAMethods from './eddsa';
+export * as EDDSAMethodTypes from './types';
+
+export default EDDSAMethods;
+
+// exporting this types for backward compatibility.
 export {
   createCombinedKey,
   createUserSignShare,
@@ -9,3 +15,5 @@ export {
   sendSignatureShare,
   encryptYShare,
 } from './eddsa';
+
+export { EncryptedYShare, DecryptableYShare, CombinedKey, SigningMaterial } from './types';
--- a/modules/sdk-core/src/bitgo/tss/index.ts
+++ b/modules/sdk-core/src/bitgo/tss/index.ts
@@ -1,5 +1,9 @@
-export { EncryptedYShare, DecryptableYShare, CombinedKey, SigningMaterial, ShareKeyPosition } from './types';
+import EDDSAMethods, { EDDSAMethodTypes } from './eddsa';
 
+export { EDDSAMethods, EDDSAMethodTypes };
+export { ShareKeyPosition } from './types';
+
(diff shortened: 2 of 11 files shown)


wesgraham (approved):
looks reasonable to me - nice work

modules/sdk-core/src/bitgo/tss/eddsa/index.ts
```diff
@@ -9,3 +9,5 @@ export {
   sendSignatureShare,
   encryptYShare,
 } from './eddsa';
+
+export { EncryptedYShare, DecryptableYShare, CombinedKey, SigningMaterial } from './types';
```
sachushajiabraham, line 13: We can do something like so that we can properly differentiate between EDDSA and ECDSA things
```suggestion
export * as EDDSAMethodTypes from './types';
```

modules/sdk-core/src/bitgo/tss/index.ts
```diff
@@ -10,4 +12,8 @@ export {
   getTxRequest,
   sendSignatureShare,
   encryptYShare,
+  EncryptedYShare,
+  DecryptableYShare,
+  CombinedKey,
+  SigningMaterial,
 } from './eddsa';
```
sachushajiabraham, line 19: Let export types like 

```
// exporting all eddsa types and making eddsa as default for backward compatibility
export * from './eddsa/types';
export {EDDSAMethodTypes} from './eddsa'
```

modules/sdk-core/src/bitgo/utils/tss/index.ts
```diff
@@ -0,0 +1,10 @@
+export * as EDDSAUtil from './eddsa';
+
+export {
+  TssUtils,
```
sachushajiabraham, line 4: Can do similar to the above suggestion for util types also

modules/sdk-core/src/bitgo/utils/tss/eddsa/index.ts
```diff
@@ -0,0 +1,9 @@
+export { TssUtils } from './eddsa';
+
+export {
+  ITssUtils,
+  PrebuildTransactionWithIntentOptions,
+  SignatureShareRecord,
```
sachushajiabraham, line 6: Can do similar to the above suggestion for util types also

tylerlevine (commented):
More description of the change in the commit message please. I have no context on this change to be able to review it, and future devs reading the commit message for it won't be able to understand anything about what's going on here. 

Specifically, please address:
* What is being changed?
* Why is it being changed?
* What is the desired end state?
* What were some of the tradeoffs or implementation decisions that were made and why?

tylerlevine (approved):

PR #2358, opened 2022-06-09
chore(bitgo-express): add unhandledRejection hook to bitgo express

ducnguyen-bitgo:
- Prevented express from crashing on unhandled promise rejections.
- This fix should be removed once BG-49996 is implemented.

Ticket: BG-48910

Starting from version 15, node process would terminate if there’s any unhandled promised rejection without the unhandledRejection hook. More info could be found [here](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V15.md#throw-on-unhandled-rejections---33021) and [here](https://nodejs.org/docs/latest-v16.x/api/cli.html#--unhandled-rejectionsmode).

### DIFF
--- a/modules/express/bin/bitgo-express
+++ b/modules/express/bin/bitgo-express
@@ -1,5 +1,13 @@
 #!/usr/bin/env node
 
+// TODO: Remove this unhandledRejection hook once BG-49996 is implemented.
+process.on('unhandledRejection', (reason, promise) => {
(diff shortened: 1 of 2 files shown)


mmcshinsky-bitgo:
Lint commit error :(

ducnguyen-bitgo:
@mmcshinsky-bitgo I just fixed the lint commit error :)

PR #2359, opened 2022-06-09
feat(account-lib): add support for chaincode for key derivation in ecdsa

sachushajiabraham:
this change add support for chaincode which is needed when we implement
key derivation for ecdsa

TICKET: BG-49979

### DIFF
--- a/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
@@ -1,5 +1,7 @@
 import * as paillierBigint from 'paillier-bigint';
+import { randomBytes } from 'crypto';
 import { bigIntToHex } from '../../../util/crypto';
+import { bigIntFromBufferBE, bigIntToBufferBE } from '../../util';
 import { Secp256k1Curve } from '../../curves';
 import Shamir from '../../shamir';
 import { NShare, PShare, KeyShare, KeyCombined } from './types';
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -67,6 +72,10 @@ export default class Ecdsa {
     // Add secret shares
     const x = allShares.map((participant) => BigInt(participant['u'])).reduce(Ecdsa.curve.scalarAdd);
 
+    // Chaincode will be used in future when we add support for key derivation for ecdsa
```
mmcshinsky-bitgo, line 75: Do we have a ticket TODO to link to this?

sachushajiabraham: Nope, Key derivation for ecdsa is not in the pipeline as of now, but this is needed to ensure backward compatiblility when it's implemented

olibeyene (approved):

wesgraham (approved):

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -67,6 +72,10 @@ export default class Ecdsa {
...
+    const chaincodes = [pShare, ...nShares].map(({ chaincode }) => bigIntFromBufferBE(Buffer.from(chaincode, 'hex')));
+    const chaincode = chaincodes.reduce((acc, chaincode) => Ecdsa.curve.scalarReduce(acc + chaincode));
```
johnoliverdriscoll, line 77: The chaincode isn't an element in the secp256k1 field. The common chaincode should be calculated as `(acc + chaincode) % (1n<<256n)`

PR #2362, opened 2022-06-09
feat: added explainTransaction and unit tests for dot

VinBoy2:
Implementing explain transaction for polkadot OVC signing.

### DIFF
--- a/modules/bitgo/src/v2/types.ts
+++ b/modules/bitgo/src/v2/types.ts
@@ -1,7 +1,79 @@
+import BigNumber from 'bignumber.js';
+
 /**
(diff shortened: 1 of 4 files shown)


modules/bitgo/src/v2/types.ts
```diff
@@ -1,7 +1,79 @@
...
 /**
  * Basic coin recipient information
  */
 export interface Recipient {
   address: string;
   amount: string;
 }
+/*
+ * Types for Polkadot
+ */
```
lcovar, line 12:
```suggestion
```

These types are not specific to polkadot so we can remove this comment.

PR #2364, opened 2022-06-09
feat: added explainTransaction and unit tests for dot

VinBoy2:
TICKET: BG-49946

### DIFF
--- a/modules/bitgo/src/v2/types.ts
+++ b/modules/bitgo/src/v2/types.ts
@@ -1,7 +1,73 @@
+import BigNumber from 'bignumber.js';
+import { ITransactionExplanation as BaseTransactionExplanation, ITransactionRecipient } from '@bitgo/sdk-core';
+
 /**
  * Basic coin recipient information
  */
 export interface Recipient {
   address: string;
   amount: string;
+}
+export interface Output extends ITransactionRecipient {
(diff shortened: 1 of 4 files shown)


VinBoy2:
Implementing explain transaction for polkadot OVC signing.

modules/bitgo/src/v2/types.ts
```diff
@@ -1,7 +1,79 @@
...
 /**
  * Basic coin recipient information
  */
 export interface Recipient {
   address: string;
   amount: string;
 }
+/*
+ * Types for Polkadot
+ */
```
lcovar, line 12:
```suggestion
```

Not specific for polkadot so we can remove this comment.

modules/bitgo/src/v2/types.ts
```diff
@@ -1,7 +1,76 @@
...
+export type ExplanationResult = {
+  sequenceId: number;
+  feeInfo?: string;
+  txid: string;
+  type?: string;
+  outputs: Output[];
+  blockNumber: number | unknown;
+  changeOutputs: [];
+  changeAmount: string;
+}
```
joey-bai, line 76: one already exists, we should use that instead. see `ITransactionExplanation`

PR #2382, opened 2022-06-13
Ducnguyen/bg 49282 refactor bitgotest to its own package

bitgoAaron:


modules/sdk-test/src/bitgo/lib/test_bitgo.ts
```diff
@@ -0,0 +1,624 @@
+//
```
bitgoAaron, line 1: we should rename this file to something different since it no longer exports TestBitgo.

modules/sdk-test/src/bitgo/lib/test_bitgo.ts
```diff
@@ -0,0 +1,624 @@
...
+      .reply(200, { status: '1', message: 'OK', result: '2400' });
+
+    return wallet;
+  };
+
+  //
+  // registerToken
+  // Register tokens
+  //
+  BitGo.prototype.registerToken = function (tokenFamily: string, createTokenConstructor: CoinConstructor) {
```
bitgoAaron, line 577: this is not necessary anymore now that we're decorating classes & each test is passing a starting reference. When we move tests over, they should use the registration methods from CoinFactory.

PR #2383, opened 2022-06-13
feat(account-lib): hbar token transfer builder

jonathancn-bitgo:
Major Changes:
1. `tokenTransferBuilder` and unit tests which built similar to `transferBuilder` with additional `tokenName` function to add tokenId using token name.
2. `toJson()` in `transaction.ts` to support deserialization of txBody, used in `tokenTransferBuilder` unit-test
3. Some utilities in `utils.ts` to support token functions
4. Added `tokenName` field in explainTransaction

Ticket: BG-49596, BG-49607


modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /**
+   * Set the token id to be of the transaction
+   * it may take the format `'<shard>.<realm>.<token>'` or `'<token>'`
+   *
+   * @param {string} tokenId the token id for the token
+   * @returns {TransferBuilder} the builder with the new parameter set
+   */
+  tokenId(tokenId: string): this {
```
jonathancn-bitgo, line 147: had some consideration of using tokenName instead, but will be complicated on the `initTransfers`, so I keep tokenId instead.
But since the input of `send` in `sdk-core/wallet.ts` use tokenName, will do the mapping of tokenName to tokenId inside bitgo-microservice

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,169 @@
...
+    if (!isValidAddress(tokenId)) {
```
jonathancn-bitgo, line 148: gonna add `isValidToken` instead

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -85,14 +86,28 @@ export class Transaction extends BaseTransaction {
    *
    * @returns {[string, string]} first element is the recipient, second element is the amount
    */
-  private getTransferData(): [string, string] {
+  private getTransferData(): [string, string, string] {
```
kevinwan-bitgo, line 89: You should update the documentation for changed functions.

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /**
+   * Initialize the transfer specific data, getting the recipient account
+   * represented by the element with a positive amount on the transfer element.
+   * The negative amount represents the source account so it's ignored.
+   *
+   * @param {proto.IAccountAmount[]} transfers array of objects which contains accountID and transferred amount
+   */
+  protected initTransfers(tokenTransfers: proto.ITokenTransferList[]): void {
```
kevinwan-bitgo, line 87: Also here as well

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
joey-bai: SGTM

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /**
+   * Set the token id to be of the transaction using token name
+   * it may take the format `'<shard>.<realm>.<token>'` or `'<token>'`
+   *
+   * @param {string} tokenName the token id for the token
+   * @returns {TransferBuilder} the builder with the new parameter set
+   */
+  tokenName(tokenName: string): this {
```
joey-bai, line 169: do we want to keep this then?

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -85,14 +86,28 @@ export class Transaction extends BaseTransaction {
...
     let transferData;
-    this._txBody.cryptoTransfer!.transfers!.accountAmounts!.forEach((transfer) => {
-      const amount = Long.fromValue(transfer.amount!);
-      if (amount.isPositive()) {
-        transferData = [stringifyAccountId(transfer.accountID!), amount.toString()];
-      }
-    });
+    const tokenTransfers = this._txBody.cryptoTransfer!.tokenTransfers!;
```
joey-bai, line 91: i know we had this in place, but let's try to avoid force unwrapping optional properties.

alternative is to explicitly check for existence or provide default values.
e.g.
```
const tokenTransfers = this._txBody.cryptoTransfer?.tokenTransfers || [];
```

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -85,14 +86,28 @@ export class Transaction extends BaseTransaction {
...
+        if (amount.isPositive()) {
+          transferData = [
+            stringifyAccountId(transfer.accountID!),
+            amount.toString(),
+            stringifyTokenId(tokenTransfers[0].token!),
+          ];
+        }
+      });
+    } else {
+      this._txBody.cryptoTransfer!.transfers!.accountAmounts!.forEach((transfer) => {
```
joey-bai, line 104: same

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -155,15 +170,15 @@ export class Transaction extends BaseTransaction {
         {
           address: txJson.to,
           value: txJson.amount,
-          coin: this._coinConfig.name,
+          coin: txJson.tokenId || this._coinConfig.name,
```
joey-bai, line 173: should use display name here instead of tokenId if possible (use a map from tokenId -> tokenName).

joey-bai: `outputs` is used in various places and we expect this to be a human readable value

modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
```diff
@@ -21,12 +22,20 @@ export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
...
+    return this.initializeBuilder(tx, new TokenTransferBuilder(this._coinConfig));
+  }
+
   /** @inheritDoc */
   from(raw: Uint8Array | string): TransactionBuilder {
     this.validateRawTransaction(raw);
     const tx = this.parseTransaction(raw);
     switch (tx.txBody.data) {
       case 'cryptoTransfer':
+        if (isTokenTransfer(tx.txBody.cryptoTransfer!)) {
```
joey-bai, line 36: no `!`

modules/account-lib/src/coin/hbar/utils.ts
```diff
@@ -51,6 +53,26 @@ export function isValidTransactionId(txId: string): boolean {
...
+ *
+ * In any form, `shard` and `realm` are assumed to be 0 if not provided.
+ *
+ * @param {string} tokenId - the address to be validated
+ * @returns {boolean} - the validation result
+ */
+export function isValidToken(tokenId: string): boolean {
+  if (_.isEmpty(tokenId) || !tokenId.match(/^[0-9]+(?:(?=\.)(\.[0-9]+){2}|(?!\.))$/)) {
+    return false;
+  }
```
joey-bai, line 67: can we also check that this is maps to a token we support?

modules/account-lib/src/coin/hbar/utils.ts
```diff
@@ -290,3 +322,55 @@ export function isValidAddressWithPaymentId(address: string): boolean {
...
+    if (coin instanceof HederaToken) {
+      return coin.network.type === network.type && coin.tokenId === tokenId;
+    }
+    return false;
+  });
+  const tokensArray = tokens.map((token) => token);
+  if (tokensArray.length >= 1) {
+    // there should never be two tokens with the same contract address, so we assert that here
+    assert(tokensArray.length === 1);
+    return tokensArray[0];
```
joey-bai, line 354: this will never return because the `assert` will fail and throw an error.

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,191 @@
...
+import {
+  isValidAddress,
+  isValidAmount,
+  isValidToken,
+  stringifyAccountId,
+  stringifyTokenId,
+  getHbarTokenFromTokenName,
+} from './utils';
+
+export class TokenTransferBuilder extends TransactionBuilder {
```
joey-bai, line 23: OK to start, but you may want to subclass `TransferBuilder` if you find a lot of similarities.

CC @noel-bitgo

joey-bai: Sorry, wrong Noel.

CC @tunneln

jonathancn-bitgo: I considered using TransferBuilder before, but I need to get `to` and `amount` fields where they are `private` in `TransferBuilder`. Do u think it is fine to change it to `protected` for example?

modules/account-lib/src/coin/hbar/transaction.ts
jonathancn-bitgo: Yeah they used `!` before, I can change to that.

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
jonathancn-bitgo: Ohh yeah I think we can remove this, since we can use the `getTokenFromName` from utils to convert name to id

modules/account-lib/src/coin/hbar/transaction.ts
jonathancn-bitgo: For the amount! and token! on line 94 and 99, is there any way to not using ! since we can't set default value for these

modules/bitgo/src/v2/coins/hbar.ts
```diff
@@ -249,6 +249,7 @@ export class Hbar extends BaseCoin {
         amount: txJson.amount.toString(),
         address: txJson.to,
         memo,
+        tokenId: txJson.tokenId,
```
jonathancn-bitgo, line 252: this is inside `explainTransaction`, do I need tokenName instead of tokenId as well?

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
jonathancn-bitgo: refactored and saved 60 lines, if adding a boolean `isToken` to `initBuilder` and `buildImplementation` is fine

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
kevinwan-bitgo: I think we should probably use tokenName here since all of our logic for converting between tokenName and tokenId is in BitgoJS. Mapping in bitgo-microservice just introduces more dependencies between the two repos. We can even do the mapping in this function.

jonathancn-bitgo: The dependancy issue is not really a problem since both map function and the tokenBuilder are in `account-lib`

But yeah tokenName input will be better

modules/bitgo/test/v2/unit/coins/hbar.ts
```diff
@@ -49,6 +49,29 @@ describe('Hedera Hashgraph:', function () {
     explain.changeAmount.should.equal('0');
   });
 
+  it('should explain a token transfer transaction', async function () {
+    const tokenTransferParam = {
```
jonathancn-bitgo, line 53: seems a bit forcing, but that's the field needed in explainTransaction.
the former is using 1 line stringified json object which is not yet producable

modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
```diff
@@ -21,12 +22,20 @@ export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
...
+    return this.initializeBuilder(tx, new TokenTransferBuilder(this._coinConfig));
+  }
+
   /** @inheritDoc */
   from(raw: Uint8Array | string): TransactionBuilder {
     this.validateRawTransaction(raw);
     const tx = this.parseTransaction(raw);
     switch (tx.txBody.data) {
       case 'cryptoTransfer':
+        if (tx.txBody.cryptoTransfer && isTokenTransfer(tx.txBody.cryptoTransfer)) {
```
tunneln, line 36: Shouldn't the txbody.data be `tokenTransfers` and not `CryptoTransfer`?
https://docs.hedera.com/guides/docs/hedera-api/miscellaneous/transactionbody

tunneln (changes requested):

modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
tunneln, line 36: Shouldn't TokenTransferBuilder be under case `tokenTransfer` and not `cryptoTransfer`.
I believe we need to update the hbar protobuf in account-lib to support these cases.

tunneln (dismissed):

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Initialize the transfer specific data, getting the recipient account
+   * represented by the element with a positive amount on the transfer element.
+   * The negative amount represents the source account so it's ignored.
+   *
+   * @param {proto.IAccountAmount[]} transfers array of objects which contains accountID and transferred amount
+   */
+  protected initTokenTransfers(tokenTransfers: proto.ITokenTransferList[]): void {
+    const tokenTransfer = tokenTransfers[0];
```
joey-bai, line 59: we only consume the first element in `tokenTransfers`, but input type is an array. is this intentional?

what happens if there are multiple elements in the array?

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -83,16 +84,33 @@ export class Transaction extends BaseTransaction {
    * Get the recipient account and the amount
    * transferred on this transaction
    *
-   * @returns {[string, string]} first element is the recipient, second element is the amount
+   * @returns {[string, string, string]} first element is the recipient, second element is the amount, third recipient is token name
    */
-  private getTransferData(): [string, string] {
+  private getTransferData(): [string, string, string] {
```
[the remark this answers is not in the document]
jonathancn-bitgo: Ohh got it, yeah it should be optional

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,104 @@
+import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import * as Long from 'long';
+import { TokenId } from '@hashgraph/sdk';
+import { proto } from '../../../resources/hbar/protobuf/hedera';
+import { BuildTransactionError, InvalidParameterValueError } from '@bitgo/sdk-core';
+import { TransferBuilder } from './transferBuilder';
+import { Transaction } from './transaction';
+import { getHbarTokenFromTokenName, isValidToken, stringifyAccountId, stringifyTokenId } from './utils';
+
+export class TokenTransferBuilder extends TransferBuilder {
```
tunneln, line 10: What's the reason for extending `TransferBuilder` here instead of `TransactionBuilder`? All of the builders thus far have extended Transaction.

tunneln (changes requested):

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
tunneln, line 10: After our conversation, I still think it'd be in bad form to have `TokenTransferBuilder` extend `TransferBuilder`. I think coupling it too tightly to the TransferBuilder may just be a headache later on.
Maybe instead of `A` we can have `B`
```
A. TransactionBuilder -> TransferBuilder -> TokenTransferBuilder

                                            CoinTransferBuilder
B. TransactionBuilder -> TransferBuilder <
                                            TokenTransferBuilder
```

btw this change doesn't have to be apart of this PR, I just wanted to share my thoughts

tunneln (changes requested):

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
jonathancn-bitgo: I'll think about it, since my current concern is CoinTransferBuilder will require a change in microservice as well due to adding coinTransferBuilder. 
And let say we want to migrate to using diff class for coin and token, then I kinda want to change to support multi-recipients as well as the flow to update into new class will be in both account-lib and ms for the native coin

and if so, then it will be just:

                                       CoinTransferBuilder
TransactionBuilder <
                                       TokenTransferBuilder

which is my plan if tokenTransfer can be finished and there's still time to update

jonathancn-bitgo: Actually, I can extend TransactionBuilder with different format from TransferBuilder (like using `send` instead of `to` and `amount`) where this can support multi recipients but currently just used to handle 1 recipient will be fine as well.

modules/account-lib/src/coin/hbar/transactionBuilderFactory.ts
```diff
@@ -21,12 +22,20 @@ export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
     return this.initializeBuilder(tx, new TransferBuilder(this._coinConfig));
   }
 
+  /** @inheritDoc */
```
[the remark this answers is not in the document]
tunneln: Just so it was understood, be sure to add a method explanation comment above `getTokenTransferBuilder`

modules/account-lib/src/coin/hbar/tokenTransferBuilder.ts
```diff
@@ -0,0 +1,104 @@
...
+      tokenNum: tokenData.num,
+      realmNum: tokenData.realm,
+      shardNum: tokenData.shard,
+    });
+  }
+
+  /** @inheritdoc */
+  initBuilder(tx: Transaction): void {
+    super.initBuilder(tx);
+    const transferData = tx.txBody.cryptoTransfer;
```
tunneln, line 45: Be sure to set the transaction type here, i.e., `this.transaction.setTransactionType(TransactionType.Send)`

PR #2389, opened 2022-06-15
fix(account-lib): fix chaincode to use correct modulo

sachushaji:
fix modulus for chaincode to be 2^256

TICKET: BG-50457


johnoliverdriscoll (dismissed):

olibeyene (approved):

tylerlevine (changes requested):
test?

tylerlevine (changes requested):
test case passes without the implementation change. Are you sure this actually tests the new behavior?

sachushaji:
@tylerlevine previous modulus was less than the current modulo, so it will pass for the earlier implementation

sachushaji:
@tylerlevine Actually,the previous test case for this was only cheking with a simple NULL. This test case correctly checks the value for chaincode. So it should be okay

tylerlevine (approved):

PR #2391, opened 2022-06-15
feat(sdk-coin-algo): refactor Algorand to its own module

bitgoAaron:
* refactors algo specific code from account-lib & bitgo into dedicated module `@bitgo/sdk-coin-algo`
* re-exports code to be backwards compatible
* provides simpler mechanism for registering tokens with CoinFactory


twtaylor:
image

modules/sdk-coin-algo/test/unit/lib/transactionBuilder/assetTransferBuilder.ts
```diff
@@ -2,9 +2,9 @@ import { coins } from '@bitgo/statics';
 import assert from 'assert';
 import should from 'should';
 import { test } from 'mocha';
```
ducnguyen-bitgo, line 4: We might need to add mocha into our devDependencies I think.

ducnguyen-bitgo:
Just one minor change in devDependencies for adding `mocha` and we should be good to go.

modules/sdk-coin-algo/test/unit/lib/transactionBuilder/assetTransferBuilder.ts
bitgoAaron: actually `mocha` needs to only be in the root package. We ran into issues with it being included in multiple places when refactoring TestBitgo.

ducnguyen-bitgo: I see. Then everything looks good to me! Approving it right now.

ducnguyen-bitgo (approved):

PR #2404, opened 2022-06-21
fix(account-lib): change etheremjs-utils-old alias

bitgoAaron:
- renames the alias to something that doesn't cause a yarn audit failure

Ticket: BG-00000


sachushaji (approved):
LGTM!

PR #2406, opened 2022-06-21
chore(sdk-api): make web entry work

bitgoAaron:
- hoist webpack dependencies to root package
- move base webpack config to root folder
- define correct entry points for sdk-api package.json
- fix lint errors in unspents
- stop web-packing account-lib

Ticket: BG-50241


modules/statics/package.json
```diff
@@ -22,8 +22,7 @@
     "directory": "modules/statics"
   },
   "devDependencies": {
-    "nyc": "^15.0.0",
-    "should": "^13.2.3"
+    "nyc": "^15.0.0"
```
tylerlevine, line 25: nit: could remove nyc here

modules/sdk-test/package.json
```diff
@@ -33,7 +33,6 @@
     "@bitgo/statics": "^6.17.0",
     "bignumber.js": "^8.0.1",
     "lodash": "^4.17.14",
-    "should": "^13.1.3",
     "should-http": "^0.1.1"
```
tylerlevine, line 36: nit: do we need `should-http` if we don't have `should`?

tylerlevine (approved):
small nits. Awesome work on this 🙏

modules/statics/package.json
bitgoAaron: There's quite a few unused packages throughout the repo. I'd like to address all of them at once in this ticket:
https://bitgoinc.atlassian.net/browse/BG-51022

modules/sdk-test/package.json
bitgoAaron: `should` is provided by the root package.

PR #2409, opened 2022-06-22
chore(express): make bitgo express first-class citizen in monorepo

ducnguyen-bitgo:
- Make bitgo-express a first-class citizen of the monorepo.
- Update Dockerfile to build a docker image with yarn.
- Update bitgo-express README.

Ticket: BG-50244


Dockerfile
```diff
@@ -1,14 +1,16 @@
...
-COPY --chown=node:node . /tmp/bitgo/
-WORKDIR /tmp/bitgo/modules/express
-USER node
+WORKDIR /tmp/bitgo
+COPY . .
 RUN \
+    # clean up unnecessary local node_modules and dist
+    rm -rf **/node_modules **/dist && \ 
     # install with dev deps so we can run the prepare script
     yarn install --frozen-lockfile && \
```
mmcshinsky-bitgo, line 10: With it not being dependent on other packages internally, but from npm, do we need two `yarn install`s still?

Dockerfile
```diff
@@ -17,12 +19,149 @@ COPY --from=builder \
...
+COPY --from=builder /tmp/bitgo/modules/express/node_modules /var/bitgo-express/node_modules/
+
+# copy individual directories that contain js executable files, and compiled js
 COPY --from=builder /tmp/bitgo/modules/express/bin /var/bitgo-express/bin/
 COPY --from=builder /tmp/bitgo/modules/express/dist /var/bitgo-express/dist/
-COPY --from=builder /tmp/bitgo/modules/express/src /var/bitgo-express/src/
+
+# copy bitgo and yarn link it
+COPY --from=builder /tmp/bitgo/modules/bitgo/package.json /var/modules/bitgo/
+COPY --from=builder /tmp/bitgo/modules/bitgo/dist /var/modules/bitgo/dist/
```
mmcshinsky-bitgo, line 35: Should we be pulling from npm as published dependencies and not internal packages?

modules/express/README.md
```diff
@@ -77,14 +77,22 @@ Base URI: https://0.0.0.0:4000
 
 ### Building the docker container
 
-If you'd like to build the BitGo Express docker container yourself from the source code, first check out the latest release branch **rel/latest**, then run `docker build` from the project root. Here's the commands:
+If you'd like to build the BitGo Express docker container yourself from the source code, first check out the latest **master** branch, then run `docker build` from the project root. Here's the commands:
```
mmcshinsky-bitgo, line 80: Are we removing `rel/latest` as the core location for companies looking to build out the latest `stable` since `master` isn't a stable build?

modules/express/tsconfig.json
```diff
@@ -6,5 +6,10 @@
     "typeRoots": ["./types", "../../types", "./node_modules/@types", "../../node_modules/@types"]
   },
   "include": ["src/**/*", "package.json"],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules"],
+  "references": [
```
mmcshinsky-bitgo, line 10: Same as questions above

.dockerignore
```diff
@@ -10,11 +10,5 @@ node_modules/
...
 lerna-debug.log
-modules/*/
-!modules/express/*
-modules/express/dist/
-modules/express/test/
-modules/express/node_modules/
-modules/express/.nyc_output/
-modules/express/.gitignore
-modules/express/.npmignore
+modules/**/node_modules
```
mmcshinsky-bitgo, line 13: Would it make sense to move the dockerignore and Dockerfile to the express module?

mmcshinsky-bitgo:
I think I'm confused is all is why the Express module is being more deeply tied into the monorepo instead of being more standalone than it currently even is?

Dockerfile
ducnguyen-bitgo: @mmcshinsky-bitgo the purpose of this ticket is to make it built with other local packages so we need to use yarn. Also, the second yarn install is to reduce the size of the root `node_modules` by getting rid of devDependencies packages after the required packages i.e bitgo, sdk-core, and etc were built.

Dockerfile
ducnguyen-bitgo: Based on this [ticket](https://bitgoinc.atlassian.net/browse/BG-50244)'s description, it needs to use yarn and internal packages if I understand it correctly.

mmcshinsky-bitgo:
> @mmcshinsky-bitgo Please refer to this [ticket](https://bitgoinc.atlassian.net/browse/BG-50244) for more information. If I understand it correctly, the main reason is to be able to build an express docker image without having to publish local packages to NPM.

That's the weird part, because all the packages are already published :/

modules/express/README.md
bitgoAaron: while that is true about the scope of this ticket, we still only want people only using `rel/latest` if they're going to build express themselves.

bitgoAaron:
> That's the weird part, because all the packages are already published :/

requiring Express to only work with published versions of mono-repo packages prevents us from making necessary changes as part of the modularization effort. Think of Express like any other package in the monorepo, except it also needs to build a working version of itself in a docker container based on the current state of the monorepo - not on the current state of what's been publicly published.

.dockerignore
bitgoAaron: we should also still ignore the other files like `.nyc_output`, `.gitignore`, `.npmignore` etc

ducnguyen-bitgo: Okay if we don't wanna copy them over during the build stage, I'll put them back. However, they are not included in the final image.

Dockerfile
```diff
@@ -17,12 +19,149 @@ COPY --from=builder \
...
+RUN cd /var/modules/account-lib && yarn link
+
+# copy blake2b and yarn link it
+COPY --from=builder /tmp/bitgo/modules/blake2b/package.json /var/modules/blake2b/
+COPY --from=builder /tmp/bitgo/modules/blake2b/index.js /var/modules/blake2b/
+COPY --from=builder /tmp/bitgo/modules/blake2b/node_modules /var/modules/blake2b/node_modules/
+RUN cd /var/modules/blake2b && yarn link
+
+# copy blake2b-wasm and yarn link it
+COPY --from=builder /tmp/bitgo/modules/blake2b-wasm/package.json /var/modules/blake2b-wasm/
```
bitgoAaron, line 64: is there anything in blake2b-wasm we _aren't_ copying over?

bitgoAaron (changes requested):
The dockerfile is pretty gnarly, and will become invalidated as soon as we add or rename another package. We need a way to automate that dockerfile so it's always up to date with the state of the repo. I'd suggest adding a script to gather all the commands for `COPY --from=builder` and `yarn link` & write to the Dockerfile as part of the root level package scripts. Anytime someone adds a package & builds the repo it'll update the Dockerfile automatically & get checked in as part of that work. 

additionally, we need to add the unit/integration test folders in express's tsconfig's `"include"` field & clean them up so we can verify these changes have produced a viable build. I did a quick smoke test on the image when I built it locally, but I'd feel a lot more comfortable with these changes if we could get those tests up & running again. 

Fantastic start, we just need to make sure this will scale with the work we have coming down the pike.

Dockerfile
ducnguyen-bitgo: Just `example.html`, `example.js`, and `test.js` aren't copied over.

PR #2410, opened 2022-06-22
fix(core): fix SOL Failed to create any transactions error BG-50572

alia-bitgo:
Original issue was SENDCOINS endpoint throwing a "Failed to create any transactions" error. 
This was resolved, but replaced with error "Tx outputs does not match with expected txParams recipients". 
The check between `filteredOutputs` and `filteredRecipients` was failing in this [file](https://github.com/BitGo/BitGoJS/blob/master/modules/bitgo/src/v2/coins/sol.ts#L154). 
The issue was `tokenName` being explicitly defined in `wallet.ts`, causing `filteredRecipients` to have an explicit `tokenName: undefined` and the comparison failing since `filteredOutputs` did not have this.
Fix was removing the explicit definition of the `tokenName` property. All unit tests are still passing.

TICKET: [BG-50572](https://bitgoinc.atlassian.net/browse/BG-50572?atlOrigin=eyJpIjoiNTYyODkxZTA0YTBlNDI4NmFiNzNlOWZkNDFkNDZkNmMiLCJwIjoiaiJ9)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1781,7 +1781,6 @@ export class Wallet implements IWallet {
       {
         address: params.address,
         amount: params.amount,
-        tokenName: params.tokenName,
```
joey-bai, line 1784: 99% sure this will break tokens.

i think the fix you want here is to conditionally set `tokenName`. maybe something like:
```
if (params.tokenName) {
  recipients[0].tokenName = params.tokenName;
}
```

joey-bai (approved):

PR #2412, opened 2022-06-22
Update coin generation template to match newly created coin implementation structure

mmcshinsky-bitgo:
Update coin generation template to match newly created coin implementation structure

BG-50943


bitgoAaron (approved):
Nice job

PR #2413, opened 2022-06-22
feat(account-lib): add support for ecdsa sigining and verification tss

sachushaji:
this change add support for ecdsa signing and verification for TSS

TICKET: BG-48874

### DIFF
--- a/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
@@ -1,11 +1,37 @@
 import * as paillierBigint from 'paillier-bigint';
-import { randomBytes } from 'crypto';
+import * as secp from '@noble/secp256k1';
+import { randomBytes, createHash } from 'crypto';
 import { bigIntToHex } from '../../../util/crypto';
-import { bigIntFromBufferBE, bigIntToBufferBE } from '../../util';
+import { bigIntFromBufferBE, bigIntToBufferBE, bigIntFromU8ABE, getPaillierPublicKey } from '../../util';
 import { Secp256k1Curve } from '../../curves';
 import Shamir from '../../shamir';
-import { NShare, PShare, KeyShare, KeyCombined } from './types';
+import {
+  NShare,
+  PShare,
+  KeyShare,
+  KeyCombined,
+  BShare,
+  AShare,
+  Signature,
+  SignConvertRT,
+  SignConvert,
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -102,4 +125,240 @@ export default class Ecdsa {
...
+      delete shareParticipant['m'];
+      delete shareToBeSend['alpha'];
+      delete shareToBeSend['mu'];
+    }
+    if (shareToBeSend['k']) {
+      const n = BigInt(shareToBeSend['n']); // Pallier pub from other signer
+      let pk = getPaillierPublicKey(n);
+      const k = BigInt(shareToBeSend['k']);
+
+      const beta0 = Ecdsa.curve.scalarRandom();
```
sachushaji, line 221: beta0 and nu0, the original range was to from 1 to n (Where n is paillier constant for its public key), But the scalar negation for noble only permits the operation within the secp256k1 curve range, so had to stick with scalar random, @johnoliverdriscoll hope its okay

olibeyene (dismissed):
lgtm!

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
johnoliverdriscoll: I've looked over another implementation and it seems to use a beta value that is within secp256k1's curve order, so I'm sure it's fine this way.

johnoliverdriscoll (dismissed):

wesgraham (approved):

bitgoAaron (approved):

modules/sdk-core/src/account-lib/mpc/util.ts
```diff
@@ -1,3 +1,4 @@
+import * as paillierBigint from 'paillier-bigint';
```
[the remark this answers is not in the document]
sachushaji: Good Point, I will have it updated

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
johnoliverdriscoll: Ok I had a review of the white paper and I have found that this value (and n0) needs to be within 1 <= beta0 < n. To negate it on the curve you'll simply have to call curve.scalarReduce on it first.

PR #2416, opened 2022-06-23
feat(sdk-core): tss ecdsa utility to create key chains

olibeyene:
Ticket: BG-49287


wesgraham (dismissed):
excellent work!

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -0,0 +1,254 @@
...
+  async createKeychains(params: {
+    passphrase: string;
+    enterprise?: string | undefined;
+    originalPasscodeEncryptionCode?: string | undefined;
+  }): Promise<KeychainsTriplet> {
+    const MPC = new Ecdsa();
+    const m = 2;
+    const n = 3;
+
+    const userKeyShare = await MPC.keyShare(1, m, n);
```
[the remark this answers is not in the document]
sachushaji: that's a good question, currently no, but I think we can add it, this especially makes sense in OVC where we have entropy generation in place

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -0,0 +1,288 @@
...
+    const userToBitgoShare = await encryptNShare(
+      bitgoKeyShare,
+      1,
+      userGpgKey.publicKey,
+      userGpgKey.privateKey
+    );
+    const backupToBitgoShare = await encryptNShare(
+      bitgoKeyShare,
+      2,
+      userGpgKey.publicKey,
```
sachushaji, line 227: Lets may be use an independent gpg key pair dedicated for backup since user may not always be the one to generate backup keys

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -0,0 +1,288 @@
...
+      coin: 'gteth',
+      coinSpecific: {},
+    };
+    wallet = new Wallet(bitgo, baseCoin, walletData);
+    tssUtils = new EcdsaUtils(bitgo, baseCoin, wallet);
+  });
+
+  describe('TSS key chains:', async function() {
+    it('should generate TSS key chains', async function () {
+      const nockedBitGoKeychain = await nockBitgoKeychain({
```
sachushaji, line 98: The nockBitgoKeychain is implementing the same logic right for key generation?, do we need to reimplement it to test it by comparing?, i think its better to test after keygen, every property is coming in as expected and may be decrypt and make sure we are getting back the proper values?

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -0,0 +1,288 @@
...
+        bitgoKeychain,
+        'passphrase',
+        'originalPasscodeEncryptionCode');
+      const backupKeychain = await tssUtils.createBackupKeychain(
+        userGpgKey,
+        userKeyShare,
+        backupKeyShare,
+        bitgoKeychain,
+        'passphrase');
+      bitgoKeychain.should.deepEqual(nockedBitGoKeychain);
```
sachushaji, line 157: Not sure we need to test it against nockKeychains, we are implementing the same logic there right?, we can just check if the shares are being actually generated with optional parameters and just check all the relevant properties are there correctly may be?

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -0,0 +1,288 @@
...
+        backupKeyShare,
+        userGpgKey,
+      });
+      const bitgoKeychain = await tssUtils.createBitgoKeychain(userGpgKey, userKeyShare, backupKeyShare);
+      bitgoKeychain.should.deepEqual(nockedBitGoKeychain);
+
+      await tssUtils.createUserKeychain(
+        userGpgKey,
+        userKeyShare,
+        await MPC.keyShare(2, 2, 3),
```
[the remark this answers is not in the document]
sachushaji: or even better, since we are only checking for difference in common key chain, instead of generating a new one, make a random modification in common keychain for an existing keyshares from before hook and use it here

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -0,0 +1,288 @@
...
+    };
+
+    nock(bgUrl)
+      .post(`/api/v2/${params.coin}/key`, _.matches({ keyType: 'tss', source: 'bitgo' }))
+      .reply(200, bitgoKeychain);
+
+    return bitgoKeychain;
+  }
+
+  async function nockUserKeychain(params: {
```
sachushaji, line 258: the nockUserKeychain and backupkeychain seems to have similar logic, then lets combine them to something like nockKeychain and pass the difference as parameters

modules/bitgo/test/v2/unit/tss/ecdsa.ts
```diff
@@ -71,61 +70,33 @@ describe('test tss helper functions', function () {
...
-        senderGpgPrivateArmor: userGpgKeypair.privateKey,
-        recipientGpgPublicArmor: bitgoGpgKeypair.publicKey,
-      }).should.be.rejectedWith('Invalid recipient');
-      await encryptNShare({
-        keyShare: bitgoKeyShare,
-        recipientIndex: 3,
-        senderGpgPrivateArmor: userGpgKeypair.privateKey,
-        recipientGpgPublicArmor: bitgoGpgKeypair.publicKey,
-      }).should.be.rejectedWith('Invalid recipient');
+      await encryptNShare(userKeyShare, 1, userGpgKeypair.privateKey, bitgoGpgKeypair.publicKey).should.be.rejectedWith('Invalid recipient');
```
sachushaji, line 84: I think since we are only switching public <--> private keys in incorrect places as far as I understand, the error should be something like invalid private key / public key given for encryption right rather than invalid recipient?

modules/bitgo/test/v2/unit/tss/ecdsa.ts
```diff
@@ -71,61 +70,33 @@ describe('test tss helper functions', function () {
...
+      await encryptNShare(backupKeyShare, 2, userGpgKeypair.privateKey, bitgoGpgKeypair.publicKey ).should.be.rejectedWith('Invalid recipient');
+      await encryptNShare(bitgoKeyShare, 3, userGpgKeypair.privateKey, bitgoGpgKeypair.publicKey ).should.be.rejectedWith('Invalid recipient');
     });
   });
 
   describe('createCombinedKey', function () {
     it('should create combined user key', async function () {
```
sachushaji, line 91: Since logic for the combine user key and backup key is similar, lets use a config and loop it to test to resuse the same logic, something like I have done [here](https://github.com/BitGo/BitGoJS/blob/df5768f1ef46af117add35f3fefd6adf125d87a3/modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts#L102)

modules/bitgo/test/v2/unit/tss/ecdsa.ts
```diff
@@ -182,23 +154,22 @@ describe('test tss helper functions', function () {
...
+      );
+      const backupToUserShare = await encryptNShare(
+        backupKeyShare,
+        1,
+        userGpgKeypair.publicKey,
+        backupGpgKeypair.privateKey,
+      );
+
+      await createCombinedKey(
+        userKeyShare,
```
sachushaji, line 171: How did userKeyShare didn't match?, did we mutate this in a previous testcase?, probably lets all the mutation within the testcase itself to have give it more clarity

modules/bitgo/test/v2/unit/tss/ecdsa.ts
```diff
@@ -236,7 +207,8 @@ describe('test tss helper functions', function () {
           recipientPrivateArmor: userGpgKeypair.privateKey,
           senderPublicArmor: backupGpgKeypair.publicKey,
         }],
-      }).should.be.rejectedWith('Error decrypting message: Session key decryption failed.');
+        'nottherightkeychain',
+      ).should.be.rejectedWith('Error decrypting message: Session key decryption failed.');
```
sachushaji, line 211: Not sure what is session key actually, may be the error should be something like decryption failed while combining, invalid private or public keys provided?

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -86,36 +86,46 @@ export async function createCombinedKey(params: {
...
+): Promise<EncryptedNShare> {
   const nShare = keyShare.nShares[recipientIndex];
   if (!nShare) {
     throw new Error('Invalid recipient');
   }
 
-  const publicShare = keyShare.pShare.y;
-  const privateShare = nShare.u;
+  const publicShare = trimHexFromString(keyShare.pShare.y) + trimHexFromString(nShare.chaincode);
+  const privateShare = trimHexFromString(nShare.u) + trimHexFromString(nShare.n);
```
sachushaji, line 107: I think n is not something which needs to be encrypted (Its paillier constant n right?), let's add it into the public share may be

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -86,36 +86,46 @@ export async function createCombinedKey(params: {
...
   return {
     i: nShare.i,
     j: nShare.j,
-    n: nShare.n,
     publicShare,
     encryptedPrivateShare,
-    chaincode: nShare.chaincode,
   };
 }
+
```
sachushaji, line 118: Lets have move this to utils where it can be reused

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -0,0 +1,254 @@
...
+        password: passphrase,
+      }),
+      originalPasscodeEncryptionCode,
+    };
+
+    return await this.baseCoin.keychains().add(userKeychainParams);
+  }
+
+  /** @inheritdoc */
+  async createBackupKeychain(
```
sachushaji, line 197: createBackup and createUser is almost having same logic right? can we combine them to a single function and resuse, provide all the necessary difference via parameters

sachushaji (changes requested):

modules/bitgo/test/v2/unit/tss/ecdsa.ts
olibeyene: this should fail to combine the keys because we pass in invalid common key chain

sachushaji (changes requested):
Small changes left

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -0,0 +1,283 @@
...
+  };
+  let nockedBitGoKeychain: Keychain;
+  let nockedUserKeychain: Keychain;
+
+  const coinName = 'gteth';
+
+  beforeEach(async function () {
+    sandbox = sinon.createSandbox();
+
+    nockedBitGoKeychain = await nockBitgoKeychain({
```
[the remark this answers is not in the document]
olibeyene: it causes problem when I added to the before. I'm not familiar with nock much to know the problem.

sachushaji:
Some comments have not been addressed ? @olibeyene

olibeyene:
> Some comments have not been addressed ? @olibeyene

sachushaji (dismissed):
LGTM, Nice work @olibeyene

sachushaji (approved):

sachushaji (approved):

modules/sdk-core/src/bitgo/utils/common.ts
```diff
@@ -0,0 +1,13 @@
+export function trimHexFromString(number: string): string {
```
sachushaji, line 1: Are we using this? If not no need for common.ts

modules/sdk-core/src/bitgo/utils/index.ts
```diff
@@ -9,5 +9,6 @@ export * from './promise-utils';
 export * from './triple';
 export * from './tss';
 export * from './util';
+export * from './common';
```
sachushaji, line 12: No need for this common import as we don't use it

sachushaji (approved):
LGTM , Nice Work @olibeyene

modules/sdk-core/src/bitgo/utils/tss/ecdsa/index.ts
```diff
@@ -0,0 +1,2 @@
+export { EcdsaUtils } from './ecdsa';
+export * as EcdsaUtilsTypes from './types';
```
bitgoAaron, line 2: can we just 
```suggestion
export * from './ecdsa';
export * from './types';
```
and ditch the extra export in the [parent index](https://github.com/BitGo/BitGoJS/pull/2416/files#diff-94348c1eaa8e1cee2a08b2b4a9ea3fdc5297c7a193113cd26e765f33bd009d2aR4)

bitgoAaron (changes requested):
lets simplify the exports if possible

bitgoAaron (approved):

PR #2418, opened 2022-06-23
fix(sdk-core): fix sol create ATA request

jonathancn-bitgo:
BG-51023


joey-bai (approved):

jonathancn-bitgo:
Will add support to create ATA for other address (used for consolidation), so recipients will be required

PR #2420, opened 2022-06-23
chore(sdk-coin-btc): project init

bitgoAaron:
- creates new module sdk-coin-btc
- moves btc, tbtc from bitgo to sdk-coin-btc
- updates imports, re-exports from bitgo
- moves all v1 types & remaining methods on BitGo into sdk-api
- adds CI check for `SKIP_CACHE` label to force yarn installl

Ticket: BG-51034


modules/sdk-coin-btc/test/unit/util/utxoCoins.ts
```diff
@@ -0,0 +1,36 @@
...
+import { BitGoAPI } from '@bitgo/sdk-api';
+import { Btc, Tbtc } from '../../../src';
+
+export const defaultBitGo = TestBitGo.decorate(BitGoAPI, { env: 'mock' });
+
+function getUtxoCoins(): AbstractUtxoCoin[] {
+  defaultBitGo.safeRegister('btc', Btc.createInstance);
+  defaultBitGo.safeRegister('tbtc', Tbtc.createInstance);
+  return [new Btc(defaultBitGo), new Tbtc(defaultBitGo)];
+}
```
OttoAllmendinger, line 16: this significantly reduces test coverage

OttoAllmendinger (changes requested):

modules/sdk-coin-btc/test/unit/util/utxoCoins.ts
bitgoAaron: I did not mean to remove the original tests from bitgo, will revert & exclude btc/tbtc from the tests that run in that package.

OttoAllmendinger: the test suites in this directory are supposed to run over *all* utxo coins though

bitgoAaron: the original tests are still in this branch. All of the utxo unit tests are unchanged:

https://github.com/BitGo/BitGoJS/tree/aloe/btcRefactor/modules/bitgo/test/v2/unit/coins/utxo

the code you're seeing in this file is the same content as what's in `modules/bitgo/test/v2/unit/coins/utxo` , except it only runs the tests for `Btc` and `Tbtc`, since those are the only coins in this module. Eventually when each coin is in its own package we will be able to completely eliminate the utxo tests in bitgo. If anything, we should exclude `Btc` and `Tbtc` from the tests run there, since they are now being duplicated.

modules/sdk-coin-btc/test/unit/recovery/backupKeyRecovery.ts
```diff
@@ -0,0 +1,298 @@
+/**
+ * @prettier
+ */
+import 'should';
+import sinon from 'sinon';
```
ducnguyen-bitgo, line 5: Should we add `sinon`, `nock`, and `bip32` in devDependencies? Also, `lodash` and `should` are used in other test files.

mmcshinsky-bitgo: Our shared devDependencies between multiple modules are hoisted to the root of the monorepo.

ducnguyen-bitgo: I saw in some other coin packages we added those test dependent packages to devDependencies so that's why I wanted to make sure what we should do here as we refactor more coins.

sachushaji (dismissed):
This PR is huge, The changes specific to ETH using interfaces from statics looks good to me, as for the UTXO related stuff @OttoAllmendinger should hopefully be the better reviewer. But a general comment is like most of the UTXO related code functions are not typed even though they exist technically in a typescript file and they use Bluebird style promises which is very ancient , May be could do something about that too in this PR or may be in follow up as anyway we are doing this big refactors

modules/sdk-coin-btc/test/unit/util/utxoCoins.ts
OttoAllmendinger: as discussed on slack: we want the test suite to live in once place only

OttoAllmendinger (changes requested):
let's figure out the canonical place for the utxo coin test suite first

bitgoAaron:
> let's figure out the canonical place for the utxo coin test suite first

I'm going to leave the unit tests in bitgo for now & address this in a separate ticket: https://bitgoinc.atlassian.net/browse/BG-51405

For now, bitgo is the canonical place for utxo coin test suite. It'll likely be another large refactor to make those test functions callable from a separate package and this PR is already too big. Removing the redundant test files in the sdk-coin-btc package makes this PR go from >100 files to 50.

ducnguyen-bitgo (approved):
LGTM.

OttoAllmendinger (approved):

sachushaji (approved):
LGTM on ETH related changes

PR #2423, opened 2022-06-24
chore: add missing props to TxRequest model

joey-bai:
TxRequest model had missing props because we added them as needed.
However, React is using the SDK so it'd be best to have this model be
complete to avoid duplicating definitions.

Ticket: BG-51084

Pretty much a copy of what's defined in wallet-platform: https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/models/v2/TransactionRequest/TransactionRequest.types.ts#L100-L119

Not sure if there's a way to share API models between WP and BitGoJS today, but it would be extremely useful.


modules/bitgo/test/v2/unit/coins/sol.ts
```diff
@@ -903,14 +903,25 @@ describe('SOL:', function () {
...
-      const rebuiltTx = {
+      const rebuiltTx: TxRequest = {
         txRequestId,
         unsignedTxs: [{
           serializedTxHex: 'deadbeef',
           signableHex: 'serializedTxHex',
           derivationPath: 'm/0',
         }],
         transactions: [],
+        date: new Date(),
```
YPAzevedo, line 914: just wondering, doesn't the tx request comes over a rest api? -- will it ever be type `Date` ? 🤔 should always be string no?

mmcshinsky-bitgo (approved):

mmcshinsky-bitgo (approved):

PR #2429, opened 2022-06-24
feat(account-lib): sol create ata for diff owner

jonathancn-bitgo:
Add owner field to Sol ATA creation in `account-lib` to support different owner than the sender

Ticket: BG-51023


modules/account-lib/src/coin/sol/ataInitializationBuilder.ts
```diff
@@ -84,16 +108,17 @@ export class AtaInitializationBuilder extends TransactionBuilder {
 
   /** @inheritdoc */
   protected async buildImplementation(): Promise<Transaction> {
-    assert(this._sender, 'Sender must be set before building the transaction');
```
joey-bai, line 87: we should leave the `assert(this._sender)` check right?

joey-bai (approved):

mmcshinsky-bitgo (approved):

PR #2442, opened 2022-06-28
fix(statics): fix algo alias

JulioAchaval-SimTLiX:
fix algo bug alias

STLX-10016

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -946,7 +946,7 @@ export const coins = CoinMap.fromCoins([
   tronToken('trx:weth', 'Tron Wrapped Ether', 18, 'TXWkP3jLBqRGojUih1ShzNyDaN5Csnebok', UnderlyingAsset.WETH),
   algoToken(
     'algo:USDC-31566704',
-    undefined,
+    'algo:USDC-31566704',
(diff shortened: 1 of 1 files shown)


modules/statics/src/coins.ts
```diff
@@ -946,7 +946,7 @@ export const coins = CoinMap.fromCoins([
   tronToken('trx:weth', 'Tron Wrapped Ether', 18, 'TXWkP3jLBqRGojUih1ShzNyDaN5Csnebok', UnderlyingAsset.WETH),
   algoToken(
     'algo:USDC-31566704',
-    undefined,
+    'algo:USDC-31566704',
```
ducnguyen-bitgo, line 949: If you change alias to be the same as the coin name, this [check](https://github.com/BitGo/BitGoJS/blob/STLX-10016-fix-token-names/modules/statics/src/map.ts#L21) will fail immediately as the coin was already defined/registered in the `CoinMap` [here](https://github.com/BitGo/BitGoJS/blob/STLX-10016-fix-token-names/modules/statics/src/coins.ts#L53) and it is trying to define/register the alias one more time.

ducnguyen-bitgo (changes requested):

PR #2445, opened 2022-06-29
feat(account-lib): add cardano ada statics

margueriteblair:


ducnguyen-bitgo (dismissed):
Can you fix this [unit test](https://github.com/BitGo/BitGoJS/blob/BG-50794-account-lib-add-statics-and-skeleton/modules/statics/test/unit/coins.ts#L37) as it's failing the checks now?

yarn.lock
```diff
@@ -4864,11 +4864,6 @@ argparse@^2.0.1:
   resolved "https://registry.yarnpkg.com/argparse/-/argparse-2.0.1.tgz#246f50f3ca78a3240f6c997e8a9bd1eac49e4b38"
   integrity sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==
 
-argv@0.0.2:
```
mmcshinsky-bitgo, line 4867: Can we remove the yarn.lock change? I don't see any package changes in the PR.

mmcshinsky-bitgo (changes requested):

yarn.lock
margueriteblair: yes, should be fine to remove that, unsure how that got in there. let me know if there's anything i'd need to do on my end for that? TIA

modules/statics/src/ada.ts
```diff
@@ -0,0 +1,80 @@
...
+  features: CoinFeature[];
+  asset: UnderlyingAsset;
+  prefix?: string;
+  suffix?: string;
+  primaryKeyCurve: KeyCurve;
+}
+
+export class Ada extends BaseCoin {
+  public static readonly DEFAULT_FEATURES = [
+    CoinFeature.UNSPENT_MODEL,
```
davidkaplanbitgo, line 17: Question about this - we make distinctions between bitcoin-like unspent models and nonbitcoin-like unspent models in other places like bitgo-microservices, is there a point to do it here or no?

ewangbitgo (approved):

modules/statics/src/ada.ts
margueriteblair: Not 100% sure, I'm not familiar with what it looks like in bitgo-microservices. Could you link to the distinction so I could get some context?

PR #2454, opened 2022-06-30
fix(account-lib): fix ecdsa tests timeout issues

sachushaji:
this change introduces mocks for pallier keys
which are very expensive to generate to fix
timeout issues in test.

TICKET: BG-51505


bitgoAaron (approved):

ducnguyen-bitgo (approved):
LGTM.

PR #2458, opened 2022-06-30
fix(account-lib): fix sol token source address

jonathancn-bitgo:
BG-51536


joey-bai (commented):
do we have a test that covers this already?

jonathancn-bitgo:
> do we have a test that covers this already?

After checking this, we can't test it in unit-test since both `Transaction` and `txData` from `toJson()` doesn't contain the mint token source address in their field, only the original sol address

joey-bai (approved):

mmcshinsky-bitgo (approved):

PR #2460, opened 2022-07-01
fix(sdk-core): make hex representation consistent

sachushaji:
this change makes hex representation consistent
across ECDSA and EDDSA shares

TICKET: BG-50107


modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -111,8 +111,8 @@ export default class Ecdsa {
         l: pShare.l,
         m: pShare.m,
         n: pShare.n,
-        y: bigIntToHex(y),
-        x: bigIntToHex(x),
+        y: y.toString(16),
```
mmcshinsky-bitgo, line 114: Is it okay that the expected output string is now a different syntax? Technically would we consider this a breaking change?

sachushaji: This module is under active development and is not yet being used. It is okay to have it made.

mmcshinsky-bitgo: Awesome, thanks :)

mmcshinsky-bitgo (approved):

PR #2464, opened 2022-07-01
feat(sdk-coin-avaxp): add into accountLib umbrella

jrafael-simtlix:
STLX-17723


mmcshinsky-bitgo (changes requested):
We won't be adding more coins to `account-lib`. Please add new or refactored coins to `bitgo`.

jrafael-simtlix:
@mmcshinsky-bitgo Casper and Algorand move towards this approach. We can go similar. We are near to release and we don’t have time for analysis.
We need AccountLib in WalletPlatform.

PR #2466, opened 2022-07-05
feat(statics)!: update terra a classic

Joelsx:
update terra a classic

BREAKING CHANGE: We are replacing the symbol of ust to usdtc same contract

BG-51357

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -732,7 +732,7 @@ export enum UnderlyingAsset {
   USDX = 'usdx',
   USG = 'usg',
   USPX = 'uspx',
-  UST = 'ust',
+  USTC = 'ustc',
   USX = 'usx',
   UTK = 'utk',
   UTK1 = 'utk1',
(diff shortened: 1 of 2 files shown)


modules/statics/src/base.ts
```diff
@@ -730,7 +730,7 @@ export enum UnderlyingAsset {
   USDX = 'usdx',
   USG = 'usg',
   USPX = 'uspx',
-  UST = 'ust',
```
mmcshinsky-bitgo, line 733: Can you update your commit to reflect this as a breaking change?

mmcshinsky-bitgo (changes requested):

modules/statics/src/base.ts
Joelsx: okay

mmcshinsky-bitgo:
It will need to be on the first line of the commit. See: https://www.conventionalcommits.org/en/v1.0.0/

mmcshinsky-bitgo (dismissed):

mmcshinsky-bitgo (approved):

PR #2468, opened 2022-07-05
BG-50243-determine-fixes-for-breaking-change-in-bit-go-version-15-to-remain-backwards-compatible-master

mmcshinsky-bitgo:
BG-50243


modules/bitgo/src/v2/index.ts
```diff
@@ -1,3 +1,17 @@
+/* eslint-disable import/no-internal-modules */
+/**
+ * @deprecated after version 14.0.0
+ * Retained for short term backwards compatibility - migrate to: @bitgo/sdk-api and @bitgo/sdk-core
+ */
+export * from '@bitgo/sdk-core/dist/src/bitgo/baseCoin/baseCoin';
```
bitgoAaron, line 6: can we just `export * from '@bitgo/sdk-core` instead of digging into dist folders?

mmcshinsky-bitgo: Yeah, it's my conflict between specific backwards compatibility or "adding more features" by exporting everything. :/

mmcshinsky-bitgo: I have some overlapping types to resolve if doing it that way :(

modules/bitgo/src/index.ts
```diff
@@ -28,7 +29,7 @@ export { Buffer } from 'buffer';
 export { CustomSigningFunction, SignedTransaction };
```
bitgoAaron, line 29: these & `export { tss }` below are redundant now that we're exporting * from core

modules/sdk-core/src/index.ts
```diff
@@ -1,6 +1,8 @@
 export * from './account-lib';
+export * as accountLib from './account-lib';
 export * from './api';
 export * from './bitgo';
+export * as bitgo from './bitgo';
```
bitgoAaron, line 5: are these still necessary if we're `exporting * from './bitgo` ? Do we still need to namespace them?

bitgoAaron (changes requested):
couple of extra exports still

modules/sdk-core/src/index.ts
mmcshinsky-bitgo: Not anymore. I'll remove them.

bitgoAaron (approved):

PR #2469, opened 2022-07-05
feat(sdk-core): BG-49288 core implement ecdsa signing

sachushaji:
Add ECDSA signing functionality to SDK core


modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -82,6 +103,158 @@ export async function createCombinedKey(
...
+  }
+
+  if (yShare.i === ShareKeyPosition.BITGO || yShare.j !== ShareKeyPosition.BITGO) {
+    throw new Error('Invalid YShare provided for sign');
+  }
+  return MPC.signShare(xShare, yShare);
+}
+
+/**
+ * Creates the Gamma Share with User WShare and AShare From BitGo
```
anshuldoshi, line 124: nit: Createes the Gamma Share and Mu Share with User WShare and AShare from BitGo

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -230,4 +234,73 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
+
+    const bitgoToUserDShare = (await ECDSAMethods.sendShareToBitgo(
+      this.bitgo,
+      this.wallet.id(),
+      txRequestId,
+      SendShareType.MUShare,
+      userToBitGoGShare.muShare as ECDSA.MUShare
+    )) as DShare;
+
+    const userOShare = await ECDSAMethods.createUserOmicronShare(userToBitGoGShare.gShare as ECDSA.GShare);
```
anshuldoshi, line 293: Here too, we are returning the user's omicron share as well as their delta share, can we rename the variable/function name to reflect this?

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -230,4 +234,73 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
+
+    const signablePayload = Buffer.from(txRequestResolved.unsignedTxs[0].signableHex, 'hex');
+
+    const userSShare = await ECDSAMethods.createUserSignatureShare(
+      userOShare.oShare,
+      bitgoToUserDShare,
+      signablePayload
+    );
+
+    await ECDSAMethods.sendShareToBitgo(this.bitgo, this.wallet.id(), txRequestId, SendShareType.SShare, userSShare);
```
anshuldoshi, line 303: When do we send the user's delta share to BitGo? Feels like we should be doing it in this step

anshuldoshi (commented):
Looks good overall, left a few nits and questions.

Also, a general question for the signing flow: how will we handle the case in which one of the round trips fails for some reason (e.g. bad network connection)? On the platform side, we decide what step we're on by the number of shares saved on the TxRequest document. I believe EDDSA has some mechanism where they delete the shares on the TxRequest document and restart, are we planning to do the same? Can we take advantage of the fact that we can cache intermediary shares (e.g. o and d shares) to improve on this? Not a blocking issue, but perhaps something to look into.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -230,4 +234,73 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
   }
+
+  /**
+   * Signs the transaction associated to the transaction request.
+   * @param {string | TxRequest} params.txRequest - transaction request object or id
+   * @param {string} params.prv - decrypted private key
+   * @param { string} params.reqId - request id
+   * @returns {Promise<TxRequest>} fully signed TxRequest object
+   */
+  async signTxRequest(params: {
```
olibeyene, line 245: I don't see any test for this method. can add test for this method?

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -82,6 +103,158 @@ export async function createCombinedKey(
...
+  }
+  if (aShare.i === ShareKeyPosition.BITGO || aShare.j !== ShareKeyPosition.BITGO) {
+    throw new Error('Invalid AShare, is not from Bitgo to User');
+  }
+
+  return MPC.signConvert({ wShare, aShare });
+}
+
+/**
+ * Creates the Omicron Share with User GShare and MUShare From BitGo
```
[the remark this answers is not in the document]
sachushaji: Right, nice find, updated.

modules/sdk-core/src/bitgo/tss/ecdsa/types.ts
```diff
@@ -1,5 +1,27 @@
...
+export type KeyShare = ECDSA.KeyShare;
+export type XShare = ECDSA.XShare;
+export type YShare = ECDSA.YShare;
+export type SignShare = ECDSA.SignShareRT;
+export type WShare = ECDSA.WShare;
+export type AShare = ECDSA.AShare;
+export type GShare = ECDSA.GShare;
+export type OShare = ECDSA.OShare;
+export type DShare = ECDSA.DShare;
+export type createUserGammaShareRT = ECDSA.SignConvertRT; // Recheck
```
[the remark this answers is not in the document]
sachushaji: Nice Catch, updated

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -230,4 +234,73 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
+
+    const bitgoToUserAShare = (await ECDSAMethods.sendShareToBitgo(
+      this.bitgo,
+      this.wallet.id(),
+      txRequestId,
+      SendShareType.KShare,
+      userSignShare.kShare
+    )) as AShare;
+
+    const userToBitGoGShare = await ECDSAMethods.createUserGammaShare(userSignShare.wShare, bitgoToUserAShare);
```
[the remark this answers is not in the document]
sachushaji: Yeah makes sense, I just followed the convention used in Eddsa.

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
```diff
@@ -86,4 +87,37 @@ export abstract class MpcUtils {
...
+      comment: params.comment,
+      nonce: params.nonce,
+      recipients: intentRecipients,
+      memo: params.memo?.value,
+      token: params.tokenName,
+    };
+    if (baseCoin.getFamily() === 'eth') {
+      // Fill in intents for eth as it finalizes
+      return {
+        ...baseIntent,
```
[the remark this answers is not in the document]
sachushaji: Updated, referred types from this link. But I think it makes sense to make the types from SDK as a single source of truth. Otherwise we would have to do copy pasting each time we update an intent

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
sachushaji: Really nice catch, Will write a test case for sign tx which I missed as Oli pointed out

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
sachushaji: Yup, Good catch @olibeyene

sachushaji:
@anshuldoshi , Thanks for the awesome review. Regarding the flow, Yeah, we could persist the encrypted versions of all the shares from client side also at WP so that in case a failure occur, we could send back the encrypted versions back to the client to restart from the failed step. I think at this phase probably, its a bit too complicated as we need to define additional endpoints and functionalities to do the same. Probably we can implement this at a later phase with this supported for eddsa as well.

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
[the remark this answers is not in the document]
anshuldoshi: agreed yeah having the types live in sdk makes more sense, we should open up a ticket for this

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -82,6 +103,165 @@ export async function createCombinedKey(
...
+      break;
+    case SendShareType.SShare:
+      if (!dShare) {
+        throw new Error('DShare not provided');
+      }
+      if (dShare.i !== ShareKeyPosition.BITGO || dShare.j === ShareKeyPosition.BITGO) {
+        throw new Error('Invalid Dshare provided');
+      }
+      const sShare = share as SShare;
+      signatureShare.share = sShare.r + sShare.s + sShare.y + dShare.delta + dShare.Gamma;
```
anshuldoshi, line 246: Is this a valid way to send multiple shares? Not a blocker since this capability is not supported yet, but this PR (https://github.com/BitGo/bitgo-microservices/pull/20947) will allow us to pass in multiple signature share records once it's merged so maybe we can change this to use that format in the future

anshuldoshi (commented):
Awesome work, thanks for addressing everything @sachushaji! Left a couple comments but otherwise LGTM

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
```diff
@@ -86,4 +87,42 @@ export abstract class MpcUtils {
...
+        feeOptions: params.feeOptions,
+        hopParams: params.hopParams,
+        isTss: params.isTss,
+      };
+    }
+
+    return {
+      ...baseIntent,
+      memo: params.memo?.value,
+      token: params.tokenName,
```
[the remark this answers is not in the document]
sachushaji: This is a common function which will be used for both eddsa and ecdsa. If you see above, we have a if condition. which will only return eth like intents avoiding memo and token params.

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
sachushaji: Okay nice, But Let's have it tracked via another ticket and handle the new case once its merged.

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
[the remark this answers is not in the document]
anshuldoshi: ahh sorry didn't see that, good point

anshuldoshi (dismissed):

rlutzbitgo (dismissed):
Nice, LGTM

olibeyene (dismissed):
lgtm!

modules/bitgo/test/v2/unit/internal/tssUtils/common.ts
```diff
@@ -0,0 +1,42 @@
+import { SignatureShareRecord } from '@bitgo/sdk-core';
```
olibeyene, line 1: duplicate code

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/types.ts
```diff
@@ -150,7 +150,7 @@ export interface SShare {
 }
 
 export interface SignCombine {
-  gShares: GShare;
+  gShare: GShare;
```
mmcshinsky-bitgo, line 153: Technically a BREAKING CHANGE due to this being out in `latest`. Is there any risk of current end users using this and having to make the change?

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -82,6 +103,165 @@ export async function createCombinedKey(
...
 }
 
+/**
+ * Creates the SignShare with User XShare and YShare Corresponding to BitGo
+ * @param {XShare} xShare User secret xShare
+ * @param {YShare} yShare YShare from Bitgo
+ * @returns {Promise<SignShare>}
+ */
+export async function createUserSignShare(xShare: XShare, yShare: YShare): Promise<SignShare> {
+  if (xShare.i === ShareKeyPosition.BITGO) {
```
olibeyene, line 113: ShareKeyPosition.BITGO should be ShareKeyPosition.User

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -82,6 +103,165 @@ export async function createCombinedKey(
...
+    throw new Error(`Invalid XShare, XShare doesn't belong to the User`);
+  }
+
+  if (yShare.i === ShareKeyPosition.BITGO || yShare.j !== ShareKeyPosition.BITGO) {
```
olibeyene, line 117: `if (yShare.i === ShareKeyPosition.BITGO || yShare.j === ShareKeyPosition.User) {`

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/types.ts
olibeyene: ecdsa is not released yet

sachushaji: Yes, this module is actively under development and is not being used yet

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
sachushaji: Hmm, When user creates the signShare it should be the users xShare to be used for this?

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
sachushaji: I didn't explicitly mention user because backup could also be used to sign as the second part instead of user share, although could be only user share since backup would probably be only used in ovc, want me to restrict only to user?

olibeyene: ah good point

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
olibeyene: yes I believe the x share should belong to user for the user signing share. we should throw an error if it's not user's x share.

sachushaji: Same point as below, considering backup also

olibeyene (approved):

mmcshinsky-bitgo (approved):

PR #2480, opened 2022-07-06
feat(account-lib): get rid of old ethereum lib

dpc:
It pulls in some old dependencies, that waste time and fail to build in our CI (and falls back to slower JS variants).

I tried to keep things minimal and as close to the previous version as possible.

Please be aware of this node.js papercut:

```
> Buffer.from('abc', 'hex').toString('hex');
'ab'
```

Ticket: BG-45514

### DIFF
--- a/modules/sdk-coin-eth/src/eth.ts
+++ b/modules/sdk-coin-eth/src/eth.ts
@@ -548,7 +548,7 @@ export class Eth extends BaseCoin {
         'ETHER',
         new optionalDeps.ethUtil.BN(optionalDeps.ethUtil.stripHexPrefix(recipient.address), 16),
         recipient.amount,
-        Buffer.from(optionalDeps.ethUtil.stripHexPrefix(recipient.data) || '', 'hex'),
+        Buffer.from(optionalDeps.ethUtil.stripHexPrefix(optionalDeps.ethUtil.padToEven(recipient.data || '')), 'hex'),
         expireTime,
         contractSequenceId,
       ],
@@ -722,6 +722,10 @@ export class Eth extends BaseCoin {
     // if no recipients in either params or txPrebuild, then throw an error
     if (!params.recipients || !Array.isArray(params.recipients)) {
       throw new Error('recipients missing or not array');
+    }
+
+    if (params.recipients.length == 0) {
+      throw new Error('recipients empty');
     }
 
     // Normally the SDK provides the first signature for an ETH tx, but occasionally it provides the second and final one.
@@ -1384,7 +1388,10 @@ export class Eth extends BaseCoin {
     const serverXpub = common.Environments[this.bitgo.getEnv()].hsmXpub;
     const serverPubkeyBuffer: Buffer = bip32.fromBase58(serverXpub).publicKey;
     const signatureBuffer: Buffer = Buffer.from(optionalDeps.ethUtil.stripHexPrefix(signature), 'hex');
-    const messageBuffer: Buffer = Buffer.from(optionalDeps.ethUtil.stripHexPrefix(id), 'hex');
+    const messageBuffer: Buffer = Buffer.from(
+      optionalDeps.ethUtil.padToEven(optionalDeps.ethUtil.stripHexPrefix(id)),
+      'hex'
+    );
 
     const sig = new Uint8Array(signatureBuffer.slice(1));
     const isValidSignature: boolean = secp256k1.ecdsaVerify(sig, messageBuffer, serverPubkeyBuffer);
@@ -1549,6 +1556,8 @@ export class Eth extends BaseCoin {
    * @returns {Boolean} True iff address is a wallet address
    */
   isWalletAddress(params: VerifyEthAddressOptions): boolean {
+    const ethUtil = optionalDeps.ethUtil;
+
     let expectedAddress;
     let actualAddress;
 
@@ -1577,7 +1586,10 @@ export class Eth extends BaseCoin {
       const forwarderImplementationAddress = ethNetwork?.forwarderImplementationAddress as string;
 
       const initcode = getProxyInitcode(forwarderImplementationAddress);
-      const saltBuffer = optionalDeps.ethUtil.setLengthLeft(coinSpecific.salt, 32);
+      const saltBuffer = ethUtil.setLengthLeft(
+        Buffer.from(ethUtil.padToEven(ethUtil.stripHexPrefix(coinSpecific.salt || '')), 'hex'),
+        32
+      );
 
       // Hash the wallet base address with the given salt, so the address directly relies on the base address
       const calculationSalt = optionalDeps.ethUtil.bufferToHex(
--- a/modules/sdk-coin-eth/src/lib/types.ts
+++ b/modules/sdk-coin-eth/src/lib/types.ts
@@ -1,3 +1,4 @@
+import assert from 'assert';
 import BigNumber from 'bignumber.js';
 import {
   TransactionFactory,
@@ -7,8 +8,7 @@ import {
   AccessListEIP2930Transaction,
 } from '@ethereumjs/tx';
 import EthereumCommon from '@ethereumjs/common';
-import { bufferToHex, bufferToInt } from '@bitgo/ethereumjs-utils-old';
-import { toBuffer, addHexPrefix } from 'ethereumjs-util';
+import { bufferToHex, bufferToInt, toBuffer, toUnsigned, addHexPrefix } from 'ethereumjs-util';
 import { BaseTxData, EIP1559TxData, EthLikeTransactionData, LegacyTxData, ETHTransactionType, TxData } from './iface';
 import { KeyPair } from './keyPair';
 
@@ -94,8 +94,8 @@ export class EthTransactionData implements EthLikeTransactionData {
   /** @inheritdoc */
   toJson(): TxData {
     const result: BaseTxData = {
-      nonce: bufferToInt(this.tx.nonce),
-      gasLimit: new BigNumber(bufferToHex(this.tx.gasLimit), 16).toString(10),
+      nonce: bufferToInt(toUnsigned(this.tx.nonce)),
+      gasLimit: new BigNumber(bufferToHex(toUnsigned(this.tx.gasLimit)), 16).toString(10),
       value: this.tx.value.toString(10),
       data: bufferToHex(this.tx.data),
     };
@@ -112,11 +112,13 @@ export class EthTransactionData implements EthLikeTransactionData {
 
     if (this.tx.verifySignature()) {
       result.from = bufferToHex(this.tx.getSenderAddress().toBuffer());
-      result.r = bufferToHex(this.tx.r);
-      result.s = bufferToHex(this.tx.s);
+      assert(this.tx.r != undefined);
+      result.r = bufferToHex(toUnsigned(this.tx.r));
+      assert(this.tx.s != undefined);
+      result.s = bufferToHex(toUnsigned(this.tx.s));
     }
     if (this.tx.v) {
-      result.v = bufferToHex(this.tx.v);
+      result.v = bufferToHex(toUnsigned(this.tx.v));
     }
     result.chainId = addHexPrefix(this.tx.common.chainIdBN().toString(16));
 
@@ -125,7 +127,7 @@ export class EthTransactionData implements EthLikeTransactionData {
     }
 
     if (this.tx instanceof LegacyTransaction) {
-      const gasPrice = new BigNumber(bufferToHex(this.tx.gasPrice), 16).toString(10);
+      const gasPrice = new BigNumber(bufferToHex(toUnsigned(this.tx.gasPrice)), 16).toString(10);
 
       return {
         ...result,
@@ -133,8 +135,10 @@ export class EthTransactionData implements EthLikeTransactionData {
         gasPrice,
       };
     } else if (this.tx instanceof FeeMarketEIP1559Transaction) {
-      const maxFeePerGas = new BigNumber(bufferToHex(this.tx.maxFeePerGas), 16).toString(10);
-      const maxPriorityFeePerGas = new BigNumber(bufferToHex(this.tx.maxPriorityFeePerGas), 16).toString(10);
+      const maxFeePerGas = new BigNumber(bufferToHex(toUnsigned(this.tx.maxFeePerGas)), 16).toString(10);
+      const maxPriorityFeePerGas = new BigNumber(bufferToHex(toUnsigned(this.tx.maxPriorityFeePerGas)), 16).toString(
+        10
+      );
 
       return {
         ...result,
--- a/modules/sdk-coin-eth/src/lib/utils.ts
+++ b/modules/sdk-coin-eth/src/lib/utils.ts
@@ -9,12 +9,13 @@ import {
   setLengthLeft,
   stripHexPrefix,
   toBuffer,
-} from '@bitgo/ethereumjs-utils-old';
-import { generateAddress2 } from 'ethereumjs-util';
+  generateAddress2,
+  padToEven,
+} from 'ethereumjs-util';
 import { BaseCoin, BaseNetwork, coins, ContractAddressDefinedToken, EthereumNetwork } from '@bitgo/statics';
 import EthereumAbi from 'ethereumjs-abi';
 import EthereumCommon from '@ethereumjs/common';
-import * as BN from 'bn.js';
+import BN from 'bn.js';
 import BigNumber from 'bignumber.js';
 import {
   ActivateMethodId,
@@ -280,12 +281,12 @@ export function decodeTokenTransferData(data: string): TokenTransferData {
   );
 
   return {
-    to: addHexPrefix(to),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(tokenContractAddress),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    expireTime: bufferToInt(expireTime as Buffer),
+    sequenceId: bufferToInt(sequenceId as Buffer),
+    signature: bufferToHex(signature as Buffer),
+    tokenContractAddress: addHexPrefix(tokenContractAddress as string),
   };
 }
 
@@ -299,7 +300,7 @@ export function decodeERC721TransferData(data: string): ERC721TransferData {
     getBufferedByteCode(sendMultisigMethodId, data)
   );
 
-  const internalDataHex = bufferToHex(internalData);
+  const internalDataHex = bufferToHex(internalData as Buffer);
   if (!internalDataHex.startsWith(ERC721SafeTransferTypeMethodId)) {
     throw new BuildTransactionError(`Invalid transfer bytecode: ${data}`);
   }
@@ -310,15 +311,15 @@ export function decodeERC721TransferData(data: string): ERC721TransferData {
   );
 
   return {
-    to: addHexPrefix(receiver),
-    from: addHexPrefix(from),
-    expireTime: bufferToInt(expireTime),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    tokenId: new BigNumber(bufferToHex(tokenId)).toFixed(),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(to),
-    userData: bufferToHex(userSentData),
+    to: addHexPrefix(receiver as string),
+    from: addHexPrefix(from as string),
+    expireTime: bufferToInt(expireTime as Buffer),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    tokenId: new BigNumber(bufferToHex(tokenId as Buffer)).toFixed(),
+    sequenceId: bufferToInt(sequenceId as Buffer),
+    signature: bufferToHex(signature as Buffer),
+    tokenContractAddress: addHexPrefix(to as string),
+    userData: bufferToHex(userSentData as Buffer),
   };
 }
 
@@ -336,7 +337,7 @@ export function decodeERC1155TransferData(data: string): ERC1155TransferData {
     getBufferedByteCode(sendMultisigMethodId, data)
   );
 
-  const internalDataHex = bufferToHex(internalData);
+  const internalDataHex = bufferToHex(internalData as Buffer);
   if (internalDataHex.startsWith(ERC1155SafeTransferTypeMethodId)) {
     let tokenId;
     let value;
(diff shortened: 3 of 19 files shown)


yarn.lock
```diff
@@ -11224,16 +11222,6 @@ karma@^5.1.1:
     ua-parser-js "0.7.22"
     yargs "^15.3.1"
 
-keccak@^1.0.2:
```
dpc, line 11227: :muscle:

modules/sdk-coin-eth/src/eth.ts
```diff
@@ -724,6 +724,10 @@ export class Eth extends BaseCoin {
       throw new Error('recipients missing or not array');
     }
 
+    if (params.recipients.length == 0) {
```
dpc, line 727: Mostly to satisfy the type system, but also seems like a good idea.

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -552,15 +562,29 @@ export function hasSignature(txData: TxData): boolean {
...
 /**
  * Get the raw data decoded for some types
  *
  * @param {string[]} types ABI types definition
  * @param {Buffer} serializedArgs encoded args
  * @returns {Buffer[]} the decoded raw
  */
-export function getRawDecoded(types: string[], serializedArgs: Buffer): Buffer[] {
-  return EthereumAbi.rawDecode(types, serializedArgs);
+export function getRawDecoded(types: string[], serializedArgs: Buffer): RecursiveBufferOrString[] {
```
dpc, line 574: The library underneath did not change, but the signature here was all wrong, and the old ethUtils were very lax about types and did a lot of auto-conversions that made up for it.

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -575,6 +599,9 @@ export function getBufferedByteCode(methodId: string, rawData: string): Buffer {
   if (splitBytecode.length !== 2) {
     throw new BuildTransactionError(`Invalid send bytecode: ${rawData}`);
   }
+  if (splitBytecode[1].length % 2 != 0) {
```
sachushaji, line 602:
```suggestion
  if (splitBytecode[1].length % 2 !== 0) {
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -552,15 +562,29 @@ export function hasSignature(txData: TxData): boolean {
...
  * Get the raw data decoded for some types
  *
  * @param {string[]} types ABI types definition
  * @param {Buffer} serializedArgs encoded args
  * @returns {Buffer[]} the decoded raw
  */
-export function getRawDecoded(types: string[], serializedArgs: Buffer): Buffer[] {
-  return EthereumAbi.rawDecode(types, serializedArgs);
+export function getRawDecoded(types: string[], serializedArgs: Buffer): RecursiveBufferOrString[] {
+  function normalize(v: any, i: number): any {
```
sachushaji, line 575:
```suggestion
  function normalize(v: unknown, i: number): unknown {
```
any turns off the type checking for that variable, more correct to use unknown as it  forces you to typecheck before use (Which you are doing btw)

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -417,8 +418,8 @@ export function decodeFlushTokensData(data: string): FlushTokensData {
   );
 
   return {
-    forwarderAddress: addHexPrefix(forwarderAddress),
-    tokenAddress: addHexPrefix(tokenAddress),
+    forwarderAddress: addHexPrefix(forwarderAddress as string),
+    tokenAddress: addHexPrefix(tokenAddress as string),
```
sachushaji, line 422:
```suggestion
    tokenAddress: addHexPrefix(tokenAddress.toString()),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -417,8 +418,8 @@ export function decodeFlushTokensData(data: string): FlushTokensData {
   );
 
   return {
-    forwarderAddress: addHexPrefix(forwarderAddress),
-    tokenAddress: addHexPrefix(tokenAddress),
+    forwarderAddress: addHexPrefix(forwarderAddress as string),
```
sachushaji, line 421:
```suggestion
    forwarderAddress: addHexPrefix(forwarderAddress.toString()),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -391,12 +392,12 @@ export function decodeNativeTransferData(data: string): NativeTransferData {
...
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    data: bufferToHex(internalData),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    expireTime: bufferToInt(expireTime as Buffer),
+    sequenceId: bufferToInt(sequenceId as Buffer),
+    signature: bufferToHex(signature as Buffer),
+    data: bufferToHex(internalData as Buffer),
```
sachushaji, line 400:
```suggestion
    data: bufferToHex(Buffer.from(internalData)),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -391,12 +392,12 @@ export function decodeNativeTransferData(data: string): NativeTransferData {
...
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    data: bufferToHex(internalData),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    expireTime: bufferToInt(expireTime as Buffer),
+    sequenceId: bufferToInt(sequenceId as Buffer),
+    signature: bufferToHex(signature as Buffer),
```
sachushaji, line 399:
```suggestion
    signature: bufferToHex(Buffer.from(signature)),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -391,12 +392,12 @@ export function decodeNativeTransferData(data: string): NativeTransferData {
...
-    to: addHexPrefix(to),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    data: bufferToHex(internalData),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    expireTime: bufferToInt(expireTime as Buffer),
+    sequenceId: bufferToInt(sequenceId as Buffer),
```
sachushaji, line 398:
```suggestion
    sequenceId: bufferToInt(Buffer.from(sequenceId as Buffer)),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -391,12 +392,12 @@ export function decodeNativeTransferData(data: string): NativeTransferData {
...
   return {
-    to: addHexPrefix(to),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    data: bufferToHex(internalData),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    expireTime: bufferToInt(expireTime as Buffer),
```
sachushaji, line 397:
```suggestion
    expireTime: bufferToInt(Buffer.from(expireTime)),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -391,12 +392,12 @@ export function decodeNativeTransferData(data: string): NativeTransferData {
...
 
   return {
-    to: addHexPrefix(to),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    data: bufferToHex(internalData),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
```
sachushaji, line 396:
```suggestion
    amount: new BigNumber(bufferToHex(Buffer.from(amount))).toFixed(),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -391,12 +392,12 @@ export function decodeNativeTransferData(data: string): NativeTransferData {
   );
 
   return {
-    to: addHexPrefix(to),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    data: bufferToHex(internalData),
+    to: addHexPrefix(to as string),
```
sachushaji, line 395:
```suggestion
    to: addHexPrefix(to.toString()),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -310,15 +311,15 @@ export function decodeERC721TransferData(data: string): ERC721TransferData {
...
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(to),
-    userData: bufferToHex(userSentData),
+    to: addHexPrefix(receiver as string),
+    from: addHexPrefix(from as string),
+    expireTime: bufferToInt(expireTime as Buffer),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    tokenId: new BigNumber(bufferToHex(tokenId as Buffer)).toFixed(),
+    sequenceId: bufferToInt(sequenceId as Buffer),
+    signature: bufferToHex(signature as Buffer),
```
sachushaji, line 320:
```suggestion
    signature: bufferToHex(Buffer.from(signature)),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -310,15 +311,15 @@ export function decodeERC721TransferData(data: string): ERC721TransferData {
...
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    tokenId: new BigNumber(bufferToHex(tokenId)).toFixed(),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(to),
-    userData: bufferToHex(userSentData),
+    to: addHexPrefix(receiver as string),
+    from: addHexPrefix(from as string),
+    expireTime: bufferToInt(expireTime as Buffer),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
```
sachushaji, line 317:
```suggestion
    amount: new BigNumber(bufferToHex(Buffer.from(amount))).toFixed(),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -299,7 +300,7 @@ export function decodeERC721TransferData(data: string): ERC721TransferData {
     getBufferedByteCode(sendMultisigMethodId, data)
   );
 
-  const internalDataHex = bufferToHex(internalData);
+  const internalDataHex = bufferToHex(internalData as Buffer);
```
sachushaji, line 303:
```suggestion
  const internalDataHex = bufferToHex(Buffer.from(internalData));
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -280,12 +281,12 @@ export function decodeTokenTransferData(data: string): TokenTransferData {
...
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(tokenContractAddress),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    expireTime: bufferToInt(expireTime as Buffer),
+    sequenceId: bufferToInt(sequenceId as Buffer),
+    signature: bufferToHex(signature as Buffer),
+    tokenContractAddress: addHexPrefix(tokenContractAddress as string),
```
sachushaji, line 289:
```suggestion
    tokenContractAddress: addHexPrefix(tokenContractAddress.toString()),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -280,12 +281,12 @@ export function decodeTokenTransferData(data: string): TokenTransferData {
...
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(tokenContractAddress),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    expireTime: bufferToInt(expireTime as Buffer),
+    sequenceId: bufferToInt(sequenceId as Buffer),
+    signature: bufferToHex(signature as Buffer),
```
sachushaji, line 288:
```suggestion
    signature: bufferToHex(Buffer.from(signature)),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -280,12 +281,12 @@ export function decodeTokenTransferData(data: string): TokenTransferData {
...
-    to: addHexPrefix(to),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(tokenContractAddress),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    expireTime: bufferToInt(expireTime as Buffer),
+    sequenceId: bufferToInt(sequenceId as Buffer),
```
sachushaji, line 287:
```suggestion
    sequenceId: bufferToInt(Buffer.from(sequenceId)),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -280,12 +281,12 @@ export function decodeTokenTransferData(data: string): TokenTransferData {
...
   return {
-    to: addHexPrefix(to),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(tokenContractAddress),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
+    expireTime: bufferToInt(expireTime as Buffer),
```
sachushaji, line 286:
```suggestion
    expireTime: bufferToInt(Buffer.from(expireTime)),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -280,12 +281,12 @@ export function decodeTokenTransferData(data: string): TokenTransferData {
...
 
   return {
-    to: addHexPrefix(to),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(tokenContractAddress),
+    to: addHexPrefix(to as string),
+    amount: new BigNumber(bufferToHex(amount as Buffer)).toFixed(),
```
sachushaji, line 285:
```suggestion
    amount: new BigNumber(bufferToHex(Buffer.from(amount))).toFixed(),
```

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -280,12 +281,12 @@ export function decodeTokenTransferData(data: string): TokenTransferData {
   );
 
   return {
-    to: addHexPrefix(to),
-    amount: new BigNumber(bufferToHex(amount)).toFixed(),
-    expireTime: bufferToInt(expireTime),
-    sequenceId: bufferToInt(sequenceId),
-    signature: bufferToHex(signature),
-    tokenContractAddress: addHexPrefix(tokenContractAddress),
+    to: addHexPrefix(to as string),
```
sachushaji, line 284:
```suggestion
    to: addHexPrefix(to.toString()),
```

modules/account-lib/src/coin/avaxc/utils.ts
```diff
@@ -50,7 +50,14 @@ export function isValidEthAddress(address: string): boolean {
  * @returns {boolean} - the validation result
  */
 export function isValidEthPrivateKey(privateKey: string): boolean {
+  if (privateKey.length != 64) {
+    return false;
+  }
   const privateKeyBuffer = Buffer.from(privateKey, 'hex');
+
+  if (privateKeyBuffer.length != 32) {
```
sachushaji, line 58: Since we are already checking for length == 64 above, this check may be redundant?

modules/account-lib/src/coin/avaxc/utils.ts
```diff
@@ -50,7 +50,14 @@ export function isValidEthAddress(address: string): boolean {
  * @returns {boolean} - the validation result
  */
 export function isValidEthPrivateKey(privateKey: string): boolean {
+  if (privateKey.length != 64) {
```
sachushaji, line 53:
```suggestion
  if (privateKey.length !== 64) {
```
enforces stricter type check and also suppresses eslint complaint

modules/sdk-coin-eth/src/lib/types.ts
```diff
@@ -112,11 +111,17 @@ export class EthTransactionData implements EthLikeTransactionData {
 
     if (this.tx.verifySignature()) {
       result.from = bufferToHex(this.tx.getSenderAddress().toBuffer());
-      result.r = bufferToHex(this.tx.r);
-      result.s = bufferToHex(this.tx.s);
+      // TODO: assert?
+      if (this.tx.r != undefined) {
+        result.r = bufferToHex(toUnsigned(this.tx.r));
```
sachushaji, line 116:
```suggestion
        result.r = bufferToHex(toUnsigned(this.tx.r as BN));
```
probably safe to remove the if condition and do something like this, incase r is undefined, current change will simply ignore which is not what we need, the above change should throw an error

sachushaji (commented):
Thanks a lot @dpc for taking this up, Commented some suggestions

modules/account-lib/src/coin/celo/stakingBuilder.ts
```diff
@@ -257,36 +257,36 @@ export class StakingBuilder {
       case StakingOperationTypes.VOTE:
         this.validateDecodedDataLength(decoded.length, 4, data);
         const [groupToVote, amount, lesser, greater] = decoded;
-        this._amount = ethUtil.bufferToHex(amount);
-        this._validatorGroup = ethUtil.addHexPrefix(groupToVote);
-        this._lesser = ethUtil.addHexPrefix(lesser);
-        this._greater = ethUtil.addHexPrefix(greater);
+        this._amount = ethUtil.bufferToHex(amount as Buffer);
+        this._validatorGroup = ethUtil.addHexPrefix(groupToVote as string);
```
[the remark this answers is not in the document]
dpc: We are always deterministically getting `string` for this field. And it would have to be `toString('hex')` to convert to hex no?

modules/account-lib/src/coin/avaxc/utils.ts
dpc: That's OK. :)

modules/sdk-coin-eth/src/lib/types.ts
dpc: >  the above change should throw an error

I thought `as` is just a type-system thing, and no-op at runtime, so it will not throw an error.

modules/account-lib/src/coin/avaxc/utils.ts
mmcshinsky-bitgo, line 58:
```suggestion
  if (privateKeyBuffer.length !== 32) { 
```

sachushaji:
> @sachushaji I don't think this to and from buffer conversion will work as it requires `'hex'`):
> 
> ```
> > Buffer.from('abcdef').toString('hex')
> '616263646566'
> ```
> 
> Also, I don't like the mechanism itself. Each value there already has a determined type, and we don't need extra conversions unifying anything. If anything I'd add some assertion instead that would throw if the type is not the one expected.

@dpc , I think you are right, Should be hex when we do the conversion. Type assertion makes sense. Since the actual type was `RecursiveBufferOrString` from the normalise function, changing it to explicitly a type was my thought for the suggestions to do the conversion since the normalise function looked like it can give either.

modules/sdk-coin-eth/src/lib/types.ts
sachushaji: Yeah, but if the value passes indeed undefined, bufferToHex / toUnsigned from eth utils should throw an error right?, currently we are just skipping if this is undefined

modules/account-lib/src/coin/celo/stakingBuilder.ts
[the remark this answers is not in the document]
sachushaji: yup you're right, it should be hex.

modules/sdk-coin-eth/src/lib/types.ts
dpc: The previous code wasn't checking for it, and I'd assume this can never happen. `r` and `s` on signatures, is not something that can just not be there.

modules/sdk-coin-eth/src/lib/utils.ts
dpc: I don't know. Then we are going to be converting everything to string, just to unconvert it to Buffer again in many cases etc.

At the core it's you decode bunch of data from a binary of transactions. You pass parameter that specify the types of data you expect. And then you get back your data in corresponding types that are guaranteed to be of the right type.

This normalize function stuff there is to fix incorrect type signature that was lying before (and we were getting away with it because this old eth lib was doing a lot of auto-conversion everywhere, and then figured out that it's a stupid idea (which is true)), and we we couldn't upgrade to new library because of that.

As long as we keep this weird and old `ethereum-abi` library, I think it's OK to just specify that `string | buffer` is returned and let the caller pick, the type that it knows the return value is in.

modules/sdk-coin-eth/src/lib/types.ts
sachushaji: Yup r and s should be there, the previous code would  throw an error in case when we passed in undefined and I think that should be the correct behaviour we would expect.

sachushaji: Previous code as in the original code without the if condition before your change, if r is undefined, the current code just skips right without throwing an error?, When you pass in undefined to the util function from the eth dependency, it will trigger an error as far as I understand. Just wanted to keep the same behaviour of the code here.

sachushaji (approved):
LGTM

PR #2481, opened 2022-07-06
feat(account-lib): update hbar txData and explainTx to support diff instructions

jonathancn-bitgo:
PR Scope for supporting multiple transaction instructions which are tokenTransfer and tokenAssociate: 
1. Updating `txData` in `account-lib/iface.ts` by adding `instructionData` field for and adding `type` interfaces for each transaction.
2. Updating `toJson()` + `loadinputsandoutputs()` in `account-lib/transaction.ts` to parse recipients or tokens into the `instructionData` field.
3. Updating `explainTransaction` in `bitgo/hbar.ts` module to update outputs and outputAmount for token transfer and associate.
4. Unit tests update for token transfer in `account-lib` and `bitgo` modules (unit-test for token associate will be in separate PR)

Ticket: BG-51280, BG-49606, BG-49600


modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -94,26 +103,41 @@ export class Transaction extends BaseTransaction {
...
+        const transferData = this.getTransferData();
+        result.instructionsData = {
+          type: HederaTransactionTypes.Transfer,
+          params: transferData,
+        };
+        result.to = result.instructionsData.params.recipients[0].address;
+        result.amount = result.instructionsData.params.recipients[0].amount;
+        break;
+      case HederaTransactionTypes.CreateAccount:
+        break;
```
joey-bai, line 117: do we need this?

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -94,26 +103,41 @@ export class Transaction extends BaseTransaction {
...
+      case HederaTransactionTypes.TokenAssociateToAccount:
+        result.instructionsData = {
+          type: HederaTransactionTypes.TokenAssociateToAccount,
+          params: this.getAccountAssociateData(),
+        };
+        break;
+      default:
+        throw new InvalidTransactionError('Invalid transaction ' + this._txBody.data);
```
joey-bai, line 125: lets make this a no-op instead to match previous behavior

modules/account-lib/src/coin/hbar/transaction.ts
```diff
@@ -189,23 +226,42 @@ export class Transaction extends BaseTransaction {
...
+          value: totalAmount.toString(),
+          coin: instruction.params.tokenName || this._coinConfig.name,
+        });
+        break;
+
+      case HederaTransactionTypes.TokenAssociateToAccount:
+        instruction.params.tokens.map((tokenName) => {
+          const tokenEntry: Entry = {
+            address: instruction.params.accountId,
+            value: TRANSFER_AMOUNT_UNKNOWN_TEXT,
```
joey-bai, line 255: shouldn't we be able to get this from the `txJson`?

jonathancn-bitgo: It's from txJson but in txJson `instructionData`, it has different format from tokenEntry, so this is just adjustment for the tokenEntry interface

modules/account-lib/src/coin/hbar/transaction.ts
joey-bai: yep

modules/account-lib/src/coin/hbar/transaction.ts
joey-bai: i meant just the `value` amount

jonathancn-bitgo: currently `instructionData` for `token associate` has `accountId` for owner and `tokens` which is list of `tokenName`s
I think adding an unknown value of amount in the `tokens` will be weird?

joey-bai: ohhhh this is for token association. lets use 0 instead?

joey-bai (approved):

joey-bai (dismissed):

modules/bitgo/src/v2/coins/hbar.ts
```diff
@@ -221,9 +222,48 @@ export class Hbar extends BaseCoin {
...
+    // TODO(BG-24809): get the memo from the toJson
+    let memo = '';
+    if (params.memo) {
+      memo = params.memo.value;
+    }
+
+    switch (txJson.instructionsData.type) {
+      case 'cryptoTransfer':
+        const recipients = txJson.instructionsData.params.recipients || [];
+        recipients.map((recipient) => {
```
bitgoAaron, line 236: did you mean to use `forEach` here? `map` should return something. Can we clean this up and either use 
```
outputs = recipients.map(...)
```
or 
````
recipients.forEach(...)

bitgoAaron (changes requested):
can we please clean up map

jonathancn-bitgo:
> can we please clean up map

Updated all maps that don't return

bitgoAaron (approved):

PR #2500, opened 2022-07-12
chore(bitgo): refactor tss utils

sachushaji:
## Summary
This change refactors tss utils for reusing common functionalities for different signature schemes. The major changes can be summarised as follows

## Changes
1. Introduced a new `BaseTssUtils` class which each signature scheme has to inherit. This class has implementations which can be reused for different schems
2. Introduced baseType and common files which will have methods and types which can be reused for every scheme
3. Introduced scheme specific naming to avoid confusions and added relevant deprecation flags to prompt people to use scheme specific names (eg TssUtils --> EDDSAUtils)
4. All changes are backward compatible

Will be porting ECDSA to use this in a separate PR for ease of review

TICKET: BG-52194

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
@@ -1,0 +1,140 @@
+import { IRequestTracer } from '../../../api';
+import { SerializedKeyPair } from 'openpgp';
+import { KeychainsTriplet, IBaseCoin } from '../../baseCoin';
+import { BitGoBase } from '../../bitgoBase';
+import { Keychain } from '../../keychain';
+import { getTxRequest } from '../../tss';
+import { IWallet } from '../../wallet';
+import { MpcUtils } from '../mpcUtils';
+import * as _ from 'lodash';
+import {
+  ITssUtils,
+  PrebuildTransactionWithIntentOptions,
+  SignatureShareRecord,
+  TxRequest,
+  TxRequestVersion,
+} from './baseTypes';
+
+/**
+ * BaseTssUtil class which different signature schemes have to extend
+ */
+export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtils<KeyShare> {
+  private _wallet?: IWallet;
+
+  constructor(bitgo: BitGoBase, baseCoin: IBaseCoin, wallet?: IWallet) {
+    super(bitgo, baseCoin);
+    this._wallet = wallet;
+  }
+
+  get wallet(): IWallet {
+    if (_.isNil(this._wallet)) {
+      throw new Error('Wallet not defined');
+    }
+    return this._wallet;
+  }
+
+  createUserKeychain(
+    userGpgKey: SerializedKeyPair<string>,
+    userKeyShare: KeyShare,
+    backupKeyShare: KeyShare,
+    bitgoKeychain: Keychain,
+    passphrase: string,
+    originalPasscodeEncryptionCode: string,
+    recipientIndex?: number | undefined
+  ): Promise<Keychain> {
+    throw new Error('Method not implemented.');
+  }
+
+  createBackupKeychain(
+    userGpgKey: SerializedKeyPair<string>,
+    userKeyShare: KeyShare,
+    backupKeyShare: KeyShare,
+    bitgoKeychain: Keychain,
+    passphrase: string
+  ): Promise<Keychain> {
+    throw new Error('Method not implemented.');
+  }
+
+  createBitgoKeychain(
+    userGpgKey: SerializedKeyPair<string>,
+    userKeyShare: KeyShare,
+    backupKeyShare: KeyShare,
+    enterprise: string
+  ): Promise<Keychain> {
+    throw new Error('Method not implemented.');
+  }
+
+  createKeychains(params: {
+    passphrase: string;
+    enterprise?: string | undefined;
+    originalPasscodeEncryptionCode?: string | undefined;
+  }): Promise<KeychainsTriplet> {
+    throw new Error('Method not implemented.');
+  }
+
+  signTxRequest(params: { txRequest: string | TxRequest; prv: string; reqId: IRequestTracer }): Promise<TxRequest> {
+    throw new Error('Method not implemented.');
+  }
+
+  prebuildTxWithIntent(
+    params: PrebuildTransactionWithIntentOptions,
+    apiVersion?: TxRequestVersion | undefined,
+    preview?: boolean | undefined
+  ): Promise<TxRequest> {
+    throw new Error('Method not implemented.');
+  }
+
+  /**
+   * Call delete signature shares for a txRequest, the endpoint delete the signatures and return them
+   *
+   * @param {string} txRequestId tx id reference to delete signature shares
+   * @returns {SignatureShareRecord[]}
+   */
+  async deleteSignatureShares(txRequestId: string): Promise<SignatureShareRecord[]> {
+    return this.bitgo
+      .del(this.bitgo.url(`/wallet/${this.wallet.id()}/txrequests/${txRequestId}/signatureshares`, 2))
+      .send()
+      .result();
+  }
+
+  /**
+   * Initialize the send procedure once Bitgo has the User To Bitgo GShare
+   *
+   * @param {String} txRequestId - the txRequest Id
+   * @returns {Promise<any>}
+   */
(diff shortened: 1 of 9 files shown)


modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -0,0 +1,120 @@
...
+export type TransactionState =
+  | 'initialized'
+  | 'pendingSignature'
+  | 'signed'
+  | 'held'
+  | 'delivered'
+  | 'invalidSignature'
+  | 'rejected';
+
+export type ITxRequest = {
```
joey-bai, line 43: if we're moving models around, i think `sdk-api` might be a better place for this + related model. this is the API model for the /txrequests resource from wallet platform

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -0,0 +1,120 @@
...
+  date: string;
+  userId: string;
+  intent: any;
+  pendingApprovalId?: string;
+  policiesChecked: boolean;
+  signatureShares?: SignatureShareRecord[];
+  pendingTxHashes?: string[];
+  txHashes?: string[];
+  // Only available in 'lite' version
+  unsignedTxs: unknown[];
```
joey-bai, line 59: why'd we change this to `unknown`?

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -0,0 +1,120 @@
...
+  // Only available in 'full' version
+  transactions: {
+    state: TransactionState;
+    unsignedTx: unknown;
```
joey-bai, line 63: same

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -0,0 +1,120 @@
...
+  BITGO = 'bitgo',
+}
+
+export interface SignatureShareRecord {
+  from: SignatureShareType;
+  to: SignatureShareType;
+  share: string;
+}
+
+export interface ITssUtils<KeyShare, TxRequest> {
```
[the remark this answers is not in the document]
sachushaji: Yeah, Because for different signature schemes or blockchains, the `unsignedTxs` /  `transactions`  `intents` will be different and seems best to have it overridden at the time of initialisation

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
sachushaji: This would be the base type for creating TxRequest object and seems like the unsignedTx would be specific to the signature scheme being used, By using unknown, we force people to define and override with a type which makes sense for the signature scheme in hand

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
```diff
@@ -0,0 +1,132 @@
...
+import { MpcUtils } from '../mpcUtils';
+import * as _ from 'lodash';
+import { ITssUtils, PrebuildTransactionWithIntentOptions, SignatureShareRecord, TxRequestVersion } from './baseTypes';
+
+/**
+ * BaseTssUtil class which different signature schemes have to extend
+ */
+export default class BaseTssUtils<KeyShare, TxRequest>
+  extends MpcUtils
+  // eslint-disable-next-line prettier/prettier
```
sachushaji, line 17: Strange eslint error to give a new line but after giving a newline, it prompts to revert back to the old and gets the same error again, so disabled eslint check

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
sachushaji: Only the model?, when I quickly checked, the types defined seem to be also used in the sdk-api, so probably have to move the methods also if we need to do this?

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -0,0 +1,132 @@
...
+    reqId: IRequestTracer;
+  }): Promise<TxRequestType>;
+  prebuildTxWithIntent(
+    params: PrebuildTransactionWithIntentOptions,
+    apiVersion?: TxRequestVersion,
+    preview?: boolean
+  ): Promise<TxRequestType>;
+  deleteSignatureShares(txRequestId: string): Promise<SignatureShareRecord[]>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  sendTxRequest(txRequestId: string): Promise<any>;
```
sachushaji, line 129: Linter started complaining for the use of `any`, would have preferred to use `unknown` but have a feeling that this change might break stuff which is already using this functions. Disabled check.

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -0,0 +1,132 @@
...
+  to: SignatureShareType;
+  share: string;
+}
+
+/**
+ * Common Interface for implementing signature scheme specific
+ * util functions
+ */
+export interface ITssUtils<KeyShare = EDDSA.KeyShare, TxRequestType = EddsaTxRequest> {
+  createUserKeychain(
```
olibeyene, line 90: ecdsa has createParticipantKeychain, can we also add that here or have that in the following pr?

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -0,0 +1,132 @@
...
+  from: SignatureShareType;
+  to: SignatureShareType;
+  share: string;
+}
+
+/**
+ * Common Interface for implementing signature scheme specific
+ * util functions
+ */
+export interface ITssUtils<KeyShare = EDDSA.KeyShare, TxRequestType = EddsaTxRequest> {
```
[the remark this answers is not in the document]
sachushaji: yes, Made to make it backward compatible, We can override it for ECDSA

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -0,0 +1,132 @@
...
+    bitgoKeychain: Keychain,
+    passphrase: string
+  ): Promise<Keychain>;
+  createBitgoKeychain(
+    userGpgKey: SerializedKeyPair<string>,
+    userKeyShare: KeyShare,
+    backupKeyShare: KeyShare,
+    enterprise: string
+  ): Promise<Keychain>;
+  createKeychains(params: {
```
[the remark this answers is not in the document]
sachushaji: Tried that, but seems like since the parameters are in a json object, need to give a dummy name it seems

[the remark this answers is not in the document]
olibeyene: if you deconstruct the object, they can be just params and they don't have to be in a param object.

olibeyene (approved):

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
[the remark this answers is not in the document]
olibeyene: `createKeychains(passphrase: string, enterprise?: string, originalPasscodeEncryptionCode?: string): Promise<KeychainsTriplet>`

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -0,0 +1,146 @@
...
+export type TxRequest = {
+  txRequestId: string;
+  walletId: string;
+  walletType: WalletType;
+  version: number;
+  enterpriseId?: string;
+  state: TxRequestState;
+  date: string;
+  userId: string;
+  intent: unknown; // Should override with sig scheme specific intent
```
sachushaji, line 73: More safe to use `unknown` as it will prompt people to do proper typechecing before use

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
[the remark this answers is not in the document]
sachushaji: TxRequest has been removed from parameters based on our discussion over slack

joey-bai (approved):

olibeyene (approved):

mmcshinsky-bitgo (approved):

PR #2502, opened 2022-07-12
feat(account-lib): token associate unit test

jonathancn-bitgo:


joey-bai (approved):
thanks for following up with this!

mmcshinsky-bitgo (approved):

PR #2506, opened 2022-07-13
chore(sdk-coin-trx): refactor trx to its own module

mmcshinsky-bitgo:
BG-48136


modules/account-lib/package.json
```diff
@@ -43,21 +41,23 @@
...
     "@bitgo/sdk-coin-stx": "^1.0.0",
+    "@bitgo/sdk-coin-trx": "^1.0.0",
     "@bitgo/sdk-coin-xtz": "^1.0.0",
     "@bitgo/sdk-core": "^1.0.1",
     "@bitgo/statics": "^6.17.0",
     "@ethereumjs/common": "^2.4.0",
     "@hashgraph/sdk": "~2.3.0",
     "@polkadot/api": "7.15.1",
     "@stablelib/hex": "^1.0.0",
+    "@solana/spl-token": "0.1.8",
```
bitgoAaron, line 52: are these still used in account-lib? I just removed these dependencies since SOL was the only coin referencing them...

mmcshinsky-bitgo: Probably from the rebase. I'll fix that

modules/sdk-coin-trx/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "nyc": {
+    "extension": [
+      ".ts"
+    ],
+    "include": [
+      "src/**/*.ts"
+    ]
+  },
+  "dependencies": {
+    "@bitgo/abstract-eth": "^1.0.0",
```
bitgoAaron, line 50: don't think `@bitgo/abstract-eth` is actually used

modules/sdk-coin-trx/src/lib/contractCallBuilder.ts
```diff
@@ -1,7 +1,7 @@
 import { createHash } from 'crypto';
 import { BaseCoin as CoinConfig, TronNetwork } from '@bitgo/statics/';
```
bitgoAaron, line 2: delete trailing slash from `@bitgo/statics/`

bitgoAaron (changes requested):
couple of NITs to cleanup imports, then it'll be good to go.

bitgoAaron (approved):
LGTM

PR #2507, opened 2022-07-13
fix(statics): update avaxp network to use number instead of bigint

jicli:
bigint cannot be parsed when download transaction
Ticket: STLX-17918

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -35,21 +35,20 @@ export interface AvalancheNetwork extends BaseNetwork {
   readonly networkID: number;
   readonly hrp: string;
   readonly vm: string;
-  readonly creationTxFee: bigint;
-  readonly createSubnetTx: bigint;
-  readonly createChainTx: bigint;
-  readonly minConsumption: number;
-  readonly maxConsumption: number;
-  readonly maxStakingDuration: bigint;
-  readonly maxSupply: bigint;
-  readonly minStake: bigint;
-  readonly minStakeDuration: number;
-  readonly maxStakeDuration: number;
-  readonly minDelegationStake: bigint;
-  readonly minDelegationFee: bigint;
+  readonly creationTxFee: string;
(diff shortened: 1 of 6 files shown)


modules/statics/src/networks.ts
```diff
@@ -169,19 +169,19 @@ class AvalancheP extends Mainnet implements AvalancheNetwork {
...
+  createSubnetTx = 1000000000; // 1 AVAX
+  createChainTx = 1000000000; // 1 AVAX
+  creationTxFee = 10000000; // 1 CENTIAVAX
   minConsumption = 0.1;
   maxConsumption = 0.12;
-  maxStakingDuration = BigInt(31536000); // 1 year
-  maxSupply = BigInt(720000000) * BigInt(1000000000); // 720 mil tokens
-  minStake = BigInt(1000000000) * BigInt(2000);
+  maxStakingDuration = 31536000; // 1 year
+  maxSupply = 720000000 * 1000000000; // 720 mil tokens
```
joey-bai, line 179: this exceeds JS number overflow limit and will likely cause you problems if you try to do math with it.
this is `7.2 * 10^17`
safe overflow limit for JS is ~ `9 * 10^15`

joey-bai: alternative is to make this a string instead, and call it `maxSupplyString`

joey-bai: when you consume it, you can use any BigNumber/BigInt library to safely do math on it.

jicli: updated it to string

joey-bai (approved):
LGTM - will defer to SDK team on usage of string instead of bigint

tylerlevine (changes requested):
nit: do not include the type in the name, that information is already in the type

Commit does need to identify this as a breaking change though, since you are making a change that could break calling code.

tylerlevine (approved):

PR #2517, opened 2022-07-14
chore(sdk-coin-xlm): move XLM to its own module

bitgoAaron:
BG-51039


modules/sdk-coin-xlm/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "@bitgo/statics": "^6.17.0",
+    "@bitgo/utxo-lib": "^2.2.2",
+    "bignumber.js": "^8.0.1",
+    "lodash": "^4.17.14",
+    "stellar-sdk": "^10.0.1",
+    "superagent": "^3.8.3"
+  },
+  "devDependencies": {
+    "@bitgo/sdk-api": "^1.0.1",
+    "@bitgo/sdk-test": "^1.0.0"
```
mmcshinsky-bitgo, line 54: Are we doing @types/lodash in devDependencies for these still or relying on root?

mmcshinsky-bitgo (approved):

modules/sdk-coin-xlm/package.json
bitgoAaron: since it's specified in root we don't need to specify it here. We should clean up all the redundant dependencies as part of https://bitgoinc.atlassian.net/browse/BG-51022

PR #2518, opened 2022-07-14
feat(sdk-coin-eth): add util for eth address from compressed point

sachushaji:
this change adds a util function to create a eth
address from a compressed point

TICKET: BG-52320


sachushaji:
we already have a method in KeyPar class for it, closing.

PR #2524, opened 2022-07-15
chore(prime): add GBP

drew-bitgo:
TICKET: BG-52392


modules/bitgo/src/v2/coinFactory.ts
```diff
@@ -27,6 +27,7 @@ import {
   Eth,
   Eth2,
   FiatEur,
+  FiatGBP,
```
mmcshinsky-bitgo, line 30: nit: any reason it's `GBP` as opposed to `Gbp` like `Eur` and `Usd`?

drew-bitgo: I decided to do it this was since GBP is an abbreviation and not a shorting (like Eur for Euro)

mmcshinsky-bitgo (approved):

PR #2529, opened 2022-07-15
chore(bitgo): improve docker build caching

dpc:
## Bench

This cuts the build time (when I try it locally) from:

```
Executed in  293.18 secs      fish           external
   usr time  349.22 millis  446.00 micros  348.77 millis
   sys time  174.63 millis  239.00 micros  174.39 millis
```

to:

```
Executed in  116.04 secs      fish           external
   usr time  348.85 millis  688.00 micros  348.16 millis
   sys time  147.11 millis    0.00 micros  147.11 millis
```

when the cache is warm (no dependency changes, just a typical source code change).

Ticket: BG-00000

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,28 @@
-FROM node:16-alpine@sha256:72a490e7ed8aed68e16b8dc8f37b5bcc35c5b5c56ee3256effcdee63e2546f93 AS builder
+
+# An elaborated scheme to build all the dependencies of all packages first in a cached layer
+# https://stackoverflow.com/a/63142468/134409
+# https://medium.com/@emilefugulin/building-a-sane-docker-image-for-typescript-lerna-and-prisma-2-76d8ff9926e4
+FROM node:16-alpine@sha256:72a490e7ed8aed68e16b8dc8f37b5bcc35c5b5c56ee3256effcdee63e2546f93 AS filter-packages-json
 MAINTAINER Tyler Levine <tyler@bitgo.com>
+
+COPY package.json yarn.lock lerna.json ./
+WORKDIR /tmp/bitgo
+COPY package.json yarn.lock lerna.json ./
+COPY modules ./modules
+# delete all the non package.json files under `./modules/`
+RUN find modules \! -name "package.json" -mindepth 2 -maxdepth 2 -print | xargs rm -rf
+
+FROM node:16-alpine@sha256:72a490e7ed8aed68e16b8dc8f37b5bcc35c5b5c56ee3256effcdee63e2546f93 AS builder
 RUN apk add --no-cache git python3 make g++ libtool autoconf automake
 WORKDIR /tmp/bitgo
+COPY --from=filter-packages-json /tmp/bitgo .
+# (skip postinstall) https://github.com/yarnpkg/yarn/issues/4100#issuecomment-388944260
+RUN NOYARNPOSTINSTALL=1 yarn install --pure-lockfile
+
 COPY . .
 RUN \
     # clean up unnecessary local node_modules and dist
-    rm -rf **/node_modules **/dist && \ 
+    rm -rf modules/**/node_modules modules/**/dist && \
     # install with dev deps so we can run the prepare script
     yarn install --frozen-lockfile && \
     # install again to prune dev deps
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,11 +1,28 @@
...
 WORKDIR /tmp/bitgo
+COPY --from=filter-packages-json /tmp/bitgo .
+# (skip postinstall) https://github.com/yarnpkg/yarn/issues/4100#issuecomment-388944260
+RUN NOYARNPOSTINSTALL=1 yarn install --pure-lockfile
+
 COPY . .
 RUN \
     # clean up unnecessary local node_modules and dist
-    rm -rf **/node_modules **/dist && \ 
+    rm -rf modules/**/node_modules modules/**/dist && \
```
dpc, line 25: This seems to be throwing away the pre-build dependencies. What was the exact purpose of it?

dpc:
This would at least work in our internal CI and locally, but maybe we can enable it for Github Actions as well:

https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching


```
     # This is a separate action that sets up buildx runner
      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v1
      # So now you can use Actions' own caching!
      - name: Cache Docker layers
        uses: actions/cache@v2
        with:
          path: /tmp/.buildx-cache
          key: ${{ runner.os }}-buildx-${{ github.sha }}
          restore-keys: |
            ${{ runner.os }}-buildx-
      - name: Login to DockerHub
        uses: docker/login-action@v1
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_TOKEN }}
      # And make it available for builds
      - name: Build and push
        uses: docker/build-push-action@v2
        with:
          context: .
          push: false
          tags: user/app:latest
          cache-from: type=local,src=/tmp/.buildx-cache
          cache-to: type=local,dest=/tmp/.buildx-cache-new
        # This ugly bit is necessary if you don't want your cache to grow forever
        # until it hits GitHub's limit of 5GB.
        # Temp fix
        # https://github.com/docker/build-push-action/issues/252
        # https://github.com/moby/buildkit/issues/1896
      - name: Move cache
        run: |
          rm -rf /tmp/.buildx-cache
          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
```

marktoda (approved):

mmcshinsky-bitgo (changes requested):
Missing ticket number in commit. Commit message description will show up in published public release notes and should be modified to reflect such.

mmcshinsky-bitgo (approved):

dpc:
Rebased because something completely unrelated was failing.

Dockerfile
```diff
@@ -1,11 +1,28 @@
...
+COPY modules ./modules
+# delete all the non package.json files under `./modules/`
+RUN find modules \! -name "package.json" -mindepth 2 -maxdepth 2 -print | xargs rm -rf
+
+FROM node:16-alpine@sha256:72a490e7ed8aed68e16b8dc8f37b5bcc35c5b5c56ee3256effcdee63e2546f93 AS builder
 RUN apk add --no-cache git python3 make g++ libtool autoconf automake
 WORKDIR /tmp/bitgo
+COPY --from=filter-packages-json /tmp/bitgo .
+# (skip postinstall) https://github.com/yarnpkg/yarn/issues/4100#issuecomment-388944260
+RUN NOYARNPOSTINSTALL=1 yarn install --pure-lockfile
```
bitgoAaron, line 20: wouldn't you want to do this in the `filter-packages-json` step? What's the point of copying that content in a separate layer?

dpc: I'm like 99% sure, but basically that is the root of the trick: Docker caching is FS content based. In `filter-packages-json` I have to copy whole `./modules` which will invalidate `filter-packages-json` and thus re-run it very often. But this step is very fast, and deletes most of the files. So by having a new docker build image, I can copy only the filtered files, which usually will not change. Thus this layer will not be invalidated, even if `filter-packages-json` was. So everything here will be cached and not rebuild 99% of the time.

dpc:
Rebased again and it's finally green.

bitgoAaron (approved):

PR #2536, opened 2022-07-18
fix(express): allow account consolidations

derickson-bitgo:
Use coin.allowsAccountConsolidations instead of coin family to validate
the route for consolidating accounts.

BG-50574

### DIFF
--- a/modules/express/test/unit/clientRoutes/consolidateAccount.ts
+++ b/modules/express/test/unit/clientRoutes/consolidateAccount.ts
@@ -1,4 +1,3 @@
-import { CoinFamily } from '@bitgo/statics';
 import * as sinon from 'sinon';
 
 import 'should-http';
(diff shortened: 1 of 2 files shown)


modules/express/test/unit/clientRoutes/consolidateAccount.ts
```diff
@@ -89,11 +68,11 @@ describe('Consolidate account', () => {
       .should.be.rejectedWith('consolidate address must be an array of addresses');
   });
 
-  function createConsolidateMocks(res, coin = CoinFamily.ALGO) {
+  function createConsolidateMocks(res, allowsAccountConsolidations = true) {
```
mmcshinsky-bitgo, line 71:
```suggestion
  function createConsolidateMocks(res, allowsAccountConsolidations = false) {
```

mmcshinsky-bitgo: Since the default is false and true is an explicit action

derickson-bitgo: fixed

joey-bai (approved):

PR #2537, opened 2022-07-18
feat(sdk-core): add more ecdsa helper methods

olibeyene:
Ticket: BG-52582

### DIFF
--- a/modules/bitgo/test/v2/unit/tss/ecdsa.ts
+++ b/modules/bitgo/test/v2/unit/tss/ecdsa.ts
@@ -7,6 +7,7 @@ import {
   SignatureShareRecord,
   getTxRequest,
   ECDSAMethods,
+  SignatureShareType,
 } from '@bitgo/sdk-core';
 import * as openpgp from 'openpgp';
 import * as should from 'should';
@@ -73,7 +74,7 @@ describe('Ecdsa tss helper functions tests', function () {
     bitgoGpgKeypair = gpgKeypairs[2];
   });
 
-  describe('encryptNShare', function () {
+  describe('encryptNShare and decryptNShare', function () {
     it('should encrypt n shares foreach user', async function () {
       for (let i = 2; i <= 3; i++) {
         const encryptedNShare = await ECDSAMethods.encryptNShare(userKeyShare, i, bitgoGpgKeypair.publicKey, userGpgKeypair.privateKey);
@@ -90,6 +91,16 @@ describe('Ecdsa tss helper functions tests', function () {
       await encryptNShare(userKeyShare, 1, userGpgKeypair.privateKey, bitgoGpgKeypair.publicKey).should.be.rejectedWith('Invalid recipient');
       await encryptNShare(backupKeyShare, 2, userGpgKeypair.privateKey, bitgoGpgKeypair.publicKey).should.be.rejectedWith('Invalid recipient');
       await encryptNShare(bitgoKeyShare, 3, userGpgKeypair.privateKey, bitgoGpgKeypair.publicKey).should.be.rejectedWith('Invalid recipient');
+    });
+
+    it('should decrypt n share', async function() {
+      const encryptedNShare = await ECDSAMethods.encryptNShare(userKeyShare, 3, bitgoGpgKeypair.publicKey, userGpgKeypair.privateKey);
+      const decryptedNShare = await ECDSAMethods.decryptNShare({ nShare: encryptedNShare, recipientPrivateArmor: bitgoGpgKeypair.privateKey, senderPublicArmor: userGpgKeypair.publicKey });
+      decryptedNShare.i.should.equal(userKeyShare.nShares[3].i);
+      decryptedNShare.j.should.equal(userKeyShare.nShares[3].j);
+      decryptedNShare.n.should.equal(userKeyShare.nShares[3].n);
+      decryptedNShare.u.should.equal(userKeyShare.nShares[3].u);
+      decryptedNShare.y.should.equal(userKeyShare.nShares[3].y);
     });
   });
 
@@ -219,7 +230,6 @@ describe('Ecdsa tss helper functions tests', function () {
     });
   });
 
-
   describe('tss signing helper function', async function () {
     const bitgo = TestBitGo.decorate(BitGo, { env: 'mock' });
     bitgo.initializeTestVars();
@@ -299,7 +309,6 @@ describe('Ecdsa tss helper functions tests', function () {
         const userShare = await ECDSAMethods.createUserOmicronAndDeltaShare(omicronAndDeltaShareCreationParams.gShare);
         userShare.should.have.properties(['dShare', 'oShare']);
         const { dShare, oShare } = userShare;
-       
         dShare?.i?.should.equal(3);
         dShare?.j?.should.equal(1);
         dShare?.should.have.property('delta').and.be.a.String();
@@ -356,13 +365,12 @@ describe('Ecdsa tss helper functions tests', function () {
             const mockSendReq = { from: 'user', to: 'bitgo', share: config[index].mockShareToSendString } as SignatureShareRecord;
             const invalidSignatureShare = { from: 'bitgo', to: 'user', share: JSON.stringify(config[index].incorrectReceivedShareString) } as SignatureShareRecord;
             const nock = await nockSendSignatureShare({ walletId: wallet.id(), txRequestId: txRequest.txRequestId, signatureShare: mockSendReq, response: invalidSignatureShare }, 200);
-            await ECDSAMethods.sendShareToBitgo(bitgo, wallet.id(), txRequest.txRequestId, config[index].sendType, config[index].mockShareToSend, config[index].optionalDShare).should.be.rejectedWith(`Invalid ${config[index].shareReceived} from Bitgo`);
+            await ECDSAMethods.sendShareToBitgo(bitgo, wallet.id(), txRequest.txRequestId, config[index].sendType, config[index].mockShareToSend, config[index].optionalDShare).should.be.rejectedWith(/Excepted share length to be greater than or equal .*/g);
             nock.isDone().should.equal(true);
           });
         });
       }
     });
-
 
     describe('getTxRequest:', async function () {
       it('should succeed to get txRequest by id', async function () {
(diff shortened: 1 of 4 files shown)


sachushaji (changes requested):
Thanks @olibeyene for taking this up, This will help remove some redundant code across different services. I had done some updates on how the shares are parsed [here](https://github.com/BitGo/BitGoJS/pull/2469) , can you update it?, also kindly add some unit test cases for the same. You can use mock data from ecdsaFixtures

wesgraham (approved):
LGTM once unit tests added

wesgraham (approved):
awesome!

modules/bitgo/test/v2/unit/tss/ecdsa.ts
```diff
@@ -382,4 +391,140 @@ describe('Ecdsa tss helper functions tests', function () {
...
+    let userSignShare: ECDSAMethodTypes.SignShare;
+    let bitgoSignShare: ECDSA.SignConvertRT;
+    let userGammaAndMuShare: ECDSAMethodTypes.CreateUserGammaAndMuShareRT;
+    let userOmicronAndDeltaShare: ECDSAMethodTypes.CreateUserOmicronAndDeltaShareRT;
+    let bitgoOmicronAndDeltaShare: ECDSAMethodTypes.CreateUserOmicronAndDeltaShareRT;
+    let userSignatureShare: ECDSAMethodTypes.SignatureShare;
+    before(async function() {
+      const txRequest = {
+        txRequestId: 'randomId',
+        unsignedTxs: [{ signableHex: 'TOO MANY SECRETS', serializedTxHex: 'randomhex2' }],
```
sachushaji, line 404: i think we can use ecdsaFixtures for this instead of doing this before hook, fixtures already have mocks for most shares and signing is a bit expensive to perform

modules/bitgo/test/v2/unit/tss/ecdsa.ts
```diff
@@ -382,4 +391,140 @@ describe('Ecdsa tss helper functions tests', function () {
...
+          j: 3,
+        },
+      });
+      userOmicronAndDeltaShare = await ECDSAMethods.createUserOmicronAndDeltaShare(userGammaAndMuShare.gShare!);
+
+      userSignatureShare = await ECDSAMethods.createUserSignatureShare(userOmicronAndDeltaShare.oShare, bitgoOmicronAndDeltaShare.dShare, signablePayload);
+    });
+
+    it('should successfully parse K share', function () {
+      const bitgoKShare = userSignShare.kShare;
```
sachushaji, line 440: These tests looks nice, Probably we can add a addition check on the parsers to make sure the share has sufficient length and to throw an error if it doesn't meet the condition also

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -295,3 +282,93 @@ export async function encryptNShare(
...
+  };
+
+  return nShare;
+}
+
+export function getPublicKey(commonKeyChain: string): string {
+  return commonKeyChain.slice(0, 66);
+}
+
+export function parseKShare(share: SignatureShareRecord): KShare {
```
sachushaji, line 309: Lets add a condition for all these parsers to throw an error if it is not meeting the minimum length for the share

modules/sdk-core/src/bitgo/tss/ecdsa/types.ts
```diff
@@ -16,9 +17,12 @@ export type SignatureShare = ECDSA.SignRT;
 export type Signature = ECDSA.Signature;
 
 export enum SendShareType {
+  AShare = 'AShare',
```
sachushaji, line 20: This interface is actually for shares which we are sending to bitgo (ie: K, Mu and SS Shares), may be we can rename it to make it more apparent

sachushaji (changes requested):
Really nice work, Small changes, also replace the core signing code and test making use of your cool parsers instead of the current logic used inline

modules/sdk-core/src/bitgo/tss/ecdsa/types.ts
olibeyene: it makes sense to keep this type only send to bitgo types. reverting this change.

modules/bitgo/test/v2/unit/tss/ecdsa.ts
```diff
@@ -380,6 +388,169 @@ describe('Ecdsa tss helper functions tests', function () {
...
+        const signature = ECDSAMethods.parseSignatureShare(share);
+        signature.i.should.equal(1);
+        signature.r.should.equal(mockSignature.r);
+        signature.s.should.equal(mockSignature.s);
+        signature.y.should.equal(mockSignature.y);
+      });
+
+      it('should successfully convert signature share to signarture share record', function () {
+        const share = {
+          to: SignatureShareType.USER,
```
sachushaji, line 545: NIT: Lets add a test for the error cases also

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -196,44 +183,21 @@ export async function sendShareToBitgo(
...
-  };
-
+  let signatureShare: SignatureShareRecord;
   let responseFromBitgo: SendShareToBitgoRT;
 
   switch (shareType) {
     case SendShareType.KShare:
       const kShare = share as KShare;
-      signatureShare.share = kShare.k + kShare.n;
+      signatureShare = convertKShare(kShare);
```
sachushaji, line 192: Looks much cleaner, Nice

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -270,7 +226,7 @@ export async function sendShareToBitgo(
  * @param recipientIndex - index of the recipient (1, 2, or 3)
  * @param recipientGpgPublicArmor - recipient's public gpg key in armor format
  * @param senderGpgPrivateArmor - sender's private gpg key in armor format
- * @returns { EncryptedNShare } encrypted Y Share
+ * @returns { EncryptedNShare } encrypted N Share
```
sachushaji, line 229: NIT: may be we can avoid specifying type here since rest of the doc is not using it any ways, if we are specifying, then correct type should be `{ Promise<EncryptedNShare> }`

modules/bitgo/test/v2/unit/tss/ecdsa.ts
```diff
@@ -356,14 +365,13 @@ describe('Ecdsa tss helper functions tests', function () {
             const mockSendReq = { from: 'user', to: 'bitgo', share: config[index].mockShareToSendString } as SignatureShareRecord;
             const invalidSignatureShare = { from: 'bitgo', to: 'user', share: JSON.stringify(config[index].incorrectReceivedShareString) } as SignatureShareRecord;
             const nock = await nockSendSignatureShare({ walletId: wallet.id(), txRequestId: txRequest.txRequestId, signatureShare: mockSendReq, response: invalidSignatureShare }, 200);
-            await ECDSAMethods.sendShareToBitgo(bitgo, wallet.id(), txRequest.txRequestId, config[index].sendType, config[index].mockShareToSend, config[index].optionalDShare).should.be.rejectedWith(`Invalid ${config[index].shareReceived} from Bitgo`);
+            await ECDSAMethods.sendShareToBitgo(bitgo, wallet.id(), txRequest.txRequestId, config[index].sendType, config[index].mockShareToSend, config[index].optionalDShare).should.be.rejectedWith(/Excepted share length to be greater than or equal .*/g);
```
sachushaji, line 368: We are not getting >= which length here in the error?

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -295,3 +251,263 @@ export async function encryptNShare(
...
+  }
+}
+
+/**
+ * parses K share from signature share record
+ * @param share - signature share record
+ * @returns K Share
+ */
+export function parseKShare(share: SignatureShareRecord): KShare {
+  validateShare(share, 2304);
```
sachushaji, line 316: N may be dynamic, let check for >1536

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -295,3 +251,263 @@ export async function encryptNShare(
...
+  };
+}
+
+/**
+ * parses A share from signature share record
+ * @param share - signature share record
+ * @returns A Share
+ */
+export function parseAShare(share: SignatureShareRecord): AShare {
+  validateShare(share, 5376, false);
```
sachushaji, line 345: n may be dynamic, lets check for > 4608

sachushaji (commented):
Small NITS, Thanks for the updates

modules/bitgo/test/v2/unit/tss/ecdsa.ts
olibeyene: since the length is different for each share, i'm using regex to get the general error part.

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
olibeyene: yep, validateShare will check for only if it's greater. I will need to pass in false.

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
olibeyene: the false param is making validateShare to only check for greater.

modules/bitgo/test/v2/unit/tss/ecdsa.ts
olibeyene: I will have follow up pr with this kind of test

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -295,3 +251,263 @@ export async function encryptNShare(
...
+  }
+}
+
+/**
+ * parses K share from signature share record
+ * @param share - signature share record
+ * @returns K Share
+ */
+export function parseKShare(share: SignatureShareRecord): KShare {
+  validateShare(share, 2304, false);
```
sachushaji, line 316: > 1536

sachushaji (approved):
LGTM, Thanks @olibeyene , Nice work

mmcshinsky-bitgo (approved):

PR #2539, opened 2022-07-18
feat(sdk-coin-avaxp): update stake validation check

jicli:
Ticket: STLX-18053

### DIFF
--- a/modules/sdk-coin-avaxp/src/lib/delegatorTxBuilder.ts
+++ b/modules/sdk-coin-avaxp/src/lib/delegatorTxBuilder.ts
@@ -56,8 +56,8 @@ export class DelegatorTxBuilder extends TransactionBuilder {
   }
 
   /**
-   *
-   * @param startTime
+   * start time of staking period
+   * @param value
    */
   startTime(value: string | number): this {
     this._startTime = new BN(value);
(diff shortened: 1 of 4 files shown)


modules/sdk-coin-avaxp/src/lib/delegatorTxBuilder.ts
```diff
@@ -124,6 +124,10 @@ export class DelegatorTxBuilder extends TransactionBuilder {
    *   unix time stamp based off seconds
    */
   validateStakeDuration(startTime: BN, endTime: BN): void {
+    const oneDayLater = new BN(Date.now()).add(new BN(86400));
```
mmcshinsky-bitgo, line 127: Wouldn't the validation need to be against an implemented DelegatorTxBuilder's startTime and not one built as a one off?

mmcshinsky-bitgo (changes requested):

modules/sdk-coin-avaxp/src/lib/delegatorTxBuilder.ts
mmcshinsky-bitgo:
```suggestion
    const oneDayLater = new BN(Date.now()).sub(new BN(86400));
```

mmcshinsky-bitgo:
Per disusssion, we're going to add a description to ``validateStakeDuration` to clarify against the ticket the purpose of this is validating the creation start time and the initial transaction building and not validating the signing after a 24 hour period in the future. Thanks @jicli!

mmcshinsky-bitgo (approved):

PR #2540, opened 2022-07-18
Bg 51022 remove all unused dependencies from package jsons fail builds instead of warn

mmcshinsky-bitgo:
Fix all dependencies in SDK with whitelists for some required dependency installs that aren't used explicitly.
Add tests to coins with no tests to fix github actions errors. 
Probably due to the sheer number of packages `parallel` was killing tests when all packages were updated. Removing `parallel` flag on `unit-test` for lerna fixed this.

BG-51022


modules/blake2b/test.js
```diff
@@ -1,116 +1,117 @@
...
-      var personal = v.personal.length === 0 ? null : hexWrite(new Uint8Array(v.personal.length / 2), v.personal)
-
-      var expected = Buffer.from(hexWrite(new Uint8Array(v.out.length / 2), v.out))
-      var actual = Buffer.from(blake2b(out.length, key, salt, personal, true).update(input).digest(out))
-
-      assert.deepEquals(actual, expected)
-    })
-    assert.end()
-  })
+    for (let i = 0; i < vectors.length; i++) {
```
agao14, line 15: Curious if there's a reason why we switched from `forEach` to a for loop?

agao14 (commented):
Left a quick question!

modules/blake2b/test.js
mmcshinsky-bitgo: At 576 vectors, `for` becomes faster than `forEach` in most environments.

bitgoAaron: interesting, I've never heard that. Do you have a source for that metric?

modules/sdk-coin-ada/package.json
```diff
@@ -43,11 +43,11 @@
   "dependencies": {
     "@bitgo/sdk-core": "^1.0.1",
     "@bitgo/statics": "^6.17.0",
-    "@emurgo/cardano-serialization-lib-browser": "^10.2.0",
```
bitgoAaron, line 46: this is used by webpack for module substitution, we're going to need to add an exception for this since the nodejs version will not work in the browser.

bitgoAaron (changes requested):
@emurgo/cardano-serialization-lib-browser will be required for cardano once WASM is approved & merged. Let's find a nice way to exclude webpack overrides like this first

modules/blake2b/test.js
mmcshinsky-bitgo: Here's something I did for it awhile ago. https://dev.to/mmcshinsky/benchmarking-javascript-loops-and-methods-part-1-19cl

It would be faster on multiple runs if it were being pre-optimized but each testing runs once as well is its' own one time run environment.

mmcshinsky-bitgo: When you get into object looping, it gets even worse: https://dev.to/mmcshinsky/benchmarking-javascript-loops-and-methods-part-2-2pfi

modules/sdk-coin-ada/package.json
mmcshinsky-bitgo: I'll add that in, thanks!

modules/blake2b/test.js
agao14: Interesting, thanks for clearing that up!

bitgoAaron (approved):

PR #2545, opened 2022-07-19
fix(sdk-coin-avaxp): fix avaxp validation issue

jicli:
updated node validation to measure the bytes length
update stake error message literal to reflect actual amount
Ticket: STLX-18165


mmcshinsky-bitgo (changes requested):
Missing ticket number in commit

mmcshinsky-bitgo (approved):

PR #2553, opened 2022-07-20
fix(bitgo): `Buffer` does not exist in browsers

dpc:
Instead of converting `BN` to `Buffer` eagerly,
just make it another case that we can return, and let
the code outside use it directly as they see fit.

Ticket: BG-00000

### DIFF
--- a/modules/sdk-coin-eth/src/lib/utils.ts
+++ b/modules/sdk-coin-eth/src/lib/utils.ts
@@ -567,7 +567,7 @@ export function hasSignature(txData: TxData): boolean {
   );
 }
 
-type RecursiveBufferOrString = string | Buffer | RecursiveBufferOrString[];
+type RecursiveBufferOrString = string | Buffer | BN | RecursiveBufferOrString[];
 
 /**
  * Get the raw data decoded for some types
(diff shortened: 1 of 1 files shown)


modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -579,7 +579,7 @@ type RecursiveBufferOrString = string | Buffer | RecursiveBufferOrString[];
 export function getRawDecoded(types: string[], serializedArgs: Buffer): RecursiveBufferOrString[] {
   function normalize(v: unknown, i: number): unknown {
     if (BN.isBN(v)) {
-      return v.toBuffer();
+      return v;
     } else if (typeof v === 'string' || Buffer.isBuffer(v)) {
```
joey-bai, line 583: can be combined into the same if block

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -579,7 +579,7 @@ type RecursiveBufferOrString = string | Buffer | RecursiveBufferOrString[];
...
       return v;
     } else if (Array.isArray(v)) {
```
jicli, line 585: are we confident that it will not fail on the following ifs?

dpc: No other problems were reported, so it seems like Avax is the only coin that somehow triggers decoding of something to `BN`.

modules/sdk-coin-eth/src/lib/utils.ts
mmcshinsky-bitgo: Will end users have to change their code as a result of this line change?

jicli: likely not, it's only change when we refactor and it's not working since then.

mmcshinsky-bitgo: Cool, wanted to make sure that we didn't need to label the commit as BREAKING CHANGE. :)

mmcshinsky-bitgo (approved):

PR #2557, opened 2022-07-21
fix: update AddWalletOptions

mmcshinsky-bitgo:
walletes.add was missing a required "label" field and optional fields
All defined parameters will now be pass to request if included

BG-51948

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallets.ts
@@ -65,23 +65,48 @@ export interface AcceptShareOptions {
 }
 
 export interface AddWalletOptions {
-  type?: string;
+  coinSpecific?: any;
+  enterprise?: string;
+  isCold?: IsCold;
+  isCustodial?: IsCustodial;
   keys?: string[];
+  keySignatures?: KeySignatures;
+  label: string;
+  multisigType?: 'onchain' | 'tss' | 'blsdkg';
+  address?: string;
   m?: number;
   n?: number;
   tags?: string[];
+  type?: string;
+  walletVersion?: number;
+  eip1559?: Eip1559;
   clientFlags?: string[];
-  isCold?: boolean;
-  isCustodial?: boolean;
-  address?: string;
+  // Additional params needed for xrp
   rootPub?: string;
+  // In XRP, XLM and CSPR this private key is used only for wallet creation purposes,
+  // once the wallet is initialized then we update its weight to 0 making it an invalid key.
+  // https://www.stellar.org/developers/guides/concepts/multi-sig.html#additional-signing-keys
   rootPrivateKey?: string;
   initializationTxs?: any;
   disableTransactionNotifications?: boolean;
   gasPrice?: number;
-  walletVersion?: number;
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -65,24 +65,49 @@ export interface AcceptShareOptions {
...
+  label: string;
+  multisigType?: 'onchain' | 'tss' | 'blsdkg';
+  address?: string;
   m?: number;
   n?: number;
   tags?: string[];
+  type?: string;
+  walletVersion?: number;
+  eip1559?: Eip1559;
   clientFlags?: string[];
```
mmcshinsky-bitgo, line 83: I've noticed that all parameters from `clientFlags` and down aren't actually specified on the API documentation.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -209,7 +175,7 @@ export class Wallets implements IWallets {
    */
   async generateWallet(params: GenerateWalletOptions = {}): Promise<WalletWithKeychains> {
     common.validateParams(params, ['label'], ['passphrase', 'userKey', 'backupXpub']);
-    if (!_.isString(params.label)) {
+    if (typeof params.label !== 'string') {
```
byhow, line 178: :nice:

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -95,9 +95,15 @@ export class Wallets implements IWallets {
    *    "n": number of keys available on the wallet (3)
    *    "keys": array of keychain ids
    */
-  async add(params: AddWalletOptions = {}): Promise<any> {
+  async add(params: AddWalletOptions): Promise<any> {
+    params = params || {};
```
byhow, line 99: nit: not sure if this is necessary since `param` is already declared non null in the function param

byhow (approved):
just some nit, LGTM

modules/sdk-core/src/bitgo/wallet/wallets.ts
mmcshinsky-bitgo: Without it, line 103 throws an typescript error

mmcshinsky-bitgo: I did make some changes after that. Lemme confirm.

modules/sdk-core/src/bitgo/wallet/iWallets.ts
bitgoAaron: That could be because the API documentation is out of date, or the implementation was removed. Best to lookup the route in wallet-platform & verify the params if we're unsure here.

bitgoAaron (approved):
Approved, but let's double check `clientFlags` in platform just to be sure

modules/sdk-core/src/bitgo/wallet/iWallets.ts
mmcshinsky-bitgo: Yup, they're undocumented deeply nested params.

PR #2559, opened 2022-07-21
fix(avaxp): use "P-" prefix in avaxp addresses

CapnMigraine:
Remove methods convert avaxp addresses by adding or removing the `P-` prefix.

Ticket: STLX-18215

### DIFF
--- a/modules/sdk-coin-avaxp/src/lib/utils.ts
+++ b/modules/sdk-coin-avaxp/src/lib/utils.ts
@@ -48,10 +48,7 @@ export class Utils implements BaseUtils {
    */
   /** @inheritdoc */
   isValidBlockId(hash: string): boolean {
-    if (hash.length !== 66 || hash.slice(0, 2) !== '0x') {
-      return false;
-    }
-    return (hash as string).match(/^[a-zA-Z0-9]+$/) !== null;
+    return this.binTools.isCB58(hash) && this.binTools.b58ToBuffer(hash).length === 36;
   }
 
   /**
--- a/modules/sdk-coin-avaxp/src/lib/utils.ts
+++ b/modules/sdk-coin-avaxp/src/lib/utils.ts
@@ -37,7 +37,7 @@ export class Utils implements BaseUtils {
    */
   /** @inheritdoc */
   isValidAddress(address: string): boolean {
-    return (address as string).match(/^[a-zA-Z0-9]+$/) !== null;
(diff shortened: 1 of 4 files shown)


modules/sdk-coin-avaxp/src/lib/utils.ts
```diff
@@ -37,7 +37,7 @@ export class Utils implements BaseUtils {
    */
   /** @inheritdoc */
   isValidAddress(address: string): boolean {
-    return (address as string).match(/^[a-zA-Z0-9]+$/) !== null;
+    return /^P-[a-zA-Z0-9]+$/.test(address);
```
jicli, line 40: https://github.com/BitGo/bitgo-microservices/pull/21198#discussion_r927758537

jicli: and also, it's not always start with `P-` cuz node uses it, can we do something like 
return `/^P-[a-zA-Z0-9]+$/.test(address); || /^NodeId-[a-zA-Z0-9]+$/.test(address)`

modules/sdk-coin-avaxp/test/unit/lib/utils.ts
```diff
@@ -3,23 +3,43 @@ import { Utils, KeyPair } from '../../../src/lib/';
...
+    it('should fail to validate invalid address', function () {
+      const address = 'fuji15jamwukfqkwhe8z26tjqxejtjd3jk9vj4kmxwa';
+      Utils.isValidAddress(address).should.be.false();
+    });
+
+    it('should validate an address', function () {
+      const address = 'P-fuji15jamwukfqkwhe8z26tjqxejtjd3jk9vj4kmxwa';
+      Utils.isValidAddress(address).should.be.true();
+    });
+
```
jicli, line 34: add a unit test for nodeid? thanks!

modules/sdk-coin-avaxp/src/lib/utils.ts
CapnMigraine: fixed

modules/sdk-coin-avaxp/test/unit/lib/utils.ts
CapnMigraine: done

jicli (approved):

mmcshinsky-bitgo (approved):

PR #2562, opened 2022-07-22
feat(statics): add new tokens

Joelsx:
add new tokens

BG-51006


mmcshinsky-bitgo:
@Joelsx  I don't see a breaking change in this code base. Can you elaborate?

Joelsx:
> @Joelsx I don't see a breaking change in this code base. Can you elaborate?

Yes. For the UNDERLYING Asset like SETHH = 'seth-h', should be SETH-H = 'seth-h', but we can put a caracter - for the underlying asset so we are paring the symbol seth-h with the close to be. It should not break anything and work fine. But If the tokens fails is a good place to start looking. 

Besides For UnderlyingAsset that are the stmbols of the tokens we should start accep the same characters of the networks.

mmcshinsky-bitgo:
@Joelsx This isn't a breaking change for the repo itself. Let's remove that from the commit message.

mmcshinsky-bitgo:
Also, since this is an object, you can pass in `SETH-H` if you bracket it.

modules/statics/src/base.ts
```diff
@@ -649,6 +650,7 @@ export enum UnderlyingAsset {
   STBU = 'stbu',
   STC = 'stc',
   STCV2 = 'stcv2',
+  SETHH = 'seth-h',
```
mmcshinsky-bitgo, line 653:
```suggestion
 'SETH-H' = 'seth-h',
```
Would that meet the constraints you are looking for in regards to the hyphen or are there too many down stream services that would need to change interpretations of the enum?

modules/statics/src/coins.ts
```diff
@@ -699,6 +700,7 @@ export const coins = CoinMap.fromCoins([
   erc20('stbu', 'Stobox Token', 18, '0x212dd60d4bf0da8372fe8116474602d429e5735f', UnderlyingAsset.STBU),
   erc20('stc', 'Student Coin', 2, '0xb8b7791b1a445fb1e202683a0a329504772e0e52', UnderlyingAsset.STC),
   erc20('stcv2', 'Student Coin V2', 18, '0x15b543e986b8c34074dfc9901136d9355a537e7e', UnderlyingAsset.STCV2),
+  erc20('seth-h', 'Staked Aave', 18, '0x65077fa7df8e38e135bd4052ac243f603729892d', UnderlyingAsset.SETHH),
```
mmcshinsky-bitgo, line 703: Then you can bracket notation it here and keep the one-to-one token name you are looking for.

modules/statics/src/base.ts
Joelsx: Well, that's the main issue we dont know the full implications. But for XML is use your recomendations so I can change to use your suggestion.

mmcshinsky-bitgo: It's a new addition `statics` wise, and it's JavaScript so any service wanting to make use of it should be able to use it seamlessly.

Joelsx: done

mmcshinsky-bitgo (approved):

PR #2563, opened 2022-07-22
perf(account-lib): remove unnecessary clamp in shamir

sachushaji:
this change removes unnecessary clamping for shamir secrets.
This change also adds some missing unit test cases for shamir
secret implementation.

TICKET: BG-52170


sachushaji:
closing in favour of #2572

PR #2564, opened 2022-07-22
refactor(account-lib): preferring exception over assertion MPC implementation

sachushaji:
This change prefers using try catch with meaningful error messages for shamir and eddsa tss implementations over asserts. While assert successfully aborts execution, the fact that the functions throw uncaught exceptions means that function termination can lead to unpredictable program behavior. Note that uncaught exceptions also concern outright program terminations executed at the top-level function, though this is not the case here.

TICKET: BG-52167


sachushaji:
Kindly review #2563 before reviewing this. Check the latest commit for changes in the meanwhile.

sachushaji:
closing in favour of #2572

PR #2568, opened 2022-07-22
refactor(account-lib): deprecate hbar to and amount function

kevinwan-bitgo:
Ticket: BG-51282

### DIFF
--- a/modules/sdk-coin-hbar/src/lib/coinTransferBuilder.ts
+++ b/modules/sdk-coin-hbar/src/lib/coinTransferBuilder.ts
@@ -1,22 +1,14 @@
 import { BaseCoin as CoinConfig } from '@bitgo/statics';
 import * as Long from 'long';
 import * as proto from '@hashgraph/proto';
-import { ExtendableError, InvalidParameterValueError } from '@bitgo/sdk-core';
+import { InvalidParameterValueError } from '@bitgo/sdk-core';
 import { Recipient } from './iface';
 import { TransferBuilder } from './transferBuilder';
 import { Transaction } from './transaction';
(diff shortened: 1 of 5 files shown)


jonathancn-bitgo (approved):

modules/sdk-coin-hbar/src/lib/coinTransferBuilder.ts
```diff
@@ -78,67 +70,13 @@ export class CoinTransferBuilder extends TransferBuilder {
...
-      Currently work for one recipient by using exatcly one of to + amount or send function
-   */
-  /**
-   * Set the destination address where the funds will be sent,
-   * it may take the format `'<shard>.<realm>.<account>'` or `'<account>'`
-   *
-   * @param {string} address - The address to transfer funds to
-   * @returns {TransferBuilder} - The builder with the new parameter set
-   */
-  to(address: string): this {
```
mmcshinsky-bitgo, line 91: Technically... as a result of a stable release, this is a breaking change. Is there any chance between our last `latest` release and whenever we do another `latest` release that these methods can be successfully used by end customers?

kevinwan-bitgo: It seems that these methods can be accessed by the end user through importing `@bitgo/sdk-coin-hbar`

mmcshinsky-bitgo: Is it expected that they'll be used before the next stable release by end users?

mmcshinsky-bitgo: If so, we need to add BREAKING CHANGES: <description> to the commit footer.

PR #2571, opened 2022-07-22
docs: remove the outdated publishing guide

byhow:
Edit: Remove it as per discussed to refer to the updated docs to avoid confusion

~~It'd be great if anyone else with access can verify that it works on their machines too~~

Ticket: BG-00000


tylerlevine (dismissed):

byhow:
Was missing reference in the previous commit, amended

tylerlevine (approved):

PR #2572, opened 2022-07-25
fix(account-lib): BG-52162 shamir sharing changes

sachushaji:
This PR contains fixes for shamir secret sharing and MPC. All changes specific to a unique ticket is a seperate commit for the ease of review. The changes in the PR are as follows.

## Changes

1.  ****BG-00000** chore(account-lib): add unit tests in shamir**  
This change adds some missing unit test cases for shamir secret implementation.

2.  ****BG-52167** refactor(account-lib): preferring try catch over asserts** 
This change prefers using try catch with meaningful error messages for shamir and eddsa tss implementations over asserts. While assert successfully aborts execution, the fact that the functions throw uncaught exceptions means that function termination can lead to unpredictable program behavior. Note that uncaught exceptions also concern outright program terminations executed at the top-level function, though this is not the case here.

3. ****BG-52163, BG-52166** fix(account-lib): shamir secret indices validity** 
This change fixes a potential shamir share leakage by restricting the valid indices to non-zero numbers.
This change also gracefully handles exception caused by combining invalid modularly non-unique shares

4. ****BG-52168** feat(account-lib): custom salt shamir share** 
This change adds an optional salt parameter which can be used via key generation so that we can have an option to get custom shares based on salt as compared to an always deterministic share based on its hash

TICKET :  BG-52162

### DIFF
--- a/modules/sdk-core/src/account-lib/mpc/shamir.ts
+++ b/modules/sdk-core/src/account-lib/mpc/shamir.ts
@@ -1,4 +1,3 @@
-const assert = require('assert');
 const crypto = require('crypto');
 import Curve from './curves';
 import { bigIntFromBufferLE, bigIntToBufferLE, clamp } from './util';
@@ -31,8 +30,14 @@ export default class Shamir {
         .fill(null)
         .map((_, i) => BigInt(i + 1));
     }
-    assert(threshold > 1);
-    assert(threshold <= numShares);
+    if (threshold < 2) {
+      throw new Error('Threshold cannot be less than two');
+    }
+
+    if (threshold > numShares) {
+      throw new Error('Threshold cannot be greater than the total number of shares');
+    }
+
     const coefs: bigint[] = [];
     for (let ind = 0; ind < threshold - 1; ind++) {
       const coeff = clamp(
--- a/modules/sdk-core/src/account-lib/mpc/shamir.ts
+++ b/modules/sdk-core/src/account-lib/mpc/shamir.ts
@@ -23,7 +23,12 @@ export default class Shamir {
   split(secret: bigint, threshold: number, numShares: number, indices?: Array<number>): Record<number, bigint> {
     let bigIndices: Array<bigint>;
     if (indices) {
-      bigIndices = indices.map((i) => BigInt(i));
+      bigIndices = indices.map((i) => {
+        if (i < 1) {
+          throw new Error('Invalid value supplied for indices');
+        }
+        return BigInt(i);
+      });
     } else {
       // make range(1, n + 1)
       bigIndices = Array(numShares)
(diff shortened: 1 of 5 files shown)


modules/sdk-core/src/account-lib/mpc/shamir.ts
```diff
@@ -17,26 +16,47 @@ export default class Shamir {
...
+
+    if (threshold > numShares) {
+      throw new Error('Threshold cannot be greater than the total number of shares');
+    }
+
     const coefs: bigint[] = [];
     for (let ind = 0; ind < threshold - 1; ind++) {
-      const coeff = clamp(
-        bigIntFromBufferLE(crypto.createHmac('sha256', ind.toString(10)).update(bigIntToBufferLE(secret, 32)).digest())
+      const coeff = bigIntFromBufferLE(
```
sachushaji, line 55: Just want to make sure here clamp doesn't bring any additional benefit as mentioned on the ticket, The eddsa tests have became flaky

johnoliverdriscoll: Are you saying that removing the clamp has made the tests flaky?

sachushaji: This is the only significant change done to the code, I am getting intermittent CI failures with eddsa signing failing, not able to reproduce locally though @johnoliverdriscoll

johnoliverdriscoll: Ok I'll see if I can reproduce... it's been a while since I worked on the SDK, what commands are you using to test locally?

sachushaji: I simply run the individual test case using vscode / intellij

johnoliverdriscoll: Which test(s) are failing intermittently?

sachushaji: @johnoliverdriscoll https://github.com/BitGo/BitGoJS/runs/7501100144?check_suite_focus=true

johnoliverdriscoll: Looking at this again, I have problems with how the issue is written it the report. It suggests that clamping is not necessary and the alternative is to use a random 32 byte buffer, however, that would produce elements that are not in the Ed25519 field. The actual alternative would be to use a curve reduction, which is arguably less performant than clamping. I'm noting performance here because that is the only benefit cited in the issue. I suggest we continue to clamp. This will resolve the flaky tests.

johnoliverdriscoll (changes requested):

trustkip:
@johnoliverdriscoll you are referring to BIG-09-007 WP3: Unnecessary clamping for Shamir coefficients in the Cure 53 report, right?

johnoliverdriscoll (approved):

byhow (approved):

mmcshinsky-bitgo (approved):

PR #2574, opened 2022-07-25
feat(sdk-coin-bsc): add transfer and transaction builders

dulhanwaduge:
Ticket: BG-52425

Transaction tests will be written in another ticket

### DIFF
--- a/modules/sdk-coin-bsc/src/lib/resources.ts
+++ b/modules/sdk-coin-bsc/src/lib/resources.ts
@@ -1,0 +1,22 @@
+import EthereumCommon from '@ethereumjs/common';
+import { coins, EthereumNetwork } from '@bitgo/statics';
+
+export const testnetCommon = EthereumCommon.forCustomChain(
+  'mainnet',
+  {
+    name: 'testnet',
+    networkId: (coins.get('tbsc').network as EthereumNetwork).chainId,
+    chainId: (coins.get('tbsc').network as EthereumNetwork).chainId,
+  },
+  'petersburg'
+);
+
+export const mainnetCommon = EthereumCommon.forCustomChain(
+  'mainnet',
+  {
+    name: 'mainnet',
+    networkId: (coins.get('bsc').network as EthereumNetwork).chainId,
+    chainId: (coins.get('bsc').network as EthereumNetwork).chainId,
+  },
+  'petersburg'
+);
--- a/modules/sdk-coin-bsc/test/unit/utils.ts
+++ b/modules/sdk-coin-bsc/test/unit/utils.ts
@@ -1,0 +1,29 @@
+import assert from 'assert';
+import should from 'should';
+import { NetworkType } from '@bitgo/statics';
+import { getCommon } from '../../src/lib/utils';
+
+describe('Network Common Configuration', () => {
+  it('getCommon for mainnet', () => {
+    const common = getCommon(NetworkType.MAINNET);
+    should.equal(common.chainName(), 'mainnet');
+    should.equal(common.hardfork(), 'petersburg');
+    should.equal(common.chainIdBN().toString(), '56');
+    should.equal(common.networkIdBN().toString(), '56');
+  });
+
+  it('getCommon for testnet', () => {
+    const common = getCommon(NetworkType.TESTNET);
+    should.equal(common.chainName(), 'testnet');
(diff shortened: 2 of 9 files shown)


modules/sdk-coin-bsc/src/lib/resources.ts
```diff
@@ -0,0 +1,22 @@
+import EthereumCommon from '@ethereumjs/common';
+import { coins, EthereumNetwork } from '@bitgo/statics';
+
+export const testnetCommon = EthereumCommon.forCustomChain(
+  'mainnet',
+  {
+    name: 'testnet',
+    networkId: 1,
```
wesgraham, line 8: this wont work for BSC -- networkId; 1 is for ETH, bsc will be 56 as per https://chainlist.org/chain/56 for mainnet, and 97 for testnet https://chainlist.org/chain/97

wesgraham: ah crap maybe networkId is fine and these should just be for chainid -- i'd just make sure we add bsc and tbsc to bitgojs/statics properly

wesgraham (dismissed):
awesome! approved so long as networkid is fine

modules/sdk-coin-bsc/src/lib/resources.ts
dulhanwaduge: Changed so that networkId and chainId are the same

modules/sdk-coin-bsc/test/unit/utils.ts
```diff
@@ -0,0 +1,29 @@
+import assert from 'assert';
+import should from 'should';
+import { NetworkType } from '@bitgo/statics';
+import { getCommon } from '../../src/lib/utils';
+
+describe('Network Common Configuration', () => {
+  it('getCommon for mainnet', () => {
+    const common = getCommon(NetworkType.MAINNET);
+    should.equal(common.chainName(), 'mainnet');
+    should.equal(common.hardfork(), 'petersburg');
```
dulhanwaduge, line 10: Not too sue about the hardfork, but used `petersburg` since most other eth-likes used that.

wesgraham (approved):

modules/sdk-coin-bsc/src/lib/resources.ts
```diff
@@ -0,0 +1,22 @@
+import EthereumCommon from '@ethereumjs/common';
+import { coins, EthereumNetwork } from '@bitgo/statics';
+
+export const testnetCommon = EthereumCommon.forCustomChain(
```
mmcshinsky-bitgo, line 4: Should these be in `statics`

dulhanwaduge: Hmm, all the other eth-like coins have this `resources.ts` file in their respective modules. Should I still move it over to statisc?

mmcshinsky-bitgo: Can you give me an example you are replicating?

mmcshinsky-bitgo: Testnet definitions would make more sense here: `modules/statics/src/networks.ts`

dulhanwaduge: An example (for Celo) is: https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-coin-celo/src/lib/resources.ts
It seems that all the eth-like coins (Etc, Celo, Rbtc, and Avaxc) all have this common object defined in `resources.ts`.

So I already have the mainnet and testnet definitions in `modules/statics/src/networks.ts`, but what I could do is add a few properties to them, and then create the common object dynamically in this function: https://github.com/BitGo/BitGoJS/blob/1c27a6343f30d341588eadd2a323b4ac2fe73646/modules/sdk-coin-bsc/src/lib/utils.ts#L15

dulhanwaduge: On second thought, I think it is best to keep it as is. It would require inconsistent changes in statics.

mmcshinsky-bitgo: Thanks for the references :)
I appreciate it

mmcshinsky-bitgo (approved):

PR #2576, opened 2022-07-25
chore(sdk-coin-avaxp): improve explain transaction

davidkaplanbitgo:
Add more information for `explainTransaction` for Avalanche P-Chain

STLX-18209

### DIFF
--- a/modules/sdk-coin-avaxp/src/avaxp.ts
+++ b/modules/sdk-coin-avaxp/src/avaxp.ts
@@ -16,7 +16,7 @@ import {
 import * as AvaxpLib from './lib';
 import { AvaxpSignTransactionOptions, TransactionFee, ExplainTransactionOptions } from './iface';
 import _ from 'lodash';
-import { BN } from 'avalanche';
+import { BN, BinTools } from 'avalanche';
 
 export class AvaxP extends BaseCoin {
   protected readonly _staticsCoin: Readonly<StaticsBaseCoin>;
@@ -84,8 +84,9 @@ export class AvaxP extends BaseCoin {
   }
 
   verifyAddress(params: VerifyAddressOptions): boolean {
-    // TODO(STLX-16574): verifyTransaction
-    return true;
+    const bintools = BinTools.getInstance();
+    const address = bintools.parseAddress(params.address, 'P');
+    return address !== undefined;
   }
 
   /**
--- a/modules/sdk-coin-avaxp/src/lib/iface.ts
+++ b/modules/sdk-coin-avaxp/src/lib/iface.ts
@@ -2,6 +2,9 @@ import { Entry, TransactionExplanation as BaseTransactionExplanation, Transactio
 
 export interface TransactionExplanation extends BaseTransactionExplanation {
   type: TransactionType;
+  rewardAddresses: string[];
+  inputs: Entry[];
+  memo?: string;
 }
 
 /**
@@ -17,13 +20,15 @@ export enum MethodNames {
  */
 export interface TxData {
   id: string;
+  inputs: Entry[];
   type: TransactionType;
   fromAddresses: string[];
   threshold: number;
   locktime: string;
   memo: string;
   signatures: string[];
-  outputs: Entry[];
+  stakedOutputs: Entry[];
+  changeOutputs: Entry[];
 }
 
 /**
--- a/modules/sdk-coin-avaxp/src/lib/transaction.ts
+++ b/modules/sdk-coin-avaxp/src/lib/transaction.ts
@@ -9,7 +9,7 @@ import {
 } from '@bitgo/sdk-core';
 import { KeyPair } from './keyPair';
 import { DecodedUtxoObj, TransactionExplanation, TxData } from './iface';
-import { AddValidatorTx, AmountOutput, BaseTx, Tx } from 'avalanche/dist/apis/platformvm';
+import { AddValidatorTx, AmountInput, AmountOutput, BaseTx, Tx } from 'avalanche/dist/apis/platformvm';
 import { BinTools, BN, Buffer as BufferAvax } from 'avalanche';
 import utils from './utils';
 import * as createHash from 'create-hash';
@@ -170,13 +170,15 @@ export class Transaction extends BaseTransaction {
     }
     return {
       id: this.id,
+      inputs: this.inputs,
       fromAddresses: this.fromAddresses,
       threshold: this._threshold,
       locktime: this._locktime.toString(),
       type: this.type,
       memo: utils.bufferToString(this.avaxPTransaction.getMemo()),
       signatures: this.signature,
-      outputs: this.outputs,
+      stakedOutputs: this.stakedOutputs,
+      changeOutputs: this.changeOutputs,
     };
   }
 
@@ -217,16 +219,18 @@ export class Transaction extends BaseTransaction {
   /**
    * Get the list of outputs. Amounts are expressed in absolute value.
    */
-  get outputs(): Entry[] {
-    if (this.type === TransactionType.addValidator) {
-      const addValidatorTx = this.avaxPTransaction as AddValidatorTx;
-      return [
-        {
-          address: addValidatorTx.getNodeIDString(),
-          value: addValidatorTx.getStakeAmount().toString(),
-        },
-      ];
-    }
+  get stakedOutputs(): Entry[] {
+    // Get staked outputs
+    const addValidatorTx = this.avaxPTransaction as AddValidatorTx;
+    return [
+      {
+        address: addValidatorTx.getNodeIDString(),
+        value: addValidatorTx.getStakeAmount().toString(),
+      },
+    ];
+  }
+
+  get changeOutputs(): Entry[] {
     // general support any transaction type, but it's scoped yet
     return this.avaxPTransaction.getOuts().map((output) => {
       const amountOutput = output.getOutput() as any as AmountOutput;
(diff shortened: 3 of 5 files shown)


modules/sdk-coin-avaxp/src/avaxp.ts
```diff
@@ -84,8 +84,7 @@ export class AvaxP extends BaseCoin {
   }
 
   verifyAddress(params: VerifyAddressOptions): boolean {
-    // TODO(STLX-16574): verifyTransaction
-    return true;
+    throw new MethodNotImplementedError();
```
jicli, line 87: im little confused on why we throwing now? it will likely break?

joey-bai: +1

modules/sdk-coin-avaxp/src/lib/iface.ts
```diff
@@ -2,6 +2,14 @@ import { Entry, TransactionExplanation as BaseTransactionExplanation, Transactio
...
 export interface TransactionExplanation extends BaseTransactionExplanation {
   type: TransactionType;
+  fromAddresses: string[];
+  rewardAddresses: string[];
+  nodeID: string | undefined;
+  totalIn: string;
+  totalStaked: string;
+  memo: string;
+  threshold: number;
+  changeOutputs: any;
```
joey-bai, line 12: can we document these parameters?

joey-bai, line 12: let's define a type for `changeOutputs`.

modules/sdk-coin-avaxp/src/lib/iface.ts
```diff
@@ -2,6 +2,14 @@ import { Entry, TransactionExplanation as BaseTransactionExplanation, Transactio
 
 export interface TransactionExplanation extends BaseTransactionExplanation {
   type: TransactionType;
+  fromAddresses: string[];
+  rewardAddresses: string[];
```
joey-bai, line 6: curious if we're duplicating information... how are `rewardAddresses` different from the addresses in `outputs`?

if we see that `type === staking`, does it mean that the addresses in `outputs` are reward addresses?

modules/sdk-coin-avaxp/src/lib/iface.ts
```diff
@@ -2,6 +2,14 @@ import { Entry, TransactionExplanation as BaseTransactionExplanation, Transactio
...
+  nodeID: string | undefined;
+  totalIn: string;
+  totalStaked: string;
```
joey-bai, line 9: same, are we duplicating info?

can `totalStaked` be derived from summing up values in `outputs` for staking txns?

modules/sdk-coin-avaxp/src/lib/transaction.ts
```diff
@@ -258,16 +261,57 @@ export class Transaction extends BaseTransaction {
...
+      totalIn = totalIn.add(new BN(utxo.amount));
+    });
+    const totalStaked = txJson.outputs.reduce((p, n) => p.add(new BN(n.value)), new BN(0));
+    const totalOut = totalStaked;
+    const totalChange = totalIn.sub(totalOut);
+    const rewardAddresses = this.rewardAddresses;
+    const nodeID =
+      this.type === TransactionType.addValidator
+        ? (this.avaxPTransaction as AddValidatorTx).getNodeIDString()
+        : undefined;
```
joey-bai, line 276: do we need this business logic in `explainTransaction`? can it be moved to `.toJson` instead?

modules/sdk-coin-avaxp/src/avaxp.ts
davidkaplanbitgo: We don't use `verifyAddress` anywhere, and the comment said it was dependent on `verifyTransaction` - I don't think it is a good idea to keep the comment in there

modules/sdk-coin-avaxp/src/lib/iface.ts
davidkaplanbitgo: `totalIn` would be the staked + fee, `totalStaked` would be the sum of the outputs. This can be removed

modules/sdk-coin-avaxp/src/lib/transaction.ts
```diff
@@ -258,14 +258,37 @@ export class Transaction extends BaseTransaction {
...
       displayOrder,
       id: txJson.id,
       outputs: txJson.outputs.map((o) => ({ address: o.address, amount: o.value, memo: txJson.memo })),
-      outputAmount: txJson.outputs.reduce((p, n) => p.add(new BN(n.value)), new BN(0)).toString(),
-      changeOutputs: [], // account based does not use change outputs
-      changeAmount: '0', // account base does not make change
+      outputAmount,
+      changeOutputs: [
+        {
+          address: this.fromAddresses.toString(),
```
tylerlevine, line 286: Isn't `fromAddresses` an array? Do we need to figure out which `fromAddress` received the change output(s)?

modules/sdk-coin-avaxp/src/lib/transaction.ts
```diff
@@ -258,14 +258,37 @@ export class Transaction extends BaseTransaction {
...
+      'fee',
+      'type',
+    ];
+
+    let totalIn = new BN(0);
+    this._utxos.map((utxo) => {
+      totalIn = totalIn.add(new BN(utxo.amount));
+    });
+    const outputAmount = txJson.outputs.reduce((p, n) => p.add(new BN(n.value)), new BN(0)).toString();
+    const changeAmount = totalIn.sub(new BN(outputAmount)).toString();
```
tylerlevine, line 277: Is this correct? Is the change amount implicit? Usually this is how the mining fee is calculated for utxo coins and not the change amount, which appears as explicit output(s) on the transaction.

tylerlevine (changes requested):

modules/sdk-coin-avaxp/src/lib/transaction.ts
davidkaplanbitgo: The type for `changeOutputs` in  [`ITransactionRecipient `](https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts#L66) is a `string` - changing the type of this address from `string` to `string | string[]` has cascading effects throughout  the SDK in other coins. When I tried to override the type for `changeOutputs`, I could not build the SDK due to errors. Instead of not extending `ITransactionExplanation`, I thought to just make the addresses into a single string and then parse it later. Is this a bad idea?

The alternative is to effectively reimplement the `ITransactionExplanation` with that single parameter changed.

modules/sdk-coin-avaxp/src/lib/transaction.ts
davidkaplanbitgo: There are [no network fees for delegating or validating for avalanche](https://docs.avax.network/quickstart/transaction-fees#fee-schedule), which are the only two things that we do with the P-chain. So the only thing contributing to the outputs are the amount that you are staking and the change left over from your original amount that gets sent back as change

modules/sdk-coin-avaxp/src/lib/transaction.ts
tylerlevine: There is no `changeOutputs` property in `ITransactionRecipient` - nor would this make sense since a recipient does not have a change output. I'm guessing you probably meant the `changeOutputs` property on `ITransactionExplanation`?

modules/sdk-coin-avaxp/src/lib/transaction.ts
tylerlevine: and that change amount is not an output on the transaction? It's left implicit as the difference between the input and output amounts?

modules/sdk-coin-avaxp/src/lib/transaction.ts
davidkaplanbitgo: Yeah - sorry my explanation was bad.

The `changeOutputs` property in `ITransactionExplanation` is the type `ITransactionRecipient[]`, and the `address` property of `ITransactionRecipient` is a `string` - changing the address property type from a `string` to `string | string[]` would mess up a lot of other coins.

tylerlevine: I understand that, but that's not what I'm asking about.

Each element in `changeOutputs` should correspond to exactly one output. We currently have a list of `fromAddresses` that's getting concatenated into one `address` property in one element in `changeOutputs`. If we have a list of more than one change outputs on the transaction, why aren't we creating separate elements in `changeOutputs` where each corresponds to one address?

If there is only one change output (and thus we should only have one element in `changeOutputs`), then why do we have a list of addresses associated with that change output instead of a single address?

jrafael-simtlix: Multisig wallet in Avalanche P and X Chains don't have unique address. They use the addresses of the signers. So those 3 addresses are one wallet.

modules/sdk-coin-avaxp/src/lib/transaction.ts
jrafael-simtlix: I see this alternatives:
1.- set the change amount to 0, as an account. This is implicit now.
2.- copy utxo behavior. See it below

https://github.com/BitGo/BitGoJS/blob/fedcb7e6c4c4078cc1425f74f068f6933a64861b/modules/core/src/v2/coins/abstractUtxoCoin.ts#L1256-L1289

tylerlevine (approved):
lgtm

mmcshinsky-bitgo (approved):

PR #2579, opened 2022-07-26
feat(account-lib): add option to pass in custom seed ecdsa

sachushaji:
this change adds capability to give custom seed for key generation for ecdsa tss

Ticket: BG-51644


modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -40,18 +40,23 @@ export default class Ecdsa {
...
+    if (seed && seed.length !== 72) {
+      throw new Error('Seed must have length 72');
+    }
     // Generate additively homomorphic encryption key.
     const { publicKey, privateKey } = await paillierBigint.generateRandomKeys(3072, true);
-    const u = Ecdsa.curve.scalarRandom();
+    const u = (seed && bigIntFromU8ABE(secp.utils.hashToPrivateKey(seed.slice(0, 40)))) ?? Ecdsa.curve.scalarRandom();
     const y = Ecdsa.curve.basePointMult(u);
-    const chaincode = randomBytes(32);
+    const chaincode = seed?.slice(40) ?? randomBytes(32);
```
johnoliverdriscoll, line 59: These lengths seem very odd. Why is the seed 72 bytes and the u value 40 bytes?

sachushaji: @johnoliverdriscoll ,https://github.com/paulmillr/noble-secp256k1/blob/d8b5d189e0788ebd26df79cc2545a097ca5cee8b/index.ts#L1520 replicated the random function as in secp256k1 library which needs 40 bytes and 32 bytes for the chaincode. So total seed length => 72 for privateKey (u) and chaincode

johnoliverdriscoll (approved):

mmcshinsky-bitgo (approved):

PR #2580, opened 2022-07-26
feat(sdk-coin-ada): incorporate sdk-coin-ada back into bitgo

margueriteblair:
Ticket: BG-53173

### DIFF
--- a/modules/bitgo/karma.conf.js
+++ b/modules/bitgo/karma.conf.js
@@ -21,6 +21,7 @@ module.exports = function (config) {
       'dist/browser/BitGoJS.min.js',
       'test/browser/karmaHelper.js',
       { pattern: 'test/browser/**/*.ts' },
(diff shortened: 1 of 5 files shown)


modules/account-lib/package.json
```diff
@@ -28,6 +28,7 @@
     "node": ">=14 <17"
   },
   "dependencies": {
+    "@bitgo/sdk-coin-ada": "^1.1.0",
```
mmcshinsky-bitgo, line 31: Only needed in `bitgo`. Can remove from `account-lib`.

mmcshinsky-bitgo (changes requested):

modules/account-lib/package.json
margueriteblair: @ewangbitgo Can you confirm that we need sdk-coin-ada in account lib?

mmcshinsky-bitgo (approved):

PR #2586, opened 2022-07-27
feat(account-lib): add ada transaction builder

ewangbitgo:
Ticket: BG-50796


margueriteblair (approved):
lgtm

modules/sdk-coin-ada/src/transaction.ts
```diff
@@ -1,17 +1,185 @@
-import { BaseTransaction, BaseKey, NotImplementedError } from '@bitgo/sdk-core';
+import {
+  BaseKey,
+  BaseTransaction,
+  Entry,
+  InvalidTransactionError,
+  TransactionType,
+  TransactionExplanation,
+} from '@bitgo/sdk-core';
+import * as CardanoWasm from '@emurgo/cardano-serialization-lib-nodejs';
```
byhow, line 9: How will we be handling using the sdk in the browser given that this seems to be a nodejs package?

modules/sdk-coin-ada/src/transactionBuilderFactory.ts
```diff
@@ -11,8 +11,21 @@ export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
   }
 
   /** @inheritdoc */
-  from(raw: Uint8Array | string): TransactionBuilder {
-    throw new NotImplementedError('from not implemented');
+  from(raw: string): TransactionBuilder {
```
byhow, line 14: Consider adding BREAKING CHANGE or exclamation marks after commit type as this changes how function parameters are passed in

modules/sdk-coin-ada/test/unit/utils.ts
```diff
@@ -27,6 +27,8 @@ describe('utils', () => {
   });
 
   it('should validate public key correctly', () => {
+    //    should.equal(utils.isValidPublicKey('0304e96b46fcb0974e94a002fda63f3ba8b97429ed27238b7b7ad52553219d4928'), true);
```
byhow, line 30: nit: remove comments if not needed

modules/sdk-coin-ada/src/transactionBuilderFactory.ts
mmcshinsky-bitgo: Side note, BitGoJS fails the `!`, so we have to use BREAKING CHANGE in the commit footer.

modules/sdk-coin-ada/src/transaction.ts
margueriteblair: We wrap it within @cardano-serialization-lib-browser which is compatible :) The polyfill code for that is in bitgo.config.js

modules/sdk-coin-ada/src/transaction.ts
```diff
@@ -1,17 +1,185 @@
-import { BaseTransaction, BaseKey, NotImplementedError } from '@bitgo/sdk-core';
+import {
```
mmcshinsky-bitgo, line 1: Let's move all transaction/transfer files to `lib` for consistency with other coins.

modules/sdk-coin-ada/src/transactionBuilderFactory.ts
byhow: Ah good to know

mmcshinsky-bitgo (approved):

PR #2592, opened 2022-07-28
fix(sdk-core): tss ecdsa keychain encryption

olibeyene:
Ticket: BG-53419


modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -506,6 +509,7 @@ describe('TSS Ecdsa Utils:', async function () {
   }): Promise<Keychain> {
 
     nock('https://bitgo.fakeurl')
```
rlutzbitgo, line 511: Should probably just set this to bgUrl as well

bitgoAaron (approved):

PR #2593, opened 2022-07-28
refactor(statics): update sol token asset

jonathancn-bitgo:
BG-52918

BREAKING CHANGE: This breaks the provided token asset used for wp in bitgo-ms, will require to change
available sokana token asset in wp once this merged.
BG-52918

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -1580,14 +1580,6 @@ export const coins = CoinMap.fromCoins([
     AccountCoin.DEFAULT_FEATURES
   ),
   solToken(
-    'sol:orca',
-    'Orca',
-    6,
-    'orcaEKTdK7LKz57vaAYr9QeNsVEPfiu6QeMU1kektZE', // https://explorer.solana.com/address/orcaEKTdK7LKz57vaAYr9QeNsVEPfiu6QeMU1kektZE
-    UnderlyingAsset.ORCA,
-    AccountCoin.DEFAULT_FEATURES
-  ),
-  solToken(
     'sol:usdc',
     'USD Coin',
     6,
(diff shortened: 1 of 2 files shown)


modules/statics/src/coins.ts
```diff
@@ -1579,14 +1579,6 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.SRM,
     AccountCoin.DEFAULT_FEATURES
   ),
-  solToken(
```
mmcshinsky-bitgo, line 1582: These are breaking changes and need to be added to the commit footer.

mmcshinsky-bitgo, line 1582: These are breaking changes and need to be added to the commit footer.

mmcshinsky-bitgo (changes requested):

modules/statics/src/coins.ts
jonathancn-bitgo: Updated the message commit, so how does breaking change gonna be handled later?

Since in my case I need to change the available sol token asset in wallet-platform after updating the static lib and do the change in indexer as well (which I already have PR open for that)

mmcshinsky-bitgo: It will cause a `major` version bump in the next release. You'll need to fix BREAKING CHANGE to be in the commit footer.

mmcshinsky-bitgo: You'll still keep the BG-number on the next line after the description in the footer.

mmcshinsky-bitgo:
```
BREAKING CHANGE: This breaks the provided token asset used for wp in bitgo-ms,
will require to change available solana token asset in wp once this merged.
BG-52918
```

mmcshinsky-bitgo (approved):

PR #2596, opened 2022-07-29
feat(utxo-bin): improve output scripts parsing

OttoAllmendinger:
Output scripts (scriptPubKey) is shown if there is no corresponding address
or if `--parseOutputScript` is set.

Issue: BG-53109


OttoAllmendinger:
wrong branchname

PR #2598, opened 2022-07-29
feat(sdk-core): add specialized enable token functions

kevinwan-bitgo:
Ticket: BG-53196

### DIFF
--- a/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
+++ b/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
@@ -321,6 +321,11 @@ export interface RecoverTokenTransaction {
   };
 }
 
+export interface TokenEnablementConfig {
+  requiresTokenEnablement: boolean;
+  supportsMultipleTokenEnablements: boolean;
+}
+
 export interface IBaseCoin {
   type: string;
   url(suffix: string): string;
@@ -337,6 +342,7 @@ export interface IBaseCoin {
   sweepWithSendMany(): boolean;
   transactionDataAllowed(): boolean;
   allowsAccountConsolidations(): boolean;
+  getTokenEnablementConfig(): TokenEnablementConfig;
   supportsTss(): boolean;
   supportsStaking(): boolean;
   supportsBlsDkg(): boolean;
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -34,11 +34,21 @@ export interface BuildConsolidationTransactionOptions extends PrebuildTransactio
   consolidateAddresses?: string[];
 }
 
+export interface TokenEnablement {
+  name: string;
+  address?: string; // Some chains like Solana require tokens to be enabled for specific address. If absent, we will enable it for the wallet's root address
+}
+
+export interface BuildTokenEnablementOptions extends PrebuildTransactionOptions {
+  tokens: TokenEnablement[];
+}
+
 export interface PrebuildTransactionOptions {
   reqId?: IRequestTracer;
   recipients?: {
     address: string;
     amount: string | number;
+    tokenName?: string;
   }[];
   numBlocks?: number;
   maxFeeRate?: number;
@@ -556,4 +566,7 @@ export interface IWallet {
   buildAccountConsolidations(params?: BuildConsolidationTransactionOptions): Promise<PrebuildTransactionResult[]>;
   sendAccountConsolidation(params?: PrebuildAndSignTransactionOptions): Promise<any>;
   sendAccountConsolidations(params?: BuildConsolidationTransactionOptions): Promise<any>;
+  buildTokenEnablements(params?: BuildTokenEnablementOptions): Promise<PrebuildTransactionResult[]>;
+  sendTokenEnablement(params?: PrebuildAndSignTransactionOptions): Promise<any>;
+  sendTokenEnablements(params?: BuildTokenEnablementOptions): Promise<any>;
 }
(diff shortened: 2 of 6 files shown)


modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -34,11 +34,16 @@ export interface BuildConsolidationTransactionOptions extends PrebuildTransactio
   consolidateAddresses?: string[];
 }
 
+export interface BuildTokenEnablementOptions extends PrebuildTransactionOptions {
+  tokens: string[];
```
joey-bai, line 38: `tokenNames`

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2268,6 +2269,83 @@ export class Wallet implements IWallet {
     }
   }
 
+  public async buildTokenEnablements(
```
joey-bai, line 2272: add JS docs for the functions you added

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2268,6 +2269,83 @@ export class Wallet implements IWallet {
...
+    }
+    if (params.recipients) {
+      throw new Error('Can not specify recipients for token enablement transactions');
+    }
+
+    if (params.reqId) {
+      this.bitgo.setRequestTracer(params.reqId);
+    }
+
+    const buildParams = _.pick(params, this.prebuildWhitelistedParams());
```
joey-bai, line 2289: do we want a separate set of whitelisted params just for token enablement?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2268,6 +2269,83 @@ export class Wallet implements IWallet {
...
+  }
+
+  public async sendTokenEnablement(params: PrebuildAndSignTransactionOptions = {}): Promise<any> {
+    if (!this.baseCoin.requiresTokenEnablement()) {
+      throw new Error(`${this.baseCoin.getFullName()} does not require token enablement transactions`);
+    }
+
+    if (typeof params.prebuildTx === 'string' || params.prebuildTx === undefined) {
+      throw new Error('Invalid build of token enablement.');
+    }
```
joey-bai, line 2310: is there a way where we can check that the `prebuildTx` is a token enablement txn?

modules/sdk-core/src/bitgo/wallet/wallet.ts
kevinwan-bitgo: `buildParams` here are the parameters that we plan on sending to `prebuildTransacation` so `prebuildWhitelistedParams` are the params that we want. We write token names into `recipients` later.

kevinwan-bitgo: Also, when we support the tss builds, we need to add the enable token params to `prebuildWhitelistedParams` for building with intent to work.

modules/sdk-core/src/bitgo/wallet/wallet.ts
kevinwan-bitgo: `prebuildTransaction` does not return the `buildParams` used so this should be pretty difficult

kevinwan-bitgo: Another thing to note is that this function does not do anything specific to token enablements. It's the same as `sendAccountConsolidation` only we omit a few checks they do

joey-bai: 👍 thanks for explaining

joey-bai (approved):

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
```diff
@@ -321,6 +321,11 @@ export interface RecoverTokenTransaction {
   };
 }
 
+export interface TokenEnablementConfig {
+  requiresTokenEnablement: boolean;
+  supportsMultipleTokenEnablements?: boolean;
```
joey-bai, line 326: add doc: "flag indicating if multiple tokens can be enabled in a single transaction"

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
```diff
@@ -321,6 +321,11 @@ export interface RecoverTokenTransaction {
   };
 }
 
+export interface TokenEnablementConfig {
+  requiresTokenEnablement: boolean;
```
joey-bai, line 325: add doc: "flag indicating if tokens need to be enabled before receiving and sending"

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
joey-bai: also, this should be required and we default to false.

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -34,11 +34,19 @@ export interface BuildConsolidationTransactionOptions extends PrebuildTransactio
   consolidateAddresses?: string[];
 }
 
+export interface BuildTokenEnablementOptions extends PrebuildTransactionOptions {
+  tokens: {
+    name: string;
+    address?: string;
+  }[];
```
joey-bai, line 41: same lets document properties.

especially address as its not always clear. e.g. "Some coins like Solana require tokens be enabled on a specific address. If absent, we assume it is for the Wallet's root address"

joey-bai (approved):

mmcshinsky-bitgo (approved):

PR #2599, opened 2022-07-29
feat(sdk-core): add includeTokens wallet.addresses parameter

michael-gallegos-bg:
We recently introduced an includeTokens param to Wallet Platform.
This makes it easier to filter out wallet addresses with/without the token property.
Adding the param as an option to the SDK.

Additions:
- includeTokens param in Wallet.addresses
- Unit test to ensure non-boolean input is rejected

TICKET: BG-53596

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -239,6 +239,7 @@ export interface AddressesOptions extends PaginationOptions {
   segwit?: boolean;
   chains?: number[];
   includeBalances?: boolean;
+  includeTokens?: boolean;
   includeTotalAddressCount?: boolean;
   returnBalancesForToken?: string;
   pendingDeployment?: boolean;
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -239,6 +239,7 @@ export interface AddressesOptions extends PaginationOptions {
   segwit?: boolean;
   chains?: number[];
   includeBalances?: boolean;
+  includeTokens?: boolean;
```
mmcshinsky-bitgo, line 242: Is there documentation for this on the api docs?

michael-gallegos-bg: Not yet, I can mark this as draft for now and add those in. Thanks :)

michael-gallegos-bg: Re-opening this PR as this has been added to the API docs. Thanks
https://github.com/BitGo/bitgo-microservices/pull/21866

mmcshinsky-bitgo (approved):

PR #2604, opened 2022-08-01
feat(sdk-core): Add enable token support for sol

kevinwan-bitgo:
Ticket: BG-53197

BREAKING CHANGE: This breaks buildTransactionWithIntent in the case where we want to enable tokens. This is only currently being used for solana tokens which are not on prod yet. We will need to add the enableToken intent type and add logic to deal with these types for solana on wp.

BG-53197

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -694,21 +694,22 @@ describe('V2 Wallet:', function () {
 
     it('Should build token enablement transactions', async () => {
       const params = {
-        tokens: [{
+        enableTokens: [{
           name: 'thbar:usdc',
         }],
       };
-      nock(bgUrl)
+      const txRequestNock = nock(bgUrl)
         .post(`/api/v2/${hbarWallet.coin()}/wallet/${hbarWallet.id()}/tx/build`)
         .reply((uri, body) => {
           const params = JSON.parse(body);
           params.recipients.length.should.equal(1);
           params.recipients[0].tokenName.should.equal('thbar:usdc');
           params.type.should.equal('enabletoken');
-          should.not.exist(params.tokens);
+          should.not.exist(params.enableTokens);
           return params;
         });
       await hbarWallet.buildTokenEnablements(params);
+      txRequestNock.isDone().should.equal(true);
     });
 
     afterEach(() => {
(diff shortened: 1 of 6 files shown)


mmcshinsky-bitgo:
It looks like there aren't any breaking changes in `sdk-coin-sol`. We would need two commits to major bump `sdk-core` (note: BREAKING CHANGE needs to be in the commit footer) and a minor bump in `sdk-coin-sol`.

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -831,6 +832,30 @@ describe('V2 Wallet:', function () {
...
+            apiVersion: 'full',
+            transactions: [{
+              unsignedTx: {
+                serializedTxHex: 'fake transaction',
+                feeInfo: 'fake fee info',
+              },
+            }],
+          };
+        });
+      await solWallet.buildTokenEnablements(params);
```
joey-bai, line 857: let's verify the nock so we know your `reply` body was actually executed.

should be able to do something like:
```
const txRequestNock = nock(bgUrl)...

await solWallet.buildTokenEnablements(params);

txRequestNock.isDone();
```

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -38,6 +43,7 @@ export interface PrebuildTransactionWithIntentOptions {
   comment?: string;
   memo?: Memo;
   tokenName?: string;
+  tokens?: TokenEnablement[];
```
joey-bai, line 46: maybe rename this to `enableTokens` to be clear about its purpose

kevinwan-bitgo: I named the field `tokens` here to match the `BuildTokenEnablementOptions` type. If we rename this type here, we should probably rename it in `BuildTokenEnablementOptions` as well.

joey-bai (approved):

mmcshinsky-bitgo (changes requested):
It looks like there aren't any breaking changes in sdk-coin-sol. We would need two commits to major bump sdk-core (note: BREAKING CHANGE needs to be in the commit footer) and a minor bump in sdk-coin-sol.

kevinwan-bitgo:
The thing that might break is if we use `sendMany` to enable tokens then wallet platform will not recognize our new intent types. But solana tokens aren't on prod and we don't plan on supporting this method so it probably won't affect anyone.

joey-bai (approved):

joey-bai:
@mmcshinsky-bitgo please let us know whether this is actually considered a breaking change or not.

PR #2607, opened 2022-08-02
feat(wp): add v1 wallet init code

gianchandania:
This PR adds logic to initialize v1 wallets

**Changes**

- Modifies `buildWalletInitializationTransaction()` method to support v1 wallets
- Modifies `decodeWalletCreationData()` method to support v1 wallets
- Added `walletVersion()` method to set the current walletVersion
- Added `getWalletV1InitData()` method to get method call data for v1 wallets
- Added walletFactoryAddress and  walletImplementationAddress for polygon in networks.ts
- Modified tests for walletInitialization for polygon

BREAKING CHANGE: Changes the return for exported method
decodeWalletCreationData() in sdk-coin-eth

Ticket: [BG-53733](https://bitgoinc.atlassian.net/browse/BG-53733)


modules/sdk-coin-eth/src/lib/transactionBuilder.ts
```diff
@@ -91,6 +95,7 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     this._value = '0';
     this._walletOwnerAddresses = [];
     this._forwarderVersion = 0;
+    this._walletVersion = 0;
```
mmcshinsky-bitgo, line 98: Should we default to `1` for the majority of wallet creations?

gianchandania: Actually majority of ethlike coins are using wallet version 0, just polygon (the new addition) is going to use version 1 alongwith eth itself, hence have set the default version to 0

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -225,31 +227,47 @@ export function isValidAmount(amount: string): boolean {
  * @param {string} data The wallet creation data to decode
  * @returns {string[]} - The list of signer addresses
  */
-export function decodeWalletCreationData(data: string): string[] {
-  if (!(data.startsWith(walletInitializationFirstBytes) || data.startsWith(v1WalletInitializationFirstBytes))) {
+export function decodeWalletCreationData(data: string): WalletInitializationData {
```
mmcshinsky-bitgo, line 230: BREAKING CHANGE missing from commit footer for this change.

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -455,7 +473,7 @@ export function classifyTransaction(data: string): TransactionType {
  */
 const transactionTypesMap = {
   [walletInitializationFirstBytes]: TransactionType.WalletInitialization,
-  [v1WalletInitializationFirstBytes]: TransactionType.WalletInitialization,
+  [v1CreateWalletMethodId]: TransactionType.WalletInitialization,
```
mmcshinsky-bitgo, line 476: What's the full impact of changing this? What's the downstream affect of changes users will have to make?

mmcshinsky-bitgo (changes requested):

modules/sdk-coin-eth/src/lib/transactionBuilder.ts
gianchandania: Infact all the ethlike coins that are using account lib don't even have a version 1 at this point

modules/sdk-coin-eth/src/lib/utils.ts
mmcshinsky-bitgo, line 230: This changes the exported return. What is the possible disruption as a result for internal and external users? Does this need a BREAKING CHANGE in the footer and a `major` bump to the package?

mmcshinsky-bitgo (changes requested):

modules/sdk-coin-eth/src/lib/transactionBuilder.ts
mmcshinsky-bitgo: Awesome, thanks!

modules/sdk-coin-eth/src/lib/utils.ts
mmcshinsky-bitgo: I'm leaning this direction due to this portion of the definition of a breaking change: `Removal of an allowed parameter, request field or response field`

modules/sdk-coin-eth/src/lib/utils.ts
gianchandania: Actually this statement was added to support v1 wallet initialization for polygon which is not yet available to customers. But this needs to be changed because there was some issue in the logic. So I don't think the customers or anybody else would be impacted by changing this line as v1 wallet init logic is not present in sdk for any other ethlike coin at this point anyways

`[v1WalletInitializationFirstBytes]: TransactionType.WalletInitialization,`

modules/sdk-coin-eth/src/lib/utils.ts
wesgraham: @mmcshinsky-bitgo This is a great note! In the case of `decodeWalletCreationData`, this method is not typically exposed directly (externally), and is only used for ETH tx building. The method is also not used in BGMS anywhere which gives us additional confidence about how seldom it is used. That being said if it is policy that any return type change constitutes a major bump then we can go through those proceedings

modules/sdk-coin-eth/src/lib/transactionBuilder.ts
```diff
@@ -748,4 +762,17 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
     this._initCode = getProxyInitcode(implementationAddress);
   }
+
+  /**
+   * Set the wallet version for wallet to be initialized
+   *
+   * @param {number} version wallet version
+   */
+  walletVersion(version: number): void {
+    if (version < 0 || version > 2) {
```
wesgraham, line 772: we support ETH wallets with version up to and including 3 https://github.com/BitGo/bitgo-microservices/blob/99011fce7f241688f32669674a19a6e12b7dbe4b/packages/api-docs/shared/CreateWallet.yaml#L61

gianchandania: updating this. Thanks for point this out

modules/sdk-coin-eth/src/lib/utils.ts
mmcshinsky-bitgo: Yeah, it's that weird spot since this is an exported function, we would have to support the possibility that there are external users who could be using it even if it doesn't make sense to us or our implementations already take care of it.

gianchandania: Ok Got it. So I think because of this possibility that external users might be using this method, I should add Breaking Change in the commit footer, is that correct?

modules/sdk-coin-eth/src/lib/transactionBuilder.ts
gianchandania: done

modules/sdk-coin-eth/src/lib/utils.ts
mmcshinsky-bitgo: Yeah, thanks :)

gianchandania: done

mmcshinsky-bitgo (approved):

PR #2515, opened 2022-07-14
feat(sdk-coin-eth): modify address init code

gianchandania:
BG-52294

PR #2609, opened 2022-08-03
fix(sdk-coin-eth): fix v1 address init code

gianchandania:
This PR modifies logic to initialize v1 addresses

**Changes**

- Modifies `buildAddressInitializationTransaction()` method to support v1 address initialization

- Modifies `setTransactionTypeFields() `method to support decoding of v1 addresses

- Added `decodeCreateV1ForwarderData()` method to support decoding of v1 addresses

- Added `getV1AddressInitializationData()` method to get method call data for v1 addresses initialization

- Added `baseAddress()` method to set the baseAddress of wallet

- Modifies `getAddressInitializationData()` by removing optional param forwarderVersion. This method now only returns the method id for createForwarder method for v0 forwarder and not v1 forwarders

- Added tests for addressInitialization for polygon

- Modifies tests because of  change in method `getAddressInitializationData()`

- Renamed `getWalletV1InitData()` added in this [PR](https://github.com/BitGo/BitGoJS/pull/2607) for consistency 

- Renamed createWalletTypes added in this [PR](https://github.com/BitGo/BitGoJS/pull/2607) for consistency

Ticket: [BG-53920](https://bitgoinc.atlassian.net/browse/BG-53920)


modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -190,12 +191,10 @@ export function flushCoinsData(): string {
...
  * Returns the create forwarder method calling data
  *
- * @param {string} forwarderVersion - forwarder version for the address to be initialized
  * @returns {string} - the createForwarder method encoded
  */
-export function getAddressInitializationData(forwarderVersion?: number): string {
-  const impliedForwarderVersion = forwarderVersion ?? defaultForwarderVersion;
-  return impliedForwarderVersion === defaultForwarderVersion ? createForwarderMethodId : v1CreateForwarderMethodId;
+export function getAddressInitializationData(): string {
+  return createForwarderMethodId;
```
gianchandania, line 197: @mmcshinsky-bitgo I am wondering if removing an optional param is a breaking change?

mmcshinsky-bitgo: Only required parameters are breaking changes, but we should clearly detail the change in the commit description.

gianchandania: Got it. Will update the description

modules/sdk-coin-eth/src/lib/utils.ts
```diff
@@ -61,8 +64,12 @@ import {
   walletInitializationFirstBytes,
   v1CreateForwarderMethodId,
   walletSimpleConstructor,
+<<<<<<< HEAD
   createWalletTypes,
   v1CreateWalletMethodId,
+=======
+  createV1ForwarderTypes,
+>>>>>>> 27bae27d3 (fix(sdk-coin-eth): fix v1 address init code)
```
mmcshinsky-bitgo, line 72: Merge conflict comments need to be removed here.

mmcshinsky-bitgo (changes requested):

modules/sdk-coin-eth/src/lib/utils.ts
gianchandania: working on it

modules/sdk-coin-eth/src/lib/utils.ts
mmcshinsky-bitgo: Does removing the optional parameter create situations where the end consumers will need to change their own code to get previous expectations that the function provided in the expected logic flow?

modules/sdk-coin-eth/src/lib/utils.ts
gianchandania: done

modules/sdk-coin-eth/src/lib/utils.ts
gianchandania: Actually to give a little background on the change, I had added this optional parameter in this [PR](https://github.com/BitGo/BitGoJS/pull/2515). But now I have made the function just like it was before this mentioned PR by removing the optional parameter. 

The method will behave in the same manner for the users who were not passing the optional parameter or were passing the default value of optional parameter. But if a non-default value was passed to the method, then it returned a different response previously, and now there are no 2 different responses. It is just a single response (i.e. for the default value).

mmcshinsky-bitgo (approved):

PR #2613, opened 2022-08-04
feat(sdk-core): create ILightning interface

sj3014:
- create ILightning interface and add lightning() function to IWallet
  interface
- create a stub implementation of the ILightning interface

Ticket: BG-54079

### DIFF
--- a/modules/sdk-coin-btc/src/btc.ts
+++ b/modules/sdk-coin-btc/src/btc.ts
@@ -34,4 +34,8 @@ export class Btc extends AbstractUtxoCoin {
   supportsBlockTarget(): boolean {
     return true;
   }
+
+  supportsLightning(): boolean {
+    return true;
+  }
 }
--- a/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
+++ b/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
@@ -268,6 +268,14 @@ export abstract class BaseCoin implements IBaseCoin {
   }
 
   /**
+   * Check whether a coin supports lightning transactions
+   * @returns {boolean}
+   */
+  supportsLightning() {
+    return false;
+  }
+
+  /**
    * Hook to add additional parameters to the wallet generation
    * @param walletParams
    * @param keychains
--- a/modules/sdk-core/src/bitgo/lightning/iLightning.ts
+++ b/modules/sdk-core/src/bitgo/lightning/iLightning.ts
@@ -1,0 +1,164 @@
+import * as t from 'io-ts';
+
+export interface CreateInvoiceParams {
+  amount: number;
+  memo?: string;
+  expiration?: number;
+}
+
+export interface CreateDepositAddressParams {
+  amount: number;
+  memo?: string;
+  expiration?: number;
+}
+
+export interface GetInvoicesParams {
+  id?: string;
+  status?: 'open' | 'closed';
+}
+
+export interface LightningWithdrawalParams {
+  amount: number;
+  address?: string;
+  sequenceId?: string;
+}
+
+export interface LightningDepositParams {
+  amount: number;
+  address?: string;
+  sequenceId?: string;
+}
+
+export interface PayInvoiceParams {
+  invoice: string;
+}
+
+export const WPTransferEntry = t.type(
+  {
+    wallet: t.string,
+    address: t.string,
+    value: t.number,
+  },
+  'WPTransferEntry'
+);
+
+// eslint-disable-next-line no-redeclare
+export type WPTransferEntry = t.TypeOf<typeof WPTransferEntry>;
+
+export const WPTransfer = t.type(
+  {
+    entries: t.array(WPTransferEntry),
+    id: t.string,
+    heightId: t.string,
+    type: t.string,
+    value: t.number,
+    state: t.string,
+  },
+  'WPTransfer'
+);
+
+// eslint-disable-next-line no-redeclare
+export type WPTransfer = t.TypeOf<typeof WPTransfer>;
+
+export const WithdrawResponse = t.strict(
+  {
+    status: t.string,
+    transfer: WPTransfer,
+  },
+  'CreateWithdrawalResponse'
+);
+
+// eslint-disable-next-line no-redeclare
+export type WithdrawResponse = t.TypeOf<typeof WithdrawResponse>;
+
+export const CreateInvoiceResponse = t.strict(
+  {
+    value: t.number,
+    memo: t.union([t.string, t.undefined]),
+    paymentHash: t.string,
+    invoice: t.string,
+    walletId: t.string,
+    status: t.string,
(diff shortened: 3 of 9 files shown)


modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -0,0 +1,155 @@
+/* eslint-disable no-redeclare */
```
mmcshinsky-bitgo, line 1: We should try not to disable eslint rules unless absolutely necessary.

mmcshinsky-bitgo, line 1: It looks like this is a pattern in `io-ts` but this may cause accidental problems when other developers update these files and cause duplicate errors. Can we implement another rule for this for the best of both worlds?

mmcshinsky-bitgo: TypeScript by itself without the eslint rule `no-redeclare` looks to handle it nicely, but we'd have to do a repo wide check for this if we want to verify this approach.

mmcshinsky-bitgo:
Unsubscribing until PR ready. Looks like there are still changes being pushed (not in `draft).

modules/sdk-core/src/bitgo/lightning/iLightning.ts
sj3014: what do you mean by "implement another rule for this" ? Should I use `// eslint-disable-next-line no-redeclare` ?

modules/sdk-coin-btc/src/btc.ts
```diff
@@ -34,4 +34,8 @@ export class Btc extends AbstractUtxoCoin {
   supportsBlockTarget(): boolean {
     return true;
   }
+
+  supportsLightning(): boolean {
+    return true;
```
tylerlevine, line 39: this is ok for now but maybe we can add a feature in `statics` for lightning support?

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
```diff
@@ -267,6 +267,14 @@ export abstract class BaseCoin implements IBaseCoin {
     return false;
   }
 
+  /**
+   * Check whether a coin supports lightning transactions
+   * @returns {boolean}
+   */
+  supportsLightning() {
```
tylerlevine, line 274: I wonder if we can replace this with more generic `getFeatures()`/`hasFeature()` functions that would look at the `features` property of the underlying statics coin definition. Open to thoughts on this.

tylerlevine (approved):
no blockers from me, just one suggestion on a more flexible way to check for these yes/no statements about what a coin can do. I think we can revisit this later if needed.

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
sj3014: I also looked into `CoinFeature` , but it seems only account based coins are making use of it. I might be missing something, but I am also open to thoughts on using features

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -0,0 +1,164 @@
...
+
+export interface GetInvoicesParams {
+  id?: string;
+  status?: 'open' | 'closed';
+}
+
+export interface LightningWithdrawalParams {
+  amount: number;
+  address?: string;
+  sequenceId?: string;
```
sangaman, line 23: We do require `sequenceId` for withdrawals (and addresses, of course). But we can have our code randomly generate a sequenceId if none is provided, and we can generate an address automatically as well, as long as the request to BitGo's API has both those fields on it. Just wanted to make sure that's the plan.

mmcshinsky-bitgo (approved):

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+export interface LightningDepositParams {
+  amount: number;
+  address?: string;
+  sequenceId?: string;
+}
+
+export interface PayInvoiceParams {
+  invoice: string;
```
sangaman, line 33: We do have `memo`, `sequenceId`, `feeLimitRatio`, and `feeLimit` as optional parameters as well when paying a lightning invoice via the API.

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -0,0 +1,164 @@
...
+  expiration?: number;
+}
+
+export interface CreateDepositAddressParams {
+  amount: number;
+  memo?: string;
+  expiration?: number;
+}
+
+export interface GetInvoicesParams {
```
sangaman, line 15: Is this going to query our transfers? I'm not sure what this call will be doing.

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -0,0 +1,164 @@
...
+export interface CreateInvoiceParams {
+  amount: number;
+  memo?: string;
+  expiration?: number;
+}
+
+export interface CreateDepositAddressParams {
+  amount: number;
+  memo?: string;
+  expiration?: number;
```
sangaman, line 12: I don't think the deposit address will ever expire so we can drop this.

sangaman: We don't have a `memo` either.

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -0,0 +1,164 @@
...
+  'PayInvoiceResponse'
+);
+
+// eslint-disable-next-line no-redeclare
+export type PayInvoiceResponse = t.TypeOf<typeof PayInvoiceResponse>;
+
+export const GetBalanceResponse = t.strict(
+  {
+    balance: t.number,
+    availableBalance: t.number,
```
sangaman, line 126: We're also returning a `maximumBalance` value - the max balance the wallet is allowed to have for lightning.

sangaman (changes requested):
Looks like my review is just a few minutes late her, but I noticed some changes I think we should make. Maybe for a follow-up PR.

modules/sdk-core/src/bitgo/lightning/iLightning.ts
tylerlevine: it also doesn't have an amount

modules/sdk-core/src/bitgo/lightning/iLightning.ts
tylerlevine: yes, that is the plan. The behavior should be to allow a user to specify if they like, and automatically do the right thing if they don't specify.

PR #2617, opened 2022-08-04
feat(abstract-utxo): add support for bigints from new utxo-lib

williamqin-bitgo:
BG-49870
BG-49872


modules/bitgo/test/v2/unit/coins/utxo/util/unspents.ts
```diff
@@ -72,28 +72,31 @@ export function mockWalletUnspent(
   };
 }
 
-export function mockUnspentReplayProtection(network: utxolib.Network): Unspent {
+export function mockUnspentReplayProtection<TNumber extends number | bigint = number>(
+  network: utxolib.Network,
+  value: TNumber = 1000 as TNumber
```
OttoAllmendinger, line 77: The cast is invalid. Let's remove this default argument.

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -251,6 +254,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       );
     }
     this._network = network;
+    this._amountType = amountType;
```
OttoAllmendinger, line 257: since this is fully dependent on network let's do this instead

```suggestion
    this.amountType = utxolib.getMainnet(network) === utxolib.networks.dogecoin ? 'bigint' : 'number';
```

also the leading underscore is not needed here

OttoAllmendinger (changes requested):
some minor nits

modules/abstract-utxo/src/abstractUtxoCoin.ts
williamqin-bitgo: oh, I thought the leading underscore was meant for `private` variables

regarding the network dependency, I think the implementation is cleaner with the `amountType` being determined by each coin's individual override code, i.e. see `modules/sdk-coin-doge/src/doge.ts`

that way, there isn't any mention of `doge` within `abstract-utxo`

OttoAllmendinger: ok fair enough

maybe we should add `utxolib.getAmountType(network): 'number' | 'bigint'`  at some point

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -241,8 +243,9 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   public altScriptHash?: number;
   public supportAltScriptDestination?: boolean;
   private readonly _network: utxolib.Network;
+  private readonly _amountType: 'number' | 'bigint';
 
-  protected constructor(bitgo: BitGoBase, network: utxolib.Network) {
+  protected constructor(bitgo: BitGoBase, network: utxolib.Network, amountType: 'number' | 'bigint' = 'number') {
```
OttoAllmendinger, line 248: shouldn't we add the type parameter at the class level?

```
export abstract class AbstractUtxoCoin<TNumber extends bigint | number = number> extends BaseCoin`
```

and then for doge we can do `Doge extends AbstractUtxoCoin<bigint>`

OttoAllmendinger (changes requested):

modules/abstract-utxo/src/abstractUtxoCoin.ts
williamqin-bitgo: We could, but this adds quite a few complications because then every single function that takes in `AbstractUtxoCoin` needs to be updated to expect either `AbstractUtxoCoin<bigint>` or `AbstractUtxoCoin<number>`. (I actually tried this first)
Additionally, this isn't necessary because there are no class properties that have to use the `TNumber` generic and we only need to change the interface for a few select functions.

OttoAllmendinger: ok I see

OttoAllmendinger (dismissed):

williamqin-bitgo:
Updated `_amountType` to be public (but still readonly) and changed name to `amountType`

modules/abstract-utxo/src/abstractUtxoCoin.ts
williamqin-bitgo: also I ended up changing it to `public` anyways, removed underscore

OttoAllmendinger (approved):

mmcshinsky-bitgo (approved):

PR #2618, opened 2022-08-04
fix(sdk-core): Fix sendTokenEnablements 

kevinwan-bitgo:
Ticket: BG-54114

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -2294,7 +2294,7 @@ export class Wallet implements IWallet {
       this.bitgo.setRequestTracer(params.reqId);
     }
 
-    const buildParams = _.pick(params, this.prebuildWhitelistedParams());
+    const buildParams: PrebuildTransactionOptions = _.pick(params, this.prebuildWhitelistedParams());
     buildParams.type = 'enabletoken';
     // Check if we build with intent
     if (this._wallet.multisigType === 'tss') {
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2351,12 +2354,10 @@ export class Wallet implements IWallet {
...
-      // fold any of the parameters we used to build this transaction into the unsignedBuild
-      const unsignedBuildWithOptions: PrebuildAndSignTransactionOptions = _.pick(
-        params,
-        this.prebuildWhitelistedParams()
-      );
-      unsignedBuildWithOptions.prebuildTx = unsignedBuild;
+      const unsignedBuildWithOptions: PrebuildAndSignTransactionOptions = {
+        ...params,
+        prebuildTx: unsignedBuild,
+      };
```
joey-bai, line 2360: we should probably filter based on `prebuildWhitelistedParams`

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2304,11 +2304,14 @@ export class Wallet implements IWallet {
       buildParams.recipients = params.enableTokens.map((token) => {
         return {
           tokenName: token.name,
-          address: token.address,
+          address: token.address || this._wallet.coinSpecific!.baseAddress!,
+          amount: '0',
```
joey-bai, line 2308: do we need to set this explicitly?
feel like for coins that want to leverage this function, they'll need to update WP accordingly to handle missing `address` and `amount` when building token enablement transactions

joey-bai: if we need this, lets do the following:

ensure user is not passing `amount`. if they are, throw an error like `Can not specify amount when enabling tokens`

additionally, lets nil-check `this._wallet.coinSpecific!.baseAddress` and throw error accordingly:
```
if (!this._wallet.coinSpecific?.baseAddress) {
   throw new Error('Wallet does not have base address, must specify with token param') ;
}
```

kevinwan-bitgo: The user cannot specify amount since we forbid them from specifying `recipients`. Only `address` and `name` are in `TokenEnablement`

joey-bai (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
kevinwan-bitgo: Filtering would remove `walletPassphrase` making us unable to sign the transaction. `sendMany` also calls `prebuildAndSignTransaction` with the full `params` with no issue.

joey-bai (approved):

bitgoAaron (approved):

PR #2623, opened 2022-08-05
fix(sdk-coin-hbar): verify transaction should work without token name

kevinwan-bitgo:
Ticket: BG-54301


joey-bai (approved):
nice find, thanks for fixing!

bitgoAaron (approved):

PR #2624, opened 2022-08-05
feat(sdk-core): add verification of private shares

islamaminBitGo:
Added a verification for share proofs. This is to verify that a u share that is being shared to the destined entity was actually sent by the sender of that u share.

### DIFF
--- a/modules/bitgo/test/v2/unit/internal/opengpgUtils.ts
+++ b/modules/bitgo/test/v2/unit/internal/opengpgUtils.ts
@@ -61,6 +61,53 @@ describe('OpenGPG Utils Tests', function () {
     });
   });
 
+  describe('verifyShareProof positive case', function () {
+    it('should be able to verify a valid proof', async function () {
+      const uValue = crypto.randomBytes(32).toString('hex');
+      const proof = await openpgpUtils.createShareProof(senderKey.privateKey, uValue);
+      // verify proof
+      const isValid = await openpgpUtils.verifyEdShareProof(senderKey.publicKey, proof, uValue);
+      isValid.should.be.true();
+    });
+  });
+
+  describe('verifyShareProof attack proof', function () {
+    it('should be able to detect sender is an attacker', async function () {
+      const uValue = crypto.randomBytes(32).toString('hex');
+      const proof = await openpgpUtils.createShareProof(otherKey.privateKey, uValue);
+      // verify proof
+      const isValid = await openpgpUtils.verifyEdShareProof(senderKey.publicKey, proof, uValue);
+      isValid.should.be.false();
+    });
+  });
+
+  describe('verifyShareProof attack u value', function () {
+    it('should be able to detect u value is corrupted', async function () {
+      const uValue = crypto.randomBytes(32).toString('hex');
+      const proof = await openpgpUtils.createShareProof(senderKey.privateKey, uValue);
+      const uValueCorrupted = crypto.randomBytes(32).toString('hex');
+      // verify proof
+      const isValid = await openpgpUtils.verifyEdShareProof(senderKey.publicKey, proof, uValueCorrupted);
+      isValid.should.be.false();
+    });
+  });
+
+  describe('verifySharedDataProof and createSharedDataProof test', function () {
+    it('should be able to detect if proof value is corrupted or not', async function () {
+      const sharedData1 = crypto.randomBytes(32).toString('hex');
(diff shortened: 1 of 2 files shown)


bitgoAaron (changes requested):
Needs unit tests

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -32,6 +32,33 @@ export async function getBitgoGpgPubKey(bitgo: BitGoBase): Promise<Key> {
...
+ * @return {boolean} whether uValue proof actually was signed by sender as part of their subkeys
+ */
+export async function verifyShareProof(
+  senderPubKey: string,
+  privateShareProof: string,
+  uValue: string
+): Promise<boolean> {
+  const decodedProof = await pgp.readKey({ armoredKey: privateShareProof });
+  const otherOvcGpgKey = await pgp.readKey({ armoredKey: senderPubKey });
+  const isValid = (await decodedProof.verifyPrimaryUser([otherOvcGpgKey]))[0].valid;
```
Logicwax, line 50: why bother making a variable if you're just going to use it once one line below?   this can just be put into an if statement

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -32,6 +32,33 @@ export async function getBitgoGpgPubKey(bitgo: BitGoBase): Promise<Key> {
...
+ * @param uValue u value from an Eddsa keyshare
+ * @return {boolean} whether uValue proof actually was signed by sender as part of their subkeys
+ */
+export async function verifyShareProof(
+  senderPubKey: string,
+  privateShareProof: string,
+  uValue: string
+): Promise<boolean> {
+  const decodedProof = await pgp.readKey({ armoredKey: privateShareProof });
+  const otherOvcGpgKey = await pgp.readKey({ armoredKey: senderPubKey });
```
Logicwax, line 49: this is the SDK, which means this can be used in other contexts (not just in OVC, but in hot wallets).   so remove the reference to "ovc" and use something akin to user/backup/bitgo since this will need to be used in hot wallets as well

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -32,6 +32,33 @@ export async function getBitgoGpgPubKey(bitgo: BitGoBase): Promise<Key> {
...
+  const isValid = (await decodedProof.verifyPrimaryUser([otherOvcGpgKey]))[0].valid;
+  if (!isValid) {
+    return false;
+  }
+  const proofSubkeys = decodedProof.getSubkeys()[1];
+  const decodedUValueProof = Buffer.from(proofSubkeys.keyPacket.publicParams['Q'].slice(1)).toString('hex');
+  const rawUValueProof = Buffer.from(
+    sodium.crypto_scalarmult_ed25519_base_noclamp(Buffer.from(uValue, 'hex'))
```
Logicwax, line 57: when we deploy ECDSA TSS, we will need the equivalent math operation here.    which means we will probably need to make separate proof functions for ed and ec.    I suggest changing the function name to something like verifyEdShareProof or something along those lines.

bitgoAaron (approved):

modules/bitgo/test/v2/unit/internal/opengpgUtils.ts
```diff
@@ -61,6 +61,58 @@ describe('OpenGPG Utils Tests', function () {
...
+    });
+  });
+
+  describe('verifySharedDataProof and createSharedDataProof test', function () {
+    it('should be able to detect if proof value is corrupted or not', async function () {
+      const sharedData1 = crypto.randomBytes(32).toString('hex');
+      const sharedData2 = crypto.randomBytes(32).toString('hex');
+      const dataToProofArray = [{ name: 's1', value: sharedData1 },
+        { name: 's2', value: sharedData2 }];
+      const dataToProofMap = new Map(
```
Logicwax, line 101: why not pass in dataToProofArray directly?

islamaminBitGo: i had written the verification code to use a map as input, i updated the input now to take in array of key val.

Logicwax (approved):

bitgoAaron (approved):

PR #2625, opened 2022-08-08
fix(sdk-core): fix lightning requests params

sj3014:
- GetInvoices and CreateDepositAddress do not take params
- add required params for LightningWithdrawal
- add optional params for PayInvoice
- add maximumBalance to response for GetBalance

Ticket: BG-54079

### DIFF
--- a/modules/sdk-core/src/bitgo/lightning/iLightning.ts
+++ b/modules/sdk-core/src/bitgo/lightning/iLightning.ts
@@ -6,17 +6,6 @@ export interface CreateInvoiceParams {
   expiration?: number;
 }
 
-export interface CreateDepositAddressParams {
-  amount: number;
-  memo?: string;
-  expiration?: number;
-}
-
-export interface GetInvoicesParams {
-  id?: string;
-  status?: 'open' | 'closed';
-}
-
 export interface LightningWithdrawalParams {
   amount: number;
   address?: string;
@@ -31,6 +20,10 @@ export interface LightningDepositParams {
 
 export interface PayInvoiceParams {
   invoice: string;
+  sequenceId?: string;
+  comment?: string;
+  feeLimitRatio?: number;
+  feeLimit?: number;
 }
 
 export const WPTransferEntry = t.type(
@@ -124,6 +117,7 @@ export const GetBalanceResponse = t.strict(
   {
     balance: t.number,
     availableBalance: t.number,
+    maximumBalance: t.number,
   },
   'GetBalanceResponse'
 );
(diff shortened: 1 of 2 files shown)


tylerlevine:
Follow up for post merge comments on https://github.com/BitGo/BitGoJS/pull/2613

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -155,10 +149,11 @@ export type GetInvoicesResponse = t.TypeOf<typeof GetInvoicesResponse>;
...
 export interface ILightning {
   createInvoice(params?: CreateInvoiceParams): Promise<CreateInvoiceResponse>;
-  createDepositAddress(params?: CreateDepositAddressParams): Promise<CreateDepositAddressResponse>;
+  createDepositAddress(): Promise<CreateDepositAddressResponse>;
   payInvoice(params?: PayInvoiceParams): Promise<PayInvoiceResponse>;
   getBalance(): Promise<GetBalanceResponse>;
   withdraw(params?: LightningWithdrawalParams): Promise<WithdrawResponse>;
   deposit(params?: LightningDepositParams): Promise<DepositResponse>;
-  getInvoices(params?: GetInvoicesParams): Promise<GetInvoicesResponse>;
+  // Implemented in lnd-middleware but not in wallet-platform
```
tylerlevine, line 157: This comment isn't helpful for those who don't have context on the lightning implementation - please remove.

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -155,10 +149,11 @@ export type GetInvoicesResponse = t.TypeOf<typeof GetInvoicesResponse>;
...
+  getInvoices(): Promise<GetInvoicesResponse>;
```
tylerlevine, line 158: I think we should just remove this function from the interface - it needs to take some params and we don't know what they are yet. We can add it later once we added the necessary backend support.

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -6,21 +6,10 @@ export interface CreateInvoiceParams {
...
-  id?: string;
-  status?: 'open' | 'closed';
-}
-
 export interface LightningWithdrawalParams {
   amount: number;
-  address?: string;
-  sequenceId?: string;
+  address: string;
+  sequenceId: string;
```
tylerlevine, line 12: These should be optional

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -49,7 +47,7 @@ export class Lightning implements ILightning {
     throw new Error('method not implemented');
   }
 
-  public async getInvoices(params?: GetInvoicesParams): Promise<GetInvoicesResponse> {
+  public async getInvoices(): Promise<GetInvoicesResponse> {
```
tylerlevine, line 50: Let's remove until we're ready to implement this and we know what the params should be.

tylerlevine (changes requested):

modules/sdk-core/src/bitgo/lightning/iLightning.ts
sj3014: Should we also remove GetInvoicesParams and GetInvoicesResponse types?

tylerlevine: yeah, let's remove all the code related to `getInvoices()` and we'll add it all back when we're ready.

tylerlevine (approved):

PR #2627, opened 2022-08-08
feat(sdk-coin-polygon): add signTx method

gianchandania:
This PR adds `signTransaction()` method in Polygon to solve issue in Polygon transfers and updates the test cases accordingly

[BG-54459](https://bitgoinc.atlassian.net/browse/BG-54459)

### DIFF
--- a/modules/sdk-coin-polygon/src/polygon.ts
+++ b/modules/sdk-coin-polygon/src/polygon.ts
@@ -11,6 +11,8 @@ import {
   optionalDeps,
   BuildTransactionParams,
   SignFinalOptions,
+  SignTransactionOptions,
+  SignedTransaction,
 } from '@bitgo/sdk-coin-eth';
 import { BaseCoin, BitGoBase, TransactionExplanation, FullySignedTransaction } from '@bitgo/sdk-core';
 import { BaseCoin as StaticsBaseCoin, coins } from '@bitgo/statics';
@@ -276,5 +278,32 @@ export class Polygon extends Eth {
     const ethTx = unsignedEthTx.sign(signingKey);
 
     return { txHex: ethTx.serialize().toString('hex') };
+  }
+
+  /**
+   * Assemble half-sign prebuilt transaction
+   * @param params
+   */
+  async signTransaction(params: SignTransactionOptions): Promise<SignedTransaction> {
+    const txBuilder = this.getTransactionBuilder();
+    txBuilder.from(params.txPrebuild.txHex);
+    txBuilder.transfer().key(new KeyPair({ prv: params.prv }).getKeys().prv!);
+    const transaction = await txBuilder.build();
+
+    const recipients = transaction.outputs.map((output) => ({ address: output.address, amount: output.value }));
+
+    const txParams = {
+      eip1559: params.txPrebuild.eip1559,
+      txHex: transaction.toBroadcastFormat(),
+      recipients: recipients,
+      expiration: params.txPrebuild.expireTime,
+      hopTransaction: params.txPrebuild.hopTransaction,
+      custodianTransactionId: params.custodianTransactionId,
+      expireTime: params.expireTime,
+      contractSequenceId: params.txPrebuild.nextContractSequenceId as number,
+      sequenceId: params.sequenceId,
+    };
+
+    return { halfSigned: txParams };
   }
 }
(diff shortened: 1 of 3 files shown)


wesgraham (dismissed):

modules/sdk-coin-polygon/src/polygon.ts
```diff
@@ -12,13 +12,53 @@ import {
...
 import { KeyPair, TransactionBuilder } from './lib';
 import _ from 'lodash';
 import type * as EthTxLib from '@ethereumjs/tx';
 
+interface EIP1559 {
+  maxPriorityFeePerGas: number;
+  maxFeePerGas: number;
+}
+
+interface TransactionPrebuild extends BaseTransactionPrebuild {
```
mmcshinsky-bitgo, line 35: Does `TransactionPrebuild` from abstract-eth work better than the one from core?

modules/sdk-coin-polygon/src/polygon.ts
```diff
@@ -12,13 +12,53 @@ import {
...
+  feeInfo?: EthTransactionFee;
+  source?: string;
+  dataToSign?: string;
+  nextContractSequenceId?: string | number;
+  expireTime?: number;
+  hopTransaction?: string;
+  eip1559?: EIP1559;
+}
+
+interface PolygonSignTransactionOptions extends SignTransactionOptions {
```
mmcshinsky-bitgo, line 45: Same with `EthSignTransactionOptions` from abstract-eth?

modules/sdk-coin-polygon/src/polygon.ts
```diff
@@ -12,13 +12,53 @@ import {
...
+  txPrebuild: TransactionPrebuild;
+  prv: string;
+  custodianTransactionId?: string;
+}
+
+interface HalfSignedEthLikeTransaction extends HalfSignedAccountTransaction {
```
mmcshinsky-bitgo, line 51: `HalfSignedEthLikeTransaction` as well. Making sure on these since `sdk-coin-polygon` has `abstract-eth` as a dependency with exposed interfaces that may align better (if desired) from there.

mmcshinsky-bitgo: And Polygon already extends Eth so it matches up from the start, generally.

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
```diff
@@ -265,9 +265,9 @@ export interface HalfSignedAccountTransaction {
 export interface HalfSignedTransaction extends HalfSignedAccountTransaction {
   halfSigned: {
     recipients: Recipient[];
-    expireTime: number;
-    contractSequenceId: number;
-    sequenceId: number;
+    expireTime?: number;
```
mmcshinsky-bitgo, line 268: If pulling from abstract-eth, you want have to affect the core logic.

modules/sdk-coin-polygon/src/polygon.ts
gianchandania: Have used TransactionPrebuild from abstract-eth

modules/sdk-coin-polygon/src/polygon.ts
gianchandania: Have updated to use EthSignTransactionOptions from abstract-eth

modules/sdk-coin-polygon/src/polygon.ts
gianchandania: Have updated to use HalfSignedEthLikeTransaction from abstract-eth

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
gianchandania: Actually will still have to update this because signTransaction method in the base class Eth is actually using HalfSignedTransaction from sdk-core. So it is giving me error 

expireTime, contractSequenceId, sequenceId fields are missing from the type. These fields are not required in abstract-eth or polygon. So have made this change

mmcshinsky-bitgo: You can use `Partial<>` to extend and make optional required fields instead of changing the original intent.

gianchandania: Have removed the changes in the sdk-core

sachushaji (approved):

mmcshinsky-bitgo (approved):

PR #2630, opened 2022-08-09
feat(sdk-coin-polygon): support tokens

mullapudipruthvik:
Ticket: BG-54521

Adding skeleton to support Polygon tokens
Tested locally as well
`let token = await bitgo.coin('DERC20');`


modules/bitgo/test/browser/browser.spec.ts
```diff
@@ -19,6 +19,7 @@ describe('Coins', () => {
             'EosToken': 1,
             'AlgoToken': 1,
             'AvaxCToken': 1,
+            'PolygonToken': 1,
```
mullapudipruthvik, line 22: Browser tests were failing if I didn't include PolygonToken here 
I see we have includes other tokens as well

wesgraham (approved):

PR #2634, opened 2022-08-09
feat(sdk-core): add createDepositAddress to lightning

sj3014:
- it generates lightning deposit address

Ticket: BG-54081


tylerlevine (approved):
Nice 👍

PR #2636, opened 2022-08-09
feat(sdk-core): add createInvoice to lightning

sj3014:
- createInvoice() in lightning generates a lightning invoice
- a test code lives in /bitgo/test

Ticket: BG-54080

### DIFF
--- a/modules/sdk-core/src/bitgo/lightning/lightning.ts
+++ b/modules/sdk-core/src/bitgo/lightning/lightning.ts
@@ -23,8 +23,15 @@ export class Lightning implements ILightning {
     this.url = this.bitgo.url(`/wallet/${walletId}/lightning`, 2);
   }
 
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -24,7 +24,14 @@ export class Lightning implements ILightning {
   }
 
   public async createInvoice(params?: CreateInvoiceParams): Promise<CreateInvoiceResponse> {
```
mmcshinsky-bitgo, line 26: Want to make sure. `value` is set as required but the params are optional. Is this intended?

sj3014: The params should be required. Fixed. Thank you!

mmcshinsky-bitgo (approved):

PR #2639, opened 2022-08-09
feat(abstract-utxo): cross chain recovery support for bigint coins (doge)

williamqin-bitgo:
BG-49873

depends on https://github.com/BitGo/BitGoJS/pull/2617

### DIFF
--- a/modules/abstract-utxo/src/config.ts
+++ b/modules/abstract-utxo/src/config.ts
@@ -1,8 +1,9 @@
 // Supported cross-chain recovery routes. The coin to be recovered is the index, the valid coins for recipient wallets
 // are listed in the array.
 export const supportedCrossChainRecoveries = {
-  btc: ['bch', 'ltc', 'bsv'],
-  bch: ['btc', 'ltc', 'bsv'],
-  ltc: ['btc', 'bch', 'bsv'],
-  bsv: ['btc', 'ltc', 'bch'],
+  btc: ['bch', 'ltc', 'bsv', 'doge'],
+  bch: ['btc', 'ltc', 'bsv', 'doge'],
+  ltc: ['btc', 'bch', 'bsv', 'doge'],
+  bsv: ['btc', 'ltc', 'bch', 'doge'],
+  doge: ['btc', 'bch', 'ltc', 'bsv'],
 };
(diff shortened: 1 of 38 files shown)


modules/abstract-utxo/src/config.ts
```diff
@@ -1,8 +1,9 @@
...
 export const supportedCrossChainRecoveries = {
-  btc: ['bch', 'ltc', 'bsv'],
-  bch: ['btc', 'ltc', 'bsv'],
-  ltc: ['btc', 'bch', 'bsv'],
-  bsv: ['btc', 'ltc', 'bch'],
+  btc: ['bch', 'ltc', 'bsv', 'doge'],
+  bch: ['btc', 'ltc', 'bsv', 'doge'],
+  ltc: ['btc', 'bch', 'bsv', 'doge'],
+  bsv: ['btc', 'ltc', 'bch', 'doge'],
+  doge: ['btc', 'bch', 'ltc', 'bsv'],
```
OttoAllmendinger, line 8: is this really a concern? do people send dogecoin to ltc addresses for instance? from what I know this was only a problem for coins with shared address prefixes

OttoAllmendinger (approved):

modules/abstract-utxo/src/config.ts
williamqin-bitgo: I believe these coins don't share address prefixes, but I'm not sure if cross chain recovery is necessary.
This was added as a task to our doge epic, and it also seems like it could be useful.

Also, could I verify with you that the fee rate of `1000` for doge is correct? Doge has a minimum fee of `0.01 DOGE/kB`, but i've been able to send transactions on a regtest node at a fee of `0.001 DOGE/kB` (which is the minimum relay fee).

OttoAllmendinger: let's stick with 1000

williamqin-bitgo:
please don't merge yet, need to wait for https://github.com/BitGo/BitGoJS/pull/2617

williamqin-bitgo:
For some reason the base branch didn't auto update

williamqin-bitgo:
rebased and force pushed to rerun checks

OttoAllmendinger (approved):

byhow (approved):

williamqin-bitgo:
don't merge this until issues in CR-686 are fixed

williamqin-bitgo:
CR-686 underlying causes fixed: https://github.com/BitGo/BitGoJS/pull/2662

mmcshinsky-bitgo (approved):

PR #2643, opened 2022-08-11
fix(sdk-core): ecdsa signing get user share

olibeyene:
Ticket: BG-52026

### DIFF
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/common.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/common.ts
@@ -34,7 +34,7 @@ export async function nockCreateTxRequest(params: { walletId: string, requestBod
     .reply(200, params.response);
 }
 
-export async function nockGetTxRequest(params: {walletId: string, txRequestId: string, response: unknown}): Promise<nock.Scope> {
+export async function nockGetTxRequest(params: {walletId: string, txRequestId: string, response: unknown, notPersist?: boolean}): Promise<nock.Scope> {
   return nock('https://bitgo.fakeurl')
     .persist(true)
     .get(`/api/v2/wallet/${params.walletId}/txrequests?txRequestIds=${params.txRequestId}&latest=true`)
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
@@ -19,8 +19,9 @@ import {
   SignatureShareRecord,
   RequestTracer,
 } from '@bitgo/sdk-core';
-import { keyShares, otherKeyShares } from '../../../fixtures/tss/ecdsaFixtures';
-import { nockGetTxRequest, nockSendSignatureShareWithResponse } from './common';
+import { keyShares, mockAShare, mockDShare, otherKeyShares } from '../../../fixtures/tss/ecdsaFixtures';
+import { nockSendSignatureShareWithResponse } from './common';
+import { nockGetTxRequest } from '../../tss/helpers';
 
 const encryptNShare = ECDSAMethods.encryptNShare;
 type KeyShare = ECDSA.KeyShare;
@@ -101,7 +102,7 @@ describe('TSS Ecdsa Utils:', async function () {
     const baseCoin = bitgo.coin(coinName);
 
     bgUrl = common.Environments[bitgo.getEnv()].uri;
-    
+
     nock(bgUrl)
       .persist()
       .get('/api/v1/client/constants')
@@ -378,14 +379,10 @@ describe('TSS Ecdsa Utils:', async function () {
         response: signatureShareThreeFromBitgo,
       });
       /* END STEP THREE */
-
-      // We could possibly avoid this request, but left for consistency with EDDSA
-      const response = { txRequests: [{ ...txRequest }] };
-      await nockGetTxRequest({ walletId: wallet.id(), txRequestId: txRequest.txRequestId, response });
-
     });
 
     it('signTxRequest should succeed with txRequest object as input', async function () {
+      await setupSignTxRequestNocks(false);
       const signedTxRequest = await tssUtils.signTxRequest({
         txRequest,
         prv: JSON.stringify({
@@ -399,10 +396,7 @@ describe('TSS Ecdsa Utils:', async function () {
     });
 
     it('signTxRequest should succeed with txRequest id as input', async function () {
-      const getTxRequest = sandbox.stub(tssUtils, 'getTxRequest');
-      getTxRequest.resolves(txRequest);
-      getTxRequest.calledWith(txRequestId);
-
+      await setupSignTxRequestNocks();
       const signedTxRequest = await tssUtils.signTxRequest({
         txRequest: txRequestId,
         prv: JSON.stringify({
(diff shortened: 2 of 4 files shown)


modules/bitgo/test/v2/unit/internal/tssUtils/common.ts
```diff
@@ -34,9 +34,9 @@ export async function nockCreateTxRequest(params: { walletId: string, requestBod
     .reply(200, params.response);
 }
 
-export async function nockGetTxRequest(params: {walletId: string, txRequestId: string, response: unknown}): Promise<nock.Scope> {
+export async function nockGetTxRequest(params: {walletId: string, txRequestId: string, response: unknown, notPersist?: boolean}): Promise<nock.Scope> {
   return nock('https://bitgo.fakeurl')
-    .persist(true)
+    .persist(params.notPersist)
```
sachushaji, line 39: This changes correct behaviour in the mocks which is reused for eddsa as well, but should be fine if all tests passes

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -452,6 +449,23 @@ describe('TSS Ecdsa Utils:', async function () {
 
       sandbox.verifyAndRestore();
     });
+
+    async function setupSignTxRequest(isTxRequest = true) {
```
sachushaji, line 453: we could rename may to the below suggestion to give more clarity on what the function does
```suggestion
    async function setupSignTxRequestMocks(isTxRequest = true) {
```

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -413,15 +406,16 @@ describe('TSS Ecdsa Utils:', async function () {
...
-
-      sandbox.verifyAndRestore();
     });
 
     it('signTxRequest should fail with invalid user prv', async function () {
       const getTxRequest = sandbox.stub(tssUtils, 'getTxRequest');
       getTxRequest.resolves(txRequest);
       getTxRequest.calledWith(txRequestId);
 
+      const response = { txRequests: [{ ...txRequest }] };
```
sachushaji, line 416: can we use setupSignTxRequest here?

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -441,6 +435,9 @@ describe('TSS Ecdsa Utils:', async function () {
...
```
sachushaji, line 438: here also

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -452,6 +449,23 @@ describe('TSS Ecdsa Utils:', async function () {
...
+      }
+      const aRecord = ECDSAMethods.convertAShare(mockAShare);
+      const signatureShares = [aRecord];
+      txRequest.signatureShares = signatureShares;
+      response = { txRequests: [{ ...txRequest }] };
+      await nockGetTxRequest({ walletId: wallet.id(), txRequestId: txRequest.txRequestId, response: response });
+      const dRecord = ECDSAMethods.convertDShare(mockDShare);
+      signatureShares.push(dRecord);
+      response = { txRequests: [{ ...txRequest }] };
+      await nockGetTxRequest({ walletId: wallet.id(), txRequestId: txRequest.txRequestId, response: response });
```
sachushaji, line 466: we could use persist here to avoid repeating the same nock twice

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -520,5 +534,10 @@ describe('TSS Ecdsa Utils:', async function () {
     return params.keyChain;
   }
 
+  async function nockGetTxRequest(params: {walletId: string, txRequestId: string, response: unknown }): Promise<nock.Scope> {
```
sachushaji, line 537: we already have a nock function defined in `modules/bitgo/test/v2/unit/internal/tssUtils/common.ts `?

modules/bitgo/test/v2/unit/tss/ecdsa.ts
```diff
@@ -344,28 +344,32 @@ describe('Ecdsa tss helper functions tests', function () {
...
             responseAShare.should.deepEqual(config[index].mockShareAsResponse);
           });
 
           it(`should fail if we get an invalid ${config[index].shareReceived} as response`, async function () {
             const mockSendReq = { from: 'user', to: 'bitgo', share: config[index].mockShareToSendString } as SignatureShareRecord;
             const invalidSignatureShare = { from: 'bitgo', to: 'user', share: JSON.stringify(config[index].incorrectReceivedShareString) } as SignatureShareRecord;
             const nock = await nockSendSignatureShare({ walletId: wallet.id(), txRequestId: txRequest.txRequestId, signatureShare: mockSendReq, response: invalidSignatureShare }, 200);
-            await ECDSAMethods.sendShareToBitgo(bitgo, wallet.id(), txRequest.txRequestId, config[index].sendType, config[index].mockShareToSend, config[index].optionalDShare).should.be.rejectedWith(/Excepted share length to be greater than or equal .*/g);
+            txRequest.signatureShares = [invalidSignatureShare];
+            const response = { txRequests: [{ ...txRequest }] };
```
sachushaji, line 370: NIT: we could reuse that helper function you defined in `modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts` , probably need to add more parameters

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -208,11 +209,50 @@ export async function sendShareToBitgo(
       }
       const sShare = share as SShare;
       signatureShare = convertSDShare({ sShare, dShare });
-      const signatureRecord = await sendSignatureShare(bitgo, walletId, txRequestId, signatureShare);
-      responseFromBitgo = parseSignatureShare(signatureRecord);
+      await sendSignatureShare(bitgo, walletId, txRequestId, signatureShare);
+      responseFromBitgo = sShare;
```
sachushaji, line 213: I think we should get a fully signed signature from bitgo at this point via get Tx request?

olibeyene: tx request will be signed and wp cleans up signature shares. at this point the share don't matter.

sachushaji: Probably good if the client get the full signature as a response as well. Think we already is calling a getTxRequest as last step in the sign function, probably only need to move here may be

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
olibeyene: the persist is causing issues even when disable it. I'll use the other nock which part of helpers

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
olibeyene: they are not the same, the responses are different

modules/bitgo/test/v2/unit/tss/ecdsa.ts
olibeyene: they are in different files. I don't think it's worth reusing that method

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
olibeyene: Yep I agree, the user gets signed tx however not that shares

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
sachushaji: But seems like the response in line 465 is reused in the nocks in lines 466 and 467?

byhow (approved):

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
olibeyene: that part is the same, so nock can handle the response one more time as it's needed for this test.

PR #2644, opened 2022-08-11
feat(sdk-core): add withdraw to lightning object

sj3014:
- it creates a new address if an address is not provided
- it creates a new sequenceId if it is not provided

Ticket: BG-54084

### DIFF
--- a/modules/bitgo/test/v2/fixtures/lightning/lightning.ts
+++ b/modules/bitgo/test/v2/fixtures/lightning/lightning.ts
@@ -8,6 +8,7 @@ export default {
     memo: 'test payment',
     invoice: 'lntb1230n1p309pp3pp54l6up2ggkzamz6t8jcntgdszcz0r3xey93grt3zt8w0kzw5u4vdsdq5w3jhxapqwpshjmt9de6qcqzpgxqrrsssp56rek9prhjj64f6xvkzjffajmaf0dw472knfykafftklv8ljhjs9q9qyyssq2uznk85spdqfmgpsckx9hunxlgsvrzhwknq7ucpcnuan2fhr3fj567psz75vksynk3q03kt7hutm37tce8258hjqndljr0evuf04eyspz5kw0z',
   },
+
   payment: {
     transfer: {
       entries: [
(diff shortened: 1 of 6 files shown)


modules/bitgo/test/v2/fixtures/lightning/lightning.ts
```diff
@@ -75,4 +76,40 @@ export default {
...
+      heightId: '999999999-62f428a4a8fe2800087908e901f26ae9',
+      date: '2022-08-10T21:52:36.147Z',
+      type: 'send',
+      value: -30000,
+      valueString: '-30000',
+      baseValue: -30000,
+      baseValueString: '-30000',
+      payGoFee: 0,
+      payGoFeeString: '0',
+      state: 'pendingApproval',
```
sangaman, line 102: Just a note, this fixture might have to change soon since we may switch from using `pendingApproval` for withdrawals at all. It's fine for now but just wanted point it out.

sangaman (dismissed):
Looks great, TY!

modules/bitgo/test/v2/unit/lightning.ts
```diff
@@ -59,4 +60,13 @@ describe('lightning API requests', function () {
     assert.deepStrictEqual(res, fixtures.payment);
     scope.done();
   });
+  
+  it('should withdraw a value from lightning wallet to regular wallet', async function () {
+    const scope = nock(bgUrl).post(`/api/v2/wallet/${wallet.id()}/lightning/withdrawal`).reply(200, fixtures.withdraw
+    );
```
OttoAllmendinger, line 66: side note - it looks like `prettier` is not set up correctly for this repo

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -59,8 +62,25 @@ export class Lightning implements ILightning {
...
+  public async withdraw(params: LightningWithdrawalParams): Promise<WithdrawResponse> {
+    const { value } = params;
+    let { destination, sequenceId } = params;
+
+    if (destination === undefined) {
+      destination = (await this.wallet.createAddress()).address;
+    }
+
+    if (sequenceId === undefined) {
+      sequenceId = `${Math.round(Math.random() * 1e10)}`;
```
OttoAllmendinger, line 74: doesn't matter much in this case
```suggestion
      sequenceId = crypto.randomBytes(16).toString('hex');
```

OttoAllmendinger (approved):

modules/sdk-core/src/bitgo/lightning/lightning.ts
sj3014: This seems a nicer way to generate `sequenceId`

sangaman: Thought about suggesting something similar. FWIW if we do go this route, 16 bytes is probably overkill since it only needs to be unique per wallet. 8 should be plenty. But that's a nit of a nit.

tylerlevine (approved):

PR #2653, opened 2022-08-12
feat(abstract-utxo): backup key recovery service for doge

williamqin-bitgo:
BG-49873


modules/bitgo/test/v2/unit/coins/utxo/recovery/backupKeyRecovery.ts
```diff
@@ -229,6 +229,10 @@ function run(
 }
 
 utxoCoins.forEach((coin) => {
+  if (coin.getFamily() !== 'doge') {
```
mmcshinsky-bitgo, line 232: Was this intentional or only for testing?

williamqin-bitgo: oh no, only for testing 🤦

mmcshinsky-bitgo (approved):

PR #2656, opened 2022-08-16
fix(account-lib): adding fee and change to sdk-coin-ada transactionBu…

margueriteblair:
…ilder

Ticket: BG-54758


modules/sdk-coin-ada/src/lib/transaction.ts
```diff
@@ -118,7 +118,7 @@ export class Transaction extends BaseTransaction {
    */
   loadInputsAndOutputs(): void {
     const outputs: Entry[] = [];
-    const inputs: Entry[] = [];
```
ewangbitgo, line 121: why change to any here?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -21,7 +22,10 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
   private _transactionOutputs: TransactionOutput[] = [];
   private _txBuilder: CardanoWasm.TransactionBuilder;
   private _signatures: Signature[] = [];
+  private _changeAddress: string;
+  private _senderBalance: string;
```
ewangbitgo, line 26: is totalInputAmount a better name?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -21,7 +22,10 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
   private _transactionOutputs: TransactionOutput[] = [];
   private _txBuilder: CardanoWasm.TransactionBuilder;
   private _signatures: Signature[] = [];
+  private _changeAddress: string;
```
ewangbitgo, line 25: is addressForChange a better name?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -59,6 +63,21 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     return this;
   }
 
+  fee(f: BigNum): BigNum {
```
ewangbitgo, line 66: do we need to set fee as we can calculating fee internally?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -152,6 +231,15 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     this._transaction = transaction;
   }
 
+  /** @inheritdoc */
+  get getFee(): string {
+    return this._fee.to_str();
+  }
+
+  get getChangeAddress(): string {
```
ewangbitgo, line 239: do we need this method?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -102,21 +121,30 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
           CardanoWasm.Address.from_bech32(output.address),
-          CardanoWasm.Value.new(CardanoWasm.BigNum.from_str(output.amount))
+          CardanoWasm.Value.new(amount)
         )
       );
+      totalAmountToSend = totalAmountToSend.checked_add(amount);
     });
+
+    // add extra output for the change
+    const changeAddress = CardanoWasm.Address.from_bech32(this._changeAddress);
```
ewangbitgo, line 138: we needs to check if _changeAddress is undefined. When we reconstruct a txn from serialized data, the change address will not be set, and we can directly use output to build, and we can calculate fee based on the size of serialized data

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
margueriteblair: I just use it to check the fee in the unit tests, which I think is useful, if you think it's not, I will delete

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
margueriteblair: you're right, deleting now

modules/sdk-coin-ada/src/lib/transaction.ts
margueriteblair: Because Entry type does not support the input format, do we have another besides any?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
margueriteblair: yes

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -102,21 +122,32 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+          CardanoWasm.Value.new(amount)
         )
       );
+      totalAmountToSend = totalAmountToSend.checked_add(amount);
     });
+
+    // add extra output for the change
+    if (this._changeAddress) {
+      const changeAddress = CardanoWasm.Address.from_bech32(this._changeAddress);
+      outputs.add(CardanoWasm.TransactionOutput.new(changeAddress, CardanoWasm.Value.new(CardanoWasm.BigNum.zero())));
```
ewangbitgo, line 141: why the amount is zero for change address?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -131,7 +162,62 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+    const txDraft = CardanoWasm.Transaction.new(txBody, witnessSet);
+
+    const linearFee = CardanoWasm.LinearFee.new(
+      CardanoWasm.BigNum.from_str('44'),
+      CardanoWasm.BigNum.from_str('155381')
+    );
+
+    // calculate the fee based off our dummy transaction
+    if (!this._fee) {
+      const fee = CardanoWasm.min_fee(txDraft, linearFee).checked_add(CardanoWasm.BigNum.from_str('308'));
```
ewangbitgo, line 174: why we have to add 308

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
margueriteblair: You start off with making a dummy change address output with zero as a placeholder, then, we calculate the fee. Once we have the fee we can calculate the change value for real

ewangbitgo: will 0 affect the final size calculation?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -131,7 +162,62 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+    const txDraft = CardanoWasm.Transaction.new(txBody, witnessSet);
+
+    const linearFee = CardanoWasm.LinearFee.new(
+      CardanoWasm.BigNum.from_str('44'),
+      CardanoWasm.BigNum.from_str('155381')
+    );
+
+    // calculate the fee based off our dummy transaction
+    if (!this._fee) {
+      const fee = CardanoWasm.min_fee(txDraft, linearFee).checked_add(CardanoWasm.BigNum.from_str('6408'));
```
ewangbitgo, line 174: txDraft needs to add at least one sigature to get correct size, we need to remove this hardcoded value

modules/sdk-coin-ada/test/unit/transactionBuilder.ts
```diff
@@ -47,25 +51,181 @@ describe('ADA Transaction Builder', async () => {
...
     const builtTx = await txBuilder.build();
     should.equal(builtTx.type, TransactionType.Send);
     should.equal(builtTx.id, testData.rawTx.txHash);
-    builtTx.inputs.length.should.equal(0);
     builtTx.outputs.length.should.equal(2);
     builtTx.outputs[0].should.deepEqual(testData.rawTx.outputAddress1);
     builtTx.outputs[1].should.deepEqual(testData.rawTx.outputAddress2);
   });
+
+  it('should use the addSignature interface', async () => {
```
ewangbitgo, line 169: the test that use koios has to be removed and move to sandbox because the test are going to run by CI and should not access outside network

margueriteblair: got it, will remove

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -59,6 +63,21 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     return this;
   }
 
+  changeAddress(addr: string): this {
+    this._changeAddress = addr;
+    return this;
+  }
+
+  senderBalance(b: string): this {
```
ewangbitgo, line 71: can we combine senderBalance and changeAddress into one method because these two value needs to be input together to be used correctly. Maybe TotalInputBalance is a better name than senderBalance because in theory inputs can be just a fraction of sender's total balance.

ewangbitgo (dismissed):

byhow (dismissed):

alia-bitgo (approved):

alia-bitgo (dismissed):

alia-bitgo (approved):

mmcshinsky-bitgo (approved):

PR #2658, opened 2022-08-16
Sync up with our fork of bitcoinjs-lib v6.0.2

brandonblack:
Issues: BG-41154


OttoAllmendinger (dismissed):

modules/utxo-lib/test/bitgo/fixtures/signature/bitcoin/p2tr/parsed-backup-bitgo-fullSigned.json
```diff
@@ -7,8 +7,8 @@
     "e21c29b4a7eeace9c7a8cefb568ca00c86ff9bf5e79e07e5442c29d4a0950d04"
   ],
   "signatures": [
-    "9f74390b6c7f5988772a909fc689a3e2dd680ceb8d54d4caffdd5907132bc415d7d964f4b25ce02690938cd4b62057f80b929aca4155c64db205a336b2e6ce79",
-    "69a969c8df93bc7894d045bc5e5d73f50d815fed7f703f11080f6b46561220156a0eed3382ec0eddbfa2737a29a7b07ba572880ae26f05f56ebca40bd7a5a987"
+    "a2074b68b115e970c8c4bee9cf75c03b82bd3e3086d58801a4adb7b31b9b5f2cea4a5615a4feef74b61db6d765bb814feaf27331fad128dda55cfe106d25d44e",
```
sangaman, line 10: Are these changes all because of the lack of extra nonce data?

sangaman (dismissed):
I just spot checked some of the changes and looked through the commits & their messages, but overall looks good.

brandonblack:
Updated to RC version of @bitgo/bitcoinjs-lib. I think this is ready for SDK team review.

modules/abstract-eth/package.json
```diff
@@ -40,7 +40,7 @@
     "@bitgo/sdk-coin-eth": "^1.1.0",
     "@bitgo/sdk-core": "^1.1.0",
     "@bitgo/statics": "^7.0.0",
-    "bignumber.js": "^8.0.1",
-    "bip32": "^2.0.6"
+    "@bitgo/utxo-lib": "^2.4.1",
```
mmcshinsky-bitgo, line 43: Can we upgrade `bip32` and use that instead of using `utxo-lib`?

brandonblack: Sadly no. BIP32 (and ECPair) can no longer be used standalone, because they require initialization with a chosen secp256k1 library. In utxo-lib, we have the code to wrap noble-secp256k1 appropriately and initialize BIP32 / ECPair, and export them for use elsewhere in the SDK.

modules/utxo-lib/src/noble_ecc.ts
```diff
@@ -0,0 +1,95 @@
...
+function isPoint(p: Uint8Array, xOnly: boolean): boolean {
+  if ((p.length === 32) !== xOnly) return false;
+  try {
+    return !!necc.Point.fromHex(p);
+  } catch (e) {
+    return false;
+  }
+}
+
+const ecc = {
```
mmcshinsky-bitgo, line 38: If this is going to be used across different coins types (eth, utxo, etc), it would seem like it would be better served not being exported from `utxo-lib`.

brandonblack: Great idea! Where do you suggest?

modules/abstract-eth/package.json
mmcshinsky-bitgo: We should deprecate adding features to `utxo-lib` and add new features to other packages like `abstract-utxo` and `sdk-core` since that (and coin specific functionality) will be migrating into.

modules/utxo-lib/src/noble_ecc.ts
mmcshinsky-bitgo: It would seem like `sdk-core` would be better than `abstract-utxo` since the implementation is coin agnostic.

modules/abstract-eth/package.json
brandonblack: What's the difference between `utxo-lib` and `abstract-utxo`?

brandonblack: Ok, that makes sense. Doesn't seem like there are implications for this PR (beyond moving the ecc stuff to core), since the code I'm importing from bitcoinjs-lib is needed by code still in `utxo-lib`.

I'll have to think a bit more about the implications of this for the PSBTification work that I'm starting on.

Possibly it means that for Doge, ZCash, Litecoin, Bitcoin, BTG, BCH and BSV any extensions needed to support each coin in `psbt` will only be made in their specific modules and no PSBT wrapper will be added to `abstract-utxo` nor `utxo-lib`.

modules/utxo-lib/test/bitgo/fixtures/signature/bitcoin/p2tr/parsed-backup-bitgo-fullSigned.json
brandonblack: Yes

modules/utxo-lib/src/noble_ecc.ts
brandonblack: Moving it to sdk-core creates a circular dependency between `utxo-lib` and `sdk-core` :cry:. Might have to live here for now.

mmcshinsky-bitgo: Ah biscuits...

modules/abstract-eth/package.json
OttoAllmendinger: > abstract-utxo is the shared utility between all the utxo coins

On wallet platform as well? The `utxo-lib` package is a pretty low-level wrapper around `bitcoinjs-lib` that adds some additional methods for other networks.

> Basically account-lib and utxo-lib will be completely deprecated as entry points into the sdk ecosystem.

The `utxo-lib` package was never meant as an entrypoint to the sdk ecosystem, but as a low-level systems library that can be used across WP/IMS/SDK etc

mmcshinsky-bitgo: `account-lib` has been stripped to the bones. I don't think `utxo-lib` will go away for quite awhile since it's a low priority item. `abstract-utxo` and `abstract-eth` have taken over the roles of `account-lib` and `utxo-lib` so while it may not be completely deprecated, it may become the same shell that `account-lib` is now.

mmcshinsky-bitgo: We also have been maintaining backwards compatibility for now, so when that does happen, it would probably end up being a shell that exports what it can without causing circular references. Something along those lines, but not on the critical development path currently.

OttoAllmendinger: I'm not sure how `abstract-utxo` has taken over the role of `utxo-lib`. From what I can tell the roles have remained steady over the last couple of years.

mmcshinsky-bitgo: `abstract-utxo` is brand new along with `abstract-eth` due to the modularization efforts. The only reason currently more of `utxo-lib` wasn't ported over was due to time constraints.

OttoAllmendinger: The `abstract-utxo` package may be new, but the `AbstractUtxo` class has existed for years. It exists to implement the `coin` interface defined in BitGoJS. 

The `utxo-lib` library serves a different purpose: it extends `bitcoinjs-lib` (with minimal dependencies) to support more values for the `network` parameter.

OttoAllmendinger (approved):

mmcshinsky-bitgo (approved):

PR #2660, opened 2022-08-17
feat(sdk-core): add getInvoices to lightning object

sj3014:
- add getInvoices to lightning object
- getInvoices route will be added to the next release of wallet-platform repo

Ticket: BG-54087

### DIFF
--- a/modules/sdk-core/src/bitgo/lightning/iLightning.ts
+++ b/modules/sdk-core/src/bitgo/lightning/iLightning.ts
@@ -25,6 +25,11 @@ export interface PayInvoiceParams {
   feeLimit?: number;
 }
 
+export interface GetInvoicesQuery {
+  status?: string;
+  limit?: number;
+}
+
 export const WPTransferEntry = t.partial(
   {
     wallet: t.string,
@@ -128,6 +133,22 @@ export const GetBalanceResponse = t.strict(
 // eslint-disable-next-line no-redeclare
 export type GetBalanceResponse = t.TypeOf<typeof GetBalanceResponse>;
 
+const InvoiceInfo = t.strict({
+  paymentHash: t.string,
+  walletId: t.string,
+  status: t.string,
+  value: t.number,
+  expiresAt: t.string,
+  createdAt: t.string,
+  updatedAt: t.string,
+  amtPaidSats: t.number,
+});
+
+export const GetInvoicesResponse = t.array(InvoiceInfo);
+
+// eslint-disable-next-line no-redeclare
+export type GetInvoicesResponse = t.TypeOf<typeof GetInvoicesResponse>;
+
 export interface ILightning {
   createInvoice(params: CreateInvoiceParams): Promise<CreateInvoiceResponse>;
   createDepositAddress(): Promise<CreateDepositAddressResponse>;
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -128,11 +133,28 @@ export const GetBalanceResponse = t.strict(
...
+  value: t.number,
+  expiresAt: t.string,
+  createdAt: t.string,
+  updatedAt: t.string,
+  amtPaidSats: t.number,
+});
+
+export const GetInvoicesResponse = t.array(InvoiceInfo);
+
+// eslint-disable-next-line no-redeclare
```
OttoAllmendinger, line 149: let's disable this in the `eslintrc` config

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -98,4 +100,18 @@ export class Lightning implements ILightning {
...
+    queryParams.status = query?.status;
+    queryParams.limit = query?.limit;
+
+    const body = await this.bitgo
+      .get(this.url + '/invoices')
+      .query(queryParams)
+      .result();
+    return decodeOrElse(GetInvoicesResponse.name, GetInvoicesResponse, body, (errors) => {
+      throw new Error(`error(s) parsing response body: ${errors}`);
+    });
```
OttoAllmendinger, line 115: can we make this more generic

```suggestion
    return this.request('get', '/invoices', queryParams, GetInvoicesResponse.name);
```

we will save a lot of boilerplate that way

(can happen in next PR)

OttoAllmendinger (approved):

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -98,4 +100,18 @@ export class Lightning implements ILightning {
       throw new Error(`error(s) parsing response body: ${errors}`);
     });
   }
+
+  public async getInvoices(query?: GetInvoicesQuery): Promise<GetInvoicesResponse> {
+    const queryParams: { status?: string; limit?: number } = {};
+    queryParams.status = query?.status;
+    queryParams.limit = query?.limit;
```
sangaman, line 107: Nit but I think this is a tiny bit cleaner:

```suggestion
    const queryParams = {
      status: query?.status,
      limit: query?.limit
    }
```
```

sangaman (approved):
Looks good, thanks.

modules/sdk-core/src/bitgo/lightning/iLightning.ts
sj3014: `eslintrc` for `sdk-core`?

sj3014: @mmcshinsky-bitgo mentioned in my earlier pr that he doesn't prefer to disable`no-redeclare` for the file since this pattern is only specific to `io-ts`

OttoAllmendinger: fair enough

mmcshinsky-bitgo (approved):

PR #2663, opened 2022-08-17
fix: add ability to use unencrypted private key

mmcshinsky-bitgo:
Users with unencrypted keys using external signing will now be able to
identify and use wallet credentials to manual sign transactions where an
encrypted key is unavailable

BG-54633


bitgoAaron (changes requested):
Can you please add a unit test that verifies this works as expected

bitgoAaron (approved):

PR #2669, opened 2022-08-18
feat(statics): add ETHw statics

fperalta-simtlix:
Add ETHw coin to statics. This is to be able to withdraw from a possible ETH PoW fork. There is no
fork confirmed yet.

BG-55420


modules/statics/src/coins.ts
```diff
@@ -108,6 +108,7 @@ export const coins = CoinMap.fromCoins([
     ETH_FEATURES_WITH_STAKING
   ),
   account('eth2', 'Ethereum 2.0', Networks.main.ethereum2, 18, UnderlyingAsset.ETH2, ETH2_FEATURES, KeyCurve.BLS),
+  account('ethw', 'Ethereum PoW', Networks.main.ethereumW, 18, UnderlyingAsset.ETHW, [...AccountCoin.DEFAULT_FEATURES]),
```
wesgraham, line 111: nit: do we want ETH_FEATURES?

modules/statics/src/networks.ts
```diff
@@ -430,6 +430,17 @@ class Ethereum2 extends Mainnet implements AccountNetwork {
   accountExplorerUrl = 'https://beaconscan.com/address';
 }
 
+class EthereumW extends Mainnet implements EthereumNetwork {
+  name = 'Ethereum PoW';
+  family = CoinFamily.ETHW;
+  explorerUrl = '';
+  accountExplorerUrl = '';
+  chainId = NaN; // This would be updated with the chain ID of a ETH PoW fork if it exists after the merge
```
wesgraham, line 438: Can we add a TODO with ticket num https://bitgoinc.atlassian.net/browse/BG-55520 BG-55520?

modules/statics/src/coins.ts
fperalta-simtlix: `ETH_FEATURES` adds these two features to the account default features, `SUPPORTS_TOKENS` and `ENTERPRISE_PAYS_FEES`. Since we only want to support recovery, I think we don't need them. Is that right?

wesgraham (approved):

mmcshinsky-bitgo:
Looks like it is missing as being a registered coin in the BitGo coin factory. Is this intended?

fperalta-simtlix:
> Looks like it is missing as being a registered coin in the `bitgo` coin factory. Is this intended?

Oh, yes, it is in this other PR https://github.com/BitGo/BitGoJS/pull/2677.

PR #2672, opened 2022-08-18
fix: allow token enablement for cold wallet

lcovar:
Allow signing cold wallet token enablement transactions. For cold wallets users have to pass their private key in plain text format in the request body like `prv: "<prvKey>"`

i.e:

```
  const result = await wallet.sendTokenEnablements({
    enableTokens: [{ name: 'thbar:usdc' }],
    prv: 'prvKey',
  });
```

TICKET: BG-0000

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -2343,9 +2343,8 @@ export class Wallet implements IWallet {
    * Signs and sends a single unsigned token enablement transaction
    * @param params
    * @returns
-   *   - The response from sending the transaction for hot wallets
+   *   - The response from sending the transaction for hot/cold wallets
    *   - The response from initiating the transaction for custodial wallets
-   *   - The included prebuilt transaction for cold wallets
    */
   public async sendTokenEnablement(params: PrebuildAndSignTransactionOptions = {}): Promise<any> {
     const teConfig = this.baseCoin.getTokenEnablementConfig();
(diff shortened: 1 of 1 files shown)


joey-bai (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2362,11 +2362,9 @@ export class Wallet implements IWallet {
     } else {
       switch (this._wallet.type) {
         case 'hot':
+        case 'cold':
```
mmcshinsky-bitgo, line 2365: Need to update the jsdocs.

vmccarty: We don't currently have anything in the upcoming [dev portal](https://developers.bitgo-dev.com/) on token-enablement transactions. Talked with Mike on Slack and it sounds like we maybe should. Will add a Jira to our epic, [BG-46930](https://bitgoinc.atlassian.net/browse/BG-46930), if such a topic is needed.

mmcshinsky-bitgo: Here's a ticket for that as well: [BG-55694](https://bitgoinc.atlassian.net/browse/BG-55694)

lcovar: @mmcshinsky-bitgo updated jsdocs

mmcshinsky-bitgo (approved):

PR #2673, opened 2022-08-19
feat(btc): keypair

argjv:
Add a KeyPair class for BTC with identical behavior to the account-based coins in an effort to generalize key management across assets.

### DIFF
--- a/modules/sdk-coin-btc/src/lib/keyPair.ts
+++ b/modules/sdk-coin-btc/src/lib/keyPair.ts
@@ -1,0 +1,108 @@
+import { randomBytes } from 'crypto';
+import * as bip32 from 'bip32';
+import {
+  AddressFormat as BaseAddressFormat,
+  DefaultKeys,
+  isPrivateKey,
+  isPublicKey,
+  isSeed,
+  KeyPairOptions,
+  NotImplementedError,
+  NotSupported,
+  Secp256k1ExtendedKeyPair,
+} from '@bitgo/sdk-core';
+import { address, crypto, networks } from 'bitcoinjs-lib';
+
+const DEFAULT_SEED_SIZE_BYTES = 16;
+const COMPRESSED_PUBLIC_KEY_BYTE_LENGTH = 33;
+
+// TODO: extend the AddressFormat enum in @bitgo/sdk-core with these values
+enum ExtendedAddressFormat {
+  mainnetBech32 = 'mainnetBech32',
+  testnetBech32 = 'testnetBech32',
+  mainnetMultisig = 'mainnetBech32',
+  testnetMultisig = 'testnetBech32',
+}
+
+export type AddressFormatType = BaseAddressFormat | ExtendedAddressFormat;
+export const AddressFormat = { ...BaseAddressFormat, ...ExtendedAddressFormat };
+
+/** Bitcoin key management. */
+export class KeyPair extends Secp256k1ExtendedKeyPair {
+  /**
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   * @param { KeyPairOptions } source Either a master seed, a private key (extended or raw), or a public key
+   *     (extended, compressed, or uncompressed)
+   */
+  constructor(source?: KeyPairOptions) {
+    super(source);
+    if (!source) {
+      const seed = randomBytes(DEFAULT_SEED_SIZE_BYTES);
+      this.hdNode = bip32.fromSeed(seed);
+    } else if (isSeed(source)) {
+      this.hdNode = bip32.fromSeed(source.seed);
+    } else if (isPrivateKey(source)) {
+      this.recordKeysFromPrivateKey(source.prv);
+    } else if (isPublicKey(source)) {
+      this.recordKeysFromPublicKey(source.pub);
+      // TODO: move this to Secp256k1ExtendedKeyPair
+      this.keyPair.compressed = Buffer.from(source.pub, 'hex').length === COMPRESSED_PUBLIC_KEY_BYTE_LENGTH;
+    } else {
+      throw new Error('Invalid key pair options');
+    }
+
+    if (this.hdNode) {
+      this.keyPair = Secp256k1ExtendedKeyPair.toKeyPair(this.hdNode);
+    }
+  }
+
+  /**
+   * Bitcoin uncompressed public and private keys in hex format.
+   *
+   * @returns { DefaultKeys } The keys in the protocol default key format
+   */
+  getKeys(): DefaultKeys {
+    return {
+      pub: this.getPublicKey({ compressed: this.keyPair.compressed }).toString('hex'),
+      prv: this.getPrivateKey()?.toString('hex'),
+    };
+  }
+
+  /**
+   * Get a bitcoin address in legacy or bech32 format.
+   *
+   * @param {AddressFormatType} format One of mainnet, testnet, mainnetBech32, or testnetBech32. mainnetMultisig is not
+   *    supported since it requires multiple key pairs.
+   * @returns {string} The address derived from the public key
+   */
+  getAddress(format?: AddressFormatType): string {
+    const compressed = this.keyPair.compressed;
+    const publicKeyHash160 = crypto.ripemd160(crypto.sha256(this.getPublicKey({ compressed })));
+
+    if (!format) {
+      // Return mainnet legacy addresses by default
+      return address.toBase58Check(publicKeyHash160, networks.bitcoin.pubKeyHash);
+    }
+
+    switch (format) {
+      case AddressFormat.mainnet:
+        return address.toBase58Check(publicKeyHash160, networks.bitcoin.pubKeyHash);
+      case AddressFormat.testnet:
+        return address.toBase58Check(publicKeyHash160, networks.testnet.pubKeyHash);
+      case AddressFormat.mainnetBech32:
+        return address.toBech32(publicKeyHash160, 0, networks.bitcoin.bech32);
+      case AddressFormat.testnetBech32:
+        return address.toBech32(publicKeyHash160, 0, networks.testnet.bech32);
+      case AddressFormat.mainnetMultisig:
+      case AddressFormat.testnetMultisig:
+        throw new NotSupported(
+          'Unsupported address format: ' +
+            format +
+            '. Multisig addresses require multiple key sets and is not supported in the KeyPair class '
+        );
+      default:
+        throw new NotImplementedError('Unsupported address format: ' + format);
+    }
+  }
+}
--- a/modules/sdk-coin-btc/src/index.ts
+++ b/modules/sdk-coin-btc/src/index.ts
@@ -1,2 +1,3 @@
 export * from './btc';
 export * from './tbtc';
+export * from './lib';
--- a/modules/sdk-coin-btc/src/lib/index.ts
+++ b/modules/sdk-coin-btc/src/lib/index.ts
@@ -1,0 +1,1 @@
+export { KeyPair } from './keyPair';
--- a/modules/sdk-coin-btc/test/unit/fixture/keys.json
+++ b/modules/sdk-coin-btc/test/unit/fixture/keys.json
@@ -1,0 +1,120 @@
+{
+  "valid": [
+    {
+      "d": "0000000000000000000000000000000000000000000000000000000000000001",
+      "Q": "0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
+      "compressed": true,
+      "network": "bitcoin",
+      "address": "1BgGZ9tcN4rm9KBzDn7KprQz87SZ26SAMH",
+      "WIF": "KwDiBf89QgGbjEhKnhXJuH7LrciVrZi3qYjgd9M7rFU73sVHnoWn"
+    },
+    {
+      "d": "0000000000000000000000000000000000000000000000000000000000000001",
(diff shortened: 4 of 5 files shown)


modules/sdk-coin-btc/src/lib/keyPair.ts
```diff
@@ -0,0 +1,96 @@
...
+enum ExtendedAddressFormat {
+  mainnetBech32 = 'mainnetBech32',
+  testnetBech32 = 'testnetBech32',
+}
+
+export type AddressFormatType = BaseAddressFormat | ExtendedAddressFormat;
+export const AddressFormat = { ...BaseAddressFormat, ...ExtendedAddressFormat };
+
+/** Bitcoin key management. */
+export class KeyPair extends Secp256k1ExtendedKeyPair {
```
mmcshinsky-bitgo, line 27: Is this something that would be good to put in `abstract-utxo` and share among utxo coins?

argjv: Absolutely, however, due to the dependency between modules I opted for having the first implemented directly in a class I can use. I'm moving the common code to the base class in a follow-up PR so subsequent UTXO coins only have to set up the network.

There are a few changes coming in `sdk-core` and `abstract-utxo` classes.

Also, the enum `AddressFormat` seems to be mixing the format and network. I think we may want to have them separated into 2 different parameters like:

```
export enum AddressFormat {
  hex = 'hex',
  base58 = 'base58',
  bech32 = 'bech32',
  bech32 = 'bech32',
}

export enum Network {
  mainnet = 'mainnet',
  testnet = 'testnet',
}
```

Then we can do something like:
`getAddress({ format: AddressFormat, network: Network })`, or even better, we can initialize the KeyPair class with the right network like `new KeyPair({ network: Network })`. Thoughts?

modules/sdk-coin-btc/src/lib/keyPair.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Get a bitcoin address in legacy or bech32 format.
+   *
+   * @param {AddressFormatType} format One of mainnet, testnet, mainnetBech32, or testnetBech32
+   * @returns {string} The address derived from the public key
+   */
+  getAddress(format?: AddressFormatType): string {
+    const compressed = this.keyPair.compressed;
+    const publicKeyHash160 = crypto.ripemd160(crypto.sha256(this.getPublicKey({ compressed })));
```
tylerlevine, line 76: why only support pay to public key hash addresses? These addresses are not compatible with bitgo multisig wallets.

modules/sdk-coin-btc/src/lib/keyPair.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+   * Bitcoin uncompressed public and private keys in hex format.
+   *
+   * @returns { DefaultKeys } The keys in the protocol default key format
+   */
+  getKeys(): DefaultKeys {
+    return {
+      pub: this.getPublicKey({ compressed: false }).toString('hex'),
```
OttoAllmendinger, line 63: why do we want to support uncompressed keys?

OttoAllmendinger (commented):
I'm not sure where and how this is supposed to be used

modules/sdk-coin-btc/src/lib/keyPair.ts
sangaman: Looks like other coins that implement this method default to `compressed: false` but I also see some that accept `compressed` as a parameter to `getKeys`. We could do that here and have it default to `true`? That way we can "support" uncompressed keys but default compressed, which seems like the more sensible default.

modules/sdk-coin-btc/src/lib/keyPair.ts
```diff
@@ -0,0 +1,96 @@
...
+      const seed = randomBytes(DEFAULT_SEED_SIZE_BYTES);
+      this.hdNode = bip32.fromSeed(seed);
+    } else if (isSeed(source)) {
+      this.hdNode = bip32.fromSeed(source.seed);
+    } else if (isPrivateKey(source)) {
+      this.recordKeysFromPrivateKey(source.prv);
+    } else if (isPublicKey(source)) {
+      this.recordKeysFromPublicKey(source.pub);
+      // TODO: move this to Secp256k1ExtendedKeyPair
+      this.keyPair.compressed = source.pub.length === 66;
```
sangaman, line 46: I believe uncompressed bitcoin keys are 65 bytes, not 66.

modules/sdk-coin-btc/src/lib/keyPair.ts
```diff
@@ -0,0 +1,96 @@
...
+   * @param {AddressFormatType} format One of mainnet, testnet, mainnetBech32, or testnetBech32
+   * @returns {string} The address derived from the public key
+   */
+  getAddress(format?: AddressFormatType): string {
+    const compressed = this.keyPair.compressed;
+    const publicKeyHash160 = crypto.ripemd160(crypto.sha256(this.getPublicKey({ compressed })));
+
+    if (!format) {
+      // Return mainnet legacy addresses by default
+      return address.toBech32(publicKeyHash160, 0, networks.bitcoin.bech32);
```
sangaman, line 80: Either the comment or the code here is a mistake, the comment says legacy address but this is returning a bech32 address.

modules/sdk-coin-btc/src/lib/keyPair.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    switch (format) {
```
sangaman, line 83: Is there any value in supporting p2wpkh-p2sh here?

modules/sdk-coin-btc/src/lib/keyPair.ts
```diff
@@ -0,0 +1,96 @@
...
+      pub: this.getPublicKey({ compressed: false }).toString('hex'),
+      prv: this.getPrivateKey()?.toString('hex'),
+    };
+  }
+
+  /**
+   * Get a bitcoin address in legacy or bech32 format.
+   *
+   * @param {AddressFormatType} format One of mainnet, testnet, mainnetBech32, or testnetBech32
+   * @returns {string} The address derived from the public key
```
OttoAllmendinger, line 72: UTXO coins use a triple of keys plus a derivation path to generate address. (I am surprised that we even have an interface to generate addresses controlled by single key)

Deriving addresses from key triples is implemented here

* https://github.com/BitGo/BitGoJS/blob/master/modules/utxo-lib/src/bitgo/wallet/WalletKeys.ts
* https://github.com/BitGo/BitGoJS/blob/master/modules/utxo-lib/src/bitgo/wallet/WalletScripts.ts

OttoAllmendinger (changes requested):

tylerlevine (changes requested):
I think unfortunately I have to NACK this as it lends itself to incorrect assumptions about how bitcoin addresses work with bitgo wallets. It doesn't support multisig in any form, nor BIP32 derivation.

I don't fully understand the use case here but this strikes me as something that would be more appropriate to implement in a fork if this single-sig address construction is desired.

modules/sdk-coin-btc/src/lib/keyPair.ts
jorge-falconx: 66 is the number of characters when in hex format.
Changed it to be buffer and make this more clear.

argjv: Ah yeah, 66 is the character length when in `hex` format.
I changed it to be in bytes and make it more clear.

modules/sdk-coin-btc/src/lib/keyPair.ts
argjv: The `KeyPair` class is designed to handle a single set of keys. In order to manage multiple `KeyPair`s you would be better to add another abstraction layer like `MultiSigKeyChain` or `KeyStore` or something like that where you can add multisig-specific logic.

```
/** Manage a multisig account and their keys  */
export class MultiSigKeyChain {
  constructor(m: number, n: number) {
    ...
  }
  
  /**  Add a key pair with or without their private material. Private material is only recommended for testing  */
  addKeyPair(sigingKeyPair: KeyPair) { ... }

  /** Sign a transaction with all stored KeyPairs or a particular one if pub is passed as a parameter. */
  sign(tx: string, pub?: PublicKey) {}

  /** Get a multisig compatible address. */
  getAddress(format: AddressFormat) { .. }
}
```

This is what's been done for Algorand (in a non-generic way).

Adding the address format and throwing an error is probably a good idea here since a multisig address cannot be generated under the 1 key primitive. I leave the multisig class implementation to you guys unless you want me to give it a stub in a follow-up PR.

modules/sdk-coin-btc/src/lib/keyPair.ts
argjv: Changing default to compressed = true

argjv: Changing the default to be `this.keyPair.compressed`

modules/sdk-coin-btc/src/lib/keyPair.ts
argjv: Good catch

modules/sdk-coin-btc/src/lib/keyPair.ts
argjv: There is but I don't think that would overload the `KeyPair` class with the responsibility to handle multiple keys instead of just one. See my comment about `MultiSigKeyChain`

modules/sdk-coin-btc/src/lib/keyPair.ts
argjv: You actually already have that multisig management here: https://github.com/BitGo/BitGoJS/blob/549fbbcded358e0207da298e0278dfdb09a4d787/modules/utxo-lib/src/bitgo/wallet/WalletKeys.ts#L73

And that's where it belongs since it is BitGo specific, so scratch the `MultiSigKeyChain` suggestion. This class is designed to deal with a single `KeyPair`.

modules/sdk-coin-btc/src/lib/keyPair.ts
argjv: Do you still need the derivation path if this class is initialized with extended keys? Keep in mind it is using `Secp256k1ExtendedKeyPair` which by default tries to use bip32.
Key deriviation was not contemplated in `Secp256k1ExtendedKeyPair` either but I guess it would make sense to take the derivation path in the constructor parameter so that this class can be initialized with it.
A `derive(...)` method can also be added to the base interface but I guess it would only be useful for a few coins only.

tylerlevine: Yes, I think so. The motivating idea behind extended key pairs is key derivation. You might as well just generate a non-extended key pair and use that, since that's functionally equivalent to using an extended key pair that only supports the `m` derivation path.

argjv:
Thanks for the reviews!

Let me give this change some more context and potential use cases, but I need you to be open minded about it. Don't think bitcoin, multisig or BitGo. Think 3rd party devs first, after all, I'm one of them now.

The `KeyPair` interface in account-lib is extremly powerful because with only 2 lines of code I can generate random keys and a receive address for every account based coin you guys support:

```
const keyPair = register(coin);
{
  address: keyPair.getAddress(),
  keys: keyPair.getKeys(),
}
```

They are all single sig, and the security implications of generating keys with this JS lib makes them unsitable for production environments. However, this is good enough for test and even regtest environments. This can be the backend of a very generic single sig faucet, one that does not need full BitGo multisig wallet support before you can use it. It can also help with receive addresses for mining rewards in a regtest environment. Or it can be used as a formating tool so you can generate addresses on multiple chains from the same pub key.
The applications for this basic class are many once you drop the assumption that if it doesn't work for BitGo multisig wallets it should be discarded.

Sure, I can create my own fork and go on my marry way, but that doesn't help others.

These more generic abstractions make your libraries more developer friendly and can unlock the development of projects that we cannot even imagine right now.

As mentioned in the other comments, adding this will not make BitGo single sig any more that they ALGO.KeyPair does not make Algo wallets single sig. Besides, you already have a lot or great multisig wrok here: https://github.com/BitGo/BitGoJS/tree/549fbbcded358e0207da298e0278dfdb09a4d787/modules/utxo-lib/src/bitgo/wallet

Let me help you develop a 3rd party developer friendly set of libraries 💙

OttoAllmendinger (changes requested):
> Let me give this change some more context and potential use cases, but I need you to be open minded about it. Don't think bitcoin, multisig or BitGo. Think 3rd party devs first, after all, I'm one of them now.

> The applications for this basic class are many once you drop the assumption that if it doesn't work for BitGo multisig wallets it should be discarded.

A new low-level cross-coin generic "key management" library sounds like an interesting project, but I don't think the BitGoJS repository in general and the `sdk-coin-btc` module in particular is the correct place for an endeavor like this.  The code in this repository pertains to BitGo services.

bitgoAaron:
> The applications for this basic class are many once you drop the assumption that if it doesn't work for BitGo multisig wallets

@argjv I think what you've provided would still be quite useful, however `@bitgo/sdk-coin-btc` is probably the wrong module for it. Would you mind moving this code to a separate utility module? We'd be happy to include it in the overall BitGoJS ecosystem as long as it doesn't confuse our users' main use-cases here for using this library to interact with BitGo multi-sig & TSS wallets. I'd be fine with moving it to `@bitgo/utxo-lib` if this is specific to just utxo coins, but if you think it will grow beyond even those then a completely different module is probably more appropriate.

modules/sdk-coin-btc/src/lib/keyPair.ts
sangaman: Ah, I misread the line too thinking this said uncompressed.

PR #2689, opened 2022-08-24
feat(sdk-core): support lnurl pay

sj3014:
- decodeLnurlPay() decodes lnurl and fetch information about the lnurl
  from ln service
- fetchLnurlPayInvoice() requests for lightning invoice and verifies it
- a user will then use the return invoice to call payInvoice()

Ticket: BG-55259


modules/bitgo/test/v2/unit/lightning.ts
```diff
@@ -84,4 +84,33 @@ describe('lightning API requests', function () {
     assert.deepStrictEqual(res, fixtures.invoices);
     scope.done();
   });
+  
+  it('should decode lnurl-pay', async function () {
+    const lnurl = 'LNURL1DP68GURN8GHJ7UM9WFMXJCM99E3K7MF0V9CXJ0M385EKVCENXC6R2C35XVUKXEFCV5MKVV34X5EKZD3EV56NYD3HXQURZEPEXEJXXEPNXSCRVWFNV9NXZCN9XQ6XYEFHVGCXXCMYXYMNSERXFQ5FNS';
```
OttoAllmendinger, line 89: Please add the code that generates the LNURL somewhere and use the generated result here instead. This makes it easier to verify what's going on.

modules/bitgo/test/v2/unit/lightning.ts
```diff
@@ -84,4 +84,33 @@ describe('lightning API requests', function () {
...
+    const res = await wallet.lightning().decodeLnurlPay(lnurl);
+
+    assert.deepStrictEqual(res, { ...fixtures.decodedLnurl, maxSendable: 5679000 });
+    balanceScope.done();
+    scope.done();
+  });
+
+  it('should process lnurl invoice payment', async function () {
+    const callback = 'https://service.com/api?q=3fc3645b439ce8e7f2553a69e5267081d96dcd340693afabe04be7b0ccd178df';
+    const invoice = 'lntb20m1pvjluezsp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygshp58yjmdan79s6qqdhdzgynm4zwqd5d7xmw5fk98klysy043l2ahrqspp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqfpp3x9et2e20v6pu37c5d9vax37wxq72un989qrsgqdj545axuxtnfemtpwkc45hx9d2ft7x04mt8q7y6t0k2dge9e7h8kpy9p34ytyslj3yu569aalz2xdk8xkd7ltxqld94u8h2esmsmacgpghe9k8';
```
OttoAllmendinger, line 106: Similar situation here - the opaque blob here makes it difficult to modify this test. This should be a generated value instead.

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +119,33 @@ export class Lightning implements ILightning {
...
+      throw new Error('invalid lnurl');
+    }
+
+    const url = Buffer.from(bech32.fromWords(words)).toString();
+    const body = await this.bitgo.get(url).result();
+    const decodedRes = decodeOrElse(LnurlPayResponse.name, LnurlPayResponse, body, (errors) => {
+      throw new Error(`error(s) parsing lnurl response: ${errors}`);
+    });
+    const { availableBalance } = await this.getBalance();
+    const domain = new URL(url).hostname.replace('www.', '');
```
OttoAllmendinger, line 136: why?

modules/sdk-core/src/bitgo/lightning/validateLnurlInvoice.ts
```diff
@@ -0,0 +1,63 @@
...
+      pubKeyHash: 0x6f,
+      scriptHash: 0xc4,
+      validWitnessVersions: [0, 1],
+    };
+  }
+
+  return undefined;
+}
+
+export function validateLnurlInvoice(coin: string, invoiceStr: string, milliSatAmount: string, metadata: string): void {
```
OttoAllmendinger, line 30: Missing docstring

It is easy to forget using a validation method. Also it is often unclear when to use it - any time you receive this set of parameters somewhere?

Better add a `parseLnurlInvoice(...): ParsedInvoice` that does the validation and ensures that the resulting type is valid.

OttoAllmendinger (changes requested):

modules/sdk-core/package.json
```diff
@@ -39,10 +39,12 @@
     "@bitgo/utxo-lib": "^2.4.1",
     "@noble/secp256k1": "1.6.0",
     "@stablelib/hex": "^1.0.0",
+    "bech32": "^2.0.0",
     "big.js": "^3.1.3",
     "bignumber.js": "^9.0.0",
     "bip32": "^3.0.1",
     "bitcoinjs-message": "^2.0.0",
+    "bolt11": "^1.4.0",
```
sangaman, line 47: Can these dependencies be moved to `sdk-coin-btc` without causing any problems?

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -30,6 +30,12 @@ export interface GetInvoicesQuery {
   limit?: number;
 }
 
+export interface PayLnurlParams {
```
sangaman, line 33: `LnurlPayParams` would go better w/ the naming elsewhere.

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -149,6 +155,18 @@ export const GetInvoicesResponse = t.array(InvoiceInfo);
 // eslint-disable-next-line no-redeclare
 export type GetInvoicesResponse = t.TypeOf<typeof GetInvoicesResponse>;
 
+export const LnurlPayResponse = t.strict({
+  tag: t.literal('payRequest'),
+  callback: t.string,
+  maxSendable: t.number,
+  minSendable: t.number,
+  metadata: t.string,
+  domain: t.union([t.string, t.undefined]),
```
sangaman, line 164: I don't think `domain` is one of the elements on the first response? Just going off of what's described here https://github.com/fiatjaf/lnurl-rfc/blob/luds/06.md#pay-to-static-qrnfclink. You should just parse it from the URL.

modules/sdk-core/src/bitgo/lightning/lightning.ts
sangaman: From https://github.com/fiatjaf/lnurl-rfc/blob/luds/06.md#pay-to-static-qrnfclink

> Additionally, a payment dialog must include:
>
> - Domain name extracted from LNURL query string.

This would be good to reference in a comment though.

sj3014: lnurl protocol specifies that we should provide domain name extracted from lnurl query string

modules/sdk-core/src/bitgo/lightning/iLightning.ts
sj3014: lnurl spec 4. specifies that we should "display" max can send, min can send, metadata, and domain. I thought it would be reasonable to include them in the decode part

modules/sdk-core/package.json
sj3014: wouldn't it cause a circular dependency since `sdk-coin-btc` has `@bitgo/sdk-core` as a dependency and `validateLnurlInvoice.ts` lives in `@bitgo/sdk-core` ? I am not sure how to move it to `sdk-coin-btc`

modules/sdk-core/src/bitgo/lightning/validateLnurlInvoice.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (millisatoshis === null || millisatoshis === undefined) {
+    throw new Error('invoice does not have an amount');
+  }
+
+  if (millisatoshis !== milliSatAmount) {
+    throw new Error('amount of invoice does not match with given amount');
+  }
+
+  const hash = crypto.createHash('sha256').update(new TextEncoder().encode(metadata)).digest('hex');
+  const found = tags.find((tag) => tag.data === hash);
```
sangaman, line 59: Rather than finding by `data`, I think it's better to find by `tagName === 'purpose_commit_hash'` and then check its value. So we don't accidentally pass it if the right value is in the wrong tag.

```suggestion
  const descriptionHash = tags.find((tag) => tag.tagName === 'purpose_commit_hash');
  if (descriptionHash !== hash)
```

modules/sdk-core/src/bitgo/lightning/iLightning.ts
sangaman: OK, I just misinterpreted how the type was being used. That's fine then. Why would it be undefined though? We should always have a domain because we'll always have a url.

sangaman (changes requested):

modules/sdk-core/src/bitgo/lightning/validateLnurlInvoice.ts
sj3014: for `domain`, type is `t.union([t.string, t.undefined)` because it `LnurlPayResponse` is first used to decode the response from ln service, which does not contain `domain` field, and later we manually add `domain` to the final return value. I have done this way so that I do not have to create another type. Or actually we can omit `domain` in `LnUrlPayResponse` and make return type be like `LnurlPayResponse & { domain: string }`

modules/sdk-core/src/bitgo/lightning/validateLnurlInvoice.ts
```diff
@@ -0,0 +1,63 @@
...
+  const { millisatoshis, tags } = decodedInvoice;
+  if (millisatoshis === null || millisatoshis === undefined) {
+    throw new Error('invoice does not have an amount');
+  }
+
+  if (millisatoshis !== milliSatAmount) {
+    throw new Error('amount of invoice does not match with given amount');
+  }
+
+  const hash = crypto.createHash('sha256').update(new TextEncoder().encode(metadata)).digest('hex');
```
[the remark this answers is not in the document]
sj3014: the lnurl spec specifies that `metadata` should be converted to byte array in UTF-8 encoding, but I just ran a test without `new TextEncoder()` part and it worked fine.

modules/bitgo/test/v2/unit/lightning.ts
sj3014: the invoice is from https://github.com/lightning/bolts/blob/master/11-payment-encoding.md bolt11 spec. Do we still want to generate an invoice for ourselves?

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +119,33 @@ export class Lightning implements ILightning {
...
+
+  public async payLnurl(params: PayLnurlParams): Promise<PayInvoiceResponse> {
+    const { callback, milliSatAmount, metadata } = params;
+    const { pr: invoice } = callback.includes('?')
+      ? await this.bitgo.get(callback + `&amount=${milliSatAmount}`).result()
+      : await this.bitgo.get(callback).query({ amount: milliSatAmount }).result();
+
+    validateLnurlInvoice(this.wallet.coin(), invoice, milliSatAmount, metadata);
+    return await this.payInvoice({ invoice });
+  }
```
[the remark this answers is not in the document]
sj3014: oh yeah thank you for the remind!

modules/bitgo/test/v2/unit/lightning.ts
OttoAllmendinger: please put it in a JSON file 

```
{
  "comment": "taken from https://...",
  "data": "lnbt20..."
}
```

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +119,33 @@ export class Lightning implements ILightning {
       throw new Error(`error(s) parsing response body: ${errors}`);
     });
   }
+
+  public async decodeLnurlPay(lnurl: string): Promise<LnurlPayResponse> {
+    const { prefix, words } = bech32.decode(lnurl, 2000);
```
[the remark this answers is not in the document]
sj3014: yes that is just a upper limit

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +118,32 @@ export class Lightning implements ILightning {
       throw new Error(`error(s) parsing response body: ${errors}`);
     });
   }
+
+  public async decodeLnurlPay(lnurl: string): Promise<LnurlPayResponse & { domain: string }> {
```
sangaman, line 122: I like the way you've done it with `LnurlPayResponse & { domain: string }` per our earlier conversation about `domain`. I might go one step further and give this type a name, maybe `type DecodedLnurlPayRequest = LnurlPayResponse & { domain: string }` which is used here. I think that's a little bit clearer.

Edit: Even better, I think we should differentiate the units on the LNURL response per the spec (millisats) from the units we're working with (sats). So ideally the decoded type should look like this:

```typescript
type DecodedLnurlPayRequest = Omit<LnurlPayResponse, 'maxSendable' | 'minSendable'> & { 
  domain: string,
  /** The maximum amount in satoshis we can pay for this LNRUL request /**
  maxSendableSats: number,
  /** The minimum amount in satoshis we can pay for this LNRUL request /**
  minSendableSats: number,
}
```

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +118,32 @@ export class Lightning implements ILightning {
...
+    const url = decodeLnurl(lnurl);
+    const body = await this.bitgo.get(url).result();
```
sangaman, line 124: I don't think we should use the `bitgo.get` here because it'll add headers and authorization that's specific to interaction with the BitGo API, whereas this is a non-BitGo server. Can we `import request from superagent` and do `request.get` here?

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +118,32 @@ export class Lightning implements ILightning {
...
+    const decodedRes = decodeOrElse(LnurlPayResponse.name, LnurlPayResponse, body, (errors) => {
+      throw new Error(`error(s) parsing lnurl response: ${errors}`);
+    });
+    const { availableBalance } = await this.getBalance();
```
sangaman, line 128: I don't think we want to look up our available balance here. That's hitting the bitgo API which again isn't ideal for internal use. And regardless of that, our balance is subject to change and doesn't need to be a limiting factor here. Let's keep this client side and just use the `maxSpendable`.

modules/sdk-core/src/bitgo/lightning/iLightning.ts
```diff
@@ -149,6 +155,17 @@ export const GetInvoicesResponse = t.array(InvoiceInfo);
 // eslint-disable-next-line no-redeclare
 export type GetInvoicesResponse = t.TypeOf<typeof GetInvoicesResponse>;
 
+export const LnurlPayResponse = t.strict({
+  tag: t.literal('payRequest'),
+  callback: t.string,
+  maxSendable: t.number,
+  minSendable: t.number,
```
sangaman, line 162:
```suggestion
  /** The maximum amount in millisatoshis we can pay for this LNRUL request */
  maxSendable: t.number,
  /** The minimum amount in millisatoshis we can pay for this LNRUL request */
  minSendable: t.number,
```

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +118,32 @@ export class Lightning implements ILightning {
...
+    const domain = new URL(url).hostname;
+
+    decodedRes.maxSendable = Math.min(decodedRes.maxSendable / 1000, availableBalance);
```
sangaman, line 131: We should round this down, if we something like 3001 millisats we don't want to have `3.001` but rather `3` in the interest of continuing to work with whole satoshi units. So make this a `Math.floor` (and get rid of `availableBalance` per my other comment).

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +118,32 @@ export class Lightning implements ILightning {
...
+    decodedRes.minSendable = decodedRes.minSendable / 1000;
```
sangaman, line 132: Round this up for the same reasons as above, so `Math.ceil`

Note: This would mean we couldn't pay LNURL requests that request an exact millisatoshi amount. We'd either need to bite the bullet and start working with milllisats, or just accept that we won't be able to pay those (in practice I figure those will be super rare). Let's discuss a bit more tho.

sangaman (changes requested):
Getting close, I still have a handful of suggestions/requests and one takeaway is that we may need to reconsider how we're dealing with millisatoshis here.

modules/sdk-core/src/bitgo/lightning/lightning.ts
sj3014: so regardless of user' balance, we should let the user be able to make lnurl pay request, and later in wp, if the user has no sufficient balance, it then throws error?

modules/bitgo/test/v2/fixtures/lightning/invoices.ts
```diff
@@ -0,0 +1,6 @@
+export default {
+  lnurlPayInvoice: {
+    comment: 'taken from https://github.com/lightning/bolts/blob/master/11-payment-encoding.md',
```
OttoAllmendinger, line 3: typescript file is OK as well, you can use `//` for comment now

sj3014: I tried to use json file, but it gave me `include` error saying all files should be included. I checked `tsconfig.json` , and it seemed everything is included.

modules/sdk-core/src/bitgo/lightning/validateLnurlInvoice.ts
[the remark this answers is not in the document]
sangaman: Yeah I figure utf-8 is just the default. I'd say let's remove this just to keep it simple.

sangaman (changes requested):
I went to do some manual tests with this branch, and one other thing I realized is that currently all these functions are tied to a wallet - like we need to authenticate and fetch a wallet before we can access these methods. That won't work for internal usage where we're not interacting with the BitGo API.

I think it's fine to have these methods there too, but maybe this needs to be in an `lnurlUtils.ts` file or something. Then the existing wallet lightning methods can call those utils.

modules/sdk-core/src/bitgo/lightning/lightning.ts
sangaman: FYI, I tried this branch before you made the change to superagent and this request was failing with `ApiResponseError: invalid response HMAC, possible man-in-the-middle-attack` because it wasn't a request to a BitGo server. With the latest I can get past that error.

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +120,33 @@ export class Lightning implements ILightning {
...
+      ...decodedRes,
+      domain,
+    };
+  }
+
+  public async fetchLnurlPayInvoice(params: LnurlPayParams): Promise<string> {
+    const { callback, milliSatAmount, metadata } = params;
+    const { pr: invoice } = callback.includes('?')
+      ? await this.bitgo.get(callback + `&amount=${milliSatAmount}`).result()
+      : await this.bitgo.get(callback).query({ amount: milliSatAmount }).result();
```
sangaman, line 146: These need to use superagent get instead of bitgo get, currently giving me that `ApiResponseError: invalid response HMAC, possible man-in-the-middle-attack` error.

sangaman (changes requested):

modules/sdk-core/src/bitgo/lightning/parseAndValidateLnurlInvoice.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+
+/**
+ * @param {string} coin - a type of coin for lightning
+ * @param {string} invoiceStr - a lightning invoice
+ * @return {ParsedLightningInvoice}
+ * @throws error for invalid lightning invoice
+ */
+export function parseLightningInvoice(coin: string, invoiceStr: string): ParsedLightningInvoice {
```
sangaman, line 56: The lnurl server might give us an undefined `invoiceStr` or even something unexpected like a number or object. We should probably have `invoiceStr` be `unknown` here and do basic checks to make sure it's a string.

sangaman (changes requested):
Ok I think this is my last round of feedback.

Besides these comments, I also think it'd make sense to combine `lnurlUtils.ts` and `parseAndValidateLnurlInvoice.ts` (which is a bit too long) into one file and maybe name it `lightningUtils.ts` since the `parseLightningInvoice` function isn't specific to lnurl.

Lastly I think it'd be good to have some more tests for failure cases and such (like mismatched description hash, wrong amount, etc) however I think that can be handled in a future PR. I've tested that this branch works against my test lnurl server.

Thanks!

modules/sdk-core/src/bitgo/lightning/parseAndValidateLnurlInvoice.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (coin === 'tbtc') {
+    // tb = testnet invoice, crt = regtest invoice
+    if (!invoiceStr.startsWith('lntb') && !invoiceStr.startsWith('lncrt') && !invoiceStr.startsWith('lntbs')) {
+      throw new Error('invoice is not for tbtc');
+    }
+  } else {
+    if (!invoiceStr.startsWith('lnbc')) {
+      throw new Error('invoice is not for mainnet btc');
+    }
+  }
```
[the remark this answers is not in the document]
sj3014: actually, I don't think we need this `coin` checking logic at all as this logic is done when a user creates a new lightning object

modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -114,4 +117,12 @@ export class Lightning implements ILightning {
       throw new Error(`error(s) parsing response body: ${errors}`);
     });
   }
+
+  public async decodeLnurlPay(lnurl: string): Promise<DecodedLnurlPayRequest> {
+    return await decodeLnurlPay(lnurl);
+  }
+
+  public async fetchLnurlPayInvoice(params: LnurlPayParams): Promise<string> {
+    return await fetchLnurlPayInvoice(params, this.wallet.coin());
```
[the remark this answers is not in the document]
sj3014: i believe we need those keywords as `decodeLnurlPay` and `fetchLnurlPayInvoice` utils return `Promise`

modules/sdk-core/src/bitgo/lightning/parseAndValidateLnurlInvoice.ts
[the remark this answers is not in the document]
sj3014: we can check the invoice prefix in a separate function, but wouldn't it still be called within the same `fetchLnurlPayInvoice` function?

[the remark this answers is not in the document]
sj3014: i actually removed this logic because it will be checked when we try to pay the invoice.

modules/sdk-core/src/bitgo/lightning/lightning.ts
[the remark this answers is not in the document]
sangaman: Yes but they can return promises without being `async`, and the promises don't need to be `await`ed here. Functionally it doesn't make much difference, I just generally try not to declare functions as `async` unless they have to be.

modules/sdk-core/src/bitgo/lightning/parseAndValidateLnurlInvoice.ts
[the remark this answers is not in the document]
sangaman: OK, I think removing it is probably fine.

sangaman (approved):
Good stuff, thanks for going through all these comments.  My only remaining nit is those couple of unnecessary `async`/`await` keywords but that is basically stylistic.

OttoAllmendinger (approved):

modules/sdk-core/src/bitgo/lightning/lightning.ts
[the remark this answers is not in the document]
byhow: I agree, we don't need to `await` it at the end. Since we are returning a promise, after `await` we need to wrap it again in a `Promise` compared to just returning the promise as is, but its really a nit

byhow (approved):
Very thorough review! Good spirit

modules/sdk-core/src/bitgo/lightning/lightningUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+  milliSatoshis: string;
+  /** hex-encoded */
+  paymentHash: string;
+  /** hex-encoded */
+  pubkey: string;
+  /** hex-encoded */
+  descriptionHash?: string;
+};
+
+export async function decodeLnurlPay(lnurl: string): Promise<DecodedLnurlPayRequest> {
```
mmcshinsky-bitgo, line 18: Already merged but these exported functions need docs.

PR #2692, opened 2022-08-24
feat(bsc): support tokens for bsc

kxl4126:
Ticket: BG-55766


modules/sdk-coin-bsc/src/bscToken.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+import { Bsc } from './bsc';
+import { TransactionPrebuild } from '@bitgo/sdk-coin-eth';
+import { EthLikeTokenConfig, tokens, coins } from '@bitgo/statics';
+import { BitGoBase, CoinConstructor, NamedCoinConstructor } from '@bitgo/sdk-core';
+
+export { EthLikeTokenConfig };
+
+export class BscToken extends Bsc {
```
wesgraham, line 12: this lgtm - something to think about to make this onboarding even faster is whether we need to duplicate token code here and extend the coin --> or whether we could have an `abstractEthLikeToken` that we extend directly or a config/array we append with bsc coin name or something to give us the same support

wesgraham (dismissed):
excellent!

wesgraham (approved):

byhow (approved):

PR #2693, opened 2022-08-24
feat(sdk-coin-sol): implement transferBuilderV2

linyah9912:
Ticket: BG-55476

Currently the SOL transfer builder only supports transferring either native asset or tokens. To help building SOL consolidation transactions, needs to support specifying a `feePayer`, and transferring both native asset and tokens.

This PR implements the new TransferBuilderV2 which set a `feePayer` and build a single transaction that contains multiple instructions.


joey-bai (approved):
Great work!

byhow (approved):

PR #2675, opened 2022-08-19
feat: adding support for message signing

jeanlarrieux:
BG-54927

PR #2697, opened 2022-08-26
Readme yarn version

byhow:
There isn't a good way to enforce package manager version to be
yarn v1. There are two possible ways to specify the version: one
is through the `engine` option, and another is the `packageManager`
option in `package.json`.

Now with the `engine` configuration, it only notify the user of the
SDK that it is compatile with `yarn` v1 installs, but for developers
of the SDK, it won't enfore the package manager version.

With `packageManager`, you can specify like this:
```
...
  "packageManager": "^yarn@1.22.19"
...
```
But if you would like to use yarn v2 on this repo with:
```
yarn set version berry
```
It will override this config and the install will not be blocked either,
let alone that the option only works after `v16.9.x` without corepack,
which is not a dependency of the SDK.

Ticket: BG-00000

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ supported versions described in the [`engines` property of the `bitgo` module pa
 
 We specifically limit our support to LTS versions of Node, not because this package won't work on other versions, but
 because LTS versions tend to be the most widely used in practice. It's possible the packages in this repository will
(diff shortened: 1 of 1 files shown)


byhow:
relates to #2675

byhow:
closing this as #2675 was closed

PR #2701, opened 2022-08-27
refactor(account-lib): modify random gen for nu0 and beta0

sachushaji:
this change expands the random number
range for nuo and beta0 in ecdsa tss
to be within [1:n] where n is maxint
of paillier keys

TICKET: BG-55963

### DIFF
--- a/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
@@ -1,4 +1,5 @@
 import * as paillierBigint from 'paillier-bigint';
+import * as bigintCryptoUtils from 'bigint-crypto-utils';
 import * as secp from '@noble/secp256k1';
 import { randomBytes, createHash } from 'crypto';
 import { hexToBigInt } from '../../../util/crypto';
@@ -28,6 +29,9 @@ import {
   XShare,
   YShare,
 } from './types';
+
+const _1n = BigInt(1);
+const _3n = BigInt(3);
 
 /**
  * ECDSA TSS implementation supporting 2:n Threshold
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -225,15 +229,18 @@ export default class Ecdsa {
...
+      shareParticipant.beta = bigIntToBufferBE(
+        Ecdsa.curve.scalarReduce(Ecdsa.curve.scalarNegate(Ecdsa.curve.scalarReduce(beta0))),
+        32
+      ).toString('hex');
       const alpha = pk.addition(pk.multiply(k, hexToBigInt(shareParticipant.gamma)), pk.encrypt(beta0));
       shareToBeSend.alpha = bigIntToBufferBE(alpha, 32).toString('hex');
 
-      const nu0 = Ecdsa.curve.scalarRandom();
-      shareParticipant.nu = bigIntToBufferBE(Ecdsa.curve.scalarNegate(nu0), 32).toString('hex');
+      const nu0 = bigintCryptoUtils.randBetween(n / _3n - _1n);
```
sachushaji, line 240: Existing init tests is already testing the validity of the signing process, no need for additional test. Max Int based on value [here](https://github.com/data61/python-paillier/blob/7d9911eb03c3c2d64399bc15405feb5e628379d1/phe/paillier.py#L90)

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -225,15 +229,18 @@ export default class Ecdsa {
       let pk = getPaillierPublicKey(n);
       const k = hexToBigInt(shareToBeSend['k']);
 
-      const beta0 = Ecdsa.curve.scalarRandom();
-      shareParticipant.beta = bigIntToBufferBE(Ecdsa.curve.scalarReduce(Ecdsa.curve.scalarNegate(beta0)), 32).toString(
-        'hex'
-      );
+      const beta0 = bigintCryptoUtils.randBetween(n / _3n - _1n);
+      shareParticipant.beta = bigIntToBufferBE(
+        Ecdsa.curve.scalarReduce(Ecdsa.curve.scalarNegate(Ecdsa.curve.scalarReduce(beta0))),
```
johnoliverdriscoll, line 234: The extra reduce is not necessary, it's enough to just use `Ecdsa.curve.scalarNegate(Ecdsa.curve.scalarReduce(beta0))`

johnoliverdriscoll (changes requested):

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
sachushaji: Nice Catch !!, Updated

wesgraham (approved):

johnoliverdriscoll (approved):

mmcshinsky-bitgo (approved):

PR #2702, opened 2022-08-29
fix(account-lib): added validate address function to sdk-coin-ada

margueriteblair:
Ticket: BG-56066

### DIFF
--- a/modules/sdk-coin-ada/src/ada.ts
+++ b/modules/sdk-coin-ada/src/ada.ts
@@ -5,6 +5,7 @@ import {
   BaseCoin,
   BaseTransaction,
   BitGoBase,
+  InvalidAddressError,
   KeyPair,
   MPCAlgorithm,
   ParsedTransaction,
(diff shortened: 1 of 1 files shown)


ewangbitgo (approved):

byhow (approved):

modules/sdk-coin-ada/src/ada.ts
```diff
@@ -105,7 +106,11 @@ export class Ada extends BaseCoin {
   }
 
   isWalletAddress(params: VerifyAddressOptions): boolean {
-    throw new Error('Method not implemented.');
+    const { address } = params;
+    if (!this.isValidAddress(address)) {
+      throw new InvalidAddressError(`Invalid Cardano Address: ${address}`);
+    }
+    return true;
```
Phani024, line 113: Shouldn't we also be checking if this address belongs to a wallet?

margueriteblair: Good question! @ewangbitgo can you confirm this? When I looked at other coins, this method was either not implemented, or just checked if it was a valid address.

PR #2704, opened 2022-08-29
build: update lerna to latest and add schema

byhow:
Lerna schema was introduced post 5.2: https://github.com/lerna/lerna/releases/tag/v5.2.0,
and some refactor work was being done on `git` commands.

To reproduce the change in `lerna.json` and `package.json`, run:
```
> npx lerna@latest init
```
Reference here: https://lerna.js.org/docs/getting-started#adding-lerna

Ticket: BG-56327

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -6,5 +6,7 @@
     "version": {
       "message": "chore(release): publish modules"
     }
-  }
+  },
+  "$schema": "node_modules/lerna/schemas/lerna-schema.json",
+  "useNx": false
 }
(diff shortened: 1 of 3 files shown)


byhow:
Tested it on https://github.com/BitGo/BitGoJS/compare/rel/rc...rel/rc-updated-lerna and it seems to fix my concurrent `git tag -S` with lerna

mmcshinsky-bitgo (approved):

PR #2565, opened 2022-07-22
chore(statics): update Sol coin name to Solana BG-52979

michael-gallegos-bg:
Update "Sol" and "Testnet Sol" in statics to be the full names "Solana" and "Testnet Solana"

Ticket: https://bitgoinc.atlassian.net/browse/BG-52979

PR #2712, opened 2022-08-30
fix(statics): update ADA fullname

ewangbitgo:
Ticket: BG-56343


byhow (changes requested):
Name change needs to be breaking change: see #2565

byhow (approved):

PR #2714, opened 2022-08-31
fix(sdk-core): fix the signatures of lnurl pay methods

sj3014:
Ticket: BG-55259


sangaman (approved):
Ah, I thought the `coin` was removed from `fetchLnurlPayInvoice` in one of the rounds of reviews from the previous PR, but I guess only the logic that used it was removed. I'm actually surprised that the build didn't fail due to an unused variable, since `coin` wasn't actually being used anywhere?

mmcshinsky-bitgo (approved):

PR #2728, opened 2022-09-02
fix(sdk-coin-eos): Fix precision for EOS:CHEX tokens

angshal:
Currently, when withdrawing CHEX tokens on EOS mainnet, the amount is shown with 4 decimals of precision on the FE.
However, the precision should be 8 decimals for CHEX specifically. This fix includes the modification of a decimalPlaces getter in eos. When the token is CHEX, this getter will retrieve the decimal place from statics.

Ticket: BG-47340

### DIFF
--- a/modules/sdk-coin-eos/src/eos.ts
+++ b/modules/sdk-coin-eos/src/eos.ts
@@ -226,6 +226,10 @@ export class Eos extends BaseCoin {
     return 1e4;
   }
 
+  get decimalPlaces() {
+    return 4;
+  }
+
   /**
    * Flag for sending value of 0
    * @returns {boolean} True if okay to send 0 value, false otherwise
@@ -364,7 +368,7 @@ export class Eos extends BaseCoin {
     const dividend = this.getBaseFactor();
     const bigNumber = new BigNumber(baseUnits).dividedBy(dividend);
     // set the format so commas aren't added to large coin amounts
-    return bigNumber.toFormat(4, null as any, { groupSeparator: '', decimalSeparator: '.' });
+    return bigNumber.toFormat(this.decimalPlaces, null as any, { groupSeparator: '', decimalSeparator: '.' });
   }
 
   /**
--- a/modules/bitgo/test/v2/unit/baseCoin.ts
+++ b/modules/bitgo/test/v2/unit/baseCoin.ts
@@ -18,9 +18,12 @@ describe('V2 Base Coin:', function () {
   let basecoinBtc;
   let basecoinXlm;
   let basecoinNear;
+  let basecoinEos;
+  let basecoinEosChex;
   let basecoinErc20TokenWithName;
   let basecoinErc20TokenWithContractHash;
   let baseCoinStellarToken;
+
 
   before(function () {
     bitgo = TestBitGo.decorate(BitGo, { env: 'test' });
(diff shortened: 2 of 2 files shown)


modules/sdk-coin-eos/src/eos.ts
```diff
@@ -360,11 +360,11 @@ export class Eos extends BaseCoin {
...
    * to big units (btc, eth, xrp, xlm)
    */
-  baseUnitsToBigUnits(baseUnits: string | number): string {
-    const dividend = this.getBaseFactor();
+  baseUnitsToBigUnits(baseUnits: string | number, chex?: boolean): string {
+    const dividend = chex ? 1e8 : this.getBaseFactor();
     const bigNumber = new BigNumber(baseUnits).dividedBy(dividend);
     // set the format so commas aren't added to large coin amounts
-    return bigNumber.toFormat(4, null as any, { groupSeparator: '', decimalSeparator: '.' });
+    return bigNumber.toFormat(chex ? 8 : 4, null as any, { groupSeparator: '', decimalSeparator: '.' });
```
joey-bai, line 367: we should only need to change the hard coded `4` to `this.decimalPlaces`.

we define decimal places in statics: https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/coins.ts#L1750-L1757

modules/sdk-coin-eos/src/eos.ts
```diff
@@ -360,11 +360,11 @@ export class Eos extends BaseCoin {
    * Convert a currency amount represented in base units (satoshi, wei, atoms, drops, stroops)
    * to big units (btc, eth, xrp, xlm)
    */
-  baseUnitsToBigUnits(baseUnits: string | number): string {
-    const dividend = this.getBaseFactor();
+  baseUnitsToBigUnits(baseUnits: string | number, chex?: boolean): string {
+    const dividend = chex ? 1e8 : this.getBaseFactor();
```
joey-bai, line 364: `this.getBaseFactor()` should be left as is. we have an `EosToken` subclass that will populate this with the correct value.

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
```diff
@@ -347,7 +347,7 @@ export interface IBaseCoin {
   supportsStaking(): boolean;
   supportsBlsDkg(): boolean;
   getBaseFactor(): number | string;
-  baseUnitsToBigUnits(baseUnits: string | number): string;
+  baseUnitsToBigUnits(baseUnits: string | number, chex?: boolean): string;
```
joey-bai, line 350: please revert, we should not be adding something coin specific. there are other ways to determine the asset if it comes down to it (like checking `this.name` in the coin class)

byhow:
What resolves us to this implementation? The reason I am asking is that I had the same implementation a couple of months ago but it wasn't recommended. see: https://github.com/BitGo/BitGoJS/pull/2290

mmcshinsky-bitgo (approved):

mmcshinsky-bitgo:
> What resolves us to this implementation? The reason I am asking is that I had the same implementation a couple of months ago but it wasn't recommended. see: #2290

It looks like EosToken is overriding everything properly so we should be good to go.

PR #2724, opened 2022-09-01
feat: update repository templates

mmcshinsky-bitgo:
* Update: Additional steps for bug and feature requests
* Add: Pull request template to set pull request expectations

BG-56523

PR #2730, opened 2022-09-04
feat(sdk-core): add recid to fully constructed signature

johnoliverdriscoll:
Issue: BG-56664

### DIFF
--- a/modules/bitgo/test/v2/fixtures/tss/ecdsaFixtures.ts
+++ b/modules/bitgo/test/v2/fixtures/tss/ecdsaFixtures.ts
@@ -319,14 +319,14 @@ export const mockDShareToBitgo = {
 export const mockSShare = {
   i: 3,
   y: '0362c0f94da66b5cb2880d1bd8c0c407cfa417be23cfc29d3d02aec83a70f6d8ab',
-  r: '181273015b236d9b4de9811a76167db16e20bd34b9bde369ccd352e7bffbbc7f',
+  R: '02181273015b236d9b4de9811a76167db16e20bd34b9bde369ccd352e7bffbbc7f',
   s: '549291233abb0a8f8cf6948b956a596c90b16f9ae15554d77e7a8f69e04ac311',
 };
 
-export const mockSignature = {
+export const mockSShareFromUser = {
   i: 1,
   y: '0362c0f94da66b5cb2880d1bd8c0c407cfa417be23cfc29d3d02aec83a70f6d8ab',
-  r: '181273015b236d9b4de9811a76167db16e20bd34b9bde369ccd352e7bffbbc7f',
+  R: '02181273015b236d9b4de9811a76167db16e20bd34b9bde369ccd352e7bffbbc7f',
   s: '4ffc4eb69700daa897662db238e06a856c697807ab22a3824ad472eb59e7d86f',
 };
 
--- a/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
@@ -22,7 +22,6 @@ import {
   SignCombineRT,
   DShare,
   OShare,
-  SignRT,
   SShare,
   SignShareRT,
   KShare,
@@ -314,9 +313,9 @@ export default class Ecdsa {
    * @param {OShare} oShare private omicron share of current participant
    * @param {DShare} dShare delta share received from the other participant
    * @param {Hash} hash hashing algorithm implementing Node`s standard crypto hash interface
-   * @returns {SignRT}
+   * @returns {SShare}
    */
-  sign(M: Buffer, oShare: OShare, dShare: DShare, hash?: Hash): SignRT {
+  sign(M: Buffer, oShare: OShare, dShare: DShare, hash?: Hash): SShare {
     const m = (hash || createHash('sha256')).update(M).digest();
 
     const delta = Ecdsa.curve.scalarAdd(hexToBigInt(oShare.delta), hexToBigInt(dShare.delta));
@@ -335,7 +334,7 @@ export default class Ecdsa {
     return {
       i: oShare.i,
       y: oShare.y,
-      r: bigIntToBufferBE(r, 32).toString('hex'),
+      R: pointR.toHex(true),
       s: bigIntToBufferBE(s, 32).toString('hex'),
     };
   }
@@ -346,21 +345,23 @@ export default class Ecdsa {
    * @returns {Signature}
    */
   constructSignature(shares: SShare[]): Signature {
-    // Every r must match.
-    const r = shares[0]['r'];
-    const isRMatching = shares.map((share) => share['r'] === r).reduce((a, b) => a && b);
+    // Every R must match.
+    const R = shares[0]['R'];
+    const isRMatching = shares.map((share) => share['R'] === R).reduce((a, b) => a && b);
     if (!isRMatching) {
-      throw new Error('r value should be consistent across all shares');
+      throw new Error('R value should be consistent across all shares');
     }
 
     let s = shares.map((share) => hexToBigInt(share['s'])).reduce(Ecdsa.curve.scalarAdd);
+    const recid = (R.slice(0, 2) === '03' ? 1 : 0) ^ (s > Ecdsa.curve.order() / BigInt(2) ? 1 : 0);
 
     // Normalize s.
     s = s > Ecdsa.curve.order() / BigInt(2) ? Ecdsa.curve.order() - s : s;
     return {
       y: shares[0]['y'],
-      r: r,
+      r: R.slice(2),
       s: bigIntToBufferBE(s, 32).toString('hex'),
+      recid: recid,
     };
   }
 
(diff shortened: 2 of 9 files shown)


modules/sdk-core/src/account-lib/mpc/tss/ecdsa/types.ts
```diff
@@ -164,13 +164,14 @@ export interface SignCombineRT {
 
 export interface Signature {
   y: string;
+  recid: number;
   r: string;
   s: string;
 }
 
 export interface SignRT {
```
sachushaji, line 172: NIT: We can remove this SIgnRT from everywhere as it is equivalent to SShare.

sachushaji: or may be for backward compatibility
```suggestion
export interface SignRT = SShare;
```

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -428,8 +428,8 @@ export function parseMuShare(share: SignatureShareRecord): MUShare {
  */
 export function convertMuShare(share: MUShare): SignatureShareRecord {
   return {
-    to: getParticapantFromIndex(share.i),
-    from: getParticapantFromIndex(share.j),
+    to: getParticipantFromIndex(share.i),
```
sachushaji, line 431: Appreciate fixing these spelling mistakes

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -335,7 +335,7 @@ export default class Ecdsa {
     return {
       i: oShare.i,
       y: oShare.y,
-      r: bigIntToBufferBE(r, 32).toString('hex'),
+      R: pointR.toHex(true),
```
sachushaji, line 338: This might be a breaking change @olibeyene , you may need to address this in WP

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -346,21 +346,23 @@ export default class Ecdsa {
...
+    // Every R must match.
+    const R = shares[0]['R'];
+    const isRMatching = shares.map((share) => share['R'] === R).reduce((a, b) => a && b);
     if (!isRMatching) {
-      throw new Error('r value should be consistent across all shares');
+      throw new Error('R value should be consistent across all shares');
     }
 
     let s = shares.map((share) => hexToBigInt(share['s'])).reduce(Ecdsa.curve.scalarAdd);
+    const recid = (R.slice(0, 2) === '03' ? 1 : 0) ^ (s > Ecdsa.curve.order() / BigInt(2) ? 1 : 0);
```
sachushaji, line 357: Can we add a small unit test for this new addition `recid` value may be by using the SShares in fixtures and testing whether we get expected `recid` value back

modules/bitgo/test/v2/fixtures/tss/ecdsaFixtures.ts
```diff
@@ -319,13 +319,13 @@ export const mockDShareToBitgo = {
 export const mockSShare = {
   i: 3,
   y: '0362c0f94da66b5cb2880d1bd8c0c407cfa417be23cfc29d3d02aec83a70f6d8ab',
-  r: '181273015b236d9b4de9811a76167db16e20bd34b9bde369ccd352e7bffbbc7f',
+  R: '02181273015b236d9b4de9811a76167db16e20bd34b9bde369ccd352e7bffbbc7f',
   s: '549291233abb0a8f8cf6948b956a596c90b16f9ae15554d77e7a8f69e04ac311',
 };
 
 export const mockSignature = {
```
sachushaji, line 326: NIT: we could rename this to avoid confusion as this is not the final signature
```suggestion
export const mockSShareFromUser = {
```

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
sachushaji: @olibeyene @anshuldoshi  you could make use of this for calculating the 'v' value

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/types.ts
johnoliverdriscoll: Done

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
johnoliverdriscoll: I cant find any unit tests for the mpc library itself which seems very strange. I could just be missing it. Instead I modified the secp256k1 curve implementation no use public key recovery to to verify the signature, so any existing signature verification will automatically test the recid field.

modules/bitgo/test/v2/fixtures/tss/ecdsaFixtures.ts
johnoliverdriscoll: Done

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
sachushaji: https://github.com/BitGo/BitGoJS/tree/master/modules/account-lib/test/unit/mpc/tss/ecdsa tests are in a different module, I understand the confusion

sachushaji (dismissed):
LGTM, Thanks @johnoliverdriscoll

sachushaji:
@olibeyene Kindly take a look at this PR

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
olibeyene: yes this will be braking change. can we keep R lower case?

olibeyene (changes requested):
I'm confused. @johnoliverdriscoll can you explain why we need R instead of r? on chain the signature is constructed by r and s. for eth it is r, s, and v(recid)

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
johnoliverdriscoll: I think it's better that fields communicate their mathematical value. In this case, the value in the field used to be `r` which is the `x` coordinate of the "random" point used in signing. For calculating the `recid` value, we need the parity of the random point's `y` coordinate. So we essentially need the full random point `R` itself. Since we only need to know the `y`-parity, we just use the compressed value and interpret its `y`-parity from its SEC1 encoding prefix byte (`02` being an even parity, `03` being an odd parity).

johnoliverdriscoll:
> I'm confused. @johnoliverdriscoll can you explain why we need R instead of r? on chain the signature is constructed by r and s. for eth it is r, s, and v(recid)

I believe I answered that here: https://github.com/BitGo/BitGoJS/pull/2730#discussion_r963753049

However, I think you may be confused as to where this change is being applied. I am *not* changing the format of signatures aside from adding the `recid` field. Please review the types field in this commit and note that the `Signature` type is now the tuple (`y`, `recid`, `r`, `s`). The `R` field change is only in the `SShare` type. These `SShares` are the contributions from each signer that are used to construct the full `Signature`.

olibeyene (approved):

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
byhow: Since this is a breaking change, we need to reflect that in the commit message for the SDK. ref: https://github.com/BitGo/BitGoJS/pull/2565

We are also working on a PR template so that we can better communicate this effort: https://github.com/BitGo/BitGoJS/pull/2724

byhow (changes requested):
LGTM except needs to address breaking change commit

johnoliverdriscoll:
Commit message updated

mmcshinsky-bitgo:
> Commit message updated

Please follow this guide for your commit: https://www.conventionalcommits.org/en/v1.0.0/. Needs `BREAKING CHANGE: <description>` in the footer.

johnoliverdriscoll:
> > Commit message updated
> 
> Please follow this guide for your commit: https://www.conventionalcommits.org/en/v1.0.0/. Needs `BREAKING CHANGE: <description>` in the footer.

@mmcshinsky-bitgo The full commit message is as follows:

> feat(sdk-core): add recid to fully constructed signature
>
> BREAKING CHANGE: The SShare type's `r` field is now `R` (33 bytes encoded as 66 hex characters).
>
> Issue: BG-56664

https://github.com/BitGo/BitGoJS/pull/2730/commits/e18c99aa7c72c01eead9dc0f51eec662dff6c1ca

Please explain how this does not follow the convention outlined in the guide you linked.

byhow (approved):

PR #2745, opened 2022-09-08
feat(sdk-coin-avaxp): update explain transaction method for export an…

jrafael-simtlix:
…d import transaction

BG-55046

### DIFF
--- a/modules/sdk-coin-avaxp/src/lib/transaction.ts
+++ b/modules/sdk-coin-avaxp/src/lib/transaction.ts
@@ -9,7 +9,7 @@ import {
 } from '@bitgo/sdk-core';
 import { KeyPair } from './keyPair';
 import { DecodedUtxoObj, TransactionExplanation, TxData } from './iface';
-import { AddValidatorTx, AmountInput, AmountOutput, BaseTx, Tx } from 'avalanche/dist/apis/platformvm';
+import { AddValidatorTx, AmountInput, BaseTx, ExportTx, ImportTx, Tx } from 'avalanche/dist/apis/platformvm';
 import { BinTools, BN, Buffer as BufferAvax } from 'avalanche';
 import utils from './utils';
 import { Credential } from 'avalanche/dist/common';
@@ -177,7 +177,7 @@ export class Transaction extends BaseTransaction {
       type: this.type,
       memo: utils.bufferToString(this.avaxPTransaction.getMemo()),
       signatures: this.signature,
-      stakedOutputs: this.stakedOutputs,
+      outputs: this.outputs,
       changeOutputs: this.changeOutputs,
     };
   }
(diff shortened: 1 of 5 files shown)


modules/sdk-coin-avaxp/src/lib/transaction.ts
```diff
@@ -218,44 +218,45 @@ export class Transaction extends BaseTransaction {
...
-        address,
-      };
-    });
+    return this.avaxPTransaction.getOuts().map(utils.mapOutputToEntry(this._network));
   }
 
   get inputs(): Entry[] {
-    const bintools = BinTools.getInstance();
-    return this.avaxPTransaction.getIns().map((input) => {
+    let ins;
```
lcovar, line 250:
```suggestion
    let inputs;
```

CapnMigraine (approved):

CapnMigraine (changes requested):
Please add a test for explaining export and import transactions to https://github.com/BitGo/BitGoJS/blob/%40bitgo/sdk-coin-avaxp%402.2.0-rc.23/modules/sdk-coin-avaxp/test/unit/avaxp.ts#L280

modules/sdk-coin-avaxp/src/lib/transaction.ts
```diff
@@ -218,44 +218,45 @@ export class Transaction extends BaseTransaction {
...
-    // Get staked outputs
-    const addValidatorTx = this.avaxPTransaction as AddValidatorTx;
-    return [
-      {
-        address: addValidatorTx.getNodeIDString(),
-        value: addValidatorTx.getStakeAmount().toString(),
-      },
-    ];
+  get outputs(): Entry[] {
+    if (this.type === TransactionType.Import) {
```
byhow, line 222: might be cleaner with switch statement instead of a list of `if`

modules/sdk-coin-avaxp/src/lib/transaction.ts
```diff
@@ -177,7 +177,7 @@ export class Transaction extends BaseTransaction {
       type: this.type,
       memo: utils.bufferToString(this.avaxPTransaction.getMemo()),
       signatures: this.signature,
-      stakedOutputs: this.stakedOutputs,
+      outputs: this.outputs,
```
byhow, line 180: backward incompatible changes will need to be `BREAKING` changes. see https://github.com/BitGo/BitGoJS/pull/2565

byhow (dismissed):

byhow (commented):
pressed `Approve` by accident. will wait for coin specific team members' approvals first

modules/sdk-coin-avaxp/src/lib/transaction.ts
jrafael-simtlix: done

modules/sdk-coin-avaxp/src/lib/transaction.ts
jrafael-simtlix: done

CapnMigraine (approved):
lgtm

modules/sdk-coin-avaxp/src/lib/transaction.ts
mmcshinsky-bitgo: Note, when we release, this will be a breaking change internally as users switch between the names, but not for the stable release so this won't bump a major version. See:

https://github.com/BitGo/BitGoJS/blob/rel/latest/modules/sdk-coin-avaxp/src/lib/iface.ts#L26
https://github.com/BitGo/BitGoJS/blob/rel/latest/modules/sdk-coin-avaxp/src/lib/transaction.ts#L178

jrafael-simtlix: Since this coin is in limited integration, so there are few users.
Explain Transaction hasn't break changes and is used in SDK and WP. 
Even so, we will watch over the dump the sdk.

mmcshinsky-bitgo (approved):

PR #2756, opened 2022-09-13
feat(bitgo): add lightning and lnurl examples

jamalchowdhury-bitgo-github:
Ticket: [BG-56923](https://bitgoinc.atlassian.net/browse/BG-56923)
Added examples in JavaScript and TypeScript showcasing Lightning/LNURL services in BitGoJS.
Includes:
- Checking a wallet's lightning balance.
- Getting a lightning deposit address and depositing on-chain funds to a wallet's lightning balance.
- Withdrawing from a wallet's lightning balance to on-chain funds.
- Making a lightning payment.
- Creating a lightning invoice.
- Paying an lnurl request.

### DIFF
--- a/modules/bitgo/example/ts/btc/make-lightning-deposit.ts
+++ b/modules/bitgo/example/ts/btc/make-lightning-deposit.ts
@@ -1,0 +1,35 @@
+/**
+ * Create lightning deposit address and deposit on-chain funds to a wallet's lightning balance.
+ *
+ * Copyright 2022 BitGo, Inc.  All Rights Reserved.
+ */
+const BitGoJS = require('bitgo');
+
+// change this to env: 'prod' when you are ready for production
+const env = 'test';
+// change coin to 'btc' when working with production
+const coin = env === 'test' ? 'tbtc' : 'btc';
+
+const bitgo = new BitGoJS.BitGo({ env });
+
+// set your access token here
+const accessToken = '';
+
+// set your wallet ID here
+const walletId = '';
+
+async function makeLightningDeposit() {
+  bitgo.authenticateWithAccessToken({ accessToken });
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  const lightning = wallet.lightning();
+
+  const initialLightningBalance = await lightning.getBalance();
+  console.log('Initial Lightning Balance:', initialLightningBalance.balance);
+  const initialBalance = wallet.balance();
+  console.log('Initial Balance:', initialBalance);
+
+  const deposit = await lightning.deposit({ amount: 100000 });
+  console.log('Deposit Response:', deposit);
+}
+
+makeLightningDeposit().catch(console.error);
--- a/modules/bitgo/example/README.md
+++ b/modules/bitgo/example/README.md
@@ -41,6 +41,15 @@ In this directory, you can find examples on how to use the BitGoJS SDK with Type
 * [Enable Token](./ts/algo/enable-token.ts)
 * [Transaction with emergency param](./ts/algo/transaction-with-emergency-param.ts)
 
+### Bitcoin Lightning (BTC)
+
+* [Check Lightning Balance](./ts/btc/check-lightning-balance.ts)
+* [Create Lightning Invoice](./ts/btc/create-lightning-invoice.ts)
+* [Withdraw Lightning Balance](./ts/btc/withdraw-lightning-balance.ts)
+* [Make Lightning Deposit](./ts/btc/make-lightning-deposit.ts)
+* [Make Lightning Payment](./ts/btc/make-lightning-payment.ts)
+* [Pay LNURL Request](./ts/btc/pay-lnurl-request.ts)
+
 ### Ethereum (ETH)
 
 * [Create Wallet Address](./ts/eth/create-wallet-address.ts)
--- a/modules/bitgo/example/ts/btc/check-lightning-balance.ts
+++ b/modules/bitgo/example/ts/btc/check-lightning-balance.ts
@@ -1,0 +1,31 @@
+/**
+ * Check the lightning balance of a wallet.
+ *
+ * Copyright 2022 BitGo, Inc.  All Rights Reserved.
+ */
+const BitGoJS = require('bitgo');
+
+// change this to env: 'prod' when you are ready for production
+const env = 'test';
+// change coin to 'btc' when working with production
+const coin = env === 'test' ? 'tbtc' : 'btc';
+
+const bitgo = new BitGoJS.BitGo({ env });
+
+// set your access token here
+const accessToken = '';
+
+// set your wallet ID here
+const walletId = '';
+
+async function checkLightningBalance() {
+  bitgo.authenticateWithAccessToken({ accessToken });
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  const lightning = wallet.lightning();
+
+  const lightningBalance = await lightning.getBalance();
+  console.log('Lightning Balance:', lightningBalance.balance);
+  console.log('Lightning Balance Available for Spending:', lightningBalance.availableBalance);
+}
+
+checkLightningBalance().catch(console.error);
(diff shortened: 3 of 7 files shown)


modules/bitgo/example/js/btc/check-lightning-balance.js
```diff
@@ -0,0 +1,31 @@
...
+async function checkLightningBalance() {
+  bitgo.authenticateWithAccessToken({ accessToken });
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  const lightning = wallet.lightning();
+
+  const lightningBalance = await lightning.getBalance();
+  console.log('Lightning Balance:', lightningBalance.balance);
+}
+
+checkLightningBalance().catch((e) => console.error(e));
```
sangaman, line 31: Nit but here (and anywhere else you're using `.catch` in these examples) you can just do
```suggestion
checkLightningBalance().catch(console.error);
```

modules/bitgo/example/js/btc/check-lightning-balance.js
```diff
@@ -0,0 +1,31 @@
...
+// https://test.bitgo.com/enterprise/XXXXXXXXX/coin/tbtc/YYYYY/transactions
+const walletId = '';
+
+async function checkLightningBalance() {
+  bitgo.authenticateWithAccessToken({ accessToken });
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  const lightning = wallet.lightning();
+
+  const lightningBalance = await lightning.getBalance();
+  console.log('Lightning Balance:', lightningBalance.balance);
```
sangaman, line 28: Let's log the `availableBalance` as well and call it "Lightning Balance Available for Spending"

The distinction fyi is that `availableBalance` subtracts payments that are still in a pending status and may fail or succeed. This number isn't allowed to go in the negative.

modules/bitgo/example/js/btc/check-lightning-balance.js
```diff
@@ -0,0 +1,31 @@
...
+ * Check the lightning balance of a wallet.
+ *
+ * Copyright 2022 BitGo, Inc.  All Rights Reserved.
+ */
+const BitGoJS = require('bitgo');
+
+// change this to env: 'production' when you are ready for production
+const env = 'test';
+// change coin to 'btc' when working with production
+const coin = 'tbtc';
```
sangaman, line 11: How about
```suggestion
const coin = env === 'prod' ? 'btc' : 'tbtc';
```

modules/bitgo/example/js/btc/create-lightning-invoice.js
```diff
@@ -0,0 +1,31 @@
...
+// https://test.bitgo.com/enterprise/XXXXXXXXX/coin/tbtc/YYYYY/transactions
+const walletId = '';
+
+async function createLightningInvoice() {
+  bitgo.authenticateWithAccessToken({ accessToken });
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  const lightning = wallet.lightning();
+
+  const walletInvoice = await lightning.createInvoice({ value: 100000, memo: 'Test Payment' });
+  console.log('Wallet Invoice:', walletInvoice.invoice);
```
sangaman, line 28: I think rather than "Wallet Invoice" it'd be more accurate/specific to call this a "Lightning Invoice"

modules/bitgo/example/js/btc/make-lightning-deposit.js
```diff
@@ -0,0 +1,39 @@
...
+
+  const lightningDepositAddress = await lightning.createDepositAddress();
+  console.log('Lightning Deposit Address:', lightningDepositAddress.address);
+
+  const initialLightningBalance = await lightning.getBalance();
+  console.log('Initial Lightning Balance:', initialLightningBalance.balance);
+  const initialBalance = wallet.balance();
+  console.log('Initial Balance:', initialBalance);
+
+  const deposit = await lightning.deposit({ amount: 100000, address: lightningDepositAddress.address });
```
sangaman, line 35: I think we can simplify this a bit and not make the separate `createDepositAddress` call. If we don't pass an `address` to `lightning.deposit` it will make one for us.

As a side note, I'm actually thinking we should remove `address` as a param for `lightning.deposit` altogether. We're not checking that the provided address is actually a lightning address, and there's no easy way to check for that.

modules/bitgo/example/js/btc/make-lightning-payment.js
```diff
@@ -0,0 +1,36 @@
...
+async function makeLightningPayment() {
+  bitgo.authenticateWithAccessToken({ accessToken });
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  const lightning = wallet.lightning();
+
+  const initialLightningBalance = await lightning.getBalance();
+  console.log('Initial Lightning Balance:', initialLightningBalance.balance);
+
+  await bitgo.unlock({ otp: '000000', duration: 3600 });
+  const walletInvoice = await lightning.createInvoice({ value: 1000, memo: 'Test Payment' });
```
sangaman, line 31: In this case the wallet would be paying itself, which we could do but I think might be confusing. Maybe just declare something like

```
const invoiceToPay = ''; // set the invoice you want to pay here
```

modules/bitgo/example/js/btc/withdraw-lightning-balance.js
```diff
@@ -0,0 +1,36 @@
...
+  bitgo.authenticateWithAccessToken({ accessToken });
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  const lightning = wallet.lightning();
+
+  const initialLightningBalance = await lightning.getBalance();
+  console.log('Initial Lightning Balance:', initialLightningBalance.balance);
+  const initialBalance = wallet.balance();
+  console.log('Initial Balance:', initialBalance);
+
+  const fixtureWithdraw = await lightning.withdraw({ value: 10000 });
```
sangaman, line 32: Why is this called `fixtureWithdraw`? I think it could just be `withdrawResponse`.

sangaman (changes requested):
I realize I made all my comments on the .js files when I meant to comment on .ts. So please make feedback updates to the ts files. The examples README says this about js:

> In this directory, you can find examples on how to use the BitGoJS SDK with JavaScript. These examples use coroutine syntax with yield.

So I dunno if you'd want to refactor the js examples to use `yield` to be consistent with that comment - or I'd also be fine with dropping them and just having `.ts` examples.

Also I think it'd be nice to include a "lightning" section in the examples README.md that links to these.

Otherwise this is awesome and will be a great resource for new BitGo lightning users.

modules/bitgo/example/ts/btc/make-lightning-deposit.ts
```diff
@@ -0,0 +1,39 @@
...
+// change coin to 'btc' when working with production
+const coin = 'tbtc';
+
+const bitgo = new BitGoJS.BitGo({ env });
+
+// set your access token here
+const accessToken = '';
+
+// set your wallet from the YYYYY parameter here in the URL on app.bitgo-test.com
+// https://test.bitgo.com/enterprise/XXXXXXXXX/coin/tbtc/YYYYY/transactions
```
mmcshinsky-bitgo, line 19: // https://api.bitgo.com/enterprise/XXXXXXXXX/coin/tbtc/YYYYY/transactions
Where is the documentation for this endpoint?

modules/bitgo/example/js/btc/check-lightning-balance.js
jamalchowdhury-bitgo-github: Added `const coin = env === 'test' ? 'tbtc' : 'btc';` because TypeScript complains the other way round.

sangaman (dismissed):
Good stuff, TY.

mmcshinsky-bitgo (approved):

mmcshinsky-bitgo:
@jamalchowdhury-bitgo-github LGTM. You need to sign the commits for this to be able to merge.

mmcshinsky-bitgo (approved):

PR #2760, opened 2022-09-13
chore: update token metadata for kylin eos network

lcovar:
TICKET: BG-57018

## Description

This change updates token metadata for the EOS kylin network. Previously we were using jungle3 but that will no longer be available. Token contract addresses stay the same except for CHEX. `testtoken111->testtoken113`


margueriteblair (approved):
:D

byhow (approved):

PR #2764, opened 2022-09-14
fix(bitgo): remove address param from lightning().deposit

jamalchowdhury-bitgo-github:
Removed `address` param for `lightning().deposit` from SDK because there is no easy way to check if address is lightning address. A user could potentially provide an arbitrary, non-lightning address which we’d send funds to and end up not adding anything to the wallet’s custodial lightning balance

Ticket: BG-57388

### DIFF
--- a/modules/sdk-core/src/bitgo/lightning/lightning.ts
+++ b/modules/sdk-core/src/bitgo/lightning/lightning.ts
@@ -91,11 +91,7 @@ export class Lightning implements ILightning {
 
   public async deposit(params: LightningDepositParams): Promise<DepositResponse> {
     const { amount } = params;
-    let { address } = params;
-
-    if (address === undefined) {
-      address = (await this.createDepositAddress()).address;
-    }
+    const address = (await this.createDepositAddress()).address;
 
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/lightning/lightning.ts
```diff
@@ -91,13 +91,8 @@ export class Lightning implements ILightning {
...
   public async deposit(params: LightningDepositParams): Promise<DepositResponse> {
     const { amount } = params;
-    let { address } = params;
 
-    if (address === undefined) {
-      address = (await this.createDepositAddress()).address;
-    }
-
-    const res = await this.wallet.send({ amount, address });
+    const res = await this.wallet.send({ amount });
```
sangaman, line 95: We still need to specify an address when we send coins, we just want to make sure that we use `createDepositAddress` to get the address.

sangaman (changes requested):

sangaman (approved):
TY!

mmcshinsky-bitgo (approved):

PR #2766, opened 2022-09-14
fix(sdk-core): ecdsa tss signing flow update

olibeyene:
Ticket: BG-57413

WP updated ecdsa tss signing flow, this pr is matching that signing flow

### DIFF
--- a/modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
@@ -29,6 +29,7 @@ import { BitGoBase } from '../../bitgoBase';
 import { KShare, MUShare, SShare } from '../../../account-lib/mpc/tss/ecdsa/types';
 import { getTxRequest, sendSignatureShare } from '../common';
 import createKeccakHash from 'keccak';
+import assert from 'assert';
 
 const MPC = new Ecdsa();
 
@@ -168,6 +169,8 @@ export async function createUserSignatureShare(
   return MPC.sign(message, oShare, dShare, createKeccakHash('keccak256'));
 }
 
+export type MuDShare = { muShare: MUShare; dShare: DShare; i: ShareKeyPosition };
+
 /**
  * Sends Share To Bitgo
  * @param {BitGoBase} bitgo - the bitgo instance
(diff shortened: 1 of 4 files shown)


byhow (approved):
has very little context on how TSS flow works, might be good to get someone else who's familiar with that to get a review as well

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -181,8 +184,8 @@ export async function sendShareToBitgo(
   walletId: string,
   txRequestId: string,
   shareType: SendShareType,
-  share: SShare | MUShare | KShare,
-  dShare?: DShare
+  share: SShare | MuDShare | KShare,
```
sachushaji, line 187: NIT: can you update the function params description also with the change

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -192,27 +195,28 @@ export async function sendShareToBitgo(
 
   switch (shareType) {
     case SendShareType.KShare:
+      assert(signerShare, `signer share must be present`);
```
sachushaji, line 198: Also I can see that signerShare also sent out in MuShare block also, so would the assert applicable for both the cases?

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -290,36 +290,42 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
     }
+
     const MPC = new Ecdsa();
     const signingKey = MPC.keyCombine(userSigningMaterial.pShare, [
       userSigningMaterial.bitgoNShare,
       userSigningMaterial.backupNShare,
     ]);
 
     const userSignShare = await ECDSAMethods.createUserSignShare(signingKey.xShare, signingKey.yShares[3]);
+    const signerShare = userSigningMaterial.bitgoNShare.u + userSigningMaterial.bitgoNShare.chaincode;
```
sachushaji, line 301: Why are we sending userSigning material encrypted if its available locally while user signs it? unless I am missing something

sachushaji (approved):
Some NITs and Question, otherwise LGTM

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
olibeyene: @johnoliverdriscoll can you explain why hsm needs the user's signing material? I'm not sure why there's a need but the first call to hsm requires that

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
olibeyene: signerShare is only required when sending K share i.e. the first part of the signing. at other stage signer share could be whatever.

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
olibeyene: we can but I don't want to hold up this pr

PR #2773, opened 2022-09-15
feat(utxo-lib): add UtxoPsbt w/BitGO P2TR signing

brandonblack:
* Extends upstream PSBT class
* Tests assert that resulting signed transactions are identical between PSBT and TransactionBuilder
* Only supports BitGo style 2-of-2 script path signing

Issue: BG-47820

Co-authored-by: davidkaplan@bitgo.com

## Type of change

# How Has This Been Tested?

# Checklist:


modules/utxo-lib/src/bitgo/Unspent.ts
```diff
@@ -110,6 +124,66 @@ export function addToTransactionBuilder<TNumber extends number | bigint>(
...
+  network: Network
+): void {
+  const { txid, vout, script, value } = toPrevOutput(u, network);
+  const { walletKeys, signerIndex, cosignerIndex } = rootSigner.deriveForChainAndIndex(u.chain, u.index);
+  const scriptType = scriptTypeForChain(u.chain);
+  psbt.addInput({
+    hash: txid,
+    index: vout,
+  });
+  const inputIndex = psbt.inputCount - 1;
```
sangaman, line 140: Why is this `- 1` and not `+ 1`? Since we're adding an input to a tx, I'd expect that input would increase the input count and go at the end of the index.

modules/utxo-lib/src/bitgo/Unspent.ts
```diff
@@ -110,6 +124,66 @@ export function addToTransactionBuilder<TNumber extends number | bigint>(
...
+    }
+    psbt.updateInput(inputIndex, { nonWitnessUtxo: u.prevTx });
+  }
+
+  if (scriptType === 'p2tr') {
+    const { controlBlock, witnessScript, leafVersion, leafHash } = createSpendScriptP2tr(walletKeys.publicKeys, [
+      walletKeys.triple[signerIndex].publicKey,
+      walletKeys.triple[cosignerIndex].publicKey,
+    ]);
+    psbt.updateInput(inputIndex, {
```
sangaman, line 160: In the case of a segwit unspent, we're calling `updateInput` twice. Once on line 142 and then again either here or in the `else` block for non-p2tr. Is there a reason we have to call it twice? Or can it be in one call? Where we'd check for either non-segwit, segwit p2tr, or segwit non-p2tr and have one updateInput call for each.

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -0,0 +1,319 @@
...
+    // Upstream checks for duplicate inputs here, but it seems to be of dubious value.
+    return psbt;
+  }
+
+  protected static newTransaction(network: Network): UtxoTransaction<bigint> {
+    return new UtxoTransaction<bigint>(network);
+  }
+
+  get tx(): Tx {
+    return (this.data.globalMap.unsignedTx as PsbtTransaction).tx as Tx;
```
sangaman, line 66: Why is the casting necessary here?

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  /**
+   * Mostly copied from bitcoinjs-lib/ts_src/psbt.ts
+   */
```
sangaman, line 71: Does this apply to both methods below? What code changed? Is it possible to just call these mostly copied methods directly since we're already doing `extends Psbt`. In the case of `validateSignaturesOfAllInputs` below for example I'm thinking we should be able to call `super()` and pass in the hardcoded validator we want.

modules/utxo-lib/src/bitgo/transaction.ts
```diff
@@ -82,6 +83,61 @@ export function setTransactionBuilderDefaults<TNumber extends number | bigint>(
...
+    case networks.zcash:
+      throw new Error('Zcash psbt not implemented');
+    default:
+      if (version !== 1) {
+        throw new Error(`invalid version`);
+      }
+  }
+  // FIXME: Always call this, because there's a bug in the upstream PSBT that
+  // defaults transactions to v2.
+  psbt.setVersion(version);
```
sangaman, line 108: Is it safe to be setting version to `undefined` in the case that it's not specified in the params? Or should we default `version = 2` in case it's not provided?

modules/utxo-lib/src/bitgo/wallet/chains.ts
```diff
@@ -102,3 +102,11 @@ export function isExternalChainCode(v: ChainCode): boolean {
 export function isInternalChainCode(v: ChainCode): boolean {
   return toChainPair(v).indexOf(v) === 1;
 }
+
+/**
+ * @return true iff chain code is a segwit address
+ */
+export function isSegwit(v: ChainCode): boolean {
+  const segwitCodes: ChainCode[] = [...chainCodesP2shP2wsh, ...chainCodesP2wsh, ...chainCodesP2tr];
```
sangaman, line 110: It might make sense to include the non-script `p2wpkh` type as well here just for completeness sake.

modules/utxo-lib/src/noble_ecc.ts
```diff
@@ -68,7 +68,11 @@ const ecc = {
     throwToNull(() => necc.utils.pointAddScalar(p, tweak, defaultTrue(compressed))),
 
   privateAdd: (d: Uint8Array, tweak: Uint8Array): Uint8Array | null =>
-    throwToNull(() => necc.utils.privateAdd(d, tweak)),
+    throwToNull(() => {
```
sangaman, line 71: What's the purpose of this change?

sangaman (commented):
Looks good overall but I made a handful of comments and questions.

modules/utxo-lib/src/bitgo/wallet/chains.ts
sangaman: Never mind I realized this is using our bitgo chaincodes and we don't have one for p2wpkh since we don't use it, so disregard.

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
brandonblack: Because the `unsignedTx` is an `bip174.Transaction` interface, but we need to tell it what concrete implementation `bitcoinjs-lib.PsbtTransaction` we were using to get the `tx` off of it.

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
brandonblack: No, just this method. There is no Taproot support upstream. We can't call super because the Taproot fields are totally disjoint from the pre-Taproot fields. The validator just checks the sigs, but the sigs have to be fetched from an entirely other place.

modules/utxo-lib/src/bitgo/transaction.ts
brandonblack: on line 89, `getDefaultTransactionVersion`, and we throw in the switch if it's not set to expected values.

I think there is a bug here that we say that we don't support zcash, but we do. Will have to look into that.

modules/utxo-lib/src/noble_ecc.ts
brandonblack: Fixes an error caused by `ECPair`'s (strongly against my preferences) `testecc` changes. Could use a comment.

modules/utxo-lib/src/bitgo/Unspent.ts
brandonblack: Could do it in one call, but it's more appropriate to do it in separate calls, because of the logic flow.

"Any segwit gets a witnessUtxo, taproot gets taproot fields, segwitv0 gets witnessScript"

`updateInput` is intended to be called many times, often by separate updaters (eg. one might have access to a full node to fetch UTXOs, and another might have access to pubkeys to generate scripts).

modules/utxo-lib/src/bitgo/Unspent.ts
brandonblack: We _just_ added it, so the index (0-base) of the one we just added is the count (1-base) minus one.

modules/utxo-lib/package.json
```diff
@@ -49,14 +49,15 @@
     "@bitgo/blake2b": "^3.0.3",
     "@noble/secp256k1": "1.6.3",
     "bip32": "^3.0.1",
+    "bip174": "npm:@bitgo/bip174@3.0.0",
     "bitcoin-ops": "^1.3.0",
-    "bitcoinjs-lib": "npm:@bitgo/bitcoinjs-lib@7.0.0-rc.0",
+    "bitcoinjs-lib": "git+https://github.com/BitGo/bitcoinjs-lib#3876701401d62a332c4901aa2a6f6b71c3cd799d",
```
mmcshinsky-bitgo, line 54: Is it possible to get another publish for this?

modules/utxo-lib/package.json
```diff
@@ -49,14 +49,15 @@
...
     "bn.js": "^5.2.1",
     "bs58check": "^2.1.2",
     "cashaddress": "^1.1.0",
     "create-hash": "^1.2.0",
     "create-hmac": "^1.1.7",
-    "ecpair": "^2.0.1",
+    "ecpair": "git+https://github.com/brandonblack/ecpair#95667235852b2c28565012c292d6c9cb3b2eee38",
```
mmcshinsky-bitgo, line 60: Is there a todo to revert when this PR is merged and published? https://github.com/bitcoinjs/ecpair/pull/11

modules/utxo-lib/package.json
brandonblack: Yes, pending a little PR merge there, we will publish.

modules/utxo-lib/package.json
brandonblack: Yep, we're discussing the options there, as the maintainer has been idle for some months. May end up having to fork off a BitGo fork.

sangaman (approved):
LGTM pending questions on dependencies/forks.

mmcshinsky-bitgo (approved):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -0,0 +1,319 @@
...
+          leafHash,
+        },
+      ],
+    });
+    return this;
+  }
+
+  private getTaprootHashForSig(
+    inputIndex: number,
+    forValidate: boolean,
```
OttoAllmendinger, line 269: unused - what was the intended purpose?

brandonblack: Exists because it exists on the upstream equivalent method. Would have to go to the upstream `getHashForSig` and read the code to remember why it exists there.

PR #2776, opened 2022-09-17
feat(sdk-core): add VSS share generation and verification

johnoliverdriscoll:
ISSUE: BG-57633

## Description

VSS (Verifiable Secret Sharing) is a method by which the Verifier can be assured that they have received a share of a secret controlled by the Prover. This allows all parties in a DKG scheme to be assured that the shares they have received from their cohorts are not spoofed and that it is possible to spend from the common key.

## Issue Number

BG-57633

## Type of change

Unsure what I'm supposed to do in this section...

# How Has This Been Tested?

`shamir.verify` is tested in account-lib unit tests and in hsm-firmware test suite.

# Checklist:


modules/sdk-core/src/account-lib/mpc/shamir.ts
```diff
@@ -27,7 +27,7 @@ export default class Shamir {
     numShares: number,
     indices?: Array<number>,
     salt = BigInt(0)
-  ): Record<number, bigint> {
+  ): [Record<number, bigint>, Array<bigint>] {
```
sachushaji, line 30: NIT: Can we update the new return type in the return type description of this function also. Also if it makes sense, instead of returning shares and `v` as an array, can we consider to return it as an object like 
```suggestion
  ): { shares: Record<number, bigint>,v:  Array<bigint>} {
```
Would be become more clearer on what exactly is being returned.

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -106,6 +107,15 @@ export default class Ecdsa {
     // Add secret shares
     const x = allShares.map((participant) => hexToBigInt(participant['u'])).reduce(Ecdsa.curve.scalarAdd);
 
+    // Verify shares.
+    for (const share of nShares) {
+      try {
+        Ecdsa.shamir.verify(hexToBigInt(share.u), [hexToBigInt(share.y), hexToBigInt(share.v)], pShare.i);
+      } catch (err) {
+        throw new Error(`Could not verify share from participant ${share.j}`);
```
sachushaji, line 115: NIT: Can we add actual the error thrown also to to the message so that it will also give info on what check at which the verification actually failed.
```suggestion
        throw new Error(`Could not verify share from participant ${share.j}. Verification Error: ${err}`);
```

sachushaji (dismissed):
Some NITs and Suggestions, Otherwise LGTM, Thanks John

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
johnoliverdriscoll: Done

modules/sdk-core/src/account-lib/mpc/types.js
```diff
@@ -0,0 +1,4 @@
+export interface SplitSecret {
```
byhow, line 1: should this be a `.ts` file?

modules/sdk-core/src/account-lib/mpc/tss/eddsa/types.ts
```diff
@@ -11,6 +11,7 @@ export interface YShare {
   i: number;
```
byhow, line 11: backward incompatible changes needs to be addressed as `BREAKING` in the commit comments, if this type change will cause incompatibility

byhow (commented):
might need a rebase against latest master to pass CI as we applied the fix at 07e8a37149cf7d09a52d56d9a3c1d454cfbb9a98

modules/sdk-core/src/account-lib/mpc/tss/eddsa/types.ts
johnoliverdriscoll: I really don't know. @sachushaji Do you know if this is a backwards incompatible change?

modules/sdk-core/src/account-lib/mpc/types.js
johnoliverdriscoll: fixed

modules/sdk-core/src/account-lib/mpc/tss/eddsa/types.ts
johnoliverdriscoll: Added breaking change info to commit footer.

modules/bitgo/test/v2/unit/tss/eddsa.ts
```diff
@@ -397,7 +398,7 @@ describe('test tss helper functions', function () {
 
     describe('offerUserToBitgoRShare:', async function() {
       it('should succeed to send Signature Share', async function() {
-        const signatureShare = { from: 'user', to: 'bitgo', share: validUserSignShare.rShares[3].r + validUserSignShare.rShares[3].R } as SignatureShareRecord;
+        const signatureShare = { from: 'user', to: 'bitgo', share: validUserSignShare.rShares[3].v + validUserSignShare.rShares[3].r + validUserSignShare.rShares[3].R } as SignatureShareRecord;
```
sachushaji, line 401: Will this change be backward compatible for existing uses who do not have "v" in keys

sachushaji (dismissed):
Thanks for addressing the changes requested, Great Work !!!. Seems like the v changes have been isolated mostly to the key generation and hopefully wouldn't break existing clients for signing. But I suppose in order for this new key gen to work, every client has to make use of the latest SDK updating from the old version to successfully generate keys shares once this change is updated to the platform. @olibeyene @joey-bai , Kindly take a note of this. Requesting review from you guys also.

modules/bitgo/test/v2/unit/tss/eddsa.ts
johnoliverdriscoll: Yes, the v field in the signing share is generated from their signing key at the time of signing.

johnoliverdriscoll:
Any updates on this review?

byhow (dismissed):

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -296,6 +296,7 @@ export async function encryptNShare(
 
   const publicShare = Buffer.concat([
```
olibeyene, line 297: is this change supported by hsm? I think this change will break key creation

olibeyene (changes requested):
Please verify this is not a breaking change with the current released WP and HSM. we only have few days to release ETH TSS.

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
johnoliverdriscoll: This is the public share, it's not sent to the HSM, at least not in this format. The HSM inputs are parameterized via object properties much the same way that the sdk-core tss module is.

johnoliverdriscoll: Unless I'm mistaken, the changes regarding sending the new v field would need to be made to WP.

olibeyene: Currently this change will have v part of the public share (y) prop that will be sent to HSM from wp. if we want to exclude the v from y then we'll need to update WP.

modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
```diff
@@ -185,7 +197,7 @@ export async function offerUserToBitgoRShare(
   const signatureShare: SignatureShareRecord = {
     from: SignatureShareType.USER,
     to: SignatureShareType.BITGO,
-    share: rShare.r + rShare.R,
+    share: rShare.v + rShare.r + rShare.R,
```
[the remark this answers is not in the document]
johnoliverdriscoll: Not sure I understand the user's V value would be undefined since the user is the one running the SDK

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -66,6 +66,7 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
       i: 1,
       j: 3,
       y: bitGoToUserShare.publicShare.slice(0, 64),
+      v: bitGoToUserShare.publicShare.slice(64, 128),
```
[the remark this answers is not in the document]
johnoliverdriscoll: I think the existing `'Eddsa tss signing helper function'` test in modules/bitgo/test/v2/unit/tss/eddsa.ts tests this? I could totally be wrong though.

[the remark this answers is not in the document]
joey-bai: you're right about this.

the signing material (TSS Key) do not have `v` properties defined.

modules/sdk-core/src/account-lib/mpc/tss/eddsa/eddsa.ts
```diff
@@ -124,6 +125,21 @@ export default class Eddsa {
     const chaincodes = [uShare, ...yShares].map(({ chaincode }) => bigIntFromBufferBE(Buffer.from(chaincode, 'hex')));
     const chaincode = chaincodes.reduce((acc, chaincode) => (acc + chaincode) % base);
 
+    // Verify shares.
+    for (const share of yShares) {
+      if ('v' in share) {
```
joey-bai, line 130: do we expect cases where `v` can exist for some yShares but not all? 

also, should we just enforce that all yShares should have `v` or non do?

joey-bai (dismissed):
lgtm for EDDSA

modules/sdk-core/src/account-lib/mpc/tss/eddsa/eddsa.ts
johnoliverdriscoll: The user key shares in wallets that are currently in production will not have the `v` value. This will be an issue for the HSM, but we have a solution in mind to address this. But we need all this VSS stuff live in the SDK before we can apply it.

johnoliverdriscoll:
@olibeyene can you please review and/or approve? These changes are compatible with the current HSM version, and we need them implemented in SDK before we can implement VSS in the HSM.

olibeyene (dismissed):

joey-bai (approved):

PR #2779, opened 2022-09-19
feat(sdk-coin-dot): initial implementation of dot recover function an…

alanliu-bitgo:
## Description

Implements the recover function for Polkadot.

## Issue Number

BG-57231

## Type of change

# How Has This Been Tested?

Unit test for the recover function


lcovar (approved):
Thanks for the tests!

modules/sdk-coin-dot/src/dot.ts
```diff
@@ -276,7 +513,7 @@ export class Dot extends BaseCoin {
   }
 
   async verifyTransaction(params: VerifyTransactionOptions): Promise<boolean> {
-    return true;
+    throw new MethodNotImplementedError();
```
lcovar, line 516: Im not 100% sure on this but I think if we throw an error here then it will break OVC which is a piece of software we use to sign txns offline. Lets keep this as true and add a TODO with a ticket number to implement this.

lcovar (changes requested):
Im not 100% sure on this but I think if we throw an error here then it will break OVC which is a piece of software we use to sign txns offline. Lets keep this as true and add a TODO with a ticket number to implement this.

https://github.com/BitGo/BitGoJS/pull/2779/files#diff-8504cbe7beeac3b47bb6246b51de11a527c99d0b2d73e8bd03d5d4a9345268adR516

lcovar (dismissed):

lcovar (approved):

byhow (approved):

PR #2783, opened 2022-09-20
feat: The client needs to generate a GPG key for their backup ...

vitalir-bg:
key share and share it with BitGo

Ticket: BG-53427

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 
+import * as pgp from 'openpgp';
 import {
   createMessage,
   decrypt,
@@ -12,9 +13,9 @@ import {
   sign,
   verify,
 } from 'openpgp';
-import * as pgp from 'openpgp';
 import * as _ from 'lodash';
 import { BitGoBase } from '../bitgoBase';
+import crypto from 'crypto';
 
 const sodium = require('libsodium-wrappers-sumo');
 
@@ -81,7 +82,7 @@ export async function verifySharedDataProof(
   const anyInvalidProof = _.some(
     // @ts-ignore
     primaryUser.user.otherCertifications[0].rawNotations,
-    (notation) => dataToVerify.find((i) => i.name == notation.name)?.value !== Buffer.from(notation.value).toString()
+    (notation) => dataToVerify.find((i) => i.name === notation.name)?.value !== Buffer.from(notation.value).toString()
   );
   return !anyInvalidProof;
 }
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -341,3 +342,34 @@ export async function verifySignature(text: string, armoredSignature: string, pu
...
 }
+
+/**
+ * Generate a GPG key pair
+ *
+ * @param: keyCurve the curve to create a key with
+ * @param: username name of the user (optional)
+ * @param: email email of the user (optional)
+ */
+export async function generateGPGKeyPair(
```
mmcshinsky-bitgo, line 353: Can this be implemented and tested into all the places where `.generateKey(` is being used?

I see about 20 calls that could be changed to use this.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -54,18 +54,7 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
-    const userGpgKey = await openpgp.generateKey({
-      userIDs: [
-        {
-          name: randomHexString,
-          email: `user-${randomHexString}@${randomHexString}.com`,
-        },
-      ],
-      curve: 'secp256k1',
-    });
+    const userGpgKey = await generateGPGKeyPair('secp256k1');
```
mmcshinsky-bitgo, line 57: Missing username and email?

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -223,19 +222,7 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
...
-    const userGpgKey = await openpgp.generateKey({
-      userIDs: [
-        {
-          name: randomHexString,
-          email: `user-${randomHexString}@${randomHexString}.com`,
-        },
-      ],
-      curve: 'secp256k1',
-    });
+    const userGpgKey = await generateGPGKeyPair('secp256k1');
```
mmcshinsky-bitgo, line 225: Missing username and email?

mmcshinsky-bitgo (changes requested):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
vitalir-bg: user name and email parameters are optional, if they are missing they are generated randomly inside this function.
In all places in the current PR the values for both name and email were randomly generated and followed the same format so it is not necessary to explicitly provide them.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
vitalir-bg: see my earlier comment

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
vitalir-bg: Not sure this will work for all cases since in tests the curve parameter is often omitted and the default value for it seems to be 'curve25519'. Not sure what the side effects gonna be if I change this.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
mmcshinsky-bitgo: Awesome, thanks!

mmcshinsky-bitgo (approved):

PR #2785, opened 2022-09-20
feat(doge): handling valueString from wallet platform response

dpkjnr:
Changed type of minValue and maxValue params to support both number and string in consolidate and fanout API.
Wallet platform will support both types for all the coins except for doge and for doge it will only support string


Phani024:
Looks good. Maybe add a couple of unit tests related to this?

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -25,7 +25,7 @@ export interface MaximumSpendableOptions {
   maxFeeRate?: number;
   recipientAddress?: string;
   limit?: number;
-  target?: number;
+  target?: number | string;
```
sachushaji, line 28: Instead of adding `| string` to every place, We can introuce a new `type NumberLike = number | string ` if not already there and relace all places where we make this change
```suggestion
  target?: NumberLike;
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -648,8 +648,8 @@ export class Wallet implements IWallet {
    * @param {Number} params.maxFeeRate - upper limit for feeRate in satoshis/kB
    * @param {Number} params.maxFeePercentage - the maximum relative portion that you're willing to spend towards fees
    * @param {Number} params.feeTxConfirmTarget - estimate the fees to aim for first confirmation with this number of blocks
-   * @param {Number} params.minValue - the minimum value of unspents to use in satoshis
-   * @param {Number} params.maxValue - the maximum value of unspents to use in satoshis
+   * @param {Number | String} params.minValue - the minimum value of unspents to use in satoshis
```
sachushaji, line 651:
```suggestion
   * @param {number | string} params.minValue - the minimum value of unspents to use in satoshis
```

sachushaji: the param should match the type, can you also make this change to all places you modified also

sachushaji (commented):
Changes LGTM overall, small suggestions.Can you also add some unit tests to make sure this actually works when passed with number in stribg format also

modules/sdk-core/src/bitgo/wallet/iWallet.ts
dpkjnr: I couldn't find any correct name for this type and I think it would be better to keep type in string | number format to keep it more readable.

modules/sdk-core/src/bitgo/wallet/wallet.ts
dpkjnr: added at other places too

dpkjnr:
> Changes LGTM overall, small suggestions.Can you also add some unit tests to make sure this actually works when passed with number in stribg format also

added unit tests to verify if string format of minValue/maxValue is passed to wallet platform

modules/sdk-core/src/bitgo/wallet/iWallet.ts
sachushaji: The suggestion was more like introducing this type if not available, but I think its okay to have it as is also too

sachushaji (dismissed):
LGTM, Nice Job @dpkjnr

saravanan7mani (changes requested):
https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/wallet.ts#L627

In the above line, consolidateUnspents API response has `unspents` objects. Currently `txInfo.unspents[i].value` (`number` type) is used for sign tx here. But `valueString` should be used.

To do: 
Parse `response` object and `valueString` and convert to `bigint` type and assign it to `txInfo.unspents[i].value`

saravanan7mani:
> > > https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/wallet.ts#L627
> > > In the above line, consolidateUnspents API response has `unspents` objects. Currently `txInfo.unspents[i].value` (`number` type) is used for sign tx here. But `valueString` should be used.
> > > To do: Parse `response` object and `valueString` and convert to `bigint` type and assign it to `txInfo.unspents[i].value`
> > 
> > 
> > where did you find this? I didn't find any such usecase
> 
> The link provides the part of the code that comes with consolidateUnspents function flow.

Sweep Wallet requires similar change

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2070,4 +2071,114 @@ describe('V2 Wallet:', function () {
       });
     });
   });
+
+  describe('unspents APIs with string type minValue and maxValue', function () {
```
OttoAllmendinger, line 2075: move these tests into a new file `unit/wallet/unspents.ts`

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2070,4 +2071,114 @@ describe('V2 Wallet:', function () {
...
+    // using the same test for fanout and consolidate unspent API
```
OttoAllmendinger, line 2076: parameterize the tests to take the method name as an argument

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2070,4 +2071,114 @@ describe('V2 Wallet:', function () {
...
+        .post(`/api/v2/${wallet.coin()}/wallet/${wallet.id()}/tx/send`, { type: 'consolidate' })
+        .reply(200, {});
+
+      await wallet.consolidateUnspents(params);
+
+      const txInfo = stub.getCall(0)?.args?.[0]?.txPrebuild?.txInfo as any;
+      txInfo?.unspents?.[0].value.should.be.a.type('bigint');
+
+      scope.isDone().should.be.True();
+      sendScope.isDone().should.be.True();
```
OttoAllmendinger, line 2103:
```suggestion
      scope.done();
      sendScope.done();
```

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2070,4 +2071,114 @@ describe('V2 Wallet:', function () {
       });
     });
   });
+
+  describe('unspents APIs with string type minValue and maxValue', function () {
+    // using the same test for fanout and consolidate unspent API
+    it('consolidateUnspents should handle string type minValue and maxValue', async function () {
+      const params = { minValue: '123', maxValue: '9000000000000000000123' };
+
+      const scope = nock(bgUrl)
```
OttoAllmendinger, line 2080:
```suggestion
      const consolidateFanoutScope = nock(bgUrl)
```

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2070,4 +2071,114 @@ describe('V2 Wallet:', function () {
...
+      await wallet.consolidateUnspents(params);
+
+      const txInfo = stub.getCall(0)?.args?.[0]?.txPrebuild?.txInfo as any;
+      txInfo?.unspents?.[0].value.should.be.a.type('bigint');
+
+      scope.isDone().should.be.True();
+      sendScope.isDone().should.be.True();
+    });
+
+    it('maxspendable should handle string type minValue and maxValue', async function () {
```
OttoAllmendinger, line 2106:
```suggestion
    it('maximumSpendable should handle string type minValue and maxValue', async function () {
```

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2070,4 +2071,114 @@ describe('V2 Wallet:', function () {
...
+          sinon.assert.match(params.target, '9000000000000000000123');
+          return params;
+        })
+        .reply(200, {});
+
+      await wallet.unspents(params);
+
+      scope.isDone().should.be.True();
+    });
+    nock.cleanAll();
```
OttoAllmendinger, line 2141: This statement does not get executed after `it` is run. It needs to be in an `after()` block

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2070,4 +2071,114 @@ describe('V2 Wallet:', function () {
       });
     });
   });
+
+  describe('unspents APIs with string type minValue and maxValue', function () {
+    // using the same test for fanout and consolidate unspent API
+    it('consolidateUnspents should handle string type minValue and maxValue', async function () {
+      const params = { minValue: '123', maxValue: '9000000000000000000123' };
```
OttoAllmendinger, line 2078: How did you arrive at that number?

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2070,4 +2071,114 @@ describe('V2 Wallet:', function () {
...
+
+      const scope = nock(bgUrl)
+        .post(`/api/v2/${wallet.coin()}/wallet/${wallet.id()}/consolidateUnspents`, body => {
+          return typeof body.minValue === 'string' && body.minValue === '123'
+              && typeof body.maxValue === 'string' && body.maxValue === '9000000000000000000123';
+        })
```
OttoAllmendinger, line 2084: This is very repetitive. Please add a method that returns a nock.

OttoAllmendinger (changes requested):

modules/bitgo/test/v2/unit/wallet.ts
dpkjnr: I was just using very large number that doesn't fit into number. Changed it to 10000000000000123

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -618,6 +618,12 @@ export class Wallet implements IWallet {
       .send(filteredParams)
       .result();
 
+    response.txInfo.unspents = response.txInfo?.unspents?.map((u) => {
```
saravanan7mani, line 621: If `txInfo` or `unspents` are `undefined`, left hand side of the assignment can throw `TypeError`?

saravanan7mani (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
saravanan7mani: `unspents[].value` is used within `signTransaction`. 
Is it safe to pass `bigint` for other UTXO coins? Does utxo-lib take care of it?

dpkjnr: Good observation @saravanan7mani , it would fail if txInfo is undefined which is unlikely to be undefined. Changed it anyway

modules/bitgo/test/v2/unit/unspents.ts
```diff
@@ -0,0 +1,148 @@
...
+          .post(`/api/v2/${wallet.coin()}/wallet/${wallet.id()}/tx/send`, { type: manageUnspentType })
+          .reply(200, {});
+
+        await wallet[manageUnspentType + 'Unspents'](params);
+
+        const txInfo = signTransactionStub.getCall(0)?.args?.[0]?.txPrebuild?.txInfo as any;
+        txInfo?.unspents?.[0].value.should.be.a.type('bigint');
+
+        consolidateUnspentsScope.isDone();
+        sendScope.isDone();
```
OttoAllmendinger, line 66:
```suggestion
        consolidateUnspentsScope.done();
        sendScope.done();
```

same with other instances

modules/bitgo/test/v2/unit/unspents.ts
```diff
@@ -0,0 +1,148 @@
...
+      '5b3424f91bf349930e34017500000000',
+      '5b3424f91bf349930e34017600000000',
+      '5b3424f91bf349930e34017700000000',
+    ],
+    coinSpecific: {},
+  };
+  const wallet = new Wallet(bitgo, basecoin, walletData);
+  const bgUrl = common.Environments[bitgo.getEnv()].uri;
+  const minValueStringMatcher = sinon.match('123');
+  const maxValueStringMatcher = sinon.match.string.and(sinon.match('10000000000000123'));
```
OttoAllmendinger, line 24: it is unclear how and why you picked that number

this is clearer:
```
// cannot be represented by a JS Number
const highPrecisionBigInt = BigInt(1e16) + 1n;
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -618,6 +618,11 @@ export class Wallet implements IWallet {
       .send(filteredParams)
       .result();
 
+    response.txInfo?.unspents?.map((u) => {
```
saravanan7mani, line 621: `map` method doesn't change the `value` to `bigint` in the `response.txInfo.unspents` array right? It returns new `bigint` array.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

modules/bitgo/test/v2/unit/unspents.ts
OttoAllmendinger: I'm not quite sure what the purpose of these sinon matchers is. You are using them with `test()` as a boolean func in one place and using them in assertion style in another place. 

Instead you can just write a method like

```
function matchMinMaxValue({ minValue: string, maxValue: string }): boolean {
  ...
}
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -821,6 +826,10 @@ export class Wallet implements IWallet {
     ]);
     this.bitgo.setRequestTracer(reqId);
     const response = await this.bitgo.post(this.url('/sweepWallet')).send(filteredParams).result();
+    response.txInfo?.unspents?.map((u) => {
```
saravanan7mani, line 829: `map` method doesn't change the `value` to `bigint` in the `response.txInfo.unspents` array right? It returns new `bigint` array.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

modules/bitgo/test/v2/unit/wallet.ts
OttoAllmendinger: unresolved

modules/bitgo/test/v2/unit/wallet.ts
OttoAllmendinger: unresolved

OttoAllmendinger (changes requested):

saravanan7mani (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -618,6 +618,11 @@ export class Wallet implements IWallet {
       .send(filteredParams)
       .result();
 
+    response.txInfo?.unspents?.map((u) => {
+      u.value = BigInt(u.valueString);
+      return u;
+    });
```
OttoAllmendinger, line 624: We have to be very careful here. We don't know where `response.txInfo` ends up precisely (mind you this is the generic code path for all utxo coins). If this value gets mixed up with regular `number` values or with a HTTP POST call, there will be problems.

Better do a conversion where it is needed.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -821,6 +826,10 @@ export class Wallet implements IWallet {
     ]);
     this.bitgo.setRequestTracer(reqId);
     const response = await this.bitgo.post(this.url('/sweepWallet')).send(filteredParams).result();
+    response.txInfo?.unspents?.map((u) => {
+      u.value = BigInt(u.valueString);
+      return u;
+    });
```
OttoAllmendinger, line 832: same

OttoAllmendinger (changes requested):

modules/bitgo/test/v2/unit/wallet.ts
dpkjnr: oh! I didn't observe function name is also different. Fixed it

modules/bitgo/test/v2/unit/wallet.ts
dpkjnr: resolved by creating common function matchMinMaxValues

modules/sdk-core/src/bitgo/wallet/wallet.ts
dpkjnr: Moved conversion to signTransaction of abstractUtxoCoin and behaviour is coin specific now.

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -78,4 +80,16 @@ export class Doge extends AbstractUtxoCoin {
...
   }
+
+  parseUnspents<TNumber extends number | bigint>(unspents: Unspent<TNumber>[]): Unspent<TNumber>[] {
+    return unspents?.map((unspent) => {
+      const newUnspent: Unspent<TNumber> = Object.assign(
+        {},
+        unspent.id,
+        unspent.address,
+        utxolib.bitgo.toTNumber((unspent as any)?.valueString, 'bigint')
+      );
```
OttoAllmendinger, line 91: `Object.assign` will assign the _properties_ of the arguments, not the arguments themselves (it cannot know the property name)

in modern JS you can use the spread syntax anyway

```suggestion
      return { ...unspent, value: BigInt(unspent.valueString); }
```

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -78,4 +80,16 @@ export class Doge extends AbstractUtxoCoin {
   ): Promise<CrossChainRecoverySigned<TNumber> | CrossChainRecoveryUnsigned<TNumber>> {
     return super.recoverFromWrongChain(params);
   }
+
+  parseUnspents<TNumber extends number | bigint>(unspents: Unspent<TNumber>[]): Unspent<TNumber>[] {
```
OttoAllmendinger, line 84: The fact that you are casting `(unspent as any)` later is a hint that our types are incorrect. 

```suggestion

// should live in abstractUtxoCoin.ts
/** Unspent data as returned by Wallet Platform */
type UnspentJSON = Unspent & { value: number; valueString: string };

  parseUnspents(unspents: UnspentJSON[]): Unspent<bigint>[] {
```

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -1374,4 +1374,8 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   valuelessTransferAllowed(): boolean {
     return false;
   }
+
+  parseUnspents<TNumber extends number | bigint>(unspents: Unspent<TNumber>[]): Unspent<TNumber>[] {
```
OttoAllmendinger, line 1378: see comment about `UnspentJSON`

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -1107,7 +1107,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
 
     const signedTransaction = signAndVerifyWalletTransaction(
       transaction,
-      txPrebuild.txInfo.unspents,
+      this.parseUnspents(txPrebuild.txInfo.unspents),
```
OttoAllmendinger, line 1110: the `txPrebuild.txinfo.unspents` type should also use `UnspentJSON`

OttoAllmendinger (changes requested):

modules/sdk-coin-doge/src/doge.ts
dpkjnr: I have to return Unspent<number | bigint> from this method as it will be used by other coins too. And when I do `  parseUnspents<TNumber extends number | bigint>(unspents: UnspentJSON[]): Unspent<TNumber>[] {
    return unspents;
  }`

modules/sdk-coin-doge/src/doge.ts
dpkjnr: done

modules/sdk-core/src/bitgo/wallet/wallet.ts
dpkjnr: fixed it

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -78,4 +81,10 @@ export class Doge extends AbstractUtxoCoin {
   ): Promise<CrossChainRecoverySigned<TNumber> | CrossChainRecoveryUnsigned<TNumber>> {
     return super.recoverFromWrongChain(params);
   }
+
+  parseUnspents(unspents: UnspentJSON[]): Unspent<number | bigint>[] {
+    return unspents?.map((unspent) => {
```
OttoAllmendinger, line 86:
```suggestion
    return unspents.map((unspent) => {
```

OttoAllmendinger (changes requested):

PR #2788, opened 2022-09-20
fix(account-lib): fix EDDSA MPC key validation for small number keys

LeoKotschenreuther:
We were converting our public keys to Buffers but were cutting off
leading zeroes during the process (because we first converted the
hex strings to BigInts). This led to us passing a buffer with less
than 32 Bytes into the sodium library as a public key. The library
then threw an error which we hid by assuming that any error means
that the signature we are verifying is invalid.
The fix is to always pass in a Buffer with a length of 32 Bytes and
to fill it with zeroes if necessary.

Ticket: BG-50442

## Type of change

# How Has This Been Tested?

# Checklist:


mmcshinsky-bitgo (approved):

joey-bai (approved):

lcovar:
Thank @LeoKotschenreuther you beautiful man

PR #2795, opened 2022-09-20
Bg 53422 add tss param support

anshuldoshi:
## Description

Add isTss flag support when building with intents

## Issue Number
bg-53422

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -93,6 +93,7 @@ export interface PrebuildTransactionOptions {
   enableTokens?: TokenEnablement[];
   nonce?: string;
   preview?: boolean;
+  isTss?: boolean;
 }
 
 export interface PrebuildAndSignTransactionOptions extends PrebuildTransactionOptions, WalletSignTransactionOptions {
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -93,6 +93,7 @@ export interface PrebuildTransactionOptions {
   enableTokens?: TokenEnablement[];
   nonce?: string;
   preview?: boolean;
+  isTss?: boolean;
```
mmcshinsky-bitgo, line 96: Can you explain how this is looking to be used?

mmcshinsky-bitgo (changes requested):

modules/sdk-core/src/bitgo/wallet/iWallet.ts
anshuldoshi: We are looking to use this flag when building eth TSS transactions through prebuildTransaction. Eventually this is passed into prebuildTxWithIntent and populateIntent so that we can have a properly formed intent when calling wallet-platform.

mmcshinsky-bitgo, line 96: Let's make this an extension like `BuildTokenEnablementOptions` for use in `wallet.ts`.

mmcshinsky-bitgo (changes requested):

mmcshinsky-bitgo (approved):

PR #2797, opened 2022-09-21
feat(bitgo): add api version input

ipelaez-simtlix:
Add API version input to prebuildTransactionTss method

[BG-57753](https://bitgoinc.atlassian.net/browse/BG-57753)

This PR changes the default version of the API from lite to full. This causes the txRequest endpoint to fail when prebuilding transactions without sending the preview param true.

- To prebuild transactions is needed to use the preview param true.
- To send transactions needed to send the OTP to unlock the token.

### DIFF
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -453,11 +453,21 @@ export class Sol extends BaseCoin {
 
     await tssUtils!.deleteSignatureShares(txRequestId);
     const recreated = await tssUtils!.getTxRequest(txRequestId);
+    let txHex = '';
+    if (recreated.unsignedTxs) {
+      txHex = recreated.unsignedTxs[0]?.serializedTxHex;
+    } else {
+      txHex = recreated.transactions ? recreated.transactions[0]?.unsignedTx.serializedTxHex : '';
+    }
+
+    if (!txHex) {
+      throw new Error('Missing serialized tx hex');
+    }
 
     return Promise.resolve({
       ...params,
       txPrebuild: recreated,
-      txHex: recreated.unsignedTxs[0].serializedTxHex,
+      txHex,
     });
   }
 
--- a/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
@@ -2,7 +2,7 @@ import { SerializedKeyPair } from 'openpgp';
 import { IRequestTracer } from '../../../api';
 import { KeychainsTriplet } from '../../baseCoin';
 import { ApiKeyShare, Keychain } from '../../keychain';
-import { Memo, WalletType } from '../../wallet';
+import { ApiVersion, Memo, WalletType } from '../../wallet';
 import { EDDSA, GShare, SignShare, YShare } from '../../../account-lib/mpc/tss';
 import { KeyShare } from './ecdsa';
 
@@ -227,6 +227,7 @@ export type TSSParams = {
   txRequest: string | TxRequest; // can be either a string or TxRequest
   prv: string;
   reqId: IRequestTracer;
+  apiVersion?: ApiVersion;
 };
 
 export type TSSParamsForMessage = TSSParams & {
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -52,6 +52,8 @@ export interface BuildTokenEnablementOptions extends PrebuildTransactionOptions 
   enableTokens: TokenEnablement[];
 }
 
+export type ApiVersion = 'lite' | 'full';
+
 export interface PrebuildTransactionOptions {
   reqId?: IRequestTracer;
   recipients?: {
@@ -104,6 +106,7 @@ export interface PrebuildTransactionOptions {
   lowFeeTxid?: string;
   isTss?: boolean;
   custodianTransactionId?: string;
+  apiVersion?: ApiVersion;
 }
 
 export interface PrebuildAndSignTransactionOptions extends PrebuildTransactionOptions, WalletSignTransactionOptions {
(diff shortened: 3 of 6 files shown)


modules/sdk-coin-sol/src/sol.ts
```diff
@@ -453,11 +453,17 @@ export class Sol extends BaseCoin {
 
     await tssUtils!.deleteSignatureShares(txRequestId);
     const recreated = await tssUtils!.getTxRequest(txRequestId);
+    let txHex = '';
+    if (recreated.unsignedTxs) {
+      txHex = recreated.unsignedTxs[0]?.serializedTxHex;
+    } else {
+      txHex = recreated.transactions[0].unsignedTx.serializedTxHex;
+    }
```
joey-bai, line 461: we should throw an error if `txHex` is empty after this block

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2505,8 +2505,14 @@ export class Wallet implements IWallet {
...
-    const apiVersion =
-      this._wallet.type === 'custodial' || this.baseCoin.getMPCAlgorithm() === 'ecdsa' ? 'full' : 'lite';
+    if (
+      params.apiVersion === 'lite' &&
+      (this._wallet.type === 'custodial' || this.baseCoin.getMPCAlgorithm() === 'ecdsa')
+    ) {
+      throw new Error(`Custodial and ECDSA MPC algorithm must always use 'full' api version`);
+    }
+
+    const apiVersion = params.apiVersion || 'full';
```
joey-bai, line 2515: we should leave the default behavior the same, only overriding it with `params.apiVersion` if its explicitly provided.

e.g.
```
const apiVersion = params.apiVersion ||
 (this._wallet.type === 'custodial' || this.baseCoin.getMPCAlgorithm() === 'ecdsa' ? 'full' : 'lite');
```

joey-bai (commented):
main call outs are around backwards compatibility. changing default behavior from `lite` to `full` will break clients

modules/bitgo/test/v2/unit/internal/tssUtils/common.ts
```diff
@@ -8,15 +8,15 @@ export async function nockSendTxRequest(params: {coin:string, walletId: string,
 }
 
 export async function nockSendSignatureShare(params: { walletId: string, txRequestId: string, signatureShare: unknown, signerShare?: string, tssType?: 'eddsa' | 'ecdsa'}, status = 200): Promise<nock.Scope> {
-  const transactions = getRoute(params.tssType);
```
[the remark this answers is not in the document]
silalvado: Because this is the [getRoute](https://github.com/BitGo/BitGoJS/blob/master/modules/bitgo/test/v2/unit/internal/tssUtils/common.ts#L45) method for tests:

```
export function getRoute(tssType: 'eddsa' | 'ecdsa' = 'eddsa'): string {
  if (tssType === 'ecdsa') {
    return '/transactions/0';
  }

  return '';
}
```
It is not taking into account if API version is full or lite
This is the [sendSignatureShare method](https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/tss/common.ts#L41), if API version is full, '/transactions/0' should be added to the URL

modules/bitgo/test/v2/unit/internal/tssUtils/common.ts
```diff
@@ -8,15 +8,15 @@ export async function nockSendTxRequest(params: {coin:string, walletId: string,
...
+  const transactions = '/transactions/0';
   return nock('https://bitgo.fakeurl')
     .persist(true)
     .post(`/api/v2/wallet/${params.walletId}/txrequests/${params.txRequestId + transactions}/signatureshares`)
     .reply(status, (status === 200 ? params.signatureShare : { error: 'some error' }));
 }
 
 export async function nockSendSignatureShareWithResponse(params: { walletId: string, txRequestId: string, signatureShare: unknown, response: unknown, tssType?: 'eddsa' | 'ecdsa'}, status = 200): Promise<nock.Scope> {
-  const transactions = getRoute(params.tssType);
+  const transactions = '/transactions/0';
```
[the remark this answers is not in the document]
silalvado: removed

modules/bitgo/test/v2/unit/tss/helpers.ts
```diff
@@ -1,10 +1,9 @@
 import { SignatureShareRecord } from '@bitgo/sdk-core';
-import { getRoute } from '../internal/tssUtils/common';
 import * as nock from 'nock';
 
 export async function nockSendSignatureShare(params: { walletId: string, txRequestId: string, signatureShare: any, signerShare?: string, response?: SignatureShareRecord, tssType?: 'ecdsa' | 'eddsa'}, status = 200): Promise<nock.Scope> {
-  const { signatureShare, signerShare, tssType } = params;
-  const transactions = getRoute(tssType);
+  const { signatureShare, signerShare } = params;
+  const transactions = '/transactions/0';
```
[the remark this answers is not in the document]
silalvado: removed

joey-bai (approved):

mmcshinsky-bitgo (dismissed):

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -227,6 +227,7 @@ export type TSSParams = {
   txRequest: string | TxRequest; // can be either a string or TxRequest
   prv: string;
   reqId: IRequestTracer;
+  apiVersion?: 'full' | 'lite';
```
joey-bai, line 230: can we have this reference `ApiVersion`?

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -52,6 +52,8 @@ export interface BuildTokenEnablementOptions extends PrebuildTransactionOptions
   enableTokens: TokenEnablement[];
 }
 
+export type ApiVersion = 'lite' | 'full' | undefined;
```
joey-bai, line 55: i think the extra `| undefined` is redundant.

you can declare a property as being optional using something like:
```
apiVersion?: ApiVersion
```

joey-bai (dismissed):

joey-bai (approved):

mmcshinsky-bitgo (approved):

PR #2798, opened 2022-09-21
fix(utxo-lib): move getValueScaled->test + rename

brandonblack:
* Found this function while working on PSBT support for bitcoin derived coins. It's unsafe, and not appropriate for production use. Also more general than necessary. Moved to testutils, and renamed for its specific purpose.

Issue: BG-47820

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/utxo-lib/test/testutil/tnumber.ts
+++ b/modules/utxo-lib/test/testutil/tnumber.ts
@@ -1,0 +1,29 @@
+/**
+ * Multiply a decimal amount by 1e8 to convert from coins to sats
+ * (optionally to a bigint converting to bigint)
+ *
+ * This function really shouldn't exist, but is used by some test code. At some
+ * point we should fix those functions to use strings or integers only to
+ * represent monetary values.
+ *
+ * Throws error if resulting value is not a safe integer number
+ *
+ * @param value - decimal amount of coins
+ * @param amountType - desired output type
+ * @return value * 1e8, as amountType
+ */
+export function decimalCoinsToSats<TNumber extends number | bigint>(
+  value: number,
+  amountType: 'number' | 'bigint' = 'number'
+): TNumber {
+  if (amountType === 'number') {
+    const scaledValue = value * 1e8;
+    if (!Number.isSafeInteger(scaledValue)) {
+      throw new Error('input value cannot be scaled to safe integer number');
+    }
+    return scaledValue as TNumber;
+  } else {
+    const [integerString, decimalString] = value.toFixed(8).split('.');
+    return (BigInt(integerString) * BigInt(1e8) + BigInt(decimalString)) as TNumber;
+  }
+}
--- a/modules/utxo-lib/src/bitgo/tnumber.ts
+++ b/modules/utxo-lib/src/bitgo/tnumber.ts
@@ -24,31 +24,3 @@ export function toTNumber<TNumber extends number | bigint>(
   }
   throw new Error('amountType must be either "number" or "bigint"');
 }
-
-/**
- * Multiply a decimal amount (optionally to a bigint)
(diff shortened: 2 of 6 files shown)


modules/utxo-lib/test/testutil/tnumber.ts
```diff
@@ -0,0 +1,22 @@
...
+): TNumber {
+  if (amountType === 'number') {
+    const scaledValue = value * 1e8;
+    if (!Number.isSafeInteger(scaledValue)) {
+      throw new Error('input value cannot be scaled to safe integer number');
+    }
+    return scaledValue as TNumber;
+  } else {
+    const [integerString, decimalString] = value.toFixed(8).split('.');
+    return (BigInt(integerString) * BigInt(1e8) + BigInt(decimalString)) as TNumber;
```
sangaman, line 20: It looks like this cast isn't needed? To be sure I checked out the branch and built after removing this and it worked

modules/utxo-lib/test/testutil/tnumber.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Multiply a decimal amount (optionally to a bigint)
+ * Throws error if resulting value is not a safe integer number
```
sangaman, line 3: This can be a little more explicit, it's not clear what we're multiplying by. How about "Multiply a decimal amount of coins by 1e8 to get an integer number of satoshis" or similar?

sangaman (commented):
`decimalToSats` is definitely a better name but still a bit confusing. Maybe `coinsToSats`? That might be confusing too tho so I'll leave it up to you.

Also, what is unsafe about this method? Or was it just unsafe the way it was implemented previously?

modules/utxo-lib/test/testutil/tnumber.ts
sangaman: Same for the cast on line 17.

modules/utxo-lib/test/testutil/tnumber.ts
brandonblack: > Also, what is unsafe about this method? Or was it just unsafe the way it was implemented previously?

Precise decimal values are commonly not representable exactly in floating point, and should _never_ be used to represent monetary values of any. https://www.exploringbinary.com/why-0-point-1-does-not-exist-in-floating-point/

This type of function papers over that type of error (either the original implementation or my simplified one) by rounding, but should really never ever ever be used in production.

modules/utxo-lib/test/testutil/tnumber.ts
brandonblack: This is test code, build will work, but tests will fail without the casts.

brandonblack:
> Also, what is unsafe about this method? Or was it just unsafe the way it was implemented previously?

https://www.exploringbinary.com/why-0-point-1-does-not-exist-in-floating-point/

Floating point numbers cannot represent precise decimal amounts. They should never be used for monetary values. Ever. Never.

modules/utxo-lib/test/testutil/tnumber.ts
sangaman: Really? The casts don't affect runtime, so execution should be the same. Are the tests using a different tsconfig or something such that ts-node is failing?

sangaman (approved):

sangaman:
Btw I like `decimalCoinsToSats` as a function name.

modules/utxo-lib/test/testutil/tnumber.ts
brandonblack: The tests are run as typescript, so aren't compiled until they are executed.

tylerlevine (approved):

PR #2802, opened 2022-09-21
'fix: change ethw family to eth and fix signTransaction for ethw for BG-57910'

benpreiser:
## Description

Fix ethw functionality so that OVC can sign a transaction for unsigned sweep from WRW.

## Issue Number

BG-57910

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -33,7 +33,7 @@ export enum CoinFamily {
   DOT = 'dot',
   ETH = 'eth',
   ETH2 = 'eth2',
-  ETHW = 'ethw',
+  ETHW = 'eth',
   ETC = 'etc',
   EOS = 'eos',
   FIAT = 'fiat',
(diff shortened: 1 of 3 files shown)


modules/statics/src/base.ts
```diff
@@ -33,7 +33,7 @@ export enum CoinFamily {
   DOT = 'dot',
   ETH = 'eth',
   ETH2 = 'eth2',
-  ETHW = 'ethw',
+  ETHW = 'eth',
```
bitgoAaron, line 36: if `ETHW` is indeed meant to just be part of the Ethereum CoinFamily, then this enum should be deleted entirely & the account ctor should be updated accordingly.

fperalta-simtlix: We should keep this as `ethw`. It is an eth fork like `etc`, then they can evolve separately so we consider them different families.
@benpreiser , as you pointed we have a mismatch in `getFamily` implementation of ethw. It should return `ethw`.

fperalta-simtlix (approved):

mmcshinsky-bitgo (approved):

PR #2807, opened 2022-09-22
Support for parsing and signing PSBTs for supported coins

brandonblack:
## Issue Number

BG-47820, BG-47821, BG-55846, BG-55845

## Type of change

# How Has This Been Tested?

Existing tests have been updated to cover the new code.

# Checklist:

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -1,12 +1,18 @@
 import { UtxoTransaction } from './UtxoTransaction';
-import { taproot, HDSigner, Psbt, PsbtTransaction, Transaction } from '../';
-import { TapBip32Derivation, Transaction as ITransaction, TransactionFromBuffer } from 'bip174/src/lib/interfaces';
+import { taproot, HDSigner, Psbt, PsbtTransaction, ScriptSignature, Stack, Transaction, TxOutput } from '../';
+import {
+  PartialSig,
+  PsbtInputUpdate,
+  TapBip32Derivation,
+  Transaction as ITransaction,
+  TransactionFromBuffer,
+} from 'bip174/src/lib/interfaces';
 import { checkForInput } from 'bip174/src/lib/utils';
 import { Psbt as PsbtBase } from 'bip174';
 import { Network } from '..';
-import { ecc as eccLib, script as bscript } from '..';
+import { ecc as eccLib, script as bscript, payments } from '..';
 import * as opcodes from 'bitcoin-ops';
-import { BufferWriter, varuint } from 'bitcoinjs-lib/src/bufferutils';
+import { BufferWriter, varuint, reverseBuffer } from 'bitcoinjs-lib/src/bufferutils';
 
 export interface HDTaprootSigner extends HDSigner {
   /**
@@ -58,6 +64,53 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
     const psbtBase = PsbtBase.fromBuffer(buffer, transactionFromBuffer);
     const psbt = this.createPsbt(opts, psbtBase);
     // Upstream checks for duplicate inputs here, but it seems to be of dubious value.
+    return psbt;
+  }
+
+  static async fromTransactionComplete(
+    transaction: UtxoTransaction<bigint>,
+    prevOutputs: TxOutput<bigint>[],
+    fetchTransactions: (txids: string[]) => Promise<Record<string, Buffer>>
+  ): Promise<UtxoPsbt<UtxoTransaction<bigint>>> {
+    const psbt = this.fromTransaction(transaction, prevOutputs);
+
+    const txidToIndex: Record<string, number> = {};
+    psbt.data.inputs.forEach((input, index) => {
+      if (!input.tapLeafScript && !input.witnessScript) {
+        txidToIndex[reverseBuffer(transaction.ins[index].hash).toString('hex')] = index;
+      }
+    });
+
+    const txHexs = await fetchTransactions(Object.keys(txidToIndex));
+    Object.entries(txidToIndex).forEach(([txid, index]) => {
+      psbt.updateInput(index, { nonWitnessUtxo: txHexs[txid] });
+    });
+
+    return psbt;
+  }
+
+  static fromTransaction(
+    transaction: UtxoTransaction<bigint>,
+    prevOutputs: TxOutput<bigint>[]
+  ): UtxoPsbt<UtxoTransaction<bigint>> {
+    if (prevOutputs.length !== transaction.ins.length) {
+      throw new Error(
+        `Transaction has ${transaction.ins.length} inputs, but ${prevOutputs.length} previous outputs provided`
+      );
+    }
+    const clonedTransaction = transaction.clone();
+    const updates = unsign(clonedTransaction, prevOutputs);
+
+    const psbtBase = new PsbtBase(new PsbtTransaction({ tx: clonedTransaction }));
+    clonedTransaction.ins.forEach(() => psbtBase.inputs.push({ unknownKeyVals: [] }));
+    clonedTransaction.outs.forEach(() => psbtBase.outputs.push({ unknownKeyVals: [] }));
+    const psbt = this.createPsbt({ network: transaction.network }, psbtBase);
+
+    updates.forEach((update, index) => {
+      psbt.updateInput(index, update);
+      psbt.updateInput(index, { witnessUtxo: prevOutputs[index] });
+    });
+
     return psbt;
   }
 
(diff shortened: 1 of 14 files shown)


modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -35,6 +41,9 @@ export interface PsbtOpts {
   network: Network;
 }
 
+// TODO: upstream does `checkInputsForPartialSigs` before doing things like
+// `setVersion`. Our inputs could have tapscriptsigs (or in future tapkeysigs)
+// and not fail that check. Do we want to do anything about that?
```
sangaman, line 46: What happens if the inputs have partial signatures in upstream?

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -317,3 +376,145 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+ *  * P2SH -> P2WPKH
+ *  * P2TR script path 2-of-2
+ *
+ * @param tx the partially signed transaction
+ *
+ * @return the removed scripts and signatures, ready to be added to a PSBT
+ */
+export function unsign(tx: UtxoTransaction<bigint>, prevOuts: TxOutput<bigint>[]): PsbtInputUpdate[] {
+  const ret = tx.ins.map((input, vin) => {
+    const prevOut = prevOuts[vin];
```
sangaman, line 398: I think we need to have some sort of check here that verifies that `prevouts.length` is equal to `tx.ins.length` otherwise this could be undefined, and we're accessing it on line 423 without any sort of undefined check.

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -317,3 +376,145 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+    const update: PsbtInputUpdate = {};
+    let redeemScript;
+    if (input.script && input.script.length) {
+      const decompiledScriptSig = bscript.decompile(input.script);
+      if (!decompiledScriptSig) {
+        throw new Error('Invalid scriptSig, failed to decompile');
+      }
+      if ((!input.witness || !input.witness.length) && decompiledScriptSig.length === 2) {
+        // P2PKH
```
sangaman, line 407: Should this be dropped since we don't support P2PKH on bitgo wallets (ditto for P2WPKH)? Although it probably doesn't hurt to have just for completeness sake since it's already implemented.

As an aside, do we need to cover P2SH-P2PK here since it's what we use for replay protection inputs?

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -317,3 +376,145 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+      update.redeemScript = redeemScript;
+      if (opcodes.OP_0 !== redeemScript[0] || ![22, 34].includes(redeemScript.length)) {
+        // P2SH -> 2-of-3
+        const hashFn = (hashType) => tx.hashForSignature(vin, redeemScript, hashType, prevOut.value);
+        update.partialSig = matchSignatures(hashFn, redeemScript, decompiledScriptSig);
+        return update;
+      }
+    }
+    if (input.witness && input.witness.length) {
+      const witness = [...input.witness];
```
sangaman, line 429: I believe this clones `input.witness`, in which case when we do `witness.pop()` it's not actually removing anything from `input.witness` if I follow correctly. But don't we want to remove the scripts and sigs from the tx within this method?

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -317,3 +376,145 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+  for (const sig of stack.slice(1)) {
+    // Ignore extra empty element re MULTISIG bug
+    if (!Buffer.isBuffer(sig) || sig.length === 0) continue;
+    const { signature, hashType } = ScriptSignature.decode(sig);
+    const hash = hashFn(hashType);
+    let signatureMatched = false;
+    for (const pubkey of publicKeys) {
+      if (!eccLib.verify(hash, pubkey, signature)) continue;
+      signatureMatched = true;
+      partialSig.push({ pubkey, signature: sig });
```
sangaman, line 513: You could `break` after this line so you don't bother checking the other public keys once you've found a match.

sangaman, line 513: Fwiw to me this is slightly easier to follow without the `continue` statement and the negated `if`, e.g
```suggestion
    for (const pubkey of publicKeys) {
      if (eccLib.verify(hash, pubkey, signature)) {
        signatureMatched = true;
        partialSig.push({ pubkey, signature: sig });
        break;
      }
```

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -0,0 +1,73 @@
...
+    this.tx.versionGroupId = getDefaultVersionGroupIdForVersion(version);
+    this.tx.consensusBranchId = getDefaultConsensusBranchIdForVersion(network, version);
+  }
+
+  private setPropertyCheckSignatures(propName: keyof ZcashTransaction<bigint>, value: unknown) {
+    if (this.tx[propName] === value) {
+      return;
+    }
+    this.checkForSignatures(propName);
+    this.tx[propName] = value as any;
```
sangaman, line 56: Do you know why we declare `value` as an `unknown` and then just cast it to any without checking its type at all? It seems to circumvent the purpose of `unknown` and we might as well just declare `value: any` up above.

I see we were already doing this in ZcashTransactionBuilder but it seems odd to me.

sangaman (commented):
Tried my best to follow all the changes and left some questions and suggestions.

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
brandonblack: It prevents modifying the PSBT if it would invalidate any sigs.

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
brandonblack: Yeah, these are useful in testing, which is why they're here.

We have a ticket open to support P2SH->P2PK. Thanks!

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
brandonblack: TBH, I forget why I ended up cloning the witness here, but I know it ended up being cleaner, and I changed to clearing the witness later: https://github.com/BitGo/BitGoJS/pull/2807/files#diff-e26e7163241b590c68c95bdc501223630eb5d85391772dcfe31ccf1bc11a9589R492

I'll try to brain spelunk and add a comment.

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
brandonblack: Yeah, actually just made that same change in response to your above comment :).

modules/utxo-lib/src/bitgo/UtxoTransaction.ts
```diff
@@ -15,13 +17,30 @@ export class UtxoTransaction<TNumber extends number | bigint = number> extends b
 
   constructor(
     public network: Network,
-    transaction: bitcoinjs.Transaction<TNumber> = new bitcoinjs.Transaction<TNumber>()
+    transaction?: bitcoinjs.Transaction<bigint | number>,
+    amountType?: 'bigint' | 'number'
   ) {
     super();
```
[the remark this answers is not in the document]
brandonblack: Super is `bitcoinjs.Transaction`, so this really ends up being equivalent to the old code, just skipping separately instantiating an instance of bitcoinjs.Transaction and copying from it.

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
brandonblack: I don't know why. Yeah, we just copied this over. Might kill it later as it's less useful in PSBT than TxBuilder.

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
brandonblack: This turned out to be concealing a bug in `UtxoTransaction.clone`. Unbugged, and uncloned here.

sangaman (approved):

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
tylerlevine: yeah, could probably type this better as `typeof keyof ZcashTransaction<bigint>`, but I think killing this is the right thing to do long term anyway

tylerlevine (approved):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -58,6 +67,53 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
   }
 
+  static async fromTransactionComplete(
+    transaction: UtxoTransaction<bigint>,
+    prevOutputs: TxOutput<bigint>[],
+    fetchTransactions: (txids: string[]) => Promise<Record<string, Buffer>>
+  ): Promise<UtxoPsbt<UtxoTransaction<bigint>>> {
+    const psbt = this.fromTransaction(transaction, prevOutputs);
+
+    const txidToIndex: Record<string, number> = {};
```
OttoAllmendinger, line 77: we should prefer `Map` over `Record` in cases like these

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -58,6 +67,53 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+    psbt.data.inputs.forEach((input, index) => {
+      if (!input.tapLeafScript && !input.witnessScript) {
+        txidToIndex[reverseBuffer(transaction.ins[index].hash).toString('hex')] = index;
+      }
+    });
```
OttoAllmendinger, line 82: it appears that this will not handle multiple inputs from the same transaction correctly (only the last one will be considered)

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -58,6 +67,53 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
     return psbt;
   }
 
+  static async fromTransactionComplete(
```
OttoAllmendinger, line 70: the docstring should probably mention the p2sh case here

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -317,3 +381,147 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+    return update;
+  });
+  tx.ins.forEach((input) => {
+    input.witness = [];
+    input.script = Buffer.alloc(0);
+  });
+  return ret;
+}
+
+function matchSignatures(hashFn: (hashType: number) => Buffer, script: Buffer, stack: Stack): PartialSig[] {
```
OttoAllmendinger, line 503: jsdoc would be useful here

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  setConsensusBranchId(consensusBranchId: number): void {
```
OttoAllmendinger, line 59: y no fluent interface?

PR #2817, opened 2022-09-26
refactor(sdk-coin): refactor common methods in TSS coins

alanliu-bitgo:
## Description

Refactor out common code amongst TSS coins into sdk-core tss directory. Updated Near and Polkadot to using the factored out methods.

## Issue Number
BG-58214

# How Has This Been Tested?

All unit tests passed. Dot recover function was tested again and transaction successfully broadcasted on Westend testnet.


modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
```diff
@@ -291,3 +300,70 @@ export async function encryptYShare(params: {
     encryptedPrivateShare,
   };
 }
+
+export async function getInitializedMpcInstance() {
```
joey-bai, line 304: add return type: `Promise<Eddsa>`.

also add doc:
```
Initializes Eddsa instance
```

joey-bai (approved):

byhow (approved):

PR #2819, opened 2022-09-26
feat(psbt): support fully verifiable change outputs

brandonblack:
Issues: BG-47821, BG-54756

### DIFF
--- a/modules/utxo-lib/src/bitgo/Unspent.ts
+++ b/modules/utxo-lib/src/bitgo/Unspent.ts
@@ -1,11 +1,11 @@
-import { TxOutput } from 'bitcoinjs-lib';
+import { TxOutput, taproot } from 'bitcoinjs-lib';
 import { Network } from '..';
 import { toOutputScript } from '../address';
-import { createOutputScript2of3, createSpendScriptP2tr } from './outputScripts';
+import { createOutputScript2of3, createSpendScriptP2tr, createPaymentP2tr, toXOnlyPublicKey } from './outputScripts';
 import { UtxoPsbt } from './UtxoPsbt';
 import { UtxoTransaction } from './UtxoTransaction';
 import { UtxoTransactionBuilder } from './UtxoTransactionBuilder';
-import { isSegwit, RootWalletKeys, scriptTypeForChain, WalletUnspent, WalletUnspentSigner } from './wallet';
+import { isSegwit, ChainCode, RootWalletKeys, scriptTypeForChain, WalletUnspent, WalletUnspentSigner } from './wallet';
 
 /**
  * Public unspent data in BitGo-specific representation.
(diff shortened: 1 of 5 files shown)


davidkaplanbitgo (dismissed):
lgtm!

modules/utxo-lib/src/bitgo/Unspent.ts
```diff
@@ -169,7 +224,7 @@ export function addToPsbt(
   } else {
     const { witnessScript, redeemScript } = createOutputScript2of3(walletKeys.publicKeys, scriptType);
     psbt.updateInput(inputIndex, {
-      bip32Derivation: [signerIndex, cosignerIndex].map((idx) => ({
+      bip32Derivation: [0, 1, 2].map((idx) => ({
```
sangaman, line 227: Why do we need all 3 derivation paths here?

modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
```diff
@@ -122,17 +123,23 @@ describe('WalletUnspent', function () {
...
-    psbt.addOutput({
-      address: getWalletAddress(walletKeys, 0, 100, network),
-      value: BigInt(unspentSum<bigint>(unspents, 'bigint')) - BigInt(100),
-    });
+    addChangeOutputToPsbt(
+      psbt,
+      walletKeys,
+      getInternalChainCode(scriptType),
+      100,
+      BigInt(unspentSum<bigint>(unspents, 'bigint')) - BigInt(100)
```
sangaman, line 135: To be clear, here `100` is both the index used for the derivation path and also the size of the non-change output used for this test? It might be worth naming these variables so it's easier to understand how they're used.

modules/utxo-lib/src/bitgo/Unspent.ts
```diff
@@ -124,6 +124,61 @@ export function addToTransactionBuilder<TNumber extends number | bigint>(
...
 }
 
+export function addChangeOutputToPsbt(
+  psbt: UtxoPsbt<UtxoTransaction<bigint>>,
+  rootWalletKeys: RootWalletKeys,
+  chain: ChainCode,
+  index: number,
+  value: bigint
+): void {
+  const walletKeys = rootWalletKeys.deriveForChainAndIndex(chain, index);
```
sangaman, line 134: We're allowing both internal and external chain codes here iiuc. Do we want to require the chaincode be internal, as I would expect for change outputs? Or alternatively we can accept a `scriptType` and then do `getInternalChainCode()` within this function.

modules/utxo-lib/src/bitgo/Unspent.ts
brandonblack: So that a "standard" PSBT signer that has the 3 xpubs saved can fully validate the script's pubkeys. Doesn't _really_ matter for the input side, but could still be valuable information if there is some funny business going on.

modules/utxo-lib/src/bitgo/Unspent.ts
brandonblack: Good call.

modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
brandonblack: Yeah, duplicated from previous code, but worth improving while we're here.

sangaman: Yep, I saw the old code was doing the same thing. But just a bit confusing when `100` is used for multiple things.

modules/utxo-lib/src/bitgo/Unspent.ts
brandonblack: Nix that. The chain is an API parameter in WP, so we need to accept chain on the addChangeOutput. I'll add a comment.

sangaman: Gotcha, maybe from WP we could have a check that the chain is internal?

sangaman (approved):

tylerlevine (approved):

PR #2821, opened 2022-09-27
feat(sdk-coin-sol): Unsigned Sweep Recovery

colin-y-wu:
Ticket: BG-57227

## Description

Implement unsigned sweep flow for Solana. Unsigned sweep flow is differentiated from a non-bitgo flow by the missing user key, backup key, and wallet passphrase from the frontend, and will return a serialized **unsigned** transaction hex string.

Unsigned sweep will always require a durable nonce account, and sign transactions with the durable nonce account.

## Issue Number

BG-57227

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Unit tested, verified that we receive a transaction hex for unsigned transactions

### DIFF
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -114,7 +114,7 @@ interface RecoveryOptions {
   backupKey?: string; // Box B
   bitgoKey: string; // Box C - this is bitgo's xpub and will be used to derive their root address
   recoveryDestination: string; // base58 address
-  walletPassphrase: string;
+  walletPassphrase?: string;
   durableNoncePK?: string;
   durableNonceSK?: string;
 }
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -110,8 +110,8 @@ interface SolDurableNonceFromNode {
 }
 
 interface RecoveryOptions {
-  userKey: string; // Box A
-  backupKey: string; // Box B
+  userKey?: string; // Box A
+  backupKey?: string; // Box B
   bitgoKey: string; // Box C - this is bitgo's xpub and will be used to derive their root address
   recoveryDestination: string; // base58 address
   walletPassphrase: string;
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-sol/test/unit/sol.ts
```diff
@@ -1482,5 +1497,27 @@ describe('SOL:', function () {
...
     });
+
+    it('should recover a txn for unsigned sweep recoveries', async function () {
+      // Unsigned Sweep Recovery
+      const unsignedSweepTxn = await basecoin.recover({
+        bitgoKey: testData.keys.bitgoKey,
+        recoveryDestination: testData.keys.destinationPubKey,
+        walletPassphrase: testData.keys.walletPassword,
+        durableNoncePK: testData.keys.durableNoncePubKey,
+        durableNonceSK: testData.keys.durableNoncePrivKey,
```
michael-gallegos-bg, line 1508: Since we're dealing with a few optional properties, it would be good to add some tests with edge cases (unexpected input, or undefined inputs) to ensure the overall method fails well.

michael-gallegos-bg (approved):

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -557,58 +557,80 @@ export class Sol extends BaseCoin {
...
     }
 
     if (!params.recoveryDestination || !this.isValidAddress(params.recoveryDestination)) {
       throw new Error('invalid recoveryDestination');
     }
 
     // Clean up whitespace from entered values
-    const userKey = params.userKey.replace(/\s/g, '');
-    const backupKey = params.backupKey.replace(/\s/g, '');
+    if (params.userKey && params.backupKey) {
```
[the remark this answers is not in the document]
colin-y-wu: should throw error in this case, will add a unit test to ensure that this is what happens

modules/sdk-coin-sol/test/unit/sol.ts
```diff
@@ -1482,5 +1497,80 @@ describe('SOL:', function () {
...
+      const unsignedSweepTxnJson = unsignedSweepTxnDeserialize.toJson();
+
+      should.equal(unsignedSweepTxnJson.nonce, testData.SolInputData.durableNonceBlockhash);
+      should.equal(unsignedSweepTxnJson.feePayer, testData.SolInputData.pubKey);
+      should.equal(unsignedSweepTxnJson.numSignatures, testData.SolInputData.unsignedSweepSignatures);
+    });
+
+    it('should handle error in recover function if a required field is missing/incorrect', async function () {
+      // missing userkey
+      await basecoin
```
michael-gallegos-bg, line 1525: We could potentially break these up into separate unit tests, but it looks good for now!

michael-gallegos-bg (approved):

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -110,11 +110,11 @@ interface SolDurableNonceFromNode {
...
-  userKey: string; // Box A
-  backupKey: string; // Box B
+  userKey?: string; // Box A
+  backupKey?: string; // Box B
   bitgoKey: string; // Box C - this is bitgo's xpub and will be used to derive their root address
   recoveryDestination: string; // base58 address
-  walletPassphrase: string;
+  walletPassphrase?: string;
   durableNoncePK?: string;
   durableNonceSK?: string;
```
joey-bai, line 119: just caught this, but since these two parameters need to be present together, can we group them in a single object?
i.e.
```
interface RecoveryOptions {
    ...
    durableNonce?: {
        publicKey: string;
        secretKey: string;
    }
}
```

this way, you just need to check for the existence of `durableNonce`. if its present, then so should `durableNonce.publicKey` and `durableNonce.secretKey`

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -557,58 +557,80 @@ export class Sol extends BaseCoin {
...
-      throw new Error('missing userKey');
+      if (!params.durableNoncePK || !params.durableNonceSK) {
+        throw new Error('missing nonce account for unsigned sweep');
+      }
     }
 
-    if (!params.backupKey) {
-      throw new Error('missing backupKey');
-    }
+    if (!isUnsignedSweep) {
```
joey-bai, line 575: You have this `isUnsignedSweep` check in multiple places, which makes it pretty difficult to follow the code.
you can shrink this down to a single check if you move it at the very end

An outline of your code should look something like:

1. derive Solana address from `params.bitgoKey`
2. build unsigned transaction; this is where you need to calculate transfer amount and figure out if you need to get a recent blockhash or one from durable nonce
3. finally, if this isn't for unsigned sweep, generate TSS signature with user & backup key and attach to transaction
4. return transaction in broadcast format

joey-bai (approved):

byhow (approved):

PR #2833, opened 2022-09-29
chore(root): add steps to automate publishing @bitgo-beta

bitgoAaron:
- adds script to rename packages to @bitgo-beta scope
- adds github action to generate a beta on demand. Will wire up to master once we confirm this works with secrets

Ticket: BG-57161

## Description
- Once the secrets are configured in GitHub we can then publish beta builds to https://www.npmjs.com/search?q=%40bitgo-beta

### DIFF
--- a/.github/workflows/beta-publish.yml
+++ b/.github/workflows/beta-publish.yml
@@ -1,0 +1,38 @@
+name: Publish @bitgo-beta
+on:
+  workflow_dispatch:
+
+jobs:
+  publish:
+    name: Publish Beta Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
+
+      - name: Install BitGoJS
+        run: yarn install --with-frozen-lockfile
+
+      - name: Configure NPM
+        run: |
+          echo "email=${{ secrets.BETA_EMAIL }}" > .npmrc
+          echo "@bitgo-beta:registry=https://registry.npmjs.org" >> .npmrc
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.BETA_TOKEN }}" >> .npmrc
+          echo "//registry.npmjs.org/:always-auth=true" >> .npmrc
+          
+      - name: Prepare Beta
+        run: |
+          rm -rfd ./modules/web-demo
+          rm -rfd ./modules/express
+          npx ts-node ./scripts/prepare-beta.ts
+      
+      - name: Rebuild Beta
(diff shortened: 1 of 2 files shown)


scripts/prepare-beta.ts
```diff
@@ -0,0 +1,97 @@
...
+            ...walk(file),
+          ];
+        }
+      } else if (['.ts', '.tsx', '.js', '.json'].includes(path.extname(file))) {
+        // Is a file
+        results.push(file);
+      }
+    });
+  } catch (error) {
+    console.error(`Error when walking dir ${dir}`, error);
```
louib-bitgo, line 53: @bitgoAaron is it safe to fail silently here? Since the releases will be automated, I feel like we should abort, unless errors are expected when walking the directories.

scripts/prepare-beta.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+  return results;
+};
+
+const edit = (filePath:string): void => {
```
louib-bitgo, line 58: @bitgoAaron `changeScopeInFile`?

.github/workflows/beta-publish.yml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  publish:
+    name: Publish Beta Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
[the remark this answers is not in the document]
bitgoAaron: eh, the checkout step doesn't really matter since we're explicitly setting the node-version to 14 on the following step.

scripts/prepare-beta.ts
bitgoAaron: yeah, good call. I'll still log the error so we know which directory/file was problematic and throw the exception instead.

.github/workflows/beta-publish.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Prepare Beta
+        run: |
+          rm -rfd ./modules/web-demo
+          rm -rfd ./modules/express
+          npx ts-node ./scripts/prepare-beta.ts
+      
+      - name: Rebuild Beta
+        run: yarn
+
+      - name: Publish Canary
```
louib-bitgo, line 37: @bitgoAaron The `dist-tag` is `beta` now, so the step's name should be updated. Not a blocker for this PR.

louib-bitgo (approved):

.github/workflows/beta-publish.yml
bitgoAaron: ah. Ok, will rename it on the next PR when we hook this up to trigger on pushes to the master branch

PR #2834, opened 2022-09-29
feat(utxo-lib): add consensusBranchId in ZcashPsbt for version 4 transactions

davidkaplanbitgo:
## Description

The Zcash `consensusBranchId` is a parameter that specifies the rules for consensus for each network upgrade. For versions below 5 the `consensusBranchId` is not serialized on the transaction (see [`toBufferV4`](https://github.com/BitGo/BitGoJS/blob/master/modules/utxo-lib/src/bitgo/zcash/ZcashBufferutils.ts#L155)). Because of this we need to add this parameter in the `global-map` of the `ZcashPsbt`. If it is a version 5 Zcash transaction, we do not need to add it to the `global-map` of the PSBT because it is already serialized on the `ZcashTransaction`.

With this change, the only data we need to send to the HSM for a Zcash transaction is the serialized PSBT - all of the information for signing will be contained on the PSBT.

## Issue Number

[BG-55848](https://bitgoinc.atlassian.net/browse/BG-55848)

## Type of change

# How Has This Been Tested?

For testing, a 

# Checklist:

### DIFF
--- a/modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
@@ -7,8 +7,18 @@ import {
 import { Network, PsbtTransaction } from '../../';
 import { Psbt as PsbtBase } from 'bip174';
 import * as types from 'bitcoinjs-lib/src/types';
-
+import { UtxoTransaction } from '../UtxoTransaction';
 const typeforce = require('typeforce');
+
+enum Subtype {
+  ZEC_CONSENSUS_BRANCH_ID = 0x00,
+}
+const CONSENSUS_BRANCH_ID_KEY = Buffer.concat([
+  Buffer.of(0xfc),
+  Buffer.of(0x05),
+  Buffer.from('BITGO', 'ascii'),
+  Buffer.of(Subtype.ZEC_CONSENSUS_BRANCH_ID),
+]);
 
 export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
   protected static transactionFromBuffer(buffer: Buffer, network: Network): ZcashTransaction<bigint> {
@@ -20,6 +30,68 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
       opts,
       data || new PsbtBase(new PsbtTransaction({ tx: new ZcashTransaction<bigint>(opts.network) }))
     );
+  }
+
+  /**
+   * In version < 5 of Zcash transactions, the consensus branch ID is not serialized in the transaction
+   * whereas in version 5 it is. If the transaction is less than a version 5, set the consensus branch id
+   * in the global map in the psbt. If it is a version 5 transaction, throw an error if the consensus
+   * branch id is set in the psbt (because it should be on the transaction already).
+   * @param buffer Psbt buffer
+   * @param opts options
+   */
+  static fromBuffer(buffer: Buffer, opts: PsbtOpts): UtxoPsbt<UtxoTransaction<bigint>> {
+    const psbt = super.fromBuffer(buffer, opts) as ZcashPsbt;
+
+    // Read `consensusBranchId` from the global-map
+    let consensusBranchId: number | undefined = undefined;
+    psbt.data.globalMap.unknownKeyVals?.forEach(({ key, value }, i) => {
+      if (key.equals(CONSENSUS_BRANCH_ID_KEY)) {
+        consensusBranchId = value.readUint32BE();
+      }
+    });
+    switch (psbt.tx.version) {
+      case 4:
+      case ZcashTransaction.VERSION4_BRANCH_CANOPY:
+      case ZcashTransaction.VERSION4_BRANCH_NU5:
+        if (!consensusBranchId || !psbt.data.globalMap.unknownKeyVals) {
+          throw new Error('Could not find consensus branch id on psbt for version 4 Zcash transaction');
+        }
+        psbt.tx.consensusBranchId = consensusBranchId;
+        psbt.data.globalMap.unknownKeyVals = psbt.data.globalMap.unknownKeyVals.filter(
+          ({ key }) => key !== CONSENSUS_BRANCH_ID_KEY
+        );
+        return psbt;
+      case 5:
+      case ZcashTransaction.VERSION5_BRANCH_NU5:
+        if (consensusBranchId) {
+          throw new Error('Found consensus branch id in psbt global-map for version 5 Zcash transaction');
+        }
+        return psbt;
+      default:
+        throw new Error(`Unsupported transaction version ${psbt.tx.version}`);
+    }
+  }
+
+  /**
+   * If it is a version 4 transaction, add the consensus branch id to
+   * the global map. If it is a version 5 transaction, just return the
+   * buffer because the consensus branch id is already serialized in
+   * the transaction.
+   */
+  toBuffer(): Buffer {
+    if (this.tx.version === 5 || this.tx.version === ZcashTransaction.VERSION5_BRANCH_NU5) {
+      return super.toBuffer();
+    }
+    const value = Buffer.alloc(4);
+    value.writeUint32BE(this.tx.consensusBranchId);
+    this.addUnknownKeyValToGlobal({ key: CONSENSUS_BRANCH_ID_KEY, value });
+    if (!this.data.globalMap.unknownKeyVals) {
+      throw new Error('Failed adding consensus branch id to unknownKeyVals');
+    }
+    const buff = super.toBuffer();
+    this.data.globalMap.unknownKeyVals.pop();
+    return buff;
   }
 
   setVersion(version: number, overwinter = true): this {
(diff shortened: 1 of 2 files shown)


modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -22,6 +26,72 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
...
+  static fromBuffer(buffer: Buffer, opts: PsbtOpts): UtxoPsbt<UtxoTransaction<bigint>> {
+    const psbt = super.fromBuffer(buffer, opts) as ZcashPsbt;
+
+    // Read `consensusBranchId` from the global-map
+    const consensusIdKey = Buffer.concat([
+      Buffer.of(0xfc),
+      Buffer.of(0x05),
+      Buffer.from('BITGO', 'ascii'),
+      Buffer.of(Subtype.ZEC_CONSENSUS_BRANCH_ID),
+    ]);
```
brandonblack, line 46: Let's make this a constant.

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -22,6 +26,72 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
...
+        psbt.tx.consensusBranchId = consensusBranchId;
+        return psbt;
+      case 5:
+      case ZcashTransaction.VERSION5_BRANCH_NU5:
+        if (consensusBranchId) {
+          throw new Error('Found consensus branch id in psbt global-map for version 5 Zcash transaction');
+        }
+        return psbt;
+    }
+    throw new Error(`Unsupported transaction version ${psbt.tx.version}`);
```
brandonblack, line 69: I think this would be more clear in a `default` on the switch.

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -22,6 +26,72 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
...
+    if (this.tx.version === 5 || this.tx.version === ZcashTransaction.VERSION5_BRANCH_NU5) {
+      return super.toBuffer();
+    }
+    const key = Buffer.concat([
+      Buffer.of(0xfc),
+      Buffer.of(0x05),
+      Buffer.from('BITGO', 'ascii'),
+      Buffer.of(Subtype.ZEC_CONSENSUS_BRANCH_ID),
+    ]);
+    const value = Buffer.from(this.tx.consensusBranchId.toString(16), 'hex');
```
brandonblack, line 88: I think we should use writeUint32BE (as we do for reading), now that I think about this more. This method could create a smaller than 4-byte buffer if the consensusBranchId were ever to have  leading 0 byte.

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -22,6 +26,72 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
...
+    this.addUnknownKeyValToGlobal({ key, value });
+    const buff = super.toBuffer();
+    this.data.globalMap.unknownKeyVals!.pop();
```
brandonblack, line 91: Let's either assert or add a comment here to explain how we know it's set.

brandonblack (changes requested):
Nice work! Few things to clean up :)

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -7,9 +7,19 @@ import {
...
+enum Subtype {
+  ZEC_CONSENSUS_BRANCH_ID = 0x00,
+}
+const CONSENSUS_BRANCH_ID_KEY = Buffer.concat([
+  Buffer.of(0xfc),
+  Buffer.of(0x05),
+  Buffer.from('BITGO', 'ascii'),
+  Buffer.of(Subtype.ZEC_CONSENSUS_BRANCH_ID),
+]);
+
```
brandonblack, line 22: Nice. When we have more keys in here (eg. for musig2-x-only-agg) we'll probably have a function to relate the enum to the buffer key, but that is future-our problem.

brandonblack (approved):
Woot!

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -22,6 +32,68 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
...
+      return super.toBuffer();
+    }
+    const value = Buffer.alloc(4);
+    value.writeUint32BE(this.tx.consensusBranchId);
+    this.addUnknownKeyValToGlobal({ key: CONSENSUS_BRANCH_ID_KEY, value });
+    if (!this.data.globalMap.unknownKeyVals) {
+      throw new Error('Failed adding consensus branch id to unknownKeyVals');
+    }
+    const buff = super.toBuffer();
+    this.data.globalMap.unknownKeyVals.pop();
```
tylerlevine, line 93: Not a blocker, but I have some dumb questions:

why do we pop the `consensusBranchId` from the `unknownKeyVals` here? Is this so we make it look like `toBuffer()` is side effect free?

Also, I'm curious why we don't add the consensus branch id to the `unknownKeyVals` upon construction (or at least, once we know it's a version 5 tx)?

tylerlevine (approved):
great PR description and clean implementation - very easy to review. Great work!

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
davidkaplanbitgo: > why do we pop the `consensusBranchId` from the `unknownKeyVals` here? Is this so we make it look like `toBuffer()` is side effect free?

Yes

> Also, I'm curious why we don't add the consensus branch id to the `unknownKeyVals` upon construction

This was the discussion that brandon and I had today because I originally had it this way. The problem with doing it this way is that the property of `consensusBranchId` lives on the transaction object, and not on the PSBT (this is a similar behavior as was seen in the [ZcashTransactionBuilder](https://github.com/BitGo/BitGoJS/blob/master/modules/utxo-lib/src/bitgo/zcash/ZcashTransactionBuilder.ts#L105). If we add the consensus branch id to the `unknownKeyVals` map, it would mean that we would have to manage (potentially) two separate states (one on the transaction object and the other in the `unknownKeyVals`). By putting this logic in the serialization phase, we only have to manage one state.

> (or at least, once we know it's a version 5 tx)?

Just to be clear we add it to the `unknownKeyVals` when it is a version 4 transaction, we don't need to for a version 5.

tylerlevine: > Just to be clear we add it to the unknownKeyVals when it is a version 4 transaction, we don't need to for a version 5.

Oops, right, I had that backwards. Thanks for answering all my questions 👍

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -22,6 +32,68 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
...
+    psbt.data.globalMap.unknownKeyVals?.forEach(({ key, value }, i) => {
+      if (key.equals(CONSENSUS_BRANCH_ID_KEY)) {
+        consensusBranchId = value.readUint32BE();
+      }
+    });
+    switch (psbt.tx.version) {
+      case 4:
+      case ZcashTransaction.VERSION4_BRANCH_CANOPY:
+      case ZcashTransaction.VERSION4_BRANCH_NU5:
+        if (!consensusBranchId || !psbt.data.globalMap.unknownKeyVals) {
```
sangaman, line 57: Do we need the second part of this OR statement? I think if `psbt.data.globalMap.unknownKeyVals` is undefined, then `consensusBranchId` will always be undefined already.

sangaman (approved):
Nice comments, just one nit.

For version 5 transactions, I see that we're throwing errors if the consensus branch id is in the psbt. What problem does having this value there cause? Would it make sense to tolerate its presence as just superfluous data?

davidkaplanbitgo:
> For version 5 transactions, I see that we're throwing errors if the consensus branch id is in the psbt. What problem does having this value there cause? Would it make sense to tolerate its presence as just superfluous data?

A potential drawback is that in the case that a customer sends over a PSBT that they built themselves, they could have different consensus id encodings from the transaction serialization and the psbt. I guess we can say in docs that we just ignore what is on the psbt in the case that it is a version 5 transaction but this I think is a bit cleaner.

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
davidkaplanbitgo: It could be that there are other values in the global map (in the future for musig this will be the case) and we do not find the consensus branch id in the map

PR #2835, opened 2022-09-29
feat(sdk-coin-dot): update dot recover function to traverse addresses

alanliu-bitgo:
## Description

Add support to sweep wallet with multiple addresses; these addresses derivation path are 0 indexed and Users must pass in the index to start the search at. User can also optionally pass in how many iteration to search for (ie iteration of 20 with starting index 0 results in index 19 being the last index searched). The default iteration to search for is 20.

Recover function now returns the serialized transaction hex plus the index of the address with the funds being swept so that the user can start at this index + 1 for the next recover call.

## Issue Number

BG-58560

## Type of change

# How Has This Been Tested?

Unit tests, and end to end test confirming that we were able to sweep address 1 of a test wallet

### DIFF
--- a/modules/sdk-coin-dot/src/dot.ts
+++ b/modules/sdk-coin-dot/src/dot.ts
@@ -23,6 +23,7 @@ import { BaseCoin as StaticsBaseCoin, coins, PolkadotSpecNameType } from '@bitgo
 import { Interface, KeyPair as DotKeyPair, Transaction, TransactionBuilderFactory, Utils } from './lib';
 import { ApiPromise, WsProvider } from '@polkadot/api';
 import { Material } from './lib/iface';
+import { isInteger } from 'lodash';
 
 export interface SignTransactionOptions extends BaseSignTransactionOptions {
   txPrebuild: TransactionPrebuild;
@@ -54,10 +55,13 @@ interface RecoveryOptions {
   recoveryDestination: string;
   krsProvider?: string;
   walletPassphrase?: string;
+  startingScanIndex?: number;
+  scan?: number;
 }
 
 interface DotTx {
   serializedTx: string;
+  scanIndex: number;
 }
 
 const dotUtils = Utils.default;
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-dot/src/dot.ts
```diff
@@ -54,9 +54,12 @@ interface RecoveryOptions {
   recoveryDestination: string;
   krsProvider?: string;
   walletPassphrase?: string;
+  startIdx?: number;
+  numIteration?: number;
```
joey-bai, line 58: lets rename `numIteration` to `scan` to be consistent with existing param name.

also, rename `startIdx` to `startingScanningIndex` (this was suggested by Ben in slack messages)

modules/sdk-coin-dot/src/dot.ts
```diff
@@ -54,9 +54,12 @@ interface RecoveryOptions {
...
 }
 
 interface DotTx {
+  addrIdx: number;
```
joey-bai, line 62: maybe rename to `scanIndex` to use same vocab as input

modules/sdk-coin-dot/src/dot.ts
```diff
@@ -336,9 +339,19 @@ export class Dot extends BaseCoin {
...
+   * @returns {DotTx} the serialized transactoin hex string and index
+   * of the address being swept
    */
   async recover(params: RecoveryOptions): Promise<DotTx> {
+    if (!params.startIdx) {
+      params.startIdx = 0;
+    }
+    if (!params.numIteration) {
+      params.numIteration = 20;
+    }
```
joey-bai, line 354: not a fan of mutating inputs.

can do something like:
```
const startIdx = startIdx ?? 0;
const numIteration = numIteration ?? 20;
```

joey-bai: lets also do some input validation.
like params.startIdx >= 0 and is an integer if inputted.
and params.numIteration > 0 and is an integer if inputted

joey-bai (dismissed):

joey-bai (approved):

PR #2836, opened 2022-09-29
feat(sdk-coin-sol): update recover to support derived wallet address

colin-y-wu:
Ticket: BG-57227

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -582,7 +582,7 @@ export class Sol extends BaseCoin {
     // Build the transaction
     const MPC = await EDDSAMethods.getInitializedMpcInstance();
     let bs58EncodedPublicKey;
-    let balance;
+    let balance = 0;
     let addressIndex;
     const feePerSignature = await this.getFees();
     const totalFee = params.durableNonce ? feePerSignature * 2 : feePerSignature;
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -103,7 +103,7 @@ export interface SolParseTransactionOptions extends BaseParseTransactionOptions 
 
 interface SolTx {
   serializedTx: string;
-  addressIndex: number;
+  scanIndex: number;
 }
 
 interface SolDurableNonceFromNode {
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-sol/src/sol.ts
```diff
@@ -574,16 +580,35 @@ export class Sol extends BaseCoin {
...
+
+      balance = await this.getAccountBalance(bs58EncodedPublicKey);
+
+      if (balance > totalFee) {
+        addressIndex = i;
+        break;
+      }
+    }
+
+    if (balance < totalFee) {
```
michael-gallegos-bg, line 603: Could we test how this would work with unexpected inputs? What if the endingDerivationIndex is zero (we might never have that case, but still) - then wouldn't balance always be undefined at this point since the for loop never triggers, and perhaps the comparison would error out?

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -562,8 +562,8 @@ export class Sol extends BaseCoin {
    */
   async recover(params: RecoveryOptions): Promise<SolTx> {
     const isUnsignedSweep = !params.userKey && !params.backupKey && !params.walletPassphrase;
-    const startingDerivationIndex = params.startingScanningIndex ? params.startingScanningIndex : 0;
-    const endingDerivationIndex = params.scan ? params.scan : 4294967295; // largest positive 32 bit int
```
colin-y-wu: using lodash to check for undefined now, because when the `starting` and `ending` indices are 0, they are actually falsey values

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -119,6 +120,8 @@ interface RecoveryOptions {
     publicKey: string;
     secretKey: string;
   };
+  startingScanningIndex?: number;
```
joey-bai, line 123: nit: `startingScanIndex`?

if its consistent with Dot + other coins, leave it as is

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -559,6 +562,9 @@ export class Sol extends BaseCoin {
    */
   async recover(params: RecoveryOptions): Promise<SolTx> {
     const isUnsignedSweep = !params.userKey && !params.backupKey && !params.walletPassphrase;
+    const startingDerivationIndex = !_.isUndefined(params.startingScanningIndex) ? params.startingScanningIndex : 0;
+    const endingDerivationIndex = !_.isUndefined(params.scan) ? params.scan : 4294967295; // largest positive 32 bit int
```
joey-bai, line 566: this should be `params.scan + startingDerivationIndex`.

`scan` defines the number of addresses we want to scan through before halting.

also, lets default scanning to 20 or something.

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -559,6 +562,9 @@ export class Sol extends BaseCoin {
    */
   async recover(params: RecoveryOptions): Promise<SolTx> {
     const isUnsignedSweep = !params.userKey && !params.backupKey && !params.walletPassphrase;
+    const startingDerivationIndex = !_.isUndefined(params.startingScanningIndex) ? params.startingScanningIndex : 0;
```
joey-bai, line 565: verify that `startingScanningIndex` is an integer and >= 0 if provided

modules/sdk-coin-sol/src/sol.ts
joey-bai, line 566: verify `params.scan` is an integer and > 0 if provided

joey-bai (approved):
lgtm but you have a merge conflict

joey-bai (approved):

PR #2838, opened 2022-09-29
fix(sdk-coin-sol): fix sol deserialize incorrect feePayer

linyah9912:
Ticket: BG-58613

Currently the sol tx deserialize doesn't set feePayer, so when the feePayer is different than the sender, it throws unknown signer error. Fixed with adding feePayer.

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/sdk-coin-sol/src/lib/transactionBuilder.ts
+++ b/modules/sdk-coin-sol/src/lib/transactionBuilder.ts
@@ -18,7 +18,7 @@ import { KeyPair } from '.';
 import { InstructionBuilderTypes } from './constants';
 import { solInstructionFactory } from './solInstructionFactory';
 import assert from 'assert';
-import { DurableNonceParams, InstructionParams, Memo, Nonce } from './iface';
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-sol/src/lib/transactionBuilder.ts
```diff
@@ -53,6 +53,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     this._transaction = tx;
     const txData = tx.toJson();
     this.sender(txData.feePayer as string);
```
joey-bai, line 55: i wonder if this is going to break anything... i don't think so, but may be worth adding a check.

joey-bai (dismissed):

modules/sdk-coin-sol/src/lib/transactionBuilder.ts
linyah9912: updated with the sender

joey-bai (approved):
thanks for fixing this!

PR #2856, opened 2022-10-03
Bg 58807 fix eth tss signing verifying

anshuldoshi:
## Description
SDK fixes for ETH TSS. This change includes 2 fixes
1. fix to/from of how we're sending user s share to pass signature share validation in platform
2. fix how we are signing eth tss txs to ensure we sign the hashed, RLP encoded tx rather than the hashed serialized tx

## Issue Number
bg-58807

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts
+++ b/modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts
@@ -121,7 +121,7 @@ describe('TSS ECDSA TESTS', function () {
   });
 
   describe('ECDSA Signing', async function () {
-    let config: { signerOne: ECDSA.KeyCombined; signerTwo: ECDSA.KeyCombined; hash?: Hash }[];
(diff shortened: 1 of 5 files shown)


modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -213,9 +217,24 @@ describe('TSS ECDSA TESTS', function () {
         // and finally signs the message using their private OShare
         // and delta share received from the other signer
 
+        const hashGenerator = (hashType?: string): Hash | undefined => {
```
anshuldoshi, line 220: we need to generate a hash each time rather than supplying the same Hash object because Hash objects hold state and the same objects.will not work for multiple uses. This would never happen in practice because each party. would be signing separately, but we run into this issue when testing

olibeyene (approved):

PR #2865, opened 2022-10-05
feat(sdk-core): add support for ETH TSS staking

noel-bitgo:
BOS-627

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -2100,7 +2100,11 @@ export class Wallet implements IWallet {
    * Create a staking wallet from this wallet
    */
   toStakingWallet(): StakingWallet {
-    return new StakingWallet(this);
+    const isEthTss =
+      this.baseCoin.getFamily() == 'eth' && this._wallet.coinSpecific?.walletVersion
+        ? this._wallet.coinSpecific.walletVersion >= 3
+        : false;
+    return new StakingWallet(this, isEthTss);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -115,6 +115,12 @@ export class Wallet implements IWallet {
     }
   }
 
+  isEthTss(): boolean {
```
mmcshinsky-bitgo, line 118: Can we do the same as https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/wallets.ts#L182 until we see the use case expand in being used in multiple places?

mmcshinsky-bitgo (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
noel-bitgo: I am not sure I understand what you're suggesting? 🤔

mmcshinsky-bitgo: Can we move this logic into a variable without adding a new method in the `build` method the same way we are doing it in `generateWallet`?

noel-bitgo: just pushed a change, let me know if this is what you meant

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2107,7 +2107,11 @@ export class Wallet implements IWallet {
     if (!this.baseCoin.supportsStaking()) {
       throw new Error(`Staking not supported for ${this.coin()}`);
     }
-    return new StakingWallet(this);
+    const isEthTss =
+      this.baseCoin.getFamily() == 'eth' && this._wallet.coinSpecific?.walletVersion
```
mmcshinsky-bitgo, line 2111: Does this need a to check for `teth` and `geth` or are they all returning `eth`?

noel-bitgo: as far as I can tell the coin family is eth

mmcshinsky-bitgo (approved):

PR #2867, opened 2022-10-05
fix(utxo-lib): fix fromTransactionComplete

OttoAllmendinger:
Prior to this change, we only could only track a single legacy input
per txid. This change allows us to track all legacy inputs.

Issue: BG-47821


brandonblack:
I found the same bug and am changing the semantics on this function to make adding non-witness UTXOs a follow-up call. Nice eye!

PR #2874, opened 2022-10-05
feat(utxo-lib): [psbt] separate adding input and nonWitnessUtxo

brandonblack:
also adds test coverage for this functionality

Issue: BG-58455

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -147,7 +147,7 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
     return new UtxoTransaction<bigint>(network);
   }
 
-  get tx(): Tx {
+  protected get tx(): Tx {
     return (this.data.globalMap.unsignedTx as PsbtTransaction).tx as Tx;
   }
 
@@ -366,7 +366,6 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
     hash: Buffer;
     sighashType: number;
   } {
-    const unsignedTx = this.tx;
     const sighashType = this.data.inputs[inputIndex].sighashType || Transaction.SIGHASH_DEFAULT;
     if (sighashTypes && sighashTypes.indexOf(sighashType) < 0) {
       throw new Error(
@@ -374,6 +373,7 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
           `sighashTypes array of whitelisted types. Sighash type: ${sighashType}`
       );
     }
+    const txInputs = this.txInputs; // These are somewhat costly to extract
     const prevoutScripts: Buffer[] = [];
     const prevoutValues: bigint[] = [];
 
@@ -383,10 +383,10 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
         // TODO: This could be costly, either cache it here, or find a way to share with super
         const nonWitnessUtxoTx = (this.constructor as typeof UtxoPsbt).transactionFromBuffer(
           input.nonWitnessUtxo,
-          unsignedTx.network
+          this.tx.network
         );
 
-        const prevoutHash = unsignedTx.ins[inputIndex].hash;
+        const prevoutHash = txInputs[inputIndex].hash;
         const utxoHash = nonWitnessUtxoTx.getHash();
 
         // If a non-witness UTXO is provided, its hash must match the hash specified in the prevout
@@ -396,7 +396,7 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
           );
         }
 
-        const prevoutIndex = unsignedTx.ins[inputIndex].index;
+        const prevoutIndex = txInputs[inputIndex].index;
         prevout = nonWitnessUtxoTx.outs[prevoutIndex];
       } else if (input.witnessUtxo) {
         prevout = input.witnessUtxo;
@@ -406,7 +406,7 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
       prevoutScripts.push(prevout.script);
       prevoutValues.push(prevout.value);
     }
-    const hash = unsignedTx.hashForWitnessV1(inputIndex, prevoutScripts, prevoutValues, sighashType, leafHash);
+    const hash = this.tx.hashForWitnessV1(inputIndex, prevoutScripts, prevoutValues, sighashType, leafHash);
     return { hash, sighashType };
   }
 }
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -7,12 +7,13 @@ import {
   Transaction as ITransaction,
   TransactionFromBuffer,
 } from 'bip174/src/lib/interfaces';
+import { getOutputIdForInput } from './Unspent';
 import { checkForInput } from 'bip174/src/lib/utils';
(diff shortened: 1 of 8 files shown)


modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -526,3 +524,21 @@ function matchSignatures(hashFn: (hashType: number) => Buffer, script: Buffer, s
   }
   return partialSig;
 }
+
+function isP2wsh(scriptPubkey: Buffer, redeemScript?: Buffer): boolean {
```
OttoAllmendinger, line 528: nit: I prefer topological order (definition before use)

modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
```diff
@@ -123,13 +127,13 @@ describe('WalletUnspent', function () {
     });
   });
 
-  function constructAndSignTransactionUsingPsbt(
+  async function constructAndSignTransactionUsingPsbt(
```
OttoAllmendinger, line 130: that's what I meant with upwards propagation...

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -68,26 +68,27 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
-        txidToIndex[reverseBuffer(transaction.ins[index].hash).toString('hex')] = index;
+  async addNonWitnessUtxos(fetchTransactions: (txids: string[]) => Promise<Record<string, Buffer>>): Promise<this> {
+    const txidToIndex = new Map<string, number[]>();
+    this.data.inputs.forEach((input, index) => {
+      if (!input.witnessUtxo) {
+        throw new Error('Must have witness UTXO for all inputs');
+      }
+      if (!isSegwit(input.witnessUtxo.script, input.redeemScript)) {
+        // Copy, then reverse buffer
+        const txid = reverseBuffer(Buffer.from(this.tx.ins[index].hash)).toString('hex');
```
OttoAllmendinger, line 79:
```suggestion
        const txid = getOutputIdForInput(this.tx.ins[index]).txid;
```

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -68,26 +68,27 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
-    prevOutputs: TxOutput<bigint>[],
-    fetchTransactions: (txids: string[]) => Promise<Record<string, Buffer>>
-  ): Promise<UtxoPsbt<UtxoTransaction<bigint>>> {
-    const psbt = this.fromTransaction(transaction, prevOutputs);
-
-    const txidToIndex: Record<string, number> = {};
-    psbt.data.inputs.forEach((input, index) => {
-      if (!input.tapLeafScript && !input.witnessScript) {
-        txidToIndex[reverseBuffer(transaction.ins[index].hash).toString('hex')] = index;
+  async addNonWitnessUtxos(fetchTransactions: (txids: string[]) => Promise<Record<string, Buffer>>): Promise<this> {
```
OttoAllmendinger, line 71: I am not a fan of `async` here. It has the nasty habit to propagate up the call chain.

I think we can realize this with a sync signature `addNonWitnessUtxos(txBuffers: Map<string, Buffer>)` instead.

It would be more consistent with the way we handle `prevOutputs: TxOutput[]`.

To help the caller figure out which transactions to fetch we can add a method `getNonWitnessInputs(): { index: number, input: TxInput }[]`

OttoAllmendinger (dismissed):
I actually preferred the previous style more, where the resulting object was complete without the caller having to do additional processing. Why the departure?

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -68,26 +68,27 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+        const indices = txidToIndex.get(txid) || [];
+        indices.push(index);
+        txidToIndex.set(txid, indices);
       }
     });
 
-    const txHexs = await fetchTransactions(Object.keys(txidToIndex));
-    Object.entries(txidToIndex).forEach(([txid, index]) => {
-      psbt.updateInput(index, { nonWitnessUtxo: txHexs[txid] });
+    const txHexs = await fetchTransactions([...txidToIndex.keys()]);
```
OttoAllmendinger, line 86: Since the values are actually `Buffer` not `string`, the name `txHexs` is a misnomer (`txBytes` maybe ?)

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
OttoAllmendinger: reason to use `Map` instead of `Record`: https://stackoverflow.com/a/18541990

OttoAllmendinger: also a docsting explaining the purpose would be nice

brandonblack:
Both for testing and for WP, having the async call on creation was gonna be a pain, as you mentioned. It was easier to deal with on a separate call, but it'll be even better w/o the async at all :-D

modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
```diff
@@ -143,7 +147,24 @@ describe('WalletUnspent', function () {
...
+      const ret: Record<string, Buffer> = {};
+      txids.forEach((txid) => {
+        const u = nonWitnessUnspents.find((u) => parseOutputId(u.id).txid === txid);
+        if (u === undefined) throw new Error('No prevtx found');
+        ret[txid] = u.prevTx;
+      });
+      return ret;
+    });
+    assert(psbt2.validateSignaturesOfAllInputs());
+    return tx; // extract transaction has a return type of Transaction instead of UtxoTransaction
```
davidkaplanbitgo, line 167: nit: dont know if this comment is necessary to have in

davidkaplanbitgo (dismissed):
Noice

davidkaplanbitgo (approved):

modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
OttoAllmendinger, line 130: this can be sync again

modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
```diff
@@ -189,7 +210,7 @@ describe('WalletUnspent', function () {
     amountType: 'number' | 'bigint' = 'number',
     testOutputAmount = toTNumber<TNumber>(defaultTestOutputAmount, amountType)
   ) {
-    it(`can be signed [scriptType=${scriptType} signer=${signer} cosigner=${cosigner} amountType=${amountType}]`, function () {
+    it(`can be signed [scriptType=${scriptType} signer=${signer} cosigner=${cosigner} amountType=${amountType}]`, async function () {
```
OttoAllmendinger, line 213: dito

modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
```diff
@@ -210,7 +231,7 @@ describe('WalletUnspent', function () {
         scriptType
       );
       if (amountType === 'bigint') {
-        const psbtTransaction = constructAndSignTransactionUsingPsbt(
+        const psbtTransaction = await constructAndSignTransactionUsingPsbt(
```
OttoAllmendinger, line 234: dito

brandonblack:
https://github.com/BitGo/BitGoJS/pull/2882

PR #2880, opened 2022-10-07
fix(statics): update underlying asset of GTETH to ETH

sachushaji:
the underlying asset for GTETH should
be ETH similar to that of 'TETH'.

TICKET: BG-59194

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -180,7 +180,6 @@ export enum UnderlyingAsset {
   EUROC = 'euroc',
   GBP = 'gbp',
   GTC = 'gtc',
-  GTETH = 'gteth',
   HBAR = 'hbar', // Hedera main coin
   LTC = 'ltc',
   NEAR = 'near',
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -103,14 +103,7 @@ export const coins = CoinMap.fromCoins([
     ...ETH_FEATURES,
     CoinFeature.DEPRECATED,
   ]),
-  account(
-    'gteth',
-    'Goerli Testnet Ethereum',
-    Networks.test.goerli,
-    18,
-    UnderlyingAsset.GTETH,
-    ETH_FEATURES_WITH_STAKING
-  ),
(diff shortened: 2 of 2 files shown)


modules/statics/src/base.ts
```diff
@@ -176,7 +176,6 @@ export enum UnderlyingAsset {
   EUROC = 'euroc',
   GBP = 'gbp',
   GTC = 'gtc',
-  GTETH = 'gteth',
```
mmcshinsky-bitgo, line 179: Untracked breaking change missing in commit message footer.

mmcshinsky-bitgo (changes requested):

modules/statics/src/base.ts
sachushaji: @mmcshinsky-bitgo Updated, Thanks for the nudge

modules/statics/src/coins.ts
```diff
@@ -100,14 +100,7 @@ export const coins = CoinMap.fromCoins([
   account('tdot', 'Testnet Polkadot', Networks.test.dot, 12, UnderlyingAsset.DOT, DOT_FEATURES, KeyCurve.Ed25519),
   account('eth', 'Ethereum', Networks.main.ethereum, 18, UnderlyingAsset.ETH, ETH_FEATURES_WITH_STAKING),
   account('teth', 'Kovan Testnet Ethereum (Deprecated)', Networks.test.kovan, 18, UnderlyingAsset.ETH, ETH_FEATURES),
-  account(
```
wesgraham, line 103: do we need to delete the entire entry? or can we just use underlyingasset.eth?

sachushaji: @wesgraham did not deleted it, only change is asset, since prettier was automatically applied, the details of the GTETH is now in one line instead of multiple lines

modules/statics/src/base.ts
mmcshinsky-bitgo: No problem :) Needs to be in the commit footer though :)

sachushaji: @mmcshinsky-bitgo Updated, Kindly check, If its still not right?, Kindly give me the correct format

mmcshinsky-bitgo (approved):

PR #2886, opened 2022-10-07
feat(sdk-core): Allow getting a staking wallet for any coin

mcalcano-bg:
allow getting a staking wallet for any coin

allow clients to send arbitrary properties when staking or unstaking

Ticket: BOS-649

### DIFF
--- a/modules/sdk-core/src/bitgo/staking/iStakingWallet.ts
+++ b/modules/sdk-core/src/bitgo/staking/iStakingWallet.ts
@@ -21,6 +21,8 @@ export interface StakingRequest {
 export interface StakeOptions {
   amount: string;
   clientId?: string;
+  validator?: string;
+  durationSeconds?: string;
 }
 
 export interface UnstakeOptions {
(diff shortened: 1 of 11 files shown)


modules/sdk-core/src/bitgo/staking/iStakingWallet.ts
```diff
@@ -21,12 +21,16 @@ export interface StakingRequest {
 export interface StakeOptions {
   amount: string;
   clientId?: string;
+  validator?: string;
+  durationSeconds?: string;
+  [otherOptions: string]: unknown;
```
bitgoAaron, line 26: this kind of breaks the strict typing we prefer to use. Can you elaborate on the use case here?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2104,9 +2104,6 @@ export class Wallet implements IWallet {
    * Create a staking wallet from this wallet
    */
   toStakingWallet(): StakingWallet {
-    if (!this.baseCoin.supportsStaking()) {
```
bitgoAaron, line 2107: this method is no longer used anywhere, please also remove from baseCoin & all the other coins that override it.

bitgoAaron (changes requested):
needs clarification on `otherOptions` and cleanup dead code.

modules/sdk-core/src/bitgo/wallet/wallet.ts
mcalcano-bg: Fixed.

joey-bai (approved):

olibeyene (approved):

PR #2890, opened 2022-10-11
fix: create txrequest for message signing

jeanlarrieux:
Ticket: BG-54920

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -1980,52 +1980,70 @@ describe('V2 Wallet:', function () {
       };
       let signTxRequestForMessage;
       const messageSigningCoins = ['teth', 'tpolygon'];
+      const message = 'test';
 
       beforeEach(async function () {
         signTxRequestForMessage = sandbox.stub(ECDSAUtils.EcdsaUtils.prototype, 'signTxRequestForMessage');
         signTxRequestForMessage.resolves(txRequestForMessageSigning);
-        // TODO(BG-59686): this is not doing anything if we don't check the return value
-        signTxRequestForMessage.calledOnceWithExactly({ txRequest: txRequestForMessageSigning, prv: 'secretKey', reqId });
+      });
+
+      afterEach(async function () {
+        sinon.restore();
       });
 
       it('should throw error for unsupported coins', async function () {
 
         await tssWallet.signMessage({
           reqId,
-          messagePrebuild: { message: 'test' },
+          messagePrebuild: { message },
           prv: 'secretKey',
         }).should.be.rejectedWith('Message signing not supported for Testnet Solana');
       });
 
       messageSigningCoins.map((coinName) => {
         tssEthWallet = new Wallet(bitgo, bitgo.coin(coinName), ethWalletData);
+        const txRequestId = txRequestForMessageSigning.txRequestId;
+
         it('should sign message', async function () {
+          const signMessageTssSpy = sinon.spy(tssEthWallet, 'signMessageTss' as any);
+          nock(bgUrl)
+            .get(`/api/v2/wallet/${tssEthWallet.id()}/txrequests?txRequestIds=${txRequestForMessageSigning.txRequestId}&latest=true`)
+            .reply(200, { txRequests: [txRequestForMessageSigning] });
 
           const signMessage = await tssEthWallet.signMessage({
             reqId,
-            messagePrebuild: { message: 'test', txRequestId: 'id' },
+            messagePrebuild: { message, txRequestId },
             prv: 'secretKey',
           });
-          signMessage.should.deepEqual({ txRequestId: 'id' } );
+          signMessage.should.deepEqual({ txRequestId } );
+          const actualArg = signMessageTssSpy.getCalls()[0].args[0];
+          actualArg.messagePrebuild.message.should.equal(`\u0019Ethereum Signed Message:\\n${message.length}${message}`);
         });
 
-        it('should fail to sign message without txRequestId', async function() {
-          await tssEthWallet.signMessage({
+        it('should sign message when txRequestId not provided', async function () {
+          const signMessageTssSpy = sinon.spy(tssEthWallet, 'signMessageTss' as any);
+          nock(bgUrl)
+            .post(`/api/v2/wallet/${tssEthWallet.id()}/txrequests`)
+            .reply(200, txRequestForMessageSigning);
+
+          const signMessage = await tssEthWallet.signMessage({
             reqId,
-            messagePrebuild: { message: '' },
+            messagePrebuild: { message },
             prv: 'secretKey',
-          }).should.be.rejectedWith('txRequestId required to sign message with TSS');
+          });
+          signMessage.should.deepEqual({ txRequestId } );
+          const actualArg = signMessageTssSpy.getCalls()[0].args[0];
+          actualArg.messagePrebuild.message.should.equal(`\u0019Ethereum Signed Message:\\n${message.length}${message}`);
         });
 
         it('should fail to sign message with empty prv', async function () {
           await tssEthWallet.signMessage({
             reqId,
-            messagePrebuild: { message: 'test', txRequestId: 'id' },
+            messagePrebuild: { message, txRequestId },
             prv: '',
           }).should.be.rejectedWith('prv required to sign message with TSS');
         });
       });
-
 
     });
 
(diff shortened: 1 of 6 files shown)


modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
```diff
@@ -163,6 +165,40 @@ export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtil
...
+    params: IntentOptionsBase,
+    apiVersion: TxRequestVersion = 'full',
+    preview?: boolean
+  ): Promise<TxRequest> {
+    const intentOptions: PopulatedIntentBase = {
+      intentType: params.intentType,
+      sequenceId: params.sequenceId,
+      comment: params.comment,
+      memo: params.memo?.value,
+      isTss: params.isTss,
```
karankurbur, line 185: should this always be true?

karankurbur:
lgtm overall

karankurbur (approved):

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
```diff
@@ -163,6 +165,40 @@ export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtil
...
+   * @param params
+   * @param apiVersion
+   * @param preview
+   */
+  async createTxRequestWithIntentForMessageSigning(
+    params: IntentOptionsBase,
+    apiVersion: TxRequestVersion = 'full',
+    preview?: boolean
+  ): Promise<TxRequest> {
+    const intentOptions: PopulatedIntentBase = {
```
karankurbur, line 180: u need an optional parameter for `custodianMessageId` to support message requests initiated from MMI. This will be stored in the txRequest model and used to send updates back to the extension.

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
jeanlarrieux: I set it to true when I call this function from wallet. I want to leave the option for this to happen for non-tss wallet in the future.

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
jeanlarrieux: Is that needed for all txrequest or just message signing?

karankurbur: just message signing -> I have created two new intents to be used in the txRequest API

There is a different field for txs -> `custodianTransactionId`

karankurbur: https://github.com/BitGo/bitgo-microservices/pull/24139

jeanlarrieux: can we handle this in a follow up PR?

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
```diff
@@ -284,6 +284,15 @@ export abstract class BaseCoin implements IBaseCoin {
     return false;
   }
 
+  /**
+   * Transform message to accommodate specific blockchain requirements.
+   * @param message the message to prepare
+   * @return string the prepared message.
+   */
+  prepareMessage(message: string): string {
```
aaldemir, line 292: instead of adding a method to the parent class, I think you can have the eth class `implement` an interface which include the `prepareMessage` fn

adding a method which is used to be a few coins to the base class is not preferrable

note that this is a non-blocking comment

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2651,21 +2652,34 @@ export class Wallet implements IWallet {
...
+      txRequest = await this.tssUtils!.createTxRequestWithIntentForMessageSigning(intentOption);
+      params.messagePrebuild.txRequestId = txRequest.txRequestId;
     }
 
     if (!params.prv) {
       throw new Error('prv required to sign message with TSS');
     }
 
     try {
+      if (!txRequest) {
```
aaldemir, line 2677: this might be more readable if its sets up with 
if there is not txRequestId -> createTxRequestWithIntenForMessageSigning
else -> getTxRequest

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -1944,28 +1942,24 @@ describe('V2 Wallet:', function () {
...
+
         it('should sign message', async function () {
+          nock(bgUrl)
+            .get(`/api/v2/wallet/${tssEthWallet.id()}/txrequests?txRequestIds=${txRequestForMessageSigning.txRequestId}&latest=true`)
+            .reply(200, { txRequests: [txRequestForMessageSigning] });
 
           const signMessage = await tssEthWallet.signMessage({
             reqId,
-            messagePrebuild: { message: 'test', txRequestId: 'id' },
+            messagePrebuild: { message: 'test', txRequestId: txRequestForMessageSigning.txRequestId },
```
aaldemir, line 1953: should there be a test for when there is a `txRequestId` provided and one for when there is not a `txRequestId` provided? It seems like that is the primary goal of this PR

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
jeanlarrieux: what if this is needed for other coins in the future?

modules/bitgo/test/v2/unit/wallet.ts
jeanlarrieux: The one for no request Id provided would have to get it from WP, which would be an integration test at that point. Unless we want to return a mock one. I didn't see the usefulness of returning a mock id.

aaldemir: I think adding a mocked test would check if/else logic in `signMessageTss()` - am I thinking about this wrong?

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
aaldemir: Then maybe coin can implement the same interface. The base class is just returning the message - it's not performing any transformation or manipulation for the message. If the base class performed some sort of common manipulation that is needed for all coins then that is different.

Again, this is not a blocking comment. Just want to bring this up so that we avoid adding a new method that does nothing to the base class every time a child class needs a new method. Unfortunately with JS a class can only inherit from a single parent.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2651,21 +2652,32 @@ export class Wallet implements IWallet {
...
-      throw new Error('txRequestId required to sign message with TSS');
-    }
-
     if (!params.prv) {
       throw new Error('prv required to sign message with TSS');
     }
 
     try {
+      let txRequest;
+      if (!params.messagePrebuild.txRequestId) {
```
aaldemir, line 2667: did you remove the check for `messagePrebuild`? If its not needed then maybe we use optional chaining just in case its not on the `params`

jeanlarrieux: I did not remove it, just moved it around.

aaldemir: oh yeah my bad, the check was existing before. resolving this.

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
jeanlarrieux: added unit test and `spy` to check that message prefix was added.

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -1980,53 +1980,71 @@ describe('V2 Wallet:', function () {
...
+        it('should sign message when txRequestId not provided', async function () {
+          const signMessageTssSpy = sinon.spy(tssEthWallet, 'signMessageTss' as any);
+          nock(bgUrl)
+            .post(`/api/v2/wallet/${tssEthWallet.id()}/txrequests`)
+            .reply(200, txRequestForMessageSigning);
+
+          const signMessage = await tssEthWallet.signMessage({
             reqId,
-            messagePrebuild: { message: '' },
+            messagePrebuild: { message },
```
aaldemir, line 2031: nice

aaldemir (approved):
LGTM

aaldemir (approved):

PR #2896, opened 2022-10-13
feat: allow the sdk to optionally send cookies with the request to cu…

vitalir-bg:
…stom domains

TICKET: BG-59381

## Description
This PR adds an optional argument to the constructor which enables sending cookies if requested but only for custom URIs.
  

## Type of change

# How Has This Been Tested?
unit tests

# Checklist:

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -26,6 +26,7 @@ import {
   defaultConstants,
   EncryptOptions,
   EnvironmentName,
+  Environments,
   getAddressP2PKH,
   getSharedSecret,
   GetSharingKeyOptions,
@@ -128,8 +129,10 @@ export class BitGoAPI implements BitGoBase {
   protected readonly _clientId?: string;
   protected readonly _clientSecret?: string;
   protected _validate: boolean;
+  public readonly cookiesPropagationEnabled: boolean;
 
   constructor(params: BitGoAPIOptions = {}) {
+    this.cookiesPropagationEnabled = false;
     if (
       !common.validateParams(
         params,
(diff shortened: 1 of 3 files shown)


byhow (dismissed):

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -128,8 +129,10 @@ export class BitGoAPI implements BitGoBase {
   protected readonly _clientId?: string;
   protected readonly _clientSecret?: string;
   protected _validate: boolean;
+  public readonly allowSendingCookies: boolean;
```
mmcshinsky-bitgo, line 132:
```suggestion
  public readonly enableCookies: boolean;
```
More clear in intent

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -183,10 +186,23 @@ export class BitGoAPI implements BitGoBase {
       if (params.stellarFederationServerUrl) {
         common.Environments[env].stellarFederationServerUrl = params.stellarFederationServerUrl;
       }
+      if (
+        params.customRootURI &&
```
mmcshinsky-bitgo, line 190: Why block cookie usage if a user specifies it?

modules/sdk-api/src/bitgoAPI.ts
vitalir-bg: Yeah, I also feel like allowSendingCookies is both long and vagueish so I am not satisfied with the name.
On the other hand `enableCookies` may sound like the SDK uses cookies to implement some internal logic, which is not the case. This change is to just allow the SDK user to send cookies with the request if they want this. Should it be `enableCookiesPropagation`?

LeoKotschenreuther (approved):
great work!

mmcshinsky-bitgo (approved):

PR #2897, opened 2022-10-13
feat(sdk-core): added large value support while calling WP

dpkjnr:
Changed type of minValue, maxValue and target params to support both number and string in different APIs.
Wallet platform will support both types for all the coins except for doge and for doge it will only support string

Ticket: BG-57258


OttoAllmendinger (approved):

sangaman (commented):
Changes LGTM for the most part, but are the tests just testing that the `IWallet` methods we define in sdk-core accept strings for these "large value" parameters? It seems like something that  ought to be checked at compile-time rather than run-time in the tests, which would follow since the changes in this PR are just to the types we've defined and not the runtime code. It's not a problem per se but might be a bit superfluous.

Also, do we need to update the balance-related fields in the `WalletData` type?

sangaman:
> balance related fields already support string type as it is required for account based coins too. Only unspents related fields required these changes

These appear to be `number` only, are we sure we don't need to update them? https://github.com/BitGo/BitGoJS/blob/870621e2bc93d15ed6f040379353d039eb17e609/modules/sdk-core/src/bitgo/wallet/iWallet.ts#L448

PR #2900, opened 2022-10-13
Bg 58908 validate istss evm

shiksb:
## Description

Make sure that TSS wallet creations for EVMs have wallet version set as 3. 

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallets.ts
@@ -205,6 +205,11 @@ export class Wallets implements IWallets {
       walletParams.enterprise = params.enterprise;
     }
 
+    // EVM TSS wallets must use wallet version 3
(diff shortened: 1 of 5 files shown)


kxl4126 (approved):

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -205,6 +205,11 @@ export class Wallets implements IWallets {
       walletParams.enterprise = params.enterprise;
     }
 
+    // EVM TSS wallets must use wallet version 3
+    if ((isTss && this.baseCoin.isEVM()) !== (params.walletVersion === 3)) {
```
rlutzbitgo, line 209: What about if a wallet is TSS, but is not an EVM wallet and we wanted to use walletVersion 3? We would say

`if ( true && false ) !== true)` 
or `if (false !== true)`

So if I try to make a non-EVM TSS wallet on walletVersion 3, this will always throw an error?

PR #2908, opened 2022-10-17
fix(utxo-lib): bigintify sign txn for doge

dpkjnr:
Ticket: BG-57258

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -393,7 +393,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
    * @returns {Array}
    */
   protected static findMissingOutputs(expectedOutputs: Output[], actualOutputs: Output[]): Output[] {
-    const keyFunc = ({ address, amount }: Output): string => `${address}:${Number(amount)}`;
+    const keyFunc = ({ address, amount }: Output): string => `${address}:${amount}`;
     const groupedOutputs = _.groupBy(expectedOutputs, keyFunc);
 
     actualOutputs.forEach((output) => {
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -63,7 +63,7 @@ import { supportedCrossChainRecoveries } from './config';
 
 const { getExternalChainCode, isChainCode, scriptTypeForChain, outputScripts, toOutput, verifySignatureWithUnspent } =
   bitgo;
-type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber>;
+export type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber>;
 
 type RootWalletKeys = bitgo.RootWalletKeys;
 export interface VerifyAddressOptions extends BaseVerifyAddressOptions {
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -63,7 +63,7 @@ import { supportedCrossChainRecoveries } from './config';
 
 const { getExternalChainCode, isChainCode, scriptTypeForChain, outputScripts, toOutput, verifySignatureWithUnspent } =
   bitgo;
-export type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber>;
+type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber>;
 
 type RootWalletKeys = bitgo.RootWalletKeys;
 export interface VerifyAddressOptions extends BaseVerifyAddressOptions {
--- a/modules/sdk-coin-doge/src/doge.ts
+++ b/modules/sdk-coin-doge/src/doge.ts
@@ -10,9 +10,48 @@ import {
   CrossChainRecoverySigned,
   CrossChainRecoveryUnsigned,
   RecoverFromWrongChainOptions,
+  Unspent,
+  TransactionInfo,
+  TransactionPrebuild,
 } from '@bitgo/abstract-utxo';
 import { BaseCoin, BitGoBase, HalfSignedUtxoTransaction, SignedTransaction } from '@bitgo/sdk-core';
 import * as utxolib from '@bitgo/utxo-lib';
+
+type UnspentJSON = Unspent<bigint> & { value: number; valueString: string };
+type TransactionInfoJSON = TransactionInfo<bigint> & { unspents: UnspentJSON[] };
+type TransactionPrebuildJSON = TransactionPrebuild<bigint> & { txInfo: TransactionInfoJSON };
+
+function parseUnspents<TNumber extends number | bigint>(
+  unspents: UnspentJSON[] | Unspent<TNumber>[]
+): Unspent<bigint>[] {
+  return unspents.map((unspent: Unspent<TNumber> | UnspentJSON): Unspent<bigint> => {
+    if (typeof unspent.value === 'bigint') {
+      return unspent as Unspent<bigint>;
+    }
+    if ('valueString' in unspent) {
+      return { ...unspent, value: BigInt(unspent.valueString) };
+    }
+    if (typeof unspent.value === 'number') {
+      throw new Error(`received Unspent<number> where Unspent<bigint> or UnspentJSON was expected`);
+    }
+    throw new Error('invalid unspent');
+  });
+}
+
+function parseTransactionInfo<TNumber extends number | bigint>(
+  txInfo: TransactionInfo<TNumber> | TransactionInfoJSON
+): TransactionInfo<bigint> {
+  return { ...txInfo, unspents: parseUnspents(txInfo.unspents) };
+}
+
+function parseTransactionPrebuild<TNumber extends number | bigint>(
+  txPrebuild: TransactionPrebuild<TNumber> | TransactionPrebuildJSON
+): TransactionPrebuild<bigint> {
+  if (txPrebuild?.txInfo) {
+    return { ...txPrebuild, txInfo: parseTransactionInfo(txPrebuild.txInfo) };
+  }
+  return txPrebuild as TransactionPrebuild<bigint>;
+}
 
 export class Doge extends AbstractUtxoCoin {
   constructor(bitgo: BitGoBase, network?: UtxoNetwork) {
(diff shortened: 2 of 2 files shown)


modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -63,7 +63,7 @@ import { supportedCrossChainRecoveries } from './config';
 
 const { getExternalChainCode, isChainCode, scriptTypeForChain, outputScripts, toOutput, verifySignatureWithUnspent } =
   bitgo;
-type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber>;
+type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber> & { valueString?: string };
```
saravanan7mani, line 66:
```suggestion
export type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber> & { valueString?: string };
```

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -64,6 +64,17 @@ export class Doge extends AbstractUtxoCoin {
   async signTransaction<TNumber extends number | bigint = bigint>(
     params: SignTransactionOptions<TNumber>
   ): Promise<SignedTransaction | HalfSignedUtxoTransaction> {
+    params.txPrebuild.txInfo.unspents = params.txPrebuild.txInfo.unspents.map((unspent) => {
```
saravanan7mani, line 67:
```
  async signTransaction<TNumber extends number | bigint = bigint>(
    params: SignTransactionOptions<TNumber>
  ): Promise<SignedTransaction | HalfSignedUtxoTransaction> {
    this.parseUnspents(params.txPrebuild.txInfo.unspents);
    return super.signTransaction(params);
  }


parseUnspents<TNumber extends number | bigint = bigint>(unspents: Array<Unspent<TNumber>>) {
    unspents.forEach((unspent) => {
      if (unspent.valueString === undefined) {
        throw new Error('missing required field valueString in unspent');
      }
      if (typeof unspent.value === 'number') {
        unspent.value = utxolib.bitgo.toTNumber(unspent.valueString!, 'bigint');
      }
    });
  }
```

saravanan7mani: Simplified.

saravanan7mani (changes requested):

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -63,7 +63,7 @@ import { supportedCrossChainRecoveries } from './config';
 
 const { getExternalChainCode, isChainCode, scriptTypeForChain, outputScripts, toOutput, verifySignatureWithUnspent } =
   bitgo;
-type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber>;
+export type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber> & { valueString?: string };
```
OttoAllmendinger, line 66: We should add a separate type `UnspentJSON` instead, like I suggested here: https://github.com/BitGo/BitGoJS/pull/2785#discussion_r991886762

This change will mix up internal representation and the representation used when communicating with WP

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -52,24 +53,34 @@ export class Doge extends AbstractUtxoCoin {
   async parseTransaction<TNumber extends number | bigint = bigint>(
     params: ParseTransactionOptions<TNumber>
   ): Promise<ParsedTransaction<TNumber>> {
+    if (params?.txPrebuild?.txInfo?.unspents) {
+      this.parseUnspents(params.txPrebuild.txInfo.unspents);
+    }
     return super.parseTransaction<TNumber>(params);
```
OttoAllmendinger, line 59: 1. if avoidable, methods should not mutate the input arguments, especially parse method
2. we should keep doge-specific code at a minimum

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -78,4 +89,15 @@ export class Doge extends AbstractUtxoCoin {
   ): Promise<CrossChainRecoverySigned<TNumber> | CrossChainRecoveryUnsigned<TNumber>> {
     return super.recoverFromWrongChain(params);
   }
+
+  parseUnspents<TNumber extends number | bigint = bigint>(unspents: Unspent<TNumber>[]): void {
```
OttoAllmendinger, line 93: I suggested a variant of this here that has no side effects https://github.com/BitGo/BitGoJS/pull/2785/files#diff-3bc4bb8653ef7242d182520184d5b1923b3363693cfb6ae0df1c4a6ea7f6263dR85

let's use this pattern instead

```
// types used for talking to WP
type UnspentJSON = { valueString: string; /* ... other fields */ }
type TxInfoJSON = { unspents: UnspentJSON[]; /* other fields */ }

// types used internally
type TxInfo<TNumber> = { unspents: Unspent<TNumber>; /* ... */ };

function decodeTxInfo<TNumber extends number | bigint>(txInfo: TxInfoJSON, amountType: 'number' | 'bigint'): TxInfo<TNumber> {
 return { ...txInfo, unspents: decodeUnspents(txInfo.unspents, amountType) };
}

// similarly we can write `encodeTxInfo` etc
```

In `AbstractUtxoCoin` we should then use `UnspentJSON` / `TxInfoJSON` as the type for WP responses and use `decodeTxInfo` / `encodeTxInfo` in `abstractUtxoCoin.ts` to get the WP data into the correct representation

OttoAllmendinger (changes requested):

jicli (dismissed):
LGTM, just one thing to note is that AVAX used to have `bigint` in network and it cause problem when front end update their method for JSON-fy the value. Not sure if this one is going to be the same.

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -10,10 +10,49 @@ import {
...
   CrossChainRecoveryUnsigned,
   RecoverFromWrongChainOptions,
+  Unspent,
+  TransactionInfo,
+  TransactionPrebuild,
 } from '@bitgo/abstract-utxo';
 import { BaseCoin, BitGoBase, HalfSignedUtxoTransaction, SignedTransaction } from '@bitgo/sdk-core';
 import * as utxolib from '@bitgo/utxo-lib';
 
+type UnspentJSON = Unspent<bigint> & { value: number; valueString: string };
```
OttoAllmendinger, line 20:
```suggestion
type UnspentJSON = Unspent<number> & { valueString: string };
```

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -10,10 +10,49 @@ import {
...
+type TransactionInfoJSON = TransactionInfo<bigint> & { unspents: UnspentJSON[] };
```
OttoAllmendinger, line 21:
```suggestion
type TransactionInfoJSON = TransactionInfo<number> & { unspents: UnspentJSON[] };
```

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -10,10 +10,49 @@ import {
...
+type TransactionPrebuildJSON = TransactionPrebuild<bigint> & { txInfo: TransactionInfoJSON };
```
OttoAllmendinger, line 22:
```suggestion
type TransactionPrebuildJSON = TransactionPrebuild<number> & { txInfo: TransactionInfoJSON };
```

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -63,7 +63,7 @@ import { supportedCrossChainRecoveries } from './config';
 
 const { getExternalChainCode, isChainCode, scriptTypeForChain, outputScripts, toOutput, verifySignatureWithUnspent } =
   bitgo;
-type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber>;
+export type Unspent<TNumber extends number | bigint = number> = bitgo.Unspent<TNumber>;
```
OttoAllmendinger, line 66: this change can be reverted

OttoAllmendinger (changes requested):

modules/sdk-coin-doge/src/doge.ts
OttoAllmendinger: it doesn't make a big difference because the parameter is masked by `&` anyway, but `number` is less confusing here

OttoAllmendinger (dismissed):

modules/sdk-coin-doge/src/doge.ts
```diff
@@ -10,13 +10,51 @@ import {
...
+    if (typeof unspent.value === 'bigint') {
+      return unspent as bitgo.Unspent<bigint>;
+    }
+    if ('valueString' in unspent) {
+      return { ...unspent, value: BigInt(unspent.valueString) };
+    }
+    if (typeof unspent.value === 'number') {
+      throw new Error(`received Unspent<number> where Unspent<bigint> or UnspentJSON was expected`);
+    }
+    throw new Error('invalid unspent');
```
sangaman, line 36: This could be a bit more specific to say that the unspent value was not in the right format.

As an aside, why does this method signature accept `TNumber` if it throws on `number` values, shouldn't it just be `Unspent<bigint>` in the parameter signature?

sangaman (dismissed):
LGTM, just one question/nit.

modules/sdk-coin-doge/src/doge.ts
OttoAllmendinger: strictly speaking yes, but TNumber was required to make typescript happy here

dpkjnr: actually we can use bigint instead of TNumber in parse functions atleast. But then we need to pass arguments as `params.txPrebuild as TransactionPrebuildJSON`.  Please check commit https://github.com/BitGo/BitGoJS/pull/2908/commits/53854dc76bd03fe958ac2d6e11eb121d1ce44c50 if it looks better or else I can revert to older commit

OttoAllmendinger: > But then we need to pass arguments as params.txPrebuild as TransactionPrebuildJSON

that would not a legitimate cast though because we cannot expect `txPrebuild` to always be of that type

dpkjnr: alright, reverting to the older commit

dpkjnr:
> LGTM, just one thing to note is that AVAX used to have `bigint` in network and it cause problem when front end update their method for JSON-fy the value. Not sure if this one is going to be the same.

by `bigint in network` you mean passing bigint values to HTTP calls?

OttoAllmendinger (approved):

saravanan7mani (approved):

PR #2914, opened 2022-10-18
implement is wallet address for SUI

lcovar:
## Description

The purpose of this PR is to implement `isWalletAddress` for SUI. The method had to be async due to the use of the MPC library therefore all coins had to be updated to change the `isWalletAddress` method to async and update respective tests. 

First commit is updating all the coin and test to async `isWalletAddress` and the second commit implements that method for sui.ts which is the reason why it needed to be changed to async.

## Issue Number
BG-59148

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeCoin.ts
+++ b/modules/abstract-eth/src/abstractEthLikeCoin.ts
@@ -135,7 +135,7 @@ export abstract class AbstractEthLikeCoin extends BaseCoin {
     return {};
   }
 
-  isWalletAddress(): boolean {
+  async isWalletAddress(): Promise<boolean> {
     throw new MethodNotImplementedError();
   }
 
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -894,7 +894,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
    * @throws {InvalidAddressDerivationPropertyError}
    * @throws {UnexpectedAddressError}
    */
-  isWalletAddress(params: VerifyAddressOptions): boolean {
+  async isWalletAddress(params: VerifyAddressOptions): Promise<boolean> {
     const { address, addressType, keychains, coinSpecific, chain, index } = params;
 
     if (!this.isValidAddress(address)) {
--- a/modules/sdk-coin-sui/src/sui.ts
+++ b/modules/sdk-coin-sui/src/sui.ts
@@ -99,7 +99,7 @@ export class Sui extends BaseCoin {
     return true;
   }
 
-  isWalletAddress(params: VerifyAddressOptions): boolean {
+  async isWalletAddress(params: VerifyAddressOptions): Promise<boolean> {
     throw new MethodNotImplementedError();
   }
 
--- a/modules/sdk-coin-sui/src/sui.ts
+++ b/modules/sdk-coin-sui/src/sui.ts
@@ -2,15 +2,16 @@ import {
   BaseCoin,
   BaseTransaction,
   BitGoBase,
+  EDDSAMethods,
+  InvalidAddressError,
   KeyPair,
-  MethodNotImplementedError,
   MPCAlgorithm,
   ParsedTransaction,
   ParseTransactionOptions,
   SignedTransaction,
   SignTransactionOptions,
   TransactionExplanation,
-  VerifyAddressOptions,
+  TssVerifyAddressOptions,
   VerifyTransactionOptions,
 } from '@bitgo/sdk-core';
 import { BaseCoin as StaticsBaseCoin, coins } from '@bitgo/statics';
@@ -18,6 +19,7 @@ import BigNumber from 'bignumber.js';
 import { Transaction, TransactionBuilderFactory } from './lib';
 import utils from './lib/utils';
 import * as _ from 'lodash';
+import * as sha3 from 'js-sha3';
 
 export interface ExplainTransactionOptions {
   txHex: string;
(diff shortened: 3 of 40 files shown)


modules/abstract-eth/src/abstractEthLikeCoin.ts
```diff
@@ -135,7 +135,7 @@ export abstract class AbstractEthLikeCoin extends BaseCoin {
     return {};
   }
 
-  isWalletAddress(): boolean {
+  async isWalletAddress(): Promise<boolean> {
```
margueriteblair, line 138: bruh did u mean to change all these???

margueriteblair: 40 files guey

ewangbitgo: why we need to make it async?

modules/sdk-core/src/coins/susd.ts
```diff
@@ -62,7 +62,7 @@ export class Susd extends BaseCoin {
     return {};
   }
 
-  isWalletAddress(params: VerifyAddressOptions): boolean {
+  isWalletAddress(params: VerifyAddressOptions): Promise<boolean> {
```
margueriteblair, line 65: why promise if we're not denoting it as async?

modules/abstract-eth/src/abstractEthLikeCoin.ts
lcovar: https://github.com/BitGo/BitGoJS/blob/feafb010da772a8fb454192a44d93bcffcff86a7/modules/sdk-coin-sui/src/sui.ts#L110

isWalletAddress for sui had to be async because of that ^. I had to change the interface to return promise and every coin module implements that interface so that change had to propagate to every coin.

modules/sdk-core/src/coins/susd.ts
lcovar: so that yarn doesnt yell. I'll add async prefix to the ones i missed though.

modules/sdk-coin-sui/src/sui.ts
```diff
@@ -92,8 +95,31 @@ export class Sui extends BaseCoin {
...
+
+    if (!keychains) {
+      throw new Error('missing required param keychains');
+    }
+
+    for (const keychain of keychains) {
+      const MPC = await EDDSAMethods.getInitializedMpcInstance();
+      const commonKeychain = keychain.commonKeychain as string;
+
+      const derivationPath = 'm/' + index;
```
joey-bai, line 113: i think this is going to break for custodial wallets, where the derivation path is randomly generated to look something like:
`m/{rand1}/{rand2}/{addressIndex}`

joey-bai (dismissed):
okay to start, but we may need to revisit this after testing custodial workflow

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
```diff
@@ -241,15 +241,15 @@ export abstract class BaseCoin implements IBaseCoin {
...
+  verifyAddress(params: VerifyAddressOptions): Promise<boolean> {
     return this.isWalletAddress(params);
   }
 
   /**
    * @param params
    * @return true iff address is a wallet address. Must return false if address is outside wallet.
    */
-  abstract isWalletAddress(params: VerifyAddressOptions): boolean;
+  abstract isWalletAddress(params: VerifyAddressOptions): Promise<boolean>;
```
margueriteblair, line 252: If we don't change this to async here, whenever we create a new coin will this method be made *without* async?

margueriteblair (dismissed):

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
lcovar: One would probably have to add async to the method when implementing but async can't be added to abstract methods.

modules/sdk-coin-sui/src/sui.ts
lcovar: Created this ticket to come back to this
BG-59928

joey-bai (dismissed):

OttoAllmendinger (dismissed):
why is the MPC instantiation async to begin with?

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -894,7 +894,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
    * @throws {InvalidAddressDerivationPropertyError}
    * @throws {UnexpectedAddressError}
    */
-  isWalletAddress(params: VerifyAddressOptions): boolean {
```
sangaman, line 897: The `async` keyword is not actually necessary here (or in other files where we've changed the `isWalletAddress` method signatures but aren't actually using `await` within the method).

sangaman (dismissed):
The utxo-related changes LGTM.

sangaman:
> The utxo-related changes LGTM, just one comment about superfluous async usage.

I was wondering this too. Is it using a worker or something?

lcovar:
@sangaman @OttoAllmendinger 

It's async because we have to run this method from the libsodium library when instantiating an MPC instance and needs to be async.

https://github.com/BitGo/BitGoJS/blob/f9f740721a91f8351df40b3b4d89f2c393acd7cf/modules/sdk-core/src/account-lib/mpc/curves/ed25519.ts#L11

lcovar:
Had to fix conflicts. Please re-review.

joey-bai (approved):

PR #2920, opened 2022-10-19
feat(sdk-core): add function to verify wallet signatures for TSS

LeoKotschenreuther:
When we create TSS wallets, BitGo returns two signatures so that
the owners of the user and backup keys are able to verify that
there was no man in the middle meddling with the involved key shares
and gpg keys.
These signatures are signed with BitGo's GPG public key and cover the
following items:
* commonKeychain
* key ID of the user GPG key
* key ID of the backup GPG key
* the derived public portion of the BitGo to User key share
* the derived public portion of the BitGo to Backup key share

Ticket: BG-59151

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/modules/bitgo/example/js/create-tss-wallet.js
+++ b/modules/bitgo/example/js/create-tss-wallet.js
@@ -1,0 +1,50 @@
+/**
+ * Create a multi-sig wallet at BitGo.
+ * This makes use of the convenience function generateWallet
+ *
+ * This tool will help you see how to use the BitGo API to easily create a wallet.
+ * In this form, it creates 2 keys on the host which runs this example.
+ * It is HIGHLY RECOMMENDED that you GENERATE THE KEYS ON SEPARATE MACHINES for real money wallets!
+ *
+ * To perform more advanced features, such as encrypting keys yourself, please look at createWalletAdvanced.js
+ *
+ * Copyright 2022, BitGo, Inc.  All Rights Reserved.
+ */
+
+const BitGoJS = require('bitgo');
+const bitgo = new BitGoJS.BitGo({ env: 'test' });
+
+// TODO: set your access token here
+const accessToken = '';
+
+// TODO: set a label for your new wallet here
+const label = 'Testing Wallet';
+
+// TODO: set your passphrase for your new wallet here
+const passphrase = 'test_wallet_passphrase';
+
+const coin = 'tsol';
+
+// Create the wallet
+async function main() {
+  bitgo.authenticateWithAccessToken({ accessToken });
+
+  const walletOptions = {
+    label,
+    passphrase,
+    multisigType: 'tss',
+  };
+
+  const wallet = await bitgo.coin(coin).wallets().generateWallet(walletOptions);
+
+  const walletInstance = wallet.wallet;
+
+  console.log(`Wallet ID: ${walletInstance.id()}`);
+  console.log(`Receive address: ${walletInstance.receiveAddress()}`);
+
+  console.log('BACK THIS UP: ');
+  console.log(`User keychain encrypted xPrv: ${wallet.userKeychain.encryptedPrv}`);
+  console.log(`Backup keychain xPrv: ${wallet.backupKeychain.prv}`);
+}
+
+main().catch((e) => console.error(e));
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
@@ -1,3 +1,4 @@
+import * as sodium from 'libsodium-wrappers-sumo';
 import * as _ from 'lodash';
 import * as nock from 'nock';
 import * as openpgp from 'openpgp';
@@ -20,7 +21,9 @@ import {
   Eddsa,
   KeyShare,
   Ed25519BIP32,
+  createSharedDataProof,
 } from '@bitgo/sdk-core';
+import { createWalletSignatures } from '../../tss/helpers';
 import { nockSendSignatureShare, nockGetTxRequest, nockCreateTxRequest, nockDeleteSignatureShare, nockSendTxRequest } from './common';
 
 openpgp.config.rejectCurves = new Set();
@@ -30,6 +33,9 @@ describe('TSS Utils:', async function () {
   let MPC: Eddsa;
   let bgUrl: string;
   let tssUtils: TssUtils;
+  let userGpgKey;
+  let backupGpgKey;
+  let bitgoGpgKey;
   let wallet: Wallet;
   let bitgoKeyShare;
   const reqId = new RequestTracer;
@@ -88,17 +94,38 @@ describe('TSS Utils:', async function () {
   before(async function () {
     bitgoKeyShare = await MPC.keyShare(3, 2, 3);
 
-    const bitGoGPGKey = await openpgp.generateKey({
+    userGpgKey = await openpgp.generateKey({
+      userIDs: [
+        {
+          name: 'test',
+          email: 'test@test.com',
+        },
+      ],
+      curve: 'secp256k1',
+    });
+
+    backupGpgKey = await openpgp.generateKey({
+      userIDs: [
+        {
+          name: 'testBackup',
+          email: 'testBackup@test.com',
+        },
+      ],
+      curve: 'secp256k1',
+    });
+
+    bitgoGpgKey = await openpgp.generateKey({
       userIDs: [
         {
           name: 'bitgo',
           email: 'bitgo@test.com',
         },
       ],
+      curve: 'secp256k1',
     });
     const constants = {
       mpc: {
-        bitgoPublicKey: bitGoGPGKey.publicKey,
+        bitgoPublicKey: bitgoGpgKey.publicKey,
       },
     };
 
@@ -136,21 +163,16 @@ describe('TSS Utils:', async function () {
     it('should generate TSS key chains', async function () {
       const userKeyShare = MPC.keyShare(1, 2, 3);
       const backupKeyShare = MPC.keyShare(2, 2, 3);
-      const userGpgKey = await openpgp.generateKey({
-        userIDs: [
-          {
-            name: 'test',
-            email: 'test@test.com',
-          },
-        ],
-        curve: 'secp256k1',
-      });
 
       const nockedBitGoKeychain = await nockBitgoKeychain({
         coin: coinName,
         userKeyShare,
         backupKeyShare,
+        bitgoKeyShare,
         userGpgKey,
+        // reusing the user gpg key as the backup gpg key, i.e. the user is their own the backup provider
+        backupGpgKey: userGpgKey,
+        bitgoGpgKey,
       });
       const nockedUserKeychain = await nockUserKeychain({ coin: coinName });
       await nockBackupKeychain({ coin: coinName });
@@ -185,21 +207,16 @@ describe('TSS Utils:', async function () {
     it('should generate TSS key chains without passphrase', async function () {
       const userKeyShare = MPC.keyShare(1, 2, 3);
       const backupKeyShare = MPC.keyShare(2, 2, 3);
-      const userGpgKey = await openpgp.generateKey({
-        userIDs: [
-          {
-            name: 'test',
-            email: 'test@test.com',
-          },
-        ],
-        curve: 'secp256k1',
-      });
 
       const nockedBitGoKeychain = await nockBitgoKeychain({
         coin: coinName,
         userKeyShare,
         backupKeyShare,
+        bitgoKeyShare,
         userGpgKey,
+        // reusing the user gpg key as the backup gpg key, i.e. the user is their own the backup provider
+        backupGpgKey: userGpgKey,
+        bitgoGpgKey,
       });
       const nockedUserKeychain = await nockUserKeychain({ coin: coinName });
       await nockBackupKeychain({ coin: coinName });
@@ -234,21 +251,16 @@ describe('TSS Utils:', async function () {
 
       const userKeyShare = MPC.keyShare(1, 2, 3);
       const backupKeyShare = MPC.keyShare(2, 2, 3);
-      const userGpgKey = await openpgp.generateKey({
-        userIDs: [
-          {
-            name: 'test',
-            email: 'test@test.com',
-          },
-        ],
-        curve: 'secp256k1',
-      });
 
       const nockedBitGoKeychain = await nockBitgoKeychain({
         coin: coinName,
         userKeyShare,
         backupKeyShare,
+        bitgoKeyShare,
         userGpgKey,
+        // reusing the user gpg key as the backup gpg key, i.e. the user is their own the backup provider
+        backupGpgKey: userGpgKey,
+        bitgoGpgKey,
       });
       const nockedUserKeychain = await nockUserKeychain({ coin: coinName });
       await nockBackupKeychain({ coin: coinName });
@@ -280,24 +292,117 @@ describe('TSS Utils:', async function () {
       should.exist(backupKeychain.encryptedPrv);
     });
 
-    it('should fail to generate TSS key chains', async function () {
+    it('should fail to generate TSS keychains when received invalid number of wallet signatures', async function () {
       const userKeyShare = MPC.keyShare(1, 2, 3);
       const backupKeyShare = MPC.keyShare(2, 2, 3);
-      const userGpgKey = await openpgp.generateKey({
-        userIDs: [
-          {
-            name: 'test',
-            email: 'test@test.com',
-          },
-        ],
-        curve: 'secp256k1',
+
+      const bitgoKeychain = await generateBitgoKeychain({
+        coin: coinName,
+        userKeyShare,
+        backupKeyShare,
+        bitgoKeyShare,
+        userGpgKey,
+        backupGpgKey,
+        bitgoGpgKey,
       });
+
+      const certsString = await createSharedDataProof(bitgoGpgKey.privateKey, userGpgKey.publicKey, []);
+      const certsKey = await openpgp.readKey({ armoredKey: certsString });
+      const finalKey = new openpgp.PacketList();
+      certsKey.toPacketList().forEach((packet) => finalKey.push(packet));
+      // the underlying function only requires two arguments but the according .d.ts file for openpgp has the further
+      // arguments marked as mandatory as well.
+      // Once the following PR has been merged and released we no longer need the ts-ignore:
+      // https://github.com/openpgpjs/openpgpjs/pull/1576
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      bitgoKeychain.walletHSMGPGPublicKeySigs = openpgp.armor(openpgp.enums.armor.publicKey, finalKey.write());
+
+      await tssUtils.verifyWalletSignatures(userGpgKey.publicKey, backupGpgKey.publicKey, bitgoKeychain, '', 1).should.be.rejectedWith('Invalid wallet signatures');
+    });
+
+    it('should fail to generate TSS keychains when wallet signature fingerprints do not match passed user/backup fingerprints', async function () {
+      const userKeyShare = MPC.keyShare(1, 2, 3);
+      const backupKeyShare = MPC.keyShare(2, 2, 3);
+
+      const bitgoKeychain = await generateBitgoKeychain({
+        coin: coinName,
+        userKeyShare,
+        backupKeyShare,
+        bitgoKeyShare,
+        userGpgKey,
+        backupGpgKey,
+        bitgoGpgKey,
+      });
+
+      // using the backup gpg here instead of the user gpg key to simulate that the first signature has a different
+      // fingerprint from the passed in first gpg key
+      await tssUtils.verifyWalletSignatures(backupGpgKey.publicKey, backupGpgKey.publicKey, bitgoKeychain, '', 1).should.be.rejectedWith(`first wallet signature's fingerprint does not match passed user gpg key's fingerprint`);
+
+      // using the user gpg here instead of the backup gpg key to simulate that the second signature has a different
+      // fingerprint from the passed in second gpg key
+      await tssUtils.verifyWalletSignatures(userGpgKey.publicKey, userGpgKey.publicKey, bitgoKeychain, '', 1).should.be.rejectedWith(`second wallet signature's fingerprint does not match passed backup gpg key's fingerprint`);
+    });
+
+    it('should fail to generate TSS keychains when wallet signature is for different key share', async function () {
+      const userKeyShare = MPC.keyShare(1, 2, 3);
+      const backupKeyShare = MPC.keyShare(2, 2, 3);
+      const customBitgoKeyShare = MPC.keyShare(3, 2, 3);
+
+      const bitgoKeychain1 = await generateBitgoKeychain({
+        coin: coinName,
+        userKeyShare,
+        backupKeyShare,
+        bitgoKeyShare,
+        userGpgKey,
+        // reusing the user gpg key as the backup gpg key, i.e. the user is their own backup provider
+        backupGpgKey: userGpgKey,
+        bitgoGpgKey,
+      });
+      const bitgoKeychain2 = await generateBitgoKeychain({
+        coin: coinName,
+        userKeyShare,
+        backupKeyShare,
+        bitgoKeyShare: customBitgoKeyShare,
+        userGpgKey,
+        // reusing the user gpg key as the backup gpg key, i.e. the user is their own backup provider
+        backupGpgKey: userGpgKey,
+        bitgoGpgKey,
+      });
+
+      // using the other bitgo keychains common keychain and walletHSMGPGPublicKeySigs so that the verification of the
+      // commmon keychain passes but fails for the bitgo to user/ backup shares
+      bitgoKeychain1.commonKeychain = bitgoKeychain2.commonKeychain;
+      bitgoKeychain1.walletHSMGPGPublicKeySigs = bitgoKeychain2.walletHSMGPGPublicKeySigs;
+
+      await tssUtils.createUserKeychain(
+        userGpgKey,
+        userKeyShare,
+        backupKeyShare,
+        bitgoKeychain1,
+      ).should.be.rejectedWith('bitgo share mismatch');
+
+      await tssUtils.createBackupKeychain(
+        userGpgKey,
+        userKeyShare,
+        backupKeyShare,
+        bitgoKeychain1,
+      ).should.be.rejectedWith('bitgo share mismatch');
+    });
+
+    it('should fail to generate TSS key chains when common keychains do not match', async function () {
+      const userKeyShare = MPC.keyShare(1, 2, 3);
+      const backupKeyShare = MPC.keyShare(2, 2, 3);
 
       const nockedBitGoKeychain = await nockBitgoKeychain({
         coin: coinName,
         userKeyShare,
         backupKeyShare,
+        bitgoKeyShare,
         userGpgKey,
+        // reusing the user gpg key as the backup gpg key, i.e. the user is their own the backup provider
+        backupGpgKey: userGpgKey,
+        bitgoGpgKey,
       });
       const bitgoKeychain = await tssUtils.createBitgoKeychain(userGpgKey, userKeyShare, backupKeyShare);
       bitgoKeychain.should.deepEqual(nockedBitGoKeychain);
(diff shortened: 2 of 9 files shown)


modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
```diff
@@ -373,3 +376,37 @@ export async function getTSSSignature(
...
+  backupKeyId: string;
+  decryptedShare?: string;
+  verifierIndex?: 1 | 2; // the index of the verifier, 1 means user, 2 means backup
+}): Promise<void> {
+  const rawNotations = await commonVerifyWalletSignature(params);
+
+  const { decryptedShare, verifierIndex } = params;
+
+  if (!decryptedShare || !verifierIndex) {
+    return;
```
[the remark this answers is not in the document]
LeoKotschenreuther: I made them mandatory after our Slack conversation

modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
```diff
@@ -373,3 +376,33 @@ export async function getTSSSignature(
...
+  backupKeyId: string;
+  decryptedShare: string;
+  verifierIndex: 1 | 2; // the index of the verifier, 1 means user, 2 means backup
+}): Promise<void> {
+  const rawNotations = await commonVerifyWalletSignature(params);
+
+  const { decryptedShare, verifierIndex } = params;
+
+  const publicShare =
+    Buffer.from(await sodium.crypto_sign_ed25519_sk_to_pk(Buffer.from(decryptedShare.slice(0, 64), 'hex'))).toString(
```
islamaminBitGo, line 399: please use the function crypto_scalarmult_ed25519_base_noclamp instead of crypto_sign_ed25519_sk_to_pk.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
```diff
@@ -565,18 +668,52 @@ describe('TSS Utils:', async function () {
...
           from: 'bitgo',
           to: 'backup',
-          publicShare: bitgoKeyShare.yShares[2].y + bitgoKeyShare.yShares[2].v + bitgoKeyShare.yShares[2].chaincode,
+          publicShare: params.bitgoKeyShare.yShares[2].y + params.bitgoKeyShare.yShares[2].v + params.bitgoKeyShare.yShares[2].chaincode,
           privateShare: encryptedBitgoToBackupMessage.toString(),
         },
       ],
     };
 
+    const userKeyId = userGpgKeyActual.keyPacket.getFingerprint().padStart(40, '0');
```
Logicwax, line 683: why the use of padding?  we don't currently do that on the HSM side of things, so this may cause mis-alignment?

modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
```diff
@@ -565,18 +668,52 @@ describe('TSS Utils:', async function () {
...
+        { name: 'backupKeyId', value: userKeyId },
+        { name: 'bitgoToUserPublicShare', value: bitgoToUserPublicShare },
+        { name: 'bitgoToBackupPublicShare', value: bitgoToBackupPublicShare },
+      ]
+    );
+
+    return bitgoKeychain;
+  }
+
+  async function nockBitgoKeychain(params: {
```
Logicwax, line 707: is 'nock' a typo for 'mock'?

modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
LeoKotschenreuther: I was following the example from the OVC:
https://github.com/BitGo/offline-vault-console/blob/master/frontend/src/pkg/bitgo/mpc-utils.ts#L46
I will remove the padding from here to match what happens in the HSM.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
LeoKotschenreuther: You gotta ask that the authors of the `nock` package :D
We are calling these helper functions `nockSomething` because they use the `nock` library to setup HTTP level mocks.

modules/sdk-core/src/bitgo/tss/common.ts
```diff
@@ -68,3 +70,43 @@ export async function sendSignatureShare(
...
+  bitgoPub: openpgp.Key;
+  commonKeychain: string;
+  userKeyId: string;
+  backupKeyId: string;
+}): Promise<{ value: ArrayBuffer }[]> {
+  const { walletSignature, bitgoPub, commonKeychain, userKeyId, backupKeyId } = params;
+
+  // it appears that the result for the key that was used to sign is always at index 1
+  // at index 0 is the result for the key that was certified
+  const isValid = (await walletSignature.verifyPrimaryUser([bitgoPub]))[1].valid;
```
islamaminBitGo, line 89: To me, this seems a little risky, is there a more theoretically robust way to check exactly the certification coming from bitgo. I was thinking if we can loop on the pairs (keyid, valid) and find the pair where keyid matches bitgo's keyid, then check if valid is true for that pair.

Logicwax: `.verifyPrimaryUser` is a openpgpjs function, with index 0 being the self-cert of the CA key, and index 1 being the cert of the sub-key (our proof in this context).

unless openpgp.js is bumped and they change this logic, then this should always be the case.    notice that `openpgp.Key` is passed in, so it will check against that public key.   

Are you making the argument that index 0 should be checked for validity?

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -32,6 +34,46 @@ import { KeychainsTriplet } from '../../../baseCoin';
...
+    const walletSignatures = await openpgp.readKeys({ armoredKeys: bitgoKeychain.walletHSMGPGPublicKeySigs });
+    if (walletSignatures.length !== 2) {
+      throw new Error('Invalid wallet signatures');
+    }
+
+    await verifyWalletSignature({
+      walletSignature: walletSignatures[0],
+      commonKeychain: bitgoKeychain.commonKeychain,
+      userKeyId,
+      backupKeyId: userKeyId,
```
Logicwax, line 60: why are you using userKeyId for the backupKeyId?

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -32,6 +34,46 @@ import { KeychainsTriplet } from '../../../baseCoin';
...
+      bitgoPub: bitgoGpgKey,
+      decryptedShare,
+      verifierIndex,
+    });
+
+    await verifyWalletSignature({
+      walletSignature: walletSignatures[1],
+      commonKeychain: bitgoKeychain.commonKeychain,
+      userKeyId,
+      backupKeyId: userKeyId,
```
Logicwax, line 70: same as above

modules/sdk-core/src/bitgo/tss/common.ts
LeoKotschenreuther: I wrote this code so that the signature has to be done with the bitgo pub to be considered valid. If I also consider index 0, I could pass in the user/ backup pub instead of the bitgo pub and would still consider the signature valid.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -32,6 +34,46 @@ import { KeychainsTriplet } from '../../../baseCoin';
...
+    decryptedShare: string,
+    verifierIndex: 1 | 2
+  ): Promise<void> {
+    assert(bitgoKeychain.commonKeychain);
+    assert(bitgoKeychain.walletHSMGPGPublicKeySigs);
+
+    const bitgoGpgKey = await getBitgoGpgPubKey(this.bitgo);
+
+    const userKeyPub = await openpgp.readKey({ armoredKey: userGpgKey.publicKey });
+    const userKeyId = userKeyPub.keyPacket.getFingerprint().padStart(40, '0');
```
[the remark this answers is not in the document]
LeoKotschenreuther: I must have missed this when I removed the other locations with the padding.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
LeoKotschenreuther: Because we haven't built anything yet that allows a third party to store the backup key shares. So far the user always  creates and stores the backup key shares and reuses the user gpg key as the backup gpg key.
Once we actually implement code to allow for a third party KRS we can change this code.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
LeoKotschenreuther: I explained above.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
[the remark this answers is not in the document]
LeoKotschenreuther: Removed the padding, I also checked for more places that I might have missed but didn't find any more.

modules/sdk-core/src/bitgo/tss/common.ts
Logicwax: index 0 should only show that the self-cert of the CA key is valid (gpg always includes a self-certification of the top CA key).

So by passing in bitgoPub, both indexes should return as valid.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
LeoKotschenreuther: Excellent point. I still think that we currently have some valid use cases where the user and backup gpg key are the same but I updated this function so that they are separate.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
```diff
@@ -565,20 +692,55 @@ describe('TSS Utils:', async function () {
...
+    ).toString('hex') + params.bitgoKeyShare.yShares[2].chaincode;
+
+    bitgoKeychain.walletHSMGPGPublicKeySigs = await createWalletSignatures(
+      params.bitgoGpgKey.privateKey,
+      params.userGpgKey.publicKey,
+      params.backupGpgKey.publicKey,
+      [
+        { name: 'commonKeychain', value: bitgoCombined.pShare.y + bitgoCombined.pShare.chaincode },
+        { name: 'userKeyId', value: userKeyId },
+        { name: 'backupKeyId', value: userKeyId },
```
Logicwax, line 724: your passing in userKeyId for backupkeyId, which should fail because on line 720 you are passing in `params.backupGpgKey.publicKey`

modules/bitgo/test/v2/unit/tss/helpers.ts
```diff
@@ -19,3 +20,28 @@ export async function nockGetTxRequest(params: {walletId: string, txRequestId: s
...
     .reply(200, params.response);
 }
+
+export async function createWalletSignatures(
+  privateKeyArmored: string,
+  publicKeyToCertArmoredUser: string,
+  publicKeyToCertArmoredBackup: string,
+  notations: { name: string; value: string }[]
+): Promise<string> {
+  const certsUserString = await createSharedDataProof(privateKeyArmored, publicKeyToCertArmoredUser, notations);
```
Logicwax, line 30: nit: if these are armored strings, then rename to something like "userWalletSigArmored`

modules/sdk-core/src/bitgo/tss/common.ts
```diff
@@ -68,3 +70,51 @@ export async function sendSignatureShare(
...
+  assert(
+    commonKeychain === Buffer.from(rawNotations[0].value).toString(),
+    'wallet signature does not match common keychain'
+  );
+  assert(userKeyId === Buffer.from(rawNotations[1].value).toString(), `wallet signature does not match user key id`);
+  assert(
+    backupKeyId === Buffer.from(rawNotations[2].value).toString(),
+    'wallet signature does not match backup key id'
+  );
+
```
Logicwax, line 118: what about the `userSharePub` and `backupSharePub`?

Remember, from the HSM, there are no uSig proofs.   Only wallet sigs.    which are just public keys of the user and backup party, with attached certifications from the HSM pubkey.   And each of those certs has notation data attached.

modules/bitgo/test/v2/unit/tss/helpers.ts
```diff
@@ -19,3 +20,28 @@ export async function nockGetTxRequest(params: {walletId: string, txRequestId: s
...
+  const certsBackupString = await createSharedDataProof(privateKeyArmored, publicKeyToCertArmoredBackup, notations);
+
+  const certsUserKey = await openpgp.readKey({ armoredKey: certsUserString });
+  const certsBackupKey = await openpgp.readKey({ armoredKey: certsBackupString });
+
+  const finalKey = new openpgp.PacketList();
```
Logicwax, line 36: nit: rename to something like 'mergedWalletKeys` or `mergedWalletSigs`.   because this is not a one key, but rather is merging two pgp keys into one armored blob.

modules/sdk-core/src/bitgo/tss/common.ts
```diff
@@ -68,3 +70,51 @@ export async function sendSignatureShare(
...
+export async function commonVerifyWalletSignature(params: {
+  walletSignature: openpgp.Key;
+  bitgoPub: openpgp.Key;
+  commonKeychain: string;
+  userKeyId: string;
+  backupKeyId: string;
+}): Promise<{ value: ArrayBuffer }[]> {
+  const { walletSignature, bitgoPub, commonKeychain, userKeyId, backupKeyId } = params;
+
+  // The result of verifyPrimaryUser contains two items:
```
Logicwax, line 87: My apologies for providing incorrect info here.    `.verifyPrimaryUser` doesn't interact with subkeys.   It only checks certifications of the main key.

the code you have below:
```
const verificationResult = await walletSignature.verifyPrimaryUser([bitgoPub]);
const isValid = verificationResult.some((result) => result.valid);
```
is fine.   As this will check that the wallet pgp pubkey has at least ONE of it's attached certifications from the bitgo hsm pubkey (which is all we care about).   I mistakenly led you astray by thinking we were dealing with u-value subkeys, which we are not.  sorry about that.   This is only dealing with pgp CA keys certifying other pgp CA keys, and attaching notifcation data.

Logicwax: I would remove the comment here regarding subkeys, as that doesnt apply here

modules/sdk-core/src/bitgo/tss/common.ts
```diff
@@ -68,3 +70,51 @@ export async function sendSignatureShare(
...
+  // * the self-cert of the CA key
+  // * the cert of the subkey (our proof that we are looking for)
+  // By comparing the fingerprints of the walletSignature and the passed bitgoPub we ensure that the passed bitgoPub is
+  // not the CA key (if they are the same, then it is).
+  if (walletSignature.keyPacket.getFingerprint() === bitgoPub.keyPacket.getFingerprint()) {
```
Logicwax, line 92: i really don't see the need for this.   especially after verifying the behavior of the `.verifyPrimaryUser` function

modules/sdk-core/src/bitgo/tss/common.ts
LeoKotschenreuther: The motivation was that we can reuse this function when we implement almost the same for ecdsa. As far as I understand it the code in this common function will remain the same but checking the user/backup share pubs will be different depending on the curve.

Logicwax: ah, fair point.   thanks for the explanation.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
LeoKotschenreuther: I bet the tests were passing because we are still reusing the same gpg key for the user and backup when we create TSS wallets in the SDK. I updated the function to pass the backupKeyId though.

modules/bitgo/test/v2/unit/tss/helpers.ts
LeoKotschenreuther: Renamed

modules/bitgo/test/v2/unit/tss/helpers.ts
LeoKotschenreuther: Done

modules/sdk-core/src/bitgo/tss/common.ts
LeoKotschenreuther: Comment is removed.

modules/sdk-core/src/bitgo/tss/common.ts
LeoKotschenreuther: I still do. If I remove this as well but leave the rest of the code the same then the following test will fail:
https://github.com/BitGo/BitGoJS/blob/79da30b69fb7be7a87d27f2ca463e751c20dda04/modules/bitgo/test/v2/unit/tss/common.ts#L38-L47
What this test is really doing is that a man in the middle is using their own gpg key (key1) to produce a proof over the BitGo pub (key2). This test should fail and it only fails if I keep this line.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -64,6 +110,14 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
 
     const bitGoToUserPrivateShare = await this.decryptPrivateShare(bitGoToUserShare.privateShare, userGpgKey);
 
+    await this.verifyWalletSignatures(
+      userGpgKey.publicKey,
+      userGpgKey.publicKey,
```
Logicwax, line 115: pass in the backupGpgKey.publicKey

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -132,6 +186,14 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
 
     const bitGoToBackupPrivateShare = await this.decryptPrivateShare(bitGoToBackupShare.privateShare, userGpgKey);
 
+    await this.verifyWalletSignatures(
+      userGpgKey.publicKey,
+      userGpgKey.publicKey,
```
Logicwax, line 191: same as above

modules/sdk-core/src/bitgo/tss/common.ts
Logicwax: that's not a MitM attack, that's not using the SDK properly.

the point of `commonVerifyWalletSignature()` is to take a pubkey, such as walletSignature, and verify if the second argument (bitgoPub in this case) certifies that same pubkey.     

in your test, you have key1 certifying key2 (you produce a key2 pub with a cert attached from key1).    then you pass in this certification (by passing in key2 pub with any certs attached), and bitgoPub/certifying key set to key2.   You're essentially asking the function to make sure that key2 has a certifcation from....key2 (and not key1, like you would expect).    Considering that bitgoPub is baked into the SDK, and should always be passed into the function, the only way to abuse this would be to modify this code (which wouldn't be mitm, which happens with the communication between parties).

I....guess I can agree it's good to keep guard rails here.   but if you're that concerned with someone using this code incorrectly, then i would argue that passing in bitgoPub at all should be removed and instead use the global defined bitgo public key.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
LeoKotschenreuther: There is no `backupGpgKey.publicKey` in this function available. Do you want me to change the entire wallet creation flow so that we use separate gpg keys for the user and the backup even though it's the same party?

LeoKotschenreuther: In that case, are you ok if I do this in a separate PR? This change is going to bleed into the ecdsa code as well so I'd rather keep this as a separate work item.

modules/sdk-core/src/bitgo/tss/common.ts
LeoKotschenreuther: I wasn't aware that bitgoPub is baked into the SDK. So far I only saw code that pulls it once from BitGo's `/api/v1/client/constants` API and then caches it.

LeoKotschenreuther: Where can I find the baked in bitgoPub?

Logicwax: wow, I thought it was.   I guess that is only OVC and on platform.

Then yeah, we *should* bake this into SDK then.   that would help solve MitM problems.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
Logicwax: yeah, i'm fine with that, as long as we track this in a ticket and not forget about it.

modules/sdk-core/src/bitgo/tss/common.ts
Logicwax: testnet: https://github.com/BitGo/bitgo-microservices/blob/56dd0f77c6a9878c8c55798ad1b903d6ecba4753/packages/wallet-platform/config/env/testnet.js#L145-L163

production:  https://github.com/BitGo/bitgo-microservices/blob/56dd0f77c6a9878c8c55798ad1b903d6ecba4753/packages/wallet-platform/config/env/production.js#L241-L254

I never liked the idea of pulling these from a remote API, since we should have a trust anchor in the SDK itself.

islamaminBitGo: I think we should look at this function in isolation and think of what it is promising, which is a robust check that walletSig has a non-self certification that is signed by bitgoPub and that the certification contains notation data that matches the wallet data that all parties have received so far in the creation of the wallet. Therefore, although baking bitgoPub would be the safest thing to do, the code as is will be passing it as a parameter, so I believe this check should stay there, otherwise if at some point bitgo's Pub key is baked as a global variable, I agree then that the check would be useless.

islamaminBitGo (dismissed):
LGTM, but please wait for Matt's approval as well.

modules/sdk-core/src/bitgo/tss/common.ts
islamaminBitGo: But also, I just realized this check protects against the case if a MiM intercepts the communication between WP and User,  then replaces walletSig with their own walletSig that is just HSM's public key without even any other certs attached to it. In that case, wouldn't checking walletSig using bitgoPub through verifyPrimaryUser return valid without the above check? Because walletSig.verifyPrimaryUser([bitgoPub]) would return valid being true for the self certification in walletSig since the attacker just passed bitgoPub in the communication.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -32,6 +34,50 @@ import { KeychainsTriplet } from '../../../baseCoin';
...
 
 export class EddsaUtils extends baseTSSUtils<KeyShare> {
+  async verifyWalletSignatures(
+    userGpgPub: string,
+    backupGpgPub: string,
+    bitgoKeychain: Keychain,
+    decryptedShare: string,
+    verifierIndex: 1 | 2
+  ): Promise<void> {
+    assert(bitgoKeychain.commonKeychain);
```
sachushaji, line 44: NIT: can add error descriptions for these asserts

sachushaji (approved):

modules/sdk-core/src/bitgo/tss/common.ts
Logicwax: @islamaminBitGo well I'm making the argument that we SHOULD bake it in.  It's important to have a trust anchor on the SDK itself, as MitM attacks could occur (this is why ovc even includes it)

> But also, I just realized this check protects against the case if a MiM intercepts the communication between WP and User, then replaces walletSig with their own walletSig that is just HSM's public key without even any other certs attached to it.

But the whole point of walletsigs, is that you get two GPG pubkeys and (after verifying the HSM has blessed them with a cert), you then cross-check that those are the gpg keys you've already learned are the two gpg keys to use (one being yourself, if you are the user, and the other being the backup).   Not only do you check that the HSM has certified those two keys, but that the certification itself includes notation data that specifies the two keyID's AND the commonPub (and also u-values).    

So if an attacker swapped out one of the walletsigs / user pubkey with the HSM one, then you should _still_ have a failure because the keyIDs are not lining up (you already know the gpg key of yourself, and you already know the gpg key of the backup provider) so you are able to check.   So if this single `if` statement is making that big of a difference, then we have evidence that our checking of these other things are not being performed correctly.

islamaminBitGo: Ah, exactly so one thing we're missing here in this check is an initial check of whether the two keys in walletSigs are indeed user's and backup's. 

Reiterating here to see if this is one way to move forward: 
Add one check in the function that dissects walletSigs into userGpgKeyWithCertAdded and backupGpgKeyWithCertAdded:
(Assume we're user and do the same thing for backup)
Check:
userGpgKeyWithCertAdded has the same fingerprint as our own key's fingerprint
backupGpgKeyWithCertAdded has the same fingerprint as that of the pub key we hold for backup

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -32,6 +34,50 @@ import { KeychainsTriplet } from '../../../baseCoin';
...
+    const userKeyPub = await openpgp.readKey({ armoredKey: userGpgPub });
+    const userKeyId = userKeyPub.keyPacket.getFingerprint();
+
+    const backupKeyPub = await openpgp.readKey({ armoredKey: backupGpgPub });
+    const backupKeyId = backupKeyPub.keyPacket.getFingerprint();
+
+    const walletSignatures = await openpgp.readKeys({ armoredKeys: bitgoKeychain.walletHSMGPGPublicKeySigs });
+    if (walletSignatures.length !== 2) {
+      throw new Error('Invalid wallet signatures');
+    }
```
islamaminBitGo, line 58: https://github.com/BitGo/BitGoJS/pull/2920/files#r1025889093

@Logicwax so we probably want to add a check here fir walletSignatures[0] and walletSignatures[1] given the ^ discussion.

modules/sdk-core/src/bitgo/tss/common.ts
Logicwax: yup, you got it.

the commonKeychain should also be checked too, but I guess that's done in the other function.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
Logicwax: yep

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
LeoKotschenreuther: Ticket created and DMed to you.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
LeoKotschenreuther: Ticket created for further work.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
LeoKotschenreuther: Just pushed some new code with the requested checks

modules/bitgo/example/js/create-tss-wallet.js
```diff
@@ -0,0 +1,49 @@
...
+ * It is HIGHLY RECOMMENDED that you GENERATE THE KEYS ON SEPARATE MACHINES for real money wallets!
+ *
+ * To perform more advanced features, such as encrypting keys yourself, please look at createWalletAdvanced.js
+ *
+ * Copyright 2022, BitGo, Inc.  All Rights Reserved.
+ */
+
+const BitGoJS = require('bitgo');
+const bitgo = new BitGoJS.BitGo({ env: 'test' });
+const Promise = require('bluebird');
```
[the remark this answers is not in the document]
LeoKotschenreuther: Change pushed and I also added a TS example

sachushaji (approved):
Thanks for addressing the comments

islamaminBitGo (commented):
LGTM, great work! will let Matt give the approval for a merge.

LeoKotschenreuther:
Had to fix another merge conflict

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -26,13 +28,64 @@ import {
...
+
+    if (userKeyId !== walletSignatures[0].keyPacket.getFingerprint()) {
```
Logicwax, line 60: im not opposed to this check (it's a good sanity check), but what should *really* be checked is if `userKeyId` is equal to the `userKeyId=` in the notation data of the signatures from the HSM.    

https://github.com/BitGo/hsm-api/blob/44c9e8295407d6df411e547c63d38a07a9edc6e2/src/api/newsolanakey.js#L299-L314

That's the notation data.    This means that when the HSM certifies the user/backup gpg key, its signing packets that (in addition to standard pgp packets of certifying another key) _also_ include that data.

so just do what you did in `commonVerifyWalletSignature()` where you check against the raw notations.    `userKeyId !== Buffer.from(rawNotations[1].value).toString(),`

LeoKotschenreuther: This function is (transitively) calling `commonVerifyWalletSignature` and passing the user and backup key ids. This means that it's already checking what you are suggesting.
It first calls `verifyWalletSignature` for each signature and passes the user and backup key ids:
https://github.com/BitGo/BitGoJS/blob/0e6840e4b9a89aea30e784e0acede2377937fe6c/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts#L68-L86
`verifyWalletSignature` then calls `commonVerifyWalletSignature` and passes the same params in:
https://github.com/BitGo/BitGoJS/blob/0e6840e4b9a89aea30e784e0acede2377937fe6c/modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts#L413
The checks you are asking for are done here:
https://github.com/BitGo/BitGoJS/blob/0e6840e4b9a89aea30e784e0acede2377937fe6c/modules/sdk-core/src/bitgo/tss/common.ts#L111-L115

Logicwax: Thanks the explanation.   resolving this.

islamaminBitGo (approved):

PR #2929, opened 2022-10-21
fix(core): properly translate tx type to transferToken intent BG-60250

aaldemir:
## Description
This PR fixes the tests and the switch/case statement that check for the intent type set when the tx type is `transfertoken`.

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/mpcUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/mpcUtils.ts
@@ -102,7 +102,7 @@ export abstract class MpcUtils {
   populateIntent(baseCoin: IBaseCoin, params: PrebuildTransactionWithIntentOptions): PopulatedIntent {
     const chain = this.baseCoin.getChain();
 
-    if (!['acceleration', 'fillNonce'].includes(params.intentType)) {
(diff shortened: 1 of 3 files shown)


anshuldoshi:
you'll want to update https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-coin-eth/src/eth.ts#L1629 as well

anshuldoshi (approved):

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
```diff
@@ -134,7 +134,7 @@ export abstract class MpcUtils {
     if (baseCoin.getFamily() === 'eth' || baseCoin.getFamily() === 'polygon') {
       switch (params.intentType) {
         case 'payment':
-        case 'tokenTransfer':
+        case 'transferToken':
```
mmcshinsky-bitgo, line 137: @aaldemir @anshuldoshi  This commit should have had a BREAKING CHANGE comment before merge.

aaldemir: the intentType is set within the SDK (not by the user) - is that still considered a breaking change?

The portion that relies on the input from the user was not changed.

PR #2939, opened 2022-10-24
fix(sdk-coin-near): fix near recover function

alanliu-bitgo:
## Description
Update Near recover function to handle the case when the address is not on the blockchain (has not partake in any transaction before), and handle the case when the fund is insufficient to construct a recover transaction.

## Issue Number

BG-60399

## Type of change

# How Has This Been Tested?

The failure scenarios that caught these bugs are now behaving as expected

### DIFF
--- a/modules/sdk-coin-near/src/near.ts
+++ b/modules/sdk-coin-near/src/near.ts
@@ -347,7 +347,16 @@ export class Near extends BaseCoin {
     for (let i = startIdx; i < numIteration + startIdx; i++) {
       const currPath = `m/${i}`;
       const accountId = MPC.deriveUnhardened(bitgoKey, currPath).slice(0, 64);
-      const availableBalance = new BigNumber(await this.getAccountBalance(accountId, storageAmountPerByte));
+      let availableBalance = new BigNumber(0);
+      try {
+        availableBalance = new BigNumber(await this.getAccountBalance(accountId, storageAmountPerByte));
+      } catch (e) {
+        // UNKNOWN_ACCOUNT error indicates that the address has not partake in any transaction so far, so we will
+        // treat it as a zero balance address
+        if (e.message !== 'UNKNOWN_ACCOUNT') {
+          throw e;
+        }
+      }
       if (availableBalance.toNumber() <= 0) {
         continue;
       }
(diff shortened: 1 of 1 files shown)


modules/sdk-coin-near/src/near.ts
```diff
@@ -366,15 +375,21 @@ export class Near extends BaseCoin {
       const totalGasWithPadding = totalGasRequired.multipliedBy(1.5);
       const feeReserve = BigNumber(Networks[this.network].near.feeReserve);
       const storageReserve = BigNumber(Networks[this.network].near.storageReserve);
-      const netAmount = availableBalance.minus(totalGasWithPadding).minus(feeReserve).minus(storageReserve).toFixed();
+      const netAmount = availableBalance.minus(totalGasWithPadding).minus(feeReserve).minus(storageReserve);
+      if (netAmount.toNumber() <= 0) {
+        throw new Error(
+          `Found address ${i} with non-zero fund but fund is insufficient to support a recover ` +
```
margueriteblair, line 381: Maybe I'm misunderstanding the code, but wouldn't ${i} here only refer to the integer starting index? Or is this how you want it as opposed to the actual address/account id?

margueriteblair (changes requested):

modules/sdk-coin-near/src/near.ts
alanliu-bitgo: Yep thats what we want. this is because WRW will eventually support letting the user specify which address index to start scanning from

modules/sdk-coin-near/src/near.ts
```diff
@@ -495,7 +510,11 @@ export class Near extends BaseCoin {
       },
     });
     if (response.status !== 200) {
-      throw new Error('Account not found');
+      throw new Error('Failed to query account information');
+    }
+    const errorCause = response.body.error?.cause.name;
```
mmcshinsky-bitgo, line 515: Will `cause` always have a `name` property? Wanted to make sure we don't need another `?.`

alanliu-bitgo: according to their doc https://docs.near.org/api/rpc/contracts it should always have a cause.name if theres an error.

margueriteblair (approved):

mmcshinsky-bitgo (approved):

PR #2944, opened 2022-10-26
feat(sdk-core): Allow preBuildTransaction to Accept Wallet Id

colin-y-wu:
Ticket: BG-57866

## Description

Allow WalletId to be passed as argument (under `recipient`) when we build txn

## Issue Number

Ticket: BG-57866

## Type of change

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -55,8 +55,9 @@ export interface BuildTokenEnablementOptions extends PrebuildTransactionOptions 
 export interface PrebuildTransactionOptions {
   reqId?: IRequestTracer;
   recipients?: {
-    address: string;
+    address?: string;
     amount: string | number;
+    walletId?: string;
     tokenName?: string;
     tokenData?: TokenTransferRecipientParams;
   }[];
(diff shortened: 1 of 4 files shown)


byhow (approved):

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -57,6 +57,7 @@ export interface PrebuildTransactionOptions {
   recipients?: {
     address: string;
     amount: string | number;
+    walletId?: string;
```
CapnMigraine, line 60: `address` is currently a required field in `recipients`

I think we should make it optional and add validation to ensure that either `address` or `walletId` is provided (XOR)

jicli: it also depends on what we do in the UI and how we parse it in WP.

If we support only one box field for user to input either address or wallet id, it will be the case that the `address` is used for both address and wallet id, we dont necessary need to add the `walletId` field, but to have address accept both? it would maybe cause some confusion.

CapnMigraine: We'll have a specific input for `walletId` in this case. I'd prefer to have `walletId` as a separate param, so we don't have to worry about address validations breaking if a wallet id is passed.

image

CapnMigraine (approved):

PR #2949, opened 2022-10-27
refactor(utxo-lib): move and rename addToPsbt, addChangeOutputToPsbt

OttoAllmendinger:
See commits for details. Includes breaking changes.

Issue: BG-47824

### DIFF
--- a/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
+++ b/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
@@ -1,14 +1,23 @@
+import { taproot } from 'bitcoinjs-lib';
+
 import { Network } from '../..';
 import { UtxoTransactionBuilder } from '../UtxoTransactionBuilder';
-import { createOutputScript2of3, scriptTypeForChain } from '../outputScripts';
+import {
+  createOutputScript2of3,
+  createPaymentP2tr,
+  createSpendScriptP2tr,
+  scriptTypeForChain,
+  toXOnlyPublicKey,
+} from '../outputScripts';
 import { toOutputScript } from '../../address';
 import { signInput2Of3, verifySignatureWithPublicKeys } from '../signature';
 import { WalletUnspentSigner } from './WalletUnspentSigner';
-import { RootWalletKeys } from './WalletKeys';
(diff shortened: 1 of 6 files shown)


davidkaplanbitgo (approved):
LGTM but @brandonblack should also take a look before merging

modules/utxo-lib/src/bitgo/wallet/WalletOutput.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Add a verifiable wallet output to the PSBT. The output and all data
+ * needed to verify it from public keys only are added to the PSBT.
+ * Typically these are change outputs
+ *
+ * @param psbt the PSBT to add change output to
+ * @param rootWalletKeys keys that will be able to spend the output
+ * @param chain chain code to use for deriving scripts (and to determine script
+ *              type) chain is an API parameter in the BitGo API, and may be
+ *              any valid ChainCode
```
sangaman, line 17: While we're moving code, we might as well get rid of this excess whitespace.

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -1,14 +1,15 @@
 import { Network } from '../..';
```
sangaman, line 1: Should this file be `WalletUnspent.ts`? It seems like it'd go along with the naming convention for the other files within this folder, and would distinguish it from the `Unspent.ts` in the parent dir.

sangaman (approved):
LGTM, couple of minor suggestions.

modules/utxo-lib/src/bitgo/wallet/WalletOutput.ts
sangaman: Oh now with a slightly wider view I see that this whitespace is there to align the description of `chain` and not just some random whitespace. So we can leave as is.

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
OttoAllmendinger: it probably should

PR #2961, opened 2022-10-31
fix: add test paths to tsconfig.json

OttoAllmendinger:
Since the tests are written in Typescript, we want `tsconfig.json` to
include them. This enables IntelliJ to work correctly.

The output now includes a new directory `dist/test/`.
However the keys "main" and "files" in `package.json` expliclity
reference `dist/src` paths, so it should not matter.

Issue: BG-47824

### DIFF
--- a/modules/utxo-lib/test/transaction_builder.spec.ts
+++ b/modules/utxo-lib/test/transaction_builder.spec.ts
@@ -8,8 +8,8 @@ import {
   script as bscript,
   Transaction,
   TransactionBuilder,
-} from '..';
-import { ECPair } from '../src/noble_ecc';
+  ECPair,
+} from '../src';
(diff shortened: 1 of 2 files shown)


sangaman (commented):
In lnd-middleware I believe we accomplish the same thing by having a base `tsconfig.json` that has no `include` (and thus, includes everything including the test files), but then the `build` npm script uses a child `tsconfig.json` in the `src` directory which we specify when building. That way running `build` only deals with files in `src` and the files in `test` don't get emitted to `test` (or compiled at all, so it goes a bit faster).

Just a suggestion, otherwise this LGTM for its stated purpose.

mmcshinsky-bitgo (approved):

PR #2993, opened 2022-11-08
feat(root): use utxolib toOutputScriptTryFormats in isValidAddress

rushilbg:
BG-60666

BREAKING CHANGE: this drops support for the legacy Litecoin format.
Litecoin addresses starting with 3 will not be supported anymore.

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -293,71 +293,20 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   }
 
   /**
-   * Helper to get the version number for an address
-   */
-  protected getAddressVersion(address: string): number | undefined {
-    // try decoding as base58 first
-    try {
-      const { version } = utxolib.address.fromBase58Check(address, this.network);
-      return version;
-    } catch (e) {
-      // try next format
-    }
-
-    // if coin does not support script types with bech32 encoding, do not attempt to parse
-    if (!this.supportsAddressType('p2wsh') && !this.supportsAddressType('p2tr')) {
-      return;
-    }
-
-    // otherwise, try decoding as bech32
-    try {
-      const { version, prefix } = utxolib.address.fromBech32(address);
-      if (_.isString(this.network.bech32) && prefix === this.network.bech32) {
-        return version;
-      }
-    } catch (e) {
-      // ignore errors, just fall through and return undefined
-    }
-  }
-
-  /**
-   * Helper to get the bech32 prefix for an address
-   */
-  protected getAddressPrefix(address: string): string | undefined {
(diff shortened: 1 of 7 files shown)


modules/abstract-utxo/src/config.ts
```diff
@@ -1,5 +1,5 @@
-// Supported cross-chain recovery routes. The coin to be recovered is the index, the valid coins for recipient wallets
-// are listed in the array.
+// Supported cross-chain recovery routes. The object index is the coin to be recovered, and the value is
+// an array of valid coins for recipient wallets of the original accidental/unintentional deposit.
```
OttoAllmendinger, line 2: let's do this in a separate PR

OttoAllmendinger (changes requested):
you should call the PR title "use utxolib toOutputScriptTryFormats in `isValidAddress`"

The fact that ecash addresses are now supported is more of a side effect (I believe this PR also enables cashaddr support for coin `bch`)

modules/abstract-utxo/src/config.ts
rushilbg: Removed

OttoAllmendinger:
Background: we moved all address parsing logic, including support for nonstandard formats, to `utxolib.addressFormat` a while ago. BitGoJS still implemented custom address validation in `isValidAddress()` which is slightly inconsistent.

This change now has these changes for `isValidAddress()` 

* rejects legacy litecoin addresses (3-prefixed)
* accept `bitcoincash:` cashaddrs for the `bch` coin
* accept `ecash:` cashaddrs for the `bcha` coin

OttoAllmendinger (approved):

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -292,72 +292,21 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
-      // ignore errors, just fall through and return undefined
-    }
-  }
-
   /**
    * Check if an address is valid
    * @param address
    * @param forceAltScriptSupport
    */
   isValidAddress(address: string, forceAltScriptSupport = false): boolean {
```
OttoAllmendinger, line 300: let's put support for non-defined formats behind a parameter

```suggestion
  isValidAddress(address: string, param?: { anyFormat: boolean } | /* legacy parameter */ boolean): boolean {
    if (typeof param === 'boolean' && param) {
      throw new Error('deprecated');
    }
    
    const formats = (param && param.anyFormat) ? ['default'] : undefined;
    
     try {
        utxolib.addressFormat.toOutputScriptTryFormats(address, this.network, formats);
        return true;
      } catch (e) {
        return false;
      }
  }  
```

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #2995, opened 2022-11-08
feat(utxo-lib): add tests for getInputUpdate

OttoAllmendinger:
Verify output of PSBT getInputUpdate using existing `parseSignatures`
methods.

Issue: BG-57748


modules/utxo-lib/test/bitgo/psbt/fromHalfSigned.ts
```diff
@@ -0,0 +1,98 @@
...
+              pubkey,
+              signature,
+            },
+          ],
+        };
+      }
+
+      switch (scriptType) {
+        case 'p2shP2pk':
+          break;
```
OttoAllmendinger, line 59: this is still TODO

the implementation of `getInputUpdate()` incorrectly parses this as `p2pkh` https://github.com/BitGo/BitGoJS/blob/fc33e6189f659f64e133af86f70f363022f18b91/modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts#L50

modules/utxo-lib/test/bitgo/psbt/fromHalfSigned.ts
```diff
@@ -0,0 +1,98 @@
...
+      prevOutputs = getPrevOutputs(scriptType, BigInt(1e8), walletKeys.triple);
+      tx = getHalfSignedTransaction2Of3(walletKeys.triple, walletKeys.user, walletKeys.bitgo, scriptType, network, {
+        prevOutputs,
+      });
+    });
+    it('matches expected value', function () {
+      const vin = 0;
+
+      const result = getInputUpdate(tx.clone() /* FIXME: make getInputUpdate non-destructive */, vin, prevOutputs[0]);
+      const parsed = parseSignatureScript(tx.ins[vin]);
```
brandonblack, line 35: Should this be using the other code, or using the fixtures? A future exists where the ParseSignatureScript code is deprecated / moved to utxo-bin, right?

OttoAllmendinger: Or we use `parseSignatureScript` in `getInputUpdate`? They almost do the same thing

brandonblack: My reason for rewriting rather than using `parseSignatureScript` was to get away from dependence on code that's been removed from upstream (classify). I think that reason still stands. So we could merge the two similar implementations, doing away with the use of classify, or consider `parseSignatureScript` deprecated, and plan to maintain and augment just what we need in `fromHalfSigned`

OttoAllmendinger: let's get rid of `classify` in `parseSignatureScript`

OttoAllmendinger: I wanted to do this for a while now https://github.com/BitGo/BitGoJS/pull/3001

modules/utxo-lib/src/bitgo/parseInput.ts
```diff
@@ -403,6 +404,7 @@ const parseP2tr2Of3: InputParser<ParsedSignatureScriptTaproot> = (p) => {
     signatures: match[':signature'] as [Buffer, Buffer],
     controlBlock,
     scriptPathLevel,
+    leafVersion: controlBlock[0] & 0xfe,
```
sangaman, line 407: Might be nice to add a citation as to where this comes from, I found this: https://en.bitcoin.it/wiki/BIP_0341#cite_note-7

Even if you don't add it to code it'll live in this PR.

modules/utxo-lib/test/bitgo/psbt/fromHalfSigned.ts
```diff
@@ -0,0 +1,98 @@
...
+        case 'p2tr':
+          assert(parsed.scriptType === scriptType);
+          const { redeemScript, witnessScript } = createOutputScript2of3(walletKeys.publicKeys, scriptType);
+          const parsedSignatures = parsed.signatures.filter((s) => !isPlaceholderSignature(s));
+          const signedBy = walletKeys.publicKeys.filter((k) => verifySignatureWithPublicKey(tx, vin, prevOutputs, k));
+
+          assert.strictEqual(parsedSignatures.length, 1);
+          assert.strictEqual(signedBy.length, 1);
+          assert(Buffer.isBuffer(parsedSignatures[0]));
+          assert.deepStrictEqual(
```
sangaman, line 72: I'm kinda surprised this isn't a set of fixtures as well, rather than generating the expected result dynamically. Any particular reason?

sangaman (approved):

modules/utxo-lib/test/bitgo/psbt/fromHalfSigned.ts
OttoAllmendinger: The reason is that I want to replace the implementation of `getInputUpdate` with one that is based on `parseTransaction` - call it implementation-driven development ;)

modules/utxo-lib/src/bitgo/parseInput.ts
OttoAllmendinger: I'll add it in the follow-up

PR #3001, opened 2022-11-09
feat(utxo-lib): remove dependency on `classify` in parseSignatureScript

OttoAllmendinger:
Our current script parsing code consists of

- decompiling a script
- checking the classification
- checking the length
- picking certain parts and validating them
- potential recursion for nested scripts

Doing this manually with hand-written conditionals and destructuring
is hard to read and error prone. Further we do not want to depend on
the `classify` methods any longer.

This change introduces a simple recursive pattern matching
implementation that can deal with the handful of scripts we need to
work with. It only depends on `opcodes` and `script.decompile`.

As a side benefit, the new implementation reduced the execution time
by ~5s on my machine.

Issue: BG-57748

### DIFF
--- a/modules/utxo-lib/src/bitgo/parseInput.ts
+++ b/modules/utxo-lib/src/bitgo/parseInput.ts
@@ -122,16 +122,31 @@ type MatchResult = {
   ':script': { buffer: Buffer; match: MatchResult }[];
 };
 
+class MatchError extends Error {
+  // this property is required to prohibit `return new Error()` when the return type demands `MatchError`
+  __type = 'MatchError';
+  constructor(message: string) {
+    super(message);
+  }
+
+  static forPatternElement(p: ScriptPatternElement): MatchError {
+    if (typeof p === 'object' && ':script' in p) {
+      return new MatchError(`error matching nested script`);
+    }
+    return new MatchError(`error matching ${p}`);
+  }
+}
+
 /**
  * @param script
  * @param pattern
  * @return MatchResult if script matches pattern. The result will contain the matched values.
  */
-function matchScript(script: DecompiledScript, pattern: ScriptPatternElement[]): MatchResult | undefined {
+function matchScript(script: DecompiledScript, pattern: ScriptPatternElement[]): MatchResult | MatchError {
   /**
    * Match a single script element with a ScriptPatternElement
    */
-  function matchElement(e: Buffer | number, p: ScriptPatternElement): boolean | MatchResult {
+  function matchElement(e: Buffer | number, p: ScriptPatternElement): MatchResult | boolean {
     switch (p) {
       case 'OP_0':
         return e === opcodes.OP_0 || (Buffer.isBuffer(e) && e.length === 0);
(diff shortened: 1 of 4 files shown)


modules/utxo-lib/src/bitgo/parseInput.ts
```diff
@@ -0,0 +1,418 @@
...
+  inputClassification: 'taproot';
+  publicKeys: [Buffer, Buffer];
+  signatures: [Buffer, Buffer];
+  controlBlock: Buffer;
+  /** Indicates the level inside the taptree. */
+  scriptPathLevel: number;
+  pubScript: Buffer;
+}
+
+export type ParsedSignatureScriptTaproot = ParsedSignatureScriptTaprootKeyPath | ParsedSignatureScriptTaprootScriptPath;
```
OttoAllmendinger, line 92: moved from `signature.ts`

modules/utxo-lib/src/bitgo/parseInput.ts
```diff
@@ -0,0 +1,418 @@
...
+        return e === opcodes.OP_0 || (Buffer.isBuffer(e) && e.length === 0);
+      case 'OP_1':
+      case 'OP_2':
+      case 'OP_3':
+      case 'OP_CHECKMULTISIG':
+      case 'OP_CHECKSIG':
+      case 'OP_CHECKSIGVERIFY':
+        return e === opcodes[p];
+      case ':pubkey':
+        return Buffer.isBuffer(e) && (e.length === 32 || e.length === 33);
```
brandonblack, line 146: Wonder if we should split pubkey and xonlypubkey as they do in most of the bips.

modules/utxo-lib/src/bitgo/parseInput.ts
```diff
@@ -0,0 +1,418 @@
...
+      case ':signature':
+        return Buffer.isBuffer(e) || isPlaceholderSignature(e);
```
brandonblack, line 148: Wonder if we should do a bit more signature format checking here (valid encoded sig with sighash, or valid raw compact sig with or without sighash)

modules/utxo-lib/src/bitgo/parseInput.ts
```diff
@@ -0,0 +1,418 @@
...
+type ScriptPatternCapture = ':pubkey' | ':signature' | ':buffer' | { ':script': ScriptPatternElement[] };
+
+type ScriptPatternElement = ScriptPatternConstant | ScriptPatternCapture;
+
+/**
+ * Result for a successful script match
+ */
+type MatchResult = {
+  ':pubkey': Buffer[];
+  ':buffer': Buffer[];
```
brandonblack, line 120: Since `:buffer` is only used for a controlblock, should we make the match a `:controlblock` that must be 33+32m in length?

brandonblack:
This is a cool, typesafe and extensible way to handle this. When we integrate it with the new PSBT code, I wonder if we'll want to export some if the individual parsers rather than just trying them all, since in PSBT context we have the prevout to inform which parsers to try.

OttoAllmendinger:
I might add the tighter constraints in the future. Right now I left them kind of loose because errors like invalid signatures etc are caught later anyway, with a good error message. The code here currently uses a lot of `return undefined` to signal mismatches, which isn't great to signal the more subtle validation errors.

OttoAllmendinger:
ok adding a `MatchError` return was pretty simple after all

brandonblack (approved):

PR #3005, opened 2022-11-10
refactor(utxo-lib): deprecate p2pkh parsing

OttoAllmendinger:
This was only needed in a specific test that parsed and encoded fully 
signed deposit transactions. We actually do not need to support p2pkh in
our Psbt transactions.

Issue: BG-57748

### DIFF
--- a/modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
+++ b/modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
@@ -47,14 +47,7 @@ export function getInputUpdate(tx: UtxoTransaction<bigint>, vin: number, prevOut
       throw new Error('Invalid scriptSig, failed to decompile');
     }
     if ((!input.witness || !input.witness.length) && decompiledScriptSig.length === 2) {
-      // P2PKH
-      const pubkey = decompiledScriptSig.pop();
-      const signature = decompiledScriptSig.pop();
-      if (!Buffer.isBuffer(pubkey) || !Buffer.isBuffer(signature)) {
-        throw new Error('Invalid pubkey or signature');
-      }
-      update.partialSig = [{ pubkey, signature }];
-      return update;
+      throw new Error();
     }
     redeemScript = decompiledScriptSig.pop();
     if (!Buffer.isBuffer(redeemScript)) {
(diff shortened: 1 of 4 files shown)


modules/utxo-lib/test/transaction_util.ts
```diff
@@ -56,24 +64,26 @@ export function parseTransactionRoundTrip<TNumber extends number | bigint, T ext
...
+      assert.strictEqual(
+        createPsbtFromBuffer(createPsbtFromTransaction(bigintTx, bigintInputs).toBuffer(), network)
+          .finalizeAllInputs()
+          .extractTransaction()
+          .toBuffer()
+          .toString('hex'),
+        buf.toString('hex')
+      );
+    }
     // Test UtxoPsbt.toBuffer() and UtxoPsbt.fromBuffer() implementation
```
sangaman, line 86: Should this comment move up a few lines closer to the `createPsbtFromBuffer()` call?

sangaman (dismissed):
LGTM, just one nit.

modules/utxo-lib/test/transaction_util.ts
OttoAllmendinger: oops indeed

modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
```diff
@@ -47,14 +47,7 @@ export function getInputUpdate(tx: UtxoTransaction<bigint>, vin: number, prevOut
...
     if ((!input.witness || !input.witness.length) && decompiledScriptSig.length === 2) {
-      // P2PKH
-      const pubkey = decompiledScriptSig.pop();
-      const signature = decompiledScriptSig.pop();
-      if (!Buffer.isBuffer(pubkey) || !Buffer.isBuffer(signature)) {
-        throw new Error('Invalid pubkey or signature');
-      }
-      update.partialSig = [{ pubkey, signature }];
-      return update;
+      throw new Error();
```
byhow, line 50: Should we have some more context on why this is throwing an emtpy error? Maybe message saying due to deprecation or some other reasoning behind this?

byhow (commented):
Just a nit. If this is a breaking change to existing APIs, the commit message needs to have `BREAKING CHANGE` in it, but if not, everything else LGTM

modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
OttoAllmendinger: I will get rid of this branch in the next PR

sangaman (approved):
I'm not sure if this should be considered a breaking change. IIUC it's removing fragments of support for a script type that was never truly supported and is not used by BitGo at all, and it does it within `bitgo` scoped code only.

PR #3011, opened 2022-11-11
added date params to getInvoice query

jamalchowdhury-bitgo-github:
Ticket: BG-61323


sangaman (changes requested):
Thank you, just one nit for how we declare `queryParams` (old code but we might as well address it while we're here).

sangaman (approved):

PR #3012, opened 2022-11-11
refactor(utxo-lib): remove unused props from ParsedSignatureScript

OttoAllmendinger:
They are either unused or redundant with `scriptType`

Remove
- isSegwitInput
- inputClassification
- p2shOutputClassification

BREAKING CHANGE: remove some properties of ParsedSignatureScript Issue:
BG-57748

----

b5ff17588
refactor(utxo-lib): rename parseP2PK -> parseP2shP2pk

More accurate

Issue: BG-57748

----
e9c524390
feat(utxo-lib): parse pubkey and signature in parseP2PK

Issue: BG-57748

----
43d3ce806
refactor(utxo-lib): deprecate p2pkh parsing

This was only needed in a specific tests that parsed end encoded fully 
signed deposit transactions. We actually do not need to support p2pkh in
our Psbt transactions.

Issue: BG-57748
BREAKING CHANGE: deprecate p2pkh parsing


sangaman (approved):
Is removing p2pkh a breaking change if it was never really supported or used anywhere outside of our own tests?

brandonblack:
> Is removing p2pkh a breaking change if it was never really supported or used anywhere outside of our own tests?
Tree forest.

PR #3019, opened 2022-11-15
fix(cspr): fix isWalletAddress

joey-bai:
when creating new CSPR addresses, they are dervived from the wallet's root address. however, the wallet's root address is randomly generated and NOT derived from the user's keychain.

Ticket: BG-56147

Additionally, verified with a test script:
```
async function csprCreateAddress() {
  const wallet = await bitgo.coin('tcspr').wallets().getWallet({ id: '61785623811e590008df8a42cb1f0bb1' });
  const newAddress = await wallet.createAddress();
  console.log(newAddress);
}
```

### DIFF
--- a/modules/sdk-coin-cspr/src/cspr.ts
+++ b/modules/sdk-coin-cspr/src/cspr.ts
@@ -11,7 +11,6 @@ import {
   BitGoBase,
   InvalidAddressError,
   InvalidTransactionError,
-  KeyIndices,
   KeyPair,
   ParsedTransaction,
   ParseTransactionOptions,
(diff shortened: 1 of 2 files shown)


garrett-bitgo (commented):
cool! Thanks for checking this out. I will let someone else approve.

lcovar (dismissed):

CapnMigraine (changes requested):
the test `should verify address with payment id` covers `verifyAddress`. It fails when it calls `isWalletAddress` with the wrong params

CapnMigraine (approved):

PR #3021, opened 2022-11-15
refactor(utxo-lib): rewrite fromHalfSigned to use parseSignature

OttoAllmendinger:
Now we parse p2shP2pk correctly

Issue: BG-57748

7f3b1ae49
feat(utxo-lib): parse redeemScript, witnessScritpt

Required for `getInputUpdate`

Issue: BG-57748

df0be5c29
feat(utxo-lib): add method `getSignaturesWithPublicKeys`

This method returns the actual signature buffers, which are required for
`getInputUpdate`.

Issue: BG-57748

191a4a0e6
feat(utxo-lib): test getInputUpdate for unsigned tx

ef783ee60
feat(utxo-lib): add method hasWitnessData(scriptType)

Issue: BG-57748


modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
```diff
@@ -104,19 +101,8 @@ describe('WalletUnspent', function () {
...
-      isUnspentWithPrevTx(u)
-    );
-    const txBufs = Object.fromEntries(
-      psbt2.getNonWitnessPreviousTxids().map((txid) => {
-        const u = nonWitnessUnspents.find((u) => parseOutputId(u.id).txid === txid);
-        if (u === undefined) throw new Error('No prevtx found');
-        return [txid, u.prevTx];
-      })
-    );
-    psbt2.addNonWitnessUtxos(txBufs);
```
OttoAllmendinger, line 119: @brandonblack do you see any issue with this?

modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
```diff
@@ -1,145 +1,88 @@
...
+        tapLeafScript: [
+          {
+            controlBlock: parsed.controlBlock,
+            script: parsed.pubScript,
+            leafVersion: parsed.leafVersion,
+          },
+        ],
+        tapScriptSig: getPartialSigs().map((obj) => ({
+          ...obj,
+          leafHash: taproot.getTapleafHash(eccLib, parsed.controlBlock, parsed.pubScript),
```
brandonblack, line 71: maybe define `leafHash` before the return statement, rather than re-hashing each loop? I know it's only 1 or 2, but...

modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
brandonblack: If we're going to do this, we should make `createPsbtFromTransaction` populate the `nonWitnessUtxo`s from the prevTx. We don't currently, and the `TxOutput` type doesn't have a field for it.

brandonblack (changes requested):

modules/utxo-lib/test/bitgo/wallet/WalletUnspent.ts
OttoAllmendinger: I actually did add support for it here: https://github.com/BitGo/BitGoJS/commit/b1c91bcda40e4a8caed8302ff0206d3dfce85f9f

modules/utxo-lib/src/bitgo/parseInput.ts
```diff
@@ -30,6 +30,8 @@ export interface ParsedSignatureScript2Of3 extends ParsedSignatureScript {
        For p2sh, the placeholder is OP_0 (number 0) */
     | [Buffer | 0, Buffer | 0, Buffer | 0];
   pubScript: Buffer;
+  redeemScript: Buffer | undefined;
+  witnessScript: Buffer | undefined;
```
sangaman, line 34: Is there any reason these use `| undefined` instead of the `?` optional operator? I guess this goes along w/ the usage of `omitUndefined` below?

modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
```diff
@@ -1,145 +1,86 @@
...
   tx: UtxoTransaction<bigint>,
   vin: number,
-  prevOut: TxOutput<bigint> & { prevTx?: Buffer }
+  prevOuts: (TxOutput<bigint> & { prevTx?: Buffer })[]
 ): PsbtInputUpdate {
-  const input = tx.ins[vin];
-  const update: PsbtInputUpdate = {};
-  if (prevOut.prevTx) {
-    update.nonWitnessUtxo = prevOut.prevTx;
+  const nonWitnessUtxo = prevOuts[vin].prevTx;
```
sangaman, line 17: How do we know this is a non-witness utxo? I realize this is old logic being refactored but I couldn't make sense of it.

modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
```diff
@@ -1,145 +1,86 @@
...
-    }
-    update.redeemScript = redeemScript;
-    if (opcodes.OP_0 !== redeemScript[0] || ![22, 34].includes(redeemScript.length)) {
-      // P2SH -> 2-of-3
-      const hashFn = (hashType) => tx.hashForSignature(vin, redeemScript, hashType, prevOut.value);
-      update.partialSig = matchSignatures(hashFn, redeemScript, decompiledScriptSig);
-      return update;
-    }
+
+  const parsed = parseSignatureScript(tx.ins[vin]);
```
sangaman, line 23:
```suggestion
  const parsedInput = parseSignatureScript(tx.ins[vin]);
```

Makes the code slightly easier to follow imo, particularly further down in the function.

modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
```diff
@@ -1,145 +1,86 @@
...
+              signature,
+            },
+          ]
+        : []
+    );
   }
-  if (input.witness && input.witness.length) {
-    const witness = input.witness;
-    if (witness.length > 2 && witness[witness.length - 1][0] === 0x50) {
-      throw new Error('Annex not supported');
```
sangaman, line 74: Do we want to keep this behavior of throwing if an annex is present?

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -35,10 +35,14 @@ export type VerificationSettings = {
...
-  signedBy: Buffer | undefined;
-};
+export type SignatureVerification =
+  | {
+      /** Set to the public key that signed for the signature */
+      signedBy: Buffer;
+      /** Set to the signature buffer */
+      signature: Buffer;
+    }
+  | { signedBy: undefined; signature: undefined };
```
sangaman, line 45: This type seems a little strange to me, feels like either it should be type above or just `undefined` entirely, not an object with a couple of `undefined` properties. And we can just return `undefined` rather than `{ signedBy: undefined, signature: undefined }` where applicable.

sangaman (commented):
Just a few nits and questions, nice changes overall and easier for me to follow than the previous code.

modules/utxo-lib/src/bitgo/outputScripts.ts
```diff
@@ -21,6 +21,13 @@ export function isScriptType2Of3(t: string): t is ScriptType2Of3 {
 
 export type ScriptType = ScriptTypeP2shP2pk | ScriptType2Of3;
 
+/**
+ * @return true iff scriptType requires witness data
+ */
+export function hasWitnessData(scriptType: ScriptType): scriptType is 'p2shP2wsh' | 'p2wsh' | 'p2tr' {
```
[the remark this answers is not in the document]
OttoAllmendinger: Yes, we can strengthen types even more. I actually went down the rabbit hole and defined

```
type ScriptTypeP2wsh = {
  scriptType: 'p2wsh',
  redeemScript: undefined;
  witnessScript: Buffer;
}

...
```

but I ran into some edge cases. Maybe some other day ;)

modules/utxo-lib/src/bitgo/parseInput.ts
OttoAllmendinger: this forces the property to be set explicitly, avoiding errors by accidental omission

modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
OttoAllmendinger: `parseSignatureScript()` is pretty strict will throw for unrecognized taproot date

modules/utxo-lib/src/bitgo/psbt/fromHalfSigned.ts
OttoAllmendinger: done

sangaman (approved):
Re-approving while it still means something.

modules/utxo-lib/src/bitgo/outputScripts.ts
```diff
@@ -21,6 +21,13 @@ export function isScriptType2Of3(t: string): t is ScriptType2Of3 {
...
+  return ['p2shP2wsh', 'p2wsh', 'p2tr'].includes(scriptType);
```
brandonblack, line 28: I kinda love cases like this where a maximum of 3 3-character comparisons are required to search in an array, so a hashed or tree-type data structure can't beat a linear array search.

modules/utxo-lib/src/bitgo/parseInput.ts
```diff
@@ -289,38 +291,34 @@ function parseP2ms(
...
 };
 
 const parseP2shP2wsh2Of3: InputParser<ParsedSignatureScript2Of3> = (p) => {
   if (!isWrappedSegwit(p)) {
     return new MatchError(`expected wrapped segwit input`);
   }
-  return parseP2ms(p.witness, {
-    scriptType: 'p2shP2wsh',
-  });
+  return { ...parseP2ms(p.witness, 'p2shP2wsh'), redeemScript: p.script[0] as Buffer };
```
brandonblack, line 314: I wonder if this should have a type check on p.script[0] to detect malformed tx deliberately.

brandonblack (approved):
One nit/question, but fine with merging as is.

modules/utxo-lib/src/bitgo/parseInput.ts
OttoAllmendinger: yup we might want to add this in the future

a malformed script will fail at least at the fullnode so I think it is low-prio

PR #3025, opened 2022-11-16
feat(sdk-coin-sui): support paySui and payAllSui transaction types

linyah9912:
Ticket: BG-61478

Mysten introduced 2 new transaction types `paySui` and `payAllSui`. This PR is to add support for the 2 types.

- `paySui` is only for SUI coin and doesn't require an additional `gasPayment` object. Should be used for SUI withdrawals.
- `payAllSui` is only for SUI coin, send all inputCoins to the recipient, and doesn't require an additional `gasPayment` object. Should be used for SUI consolidation.
- The existing `pay` is for both SUI coin and tokens, and requires an additional `gasPayment` object that's not in the inputCoins. Should be used for token withdrawals and consolidation.

Note: this PR also involves some refactoring (cleaned up transaction.ts and moved constants and bcs related code to new files)


modules/sdk-coin-sui/src/lib/transaction.ts
```diff
@@ -254,46 +201,63 @@ export class Transaction extends BaseTransaction {
...
-    };
+    const txData = this.getTxData();
 
     const dataBytes = bcs.ser('TransactionData', txData).toBytes();
     const serialized = new Uint8Array(TYPE_TAG.length + dataBytes.length);
     serialized.set(TYPE_TAG);
     serialized.set(dataBytes, TYPE_TAG.length);
     if (this._signature !== undefined) {
-      this.serializeSignedTx(this._signature.publicKey, this._signature.signature);
+      this._id = Buffer.from(sha3.sha3_256(serialized), 'hex').toString('base64');
```
joey-bai, line 260: this looks a bit suspicious. before, we needed the signature to generate the ID, but now we don't?

joey-bai (commented):
thank you for refactoring a code into `bcs.ts` and `constants.ts`.
it helps immensely with code organization and readibility.

modules/sdk-coin-sui/src/lib/transaction.ts
linyah9912: The serialized unsigned and signed tx are the same because of SUI's detached signature. The original `serializeSignedTx` was repeating this `serialize` and setting the same value to `this._id`, so I removed it and set `this._id` here only if the signature exists. If the signature doesn't exist, `this._id` is always UNAVAILABLE.

modules/sdk-coin-sui/src/lib/transaction.ts
```diff
@@ -302,31 +266,79 @@ export class Transaction extends BaseTransaction {
     const data = Buffer.from(serializedHex, 'hex');
     const trimmedData = new Uint8Array(data.subarray(TYPE_TAG.length));
     const k = bcs.de('TransactionData', trimmedData);
-    this.updateProperTransactionData(k);
+
+    let type: SuiTransactionType;
+    const txDetails: TxDetails = k.kind.Single;
+    if (txDetails.hasOwnProperty('Pay')) {
```
lcovar, line 272: :nit:

one way to avoid chained if-else is to put the accepted types into some sort of set and check if the deserialized type is within that set, otherwise throw an error.

lcovar (approved):

joey-bai (approved):

modules/sdk-coin-sui/src/lib/transaction.ts
```diff
@@ -302,31 +266,79 @@ export class Transaction extends BaseTransaction {
...
+      gasPayment: {
+        objectId: utils.normalizeHexId(k.gasPayment.objectId),
+        version: k.gasPayment.version.toNumber(),
+        digest: k.gasPayment.digest,
+      },
     };
   }
 
-  static updateProperTransactionData(k: any): void {
+  static getProperTxDetails(
```
mmcshinsky-bitgo, line 302: Can we get documentation for this?

mmcshinsky-bitgo (approved):

PR #3026, opened 2022-11-16
fix: multiple issues with message signing

jeanlarrieux:
Ticket: BG-61620

## Description

- Added a new function in eth.ts to construct final message hash. This is necessary as the `recid` gets handle in a specific way for evm.
- Added a new interface and a function to check if a class implements this interface. This is a better way to check for support instead of adding another method to basecoin.
- Updated getBitgoToUserLatestShare to include the requestType (either RequestType.tx or RequestType.message)
- Added a new Intent called IntentOptionsForMessage that extends IntentOptionsBase as it is necessary to send both the encoded and raw message since security team wants HSM to do their own encoding of the raw message but WP needs encoded message to MPC.verify (since both the sdk and hsm are signing the encoded version)

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -22,7 +22,13 @@ import { drawKeycard } from '../internal/keycard';
 import { Keychain } from '../keychain';
 import { IPendingApproval, PendingApproval } from '../pendingApproval';
 import { TradingAccount } from '../trading/tradingAccount';
-import { inferAddressType, RequestTracer, TxRequest, EddsaUnsignedTransaction, IntentOptionsBase } from '../utils';
+import {
+  inferAddressType,
+  RequestTracer,
+  TxRequest,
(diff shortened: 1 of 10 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2605,6 +2631,7 @@ export class Wallet implements IWallet {
     let unsignedTx: EddsaUnsignedTransaction;
 
     if (txRequest.apiVersion === 'full') {
+      assert(txRequest.transactions);
```
mmcshinsky-bitgo, line 2634: assert won't let the user what went wrong. Can we throw this with a useful message?

Or remove it and add a `?.` to line 2635 if that is all that is needed.

mmcshinsky-bitgo (changes requested):
All the asserts could be try catch with a useful message to the user to help them understand the error, or if the code can continue, in some cases object chaining looks like it would be appropriate.

mmcshinsky-bitgo (approved):

PR #3031, opened 2022-11-18
chore(statics): add generic NFTs for eth, polygon, bsc BG-62198

aaldemir:
Adds generic NFTs to Statics so that `coins.get()` can handle generic NFT names (which are dummy objects meant to be a place holder for unsupported NFTs).


aaldemir:
will squash commits given the build is green

mohammadalfaiyazbitgo (approved):

sachushaji (approved):

modules/statics/src/base.ts
```diff
@@ -143,6 +143,10 @@ export enum CoinFeature {
    * This coin is deprecated
    */
   DEPRECATED = 'deprecated',
+  /**
+   * This coin is a dummy object meant to be a placeholder for an unsupported token
+   */
+  GENERIC = 'generic',
```
mmcshinsky-bitgo, line 149: This field looks too "generic" to be used for tokens on the base definitions.

mmcshinsky-bitgo: Since the verbage everywhere here is "nonstandard", maybe something closer to `TOKEN_NONSTANDARD`?

mmcshinsky-bitgo: I'm not sure I fully understand, but this seems like a work around rather than the needed solution.

aaldemir: `nonstandard` is meant to describe which ERC or smart contract standard the token implements. Generic on the other hand is meant to be an internal descriptor; it has no indication of the smart contract contract standards and only is meant to indicate if a token object represents a specific token or a dummy (catch all) token. The need for this is due the fact that BitGo can "support" all NFTs without having to whitelabel each NFT that comes into a BitGo wallet.

mmcshinsky-bitgo: `GENERIC` here is still too wide of a name for all coins and tokens to represent a dummy token specifically. Needs a more focused name. Something like `TOKEN_GENERIC` feels closer to represent a more niche token feature of coin features.

mmcshinsky-bitgo: Especially since may coins don't have tokens.

aaldemir: The needed solution is likely a larger scope service that is the single source of truth for tokens which is updated/expanded dynamically and automatically as opposed to tokens listed manually across our tech stack. So yes this is definitely more of a work around to allow us to show customers their transfers for NFTs even if the NFT is not "supported".

aaldemir: Ah okay I see what you are saying. I will crate a follow up PR to use a more accurate name.

mmcshinsky-bitgo: Thanks, I appreciate it. It looks like `GENERIC_TOKEN_FEATURES` is used lower so maybe `GENERIC_TOKEN` would match that?

aaldemir: PR to use more semantic naming https://github.com/BitGo/BitGoJS/pull/3033

PR #3047, opened 2022-11-23
fix(sdk-coin-sui): fix sui gasPayment in inputCoins

linyah9912:
Ticket: [BG-62616](https://bitgoinc.atlassian.net/browse/BG-62616)

`Pay` doesn’t allow the gas payment object in inputCoins, but `PaySui` and `PayAllSui` require `gasPayment`=`inputCoins[0]`
This PR is to fix the validation for input coins and gas payment

### DIFF
--- a/modules/sdk-coin-sui/src/lib/transactionBuilder.ts
+++ b/modules/sdk-coin-sui/src/lib/transactionBuilder.ts
@@ -173,21 +173,36 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
   }
 
   protected buildSuiTransaction(): SuiTransaction {
-    assert(this._type, new BuildTransactionError('type is required before building'));
-    assert(this._sender, new BuildTransactionError('sender is required before building'));
-    assert(this._payTx, new BuildTransactionError('payTx is required before building'));
-    assert(this._gasBudget, new BuildTransactionError('gasBudget is required before building'));
-    assert(this._gasPrice, new BuildTransactionError('gasPrice is required before building'));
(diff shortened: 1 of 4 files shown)


modules/sdk-coin-sui/src/lib/transactionBuilder.ts
```diff
@@ -285,31 +280,26 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
-    if (coinIds.includes(this._gasPayment.objectId)) {
+    if (coinIds.includes(this._gasPayment.objectId) && this._type === SuiTransactionType.Pay) {
       throw new BuildTransactionError(
-        `Invalid gas Payment ${this._gasPayment.objectId}, cannot be one of the inputCoins`
+        `Invalid gas payment ${this._gasPayment.objectId}: cannot be one of the inputCoins`
       );
     }
+    if (coinIds[0] !== this._gasPayment.objectId && this._type !== SuiTransactionType.Pay) {
+      throw new BuildTransactionError(`Invalid gas payment: must use the first input coin as the gas payment`);
+    }
```
joey-bai, line 302: does it make sense to re-order the input coins so the gasPayment object is always first for PaySui and PayAllSui instead of erroring out?

linyah9912: The re-order logic is in bgms, and it should pass the correct inputs to SDK.
But now thinking of this and consider external clients using SDK directly, I should update it to re-order instead of errorring out

linyah9912: Updated!

joey-bai (approved):

PR #3050, opened 2022-11-23
feat(bitgo): sdk script to get token balance from wallet

margueriteblair:
A script to get token balances from a wallet. We currently only have a script that can get native asset balances.
Ticket: BG-62315

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -497,6 +497,7 @@ export interface WalletData {
   };
   multisigType: 'onchain' | 'tss';
   type?: WalletType;
+  tokens?: Record<string, any>[];
 }
 
 export interface RecoverTokenOptions {
--- a/modules/bitgo/example/js/get-wallet-token-balance.js
+++ b/modules/bitgo/example/js/get-wallet-token-balance.js
@@ -1,0 +1,33 @@
+/**
+ * Get the token balance of a multi-sig wallet at BitGo.
+ * This makes use of the convenience function wallets().get()
+ *
+ * This tool will help you see how to use the BitGo API to easily get
+ * information about a wallet.
+ *
+ * Copyright 2022, BitGo, Inc.  All Rights Reserved.
+ */
+
+const BitGoJS = require('bitgo');
+const bitgo = new BitGoJS.BitGo({ env: 'test' });
+const Promise = require('bluebird');
+
+const coin = 'tsol';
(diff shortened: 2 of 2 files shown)


alia-bitgo (approved):
lgtm :)

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -497,6 +497,7 @@ export interface WalletData {
   };
   multisigType: 'onchain' | 'tss';
   type?: WalletType;
+  tokens: Record<string, any>[];
```
mmcshinsky-bitgo, line 500: Will a tokens array always be returned all all coin wallets?

margueriteblair: craaap good catch, forgot to make optional

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -279,6 +279,13 @@ export class Wallet implements IWallet {
     return this._wallet.receiveAddress.address;
   }
 
+  /**
+   * Get a list of tokens and their balances from a wallet
+   */
+  public tokens(): Record<string, any>[] | undefined {
```
mmcshinsky-bitgo, line 285: To make sure, is this a convenience function over `toJSON` or is there another intent as well?

margueriteblair: it's a getter for the tokens because they're not accessible otherwise

margueriteblair: Same way the functions above operate pretty much :)

mmcshinsky-bitgo: But I can also do that via `toJson`, yes?

margueriteblair: oh shoot, yeah you can i see, hadnt seen that in the wallet class before, so i guess it is entirely just a convenience function then

margueriteblair: happy to change that if needed!

mmcshinsky-bitgo: Yeah lets use what we have for now. Thanks!

michaelmcshinsky: You can still leave the tokens type on there. Will help for clarity and documentation

PR #3052, opened 2022-11-24
chore(sdk-coin-avaxp): disable tx type verification

CapnMigraine:
`verifyTransaction` fails when the tx type is `importToC` because the type is not present in the TransactionType enum

Ticket: BG-62009


jicli (approved):

colin-y-wu (approved):
looks good!

PR #3063, opened 2022-11-28
fix: update XLM USD token name

joey-bai:
update name to match chain, AnchorUSD

Ticket: BG-62767

BREAKING CHANGE: updated name of XLM USD token


mmcshinsky-bitgo (changes requested):
Commit footer needs a BREAKING CHANGE: description

mmcshinsky-bitgo (approved):

PR #3073, opened 2022-12-01
feat(sdk-coin-polygon): crossChainRecovery support

mullapudipruthvik:
Ticket: BG-62984
Support transaction building for cross chain recovery


modules/sdk-core/src/account-lib/baseCoin/enum.ts
```diff
@@ -48,6 +48,8 @@ export enum TransactionType {
   // Cross chain transfer at Avax
   Export,
   Import,
+  // Cross chain Recovery for evm like chains
+  RecoveryWalletDeployment,
```
sachushaji, line 52: CrossChainRecovery is more meaningFull?

mullapudipruthvik: I think TransactionType can help to indicate the type of transaction like send,deployment etc.
so accordingly I feel `crossChainRecoveryWalletDeployment` is a bit complex

byhow (approved):

PR #3079, opened 2022-12-02
Bg 62806 eip712 encoding

jeanlarrieux:
[EIP-712 standard](https://eips.ethereum.org/EIPS/eip-712) describe how to encode the data before hashing. This is usually handled by the `@metamask/eth-sig-util` library. Normally we would just call the `eip712hash` function of that library, however that function applies the `keccak` transformation to the encoded data. In our case, we just needed the raw encoding, so we ended up doing everything shy of the `keccak`

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -20,6 +20,10 @@ import {
   SignatureShareType,
   Ecdsa,
   Keychains,
+  TypedData,
+  TypedMessage,
+  MessageTypes,
+  SignTypedDataVersion,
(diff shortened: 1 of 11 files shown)


bitgoAaron (dismissed):
sdk-core changes LGTM, but do not merge until ETH team approves.

olibeyene (approved):
can you update description of the pr?

rlutzbitgo:
Do we care about the socket-security dependency complaint?

jeanlarrieux:
> Do we care about the socket-security dependency complaint?
@bitgoAaron and I spoke about it.

bitgoAaron (approved):

PR #3082, opened 2022-12-05
fix(sdk-coin-sui): add signed data to serialized str

linyah9912:
Ticket: BG-61478


linyah9912:
> surprised that this doesn't cause tests to break. could you please double check and make sure we have tests that cover `txid` generation?

updatd test to confirm txid! 'UNAVAILABLE' for unsigned tx, a tx digest for signed tx

lcovar (approved):

joey-bai (approved):

PR #3083, opened 2022-12-05
feat(sdk-core): derive unhardened method for ecdsa

shiksb:
BG-60727

### DIFF
--- a/modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts
+++ b/modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts
@@ -17,6 +17,7 @@ describe('TSS ECDSA TESTS', function () {
     '4f7e914dc9ec696398675d1544aab61cb7a67662ffcbdb4079ec5d682be565d87c1b2de75c943dec14c96586984860268779498e6732473aed9ed9c2538f50bea0af926bdccc0134',
     'hex',
   );
+  let A: ECDSA.KeyShare, B: ECDSA.KeyShare, C: ECDSA.KeyShare;
   before(async () => {
     const pallierMock = sinon
       .stub(paillierBigint, 'generateRandomKeys')
@@ -32,7 +33,7 @@ describe('TSS ECDSA TESTS', function () {
       .resolves(paillerKeys[1] as unknown as paillierBigint.KeyPair)
       .onCall(5)
       .resolves(paillerKeys[2] as unknown as paillierBigint.KeyPair);
-    const [A, B, C] = await Promise.all([MPC.keyShare(1, 2, 3), MPC.keyShare(2, 2, 3), MPC.keyShare(3, 2, 3)]);
+    [A, B, C] = await Promise.all([MPC.keyShare(1, 2, 3), MPC.keyShare(2, 2, 3), MPC.keyShare(3, 2, 3)]);
 
     // Needs to run this serially for testing deterministic key generation
     // to get specific pallier keys to be assigned
@@ -136,6 +137,28 @@ describe('TSS ECDSA TESTS', function () {
         }
       }
     });
+
+    it('should derive unhardened child keys', async function () {
+      // parent key
+      const aKeyCombine = keyShares[0];
+      const commonKeychain = aKeyCombine.xShare.y + aKeyCombine.xShare.chaincode;
+
+      for (let index = 0; index < 10; index++) {
+        const path = `m/0/0/${index}`;
+
+        const subkey = MPC.keyDerive(A.pShare, [B.nShares[1], C.nShares[1]], path);
+
+        const derive1: string = MPC.deriveUnhardened(commonKeychain, path);
+        const derive2: string = MPC.deriveUnhardened(commonKeychain, path);
+
+        derive1.should.equal(derive2, 'derivation should be deterministic');
+
+        (subkey.xShare.y + subkey.xShare.chaincode).should.equal(
+          derive1,
+          'subkey common keychain should match derived keychain',
+        );
+      }
+    });
   });
 
   describe('ECDSA Signing', async function () {
@@ -165,7 +188,7 @@ describe('TSS ECDSA TESTS', function () {
       ];
     });
 
-    for (let index = 0; index < 8; index++) {
+    for (let index = 0; index < 9; index++) {
       it(`should properly sign the message case ${index}`, async function () {
         // Step One
         // signerOne, signerTwo have decided to sign the message
--- a/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
@@ -230,6 +230,34 @@ export default class Ecdsa {
   }
 
   /**
+   * Derives a child common keychain from common keychain
+   *
+   * @param {commonKeychain} The common keychain as a hex string.
+   * @param {path} The BIP-32 path to derive.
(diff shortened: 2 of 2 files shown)


modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -229,6 +236,34 @@ export default class Ecdsa {
...
   }
 
+  /**
+   * Derives a child common keychain from common keychain
+   *
+   * @param commonKeychain - common keychain as a hex string
+   * @param path - bip32 path
+   * @return {string} derived common keychain as a hex string
+   */
+  deriveUnhardened(commonKeychain: string, path: string): string {
```
brandonblack, line 246: Raising a flag here: We shouldn't be re-implementing BIP32. Just use `import { bip32 } from '@bitgo/utxo-lib';`

This applies to the code already added to `hdTree.ts`. That file should be moved down into an eddsa specific location. For ecdsa the functions are readily available in open source libraries we already depend on.

shiksb: This PR only implements a function to derive a child key from an existing hdTree. I don't have context behind why a BIP32 class was implemented for ECDSA instead of using utxo-lib https://github.com/BitGo/BitGoJS/pull/3029/files
Maybe @johnoliverdriscoll or @sachushaji can provide context here

johnoliverdriscoll: For TSS we need to derive the public and private keys separately. The user can derive the common key themselves in standard BIP32, but each level of derivation must be added to a *share* of the private key, not the full private key (which never exists in a fully constructed form).

johnoliverdriscoll: Just to explicitly state it, existing open source libraries are not compatible with BIP32 derivation for TSS shares.

brandonblack: The unhardened derivation is identical. Seems like the cleaner solution would be to fork the `bip32` library and add the slight difference (which appears to be only that the `sk` portion isn't directly the sk for the combined key) as a mode there, rather than taking on maintenance of a separate bip32 implementation.

johnoliverdriscoll: I don't think forking a much larger library that includes features like serialization and fingerprinting that we don't use here would be a cleaner solution than implementing the simple BIP-32 derivation algorithm. Unless you expect the decade old BIP to be revised anytime soon I don't understand this being an issue of maintenance coste.

brandonblack: It's a 336 line (by sloccount on the .js) library, and we already have forks of several others from the bitcoinjs team (`ecpair`, `bip174`, and `bitcoinjs-lib`).

johnoliverdriscoll: Either way I think this is no longer a relevant argument for this PR. There is no way to use an existing BIP-32 implementation as is because of the secret key derivation for shares. If you want the existing implementation that *is* compatible with TSS scrapped in favor of making an entirely new dependency for this single task, you should make a ticket for it.

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -229,6 +236,34 @@ export default class Ecdsa {
     return shares;
   }
 
+  /**
+   * Derives a child common keychain from common keychain
+   *
+   * @param commonKeychain - common keychain as a hex string
+   * @param path - bip32 path
```
sachushaji, line 243: Kindly make the JSDOC statements follow the same convention as the other functions in the file
```suggestion
   * @param {string} commonKeychain - common keychain as a hex string
   * @param {string} path - bip32 path
```

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -229,6 +236,34 @@ export default class Ecdsa {
...
+    }
+
+    const keychain = Buffer.from(commonKeychain, 'hex');
+
+    const derivedPublicKeychain = Ecdsa.hdTree.publicDerive(
+      {
+        pk: bigIntFromBufferLE(keychain.slice(0, 32)),
+        chaincode: bigIntFromBufferBE(keychain.slice(32)),
+      },
+      path
```
sachushaji, line 258: Kindly add sufficient unit test cases for this new function [here](https://github.com/BitGo/BitGoJS/blob/BG-61874-chain-fi-inc-forwarder-version-issue/modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts#L0-L1)

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -229,6 +236,34 @@ export default class Ecdsa {
...
+  deriveUnhardened(commonKeychain: string, path: string): string {
+    if (Ecdsa.hdTree === undefined) {
+      throw new Error("Can't derive key without HDTree implementation");
+    }
+
+    const keychain = Buffer.from(commonKeychain, 'hex');
+
+    const derivedPublicKeychain = Ecdsa.hdTree.publicDerive(
+      {
+        pk: bigIntFromBufferLE(keychain.slice(0, 32)),
```
sachushaji, line 255: I suppose for Ed25519, the library we use makes of Little Endian for its underlying operations internally and I suppose secp256k1 uses Big Endian, we probably should use bigIntFromBufferBE for this also. @johnoliverdriscoll kindly correct me if I'm wrong

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -229,6 +236,34 @@ export default class Ecdsa {
...
+        chaincode: bigIntFromBufferBE(keychain.slice(32)),
+      },
+      path
+    );
+
+    const derivedPk = bigIntToBufferLE(derivedPublicKeychain.pk, 32).toString('hex');
```
sachushaji, line 261: Here too
```suggestion
    const derivedPk = bigIntToBufferBE(derivedPublicKeychain.pk, 32).toString('hex');
```

modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -136,6 +137,23 @@ describe('TSS ECDSA TESTS', function () {
...
+        const derive1: string = MPC.deriveUnhardened(commonKeychain, path);
+        const subkey = MPC.keyDerive(A.pShare, [B.nShares[1], C.nShares[1]], path);
+        const derive2 = MPC.deriveUnhardened(commonKeychain, path);
+        // const derivedPk = derive1.slice(0, 64);
+
+        derive1.should.equal(derive2, 'derivation should be deterministic');
+        (subkey.xShare.y + subkey.xShare.chaincode).should.equal(derive1);
+      }
+    });
   });
```
shiksb, line 157: not sure why this test is failing on.. any ideas? 🤔 
`(subkey.xShare.y + subkey.xShare.chaincode).should.equal(derive1);`

PR #3085, opened 2022-12-05
Fixes for Zcash PSBTs to work with HSM

brandonblack:
Ideally some of these issues should have been caught in tests here, but it's tricky since here they are mostly round-trip tests, and everything did round-trip properly.

Issues: BG-55848, BG-56123


davidkaplanbitgo (approved):
Noice

PR #3086, opened 2022-12-05
feat(utxo-lib): add psbt getSignatureValidationArray function

saravanan7mani:
To check given pub keys are used to sign given PSBT transaction input.

TICKET: BG-58455

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -218,10 +218,22 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
     return results.reduce((final, res) => res && final, true);
   }
 
-  validateTaprootSignaturesOfInput(inputIndex: number): boolean {
+  validateTaprootSignaturesOfInput(inputIndex: number, pubkey?: Buffer): boolean {
     const input = this.data.inputs[inputIndex];
-    const mySigs = (input || {}).tapScriptSig;
-    if (!input || !mySigs || mySigs.length < 1) throw new Error('No signatures to validate');
+    const tapSigs = (input || {}).tapScriptSig;
+    if (!input || !tapSigs || tapSigs.length < 1) {
+      throw new Error('No signatures to validate');
+    }
+    let mySigs;
+    if (pubkey) {
+      const xOnlyPubkey = toXOnlyPublicKey(pubkey);
+      mySigs = tapSigs.filter((sig) => sig.pubkey.equals(xOnlyPubkey));
+      if (mySigs.length < 1) {
+        throw new Error('No signatures for this pubkey');
+      }
+    } else {
+      mySigs = tapSigs;
+    }
     const results: boolean[] = [];
 
     for (const pSig of mySigs) {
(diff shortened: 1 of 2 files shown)


modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -218,10 +218,18 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
     return results.reduce((final, res) => res && final, true);
   }
 
-  validateTaprootSignaturesOfInput(inputIndex: number): boolean {
+  validateTaprootSignaturesOfInput(inputIndex: number, pubkey?: Buffer): boolean {
     const input = this.data.inputs[inputIndex];
-    const mySigs = (input || {}).tapScriptSig;
-    if (!input || !mySigs || mySigs.length < 1) throw new Error('No signatures to validate');
+    const tapSigs = (input || {}).tapScriptSig;
+    if (!input || !tapSigs || tapSigs.length < 1) throw new Error('No signatures to validate');
```
rushilbg, line 224: Nit: For readability and consistency in the repo.
```suggestion
    if (!input || !tapSigs || tapSigs.length < 1) {
      throw new Error('No signatures to validate');
    }
```

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -218,10 +218,18 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+    let mySigs;
+    if (pubkey) {
+      const xOnlyPubkey = toXOnlyPublicKey(pubkey);
+      mySigs = tapSigs.filter((sig) => sig.pubkey.equals(xOnlyPubkey));
+      if (mySigs.length < 1) throw new Error('No signatures for this pubkey');
```
rushilbg, line 229: Same here

rushilbg (dismissed):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -241,6 +253,24 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
     return results.every((res) => res === true);
```
OttoAllmendinger, line 253: What's interesting is that `results.every()` returns `true` for empty arrays, which IMO is a bug even without this change. Maybe we should change this to `0 < results.length && results.every`

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -241,6 +253,24 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+    const input = checkForInput(this.data.inputs, inputIndex);
+    const isP2tr = input.tapScriptSig?.length;
+    return publicKeys.map((publicKey) => {
+      try {
+        return isP2tr
+          ? this.validateTaprootSignaturesOfInput(inputIndex, publicKey)
+          : this.validateSignaturesOfInput(inputIndex, (p, m, s) => eccLib.verify(m, p, s), publicKey);
+      } catch (err) {
+        return false;
+      }
```
OttoAllmendinger, line 270: Not a fan of this overly broad `catch` here. There are a number of errors that can stem from implementation errors unrelated to sig verification that we want to know about. Without the `throw` in `this.validateTaprootSignaturesOfInput` I believe we can do without the `catch` here.

OttoAllmendinger (changes requested):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
OttoAllmendinger: actually to stay compatible with with https://github.com/BitGo/bitcoinjs-lib/blob/master/ts_src/psbt.ts#L435 let's leave it as-is

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -218,10 +218,22 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
...
+    let mySigs;
+    if (pubkey) {
+      const xOnlyPubkey = toXOnlyPublicKey(pubkey);
+      mySigs = tapSigs.filter((sig) => sig.pubkey.equals(xOnlyPubkey));
+      if (mySigs.length < 1) {
+        throw new Error('No signatures for this pubkey');
+      }
+    } else {
+      mySigs = tapSigs;
+    }
```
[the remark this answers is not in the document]
OttoAllmendinger: actually to be compatible with https://github.com/BitGo/bitcoinjs-lib/blob/master/ts_src/psbt.ts#L435 let's throw instead

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
OttoAllmendinger: this comment remains valid - to be compatible with https://github.com/BitGo/bitcoinjs-lib/blob/master/ts_src/psbt.ts#L435 we should remove the `catch`

saravanan7mani: Errors for unsigned input or input with no matching signature for given public key are caught and returned as false. All other errors, propagating the error.

OttoAllmendinger (approved):

PR #3090, opened 2022-12-06
fix: resolve express and qs dependencies

mmcshinsky-bitgo:
* Resolve github advisories for express and qs
* Advisory: https://github.com/advisories/GHSA-hrpp-h998-j3pp

BG-63378

### DIFF
--- a/modules/express/package.json
+++ b/modules/express/package.json
@@ -45,7 +45,7 @@
     "connect-timeout": "^1.9.0",
     "debug": "^3.1.0",
     "dotenv": "^16.0.0",
-    "express": "^4.16.4",
+    "express": "^4.17.3",
     "lodash": "^4.17.20",
     "morgan": "^1.9.1",
     "superagent": "^3.8.3"
--- a/yarn.lock
+++ b/yarn.lock
@@ -24,38 +24,38 @@
   dependencies:
     "@babel/highlight" "^7.18.6"
 
-"@babel/compat-data@^7.17.7", "@babel/compat-data@^7.19.3", "@babel/compat-data@^7.19.4":
-  version "7.19.4"
-  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.19.4.tgz#95c86de137bf0317f3a570e1b6e996b427299747"
-  integrity sha512-CHIGpJcUQ5lU9KrPHTjBMhVwQG6CQjxfg36fGXl3qk/Gik1WwWachaXFuo0uCWJT/mStOKtcbFJCaVLihC1CMw==
+"@babel/compat-data@^7.17.7", "@babel/compat-data@^7.20.0", "@babel/compat-data@^7.20.1":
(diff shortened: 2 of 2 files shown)


modules/express/package.json
```diff
@@ -73,6 +73,9 @@
     "supertest-as-promised": "https://github.com/BitGo/supertest-as-promised/archive/a7f4b612b9fa090ae33a9616c41862aec2b25c7e.tar.gz",
     "typescript": "^4.2.4"
   },
+  "resolutions": {
```
bitgoAaron, line 76: I don't think npm honors the resolutions field, that might be a yarn specific thing. Instead we'd want to use `overrides` 
https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides.

if we were going to use resolutions, I think that should go in the root package.json, which _is_ managed by yarn.

bitgoAaron (changes requested):
see comments

modules/express/package.json
byhow: override is only supported since 8.3.0 right? for node 14 I think it is npm v6. ref: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

modules/express/package.json
```diff
@@ -73,6 +73,9 @@
     "supertest-as-promised": "https://github.com/BitGo/supertest-as-promised/archive/a7f4b612b9fa090ae33a9616c41862aec2b25c7e.tar.gz",
     "typescript": "^4.2.4"
   },
+  "overrides": {
```
byhow, line 76: only supported since v8.3: https://github.com/npm/cli/releases/tag/v8.3.0

byhow (changes requested):

modules/express/package.json
mmcshinsky-bitgo: We can:
* Put resolution in the root for yarn installers
* Put override in modules/express

In testing with v6.14.17. Looks to resolve correctly.

mmcshinsky-bitgo: `build-docker-express` released for customers pull from the root node_modules so that will be correct as well.

modules/express/package.json
mmcshinsky-bitgo: We can:
* Put resolution in the root for yarn installers
* Put override in modules/express

In testing with v6.14.17. Looks to resolve correctly.

`build-docker-express` released for customers pull from the root node_modules so that will be correct as well.

mmcshinsky-bitgo: Pending `ci` tests (for removed `yarn.lock`), we only need to update `express`.

modules/express/package.json
mmcshinsky-bitgo: Pending `ci` tests (for removed `yarn.lock`), we only need to update `express`.

byhow (approved):

bitgoAaron (approved):

PR #3092, opened 2022-12-06
chore: incremental improvements

mmcshinsky-bitgo:
* Update examples to use modularization

BG-000000


examples/ts/get-maximum-spendable.ts
```diff
@@ -10,23 +10,25 @@
...
-
-// TODO: get the wallet with this id
-const id = '';
+const bitgo = new BitGoAPI({
+  accessToken: process.env.TESTNET_ACCESS_TOKEN,
+  env: 'test',
+});
 
 const coin = 'tbtc';
+bitgo.register(coin, Tbtc.createInstance);
```
sachushaji, line 23: NIT: I was just wondering, we already is loading the register function with an instance of the chain we want to interact and the instance has the name of the chain in the getChain method. May be we could simplify this registration to remove passing coin name altogether?, Seems redundant to me unless there is a specific reason for this

sachushaji (approved):

PR #3094, opened 2022-12-07
feat(sdk-coin-avaxp): implememnt verify signature without private key

jrafael-simtlix:
BG-59774

### DIFF
--- a/modules/sdk-coin-avaxp/src/lib/utils.ts
+++ b/modules/sdk-coin-avaxp/src/lib/utils.ts
@@ -192,11 +192,10 @@ export class Utils implements BaseUtils {
     network: AvalancheNetwork,
     message: BufferAvax,
     signature: BufferAvax,
-    prv: BufferAvax
+    pubkey: BufferAvax
   ): boolean {
     const ky = new KeyPairAvax(network.hrp, network.networkID.toString());
-    ky.importKey(prv);
-    return ky.verify(message, signature);
+    return ky.recover(message, signature).equals(pubkey);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-avaxp/src/lib/utils.ts
```diff
@@ -192,27 +192,26 @@ export class Utils implements BaseUtils {
...
    * Avaxp wrapper to verify signature
    * @param network
    * @param message
    * @param signature
-   * @param prv
+   * @param pubkey
    * @return true if it's verify successful
    */
-  verifySignature(network: AvalancheNetwork, message: Buffer, signature: Buffer, prv: Buffer): boolean {
+  verifySignature(network: AvalancheNetwork, message: Buffer, signature: Buffer, pubkey: Buffer): boolean {
```
mmcshinsky-bitgo, line 209: Missing breaking change documentation in commit.

jrafael-simtlix: Surely, if the method is used... but since the verification in WP is done with the public key, this method is not being used anywhere. However, if you still request, I'll add the breaking changes doc.

mmcshinsky-bitgo: Yes, due to this being an open source repository, these must be recorded as breaking changes for our end users.

mmcshinsky-bitgo:
You'll need to switch the BREAKING CHANGE and put it in the commit footer.

jrafael-simtlix:
Move to PR https://github.com/BitGo/BitGoJS/pull/3095 where scope is changed

PR #3102, opened 2022-12-08
feat(sdk-core): get bitgo public key based on coin and feature flags

pranavjain97:
Ticket: [BG-63193](https://bitgoinc.atlassian.net/browse/BG-63193)

See commit messages for details.

### Test
1. Pointed a tss wallet creation script to use the local BitgoJS with these changes.
2. TSS ECDSA wallet creations working

[BG-63193]: https://bitgoinc.atlassian.net/browse/BG-63193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BG-63193]: https://bitgoinc.atlassian.net/browse/BG-63193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
@@ -20,7 +20,7 @@ import {
   SignatureShareRecord,
   RequestTracer,
   BitgoHeldBackupKeyShare,
-  BackupKeyShare,
+  BackupKeyShare, BitgoGPGPublicKey,
 } from '@bitgo/sdk-core';
 import { keyShares, mockAShare, mockDShare, otherKeyShares } from '../../../fixtures/tss/ecdsaFixtures';
 import { nockSendSignatureShareWithResponse } from './common';
@@ -177,6 +177,20 @@ describe('TSS Ecdsa Utils:', async function () {
       should.not.exist(backupKeyShares.bitGoHeldKeyShares);
     });
 
+    it('should get the correct bitgo gpg key based on coin and feature flags', async function() {
+      const nitroGPGKeypair = await openpgp.generateKey({
+        userIDs: [
+          {
+            name: 'bitgo nitro',
+            email: 'bitgo@test.com',
+          },
+        ],
+      });
+      const nockGPGKey = await nockGetBitgoPublicKeyBasedOnFeatureFlags(coinName, 'enterprise_id', nitroGPGKeypair);
+      const bitgoGpgPublicKey = await tssUtils.getBitgoGpgPubkeyBasedOnFeatureFlags('enterprise_id');
+      should.equal(nockGPGKey.publicKey, bitgoGpgPublicKey.armor());
+    });
+
     it('getBackupEncryptedNShare should get valid encrypted n shares based on provider', async function() {
       // Backup key held by third party
       const bitgoHeldBackupKeyShare = await createIncompleteBitgoHeldBackupKeyShare(userGpgKey, backupKeyShare, bitGoGPGKeyPair);
@@ -241,8 +255,21 @@ describe('TSS Ecdsa Utils:', async function () {
 
     it('should generate TSS key chains when backup provider is BitGo', async function () {
       const backupProvider = 'BitGoKRS';
+
+      const nitroGPGKeypair = await openpgp.generateKey({
+        userIDs: [
+          {
+            name: 'bitgo nitro',
+            email: 'bitgo@test.com',
+          },
+        ],
+      });
+
+      await nockGetBitgoPublicKeyBasedOnFeatureFlags(coinName, 'enterprise_id', nitroGPGKeypair);
+      const bitgoGpgPublicKey = await tssUtils.getBitgoGpgPubkeyBasedOnFeatureFlags('enterprise_id');
+
       const isThirdPartyBackup = tssUtils.isValidThirdPartyBackupProvider('BitGoKRS');
-      const bitgoHeldBackupShares = await createIncompleteBitgoHeldBackupKeyShare(userGpgKey, backupKeyShare, bitGoGPGKeyPair);
+      const bitgoHeldBackupShares = await createIncompleteBitgoHeldBackupKeyShare(userGpgKey, backupKeyShare, nitroGPGKeypair);
       const backupShareHolder: BackupKeyShare = {
         bitGoHeldKeyShares: bitgoHeldBackupShares,
       };
@@ -252,11 +279,11 @@ describe('TSS Ecdsa Utils:', async function () {
         backupKeyShare: backupShareHolder,
         enterprise: undefined,
         isThirdPartyBackup,
-        bitgoPublicGpgKey: bitgoPublicKey,
+        bitgoPublicGpgKey: bitgoGpgPublicKey,
       });
       assert(bitgoKeychain.commonKeychain);
 
-      await nockFinalizeBitgoHeldBackupKeyShare(coinName, bitgoHeldBackupShares, bitgoKeychain.commonKeychain, userKeyShare, bitGoGPGKeyPair, bitgoKeychain);
+      await nockFinalizeBitgoHeldBackupKeyShare(coinName, bitgoHeldBackupShares, bitgoKeychain.commonKeychain, userKeyShare, nitroGPGKeypair, bitgoKeychain);
 
       const userBackupKeyChainPromises = [tssUtils.createUserKeychain({
         userGpgKey,
@@ -266,14 +293,14 @@ describe('TSS Ecdsa Utils:', async function () {
         passphrase: 'passphrase',
         enterprise: undefined,
         isThirdPartyBackup,
-        bitgoPublicGpgKey: bitgoPublicKey,
+        bitgoPublicGpgKey: bitgoGpgPublicKey,
       }), tssUtils.createBackupKeychain({
         userGpgKey,
         userKeyShare,
         backupKeyShare: backupShareHolder,
         bitgoKeychain,
         enterprise: undefined,
-        bitgoPublicGpgKey: bitgoPublicKey,
+        bitgoPublicGpgKey: bitgoGpgPublicKey,
         backupProvider,
       })];
       const [userKeychain, backupKeychain] = await Promise.all(userBackupKeyChainPromises);
@@ -285,7 +312,7 @@ describe('TSS Ecdsa Utils:', async function () {
     });
 
     it('should generate TSS key chains with optional params', async function () {
-      const enterprise = 'enterprise';
+      const enterprise = 'enterprise_id';
       const backupShareHolder: BackupKeyShare = {
         userHeldKeyShare: backupKeyShare,
       };
@@ -338,7 +365,7 @@ describe('TSS Ecdsa Utils:', async function () {
       bitgoKeychain.should.deepEqual(nockedBitGoKeychain);
       const testKeyShares = otherKeyShares;
       const testCasesPromises = [
-        await tssUtils.createParticipantKeychain(
+        tssUtils.createParticipantKeychain(
           userGpgKey,
           bitgoPublicKey,
           1,
@@ -346,7 +373,7 @@ describe('TSS Ecdsa Utils:', async function () {
           testKeyShares[0],
           bitgoKeychain,
           'passphrase').should.be.rejectedWith('Common keychains do not match'),
-        await tssUtils.createParticipantKeychain(
+        tssUtils.createParticipantKeychain(
           userGpgKey,
           bitgoPublicKey,
           1,
@@ -355,7 +382,7 @@ describe('TSS Ecdsa Utils:', async function () {
           bitgoKeychain,
           'passphrase')
           .should.be.rejectedWith('Common keychains do not match'),
-        await tssUtils.createParticipantKeychain(
+        tssUtils.createParticipantKeychain(
           backupGpgKey,
           bitgoPublicKey,
           2,
@@ -364,7 +391,7 @@ describe('TSS Ecdsa Utils:', async function () {
           bitgoKeychain,
           'passphrase')
           .should.be.rejectedWith('Common keychains do not match'),
-        await tssUtils.createParticipantKeychain(
+        tssUtils.createParticipantKeychain(
           backupGpgKey,
           bitgoPublicKey,
           2,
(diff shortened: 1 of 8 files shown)


modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -301,35 +356,44 @@ describe('TSS Ecdsa Utils:', async function () {
...
-      const bitgoKeychain = await tssUtils.createBitgoKeychain(userGpgKey, userKeyShare, backupShareHolder);
+      const bitgoKeychain = await tssUtils.createBitgoKeychain({
+        userGpgKey,
+        userKeyShare,
+        backupKeyShare: backupShareHolder,
+        bitgoPublicGpgKey: bitgoPublicKey,
+      });
       bitgoKeychain.should.deepEqual(nockedBitGoKeychain);
       const testKeyShares = otherKeyShares;
       const testCasesPromises = [
```
vitalir-bg, line 367: are the values here still promises?

modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
```diff
@@ -221,17 +224,19 @@ describe('TSS Utils:', async function () {
       const nockedUserKeychain = await nockUserKeychain({ coin: coinName });
       await nockBackupKeychain({ coin: coinName });
 
-      const bitgoKeychain = await tssUtils.createBitgoKeychain(userGpgKey, userKeyShare, backupKeyShare);
-      const userKeychain = await tssUtils.createUserKeychain(
+      const bitgoKeychain = await tssUtils.createBitgoKeychain({ userGpgKey, userKeyShare, backupKeyShare: backupKeyShare });
```
vitalir-bg, line 227: `backupKeyShare: backupKeyShare` seems to be unnecessary since the name matches the key name

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -183,18 +191,24 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
+    userGpgKey,
+    userKeyShare,
+    backupKeyShare,
+    bitgoKeychain,
+    passphrase,
+    bitgoPublicGpgKey,
+    originalPasscodeEncryptionCode,
+    isThirdPartyBackup = false,
+  }: CreateEcdsaKeychainParams): Promise<Keychain> {
+    assert(bitgoKeychain);
```
vitalir-bg, line 204: is this assertion here only to satisfy TS compiler? If not it would be good to have an explicit error message

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
pranavjain97: They're not lol. I remember adding an `await` because the IDE was having an eslint complaint. I'll remove them.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
pranavjain97: Yes it is to only satisfy the TS compiler

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
```diff
@@ -268,12 +248,10 @@ export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtil
       preview,
     };
 
-    const txRequest = (await this.bitgo
+    return (await this.bitgo
```
byhow, line 251: You don't have to `await` here since `.result()` is already returning a promise

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
```diff
@@ -334,8 +312,22 @@ export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtil
    * Checks whether the third party backup provider is valid/supported
    * @param backupProvider - the backup provider client selected
    */
-  isValidThirdPartyBackupProvider(backupProvider: string | undefined): boolean {
+  public isValidThirdPartyBackupProvider(backupProvider: string | undefined): boolean {
```
byhow, line 315: class functions are public by default

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
```diff
@@ -334,8 +312,22 @@ export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtil
...
+   * combination of coin and the feature flags on the user and their enterprise if set.
+   * @param enterpriseId - enterprise under which user wants to create the wallet
+   */
+  public async getBitgoGpgPubkeyBasedOnFeatureFlags(enterpriseId: string | undefined): Promise<Key> {
+    const response: BitgoGPGPublicKey = await this.bitgo
+      .get(this.baseCoin.url('/tss/pubkey'))
+      .query({ enterpriseId })
+      .result();
+    const bitgoPublicKeyStr = response.publicKey as string;
+    return await readKey({ armoredKey: bitgoPublicKeyStr });
```
byhow, line 331: ditto on `await`

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -222,6 +222,14 @@ export type TxRequest = {
   latest: boolean;
 };
 
+export type CreateKeychainParamsBase = {
+  userGpgKey: SerializedKeyPair<string>;
```
byhow, line 226: This type seems a lot looser than the previous parameter type; not sure if this is intended but a bit more context behind this is helpful

byhow (commented):
Have some questions on the parameter generalization clean up, but overall LGTM

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -177,6 +177,20 @@ describe('TSS Ecdsa Utils:', async function () {
...
+      const nitroGPGKeypair = await openpgp.generateKey({
+        userIDs: [
+          {
+            name: 'bitgo nitro',
+            email: 'bitgo@test.com',
+          },
+        ],
+      });
+      const nockGPGKey = await nockGetBitgoPublicKeyBasedOnFeatureFlags(coinName, 'enterprise', nitroGPGKeypair);
+      const bitgoGpgPublicKey = await tssUtils.getBitgoGpgPubkeyBasedOnFeatureFlags('enterprise');
```
vitalir-bg, line 190: 'enterprise' value here is confusing since it looks like a magic value. It would be more readable if it looked more like an enterprise id.

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -241,8 +255,21 @@ describe('TSS Ecdsa Utils:', async function () {
...
+        userIDs: [
+          {
+            name: 'bitgo nitro',
+            email: 'bitgo@test.com',
+          },
+        ],
+      });
+
+      await nockGetBitgoPublicKeyBasedOnFeatureFlags(coinName, 'enterprise', nitroGPGKeypair);
+      const bitgoGpgPublicKey = await tssUtils.getBitgoGpgPubkeyBasedOnFeatureFlags('enterprise');
```
vitalir-bg, line 269: same here

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -663,6 +690,20 @@ describe('TSS Ecdsa Utils:', async function () {
     };
   }
 
+  async function nockGetBitgoPublicKeyBasedOnFeatureFlags(coin: string, enterprise: string, bitgoGpgKeyPair: openpgp.SerializedKeyPair<string>): Promise<BitgoGPGPublicKey> {
+    const bitgoGPGPublicKeyResponse: BitgoGPGPublicKey = {
+      name: 'irrelevant',
+      publicKey: bitgoGpgKeyPair.publicKey,
+      enterpriseId: 'enterprise id',
```
vitalir-bg, line 697: `enterpriseId` here won't match the mocked enterprise id value specified in parameters

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
vitalir-bg: Is this only because we want to use this type for `createBitgoKeychain`?
Maybe to be base CreateKeychainParamsBase shouldn't include this property and be extended when necessary? It is always easier to deal with concrete types, this requires less assertions etc.

vitalir-bg (commented):
Overall looks good to me, just several nits

olibeyene (approved):

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
pranavjain97: I didn't make any changes here. You're suggestion makes sense, but i'd rather not change code not related to this PR.

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
pranavjain97: @byhow are you specifically talking about `userGpgKey: SerializedKeyPair<string>;`, because that's still the same. 

About the need for creating `CreateKeychainParamsBase` - this was because we previously had types for individual fields in a specific order, which wasn't working for all create keychain methods for both eddsa/ecdsa.

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
pranavjain97: Updated it to `enterprise_id`

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
byhow: I am more talking about `CreateKeychainParamsBase` here with the new optional types. 

Also, thanks for explaining the context behind it. I am ok with loosing the types for business needs, but if we can consider @vitalir-bg 's suggestion to extend types to differentiate between different create keychain methods, that will be more helpful. 

If we specifically want to fit the parameters to apply to all create keychain methods at the same time, I am not sure if this is a good approach for different types of creating keychain methods. But if that really is the business need, if this is a breaking change, it needs to follow the commit standards to include breaking change in the commit message.

pranavjain97: It's not really a breaking change. If you look back at the logic of the previous create(User/Backup/Bitgo)Keychain methods-
1. `passphrase` and `originalPasscodeEncryptionCode` were always optional anyway, the only exception being createUserKeychain for ecdsa, for which I already have a concrete error msg if it's not provided. 
2.  About `bitgoKeychain` being optional, that had to be done because there is no bitgoKeychain passed in for `createBitgoKeychain` methods, everywhere else it's needed. I'll add a new type for it so we dont have to add any asserts there. 

See the latest commit for the changes.

pranavjain97: > If we specifically want to fit the parameters to apply to all create keychain methods at the same time
This was what was being done before. I changed that to create a base type, then extended new types for eddsa/ecdsa.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
byhow, line 204: Will this assertion still be needed after making `bitgoKeychain` no optional?

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -214,21 +229,24 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
+  async createBackupKeychain({
+    userGpgKey,
+    userKeyShare,
+    backupKeyShare,
+    bitgoKeychain,
+    bitgoPublicGpgKey,
+    passphrase,
+    backupProvider,
+  }: CreateEcdsaKeychainParams): Promise<Keychain> {
+    assert(bitgoKeychain);
```
byhow, line 241: ditto

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -97,14 +97,15 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
...
-  ): Promise<Keychain> {
+  async createUserKeychain({
+    userGpgKey,
+    userKeyShare,
+    backupKeyShare,
+    bitgoKeychain,
+    passphrase,
+    originalPasscodeEncryptionCode,
+  }: CreateEddsaKeychainParams): Promise<Keychain> {
+    assert(bitgoKeychain);
```
byhow, line 108: ditto

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -174,13 +175,14 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
...
-    passphrase?: string
-  ): Promise<Keychain> {
+  async createBackupKeychain({
+    userGpgKey,
+    userKeyShare,
+    backupKeyShare,
+    bitgoKeychain,
+    passphrase,
+  }: CreateEddsaKeychainParams): Promise<Keychain> {
+    assert(bitgoKeychain);
```
byhow, line 185: ditto

byhow (commented):
just a nit, otherwise LGTM

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
pranavjain97: Nop not needed anymore after making it required. Good catch

byhow (approved):

PR #3112, opened 2022-12-13
fix(sdk-coin-avaxp): fix parseTransaction throwing error for pending

jicli:
approval
Ticket: BG-63957


CapnMigraine (dismissed):
This is a good temporary solution. 
We'll need another fix to implement verification for inputs and outputs
e.g. https://github.com/BitGo/BitGoJS/blob/master/modules/abstract-utxo/src/abstractUtxoCoin.ts#L399

jicli:
@CapnMigraine not every coins implement this method, we should determine if we're interested in this method for Avaxp

CapnMigraine (approved):

PR #3115, opened 2022-12-13
chore(statics): add filecoin erc20 contract

garrett-bitgo:
Ticket: BG-63111

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -615,6 +615,7 @@ export const coins = CoinMap.fromCoins([
   erc20('ff1', 'Two Prime FF1', 18, '0x59af0356cdebd1fa23ae5dadff9170bbfc31278c', UnderlyingAsset.FF1),
   erc20('fft', 'Fight to Fame Token', 18, '0xdea05e09f5b0e102616bb145a0e4772a9b5ab193', UnderlyingAsset.FFT),
   erc20('fida', 'Bonfida Token', 6, '0xf40d9507a7d4850c52a45698c9410e2c345f7a94', UnderlyingAsset.FIDA),
+  erc20('fil', 'Filecoin', 18, '0x0d8ce2a99bb6e3b7db580ed848240e4a0f9ae153', UnderlyingAsset.FIL),
   erc20('fire', 'Ceramic Token', 18, '0x2033e559cddff6dd36ec204e3014faa75a01052e', UnderlyingAsset.FIRE),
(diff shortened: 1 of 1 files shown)


modules/statics/src/coins.ts
```diff
@@ -615,6 +615,7 @@ export const coins = CoinMap.fromCoins([
   erc20('ff1', 'Two Prime FF1', 18, '0x59af0356cdebd1fa23ae5dadff9170bbfc31278c', UnderlyingAsset.FF1),
   erc20('fft', 'Fight to Fame Token', 18, '0xdea05e09f5b0e102616bb145a0e4772a9b5ab193', UnderlyingAsset.FFT),
   erc20('fida', 'Bonfida Token', 6, '0xf40d9507a7d4850c52a45698c9410e2c345f7a94', UnderlyingAsset.FIDA),
+  erc20('fil', 'Filecoin', 18, '0x0d8ce2a99bb6e3b7db580ed848240e4a0f9ae153', UnderlyingAsset.FIL),
```
byhow, line 618: this looks like an EOA address instead of the FIL contract address. Do you have a reference for what this address is used for?

byhow (changes requested):

modules/statics/src/coins.ts
garrett-bitgo: @byhow interesting. I grabbed what I thought was the contract address from here: https://bscscan.com/address/0x0d8ce2a99bb6e3b7db580ed848240e4a0f9ae153.

I assume that this is incorrect contract address?

byhow: I was looking at [this](https://etherscan.io/address/0x0d8ce2a99bb6e3b7db580ed848240e4a0f9ae153) since it shows that this is an `erc20()` token. If this is meant to be the BSC-pegged file coin contract, would you mind using `bscToken()`? They have different explorer/chainIds

garrett-bitgo: In that case, should I remove the `ofcerc20('ofcfil', ...)` declaration in this file as well? I was under the impression that this was an erc20 token according to https://bitgoinc.atlassian.net/browse/BG-63111

byhow: The altcoins/eth or token onboarding team might have more insights on what the expected address here needs to be, but if this is a bscToken instead, I would advise not to use `erc20()` but use that instead for accuracy since they have different network [configs](https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/account.ts#L820). In the case of `ofcerc20`, I am not sure what the implication is since ofc is not really on any of the blockchains. It might help creating a new `ofcbsctoken` or some other initialization functions.

PR #3118, opened 2022-12-15
feat(root): add common registration method

bitgoAaron:
## Description
- makes all coin modules easier to import & register through a common registration pattern
- updates web-demo to demonstrate how to use dynamic import for each coin

## Issue Number

Ticket: BG-64112

## Type of change

# How Has This Been Tested?

- see web-demo /coins page

# Checklist:


modules/web-demo/src/components/Coins/coinFactory.ts
```diff
@@ -0,0 +1,196 @@
+import { BitGoAPI } from '@bitgo/sdk-api';
+import { BaseCoin } from '@bitgo/sdk-core';
+
+class CoinFactory {
```
joey-bai, line 4: should we create a CoinFactory in `sdk-api`?
`sdk-api` already depends on `sdk-core`

mmcshinsky-bitgo:
Am I correct in understanding that you can in essence reverse register coins via the coin package instead of the sdk method for one off coins as well?
```javascript
import { register } from '@bitgo/sdk-coin-btc';

const sdk = new BitGoAPI()
register(sdk);
```

bitgoAaron:
> Am I correct in understanding that you are reverse registering coins via the coin package instead of the sdk method by referencing sdk variable reference?
> 
> ```js
> import { BitGoAPI } from '@bitgo/sdk-api';
> import { register } from '@bitgo/sdk-coin-btc';
> 
> const sdk = new BitGoAPI()
> register(sdk);
> ```
> 
> For multiple coins I'd have to do something like:
> 
> ```js
> import { register as registerBtc } from '@bitgo/sdk-coin-btc';
> import { register as registerLtc } from '@bitgo/sdk-coin-ltc';
> ```

Correct. It's meant to make dynamic imports easier by ensuring every coin has the same method signature for registration. [See the web-demo factory for how we will implement this in bgms](https://github.com/BitGo/BitGoJS/pull/3118/files#diff-ee9de2a4b053356a8a88521062e3151234323f21180129ea8e3f605a960cf737R30)

modules/web-demo/src/components/Coins/coinFactory.ts
bitgoAaron: No, `sdk-api` can't import coin specific modules - it must remain decoupled. This class is purely for demonstration purposes and has no bearing on the published code.

modules/web-demo/src/components/Coins/coinFactory.ts
```diff
@@ -0,0 +1,196 @@
...
+    sdk: BitGoAPI,
+    register: (sdk: BitGoAPI) => void,
+  ) {
+    register(sdk);
+    const coin = sdk.coin(name);
+    this.coinCache.set(name, coin);
+    return coin;
+  }
+
+  async getCoin(name: string, sdk: BitGoAPI): Promise<BaseCoin | undefined> {
```
lifeiscontent, line 22: shouldn't this always return a BaseCoin? if not, it should  throw an error or something, no?

```suggestion
  async getCoin(name: string, sdk: BitGoAPI): Promise<BaseCoin> {
```

OttoAllmendinger (approved):

mmcshinsky-bitgo (approved):

modules/sdk-core/src/bitgo/coinFactory.ts
```diff
@@ -37,9 +37,6 @@ export class CoinFactory {
    * @throws Error
    */
   public register(name: string, coin: CoinConstructor): void {
-    if (this.coinConstructors.has(name)) {
```
mmcshinsky-bitgo, line 40: Do we run the risk of stored run time variable references being lost that may be needed by overwriting the last value in the Map?

bitgoAaron: Not sure I follow, but if some client code has reference to an older coin and then re-registers it, the first reference they have would still be valid. It's up to the client to manage these with the modular approach, we just need to make sure we're not throwing unnecessary exceptions anymore.

PR #3120, opened 2022-12-15
chore: onboard new batch of tokens

aarvay:
Ticket: BG-63937

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -419,6 +419,13 @@ export const coins = CoinMap.fromCoins([
     '0xd829664cdbf3195b2ce76047a65de29e7ed0a9a8',
     UnderlyingAsset.ALTBULL
   ),
+  erc20(
+    'amkt',
+    'Alongside Crypto Market Index',
+    18,
+    '0xf17a3fe536f8f7847f1385ec1bc967b2ca9cae8d',
+    UnderlyingAsset.AMKT
+  ),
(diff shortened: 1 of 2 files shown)


modules/statics/src/coins.ts
```diff
@@ -419,6 +419,7 @@ export const coins = CoinMap.fromCoins([
     '0xd829664cdbf3195b2ce76047a65de29e7ed0a9a8',
     UnderlyingAsset.ALTBULL
   ),
+  erc20('amkt', 'Alongside Index', 18, '0xf17a3fe536f8f7847f1385ec1bc967b2ca9cae8d', UnderlyingAsset.AMKT),
```
sachushaji, line 422: Double checking, When I checked the contract name it is `Alongside Crypto Market Index` shouldn't we use the same name?

mmcshinsky-bitgo, line 422:
```suggestion
  erc20('amkt', 'Alongside Crypto Market Index', 18, '0xf17a3fe536f8f7847f1385ec1bc967b2ca9cae8d', UnderlyingAsset.AMKT),
```

mmcshinsky-bitgo (changes requested):

modules/statics/src/coins.ts
aarvay: @sachushaji thanks for catching this. It's changed.

mmcshinsky-bitgo (approved):

PR #3129, opened 2022-12-16
refactor(sdk-coin-polygon): add tss recover flow to polygon

kxl4126:
Ticket: BG-64264

This ticket contains the breaking change of removing `recover` method from polygon, as it now inherits the logic from the parent eth controller. The existing `recover` logic in polygon is moved to a method called `recoverEth`, and will be used in non-TSS recoveries.


modules/sdk-coin-polygon/src/polygon.ts
```diff
@@ -325,7 +325,7 @@ export class Polygon extends Eth {
    * @param {String} params.recoveryDestination target address to send recovered funds to
    * @returns {Promise<RecoveryInfo | OfflineVaultTxInfo>}
    */
-  async recover(params: RecoverOptions): Promise<RecoveryInfo | OfflineVaultTxInfo> {
+  async recoverEth(params: RecoverOptions): Promise<RecoveryInfo | OfflineVaultTxInfo> {
```
mmcshinsky-bitgo, line 328: This is a breaking change. Do you want to do a full version bump or can this stay as the same method name?

kxl4126: I think this needs to change to `recoverEth` in order to correctly support the recover flow for TSS recoveries. Right now, it should inherit the `recover` method from Eth controller, so users should still be using the `recover` method in the SDK and pass in their usual params. Ideally, it shouldn't be a breaking change, because it should still work with the exact same params and the same method name (it inherits `recover` from eth). I can make this method private for clarification though. Unless I'm missing something>

mmcshinsky-bitgo: This is a published method, so if you are going to remove it, it needs to be recorded as a breaking change.

kxl4126: Okay, even if `polygon.recover` is still supported because it's inherited from Eth?

mmcshinsky-bitgo: If the inherited method is working the way you want it to work, then yes, you are good to go.

mmcshinsky-bitgo (approved):

modules/sdk-coin-polygon/src/polygon.ts
mullapudipruthvik: `recover` method would be used in WRW, so this would break WRW if we bump sdk
also `recoverEth` doesn't make sense in polygon class

wesgraham (approved):
code changes look good once we get clarity on recoverETH situation

modules/sdk-coin-polygon/src/polygon.ts
kxl4126: `recover` method should still be inherited from the eth class so that should still work as intended (unit tests for `recover` all still pass). But the naming is fair, maybe something like `recoverEthLike` is more apt. Can change that.

wesgraham (approved):

byhow (approved):

PR #3131, opened 2022-12-16
fix!: remove final sig construction for message signing

jeanlarrieux:
Recovery ID (v) is required for Ethereum message signing, in addition to the r and s values from the ECDSA standard. Previously this [was being computed and concatenated](https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-coin-eth/src/eth.ts#L1761) inside of SDK, however MMI does not use SDK but still need the final form of the message hash. Hence this computation[ is being moved to WP](https://github.com/BitGo/bitgo-microservices/pull/27035), and being deleted here.

Ticket: BG-64064

### DIFF
--- a/modules/sdk-coin-eth/src/eth.ts
+++ b/modules/sdk-coin-eth/src/eth.ts
@@ -19,7 +19,6 @@ import {
   common,
   ECDSA,
   Ecdsa,
-  ECDSAMethods,
   ECDSAMethodTypes,
   EthereumLibraryUnavailableError,
   FeeEstimateOptions,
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-eth/src/eth.ts
```diff
@@ -2084,25 +2083,6 @@ export class Eth extends BaseCoin {
     return prefix.concat(message);
   }
 
-  /**
-   *  Create the final signed message hash from the combine share
-   * @param combineShare
-   */
-  constructFinalSignedMessageHash(combineShare: string): string {
```
mmcshinsky-bitgo, line 2091: Unrecorded breaking change.

mmcshinsky-bitgo (changes requested):

modules/sdk-coin-eth/src/eth.ts
mmcshinsky-bitgo: Is there any benefit for our end users having access to this for their own purposes rather than deleting it?

olibeyene (approved):

modules/sdk-coin-eth/src/eth.ts
jeanlarrieux: None whatsoever. Our users care about the final message hash, whether it has the extra bit required by the EVM or not (for non EVM chains) should be an implementation detail (as long as final message hash conform to what the user need).

olibeyene (approved):

mmcshinsky-bitgo (approved):

PR #3137, opened 2022-12-20
feat(utxo-lib): support for unsigned tx with pub scripts

saravanan7mani:
Usually, unsigned tx doesn’t have conditional scripts. But WP recover wallet API requires such requirement. It forces tx to psbt conversion to support the same.

TICKET: BG-58455


davidkaplanbitgo (approved):
lgtm but @OttoAllmendinger should look at this as well

OttoAllmendinger (approved):

PR #3144, opened 2022-12-22
chore(sdk-coin-sui): make sui tx builder gasPayment optional for PayAllSui

linyah9912:
Ticket: BG-64629
Make sui tx builder gasPayment optional for PayAllSui


modules/sdk-coin-sui/src/lib/transactionBuilder.ts
```diff
@@ -91,12 +91,6 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     return this;
   }
 
-  gasPrice(gasPrice: number): this {
```
mmcshinsky-bitgo, line 94: Removal of publicly exposed method in the stable release requires a BREAKING CHANGE: description in the commit footer.

mmcshinsky-bitgo (changes requested):

modules/sdk-coin-sui/src/lib/transactionBuilder.ts
linyah9912: Got it, added the description!
Since Sui is still in progress and not announced to the public, so it probably won't cause too much trouble.

mmcshinsky-bitgo (approved):

PR #3146, opened 2022-12-22
feat(examples): add conslidation examples for trx

derickson-bitgo:
Ticket: BG-62371

### DIFF
--- a/examples/js/trx/consolidate-account-balances.js
+++ b/examples/js/trx/consolidate-account-balances.js
@@ -1,0 +1,57 @@
+/**
+ * Tron wallet can't send funds from their receive addresses.
+ * Account balance consolidation is used to sweep funds from the
+ * receive addresses into the wallet's base address for sending.
+ * @see {@link https://app.bitgo.com/docs/#operation/v2.wallet.consolidateaccount.build}
+ */
+const BitGoJS = require('bitgo');
+
+// TODO: change to 'production' for mainnet
+const bitgo = new BitGoJS.BitGo({ env: 'test' });
+
+// TODO: change to 'trx' for mainnet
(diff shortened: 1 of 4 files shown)


examples/js/trx/create-address.js
```diff
@@ -0,0 +1,24 @@
...
+const coin = 'ttrx';
+
+// TODO: set your wallet id
+const walletId = '';
+
+// TODO: set your access token here
+// You can get this from User Settings > Developer Options > Add Access Token
+const accessToken = '';
+
+Promise.coroutine(function *() {
```
sachushaji, line 18: Using Coroutines / blue bird is not required, we can use simple async / awaits.

examples/js/trx/consolidate-account-balances.js
```diff
@@ -0,0 +1,56 @@
...
+const walletPassphrase = '';
+
+// TODO: set OTP code
+const otp = '000000';
+
+// TODO: set your access token here
+// You can get this from User Settings > Developer Options > Add Access Token
+const accessToken = '';
+
+Promise.coroutine(function *() {
```
sachushaji, line 29: Using Coroutines / blue bird is not required, we can use simple async / awaits.

examples/js/trx/create-address.js
derickson-bitgo: ok, updated.

examples/js/trx/consolidate-account-balances.js
derickson-bitgo: Ok, updated.

sachushaji (approved):
Thanks for addressing the changes

PR #3167, opened 2023-01-10
Add u value proof ec dsa tss

jicli:
## Description

Add v value proof for EcDSA TSS

## Issue Number
BG-65238

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

# Checklist:


jicli:
Please review it and make comments, so i can improve on the PR as I continue to add more tests.

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -156,34 +157,53 @@ export async function createSharedDataProof(
 }
 
 /**
- * Creates an Eddsa KeyShare Proof by appending an ed25519 subkey to an armored gpg private key.
+ * Creates a KeyShare Proof based on given algo.
+ *
+ * Creates an EdDSA KeyShare Proof by appending an ed25519 subkey to an armored gpg private key.
```
Logicwax, line 162: nit: since you mentioned `(auth)` below, put it here as well.   

even smaller nit:  its `ECDSA` not `EcDSA`  (its because in EdDSA it stands for edwards curve, whereas in EC it stands for elliptic curve)

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -156,34 +157,53 @@ export async function createSharedDataProof(
...
+ * Creates an EcDSA KeyShare Proof by Append a secp256k1 subkey (auth) to a PGP keychain.
+ *
+ * For EdDSA:
+ * uValue itself is used to construct by EDDSA signature
```
Logicwax, line 166: this is a confusing section (also misspelled signature)

in both cases (ECDSA and EdDSA), the public form of the u-value is the public params of the auth subkey.   It *IS* is the public subkey.  The proof itself, is really the PGP signature/certification of these subkeys (the fact that ecdsa is used underneath, is really an implementation detail of PGP and probably not worth mentioning as that may be confusing to those using this function and its `algo` param)

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -156,34 +157,53 @@ export async function createSharedDataProof(
...
+  if (algo === 'eddsa') {
+    await sodium.ready;
+    const subKeyVal = Buffer.from(
+      sodium.crypto_scalarmult_ed25519_base_noclamp(Buffer.from(uValue, 'hex'), 'uint8array')
+    );
+    // Sub-key (encryption key) packet.
+    oid = [0x2b, 0x06, 0x01, 0x04, 0x01, 0xda, 0x47, 0x0f, 0x01];
+    // @ts-ignore
+    oid.write = () => new Uint8Array(Buffer.from('092b06010401da470f01', 'hex'));
+    Q = new Uint8Array([...new Uint8Array([0x40]), ...new Uint8Array(subKeyVal)]);
```
Logicwax, line 197: you dont need the first `...`
it could be instead: `[new Uint8Array([0x40])`

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
jicli: ok, i will remove this section

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
jicli: cannot remove it, overloading prevents me from this

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -308,13 +312,15 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
           vssProof: backupToBitgoShare.vssProof,
+          privateShareProof: backupToBitgoShare.privateShareProof,
         },
       ],
       userGPGPublicKey: userGpgKey.publicKey,
       // BitGo is the only supported third party backup as of now, so the
       // backup GPG key is the same as bitgo GPG key. Else user holds backup.
       backupGPGPublicKey: isThirdPartyBackup ? bitgoPublicGpgKey.armor() : userGpgKey.publicKey,
       enterprise: enterprise,
+      algoUsed: 'ecdsa',
```
islamaminBitGo, line 323: I dont think WP needs this parameter, what's the motivation for this?

jicli: Hi @islamaminBitGo 
since HSM is upgraded to support uSig value soon, our clients are gonna to be on old SDK, which means they are not going to provide uSig value. But eddsa usig is already in prod.

Thus, we need to know which algo it is in order to filter in wp (not passing usig to HSM) and make it work with HSM. 
In Wp https://github.com/BitGo/bitgo-microservices/pull/27889 we will ignore the uSig value for prod ecdsa. 

Does it make sense, or am i understanding the timeline and flow in HSM and SDK wrongly?

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
Logicwax: why would spread syntax be required?   you can absolutely do this (i do this already in hsm-api):
`Q = new Uint8Array([new Uint8Array([0x40]), ...new Uint8Array(subKeyVal)]);`

jicli: you have in JS, but i have in TS. 
image

islamaminBitGo: why do this anyways `...new Uint8Array([0x40]) ` then, when we can do ` [0x40, ...new Uint8Array(subKeyVal)]`

islamaminBitGo: @Logicwax is there a reason why you do an array of an array and numbers, that is [[0x40], x, y, z] in instead of [0x40, x, y, z]? Is the former required by openpgp?

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
islamaminBitGo: ah i see, gotcha

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
jicli:
```
    const subKeyVal = Buffer.from(sodium.crypto_scalarmult_ed25519_base_noclamp(Buffer.from(uValue, 'hex')));
    Q = new Uint8Array([0x40, ...subKeyVal]);
```
or the simplest, i think they comes down to the same flat array. 
so they really doesnt make much difference?

Logicwax: @islamaminBitGo yeah you're right, that works too.

@jicli this could simply be `Q = new Uint8Array([0x40, ...subKeyVal])`   does that work in TS?

jicli: yes it would work, updated

jicli:
not sure why the builds are failing.. it was green before..

Logicwax (approved):

modules/bitgo/package.json
```diff
@@ -94,6 +94,7 @@
     "fs-extra": "^9.1.0",
     "lodash": "^4.17.14",
     "openpgp": "5.1.0",
+    "secp256k1": "^4.0.2",
```
brandonblack, line 97: Let's not add another secp256k1 library. We already have a secp256k1 library (`noble-secp256k1`) in the SDK, please use it.

`import { ecc as secp256k1 } from '@bitgo/utxo-lib';`

brandonblack (dismissed):

brandonblack (approved):

PR #3168, opened 2023-01-11
fix(sdk-coin-avaxp): refactor and simplify

jrafael-simtlix:
## Description

The original scope was made agnostic builder. Instead it's a refractor and improve code without a breaking changes. 
The improvements made are discussed below.

### Coin config cast only ones. 
Static network has info required. To avoid casting the network several times, _coinConfig is overridden with AVAXPCoin. So only the constructor cast coinConfig.

https://github.com/BitGo/BitGoJS/blob/a572cbc8e79d2ee2d850730e28b80573085761cd/modules/sdk-coin-avaxp/src/lib/transactionBuilder.ts#L17

https://github.com/BitGo/BitGoJS/blob/a572cbc8e79d2ee2d850730e28b80573085761cd/modules/sdk-coin-avaxp/src/lib/transactionBuilder.ts#L34-L35

### Native transaction load ones
The transaction loads a native transaction as a single point of switch between them. An avoid casting and dynamically loading data from native transactions

https://github.com/BitGo/BitGoJS/blob/a572cbc8e79d2ee2d850730e28b80573085761cd/modules/sdk-coin-avaxp/src/lib/transaction.ts#L308-L335

### Minimize from raw transaction builder
This case is focused on coins where transaction payload data must change, so a raw transaction can only be signed.
Avoid builders fully recovering states from a transaction.

https://github.com/BitGo/BitGoJS/blob/a572cbc8e79d2ee2d850730e28b80573085761cd/modules/sdk-coin-avaxp/src/lib/transactionBuilderFactory.ts#L18-L19

### Mixins from reusable builder fields
Avalanche, in particular, has repeated fields in transactions where an inheritance path cannot be applied. To avoid repeat builder fields in each transaction, a mixins for fields is created.

https://github.com/BitGo/BitGoJS/blob/a572cbc8e79d2ee2d850730e28b80573085761cd/modules/sdk-coin-avaxp/src/lib/exportInCTxBuilder.ts#L14-L16

## Issue Number

Ticket [BG-58639](https://bitgoinc.atlassian.net/browse/BG-58639)

[BG-58639]: https://bitgoinc.atlassian.net/browse/BG-58639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sachushaji (commented):
Can you break down the PR into small commits / multiple PRs, I think its a bit hard to review with these many changes at once.

lcovar (commented):
commenting to flush review queue. pls re-request review once you're back

PR #3169, opened 2023-01-11
fix(utxo-lib): fix (de)serialization methods in UtxoPsbt

davidkaplanbitgo:
There are a few things that are fixed in this PR to ensure proper serialization and deserialization of PSBTs:
 * Format BIP32 Derivation path so that it can be deserialized with a relative path
 * Changes he `UtxoPsbt.fromBuffer` method to use `this.fromHex` instead of `UtxoPsbt.fromHex` so that it returns the right class (`DashPsbt`, `ZcashPsbt`, or `UtxoPsbt`)

Associated previous PRs:
* https://github.com/BitGo/bitcoinjs-lib/pull/54
* https://github.com/BitGo/bip174/pull/1

[BG-54756](https://bitgoinc.atlassian.net/browse/BG-54756)

[BG-54756]: https://bitgoinc.atlassian.net/browse/BG-54756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -37,6 +37,7 @@ export interface TaprootSigner {
 export interface PsbtOpts {
   network: Network;
   maximumFeeRate?: number; // [sat/byte]
+  bip32PathsAbsolute?: boolean;
 }
 
 // TODO: upstream does `checkInputsForPartialSigs` before doing things like
@@ -59,14 +60,16 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
       const tx = this.transactionFromBuffer(buffer, opts.network);
       return new PsbtTransaction({ tx });
     };
-    const psbtBase = PsbtBase.fromBuffer(buffer, transactionFromBuffer);
+    const psbtBase = PsbtBase.fromBuffer(buffer, transactionFromBuffer, {
+      bip32PathsAbsolute: opts.bip32PathsAbsolute,
+    });
     const psbt = this.createPsbt(opts, psbtBase);
     // Upstream checks for duplicate inputs here, but it seems to be of dubious value.
     return psbt;
   }
 
   static fromHex(data: string, opts: PsbtOpts): UtxoPsbt<UtxoTransaction<bigint>> {
-    return UtxoPsbt.fromBuffer(Buffer.from(data, 'hex'), opts);
+    return this.fromBuffer(Buffer.from(data, 'hex'), opts);
   }
 
   toHex(): string {
@@ -145,7 +148,7 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
 
     updates.forEach((update, index) => {
       psbt.updateInput(index, update);
-      psbt.updateInput(index, { witnessUtxo: prevOutputs[index] });
+      psbt.updateInput(index, { witnessUtxo: { script: prevOutputs[index].script, value: prevOutputs[index].value } });
     });
 
     return psbt;
--- a/modules/utxo-lib/src/bitgo/transaction.ts
+++ b/modules/utxo-lib/src/bitgo/transaction.ts
@@ -37,7 +37,11 @@ export function createTransactionFromBuffer<TNumber extends number | bigint = nu
   throw new Error(`invalid network`);
 }
 
-export function createPsbtFromBuffer(buf: Buffer, network: Network): UtxoPsbt<UtxoTransaction<bigint>> {
+export function createPsbtFromBuffer(
+  buf: Buffer,
+  network: Network,
+  bip32PathsAbsolute = false
+): UtxoPsbt<UtxoTransaction<bigint>> {
   switch (getMainnet(network)) {
     case networks.bitcoin:
     case networks.bitcoincash:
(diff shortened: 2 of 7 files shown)


modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -145,7 +145,21 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
 
     updates.forEach((update, index) => {
       psbt.updateInput(index, update);
-      psbt.updateInput(index, { witnessUtxo: prevOutputs[index] });
+      psbt.updateInput(index, { witnessUtxo: { script: prevOutputs[index].script, value: prevOutputs[index].value } });
+    });
+
+    // Remove unknownKeyVals from ins and outs if they are empty.
+    // We do this because we want to have the psbt be changeless from the transaction it is wrapping.
```
brandonblack, line 152: Not sure I understand this reasoning. An empty list-map is equivalent to it not existing in the PSBT serialization format.

brandonblack (changes requested):

modules/utxo-lib/src/bitgo/wallet/chains.ts
```diff
@@ -110,3 +110,11 @@ export function isSegwit(v: ChainCode): boolean {
...
   return segwitCodes.includes(v);
 }
+
+/**
+ * Prepend bip32 derivation path with 'm/' if it is not there
+ * @param path
+ */
+export function formatCorrectBip32DerivationPath(path: string): string {
+  return path.slice(0, 2) !== 'm/' ? 'm/' + path : path;
+}
```
OttoAllmendinger, line 120: as discussed on slack - paths that don't start with `m/` must be supported, just adding `m/` is not a solution here

OttoAllmendinger (dismissed):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
davidkaplanbitgo: It's true - but when you deserialize a psbt hex, these empty arrays do not show up in the `unspents` and `outputs`. If you were to do a comparison of 

```typescript
...
const psbtHex = originalPsbt.toHex();
const deserializedPsbt = UtxoPsbt.fromHex(psbtHex, { network });
assert.deepStrictEqual(deserializedPsbt, originalPsbt)
```
This would fail because of these empty `unknownKeyVals`

modules/utxo-lib/src/bitgo/wallet/chains.ts
davidkaplanbitgo: Looking upstream now

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
brandonblack: Hmm, that seems like something to fix in test code rather than here in production code then.

brandonblack: Or something to fix in a better written comparison (for example, one way to compare PSBTs is to serialize them to a buffer and compare the buffers, since _that_ is a canonical representation).

modules/utxo-lib/src/bitgo/transaction.ts
```diff
@@ -37,25 +37,37 @@ export function createTransactionFromBuffer<TNumber extends number | bigint = nu
   throw new Error(`invalid network`);
 }
 
-export function createPsbtFromBuffer(buf: Buffer, network: Network): UtxoPsbt<UtxoTransaction<bigint>> {
+export function createPsbtFromBuffer(
+  buf: Buffer,
+  network: Network,
+  bip32PathsAbsolute = true
```
brandonblack, line 43: This is a BitGo library, I think we should make this false here.

brandonblack (changes requested):
Couple things to improve, nice work on getting the @bitgo-forks setup and sorting through all this dependency fork blah blah blah to get this fixed :)

davidkaplanbitgo:
Should be good now.

modules/utxo-lib/src/bitgo/transaction.ts
```diff
@@ -37,25 +37,37 @@ export function createTransactionFromBuffer<TNumber extends number | bigint = nu
...
   }
 
   /* istanbul ignore next */
   throw new Error(`invalid network`);
 }
 
+export function createPsbtFromHex(
+  hex: string,
+  network: Network,
+  bip32PathsAbsolute = true
```
brandonblack, line 66: default false here too?

brandonblack (changes requested):
1 nit

davidkaplanbitgo:
NOW it should be good lol

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -128,6 +124,9 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
     this.tx.consensusBranchId = getDefaultConsensusBranchIdForVersion(network, version);
   }
 
+  // For Zcash transactions, we do not have to have non-witness utxo data for non-segwit
+  // transactions because zcash hashes the value directly and therefore does not need to
+  // check the previous transaction.
```
davidkaplanbitgo, line 129: Realized [this comment](https://github.com/BitGo/BitGoJS/pull/3210/files#r1087261135) after merging the other PR, adding the comment here

brandonblack (approved):

PR #3171, opened 2023-01-11
bug(sdk-core): send the passphrase as correct field for eddsa

pranavjain97:
Ticket: BG-00000
This was a bug introduced as part of a refactor from this PR: https://github.com/BitGo/BitGoJS/pull/3102


LeoKotschenreuther (dismissed):
I wonder whether we could add a unit test that could detect this? If anything let's do that as a followup.

joey-bai:
+1 to Leo.

We have a test, which uses Sinon to stub functions: https://github.com/BitGo/BitGoJS/blob/master/modules/bitgo/test/v2/unit/keychains.ts#L271-L282

I think this could have been caught if we verified the stubbed functions were called with specific arguments (guide: https://masteringjs.io/tutorials/sinon/stub-called-with)

joey-bai (approved):

byhow (approved):

PR #3177, opened 2023-01-17
feat(prime): add ofc stellar usdc token

ppong:
Ticket: BG-62133

### DIFF
--- a/modules/statics/src/ofc.ts
+++ b/modules/statics/src/ofc.ts
@@ -237,3 +237,98 @@ export function tofcerc20(
     })
   );
 }
+
+/**
+ * Factory function for ofc stellar token instances.
+ *
+ * @param name unique identifier of the coin
+ * @param fullName Complete human-readable name of the coin
+ * @param network Network object for this coin
+ * @param decimalPlaces Number of decimal places this coin supports (divisibility exponent)
+ * @param asset Asset which this coin represents. This is the same for both mainnet and testnet variants of a coin.
+ * @param kind Differentiates coins which represent fiat assets from those which represent crypto assets
+ * @param prefix? Optional coin prefix. Defaults to empty string
+ * @param suffix? Optional coin suffix. Defaults to coin name.
+ * @param isToken? Whether or not this account coin is a token of another coin
+ * @param features? Features of this coin. Defaults to the DEFAULT_FEATURES defined in `OfcCoin`
+ * @param primaryKeyCurve The elliptic curve for this chain/token
+ */
+export function ofcStellarToken(
+  name: string,
+  fullName: string,
+  decimalPlaces: number,
+  asset: UnderlyingAsset,
+  kind: CoinKind = CoinKind.CRYPTO,
+  features: CoinFeature[] = OfcCoin.DEFAULT_FEATURES,
+  prefix = '',
+  suffix: string = name.toUpperCase(),
+  network: OfcNetwork = Networks.main.ofc,
+  isToken = true,
+  addressCoin = 'xlm',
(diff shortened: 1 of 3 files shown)


garrett-bitgo:
This looks good to me but I won't approve it just yet. I've not gone through the process of adding new coin pairs so I'd like to see if either @jeremypfeifer or @drew-bitgo could take a look when they have time.

modules/statics/src/ofc.ts
```diff
@@ -237,3 +237,98 @@ export function tofcerc20(
...
+ */
+export function ofcStellarToken(
+  name: string,
+  fullName: string,
+  decimalPlaces: number,
+  asset: UnderlyingAsset,
+  kind: CoinKind = CoinKind.CRYPTO,
+  features: CoinFeature[] = OfcCoin.DEFAULT_FEATURES,
+  prefix = '',
+  suffix: string = name.replace(/^ofc/, '').toUpperCase(),
```
amit-yeles-bitgo, line 264: Hm I am not super familiar with adding tokens that aren't erc20s, so I don't have a lot of useful feedback here. One thing though is I noticed `name.replace(/^ofc/, '')` which is something we want to get rid of. In this case it is fine but I see that it is used all over this ofc file. @drew-bitgo Is this something we need to also remove at some point?

drew-bitgo: We dont use `suffix` for anything today, to my knowledge. This is more of a legacy thing that we have due to the base coin types in the sdk. Since its our team that is doing the conversions from ofc to backing 90% of the time, I dont think its a problem, but I would rather have this just be the name, so we dont fall into that trap (and this way, it may help us find places that are using suffix incorrectly)
```suggestion
  suffix: string = name.toUpperCase(),
```

modules/statics/src/ofc.ts
```diff
@@ -237,3 +237,98 @@ export function tofcerc20(
...
+ */
+export function tofcStellarToken(
+  name: string,
+  fullName: string,
+  decimalPlaces: number,
+  asset: UnderlyingAsset,
+  kind: CoinKind = CoinKind.CRYPTO,
+  features: CoinFeature[] = OfcCoin.DEFAULT_FEATURES,
+  prefix = '',
+  suffix: string = name.replace(/^ofc/, '').toUpperCase(),
```
drew-bitgo, line 311:
```suggestion
  suffix: string = name.toUpperCase(),
```

jeremypfeifer (dismissed):

modules/statics/src/ofc.ts
ppong: fixed

modules/statics/src/ofc.ts
ppong: fixed

modules/statics/src/ofc.ts
```diff
@@ -237,3 +237,98 @@ export function tofcerc20(
...
+ */
+export function ofcStellarToken(
+  name: string,
+  fullName: string,
+  decimalPlaces: number,
+  asset: UnderlyingAsset,
+  kind: CoinKind = CoinKind.CRYPTO,
+  features: CoinFeature[] = OfcCoin.DEFAULT_FEATURES,
+  prefix = '',
+  suffix: string = name.toUpperCase(),
```
drew-bitgo, line 264: For onlookers, historically `suffix` was used in place of `backing coin`, to prevent this confusion we will NOT be removing 'ofc' from the name for the suffix. While we work on implementing these new tokens, we will surface any issues found, if any, and resolve them, before updating erc20 and base ofc suffix

drew-bitgo (approved):

PR #3178, opened 2023-01-18
fix: removing generic typing from typeddata associated with eip712

jeanlarrieux:
`HSM` is now expecting a stringify JSON instead of JSON object. Also need `messageEncoded` to be a hex string of the Buffer instead of just a string represenation, as string representation leads to issues recreating the Buffer in `WP` due to encoding.

Ticket: BG-63009

### DIFF
--- a/modules/sdk-coin-eth/src/eth.ts
+++ b/modules/sdk-coin-eth/src/eth.ts
@@ -54,7 +54,7 @@ import type * as EthCommon from '@ethereumjs/common';
 import { calculateForwarderV1Address, getProxyInitcode, KeyPair as KeyPairLib } from './lib';
 import { addHexPrefix, stripHexPrefix } from 'ethereumjs-util';
 import BN from 'bn.js';
-import { TypedDataUtils, SignTypedDataVersion } from '@metamask/eth-sig-util';
+import { TypedDataUtils, SignTypedDataVersion, TypedMessage } from '@metamask/eth-sig-util';
 
 export { Recipient, HalfSignedTransaction, FullySignedTransaction };
 
@@ -2096,13 +2096,13 @@ export class Eth extends BaseCoin {
    * @param typedData the typed data to prepare
    * @return a buffer of the result
    */
-  encodeTypedData(typedData: TypedData<never>): Buffer {
+  encodeTypedData(typedData: TypedData): Buffer {
     const version = typedData.version;
     if (version === SignTypedDataVersion.V1) {
       throw new Error('SignTypedData v1 is not supported due to security concerns');
     }
-
-    const sanitizedData = TypedDataUtils.sanitizeData(typedData.typedDataRaw);
+    const typedDataRaw = JSON.parse(typedData.typedDataRaw);
+    const sanitizedData = TypedDataUtils.sanitizeData(typedDataRaw as unknown as TypedMessage<any>);
     const parts = [Buffer.from('1901', 'hex')];
     const eip712Domain = 'EIP712Domain';
     parts.push(TypedDataUtils.hashStruct(eip712Domain, sanitizedData.domain, sanitizedData.types, version));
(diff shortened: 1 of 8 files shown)


modules/sdk-coin-eth/src/eth.ts
```diff
@@ -2096,15 +2096,16 @@ export class Eth extends BaseCoin {
...
     if (version === SignTypedDataVersion.V1) {
       throw new Error('SignTypedData v1 is not supported due to security concerns');
     }
-
-    const sanitizedData = TypedDataUtils.sanitizeData(typedData.typedDataRaw);
+    const typedDataRaw = JSON.parse(typedData.typedDataRaw);
+    const sanitizedData = TypedDataUtils.sanitizeData(typedDataRaw as unknown as TypedMessage<any>);
     const parts = [Buffer.from('1901', 'hex')];
     const eip712Domain = 'EIP712Domain';
+    console.log('jeannius typedDataRaw: ', typedDataRaw);
```
mmcshinsky-bitgo, line 2108: console.log

modules/sdk-coin-eth/src/eth.ts
```diff
@@ -2096,15 +2096,16 @@ export class Eth extends BaseCoin {
    * @param typedData the typed data to prepare
    * @return a buffer of the result
    */
-  encodeTypedData(typedData: TypedData<never>): Buffer {
+  encodeTypedData(typedData: TypedData): Buffer {
```
mmcshinsky-bitgo, line 2099: Changes to a required parameters on a public method is a breaking change. Needs to be documented in its' own commit for `sdk-coin-eth` separate from other modules.

mmcshinsky-bitgo (changes requested):

modules/sdk-coin-eth/src/eth.ts
```diff
@@ -2096,13 +2096,13 @@ export class Eth extends BaseCoin {
    * @param typedData the typed data to prepare
    * @return a buffer of the result
    */
-  encodeTypedData(typedData: TypedData<never>): Buffer {
```
olibeyene, line 2099: this is a breaking change

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -119,8 +119,8 @@ export interface PopulatedIntentForMessageSigning extends PopulatedIntentBase {
   custodianMessageId?: string;
 }
 
-export interface PopulatedIntentForTypedDataSigning<T extends MessageTypes> extends PopulatedIntentBase {
-  messageRaw: TypedMessage<T>;
+export interface PopulatedIntentForTypedDataSigning extends PopulatedIntentBase {
+  messageRaw: string;
```
olibeyene, line 123: this is also a breaking change

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2788,7 +2788,7 @@ export class Wallet implements IWallet {
           intentType: 'signMessage',
           isTss: true,
           messageRaw: params.message.messageRaw,
-          messageEncoded: params.message?.messageEncoded ?? '',
+          messageEncoded: Buffer.from(params.message?.messageEncoded ?? '').toString('hex'),
```
olibeyene, line 2791: if params.message?.messageEncoded is null, you should throw an error

olibeyene (changes requested):
most of these change are breaking changes. please follow the developer relations guideline

modules/sdk-core/src/bitgo/wallet/wallet.ts
jeanlarrieux: well the point was that future evm chains may not have encoding for message, hence why I made `messageEncoded` optional. If we are to throw an error, might as well make it mandatory.

olibeyene: gotcha

modules/sdk-coin-eth/src/eth.ts
olibeyene: resolving this request since this feature are not live even though they are breaking changes

olibeyene (dismissed):

mmcshinsky-bitgo (approved):

olibeyene (approved):

PR #3180, opened 2023-01-18
chore(sdk-evm): project init

jeanlarrieux:
Ticket: BG-62333

### DIFF
--- a/modules/sdk-evm/package.json
+++ b/modules/sdk-evm/package.json
@@ -10,7 +10,16 @@
     "check-fmt": "prettier --check .",
     "clean": "rm -r ./dist",
     "lint": "eslint --quiet .",
-    "prepare": "npm run build"
+    "prepare": "npm run build",
+    "fmt": "prettier --write '{src,test}/**/*.{ts,js,json}'"
+  },
+  "dependencies": {
+    "@bitgo/sdk-core": "^5.0.0",
+    "@bitgo/statics": "^13.1.0"
+  },
+  "devDependencies": {
+    "@bitgo/sdk-api": "^1.9.5",
+    "@bitgo/sdk-test": "^1.2.10"
   },
   "author": "Bitgo DEFI Team <defi-team@bitgo.com>",
   "license": "MIT",
--- a/modules/sdk-evm/package.json
+++ b/modules/sdk-evm/package.json
@@ -1,0 +1,39 @@
+{
+  "name": "@bitgo/sdk-evm",
+  "version": "1.0.0",
+  "description": "Bitgo SDK library for EVM Wallets",
+  "main": "./dist/src/index.js",
+  "type": "./dist/src/index.d.ts",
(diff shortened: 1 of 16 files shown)


modules/sdk-evm/src/evmWallets.ts
```diff
@@ -0,0 +1,107 @@
...
+    return newWallet;
+  }
+
+  /**
+   * Calls Bitgo API to create an EVM wallet.
+   * @param params
+   * @private
+   */
+  private async createEVMWallet(params: EvmWalletParams): Promise<EvmWallet> {
+    const urlPath = 'api/v2/evmwallets';
```
karankurbur, line 104: api/v2/wallets/evm

modules/sdk-evm/src/types.ts
```diff
@@ -0,0 +1,17 @@
+export const supportedCoins: string[] = ['eth', 'bsc', 'polygon', 'gteth', 'tbsc', 'tpolygon'];
```
bitgoAaron, line 1: this looks like it might be better as an array of `CoinFamily`.

```suggestion
import { CoinFamily } from '@bitgo/statics';
export const supportedCoins: CoinFamily[] = [CoinFamily.ETH, CoinFamily.BSC, CoinFamily.POLYGON]
```

modules/bitgo/package.json
```diff
@@ -83,6 +83,7 @@
     "@bitgo/sdk-coin-xtz": "^1.4.10",
     "@bitgo/sdk-coin-zec": "^1.3.10",
     "@bitgo/sdk-core": "^5.0.0",
+    "@bitgo/sdk-evm": "^1.0.0",
```
bitgoAaron, line 86: let's avoid adding this here until we're ready to release it.

bitgoAaron (changes requested):
See comments

modules/bitgo/package.json
jeanlarrieux: I won't be able to have unit test without adding this to `modules/bitgo`

modules/sdk-evm/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "fmt": "prettier --write .",
+    "check-fmt": "prettier --check .",
+    "clean": "rm -r ./dist",
+    "lint": "eslint --quiet .",
+    "prepare": "npm run build"
+  },
+  "dependencies": {
+    "@bitgo/sdk-api": "^1.9.5",
+    "@bitgo/sdk-core": "^5.0.0",
+    "@bitgo/sdk-test": "^1.2.10",
```
bitgoAaron, line 18: sdk-test is a private package, should go in the `devDependencies`. `sdk-api` is only used by your tests, so it should also go in `devDependencies`

bitgoAaron (approved):

bitgoAaron (approved):

PR #3182, opened 2023-01-18
feat(utxo-lib): do not require previous transactions for zcash

davidkaplanbitgo:
Treat Zcash transactions as Bitcoin Segwit transactions to not using previous transactions because the value is hashed in non-segwit transactions

BG-65554

### DIFF
--- a/modules/utxo-lib/test/bitgo/wallet/util.ts
+++ b/modules/utxo-lib/test/bitgo/wallet/util.ts
@@ -1,4 +1,4 @@
-import { Network } from '../../../src';
+import { getMainnet, Network, networks } from '../../../src';
 import {
   formatOutputId,
   WalletUnspent,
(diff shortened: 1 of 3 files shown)


modules/utxo-lib/test/bitgo/wallet/util.ts
```diff
@@ -118,7 +118,10 @@ export function mockWalletUnspent<TNumber extends number | bigint>(
       BigInt(value),
       network
     );
-    const unspent = isSegwit(chain) ? fromOutput(prevTransaction, vout) : fromOutputWithPrevTx(prevTransaction, vout);
+    const unspent =
```
brandonblack, line 121: Is there also some test code (or fixtures) that are creating prevtxs for zcash unnecessarily?

brandonblack (commented):
1 Question, otherwise LGTM

modules/utxo-lib/test/bitgo/wallet/util.ts
davidkaplanbitgo: AFAIK no

brandonblack (approved):

PR #3183, opened 2023-01-18
fix(ecdsa): allow sync paillier key generation for ovc

islamaminBitGo:
Some UI components like OVC that uses electron can not use node Workers. This PR adds a way to generate a public/private key pair in a Paillier cryptosystem synchronously, so that OVC can use the keyShare function.


Logicwax (approved):

bitgoAaron (changes requested):
can you update the JSDoc for this function that describes what this is doing

islamaminBitGo:
> can you update the JSDoc for this function that describes what this is doing

good point, added an explanation for the sync parameter.

bitgoAaron (approved):

PR #3184, opened 2023-01-18
refactor(tss): always use a distinct backup gpg key

brekiy:
Modifies the `createKeychains` functions for EdDSA, ECDSA, and BLS-DKG to generate and use distinct user and backup GPG keys.

BG-62364


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -231,6 +235,7 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
   async createBackupKeychain({
     userGpgKey,
```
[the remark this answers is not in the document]
LeoKotschenreuther: On second thought it might get used soon again once we verify the u value proofs.

LeoKotschenreuther (commented):
Just some nits, otherwise great work!

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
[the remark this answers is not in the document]
brekiy: Yep I figured it would be safer to leave it in place for any other changes in case they're required.

brekiy:
Rebased to fix merge conflict

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -317,8 +323,8 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       ],
       userGPGPublicKey: userGpgKey.publicKey,
       // BitGo is the only supported third party backup as of now, so the
-      // backup GPG key is the same as bitgo GPG key. Else user holds backup.
-      backupGPGPublicKey: isThirdPartyBackup ? bitgoPublicGpgKey.armor() : userGpgKey.publicKey,
+      // backup GPG key is the same as bitgo GPG key. Else use the provided backupGpgKey.
+      backupGPGPublicKey: isThirdPartyBackup ? bitgoPublicGpgKey.armor() : backupGpgKey.publicKey,
```
Logicwax, line 327: im confused about this logic here.  if `isThirdPartyBackup` is true (implying theres a third party backup provider), don't we want to use `backupGpgKey.publicKey`?

brekiy: The comment about `Bitgo is the only supported third party` confused me. I only changed the false condition since we're now passing in the backup key. In that case, will the user still hold the backup? 
```suggestion
      backupGPGPublicKey: isThirdPartyBackup ? backupGpgKey.publicKey : userGpg.publicKey,
```

LeoKotschenreuther (approved):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
Logicwax: that still doesn't clear up my confusion though.   even if the HSM generates the backup share, wouldn't you want this logic flipped?    as in, if `isThirdPartyBackup` is true, then use the backup key.   otherwise, use bitgo's public key if it's not.

>I think this line will have to change soon once BitGo Trust takes over the storage of the backup keys but for now I think that making this change is unrelated to the current PR/ Jira ticket and hence I'm ok with merging as is.

when trust takes over holding the backup share, they should be treated no different than any other backup provider (such as in the case of all OVC wallets, for now, which have a separate backup key for the backup party).    the backup key should always be the backup key that is provided, regardless.

Logicwax: considering this PR was merged already, I think another PR needs to be made to correct this.   We need to make sure a distinct backup key is always used.   There are no cases where the `userGpg.publicKey` should be used whatsoever.   

Please change this to: `backupGPGPublicKey: backupGpgKey.publicKey`

PR #3188, opened 2023-01-19
feat(sdk-core): add keepAlive as a whitelisted consolidate param

michael-gallegos-bg:
KeepAlive is a flag used currently in the Polkadot consolidation flow. It determines whether or not to sweep the minimum funding amount of an address during a consolidation tx.

TICKET: BG-66035

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -211,7 +211,16 @@ export class Wallet implements IWallet {
    * This is a strict sub-set of prebuildWhitelistedParams
    */
   prebuildConsolidateAccountParams(): string[] {
-    return ['consolidateAddresses', 'feeRate', 'maxFeeRate', 'memo', 'validFromBlock', 'validToBlock', 'preview'];
+    return [
+      'consolidateAddresses',
+      'feeRate',
+      'maxFeeRate',
+      'memo',
+      'validFromBlock',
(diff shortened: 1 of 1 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -211,7 +211,16 @@ export class Wallet implements IWallet {
...
+    return [
+      'consolidateAddresses',
+      'feeRate',
+      'maxFeeRate',
+      'memo',
+      'validFromBlock',
+      'validToBlock',
+      'preview',
+      'keepAlive',
+    ];
```
byhow, line 223: Does this need to be in `core` instead of `sdk-coin-dot` implementation? Maybe through inheritance since this is dot only?

michael-gallegos-bg: This is to update the list of whitelisted params when building/sending account consolidations at the wallet level.
Example call via an SDK script:
```
const consolidationTxs = await walletInstance.sendAccountConsolidations({
    walletPassphrase: walletPassphrase,
    keepAlive: false,
  });
```

If there is somewhere I should move this param flag to let me know.

michael-gallegos-bg: Just pushed an update that adds it to the appropriate jsdocs as well

byhow (approved):

PR #3189, opened 2023-01-19
feat(sdk-coin-atom): create atom module

michael-gallegos-bg:
Create the module and skeleton class for the Cosmos token ATOM.

TICKET: BG-64363

### DIFF
--- a/modules/sdk-coin-atom/src/atom.ts
+++ b/modules/sdk-coin-atom/src/atom.ts
@@ -1,0 +1,97 @@
+import {
+  BaseCoin,
+  BitGoBase,
+  ExplanationResult,
+  KeyPair,
+  MPCAlgorithm,
+  ParsedTransaction,
+  ParseTransactionOptions,
+  SignedTransaction,
+  SignTransactionOptions,
+  UnsignedTransaction,
+  VerifyAddressOptions,
+  VerifyTransactionOptions,
+} from '@bitgo/sdk-core';
+
+import { BaseCoin as StaticsBaseCoin, CoinFamily } from '@bitgo/statics';
+
+export class Atom extends BaseCoin {
+  protected readonly _staticsCoin: Readonly<StaticsBaseCoin>;
+  protected constructor(bitgo: BitGoBase, staticsCoin?: Readonly<StaticsBaseCoin>) {
+    super(bitgo);
+
+    if (!staticsCoin) {
+      throw new Error('missing required constructor parameter staticsCoin');
+    }
+
+    this._staticsCoin = staticsCoin;
+  }
+
+  static createInstance(bitgo: BitGoBase, staticsCoin?: Readonly<StaticsBaseCoin>): BaseCoin {
+    return new Atom(bitgo, staticsCoin);
+  }
+
+  /**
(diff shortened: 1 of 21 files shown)


modules/sdk-coin-atom/src/atom.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  public getFullName(): string {
+    return 'Cosmos';
+  }
+
+  /** @inheritDoc */
+  supportsTss(): boolean {
+    return true;
+  }
```
joey-bai, line 56: is there some account features prop we can use from static coin?

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  getMPCAlgorithm(): MPCAlgorithm {
+    return 'eddsa';
+  }
```
joey-bai, line 60: should this be ecdsa?
also, can we have this be derived from static coin?

modules/sdk-coin-atom/src/tatom.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Testnet Cosmos
+ *
+ * @format
+ */
+import { BaseCoin, BitGoBase } from '@bitgo/sdk-core';
+import { Atom } from './atom';
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+
+export class Tatom extends Atom {
```
joey-bai, line 10: similar comments on leveraging static coin where possible.

modules/sdk-coin-atom/src/atom.ts
michael-gallegos-bg: Yes it should, and as for deriving from the static coin type not as simply as I can tell - we could make it an attribute of the StaticCoin type in the future if we'd like to do a bit of refactoring.

modules/sdk-coin-atom/src/atom.ts
michael-gallegos-bg: Just commented on the other comment - but I think it might be good to add these as statics attributes in the near future. We can follow that up in a separate PR. I created a ticket - https://bitgoinc.atlassian.net/browse/BG-66189

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -0,0 +1,97 @@
...
+    return 'atom';
+  }
+
+  public getFamily(): string {
+    return 'atom';
+  }
+
+  public getFullName(): string {
+    return 'Cosmos';
+  }
```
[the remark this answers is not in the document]
michael-gallegos-bg: Yes

joey-bai (approved):

modules/sdk-coin-atom/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './atom';
+export * from './tatom';
```
mmcshinsky-bitgo, line 2: Needs new `register` method added to other coin modules.

michael-gallegos-bg: Already working on including it in my next PR, ty!

PR #3194, opened 2023-01-21
feat: send vss proof and privateShareProof to bitgo as backup

pengyuc-bitgo:
also patch some places the vss proof is lost

Ticket: BG-66133, BG-65793


jicli (approved):

olibeyene (commented):
can you add unit test for this changes? non blocker but it would be great to have test

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -503,6 +505,7 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
           encryptedPrivateShare: bitGoToRecipientShare.privateShare,
           n: bitGoToRecipientShare.n!,
           vssProof: bitGoToRecipientShare.vssProof,
+          privateShareProof: bitGoToRecipientShare.privateShareProof,
```
jicli, line 508: i was looking in the code, was this really necessary, or is it going to be needed in WP to be stored to DB?

pengyuc-bitgo: I think for this line, its probably just make it consistent so that anywhere the vssProof is set, the privateShareProof is also set.

PR #3196, opened 2023-01-23
feat(express): support external signer TSS EdDSA in sendMany express endpoint

islamaminBitGo:
Ticket: BG-65857

Test: 

The following transaction was performed using external signer successfully https://explorer.solana.com/tx/gdj8H7jkNZbkAwZeEcNMK1kymmiDAcRR3gXV4Kn1h6DRVTSecUU7gBGXHfpeU4Bh8oHHCjjukkTmANGEfDwuhG4?cluster=devnet

Request:
image

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -729,7 +729,11 @@ async function handleV2SendMany(req: express.Request) {
   req.body.reqId = reqId;
   let result;
   try {
-    result = await wallet.sendMany(createSendParams(req));
+    if (wallet._wallet.multisigType === 'tss') {
+      result = await wallet.sendMany(createTSSSendParams(req));
+    } else {
+      result = await wallet.sendMany(createSendParams(req));
+    }
   } catch (err) {
     err.status = 400;
     throw err;


jicli (approved):

olibeyene (changes requested):
eddsa already support express endpoint. I don't believe this change is needed. Please confirm

islamaminBitGo:
> eddsa already support express endpoint. I don't believe this change is needed. Please confirm

it's not supported in external signer mode where express would need to call an external signer instance to generate signature shares.

modules/express/src/clientRoutes.ts
```diff
@@ -729,7 +729,11 @@ async function handleV2SendMany(req: express.Request) {
   req.body.reqId = reqId;
   let result;
   try {
-    result = await wallet.sendMany(createSendParams(req));
+    if (wallet._wallet.multisigType === 'tss') {
+      result = await wallet.sendMany(createTSSSendParams(req));
```
olibeyene, line 733: can you verify this tss param will work for ECDSA?

islamaminBitGo: we do not support ECDSA yet.

olibeyene (approved):

PR #3197, opened 2023-01-23
feat(sdk-coin-atom): add keyPair management to cosmos

michael-gallegos-bg:
TICKET: BG-66310

### DIFF
--- a/modules/sdk-coin-atom/src/lib/keyPair.ts
+++ b/modules/sdk-coin-atom/src/lib/keyPair.ts
@@ -1,0 +1,73 @@
+import { randomBytes } from 'crypto';
+import { bip32 } from '@bitgo/utxo-lib';
+import { pubkeyToAddress } from '@cosmjs/amino';
+import { toBase64 } from '@cosmjs/encoding';
+import {
+  DefaultKeys,
+  isPrivateKey,
+  isPublicKey,
+  isSeed,
+  KeyPairOptions,
+  Secp256k1ExtendedKeyPair,
+} from '@bitgo/sdk-core';
+import { DEFAULT_SEED_SIZE_BYTES } from './constants';
+
+/**
+ * Cosmos keys and address management.
+ */
+export class KeyPair extends Secp256k1ExtendedKeyPair {
+  /**
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   * @param { KeyPairOptions } source Either a master seed, a private key (extended or raw), or a public key
+   *     (extended, compressed, or uncompressed)
+   */
+  constructor(source?: KeyPairOptions) {
+    super(source);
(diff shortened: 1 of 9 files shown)


modules/sdk-coin-atom/src/lib/keyPair.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  /**
+   * Cosmos default keys format: raw private and compressed public key
+   *
+   * @returns { DefaultKeys } The keys in the protocol default key format
+   */
+  getKeys(): DefaultKeys {
+    return {
+      pub: this.getPublicKey({ compressed: true }).toString('hex'),
```
michael-gallegos-bg, line 53: > A compressed format is used for secp256k1 and secp256r1 serialization.
https://docs.cosmos.network/v0.46/basics/accounts.html

Need to double check if this is what is wanted

modules/sdk-coin-atom/src/lib/keyPair.ts
```diff
@@ -0,0 +1,74 @@
...
+import {
+  DefaultKeys,
+  isPrivateKey,
+  isPublicKey,
+  isSeed,
+  KeyPairOptions,
+  Secp256k1ExtendedKeyPair,
+} from '@bitgo/sdk-core';
+
+const DEFAULT_SEED_SIZE_BYTES = 16;
```
Vijay-Jagannathan, line 14: Is this common across chains (does seem like it but confirming) ? If yes, do we need to put it somewhere in the `Secp256k1ExtendedKeyPair` class or any other appropriate file ?

modules/sdk-coin-atom/src/lib/keyPair.ts
Vijay-Jagannathan: While I was trying to create a wallet address using the `@cosmjs/amino` library, it complained that I was having an `Invalid Secp256k1 pubkey length (compressed): 48` and that the correct compressed length was 33 bytes - [Link 1](https://github.com/cosmos/cosmjs/blob/main/packages/amino/src/pubkeys.ts#L42-L53), [Link 2](https://github.com/cosmos/cosmjs/blob/main/packages/amino/src/addresses.ts#L17). From this I believe compressed is what's expected but I'll defer to another pair of eyes.

modules/sdk-coin-atom/src/lib/keyPair.ts
michael-gallegos-bg: You are correct, it seems most commonly 16, but some chains have it at 64.
For our purposes, I will set it in a coin constants file.

Vijay-Jagannathan (approved):

sachushaji (approved):

PR #3201, opened 2023-01-24
feat(key-card): add utils function to split keys

LeoKotschenreuther:
This function will later be used to split data into multiple pieces before converting it to QR codes on the key cards.

Ticket: BG-64432

### DIFF
--- a/modules/key-card/test/unit/utils.ts
+++ b/modules/key-card/test/unit/utils.ts
@@ -1,0 +1,30 @@
+import 'should';
+import { splitKeys } from '../../src';
+
+describe('getSplitKeys', function () {
+  it('only returns one item when input shorter than limit', function () {
+    const key = 'hello world';
+    const limit = 20;
+    const keys = splitKeys(key, limit);
+    keys.length.should.equal(1);
+    keys[0].should.equal(key);
+  });
+
+  it('splits key up into multiple items when exceeding the limit', function () {
+    const key = 'hello world this is a short message';
+    const limit = 5;
+    const keys = splitKeys(key, limit);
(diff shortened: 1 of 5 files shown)


modules/key-card/test/unit/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  it('splits key up into multiple items when exceeding the limit', function () {
+    const key = 'hello world this is a short message';
+    const limit = 5;
+    const keys = splitKeys(key, limit);
+    keys.length.should.equal(7);
+    for (const splitKey of keys.slice(0, keys.length - 1)) {
+      splitKey.length.should.equal(limit);
+    }
+    keys[keys.length - 1].length.should.be.lessThanOrEqual(limit);
```
rlutzbitgo, line 21: We could add 
```
keys[0].should.equal('hello')
keys[1].should.equal(' worl')
keys[2].should.equal('d thi')
keys[3].should.equal('s is ')
keys[4].should.equal('a sho')
keys[5].should.equal('rt me')
keys[6].should.equal('ssage')
```
If we really wanted to be safe. I think it's fine though, up to you

rlutzbitgo (approved):

modules/key-card/test/unit/utils.ts
LeoKotschenreuther: Good idea, I will make that change.

LeoKotschenreuther: Added

zahin-mohammad (approved):

byhow (approved):

PR #3205, opened 2023-01-25
Bg 66142 create evm tss utils

jeanlarrieux:
Finish creating necessary code to interact with `wp` in order to create an `evm wallet`


jicli (dismissed):
lgtm

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/ecdsaUnifiedWalletUtils.ts
```diff
@@ -0,0 +1,407 @@
...
+
+  private url(suffix: string): string {
+    return this._keychains.getCoin().url(suffix);
+  }
+
+  private async setBitgoGpgPubKey(bitgo) {
+    this.bitgoPublicGpgKey = await getBitgoGpgPubKey(bitgo);
+  }
+
+  async createKeychains(params: {
```
jicli, line 45: on second thought this class looks like almost exactly like `ecdsa` 
Two question
1: what would be the major difference between ecdsa-unified-wallet and ecdsa
2. would ecdsa-unified-wallet support for example vss, uSig and etc just like ecdsa

jicli: Also, would in future ecdsa-unified-wallet be supported for cutodial hot and trust just like Nike in ecdsa now.

So we would have backupProvider as `TrustAsKRS` then this class needs exact same modification from `ecdsa` which would be a problem to keep track the difference here and there.

jeanlarrieux: Yes it would support those things. We wanted to separate key generation from wallet creation. It looks like the wallet argument is optional for the `EcdsaUtils` class. so I was able to replace it.

modules/sdk-unified-wallet/package.json
```diff
@@ -14,10 +14,12 @@
   },
   "dependencies": {
     "@bitgo/sdk-core": "^5.2.0",
-    "@bitgo/statics": "^14.1.0"
+    "@bitgo/statics": "^14.1.0",
+    "bitgo": "^16.4.0",
```
bitgoAaron, line 18: please remove. Instead write your unit tests with TestBitGo and BitGoAPI. See other coin module unit tests for example on how to do this, but adding this dependency is going to 1000x your bundle size because now you depend on every other coin in the repo.

modules/sdk-unified-wallet/test/unit/ecdsaUnifiedWallets.ts
```diff
@@ -0,0 +1,99 @@
+import { TestBitGo } from '@bitgo/sdk-test';
+import { EcdsaUnifiedWallets, GenerateUnifiedWalletOptions, UnifiedWallet } from '../../src';
+import { BitgoGPGPublicKey, common, ECDSAUtils } from '@bitgo/sdk-core';
+import * as openpgp from 'openpgp';
+import { bitgoKeyChain, backupKeychain, userKeyChain } from '../fixtures/ecdsaUnifiedWalletFixtures';
+import * as sinon from 'sinon';
+import nock = require('nock');
+import { BitGo } from 'bitgo';
```
bitgoAaron, line 8: don't do this

bitgoAaron (changes requested):
see comments

modules/sdk-unified-wallet/package.json
```diff
@@ -14,10 +14,11 @@
...
+    "openpgp": "5.1.0"
```
bitgoAaron, line 19: this belongs in devDependencies since it's only used by your unit tests, please move there.

modules/sdk-unified-wallet/package.json
```diff
@@ -14,10 +14,11 @@
...
   },
   "devDependencies": {
-    "@bitgo/sdk-api": "^1.9.7",
```
bitgoAaron, line 20: you're going to need this back in so you can write your unit tests without the `bitgo` module

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/types.ts
```diff
@@ -1,4 +1,4 @@
 import { CoinFamily } from '@bitgo/statics';
 
-export const supportedCoins: CoinFamily[] = [CoinFamily.ETH, CoinFamily.BSC, CoinFamily.POLYGON];
+export const supportedCoins: string[] = [CoinFamily.ETH, CoinFamily.BSC, CoinFamily.POLYGON];
```
bitgoAaron, line 3: why relax the type here?

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/ecdsaUnifiedWallets.ts
```diff
@@ -1,16 +1,22 @@
...
+  Keychains,
+} from '@bitgo/sdk-core';
 import { UnifiedWalletID, UnifiedWallet, GenerateUnifiedWalletOptions } from '../types';
 import { supportedCoins } from './types';
 import { UnifiedWallets } from '../unifiedWallets';
 
 export class EcdsaUnifiedWallets extends UnifiedWallets {
-  constructor(bitgo: BitGoBase) {
-    super(bitgo);
+  constructor(bitgo: BitGoBase, coinName = 'eth') {
```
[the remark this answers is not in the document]
jeanlarrieux: The only reason this is there is for testing purposes (teth) but if it wasn't for that it would just not be exposed. The reason is we just need an 'exemplar' ecdsa coin to generate the keys. Wallet generation happens after ecdsa keys are done being generated. We then used the same set of key (from the examplar coin) to create wallets for each of the coins we want (the `supportedCoins`)

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/ecdsaUnifiedWallets.ts
```diff
@@ -1,16 +1,22 @@
...
+  BitGoBase,
+  KeychainsTriplet,
+  SupplementGenerateWalletOptions,
+  Keychains,
+} from '@bitgo/sdk-core';
 import { UnifiedWalletID, UnifiedWallet, GenerateUnifiedWalletOptions } from '../types';
 import { supportedCoins } from './types';
 import { UnifiedWallets } from '../unifiedWallets';
 
 export class EcdsaUnifiedWallets extends UnifiedWallets {
```
jicli, line 12: perhaps `EcdsaEVMUnifiedWallets` name would be more approriate? since looks like we are fixing this to be EVM specific

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/types.ts
jicli: instead of just changing this to `string` and add extra param in generateUnifiedWallet
Can we just add test coins in the list?

jicli: Or if it's only for testing purpose, we could hard modify in the test, but i really dont think we should add a param just for test

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/ecdsaEVMUnifiedWallets.ts
```diff
@@ -1,12 +1,18 @@
...
+  Keychains,
+} from '@bitgo/sdk-core';
 import { UnifiedWalletID, UnifiedWallet, GenerateUnifiedWalletOptions } from '../types';
 import { supportedCoins } from './types';
 import { UnifiedWallets } from '../unifiedWallets';
 
-export class EcdsaUnifiedWallets extends UnifiedWallets {
-  constructor(bitgo: BitGoBase) {
-    super(bitgo);
+export class EcdsaEVMUnifiedWallets extends UnifiedWallets {
```
jicli, line 12: weird, why doesnt this file has keyword colors...

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/types.ts
```diff
@@ -1,4 +1,4 @@
 import { CoinFamily } from '@bitgo/statics';
 
-export const supportedCoins: CoinFamily[] = [CoinFamily.ETH, CoinFamily.BSC, CoinFamily.POLYGON];
+export const supportedCoins: string[] = [CoinFamily.ETH, CoinFamily.POLYGON];
```
jicli, line 3: why did we remove BSC? not supported?

jeanlarrieux: TSS not yet supported on BSC and will lead to error when trying to create a wallet.

jicli (approved):

jicli (approved):

bitgoAaron (approved):

PR #3209, opened 2023-01-25
fix(sdk-core): send vss share to wp

olibeyene:
Ticket: BG-61036

### DIFF
--- a/modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
+++ b/modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
@@ -150,21 +150,19 @@ export async function createUserToBitGoGShare(
     throw new Error('Invalid YShare, is not backup key');
   }
 
-  let v, r, R;
+  const r = bitgoToUserRShare.share.substring(0, 64);
+  const R = bitgoToUserRShare.share.substring(64, 128);
+
+  let vssProof;
   if (bitgoToUserRShare.share.length > 128) {
-    v = bitgoToUserRShare.share.substring(0, 64);
-    r = bitgoToUserRShare.share.substring(64, 128);
-    R = bitgoToUserRShare.share.substring(128, 192);
-  } else {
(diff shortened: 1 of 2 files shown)


pranavjain97:
How do we know that vss signing is fixed?

pranavjain97 (approved):

modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
```diff
@@ -200,8 +198,7 @@ export async function offerUserToBitgoRShare(
   const signatureShare: SignatureShareRecord = {
     from: SignatureShareType.USER,
     to: SignatureShareType.BITGO,
-    // TODO(BG-61037): Fix signing with VSS
-    share: /* rShare.v + */ rShare.r + rShare.R,
+    share: rShare.r + rShare.R + rShare.v,
```
pengyuc-bitgo, line 201: Why are we putting them all in one string?

olibeyene: this the pattern how tss exchanges shares with WP <> sdk. shares are contacted before sending to wp

jicli (approved):

modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
jicli: would hope we had delimiter instead of counting bytes

pengyuc-bitgo (approved):

sachushaji (commented):
Can you add a breaking changes mentioned in the commit message as we are including v value now which prior to the change was not passed. Check [here](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=in%20Semantic%20Versioning\).-,BREAKING,-CHANGE%3A%20a)

olibeyene:
Closing this pr since [pr 3285 handles eddsa vss share](https://github.com/BitGo/BitGoJS/pull/3285 )

PR #3210, opened 2023-01-25
feat(utxo-lib): zcash non-segwit unspents can be signed

davidkaplanbitgo:
Zcash non-segwit unspents can be signed without a nonWitnessUtxo

[BG-66488](https://bitgoinc.atlassian.net/browse/BG-66488)

[BG-66488]: https://bitgoinc.atlassian.net/browse/BG-66488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
@@ -4,7 +4,7 @@ import {
   getDefaultVersionGroupIdForVersion,
   ZcashTransaction,
 } from './ZcashTransaction';
-import { Network, PsbtTransaction } from '../../';
+import { Network, PsbtTransaction, Signer } from '../../';
 import { Psbt as PsbtBase } from 'bip174';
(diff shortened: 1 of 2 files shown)


brandonblack (approved):

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -128,6 +128,15 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
     this.tx.consensusBranchId = getDefaultConsensusBranchIdForVersion(network, version);
   }
 
+  signInput(inputIndex: number, keyPair: Signer, sighashTypes?: number[]): this {
+    (this as any).__CACHE.__UNSAFE_SIGN_NONSEGWIT = true;
```
brandonblack, line 132: Maybe we should throw a comment in here about why this is safe (zcash hashes the value directly)?

PR #3211, opened 2023-01-25
feat(sdk-core): use trust backup gpg for encrypting

jicli:
Ticket: BG-66235

### DIFF
--- a/modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
@@ -303,7 +303,7 @@ export async function getBitgoToUserLatestShare(
  * @param keyShare - TSS key share of the party preparing exchange materials
  * @param recipientIndex - index of the recipient (1, 2, or 3)
  * @param recipientGpgPublicArmor - recipient's public gpg key in armor format
- * @param gpgKey - ephemeral GPG key to encrypt / decrypt sensitve data exchanged between user and server
+ * @param senderGpgKey - ephemeral GPG key to encrypt / decrypt sensitve data exchanged between user and server
  * @param isbs58Encoded - is bs58 encoded or not
  * @returns encrypted N Share
  */
@@ -311,7 +311,7 @@ export async function encryptNShare(
   keyShare: KeyShare,
   recipientIndex: number,
   recipientGpgPublicArmor: string,
-  gpgKey: pgp.SerializedKeyPair<string>,
+  senderGpgKey: pgp.SerializedKeyPair<string>,
   isbs58Encoded = true
 ): Promise<EncryptedNShare> {
   const nShare = keyShare.nShares[recipientIndex];
@@ -346,7 +346,7 @@ export async function encryptNShare(
     encryptedPrivateShare,
     n: nShare.n,
     vssProof: nShare.v,
-    privateShareProof: await createShareProof(gpgKey.privateKey, nShare.u, 'ecdsa'),
+    privateShareProof: await createShareProof(senderGpgKey.privateKey, nShare.u, 'ecdsa'),
   };
 }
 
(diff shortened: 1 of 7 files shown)


jicli:
Please review it while im adding unit tests. Thanks

olibeyene (dismissed):
lgtm!

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -123,7 +134,9 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
 
     const userKeyShare = await MPC.keyShare(1, m, n);
     const userGpgKey = await generateGPGKeyPair('secp256k1');
-    const backupGpgKey = await generateGPGKeyPair('secp256k1');
+    const userLocalBackupGpgKey = await generateGPGKeyPair('secp256k1');
+    const krsConstants = await this.getTrustBackupConstants();
+    const thirdPartyBackupPublicGpgKey = await readKey({ armoredKey: krsConstants.trustBackupGPGPubKey });
```
pengyuc-bitgo, line 139: We don't usually need to get this Trust backup key. That is only for when we use the trust as KRS.
I don't think the `thirdPartyBackupPublicGpgKey` and `userLocalBackupGpgKey` would usually co-exist.
So, maybe we can at least get this after the `isThirdPartyBackup` and only get this if it is true.

The `createBackupKeyShares` a few lines below may be a okay example to follow to avoid having both `userLocalBackupGpgKey` and `thirdPartyBackupPublicGpgKey` parmeters in every methods that uses the backup public key.

I wish there is a better refactor than even that, though. :/

pengyuc-bitgo (dismissed):

pengyuc-bitgo (approved):

brekiy (dismissed):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/types.ts
```diff
@@ -1,19 +1,19 @@
...
 export type DecryptableNShare = ECDSAMethodTypes.DecryptableNShare;
 
 export type CreateEcdsaKeychainParams = CreateKeychainParamsBase & {
   userKeyShare: ECDSA.KeyShare;
-  backupGpgKey: SerializedKeyPair<string>;
   backupKeyShare: BackupKeyShare;
   isThirdPartyBackup?: boolean;
   backupProvider?: BackupProvider;
   bitgoPublicGpgKey: Key;
+  backupGpgKey: backupGpgKey;
```
pranavjain97, line 16: nit: can we make this consistent to call `backupPublicGpgKey`?

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -498,22 +518,29 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
     }
 
-    await this.verifyWalletSignatures(
-      userGpgKey.publicKey,
-      backupGpgKey.publicKey,
-      bitgoKeychain,
-      bitGoToRecipientShare.publicShare,
-      recipientIndex
-    );
+    // await this.verifyWalletSignatures(
```
jicli, line 521: todo uncomment ofc after figuring out what to do

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -303,15 +303,15 @@ export async function getBitgoToUserLatestShare(
...
+ * @param senderGpgKey - ephemeral GPG key to encrypt / decrypt sensitve data exchanged between user and server
  * @param isbs58Encoded - is bs58 encoded or not
  * @returns encrypted N Share
  */
 export async function encryptNShare(
   keyShare: KeyShare,
   recipientIndex: number,
   recipientGpgPublicArmor: string,
-  gpgKey: pgp.SerializedKeyPair<string>,
+  senderGpgKey: pgp.SerializedKeyPair<string>,
```
byhow, line 314: breaking name changes need to have `BREAKING: ` in the commit message. i.e. https://github.com/BitGo/BitGoJS/pull/2565

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -35,6 +35,18 @@ export async function getBitgoGpgPubKey(bitgo: BitGoBase): Promise<Key> {
...
+/**
+ * Fetches Trust pub key string
+ * @param bitgo
+ */
+export async function getTrustGpgPubKey(bitgo: BitGoBase): Promise<Key> {
+  const constants = await bitgo.fetchConstants();
+  if (!constants.trustPubKey) {
+    throw new Error('Unable to get trustPubKey');
+  }
+  return await readKey({ armoredKey: constants.trustPubKey });
```
byhow, line 47: nit: extraneous `await`

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
byhow: do we have a ticket for this todo?

pranavjain97 (commented):
Lgtm; small nits.. also I think we should explore ways to avoid needing to do `as` in the types everywhere. Please address in a follow-up. 

Why are we commenting out `verifyWalletSignatures`?

byhow (changes requested):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/types.ts
jicli: it's not necessarily just the public, it can be either a key pair or just a pub key

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
jicli: is this really a breaking change? it's not a typed params so whatever people passing in the original order will still be the same order and nothing will change

byhow: ah, fair enough, my bad

byhow (approved):

byhow (approved):

PR #3215, opened 2023-01-26
fix(bitgo): remove enforcement of HMAC verification on all non-prod envs

aaggarwal10:
Ticket: BG-66645

### DIFF
--- a/modules/bitgo/test/unit/bitgo.ts
+++ b/modules/bitgo/test/unit/bitgo.ts
@@ -78,17 +78,17 @@ describe('BitGo Prototype Methods', function () {
   });
 
   describe('HMAC request verification', () => {
-    it('throws if HMAC request verification is disabled for non-dev environments', function () {
+    it('throws if HMAC request verification is disabled for non-prod environments', function () {
       (() => TestBitGo.decorate(BitGo, { env: 'prod', hmacVerification: false }))
         .should.throw(/Cannot disable request HMAC verification in environment/);
       (() => TestBitGo.decorate(BitGo, { env: 'test', hmacVerification: false }))
-        .should.throw(/Cannot disable request HMAC verification in environment/);
+        .should.not.throw(/Cannot disable request HMAC verification in environment/);
       (() => TestBitGo.decorate(BitGo, { env: 'adminProd', hmacVerification: false }))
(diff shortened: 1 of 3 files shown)


modules/bitgo/test/unit/bitgo.ts
```diff
@@ -81,14 +81,8 @@ describe('BitGo Prototype Methods', function () {
     it('throws if HMAC request verification is disabled for non-dev environments', function () {
```
bitgoAaron, line 81: let's update this description to `non-prod environments`

modules/bitgo/test/unit/bitgo.ts
```diff
@@ -81,14 +81,8 @@ describe('BitGo Prototype Methods', function () {
...
       (() => TestBitGo.decorate(BitGo, { env: 'prod', hmacVerification: false }))
         .should.throw(/Cannot disable request HMAC verification in environment/);
-      (() => TestBitGo.decorate(BitGo, { env: 'test', hmacVerification: false }))
-        .should.throw(/Cannot disable request HMAC verification in environment/);
```
bitgoAaron, line 85: instead of deleting, can we just change the assertion to call `.should.not.throw` ?

bitgoAaron (changes requested):
see comments

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -256,7 +256,7 @@ export class BitGoAPI implements BitGoBase {
     this._validate = params.validate === undefined ? true : params.validate;
 
     if (!params.hmacVerification && params.hmacVerification !== undefined) {
-      if (common.Environments[env].hmacVerificationEnforced) {
+      if ((env == 'prod' || env == 'adminProd') && common.Environments[env].hmacVerificationEnforced) {
```
bitgoAaron, line 259: we should also change the value of `hmacVerificationEnforced` for testnetBase to false over in https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/environments.ts#L132

bitgoAaron (approved):

bitgoAaron (approved):

PR #3216, opened 2023-01-26
fix(sdk-core): get low balance forwarder

olibeyene:
Ticket: BG-66642

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -324,6 +324,16 @@ export interface FlushForwarderTokenOptions {
   };
 }
 
+export interface ForwarderBalance {
+  address: string;
+  balance: string;
+}
+
+export interface ForwarderBalanceOptions {
+  minimumBalance?: number;
+  maximumBalance?: number;
+}
+
 export interface CreateAddressOptions {
   chain?: number;
   gasPrice?: number | string;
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2922,4 +2924,25 @@ export class Wallet implements IWallet {
     this._wallet = await this.bitgo.post(url).send(params).result();
     return this._wallet;
   }
+
+  /**
+   * Gets forwarder's balance
+   * @param params - optional query parameters
+   * @returns List of forwarder address and balance,
+   * if params is not set the list of forwarder are the identified low balance forwarder
```
olibeyene, line 2932: if params is not set then returns gets low balance forwarders

shiksb (commented):
LGTM, can you add a unit test?

shiksb (dismissed):

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -599,6 +609,7 @@ export interface IWallet {
   updateForwarders(forwarderFlags?: any): Promise<any>;
   deployForwarders(params: DeployForwardersOptions): Promise<any>;
   flushForwarderToken(params: FlushForwarderTokenOptions): Promise<any>;
+  getForwarderBalance(params: ForwarderBalanceOptions): Promise<ForwarderBalance[]>;
```
byhow, line 612: seems this param is optional in the implementation
```suggestion
  getForwarderBalance(params?: ForwarderBalanceOptions): Promise<ForwarderBalance[]>;
```

byhow (approved):

PR #3218, opened 2023-01-27
feat(key-card): add function to generate key card faqs

LeoKotschenreuther:
Ticket: BG-64432

### DIFF
--- a/modules/key-card/test/unit/faq.ts
+++ b/modules/key-card/test/unit/faq.ts
@@ -1,0 +1,14 @@
+import { generateFaq } from '../../src/faq';
+
+describe('generateFaq', function () {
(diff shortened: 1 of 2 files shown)


modules/key-card/test/unit/faq.ts
```diff
@@ -0,0 +1,14 @@
+import { generateFaq } from '../../src/faq';
+
+describe('generateFaq', function () {
```
LeoKotschenreuther, line 3: I know that these tests are very rudimentary. If somebody has ideas how I can improve the test(s) in a meaningful way then I'm all ears.

brekiy (approved):

PR #3219, opened 2023-01-27
chore: add narrow types for backupProvider

zahin-mohammad:
## Description

This PR adds a narrow type for BackupProvider to make it easier to understand all the valid values it can hold.

## Issue Number

TICKET: BG-66601

## Type of change

# How Has This Been Tested?

This PR does not change any existing functionality. 

# Checklist:

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
@@ -4,7 +4,7 @@ import { IBaseCoin, KeychainsTriplet } from '../../baseCoin';
 import { BitGoBase } from '../../bitgoBase';
 import { Keychain } from '../../keychain';
 import { getTxRequest } from '../../tss';
-import { IWallet } from '../../wallet';
+import { IWallet, BackupProvider } from '../../wallet';
 import { MpcUtils } from '../mpcUtils';
 import * as _ from 'lodash';
 import {
@@ -89,6 +89,7 @@ export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtil
     enterprise?: string | undefined;
     originalPasscodeEncryptionCode?: string | undefined;
     isThirdPartyBackup?: boolean;
+    backupProvider?: BackupProvider;
   }): Promise<KeychainsTriplet> {
     throw new Error('Method not implemented.');
   }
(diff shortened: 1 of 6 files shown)


modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
```diff
@@ -314,7 +315,9 @@ export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtil
    * Checks whether the third party backup provider is valid/supported
    * @param backupProvider - the backup provider client selected
    */
-  isValidThirdPartyBackupProvider(backupProvider: string | undefined): boolean {
+  isValidThirdPartyBackupProvider(
+    backupProvider: BackupProvider | string | undefined
```
jicli, line 319: should this not have string then

modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -21,16 +21,17 @@ export interface GenerateMpcWalletOptions {
   originalPasscodeEncryptionCode?: string;
   enterprise?: string;
   walletVersion?: number;
-  backupProvider?: string;
+  backupProvider?: BackupProvider;
 }
 
+export type BackupProvider = 'BitGoTrustAsKrs';
```
pranavjain97, line 27: If you're adding better types, can you instead have `BitGoTrustAsKrs` as const type, so we can more in the future. Something similar to what we did for hsm types.
```
export const hsmTypes = ['institutional', 'nitro'] as const;

export const HSMTypes = createStringObject(hsmTypes);
export type HSMType = typeof hsmTypes[number];
```

pranavjain97 (changes requested):

modules/sdk-core/src/bitgo/wallet/iWallets.ts
zahin-mohammad: This is not trivial to do because there is no util `createStringObject` in the sdk

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
zahin-mohammad: Its fine to have a string here, if this function returns true it just says that the input string is a valid BackupProvider.

pranavjain97 (approved):

modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -21,16 +21,17 @@ export interface GenerateMpcWalletOptions {
   originalPasscodeEncryptionCode?: string;
   enterprise?: string;
   walletVersion?: number;
-  backupProvider?: string;
+  backupProvider?: BackupProvider;
 }
-
+export const backupProviders = ['BitGoTrustAsKrs'] as const;
+export type BackupProvider = typeof backupProviders[number];
```
sachushaji, line 27: NIT. This could be a bit simpler
```suggestion
export type BackupProvider = 'BitGoTrustAsKrs'
```
Adding more type would simply be 
`export type BackupProvider = 'BitGoTrustAsKrs' | 'next type'`

sachushaji (approved):

PR #3220, opened 2023-01-27
feat(sdk-coin-atom): skeleton for transaction building and account-lib ATOM

michael-gallegos-bg:
Adds basic functionality for decoding transactions for ATOM transfers.
Also added skeleton for transaction builder factory, transfer builder, and account-lib exposure.

Will be working on adding additional testing in follow up PRs, working out skeleton right now to unblock account-lib KeyPair access on the WP side.

TICKET: BG-64363

### DIFF
--- a/modules/sdk-coin-atom/src/lib/constants.ts
+++ b/modules/sdk-coin-atom/src/lib/constants.ts
@@ -1,5 +1,4 @@
 export const DEFAULT_SEED_SIZE_BYTES = 16;
-export const testnetRpc = 'rpc.sentry-01.theta-testnet.polypore.xyz:26657';
 export enum AtomTransactionType {
   Pay = 'Pay',
 }
--- a/modules/sdk-coin-atom/src/lib/constants.ts
+++ b/modules/sdk-coin-atom/src/lib/constants.ts
@@ -1,1 +1,7 @@
 export const DEFAULT_SEED_SIZE_BYTES = 16;
+export const testnetRpc = 'rpc.sentry-01.theta-testnet.polypore.xyz:26657';
+export enum AtomTransactionType {
+  Pay = 'Pay',
+}
+export const sendMsgType = '/cosmos.bank.v1beta1.MsgSend';
+export const thetaTestnetChainId = 'theta-testnet-001';
--- a/modules/sdk-coin-atom/src/lib/iface.ts
+++ b/modules/sdk-coin-atom/src/lib/iface.ts
@@ -1,0 +1,35 @@
+import { TransactionExplanation as BaseTransactionExplanation, TransactionType } from '@bitgo/sdk-core';
+import { Coin, SignerData } from '@cosmjs/stargate';
+
+export interface TransactionExplanation extends BaseTransactionExplanation {
+  type: TransactionType;
+}
+
+export interface MessageData {
+  typeUrl: string;
+  value: {
+    fromAddress: string;
+    toAddress: string;
+    amount: Coin[];
+  };
+}
+
+export interface GasFeeLimitData {
+  amount: Coin[];
+  gas: number; // gas limit
+}
+
+/**
+ * The transaction data returned from the toJson() function of a transaction
+ */
+export interface TxData extends AtomTransaction {
+  id?: string;
+}
+
+export interface AtomTransaction {
+  type: string;
+  signerAddress: string;
+  explicitSignerData: SignerData;
+  sendMessages: MessageData[];
+  gasBudget: GasFeeLimitData;
+}
--- a/modules/sdk-coin-atom/src/lib/transaction.ts
+++ b/modules/sdk-coin-atom/src/lib/transaction.ts
@@ -1,0 +1,185 @@
+import {
+  BaseKey,
+  BaseTransaction,
+  InvalidTransactionError,
+  ParseTransactionError,
+  PublicKey as BasePublicKey,
+  Signature,
+  TransactionRecipient,
+  TransactionType,
+} from '@bitgo/sdk-core';
+import { AtomTransaction, TransactionExplanation, TxData } from './iface';
+import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import { Buffer } from 'buffer';
+import { AtomTransactionType, sendMsgType } from './constants';
+import utils from './utils';
+import { encodePubkey, makeSignBytes } from '@cosmjs/proto-signing';
+import { encodeSecp256k1Pubkey } from '@cosmjs/amino';
+
+export class Transaction extends BaseTransaction {
+  private _atomTransaction: AtomTransaction;
+  private _signature: Signature;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  get atomTransaction(): AtomTransaction {
+    return this._atomTransaction;
+  }
+
+  setAtomTransaction(tx: AtomTransaction): void {
+    this._atomTransaction = tx;
+  }
+
+  /** @inheritDoc **/
+  get id(): string {
+    return this._id || 'UNAVAILABLE_TEXT';
+  }
+
+  addSignature(publicKey: BasePublicKey, signature: Buffer): void {
+    this._signatures.push(signature.toString('hex'));
+    this._signature = { publicKey, signature };
+    this.serialize();
+  }
+
+  get atomSignature(): Signature {
+    return this._signature;
+  }
+
+  /** @inheritdoc */
+  canSign(key: BaseKey): boolean {
+    return true;
+  }
+
+  /** @inheritdoc */
+  toBroadcastFormat(): string {
+    if (!this._atomTransaction) {
+      throw new InvalidTransactionError('Empty transaction');
+    }
+    return this.serialize();
+  }
+
+  /** @inheritdoc */
+  toJson(): TxData {
+    if (!this._atomTransaction) {
+      throw new ParseTransactionError('Empty transaction');
+    }
+    const tx = this._atomTransaction;
+    return {
+      id: this._id,
+      type: tx.type,
+      signerAddress: tx.signerAddress,
+      explicitSignerData: tx.explicitSignerData,
+      sendMessages: tx.sendMessages,
+      gasBudget: tx.gasBudget,
+    };
+  }
+
+  /** @inheritDoc */
+  explainTransaction(): TransactionExplanation {
+    const result = this.toJson();
+    const displayOrder = ['id', 'outputs', 'outputAmount', 'changeOutputs', 'changeAmount', 'fee', 'type'];
+    const outputs: TransactionRecipient[] = [];
+
+    const explanationResult: TransactionExplanation = {
+      displayOrder,
+      id: this.id,
+      outputs,
+      outputAmount: '0',
+      changeOutputs: [],
+      changeAmount: '0',
+      fee: { fee: this.atomTransaction.gasBudget.gas.toString() }, // TODO - explore the fee explanation more, do we need to show limit or base amount
+      type: this.type,
+    };
+
+    switch (this.type) {
+      case TransactionType.Send:
+        return this.explainTransferTransaction(result, explanationResult);
+      default:
+        throw new InvalidTransactionError('Transaction type not supported');
+    }
+  }
+
+  /**
+   * Set the transaction type.
+   *
+   * @param {TransactionType} transactionType The transaction type to be set.
+   */
+  transactionType(transactionType: TransactionType): void {
+    this._type = transactionType;
+  }
+
+  /**
+   * Sets this transaction payload
+   *
+   * @param rawTransaction
+   */
+  fromRawTransaction(rawTransaction: string): void {
+    try {
+      this._atomTransaction = Transaction.deserializeAtomTransaction(rawTransaction);
+      this._type = TransactionType.Send;
+    } catch (e) {
+      throw e;
+    }
+  }
+
+  serialize(): string {
+    const pubkey = encodePubkey(
+      encodeSecp256k1Pubkey(Buffer.from(this._atomTransaction.sendMessages[0].value.fromAddress, 'hex'))
+    ) as unknown as string;
+    const signDoc = utils.createSignDocFromAtomTransaction(pubkey, this.atomTransaction);
+    if (!this._signature) {
+      // no signature yet, get the signable bytes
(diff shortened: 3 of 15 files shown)


modules/sdk-coin-atom/src/lib/iface.ts
```diff
@@ -0,0 +1,35 @@
...
+export interface GasFeeLimitData {
+  amount: Coin[];
+  gas: number; // gas limit
+}
+
+/**
+ * The transaction data returned from the toJson() function of a transaction
+ */
+export interface TxData extends AtomTransaction {
+  id?: string;
```
michael-gallegos-bg, line 26: Need to refactor this in the future to rename the fields if we wish

sachushaji (dismissed):
Package.json and lock changes looks good to me

modules/sdk-coin-atom/src/lib/iface.ts
```diff
@@ -0,0 +1,35 @@
...
+  value: {
+    fromAddress: string;
+    toAddress: string;
+    amount: Coin[];
+  };
+}
+
+export interface GasFeeLimitData {
+  amount: Coin[];
+  gas: number; // gas limit
```
Vijay-Jagannathan, line 19: Based on our sandbox scripts, `gasLimit` seems like a `string`

modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export class Transaction extends BaseTransaction {
+  private _atomTransaction: AtomTransaction;
+  private _signature: Signature;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  get atomTransaction(): AtomTransaction {
```
Vijay-Jagannathan, line 27: Would be great if we had something like [Lombok](https://projectlombok.org/features/GetterSetter) in typescript to reduce boilerplate and look easy on the eye, but I'm not sure if there's a decent library that does that

modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  /**
+   * Returns a complete explanation for a transfer transaction
+   * Currently only supports one message per transfer.
+   * @param {TxData} json The transaction data in json format
+   * @param {TransactionExplanation} explanationResult The transaction explanation to be completed
+   * @returns {TransactionExplanation}
+   */
+  explainTransferTransaction(json: TxData, explanationResult: TransactionExplanation): TransactionExplanation {
+    const outputAmount = this.atomTransaction.sendMessage.value.amount[0].amount;
```
Vijay-Jagannathan, line 180: More of a question for my understanding since I'm new to typescript:

I see that we've initialized a class variable called `_atomTransaction` with a leading underscore to indicate that it's private. However, in line 180 and 183, I see that underscore is not present. So my question is whether this is interchangeable or a typo ? If it's interchangeable, could we stick to one way of referencing the class variable (either with _ or the regular way) ?

modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -0,0 +1,193 @@
...
+    this.serialize();
+  }
+
+  get atomSignature(): Signature {
+    return this._signature;
+  }
+
+  /** @inheritdoc */
+  canSign(key: BaseKey): boolean {
+    return true;
```
Vijay-Jagannathan, line 52: Is this temporary or is it always supposed to return true ?

modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -0,0 +1,193 @@
...
+    // TODO - support many messages decoding, right now we are just taking one message for send tx type
+    const encodedMessages = decodedTx.body.messages as EncodeObject[];
+    const decodedMessage = registry.decode(encodedMessages[0]); // TODO - do we need this?
+    let type: AtomTransactionType;
+    if (encodedMessages[0].typeUrl === sendMsgType) {
+      type = AtomTransactionType.Pay;
+    } else {
+      throw new Error('Transaction type not supported: ' + encodedMessages[0].typeUrl);
+    }
+    const sendMessageData: MessageData = {
```
Vijay-Jagannathan, line 145: Could we have a helper method which does this `MessageData` construction for us instead of the `deseralizeAtomTransaction` method doing it ?

modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -0,0 +1,193 @@
...
+    };
+    const sequence = Number(decodedTx.authInfo.signerInfos[0].sequence);
+    const explicitSignerData: SignerData = {
+      accountNumber: 100, // TODO - get this from tx data somehow
+      sequence,
+      chainId: thetaTestnetChainId, // TODO - figure out if we can get this from tx data
+    };
+    const gasBudget: GasFeeLimitData = {
+      amount: decodedTx.authInfo.fee?.amount as Coin[],
+      gas: Number(decodedTx.authInfo.fee?.gasLimit),
```
Vijay-Jagannathan, line 161: Similar to the comment in the interface file

modules/sdk-coin-atom/src/lib/iface.ts
michael-gallegos-bg: I think it might be getting converted in their SDK, because in my debugging when deserializing from an encoded base64 string, it came back as a number.

modules/sdk-coin-atom/src/lib/transaction.ts
michael-gallegos-bg: This will be abstracted more anyways via the transaction builder and the factory.

modules/sdk-coin-atom/src/lib/transaction.ts
michael-gallegos-bg: Yes, this was just an initial draft, will be modularizing it today.

modules/sdk-coin-atom/src/lib/transaction.ts
michael-gallegos-bg: From my comment before, it was coming back as a Long, so I cast it to a number type, but we can keep it as a long or even convert to a string if we wish.

modules/sdk-coin-atom/src/lib/transaction.ts
Vijay-Jagannathan: Ah gotcha, I read the `get atomTransaction()` as `getAtomTransaction()`. Now it makes sense

modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  get atomTransaction(): AtomTransaction {
+    return this._atomTransaction;
+  }
+
+  setAtomTransaction(tx: AtomTransaction): void {
```
Vijay-Jagannathan, line 31: Based on the getter method, it seems like `set atomTransaction(...)` would be better in terms of consistency. Not a blocking change though, can be addressed in a follow up.

modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  /** @inheritDoc */
+  explainTransaction(): TransactionExplanation {
+    const result = this.toJson();
+    const displayOrder = ['id', 'outputs', 'outputAmount', 'changeOutputs', 'changeAmount', 'fee', 'type'];
+    const outputs: TransactionRecipient[] = [];
+
+    const explanationResult: TransactionExplanation = {
+      displayOrder,
```
Vijay-Jagannathan, line 86: Question: It does seem like this `TransactionExplanation` object is constructed in such a way that there is inherent ordering as per the `displayOrder` constant. Do we still need the constant for some other purpose ?

sachushaji (commented):
Flushing review queue, Re-request when ready

modules/sdk-coin-atom/src/lib/iface.ts
dpkjnr: if it came back as long then BigInt would be better option I think. JS number type is notorious for precision issues

modules/sdk-coin-atom/src/lib/transaction.ts
michael-gallegos-bg: I'm not entirely sure, we can ask the greater team and see if we can get more insight, but the constant is present for all other coins' implementation of this specific method.

modules/sdk-coin-atom/test/unit/transaction.ts
```diff
@@ -0,0 +1,44 @@
...
+    });
+  });
+
+  describe('From raw transaction', () => {
+    it('should build a transfer from raw signed base64', function () {
+      tx.fromRawTransaction(testData.TEST_TX.signedTxBase64);
+      const json = tx.toJson();
+      should.equal(json.signerAddress, testData.TEST_TX.sender);
+    });
+    // it('should build a transfer from raw unsigned base64', function () {
```
michael-gallegos-bg, line 27: Will be adding more testing in a follow up PR, right now getting the skeleton's out to get closer to unblocking wallet creation via account lib in WP

modules/sdk-coin-atom/src/lib/transferBuilder.ts
```diff
@@ -0,0 +1,13 @@
+import { TransactionBuilder } from './transactionBuilder';
+import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import { TransactionType } from '@bitgo/sdk-core';
+
+export class TransferBuilder extends TransactionBuilder {
```
Vijay-Jagannathan, line 5: Question: Any specific reason why we have a separate `TransferBuilder` class to just do a get `transactionType` or is this how other coins are doing it ?

Vijay-Jagannathan (approved):

modules/sdk-coin-atom/src/lib/constants.ts
```diff
@@ -1 +1,7 @@
 export const DEFAULT_SEED_SIZE_BYTES = 16;
+export const testnetRpc = 'rpc.sentry-01.theta-testnet.polypore.xyz:26657';
```
sachushaji, line 2: Whats the need of hardcoding a RPC endpoint here?

modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -0,0 +1,228 @@
...
+  getGasBudgetFromDecodedTx(decodedTx: DecodedTxRaw): GasFeeLimitData {
+    return {
+      amount: decodedTx.authInfo.fee?.amount as Coin[],
+      gas: Number(decodedTx.authInfo.fee?.gasLimit),
+    };
+  }
+
+  getExplicitSignerDataFromDecodedTx(decodedTx: DecodedTxRaw): SignerData {
+    const sequence = this.getSequenceFromDecodedTx(decodedTx);
+    return {
```
sachushaji, line 172: Can you assign ticket numbers for this TODOs?

modules/sdk-coin-atom/test/unit/transaction.ts
sachushaji: Can you add ticket numbers for this follow ups

modules/sdk-coin-atom/src/lib/constants.ts
sachushaji, line 2: Whats the need of hardcoding a RPC endpoint here?

modules/sdk-coin-atom/src/lib/utils.ts
sachushaji, line 172: Can you assign ticket numbers for this TODOs?

modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -0,0 +1,228 @@
...
+  getMessageDataFromDecodedTx(decodedTx: DecodedTxRaw): MessageData[] {
+    const encodedMessages = this.getEncodedMessagesFromDecodedTx(decodedTx);
+    return this.getMessageDataFromEncodedMessages(encodedMessages);
+  }
+
+  createSignDocFromAtomTransaction(pubkey: any, atomTransaction: AtomTransaction) {
+    // TODO - validate that this is a valid atom transaction class
+    const register = new Registry(defaultRegistryTypes);
+    const txBodyBytes = register.encodeTxBody(atomTransaction.sendMessages as any);
+    console.log(register.decodeTxBody(txBodyBytes));
```
sachushaji, line 192: Kindly remove Console statements

michael-gallegos-bg: Good call! Thank you

modules/sdk-coin-atom/src/lib/utils.ts
michael-gallegos-bg: Yes will do shortly

modules/sdk-coin-atom/src/lib/transferBuilder.ts
michael-gallegos-bg: I think when we get further along the line, we will need to decorate the transaction depending on the type we are building (transfer, staking, other)

And yes this is how other coins do this

sachushaji (approved):
Thanks for the hard work @michael-gallegos-bg. One thong to not is that the PR has only minimal tests for a lot of functionalities which you added. I think you can add more for the functions which you have implemented, But I see a follow up ticket mentioned in the test file for it, So I'm not blocking this PR.

modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -0,0 +1,226 @@
...
+  validateRawTransaction(rawTransaction: string): void {
+    if (!rawTransaction) {
+      throw new ParseTransactionError('Invalid raw transaction: Undefined');
+    }
+    if (!this.isValidRawTransaction(rawTransaction)) {
+      throw new ParseTransactionError('Invalid raw transaction');
+    }
+  }
+
+  isValidAddress(address: string): boolean {
```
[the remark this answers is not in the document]
michael-gallegos-bg: Yes this is a follow up ticket that we have, this is the next step we will be working on - just wanted to get this out in account-lib for now to unblock KeyPair access on the WP side. Thanks!

PR #3221, opened 2023-01-27
feat(sdk-core): ecdsa signing use derivation path

olibeyene:
Ticket: BG-66641

### DIFF
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
@@ -730,18 +730,19 @@ describe('TSS Ecdsa Utils:', async function () {
     });
 
     async function setupSignTxRequestNocks(isTxRequest = true) {
-      let response = { txRequests: [{ ...txRequest, transactions: [{ ...txRequest, unsignedTx: { signableHex: txRequest.unsignedTxs[0].signableHex, serializedTxHex: txRequest.unsignedTxs[0].serializedTxHex } }] }] };
+      const derivationPath = '';
+      let response = { txRequests: [{ ...txRequest, transactions: [{ ...txRequest, unsignedTx: { signableHex: txRequest.unsignedTxs[0].signableHex, serializedTxHex: txRequest.unsignedTxs[0].serializedTxHex, derivationPath } }] }] };
       if (isTxRequest) {
         await nockGetTxRequest({ walletId: wallet.id(), txRequestId: txRequest.txRequestId, response: response });
       }
       const aRecord = ECDSAMethods.convertAShare(mockAShare);
       const signatureShares = [aRecord];
       txRequest.signatureShares = signatureShares;
-      response = { txRequests: [{ ...txRequest, transactions: [{ ...txRequest, unsignedTx: { signableHex: txRequest.unsignedTxs[0].signableHex, serializedTxHex: txRequest.unsignedTxs[0].serializedTxHex } }] }] };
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -583,11 +583,14 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
-    const signingKey = MPC.keyCombine(userSigningMaterial.pShare, [
-      userSigningMaterial.bitgoNShare,
-      userSigningMaterial.backupNShare,
-    ]);
+    const keyDerive = MPC.keyDerive(
+      userSigningMaterial.pShare,
+      [userSigningMaterial.bitgoNShare, userSigningMaterial.backupNShare],
+      derivationPath
+    );
+    const signingKey = MPC.keyCombine(userSigningMaterial.pShare, [keyDerive.nShares[3], keyDerive.nShares[2]]);
```
johnoliverdriscoll, line 593: Not necessary for the user to call `keyCombine` here. In fact, the signingKey will not produce valid signing shares for the expected key. The user already has their signing material as `keyDerive.xShare`.

johnoliverdriscoll (changes requested):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -612,18 +612,44 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
+      );
+      const unsignedTx =
+        txRequestResolved.apiVersion === 'full'
+          ? txRequestResolved.transactions![0].unsignedTx
+          : txRequestResolved.unsignedTxs[0];
+      signablePayload = Buffer.from(unsignedTx.signableHex, 'hex');
+      derivationPath = txRequestResolved.transactions![0].unsignedTx.derivationPath;
+    } else if (requestType === RequestType.message) {
+      signablePayload = (params as TSSParamsForMessage).bufferToSign;
+      // TODO Message singing with derivation path
```
olibeyene, line 631: @jeanlarrieux  can you take look at what it takes to add derivation path to message signing?

wesgraham (approved):

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -730,18 +730,19 @@ describe('TSS Ecdsa Utils:', async function () {
     });
 
     async function setupSignTxRequestNocks(isTxRequest = true) {
-      let response = { txRequests: [{ ...txRequest, transactions: [{ ...txRequest, unsignedTx: { signableHex: txRequest.unsignedTxs[0].signableHex, serializedTxHex: txRequest.unsignedTxs[0].serializedTxHex } }] }] };
+      const derivationPath = '';
```
jicli, line 733: ideally would love to see an example with actual derivation path test case

jicli (approved):

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
sachushaji: +1

johnoliverdriscoll (approved):
lgtm, given that this does not actually transmit a derivation path, and simply provides a no-op parameter

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
olibeyene: I will add additional test in follow up pr BG-67621

byhow (approved):

PR #3223, opened 2023-01-30
fix(sdk-coin-bsc): build from txHex

mullapudipruthvik:
Ticket: BG-66898

Allow txs to build from txHex for coins without
factoryImplementationAddress

### DIFF
--- a/modules/sdk-coin-eth/src/lib/transactionBuilder.ts
+++ b/modules/sdk-coin-eth/src/lib/transactionBuilder.ts
@@ -236,7 +236,11 @@ export class TransactionBuilder extends BaseTransactionBuilder {
           this.forwarderVersion(1);
           this.baseAddress(baseAddress);
           this.salt(addressCreationSalt);
-          this.initCode((this._coinConfig.network as EthereumNetwork).forwarderImplementationAddress as string);
+          const forwarderImplementationAddress = (this._coinConfig.network as EthereumNetwork)
+            .forwarderImplementationAddress as string;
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-eth/src/lib/transactionBuilder.ts
```diff
@@ -236,7 +236,11 @@ export class TransactionBuilder extends BaseTransactionBuilder {
           this.forwarderVersion(1);
           this.baseAddress(baseAddress);
           this.salt(addressCreationSalt);
-          this.initCode((this._coinConfig.network as EthereumNetwork).forwarderImplementationAddress as string);
+          const forwarderImplementationAddress = (this._coinConfig.network as EthereumNetwork)
```
olibeyene, line 239: BSC coin doesn't need address initialization and forwarders are also not implemented yet. I don't understand this pr. Can you elaborate the goal of this pr?

olibeyene (changes requested):
BSC is not enabled, please explain the goal of this pr.

modules/sdk-coin-eth/src/lib/transactionBuilder.ts
mullapudipruthvik: Hey we need bsc txBuilder the ability to build `AddressInitialization`, `WalletInitialization` similar to eth txBuilder.
This is required for cross chain recovery as we need to deploy v0/v1 wallet and forwarder to perform recovery

olibeyene (approved):

PR #3224, opened 2023-01-30
feat(sdk-core): add ecdsa to verifyShareProof()

brekiy:
Adds an ECDSA path to the verifyShareProof function to be used for verifying
ECDSA keyshares.

Ticket: BG-65151

BREAKING CHANGE: The `verifyEdShareProof` function is renamed to
`verifyShareProof` and now expects a 4th parameter which must be a
string of either `eddsa` or `ecdsa`.

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
@@ -48,17 +48,19 @@ export async function getTrustGpgPubKey(bitgo: BitGoBase): Promise<Key> {
 }
 
 /**
- * Verify an Eddsa KeyShare Proof.
+ * Verify an Eddsa or Ecdsa KeyShare Proof.
  *
  * @param senderPubKey public key of the sender of the privateShareProof
  * @param privateShareProof u value proof
  * @param uValue u value from an Eddsa keyshare
+ * @param algo
  * @return {boolean} whether uValue proof actually was signed by sender as part of their subkeys
  */
-export async function verifyEdShareProof(
+export async function verifyShareProof(
   senderPubKey: string,
   privateShareProof: string,
-  uValue: string
+  uValue: string,
+  algo: 'eddsa' | 'ecdsa'
 ): Promise<boolean> {
   const decodedProof = await pgp.readKey({ armoredKey: privateShareProof });
   const senderGpgKey = await pgp.readKey({ armoredKey: senderPubKey });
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -61,6 +61,30 @@ export async function verifyEdShareProof(
...
 
+/**
+ * Verify an Ecdsa KeyShare Proof.
+ *
+ * @param senderPubKey public key of the sender of the privateShareProof
+ * @param privateShareProof u value proof
+ * @param uValue u value from an Ecdsa keyshare
+ * @return {boolean} whether uValue proof actually was signed by sender as part of their subkeys
+ */
+export async function verifyEcShareProof(
```
jicli, line 72: since majority of the verify is the same 
we can do with `verifyShareProof(..., algo)`?

brekiy: good point

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -36,29 +36,39 @@ export async function getBitgoGpgPubKey(bitgo: BitGoBase): Promise<Key> {
...
-  return decodedUValueProof === rawUValueProof;
+  if (algo === 'eddsa') {
+    const decodedUValueProof = Buffer.from(proofSubkeys.keyPacket.publicParams['Q'].slice(1)).toString('hex');
+    const rawUValueProof = Buffer.from(
+      sodium.crypto_scalarmult_ed25519_base_noclamp(Buffer.from(uValue, 'hex'))
+    ).toString('hex');
+    return decodedUValueProof === rawUValueProof;
+  } else if (algo === 'ecdsa') {
+    const decodedUValueProof = Buffer.from(proofSubkeys.keyPacket.publicParams['Q']).toString('hex');
+    const rawUValueProof = secp256k1.pointFromScalar(Buffer.from(uValue, 'hex'), false);
```
Logicwax, line 67: is `secp256k1.publicKeyCreate()` not available?

brekiy: Not in the used library (`modules/utxo-lib-src/noble_ecc.ts`)

brekiy: This is `pointFromScalar()`'s definition:
https://github.com/BitGo/BitGoJS/blob/master/modules/utxo-lib/src/noble_ecc.ts#L50-L51

It uses `getPublicKey()` from `noble-secp256k1`.
https://www.npmjs.com/package/@noble/secp256k1#getpublickeyprivatekey

Logicwax (approved):

brekiy:
Rebased to fix merge conflict

islamaminBitGo (approved):

byhow (changes requested):
Changes LGTM; breaking changes for extra parameters that are added to the function needs to have `BREAKING:` in the commit message unless this function is not public facing

islamaminBitGo:
> Changes LGTM; breaking changes for extra parameters that are added to the function needs to have `BREAKING:` in the commit message unless this function is not public facing

i can confirm that it is public facing as im able to use it outside of the sdk. I agree with adding `BREAKING:` in the commit message

byhow (approved):
Appreciate the change!

PR #3230, opened 2023-01-31
fix(sdk-core): correct ecdsa verifyWalletSigs()

brekiy:
Corrects verification of wallet signatures received from the HSM during TSS ECDSA wallet
generation.

BG-68087

### DIFF
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
@@ -26,6 +26,7 @@ import {
 import { keyShares, mockAShare, mockDShare, otherKeyShares } from '../../../fixtures/tss/ecdsaFixtures';
 import { nockSendSignatureShareWithResponse } from './common';
 import { createWalletSignatures, nockGetTxRequest } from '../../tss/helpers';
+import { ecc } from '@bitgo/utxo-lib';
 
 const encryptNShare = ECDSAMethods.encryptNShare;
 type KeyShare = ECDSA.KeyShare;
@@ -909,6 +910,7 @@ describe('TSS Ecdsa Utils:', async function () {
           privateShare: userToBitgoShare.encryptedPrivateShare,
           n: userToBitgoShare.n,
           vssProof: userToBitgoShare.vssProof,
+          privateShareProof: userToBitgoShare.privateShareProof,
         },
         {
           from: 'bitgo',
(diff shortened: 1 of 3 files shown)


modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -909,8 +911,8 @@ describe('TSS Ecdsa Utils:', async function () {
         { name: 'commonKeychain', value: bitgoCombined.xShare.y + bitgoCombined.xShare.chaincode },
         { name: 'userKeyId', value: userKeyId },
         { name: 'backupKeyId', value: backupKeyId },
-        { name: 'bitgoToUserPublicShare', value: userToBitgoShare.publicShare },
-        { name: 'bitgoToBackupPublicShare', value: backupToBitgoShare.publicShare },
+        { name: 'bitgoToUserPublicShare', value: userToBitgoShare.privateShareProof as string },
```
brekiy, line 914: This `privateShareProof` is generated via `createShareProof()` in `opengpgutils`:
https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/utils/opengpgUtils.ts#L170

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -748,14 +748,19 @@ export async function verifyWalletSignature(params: {
...
   userKeyId: string;
   backupKeyId: string;
-  publicShare: string;
+  decryptedShare: string;
   verifierIndex: 1 | 2;
 }): Promise<void> {
   const rawNotations = await commonVerifyWalletSignature(params);
   const publicShareRawNotationIndex = 2 + params.verifierIndex;
 
+  // Read the u value of the share
```
Logicwax, line 757: wouldn't this be "Derive public form of u-value" ?

brekiy:
separating some work to rebase on master

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -748,14 +748,19 @@ export async function verifyWalletSignature(params: {
...
   const rawNotations = await commonVerifyWalletSignature(params);
-  const publicShareRawNotationIndex = 2 + params.verifierIndex;
+  const publicUValueRawNotationIndex = 2 + params.verifierIndex;
 
+  // Derive public form of u-value
+  const publicUValue = ecc.pointFromScalar(Buffer.from(params.decryptedShare.slice(0, 64), 'hex'), false);
+  // Verify that the u value + chaincode is equal to the proof retrieved from the raw notations
   assert(
-    params.publicShare === Buffer.from(rawNotations[publicShareRawNotationIndex].value).toString(),
+    publicUValue !== null &&
```
Logicwax, line 761: one thing missing in this function is to check the gpg pubkey Ids in the notation data.     in `verifyWalletSignatures()` you verified that the wallet sigs contain two distinct pgp keychains.      and you checked that the pubkeys themselves match the two we are interested in.

but in each keychain ([0] and [1] in verifyWalletSignatures()), there is notation data of not just the pub u-value and commonkeychain, but also the two gpg key Ids used as well.    

add a check to make sure those match up.

brekiy: This should actually be covered under `commonVerifyWalletSignature()`, which checks the first 3 notation data:
https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/tss/common.ts#L125-L150

byhow (approved):
code change LGTM but have no context on the implementation; will defer to @Logicwax for final approval

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
Logicwax: oh, perfect.

Logicwax (approved):
lgtm

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -748,14 +748,19 @@ export async function verifyWalletSignature(params: {
...
+      publicUValue.toString() + params.decryptedShare.slice(64) ===
```
islamaminBitGo, line 762: Since the chaincode is being provided in hex encoding, the publicUValue buffer needs to also be in hex encoding so please use .toString('hex') instead of .toString()
@Logicwax : context is that this fails right now while testing as the encoding for this is not correct. 
@brekiy is creating a new PR to fix this.

PR #3237, opened 2023-02-02
feat(sdk-coin-atom): create wallet script

mukeshsp:
Ticket: BG-67048

### DIFF
--- a/examples/js/atom/create-wallet.js
+++ b/examples/js/atom/create-wallet.js
@@ -1,0 +1,56 @@
+/**
+ * Create a TSS atom wallet at BitGo.
+ * This makes use of the convenience function generateWallet
+ *
+ * This tool will help you see how to use the BitGo API to easily create a wallet.
+ * In this form, it creates 2 keys on the host which runs this example.
+ * It is HIGHLY RECOMMENDED that you GENERATE THE KEYS ON SEPARATE MACHINES for real money wallets!
+ *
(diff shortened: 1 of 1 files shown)


Vijay-Jagannathan (approved):

examples/js/atom/create-wallet.js
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * To perform more advanced features, such as encrypting keys yourself, please look at createWalletAdvanced.js
+ *
+ * Copyright 2023, BitGo, Inc.  All Rights Reserved.
+ */
+const BitGo = require('bitgo');
+const bitgo = new BitGo.BitGo({
+  env: 'custom',
+  customRootURI: '',
+});
```
sachushaji, line 17: NIT: probably best use test env as this would be referred by the public and they almost never will have to use a custom URI
```suggestion
const bitgo = new BitGo.BitGo({
  env: 'test',
});
```

sachushaji (commented):
Can you also add a typescript example also

examples/js/atom/create-wallet.js
mukeshsp: sure, I will do it in next PR 
here is ticket : https://bitgoinc.atlassian.net/browse/BG-67328

sachushaji: Okay

sachushaji (approved):

PR #3239, opened 2023-02-02
feat(key-card): add code to select qrData

LeoKotschenreuther:
The added functions receive the keycards and some optional extra fields as an input and return an object that can then be passed to the drawKeycard function (to be added in a future PR) to fill the data for the different boxes.

Ticket: BG-64432

### DIFF
--- a/modules/key-card/src/generateQrData.ts
+++ b/modules/key-card/src/generateQrData.ts
@@ -1,0 +1,164 @@
+import { BaseCoin } from '@bitgo/statics';
+import { Keychain } from '@bitgo/sdk-core';
+import { encrypt } from '@bitgo/sdk-api';
+import * as assert from 'assert';
+
+export interface GenerateQrDataParams {
+  // The backup keychain as it is returned from the BitGo API upon creation
+  backupKeychain: Keychain,
+  // The 3rd party provider of the backup key if neither the user nor BitGo stores it
+  backupKeyProvider?: string;
+  // The key id of the backup key, only used for cold keys
+  backupMasterKey?: string;
+  // The BitGo keychain as it is returned from the BitGo API upon creation
+  bitgoKeychain: Keychain,
+  // The coin of the wallet that was/ is about to be created
+  coin: Readonly<BaseCoin>;
+  // A code that can be used to encrypt the wallet password to.
+  // If both the passphrase and passcodeEncryptionCode are passed, then this code encrypts the passphrase with the
+  // passcodeEncryptionCode and puts the result into Box D. Allows recoveries of the wallet password.
+  passcodeEncryptionCode?: string;
+  // The wallet password
+  // If both the passphrase and passcodeEncryptionCode are passed, then this code encrypts the passphrase with the
+  // passcodeEncryptionCode and puts the result into Box D. Allows recoveries of the wallet password.
+  passphrase?: string;
+  // The user keychain as it is returned from the BitGo API upon creation
+  userKeychain: Keychain;
+  // The key id of the user key, only used for cold keys
+  userMasterKey?: string;
+}
+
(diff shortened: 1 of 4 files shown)


LeoKotschenreuther:
> does this account for tss keychains? IIRC they need to split because the data is too large for one qr image.

That will be done in another PR. This PR is only about selecting the data to put on the key card, data splitting will happen during a future PR where I introduce the code that will actually draw the key card

LeoKotschenreuther:
If it helps, I'm recreating roughly the following code here:
https://github.com/BitGo/bitgo-microservices/blob/c7269431b91295308d0765862b3663ed35852502/packages/client-app-dashboard/app/shared/components/add-wallet-form/key-card-pdf.ts#L165-L285

I removed the OFC specific things (because we don't have key cards for trading accounts) as well as the stellar & ripple specific code because @zahin-mohammad said it was no longer used.

sachushaji (dismissed):

modules/key-card/src/generateQrData.ts
```diff
@@ -0,0 +1,171 @@
...
+function getPubFromKey(key: Keychain): string | undefined {
+  if (key.type === 'blsdkg' && key.commonPub) {
+    return key.commonPub;
+  }
+  if (key.type === 'tss' && key.commonKeychain) {
+    return key.commonKeychain;
+  }
+  if (key.type === 'independent' && key.pub) {
+    return key.pub;
+  }
```
mohammadalfaiyazbitgo, line 55:
```suggestion
  switch(key.type) {
      case 'blsdkg':
         return key.commonPub;
       ....
  }
```
nit: the check to see if `commonPub` and other fields exists is redundant since this would return `undefined` for those cases anyways.

modules/key-card/src/generateQrData.ts
```diff
@@ -0,0 +1,171 @@
...
+      title: 'B: Backup Key',
+      image: '.qrPublicBackupKey',
+      description:
+        'This is the public key held at ' +
+        backupKeyProvider +
+        ', an ' +
+        coin.name +
+        ' recovery service. If you lose\r\nyour key, ' +
+        backupKeyProvider +
+        ' will be able to sign transactions to recover funds.',
```
mohammadalfaiyazbitgo, line 111: this could be string interpolated no?

modules/key-card/src/generateQrData.ts
```diff
@@ -0,0 +1,171 @@
...
+  return {
+    title: 'B: Provided Backup Key',
+    image: '.qrEncryptedUserProvidedXpub',
+    description: 'This is the public key you provided for your wallet.',
+    data: pub,
+    publicMasterKey: backupMasterKey,
+  };
+}
+
+function generateBitGoQrData(coin: Readonly<BaseCoin>, bitgoKeychain: Keychain): QrDataEntry | undefined {
```
mohammadalfaiyazbitgo, line 125: this would never return `undefined` ?

mohammadalfaiyazbitgo: also not sure where `coin` is used here.

modules/key-card/src/generateQrData.ts
LeoKotschenreuther: Great idea, will do

modules/key-card/src/generateQrData.ts
LeoKotschenreuther: Good catch. Both were leftover from when I was supporting the OFC special case. Removing them now.

mohammadalfaiyazbitgo (approved):

zahin-mohammad (approved):

sachushaji (approved):

PR #3244, opened 2023-02-03
feat(sdk-core): rename type

LeoKotschenreuther:
Types should start with an upper case letter, otherwise they can
be confused with variable names which usually start with a lower case
letter.

Ticket: BG-64432
BREAKING CHANGE: rename type backupGpgKey to BackupGpgKey

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
@@ -280,7 +280,7 @@ export type Challenge = {
   h2: string;
 };
 
-export type backupGpgKey = SerializedKeyPair<string> | Key;
+export type BackupGpgKey = SerializedKeyPair<string> | Key;
 
 /**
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -280,7 +280,7 @@ export type Challenge = {
   h2: string;
 };
 
-export type backupGpgKey = SerializedKeyPair<string> | Key;
```
sachushaji, line 283: Nice catch!!!, But I think this can be considered as a breaking change since its exported out. Can you add the breaking change into the commit ([Message format](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=in%20Semantic%20Versioning\).-,BREAKING,-CHANGE%3A%20a))

sachushaji (approved):
Thanks Leo for addressing the changes!!

PR #3252, opened 2023-02-06
refactor(sdk-coin-atom): only decode sequence from raw atom tx

michael-gallegos-bg:
We found that we no longer need the accountNumber or the chainId to properly construct a raw tx.

Also adds more testing regarding transactionBuilder and factory building.

TICKET: BG-67573

### DIFF
--- a/modules/sdk-coin-atom/package.json
+++ b/modules/sdk-coin-atom/package.json
@@ -44,7 +44,6 @@
     "@bitgo/statics": "^14.1.0",
     "@bitgo/utxo-lib": "^7.5.0",
     "@cosmjs/amino": "^0.29.5",
-    "@cosmjs/crypto": "^0.29.5",
     "@cosmjs/encoding": "^0.29.5",
     "@cosmjs/proto-signing": "^0.29.5",
     "@cosmjs/stargate": "^0.29.5",
--- a/modules/sdk-coin-atom/src/lib/iface.ts
+++ b/modules/sdk-coin-atom/src/lib/iface.ts
@@ -1,5 +1,5 @@
 import { TransactionExplanation as BaseTransactionExplanation, TransactionType } from '@bitgo/sdk-core';
-import { Coin, SignerData } from '@cosmjs/stargate';
+import { Coin } from '@cosmjs/stargate';
 
 export interface TransactionExplanation extends BaseTransactionExplanation {
   type: TransactionType;
@@ -29,7 +29,7 @@ export interface TxData extends AtomTransaction {
 export interface AtomTransaction {
   type: string;
   signerAddress: string;
-  explicitSignerData: SignerData;
+  sequence: number;
   sendMessages: MessageData[];
   gasBudget: GasFeeLimitData;
 }
--- a/modules/sdk-coin-atom/src/lib/transaction.ts
+++ b/modules/sdk-coin-atom/src/lib/transaction.ts
@@ -13,8 +13,10 @@ import { BaseCoin as CoinConfig } from '@bitgo/statics';
 import { Buffer } from 'buffer';
 import { AtomTransactionType, sendMsgType } from './constants';
 import utils from './utils';
-import { encodePubkey, makeSignBytes } from '@cosmjs/proto-signing';
+import { encodePubkey } from '@cosmjs/proto-signing';
 import { encodeSecp256k1Pubkey } from '@cosmjs/amino';
+import { Any } from 'cosmjs-types/google/protobuf/any';
+import { fromHex } from '@cosmjs/encoding';
 
 export class Transaction extends BaseTransaction {
   private _atomTransaction: AtomTransaction;
@@ -70,7 +72,7 @@ export class Transaction extends BaseTransaction {
       id: this._id,
       type: tx.type,
       signerAddress: tx.signerAddress,
-      explicitSignerData: tx.explicitSignerData,
+      sequence: tx.sequence,
       sendMessages: tx.sendMessages,
       gasBudget: tx.gasBudget,
     };
@@ -118,6 +120,10 @@ export class Transaction extends BaseTransaction {
   fromRawTransaction(rawTransaction: string): void {
     try {
       this._atomTransaction = Transaction.deserializeAtomTransaction(rawTransaction);
+      if (utils.isSignedRawTx(rawTransaction)) {
+        const signerInfo = utils.getSignerInfoFromRawSignedTx(rawTransaction);
+        this.addSignature(signerInfo.pubKey, signerInfo.signature);
+      }
       this._type = TransactionType.Send;
     } catch (e) {
       throw e;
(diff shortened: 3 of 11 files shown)


modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -125,6 +125,8 @@ export class Transaction extends BaseTransaction {
   }
 
   serialize(): string {
+    // TODO - Check if we can get the pub key from the encoded/decoded tx data
```
jicli, line 128: add a ticket for it plz

michael-gallegos-bg: this is a draft PR, consulting with cosmos team on how we wish to proceed

michael-gallegos-bg:
From @jicli : "but we just changed to sequence id?"
Yes, so we used to implement this using the Stargate Client from the Cosmos JS SDK, which requires an account number/chain ID as inputs. However, we do not need this as we are using TSS, so only need the sequence number.
I refactored the code to account for this change.

modules/sdk-coin-atom/src/lib/transactionBuilder.ts
```diff
@@ -153,14 +155,14 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
-      }
-    }
+    // for (const amount of amountArray) {
+    //   if (!amount.amount) {
+    //     throw new BuildTransactionError('Invalid amount: undefined');
+    //   }
+    //   if (!amount.denom) {
+    //     throw new BuildTransactionError('Invalid denom: undefined');
+    //   }
+    // }
```
mukeshsp, line 165: Please remove the commented code and add the ticket number in todo.

modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -203,17 +197,17 @@ export class Utils implements BaseUtils {
...
+    // @ts-ignore
+    return makeSignDoc(txBodyBytes, authInfoBytes, null, null);
   }
 
-  createSignedTxRaw(pubkey: string, signDoc: SignDoc, signature: any): TxRaw {
-    const signatureBytes = new Secp256k1Signature(fromHex(signature.r), fromHex(signature.s)).toFixedLength();
-    const stdSignature = encodeSecp256k1Signature(fromHex(pubkey), signatureBytes);
+  createSignedTxRaw(atomSignature: Signature, signDoc: SignDoc): TxRaw {
+    console.log('atom sig at this point: ' + atomSignature.signature);
+    console.log('hex of that signature: ' + atomSignature.signature.toString('hex'));
```
mukeshsp, line 208: remove console log

modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -203,17 +197,17 @@ export class Utils implements BaseUtils {
...
+    const stdSignature = encodeSecp256k1Signature(fromHex(atomSignature.publicKey.pub), atomSignature.signature);
+    console.log('stdSig: ' + stdSignature.signature);
```
mukeshsp, line 210: remove console

modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -225,6 +219,31 @@ export class Utils implements BaseUtils {
     const signedTxBytes = TxRaw.encode(signedTxRaw).finish();
     return toBase64(signedTxBytes);
   }
+
+  isSignedRawTx(rawTransaction: string): boolean {
+    // if (!this.isValidRawTransaction(rawTransaction)) {
+    //   return false;
+    // }
```
Vijay-Jagannathan, line 226: Good to remove

modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -225,6 +219,31 @@ export class Utils implements BaseUtils {
...
+    const decodedTx = this.getDecodedTxFromRawBase64(rawTransaction);
+    if (decodedTx.authInfo.signerInfos.length > 0) {
+      return true;
+    }
+    return false;
+  }
+
+  getSignerInfoFromRawSignedTx(rawTransaction: string) {
+    if (!this.isSignedRawTx(rawTransaction)) {
```
Vijay-Jagannathan, line 229: This method gets called within `fromRawTransaction` in the transaction class. Before this method is called, we already check if the transaction is indeed signed or not (Line 123 in transaction class), so we might not need line 235 (unless this method can be called from elsewhere where we need to check if the transaction is signed)

mukeshsp: Let's keep this function for now, as we may call it in another class in the future!

modules/sdk-coin-atom/package.json
```diff
@@ -44,7 +44,6 @@
     "@bitgo/statics": "^14.1.0",
     "@bitgo/utxo-lib": "^7.5.0",
     "@cosmjs/amino": "^0.29.5",
-    "@cosmjs/crypto": "^0.29.5",
```
sachushaji, line 47: No package lock changes?

modules/sdk-coin-atom/src/lib/iface.ts
```diff
@@ -29,7 +29,7 @@ export interface TxData extends AtomTransaction {
 export interface AtomTransaction {
   type: string;
   signerAddress: string;
-  explicitSignerData: SignerData;
+  sequence: number;
```
sachushaji, line 32: Seems to be breaking change?, mention in the commit message ([format](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=in%20Semantic%20Versioning\).-,BREAKING,-CHANGE%3A%20a))

sachushaji (changes requested):

modules/sdk-coin-atom/package.json
michael-gallegos-bg: Isn't the lock-file automatically generated?
I do not see it in other coin modules
https://github.com/BitGo/BitGoJS/tree/master/modules/sdk-coin-atom
https://github.com/BitGo/BitGoJS/tree/master/modules/sdk-coin-sol

modules/sdk-coin-atom/src/lib/transactionBuilder.ts
michael-gallegos-bg: Marking as resolved

modules/sdk-coin-atom/src/lib/utils.ts
michael-gallegos-bg: Marking as resolved

modules/sdk-coin-atom/src/lib/utils.ts
michael-gallegos-bg: Marking as resolved

modules/sdk-coin-atom/src/lib/utils.ts
michael-gallegos-bg: Good call - I was trying to explicitly show in this version that currently we are only supporting signed tx data, and are working on unsigned with the HSM team

modules/sdk-coin-atom/src/lib/iface.ts
michael-gallegos-bg: Updated the commit message for the breaking change commit!

mukeshsp (approved):

modules/sdk-coin-atom/package.json
sachushaji: Yeah, Lock files are autogenerated when you do `yarn`  again if you made changes to dependencies. I checkout locally and seems like it's a peer dependencies for some of the packages like `@cosmjs/amino` etc, probably explains why lock file was not modified.

sachushaji (approved):

modules/sdk-coin-atom/package.json
michael-gallegos-bg: Oh yes - they're dependencys of each other, thanks for checking!

PR #3253, opened 2023-02-07
chore(utxo-lib): wrap verifySignature checks for zcash as unsafe

davidkaplanbitgo:
For Zcash transactions, we do not have to have non-witness UTXO data for non-segwit transactions because zcash hashes the value directly. Thus, it is unnecessary to have the previous transaction hash on the unspent. We did this for signing inputs and we need to do it for validating inputs as well.

BG-66521

### DIFF
--- a/modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
@@ -8,6 +8,7 @@ import { Network, PsbtTransaction, Signer } from '../../';
 import { Psbt as PsbtBase } from 'bip174';
 import * as types from 'bitcoinjs-lib/src/types';
 import { UtxoTransaction } from '../UtxoTransaction';
+import { ValidateSigFunction } from 'bitcoinjs-lib/src/psbt';
 const typeforce = require('typeforce');
 
 enum Subtype {
@@ -128,9 +129,17 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
   // transactions because zcash hashes the value directly. Thus, it is unnecessary to have
   // the previous transaction hash on the unspent.
   signInput(inputIndex: number, keyPair: Signer, sighashTypes?: number[]): this {
+    return this.withUnsafeSignNonSegwitTrue(super.signInput.bind(this, inputIndex, keyPair, sighashTypes));
+  }
+
+  validateSignaturesOfInput(inputIndex: number, validator: ValidateSigFunction, pubkey?: Buffer): boolean {
+    return this.withUnsafeSignNonSegwitTrue(super.validateSignaturesOfInput.bind(this, inputIndex, validator, pubkey));
+  }
+
+  private withUnsafeSignNonSegwitTrue<T>(fn: () => T): T {
     (this as any).__CACHE.__UNSAFE_SIGN_NONSEGWIT = true;
     try {
-      return super.signInput(inputIndex, keyPair, sighashTypes);
+      return fn();
     } finally {
       (this as any).__CACHE.__UNSAFE_SIGN_NONSEGWIT = false;
     }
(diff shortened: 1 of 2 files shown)


modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -128,14 +129,30 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
   // transactions because zcash hashes the value directly. Thus, it is unnecessary to have
   // the previous transaction hash on the unspent.
   signInput(inputIndex: number, keyPair: Signer, sighashTypes?: number[]): this {
-    (this as any).__CACHE.__UNSAFE_SIGN_NONSEGWIT = true;
+    this.setUnsafeSignNonSegwit(true);
```
brandonblack, line 132: Let's do this as a `with` now that we're making it cleaner.

brandonblack (changes requested):

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
davidkaplanbitgo: Is it depreciated? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with

lcovar (dismissed):

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -128,9 +129,17 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
...
+
+  validateSignaturesOfInput(inputIndex: number, validator: ValidateSigFunction, pubkey?: Buffer): boolean {
+    return this.withUnsafeSignNonSegwitTrue(super.validateSignaturesOfInput.bind(this, inputIndex, validator, pubkey));
+  }
+
+  private withUnsafeSignNonSegwitTrue(callback: () => any): any {
     (this as any).__CACHE.__UNSAFE_SIGN_NONSEGWIT = true;
     try {
-      return super.signInput(inputIndex, keyPair, sighashTypes);
+      return callback();
```
brandonblack, line 142: hate to be that guy, but callback is a confusing name here. `fn`, `func`, `call`, something that doesn't already have an accepted meaning different from what we're doing.

brandonblack (changes requested):

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
```diff
@@ -128,9 +129,17 @@ export class ZcashPsbt extends UtxoPsbt<ZcashTransaction<bigint>> {
...
   // the previous transaction hash on the unspent.
   signInput(inputIndex: number, keyPair: Signer, sighashTypes?: number[]): this {
+    return this.withUnsafeSignNonSegwitTrue(super.signInput.bind(this, inputIndex, keyPair, sighashTypes));
+  }
+
+  validateSignaturesOfInput(inputIndex: number, validator: ValidateSigFunction, pubkey?: Buffer): boolean {
+    return this.withUnsafeSignNonSegwitTrue(super.validateSignaturesOfInput.bind(this, inputIndex, validator, pubkey));
+  }
+
+  private withUnsafeSignNonSegwitTrue(fn: () => any): any {
```
brandonblack, line 139:
```suggestion
  private withUnsafeSignNonSegwitTrue<T>(fn: () => T): T {
```
Does this work to improve type checking?

brandonblack (changes requested):

modules/utxo-lib/src/bitgo/zcash/ZcashPsbt.ts
davidkaplanbitgo: Yes - this works - and a lot better than how it was before

brandonblack (approved):
Wow, that turned out clean-a-f :-D

PR #3256, opened 2023-02-07
feat(atom): implement basic util methods for atom

mukeshsp:
Ticket: BG-67049

### DIFF
--- a/modules/sdk-coin-atom/src/lib/utils.ts
+++ b/modules/sdk-coin-atom/src/lib/utils.ts
@@ -19,7 +19,7 @@ import { TxRaw, SignDoc } from 'cosmjs-types/cosmos/tx/v1beta1/tx';
 export class Utils implements BaseUtils {
   /** @inheritdoc */
   isValidBlockId(hash: string): boolean {
-    throw new Error('Method not implemented.');
+    return this.validateBlake2b(hash);
   }
 
   /** @inheritdoc */
@@ -39,9 +39,15 @@ export class Utils implements BaseUtils {
 
   /** @inheritdoc */
   isValidTransactionId(txId: string): boolean {
-    throw new Error('Method not implemented.');
+    return this.validateBlake2b(txId);
   }
 
+  validateBlake2b(hash: string): boolean {
+    if (hash?.length !== 64) {
+      return false;
+    }
+    return hash.match(/^[a-zA-Z0-9]+$/) !== null;
+  }
   /**
    * Checks if raw transaction can be deserialized
    *
--- a/modules/sdk-coin-atom/test/unit/utils.ts
+++ b/modules/sdk-coin-atom/test/unit/utils.ts
@@ -1,0 +1,43 @@
+import should from 'should';
+import { address, blockHash, txIds } from '../resources/atom';
+import utils from '../../src/lib/utils';
+
+describe('utils', () => {
+  it('should validate addresses correctly', () => {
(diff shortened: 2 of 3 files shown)


modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -39,9 +39,18 @@ export class Utils implements BaseUtils {
...
+    return this.validateBlake2b(txId);
   }
 
+  validateBlake2b(hash: string): boolean {
+    if (!hash) {
+      return false;
+    }
+    if (hash.length !== 64) {
+      return false;
+    }
```
jicli, line 51: they can be merged into `if(!hash || hash.length !== 64)` or `if(hash?.length !== 64)`

mukeshsp: updated !!

jicli (approved):

modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -39,9 +39,15 @@ export class Utils implements BaseUtils {
 
   /** @inheritdoc */
   isValidTransactionId(txId: string): boolean {
-    throw new Error('Method not implemented.');
+    return this.validateBlake2b(txId);
   }
 
+  validateBlake2b(hash: string): boolean {
```
dpkjnr, line 45: this function can be static. We won't require redundant `this.`  Also, blake2 and sha256 hash has the same size of hash, not sure if we need separate function

modules/sdk-coin-atom/test/unit/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+    should.equal(utils.isValidAddress(undefined as unknown as string), false);
+    should.equal(utils.isValidAddress(''), false);
+  });
+
+  it('should validate block hash correctly', () => {
+    should.equal(utils.isValidBlockId(blockHash.hash1), true);
+    should.equal(utils.isValidBlockId(blockHash.hash2), true);
+    // param is coming as undefined so it was causing an issue
+    should.equal(utils.isValidBlockId(undefined as unknown as string), false);
+    should.equal(utils.isValidBlockId(''), false);
```
Vijay-Jagannathan, line 21: Seems to be covered in the `should validate invalid block hash correctly` block below

modules/sdk-coin-atom/src/lib/utils.ts
Vijay-Jagannathan: +1. Also, seems like we're just checking length along with regex character matching. Any thoughts on why the method is named this way ?

modules/sdk-coin-atom/test/unit/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+    should.equal(utils.isValidAddress(address.address1), true);
+    should.equal(utils.isValidAddress(address.address2), true);
+    should.equal(utils.isValidAddress(address.address3), false);
+    should.equal(utils.isValidAddress(address.address4), true);
+    should.equal(utils.isValidAddress('dfjk35y'), false);
+    should.equal(utils.isValidAddress(undefined as unknown as string), false);
+    should.equal(utils.isValidAddress(''), false);
+  });
+
+  it('should validate block hash correctly', () => {
```
Vijay-Jagannathan, line 16: We can probably name this suite as `should return true for correct block hash` and the following one as `should return false for incorrect block hash` to avoid confusion

PR #3261, opened 2023-02-07
feat(sdk-unified-wallet): added get coin wallet

jeanlarrieux:
Ticket: BG-65106


modules/sdk-unified-wallet/package.json
```diff
@@ -14,7 +14,8 @@
   },
   "dependencies": {
     "@bitgo/sdk-core": "^5.2.0",
-    "@bitgo/statics": "^14.1.0"
+    "@bitgo/statics": "^14.1.0",
+    "lodash": "^4.17.21"
```
sachushaji, line 18: No package lock updates for dependency change?

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/ecdsaEVMUnifiedWallets.ts
```diff
@@ -4,12 +4,16 @@ import {
   KeychainsTriplet,
   SupplementGenerateWalletOptions,
   Keychains,
+  Wallet,
 } from '@bitgo/sdk-core';
 import { UnifiedWalletID, UnifiedWallet, GenerateUnifiedWalletOptions } from '../types';
-import { supportedCoins } from './types';
+import { supportedCoins, EvmWalletCoins } from './types';
 import { UnifiedWallets } from '../unifiedWallets';
+import _ from 'lodash';
```
sachushaji, line 12: NIT: You could avoid using lodash as most of the functionalities are supported natively. Take a look at [this](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_isempty) for some suggestions

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/ecdsaEVMUnifiedWallets.ts
```diff
@@ -72,4 +77,28 @@ export class EcdsaEVMUnifiedWallets extends UnifiedWallets {
     });
     return keychainsTriplet;
   }
+
+  async getCoinWalletById(id: string, coinName: string): Promise<Wallet> {
+    if (isEmpty(id)) {
```
karankurbur, line 82: Why do we need this function, can't we just do `if (!id)`

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/ecdsaEVMUnifiedWallets.ts
```diff
@@ -72,4 +77,28 @@ export class EcdsaEVMUnifiedWallets extends UnifiedWallets {
     });
     return keychainsTriplet;
   }
+
+  async getCoinWalletById(id: string, coinName: string): Promise<Wallet> {
```
karankurbur, line 81: Can we modify this method signature to something like:
```
await evmWallets.getUnifiedWalletById(id).coin("eth");
```

karankurbur: Trying to follow how we access wallets else where in the SDK example:

```
  const walletInstance = await basecoin.wallets().get({ id: walletId });
```

Open to suggestions / pushback on this

jeanlarrieux: I see your point, but I don't believe this need to be similar as EVMUnifiedWallet is a group of coins having the same address. We already know what those coins are, so better to pass in the coin we want and get the specific wallet instead of making a 2-step process (getWallet, then getCoin). I believe our current 1 step is better (getEvmWalletBasedOnId but for that specific coin.

sachushaji (dismissed):

karankurbur (approved):

modules/sdk-unified-wallet/src/ecdsa-unified-wallet/ecdsaEVMUnifiedWallets.ts
```diff
@@ -72,4 +77,28 @@ export class EcdsaEVMUnifiedWallets extends UnifiedWallets {
...
+    if (!id) {
+      throw new Error('Id field cannot be empty');
+    }
+    if (!supportedCoins.includes(coinName)) {
+      throw new Error(`unsupported coin ${coinName}`);
+    }
+    if (typeof this.unifiedWallet === 'undefined') {
```
jicli, line 88: why does it need to be `typeof` to find out if it's undefined?

karankurbur (approved):

PR #3263, opened 2023-02-08
fix(sdk-core): mark pub as optional in Keychain interface

LeoKotschenreuther:
This makes sure that the type matches what is returned from BitGo. TSS and BLSDKG keychains don't have the pub property set.

Ticket: BG-00000
BREAKING CHANGE: Keychain.pub is now optional

### DIFF
--- a/modules/sdk-coin-xlm/src/xlm.ts
+++ b/modules/sdk-coin-xlm/src/xlm.ts
@@ -1,3 +1,4 @@
+import assert from 'assert';
 import * as _ from 'lodash';
 import * as utxolib from '@bitgo/utxo-lib';
 import * as querystring from 'querystring';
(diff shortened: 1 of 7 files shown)


modules/sdk-coin-xlm/src/xlm.ts
```diff
@@ -1042,9 +1043,11 @@ export class Xlm extends BaseCoin {
         throw new Error('keychains are required, but could not be fetched');
       }
 
+      assert(keychains.backup.pub);
```
margueriteblair, line 1046: why would only xlm need to be changed for this PR? is this our only onchain multisig coin?

LeoKotschenreuther: Apparently it was the only coin that did something with that pub that no longer works if it can be undefined.

margueriteblair (approved):

sachushaji (approved):

davidkaplanbitgo (approved):

PR #3271, opened 2023-02-08
chore(sdk-core): generate safe primes for range proof ntilde

johnoliverdriscoll:
Ticket: BG-63925

The GG18 paper mentions that ntilde should be the produced from 2 safe primes, where a safe prime is in the form `p = 2q + 1` where `q` is also prime.


sachushaji (commented):
Changes LGTM, But seems like the tests are timing out though. We may need to either increase timeouts or mock out the time consuming functions with some expected static values.

PR #3275, opened 2023-02-09
feat(utxo-lib): signature validation array signers come from globalXpubs

davidkaplanbitgo:
BG-67822

### DIFF
--- a/modules/utxo-lib/test/bitgo/psbt/fromHalfSigned.ts
+++ b/modules/utxo-lib/test/bitgo/psbt/fromHalfSigned.ts
@@ -11,6 +11,7 @@ import { getDefaultWalletKeys, getKeyName } from '../../testutil';
 
 import { readFixture } from '../../fixture.util';
 import { normDefault } from '../../testutil/normalize';
+import * as bs58check from 'bs58check';
 
 function getScriptTypes(): outputScripts.ScriptType[] {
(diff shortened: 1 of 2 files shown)


modules/utxo-lib/test/bitgo/psbt/fromHalfSigned.ts
```diff
@@ -95,6 +96,18 @@ function runTest(
...
       const psbt = UtxoPsbt.fromTransaction(startTx, prevOutputs);
+      psbt.updateGlobal({
+        globalXpub: walletKeys.triple.map((bip32) => {
+          const masterFingerprint = Buffer.alloc(4);
+          masterFingerprint.writeUInt32BE(bip32.parentFingerprint);
+          const extendedPubkey = bip32.neutered().toBase58();
+          return {
+            extendedPubkey: bs58check.decode(extendedPubkey),
+            masterFingerprint,
+            path: 'm',
```
JamesJo06, line 107: quick question here. Is this the only place where `globalXpub` get updated?

davidkaplanbitgo: xpubs are added to the psbt during the build route in WP. I was thinking of making this a function on `UtxoPsbt`, maybe this should be done now with this PR

brandonblack (approved):

PR #3277, opened 2023-02-09
chore(sdk-core): require privateShareProof from third party

jicli:
Ticket: BG-65434

### DIFF
--- a/modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
@@ -369,8 +369,7 @@ export async function buildNShareFromAPIKeyShare(keyShare: ApiKeyShare): Promise
     publicShare: keyShare.publicShare,
     encryptedPrivateShare: keyShare.privateShare,
     n: keyShare.n ?? '', // this is not currently needed for key creation
-    // TODO (BG-65434) : make it as mandatory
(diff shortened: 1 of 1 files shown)


modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
```diff
@@ -369,8 +369,7 @@ export async function buildNShareFromAPIKeyShare(keyShare: ApiKeyShare): Promise
     publicShare: keyShare.publicShare,
     encryptedPrivateShare: keyShare.privateShare,
     n: keyShare.n ?? '', // this is not currently needed for key creation
-    // TODO (BG-65434) : make it as mandatory
-    privateShareProof: keyShare.privateShareProof ?? '',
+    privateShareProof: keyShare.privateShareProof,
```
sachushaji, line 372: This change only makes it return undefined if it's not defined, doesn't make it mandatory as commented earlier. The type for `privateShareProof` is still optional. Is this the change you intended?

sachushaji (approved):

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsa.ts
jicli: yes that's good, so in wp we would be checking `if (privateShareProof)` .. it would throwing in errors anyways but in different errors.

PR #3278, opened 2023-02-10
chore: move pt2r and taproot files from bitcoinjs-lib

lcovar:
TICKET: BG-67837

### DIFF
--- a/modules/utxo-lib/src/transaction_builder.ts
+++ b/modules/utxo-lib/src/transaction_builder.ts
@@ -1,5 +1,5 @@
 import * as types from 'bitcoinjs-lib/src/types';
-import { address as baddress } from './';
+import { address as baddress, p2trPayments } from './';
 import * as bufferutils from 'bitcoinjs-lib/src/bufferutils';
 import * as classify from './classify';
 import { crypto as bcrypto } from './';
@@ -520,7 +520,7 @@ function expandInput<TNumber extends number | bigint = number>(
     }
 
     case SCRIPT_TYPES.P2TR_NS: {
-      const { n, pubkeys, signatures } = payments.p2tr_ns(
+      const { n, pubkeys, signatures } = p2trPayments.p2tr_ns(
         {
           // Witness signatures are reverse of pubkeys, because it's a stack
           signatures: witnessStack.length ? witnessStack.reverse() : undefined,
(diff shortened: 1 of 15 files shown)


sachushaji (dismissed):
Package.json changes LGTM

davidkaplanbitgo (dismissed):
The changes look good to me - we're doing this because we want to be able to make changes in taproot for ordinals?

JamesJo06:
are the source codes in payments directory directly from bitcoinjs-lib?

lcovar:
> The changes look good to me - we're doing this because we want to be able to make changes in taproot for ordinals?

Brandon wanted to move p2tr.ts and its associated code over to utxo-lib because upstream and our implementations of these have diverged so there's no reason to have them up there any more. Will also make adding musig key agg easier since I only have to work with utxo-lib

lcovar:
> are the source codes in payments directory directly from bitcoinjs-lib?

Yes, with some minor changes some it compiles (like no non null assertion changes). I updated bitcoinjs-lib to export the `lazy` file so I'll update that in a follow up and import lazy from there instead.

sachushaji (dismissed):
Package.json changes LGTM

modules/utxo-lib/src/payments/lazy.ts
```diff
@@ -0,0 +1,29 @@
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function prop(object: any, name: string, f: () => any): void {
```
brandonblack, line 2: Didn't we expor this from bitcoinjs-lib instead?

brandonblack (changes requested):

modules/utxo-lib/src/transaction_builder.ts
```diff
@@ -1,5 +1,5 @@
 import * as types from 'bitcoinjs-lib/src/types';
-import { address as baddress } from './';
+import { address as baddress, p2tr, p2tr_ns } from './';
```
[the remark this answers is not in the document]
lcovar: https://github.com/BitGo/BitGoJS/pull/3278/files#diff-fa1283e60c39bb1ebf8337d3389acc9246b1e313c3d7bb50c6c8967ee4e3f45dR15

Had to export payments module as `p2trPayments` since we're doing `export * from bitcoinjs-lib` which is exporting a `payments` module.

modules/utxo-lib/src/payments/lazy.ts
lcovar: updated

brandonblack (approved):

sachushaji (approved):

PR #3280, opened 2023-02-10
fix(sdk-core): ecdsa signing vss share

olibeyene:
Ticket: BG-67907


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -680,6 +680,9 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       },
       encryptionKeys: [bitgoGpgKey],
     })) as string;
+    const userGpgKey = await generateGPGKeyPair('secp256k1');
```
islamaminBitGo, line 683: the public version of this GPG key needs to be provided to WP as well so that in the future hsm can do a uSig authentication check on the signerHDShare

islamaminBitGo (changes requested):

pengyuc-bitgo (approved):

islamaminBitGo (approved):

byhow (approved):

PR #3283, opened 2023-02-10
feat: support ordinary key musig aggregation

lcovar:


modules/utxo-lib/src/payments/p2tr.ts
```diff
@@ -19,6 +21,19 @@ const BITCOIN_NETWORK = networks.bitcoin;
 const H = Buffer.from('50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0', 'hex');
 const EMPTY_BUFFER = Buffer.alloc(0);
 
+function isMusigPubKey(pubKey: Uint8Array): boolean {
```
brandonblack, line 24: Rename to `isOrdinaryPubkey` or such (nothing specific to MuSig about them).

byhow (dismissed):

modules/utxo-lib/src/payments/p2tr.ts
```diff
@@ -18,6 +20,19 @@ const BITCOIN_NETWORK = networks.bitcoin;
 const H = Buffer.from('50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0', 'hex');
 const EMPTY_BUFFER = Buffer.alloc(0);
 
+function isOrdinaryPubkey(pubKey: Uint8Array): boolean {
+  if (pubKey.length !== 33) return false;
+  try {
+    return !!necc.Point.fromHex(pubKey);
```
OttoAllmendinger, line 26: haven't we decoded from hex here already? shouldn't this be `fromBuffer` instead?

modules/utxo-lib/src/noble_crypto.ts
```diff
@@ -0,0 +1,80 @@
...
+// @ts-ignore base_crypto is exported as a subPath export, ignoring since compiler complains about importing like this
+import * as base_crypto from '@brandonblack/musig/base_crypto';
+
+const crypto = {
+  ...base_crypto,
+  pointMultiplyUnsafe: (p: Uint8Array, a: Uint8Array, compress: boolean): Uint8Array | null => {
+    try {
+      const product = necc.Point.fromHex(p).multiplyAndAddUnsafe(
+        necc.Point.ZERO,
+        BigInt(`0x${Buffer.from(a).toString('hex')}`),
```
OttoAllmendinger, line 14: this is a bit hacky - let's add a ``toBigInt(b: Uint8Array | Buffer): BigInt` instead that uses https://nodejs.org/api/buffer.html#bufreadbiguint64beoffset (or LE)

OttoAllmendinger (changes requested):

modules/utxo-lib/src/payments/p2tr.ts
lcovar: I was basically reusing the check we use for keys of length 32. Also uses `!!necc.Point.fromHex(pubKey)` and looks like the fromHex method takes this type `declare type Hex = Uint8Array | string;` so either a string or Uint8Array (what we're passing)


https://github.com/BitGo/BitGoJS/blob/862850781b2e8b36c71608c5ae71424b9ebe9dee/modules/utxo-lib/src/noble_ecc.ts#L29-L36

modules/utxo-lib/src/noble_crypto.ts
```diff
@@ -0,0 +1,76 @@
+import * as createHash from 'create-hash';
+import * as necc from '@noble/secp256k1';
+import { MuSigFactory, MuSig } from '@brandonblack/musig';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore base_crypto is exported as a subPath export, ignoring since compiler complains about importing like this
+import * as base_crypto from '@brandonblack/musig/base_crypto';
```
OttoAllmendinger, line 6:
```suggestion
import * as baseCrypto from '@brandonblack/musig/base_crypto';
```

modules/utxo-lib/src/noble_crypto.ts
```diff
@@ -0,0 +1,76 @@
...
+
+function toBigInt(b: Uint8Array | Buffer): bigint {
+  return Buffer.from(b).readBigUint64BE();
+}
+
+const crypto = {
+  ...base_crypto,
+  pointMultiplyUnsafe: (p: Uint8Array, a: Uint8Array, compress: boolean): Uint8Array | null => {
```
OttoAllmendinger, line 14:
```suggestion
  pointMultiplyUnsafe(p: Uint8Array, a: Uint8Array, compress: boolean): Uint8Array | null {
```

for the other methods too

modules/utxo-lib/src/noble_crypto.ts
```diff
@@ -0,0 +1,76 @@
...
+    try {
+      const product = necc.Point.fromHex(p).multiplyAndAddUnsafe(necc.Point.ZERO, toBigInt(a), BigInt(1));
+      if (!product) return null;
+      return product.toRawBytes(compress);
+    } catch {
+      return null;
+    }
```
OttoAllmendinger, line 21: why are we doing these broad catch-alls instead of propagating the error?

the interface states 

https://github.com/brandonblack/musig-js/blob/master/index.ts#L248-L257

> null if the result is the point at infinity.

But your implementation may return `null` for any number of reasons

modules/utxo-lib/src/payments/p2tr.ts
```diff
@@ -18,6 +20,19 @@ const BITCOIN_NETWORK = networks.bitcoin;
 const H = Buffer.from('50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0', 'hex');
 const EMPTY_BUFFER = Buffer.alloc(0);
 
+function isOrdinaryPubkey(pubKey: Uint8Array): boolean {
```
OttoAllmendinger, line 23: the Musig2 BIP uses "Plain" instead of "Ordinary" these days

```suggestion
function isPlainPubkeys(pubKey: Uint8Array): boolean {
```

modules/utxo-lib/src/payments/p2tr.ts
```diff
@@ -41,7 +56,7 @@ export function p2tr(a: Payment, opts?: PaymentOpts): Payment {
       pubkey: typef.maybe(ecc.isXOnlyPoint),
       // the pub key(s) used for keypath signing.
       // aggregated with MuSig2* if > 1
-      pubkeys: typef.maybe(typef.arrayOf(ecc.isXOnlyPoint)),
+      pubkeys: typef.maybe(typef.arrayOf(typef.anyOf(ecc.isXOnlyPoint, isOrdinaryPubkey))),
```
OttoAllmendinger, line 59: mixing and matching doesn't seem right
```suggestion
      pubkeys: typef.maybe(typef.anyOf(typef.arrayOf(ecc.isXOnlyPoint), typef.arrayOf(isOrdinaryPubkey))),
```

modules/utxo-lib/test/fixtures/p2tr.json
```diff
@@ -134,6 +134,38 @@
         ]
       }
     },
+    {
+      "description": "p2tr, out (from scripts & aggregate key) -- Musig",
```
OttoAllmendinger, line 138:
```suggestion
      "description": "p2tr, out (from scripts & aggregate key) -- Musig2",
```

with plain pubkeys?

OttoAllmendinger (changes requested):

modules/utxo-lib/src/noble_crypto.ts
lcovar: I am using [this](https://github.com/brandonblack/musig-js/blob/b2daacfa72e2488dab0f40e470a8ebd6ae2937bf/test/utils.ts#L86-L96) impl of the crypto interface.

I'll look into returning null only when it's point is at infinity.

modules/utxo-lib/src/noble_crypto.ts
```diff
@@ -0,0 +1,76 @@
+import * as createHash from 'create-hash';
```
brandonblack, line 1: Is there a good reason not to combine this with `noble_ecc` and have one file that exports all of these related cryptographic primitives?

lcovar: No, no strong opinions here on this. Updating it so that it's combined with noble_ecc

modules/utxo-lib/src/noble_crypto.ts
brandonblack: This is, roughly, to drag tiny- and noble- secp256k1s together. Because tiny is a native / wasm wrapper, it doesn't through, but returns null on failure. It's still the more common implementation, so I dragged noble in that direction instead of the reverse.

brandonblack (dismissed):

byhow (dismissed):
package changes LGTM

OttoAllmendinger (dismissed):
not sure why the tests succeed with `isOrdinaryPubkey`

modules/utxo-lib/test/fixtures/p2tr.json
```diff
@@ -287,7 +319,7 @@
         "network": "testnet"
       },
       "options": {},
-      "exception": "Expected property \"pubkey\" of type \\?isXOnlyPoint"
+      "exception": "Expected property \"pubkey\" of type \\?isXOnlyPoint|[isOrdinaryPubkey]"
```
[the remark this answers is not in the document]
OttoAllmendinger: please check to make sure this test is actually executed

brandonblack (approved):

byhow (approved):

PR #3284, opened 2023-02-11
fix(sdk-core): reduce size of beta0 and nu0 per revised whitepaper

johnoliverdriscoll:
Ticket: BG-63925


modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -653,7 +652,7 @@ export default class Ecdsa {
             s: bigIntToBufferBE(proof.s, 384).toString('hex'),
             s1: bigIntToBufferBE(proof.s1, 96).toString('hex'),
             s2: bigIntToBufferBE(proof.s2, 480).toString('hex'),
-            t1: bigIntToBufferBE(proof.t1, 416).toString('hex'),
```
sachushaji, line 656: Seems this is a breaking change. Can you kindly mention all breaking changes in the commit footer

sachushaji (commented):
Seems PR is failing at dependency audit step, Can you try rebasing and if still facing the same issue, kindly notify the SDK team for a fix.

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
johnoliverdriscoll: This is not a breaking change as this would only affect the HSM and the HSM code is still in review

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/rangeproof.ts
```diff
@@ -253,7 +253,11 @@ export function verifyWithCheck(
   const modulusBytes = Math.floor((modulusBits + 7) / 8);
   const q = curve.order();
   const q3 = q ** BigInt(3);
-  if (proof.s1 == q3) {
+  const q7 = q ** BigInt(7);
+  if (proof.s1 > q3) {
```
rlutzbitgo, line 257: Just wanted to be sure that this should be gt and not gte, but lgtm

johnoliverdriscoll: The whitepaper allows for $s_1 <= q^3$ so gt is intentional here

rlutzbitgo (approved):

byhow (approved):

PR #3285, opened 2023-02-11
fix(sdk-core): eddsa signing vss share

olibeyene:
Ticket: BG-67907

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
@@ -683,6 +683,8 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
     const userGpgKey = await generateGPGKeyPair('secp256k1');
     const privateShareProof = await createShareProof(userGpgKey.privateKey, signingKey.nShares[bitgoIndex].u, 'ecdsa');
     const vssProof = signingKey.nShares[bitgoIndex].v;
+    const userPublicGpgKey = userGpgKey.publicKey;
+    const publicShare = signingKey.nShares[bitgoIndex].y + signingKey.nShares[bitgoIndex].chaincode;
 
     // signing stage one with K share send to bitgo and receives A share
     const bitgoToUserAShare = (await ECDSAMethods.sendShareToBitgo(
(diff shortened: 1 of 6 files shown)


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -681,7 +681,7 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       encryptionKeys: [bitgoGpgKey],
     })) as string;
     const userGpgKey = await generateGPGKeyPair('secp256k1');
-    const privateShareProof = await createShareProof(userGpgKey.privateKey, signingKey.nShares[bitgoIndex].u, 'ecdsa');
+    const privateShareProof = await createShareProof(userGpgKey.privateKey, signingKey.nShares[bitgoIndex].y, 'ecdsa');
```
pengyuc-bitgo, line 684: This is not right, actually. You should still use the `u` value there and add another line
```
    const publicShare = signingKey.nShares[bitgoIndex].y + signingKey.nShares[bitgoIndex].chaincode; // add chaincode to be consistent with other publicShare
```
And add the `publicShare` in as another field with `vssProof` and `privateShareProof`.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -510,17 +510,26 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
...
     const userSignShare = await createUserSignShare(signablePayload, signingKey.pShare);
 
-    const signerShare = signingKey.yShares[3].u + signingKey.yShares[3].chaincode;
+    const bitgoIndex = 3;
+    const signerShare = signingKey.yShares[bitgoIndex].u + signingKey.yShares[bitgoIndex].chaincode;
     const bitgoGpgKey = await getBitgoGpgPubKey(this.bitgo);
     const encryptedSignerShare = await encryptText(signerShare, bitgoGpgKey);
 
+    const userGpgKey = await generateGPGKeyPair('secp256k1');
+    const privateShareProof = await createShareProof(userGpgKey.privateKey, signingKey.yShares[bitgoIndex].y, 'eddsa');
```
pengyuc-bitgo, line 519: Same here should be
```
    const privateShareProof = await createShareProof(userGpgKey.privateKey, signingKey.yShares[bitgoIndex].u, 'eddsa');

```

byhow (approved):

PR #3287, opened 2023-02-13
chore(sdk-coin-celo): bump celo packages

rlutzbitgo:
TICKET: BG-68073


rlutzbitgo:
No versions of affected celo packages don't still cause this to trigger, but upon investigation looks like a false positive that was removed from GHSA anyway

PR #3300, opened 2023-02-15
chore(sdk-core): make range proofs mandatory

johnoliverdriscoll:
Ticket: BG-68409


rlutzbitgo (approved):

byhow (approved):

modules/sdk-coin-eth/src/eth.ts
```diff
@@ -1278,7 +1300,7 @@ export class Eth extends BaseCoin {
     }
 
     const signableHex = tx.getMessageToSign(false).toString('hex');
-
+    assert(userKeyCombined && backupKeyCombined);
```
olibeyene, line 1303: if (!userKeyCombined  || !backupKeyCombined) {
 return someUseful error.
}

modules/sdk-coin-eth/src/eth.ts
```diff
@@ -55,6 +55,7 @@ import { calculateForwarderV1Address, getProxyInitcode, KeyPair as KeyPairLib }
 import { addHexPrefix, stripHexPrefix } from 'ethereumjs-util';
 import BN from 'bn.js';
 import { TypedDataUtils, SignTypedDataVersion, TypedMessage } from '@metamask/eth-sig-util';
+import assert from 'assert';
```
olibeyene, line 58: don't need to use assert in typescript. if statement is usually does the same

olibeyene (changes requested):
Overall lgtm! minor nit

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -321,7 +321,7 @@ export default class Ecdsa {
    * @returns {SignShareRT} Returns the participant private w-share
    * and k-share to be distributed to other participant signer
    */
-  async signShare(xShare: XShare | XShareWithNTilde, yShare: YShare | YShareWithNTilde): Promise<SignShareRT> {
+  async signShare(xShare: XShare | XShareWithNTilde, yShare: YShareWithNTilde): Promise<SignShareRT> {
```
pranavjain97, line 324: any reason to still have `XShare`? I think we can remove it since we're making it mandatory with range proofs.

pranavjain97: For this, you'll also need to make few more changes, see the diff
Screen Shot 2023-02-16 at 1 21 06 PM

pranavjain97 (changes requested):

olibeyene (approved):

bitgoAaron (approved):

PR #3301, opened 2023-02-16
feat(statics): enable getting asset by id

CapnMigraine:
Ticket: BG-68438

### DIFF
--- a/modules/statics/src/errors.ts
+++ b/modules/statics/src/errors.ts
@@ -26,6 +26,13 @@ export class DuplicateCoinDefinitionError extends BitGoStaticsError {
   }
 }
 
+export class DuplicateCoinIdDefinitionError extends BitGoStaticsError {
+  public constructor(id: string) {
+    super(`coin with id '${id}' is already defined`);
+    Object.setPrototypeOf(this, DuplicateCoinIdDefinitionError.prototype);
+  }
+}
+
 export class DisallowedCoinFeatureError extends BitGoStaticsError {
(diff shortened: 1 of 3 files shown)


modules/statics/src/map.ts
```diff
@@ -44,7 +51,7 @@ export class CoinMap {
   }
 
   public has(key: string): boolean {
-    return this._map.has(key) || this._coinByAliases.has(key);
+    return this._map.has(key) || this._coinByIds.has(key) || this._coinByAliases.has(key);
```
sachushaji, line 54: Can you kindly add a test case for this?

modules/statics/src/errors.ts
```diff
@@ -26,6 +26,13 @@ export class DuplicateCoinDefinitionError extends BitGoStaticsError {
   }
 }
 
+export class DuplicateCoinIdDefinitionError extends BitGoStaticsError {
```
sachushaji, line 29: Can you kindly add a test case for this error flow?

modules/statics/src/map.ts
sachushaji, line 54: Can you kindly add a test case for this?

CapnMigraine: done

modules/statics/src/errors.ts
CapnMigraine: done

sachushaji (approved):

PR #3313, opened 2023-02-17
fix(sdk-core): fix parameter misordering for user gpg key

rlutzbitgo:
TICKET: BG-00000

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
@@ -697,8 +697,8 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       encryptedSignerShare,
       vssProof,
       privateShareProof,
-      userPublicGpgKey,
-      publicShare
+      publicShare,
+      userPublicGpgKey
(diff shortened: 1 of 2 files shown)


zahin-mohammad (dismissed):

wesgraham (approved):

olibeyene (dismissed):

rlutzbitgo:
We did fix WP to make this still work on SDK 18.0.1, but we should still fix this bug here. WP will handle both cases.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -697,8 +697,8 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       encryptedSignerShare,
       vssProof,
       privateShareProof,
-      userPublicGpgKey,
-      publicShare
+      publicShare,
+      userPublicGpgKey
```
sachushaji, line 701: Can you add a testcase to test this fix?,

zahin-mohammad: The API will support both cases to be backwards compatible with the previous SDK version that was already released.

zahin-mohammad (approved):

rlutzbitgo:
Feel free to merge this once it gets developer-relations-team approval

sachushaji (approved):

PR #3317, opened 2023-02-21
chore: add example of using sdk with proxy

brekiy:
Adds an example of using the SDK with a backend proxy to communicate with the BitGo platform.

BG-68680


zahin-mohammad (approved):

examples/ts/proxy/create-wallet.ts
```diff
@@ -0,0 +1,33 @@
...
+  // This value is ignored in favor of your real developer token, stored in your backend proxy machine.
+  accessToken: 'unusedValue',
+  // Set as prod/test as needed for whatever BitGo environment you want to use.
+  // This *must* match the BitGo platform API your proxy instance is using.
+  env: 'test',
+  // In your real setup this would be <your.custom.backend.urlroot>, where you host the proxy instance.
+  customRootURI: 'localhost:3000',
+});
+
+async function createHotMultiSigWalletSimple() {
```
jicli, line 15: dont we already have this in our examples

jicli: how is this different from https://github.com/BitGo/BitGoJS/blob/3e6b9a0309a2d12fed17877a04efaf418adba2c9/examples/ts/create-wallet.ts

brekiy: This is just a packaged example to let a client run it more or less out of the box.

examples/ts/proxy/server.js
```diff
@@ -0,0 +1,33 @@
+const express = require('express');
```
sachushaji, line 1: examples in ts folder should be typescript? seems you are using `server.js`

sachushaji (changes requested):

examples/ts/proxy/README.md
```diff
@@ -0,0 +1,8 @@
+This directory contains a very simple proxy setup to allow a front-end to use BitGo SDK methods with a non-BitGo back-end.
```
mmcshinsky-bitgo, line 1: This should point out the difference between why you would choose this option vs. using the already built out express app in the modules folder.

examples/ts/proxy/create-wallet.ts
mmcshinsky-bitgo: You could add an export to the other example and import it here if you want to avoid duplication.

examples/ts/proxy/README.md
```diff
@@ -0,0 +1,9 @@
+This directory contains a very simple example proxy setup that allows developers to use BitGo SDK methods with a non-BitGo back-end.
+This is in contrast to the BitGo Express app which has its own routes that do not intersect with the SDK.
```
mmcshinsky-bitgo, line 2:
```suggestion
This directory contains a very simple example proxy setup that allows developers to use local BitGo SDK methods with a non-BitGo back-end.
This is in contrast to the [BitGo Express module](https://github.com/BitGo/BitGoJS/tree/master/modules/express) which implements the SDK on the server that you host in order to provide specific SDK logic before proxying requests to BitGo APIs for your convenience.

Please take the time to review your use case as you choose between these options.
```

examples/ts/proxy/create-wallet.ts
```diff
@@ -0,0 +1,35 @@
+import { BitGo } from 'bitgo';
```
mmcshinsky-bitgo, line 1: New examples should incorporate modularization methods. See: https://github.com/BitGo/BitGoJS/blob/master/examples/ts/get-wallet.ts

mmcshinsky-bitgo (changes requested):

mmcshinsky-bitgo (approved):
Thanks for making those updates. Appreciate it :)

PR #3321, opened 2023-02-23
feat(utxo-lib): tweaks to UtxoPsbt

OttoAllmendinger:
Fix some papercuts with UtxoPsbt class.

See commits for details.

Issue: BG-68135

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -458,4 +458,8 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
     const hash = this.tx.hashForWitnessV1(inputIndex, prevoutScripts, prevoutValues, sighashType, leafHash);
     return { hash, sighashType };
   }
+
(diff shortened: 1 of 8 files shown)


modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -414,7 +414,6 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint>> extends Psbt {
 
   private getTaprootHashForSig(
     inputIndex: number,
-    forValidate: boolean,
```
davidkaplanbitgo, line 413: do you know why this was in the signature to begin with?

davidkaplanbitgo (approved):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
OttoAllmendinger: @brandonblack ?

brandonblack: Yeah, to match upstream.

PR #3322, opened 2023-02-23
feat(utxo-lib): improve signature of `createTransactionFromBuffer`

OttoAllmendinger:
Make `amountType` a named argument. Improve ergonomics.

Issue: BG-68135

### DIFF
--- a/modules/utxo-lib/src/bitgo/transaction.ts
+++ b/modules/utxo-lib/src/bitgo/transaction.ts
@@ -15,9 +15,20 @@ import { ZcashNetwork, ZcashTransaction } from './zcash/ZcashTransaction';
 export function createTransactionFromBuffer<TNumber extends number | bigint = number>(
   buf: Buffer,
   network: Network,
-  { version }: { version?: number } = {},
-  amountType: 'number' | 'bigint' = 'number'
+  { version, amountType }: { version?: number; amountType?: 'number' | 'bigint' } = {},
+  deprecatedAmountType?: 'number' | 'bigint'
 ): UtxoTransaction<TNumber> {
+  if (amountType) {
+    if (deprecatedAmountType && amountType !== deprecatedAmountType) {
+      throw new Error(`invalid arguments`);
+    }
+  } else {
+    if (deprecatedAmountType) {
+      amountType = deprecatedAmountType;
+    } else {
+      amountType = 'number';
+    }
+  }
   switch (getMainnet(network)) {
     case networks.bitcoin:
     case networks.bitcoincash:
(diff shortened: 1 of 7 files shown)


davidkaplanbitgo (dismissed):

modules/utxo-lib/src/bitgo/transaction.ts
```diff
@@ -15,9 +15,20 @@ import { ZcashNetwork, ZcashTransaction } from './zcash/ZcashTransaction';
...
+    if (deprecatedAmountType && amountType !== deprecatedAmountType) {
+      throw new Error(`invalid arguments`);
+    }
+  } else {
+    if (deprecatedAmountType) {
+      amountType = deprecatedAmountType;
+    } else {
+      amountType = 'number';
+    }
+  }
```
sachushaji, line 31: NIT,  can avoid some nesting if else
```suggestion
  if (amountType && deprecatedAmountType && amountType !== deprecatedAmountType) {
      throw new Error(`invalid arguments`);
  } else if (deprecatedAmountType)  {
      amountType = deprecatedAmountType;
    } else {
      amountType = 'number';
  }
```

modules/utxo-lib/src/bitgo/transaction.ts
```diff
@@ -15,9 +15,20 @@ import { ZcashNetwork, ZcashTransaction } from './zcash/ZcashTransaction';
 export function createTransactionFromBuffer<TNumber extends number | bigint = number>(
   buf: Buffer,
   network: Network,
-  { version }: { version?: number } = {},
-  amountType: 'number' | 'bigint' = 'number'
```
sachushaji, line 19: Seems to be breaking change?, Can you kindly mention in the commit message ([format](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=in%20Semantic%20Versioning\).-,BREAKING,-CHANGE%3A%20a))

OttoAllmendinger: it's not, old callsites will behave as before

modules/utxo-lib/src/bitgo/transaction.ts
OttoAllmendinger: This is not equivalent - if only `amountType` is set it would be overwritten in the `else` branch. 

There are four cases to deal with (two cases for both arguments). Every case is different. The original code makes it easier to reason about.

sachushaji: Ohh, yeah. Got it!!

modules/utxo-lib/src/bitgo/transaction.ts
sachushaji: Got it!!, thanks a lot for making it non-breaking

sachushaji (approved):
LGTM, thanks @OttoAllmendinger

davidkaplanbitgo (approved):

PR #3331, opened 2023-02-27
chore: add new tokens

aarvay:
Ticket: BG-68301

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -3501,6 +3501,14 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.LYN
   ),
   erc20(
+    'da39727e-150b-4995-bab7-88b60ef2dbd9',
+    'magic',
+    'Magic',
+    18,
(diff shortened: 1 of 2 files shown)


modules/statics/src/coins.ts
```diff
@@ -4725,6 +4741,14 @@ export const coins = CoinMap.fromCoins([
     '0xaea8e1b6cb5c05d1dac618551c76bcd578ea3524',
     UnderlyingAsset.SGR
   ),
+  erc20(
+    '518fe28c-87fc-4b9f-b07e-a2b8f63f4686',
+    'sgt',
+    'Suzu Governance Token',
```
sachushaji, line 4747: In contract, the name of the token seems to be `SGT`, Double checking if this is correct?

sachushaji (approved):

modules/statics/src/coins.ts
aarvay: Yes, it's added as sgt only.

PR #3333, opened 2023-02-27
feat(utxo-lib): add `utxolib/src/ord` with some basic classes

OttoAllmendinger:
* SatRange for describing satoshi ranges
* OrdOutput with `join` and `split` methods

Issue: BG-68315


davidkaplanbitgo (approved):
I feel like I understand it better after reading the code ✅

lcovar (approved):

PR #3334, opened 2023-02-27
feat(sdk-core): add signedChildPsbt to halfSigned object

JamesJo06:
- `halfSigned` in `params` of `submitTransaction` accepts `signedChildPsbt`
- `signedChildPsbt` is PSBT that has the output of `txHex` in `halfSigned`
- as input with a signature
- parsing and validations will be done in the backend server

Ticket: BG-68142

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -420,6 +420,7 @@ export interface SubmitTransactionOptions {
     txHex?: string; // Transaction in any format required by each coin, i.e. in Tron it is a stringifyed JSON
     payload?: string;
     txBase64?: string;
+    signedChildPsbt?: string; // PSBT that has the output of txHex as input with a signature
   };
   comment?: string;
   txRequestId?: string;


modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -420,6 +420,7 @@ export interface SubmitTransactionOptions {
     txHex?: string; // Transaction in any format required by each coin, i.e. in Tron it is a stringifyed JSON
     payload?: string;
     txBase64?: string;
+    signedChildPsbt?: string; // PSBT that has the output of txHex as input with a signature
```
sachushaji, line 423: This seems to be a coin specific option? is it necessary to add like this here?. Also can we get a additional review from someone from your team. Thanks

JamesJo06: we are currently making changes in wp's `/send` to accept `signedChildPsbt` in `halfSigned` We did this way because there is no such thing as `coinSpecific` in `/send`'s body parameter. Here is the pr that makes these changes in wp: https://github.com/BitGo/bitgo-microservices/pull/29459

JamesJo06: Added @brandonblack as an additional review 👍

brandonblack (approved):

byhow (approved):

modules/sdk-core/src/bitgo/wallet/iWallet.ts
sachushaji: Okay, Then I would recommend to use some thing like `coinSpecific?: Record<string, unknown>;` so that it becomes more generic. You can checking this property `signedChildPsbt` to be present in `coinSpecific` in the API side for UTXO specific coins to do the custom logic for it. In this way, other coins in the future can also utilise this.

sachushaji: Okay, Then I would recommend to use some thing like `coinSpecific?: Record<string, unknown>;` so that it becomes more generic. You can checking this property `signedChildPsbt` to be present in `coinSpecific` in the API side for UTXO specific coins to do the custom logic for it. In this way, other coins in the future can also utilise this.

PR #3335, opened 2023-02-28
Bg 69333

jeanlarrieux:
In an effort to support various clients that need a way to interact with `Dapp` (on an `evm` chain as an example). `Defi` team decided to create this `wrapper` to allow `evm RPC` calls to be passed directly to the `SDK` and match to an actual `SDK` method in `wallet`. 
The pre-requisite are a BitGo account, with corresponding enterprise number.


byhow (dismissed):

modules/sdk-evm-rpc-wrapper/test/unit/evmRPCWrapper.ts
```diff
@@ -0,0 +1,91 @@
...
+  it('personal_sign should call wallet signMessage', async function () {
+    signTxRequestForMessage.resolves(txRequestForMessageSigning);
+    const messageRaw = 'test message';
+    params = {
+      method: 'personal_sign',
+      params: [messageRaw],
+    };
+    const expected: SignedMessage = { txRequestId: reqId.toString(), txHash, messageRaw, coin };
+    const signStub = sandbox.stub(Wallet.prototype, 'signMessage').resolves(expected);
+    nock(bgUrl)
```
sachushaji, line 54: Its always a good practice to make sure that at the end of the test, the nock was actually utilised, Check [this](https://github.com/nock/nock#isdone). This will make sure that nocks won't leak out to other tests and you could avoid using `nock.cleanAll();`

modules/sdk-evm-rpc-wrapper/test/unit/evmRPCWrapper.ts
```diff
@@ -0,0 +1,91 @@
...
+import * as sinon from 'sinon';
+import {
+  ethWalletData,
+  txRequestForMessageSigning,
+  reqId,
+  typedMessage,
+  txRequestForTypedDataSigning,
+} from '../fixtures/evmRPCWrapperFixtures';
+import nock = require('nock');
+import { SignTypedDataVersion } from '@metamask/eth-sig-util';
```
sachushaji, line 16: I noticed that SignTypedDataVersion is also available to bbe imported from SDK-core, if thats the case you can use it from there and avoid adding an extra dependency in this package

sachushaji (commented):
Requesting an additional review fron @BitGo/defi-team . Also, Would it makes sense to make this package generic, meaning having it as sdk-rpc-wrapper instead of making it evm specific so that you can extend this to add support for non - evm coins as well in the future if the need arises?. 

Also do add a` index.ts` file and have the things you want people to use externally export it out their which will make it easier for people to import it from the package.

Kindly add an example usage of this package in the sdk examples folder may be as follow up also.

sachushaji (approved):
thanks for addressing the feedbacks, Nice Work!!

byhow (approved):

PR #3339, opened 2023-03-01
Bg 69090 evm wallets pagination

jeanlarrieux:

### DIFF
--- a/modules/sdk-unified-wallet/src/types.ts
+++ b/modules/sdk-unified-wallet/src/types.ts
@@ -21,3 +21,11 @@ export type CurveType = 'Ecdsa' | 'Eddsa';
 export interface GenerateUnifiedWalletOptions extends GenerateWalletOptions {
   curve: CurveType;
 }
+
+export type PaginationOrder = 'ASC' | 'DESC';
+
+export interface PaginationOptions {
+  prevPage?: number;
+  limit?: number;
+  order: PaginationOrder;
+}
--- a/modules/sdk-unified-wallet/src/unifiedWallets.ts
+++ b/modules/sdk-unified-wallet/src/unifiedWallets.ts
@@ -95,7 +95,7 @@ export abstract class UnifiedWallets implements IUnifiedWallets {
     return newWallet;
   }
 
-  async getAllUnifiedWallets(): Promise<UnifiedWallet[]> {
+  async getUnifiedWallets(): Promise<UnifiedWallet[]> {
     return await this.bitgo.get(this.bitgo.url(this.urlPath, 2)).result();
   }
 
--- a/modules/sdk-unified-wallet/src/unifiedWallets.ts
+++ b/modules/sdk-unified-wallet/src/unifiedWallets.ts
@@ -1,5 +1,5 @@
 import { IUnifiedWallets } from './iUnifiedWallets';
-import { UnifiedWalletParams, UnifiedWallet, GenerateUnifiedWalletOptions } from './types';
+import { UnifiedWalletParams, UnifiedWallet, GenerateUnifiedWalletOptions, PaginationOptions } from './types';
 import {
   SupplementGenerateWalletOptions,
   BitGoBase,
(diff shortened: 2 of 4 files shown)


modules/sdk-unified-wallet/src/types.ts
```diff
@@ -21,3 +21,11 @@ export type CurveType = 'Ecdsa' | 'Eddsa';
 export interface GenerateUnifiedWalletOptions extends GenerateWalletOptions {
   curve: CurveType;
 }
+
+export type PaginationOrder = 'ASC' | 'DESC';
+
+export interface PaginationOptions {
+  prevPage?: number;
+  limit?: number;
+  order: PaginationOrder;
```
jakeyangbitgo, line 30: what is this order by? wallet balance?

jeanlarrieux: Order by is the type of ordering (Ascending vs Descending). This is one of the parameters the `WP` [API requires](https://github.com/BitGo/bitgo-microservices/blob/904ca39df91bbe2f8048dc9a8b39380734f40c6d/packages/wallet-platform/app/controllers/api/v2/coins/features/EvmWallet.ts#L122)

modules/sdk-unified-wallet/test/unit/ecdsaUnifiedWallets.ts
```diff
@@ -125,7 +125,7 @@ describe('EVM Wallets:', function () {
 
     it('should get all unified wallets', async function () {
       nock(bgUrl).get('/api/v2/wallet/evm').reply(200, [expected]);
-      const result = await evmWallets.getAllUnifiedWallets();
+      const result = await evmWallets.getUnifiedWallets();
```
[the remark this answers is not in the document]
jeanlarrieux: Pagination was[ unit tested](https://github.com/BitGo/bitgo-microservices/blob/904ca39df91bbe2f8048dc9a8b39380734f40c6d/packages/wallet-platform/test/api/v2/EvmWallet/EvmWallet.ts#L363) in `wp`, since this is an API call we could test in an integration test.

modules/sdk-unified-wallet/src/types.ts
jakeyangbitgo: Ascending vs descending of what? it seems like wallet id?

modules/sdk-unified-wallet/src/unifiedWallets.ts
```diff
@@ -95,8 +95,12 @@ export abstract class UnifiedWallets implements IUnifiedWallets {
     return newWallet;
   }
 
-  async getAllUnifiedWallets(): Promise<UnifiedWallet[]> {
-    return await this.bitgo.get(this.bitgo.url(this.urlPath, 2)).result();
+  async getUnifiedWallets(
+    paginationOptions: PaginationOptions = { prevPage: 0, limit: 5, order: 'ASC' }
```
arnold-yip, line 99: default limit should not be 5, too small for each page
default order should not be ascending, should be descending order by creation date

arnold-yip (changes requested):

modules/sdk-unified-wallet/src/unifiedWallets.ts
jeanlarrieux: I also was concern that too big of a default could create some issues in the front end. Any preference what it should be? (from reading the [BitGo guidelines](https://bitgoinc.atlassian.net/wiki/spaces/ENG/pages/193495049/REST+API+conventions), I didn't see suggested default)

modules/sdk-unified-wallet/src/types.ts
jeanlarrieux: According to @karankurbur , by default `mongo` returns by order of creation. So it seems like the ordering is based on creation order.

steven-isbell (dismissed):

modules/sdk-unified-wallet/src/unifiedWallets.ts
```diff
@@ -95,8 +95,12 @@ export abstract class UnifiedWallets implements IUnifiedWallets {
     return newWallet;
   }
 
-  async getAllUnifiedWallets(): Promise<UnifiedWallet[]> {
-    return await this.bitgo.get(this.bitgo.url(this.urlPath, 2)).result();
+  async getUnifiedWallets(
+    paginationOptions: PaginationOptions = { prevPage: 0, limit: 15, order: 'DESC' }
+  ): Promise<UnifiedWallet[]> {
+    const { limit, order, prevPage } = paginationOptions;
```
jakeyangbitgo, line 101: nit: you can actually do `const { limit, order, prevPage: page } = paginationOptions;`

jeanlarrieux: thanks for the tip :) ... implemented

steven-isbell (approved):

arnold-yip (approved):

PR #3341, opened 2023-03-02
feat(statics): add expiring tx CoinFeature

jayfgrollman:
Adds a CoinFeature called expiring_txs to statics for coins like DOT

CLOSES TICKET: BG-67958

We currently have coin specific logic in the trust admin tool, which is a no no. The logic involves DOT, which has expiring transaction times. Instead of using coin specific logic, the best path forward is to add a coinfeature in statics, especially because there may be more coins like this in the future. This PR adds an expiring txs coinfeature, and adds that feature to DOT.


modules/statics/src/base.ts
```diff
@@ -181,6 +181,10 @@ export enum CoinFeature {
    * This coin supports custody in BitGo Switzerland entities
    */
   CUSTODY_BITGO_SWITZERLAND = 'custody-bitgo-switzerland',
+  /*
+   * This coin has transactions that expire after a certain amount of time.
+   */
+  EXPIRING_TXS = 'expiring-txs',
```
sachushaji, line 187:
```suggestion
  EXPIRING_TRANSACTIONS = 'expiring-transactions',
```
Lets avoid shorthands for a coin feature

modules/statics/src/coins.ts
```diff
@@ -101,7 +101,7 @@ const ALGO_FEATURES = [
   CoinFeature.CUSTODY_BITGO_SWITZERLAND,
   CoinFeature.CUSTODY_BITGO_GERMANY,
 ];
-const DOT_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.TSS, CoinFeature.STAKING];
+const DOT_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.TSS, CoinFeature.STAKING, CoinFeature.EXPIRING_TXS];
```
sachushaji, line 104:
```suggestion
const DOT_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.TSS, CoinFeature.STAKING, CoinFeature. EXPIRING_TRANSACTIONS];
```

modules/statics/src/base.ts
jayfgrollman: Done!

modules/statics/src/coins.ts
jayfgrollman: Done!

sachushaji (approved):

PR #3345, opened 2023-03-03
feat(utxo-lib): add methods for creating PSBTs for inscription txs

OttoAllmendinger:
These are wrappers around the layout methods that do some fee calculation
and psbt construction.

Issue: BG-68315

### DIFF
--- a/modules/utxo-ord/package.json
+++ b/modules/utxo-ord/package.json
@@ -26,7 +26,8 @@
     "directory": "modules/utxo-ord"
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


sachushaji (approved):
Hi @OttoAllmendinger , Would recommend to add the BTC team as owners for this package in CODEOWNER file that relevant team members with context can review your changes. Package.json looks good to me.

OttoAllmendinger:
good catch, added #3351

davidkaplanbitgo (approved):

PR #3347, opened 2023-03-03
refactor(utxo-lib): deprecate `network` parameter in some psbt methods

OttoAllmendinger:
We do have a `psbt.network` property so let's use it

Issue: BG-68315


saravanan7mani (approved):
LGTM

PR #3349, opened 2023-03-03
Add EdDSA commitments

johnoliverdriscoll:
Ticket: BG-69663

Both signers must commit to their `RShare.r` values before exchanging them. These commitments will be produced via `signShare` as the `commitment` property on the returned `SignShare` type. Signers then exchange these commitments before proceeding to exchange their `SignShare.rShare`.

BREAKING CHANGES:
* `commitment` is now a required parameter to `sign`.
* `jShares` parameter to `signShare` is now `jShare` as we only support 2-of-3 signing.
* `rShares` parameter to `signShare` is now `rShare` as we only support 2-of-3 signing.

Please see TODO, as we still need some WP work and plumbing to store commitments.

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
@@ -25,7 +25,7 @@ import {
   TSSParams,
   TxRequest,
 } from '../baseTypes';
-import { CreateEddsaBitGoKeychainParams, CreateEddsaKeychainParams, KeyShare, YShare } from './types';
+import { CreateEddsaBitGoKeychainParams, CreateEddsaKeychainParams, KeyShare, YShare, Commitment } from './types';
 import baseTSSUtils from '../baseTSSUtils';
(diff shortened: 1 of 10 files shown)


modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -534,10 +536,12 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
       publicShare
     );
 
+    const bitgoToUserCommitment = {} as Commitment; // TODO: Not sure where to source this from -JD
```
sachushaji, line 539: Probably a txrequest endpoint to retrieve it from BitGo or as a response to the last step?, Need to finalise the API request flow for this if not discussed previously

sachushaji (commented):
Thanks @johnoliverdriscoll for the changes, Seems Ada testcases also needs to be fixed though. Also need as mentioned by you, WP changes and How we get back commitment from bitgo has to be finalised and reflected to this PR before we can actually merge this.

PR #3350, opened 2023-03-07
feat: create inscription address

lcovar:
TICKET: BG-68137

### DIFF
--- a/modules/utxo-ord/src/inscriptions.ts
+++ b/modules/utxo-ord/src/inscriptions.ts
@@ -1,0 +1,64 @@
+/*
+Functions for dealing with inscriptions.
+
+See https://docs.ordinals.com/inscriptions.html
+*/
+
+import * as assert from 'assert';
+import { p2trPayments as payments, ecc as eccLib, script as bscript, Payment } from '@bitgo/utxo-lib';
+
+const OPS = bscript.OPS;
+const MAX_LENGTH_TAP_DATA_PUSH = 520;
+
+/**
+ * The max size of an individual OP_PUSH in a Taproot script is 520 bytes. This
+ * function splits inscriptionData into an array buffer of 520 bytes length.
+ * https://docs.ordinals.com/inscriptions.html
+ * @param inscriptionData
+ * @param chunkSize
+ */
+function splitBuffer(inscriptionData: Buffer, chunkSize: number) {
+  const pushDataBuffers: Buffer[] = [];
+  for (let i = 0; i < inscriptionData.length; i += chunkSize) {
+    pushDataBuffers.push(inscriptionData.slice(i, i + chunkSize));
+  }
+
+  return pushDataBuffers;
+}
+
+/**
+ * @param userPubkey
+ * @param contentType
+ * @param inscriptionData
+ * @returns inscription address
+ */
+export function createOutputScriptForInscription(
+  userPubkey: Buffer,
+  contentType: string,
+  inscriptionData: Buffer
+): Buffer {
+  const dataPushBuffers = splitBuffer(inscriptionData, MAX_LENGTH_TAP_DATA_PUSH);
+
+  const uncompiledScript = [
+    userPubkey,
+    OPS.OP_CHECKSIG,
+    OPS.OP_FALSE,
+    OPS.OP_IF,
+    Buffer.from('ord', 'ascii'),
+    OPS.OP_1,
+    Buffer.from(contentType, 'ascii'),
+    OPS.OP_0,
+    ...dataPushBuffers,
+    OPS.ENDIF,
+  ];
+
+  const compiledScript = bscript.compile(uncompiledScript);
+  const redeem: Payment = {
+    output: compiledScript,
+    depth: 0,
+  };
+  const payment = payments.p2tr({ redeems: [redeem] }, { eccLib });
+
+  assert(payment.output, 'Failed to create inscription output script');
+  return payment.output;
+}
--- a/modules/utxo-ord/src/index.ts
+++ b/modules/utxo-ord/src/index.ts
@@ -7,3 +7,4 @@ export * from './OrdOutput';
 export * from './OutputLayout';
 export * from './SatPoint';
 export * from './psbt';
+export * as inscriptions from './inscriptions';
--- a/modules/utxo-ord/test/inscription.ts
+++ b/modules/utxo-ord/test/inscription.ts
@@ -1,0 +1,37 @@
+import * as assert from 'assert';
+import { inscriptions } from '../src';
+import { address, networks } from '@bitgo/utxo-lib';
+
+describe('inscriptions', () => {
+  describe('Inscription Output Script', () => {
+    const contentType = 'text/plain';
+    const pubKey = 'af455f4989d122e9185f8c351dbaecd13adca3eef8a9d38ef8ffed6867e342e3';
+    const pubKeyBuffer = Buffer.from(pubKey, 'hex');
+
+    function testInscriptionScript(inscriptionData: Buffer, expectedScriptHex: string, expectedAddress: string) {
+      const outputScript = inscriptions.createOutputScriptForInscription(pubKeyBuffer, contentType, inscriptionData);
+      assert.strictEqual(outputScript.toString('hex'), expectedScriptHex);
+      assert.strictEqual(address.fromOutputScript(outputScript, networks.testnet), expectedAddress);
+    }
+
+    it('should generate an inscription address', () => {
+      const inscriptionData = Buffer.from('Never Gonna Give You Up', 'ascii');
+
+      testInscriptionScript(
+        inscriptionData,
+        '5120e0db418d51573f593389816568e86f96b65cf474712d085b1b0461645639f2fb',
+        'tb1purd5rr232ul4jvufs9jk36r0j6m9ear5wyksskcmq3skg43e7tasdjpqs4'
+      );
+    });
+
+    it('should generate an inscription address when data length is > 520', () => {
+      const inscriptionData = Buffer.from('Never Gonna Let You Down'.repeat(100), 'ascii');
+
+      testInscriptionScript(
+        inscriptionData,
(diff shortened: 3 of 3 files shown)


modules/utxo-lib/src/index.ts
```diff
@@ -16,6 +16,8 @@ export * from './noble_ecc';
 
 export * as p2trPayments from './payments';
 
+export * as inscriptions from './inscriptions';
```
OttoAllmendinger, line 19: let's add this to `modules/utxo-ord` instead

modules/utxo-lib/src/inscriptions.ts
```diff
@@ -0,0 +1,66 @@
+import * as assert from 'assert';
```
OttoAllmendinger, line 1:
```suggestion
/*
Functions for dealing with inscriptions.

See https://docs.ordinals.com/inscriptions.html
*/

import * as assert from 'assert';
```

modules/utxo-lib/src/inscriptions.ts
```diff
@@ -0,0 +1,66 @@
...
+const OPS = bscript.OPS;
+const MAX_LENGTH_TAP_DATA_PUSH = 520;
+
+/**
+ * The max size of an individual OP_PUSH in a Taproot script is 520 bytes. This
+ * function splits inscriptionData into an array buffer of 520 bytes length.
+ * https://docs.ordinals.com/inscriptions.html
+ * @param inscriptionData
+ */
+function sliceInscriptionData(inscriptionData: Uint8Array) {
```
OttoAllmendinger, line 15: We typically use `Buffer` (we are even using it here)

There is nothing really inscription-specific here. Let's just call it `splitBuffer`. This makes the variable names a bit simpler as well.

```suggestion
function splitBuffer(buf: Buffer, chunkSize: number) {
```

modules/utxo-lib/src/inscriptions.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (inscriptionData.length <= MAX_LENGTH_TAP_DATA_PUSH) {
+    return [Buffer.from(inscriptionData)];
+  }
+
```
OttoAllmendinger, line 19: the loop takes care of this
```suggestion
```

modules/utxo-lib/src/inscriptions.ts
```diff
@@ -0,0 +1,66 @@
...
+  const pushDataBuffers: Buffer[] = [];
+  for (let i = 0; i < inscriptionData.length; i += MAX_LENGTH_TAP_DATA_PUSH) {
+    const fixedLengthChunk = inscriptionData.slice(i, i + MAX_LENGTH_TAP_DATA_PUSH);
+    pushDataBuffers.push(Buffer.from(fixedLengthChunk));
```
OttoAllmendinger, line 23: the result is a buffer already
```suggestion
    pushDataBuffers.push(inscriptionData.slice(i, i + MAX_LENGTH_TAP_DATA_PUSH));
```

modules/utxo-lib/src/inscriptions.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  return pushDataBuffers;
+}
+
+/**
+ * Generate an inscription address.
```
OttoAllmendinger, line 30: To stay more consistent with our other address generation methods (`src/bitgo/outputScripts.ts`), let's make the result an output script (`Buffer`) here as well.

One of the advantages is that output scripts are not coin-dependent so the caller can convert to the desired address format. This code here assumes bitcoin (mainnet?).

modules/utxo-lib/src/inscriptions.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @param userPubkey
+ * @param inscriptionData
+ * @param contentType
+ * @returns inscription address
+ */
+export function makeInscriptionAddress(
```
OttoAllmendinger, line 36:
```suggestion
/**
 * @param userPubkey
 * @param inscriptionData
 * @param contentType
 * @returns output script for inscription
 */
export function createOutputScriptForInscription(
```

modules/utxo-lib/src/inscriptions.ts
```diff
@@ -0,0 +1,66 @@
...
+  userPubkey: Uint8Array,
+  inscriptionData: Uint8Array,
+  contentType: string
+): string {
```
OttoAllmendinger, line 40:
```suggestion
): Buffer {
```

modules/utxo-lib/src/inscriptions.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const compiledScript = bscript.compile(uncompiledScript);
+  const redeem: Payment = {
+    output: compiledScript,
+    depth: 0,
+  };
+  const payment = p2tr({ redeems: [redeem], pubkey: pubKeyBuffer }, { eccLib });
+
+  assert(payment.address, 'Failed to create inscription address');
+  return payment?.address;
```
OttoAllmendinger, line 65: I'm a bit confused by this - the `assert` assumes that `payment` exists, but the `return` statement 
```suggestion
  assert(payment.output, 'Failed to create inscription output script');
  return payment.output;
```

modules/utxo-lib/src/inscriptions.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Generate an inscription address.
+ * @param userPubkey
+ * @param inscriptionData
+ * @param contentType
+ * @returns inscription address
+ */
+export function makeInscriptionAddress(
+  userPubkey: Uint8Array,
+  inscriptionData: Uint8Array,
+  contentType: string
```
OttoAllmendinger, line 39: the more conventional order is: content type, then data
```suggestion
  userPubkey: Buffer,
  contentType: string
  inscriptionData: Buffer,
```

modules/utxo-lib/test/inscription.ts
```diff
@@ -0,0 +1,25 @@
...
+  describe('Inscription Address', () => {
+    const pubKey = 'af455f4989d122e9185f8c351dbaecd13adca3eef8a9d38ef8ffed6867e342e3';
+    const pubKeyBuffer = Buffer.from(pubKey, 'hex');
+
+    it('should generate an inscription address', () => {
+      const contentType = 'text/plain';
+      const inscriptionData = new Uint8Array(Buffer.from('Never Gonna Give You Up', 'ascii'));
+
+      const address = inscriptions.makeInscriptionAddress(pubKeyBuffer, inscriptionData, contentType);
+      assert(address);
```
OttoAllmendinger, line 14: The `assert` is not really doing anything - the TS signature already guarantees that there will be a return value.

Let's instead make sure it matches some value

```suggestion
      const script = inscriptions.createOutputScriptForInscription(pubKeyBuffer, contentType, inscription data);
      assert.strictEquals(script.toString('hex'), 'deadbeef');
      assert.strictEquals(fromOutputScript(script, networks.testnet), 'whatever');
```

modules/utxo-lib/test/inscription.ts
```diff
@@ -0,0 +1,25 @@
...
+    });
+
+    it('should generate an inscription address when data length is > 520', () => {
+      const contentType = 'text/plain';
+      const inscriptionData = Buffer.from('Never Gonna Let You Down'.repeat(100), 'ascii');
+
+      const address = inscriptions.makeInscriptionAddress(pubKeyBuffer, inscriptionData, contentType);
+      assert(address);
+    });
```
OttoAllmendinger, line 23: let's write a func `testInscriptionScript(inscriptionData, expectedScriptHex, espectedAddress)`

OttoAllmendinger (dismissed):

modules/utxo-lib/src/inscriptions.ts
OttoAllmendinger: Also we can already think about spending from the output. We can extract a method `createRedeemScriptForInscription(...): Buffer` from here.

modules/utxo-lib/src/inscriptions.ts
lcovar: good catch, i'm asserting the wrong field here

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -0,0 +1,65 @@
+/*
+Functions for dealing with inscriptions.
+
+See https://docs.ordinals.com/inscriptions.html
+*/
+
+import * as assert from 'assert';
+import { Payment, script as bscript } from 'bitcoinjs-lib';
```
brandonblack, line 8: You can also import these from `utxo-lib` - it reexports everything from `bitcoinjs-lib`.

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -0,0 +1,65 @@
...
+    ...dataPushBuffers,
+    OPS.ENDIF,
+  ];
+
+  const compiledScript = bscript.compile(uncompiledScript);
+  const redeem: Payment = {
+    output: compiledScript,
+    depth: 0,
+  };
+  const payment = payments.p2tr({ redeems: [redeem], pubkey: userPubkey }, { eccLib });
```
brandonblack, line 61: This an interesting trade-off. As written, the user can spend this address _either_ by revealing the inscription or not. I _think_ we should not specify `pubkey` and instead leave the user only able to spend by revealing the inscription, but ... not positive. Might be a nice escape hatch to get most of their coin back w/o revealing.

brandonblack (dismissed):
Couple of comments that hint at potential changes, but I think this is good for this task.

modules/utxo-ord/src/inscriptions.ts
lcovar: Wasn't aware of the implications of passing pubKey here. I'm going to remove it since and stick the the original plan where the inscription needs to be revealed.

modules/utxo-ord/src/inscriptions.ts
lcovar: ah didnt even have the bitcoinjs-lib module imported. Updated it import from utxo-lib instead since we have this dep already.

brandonblack (approved):

PR #3352, opened 2023-03-07
feat: express route for signing arbitrary payloads

jonball4:
This endpoint will enable our Clients and Partners to leverage
BitGo Express to sign arbitrary http post request payloads, with
their trading account key, in order to integrate with our systems.

Ticket: BG-66430

image

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -461,6 +461,41 @@ export async function handleV2Sign(req: express.Request) {
   }
 }
 
+export async function handleV2OFCSignPayload(req: express.Request): Promise<{ payload: string; signature: string }> {
+  const walletId = req.body.walletId;
+  const payload = req.body.payload;
+  const ofcCoinName = 'ofc';
+
+  if (!payload) {
+    throw new ApiResponseError('Missing required field: payload', 400);
+  }
+
+  if (!walletId) {
+    throw new ApiResponseError('Missing required field: walletId', 400);
+  }
+
+  const bitgo = req.bitgo;
+
+  // This is to set us up for multiple trading accounts per enterprise
+  const wallet = await bitgo.coin(ofcCoinName).wallets().get({ id: walletId });
+
+  if (wallet === undefined) {
+    throw new ApiResponseError(`Could not find OFC wallet ${walletId}`, 404);
+  }
+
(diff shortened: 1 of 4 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -461,6 +462,43 @@ export async function handleV2Sign(req: express.Request) {
...
+  if (!walletId) {
+    throw new Error('Missing required field: walletId');
+  }
+
+  const bitgo = req.bitgo;
+
+  const wallets = (await bitgo.coin(ofcCoinName).wallets().list()).wallets;
+
+  // This is to set us up for multiple trading accounts per enterprise
+  const wallet = wallets.find((w: Wallet) => w.id() === walletId);
```
mmcshinsky-bitgo, line 485:
```suggestion
const wallet = await bitgo.coin(ofcCoinName).wallets().get({ id: walletId });
```

mmcshinsky-bitgo (changes requested):

modules/express/src/clientRoutes.ts
jonball4: took this suggestion -thanks!

modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
```diff
@@ -16,7 +16,7 @@ export interface BuildPayloadAmounts {
 }
 
 export interface SignPayloadParameters {
-  payload: Payload;
+  payload: unknown;
```
mmcshinsky-bitgo, line 19: Can we type this? We get regular complaints on `any` and `unknown` from customers.

jonball4: I can make it an `object` - it can be anything compatible with `JSON.stringify`

jonball4: (that takes an `any`)

jonball4: the point of the function is to take any payload, so typing it to something more constrained would be counter-intuitive

jonball4: I changed it to `string | Record<string, unknown>` - `object` is a banned type.

mmcshinsky-bitgo (approved):

PR #3355, opened 2023-03-08
feat(sdk-rpc-wrapper): add send transaction

jeanlarrieux:
Adding `eth_sendTransaction` method 

Ticket: BG-69639

### DIFF
--- a/modules/sdk-rpc-wrapper/src/evmRPCWrapper.ts
+++ b/modules/sdk-rpc-wrapper/src/evmRPCWrapper.ts
@@ -1,5 +1,13 @@
-import { EVMRPCRequest, EVMRPCResult } from './types';
-import { Wallet, WalletSignMessageOptions, WalletSignTypedDataOptions, SignTypedDataVersion } from '@bitgo/sdk-core';
+import { EVMRPCRequest, EVMRPCResult, EVMRPCTransactionOptions } from './types';
+import {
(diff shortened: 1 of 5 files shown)


modules/sdk-rpc-wrapper/src/evmRPCWrapper.ts
```diff
@@ -51,4 +67,28 @@ export class EvmRPCWrapper {
       result,
     };
   }
+
+  private async sendTransaction(options: EVMRPCTransactionOptions): Promise<any> {
+    const { to, data, gasPrice, gasLimit, value } = options;
+
+    const sendManyOptions: SendManyOptions = {
+      recipients: [
```
Omurshid, line 75: I'm assuming `sendTransaction` is for sending 1 transaction - why is `recipients` a list here?

jeanlarrieux: [SenMany Method](https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/wallet.ts#L1989)

[SendManyOptions](https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/iWallet.ts#L447)

Omurshid (approved):

PR #3356, opened 2023-03-08
feat: prepare and sign inscription reveal transaction 

lcovar:
TICKET: BG-68138

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -217,6 +217,29 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
     return this;
   }
 
+  finalizeTapInputWithSingleLeafScriptAndSignature(inputIndex: number): this {
+    const input = checkForInput(this.data.inputs, inputIndex);
+    if (input.tapLeafScript?.length !== 1) {
+      throw new Error('Only one leaf script supported for finalizing');
+    }
+    if (input.tapScriptSig?.length !== 1) {
+      throw new Error('Could not find signatures in Script Sig.');
+    }
+
+    const { controlBlock, script } = input.tapLeafScript[0];
+    const witness: Buffer[] = [input.tapScriptSig[0].signature, script, controlBlock];
+    const witnessLength = witness.reduce((s, b) => s + b.length + varuint.encodingLength(b.length), 1);
+
+    const bufferWriter = BufferWriter.withCapacity(witnessLength);
+    bufferWriter.writeVector(witness);
+    const finalScriptWitness = bufferWriter.end();
+
+    this.data.updateInput(inputIndex, { finalScriptWitness });
+    this.data.clearFinalizedInput(inputIndex);
+
+    return this;
+  }
+
   /**
    * Mostly copied from bitcoinjs-lib/ts_src/psbt.ts
    *
--- a/modules/utxo-ord/src/inscriptions.ts
+++ b/modules/utxo-ord/src/inscriptions.ts
@@ -5,7 +5,19 @@ See https://docs.ordinals.com/inscriptions.html
 */
 
 import * as assert from 'assert';
-import { p2trPayments as payments, ecc as eccLib, script as bscript, Payment } from '@bitgo/utxo-lib';
+import {
+  p2trPayments as payments,
+  ecc as eccLib,
+  script as bscript,
+  Payment,
+  Network,
+  bitgo,
+  address,
+  taproot,
+  ECPair,
+  Transaction,
+} from '@bitgo/utxo-lib';
+import { TapLeafScript } from 'bip174/src/lib/interfaces';
 
 const OPS = bscript.OPS;
 const MAX_LENGTH_TAP_DATA_PUSH = 520;
@@ -27,20 +39,17 @@ function splitBuffer(inscriptionData: Buffer, chunkSize: number) {
 }
 
 /**
- * @param userPubkey
+ *
+ * @returns inscription payment object
+ * @param pubkey
  * @param contentType
  * @param inscriptionData
- * @returns inscription address
  */
-export function createOutputScriptForInscription(
-  userPubkey: Buffer,
-  contentType: string,
-  inscriptionData: Buffer
-): Buffer {
+function createPaymentForInscription(pubkey: Buffer, contentType: string, inscriptionData: Buffer): Payment {
   const dataPushBuffers = splitBuffer(inscriptionData, MAX_LENGTH_TAP_DATA_PUSH);
 
   const uncompiledScript = [
-    userPubkey,
+    pubkey,
     OPS.OP_CHECKSIG,
     OPS.OP_FALSE,
     OPS.OP_IF,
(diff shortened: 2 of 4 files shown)


modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +66,72 @@ export function createOutputScriptForInscription(
...
+ * @returns InscriptionRevealData
+ * @param createInscriptionParams & network
+ */
+export function createInscriptionRevealData(
+  createInscriptionParams: CreateInscriptionParams & { network: Network }
+): InscriptionRevealData {
+  const payment = createPaymentForInscription(createInscriptionParams);
+
+  const commitRecipient = payments.p2tr(payment, { eccLib });
+  const { output: commitOutput } = commitRecipient;
```
brandonblack, line 89:
```suggestion
  const commitRecipient = payments.p2tr(payment, { eccLib });
  const { output: commitOutput } = commitRecipient;
```
const { output: commitOutput } = payment;

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +66,72 @@ export function createOutputScriptForInscription(
...
+  assert(commitOutput);
+  const commitAddress = address.fromOutputScript(commitOutput, createInscriptionParams.network);
+
+  // create fake dummy txn for size
+  const psbt = bitgo.createPsbtForNetwork({ network: createInscriptionParams.network as Network });
+  const output = address.toOutputScript(commitAddress, createInscriptionParams.network);
```
brandonblack, line 95: Why do this when we have `commitOutput` above with the same content?

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +66,72 @@ export function createOutputScriptForInscription(
...
+  psbt.addInput({
+    hash: Buffer.alloc(32),
+    index: 0,
+    witnessUtxo: { script: commitOutput, value: BigInt(10239812038) },
+  });
+  psbt.addOutput({ script: output, value: BigInt(10239812038) });
```
brandonblack, line 101: `value: 10000n` (standard inscribed output size)

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +66,72 @@ export function createOutputScriptForInscription(
...
+  psbt.signAllInputs({
+    publicKey: Buffer.alloc(64),
+    sign(hash: Buffer, lowR?: boolean): Buffer {
```
brandonblack, line 104: `signSchnorr` I think.

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +66,84 @@ export function createOutputScriptForInscription(
...
+  // });
+
+  // Fails withError: Input #0 has witnessUtxo but non-segwit script: 51204ae1d7b9cdd20c848b3ae7db4f32f4ec841cf05859e53650c64970384caa622a
+  psbt.signInput(0, {
+    publicKey: Buffer.alloc(64),
+    sign(hash: Buffer, lowR?: boolean): Buffer {
+      return Buffer.alloc(64);
+    },
+  });
+  const revealSize = psbt.extractTransaction().virtualSize();
```
OttoAllmendinger, line 120: This is a bit over the top - pretty sure we can add a simple func that adds up the various input components and returns the vsize. This code here can be used for verification in tests.

Let's extend https://github.com/BitGo/BitGoJS/blob/8a0f0841eabd07478b6f40129e15e83954743fc9/modules/unspents/src/inputWeights.ts instead

OttoAllmendinger (changes requested):
instead of calculating the fee, let's calculate the virtual size of a reveal input

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +66,84 @@ export function createOutputScriptForInscription(
     output: compiledScript,
     depth: 0,
   };
-  const payment = payments.p2tr({ redeems: [redeem] }, { eccLib });
+
+  // cal fees here
```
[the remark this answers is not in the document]
lcovar: This should be deleted

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +67,105 @@ export function createOutputScriptForInscription(
...
+  // create fake dummy txn for size
+  const psbt = bitgo.createPsbtForNetwork({ network: createInscriptionParams.network as Network });
+  psbt.addInput({
+    hash: Buffer.alloc(32),
+    index: 0,
+    witnessUtxo: { script: commitOutput, value: BigInt(10239812038) },
+    tapLeafScript: [
+      {
+        controlBlock,
+        script: commitOutput,
```
brandonblack, line 102: script here should be the inscription reveal script, which I think ends up in `payment.redeem.output`.

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +67,105 @@ export function createOutputScriptForInscription(
...
+  //   sign(hash: Buffer): Buffer {
+  //     return Buffer.alloc(64);
+  //   },
+  //   signSchnorr(hash: Buffer): Buffer {
+  //     return Buffer.alloc(64);
+  //   },
+  // });
+
+  // Error: Input #0 has witnessUtxo but non-segwit script: 51204ae1d7b9cdd20c848b3ae7db4f32f4ec841cf05859e53650c64970384caa622a
+  psbt.signInput(0, {
```
brandonblack, line 136: This is the correct signing method for this type of situation.

brandonblack: Er what about `signTaprootInput` actually?

modules/utxo-ord/src/inscriptions.ts
brandonblack: Yeah, need to call `signTaprootInput` and then specify `signSchnorr`

modules/utxo-ord/src/inscriptions.ts
brandonblack: We're gonna need (essentially) this PSBT built to spend the thing any way, so why not use the _real_ size rather than the error prone attempt to add up the pieces?

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +63,98 @@ export function createOutputScriptForInscription(
...
+
+  return payments.p2tr({ redeems: [redeem], redeemIndex: 0 }, { eccLib });
+}
+
+/**
+ * Creates a dummy signed transaction to get the size of the reveal transaction
+ * used to calculate fees.
+ * @param network
+ * @param controlBlock
+ * @param commitOutput
```
OttoAllmendinger, line 75:
```suggestion
 * @param payment
 * @param controlBlock
 * @param commitOutput
 * @param network
 * @return virtual size of a transaction with a single inscription reveal input and a single commitOutput
```

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +63,98 @@ export function createOutputScriptForInscription(
...
+      publicKey: Buffer.alloc(32),
+      signSchnorr(hash: Buffer): Buffer {
+        return Buffer.alloc(64);
+      },
+    },
+    [leafHash]
+  );
+
+  psbt.finalizeInscriptionInput(0);
+  return psbt.extractTransaction(true).virtualSize();
```
OttoAllmendinger, line 113:
```suggestion
  return psbt.extractTransaction(/* disableFeeCheck */ true).virtualSize();
```

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +63,98 @@ export function createOutputScriptForInscription(
...
+    ],
+  });
+  psbt.addOutput({ script: commitOutput, value: BigInt(10000) });
+
+  psbt.signTaprootInput(
+    0,
+    {
+      publicKey: Buffer.alloc(32),
+      signSchnorr(hash: Buffer): Buffer {
+        return Buffer.alloc(64);
```
OttoAllmendinger, line 106:
```suggestion
        // dummy schnorr-sized signature
        return Buffer.alloc(64);
```

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +63,98 @@ export function createOutputScriptForInscription(
...
+  network: Network
+): number {
+  const psbt = bitgo.createPsbtForNetwork({ network });
+  const parsedControlBlock = taproot.parseControlBlock(eccLib, controlBlock);
+  const leafHash = taproot.getTapleafHash(eccLib, parsedControlBlock, payment.redeem?.output as Buffer);
+
+  psbt.addInput({
+    hash: Buffer.alloc(32),
+    index: 0,
+    witnessUtxo: { script: commitOutput, value: BigInt(10239812038) },
```
OttoAllmendinger, line 90:
```suggestion
    witnessUtxo: { script: commitOutput, value: BigInt(100_000) },
```

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +63,98 @@ export function createOutputScriptForInscription(
...
+    tapLeafScript: [
+      {
+        controlBlock,
+        script: payment.redeem?.output as Buffer,
+        leafVersion: taproot.INITIAL_TAPSCRIPT_VERSION,
+      },
+    ],
+  });
+  psbt.addOutput({ script: commitOutput, value: BigInt(10000) });
```
OttoAllmendinger, line 99:
```suggestion
  psbt.addOutput({ script: commitOutput, value: BigInt(10_000) });
```

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +63,98 @@ export function createOutputScriptForInscription(
...
+export function createOutputScriptForInscription(pubkey: Buffer, contentType: string, inscriptionData: Buffer): Buffer {
+  const payment = createPaymentForInscription(pubkey, contentType, inscriptionData);
 
   assert(payment.output, 'Failed to create inscription output script');
   return payment.output;
 }
+
+export function getFeesForReveal(revealSize: number, feeRate: number): bigint {
+  return BigInt(revealSize * revealSize);
+}
```
OttoAllmendinger, line 160: wrong implementation, missing tests and missing jsdoc (feeRate needs a unit)

given that we don't seem to use this method let's get rid of it. We already have methods to multiple virtual size with feeRate.

```suggestion
```

modules/utxo-ord/test/inscription.ts
```diff
@@ -34,4 +34,37 @@ describe('inscriptions', () => {
...
+        pubKeyBuffer,
+        contentType,
+        inscriptionData,
+        networks.testnet
+      );
+      const feeRate = 1.4;
+
+      const feeEstimate = inscriptions.getFeesForReveal(revealSize, feeRate);
+      assert.strictEqual(revealSize, 121);
+      assert.strictEqual(feeEstimate, BigInt(14641));
```
OttoAllmendinger, line 67: I love palindromes ;)

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +63,98 @@ export function createOutputScriptForInscription(
...
+    [leafHash]
+  );
+
+  psbt.finalizeInscriptionInput(0);
+  return psbt.extractTransaction(true).virtualSize();
+}
+
+export type InscriptionRevealData = {
+  address: string;
+  revealSize: number;
```
OttoAllmendinger, line 118:
```suggestion
  revealTransactionVSize: number;
```

OttoAllmendinger (dismissed):
Please add a test for the VSize calculation by checking them against a valid, fully-signed reveal tx.

modules/utxo-ord/test/inscription.ts
```diff
@@ -34,4 +34,29 @@ describe('inscriptions', () => {
...
+        contentType,
+        inscriptionData,
+        networks.testnet
+      );
+
+      const randomHash = '96b2376fb0ccfdbcc9472489ca3ec75df1487b08a0ea8d9d82c55da19d8cceea';
+      const fullySignedRevealTransaction = inscriptions.signRevealTransaction(
+        ecPair.privateKey as Buffer,
+        tapLeafScript,
+        randomHash,
```
brandonblack, line 53: Don't we need to specify the `vout` / `index` here in addition to the `hash` for the input from the commit transaction?

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -217,6 +217,26 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
...
   }
 
+  finalizeTapInputWithSingleLeafScript(inputIndex: number): this {
+    const input = checkForInput(this.data.inputs, inputIndex);
+    if (input.tapLeafScript?.length !== 1) {
+      throw new Error('Only one leaf script supported for finalizing');
+    }
+
+    const { controlBlock, script } = input.tapLeafScript[0];
+    const witness: Buffer[] = [script, controlBlock];
```
brandonblack, line 227: signature needs to be in here

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -217,6 +217,26 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
     return this;
   }
 
+  finalizeTapInputWithSingleLeafScript(inputIndex: number): this {
```
brandonblack, line 220: `finalizeTapInputWithSingleLeafScriptAndSignature` ;)

brandonblack (changes requested):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -217,6 +217,29 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
     return this;
   }
 
+  finalizeTapInputWithSingleLeafScriptAndSignature(inputIndex: number): this {
+    const input = checkForInput(this.data.inputs, inputIndex);
+    if (input.tapLeafScript?.length !== 1) {
+      throw new Error('Only one leaf script supported for finalizing');
+    }
+    if (!input.tapScriptSig) {
```
brandonblack, line 225: `input.tapScriptSig?.length !== 1`

brandonblack: technically we "should" make sure that the script sig is for the same script as we found above, but DGAF for this since we know how it's gonna be used.

modules/utxo-ord/src/inscriptions.ts
```diff
@@ -57,8 +66,148 @@ export function createOutputScriptForInscription(
...
+  commitAddress: string,
+  commitTxnHash: string,
+  vout: number,
+  network: Network
+): Transaction<bigint> {
+  const psbt = bitgo.createPsbtForNetwork({ network });
+  const commitOutput = address.toOutputScript(commitAddress, network);
+  psbt.addInput({
+    hash: commitTxnHash,
+    index: 0,
```
brandonblack, line 199:
```suggestion
    index: vout,
```

brandonblack (dismissed):

brandonblack (approved):

sachushaji (approved):
package.json changes LGTM

PR #3365, opened 2023-03-09
feat(utxo-ord): use stepwise solution algorithm for inscription tx

OttoAllmendinger:
Replace incomplete constraint resolution algorithm with a more
hand-crafted solution.

This now correctly builds transactions for inscriptions that are in
the middle or close to the end of an input.

Also remove the `maxChangeOutput` parameter since practically we don't
care much about the maximum size of change outputs for inscription
passing transactions.

Issue: BG-68315

### DIFF
--- a/modules/utxo-ord/src/OutputLayout.ts
+++ b/modules/utxo-ord/src/OutputLayout.ts
@@ -51,7 +51,6 @@ Broadly speaking, there are four scenarios:
 
 (3) Large inscription input with inscription close to end of input.
     Change output padding start followed by inscription output.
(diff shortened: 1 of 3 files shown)


brandonblack (approved):
Why nulls instead of undefineds?

I'm hoping we don't need it, but I'm glad that the _interface_ to this supports inputs with multiple ordinals of interest on them.

OttoAllmendinger:
> Why nulls instead of undefineds?

`return null` looks a little bit more explicit than just `return` - but it's not a firmly held opinion

PR #3369, opened 2023-03-10
feat(utxo-ord): add support for additional inputs

OttoAllmendinger:
A single input might be too small to allow building an inscription
passing transaction

Issue: BG-68315

### DIFF
--- a/modules/utxo-ord/src/psbt.ts
+++ b/modules/utxo-ord/src/psbt.ts
@@ -5,6 +5,9 @@ import { OrdOutput } from './OrdOutput';
 import { parseSatPoint, SatPoint } from './SatPoint';
 import { SatRange } from './SatRange';
 import { getOrdOutputsForLayout, OutputLayout, toArray, findOutputLayout } from './OutputLayout';
+import { powerset } from './combinations';
+
+type WalletUnspent = bitgo.WalletUnspent<bigint>;
 
 export type WalletOutputPath = {
   chain: bitgo.ChainCode;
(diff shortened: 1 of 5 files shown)


davidkaplanbitgo (dismissed):

modules/utxo-ord/src/psbt.ts
```diff
@@ -115,8 +120,11 @@ export function findOutputLayoutForWalletUnspents(
     maxInscriptionOutput = BigInt(20_000),
   } = constraints;
 
-  const [input] = inputs;
-  const inscriptionOutput = new OrdOutput(input.value, [toSatRange(satPoint)]);
+  // Join all the inputs into a single inscriptionOutput.
+  // For the purposes of finding a layout there is no difference.
+  const inscriptionOutput = OrdOutput.joinAll(
+    inputs.map((i) => new OrdOutput(i.value, i === inputs[0] ? [toSatRange(satPoint)] : []))
```
brandonblack, line 126: Hmm, this assumes that the order of inputs is set by the caller. kinda seems like we'd want to do something more like having the inscribed input in one param and then the other available inputs in a separate param so they can be added where needed to make reasonable sized fee and change outs.

brandonblack (changes requested):

modules/utxo-ord/src/psbt.ts
OttoAllmendinger: fixed

OttoAllmendinger: * new argument `supplementalUnspents: WalletUnspent[]`
* new argument `minimizeInputs`

davidkaplanbitgo (approved):

modules/utxo-ord/src/psbt.ts
```diff
@@ -135,19 +159,84 @@ export function findOutputLayoutForWalletUnspents(
...
+  }: {
+    supplementaryUnspents?: WalletUnspent[];
+    minimizeInputs?: boolean;
+  } = {}
 ): bitgo.UtxoPsbt {
-  const layout = findOutputLayoutForWalletUnspents(unspents, satPoint, outputs, constraints);
-  if (!layout) {
+  // support for legacy call style
+  if (Array.isArray(unspent)) {
+    if (unspent.length !== 1) {
```
mmcshinsky-bitgo, line 223: Can we ignore the latter unspents and just take the first one without throwing an error?

brandonblack (approved):

PR #3370, opened 2023-03-10
Bg 70036 sol unified wallet

jeanlarrieux:
Implementing `unified Wallet` for `EDDSA` coins. However, `WP` is not currently able to create `EDDSA Unified Wallet`, so created some JIRA tasks and added `TODOs` to the relevant part of the code.


sachushaji (approved):
Dev rel owned files looks good to me

Omurshid (approved):

PR #3374, opened 2023-03-14
feat: add inscription class 

lcovar:
## Description

Add inscription class to abstract utxo-coin and the ability to get that inscription instance from basecoin.

### Example Usage
```
  const walletInstance = await bitgo.coin(coin).wallets().get({ id: walletId });
  const inscriptionBuilder = bitgo.coin(coin).getInscriptionBuilder(walletInstance);

  const { address, tapLeafScript, revealTransactionVSize } = await inscriptionBuilder.prepareReveal(Buffer.from('hey', 'ascii'), 'text/plain');
  const feeRateSatPerKb = 25;
  const sendAmount = (revealTransactionVSize / 1000) * feeRateSatPerKb + 10000

  const builtTxn = await walletInstance.prebuildTransaction({
    recipients: [{
      amount: sendAmount,
      address: address,
    }],
  });

  const unsignedCommitTx = builtTxn.txHex;
  const unspents = builtTxn.txInfo.unspents;

  const signedTx = await inscriptionBuilder.signAndSendReveal(walletPassphrase, tapLeafScript,
    address, Buffer.from(unsignedCommitTx, 'hex'), unspents, '2NADyKQFC42HsRxBS8nd7WTWQraiAJcroYj');
```

### DIFF
--- a/modules/sdk-coin-btc/src/btc.ts
+++ b/modules/sdk-coin-btc/src/btc.ts
@@ -3,8 +3,10 @@ import {
   BitGoBase,
   BaseCoin,
   VerifyRecoveryTransactionOptions as BaseVerifyRecoveryTransactionOptions,
+  Wallet,
 } from '@bitgo/sdk-core';
 import * as utxolib from '@bitgo/utxo-lib';
+import { InscriptionBuilder } from './inscriptionBuilder';
 
 export interface VerifyRecoveryTransactionOptions extends BaseVerifyRecoveryTransactionOptions {
   transactionHex: string;
@@ -37,5 +39,9 @@ export class Btc extends AbstractUtxoCoin {
 
   supportsLightning(): boolean {
     return true;
+  }
+
+  getInscriptionBuilder(wallet: Wallet): InscriptionBuilder {
+    return new InscriptionBuilder(wallet, this);
   }
 }
--- a/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
+++ b/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
@@ -22,6 +22,8 @@ import { encodePsbtMusig2ParticipantsKeyValData } from '../Musig2';
 export interface WalletUnspent<TNumber extends number | bigint = number> extends Unspent<TNumber> {
   chain: ChainCode;
   index: number;
+  witnessScript?: string;
+  valueString?: string;
 }
 
 export interface NonWitnessWalletUnspent<TNumber extends number | bigint = number>
--- a/modules/utxo-ord/test/inscription.ts
+++ b/modules/utxo-ord/test/inscription.ts
@@ -35,7 +35,9 @@ describe('inscriptions', () => {
     });
   });
 
-  describe('Inscription Reveal Data', () => {
+  // TODO(BG-70861): update method with valid data. failing since we need a unsigned commit txn
+  // with valid outs matching commit output
+  xdescribe('Inscription Reveal Data', () => {
     it('should sign reveal transaction and validate reveal size', () => {
       const ecPair = ECPair.makeRandom();
       const inscriptionData = Buffer.from('And Desert You', 'ascii');
@@ -46,16 +48,18 @@ describe('inscriptions', () => {
         networks.testnet
       );
 
-      const randomHash = '96b2376fb0ccfdbcc9472489ca3ec75df1487b08a0ea8d9d82c55da19d8cceea';
       const fullySignedRevealTransaction = inscriptions.signRevealTransaction(
         ecPair.privateKey as Buffer,
         tapLeafScript,
         address,
-        randomHash,
-        2,
+        '2N9R3mMCv6UfVbWEUW3eXJgxDeg4SCUVsu9',
+        Buffer.from(
+          '01000000014b3ea9504d1909063ef15f8b3dbe3dc4c5a129a6d0e31d7ab79a2e9e1bbe38ba0100000000ffffffff02a18601000000000022512087daa0f42694fd0536d3413cb0eff2fa63068c37a312695d92b76a0da2dd8ef55df2100000000000220020857ba44c62320fc9ed0e8a89c4dcdcb211934bcecb8f9778088a482978d77c0000000000',
+          'hex'
+        ),
         networks.testnet
       );
-      const actualVirtualSize = fullySignedRevealTransaction.virtualSize();
+      const actualVirtualSize = fullySignedRevealTransaction.extractTransaction(true).virtualSize();
 
       assert.strictEqual(revealTransactionVSize, actualVirtualSize);
     });
--- a/modules/sdk-coin-btc/package.json
+++ b/modules/sdk-coin-btc/package.json
@@ -42,7 +42,8 @@
   "dependencies": {
     "@bitgo/abstract-utxo": "^2.0.1",
     "@bitgo/sdk-core": "^7.0.1",
-    "@bitgo/utxo-lib": "^7.7.0"
+    "@bitgo/utxo-lib": "^7.7.0",
+    "@bitgo/utxo-ord": "^1.0.0"
   },
   "devDependencies": {
     "@bitgo/sdk-api": "^1.9.10",
--- a/modules/sdk-coin-btc/src/inscriptionBuilder.ts
+++ b/modules/sdk-coin-btc/src/inscriptionBuilder.ts
@@ -1,0 +1,77 @@
+import { AbstractUtxoCoin } from '@bitgo/abstract-utxo';
+import {
+  HalfSignedUtxoTransaction,
+  IInscriptionBuilder,
+  IWallet,
+  KeyIndices,
+  PreparedInscriptionRevealData,
+  SubmitTransactionResponse,
+  xprvToRawPrv,
+} from '@bitgo/sdk-core';
+import * as utxolib from '@bitgo/utxo-lib';
+import { inscriptions } from '@bitgo/utxo-ord';
+import assert from 'assert';
+
+export class InscriptionBuilder implements IInscriptionBuilder {
+  private readonly wallet: IWallet;
+  private readonly coin: AbstractUtxoCoin;
+
+  constructor(wallet: IWallet, coin: AbstractUtxoCoin) {
+    this.wallet = wallet;
+    this.coin = coin;
+  }
+
+  async prepareReveal(inscriptionData: Buffer, contentType: string): Promise<PreparedInscriptionRevealData> {
+    const user = await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[KeyIndices.USER] });
+    assert(user.pub);
+    const pubkey = Buffer.from(user.pub, 'hex');
+
+    return inscriptions.createInscriptionRevealData(pubkey, contentType, inscriptionData, this.coin.network);
+  }
+
+  /**
+   *
+   * @param walletPassphrase
+   * @param tapLeafScript
+   * @param commitAddress
+   * @param unsignedCommitTx
+   * @param commitTransactionUnspents
+   * @param recipientAddress
+   */
+  async signAndSendReveal(
+    walletPassphrase: string,
+    tapLeafScript: utxolib.bitgo.TapLeafScript,
+    commitAddress: string,
+    unsignedCommitTx: Buffer,
+    commitTransactionUnspents: utxolib.bitgo.WalletUnspent[],
+    recipientAddress: string
+  ): Promise<SubmitTransactionResponse> {
+    const userKeychain = await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[KeyIndices.USER] });
+    const xprv = await this.wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const prv = xprvToRawPrv(xprv);
+
+    const halfSignedCommitTransaction = (await this.wallet.signTransaction({
+      prv: xprv,
+      txPrebuild: {
+        txHex: unsignedCommitTx.toString('hex'),
+        txInfo: { unspents: commitTransactionUnspents },
+      },
+    })) as HalfSignedUtxoTransaction;
+
+    const fullySignedRevealTransaction = await inscriptions.signRevealTransaction(
+      Buffer.from(prv, 'hex'),
+      tapLeafScript,
+      commitAddress,
+      recipientAddress,
+      unsignedCommitTx,
+      this.coin.network
+    );
+
+    return this.wallet.submitTransaction({
+      halfSigned: {
+        txHex: halfSignedCommitTransaction.txHex,
+        signedChildPsbt: fullySignedRevealTransaction.toHex(),
+      },
+    });
+  }
+}
--- a/modules/sdk-coin-btc/tsconfig.json
+++ b/modules/sdk-coin-btc/tsconfig.json
@@ -24,6 +24,9 @@
     },
     {
       "path": "../utxo-lib"
+    },
+    {
+      "path": "../utxo-ord"
     }
   ]
 }
--- a/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
+++ b/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
@@ -39,6 +39,7 @@ import {
   VerifyAddressOptions,
   VerifyTransactionOptions,
 } from './iBaseCoin';
+import { IInscriptionBuilder } from '../inscriptionBuilder';
 
 export abstract class BaseCoin implements IBaseCoin {
   protected readonly bitgo: BitGoBase;
@@ -456,5 +457,9 @@ export abstract class BaseCoin implements IBaseCoin {
 
   async recoverToken(params: RecoverWalletTokenOptions): Promise<RecoverTokenTransaction> {
     throw new NotImplementedError('recoverToken is not supported for this coin');
+  }
+
+  getInscriptionBuilder(wallet: Wallet): IInscriptionBuilder {
+    throw new NotImplementedError('Inscription Builder is not supported for this coin');
   }
 }
--- a/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
+++ b/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
@@ -12,6 +12,7 @@ import { IWallet, IWallets, Wallet, WalletData } from '../wallet';
 import { IWebhooks } from '../webhook/iWebhooks';
 import { BaseTokenConfig } from '@bitgo/statics';
 import { TransactionType } from '../../account-lib';
+import { IInscriptionBuilder } from '../inscriptionBuilder';
 
 export interface Output extends ITransactionRecipient {
   address: string;
@@ -465,4 +466,5 @@ export interface IBaseCoin {
   getMPCAlgorithm(): MPCAlgorithm;
   // TODO - this only belongs in eth coins
   recoverToken(params: RecoverWalletTokenOptions): Promise<RecoverTokenTransaction>;
+  getInscriptionBuilder(wallet: Wallet): IInscriptionBuilder;
 }
--- a/modules/sdk-core/src/bitgo/index.ts
+++ b/modules/sdk-core/src/bitgo/index.ts
@@ -11,6 +11,7 @@ export * from './ecdh';
 export * from './enterprise';
 export * from './environments';
 export * from './errors';
+export * from './inscriptionBuilder';
 export * from './internal';
 export * from './keychain';
 export * as bitcoin from './legacyBitcoin';
--- a/modules/sdk-core/src/bitgo/inscriptionBuilder/iInscriptionBuilder.ts
+++ b/modules/sdk-core/src/bitgo/inscriptionBuilder/iInscriptionBuilder.ts
@@ -1,0 +1,58 @@
+import * as utxolib from '@bitgo/utxo-lib';
+
+export interface SubmitTransactionResponse {
+  transfer: Transfer;
+  txid: string;
+  tx: string;
+  status: string;
+}
+
+interface Entry {
+  address: string;
+  wallet: string;
+  value: number;
+  valueString: string;
+  isChange?: boolean;
+  isPayGo?: boolean;
+}
+
+interface Transfer {
+  entries: Entry[];
+  id: string;
+  coin: string;
+  wallet: string;
+  enterprise: string;
+  txid: string;
+  height: number;
+  heightId: string;
+  type: string;
+  value: number;
+  valueString: string;
+  baseValue: number;
+  baseValueString: string;
+  feeString: string;
+  payGoFee: number;
+  payGoFeeString: string;
+  state: string;
+  vSize: number;
+}
+
+export type PreparedInscriptionRevealData = {
+  address: string;
+  revealTransactionVSize: number;
+  tapLeafScript: utxolib.bitgo.TapLeafScript;
+};
+
+export interface IInscriptionBuilder {
+  prepareReveal(inscriptionData: Buffer, contentType: string): Promise<PreparedInscriptionRevealData>;
+
+  // same response as wallet.submitTransaction`
+  signAndSendReveal(
+    walletPassphrase: string,
+    tapLeafScript: utxolib.bitgo.TapLeafScript,
+    commitAddress: string,
+    unsignedCommitTx: Buffer,
+    commitTransactionUnspents: utxolib.bitgo.WalletUnspent[],
+    recipientAddress: string
+  ): Promise<SubmitTransactionResponse>;
+}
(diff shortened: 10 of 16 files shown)


modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,84 @@
...
+    commitTransactionUnspents: any[],
+    commitAddress: string,
+    recipientAddress: string
+  ): Promise<any> {
+    const userKeychain = await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[KeyIndices.USER] });
+    const xprv = await this.wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const prv = xprvToRawPrv(xprv);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore it exists;
+    const network = this.bitgo.coin(this.wallet.coin()).network;
```
lcovar, line 56: Unsure if there is a more elegant way of getting network here.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2596,6 +2597,16 @@ export class Wallet implements IWallet {
     return new Lightning(this.bitgo, this);
   }
 
+  /**
+   * Create inscriptions for btc/tbtc from this wallet
+   */
+  public inscription(): Inscription {
```
rushilbg, line 2603: I know we're trying to keep up with the `lightning` convention here, but using only `inscription` doesn't clearly let the reader know what the function is doing, which is to create an Inscription instance. Can we rename it to `getInscriptionInstance` or `createInscriptionInstance`?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2596,6 +2597,16 @@ export class Wallet implements IWallet {
...
+    if (!this.baseCoin.supportsInscriptions()) {
+      throw new Error(`Lightning not supported for ${this.coin()}`);
```
rushilbg, line 2605:
```suggestion
      throw new Error(`Inscription not supported for ${this.coin()}`);
```

modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,84 @@
...
+   * @param commitAddress
+   * @param recipientAddress
+   */
+  async signAndSend(
+    walletPassphrase: string,
+    tapLeafScript: TapLeafScript,
+    unsignedCommitTx: string,
+    commitTransactionUnspents: any[],
+    commitAddress: string,
+    recipientAddress: string
```
rushilbg, line 49: Nit: Although I'd have preferred to condense the arguments but if we're not doing that lets at least reorder the arguments to keep properties of same object next to each other.
Eg: `tapLeafScript` & `commitAddress` are part of same revealData object, similarly `unsignedCommitTx` & `commitTransactionUnspents`.

modules/sdk-core/src/bitgo/inscription/iInscription.ts
```diff
@@ -0,0 +1,15 @@
+import { PreparedInscriptionRevealData, TapLeafScript } from '@bitgo/utxo-ord';
+
+export interface IInscription {
+  prepareReveal(inscriptionData: Buffer, contentType: string, feeRate: number): Promise<PreparedInscriptionRevealData>;
```
rushilbg, line 4: Should the UI use `FileReader.readAsArrayBuffer()` to get file data and pass it here as `inscriptionData`?
Ref: https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsArrayBuffer

rushilbg (changes requested):

modules/sdk-core/src/bitgo/inscription/iInscription.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  // same response as wallet.submitTransaction
+  signAndSend(
+    walletPassphrase: string,
+    tapLeafScript: TapLeafScript,
+    unsignedCommitTx: string,
+    commitTransactionUnspents: any[],
```
davidkaplanbitgo, line 11: this should have a stricter type than `any`

modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,84 @@
...
+    inscriptionData: Buffer,
+    contentType: string,
+    feeRate: number
+  ): Promise<PreparedInscriptionRevealData> {
+    const user = await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[KeyIndices.USER] });
+    assert(user.pub);
+    const pubkey = Buffer.from(user.pub);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore it exists
+    const network = this.bitgo.coin(this.wallet.coin()).network;
```
davidkaplanbitgo, line 29:
```suggestion
    const network = getNetwork(this.wallet.coin() as V1Network)
```

davidkaplanbitgo (changes requested):

modules/sdk-core/src/bitgo/inscription/inscription.ts
davidkaplanbitgo: try this? - this should work because we are only using inscriptions with `"bitcoin"` and `"testnet"`

modules/sdk-core/src/bitgo/inscription/iInscription.ts
lcovar: Not sure if that exact method but something along those lines yes, read the file as a buffer.

modules/sdk-coin-btc/src/btc.ts
```diff
@@ -38,4 +38,8 @@ export class Btc extends AbstractUtxoCoin {
   supportsLightning(): boolean {
     return true;
   }
+
+  supportsInscriptions(): boolean {
+    return true;
+  }
```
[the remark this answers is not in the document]
lcovar: Created internal ticket for this. BG-70561

modules/sdk-core/src/bitgo/wallet/wallet.ts
lcovar: Hmm personally prefer just inscription(). Lets vote on this on our channel.

modules/sdk-core/src/bitgo/inscription/inscription.ts
lcovar: ended up writing a utility method to get network based on coin name. If we support doginals getting the network like this ^ wont work 😢

modules/sdk-core/src/bitgo/inscription/iInscription.ts
lcovar: I used a sample response from /build to see what unspent type should be. Not sure if that covers all the cases though

lcovar: since we didnt have an existing type. We have txInfo set as unknown atm.

modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   *
+   * @param walletPassphrase
+   * @param tapLeafScript
+   * @param unsignedCommitTx
+   * @param commitTransactionUnspents
+   * @param commitAddress
+   * @param recipientAddress
+   */
+  async signAndSend(
```
davidkaplanbitgo, line 38: Possibly rename to 
```suggestion
  async signAndSendReveal(
```
Just so that it does not need to be renamed once I start. I was thinking `signAndSendTransfer`, but I am not tied to the name

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2596,6 +2597,16 @@ export class Wallet implements IWallet {
     return new Lightning(this.bitgo, this);
   }
 
+  /**
+   * Create inscriptions for btc/tbtc from this wallet
+   */
+  public createInscriptionInstance(): Inscription {
```
davidkaplanbitgo, line 2603: 👍

davidkaplanbitgo (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
lcovar: updated to createInscriptionInstance

modules/sdk-core/src/bitgo/inscription/iInscription.ts
```diff
@@ -0,0 +1,25 @@
+import { PreparedInscriptionRevealData, TapLeafScript } from '@bitgo/utxo-ord';
+
+export type TxInfoUnspents = {
```
rushilbg, line 3: No existing type that could be reused instead?

modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,75 @@
...
+    return inscriptions.createInscriptionRevealData(pubkey, contentType, inscriptionData, network);
+  }
+
+  /**
+   *
+   * @param walletPassphrase
+   * @param tapLeafScript
+   * @param unsignedCommitTx
+   * @param commitTransactionUnspents
+   * @param commitAddress
```
rushilbg, line 35: Order change needed here too.

rushilbg (dismissed):

rushilbg:
> ### Example Usage
> ```
> 
>   const walletInstance = await bitgo.coin(coin).wallets().get({ id: walletId });
>   const inscription = walletInstance.createInscriptionIntance();
>   
>   const { address, tapLeafScript, revealTransactionVSize } = 
>     await inscription.prepareReveal(Buffer.from('Ongo Gablogian, Art Collector', 'ascii'), 'text/plain');
>   const feeRate = 25;
>   const sendAmount = revealTransactionVSize * feeRate;
> 
>   const builtTxn = await walletInstance.prebuildTransaction({
>     recipients: [{
>       amount: sendAmount,
>       address: address,
>     }],
>   });
> 
>   const unsignedCommitTx = builtTxn.txHex;
>   const unspents = builtTxn.txInfo.unspents;
> 
>   const signedTx = await inscription.signAndSend(walletPassphrase, revealData.tapLeafScript,
>   unsignedCommitTx, unspents, address, '<recipientAddress>');
> ```

The client passes `numBlocks` parameter to `prebuildTransaction` usually. `feeRate` is passed only in case customFeeRate is inputted by the user, otherwise the SDK uses `numBlocks`. Since, `sendAmount` is calculated here using `feeRate`, the relationship b/w numBlocks and feeRate will need to be exposed to client now, which was internal to SDK until now.

Let's discuss this before merging.

modules/sdk-core/src/bitgo/inscription/iInscription.ts
OttoAllmendinger: @lcovar check if `utxolib.bitgo.WalletUnspent` can be used here

modules/sdk-core/src/bitgo/inscription/iInscription.ts
```diff
@@ -0,0 +1,25 @@
...
+  chain: number;
+  index: number;
+  witnessScript?: string;
+  id: string;
+  address: string;
+  value: number;
+  valueString: string;
+};
+
+export interface IInscription {
```
OttoAllmendinger, line 13:
```suggestion
export interface IInscriptionBuilder {
```

modules/sdk-core/src/bitgo/inscription/iInscription.ts
```diff
@@ -0,0 +1,25 @@
...
+  prepareReveal(inscriptionData: Buffer, contentType: string): Promise<PreparedInscriptionRevealData>;
+
+  // same response as wallet.submitTransaction
+  signAndSendReveal(
+    walletPassphrase: string,
+    tapLeafScript: TapLeafScript,
+    commitAddress: string,
+    unsignedCommitTx: string,
```
OttoAllmendinger, line 21: `Buffer`, maybe even `UtxoPsbt`

modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,75 @@
...
+import { KeyIndices } from '../keychain';
+import { xprvToRawPrv } from '../../account-lib';
+import { HalfSignedUtxoTransaction } from '../baseCoin';
+import { inferNetworkFromCoinName } from '../utils';
+
+export class Inscription implements IInscription {
+  private readonly bitgo: BitGoBase;
+  private readonly wallet: IWallet;
+
+  constructor(bitgo: BitGoBase, wallet: IWallet) {
```
OttoAllmendinger, line 15: let's pass in `coin: AbstractUtxoCoin` here so we can use `this.coin.network` normally

OttoAllmendinger (dismissed):

modules/sdk-core/src/bitgo/inscription/inscription.ts
davidkaplanbitgo: `this.coin = getUtxoCoin(wallet.coin());`

This is typed as an `AbstractUtxoCoin`

lcovar: AbstractUtxoCoin type is not imported in sdk-core, I can't do this.

modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,75 @@
...
+   * @param recipientAddress
+   */
+  async signAndSendReveal(
+    walletPassphrase: string,
+    tapLeafScript: TapLeafScript,
+    commitAddress: string,
+    unsignedCommitTx: string,
+    commitTransactionUnspents: TxInfoUnspents[],
+    recipientAddress: string
+  ): Promise<any> {
```
mmcshinsky-bitgo, line 45: Can we type this `any` as well?

modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,75 @@
...
+    const fullySignedRevealTransaction = await inscriptions.signRevealTransaction(
+      Buffer.from(prv, 'hex'),
+      tapLeafScript,
+      commitAddress,
+      recipientAddress,
+      unsignedCommitTx,
+      network
+    );
+
+    return await this.wallet.submitTransaction({
```
mmcshinsky-bitgo, line 68: Can remove await since it's already a returned promise

modules/sdk-core/src/bitgo/utils/abstractUtxoCoinUtil.ts
```diff
@@ -6,3 +6,14 @@ export function inferAddressType(addressDetails: { chain: number }): ScriptType2
     ? utxolib.bitgo.scriptTypeForChain(addressDetails.chain)
     : null;
 }
+
+export function inferNetworkFromCoinName(coinName: string): utxolib.Network {
```
mmcshinsky-bitgo, line 10: Is this function intentionally scoped to bitcoin only?

modules/sdk-core/src/bitgo/inscription/iInscription.ts
```diff
@@ -0,0 +1,25 @@
...
+    commitTransactionUnspents: TxInfoUnspents[],
+    recipientAddress: string
+  ): Promise<any>;
```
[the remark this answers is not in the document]
mmcshinsky-bitgo: +1 to typing

modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const halfSignedCommitTransaction = (await this.wallet.signTransaction({
+      prv: xprv,
+      txPrebuild: {
+        txHex: unsignedCommitTx,
+        txInfo: { unspents: commitTransactionUnspents },
+      },
+    })) as HalfSignedUtxoTransaction;
+
+    const network = inferNetworkFromCoinName(this.wallet.coin());
```
mmcshinsky-bitgo, line 58:
```suggestion
    const network = this.wallet.network()
```
Is this available and does what you are looking to do?

mmcshinsky-bitgo (dismissed):

modules/sdk-core/src/bitgo/inscription/inscription.ts
lcovar: Unfortunately not available. What would help here was if I had AbstractUtxo type so I can do something like (this.wallet.coin() as AbstractUtxo).nework()

modules/sdk-core/src/bitgo/inscription/iInscription.ts
lcovar: Curious, so that I understand this better, why do prefer to pass around a Buffer over a tx hex string?

lcovar:
@OttoAllmendinger WalletUnspent looks like it will work fine, added 2 missing fields. Good looking out

modules/sdk-core/src/bitgo/inscription/inscription.ts
lcovar: @OttoAllmendinger @mmcshinsky-bitgo  added a type for this. lmk if it looks good. Only thing is im not sure which fields should be optional. Used a response from submit transaction to create the type.

modules/sdk-core/src/bitgo/inscription/inscription.ts
OttoAllmendinger: `inscription.ts` should live in `AbstractUtxoCoin`. It is a utxo-specific concept after all.

lcovar: Ideally yes, but these methods need to be called within a wallet for our users. We run into this same issue where we only have basecoin to work with in sdk-core. We still won't be able to do something like wallet.coin().inscriptionMethodHere().

davidkaplanbitgo (dismissed):
nit

modules/sdk-core/src/bitgo/inscription/inscription.ts
OttoAllmendinger: ok can't we have a bunch of methods in `AbstractUtxoCoin` then that take the wallet as a parameter?

modules/sdk-core/src/bitgo/inscription/iInscription.ts
OttoAllmendinger: `string` is maybe the weakest type in JS. We want strongly typed not https://wiki.c2.com/?StringlyTyped

modules/sdk-core/src/bitgo/inscription/iInscription.ts
```diff
@@ -0,0 +1,53 @@
+import { PreparedInscriptionRevealData, TapLeafScript } from '@bitgo/utxo-ord';
```
OttoAllmendinger, line 1: adding coin-specific libraries to `sdk-core` kind of defeats the purpose of modularizing the SDK

modules/sdk-core/src/bitgo/inscription/inscription.ts
```diff
@@ -0,0 +1,76 @@
...
+import { BitGoBase } from '../bitgoBase';
+import { IWallet } from '../wallet';
+import { inscriptions, PreparedInscriptionRevealData, TapLeafScript } from '@bitgo/utxo-ord';
+import { KeyIndices } from '../keychain';
+import { xprvToRawPrv } from '../../account-lib';
+import { HalfSignedUtxoTransaction } from '../baseCoin';
+import { inferNetworkFromCoinName } from '../utils';
+import * as utxolib from '@bitgo/utxo-lib';
+
+export class Inscription implements IInscription {
```
OttoAllmendinger, line 12:
```suggestion
export class InscriptionBuilder implements IInscriptionBuilder {
```

IF we actually need the interface - still unclear why

I don't understand why we just add a `class InscriptionBuilder` that lives in `modules/abstract-utxo/`

modules/sdk-core/src/bitgo/inscription/inscription.ts
lcovar: Added it to abstract utxo coin. baseCoin still needs this since we need to keep in mind UI usage, we do not use coin specific methods.

modules/abstract-utxo/src/inscriptionBuilder.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const halfSignedCommitTransaction = (await this.wallet.signTransaction({
+      prv: xprv,
+      txPrebuild: {
+        txHex: unsignedCommitTx.toString('hex'),
+        txInfo: { unspents: commitTransactionUnspents },
+      },
+    })) as HalfSignedUtxoTransaction;
+
+    const fullySignedRevealTransaction = await inscriptions.signRevealTransaction(
```
rushilbg, line 61: +1 on verbose naming!

rushilbg (dismissed):

davidkaplanbitgo (dismissed):

byhow (dismissed):

JamesJo06 (dismissed):

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -1362,4 +1363,8 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   getRecoveryProvider(apiToken?: string): RecoveryProvider {
     return forCoin(this.getChain(), apiToken);
   }
+
```
brandonblack, line 1366: Should we push this all the way down to btc, since we only support this on btc so far? We will probably do litecoin support at some point, but wouldn't expect us to do other utxo coins (even though doginals are a thing, they are totally different scripts). (can be follow-up PR too)

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -22,6 +22,8 @@ import { encodePsbtMusig2ParticipantsKeyValData } from '../Musig2';
 export interface WalletUnspent<TNumber extends number | bigint = number> extends Unspent<TNumber> {
   chain: ChainCode;
   index: number;
+  witnessScript?: string;
```
brandonblack, line 25: Are these just to let us put these properties on when we find them? How do we find witnessScript if the unspent returned from WP doesn't have it (does that ever happen?)?

modules/utxo-ord/test/inscription.ts
```diff
@@ -35,7 +35,9 @@ describe('inscriptions', () => {
     });
   });
 
-  describe('Inscription Reveal Data', () => {
+  // TODO: update method with valid data. failing since we need a unsigned commit txn
+  // with valid outs matching commit output
+  xdescribe('Inscription Reveal Data', () => {
```
brandonblack, line 40: Will this be a follow-up PR?

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
lcovar: These are the properties that _may_ returned to us when we build a btc transaction, under txInfo. I left it optional since I am not too sure if every response to /build for **every** btc transaction returns unspent info with all these fields.

brandonblack: Right, so my question is: are we sure they are present in all the cases we need them?

modules/abstract-utxo/src/abstractUtxoCoin.ts
lcovar: Ok, i think we should. I can quickly do it in this PR.

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
lcovar: We actually dont _need_ this field for our use case. We need txInfo.unspents to sign the unsigned commit txn hex and wanted to improve the type using the response from /build. However, during [signing](https://github.com/BitGo/BitGoJS/blob/214eafe48e8d12fd5d58efac289bab33bbd46fd3/modules/utxo-lib/src/bitgo/wallet/Unspent.ts#L37-L61) we care about the unspent index/chain/value/address.


```
 "unspents": [
      {
        "chain": 21,
        "index": 80,
        "**witnessScript**": "522102b2390579afa56499ebb1d5f92da6942c15e6acaae501a5d84c6cca433764d1bb21021fea984d5c20c716c14100bf4cd0e4bd0b851b540c22aeb2f010114eb90aeae121038299d563f8ee1aea4de65b1c23c352e3443ddb7f4355c39f989b0aaed7e2af8553ae",
        "id": "29182610ce831c9390b45bf0452a018ad39347c53289bf13cd996ee5ef2d4037:1",
        "address": "tb1q44kr6nhkgdxuu89n82sjpcskmltc7qgarzv0ftn7qgu3efw9n48s89egv0",
        "value": 1353750,
        "**valueString**": "1353750"
      }
    ],
```

brandonblack (approved):

byhow (approved):

PR #3383, opened 2023-03-20
fix(statics):  add tsol valid usdc for devnet token data

alebusse:
added solana valid USDC for devnet  as tsol:usdcv2 token data

BG-0000


ewangbitgo (dismissed):

ewangbitgo (approved):

sachushaji (approved):
If you are adding a new token, you can ideally prefix the PR with feat rather than fix

PR #3386, opened 2023-03-20
feat: add support finalizeAllInputs for reveal transactions

lcovar:
TICKET: BG-70666

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -282,7 +282,13 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
   finalizeAllInputs(): this {
     checkForInput(this.data.inputs, 0); // making sure we have at least one
     this.data.inputs.map((input, idx) => {
-      return input.tapScriptSig?.length ? this.finalizeTaprootInput(idx) : this.finalizeInput(idx);
+      if (input.tapScriptSig) {
+        return input.tapScriptSig.length === 1
+          ? this.finalizeTapInputWithSingleLeafScriptAndSignature(idx)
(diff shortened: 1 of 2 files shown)


modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -282,7 +282,12 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
   finalizeAllInputs(): this {
     checkForInput(this.data.inputs, 0); // making sure we have at least one
     this.data.inputs.map((input, idx) => {
-      return input.tapScriptSig?.length ? this.finalizeTaprootInput(idx) : this.finalizeInput(idx);
+      try {
+        return input.tapScriptSig?.length ? this.finalizeTaprootInput(idx) : this.finalizeInput(idx);
+      } catch (e) {
+        // Likely a single leaf script and single sig reveal txn, try finalizing with the respective method
```
OttoAllmendinger, line 288: not sure why you assume that is likely - there can be any number of reasons why finalization fails

can you think of some other way to detect the `SingleLeafScriptAndSignature` condition?

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -282,7 +282,12 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
...
+        return this.finalizeTapInputWithSingleLeafScriptAndSignature(0);
```
OttoAllmendinger, line 289: the argument `0` looks suspicious - shouldn't it be `idx`?

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #3391, opened 2023-03-21
chore(bg-express): add example script for sending from a bg wallet

anshuldoshi:
Context: This script was used for the validation of trading account to trading account withdrawals, so adding this to BitGo Express as an example script for future reference. Although this was used for trading account testing, this script is general to all bitgo wallets.

Ticket: bg-70456

### DIFF
--- a/modules/express/examples/send_from_bitgo_wallet.py
+++ b/modules/express/examples/send_from_bitgo_wallet.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python3
+
+import json
+import requests
+import time
+import pyotp
+
+"""
+This script requires BitGo Express to be running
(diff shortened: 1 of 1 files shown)


sachushaji (commented):
Hey @anshuldoshi ,We have a dedicated examples folder at the root for all examples. I think it would makes sense to add it there. Also I have only seen examples written in JS for almost all usecases, Can we also do the same for this example as well for consistency sake?, Although this one in particular don't require the SDK to be installed as a dependency.

anshuldoshi:
Closing this because a very similar script already exists https://github.com/BitGo/BitGoJS/blob/master/examples/ts/send-wallet-transaction.ts

PR #3392, opened 2023-03-21
feat(utxo-lib): support p2trMusig2 key path finalize

saravanan7mani:
Finalize p2trMusig2 key path PSBT input.

TICKET: BG-66940

### DIFF
--- a/modules/utxo-lib/src/bitgo/Musig2.ts
+++ b/modules/utxo-lib/src/bitgo/Musig2.ts
@@ -10,6 +10,7 @@ import { ecc, musig } from '../noble_ecc';
 import { Tuple } from './types';
 import { calculateTapTweak, tapTweakPubkey } from '../taproot';
 import { SessionKey } from '@brandonblack/musig';
+import { Transaction } from '../index';
 
 /**
  *  Participant key value object.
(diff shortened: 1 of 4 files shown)


modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -322,6 +327,37 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
     return this;
   }
 
+  /**
+   * Finalizes a taproot musig2 input by aggregating all partial sigs.
+   * IMPORTANT: Always call validate* function before finalizing.
```
OttoAllmendinger, line 332: maybe we should do it inside this method then?

what happens if the caller forgets about it?

OttoAllmendinger (dismissed):

modules/utxo-lib/src/bitgo/Musig2.ts
```diff
@@ -413,3 +414,29 @@ export function assertPsbtMusig2Nonces(
     }
   });
 }
+
+/**
+ * @returns Input object but  sig hash type data is taken out from partialSig field.
```
rushilbg, line 419: nit:
```suggestion
 * @returns Input object but sig hash type data is taken out from partialSig field.
```

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
saravanan7mani: Just following the [bitcoinjs-lib](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/ts_src/psbt.ts#L114) Psbt.ts standard flow which is consistence across other input types.

The validate* function checks script, pub keys, sigs and other meta data completely.
Finalize* is a dumb function. Just moves everything to different fields, so later extraction of network tx will be quicker, cheaper and easier.

If the caller doesn't ensure validity with validate* function, there are different possible outcomes:
1. Finalize will be successful if all the signatures exist (but it can be wrong signature or pub keys too)
2. Finalize will throw error if invalid number of sigs exist or few metadata like tapInternalKey is missing.

Upstream requires changes to make the requested behaviour consistent across inputs.

modules/utxo-lib/src/bitgo/Musig2.ts
saravanan7mani: Done

OttoAllmendinger (approved):

PR #3396, opened 2023-03-23
feat(sdk-core): enable fillNonce and acceleration tx for receive address

gianchandania:
This PR adds support for nonce hole filling and consolidation accelerations on an ETH TSS receive address

BG-69374
BG-69376

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -326,6 +326,10 @@ describe('V2 Wallet:', function () {
       ethWallet = new Wallet(bitgo, bitgo.coin('teth'), walletData);
     });
 
+    afterEach(async function() {
+      nock.cleanAll();
+    });
+
     it('should error eip1559 and gasPrice are passed', async function () {
       const params = {
         gasPrice: 100,
@@ -435,6 +439,22 @@ describe('V2 Wallet:', function () {
         // test is successful if nock is consumed, HMAC errors expected
       }
       response.isDone().should.be.true();
+    });
+
+    it('should not pass recipients in sendMany when transaction type is fillNonce', async function () {
+      const recipientAddress = '0x7db562c4dd465cc895761c56f83b6af0e32689ba';
+      const recipients = [{
+        address: recipientAddress,
+        amount: 0,
+      }];
+      const sendManyParams = { recipients, type: 'fillNonce', isTss: true, nonce: '13' };
+
+      try {
+        await ethWallet.sendMany(sendManyParams);
+      } catch (e) {
+        e.message.should.equal('cannot provide recipients for transaction type fillNonce');
+        // test is successful if nock is consumed, HMAC errors expected
+      }
     });
 
     it('should use a custom signing function if provided', async function () {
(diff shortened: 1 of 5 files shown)


modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -1983,6 +1983,43 @@ describe('V2 Wallet:', function () {
...
+        });
+
+        sinon.assert.calledOnce(prebuildTxWithIntent);
+        const args = prebuildTxWithIntent.args[0];
+        args[0]!.should.not.have.property('recipients');
+        args[0]!.feeOptions!.should.deepEqual(feeOptions);
+        args[0]!.lowFeeTxid!.should.equal(lowFeeTxid);
+        args[0]!.receiveAddress!.should.equal(receiveAddress);
+        args[0]!.intentType.should.equal('acceleration');
+        args[1]!.should.equal('full');
```
wesgraham, line 2020: can we have any assertions on the recipients array params being present in the prebuilt response? or the token name/id even?

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2013,6 +2050,39 @@ describe('V2 Wallet:', function () {
         args[1]!.should.equal('full');
       });
 
+      it('should call prebuildTxWithIntent with the correct params for eth fillNonce for receive address nonce filling tx', async function () {
+        const feeOptions = {
```
sachushaji, line 2054: Just checking whether this will work with legacy fee options also for BSC transfers?

sachushaji (approved):

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -1983,6 +1983,43 @@ describe('V2 Wallet:', function () {
         args[1]!.should.equal('full');
       });
 
+      it('should call prebuildTxWithIntent with the correct params for eth accelerations for receive address', async function () {
+        const recipients = [{
```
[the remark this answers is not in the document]
gianchandania: Actually this testcase has a assertion that didn't expect recipients in the output because for intentType fillNonce and accleration we are not passing recipients in the method prebuildTxWithIntent

`args[0]!.should.not.have.property('recipients');`

modules/bitgo/test/v2/unit/wallet.ts
gianchandania: As mentioned above, here we actually checking that recipients should not be there in the response. Apart from this as discussed I have also added a check where for intentType fillNonce and acceleration, users will be thrown an error if they pass recipients

modules/bitgo/test/v2/unit/wallet.ts
gianchandania: Based on this [method](https://github.com/BitGo/bitgo-microservices/blob/b0b1e2d4fdd8937674ef50d33cd594600dcb432f/packages/wallet-platform/app/controllers/api/v2/coins/abstractEthLikeCommon.ts#L2053) in platform, it seems like it should work for legacy fee options as well for BSC.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -188,6 +188,7 @@ export class Wallet implements IWallet {
       'numBlocks',
       'nonce',
       'preview',
+      'receiveAddress',
```
mmcshinsky-bitgo, line 191: Is this documented for user visibility in the developer portal?

sachushaji (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
gianchandania: @mmcshinsky-bitgo We are yet to make any changes to the API docs related to Eth TSS Receive Addresses. Will keep track of this and accomodate it when we make changes to api docs
cc: @wesgraham

PR #3397, opened 2023-03-23
feat(sdk-coin-atom): add function to load inputs and outputs of a txn

dpkjnr:
BaseTransaction has five properties: id, type, inputs, outputs and signature. There is coin specific txn object is also present in extended transaction classes, AtomTransaction in this case. This PR adds a function to transaction class to load inputs and outputs from AtomTransaction. This PR also makes sure that `loadInputsAndOutputs` is called while building the transaction in TransactionBuilder

Ticket: BG-70311

### DIFF
--- a/modules/sdk-coin-atom/src/lib/transaction.ts
+++ b/modules/sdk-coin-atom/src/lib/transaction.ts
@@ -1,6 +1,7 @@
 import {
   BaseKey,
   BaseTransaction,
+  Entry,
   InvalidTransactionError,
   ParseTransactionError,
   TransactionRecipient,
@@ -121,7 +122,7 @@ export class Transaction extends BaseTransaction {
    * Set the transaction type.
    * @param {TransactionType} transactionType The transaction type to be set.
    */
-  transactionType(transactionType: TransactionType): void {
+  set transactionType(transactionType: TransactionType) {
     this._type = transactionType;
   }
 
@@ -138,7 +139,12 @@ export class Transaction extends BaseTransaction {
     if (this.atomTransaction.signature) {
       this.addSignature(Buffer.from(this.atomTransaction.signature).toString('hex'));
     }
-    this._type = utils.getTransactionTypeFromTypeUrl(this._atomTransaction.sendMessages[0].typeUrl);
+    const typeUrl = this.atomTransaction.sendMessages[0].typeUrl;
+    const transactionType = utils.getTransactionTypeFromTypeUrl(typeUrl);
+    if (transactionType === undefined) {
+      throw new Error('Transaction type is not supported ' + typeUrl);
+    }
+    this.transactionType = transactionType;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


Vijay-Jagannathan:
Would be easier if you could add some context on what this change intends to achieve

modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -138,7 +139,12 @@ export class Transaction extends BaseTransaction {
     if (this.atomTransaction.signature) {
       this.addSignature(Buffer.from(this.atomTransaction.signature).toString('hex'));
     }
-    this._type = utils.getTransactionTypeFromTypeUrl(this._atomTransaction.sendMessages[0].typeUrl);
+    const typeUrl = this.atomTransaction.sendMessages[0].typeUrl;
+    const transactionType = utils.getTransactionTypeFromTypeUrl(typeUrl);
+    if (transactionType === undefined) {
```
Vijay-Jagannathan, line 144: would `if (!transactionType)` be cleaner or do we want to only check for undefined ?

modules/sdk-coin-atom/test/unit/transaction.ts
```diff
@@ -38,6 +39,21 @@ describe('Atom Transaction', () => {
       );
       should.equal(Buffer.from(json.signature as any).toString('base64'), testData.TEST_SEND_TX.signature);
       should.equal(tx.type, TransactionType.Send);
+      tx.loadInputsAndOutputs();
+      should.deepEqual(tx.inputs, [
```
Vijay-Jagannathan, line 43: More of a question since this is the first time i'm seeing this. Are we using `deepEqual` instead of `equal` because we're comparing a custom object with child properties ?

modules/sdk-coin-atom/test/unit/transaction.ts
```diff
@@ -38,6 +39,21 @@ describe('Atom Transaction', () => {
...
+        {
+          address: testData.TEST_SEND_TX.sender,
+          value: testData.TEST_SEND_TX.sendMessage.value.amount[0].amount,
+          coin: 'tatom',
```
Vijay-Jagannathan, line 47: nit: We can replace all instances of `'tatom'` with a constant

modules/sdk-coin-atom/src/lib/transaction.ts
dpkjnr: This is tricky with enums(learned it the hard way 😅 ), first entry in the enum has value 0 which is falsy. So condition `if(!transactionType)` would resolve to true for TransactionType.Send

modules/sdk-coin-atom/test/unit/transaction.ts
```diff
@@ -109,6 +170,22 @@ describe('Atom Transaction', () => {
       );
       should.equal(Buffer.from(json.signature as any).toString('base64'), testData.TEST_WITHDRAW_REWARDS_TX.signature);
       should.equal(tx.type, TransactionType.StakingWithdraw);
+
+      tx.loadInputsAndOutputs();
+      should.deepEqual(tx.inputs, [
+        {
+          address: testData.TEST_WITHDRAW_REWARDS_TX.delegator,
+          value: 'UNAVAILABLE',
```
Vijay-Jagannathan, line 178: Any reason why this value is 'UNAVAILABLE' ?

modules/sdk-coin-atom/src/lib/transaction.ts
```diff
@@ -232,4 +238,59 @@ export class Transaction extends BaseTransaction {
...
+          address: delegateMessage.validatorAddress,
+          value: delegateMessage.amount.amount,
+          coin: this._coinConfig.name,
+        });
+        break;
+      case TransactionType.StakingWithdraw:
+        const withdrawMessage = this.atomTransaction.sendMessages[0].value as WithdrawDelegatorRewardsMessage;
+        inputs.push({
+          address: withdrawMessage.delegatorAddress,
+          value: UNAVAILABLE_TEXT,
```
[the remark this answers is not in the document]
dpkjnr: value is not an optional field in Entry type, and WithdrawRewards transaction doesn't have amount in Atom

modules/sdk-coin-atom/src/lib/transactionBuilder.ts
```diff
@@ -143,7 +143,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
 
   /** @inheritdoc */
   protected async buildImplementation(): Promise<Transaction> {
-    this.transaction.transactionType(this.transactionType);
+    this.transaction.transactionType = this.transactionType;
```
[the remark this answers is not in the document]
dpkjnr: I changed function `transactionType` to as setter function, and syntax is different

modules/sdk-coin-atom/src/lib/transaction.ts
Vijay-Jagannathan: Ah I see

modules/sdk-coin-atom/test/unit/transaction.ts
Vijay-Jagannathan: Assuming this is similar to the previous comment (re: not optional and therefore defaulting to some text)

Vijay-Jagannathan (approved):

modules/sdk-coin-atom/test/unit/transaction.ts
dpkjnr: thats correct

PR #3398, opened 2023-03-23
chore(sdk-core): export ecdsa and bip32 classes

alia-bitgo:
Exporting ecdsa and bip32 classes to use in WP.

TICKET: BG-69708

### DIFF
--- a/modules/sdk-core/src/account-lib/mpc/index.ts
+++ b/modules/sdk-core/src/account-lib/mpc/index.ts
@@ -1,4 +1,4 @@
-import HDTree, { Ed25519BIP32 } from './hdTree';
+import HDTree, { Ed25519BIP32, BIP32 } from './hdTree';
 import { EDDSA } from './tss';
 import ShamirSecret from './shamir';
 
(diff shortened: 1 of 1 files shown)


olibeyene (approved):

modules/account-lib/src/index.ts
```diff
@@ -8,8 +8,10 @@ import {
   Ed25519BIP32,
   Ed25519KeyDeriver,
   Eddsa,
+  Ecdsa,
+  BIP32,
 } from '@bitgo/sdk-core';
-export { Eddsa, Ed25519BIP32 };
+export { Eddsa, Ed25519BIP32, Ecdsa, BIP32 };
```
sachushaji, line 14: Ecdsa, BIP32 is available from sdk core, Can you directly use it from sdk-core package, I don't think we need to export here also

alia-bitgo: Only exporting BIP32 now

sachushaji (approved):

PR #3400, opened 2023-03-26
feat(utxo-lib): add integration test for p2trMusig2

saravanan7mani:
Add deposit and spend integration test for key path and script path `p2trMusig2`.
This PR tests PSBT build, nonce creation, sign, verify and finalizer works against a `regtest` node.
Testing p2trMusig2 with parse test suite is not covered in this PR.
p2trMusig2 support for `parseSignatureScript`, `UtxoPsbt.fromTransaction` are required here.
And now those are not that critical, so will revisit parse test case later when required.

TICKET: BG-71051

### DIFF
--- a/modules/utxo-lib/test/integration_local_rpc/generate/main.ts
+++ b/modules/utxo-lib/test/integration_local_rpc/generate/main.ts
@@ -8,9 +8,11 @@ import { getRegtestNode, getRegtestNodeUrl, Node, getRegtestNodeHelp } from './r
 import {
   createScriptPubKey,
   createSpendTransaction,
+  createPsbtSpendTransaction,
   isSupportedDepositType,
   ScriptType,
   scriptTypes,
+  getP2trMusig2Index,
 } from './outputScripts.util';
 import { RpcClient } from './RpcClient';
 import {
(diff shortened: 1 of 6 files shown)


modules/utxo-lib/test/integration_local_rpc/generate/main.ts
```diff
@@ -142,24 +156,48 @@ async function createTransactionsForScriptType(
...
+      } else {
+        spendTx = createSpendTransaction(
+          keys,
+          scriptType,
+          [deposit1Tx, deposit2Tx],
+          script,
+          protocol.network,
+          protocol.version
+        );
+      }
```
OttoAllmendinger, line 185: let's move this to a `getSpendTx` func

modules/utxo-lib/test/integration_local_rpc/generate/outputScripts.util.ts
```diff
@@ -158,3 +168,92 @@ export function createSpendTransaction<TNumber extends number | bigint = number>
...
+}
+
+/**
+ * @returns BIP32 hardcoded index for p2trMusig2 spend type. 0 for key path and 100 for script path.
+ * For same fixture key triple and script type (p2trMusig2),
+ * we need 2 different deposit and spend tx fixtures.
+ */
+export function getP2trMusig2Index(spendType: 'KeyPath' | 'ScriptPath'): number {
+  return spendType === 'KeyPath' ? 0 : 100;
+}
```
OttoAllmendinger, line 259: nit: we tend to use camelcase with lowercase initial character in utxolib

```suggestion

export type Musig2SpendType = 'keyPath' | 'scriptPath`;

/**
 * @returns BIP32 hardcoded index for p2trMusig2 spend type. 0 for key path and 100 for script path.
 * For same fixture key triple and script type (p2trMusig2),
 * we need 2 different deposit and spend tx fixtures.
 */
export function getP2trMusig2Index(spendType: SpendPath): number {
  return spendType === 'keyPath' ? 0 : 100;
}
```

OttoAllmendinger (changes requested):
two nits

modules/utxo-lib/test/integration_local_rpc/generate/outputScripts.util.ts
saravanan7mani: Done

modules/utxo-lib/test/integration_local_rpc/generate/main.ts
saravanan7mani: Done

OttoAllmendinger (approved):

PR #3404, opened 2023-03-27
feat(sdk-core): method to parse json from OVC to create TSS bitgo key

pranavjain97:
This exposes a new SDK function that 
1. Takes the JSON input from the [OVC (offline vault console) ](https://github.com/BitGo/offline-vault-console), which has the shares _for BitGo._
2. Parses the JSON input. 
3. Creates a corresponding bitgo key
4. Returns the shares from BitGo -> OVCs.

### DIFF
--- a/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
+++ b/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
@@ -1,6 +1,7 @@
 import { IRequestTracer } from '../../api';
 import { KeyPair, KeychainsTriplet } from '../baseCoin';
 import { BackupProvider, IWallet } from '../wallet';
+import { BitGoKeyFromOvcShares, OvcToBitGoJSON } from './ovcJsonCodec';
 
 export type KeyType = 'tss' | 'independent' | 'blsdkg';
 
@@ -130,58 +131,6 @@ export enum KeyIndices {
   BACKUP = 1,
   BITGO = 2,
 }
-
-export type BitGoKeyFromOvcShares = {
-  bitGoOutputJsonForOvc: BitGoToOvcJSON;
-  bitGoKeyId: string;
-};
-
-export type OvcToBitGoJSON = {
-  tssVersion: string;
-  walletType: string;
-  coin: string;
-  state: number;
-  ovc: {
-    1: {
-      gpgPubKey: string;
-      ovcToBitgoShare: OvcToOtherShare;
-    };
-    2: {
-      gpgPubKey: string;
-      ovcToBitgoShare: OvcToOtherShare;
-      ovcToOvcShare: OvcToOtherShare;
-    };
-  };
-};
-
-export type BitGoToOvcJSON = OvcToBitGoJSON & {
-  platform: {
-    commonKeychain: string;
-    walletHSMGPGPublicKeySigs: string;
-    ovc: {
-      // BitGo to User (OVC-1)
-      1: {
-        bitgoToOvcShare: OvcShare;
-      };
-      // BitGo to Backup (OVC-2)
-      2: {
-        bitgoToOvcShare: OvcShare;
-      };
-    };
-  };
-};
-
-export type OvcShare = {
-  publicShare: string;
-  privateShare: string;
-  vssProof: string;
-  i: number;
-  j: number;
-};
-
-export type OvcToOtherShare = OvcShare & {
-  uSig: number;
-};
 
 export interface IKeychains {
   get(params: GetKeychainOptions): Promise<Keychain>;
--- a/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
+++ b/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
@@ -131,6 +131,58 @@ export enum KeyIndices {
   BITGO = 2,
 }
 
+export type BitGoKeyFromOvcShares = {
+  bitGoOutputJsonForOvc: BitGoToOvcJSON;
+  bitGoKeyId: string;
+};
+
+export type OvcToBitGoJSON = {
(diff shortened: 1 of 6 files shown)


byhow (dismissed):

modules/sdk-core/src/bitgo/keychain/keychains.ts
```diff
@@ -305,4 +310,121 @@ export class Keychains implements IKeychains {
...
   }
+
+  /**
+   * It parses the JSON downloaded from the OVC for platform (BitGo),
+   * and creates a corresponding TSS BitGo key. It also returns the JSON that needs
+   * to be uploaded back to the OVCs containing the BitGo -> OVC shares.
+   * @param ovcOutputJson JSON format of the file downloaded from the OVC for platform
+   * @returns {BitGoKeyFromOvcShares}
+   */
+  async createTssBitGoKeyFromOvcShares(ovcOutputJson: OvcToBitGoJSON): Promise<BitGoKeyFromOvcShares> {
```
zahin-mohammad, line 321: A lot of the validation done here can be done via io-ts. 

We have io-ts available in sdk-core, lets instead create a strict codec and decode the json file using the codec.

zahin-mohammad (changes requested):

modules/sdk-core/src/bitgo/keychain/keychains.ts
pranavjain97: I need custom errors for each field that is missing, and io-ts decode will not support that and throw an arbitary error if it doesn't match the coded. In addition, i do already have a type defined for the JSON and if(xyz) will check for a non-empty string. I can add tests for some non-happy paths to test this.

modules/sdk-core/src/bitgo/keychain/keychains.ts
```diff
@@ -305,4 +312,110 @@ export class Keychains implements IKeychains {
...
   }
+
+  /**
+   * It parses the JSON downloaded from the OVC for platform (BitGo),
+   * and creates a corresponding TSS BitGo key. It also returns the JSON that needs
+   * to be uploaded back to the OVCs containing the BitGo -> OVC shares.
+   * @param ovcOutputJson JSON format of the file downloaded from the OVC for platform
+   * @returns {BitGoKeyFromOvcShares}
+   */
+  async createTssBitGoKeyFromOvcShares(ovcOutputJson: OvcToBitGoJson): Promise<BitgoKeyFromOvcShares> {
```
zahin-mohammad, line 323:
```suggestion
  async createTssBitGoKeyFromOvcShares(ovcOutputJson: unknown): Promise<BitgoKeyFromOvcShares> {
```

Since we are using io-ts, we can rely on the decoded value to have a type.

modules/sdk-core/src/bitgo/keychain/iKeychains.ts
```diff
@@ -131,6 +132,85 @@ export enum KeyIndices {
...
+  ],
+  'OvcToOtherShare'
+);
+
+export const OvcToBitGoJSON = t.strict(
+  {
+    tssVersion: t.string,
+    walletType: t.string,
+    coin: t.string,
+    state: t.number,
```
zahin-mohammad, line 161:
```suggestion
    state: t.literal(1),
```

modules/sdk-core/src/bitgo/keychain/iKeychains.ts
```diff
@@ -131,6 +132,85 @@ export enum KeyIndices {
...
+
+export const BitGoKeyFromOvcShares = t.strict(
+  {
+    bitGoOutputJsonForOvc: BitGoToOvcJSON,
+    bitGoKeyId: t.string,
+  },
+  'BitGoKeyFromOvcShares'
+);
+
+export type BitgoKeyFromOvcShares = t.TypeOf<typeof BitGoKeyFromOvcShares>;
```
zahin-mohammad, line 212:
```suggestion
export type BitGoKeyFromOvcShares = t.TypeOf<typeof BitGoKeyFromOvcShares>;
```

zahin-mohammad (changes requested):
lgtm, but some small requested changes.

modules/sdk-core/src/bitgo/keychain/keychains.ts
```diff
@@ -305,4 +312,110 @@ export class Keychains implements IKeychains {
...
+    const decodedOvcOutput = decodeOrElse(OvcToBitGoJSON.name, OvcToBitGoJSON, ovcOutputJson, (errors) => {
+      throw new Error(`Error(s) parsing OVC JSON: ${errors}`);
+    });
+
+    if (decodedOvcOutput.state !== 1) {
+      throw new Error('State expected to be "1". Please complete the first two OVC operations');
+    }
```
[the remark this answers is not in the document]
pranavjain97: It can be, but i want a custom error message for it

modules/sdk-core/src/bitgo/keychain/iKeychains.ts
```diff
@@ -131,6 +132,85 @@ export enum KeyIndices {
...
+        gpgPubKey: t.string,
+        ovcToBitgoShare: OvcToOtherShare,
+        ovcToOvcShare: OvcToOtherShare,
+      }),
+    }),
+  },
+  'OvcToBitGoJSON'
+);
+
+export type OvcToBitGoJson = t.TypeOf<typeof OvcToBitGoJSON>;
```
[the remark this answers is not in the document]
pranavjain97: For some reason, es-lint for BitGoJS isnt allowing that pattern to use the same name for type/const. I know we do that in bgms.

zahin-mohammad (approved):

PR #3423, opened 2023-03-31
refactor(utxo-lib): tighten constraints in validateScript

OttoAllmendinger:
We don't need to consider the `| undefined` case for `psbtParsed`

Issue: BG-66942


saravanan7mani (approved):
LGTM

PR #3424, opened 2023-03-31
feat(utxo-lib): introduce ParsedScriptType

saravanan7mani:
BREAKING CHANGE

When parsing inputs, we cannot always map the parsed script type to a single outputScript.

We introduce a new type to represent the parse result.

TICKET: BG-66942

### DIFF
--- a/modules/utxo-lib/src/bitgo/parseInput.ts
+++ b/modules/utxo-lib/src/bitgo/parseInput.ts
@@ -1,7 +1,7 @@
+/* eslint no-redeclare: 0 */
 import * as opcodes from 'bitcoin-ops';
 import { TxInput, script as bscript } from 'bitcoinjs-lib';
 
-import { isScriptType2Of3, ScriptType, ScriptType2Of3 } from './outputScripts';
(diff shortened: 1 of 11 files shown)


modules/utxo-lib/src/bitgo/parseInput.ts
```diff
@@ -45,12 +45,22 @@ export function getLeafVersion(controlBlock: Buffer): number {
...
 }
 
-interface ParsedScript {
-  scriptType: ScriptType;
-}
+export type ParsedScriptType2Of3 =
+  | 'p2sh'
+  | 'p2shP2wsh'
+  | 'p2wsh'
+  | 'taprootKeyPathSpend' // can be for either p2tr or p2trMusig2 output script
```
OttoAllmendinger, line 52:
```suggestion
  | 'taprootKeyPathSpend' // only implemented for p2trMusig2
```

OttoAllmendinger (dismissed):
1 nit

OttoAllmendinger (approved):

PR #3429, opened 2023-03-31
fix: use p2wsh and internalChainCode for changeAddress creation

lcovar:
TICKET: BG-72015

### DIFF
--- a/modules/sdk-coin-btc/src/inscriptionBuilder.ts
+++ b/modules/sdk-coin-btc/src/inscriptionBuilder.ts
@@ -62,16 +62,16 @@ export class InscriptionBuilder implements IInscriptionBuilder {
       signer = 'user',
       cosigner = 'bitgo',
       inscriptionConstraints = DefaultInscriptionConstraints,
-      changeAddressType = 'p2tr',
+      changeAddressType = 'p2wsh',
     }: {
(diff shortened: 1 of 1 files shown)


JamesJo06 (approved):

modules/sdk-coin-btc/src/inscriptionBuilder.ts
```diff
@@ -62,16 +62,16 @@ export class InscriptionBuilder implements IInscriptionBuilder {
       signer = 'user',
       cosigner = 'bitgo',
       inscriptionConstraints = DefaultInscriptionConstraints,
-      changeAddressType = 'p2tr',
+      changeAddressType = 'p2wsh',
     }: {
-      signer: utxolib.bitgo.KeyName;
-      cosigner: utxolib.bitgo.KeyName;
-      inscriptionConstraints: {
+      signer?: utxolib.bitgo.KeyName;
```
brandonblack, line 67: why these optional? Just not needed for p2wsh but were for p2tr?

JamesJo06: I had to pass in all those parameters because they were not optional even though they have default values

brandonblack (approved):

PR #3436, opened 2023-04-03
feat(sdk-core): add sha256 hash support in ecdsa tss signing

dpkjnr:
Ticket: BG-72077

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -88,6 +88,8 @@ import { Lightning } from '../lightning';
 import EddsaUtils from '../utils/tss/eddsa';
 import { EcdsaUtils } from '../utils/tss/ecdsa';
 import { getTxRequest } from '../tss';
+import { createHash, Hash } from 'crypto';
+import createKeccakHash from 'keccak';
 
 const debug = require('debug')('bitgo:v2:wallet');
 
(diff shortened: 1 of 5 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2804,11 +2806,20 @@ export class Wallet implements IWallet {
...
     }
     try {
+      const hashAlgorithm = this.coinSpecific()?.hashAlgorithm;
+      let hash: Hash;
+      if (hashAlgorithm && hashAlgorithm.startsWith('sha')) {
+        hash = createHash(hashAlgorithm);
+      } else {
+        hash = createKeccakHash('keccak256');
+      }
+
```
sachushaji, line 2816:
```suggestion
      const hashAlgorithm = this.coinSpecific()?.hashAlgorithm || 'keccak256';
      const hash = createHash(hashAlgorithm);
```

sachushaji: I think this might be a bit better approach to do this

mukeshsp (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
sachushaji: Had a conversation with @dpkjnr , Seems Keccack is not readily available with node crypto and have to use an external dependency to support the same

sachushaji (approved):

byhow (approved):

PR #3438, opened 2023-04-03
feat(sdk-coin-sol): add create ATA to token transfer for SOL

alebusse:
added create ATA instruction for token transfer for SOL

BG-70372


modules/sdk-coin-sol/src/lib/tokenTransferBuilder.ts
```diff
@@ -64,12 +81,31 @@ export class TokenTransferBuilder extends TransactionBuilder {
...
   }
 
+  /**
+   *
+   * @param {TokenAssociateRecipient} recipient - recipient of the associated token account creation
+   * @param {string} recipient.ownerAddress - owner of the associated token account
+   * @param {string} recipient.tokenName - name of the token that is intended to associate
+   * @returns {TransactionBuilder} This transaction builder
+   */
+  createAssociatedTokenAccount(recipient: TokenAssociateRecipient): this {
```
joey-bai, line 91: do we need a separate function for this?

was thinking to update `send` to take a new param: `fundAtaCreation: boolean`.
when true, we would attach the create ATA instruction.

joey-bai (approved):
overall lgtm, one comment/question about function design, but not blocking.
thanks for adding tests too!

alebusse:
Testing with an Alpha release before merging

PR #3440, opened 2023-04-03
chore: only check for address details if address is not in recipient

lcovar:
TICKET: BG-72170

### DIFF
--- a/modules/abstract-utxo/src/parseOutput.ts
+++ b/modules/abstract-utxo/src/parseOutput.ts
@@ -220,6 +220,7 @@ export async function parseOutput({
   debug('Parsing address details for %s', currentAddress);
   let currentAddressDetails = undefined;
   let currentAddressType: string | undefined = undefined;
+  const RECIPIENT_THRESHOLD = 1000;
   try {
     /**
      * The only way to determine whether an address is known on the wallet is to initiate a network request and
@@ -228,6 +229,21 @@ export async function parseOutput({
      * details are fetched on the wallet, a local address validation is run, whose errors however are generated
      * client-side and can therefore be analyzed with more granularity and type checking.
      */
+
+    /**
+     * In order to minimize API requests, we assume that explicit recipients are always external when the
+     * recipient list is > 1000 This is not always a valid assumption and could lead greater apparent spend (but never lower)
+     */
+    if (txParams.recipients !== undefined && txParams.recipients.length > RECIPIENT_THRESHOLD) {
+      const isCurrentAddressInRecipients = txParams.recipients.some((recipient) =>
+        recipient.address.includes(currentAddress)
+      );
+
+      if (isCurrentAddressInRecipients) {
+        return { ...currentOutput, external: true };
+      }
+    }
+
     const addressDetails = await fetchAddressDetails({
       reqId,
       addressDetailsVerification,


modules/abstract-utxo/src/parseOutput.ts
```diff
@@ -228,14 +228,20 @@ export async function parseOutput({
...
+          reqId,
+          addressDetailsVerification,
+          addressDetailsPrebuild,
+          currentAddress,
+          disableNetworking,
+          wallet,
+        });
     // verify that the address is on the wallet. verifyAddress throws if
     // it fails to correctly rederive the address, meaning it's external
     currentAddressType = AbstractUtxoCoin.inferAddressType(addressDetails) || undefined;
```
OttoAllmendinger, line 247: I'm surprised this compiles - the argument to `inferAddressType` does not allow `undefined` here. This is probably because `fetchAddressDetails` returns `any`.

I'm pretty sure that `inferAddressType(undefined)` fails and we end up in that overly broad `catch` block.

modules/abstract-utxo/src/parseOutput.ts
```diff
@@ -228,14 +228,20 @@ export async function parseOutput({
...
-      reqId,
-      addressDetailsVerification,
-      addressDetailsPrebuild,
-      currentAddress,
-      disableNetworking,
-      wallet,
-    });
+    const isCurrentAddressInRecipients = txParams.recipients?.some((recipient) =>
+      recipient.address.includes(currentAddress)
+    );
```
OttoAllmendinger, line 233: The rest of this func doesn't do anything useful if this is `false`. I believe we should just `return currentOutput` in that case.

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -387,8 +387,8 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
    * Determine an address' type based on its witness and redeem script presence
    * @param addressDetails
    */
-  static inferAddressType(addressDetails: { chain: number }): ScriptType2Of3 | null {
-    return isChainCode(addressDetails.chain) ? scriptTypeForChain(addressDetails.chain) : null;
+  static inferAddressType(addressDetails: { chain?: number }): ScriptType2Of3 | null {
+    return isChainCode(addressDetails?.chain) ? scriptTypeForChain(addressDetails?.chain) : null;
```
OttoAllmendinger, line 391: unclear why you are doing this change

OttoAllmendinger (dismissed):
see internal slack

modules/abstract-utxo/src/abstractUtxoCoin.ts
lcovar: Because of this https://github.com/BitGo/BitGoJS/pull/3440#discussion_r1156879526 

It was throwing before since it did no allow undefined. With this, undefined is accepted.

modules/abstract-utxo/src/parseOutput.ts
lcovar: https://github.com/BitGo/BitGoJS/pull/3440#discussion_r1157514684

modules/abstract-utxo/src/parseOutput.ts
lcovar: Tried this but run into test failure where we don't decorate a change address with external true if I return early.

modules/abstract-utxo/src/abstractUtxoCoin.ts
OttoAllmendinger: But then you would also have to change the type signature here.

But instead of allowing `undefined` here rather just handle it at the callsite.

modules/abstract-utxo/src/parseOutput.ts
OttoAllmendinger: please rewrite the code so that we return early and not in the `catch` block

modules/abstract-utxo/src/parseOutput.ts
lcovar: Skipped the test for now since test expects to decorate an output when the address is part of the recipients which is no longer the case with this chance. Added a TODO to unskip.

modules/abstract-utxo/src/parseOutput.ts
lcovar: done

modules/bitgo/test/v2/unit/coins/abstractUtxoCoin.ts
```diff
@@ -94,9 +94,10 @@ describe('Abstract UTXO Coin:', () => {
       return runClassifyOutputsTest(changeAddress, verification, false, { changeAddress, recipients: [] });
     });
 
-    it('should classify outputs with external address in recipients as explicit', async function () {
+    // TODO(BG-72251): unskip when we use psbt and dont need to make network calls to get address details
+    xit('should classify outputs with external address in recipients as explicit', async function () {
       const externalAddress = '2NAuziD75WnPPHJVwnd4ckgY4SuJaDVVbMD';
-      return runClassifyOutputsTest(externalAddress, verification, true, { recipients: [{ address: externalAddress, amount: outputAmount }] });
+      return runClassifyOutputsTest(externalAddress, verification, false, { recipients: [{ address: externalAddress, amount: outputAmount }] });
     });
```
zahin-mohammad, line 101: Not sure why wallets team owns this file 🤔

zahin-mohammad (commented):
commenting to flush review Q

OttoAllmendinger (commented):
we should consider alternative approaches

modules/abstract-utxo/src/parseOutput.ts
```diff
@@ -228,6 +230,22 @@ export async function parseOutput({
      * details are fetched on the wallet, a local address validation is run, whose errors however are generated
      * client-side and can therefore be analyzed with more granularity and type checking.
      */
+
+    /**
+     * In order to minimize API requests, we assume that explicit recipients are always external when the
+     * recipient list is > 1000 This is not always a valid assumption and can lead to unexpected behavior.
```
brandonblack, line 236:
```suggestion
     * recipient list is > 1000 This is not always a valid assumption and could lead to greater apparent spend (but never lower).
```

modules/abstract-utxo/src/parseOutput.ts
```diff
@@ -228,6 +230,22 @@ export async function parseOutput({
...
+     */
+    assert(txParams.recipients);
+    if (txParams.recipients?.length > RECIPIENT_THRESHOLD) {
+      const isCurrentAddressInRecipients = txParams.recipients?.some((recipient) =>
```
brandonblack, line 240: Shouldn't need the `?`s with the assert
```suggestion
    if (txParams.recipients.length > RECIPIENT_THRESHOLD) {
      const isCurrentAddressInRecipients = txParams.recipients.some((recipient) =>
```

brandonblack (changes requested):

brandonblack (approved):

modules/abstract-utxo/src/parseOutput.ts
```diff
@@ -228,6 +230,22 @@ export async function parseOutput({
      * details are fetched on the wallet, a local address validation is run, whose errors however are generated
      * client-side and can therefore be analyzed with more granularity and type checking.
      */
+
+    /**
+     * In order to minimize API requests, we assume that explicit recipients are always external when the
+     * recipient list is > 1000 This is not always a valid assumption and could lead greater apparent spend (but never lower)
+     */
+    assert(txParams.recipients);
+    if (txParams.recipients.length > RECIPIENT_THRESHOLD) {
```
brandonblack, line 239:
```suggestion
    if (txParams.recipients !== undefined && txParams.recipients.length > RECIPIENT_THRESHOLD) {
```

brandonblack (changes requested):

examples/v1btc/addys.txt
```diff
@@ -0,0 +1,245 @@
+tb1qcgcw9hzj3wmvnw9l335eyt6yk2zg0wa39z622lsm0nptw0980kks0yxyn9
```
brandonblack, line 1: extra file snuck in?

brandonblack (changes requested):

brandonblack (dismissed):

brandonblack:
rebased

brandonblack (dismissed):

modules/abstract-utxo/src/parseOutput.ts
```diff
@@ -228,6 +229,21 @@ export async function parseOutput({
      * details are fetched on the wallet, a local address validation is run, whose errors however are generated
      * client-side and can therefore be analyzed with more granularity and type checking.
      */
+
+    /**
+     * In order to minimize API requests, we assume that explicit recipients are always external when the
```
mmcshinsky-bitgo, line 234: Is this something that would also be helpful to add in developers.bitgo.com?

brandonblack: Maybe, but it might only add to confusion.

brandonblack (approved):

PR #3442, opened 2023-04-04
feat(sdk-core): add openssl wasm support for fast safe prime generation

rlutzbitgo:
TICKET: BG-72158


rlutzbitgo:
Well, that vulnerability report isn't great if it's not a false positive. I guess I'll just work on setting up a new project/pipeline to cross-compile openssl from source to wasm via emscripten or zig.

PR #3450, opened 2023-04-06
feat: use isCpfp flag during send on accelerateTransaction

lcovar:
See internal ticket BG-23844

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -1885,7 +1885,7 @@ export class Wallet implements IWallet {
     // We must pass the build params through to submit in case the CPFP tx ever has to be rebuilt.
     const submitParams = Object.assign(params, await this.prebuildAndSignTransaction(params));
     delete (submitParams as any).wallet;
-    return await this.submitTransaction(submitParams);
+    return await this.submitTransaction({ ...submitParams, isCpfp: true });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1885,7 +1885,7 @@ export class Wallet implements IWallet {
     // We must pass the build params through to submit in case the CPFP tx ever has to be rebuilt.
     const submitParams = Object.assign(params, await this.prebuildAndSignTransaction(params));
     delete (submitParams as any).wallet;
-    return await this.submitTransaction(submitParams);
+    return await this.submitTransaction({ ...submitParams, isCpfp: true });
```
OttoAllmendinger, line 1888: this is redundant with `cpfpTxIds` being set

OttoAllmendinger (changes requested):

OttoAllmendinger (changes requested):
We have similar issues with other transaction types. How about we send all the original build parameters in `submitTransaction`?

OttoAllmendinger:
alternatively consider passing `reserveInputs: true` which better communicates what happens on the other end

LeoKotschenreuther (commented):
I have no context on this. I can rubber stamp in case you need a review from me once the btc team has given their ok.

lcovar:
> We have similar issues with other transaction types. How about we send all the original build parameters in `submitTransaction`?

Converted to draft, i'll make this change. Better solution.

lcovar:
closed, redundant. thanks @OttoAllmendinger

PR #3443, opened 2023-04-04
Bg 72158 safeprime generation 2

rlutzbitgo:
TICKET: BG-72158

# Please be aware of the following when making your pull request:

## Description

We need to generate safe primes for ECDSA TSS signing. This is supported natively by the crypto lib in Node versions >=15. Node 14 is EOL in a few weeks, so this PR gets a head start on that by removing support for Node 14 and only supporting 16 (while we take the time to check support for Node versions [16, 18]). It also adds support for very fast generation of safe primes in the use of ntilde creation for ECDSA TSS. 

I plan on attempting to add support for 18 in a fast follow-up.

## Type of change

Please delete options that are not relevant.

- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [x] My code compiles correctly for both Node and Browser environments
- [ ] I have commented my code, particularly in hard-to-understand areas
- [X] My commits follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) and I have properly described any BREAKING CHANGES
- [X] The ticket or github issue was included in the commit message as a reference
- [x] I have made corresponding changes to the documentation and on any new/updated functions and/or methods - [jsdoc](https://jsdoc.app/)
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

PR #3452, opened 2023-04-06
feat(sdk-core): generate and use safe primes

rlutzbitgo:
TICKET: BG-72158

There is another PR which enables safe prime generation and also updates the SDK to deprecate Node 14 support, but it likely would have caused issues with things dependent on the SDK which haven't been migrated off yet. This PR enables safe prime generation on Node versions >=15 but throws on older versions. https://github.com/BitGo/BitGoJS/pull/3443

My hope is that we can migrate dependent projects to Node 16 and then revisit this other PR to change SDK to support Node [16, 18].

### DIFF
--- a/modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts
+++ b/modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts
@@ -12,6 +12,8 @@ import {
   mockFKeyShare,
 } from '../fixtures/ecdsa';
 import { Hash, randomBytes } from 'crypto';
+import * as assert from 'assert';
+import { bitLength, isProbablyPrime } from 'bigint-crypto-utils';
 /**
  * @prettier
  */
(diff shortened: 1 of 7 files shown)


modules/account-lib/test/unit/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -81,6 +83,32 @@ describe('TSS ECDSA TESTS', function () {
...
+    it('should generate safe primes of a specified bitlength', async function () {
+      const bitlength = 128; // this should be small enough to be reasonably fast
+      if (NODE_MAJOR_VERSION === NaN || NODE_MAJOR_VERSION <= 14) {
+        await assert.rejects(rangeProof.generateSafePrime(bitlength));
+      } else {
+        const p = await rangeProof.generateSafePrime(bitlength);
+        (await isProbablyPrime(p)).should.be.true();
+        const q = (p - BigInt(1)) / BigInt(2);
+        (await isProbablyPrime(q)).should.be.true();
+        bitLength(p).should.equal(bitlength);
```
sachushaji, line 97: I may be missing something here, But does q always resolves to a prime number given p is prime? (eg: when p =17)

LeoKotschenreuther (commented):
Automatic comment to flush this PR from my review queue. 🚽🚽🚽
Please re-request another review from the respective team(s) when you are ready for reviews.

PR #3500, opened 2023-04-20
feat(sdk-core): add isRebuild param to tss prebuild transaction

CapnMigraine:
Add a rebuild flag to allow wallet platform determine whether a pending approval nonce handler can be used to determine the nonce for the transaction

Ticket: BG-74457


LeoKotschenreuther (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2737,6 +2742,7 @@ export class Wallet implements IWallet {
             nonce: params.nonce,
             receiveAddress: params.receiveAddress,
             feeOptions,
+            isRebuild: !!params.isRebuild,
```
sachushaji, line 2745: This should always be false right since fill nonce is trying to fill a specific nonce?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2723,6 +2727,7 @@ export class Wallet implements IWallet {
             lowFeeTxid: params.lowFeeTxid,
             receiveAddress: params.receiveAddress,
             feeOptions,
+            isRebuild: !!params.isRebuild,
```
sachushaji, line 2730: Also false for acceleration since we are trying to accelerate the existing transaction and no need to consider pending approvals?

PR #3501, opened 2023-04-20
chore: use utxolib p2tr payment function to retrieve address from script

lcovar:
We added p2tr payments from our fork bitcoinjs-lib to utxo-lib instead. Our bitcoinjs fork will no longer have p2tr support so utxo-lib should take care of retriving an p2tr address from outputScript and bitcoinjs-lib can handle the other types.

TICKET: BG-74542

### DIFF
--- a/modules/utxo-lib/src/address.ts
+++ b/modules/utxo-lib/src/address.ts
@@ -3,11 +3,22 @@ import { Base58CheckResult, Bech32Result } from 'bitcoinjs-lib/src/address';
 
 import * as zcashAddress from '../src/bitgo/zcash/address';
 import { isValidNetwork, isZcash, Network } from './networks';
+import { p2trPayments } from './index';
 
 export function fromOutputScript(outputScript: Buffer, network: Network): string {
   if (isValidNetwork(network) && isZcash(network)) {
     return zcashAddress.fromOutputScript(outputScript, network);
   }
+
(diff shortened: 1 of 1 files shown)


OttoAllmendinger (changes requested):
see other comment

modules/utxo-lib/src/address.ts
```diff
@@ -3,11 +3,18 @@ import { Base58CheckResult, Bech32Result } from 'bitcoinjs-lib/src/address';
...
 import * as zcashAddress from '../src/bitgo/zcash/address';
 import { isValidNetwork, isZcash, Network } from './networks';
+import { p2trPayments } from './index';
 
 export function fromOutputScript(outputScript: Buffer, network: Network): string {
   if (isValidNetwork(network) && isZcash(network)) {
     return zcashAddress.fromOutputScript(outputScript, network);
   }
+  try {
+    return p2trPayments.p2tr({ output: outputScript, network }).address as string;
```
OttoAllmendinger, line 13: please repeat the explanation in the commit message here

OttoAllmendinger (changes requested):

modules/utxo-lib/src/address.ts
lcovar: done

OttoAllmendinger (approved):

PR #3502, opened 2023-04-24
feat(sdk-core): add proof of paillier correctness

johnoliverdriscoll:
Ticket: BG-74325

This is an implementation of the protocol found in [Efficient Noninteractive Certification of RSA Moduli and Beyond](https://eprint.iacr.org/2018/057.pdf#page6) Section 3.2.

This is needed to ensure the Paillier key used during share conversion isn't a spoofed value that allows extraction of signing material.

BREAKING CHANGES: new required fields for ECDSA share types

### DIFF
--- a/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/paillierproof.ts
+++ b/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/paillierproof.ts
@@ -1,0 +1,67 @@
+import { bitLength, randBits } from 'bigint-crypto-utils';
+import { bigIntFromBufferBE } from '../../util';
+import { gcd, modInv, modPow } from 'bigint-mod-arith';
+
+// Security parameters.
+const k = 128;
+const alpha = 2;
+
+// Generate a random coprime less than x with the same bit depth.
+async function randomCoPrimeLessThan(x: bigint): Promise<bigint> {
+  while (true) {
+    const y = bigIntFromBufferBE(Buffer.from(await randBits(bitLength(x), true)));
+    if (y > BigInt(0) && y < x && gcd(x, y) === BigInt(1)) {
+      return y;
+    }
+  }
+}
+
+export async function generateP(n: bigint): Promise<Array<bigint>> {
+  const m = k / Math.log2(alpha);
+  return Promise.all(
+    Array(m)
+      .fill(null)
+      .map(() => randomCoPrimeLessThan(n))
+  );
+}
+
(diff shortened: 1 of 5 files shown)


sachushaji (commented):
Seems build failure is genuine

johnoliverdriscoll:
I still have to code this feature in hsm fw. I was able to build sdk-core and run the tests in account-lib, which is enough for me to code fw tests. I'm going to need to rely on people with more knowledge of sdk to integrate these changes with the rest of the sdk codebase.

LeoKotschenreuther (commented):
Unsubscribe, lack of context. If you need help I suggest you ask in Slack.

zahin-mohammad:
Hey @johnoliverdriscoll , ill help out with this PR. I will move this to `sdk-lib-mpc` which I am introducing here https://github.com/BitGo/BitGoJS/pull/3533. 

It should make it easier to introduce core math/protocol functions, without the bloat of the rest of bitgo.

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/paillierproof.ts
```diff
@@ -0,0 +1,67 @@
+import { bitLength, randBits } from 'bigint-crypto-utils';
+import { bigIntFromBufferBE } from '../../util';
+import { gcd, modInv, modPow } from 'bigint-mod-arith';
+
+// Security parameters.
+const k = 128;
+const alpha = 2;
```
islamaminBitGo, line 7: According to https://eprint.iacr.org/2018/057.pdf#page6 section 5, we can pick $\alpha$ to be a value like 8009 and get down the number of values m in the proof to be 10 and still achieve statistical soundness of $2^{-128}$. There is a cost of added checks of divisibility of N by primes less than $\alpha$, but the paper has done a cost benefit analysis of changing alpha that we might benefit from considering.

islamaminBitGo (commented):
The math of this checks out as outlined by the paper, I left an optimization comment to be considered.

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/paillierproof.ts
islamaminBitGo: Note they have found the optimal $\alpha$ to be $319567$ which yields an $m = 7$ in our case.

zahin-mohammad:
final PR for this is here https://github.com/BitGo/BitGoJS/pull/3602

PR #3505, opened 2023-04-25
feat(sdk-core): add enterprise method to fetch ecdsa config

pranavjain97:
Ticket: BG-72153

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -1896,7 +1896,7 @@ export class BitGoAPI implements BitGoBase {
     return _.merge({}, defaultConstants(this.getEnv()), BitGoAPI._constants[this.getEnv()]);
   }
 
-  async getBitgoChallengesForEcdsaSigning(walletId: string): Promise<any> {
+  async getChallengesForEcdsaSigning(walletId: string): Promise<any> {
     const urlPath = `/wallet/${walletId}/challenges`;
     return await this.get(this.url(urlPath, 2)).query({}).result();
   }
--- a/modules/sdk-core/src/bitgo/enterprise/enterprise.ts
+++ b/modules/sdk-core/src/bitgo/enterprise/enterprise.ts
@@ -148,4 +148,23 @@ export class Enterprise implements IEnterprise {
       challenge
     );
   }
+
+  /**
+   * Fetches the existing TSS ECDSA enterprise challenge if one exists.
+   * Can be used with uploadAndEnableTssEcdsaSigning to re-sign the
+   * enterprise challenge with new signatures.
+   */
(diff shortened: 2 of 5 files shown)


modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -1896,7 +1896,7 @@ export class BitGoAPI implements BitGoBase {
     return _.merge({}, defaultConstants(this.getEnv()), BitGoAPI._constants[this.getEnv()]);
   }
 
-  async getBitgoChallengesForEcdsaSigning(walletId: string): Promise<any> {
+  async getChallengesForEcdsaSigning(walletId: string): Promise<any> {
```
pranavjain97, line 1899: The name was a bit misleading before, since it fetches both enterprise and bitgo challenges.

modules/sdk-core/src/bitgo/enterprise/enterprise.ts
```diff
@@ -148,4 +148,23 @@ export class Enterprise implements IEnterprise {
       challenge
     );
   }
+
+  /**
+   * Fetches the existing TSS ECDSA enterprise challenge if one exists.
+   * Can be used with uploadAndEnableTssEcdsaSigning to re-sign the
+   * enterprise challenge with new signatures.
+   */
+  async getExistingTssEcdsaChallenge(): Promise<NTilde> {
```
zahin-mohammad, line 157:
```suggestion
  async getTssEcdsaChallenge(): Promise<NTilde> {
```

`existing` seems redundant here.

modules/sdk-core/src/bitgo/enterprise/enterprise.ts
```diff
@@ -148,4 +148,23 @@ export class Enterprise implements IEnterprise {
...
+    const urlPath = `/api/v2/enterprise/${this.id}/tssconfig`;
+    const tssConfig = await this.bitgo.get(this.bitgo.url(urlPath, 2)).send().result();
```
zahin-mohammad, line 159: Can we reuse bitgoapi's `getChallengesForEcdsaSigning` to do this request?

pranavjain97: No, thats a completely diff endpoint

modules/sdk-core/src/bitgo/enterprise/enterprise.ts
pranavjain97: It's just to clarify it doesn't create a new one if it doesnt exist

zahin-mohammad (approved):

modules/sdk-core/src/bitgo/enterprise/enterprise.ts
pranavjain97: Given its only gonna be used for enterprise, i think its fine as is

PR #3506, opened 2023-04-25
fix incorrect check verifying range proof

zhongxishen:
BG-74332

### DIFF
--- a/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/rangeproof.ts
+++ b/modules/sdk-core/src/account-lib/mpc/tss/ecdsa/rangeproof.ts
@@ -258,7 +258,7 @@ export function verify(
   const modulusBytes = Math.floor((modulusBits + 7) / 8);
   const q = curve.order();
   const q3 = q ** BigInt(3);
-  if (proof.s1 == q3) {
+  if (proof.s1 > q3) {
     return false;
   }
   const hash = createHash('sha256');
--- a/modules/sdk-core/test/unit/rangeproof.ts
+++ b/modules/sdk-core/test/unit/rangeproof.ts
@@ -1,0 +1,106 @@
+import * as sinon from 'sinon';
+import * as paillierBigint from 'paillier-bigint';
+import Ecdsa from '../../src/account-lib/mpc/tss/ecdsa/ecdsa';
+import * as rangeProof from '../../src/account-lib/mpc/tss/ecdsa/rangeproof';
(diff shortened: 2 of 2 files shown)


sachushaji (commented):
Can you add a unit test for this fix which would fail if this fix was not made

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/rangeproof.ts
```diff
@@ -258,7 +258,8 @@ export function verify(
   const modulusBytes = Math.floor((modulusBits + 7) / 8);
   const q = curve.order();
   const q3 = q ** BigInt(3);
-  if (proof.s1 == q3) {
+  // if s1 not <= q3 return false
```
zahin-mohammad, line 261: not sure we need this comment, the if statement is pretty clear

zahin-mohammad (commented):
+1, on the test comment. Is this something we can easily test/reproduce?

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/rangeproof.ts
zhongxishen: removed

modules/sdk-core/test/unit/rangeproof.ts
```diff
@@ -0,0 +1,106 @@
...
+
+describe('MtA range proof', function () {
+  let switchPrime = false;
+  let safePrimeMock: sinon.SinonStub;
+
+  let paillierKeyPair: paillierBigint.KeyPair;
+  let ntilde: NTilde;
+
+  before('set up paillier and ntile', async function () {
+    safePrimeMock = sinon.stub(OpenSSL.prototype, 'generateSafePrime').callsFake(async (bitlength: number) => {
```
zhongxishen, line 16: Stubbing `rangeProof.generateSafePrimes()` is more straightforward, but it doesn't work with node 14 for some reason.

sachushaji (approved):

modules/sdk-core/test/unit/rangeproof.ts
zahin-mohammad: `generateSafePrime` doesn't work on node 14 unit tests. Are you sure ur stubs are being called then?

zahin-mohammad (approved):

islamaminBitGo (approved):

PR #3513, opened 2023-04-27
feat(sdk-coin-atom): add recover mechanism for ATOM

Vijay-Jagannathan:
Ticket: [BG-69146
](https://bitgoinc.atlassian.net/browse/BG-69146)


modules/sdk-core/src/bitgo/environments.ts
```diff
@@ -95,6 +96,7 @@ const mainnetBase: EnvironmentTemplate = {
   nearNodeUrls: ['https://rpc.mainnet.near.org'],
   solNodeUrl: 'https://api.mainnet-beta.solana.com',
   adaNodeUrl: 'https://api.koios.rest/api/v0',
+  atomNodeUrl: 'https://rest-cosmoshub.ecostake.com',
```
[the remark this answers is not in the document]
mukeshsp: is this working ?  did you try to hit using the postman ?

modules/sdk-core/src/bitgo/environments.ts
```diff
@@ -124,6 +126,7 @@ const testnetBase: EnvironmentTemplate = {
   nearNodeUrls: ['https://rpc.testnet.near.org'],
   solNodeUrl: 'https://api.devnet.solana.com',
   adaNodeUrl: 'https://preprod.koios.rest/api/v0',
+  atomNodeUrl: 'https://rpc.sentry-02.theta-testnet.polypore.xyz/',
```
mukeshsp, line 129:
```suggestion
  atomNodeUrl: 'https://rest.sentry-01.theta-testnet.polypore.xyz',
```

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -226,6 +256,221 @@ export class Atom extends BaseCoin {
...
+    try {
+      return await request.get(fullEndpoint).send();
+    } catch (e) {
+      console.debug(e);
+    }
+    throw new Error(`Unable to call endpoint ${getBalancePath + senderAddress} from node: ${nodeUrl}`);
+  }
+
+  /**
+   * Helper to fetch balance
```
sachushaji, line 401: Seems this function is fetching chainID?

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -226,6 +256,221 @@ export class Atom extends BaseCoin {
...
+  protected async getBalanceFromNode(senderAddress: string): Promise<request.Response> {
+    const nodeUrl = this.getPublicNodeUrl();
+    const getBalancePath = '/cosmos/bank/v1beta1/balances/';
+    const fullEndpoint = nodeUrl + getBalancePath + senderAddress;
+    try {
+      return await request.get(fullEndpoint).send();
+    } catch (e) {
+      console.debug(e);
+    }
+    throw new Error(`Unable to call endpoint ${getBalancePath + senderAddress} from node: ${nodeUrl}`);
```
sachushaji, line 397: Will this response directly give you the balance, if not, Try to get the balance and return only the balance so that the logic is abstracted within this function

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -226,6 +256,221 @@ export class Atom extends BaseCoin {
...
+   */
+  protected async getChainIdFromNode(): Promise<request.Response> {
+    const nodeUrl = this.getPublicNodeUrl();
+    const getLatestBlockPath = 'cosmos/base/tendermint/v1beta1/blocks/latest';
+    const fullEndpoint = nodeUrl + getLatestBlockPath;
+    try {
+      return await request.get(fullEndpoint).send();
+    } catch (e) {
+      console.debug(e);
+    }
```
sachushaji, line 422: Better to return the chainId parsed out of the response

mukeshsp (changes requested):

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -226,6 +256,221 @@ export class Atom extends BaseCoin {
...
+      if (Number(balance) <= 0) {
+        continue;
+      }
+
+      const accountNumber = await this.getAccountNumber(senderAddress);
+      const chainId = await this.getChainId();
+      const amount: Coin[] = [
+        {
+          denom: 'uatom',
+          amount: balance,
```
mukeshsp, line 317: subtract with require fee amount

modules/sdk-core/src/bitgo/environments.ts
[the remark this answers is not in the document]
Vijay-Jagannathan: Afaik no, thats why I need to double check the correct REST endpoint (took this of the docs). Testnet is working as expected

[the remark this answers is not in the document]
Vijay-Jagannathan: Verified with the correct mainnet endpoint, will change to the correct one

modules/sdk-coin-atom/src/atom.ts
Vijay-Jagannathan: This is actually done one step above in the chain. So the flow goes like this:

* Recover function calls `getAccountBalance(address)` and expects a balance string as response.
* getAccountBalance(address) calls another helper function called `getBalanceFromNode` to actually make the call to the full node. Once the helper responds, `getAccountBalance(address)` checks if the response is a 200 or otherwise; if it is a 200 indeed, then it parses the response here since the helper's job is to return a response alone
* The helper meanwhile constructs the REST API call based of the parameters that were provided and simply makes the call and gives back the response

This flow applies for all calls we make to the full node to fetch the different parameters. We could probably do this in one single call instead but this way, we can abstract whats needed/not needed

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -226,6 +257,223 @@ export class Atom extends BaseCoin {
...
+    throw new Error('Did not find an address with funds to recover');
+  }
+
+  /**
+   * Get balance from public node
+   */
+  protected async getBalanceFromNode(senderAddress: string): Promise<request.Response> {
+    const nodeUrl = this.getPublicNodeUrl();
+    const getBalancePath = '/cosmos/bank/v1beta1/balances/';
+    const fullEndpoint = nodeUrl + getBalancePath + senderAddress;
```
mukeshsp, line 394: I think, fullnode URL ending with forward slash and getBalancePath is starting with forward slash, will it create an issue ?

Vijay-Jagannathan: good catch, yeah that will be a problem; will edit

Vijay-Jagannathan:
> **unit-test-all (14.x)**: failing due to `Error while running start function: RuntimeError: JsValue(TypeError: wasm function signature contains illegal type TypeError: wasm function signature`. This is due to node 14 and not observed in node 16 **unit-test-all (16.x)**: The job was canceled because "_14_x" failed.
> 
> Is it possible to cancel the 14.x test altogether or is it necessary ?

Based on zahin's comment, implemented a workaround for this; which is to mock/stub the `rangeProof.generateNtilde` method. Not only does this help with the node 14 issue, we can also speed up the test since this method typically takes a long time.

modules/sdk-coin-atom/test/unit/atom.ts
```diff
@@ -336,4 +341,51 @@ describe('ATOM', function () {
...
+        bitgoKey: wrwUser.bitgoKey,
+        walletPassphrase: wrwUser.walletPassphrase,
+        recoveryDestination: destinationAddress,
+      });
+      res.should.not.be.empty();
+      res.should.hasOwnProperty('serializedTx');
+      res.should.hasOwnProperty('scanIndex');
+      sandBox.assert.calledOnce(basecoin.getAccountBalance);
+      sandBox.assert.calledOnce(basecoin.getAccountNumber);
+      sandBox.assert.calledOnce(basecoin.getChainId);
```
sachushaji, line 381: I think can you add a test case where you have n number of receive address with some funds and you get back an array of recovery transactions build for addresses which have funds

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -226,6 +257,223 @@ export class Atom extends BaseCoin {
...
+    const MPC = new Ecdsa();
+
+    // Step 6: For each account from the starting index till numIterations + starting index, fetch the build tx details
+    for (let i = startIdx; i < numIteration + startIdx; i++) {
+      const currPath = `m/${i}`;
+      publicKey = MPC.deriveUnhardened(bitgoKey, currPath).slice(0, 66);
+      const senderAddress = this.getAddressFromPublicKey(publicKey);
+      const balance = await this.getAccountBalance(senderAddress);
+
+      if (Number(balance) <= 0) {
```
Phani024, line 309: Will the `balance` always be within the number range?

Phani024 (commented):
Is the unsigned sweep not supported? Or is there a separate ticket for it?

zahin-mohammad:
> **unit-test-all (14.x)**: failing due to `Error while running start function: RuntimeError: JsValue(TypeError: wasm function signature contains illegal type TypeError: wasm function signature`. This is due to node 14 and not observed in node 16 **unit-test-all (16.x)**: The job was canceled because "_14_x" failed.
> 
> Is it possible to cancel the 14.x test altogether or is it necessary ?

Please mock the challenge values.

zahin-mohammad (changes requested):

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -226,6 +257,223 @@ export class Atom extends BaseCoin {
...
+
+  /**
+   * Helper to fetch account balance
+   */
+  protected async getAccountBalance(senderAddress: string): Promise<string> {
+    const response = await this.getBalanceFromNode(senderAddress);
+    if (response.status !== 200) {
+      throw new Error('Account not found');
+    }
+    return response.body.balances[0].amount;
```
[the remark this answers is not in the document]
Vijay-Jagannathan: > Is the unsigned sweep not supported? Or is there a separate ticket for it?

Not at the moment, AFAIK

> Is `balances` always expected to have records?

Yes, if the response is a 200, then balance is >= 0

modules/sdk-coin-atom/src/atom.ts
Vijay-Jagannathan: Changed to `BigNumber` to avoid problems

modules/sdk-coin-atom/test/unit/atom.ts
```diff
@@ -336,4 +341,51 @@ describe('ATOM', function () {
...
+
+      const chainId = sandBox.stub(Atom.prototype, 'getChainId' as keyof Atom);
+      chainId.withArgs().resolves(testChainId);
+
+      const serializedEntChallenge = mockChallengeA;
+      const deserializedEntChallenge = Ecdsa.deserializeNtilde(serializedEntChallenge);
+      sinon.stub(rangeProof, 'generateNtilde').resolves(deserializedEntChallenge);
+    });
+
+    it('should recover a txn for non-bitgo recoveries', async function () {
```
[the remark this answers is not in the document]
Vijay-Jagannathan: Added a few

modules/sdk-coin-atom/.mocharc.yml
```diff
@@ -1,5 +1,5 @@
 require: 'ts-node/register'
-timeout: '60000'
+timeout: '180000'
```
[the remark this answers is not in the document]
Vijay-Jagannathan: Initially the local test suite was timing out around challenge generation (which took somewhere around 2 - 3 mins); To avoid that, I had bumped the limit but now that we're mocking the challenge generation part, we wouldn't have to increase this limit. Made the change to bring it back to default

mukeshsp (dismissed):
Nice work !! Thanks

mukeshsp (approved):

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -309,4 +538,146 @@ export class Atom extends BaseCoin {
     }
     return memoIdNumber.gte(0);
   }
+
+  private getKeyCombinedFromTssKeyShares(
+    userPublicOrPrivateKeyShare: string,
+    backupPrivateOrPublicKeyShare: string,
+    walletPassphrase?: string
+  ): [ECDSAMethodTypes.KeyCombined, ECDSAMethodTypes.KeyCombined] {
+    let backupPrv;
```
zahin-mohammad, line 547: Note to self: we need to refactor this with eth tss to reduce code duplication.

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -226,6 +255,206 @@ export class Atom extends BaseCoin {
...
+   * (maybe) sign the transaction
+   *
+   * @returns {AtomTx} the serialized transaction hex string and index
+   * of the address being swept
+   */
+  async recover(params: RecoveryOptions): Promise<AtomTx> {
+    // Step 1: Check if params contains the required parameters
+    if (!params.bitgoKey) {
+      throw new Error('missing bitgoKey');
+    }
```
zahin-mohammad, line 270: Can we please use the user key instead? If we just need the common key chain, the user key should be sufficient. We do not want to couple bitgo's key in a user <> backup recovery.

modules/sdk-coin-atom/src/atom.ts
```diff
@@ -226,6 +255,206 @@ export class Atom extends BaseCoin {
...
+      throw new Error('missing userKey');
+    }
+
+    if (!params.backupKey) {
+      throw new Error('missing backupKey');
+    }
+
+    if (!params.walletPassphrase) {
+      throw new Error('missing wallet passphrase');
+    }
```
zahin-mohammad, line 286: A lot of this validation can be simplified via an io-ts codec.

zahin-mohammad (approved):
non blocking comments, but please address in a future PR.

modules/sdk-coin-atom/src/atom.ts
Vijay-Jagannathan: We'll address this in an followup

PR #3524, opened 2023-04-30
feat(utxo-lib): add test util function to create psbt

saravanan7mani:
A generic Create PSBT function as a test util.

Ticket: BG-74843

### DIFF
--- a/modules/utxo-lib/src/testutil/psbt.ts
+++ b/modules/utxo-lib/src/testutil/psbt.ts
@@ -1,0 +1,188 @@
+import * as assert from 'assert';
+import {
+  createOutputScriptP2shP2pk,
+  ScriptType,
+  ScriptType2Of3,
+  scriptTypeP2shP2pk,
+  scriptTypes2Of3,
+} from '../bitgo/outputScripts';
+import {
+  addReplayProtectionUnspentToPsbt,
+  addWalletOutputToPsbt,
+  addWalletUnspentToPsbt,
+  createPsbtForNetwork,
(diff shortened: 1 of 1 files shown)


modules/utxo-lib/src/testutil/psbt.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+}
+
+/**
+ * returns signer and cosigner names for InputScriptType.
+ * user and undefined as signer and cosigner respectively for p2shP2pk.
+ * user and backup as signer and cosigner respectively for p2trMusig2.
+ * user and bitgo as signer and cosigner respectively for other input script types.
+ */
+export function getSigners(inputType: InputScriptType): { signerName: KeyName; cosignerName?: KeyName } {
```
brandonblack, line 90: To be able to test MuSig2 signing, shouldn't this return `bitgo` for all types other than `p2shP2pk`?

brandonblack (changes requested):

modules/utxo-lib/src/testutil/psbt.ts
saravanan7mani: Yes. Thats how it works. Only `p2trMusig2` will return `backup`for cosigner since we assume it as `script path` (for simplicity). In hsm-api test suite, we will skip `p2trMusig2`.  Script type `taprootKepPathSpend` is `p2trMusig2 key path`.

brandonblack (approved):

PR #3527, opened 2023-05-01
fix(sdk-core): fix time issue when checking gpg wallet signatures fro…

islamaminBitGo:
…m bitgo
Passing in null for date skips date checks all-together, which is fine for this use case as the gpg keys used in creating wallets are all ephemeral, hence the concern of bypassing a check on an expired key is of no relevance as the keys are short-lived.

### DIFF
--- a/modules/sdk-core/src/bitgo/tss/common.ts
+++ b/modules/sdk-core/src/bitgo/tss/common.ts
@@ -1,7 +1,7 @@
 import assert from 'assert';
 
 import { BitGoBase } from '../bitgoBase';
-import { RequestType, TxRequest } from '../utils';
+import { RequestType, TxRequest, verifyPrimaryUserWrapper } from '../utils';
 import { SignatureShareRecord } from '../utils/tss/baseTypes';
 import openpgp from 'openpgp';
 import { SerializedNtilde } from '../../account-lib/mpc/tss/ecdsa/types';
@@ -95,7 +95,7 @@ export async function commonVerifyWalletSignature(params: {
     throw new Error('Invalid HSM GPG signature');
   }
 
-  const verificationResult = await walletSignature.verifyPrimaryUser([bitgoPub]);
+  const verificationResult = await verifyPrimaryUserWrapper(walletSignature, bitgoPub, false);
   const isValid = verificationResult.some((result) => result.valid);
   if (!isValid) {
     throw new Error('Invalid HSM GPG signature');
--- a/modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
@@ -20,6 +20,11 @@ import crypto from 'crypto';
 
 const sodium = require('libsodium-wrappers-sumo');
 
+export type KeyValidityDict = {
+  keyID: pgp.KeyID;
+  valid: boolean | null;
+}[];
+
 /**
  * Fetches BitGo's pubic gpg key used in MPC flows
  * @param {BitGo} bitgo BitGo object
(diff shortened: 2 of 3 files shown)


modules/sdk-core/src/bitgo/tss/common.ts
```diff
@@ -95,7 +95,7 @@ export async function commonVerifyWalletSignature(params: {
     throw new Error('Invalid HSM GPG signature');
   }
 
-  const verificationResult = await walletSignature.verifyPrimaryUser([bitgoPub]);
+  const verificationResult = await walletSignature.verifyPrimaryUser([bitgoPub], null as unknown as undefined);
```
zahin-mohammad, line 98:
```suggestion
  const verificationResult = await walletSignature.verifyPrimaryUser([bitgoPub], undefined);
```

confused on why we can't just pass in undefined. Also why this fixes the issue.

or alternatively change the types for `verifyPrimaryUser` to accept null as input?

zahin-mohammad, line 98: From our convo on slack: 

There are many places we call `verifyPrimaryUser`. To make it explicit I think we should create some wrapper methods in our SDK like `verifyPrimaryUserWIthDateCheck` and `verifyPrimaryUserWithoutDateCheck`. 

The behaviour of the existing usages would remain the same via `verifyPrimaryUserWIthDateCheck` but it will be more clear to readers of the code what is happening in case this same bug re-appears in other places.

zahin-mohammad (changes requested):

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -35,6 +40,26 @@ export async function getBitgoGpgPubKey(bitgo: BitGoBase): Promise<Key> {
   return await readKey({ armoredKey: bitgoPublicKeyStr });
 }
 
+/**
+ * Verifies the primary user on a GPG key using a reference key representing the user to be checked.
+ * Allows a verification without a date check by wrapping verifyPrimaryUser of openpgp.
+ * @param {Key} pubKey gpg key to check the primary user of.
+ * @param {Key} primaryUser gpg key of the user to check.
+ * @param {boolean} checkDates If false, disable date checks in the openpgp call to check the primary user.
+ * @return {Key} public gpg key
```
islamaminBitGo, line 49: oops wrong return doc, will fix that

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -35,6 +40,26 @@ export async function getBitgoGpgPubKey(bitgo: BitGoBase): Promise<Key> {
...
 
+/**
+ * Verifies the primary user on a GPG key using a reference key representing the user to be checked.
+ * Allows a verification without a date check by wrapping verifyPrimaryUser of openpgp.
+ * @param {Key} pubKey gpg key to check the primary user of.
+ * @param {Key} primaryUser gpg key of the user to check.
+ * @param {boolean} checkDates If false, disable date checks in the openpgp call to check the primary user.
+ * @return {KeyValidityDict} list of users checked and whether each passed as a primary user in pubKey or not.
+ */
+export async function verifyPrimaryUserWrapper(
```
zahin-mohammad, line 51: Nit, the naming can be improved. `Wrapper` does not actually tell callers what this is doing or why its being wrapped. 

`verifyPrimaryUserWithDateCheck` might be better.

modules/sdk-core/src/bitgo/tss/common.ts
```diff
@@ -95,7 +95,7 @@ export async function commonVerifyWalletSignature(params: {
     throw new Error('Invalid HSM GPG signature');
   }
 
-  const verificationResult = await walletSignature.verifyPrimaryUser([bitgoPub]);
+  const verificationResult = await verifyPrimaryUserWrapper(walletSignature, bitgoPub, false);
```
zahin-mohammad, line 98: Nit: comment on why we dont check the date here.

zahin-mohammad (approved):

Logicwax (approved):

PR #3541, opened 2023-05-05
feat(utxo-lib): add musig2 nonce at input index

saravanan7mani:
BREAKING CHANGE

For tests, adding nonce at given input index is useful.

Ticket: BG-76925

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -992,52 +992,80 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
     return { tapOutputKey, participantPubKey, pubNonce: Buffer.from(pubNonce) };
   }
 
-  private setMusig2NoncesInner(keyPair: BIP32Interface, keyType: 'root' | 'derived', sessionId?: Buffer): this {
+  private setMusig2NoncesInner(
+    keyPair: BIP32Interface,
+    keyType: 'root' | 'derived',
+    sessionId?: Buffer,
+    inputIndex?: number
+  ): this {
     if (keyPair.isNeutered()) {
       throw new Error('private key is required to generate nonce');
     }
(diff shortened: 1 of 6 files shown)


davidkaplanbitgo (changes requested):
I think if you're going to make these changes you should make them go by the same code structure as `validateSignaturesOfAllInputs` and `validateSignatureOfInput`, etc. --> maybe something like`setMusig2NoncesOfInput` and `setMusig2NoncesOfAllInputs`

brandonblack (approved):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -992,52 +992,80 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
     return { tapOutputKey, participantPubKey, pubNonce: Buffer.from(pubNonce) };
   }
 
-  private setMusig2NoncesInner(keyPair: BIP32Interface, keyType: 'root' | 'derived', sessionId?: Buffer): this {
+  private setMusig2NoncesInner(
+    keyPair: BIP32Interface,
+    keyType: 'root' | 'derived',
+    sessionId?: Buffer,
+    inputIndex?: number
```
OttoAllmendinger, line 999: if you change this to `inputs = this.data.inputs` and pass `[this.data.inputs[inputIndex]]` at the callsite, it is a bit safer because the `checkForInput()` calls are there

OttoAllmendinger (approved):
nit

davidkaplanbitgo (approved):

davidkaplanbitgo:
Merging in because @saravanan7mani  is on vacation and my PR depends on these changes

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
saravanan7mani: Agreed. It's merged already. We will do it the recommended change later. I guess, for now it's safer, since it is private method.

PR #3542, opened 2023-05-05
fix(utxo-lib): witness utxo should not be added for non-segwit PSBT i…

saravanan7mani:
…nput

As per BIP-174, witnessUtxo should not be added for non-segwit inputs. So removing it.

Ticket: BG-76926


davidkaplanbitgo (changes requested):
what inputs would go through `addWalletUnspentToPsbt` that dont have a witness?

saravanan7mani:
> what inputs would go through `addWalletUnspentToPsbt` that dont have a witness?

p2sh

brandonblack (approved):

OttoAllmendinger (approved):

davidkaplanbitgo (approved):

PR #3543, opened 2023-05-05
feat(sdk-core): stop calling staking service send api for full version

dpkjnr:
Ticket: BG-62904

For ECDSA TSS signing, TxRequest version has to be full. Before Atom, Eth was the only coin with ecdsa TSS and it was handled separately for eth tss to use api version full(i.e. final send API call is not required). With this change for all ecdsa tss coin full api version will be used.


OttoAllmendinger (commented):
flush - insufficient context

LeoKotschenreuther (approved):

PR #3544, opened 2023-05-05
feat(sdk-core): add more whitelisted build params

davidkaplanbitgo:
The extra build params added:

```
[
  'addressType',          'apiVersion',
  'consolidateAddresses', 'cpfpFeeRate',
  'cpfpTxIds',            'gasLimit',
  'hopParams',            'idfSignedTimestamp',
  'idfUserId',            'idfVersion',
  'maxFee',               'nonce',
  'preview',              'receiveAddress',
  'reservation',          'strategy',
  'nonParticipation',     'validFromBlock',
  'validToBlock',         'messageKey',
  'eip1559',              'keyregTxBase64',
  'closeRemainderTo',     'tokenName',
  'enableTokens',         'emergency'
]
```

The problem is the cpfp parameters not being passed through, but there is also not a real reason to not use the full list

BG-76974

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -2053,37 +2053,7 @@ export class Wallet implements IWallet {
       return this.sendManyTss(params);
     }
 
-    const selectParams = _.pick(params, [
-      'recipients',
-      'numBlocks',
-      'feeRate',
-      'maxFeeRate',
-      'minConfirms',
-      'enforceMinConfirmsForChange',
-      'targetWalletUnspents',
-      'message',
-      'minValue',
-      'maxValue',
-      'sequenceId',
-      'lastLedgerSequence',
-      'ledgerSequenceDelta',
-      'gasPrice',
-      'noSplitChange',
-      'unspents',
-      'comment',
-      'otp',
-      'changeAddress',
-      'instant',
-      'memo',
-      'type',
-      'trustlines',
-      'transferId',
-      'stakingOptions',
-      'hop',
-      'type',
-      'sourceChain',
-      'destinationChain',
-    ]);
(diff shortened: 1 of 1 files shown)


brandonblack (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2053,37 +2053,7 @@ export class Wallet implements IWallet {
       return this.sendManyTss(params);
     }
 
-    const selectParams = _.pick(params, [
```
mmcshinsky-bitgo, line 2056: `prebuildWhitelistedParams` is missing `comment`, `otp`, and `hop`.

mmcshinsky-bitgo (changes requested):

CapnMigraine (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -168,6 +168,9 @@ export class Wallet implements IWallet {
       'consolidateAddresses',
       'cpfpFeeRate',
       'cpfpTxIds',
+      'comment',
```
brandonblack, line 171: Do these belong on the build params?

brandonblack (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
davidkaplanbitgo: moved to within

mmcshinsky-bitgo: They were part of `selectParams`, so whatever the solution, they need to be included.

davidkaplanbitgo: they are

mmcshinsky-bitgo (approved):

PR #3550, opened 2023-05-08
feat(wp): migrate deterministic signing into utxopsbt

davidkaplanbitgo:
BREAKING CHANGE

Move deterministic signing functionality within UtxoPsbt
and out of testutils.

Set deterministic signing and nonce generation as an optional flag that defaults to non-deterministic signing for the cosigner. 

BG-74624

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -35,6 +35,8 @@ import {
   musig2PartialSigVerify,
   musig2AggregateSigs,
   getSigHashTypeFromSigs,
+  musig2DeterministicSign,
+  createMusig2DeterministicNonce,
 } from './Musig2';
 import { isTuple, Tuple } from './types';
 
@@ -45,6 +47,25 @@ export enum ProprietaryKeySubtype {
   MUSIG2_PARTICIPANT_PUB_KEYS = 0x01,
   MUSIG2_PUB_NONCE = 0x02,
   MUSIG2_PARTIAL_SIG = 0x03,
+}
+
+type SignatureParams = {
+  /** When true, and add the second (last) nonce and signature for a taproot key
+   * path spend deterministically. Throws an error if done for the first nonce/signature
+   * of a taproot keypath spend. Ignore for all other input types.
+   */
+  deterministic: boolean;
+  /** Allowed sighash types */
+  sighashTypes: number[];
+};
+
+const defaultSignatureParams = {
+  deterministic: false,
+  sighashTypes: [Transaction.SIGHASH_DEFAULT, Transaction.SIGHASH_ALL],
+};
+
+function toSignatureParams(v: Partial<SignatureParams> | number[]): SignatureParams {
+  return Array.isArray(v) ? toSignatureParams({ sighashTypes: v }) : { ...defaultSignatureParams, ...v };
 }
 
 export interface HDTaprootSigner extends HDSigner {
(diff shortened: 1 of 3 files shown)


modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -608,7 +632,7 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
   signTaprootInputHD(
     inputIndex: number,
     hdKeyPair: HDTaprootSigner | HDTaprootMusig2Signer,
-    sighashTypes: number[] = [Transaction.SIGHASH_DEFAULT, Transaction.SIGHASH_ALL]
+    { sighashTypes = [Transaction.SIGHASH_DEFAULT, Transaction.SIGHASH_ALL], deterministic = false } = {}
```
OttoAllmendinger, line 635: this is the way

OttoAllmendinger (changes requested):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -576,27 +578,49 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
     });
   }
 
+  private parseSignInputsParams(
```
OttoAllmendinger, line 581: this is static and would benefit from a toplevel type

```
type SignatureParams = { deterministic?: boolean; sighashTypes: number[] }


const defaultSignatureParams = {
  deterministic: false,
  sighashTypes: [Transaction.SIGHASH_DEFAULT, Transaction.SIGHASH_ALL]
}

function toSignatureParams(v: SignatureParams | number[]): SignatureParams {
  return Array.isArray(v) ? toSignatureParams({ sighashTypes: v }) : { ...defaultSignatureParams, ...v };
}

```

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -576,27 +578,49 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
...
+    params: { sighashTypes?: number[]; deterministic?: boolean } | number[] = {
+      sighashTypes: [Transaction.SIGHASH_DEFAULT, Transaction.SIGHASH_ALL],
+      deterministic: false,
+    }
```
OttoAllmendinger, line 585:
```suggestion
    params?: number[] | SignatureParams = defaultSignatureParams
```

OttoAllmendinger (changes requested):

OttoAllmendinger (changes requested):

modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -47,6 +49,17 @@ export enum ProprietaryKeySubtype {
   MUSIG2_PARTIAL_SIG = 0x03,
 }
 
+type SignatureParams = { deterministic: boolean; sighashTypes: number[] };
```
OttoAllmendinger, line 52:
```suggestion
type SignatureParams = {
  /** When true and adding the second (last) signature, use deterministic nonce generation method.
     *  When true and not adding last signature, throw an error
     */ 
  deterministic: boolean;
  /** Allowed sighash types */
  sighashTypes: number[]
};
```

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #3560, opened 2023-05-12
chore(sdk-lib-mpc): proxy rangeProof (sdk-core) methods to EcdsaRangeProof (sdk-lib-mpc)

zahin-mohammad:
## Description

One pr of many of migrating functions to sdk-lib-mpc.

## Issue Number

BG-77573

# Checklist:


modules/sdk-core/src/account-lib/mpc/tss/ecdsa/ecdsa.ts
islamaminBitGo, line 1: rangeProof is still being used to call generateNtilde, can we replace that with EcdsaRangeProof and remove the ./rangeProof import from this file?

zahin-mohammad: Good catch.

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/rangeproof.ts
```diff
@@ -2,65 +2,24 @@
...
-async function generateModulus(bitlength: number): Promise<RSAModulus> {
-  if (bitlength < 3072) {
-    // https://www.keylength.com/en/6/
-    console.warn('Generating a modulus with less than 3072 is not recommended!');
-  }
-  const bitlengthP = Math.floor(bitlength / 2);
-  const bitlengthQ = bitlength - bitlengthP;
-  const [p, q] = await generateSafePrimes([bitlengthP, bitlengthQ]);
-  const n = p * q;
-  // We never expect this to happen unless something went wrong with the wasm/openssl module
```
rlutzbitgo, line 43: I still think this comment is misleading at best (if it was copied into the new location for the generateModulus function). If we want to throw instead of just auto-retrying, I think that's fine, but you can't guarantee that multiplying two 1536 bit numbers together will result in a 3072 bit number. 

1111 * 1000 = 1111000 (7 bits)
1111 * 1111 = 11100001 (8 bits)

rlutzbitgo (dismissed):
Looks good to me. I restarted your node 16 tests because one of them timed out (probably worth considering bumping some of these timeouts given how the variance is with things like safe prime generation)

rlutzbitgo:
Given that it failed twice in a row, should probably just go ahead and bump the timeouts for sdk-coin-atom:
```
1) ATOM
@bitgo/sdk-coin-atom:        Recover transactions: success path
@bitgo/sdk-coin-atom:          should recover funds for non-bitgo recoveries:
@bitgo/sdk-coin-atom:      Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

islamaminBitGo (dismissed):

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/rangeproof.ts
islamaminBitGo: You bring up a good point, I think we should retry if it doesnt hit 3072 instead of throwing. Our maximum needed bits will always be bitlengthP + bitlengthQ so we can keep retrying until the modulus hits that.

zahin-mohammad:
> > should recover funds for non-bitgo recoveries:
> 
> 🙄 I think someone forgot to mock challenge generation, let me TAL

That person turned out to be me. Now that things are in the new library, the stubs need to be updated.

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/rangeproof.ts
zahin-mohammad: Agreed, but lets make the retry count an optional input. We can add a default value if wed like.

islamaminBitGo (dismissed):

zahin-mohammad:
Please merge once this gets all the needed approvals.

modules/sdk-core/src/account-lib/mpc/curves/secp256k1.ts
```diff
@@ -4,6 +4,9 @@ import * as secp from '@noble/secp256k1';
 
 const order = secp.CURVE.n;
 
+/**
+ * @deprecated use Secp256k1Curve from sdk-lib-mpc instead
+ */
 export class Secp256k1Curve implements BaseCurve {
```
sachushaji, line 10: Why can't we also export the Secp256k1Curve from sdk-lib-mpc along side the deprecation notice?, since making changes in the Secp256k1Curve should also be reflected here until we remove it from here

modules/sdk-core/src/account-lib/mpc/tss/ecdsa/rangeproof.ts
```diff
@@ -2,65 +2,24 @@
...
-    return openSSL.generateSafePrime(bitlength);
-  });
-  return await Promise.all(promises);
-}
-
-/**
- * @deprecated Use generateModulus from sdk-lib-mpc instead
- * @param bitlength
- */
-async function generateModulus(bitlength: number): Promise<RSAModulus> {
```
sachushaji, line 34: Are we not adding a generateModulus function exported from sdk-lib-mpc here?

modules/sdk-core/src/account-lib/mpc/util.ts
```diff
@@ -31,6 +31,10 @@ export function bigIntFromBufferBE(buf: Buffer): bigint {
   return BigInt('0x' + buf.toString('hex'));
 }
 
+/**
+ * @deprecated Use bigIntFromU8ABE from sdk-lib-mpc instead
+ * @param buf
+ */
 export function bigIntFromU8ABE(buf: Uint8Array): bigint {
```
sachushaji, line 38: Same, lets use it from sdk-lib-mpc and export here along side the deprecation notice so that any changes made to this function in the sdk-lib-mpc would also be reflected here

modules/sdk-lib-mpc/src/tss/index.ts
```diff
@@ -1 +1 @@
-export * as Ecdsa from './ecdsa';
+export * from './ecdsa';
```
sachushaji, line 1: Are we only going to use this library for ecdsa, Named export can help to distinguish between similar types/ functions between Ecdsa and Eddsa, without it can be confusing

sachushaji: We can also squash the commits

modules/sdk-core/src/account-lib/mpc/curves/secp256k1.ts
zahin-mohammad: Good idea, didn't think about re-exporting from the deprecated file.

zahin-mohammad: Fixed.

modules/sdk-core/src/account-lib/mpc/util.ts
zahin-mohammad: Fixed.

islamaminBitGo (approved):

islamaminBitGo (dismissed):

islamaminBitGo (approved):

sachushaji (approved):
LGTM, Thanks @zahin-mohammad

CapnMigraine (approved):

PR #3561, opened 2023-05-12
fix(utxo-lib): calculate signature count for psbt and tx

saravanan7mani:
To identify signature status, it willl be useful.

Ticket: BG-77624

### DIFF
--- a/modules/utxo-lib/src/testutil/transaction.ts
+++ b/modules/utxo-lib/src/testutil/transaction.ts
@@ -1,0 +1,183 @@
+import * as assert from 'assert';
+
+import { ScriptType, ScriptType2Of3, scriptTypeP2shP2pk } from '../bitgo/outputScripts';
+import {
+  getExternalChainCode,
+  isWalletUnspent,
+  KeyName,
+  getInternalChainCode,
+  RootWalletKeys,
+  Unspent,
+  UtxoTransactionBuilder,
+  createTransactionBuilderForNetwork,
+  addToTransactionBuilder,
+  getWalletAddress,
+  signInputP2shP2pk,
+  signInputWithUnspent,
+  WalletUnspentSigner,
+} from '../bitgo';
+import { Network } from '../networks';
+import { mockReplayProtectionUnspent, mockWalletUnspent } from './mock';
+
+/**
+ * input script type and value.
+ */
+export type TxnInputScriptType = Exclude<ScriptType, 'p2trMusig2'>;
+export type TxnOutputScriptType = Exclude<ScriptType2Of3, 'p2trMusig2'>;
+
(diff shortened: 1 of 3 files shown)


modules/utxo-lib/test/bitgo/psbt/Psbt.ts
```diff
@@ -40,6 +49,48 @@ function getScriptTypes2Of3() {
...
+    const outputs = txnOutputScriptTypes.map((scriptType) => ({ scriptType, value: BigInt(900) }));
+
+    (['unsigned', 'halfsigned', 'fullsigned'] as const).forEach((sign, signatureCount) => {
+      const txb = constructTxnBuilder(inputs, outputs, network, rootWalletKeys, sign);
+      const tx = sign === 'fullsigned' ? txb.build() : txb.buildIncomplete();
+      assert.strictEqual(getSignatureCount(tx), signatureCount);
+      tx.ins.forEach((input, inputIndex) => {
+        assert.strictEqual(
+          getSignatureCount([tx.ins[inputIndex]], 0),
+          inputIndex === 4 && signatureCount > 0 ? 1 : signatureCount
```
brandonblack, line 64: This weirdness is because the taprootMusig2 (or is this a p2sh-pk?) only has 1 signature when fullsigned?

modules/utxo-lib/src/testutil/transaction.ts
```diff
@@ -0,0 +1,183 @@
...
+): void {
+  inputs.forEach((input, index) => {
+    signTxnInput(txb, input, index, rootWalletKeys, sign);
+  });
+}
+
+/**
+ * construct transaction for given inputs, outputs, network and root wallet keys.
+ */
+export function constructTxnBuilder<TNumber extends number | bigint>(
```
brandonblack, line 141: Don't we already have these functions somewhere else? Feels like we're duplicating test code we already have elsewhere, but I don't know where.

modules/utxo-lib/test/bitgo/psbt/Psbt.ts
saravanan7mani: p2sh-pk. Comment added.

modules/utxo-lib/src/testutil/transaction.ts
saravanan7mani: Yes, but they are not flexible and generic to fit to all use cases. Thats why we have introduced these 2 main code testutil functions - `constructPsbt` and `constructTxnBuilder` and other utility functions. In follow up PRs, we should gradually replace other functions duplicate functions with these 2.

davidkaplanbitgo (approved):

PR #3562, opened 2023-05-15
feat: add psbt and musig2 support to sdk-api

saravanan7mani:
Adding sdk-api support for PSBT format and taproot keypath spend inputs for bitcoin transactions.

Added utility functions within utxo-lib as well.

BG-66947

### DIFF
--- a/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
+++ b/modules/utxo-lib/src/bitgo/UtxoPsbt.ts
@@ -38,7 +38,7 @@ import {
   musig2DeterministicSign,
   createMusig2DeterministicNonce,
 } from './Musig2';
-import { isTuple, Tuple } from './types';
+import { isTuple, Triple, Tuple } from './types';
 import { getTaprootOutputKey } from '../taproot';
 
 export const PSBT_PROPRIETARY_IDENTIFIER = 'BITGO';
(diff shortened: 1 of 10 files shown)


modules/utxo-lib/src/bitgo/UtxoPsbt.ts
```diff
@@ -652,7 +665,7 @@ export class UtxoPsbt<Tx extends UtxoTransaction<bigint> = UtxoTransaction<bigin
       return [false, false, false];
     }
     const input = checkForInput(this.data.inputs, inputIndex);
-    return this.data.globalMap.globalXpub.map((xpub) => {
+    const result = this.data.globalMap.globalXpub.map((xpub) => {
```
brandonblack, line 668: `return [...] as Triple<boolean>;`

modules/utxo-lib/src/bitgo/wallet/Psbt.ts
```diff
@@ -445,3 +445,10 @@ export function getSignatureCount(
     .map((index, _) => getInputSignatureCount(constructParam(tx, index)))
     .reduce((prev, curr) => (curr > prev ? curr : prev), 0);
 }
+
+/** @return true iff input starts with magic PSBT byte sequence */
+export function isPsbt(data: Buffer): boolean {
+  // https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#specification
+  // 0x70736274 - ASCII for 'psbt'. 0xff - separator
+  return 5 <= data.length && data.readUInt32BE(0) === 0x70736274 && data.readUInt8(4) === 0xff;
```
brandonblack, line 453:
```suggestion
  if (typeof data === 'string') {
    if (data.length < 10) return false;
    data = Buffer.from(data.slice(0, 10), 'hex');
  }
  return 5 <= data.length && data.readUInt32BE(0) === 0x70736274 && data.readUInt8(4) === 0xff;
```

this, or something similar - just so we're not converting whole PSBTs (potentially over a megabyte) just to check the first 5 bytes.

brandonblack (changes requested):

PR #3563, opened 2023-05-15
chore(sdk-core): Refactor & split getVerifyAndSignBitGoChallenges

mohammadalfaiyazbitgo:
- Add method to sign bitgo challenge without verification. To be used in contexts where the caller will verify challenges/proofs separately from signing.


zahin-mohammad (dismissed):
lgtm. Existing function was split up, so the existing tests still cover these changes.

package.json
```diff
@@ -69,7 +69,7 @@
     "parse-path": "^5.0.0",
     "parse-url": "8.1.0",
     "terser": "^5.14.2",
-    "vm2": "^3.9.17",
```
pranavjain97, line 72: was this intentional?

mmcshinsky-bitgo: Security audit update.

zahin-mohammad (dismissed):

package.json
zahin-mohammad: https://bitgo.slack.com/archives/C010AEXLLCR/p1684247091685949?thread_ts=1684246605.689579&cid=C010AEXLLCR

zahin-mohammad: @mohammadalfaiyazbitgo shouldn't there be a package-lock change as well with this? Did you run `yarn`?

mohammadalfaiyazbitgo: yeah forgot that, I'll update

zahin-mohammad (approved):

PR #3567, opened 2023-05-16
feat(sdk-coin-xrp): add backup signing for xrp

alia-bitgo:
Currently, the sdk only has support for signing once with the sdk. But for the case of recovery transactions for smc wallets, signing may be need to be done twice with both the client's user and backup key. The appropriate changes were made to add support for this.

TICKET: BG-77113

### DIFF
--- a/modules/sdk-coin-xrp/src/ripple.ts
+++ b/modules/sdk-coin-xrp/src/ripple.ts
@@ -40,9 +40,6 @@ const signWithPrivateKey = function (txHex, privateKey, options) {
       throw new Error('txHex needs to be either hex or JSON string for XRP');
     }
   }
-  if (tx.TxnSignature || tx.Signers) {
-    throw new Error('transaction must not contain "TxnSignature" or "Signers" properties');
-  }
 
   tx.SigningPubKey = options && options.signAs ? '' : publicKey;
 
@@ -56,7 +53,11 @@ const signWithPrivateKey = function (txHex, privateKey, options) {
       SigningPubKey: publicKey,
       TxnSignature: computeSignature(tx, privateKey, options.signAs),
     };
-    tx.Signers = [{ Signer: signer }];
+    if (tx.TxnSignature || tx.Signers) {
+      tx.Signers.unshift({ Signer: signer });
+    } else {
+      tx.Signers = [{ Signer: signer }];
+    }
   } else {
     tx.TxnSignature = computeSignature(tx, privateKey, undefined);
   }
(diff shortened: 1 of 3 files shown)


lcovar (approved):

modules/sdk-coin-xrp/src/ripple.ts
```diff
@@ -56,7 +53,11 @@ const signWithPrivateKey = function (txHex, privateKey, options) {
       SigningPubKey: publicKey,
       TxnSignature: computeSignature(tx, privateKey, options.signAs),
     };
-    tx.Signers = [{ Signer: signer }];
+    if (tx.TxnSignature || tx.Signers) {
+      tx.Signers.unshift({ Signer: signer });
+    } else {
+      tx.Signers = [{ Signer: signer }];
+    }
```
zahin-mohammad, line 60: Some comments on ordering between sub-seqeuent function calls would be nice. (when do we expect the if case, when do we expect the else case)

modules/sdk-coin-xrp/src/xrp.ts
```diff
@@ -289,23 +304,22 @@ export class Xrp extends BaseCoin {
    * @param params
    */
   async explainTransaction(params: ExplainTransactionOptions = {}): Promise<TransactionExplanation> {
-    if (!params.txHex) {
+    let transaction;
```
zahin-mohammad, line 307: Is there a type we can add here?

modules/sdk-coin-xrp/src/xrp.ts
```diff
@@ -318,7 +332,7 @@ export class Xrp extends BaseCoin {
         fee: {
           fee: transaction.Fee,
           feeRate: null,
-          size: txHex.length / 2,
+          size: txHex!.length / 2,
```
zahin-mohammad, line 335: Is there an issues with floats or integer division here?

```suggestion
size: txHex.length / 2,
```

modules/sdk-coin-xrp/src/xrp.ts
```diff
@@ -255,10 +263,17 @@ export class Xrp extends BaseCoin {
...
+    const tx = rippleLib.signWithPrivateKey(txPrebuild.txHex, userPrivateKey.toString('hex'), {
       signAs: userAddress,
     });
-    return { halfSigned: { txHex: halfSigned.signedTransaction } };
+
+    // Normally the SDK provides the first signature for an XRP tx, but occasionally it provides the second and final one
+    // for recovery transactions
+    if (isLastSignature) {
+      return { txHex: tx.signedTransaction };
+    }
```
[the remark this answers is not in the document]
alia-bitgo: it's a parameter that's already always passed from ovc, so i thought may as well just use it for this case.

PR #3571, opened 2023-05-17
Rel/latest

mmcshinsky-bitgo:


brandonblack (dismissed):

sachushaji (commented):
Unit test seems to be failing

modules/abstract-utxo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bitgo/abstract-utxo",
-  "version": "2.1.2",
+  "version": "2.2.1",
```
LeoKotschenreuther, line 3: If we are going from version 2.1.2 and adding features (but no breaking changes), shouldn't the new version be 2.2.0?

LeoKotschenreuther (dismissed):
I got confused about the versioning but this is not blocking.
Did we do a few releases since last changing our package.json files that could explain us skipping a few versions?

aarvay (commented):
flushing from queue. required build step failing.

modules/abstract-utxo/package.json
mmcshinsky-bitgo: I'd have to look at all the commits. This PR has a couple stable releases in it, so 2.2.0 was probably farther back in the commit history

zahin-mohammad (dismissed):

brandonblack (dismissed):

sachushaji (dismissed):

mukeshsp (dismissed):

zahin-mohammad (approved):

margueriteblair (approved):

rlutzbitgo (approved):

brandonblack (approved):

PR #3575, opened 2023-05-19
chore(root): remove 14.x from unit-test-all

zahin-mohammad:
## Description

One of our Solana dependencies make use of `||=` which is not supported by node 14. As a result the `unit-test-all` will consistently fail for node `14.x`. Node 14 is also end-of-life, as such this test is not adding any value.

## Issue Number

TICKET: BG-78094

## Type of change
- [x] CI

### DIFF
--- a/.github/workflows/test_package_updates.yml
+++ b/.github/workflows/test_package_updates.yml
@@ -25,7 +25,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [16.x]
 
     steps:
       - uses: actions/checkout@v3


alebusse:
@zahin-mohammad Should you also update the .nvmrc to a v16?

zahin-mohammad:
> @zahin-mohammad Should you also update the .nvmrc to a v16?

I think we should still keep that nvmrc as is until BGMS upgrades. We still need node 14, but 14.x in this CI job adds no value.

alebusse (approved):

alebusse:
What about this one https://github.com/BitGo/BitGoJS/blob/master/.github/workflows/ci.yml#L19?

zahin-mohammad:
> What about this one [`master`/.github/workflows/ci.yml#L19](https://github.com/BitGo/BitGoJS/blob/master/.github/workflows/ci.yml?rgh-link-date=2023-05-19T20%3A17%3A38Z#L19)?

That for some reason is fine. My guess is that its because this specific step runs all package tests, regardless of it changed or not, whereas the other one doesn't actually build the SOL one?

PR #3581, opened 2023-05-24
fix(sdk-lib-mpc): add path to tsconfig.packages

ajays97:


zahin-mohammad (dismissed):

tsconfig.packages.json
```diff
@@ -154,6 +154,9 @@
     },
     {
       "path": "./modules/web-demo"
+    },
+    {
+      "path": "./modules/sdk-lib-mpc"
     }
```
zahin-mohammad, line 160: nit: would be nice to keep this list sorted alphabetically.

zahin-mohammad (approved):

tsconfig.packages.json
ajays97: sorted

zahin-mohammad (approved):

PR #3587, opened 2023-05-25
chore: release beta from latest version when publishing first time

ajays97:
Ticket: BG-77950

### DIFF
--- a/scripts/prepare-release.ts
+++ b/scripts/prepare-release.ts
@@ -29,9 +29,9 @@ const getLernaModules = async (): Promise<void> => {
   const { stdout: lernaBinary } = await execa('yarn', ['bin', 'lerna'], { cwd: process.cwd() });
 
   const lerna = getLernaRunner(lernaBinary);
-  const modules: Array<{name: string, location: string}> = JSON.parse(await lerna('list', ['--loglevel', 'silent', '--json', '--all']));
+  const modules: Array<{ name: string, location: string }> = JSON.parse(await lerna('list', ['--loglevel', 'silent', '--json', '--all']));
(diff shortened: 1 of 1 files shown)


scripts/prepare-release.ts
```diff
@@ -119,36 +119,35 @@ const incrementVersions = async (preid: string = 'beta') => {
...
-          JSON.stringify(json, null, 2) + '\n'
-        );
-        // since we're manually setting new versions, we must also reconcile all other lerna packages to now use the 'next' version for this module
-        lernaModuleLocations.forEach((otherModulePath) => {
-          // skip it for the current version
-          if (otherModulePath === modulePath) {
-            return;
+
+      const next = inc(tags[preid] || tags['latest'], 'prerelease', undefined, preid);
+      console.log(`Setting next version for ${json.name} to ${next}`);
```
sachushaji, line 124: I can see that now some packages like web-demo, sdk-test etc which seems not published also are going through this logic flow. Probably the check was introduced to gate these internal packages probably. Lets wait for sdk teams feedback

LeoKotschenreuther (commented):
Automatic comment to flush this PR from my review queue. 🚽🚽🚽
Please re-request another review from the respective team(s) when you are ready for reviews.

scripts/prepare-release.ts
zahin-mohammad: Web-demo and express should be ignored. They where deleted via github actions before this script is run.

mmcshinsky-bitgo: Lerna should ignore those set the aren't set to public.

mmcshinsky-bitgo (approved):

PR #3590, opened 2023-05-26
feat(sdk-lib-mpc): add proof of paillier correctness & refactor appendChallenge (1/4)

zahin-mohammad:
Note: this is a refactor of #3502

# Please be aware of the following when making your pull request:

## Description

- introduces the primitives for pallier proofs from the PR listed above
- refactors appendChallenge to be simpler

## Issue Number

TICKET: BG-78338

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My code compiles correctly for both Node and Browser environments
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My commits follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) and I have properly described any BREAKING CHANGES
- [x] The ticket or github issue was included in the commit message as a reference
- [x] I have made corresponding changes to the documentation and on any new/updated functions and/or methods - [jsdoc](https://jsdoc.app/)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #3593, opened 2023-05-26
feat(sdk-lib-mpc): pallier proof plumbing (deprecated) (2/2)

zahin-mohammad:
Follow up to #3590 

TICKET: BG-38338

summary:
- makes range proof challenge mandatory for `appendChallenge`


zahin-mohammad:
moved to https://github.com/BitGo/BitGoJS/pull/3598

PR #3607, opened 2023-05-31
fix(sdk-core): fix pending approval for consolidation

davidkaplanbitgo:
When doing an unspent consolidation, there are no recipients in the `buildParams` because they are generated within WP. This causes errors reconstructing the consolidation transaction during pending approval. This new check sets the `canRecreateTransaction=false` if there are no recipients. Instead, it will use the txHex of the transaction passed into `/send` and use that. 

This should unblock customers who have policies on BTC wallets that want to consolidate.

BG-78492

### DIFF
--- a/modules/bitgo/test/v2/unit/pendingApproval.ts
+++ b/modules/bitgo/test/v2/unit/pendingApproval.ts
@@ -89,6 +89,51 @@ describe('Pending Approvals:', () => {
     scope.done();
   });
 
+  it('should no recreate the transaction during approving a pending approval if there are no recipients', async () => {
+    const pendingApprovalDataTemp: PendingApprovalData = {
+      id: 'pa0',
+      info: {
+        type: Type.TRANSACTION_REQUEST,
+        transactionRequest: {
+          coinSpecific: {
+            [coin]: { txHex: 'gabagool' },
+          },
+          recipients: [],
+          buildParams: {},
+          sourceWallet: walletId,
+        },
+      },
+      state: State.PENDING,
+      creator: 'test',
+    };
+
+    const walletDataTemp = {
+      id: walletId,
+      coin,
+      pendingApprovals: [pendingApprovalDataTemp],
+    };
+    const walletTemp = new Wallet(bitgo, basecoin, walletDataTemp);
+    (pendingApprovalDataTemp as any).wallet = walletTemp;
+
+    const pendingApprovals = walletTemp.pendingApprovals();
+    pendingApprovals.should.have.length(1);
+    const pendingApproval = pendingApprovals[0];
+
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
```diff
@@ -178,6 +178,11 @@ export class PendingApproval implements IPendingApproval {
       canRecreateTransaction = false;
     }
 
+    // If there are no recipients, then the transaction cannot be recreated
+    if (!!this.info()?.transactionRequest?.buildParams?.recipients) {
+      canRecreateTransaction = false;
+    }
```
zahin-mohammad, line 184: Can we add a test for this?

zahin-mohammad (changes requested):

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
```diff
@@ -178,6 +178,11 @@ export class PendingApproval implements IPendingApproval {
       canRecreateTransaction = false;
     }
 
+    // If there are no recipients, then the transaction cannot be recreated
+    if (!!this.info()?.transactionRequest?.buildParams?.recipients) {
```
brandonblack, line 182: I think this should be `if (!this.info()?.transactionRequest?.buildParams?.recipients?.length && this.info()?.transactionRequest?.buildParams?.type !== 'consolidate') {`

Or possibly nicer:
```
const recipients = this.info()?.transactionRequest?.buildParams?.recipients || [];
const type = this.info()?.transactionRequest?.buildParams?.type;
if (recipients.length === 0 && type !== 'consolidate') {
```

brandonblack (changes requested):

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
davidkaplanbitgo: Added in latest push

brandonblack (approved):

angshal (approved):

modules/bitgo/test/v2/unit/pendingApproval.ts
```diff
@@ -89,6 +89,51 @@ describe('Pending Approvals:', () => {
     scope.done();
   });
 
+  it('should no recreate the transaction during approving a pending approval if there are no recipients', async () => {
```
zahin-mohammad, line 92:
```suggestion
  it('should not recreate the transaction during approving a pending approval if there are no recipients', async () => {
```

zahin-mohammad (approved):

PR #3614, opened 2023-06-02
feat(statics): add cosmos side chains config

mukeshsp:
Ticket: BG-78915

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -42,6 +42,7 @@ export enum CoinFamily {
   ETC = 'etc',
   EOS = 'eos',
   FIAT = 'fiat',
+  HASH = 'hash', // Provenance
   HBAR = 'hbar',
   LTC = 'ltc',
   POLYGON = 'polygon',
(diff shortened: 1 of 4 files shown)


modules/statics/src/networks.ts
```diff
@@ -745,18 +796,21 @@ export const Networks = {
...
     fiat: Object.freeze(new Fiat()),
+    hash: Object.freeze(new Hash()),
     hedera: Object.freeze(new Hedera()),
     litecoin: Object.freeze(new Litecoin()),
     polygon: Object.freeze(new Polygon()),
     ofc: Object.freeze(new Ofc()),
     osmo: Object.freeze(new Osmo()),
     rbtc: Object.freeze(new Rbtc()),
     stellar: Object.freeze(new Stellar()),
+    sei: Object.freeze(new SEI()),
```
DinshawKothari, line 807: Why SEI and not Sei() here ??

modules/statics/src/base.ts
```diff
@@ -1258,6 +1266,10 @@ export enum BaseUnit {
   OFC = 'ofcCoin',
   OSMO = 'uosmo',
   FIAT = 'fiatCoin',
+  TIA = 'utia',
+  HASH = 'nHASH',
```
DinshawKothari, line 1270: make it nhash

modules/statics/src/networks.ts
mukeshsp: updated !!

modules/statics/src/base.ts
mukeshsp: updated !

DinshawKothari (approved):

PR #3617, opened 2023-06-05
fix(sdk-core): add previewPendingApprovals flag

gianchandania:
This PR adds previewPendingApprovals flag in the approve method. We are modifying buildTransaction API for account based coins in such a manner that we check the pending approvals before building a new tx. During approving the pending approvals, we were getting a false error that the wallet has insufficient funds because of pending transaction, because we rebuild the transaction during approval of hot wallet transactions and we were checking the pending approval transactions during buildTx. Suppose if a user has 0.1 gteth and they have generated a transaction of amount 0.1 gteth which is currently in pending approval. Now when the admin tries to approve it, buildTx is called and it checks if there are any pending approvals for the wallet. It then finds the same pending approval that we are trying to approve and will throw an error. 

Before creating any pending approval we are checking if the wallet has enough funds or not. So it is not needed to check it again at the time of approval process. Hence we are adding the flag, so that we can skip checking the pending approvals at the time of approval process

[EA-154
](https://bitgoinc.atlassian.net/browse/EA-154)

### DIFF
--- a/modules/sdk-core/src/bitgo/pendingApproval/iPendingApproval.ts
+++ b/modules/sdk-core/src/bitgo/pendingApproval/iPendingApproval.ts
@@ -30,6 +30,7 @@ export interface ApproveOptions {
   otp?: string;
   tx?: string;
   xprv?: string;
+  previewPendingApprovals?: boolean;
 }
 
 export interface PendingApprovalInfo {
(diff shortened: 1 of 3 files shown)


aarvay (commented):
I don't see the usage of this flag in the WP endpoint. Is that to be implemented ?

gianchandania:
> I don't see the usage of this flag in the WP endpoint. Is that to be implemented ?

Yes, I am working on that now

aarvay (approved):

PR #3618, opened 2023-06-05
feat(sdk-coin-trx): add receive address tron recovery

sachushaji:
this changes adds support for tron receiveaddress recovery. Tron receive addressesare single sig addresses derived from userkey. The change refactors the recover functionsuch that it will scan receive addresses untilsome funds is found to flush and generate transaction and throw error otherwise.

Ticket: EA-124

### DIFF
--- a/modules/sdk-coin-trx/src/trx.ts
+++ b/modules/sdk-coin-trx/src/trx.ts
@@ -30,6 +30,7 @@ import {
 import { Interface, Utils, WrappedBuilder } from './lib';
 import { getBuilder } from './lib/builder';
 import { TransactionReceipt } from './lib/iface';
+import { isInteger, isUndefined } from 'lodash';
 
 export const MINIMUM_TRON_MSIG_TRANSACTION_FEE = 1e6;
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-trx/src/trx.ts
```diff
@@ -538,20 +548,88 @@ export class Trx extends BaseCoin {
       throw new Error('Invalid destination address!');
     }
 
+    let startIdx = params.startingScanIndex;
+    if (isUndefined(startIdx)) {
+      startIdx = 1;
+    } else if (!isInteger(startIdx) || startIdx < 0) {
```
mukeshsp, line 554: is it possible to have startIdx = 0 ?

mukeshsp (approved):

modules/sdk-coin-trx/src/trx.ts
sachushaji: Receive address is seems to be starting with index 1. For base address we are already checking it seperately

PR #3619, opened 2023-06-05
add injective statics

mukeshsp:

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -45,6 +45,7 @@ export enum CoinFamily {
   FIAT = 'fiat',
   HASH = 'hash', // Provenance
   HBAR = 'hbar',
+  INJ = 'inj', // Injective
   LTC = 'ltc',
   POLYGON = 'polygon',
   NEAR = 'near',
@@ -1270,6 +1271,7 @@ export enum BaseUnit {
   HASH = 'nhash',
   BLD = 'ubld',
   SEI = 'usei',
+  INJ = 'inj',
 }
 
 export interface BaseCoinConstructorOptions {
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -45,7 +45,7 @@ export enum CoinFamily {
   FIAT = 'fiat',
   HASH = 'hash', // Provenance
   HBAR = 'hbar',
-  INJ = 'inj', // Injective
+  INJ = 'injective',
   LTC = 'ltc',
   POLYGON = 'polygon',
   NEAR = 'near',
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -819,7 +819,7 @@ export const coins = CoinMap.fromCoins([
   ),
   account(
     '5f9506c5-f10a-43c2-92d3-52941083bbc7',
-    'inj',
+    'injective',
     'Injective',
     Networks.main.inj,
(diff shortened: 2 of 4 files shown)


modules/statics/src/base.ts
```diff
@@ -45,6 +45,7 @@ export enum CoinFamily {
   FIAT = 'fiat',
   HASH = 'hash', // Provenance
   HBAR = 'hbar',
+  INJ = 'Injective',
```
gianchandania, line 48:
```suggestion
  INJ = 'injective',
```

modules/statics/src/coins.ts
```diff
@@ -816,6 +817,26 @@ export const coins = CoinMap.fromCoins([
     BaseUnit.SEI,
     SEI_FEATURES
   ),
+  account(
+    '5f9506c5-f10a-43c2-92d3-52941083bbc7',
+    'Injective',
+    'injective',
```
gianchandania, line 823:
```suggestion
    'Injective',
```

modules/statics/src/coins.ts
```diff
@@ -816,6 +817,26 @@ export const coins = CoinMap.fromCoins([
     BaseUnit.SEI,
     SEI_FEATURES
   ),
+  account(
+    '5f9506c5-f10a-43c2-92d3-52941083bbc7',
+    'Injective',
```
gianchandania, line 822:
```suggestion
    'injective',
```

gianchandania: This seems to be the ticker and the below one is name of the coin

modules/statics/src/networks.ts
```diff
@@ -633,6 +633,18 @@ class SeiTestnet extends Testnet implements AccountNetwork {
   explorerUrl = 'https://sei.explorers.guru/transaction/';
 }
 
+class INJ extends Mainnet implements AccountNetwork {
```
gianchandania, line 636:
```suggestion
class Injective extends Mainnet implements AccountNetwork {
```

Let's maintain the convention followed for other coins. I think the convention is that whatever is the name of the network, is the class name as well

modules/statics/src/networks.ts
```diff
@@ -633,6 +633,18 @@ class SeiTestnet extends Testnet implements AccountNetwork {
...
+  name = 'Injective';
+  family = CoinFamily.INJ;
+  explorerUrl = 'https://explorer.injective.network/transaction/';
+}
+
+class InjTestnet extends Testnet implements AccountNetwork {
+  name = 'Testnet Injective';
```
gianchandania, line 643:
```suggestion
  name = 'InjectiveTestnet'; 
```
I think this should be class name as well

modules/statics/src/coins.ts
```diff
@@ -816,6 +817,26 @@ export const coins = CoinMap.fromCoins([
...
+    'injective',
+    Networks.main.inj,
+    18,
+    UnderlyingAsset.INJ,
+    BaseUnit.INJ,
+    INJ_FEATURES
+  ),
+  account(
+    '6ae81d6a-011c-499c-a3c8-15ac7dcac48a',
+    'tInjective',
```
gianchandania, line 832:
```suggestion
    'tinjective',
```

gianchandania (changes requested):

gianchandania (approved):

PR #3631, opened 2023-06-07
feat(sdk-coin-tia): celestia sdk along with testcases

DinshawKothari:
Ticket: BG-78116

Context:
This PR contains complete implementation of celestia (tia) sdk.

All **66 testcases are passing**. It covers the below things -
- keyPair validations
- address validations
- txn validations
- block validations
- txn building and signing testcases
- staking, unstaking, withdrawReward testcases

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -136,6 +136,8 @@ COPY --from=builder /tmp/bitgo/modules/sdk-coin-osmo /var/modules/sdk-coin-osmo/
 RUN cd /var/modules/sdk-coin-osmo && yarn link
 COPY --from=builder /tmp/bitgo/modules/abstract-cosmos /var/modules/abstract-cosmos/
 RUN cd /var/modules/abstract-cosmos && yarn link
+COPY --from=builder /tmp/bitgo/modules/sdk-coin-tia /var/modules/sdk-coin-tia/
+RUN cd /var/modules/sdk-coin-tia && yarn link
 COPY --from=builder /tmp/bitgo/modules/sdk-coin-xlm /var/modules/sdk-coin-xlm/
 RUN cd /var/modules/sdk-coin-xlm && yarn link
 COPY --from=builder /tmp/bitgo/modules/sdk-coin-xrp /var/modules/sdk-coin-xrp/
@@ -195,15 +197,16 @@ RUN cd /var/bitgo-express && \
     yarn link @bitgo/sdk-coin-ltc && \
     yarn link @bitgo/sdk-coin-osmo && \
     yarn link @bitgo/abstract-cosmos && \
+    yarn link @bitgo/sdk-coin-tia && \
     yarn link @bitgo/sdk-coin-xlm && \
     yarn link @bitgo/sdk-coin-xrp && \
     yarn link @bitgo/sdk-coin-zec
 #LINK_END
 
 #LABEL_START
-LABEL created="Mon, 05 Jun 2023 16:57:03 GMT"
+LABEL created="Tue, 06 Jun 2023 10:54:43 GMT"
 LABEL version=9.27.0
-LABEL git_hash=9efc360c50ab1809e58cb5b0024850f14d62943a
+LABEL git_hash=82d33792e150f3ad67bda641c6c12f5e4d7322e0
 #LABEL_END
 
 USER node
--- a/modules/account-lib/package.json
+++ b/modules/account-lib/package.json
@@ -46,6 +46,7 @@
     "@bitgo/sdk-coin-sol": "^2.5.1",
     "@bitgo/sdk-coin-stx": "^2.0.7",
     "@bitgo/sdk-coin-sui": "^3.0.7",
+    "@bitgo/sdk-coin-tia": "^1.0.0",
     "@bitgo/sdk-coin-trx": "^1.8.0",
     "@bitgo/sdk-coin-xtz": "^1.4.23",
     "@bitgo/sdk-core": "^8.7.0",
--- a/modules/sdk-coin-tia/package.json
+++ b/modules/sdk-coin-tia/package.json
@@ -1,0 +1,56 @@
+{
+  "name": "@bitgo/sdk-coin-tia",
+  "version": "1.0.0",
+  "description": "BitGo SDK coin library for Celestia",
+  "main": "./dist/src/index.js",
+  "types": "./dist/src/index.d.ts",
+  "scripts": {
+    "build": "yarn tsc --build --incremental --verbose .",
+    "fmt": "prettier --write .",
+    "check-fmt": "prettier --check .",
+    "clean": "rm -r ./dist",
+    "lint": "eslint --quiet .",
+    "prepare": "npm run build",
+    "test": "npm run coverage",
+    "coverage": "nyc -- npm run unit-test",
+    "unit-test": "mocha"
+  },
+  "author": "BitGo SDK Team <sdkteam@bitgo.com>",
+  "license": "MIT",
+  "engines": {
+    "node": ">=14 <17"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/BitGo/BitGoJS.git",
+    "directory": "modules/sdk-coin-tia"
+  },
+  "lint-staged": {
+    "*.{js,ts}": [
+      "yarn prettier --write",
+      "yarn eslint --fix"
+    ]
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "nyc": {
+    "extension": [
+      ".ts"
+    ]
+  },
+  "dependencies": {
+    "@bitgo/abstract-cosmos": "^1.1.0",
+    "@bitgo/sdk-core": "^8.7.0",
+    "@bitgo/statics": "^16.1.0",
+    "@cosmjs/amino": "^0.29.5",
+    "@cosmjs/encoding": "^0.29.5",
+    "@cosmjs/stargate": "^0.29.5",
+    "bignumber.js": "^9.1.1"
+  },
+  "devDependencies": {
+    "@bitgo/sdk-api": "^1.10.5",
+    "@bitgo/sdk-test": "^1.2.23",
+    "@types/lodash": "^4.14.183"
+  }
+}
--- a/modules/sdk-coin-tia/src/lib/keyPair.ts
+++ b/modules/sdk-coin-tia/src/lib/keyPair.ts
@@ -1,0 +1,25 @@
+import { KeyPairOptions } from '@bitgo/sdk-core';
+import { pubkeyToAddress } from '@cosmjs/amino';
+
+import { CosmosKeyPair } from '@bitgo/abstract-cosmos';
+
+/**
+ * Celestia keys and address management.
+ */
+export class KeyPair extends CosmosKeyPair {
+  constructor(source?: KeyPairOptions) {
+    super(source);
+  }
+
+  /** @inheritdoc */
+  getAddress(): string {
+    const base64String = Buffer.from(this.getKeys().pub.slice(0, 66), 'hex').toString('base64');
+    return pubkeyToAddress(
+      {
+        type: 'tendermint/PubKeySecp256k1',
+        value: base64String,
+      },
+      'celestia'
+    );
+  }
+}
--- a/modules/sdk-coin-tia/src/lib/transactionBuilderFactory.ts
+++ b/modules/sdk-coin-tia/src/lib/transactionBuilderFactory.ts
@@ -1,0 +1,78 @@
+import {
+  CosmosTransaction,
+  CosmosTransactionBuilder,
+  CosmosTransferBuilder,
+  StakingActivateBuilder,
+  StakingDeactivateBuilder,
+  StakingWithdrawRewardsBuilder,
+} from '@bitgo/abstract-cosmos';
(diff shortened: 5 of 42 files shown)


modules/sdk-coin-tia/src/lib/keyPair.ts
```diff
@@ -0,0 +1,25 @@
...
+    const base64String = Buffer.from(this.getKeys().pub.slice(0, 66), 'hex').toString('base64');
+    return pubkeyToAddress(
+      {
+        type: 'tendermint/PubKeySecp256k1',
+        value: base64String,
+      },
+      'celestia'
+    );
+  }
+}
```
Vijay-Jagannathan, line 25: Will this method be similar for the other chains as well, apart from the prefix ? 

If yes, could we put this method in a common class and pass in the prefix instead ?

modules/sdk-coin-tia/src/lib/transaction.ts
```diff
@@ -0,0 +1,33 @@
...
+      this.addSignature(Buffer.from(this.cosmosLikeTransaction.signature).toString('hex'));
+    }
+    const typeUrl = this.cosmosLikeTransaction.sendMessages[0].typeUrl;
+    const transactionType = tiaUtils.getTransactionTypeFromTypeUrl(typeUrl);
+
+    if (transactionType === undefined) {
+      throw new Error('Transaction type is not supported ' + typeUrl);
+    }
+    this.transactionType = transactionType;
+  }
```
Vijay-Jagannathan, line 32: Seems pretty similar to [ATOM](url). If this is the case for osmosis/other chains, I'd say we'll need this in the common class instead (since we're anyway utilising `cosmosLikeTransaction`, which is a super type)

modules/sdk-coin-tia/src/lib/transactionBuilder.ts
```diff
@@ -0,0 +1,153 @@
...
+    utils.validateGasBudget(gasBudget);
+    this._gasBudget = gasBudget;
+    return this;
+  }
+
+  /**
+   * Initialize the transaction builder fields using the decoded transaction data
+   * @param {TiaTransaction} tx the transaction data
+   */
+  initBuilder(tx: TiaTransaction): void {
```
Vijay-Jagannathan, line 57: seems like a method which can be abstracted into the parent builder class. We can pass in the `tx`(of type TiaTransaction) to the parent builderclass, which takes in any `CosmosTransaction` and initializes the builder

modules/sdk-coin-tia/src/lib/transactionBuilder.ts
```diff
@@ -0,0 +1,153 @@
...
+  /** @inheritdoc */
+  protected fromImplementation(rawTransaction: string): CosmosTransaction {
+    const tx = new TiaTransaction(this._coinConfig);
+    tx.enrichTransactionDetailsFromRawTransaction(rawTransaction);
+    this.initBuilder(tx);
+    return this.transaction;
+  }
+
+  /** @inheritdoc */
+  protected async buildImplementation(): Promise<CosmosTransaction> {
```
Vijay-Jagannathan, line 85: same thoughts as above, seems like we can abstract this better

modules/sdk-coin-tia/src/lib/transactionBuilder.ts
DinshawKothari: There are few methods that uses validateAmount() and isValidAddress() methods which are coin specific. And hence I had to keep them in their sdks, rest all methods are abstracted into abstract-cosmos sdk.

modules/sdk-coin-tia/src/lib/transactionBuilderFactory.ts
```diff
@@ -0,0 +1,75 @@
+import { BaseTransactionBuilderFactory, InvalidTransactionError, TransactionType } from '@bitgo/sdk-core';
+import { TiaTransactionBuilder } from './transactionBuilder';
+import { TiaTransferBuilder } from './transferBuilder';
+import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import { TiaTransaction } from './transaction';
+import { StakingActivateBuilder } from './StakingActivateBuilder';
+import { StakingDeactivateBuilder } from './StakingDeactivateBuilder';
+import { StakingWithdrawRewardsBuilder } from './StakingWithdrawRewardsBuilder';
+
+export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
```
Vijay-Jagannathan, line 10: This entire class should extend from `CosmosTransactionBuilderFactory`; we should have one if we don't already

modules/sdk-coin-tia/src/lib/utils.ts
```diff
@@ -0,0 +1,237 @@
...
+   * Validates if the address matches with regex @see accountAddressRegex
+   *
+   * @param {string} address
+   * @returns {boolean} - the validation result
+   */
+  isValidValidatorAddress(address: string): boolean {
+    return constants.validatorAddressRegex.test(address);
+  }
+
+  validateWithdrawRewardsMessage(withdrawRewardsMessage: WithdrawDelegatorRewardsMessage) {
```
Vijay-Jagannathan, line 41: for this and all the methods in this class which perform basic validations common across chains, they could be put into a common parent utility class

modules/sdk-coin-tia/src/tia.ts
```diff
@@ -0,0 +1,205 @@
...
+      throw new Error('Invalid transaction: ' + e.message);
+    }
+  }
+
+  /**
+   * Sign a transaction with a single private key
+   * @param params parameters in the form of { txPrebuild: {txHex}, prv }
+   * @returns signed transaction in the form of { txHex }
+   */
+  async signTransaction(
```
Vijay-Jagannathan, line 98: Same idea as previous files: most methods here seem very similar/common and could exist in a common class

Vijay-Jagannathan (commented):
Seems like we could refactor a lot of code here into the respective parent classes, which will greatly enhance the current as well as future cosmos side chain onboarding. 

I'd suggest having a refactor PR out first and making it work with osmosis, then rebase this PR with those changes. This is kinda the same approach that we'll be adopting in WP as well.

modules/sdk-coin-tia/src/lib/keyPair.ts
dpkjnr: getAddress function is defined in the parent class thats why we can't add a parameter to it

DinshawKothari: Yes deepak, that's correct. That is why I had to move this method out of abstract cosmos sdk.

modules/sdk-coin-tia/src/lib/transaction.ts
dpkjnr: apart from a few utility functions rest of the code looks exactly same for all these chains. I think we need to find some better way of doing it

zahin-mohammad (commented):
deferring review to coins team due to lack of context.

modules/account-lib/package.json
```diff
@@ -45,6 +45,7 @@
     "@bitgo/sdk-coin-sol": "^2.5.1",
     "@bitgo/sdk-coin-stx": "^2.0.7",
     "@bitgo/sdk-coin-sui": "^3.0.7",
+    "@bitgo/sdk-coin-tia": "^1.0.0",
```
mmcshinsky-bitgo, line 48: Is this required to be part of account-lib for an internal service?

modules/sdk-coin-tia/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@bitgo/sdk-coin-tia",
+  "version": "1.0.0",
```
mmcshinsky-bitgo, line 3: Is this ready for release upon merge? If not, let's zero out the version number or change access value.

modules/statics/src/coins.ts
```diff
@@ -740,7 +740,7 @@ export const coins = CoinMap.fromCoins([
   account(
     '6e35c0dc-7ee8-4273-985d-254aa3641c9d',
     'tia',
-    'celestia',
+    'Celestia',
```
mmcshinsky-bitgo, line 743: This would be a breaking change that would need to be separated into its' own commit.

tsconfig.packages.json
```diff
@@ -142,9 +145,6 @@
     {
       "path": "./modules/sdk-rpc-wrapper"
     },
-    {
-      "path": "./modules/sdk-unified-wallet"
```
dpkjnr, line 146: why to remove this? 🤔

Dockerfile
```diff
@@ -195,15 +197,16 @@ RUN cd /var/bitgo-express && \
...
     yarn link @bitgo/abstract-cosmos && \
+    yarn link @bitgo/sdk-coin-tia && \
     yarn link @bitgo/sdk-coin-xlm && \
     yarn link @bitgo/sdk-coin-xrp && \
     yarn link @bitgo/sdk-coin-zec
 #LINK_END
 
 #LABEL_START
-LABEL created="Mon, 05 Jun 2023 16:57:03 GMT"
+LABEL created="Tue, 06 Jun 2023 10:54:43 GMT"
```
Vijay-Jagannathan, line 207: are these updated automatically ?

modules/sdk-coin-tia/src/lib/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Coin } from '@cosmjs/stargate';
+import BigNumber from 'bignumber.js';
+
+import { CosmosUtils } from '@bitgo/abstract-cosmos';
+import * as constants from './constants';
+
+export class TiaUtils extends CosmosUtils {
+  /** @inheritdoc */
+  isValidAddress(address: string): boolean {
+    return constants.accountAddressRegex.test(address);
```
Vijay-Jagannathan, line 11: nit(non-blocking): regex seems similar to osmo/side chains but not a blocker

modules/sdk-coin-tia/src/lib/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  /** @inheritdoc */
+  isValidValidatorAddress(address: string): boolean {
+    return constants.validatorAddressRegex.test(address);
+  }
+
+  /** @inheritdoc */
+  validateAmount(amount: Coin): void {
```
Vijay-Jagannathan, line 20: similar to osmo: https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-coin-osmo/src/lib/utils.ts#L20

modules/sdk-coin-tia/test/integration/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+describe('TIA', function () {
+  let bitgo: TestBitGoAPI;
+  before(function () {
+    bitgo = TestBitGo.decorate(BitGoAPI, { env: 'mock' });
+    bitgo.safeRegister('tia', Tia.createInstance);
+    bitgo.safeRegister('ttia', Ttia.createInstance);
+    bitgo.initializeTestVars();
+  });
+});
```
Vijay-Jagannathan, line 16: Seems like a duplicate of what's already there in `tia.ts` unit test file. We can delete this file unless this has been added here for a follow up PR

modules/sdk-coin-tia/test/unit/tia.ts
```diff
@@ -0,0 +1,339 @@
...
+  let basecoin;
+  before(function () {
+    bitgo = TestBitGo.decorate(BitGoAPI, { env: 'mock' });
+    bitgo.safeRegister('tia', Tia.createInstance);
+    bitgo.safeRegister('ttia', Ttia.createInstance);
+    bitgo.initializeTestVars();
+    basecoin = bitgo.coin('ttia');
+  });
+
+  it('should retun the right info', function () {
```
Vijay-Jagannathan, line 28: nit: typo; `retun` -> `return`

modules/sdk-coin-tia/test/unit/tia.ts
```diff
@@ -0,0 +1,339 @@
...
+      (() => {
+        basecoin.getAddressDetails(address.invalidMemoIdAddress);
+      }).should.throw();
+    });
+
+    it('should throw on multiple memo id address', () => {
+      (() => {
+        basecoin.getAddressDetails(address.multipleMemoIdAddress);
+      }).should.throw();
+    });
```
Vijay-Jagannathan, line 66: could we parametrize this ?

modules/sdk-coin-tia/test/unit/tia.ts
```diff
@@ -0,0 +1,339 @@
...
+  describe('Parse Transactions: ', () => {
+    const transferInputsResponse = {
+      address: TEST_SEND_TX.recipient,
+      amount: new BigNumber(TEST_SEND_TX.sendAmount).plus(TEST_SEND_TX.gasBudget.amount[0].amount).toFixed(),
+    };
+
+    const transferOutputsResponse = {
+      address: TEST_SEND_TX.recipient,
+      amount: TEST_SEND_TX.sendAmount,
+    };
```
Vijay-Jagannathan, line 319: nit: seems like this is only used in the test present line 321

dpkjnr (approved):
source files look good to me. Couldn't go through all test files

PR #3638, opened 2023-06-09
feat(abstract-cosmos): sdk abstract-cosmos and sdk osmo refactored

DinshawKothari:
Ticket: BG-79223


modules/abstract-cosmos/src/lib/transaction.ts
```diff
@@ -22,15 +22,18 @@ import {
...
 
 export class CosmosTransaction extends BaseTransaction {
   protected _cosmosLikeTransaction: CosmosLikeTransaction;
   protected _accountNumber: number;
   protected _chainId: string;
 
-  constructor(_coinConfig: Readonly<CoinConfig>) {
+  protected _utils: CosmosUtils;
+
+  constructor(_coinConfig: Readonly<CoinConfig>, utils: CosmosUtils) {
```
dpkjnr, line 34: make this also protected, no one else then child classes needs to call this

modules/account-lib/src/index.ts
```diff
@@ -65,7 +65,7 @@ export { Algo };
 import * as Atom from '@bitgo/sdk-coin-atom';
 export { Atom };
 
-import * as Osmo from '@bitgo/sdk-coin-osmo';
+import * as Osmo from '@bitgo/abstract-cosmos';
```
dpkjnr, line 68: why this? we won't have coin specifi module?

DinshawKothari: Here we are using TxnBuilderFactory which we have moved in abstract-cosmos.

sachushaji: Lets have it from the osmo module itself, have the `@bitgo/abstract-cosmos` exported from osmo and exported here. Please export transaction builder from the osmos module also for backward compatibility since devs will be expecting to have the transaction builder available in the coin specific module itself

modules/sdk-coin-osmo/src/lib/index.ts
```diff
@@ -1,10 +1,6 @@
 import * as Constants from './constants';
 import * as Utils from './utils';
 
-export { OsmoTransaction } from './transaction';
-export { OsmoTransactionBuilder } from './transactionBuilder';
-export { TransactionBuilderFactory } from './transactionBuilderFactory';
-export { OsmoTransferBuilder } from './transferBuilder';
```
sachushaji, line 7: Lets have the same transaction builder modules exported from cosmos abstract module for backward compatibility.

modules/abstract-cosmos/src/lib/transaction.ts
```diff
@@ -22,15 +22,18 @@ import {
...
-import utils from './utils';
+import { CosmosUtils } from './utils';
 
 export class CosmosTransaction extends BaseTransaction {
   protected _cosmosLikeTransaction: CosmosLikeTransaction;
   protected _accountNumber: number;
   protected _chainId: string;
 
-  constructor(_coinConfig: Readonly<CoinConfig>) {
+  protected _utils: CosmosUtils;
```
Vijay-Jagannathan, line 32: Can we use the utils file without having to pass it to the constructor ? Also if we are using CosmosUtils for all of the chains, then do we really need to pass it ?

Vijay-Jagannathan: Just looked, seems like we have side chain specific utility methods which extend from cosmos utils, so I'm assuming we're passing coin specific utility methods. Ignore the above comment in that case

modules/abstract-cosmos/src/lib/transactionBuilderFactory.ts
```diff
@@ -1,20 +1,24 @@
...
+import { CosmosTransaction } from './transaction';
+import { CosmosTransactionBuilder } from './transactionBuilder';
+import { CosmosTransferBuilder } from './transferBuilder';
+import { CosmosUtils } from './utils';
 
 export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
-  constructor(_coinConfig: Readonly<CoinConfig>) {
+  protected utils: CosmosUtils;
+
+  constructor(_coinConfig: Readonly<CoinConfig>, utils: CosmosUtils) {
```
dpkjnr, line 14: I think this would cause problem, is it possible to create utils instance here based on the coinConfig?

modules/account-lib/src/index.ts
Vijay-Jagannathan: Yes, let's have the transaction/transaction builders in the respective chains. They can extend from abstract cosmos transaction/transaction builders

Vijay-Jagannathan (commented):
Are we renaming the KeyPair classes in this PR or is that separate ?

modules/abstract-cosmos/src/lib/transactionBuilder.ts
```diff
@@ -128,4 +135,122 @@ export abstract class CosmosTransactionBuilder extends BaseTransactionBuilder {
...
   }
+
+  /**
+   * Sets gas budget of this transaction
+   * Gas budget consist of fee amount and gas limit. Division feeAmount/gasLimit represents
+   * the gas-fee and it should be more than minimum required gas-fee to process the transaction
+   * @param {FeeData} gasBudget
+   * @returns {TransactionBuilder} this transaction builder
+   */
+  gasBudget(gasBudget: FeeData): this {
```
mmcshinsky-bitgo, line 146:
```suggestion
  setGasBudget(gasBudget: FeeData): this {
```

modules/abstract-cosmos/src/lib/utils.ts
```diff
@@ -381,15 +379,209 @@ export class CosmosLikeUtils implements BaseUtils {
...
-    const txRaw = utils.createTxRawFromCosmosLikeTransaction(cosmosLikeTransaction);
+    const txRaw = this.createTxRawFromCosmosLikeTransaction(cosmosLikeTransaction);
     return makeSignDoc(txRaw.bodyBytes, txRaw.authInfoBytes, chainId, accountNumber);
   }
 
   isValidHexString(hexString: string): boolean {
     return /^[0-9A-Fa-f]*$/.test(hexString);
   }
+
+  validateWithdrawRewardsMessage(withdrawRewardsMessage: WithdrawDelegatorRewardsMessage) {
```
mmcshinsky-bitgo, line 390: Can we get jsdocs for this?

modules/abstract-cosmos/src/lib/utils.ts
```diff
@@ -381,15 +379,209 @@ export class CosmosLikeUtils implements BaseUtils {
...
+      );
+    }
+    if (!withdrawRewardsMessage.delegatorAddress || !this.isValidAddress(withdrawRewardsMessage.delegatorAddress)) {
+      throw new InvalidTransactionError(
+        `Invalid WithdrawDelegatorRewardsMessage delegatorAddress: ` + withdrawRewardsMessage.delegatorAddress
+      );
+    }
+  }
+
+  validateDelegateOrUndelegateMessage(delegateMessage: DelegateOrUndelegeteMessage) {
```
mmcshinsky-bitgo, line 406: Can we get jsdocs for this?

mmcshinsky-bitgo (commented):
nit - non-blockers. Would be nice if methods had documentation.

modules/abstract-cosmos/src/cosmosCoin.ts
```diff
@@ -145,6 +220,91 @@ export class CosmosCoin extends BaseCoin {
     };
   }
 
+  /**
+   * Get the public node url from the Environments constant we have defined
+   */
+  protected getPublicNodeUrl(): string {
```
Vijay-Jagannathan, line 226: we don't need this method in this class

modules/abstract-cosmos/src/cosmosCoin.ts
```diff
@@ -145,6 +220,91 @@ export class CosmosCoin extends BaseCoin {
...
+    throw new Error('Method not implemented.');
+  }
+
+  /**
+   * Get account number from public node
+   */
+  protected async getAccountFromNode(senderAddress: string): Promise<request.Response> {
```
Vijay-Jagannathan, line 233: getAccountFromNode, getBalanceFromNode, getChainIdFromNode, getAccountBalance, getChainId and getAccountDetails are also not needed I guess; We prolly need this only when it comes to WRW recovery

Vijay-Jagannathan (commented):
I don't see something along the lines of `CosmosTransactionBuilderFactory` here. Is that already abstracted out ? 

**Context**: Inside coin controller in WP, we have a custom method (non-abstract method) which takes in the factory object to construct the respective transaction object. The way I see it would work is something like this: 

The individual coin controllers will pass in their respective factory objects that reside in the SDK. This method is then passed into the parent cosmos controller class, where we see this coin factory object as a super type (meaning individual coin factory classes should extend from a cosmos top level factory) and it takes this object and instantiates the respective transaction objects for the corresponding coin.

zahin-mohammad (commented):
lacking context on changes; flushing from review Q

modules/abstract-cosmos/src/lib/utils.ts
```diff
@@ -381,15 +379,209 @@ export class CosmosLikeUtils implements BaseUtils {
...
+      }
+      if (value.delegatorAddress == null) {
+        throw new InvalidTransactionError(`Invalid MessageData value.delegatorAddress: ` + value.delegatorAddress);
+      }
+    } else {
+      throw new InvalidTransactionError(`Invalid MessageData TypeUrl is not supported: ` + messageData.typeUrl);
+    }
+  }
+
+  validateTransaction(tx: CosmosLikeTransaction): void {
```
sachushaji, line 458: Kindly add jsdocs for functions without one

modules/sdk-coin-osmo/src/lib/index.ts
```diff
@@ -1,10 +1,7 @@
 import * as Constants from './constants';
 import * as Utils from './utils';
 
-export { OsmoTransaction } from './transaction';
-export { OsmoTransactionBuilder } from './transactionBuilder';
+export { CosmosTransaction, CosmosTransactionBuilder } from '@bitgo/abstract-cosmos';
```
sachushaji, line 4:
```suggestion
export { CosmosTransaction as Transaction, CosmosTransactionBuilder as TransactionBuilder} from '@bitgo/abstract-cosmos';
```

modules/abstract-cosmos/src/lib/utils.ts
```diff
@@ -381,15 +379,209 @@ export class CosmosLikeUtils implements BaseUtils {
...
+      throw new InvalidTransactionError(`Invalid SendMessage toAddress: ` + sendMessage.toAddress);
+    }
+    if (!sendMessage.fromAddress || !this.isValidAddress(sendMessage.fromAddress)) {
+      throw new InvalidTransactionError(`Invalid SendMessage fromAddress: ` + sendMessage.fromAddress);
+    }
+    this.validateAmountData(sendMessage.amount);
+  }
+
+  // coin specific placeholder methods -
+  validateAmount(amount: Coin): void {
```
dpkjnr, line 572: make these three function abstract

modules/abstract-cosmos/src/cosmosCoin.ts
DinshawKothari: Yes i have added it as you will need it as part of recovery task. Adding this in cosmosCoin file will remove duplicacy of code, and the child class can override them in case they have different implementations.

modules/abstract-cosmos/src/lib/transactionBuilder.ts
DinshawKothari: Other setter methods are also not using set prefix. Will take care of this in a separate PR.

modules/abstract-cosmos/src/cosmosCoin.ts
```diff
@@ -100,16 +106,82 @@ export class CosmosCoin extends BaseCoin {
...
+  isValidAddress(address: string): boolean {
     throw new Error('Method not implemented.');
   }
 
-  isValidAddress(address: string): boolean {
-    throw new Error('Method not implemented.');
+  /** @inheritDoc **/
+  async verifyTransaction(params: VerifyTransactionOptions): Promise<boolean> {
+    let totalAmount = new BigNumber(0);
+    // const coinConfig = coins.get(this.getChain());
```
sachushaji, line 116: Remove if not needed

modules/abstract-cosmos/src/lib/transactionBuilder.ts
sachushaji: Kindly add a todo ticket for this

modules/abstract-cosmos/src/lib/utils.ts
```diff
@@ -341,6 +353,10 @@ export class CosmosLikeUtils implements BaseUtils {
     return false;
   }
 
+  /**
+   * Returns whether or not the string is a valid protocol public key
+   * @param {string} publicKey - the  public key to be validated
```
sachushaji, line 358:
```suggestion
   * @param {string | undefined} publicKey - the  public key to be validated
```

sachushaji (approved):
Kindly do smaller PRs from next time :-)

dpkjnr (approved):

Vijay-Jagannathan (approved):

modules/abstract-cosmos/src/cosmosCoin.ts
DinshawKothari: Will remove in the next PR.

modules/abstract-cosmos/src/lib/utils.ts
DinshawKothari: Will remove in the next PR.

PR #3646, opened 2023-06-12
fix(bitgo-js): replace isValidHexAddress method for correct pattern

evalenti-simtlix:
replace isValidHexAddress method for correct pattern

BG-387


wesgraham (dismissed):

modules/sdk-coin-trx/src/trx.ts
```diff
@@ -176,7 +177,7 @@ export class Trx extends BaseCoin {
    * @param address hex address
    */
   isValidHexAddress(address: string): boolean {
-    return address.length === 42 && /^(0x)?([0-9a-f]{2})+$/i.test(address);
+    return /^41[0-9a-f]{40}$/i.test(address);
```
mukeshsp, line 180: it would be good if you can add the testcase.

margueriteblair (approved):

modules/sdk-coin-trx/src/trx.ts
evalenti-simtlix: added new cases for the existing test

PR #3661, opened 2023-06-14
feat(express): implement EdDSA commitments for external signer

alebusse:
implemented EdDSA commitment step for external signer added unit test

WP-94

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
@@ -16,11 +16,13 @@ import {
   getTxRequest,
   offerUserToBitgoRShare,
   sendUserToBitgoGShare,
+  ShareKeyPosition,
   SigningMaterial,
 } from '../../../tss';
 import {
   CommitmentShareRecord,
   CommitmentType,
+  CustomCommitmentGeneratingFunction,
   CustomGShareGeneratingFunction,
   CustomRShareGeneratingFunction,
   EncryptedSignerShareRecord,
@@ -364,10 +366,16 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
     return keychains;
   }
 
-  async createRShareFromTxRequest(params: {
+  async createCommitmentShareFromTxRequest(params: {
     txRequest: TxRequest;
     prv: string;
-  }): Promise<{ rShare: SignShare; signingKeyYShare: YShare }> {
+    walletPassphrase: string;
+  }): Promise<{
+    userToBitgoCommitment: CommitmentShareRecord;
+    encryptedSignerShare: EncryptedSignerShareRecord;
+    encryptedUserToBitgoRShare: EncryptedSignerShareRecord;
+  }> {
+    const bitgoIndex = ShareKeyPosition.BITGO;
     const { txRequest, prv } = params;
     const txRequestResolved: TxRequest = txRequest;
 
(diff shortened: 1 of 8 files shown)


modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -364,10 +364,16 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
...
     txRequest: TxRequest;
     prv: string;
-  }): Promise<{ rShare: SignShare; signingKeyYShare: YShare }> {
+    walletPassphrase: string;
+  }): Promise<{
+    userToBitgoCommitment: CommitmentShareRecord;
+    encryptedSignerShare: EncryptedSignerShareRecord;
+    encryptedUserToBitgoRShare: EncryptedSignerShareRecord;
+  }> {
+    const bitgoIndex = 3;
```
islamaminBitGo, line 376: im not sure but i think there are some defined constants in the SDK for the indices of bitgo, user, and backup. We could make use of here, instead of directly using 3.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -394,8 +401,67 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
...
+        : txRequestResolved.unsignedTxs[0];
+
+    const signingKey = MPC.keyDerive(
+      userSigningMaterial.uShare,
+      [userSigningMaterial.bitgoYShare, userSigningMaterial.backupYShare],
+      unsignedTx.derivationPath
+    );
+
+    let rShare: SignShare;
+    if (walletPassphrase && encryptedUserToBitgoRShare) {
```
islamaminBitGo, line 450: Do we need to worry about backwards compatibility here? No, right? Since we control the full stack, i.e. both instances of express, iIthink this path wouldnt need to produce signingKeyYShare to BitGo. As that would have been returned by the previous call and already used by BitGo in the first commitment round. So, technically, we could do without this if condition and always use a flow that only returns the user's R share at this decommitment step. As a consequence, we wouldnt even need the prv key for this step, just the wallet password.

alebusse: Hmm, maybe we can remove the compatibility later, after we enforce it. The problem with making a breaking change now, is that if something else needs to be released too they gonna need to upgrade the express version and hiting this breaking change.

islamaminBitGo: But what exactly is it breaking? Once a client upgrades express they get both express (external signer) and express (platform) working well together since the new flow would be available to both instances. And clients with older express versions should still be fine due to backwards compatibility support within our stack for now.

alebusse: The clients may want to upgrade express to bring other changes and not fix this if we make a breaking change.

islamaminBitGo: but that upgrade will bring in all your new changes with them that include commitment and the whole thing would still work no? Clients dont have the option to upgrade only certain parts of express, and there is no reason a client should not want to have commitments included.

sachushaji (commented):
Flushing queue

sachushaji (commented):
Flushing as I don't have much context on eddsa commitments unfortunately and also seems there is no files owned by my teams

islamaminBitGo (dismissed):

islamaminBitGo (approved):

zahin-mohammad (approved):

PR #3663, opened 2023-06-15
chore(account-lib): export abstract cosmos

Vijay-Jagannathan:
Ticket: [WIN-15](https://bitgoinc.atlassian.net/browse/WIN-15)

Exporting abstract cosmos files so that they can be consumed inside WP (for accessing message types and other constructs specific to abstract cosmos coins) via account-lib using the individual coin sdk 

[WIN-15]: https://bitgoinc.atlassian.net/browse/WIN-15?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-coin-osmo/src/index.ts
+++ b/modules/sdk-coin-osmo/src/index.ts
@@ -2,3 +2,4 @@ export * from './lib';
 export * from './osmo';
 export * from './tosmo';
 export * from './register';
+export * as AbstractCosmos from '@bitgo/abstract-cosmos'; // <TODO: Restrict abstract cosmos export - https://bitgoinc.atlassian.net/browse/WIN-16>


Vijay-Jagannathan:
Missing `abstract-cosmos` in `package.json` of account-lib; pushing fix

mukeshsp (dismissed):

aarvay (commented):
No lockfile changes?

modules/account-lib/src/index.ts
```diff
@@ -92,6 +92,9 @@ export { Polygon };
 import * as Sui from '@bitgo/sdk-coin-sui';
 export { Sui };
 
+import * as Cosmos from '@bitgo/abstract-cosmos';
```
DinshawKothari, line 95: I think we should export classes only from coin sdks.
If you think there is something like iface not exported in sdk-coin, we can export it there and use it from that.

This will help in backward compatibility and consistency and WP not dependent on abstract-cosmos. As abstract-cosmos is just facilitating abstraction for cosmos side chains.

DinshawKothari (changes requested):

Vijay-Jagannathan:
> No lockfile changes?

didn't see any afaik, will check 👀

modules/account-lib/src/index.ts
Vijay-Jagannathan: +1 That's the ideal approach. I'm okay with this as long as we can sufficiently abstract tx building process. I'll try to export what we need via the individual coins only and see if works as intended. Will fall back to this approach if coin specific export doesn't help.

Vijay-Jagannathan:
exporting abstract cosmos files via the individual SDK coin(s) now (this change will need to be made for each coin). Any code which will need to access the abstract cosmos constructs, will now have to do it via the individual sdk coin object.

modules/sdk-coin-osmo/src/index.ts
```diff
@@ -2,3 +2,4 @@ export * from './lib';
 export * from './osmo';
 export * from './tosmo';
 export * from './register';
+export * as AbstractCosmos from '@bitgo/abstract-cosmos';
```
DinshawKothari, line 5: The idea is to not have Abstract Cosmos visible to WP. Instead, whatever is required should be directly imported from coin sdk. We can handle it later as discussed.
Please mark a TODO comment.

Vijay-Jagannathan: Added

DinshawKothari (approved):

PR #3666, opened 2023-06-16
feat(utxo-lib): various improvements to utxo-lib and related libs

OttoAllmendinger:
See commits for details

Issue: BTC-0

### DIFF
--- a/modules/utxo-lib/src/bitgo/index.ts
+++ b/modules/utxo-lib/src/bitgo/index.ts
@@ -1,3 +1,5 @@
+export { PsbtInput, PsbtOutput } from 'bip174/src/lib/interfaces';
+
 export * as bcashAddress from './bitcoincash';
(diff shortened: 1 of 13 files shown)


modules/utxo-lib/src/bitgo/index.ts
```diff
@@ -1,3 +1,5 @@
+export { PsbtInput, PsbtOutput } from 'bip174/src/lib/interfaces';
```
mmcshinsky-bitgo, line 1:
```suggestion
export { PsbtInput, PsbtOutput } from 'bip174/;
```
Can we get to these without going into `src`?

OttoAllmendinger: unfortunately not (hence the re-export here)

saravanan7mani (dismissed):

saravanan7mani (approved):

PR #3667, opened 2023-06-16
feat(sdk-coin-xlm): adding txHex to explainTransaction params for xlm

margueriteblair:
For the migration to HSMv3, we have a new wallet recovery flow. In BGA we've made changes to the recovery flow for v3 that involve using `explainTransaction` and expect a transaction hex as input. We need to adapt XLM for this reason to accept `txHex` as input in it's coin specific `explainTransaction` method.

### DIFF
--- a/modules/sdk-coin-xlm/src/xlm.ts
+++ b/modules/sdk-coin-xlm/src/xlm.ts
@@ -100,7 +100,8 @@ interface SupplementGenerateWalletOptions {
 }
 
 interface ExplainTransactionOptions {
-  txBase64: string;
+  txHex?: string;
+  txBase64?: string;
 }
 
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-xlm/src/xlm.ts
```diff
@@ -817,14 +818,26 @@ export class Xlm extends BaseCoin {
    * @param params
    */
   async explainTransaction(params: ExplainTransactionOptions): Promise<TransactionExplanation> {
```
mmcshinsky-bitgo, line 820: Can we update the docs around the accepts params for this?

sachushaji (approved):

PR #3668, opened 2023-06-16
fix(wp): recreate tx for account coins

davidkaplanbitgo:
For non-utxo coins, they need to be able rebuild payment txs
too. For coins like SOL, there is a very small validity
window so you need to rebuild so that it will be
valid when broadcasted to the network.

Issue: EA-561

### DIFF
--- a/modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
+++ b/modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
@@ -182,7 +182,10 @@ export class PendingApproval implements IPendingApproval {
     // If there are no recipients, then the transaction cannot be recreated
     const recipients = this.info()?.transactionRequest?.buildParams?.recipients || [];
     const type = this.info()?.transactionRequest?.buildParams?.type;
-    if (recipients.length === 0 && type !== 'consolidate') {
+
+    // We only want to not recreate transactions with no recipients if it is an UTXO coin.
+    // There is no easy way to do this in sdk-core, but `network` is only set on UTXO coins.
+    if ('network' in this.baseCoin && recipients.length === 0 && type !== 'consolidate') {
       canRecreateTransaction = false;
     }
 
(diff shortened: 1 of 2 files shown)


brandonblack:
@davidkaplanbitgo 2 commit message things: 1) can you add the `Issue: ` in front of the issue, and 2) can you add the why that was explained in slack? (I don't quite understand why there aren't recipients, but I know that for Solana transactions are only valid for a brief time after creation).

davidkaplanbitgo:
> @davidkaplanbitgo 2 commit message things: 1) can you add the `Issue: ` in front of the issue, and 2) can you add the why that was explained in slack? (I don't quite understand why there aren't recipients, but I know that for Solana transactions are only valid for a brief time after creation).

Fixed commit message in last push.

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
```diff
@@ -182,7 +182,10 @@ export class PendingApproval implements IPendingApproval {
     // If there are no recipients, then the transaction cannot be recreated
     const recipients = this.info()?.transactionRequest?.buildParams?.recipients || [];
     const type = this.info()?.transactionRequest?.buildParams?.type;
-    if (recipients.length === 0 && type !== 'consolidate') {
+
+    // We only want to not recreate transactions with no recipients if it is an UTXO coin.
+    // There is no easy way to do this in sdk-core, but `network` is only set on UTXO coins.
```
brandonblack, line 187: disgusting. nice hate letter to us.

brandonblack (approved):

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
davidkaplanbitgo: an option is to make `isUtxoModel` on `AbstractCoin` and we could use it here. But this PR seemed urgent so it is this way for now.

CapnMigraine (approved):

dszheng92 (approved):

PR #3670, opened 2023-06-18
fix(sdk-coin-ada): refine ada pledge parsing logic

theMobiusStrip:
- Refine transaction type parsing for pledge since existing pledge update might only take 1 pool registration certificate
- Add method to add user pre-signed node key witness

Ticket: EA-420


sachushaji (commented):
Changes LGTM, can you update the commit message with breaking changes to the pledge tx builder mentioned

noel-bitgo (approved):

PR #3675, opened 2023-06-19
fix(statics): update XTZ/TXTZ block explorer

yashrohatgi-bitgo:
- update explorer url for XTZ/TXTZ

Ticket: [CE-286](https://bitgoinc.atlassian.net/browse/CE-286)

[CE-286]: https://bitgoinc.atlassian.net/browse/CE-286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -712,15 +712,13 @@ class XrpTestnet extends Testnet implements AccountNetwork {
 class Xtz extends Mainnet implements AccountNetwork {
   name = 'Xtz';
   family = CoinFamily.XTZ;
-  explorerUrl = 'https://tezblock.io/transaction/';
-  accountExplorerUrl = 'https://tezblock.io/account/';
+  explorerUrl = 'https://tzstats.com/';
 }
 
 class XtzTestnet extends Testnet implements AccountNetwork {
   name = 'XtzTestnet';
   family = CoinFamily.XTZ;
-  explorerUrl = 'https://carthagenet.tezblock.io/transaction/';
-  accountExplorerUrl = 'https://carthagenet.tezblock.io/account/';
+  explorerUrl = 'https://mumbai.tzstats.com/';
 }
 
 class ZCash extends Mainnet implements UtxoNetwork {


modules/statics/src/networks.ts
```diff
@@ -712,15 +712,13 @@ class XrpTestnet extends Testnet implements AccountNetwork {
 class Xtz extends Mainnet implements AccountNetwork {
   name = 'Xtz';
   family = CoinFamily.XTZ;
-  explorerUrl = 'https://tezblock.io/transaction/';
```
yashrohatgi-bitgo, line 715: I'm not sure what to use for the account explorer url, based on what I see I think its just the base url with an account id appended

modules/statics/src/networks.ts
```diff
@@ -712,15 +712,13 @@ class XrpTestnet extends Testnet implements AccountNetwork {
...
-  accountExplorerUrl = 'https://tezblock.io/account/';
+  explorerUrl = 'https://tzstats.com/';
 }
 
 class XtzTestnet extends Testnet implements AccountNetwork {
   name = 'XtzTestnet';
   family = CoinFamily.XTZ;
-  explorerUrl = 'https://carthagenet.tezblock.io/transaction/';
-  accountExplorerUrl = 'https://carthagenet.tezblock.io/account/';
+  explorerUrl = 'https://mumbai.tzstats.com/';
```
yashrohatgi-bitgo, line 721: https://tezos.gitlab.io/introduction/test_networks.html

carthagenet stopped being maintained, I think mumbainet is the current testnet we should use based on the above link

zahin-mohammad (approved):

modules/statics/src/networks.ts
mmcshinsky-bitgo: I would ask stake holders to define the exact links that should be used.

mmcshinsky-bitgo (approved):

PR #3684, opened 2023-06-22
Bg 00000 inj sdk issue

DinshawKothari:


modules/statics/src/base.ts
```diff
@@ -45,7 +45,7 @@ export enum CoinFamily {
   FIAT = 'fiat',
   HASH = 'hash', // Provenance
   HBAR = 'hbar',
-  INJ = 'injective',
+  INJECTIVE = 'injective',
```
mmcshinsky-bitgo, line 48: What's the thought process behind switching this?

PR #3685, opened 2023-06-22
feat: add cold wallet feature flags

bitgoAaron:
- identify precisely which coins support cold wallets

Ticket: CE-1156


modules/statics/src/coins.ts
```diff
@@ -286,7 +314,7 @@ export const coins = CoinMap.fromCoins([
     Networks.test.bitcoinGold,
     UnderlyingAsset.BTG,
     BaseUnit.BTC,
-    BTG_FEATURES
+    BTG_FEATURES.filter((f) => f !== CoinFeature.MULTISIG_COLD)
```
bitgoAaron, line 317: technically we shouldn't even have this as an option in statics, since there is no such coin implementation for `tbtg`, only the mainnet one but I'd rather not open this change up to also include breaking changes if not necessary.

zahin-mohammad (dismissed):

mmcshinsky-bitgo (dismissed):

mmcshinsky-bitgo (approved):

PR #3686, opened 2023-06-22
feat(sdk-coin-injective): injective sdk init along with testcases

DinshawKothari:
Ticket: BG-79064


modules/statics/src/coins.ts
```diff
@@ -868,21 +868,21 @@ export const coins = CoinMap.fromCoins([
     '5f9506c5-f10a-43c2-92d3-52941083bbc7',
     'injective',
```
mmcshinsky-bitgo, line 869: Not sure why this was changed. Our pattern would be INJ and inj. Same with Networks.main.inj

mmcshinsky-bitgo: I would recommend keeping to established patterns and fixing these.

modules/account-lib/package.json
```diff
@@ -41,6 +41,7 @@
     "@bitgo/sdk-coin-eth2": "^1.2.27",
     "@bitgo/sdk-coin-hash": "^1.1.0",
     "@bitgo/sdk-coin-hbar": "^1.3.27",
+    "@bitgo/sdk-coin-injective": "^1.1.2",
```
mmcshinsky-bitgo, line 44: Is this required in account-lib for other services or can we remove it?

DinshawKothari: Yes this will be required for Wallet Platform.

modules/statics/src/coins.ts
DinshawKothari: We have inj coin name already used for erc20 inj token. That is why we have used injective for this.

mmcshinsky-bitgo: Gotcha. Another name collision :(

modules/sdk-coin-injective/src/tinjective.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Testnet Tia
```
Vijay-Jagannathan, line 2: nit: typo; should be injective

Vijay-Jagannathan (approved):

PR #3687, opened 2023-06-22
fix(sdk-core): do not treat sol as a utxo-coin for approval

brandonblack:
Solana _and_ Bitcoin-like coins have `network` property. This will now only detect Bitcoin-like coins as intended.

Should fix approving pending approvals for Solana.

Issue: BG-79393

### DIFF
--- a/modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
+++ b/modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
@@ -3,6 +3,7 @@
  */
 import * as _ from 'lodash';
 import * as common from '../../common';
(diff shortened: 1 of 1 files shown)


davidkaplanbitgo (approved):
We really need to fix this

brandonblack:
Yeah, was thinking same. Need a more robust way to decide whether a transaction can/should be rebuilt :cry:

Also should just be able to rebuild even without recipients, but that's a whole other thing around communicating intent consistently through the build/approve/sign/approve/send flow.

alebusse (approved):

linyah9912 (approved):

PR #3690, opened 2023-06-23
fix(statics): update ADA block explorer url

yashrohatgi-bitgo:
- explorer url was not linking to transactions correctly, added to url accordingly

Ticket: [CE-985](https://bitgoinc.atlassian.net/browse/CE-985)

[CE-985]: https://bitgoinc.atlassian.net/browse/CE-985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -120,7 +120,7 @@ class Ada extends Mainnet implements AdaNetwork {
   utxolibName = 'cardano';
   poolDeposit = 500000000;
   stakeKeyDeposit = 2000000;
-  explorerUrl = 'https://explorer.cardano.org/en/';
+  explorerUrl = 'https://explorer.cardano.org/en/transaction?id=';
   coinsPerUtxoWord = 34482;
   maxTransactionSize = 8000;
   maxValueSize = 4000;


CapnMigraine (dismissed):

modules/statics/src/networks.ts
```diff
@@ -120,7 +120,7 @@ class Ada extends Mainnet implements AdaNetwork {
   utxolibName = 'cardano';
   poolDeposit = 500000000;
   stakeKeyDeposit = 2000000;
-  explorerUrl = 'https://explorer.cardano.org/en/';
+  explorerUrl = 'https://explorer.cardano.org/en/transaction?id=';
```
alebusse, line 123: This gonna is gonna be parsed by the UI as
`https://explorer.cardano.org/en/transaction + txId + id=`

mmcshinsky-bitgo: This should be fixed in that repo, not here as this is the base definition.

yashrohatgi-bitgo: Ah I see that now. Not sure how I'd work around this though aside from hard coding something in the UI

alebusse (approved):

mmcshinsky-bitgo (approved):

PR #3708, opened 2023-07-07
feat(osmo): add contract address to valid addresses

dpkjnr:
Ticket: WIN-23

### DIFF
--- a/modules/sdk-coin-osmo/test/unit/osmo.ts
+++ b/modules/sdk-coin-osmo/test/unit/osmo.ts
@@ -97,6 +97,24 @@ describe('OSMO', function () {
       should.equal(utils.isValidValidatorAddress(undefined as unknown as string), false);
       should.equal(utils.isValidValidatorAddress(''), false);
     });
+
+    it('should validate contract addresses correctly', () => {
+      should.equal(utils.isValidContractAddress(address.contractAddress1), true);
+      should.equal(utils.isValidContractAddress(address.contractAddress2), false);
+      should.equal(utils.isValidContractAddress('dfjk35y'), false);
+      should.equal(utils.isValidContractAddress(undefined as unknown as string), false);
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-osmo/test/unit/osmo.ts
```diff
@@ -97,6 +97,16 @@ describe('OSMO', function () {
...
+
+    it('should all type of address on Osmo coin address validation', () => {
+      const tomso = bitgo.coin('tosmo');
+      should.equal(tomso.isValidAddress(address.address1), true);
+      should.equal(tomso.isValidAddress(address.validatorAddress1), true);
+      should.equal(tomso.isValidAddress(address.contractAddress1), true);
+      should.equal(tomso.isValidAddress('dfjk35y'), false);
+      // @ts-expect-error - testing invalid address
+      should.equal(tomso.isValidAddress(undefined), false);
+    });
```
DinshawKothari, line 109: it('should validate account addresses correctly', () => {
}

check this above. Add contractAddress also there itself. No need for a new case.

dpkjnr: there we are checking `utils.isValidAddress` that only check for account address, it will fail for validator address and contract address. I think we should rename it to isValidAccountAddress.
In my new test case I am checking `osmo.isValidAddress` that works for all three kinds of addresses

DinshawKothari: Got it, so
Osmo.isValidAddress() contains all three types of check that is there in utils file.

So its more or less the same thing.

But instead of repeated checks, can we just have some cases for isValidContractAddress() particularly.

If you want you can keep this as well. Just change the description that is easy to understand.

DinshawKothari (approved):

PR #3711, opened 2023-07-10
fix(sdk-core): fix prebuildTransactionTss method

alebusse:
Fixed prebuildTransactionTss method to enforce using txRequest full api version for external signer.

WP-208

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
@@ -493,6 +493,8 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
       txRequestId = txRequest.txRequestId;
     }
 
+    const { apiVersion } = txRequestResolved;
+
     const { userToBitgoCommitment, encryptedSignerShare, encryptedUserToBitgoRShare } =
       await externalSignerCommitmentGenerator({ txRequest: txRequestResolved });
 
@@ -502,7 +504,7 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
       txRequestId,
       userToBitgoCommitment,
       encryptedSignerShare,
-      'full'
+      apiVersion
     );
 
     const { rShare } = await externalSignerRShareGenerator({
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2631,17 +2631,28 @@ export class Wallet implements IWallet {
...
+
+    if (
+      _.isFunction(params.customCommitmentGeneratingFunction) &&
+      _.isFunction(params.customGShareGeneratingFunction) &&
+      _.isFunction(params.customRShareGeneratingFunction)
+    ) {
+      // force txRequest full API version for custom signing
+      params.apiVersion = 'full';
+    }
+
```
pengyuc-bitgo, line 2643: Should we fail the call with a error message similar to below ("using custom signing must use 'full' api version")?

alebusse: The use case for external signing has always been using full version, https://github.com/BitGo/BitGoJS/blob/e15f9b8d09e50649121f5a5c6bf5a8f4bad0b283/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts#L480-L524.

alebusse: So instead of adapting the signing process to keep splitting to lite vs full, i rather force all to full.

margueriteblair (approved):

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -493,6 +493,8 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
       txRequestId = txRequest.txRequestId;
     }
 
+    const { apiVersion } = txRequestResolved;
```
islamaminBitGo, line 496: apiVersion seems to be marked as an optional parameter in the type TxRequest. What do we do if it is not provided.

islamaminBitGo (changes requested):
apiVersion seems like an optional parameter

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
alebusse: `txRequestResolved` is actually a `txRequest` we get from wallet-platform or from the build result, so is not optional like the `buildParams` `apiVersion`

islamaminBitGo (approved):

PR #3713, opened 2023-07-10
feat(express): improve fetchEncryptedPrivKeys script

alebusse:
improve fetchEncryptedPrivKeys script

WP-186


mmcshinsky-bitgo:
Is this if the user also wants to store their private keys on their server as well instead of fetching them?

mmcshinsky-bitgo (approved):

PR #3714, opened 2023-07-10
fix(eth): add prefix for EIP-191 to workaround HSM bug

rlutzbitgo:
TICKET: HSM-104

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -2624,7 +2624,12 @@ describe('V2 Wallet:', function () {
       let signTxRequestForMessage;
       const messageSigningCoins = ['teth', 'tpolygon'];
       const messageRaw = 'test';
-      const expected: SignedMessage = { txRequestId: reqId.toString(), txHash, messageRaw, coin: 'teth' };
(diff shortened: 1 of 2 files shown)


OttoAllmendinger (commented):
flush (insufficient context)

rlutzbitgo:
Think this makes more sense to just do in WP anyway

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -2624,7 +2624,12 @@ describe('V2 Wallet:', function () {
       let signTxRequestForMessage;
       const messageSigningCoins = ['teth', 'tpolygon'];
       const messageRaw = 'test';
-      const expected: SignedMessage = { txRequestId: reqId.toString(), txHash, messageRaw, coin: 'teth' };
+      const expected: SignedMessage = {
+        txRequestId: reqId.toString(),
+        txHash,
+        messageRaw: `\u0019Ethereum Signed Message:\n${messageRaw.length}${messageRaw}`,
```
pranavjain97, line 2630: Please add a comment to the code as to why we need a prefix here i.e a short summary of what is explained int the ticket. Otherwise it would look very sus

PR #3723, opened 2023-07-13
fix(sdk-core): Use PrebuiltTx with Tss Full SendMany

mohammadalfaiyazbitgo:
This pr enables rebuilding a prebuiltTx when used with SendMany on TSS Full wallets. Previously this method would not work because the SDK would be looking for a non-existent txRequestID from the prebuiltTx.

Ticket: WP-139

### DIFF
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
@@ -5,11 +5,17 @@ import * as openpgp from 'openpgp';
 import * as should from 'should';
 import * as sinon from 'sinon';
 
-import { mockSerializedChallengeWithProofs, mockSerializedChallengeWithProofs2, TestBitGo } from '@bitgo/sdk-test';
-import { BitGo, createSharedDataProof, RequestType } from '../../../../../src';
+import {
+  mockSerializedChallengeWithProofs,
+  mockSerializedChallengeWithProofs2,
+  TestableBG,
+  TestBitGo,
+} from '@bitgo/sdk-test';
+import { BitGo, createSharedDataProof, TssUtils, RequestType } from '../../../../../src';
 import {
   BackupGpgKey,
   BackupKeyShare,
+  BaseCoin,
   BitgoGPGPublicKey,
   BitgoHeldBackupKeyShare,
   common,
@@ -58,6 +64,8 @@ describe('TSS Ecdsa Utils:', async function () {
   let bgUrl: string;
   let tssUtils: ECDSAUtils.EcdsaUtils;
   let wallet: Wallet;
+  let bitgo: TestableBG & BitGo;
+  let baseCoin: BaseCoin;
   let bitgoKeyShare;
   let userKeyShare: KeyShare;
   let backupKeyShare: KeyShare;
@@ -143,10 +151,10 @@ describe('TSS Ecdsa Utils:', async function () {
       },
     };
 
-    const bitgo = TestBitGo.decorate(BitGo, { env: 'mock' });
+    bitgo = TestBitGo.decorate(BitGo, { env: 'mock' });
     bitgo.initializeTestVars();
 
-    const baseCoin = bitgo.coin(coinName);
+    baseCoin = bitgo.coin(coinName);
 
     bgUrl = common.Environments[bitgo.getEnv()].uri;
 
@@ -173,6 +181,7 @@ describe('TSS Ecdsa Utils:', async function () {
       enterprise: enterpriseId,
       coin: coinName,
       coinSpecific: {},
+      multisigType: 'tss',
     };
     wallet = new Wallet(bitgo, baseCoin, walletData);
     tssUtils = new ECDSAUtils.EcdsaUtils(bitgo, baseCoin, wallet);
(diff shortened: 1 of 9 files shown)


modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
```diff
@@ -369,4 +369,16 @@ export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtil
     const bitgoPublicKeyStr = response.publicKey as string;
     return readKey({ armoredKey: bitgoPublicKeyStr });
   }
+
+  /**
+   * Returns supported TxRequest versions for this wallet
+   */
+  public supportedTxRequestVersions(): TxRequestVersion[] {
```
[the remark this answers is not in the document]
zahin-mohammad: Tbh given these are tss utils, it should probablly happen at the consutrctor, but doesnt seem like we do that.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1818,8 +1818,26 @@ export class Wallet implements IWallet {
...
+    if (
+      // verify the wallet must use txRequest Full api and must rebuild the tx before submitting
+      this._wallet.multisigType === 'tss' &&
+      !this.tssUtils?.supportedTxRequestVersions().includes('lite') &&
+      params.prebuildTx &&
+      !(typeof params.prebuildTx === 'string') &&
+      params.prebuildTx.buildParams &&
+      'preview' in params.prebuildTx.buildParams &&
+      params.prebuildTx.buildParams.preview
+    ) {
```
zahin-mohammad, line 1831: This condition is hard to read, please extract to a helper fn

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2990,7 +3008,7 @@ export class Wallet implements IWallet {
     }
 
     // ECDSA TSS uses TxRequestFull
-    if (this.baseCoin.getMPCAlgorithm() === 'ecdsa' || params.apiVersion === 'full' || this._wallet.type === 'cold') {
+    if (!this.tssUtils?.supportedTxRequestVersions().includes('lite') || params.apiVersion === 'full') {
```
zahin-mohammad, line 3011: Can we just define the apiVersion we are using at the top of this fn. would be easier to read/reuse.

zahin-mohammad (changes requested):

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsa.ts
```diff
@@ -1353,6 +1362,40 @@ describe('TSS Ecdsa Utils:', async function () {
     });
   });
 
+  describe('supportedTxRequestVersions', function () {
+    it('returns only full for hot, cold, custodial wallet types', function () {
```
zahin-mohammad, line 1366: This should be separated into 3 separate tests.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts
```diff
@@ -870,6 +875,83 @@ describe('TSS Utils:', async function () {
     });
   });
 
+  describe('supportedTxRequestVersions', function () {
+    it('should return full for custodial and cold wallets', async function () {
```
zahin-mohammad, line 879: This should be two separate tests.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -284,6 +294,14 @@ export class Wallet implements IWallet {
     return this._wallet.coin;
   }
 
+  type(): WalletType | undefined {
+    return this._wallet.type;
```
zahin-mohammad, line 298: We can probably default to 'hot' when this value is undefined.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1818,8 +1836,27 @@ export class Wallet implements IWallet {
       throw error;
     }
 
+    let txPrebuildQuery;
```
zahin-mohammad, line 1839: Is there a type we can assign here?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1818,8 +1836,27 @@ export class Wallet implements IWallet {
...
+    const isTssWallet = this._wallet.multisigType === 'tss';
+    const supportedTxRequestVersions = this.tssUtils?.supportedTxRequestVersions() || [];
+    const mustUseTxRequestFull = supportedTxRequestVersions.length === 1 && supportedTxRequestVersions.includes('full');
+
+    if (
+      // verify the wallet must use txRequest Full api and must rebuild the tx before submitting
+      isTssWallet &&
+      mustUseTxRequestFull &&
```
zahin-mohammad, line 1847:
```
isTssWallet &&
mustUseTxRequestFull &&
```

seems redundant, 

can just use `mustUseTxRequestFull`

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2990,7 +3028,10 @@ export class Wallet implements IWallet {
     }
 
     // ECDSA TSS uses TxRequestFull
-    if (this.baseCoin.getMPCAlgorithm() === 'ecdsa' || params.apiVersion === 'full' || this._wallet.type === 'cold') {
+    const supportedTxRequestVersions = this.tssUtils?.supportedTxRequestVersions() ?? [];
+    const onlySupportsTxRequestFull =
+      supportedTxRequestVersions.length === 1 && supportedTxRequestVersions.includes('full');
+    if (apiVersion === 'full' || onlySupportsTxRequestFull) {
```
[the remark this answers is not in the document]
zahin-mohammad: We should throw an error in this case.

zahin-mohammad (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
mohammadalfaiyazbitgo: that's not gonna match the typing on the class.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2990,7 +3028,10 @@ export class Wallet implements IWallet {
     }
 
     // ECDSA TSS uses TxRequestFull
-    if (this.baseCoin.getMPCAlgorithm() === 'ecdsa' || params.apiVersion === 'full' || this._wallet.type === 'cold') {
+    const supportedTxRequestVersions = this.tssUtils?.supportedTxRequestVersions() ?? [];
+    const onlySupportsTxRequestFull =
+      supportedTxRequestVersions.length === 1 && supportedTxRequestVersions.includes('full');
```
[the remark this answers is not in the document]
mohammadalfaiyazbitgo: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types , compiler will complain if the var is being used in ways that don't make sense with the type.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1818,8 +1836,27 @@ export class Wallet implements IWallet {
       throw error;
     }
 
+    let txPrebuildQuery;
+    const isTssWallet = this._wallet.multisigType === 'tss';
+    const supportedTxRequestVersions = this.tssUtils?.supportedTxRequestVersions() || [];
+    const mustUseTxRequestFull = supportedTxRequestVersions.length === 1 && supportedTxRequestVersions.includes('full');
```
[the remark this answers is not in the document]
mohammadalfaiyazbitgo: not magic strings, both 'tss' and 'full' are valid constrained types: https://betterprogramming.pub/understanding-string-literals-in-typescript-72862ecfca3e

modules/sdk-core/src/bitgo/wallet/wallet.ts
mohammadalfaiyazbitgo: `Promise<PrebuildTransactionResult | string>`

modules/sdk-core/src/bitgo/wallet/wallet.ts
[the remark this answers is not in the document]
mohammadalfaiyazbitgo: will follow up with a refactoring on the types

zahin-mohammad (approved):

PR #3726, opened 2023-07-14
feat(sdk-coin-osmo): allow amount in execute contract transaction

dpkjnr:
Ticket: WIN-23

### DIFF
--- a/modules/abstract-cosmos/src/lib/transaction.ts
+++ b/modules/abstract-cosmos/src/lib/transaction.ts
@@ -224,13 +224,14 @@ export class CosmosTransaction extends BaseTransaction {
       case TransactionType.ContractCall:
         explanationResult.type = TransactionType.ContractCall;
         message = json.sendMessages[0].value as ExecuteContractMessage;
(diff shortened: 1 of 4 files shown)


modules/abstract-cosmos/src/lib/transaction.ts
```diff
@@ -224,13 +224,14 @@ export class CosmosTransaction extends BaseTransaction {
       case TransactionType.ContractCall:
         explanationResult.type = TransactionType.ContractCall;
         message = json.sendMessages[0].value as ExecuteContractMessage;
+        outputAmount = message.funds?.[0]?.amount ?? '0';
```
Vijay-Jagannathan, line 227: what changed suddenly and could you add some description as well ? ticket doesn't indicate why this was needed

DinshawKothari (approved):

modules/abstract-cosmos/src/lib/transaction.ts
DinshawKothari: Changes LGTM. But yes +1 on the above comment.

PR #3728, opened 2023-07-17
feat(utxo-lib): add updateWalletOutputForPsbt

davidkaplanbitgo:
Refactor out output updating from adding in the same way as we do for inputs. This is useful for partial updating of wallet outputs for hydration.

TICKET: BTC-284

### DIFF
--- a/modules/utxo-lib/src/bitgo/wallet/WalletOutput.ts
+++ b/modules/utxo-lib/src/bitgo/wallet/WalletOutput.ts
@@ -1,4 +1,5 @@
 import { taproot } from 'bitcoinjs-lib';
+import { PsbtOutputUpdate } from 'bip174/src/lib/interfaces';
 import { UtxoPsbt } from '../UtxoPsbt';
 import { RootWalletKeys } from './WalletKeys';
 import { ChainCode, scriptTypeForChain } from './chains';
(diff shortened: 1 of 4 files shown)


modules/utxo-lib/src/bitgo/wallet/WalletOutput.ts
```diff
@@ -29,14 +30,67 @@ export function addWalletOutputToPsbt(
...
+  if (psbt.data.outputs.length <= outputIndex) {
+    throw new Error(
+      `outputIndex (${outputIndex}) is too large for the number of outputs (${psbt.data.outputs.length})`
+    );
+  }
+
+  // It is expensive to attempt to compute every output address using psbt.txOutputs[outputIndex]
+  // to then just get the script. Here, we are doing the same thing as what txOutputs() does in
+  // bitcoinjs-lib, but without iterating over each output.
+  const outputScript: Buffer = (psbt as any).__CACHE.__TX.outs[outputIndex].script;
```
davidkaplanbitgo, line 72: This is kind of ugly, but it allows us to not call `const outputScript = psbt.txOutputs[outputIndex]` for each iteration. Additionally, we do not even try to make any addresses from the scripts, so it is very efficient

davidkaplanbitgo:
Failing due to vulnerabilities, nothing wrong with the PR

brandonblack (dismissed):

davidkaplanbitgo:
Rebased off of https://github.com/BitGo/BitGoJS/pull/3729

modules/utxo-lib/src/bitgo/wallet/WalletOutput.ts
OttoAllmendinger: let's add a method `getOutputScript(outputIndex): Buffer` to UtxoPsbt then (maybe we can upstream it)

OttoAllmendinger (changes requested):

PR #3730, opened 2023-07-18
chore(statics): onboard new batch of tokens

hitansh-madan:
Ticket: EA-889


mukeshsp (dismissed):

mukeshsp (dismissed):

zahin-mohammad (dismissed):

modules/statics/src/coins.ts
```diff
@@ -9594,6 +9610,24 @@ export const coins = CoinMap.fromCoins([
...
+    'Serum (Wormhole)',
+    6,
+    'xnorPhAzWXUczCP3KjU5yDxmKKZi5cSbxytQ1LgE3kG',
+    UnderlyingAsset.SRM,
+    AccountCoin.DEFAULT_FEATURES
+  ),
+  solToken(
+    'cd7d9622-1467-405d-af1f-09edd1d431d6',
+    'sol:usdc-wormhole',
+    'USD Coin (PoS) (Wormhole)',
```
mmcshinsky-bitgo, line 9625:
```suggestion
    'USD Coin (PoS) (Wormhole) (USDC)',
```
Do we need to match solscan's exact naming?

hitansh-bitgo: Seems like solscan appends the token symbol while displaying the name, we generally do not include that in the name so I left it out as well

mukeshsp (approved):

PR #3732, opened 2023-07-18
fix(bitgo): pin dependency for audit resolution

mmcshinsky-bitgo:
CE-1454


zahin-mohammad (commented):
can we please use a valid ticket reference?

yarn.lock
```diff
@@ -18690,7 +18689,7 @@ vm-browserify@~0.0.1:
   dependencies:
     indexof "0.0.1"
 
-vm2@^3.9.17, vm2@^3.9.18:
+vm2@^3.9.18:
```
dpkjnr, line 18692: why do we still need this?

mmcshinsky-bitgo: Probably from not removing node_modules during a local install first. I'll run that and see

dpkjnr (dismissed):

zahin-mohammad (approved):

PR #3737, opened 2023-07-19
fix(utxo-lib): allow uncompressed pubkeys in utxo-lib

OttoAllmendinger:
There are some transactions built by very old versions of utxo-lib that contain
uncompressed pubkeys. Currently those cannot be parsed. This commit fixes that.

Issue: BTC-0


rushilbg (approved):
The change for showing who signed is in another PR?

PR #3738, opened 2023-07-19
chore(utxo-lib): improve readme of utxo-bin

rushilbg:
Add steps for upgrading the bin tool

BTC-0

### DIFF
--- a/modules/utxo-bin/README.md
+++ b/modules/utxo-bin/README.md
@@ -10,6 +10,7 @@ npm install --global @bitgo/utxo-bin
 ```
 
 # Sample Usage
+Either run using ``yarn ts-node bin/index.ts ...`` or install globally again using previous npm command.
 
 ## parseTx
 
--- a/modules/utxo-bin/src/commands.ts
+++ b/modules/utxo-bin/src/commands.ts
@@ -207,7 +207,7 @@ export const cmdParseTx = {
         throw new Error(`conflicting arguments`);
       }
       console.log('Reading from stdin. Please paste hex-encoded transaction data.');
-      console.log('Press Ctrl-D to finish, or Ctrl-C to cancel.');
+      console.log('After inserting data, press Ctrl-D to finish. Press Ctrl-C to cancel.');
       if (process.stdin.isTTY) {
         data = await readStdin();
       } else {


saravanan7mani (dismissed):

modules/utxo-bin/README.md
```diff
@@ -9,6 +9,14 @@ formatting them for terminal output.
 npm install --global @bitgo/utxo-bin
 ```
 
+# Upgrade
+```
+cd modules/utxo-bin
+yarn build
+```
+
```
OttoAllmendinger, line 17: I don't think we specific instructions for this module - the toplevel README should over it

doesn't `yarn install` on the toplevel work as well?

modules/utxo-bin/README.md
```diff
@@ -9,6 +9,14 @@ formatting them for terminal output.
...
+Either run using ``yarn ts-node bin/index.ts ...`` or install globally again using previous npm command.
```
OttoAllmendinger, line 18: the `yarn ts-node bin/index.ts` should be mentioned in a section `Usage` / `From Repo`

we should probably add a new section `Installation / From Repo` with `cd modules/utxo-bin; npm install --global`

OttoAllmendinger (changes requested):

modules/utxo-bin/README.md
rushilbg: No it doesn't

OttoAllmendinger: Does `yarn build` on the toplevel do the right thing?

modules/utxo-bin/src/commands.ts
```diff
@@ -183,7 +183,7 @@ export const cmdParseTx = {
         throw new Error(`conflicting arguments`);
       }
       console.log('Reading from stdin. Please paste hex-encoded transaction data.');
-      console.log('Press Ctrl-D to finish, or Ctrl-C to cancel.');
+      console.log('Press Ctrl-D to finish (after pasting tx hex), or Ctrl-C to cancel.');
```
OttoAllmendinger, line 186:
```suggestion
      console.log('After inserting data, press Ctrl-D to finish. Press Ctrl-C to cancel.');
```

OttoAllmendinger (changes requested):

modules/utxo-bin/README.md
rushilbg: > error Command "build" not found.

OttoAllmendinger: The `yarn install` (or just `yarn`) command on the toplevel should build all packages. If it doesn't work that after that we have a bug that we need to fix separately.

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #3740, opened 2023-07-19
feat(abstract-cosmos): add wallet recovery for abstract cosmos

Vijay-Jagannathan:
Ticket: WIN-144

### Context:

- Adds the wallet recovery changes for abstract cosmos and osmosis module
- Tested by using the WRW changes locally and successfully [recovered](https://testnet.osmosis.explorers.guru/transaction/9F999C59CE9F51A485E2DD560FCE2865EDCC2C8428515F5D143E24A7CB4F8A27) a test wallet

### Follow up changes:
- WRW changes will be part of a separate PR for clarity
- Recovery changes for other coins will involve minimal coin controller changes along with relevant unit tests for recovery


modules/abstract-cosmos/src/cosmosCoin.ts
```diff
@@ -110,6 +125,270 @@ export class CosmosCoin extends BaseCoin {
...
+      throw new Error('missing wallet passphrase');
+    }
+
+    // Step 2: Fetch the bitgo key from params
+    const bitgoKey = params.bitgoKey.replace(/\s/g, '');
+
+    // Step 3: Instantiate the ECDSA signer and fetch the address details
+    const MPC = new Ecdsa();
+    const chainId = await this.getChainId();
+    const currPath = 'm/0';
```
zahin-mohammad, line 164: Please define this as a constant somewhere and export it. The SDK and WP should be alligned on what the base path is, and the easiest way to do that is to use the exported SDK value.

modules/abstract-cosmos/src/cosmosCoin.ts
```diff
@@ -110,6 +125,270 @@ export class CosmosCoin extends BaseCoin {
...
+      userKeyDerivedCombined.xShare.y !== backupKeyCombined.xShare.y ||
+      userKeyDerivedCombined.xShare.chaincode !== backupKeyCombined.xShare.chaincode
+    ) {
+      throw new Error('Common keychains do not match');
+    }
+
+    return [userKeyDerivedCombined, backupKeyCombined];
+  }
+
+  private async signRecoveryTSS(
```
zahin-mohammad, line 303: Out of scope for this PR, but we should eventually move some of these methods to sdk-lib-mpc.

modules/abstract-cosmos/src/cosmosCoin.ts
Vijay-Jagannathan: added

modules/sdk-coin-osmo/test/unit/osmo.ts
```diff
@@ -394,4 +399,131 @@ describe('OSMO', function () {
...
+        recoveryDestination: destinationAddress,
+      });
+      res.should.not.be.empty();
+      res.should.hasOwnProperty('serializedTx');
+      res.should.hasOwnProperty('scanIndex');
+      sandBox.assert.calledOnce(basecoin.getAccountBalance);
+      sandBox.assert.calledOnce(basecoin.getAccountDetails);
+      sandBox.assert.calledOnce(basecoin.getChainId);
+
+      const atomTxn = new CosmosTransaction(coin, utils);
```
DinshawKothari, line 446: osmoTxn or just txn may be.

modules/abstract-cosmos/src/cosmosCoin.ts
DinshawKothari: nit: If we are using currPath as 'm/0'
Then we dont need 'Starting Scan Index' and 'Address Scanning Factor' field in WRW form.
As we have restricted it to 0 in the path.

Not a blocker for this PR.

Vijay-Jagannathan:
@DinshawKothari 

that's a good point actually, its not needed at all. Will make the necessary changes - will likely need the same for ATOM too

DinshawKothari (approved):

PR #3744, opened 2023-07-24
fix(sdk-coin-osmo): allow zero amount on contract call

Vijay-Jagannathan:
Ticket: WIN-155

### Context:

We observed during testing that - for contract call execution, we were throwing an error for 0 amount, which is not what we want. Because we still want to execute contract without any amount. This PR amends the validation flow for amount, so that we can account for 0 amount to be passed in the respective contract call builder.
<p align="center">

</p>

### DIFF
--- a/modules/abstract-cosmos/src/lib/utils.ts
+++ b/modules/abstract-cosmos/src/lib/utils.ts
@@ -646,10 +646,11 @@ export class CosmosUtils implements BaseUtils {
   /**
    * Validates an array of coin amounts.
    * @param {Coin[]} amountArray - The array of coin amounts to validate.
+   * @param {TransactionType} transactionType - optional field for transaction type
    */
-  validateAmountData(amountArray: Coin[]): void {
+  validateAmountData(amountArray: Coin[], transactionType?: TransactionType): void {
     amountArray.forEach((coinAmount) => {
-      this.validateAmount(coinAmount);
+      this.validateAmount(coinAmount, transactionType);
     });
   }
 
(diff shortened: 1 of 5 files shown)


modules/abstract-cosmos/src/lib/utils.ts
```diff
@@ -646,10 +646,11 @@ export class CosmosUtils implements BaseUtils {
   /**
    * Validates an array of coin amounts.
    * @param {Coin[]} amountArray - The array of coin amounts to validate.
+   * @param {number} transactionType - optional field for transaction type
    */
-  validateAmountData(amountArray: Coin[]): void {
+  validateAmountData(amountArray: Coin[], transactionType?: number): void {
```
DinshawKothari, line 651: why type is number ??

Vijay-Jagannathan: Unsure of the historical reasons but `TransactionType.ContractCall` is a number, hence using the concrete type

Vijay-Jagannathan: changing it to enum to look cleaner, is that what you meant ?

DinshawKothari: yes using TransactionType looks better.

modules/abstract-cosmos/src/lib/utils.ts
```diff
@@ -719,9 +721,10 @@ export class CosmosUtils implements BaseUtils {
   /**
    * Validates a execute contract message
    * @param {ExecuteContractMessage} message - The execute contract message to validate
+   * @param {number} transactionType - optional field for transaction type
```
DinshawKothari, line 724: change here also.

Vijay-Jagannathan: good catch, just pushed

grseetha (approved):

dpkjnr (approved):

PR #3748, opened 2023-07-24
chore: remove xlm from testnet

alexzhengbitgo:
prime team has decided to remove the coin xlm from the testnet. this is because the wallet expires every 3 months, and thus we would need to keep renewing it. we decided the best course of action is to remove it from the testnet while ensuring it works gatekept on prod
Ticket: PX-1669

### DIFF
--- a/modules/statics/src/account.ts
+++ b/modules/statics/src/account.ts
@@ -965,36 +965,6 @@ export function stellarToken(
 }
 
 /**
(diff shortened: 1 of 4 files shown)


mmcshinsky-bitgo:
Statics wouldn't necessarily be the gate keeper here, but the platform providing the coins correct? Can't we do this on the app itself and leave statics clean?

PR #3749, opened 2023-07-24
feat(sdk-core): add pallier pub to bitgo tss key creation

kxl4126:
Ticket: EA-37


margueriteblair (commented):
Changes look ok to me, once you fix those affected tests can approve unless someone w/ more TSS knowledge says otherwise

margueriteblair (approved):

PR #3759, opened 2023-07-25
chore(sdk-core): add clarity to other commonKeychain util

kxl4126:
Ticket: BG-0000

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
@@ -1183,7 +1183,13 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       .result();
   }
 
-  static publicKeyFromCommonKeychain(commonKeychain: string): string {
+  /**
+   * util function that checks that a commonKeychain is valid and can ultimately resolve to a valid public key
+   * @param commonKeychain - a user uploaded commonKeychain string
+   * @throws if the commonKeychain is invalid length or invalid format
+   */
+
+  static validateCommonKeychainPublicKey(commonKeychain: string) {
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -1183,7 +1183,12 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       .result();
   }
 
-  static publicKeyFromCommonKeychain(commonKeychain: string) {
+  /**
+   * util function that checks that a commonKeychain is valid and can ultimately resolve to a valid public key
+   * @param commonKeychain - a user uploaded commonKeychain string
```
zahin-mohammad, line 1188: Does this function throw any errors?

zahin-mohammad: Can you also please add non-happy path tests?

zahin-mohammad (changes requested):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -1183,7 +1183,13 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
+   * @throws if the commonKeychain is invalid length or invalid format
+   */
+
+  static getUncompressedPublicKeyFromCommonKeychain(commonKeychain: string) {
```
zahin-mohammad, line 1192: This naming is little clearer
```suggestion
  static validateCommonKeychainPublicKey(commonKeychain: string) {
```

zahin-mohammad (approved):

alebusse (commented):
I dont think you need `chai`, you could use `should` in a similar way

alebusse (approved):

PR #3765, opened 2023-07-27
chore(statics): add block explorer

mullapudipruthvik:
Ticket: CS-768


alebusse (approved):
Looks good but could you improve pr description or ticket so we have more context on why we need a block explorer url?

mullapudipruthvik:
got it! we have to support transfers created using blocks, therefore we need to update the UI to redirect to block url

PR #3796, opened 2023-08-10
feat(bitgo): calculate fees explicitly for psbts

davidkaplanbitgo:
It is more correct to calculate the miner fee explicitly with a psbt because we have all the information needed. It also means we do not have to depend on wallet platform to tell us the fees

BTC-0


lcovar (dismissed):
If we have tests for explain transaction let’s add another that asserts fee is part of the output.

davidkaplanbitgo:
Added test

lcovar (approved):

aarvay (approved):

PR #3802, opened 2023-08-14
fix(abstract-cosmos): allow valueless transfer for cosmos coin

ajays97:
Ticket: BG-78370, WP-423

### DIFF
--- a/modules/sdk-coin-osmo/src/osmo.ts
+++ b/modules/sdk-coin-osmo/src/osmo.ts
@@ -65,4 +65,9 @@ export class Osmo extends CosmosCoin {
   getAddressFromPublicKey(publicKey: string): string {
     return new KeyPair({ pub: publicKey }).getAddress();
   }
+
+  /** @inheritDoc **/
+  valuelessTransferAllowed(): boolean {
+    return true;
+  }
 }


OttoAllmendinger (dismissed):

modules/abstract-cosmos/src/cosmosCoin.ts
```diff
@@ -126,6 +126,10 @@ export class CosmosCoin extends BaseCoin {
     throw new Error('Method not implemented.');
   }
 
+  valuelessTransferAllowed(): boolean {
+    return true;
+  }
```
mukeshsp, line 131: @ajays97 it should be in osmo module.

DinshawKothari: yes.

DinshawKothari: Mark it false here, and true in osmo module.

DinshawKothari (changes requested):

modules/abstract-cosmos/src/cosmosCoin.ts
ajays97: Makes sense. Adding it to only osmo for now.

mukeshsp (approved):

Vijay-Jagannathan (approved):

DinshawKothari (approved):

PR #3805, opened 2023-08-14
feat(sdk-core): add, use SendTransactionRequest codec

OttoAllmendinger:
Foray into io-ts codecs in SDK. This is a first step towards validating
the request body before sending it to the server.

We are phasing it in gradually, so that we discover any issues with
the codec in a controlled manner.

Issue: BTC-368


davidkaplanbitgo (approved):

alebusse (approved):
Should we put all the types together in a different folder/file structure  or use a file naming convention ?

PR #3808, opened 2023-08-15
feat(sdk-core): add postWithCodec utility function

OttoAllmendinger:
This is a helper method for the development of request codecs.

Since it is difficult to get the codec right the first time, this method 
allows to send the request with the original body and set some header flags
to indicate that the codec failed to decode or encode the body.

Issue: BTC-368

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/postWithCodec.ts
+++ b/modules/sdk-core/src/bitgo/utils/postWithCodec.ts
@@ -1,0 +1,77 @@
+import * as t from 'io-ts';
+import { BitGoBase } from '../bitgoBase';
+import { BitGoRequest } from '../../api';
+import { SuperAgent, SuperAgentRequest } from 'superagent';
+import { isLeft } from 'fp-ts/Either';
+
+/**
+ * @param body
+ * @param encodedBody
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/utils/postWithCodec.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @param encodedBody
+ * @returns a list of unknown properties that are present in the body but not the codec.
+ */
+function getUnknownProperties(body: Record<string, unknown>, encodedBody: Record<string, unknown>): string[] {
+  const unknownProperties: string[] = [];
+  if (body && encodedBody) {
+    const bodyKeys = Object.keys(body);
+    const encodedBodyKeys = Object.keys(encodedBody);
+    const unknownKeys = bodyKeys.filter((key) => !encodedBodyKeys.includes(key));
+    unknownProperties.push(...unknownKeys);
```
davidkaplanbitgo, line 18:
```suggestion
    unknownProperties = unknownKeys.map((k) => body[k]);
```
This might make more sense?

alebusse (approved):

PR #3809, opened 2023-08-15
fix(sdk-coin-dot): remove check for `controller` to determine staking tx

noel-bitgo:
EA-1060


lcovar:
Flushing from queue

lcovar (commented):
Flushing PR to empty my review queue.

PR #3811, opened 2023-08-16
fix(sdk-core): handle ContractCall intent for TSS transaction type

ajays97:
Ticket: BG-78370


Vijay-Jagannathan (approved):
LGTM, just wondering if it's possible to add any integ test for this flow

PR #3812, opened 2023-08-16
fix(sdk-core): handle ContractCall intent for TSS transaction type

ajays97:
Ticket: BG-78370


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2712,6 +2712,7 @@ export class Wallet implements IWallet {
     }
 
     let txRequest: TxRequest;
+    console.log(`Got params.type: ${JSON.stringify(params)}`);
```
mmcshinsky-bitgo, line 2715:
```suggestion
```

PR #3819, opened 2023-08-17
fix(sdk-coin-dot): default to sender address when controller is missing

noel-bitgo:
EA-1060


margueriteblair (approved):
lgtm as long as previous tests are passing

PR #3820, opened 2023-08-17
docs(root): update SMC scripts

alebusse:
updated SMC scripts ts-node version and docs

BG-79903


margueriteblair (dismissed):

examples/ts/tss-smc/README.md
```diff
@@ -14,10 +15,10 @@ $ nvm use
 $ npm install
 ```
 
-Use `npx ts-node` to run the desired example:
+Use `npm` to run the desired step. For example, to run step 1:
```
mmcshinsky-bitgo, line 18: This is used so that a user can run a `ts` file example directly. Maybe identify the list of scripts in their own section of the readme as well as in the example directly?

margueriteblair (approved):

PR #3821, opened 2023-08-18
fix(sdk-core): handle txRequest full PA before signing

alebusse:
In the txRequest full flow, we have to check if the txRequest is pending approval before trying to sign.

WP-439

TICKET: WP-439

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -1855,6 +1855,14 @@ export class Wallet implements IWallet {
       });
     }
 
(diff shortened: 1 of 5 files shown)


zahin-mohammad (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1855,6 +1855,13 @@ export class Wallet implements IWallet {
       });
     }
 
+    if (mustUseTxRequestFull && signingParams.txPrebuild.txRequestId) {
```
zahin-mohammad, line 1858: Can we add a test for this function also? An e2e test for pending approvals

zahin-mohammad (approved):

PR #3822, opened 2023-08-18
feat(sdk-api): update & test create & sign transaction functions for v1 safe wallets with uncompressed public keys

rushilbg:
Note: This is for legacy v1 safe btc wallets only.


modules/sdk-api/test/unit/v1/wallet.ts
```diff
@@ -365,26 +365,41 @@ describe('Wallet Prototype Methods', function () {
...
         validate: true,
       });
       halfSignedTx.tx.should.equal(halfSignedTxHex);
+
+      const fullSignedTx = await fakeTestWallet.signTransaction({
+        transactionHex: halfSignedTxHex,
+        signingKey: bitgoSigningKey,
+        unspents,
+        validate: true,
+        fullLocalSigning: true,
```
OttoAllmendinger, line 450: what does this mean? Doesn't this have the effect of calling `.build()`?

OttoAllmendinger (approved):

modules/sdk-api/test/unit/v1/wallet.ts
rushilbg: No, it's just for verifying the count of signatures. If set, then 2 sigs are needed otherwise 1.

OttoAllmendinger (approved):

modules/sdk-api/src/v1/transactionBuilder.ts
```diff
@@ -802,16 +812,21 @@ const estimateTransactionSize = function (params) {
...
-  const estimatedSize =
-    VirtualSizes.txP2shInputSize * params.nP2shInputs +
+  // The size of an uncompressed public key is 32 bytes more than the compressed key,
+  // and hence, needs to be accounted for in the transaction size estimation.
+  const uncompressedPublicKeysTripleCorrectionFactor = 32 * 3;
+
+  return (
+    (VirtualSizes.txP2shInputSize +
+      (params.containsUncompressedPublicKeys ? uncompressedPublicKeysTripleCorrectionFactor : 0)) *
+      params.nP2shInputs +
```
OttoAllmendinger, line 822:
```suggestion
   // This is not quite accurate - if there is a mix of inputs scripts where some used
   // compressed keys and some used uncompressed keys, we would overestimate the size.
   // Since we don't have mixed input sets, this should not be an issue in practice.
    (VirtualSizes.txP2shInputSize +
      (params.containsUncompressedPublicKeys ? uncompressedPublicKeysTripleCorrectionFactor : 0)) *
      params.nP2shInputs +
```

OttoAllmendinger (dismissed):

OttoAllmendinger (dismissed):

rushilbg:
@islamaminBitGo FYI for the flaky test I disabled. PTAL.

pranavjain97 (approved):

davidkaplanbitgo (approved):

PR #3826, opened 2023-08-21
chore(statics) : onboard new tokens

hitansh-madan:
Ticket: EA-1314


OttoAllmendinger (commented):
flush

modules/statics/src/coins.ts
```diff
@@ -5034,6 +5050,14 @@ export const coins = CoinMap.fromCoins([
     '0xa31b1767e09f842ecfd4bc471fe44f830e3891aa',
     UnderlyingAsset.ROOBEE
   ),
+  erc20(
+    'a72f89b9-7c43-4547-8d4e-9f6d10867347',
+    'rpl',
+    'Rocket Pool Protocol',
```
aarvay, line 5056:
```suggestion
    'Rocket Pool',
```

aarvay (changes requested):

modules/statics/src/coins.ts
hitansh-bitgo: Does this not need to match the value for name() function from the contract?

aarvay: Not a hard rule.

aarvay (approved):

PR #3827, opened 2023-08-21
feat(utxo-lib): allow both compressed & uncompressed pubkeys

rushilbg:
in output script creation of v1 safe wallets

Also export legacySafe functions from utxolib

BTC-335

### DIFF
--- a/modules/utxo-lib/src/bitgo/legacysafe/index.ts
+++ b/modules/utxo-lib/src/bitgo/legacysafe/index.ts
@@ -48,8 +48,12 @@ export function createLegacySafeOutputScript2of3(
   }
 
   pubkeys.forEach((key) => {
-    if (key.length !== 65) {
-      throw new Error(`Unexpected key length ${key.length}. Must use uncompressed keys.`);
+    if (key.length !== pubkeys[0].length) {
+      throw new Error(`all pubkeys must have the same length`);
+    }
+    if (key.length !== 65 && key.length !== 33) {
+      // V1 Safe BTC wallets could contain either uncompressed or compressed pubkeys
+      throw new Error(`Unexpected key length ${key.length}, neither compressed nor uncompressed.`);
     }
   });
 
--- a/modules/utxo-lib/test/bitgo/legacysafe/index.ts
+++ b/modules/utxo-lib/test/bitgo/legacysafe/index.ts
@@ -31,12 +31,11 @@ describe('public key conversion', function () {
 
 describe('legacy safe scripts creation', function () {
   const compressedKeyPairs = getKeyTriple('utxo');
-  const compressedPubKeys = compressedKeyPairs.map((keyPair) => keyPair.publicKey);
 
-  it('throws error for compressed public keys', function () {
+  it('throws error for invalid public keys', function () {
     assert.throws(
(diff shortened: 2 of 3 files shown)


modules/utxo-lib/test/bitgo/legacysafe/index.ts
```diff
@@ -34,9 +34,9 @@ describe('legacy safe scripts creation', function () {
   const compressedPubKeys = compressedKeyPairs.map((keyPair) => keyPair.publicKey);
 
   it('throws error for compressed public keys', function () {
-    assert.throws(
+    assert.doesNotThrow(
```
OttoAllmendinger, line 37: the second arg here is not required anymore

you could add a test with a empty buffer or something

OttoAllmendinger (changes requested):

modules/utxo-lib/src/bitgo/legacysafe/index.ts
```diff
@@ -48,8 +48,9 @@ export function createLegacySafeOutputScript2of3(
   }
 
   pubkeys.forEach((key) => {
-    if (key.length !== 65) {
-      throw new Error(`Unexpected key length ${key.length}. Must use uncompressed keys.`);
+    // V1 Safe BTC wallets could contain both uncompressed and compressed pubkeys
```
OttoAllmendinger, line 51: technically I believe they must either all be compressed or uncompressed but we probably won't encounter a mixed-length scenario

OttoAllmendinger (dismissed):

modules/utxo-lib/src/bitgo/legacysafe/index.ts
```diff
@@ -48,8 +48,9 @@ export function createLegacySafeOutputScript2of3(
...
+    if (key.length !== 65 && key.length !== 33) {
```
OttoAllmendinger, line 52:
```suggestion
    if (key.length !== keys[0].length) { throw new Error(`inconsistent key lengths`); }
    if (key.length !== 65 && key.length !== 33) {
```
    if (key.length !== 65 && key.length !== 33) {
```

**Review by OttoAllmendinger — DISMISSED** [review]

File: modules/utxo-lib/test/bitgo/legacysafe/index.ts [hunk 4]
```diff
@@ -31,12 +31,11 @@ describe('public key conversion', function () {
...
   const compressedKeyPairs = getKeyTriple('utxo');
-  const compressedPubKeys = compressedKeyPairs.map((keyPair) => keyPair.publicKey);
 
-  it('throws error for compressed public keys', function () {
+  it('throws error for invalid public keys', function () {
     assert.throws(
-      () => createLegacySafeOutputScript2of3(compressedPubKeys, networks.bitcoin),
-      /^Error: Unexpected key length 33. Must use uncompressed keys.$/
+      () => createLegacySafeOutputScript2of3([Buffer.alloc(10), Buffer.alloc(11), Buffer.alloc(12)], networks.bitcoin),
+      /^Error: Unexpected key length 10, neither compressed nor uncompressed.$/
```
OttoAllmendinger, line 38: shouldn't this be a different error now

OttoAllmendinger (approved):

PR #3828, opened 2023-08-21
fix(utxo-lib): remove prev tx from psbt

saravanan7mani:
Large and unknown type prevTx (nonWitnessUtxo) causes issues in hsm
This util function will help to remove it in WP and tests.

Ticket: BTC-376

### DIFF
--- a/modules/utxo-lib/src/testutil/psbt.ts
+++ b/modules/utxo-lib/src/testutil/psbt.ts
@@ -114,8 +114,12 @@ export function signPsbtInput(
   inputIndex: number,
   rootWalletKeys: RootWalletKeys,
   sign: 'halfsigned' | 'fullsigned',
-  signers?: { signerName: KeyName; cosignerName?: KeyName }
+  params?: {
+    signers?: { signerName: KeyName; cosignerName?: KeyName };
+    deterministic?: boolean;
+  }
 ): void {
+  const { signers, deterministic } = params ?? {};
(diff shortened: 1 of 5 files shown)


modules/utxo-lib/src/testutil/psbt.ts
```diff
@@ -108,14 +108,16 @@ export function getSigners(inputType: InputScriptType): { signerName: KeyName; c
...
-  signers?: { signerName: KeyName; cosignerName?: KeyName }
-): void {
+export function signPsbtInput(params: {
+  psbt: UtxoPsbt;
+  input: Input;
+  inputIndex: number;
+  rootWalletKeys: RootWalletKeys;
+  sign: 'halfsigned' | 'fullsigned';
+  signers?: { signerName: KeyName; cosignerName?: KeyName };
+  deterministic?: boolean;
```
OttoAllmendinger, line 118: it would probably be enough to make only the last two params named, no?

saravanan7mani: Few more params are incoming in a follow up PR. So becoming harder to pass arguments.

margueriteblair (dismissed):

modules/utxo-lib/src/testutil/psbt.ts
saravanan7mani: Done

davidkaplanbitgo (approved):

saravanan7mani:
Duplicate of PR #3830

PR #3829, opened 2023-08-21
refactor(sdk-lib-mpc): move fn's from sdk-core

zahin-mohammad:


modules/sdk-lib-mpc/test/unit/shamir.ts
```diff
@@ -1,17 +1,19 @@
-import { Ed25519Curve, ShamirSecret, Secp256k1Curve } from '@bitgo/sdk-core';
+import 'should';
```
alebusse, line 1:
```suggestion
```
Should you maybe replace any `should` use with `assert`?

zahin-mohammad: This PR doesn't actually change the tests/ src code in any meaningful way, just moving things around + some renaming,

lcovar (approved):

alebusse (approved):

zhongxishen (approved):

PR #3832, opened 2023-08-22
feat(sdk-coin-sol): return ovc compatible format for sol sweep txns

alia-bitgo:
TICKET: [WP-463](https://bitgoinc.atlassian.net/browse/WP-463?atlOrigin=eyJpIjoiZWViNDVjYzE4ZjQ2NDZhYzllNWE5NWVlOTNiYzVkNmUiLCJwIjoiaiJ9)

[WP-463]: https://bitgoinc.atlassian.net/browse/WP-463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -105,6 +105,13 @@ interface SolTx {
   serializedTx: string;
   scanIndex: number;
   coin?: string;
+  signableHex?: string;
+  derivationPath?: string;
+  parsedTx?: ParsedTransaction;
+}
+
+interface SolTxs {
+  transactions: SolTx[];
 }
 
 interface SolDurableNonceFromNode {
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-sol/src/sol.ts
```diff
@@ -714,17 +721,40 @@ export class Sol extends BaseCoin {
...
+      {
+        address: completedTransaction.inputs[0].address,
+        valueString: completedTransaction.inputs[0].value,
+        value: new BigNumber(completedTransaction.inputs[0].value).toNumber(),
+      },
+    ];
+    const outputs = [
+      {
+        address: completedTransaction.outputs[0].address,
+        valueString: completedTransaction.inputs[0].value,
```
rlutzbitgo, line 735: I know that this is what was specified, but kinda curious why we want value and valueString for inputs but not for outputs

rlutzbitgo (approved):

modules/sdk-coin-sol/src/sol.ts
alia-bitgo: yeah i just followed exactly what ovc specified, let me follow up with Islam

alia-bitgo: just keeping it as is for now

PR #3841, opened 2023-08-24
feat(express): add external signer support for signig with derivation…

islamaminBitGo:
… paths

TICKET: EA-1383

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -1614,7 +1614,12 @@ export class Wallet implements IWallet {
       if (typeof this.baseCoin.signWithCustomSigningFunction === 'function') {
         return this.baseCoin.signWithCustomSigningFunction(params.customSigningFunction, signTransactionParams);
       }
-      return params.customSigningFunction(signTransactionParams);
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1614,10 +1614,11 @@ export class Wallet implements IWallet {
       if (typeof this.baseCoin.signWithCustomSigningFunction === 'function') {
         return this.baseCoin.signWithCustomSigningFunction(params.customSigningFunction, signTransactionParams);
       }
-      const userKey = await this.baseCoin.keychains().getKeysForSigning({ wallet: this })[0];
+      const keys = await this.baseCoin.keychains().getKeysForSigning({ wallet: this });
+      const derivationSeed = keys[0] ? keys[0].derivedFromParentWithSeed : undefined;
```
alebusse, line 1618: maybe find by `source === user` instead of index?

Otherwise just do `derivationSeed: userKey?.derivedFromParentWithSeed` in line 1620

islamaminBitGo: Did not find source in the type of key in the sdk, im assuming WP provides it but we should do the typing of this as a follow up.

evanzbitgo (approved):

alebusse (approved):

PR #3853, opened 2023-08-28
feat(sdk-core): add getDerivationPath method for smc wallets

alia-bitgo:
TICKET: WP-521

### DIFF
--- a/modules/sdk-lib-mpc/src/util.ts
+++ b/modules/sdk-lib-mpc/src/util.ts
@@ -1,6 +1,7 @@
 import { PublicKey } from 'paillier-bigint';
 import { bitLength, randBits } from 'bigint-crypto-utils';
 import { gcd } from 'bigint-mod-arith';
+import crypto from 'crypto';
 
 /**
  * Returns a bigint array from a hex string array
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/tss/eddsa/utils.ts
```diff
@@ -0,0 +1,20 @@
+import crypto from 'crypto';
+
+/**
+ * @param seed - used to construct derivation path deterministically
+ * @param isMaster - if set, path starts with prefix `m/`
+ * @return path `(m/)/999999/a/b` where `a` and `b` are 7-byte pseudorandom numbers based on seed
+ */
+export function getDerivationPath(seed: string, isMaster = true): string {
```
zahin-mohammad, line 8: Nothing here seems eddsa specific, can this be a generic util instead? Or conversely can we verify that the logic is the same or different with ecdsa?

modules/sdk-core/src/bitgo/tss/utils.ts
```diff
@@ -0,0 +1,20 @@
+import crypto from 'crypto';
+
+/**
+ * @param seed - used to construct derivation path deterministically
+ * @param isMaster - if set, path starts with prefix `m/`
+ * @return path `(m/)/999999/a/b` where `a` and `b` are 7-byte pseudorandom numbers based on seed
+ */
+export function getDerivationPath(seed: string, isMaster = true): string {
```
zahin-mohammad, line 8: Please add tests for this function.

zahin-mohammad (changes requested):

modules/sdk-core/src/bitgo/tss/eddsa/utils.ts
zahin-mohammad: Just want to avoid devs re-creating this function and introducing differences in logic.

modules/sdk-lib-mpc/src/util.ts
```diff
@@ -137,3 +138,22 @@ export async function randomPositiveCoPrimeLessThan(x: bigint): Promise<bigint>
...
+
+/**
+ * @param seed - used to construct derivation path deterministically
+ * @param isMaster - if set, path starts with prefix `m/`
+ * @return path `(m/)/999999/a/b` where `a` and `b` are 7-byte pseudorandom numbers based on seed
+ */
+export function getDerivationPath(seed: string, isMaster = true): string {
+  function sha256(input: crypto.BinaryLike): Buffer {
+    return crypto.createHash('sha256').update(input).digest();
+  }
```
zahin-mohammad, line 150: Can we move this function outside? There is no reason to re-create this function every time we call `getDerivationPath`

zahin-mohammad: It can remain a a private/un-exported function, but define outside of `getDerivationPath`

modules/sdk-lib-mpc/test/unit/util.ts
```diff
@@ -75,4 +81,10 @@ describe('mpc utils', () => {
       }
     });
   });
+
+  it('should get derivation path from a provided seed', function () {
+    const seed = '1';
+    const derivationPath = getDerivationPath(seed);
+    assert.strictEqual(derivationPath, 'm/999999/163767512/266960264');
+  });
```
zahin-mohammad, line 89: Missing tests with `isMaster=false`. 

Also missing tests around some edge cases. What happens when `seed` is an empty string?

zahin-mohammad (changes requested):

zahin-mohammad (approved):

PR #3855, opened 2023-08-28
fix(sdk-lib-mpc): fix 0 values attack on range proof

zhongxishen:
TICKET: HSM-159


aarvay (commented):
flushing due to lack of context. also unsure why custody-exp is part of the codeowners

modules/sdk-lib-mpc/package.json
```diff
@@ -6,7 +6,7 @@
   "types": "./dist/src/index.d.ts",
   "scripts": {
     "test": "yarn unit-test",
-    "unit-test": "nyc -- mocha",
+    "unit-test": "nyc -- mocha --recursive test",
```
zahin-mohammad, line 9: Could add context on to why we need these flags + the etension change in the .mocharc?

zahin-mohammad (commented):
typescript changes lgtm, just had a comment on the config changes

modules/sdk-lib-mpc/package.json
zhongxishen: `spec: ['test/unit/*.ts']` didn't run tests under sub-directories of `unit/`.

it's after the same change from #3807

zahin-mohammad: I think you can just change the spec to be. Is there a reason to have `.js` files here?
```
['test/unit/**/*.ts']
```

zahin-mohammad: WIth this the `--recursive` flag might be redundant?

zhongxishen: I don't disagree with `['test/unit/**/*.ts']`, but with #3807 already in `master`, it's better to be consistent IMO.

johnoliverdriscoll (approved):

PR #3879, opened 2023-09-03
fix(statics): fix avaxc bridged asset names

hitansh-madan:
Ticket: WIN-329

Change token names for AVAXC bridged tokens since mongo doesn't support keys with period symbol


mukeshsp (approved):

aarvay (approved):

mmcshinsky-bitgo:
Commit needed to be marked as a breaking change in the commit footer.

PR #3885, opened 2023-09-05
feat(sdk-coin-ada): add consolidation method for ada recovery

alia-bitgo:
TICKET: WP-449

### DIFF
--- a/modules/sdk-coin-ada/src/ada.ts
+++ b/modules/sdk-coin-ada/src/ada.ts
@@ -29,6 +29,8 @@ import * as request from 'superagent';
 import BigNumber from 'bignumber.js';
 import { getDerivationPath } from '@bitgo/sdk-lib-mpc';
 
+export const DEFAULT_SCAN_FACTOR = 20; // default number of receive addresses to scan for funds
+
 export interface TransactionPrebuild {
   txHex: string;
 }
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-ada/src/ada.ts
```diff
@@ -394,161 +416,233 @@ export class Ada extends BaseCoin {
...
+        recoveryDestination: baseAddress,
+        seed: params.seed,
+        index: i,
+      };
+
+      let recoveryTransaction;
+      try {
+        recoveryTransaction = await this.recover(recoverParams);
+      } catch (e) {
+        if (e.message == 'Did not find address with funds to recover') {
```
rlutzbitgo, line 594: nit: could probably use === here

modules/sdk-coin-ada/src/ada.ts
```diff
@@ -394,161 +416,233 @@ export class Ada extends BaseCoin {
...
-        if (!params.walletPassphrase) {
-          throw new Error('missing wallet passphrase');
+    const addressParams = {
+      bitgoKey: params.bitgoKey,
+      index: 0,
+      seed: params.seed,
+    };
+    const { address: baseAddress } = await this.getAdaAddressAndAccountId(addressParams);
+    const consolidationTransactions: any[] = [];
+    for (let i = startIdx; i < endIdx; i++) {
```
rlutzbitgo, line 579: I think there's some argument made to change this to be inclusive so that like, first run I put in 1, 20, next run 21, 40, etc. I'm fine with it though

rlutzbitgo (dismissed):

modules/sdk-coin-ada/src/ada.ts
alia-bitgo: since tron is exclusive too right now, ill leave it as is. but good to note and follow up on later, easy change.

rlutzbitgo (approved):

PR #3887, opened 2023-09-06
feat(unspents): add ZCash ZIP-0317 fee calculation

OttoAllmendinger:
Issue: BTC-404


davidkaplanbitgo (approved):
Zcash is such a bloody mess

PR #3891, opened 2023-09-06
feat(sdk-coin-dot): add lastScanIndex for consolidation support

rlutzbitgo:
TICKET: EA-1324


alia-bitgo:
For the `should build signed consolidation recoveries` test in dot.ts, we should add in a line: `res.lastScanIndex.should.equal(3);`

alia-bitgo (commented):
For the `should build signed consolidation recoveries` test in dot.ts, we should add in a line: `res.lastScanIndex.should.equal(3);`

alia-bitgo (approved):

PR #3899, opened 2023-09-08
docs(root): add an example to generate nonce accounts

alebusse:
Added an example with a script on how to generate multiple nonce accounts for solana WRW recovery

WP-604

TICKET: WP-604


examples/ts/sol/utils/nonce-account-creation/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "dependencies": {
+    "@bitgo/sdk-coin-sol": "2.15.0",
+    "@bitgo/statics": "24.0.0",
+    "@solana/web3.js": "1.29.2"
+  },
+  "devDependencies": {
+    "ts-node": "10.9.1",
+    "typescript": "5.1.6"
+  }
+}
```
mmcshinsky-bitgo, line 18: Dependencies pull from root. You should be able to run any example without this.

alia-bitgo (approved):

PR #3905, opened 2023-09-12
fix(sdk-lib-mpc): fix collision in hash values for dlog proofs

islamaminBitGo:
TICKET: HSM-70

BREAKING CHANGE: old challenges will fail this new verification.


modules/sdk-lib-mpc/src/tss/ecdsa/rangeproof.ts
```diff
@@ -104,6 +104,39 @@ export async function generateNtilde(bitlength = minModulusBitLength): Promise<D
...
+ */
+function generateNIZKChallenge(ntilde: DeserializedNtilde, alpha: bigint[]): Buffer {
+  if (alpha.length !== ITERATIONS) {
+    throw Error(`List of alpha values has to be of length ${ITERATIONS}`);
+  }
+  const modulus_bytes: number = Math.ceil(bitLength(ntilde.ntilde) / 8);
+  const modulus_bytes_buf: Buffer = bigIntToBufferBE(BigInt(modulus_bytes));
+  let msgToHash: Buffer = Buffer.concat([
+    modulus_bytes_buf,
+    bigIntToBufferBE(ntilde.h1, modulus_bytes),
```
zhongxishen, line 121: I don't think it's of any particular concern here, but `bigIntToBufferBE` pads up to `minBytes` if the input is small/short, however it doesn't truncate if input is longer.  Just something to keep in mind.

PR #3918, opened 2023-09-19
Revert sol

noel-bitgo:
getting this error when staking SOL `Error processing Instruction 2: incorrect program id for instruction` 
reverting the recent SOL changes to rule that out


margueriteblair (approved):
Looks good, can we just put a description in the PR why we're reverting?

PR #3921, opened 2023-09-20
fix(sdk-api): drop `superagent-proxy` dependency

bitgopatmcl:
See ticket for more info.

[superagent-proxy](https://www.npmjs.com/package/superagent-proxy) hasn't had a release in two years. It appears to be a [thin wrapper around proxy-agent](https://www.npmjs.com/package/proxy-agent), which is more actively maintained. This change drops the unmaintained dependency in favor of just implementing `proxy-agent` directly. This ends up being a major version bump, so the [api is slightly different than what is implemented by superagent-proxy](https://github.com/TooTallNate/superagent-proxy/blob/master/index.js#L71-L74), but it should behave in a similar manner.

TICKET: WP-559

### DIFF
--- a/modules/sdk-api/package.json
+++ b/modules/sdk-api/package.json
@@ -12,7 +12,7 @@
     }
   },
   "browser": {
-    "superagent-proxy": false
+    "proxy-agent": false
   },
   "scripts": {
     "build": "yarn tsc --build --incremental --verbose .",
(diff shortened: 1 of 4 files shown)


theMobiusStrip (approved):
LGTM

alebusse (approved):

PR #3922, opened 2023-09-20
fix(sdk-core): do not sign txRequest full with PA

zahin-mohammad:
Previously we had a check against `mustUseTxRequestFull` before we checked if we should skip signing in the preBuildAndTransaction function. This works fine for ecdsa, but fails for eddsa hot wallets that are using apiVersion === full since they also support lite. 

Reading through the code, we shouldn't need to restrict flows if a coin supports only txRequestFull, instead we should base the flows based on what version of the txRequest we are dealing with. 

TICKET: WP-664

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -1782,6 +1782,12 @@ export class Wallet implements IWallet {
       error.code = 'recipients_not_allowed_for_fillnonce_and_acceleration_tx_type';
       throw error;
     }
+    const supportedTxRequestVersions = this.tssUtils?.supportedTxRequestVersions() || [];
+    if (params.apiVersion && !supportedTxRequestVersions.includes(params.apiVersion)) {
+      throw new Error(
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1801,15 +1807,8 @@ export class Wallet implements IWallet {
     }
 
     let txPrebuildQuery: Promise<PrebuildTransactionResult | string>;
-    const supportedTxRequestVersions = this.tssUtils?.supportedTxRequestVersions() || [];
-    const mustUseTxRequestFull = supportedTxRequestVersions.length === 1 && supportedTxRequestVersions.includes('full');
-
-    if (
-      // verify the wallet must use txRequest Full api and must rebuild the tx before submitting
-      mustUseTxRequestFull &&
```
zahin-mohammad, line 1809: Not sure why this was specific to txRequest full 

cc @mohammadalfaiyazbitgo

zahin-mohammad:
Please merge if this is approved and the build is green.

alebusse (approved):

PR #3947, opened 2023-09-29
feat(key-card): add keycard generation for cold tss key creation

alebusse:
added keycard generation for cold tss key creation

WP-722

TICKET: WP-722


modules/key-card/src/generateQrData.ts
```diff
@@ -1,19 +1,19 @@
...
+  backupKeychain?: Keychain;
   // The name of the 3rd party provider of the backup key if neither the user nor BitGo stores it
   backupKeyProvider?: string;
   // The key id of the backup key, only used for cold keys
   backupMasterKey?: string;
   // The BitGo keychain as it is returned from the BitGo API upon creation
   bitgoKeychain: Keychain;
   // The coin of the wallet that was/ is about to be created
-  coin: Readonly<BaseCoin>;
+  coin?: Readonly<BaseCoin>;
```
abanka27, line 16: Instead of marking fields as optional, can we instead make the params be a union of accepted types?:
```
type GenerateQrDataParams = GenerateQrDataForKeychainType1Params | GenerateQrDataForKeychainType2Params ...
```

modules/key-card/src/generateQrData.ts
```diff
@@ -149,6 +151,9 @@ export function generateQrData({
   userKeychain,
   userMasterKey,
 }: GenerateQrDataParams): QrData {
+  assert(userKeychain, 'userKeychain is required');
+  assert(backupKeychain, 'backupKeychain is required');
+  assert(coin, 'coin is required');
```
abanka27, line 156: exactly my point as above -- if there are fields expected by one flow and not the other, we should let typescript/static checking do the majority of the work for us

abanka27 (changes requested):

modules/key-card/src/generateQrData.ts
zahin-mohammad: +1, this makes it easier to identify valid flows.

modules/key-card/src/generateParamsForKeyCreation.ts
```diff
@@ -0,0 +1,43 @@
...
+  assert(bitgoKeychain.commonKeychain, 'bitgoKeychain.commonKeychain is required');
+  return {
+    walletLabel,
+    keyCardImage,
+    curve,
+    qrData: {
+      user: {
+        title: 'A: Common Keychain',
+        data: bitgoKeychain.commonKeychain,
+        description: 'This is the common pub which is the equivalent of xpub (public key)\r\nof the key generated',
```
zahin-mohammad, line 19: Just checking, did product/technical-writers create these copies?

zahin-mohammad (approved):

modules/key-card/src/generateParamsForKeyCreation.ts
alebusse: Yes, Akshay did the copy and corrections.

abanka27 (approved):

PR #3958, opened 2023-10-05
feat(statics): add Berachain & Haqq

hitansh-madan:
Ticket: WIN-458


modules/statics/src/coins.ts
```diff
@@ -979,6 +979,46 @@ export const coins = CoinMap.fromCoins([
...
+  ),
+  account(
+    '02eced2c-cf1d-4660-832c-858685ae7107',
+    'tislm',
+    'Testnet Islamic Coin',
+    Networks.test.islm,
+    18,
+    UnderlyingAsset.ISLM,
+    BaseUnit.ISLM,
+    [...COSMOS_SIDECHAIN_FEATURES]
```
ajays97, line 1020:
```suggestion
    COSMOS_SIDECHAIN_FEATURES
```
nit: Can we do this since we are not spreading `COSMOS_SIDECHAIN_FEATURES` for the coins above this? For consistent readability.

hitansh-bitgo: Ooh, missed that. Fixed.

modules/statics/src/base.ts
```diff
@@ -1376,6 +1380,8 @@ export enum BaseUnit {
   KAVA = 'ukava',
   CORE = 'ucore',
   TESTCORE = 'utestcore', // Coreum testnet uses different name for native coin
+  BERA = 'abera',
+  ISLM = 'aISLM',
```
dpkjnr, line 1384: why this is in capital case? all others are in small case

hitansh-bitgo: That's just the format expected by this particular chain, so we'll need it that way

alebusse (approved):

PR #3968, opened 2023-10-09
feat(sdk-coin-islm): add Islamic Coin

hitansh-madan:
Ticket: WIN-490

- adds support for different/custom pubkey types (e.g. `ethsecp256k1`) rather than cosmos default `secp256k1`
- adds support for Islamic coin


modules/abstract-cosmos/src/cosmosCoin.ts
```diff
@@ -382,8 +382,8 @@ export class CosmosCoin extends BaseCoin {
     const MESSAGE = Buffer.from(txHex, 'hex');
 
     const [signA, signB] = [
-      MPC.sign(MESSAGE, signCombineOne.oShare, signCombineTwo.dShare, createHash('sha256')),
-      MPC.sign(MESSAGE, signCombineTwo.oShare, signCombineOne.dShare, createHash('sha256')),
+      MPC.sign(MESSAGE, signCombineOne.oShare, signCombineTwo.dShare, this.getHashFunction()),
```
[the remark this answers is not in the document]
hitansh-bitgo: I think it wouldn't, since this uses `keccak256`. And it seems in OVC we default to `keccak256` unless explicitly [added here](https://github.com/BitGo/offline-vault-console/blob/d22166c5384dfd5fb4aeac7c77a54430615a24b9/frontend/src/pkg/bitgo/utils.ts#L182C1-L182C1)

modules/account-lib/package.json
```diff
@@ -42,6 +42,7 @@
     "@bitgo/sdk-coin-hash": "^1.4.11",
     "@bitgo/sdk-coin-hbar": "^1.5.11",
     "@bitgo/sdk-coin-injective": "^1.4.11",
+    "@bitgo/sdk-coin-islm": "^1.0.0",
```
mmcshinsky-bitgo, line 45: Unless there is a specific use case for using this in account-lib elsewhere, let's remove it.

hitansh-bitgo: It's needed for wallet-platform

Vijay-Jagannathan (commented):
I think both test code and service code can be squashed into a single commit. If we do really need to split them, then we'll have the rename the test commits to be prefixed with `test` instead

alebusse (dismissed):
Changes outside of sdk-coin-islm look good too me.

modules/sdk-coin-islm/src/lib/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const validDenoms = ['aISLM'];
```
DinshawKothari, line 1: There must be other denoms as well, like ISLMT as seen here -
https://testnet.ping.pub/haqq/tx/87C5D925D599D1E487523E1295566475BCDE7F94846A50B32413B4580497C6A3
Can you add all denoms. 
nit: should be in lowercase.

modules/sdk-coin-islm/src/islm.ts
```diff
@@ -0,0 +1,72 @@
...
+    return new KeyPair({ pub: publicKey });
+  }
+
+  /** @inheritDoc **/
+  getAddressFromPublicKey(publicKey: string): string {
+    return new KeyPair({ pub: publicKey }).getAddress();
+  }
+
+  /** @inheritDoc **/
+  getHashFunction(): Hash {
```
DinshawKothari, line 69: getAccountDetails() should be also implemented here like injective.

modules/sdk-coin-islm/src/lib/constants.ts
hitansh-bitgo: Oh, ISLMT is just how the explorer shows it, the denom in tx json still is aISLM. The denom is expected in this format on chain so lowercase wouldn't have worked.

modules/sdk-coin-islm/src/islm.ts
hitansh-bitgo: updated

DinshawKothari (approved):

alebusse (approved):

PR #3971, opened 2023-10-09
feat(sdk-coin-arbeth): add arbeth sdk skeleton

gianchandania:
This PR adds sdk-coin-arbeth module 

WIN-694
TICKET: WIN-694

### DIFF
--- a/modules/account-lib/src/index.ts
+++ b/modules/account-lib/src/index.ts
@@ -62,6 +62,9 @@ export { Stx };
 import { AlgoLib as Algo } from '@bitgo/sdk-coin-algo';
 export { Algo };
 
+import * as Arbeth from '@bitgo/sdk-coin-arbeth';
+export { Arbeth };
+
 import * as Atom from '@bitgo/sdk-coin-atom';
 export { Atom };
 
@@ -165,6 +168,8 @@ const coinBuilderMap = {
   tzeta: Zeta.TransactionBuilderFactory,
   core: Core.TransactionBuilderFactory,
   tcore: Core.TransactionBuilderFactory,
+  arbeth: Arbeth.TransactionBuilder,
+  tarbeth: Arbeth.TransactionBuilder,
 };
 
 /**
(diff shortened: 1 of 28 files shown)


modules/account-lib/src/index.ts
```diff
@@ -62,6 +62,9 @@ export { Stx };
 import { AlgoLib as Algo } from '@bitgo/sdk-coin-algo';
 export { Algo };
 
+import * as Arbeth from '@bitgo/sdk-coin-arbeth';
```
mmcshinsky-bitgo, line 65: Unless there is a specific use case for using this in account-lib elsewhere, let's remove it.

gianchandania: We are using this getBuilder [method](https://github.com/BitGo/BitGoJS/pull/3971/files#diff-9189ca44944c5c699c945493c6c19b77b239dab907b726f164eb76fcd976a8a4R181) in wallet platform, so we will have to add this coin here

modules/sdk-coin-arbeth/src/arbeth.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  /** @inheritDoc */
+  supportsTss(): boolean {
+    return true;
+  }
+
+  /** @inheritDoc */
+  getMPCAlgorithm(): MPCAlgorithm {
+    return 'ecdsa';
+  }
```
sachushaji, line 28: I suppose this should be default for all eth chains

sachushaji (approved):
LGTM, nice work !!!

modules/sdk-coin-arbeth/src/arbeth.ts
gianchandania: I will add this in the new class that we discussed about abstractEthLikeTssCoin because currently abstractEthLikeCoin class doesn't have it and I think it won't make sense to add it there

alebusse (approved):
Changes outside the module sdk-coin-arbeth look good

linyah9912 (approved):

PR #3977, opened 2023-10-11
feat(statics): add holesky config

dpkjnr:
Ticket: EA-802

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -438,6 +438,19 @@ class Goerli extends Testnet implements EthereumNetwork {
   forwarderImplementationAddress = '0x80d5c91e8cc21df69fc4d64f21dc2d83121c3999';
 }
 
+class Holesky extends Testnet implements EthereumNetwork {
+  name = 'Holesky';
+  family = CoinFamily.ETH;
+  explorerUrl = 'https://holesky.beaconcha.in/tx/';
+  accountExplorerUrl = 'https://holesky.beaconcha.in/address/';
+  blockExplorerUrl = 'https://holesky.beaconcha.in/block/';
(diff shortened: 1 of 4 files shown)


modules/statics/src/networks.ts
```diff
@@ -438,6 +438,19 @@ class Goerli extends Testnet implements EthereumNetwork {
   forwarderImplementationAddress = '0x80d5c91e8cc21df69fc4d64f21dc2d83121c3999';
 }
 
+class Holesky extends Testnet implements EthereumNetwork {
+  name = 'Holesky';
+  family = CoinFamily.ETH;
+  explorerUrl = 'https://holesky.beaconcha.in/tx/';
+  accountExplorerUrl = 'https://holesky.beaconcha.in/address/';
+  blockExplorerUrl = 'https://holesky.beaconcha.in/block/';
+  // https://chainlist.org/chain/17000
```
DinshawKothari, line 447: nit: can add this as Ref - https://github.com/eth-clients/holesky#metadata

modules/statics/test/unit/coins.ts
```diff
@@ -148,6 +148,14 @@ const custodyFeatures: Record<string, { features: CoinFeature[] }> = {
       CoinFeature.CUSTODY_BITGO_FRANKFURT,
     ],
   },
+  hteth: {
```
DinshawKothari, line 151: nit: can we add hteth above only with gteth. Or the idea is to remove gteth later.

DinshawKothari (approved):

aarvay (approved):

modules/statics/src/networks.ts
dpkjnr: yeah, that looks better. I will do it along with the next PR

PR #3978, opened 2023-10-12
add abstract-eth classes

gianchandania:
This PR adds AbstractEthLikeMPCCoin and EthLikeMPCToken classes which will have all the common methods used by Ethlike MPC coins

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeMPCCoin.ts
+++ b/modules/abstract-eth/src/abstractEthLikeMPCCoin.ts
@@ -1,0 +1,17 @@
+/**
+ * @prettier
+ */
+import { MPCAlgorithm } from '@bitgo/sdk-core';
+import { AbstractEthLikeCoin } from './abstractEthLikeCoin';
+
+export abstract class AbstractEthLikeMPCCoin extends AbstractEthLikeCoin {
+  /** @inheritDoc */
+  supportsTss(): boolean {
+    return true;
+  }
+
+  /** @inheritDoc */
+  getMPCAlgorithm(): MPCAlgorithm {
+    return 'ecdsa';
+  }
+}
--- a/modules/abstract-eth/src/ethLikeMPCToken.ts
+++ b/modules/abstract-eth/src/ethLikeMPCToken.ts
@@ -1,0 +1,17 @@
+/**
+ * @prettier
+ */
+import { MPCAlgorithm } from '@bitgo/sdk-core';
+import { EthLikeToken } from './ethLikeToken';
+
+export abstract class EthLikeMPCToken extends EthLikeToken {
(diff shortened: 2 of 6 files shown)


modules/abstract-eth/src/ethLikeToken.ts
```diff
@@ -4,14 +4,14 @@
...
-import { AbstractEthLikeCoin } from './abstractEthLikeCoin';
+import { AbstractEthLikeTssCoin } from './abstractEthLikeTssCoin';
 import { TransactionBuilder as EthTransactionBuilder, TransactionPrebuild } from '@bitgo/sdk-coin-eth';
 
 export type CoinNames = {
   [network: string]: string;
 };
 
-export class EthLikeToken extends AbstractEthLikeCoin {
+export class EthLikeToken extends AbstractEthLikeTssCoin {
```
sachushaji, line 14: We would need an EthLikeTokenTss? will this mean all the token implementation even for multisig will have tss enabled?

modules/abstract-eth/src/abstractEthLikeTssCoin.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @prettier
+ */
+import { MPCAlgorithm } from '@bitgo/sdk-core';
+import { AbstractEthLikeCoin } from './abstractEthLikeCoin';
+
+export abstract class AbstractEthLikeTssCoin extends AbstractEthLikeCoin {
```
mohammadalfaiyazbitgo, line 7:
```suggestion
export abstract class AbstractEcdsaCoin extends AbstractEthLikeCoin {
```

modules/abstract-eth/src/ethLikeToken.ts
mohammadalfaiyazbitgo: +1 on above comment

mohammadalfaiyazbitgo (commented):
Using ECDSA in the class name is more descriptive than using `EthLike` imo, and do these changes need to be extended to `polygon` as well?

modules/abstract-eth/src/ethLikeToken.ts
gianchandania: added EthLikeMPCToken class

gianchandania:
> Using ECDSA in the class name is more descriptive than using `EthLike` imo, and do these changes need to be extended to `polygon` as well?

Actually this class will be specific to EthLike coins i.e. EVM compatible coins only like arbitrum, optimism, etc. Using ECDSA in the name will give an impression that it is generic because there can be other ecdsa coins as well like cosmos, hence the name

sachushaji (approved):

PR #3982, opened 2023-10-12
feat(sdk-core): whitelist distributed custody params

davidkaplanbitgo:
We need to be able to pass in `isDistributedCustody` in for creating a bitgoKey.

TICKET: BTC-396


margueriteblair (approved):
lgtm

PR #3984, opened 2023-10-12
feat(sdk-core): add distributed custody params to wallet creation

davidkaplanbitgo:
TICKET: BTC-396

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallets.ts
@@ -173,6 +173,9 @@ export class Wallets implements IWallets {
    * @param params.disableKRSEmail
    * @param params.walletVersion
    * @param params.multisigType optional multisig type, 'onchain' or 'tss' or 'blsdkg'; if absent, we will defer to the coin's default type
+   * @param params.isDistributedCustody optional parameter for creating bitgo key. This is only necessary if you want to create
+   *                                    a distributed custody wallet. If provided, you must have the enterprise license and pass in
+   *                                    `params.enterprise` into `generateWallet` as well.
    * @returns {*}
    */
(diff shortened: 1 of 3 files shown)


lcovar (dismissed):

alebusse:
Should add a check to only allow `multisig` type for `distributedCustody` somewhere?

davidkaplanbitgo:
> Should add a check to only allow `multisig` type for `distributedCustody` somewhere? Also, it must be a `isCold===true`, right?

It is true, but those checks are done on the API. Would be good to keep those checks in one place I think

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -254,6 +257,11 @@ export class Wallets implements IWallets {
       return this.generateMpcWallet({ multisigType: 'blsdkg', label, passphrase });
     }
 
+    // Handle distributed custody
+    if (params.isDistributedCustody && !params.enterprise) {
+      throw new Error('must provide enterprise when creating distributed custody wallet');
+    }
```
rushilbg, line 263: Why to make a network call when it can be decided here not to proceed with creation if not a cold wallet?

alebusse: Yes, you should check that everything is correct before hand, otherwise if i recall correctly it will create the keys and then fail to create the wallet.

davidkaplanbitgo: that makes sense - updated the PR

alebusse (changes requested):
Could you please add tests too?

lcovar (approved):
Personally ok with adding test as a follow up @alebusse as we have a tight deadline. Test nock responses anyway so we're not truly testing this behavior.

alebusse (approved):

PR #3988, opened 2023-10-13
feat(statics): add more sol token

ewangbitgo:
TICKET: EA-1955

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -1324,6 +1324,37 @@ export enum UnderlyingAsset {
   // Cardano Token
   adaTestnetToken = 'temporary-placeholder',
 
+  // solana token
+  '3uejh-usdc' = '3uejh-usdc',
+  'avax-usdc' = 'avax-usdc',
+  'bop-usdc' = 'bop-usdc',
+  'elu-usdt' = 'elu-usdt',
+  'fida-usdc' = 'fida-usdc',
+  'fida-usdt' = 'fida-usdt',
+  'ftt-ftt' = 'ftt-ftt',
+  'link-usdc' = 'link-usdc',
+  'lqid-usdc' = 'lqid-usdc',
+  'maticpo-usdc' = 'maticpo-usdc',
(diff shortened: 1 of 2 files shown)


modules/statics/src/coins.ts
```diff
@@ -10288,6 +10622,7 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.SLND,
     AccountCoin.DEFAULT_FEATURES
   ),
+
```
mishkinf-bitgo, line 10625: not a big issue but this space is not necessary

mishkinf-bitgo (dismissed):
lgtm

modules/statics/src/base.ts
```diff
@@ -1324,6 +1324,42 @@ export enum UnderlyingAsset {
...
+  'tuAtlas' = 'tuAtlas',
+  'tuSAMO' = 'tuSAMO',
+  'tuDFL' = 'tuDFL',
+  'tuLIKE' = 'tuLIKE',
+  'tuPOLIS' = 'tuPOLIS',
+  'tuSRM' = 'tuSRM',
+  'tuREAL' = 'tuREAL',
+  'tuMBS' = 'tuMBS',
+  'tuGENE' = 'tuGENE',
+  'tuCOPE' = 'tuCOPE',
```
CapnMigraine, line 1361: Could we use lowercase for all token names? these combinations seem a bit arbitrary
Other sol tokens use uppercase for the key and lowercase for the value

modules/statics/src/coins.ts
```diff
@@ -10207,6 +10207,340 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.QCAD,
     AccountCoin.DEFAULT_FEATURES
   ),
+
+  solToken(
+    '170073d0-3641-42d1-b44a-73509e8bfb23',
+    'SOL:SHDW-USDC',
```
CapnMigraine, line 10213: I think these should be all lower case

alebusse (approved):

PR #3991, opened 2023-10-16
chore(root): lock dot libs versions

alebusse:
locked resolution of dot libs to enforce the right version on installation since they have minor changes with breaking changes

WP-837

TICKET: WP-837


noel-bitgo (approved):

margueriteblair (approved):

package.json
```diff
@@ -62,6 +62,9 @@
     "@types/keyv": "3.1.4",
     "@types/react": "17.0.24",
     "@types/react-dom": "17.0.16",
+    "@polkadot/api": "10.9.1",
+    "@polkadot/types": "10.9.1",
+    "@substrate/txwrapper-core": "7.0.1",
```
mmcshinsky-bitgo, line 67: `@substrate/txwrapper-substrate` needs to be `7.0.1` as well

mmcshinsky-bitgo: Also need to confirm that packages like `@bitgo-beta/sdk-doin-dot` when installed don't include the unintended versions.

PR #3994, opened 2023-10-17
feat(sdk-coin-ada): add support for multi asset in tx building

Vijay-Jagannathan:
Ticket: WIN-736

### Context

 - Previously, if a receive address were to receive some cardano token as part of a transaction, it wouldn't be possible to either use this token (because we don't support), or send it to the root address (because we didn't support token consolidation)
 - Because of this, consolidation was failing for even ADA, since some UTXOs had both tokens and ADA in them
 - This PR adds support for processing tokens (present as asset_list inside the actual UTXO data) while initiating consolidation from the receive address. This would ensure that the root address can receive both ADA as well as the token (though token won't be spendable as we don't support) when consolidating


modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -10,11 +10,12 @@ import {
   TransactionType,
   UtilsError,
 } from '@bitgo/sdk-core';
-import { Transaction, TransactionInput, TransactionOutput, Withdrawal } from './transaction';
+import { AssetList, Transaction, TransactionInput, TransactionOutput, Withdrawal } from './transaction';
 import { KeyPair } from './keyPair';
 import util from './utils';
 import * as CardanoWasm from '@emurgo/cardano-serialization-lib-nodejs';
 import { BigNum } from '@emurgo/cardano-serialization-lib-nodejs';
+import { Buffer } from 'buffer';
```
margueriteblair, line 18: What is this Buffer import? We can't use native JS Buffer?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -255,6 +264,30 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+
+      // should coin value be 0 since this output is related to token ?
+      const zeroValue = '0';
+      txOutputAmountBuilder = txOutputAmountBuilder.with_coin_and_asset(
+        CardanoWasm.BigNum.from_str(zeroValue),
+        multiAsset
+      );
+
+      const txOutput = txOutputAmountBuilder.build();
+      outputs.add(txOutput);
```
margueriteblair, line 288: Would `outputs.add(txOutput)` work if there were multiple outputs? Or is it already assuming there's multiple outputs?

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
Vijay-Jagannathan: got auto added I believe, when I was referencing code from stack overflow; removing

modules/sdk-coin-ada/src/lib/transactionBuilder.ts
Vijay-Jagannathan: yes, had previously assumed a single assetList; just pushed the changes for multi asset list, which should add each output separately

modules/sdk-coin-ada/src/lib/transaction.ts
```diff
@@ -15,9 +15,15 @@ export interface TransactionInput {
...
 
+export interface AssetList {
+  policy_id: string;
+  asset_name: string;
+  quantity: string;
+}
 export interface TransactionOutput {
   address: string;
   amount: string;
+  multiAssets?: CardanoWasm.MultiAsset;
```
margueriteblair, line 26: Can we make this a non CardanoWasm type? Define our own `MultiAsset` interface?

modules/sdk-coin-ada/src/lib/transaction.ts
```diff
@@ -15,9 +15,15 @@ export interface TransactionInput {
   transaction_index: number;
 }
 
+export interface AssetList {
```
margueriteblair, line 18: Do we want this to be called `AssetList` ? Or would `Asset` make more sense?

modules/sdk-coin-ada/src/lib/transaction.ts
```diff
@@ -140,6 +146,7 @@ export class Transaction extends BaseTransaction {
       result.outputs.push({
         address: output.address().to_bech32(),
         amount: output.amount().coin().to_str(),
+        multiAssets: output.amount().multiasset(),
```
margueriteblair, line 149: What happens when the multi asset doesn't exist on the output?

modules/sdk-coin-ada/test/unit/transactionBuilder.ts
```diff
@@ -39,6 +39,123 @@ describe('ADA Transaction Builder', async () => {
...
+    txData.outputs.length.should.equal(4);
+    txData.outputs[0].address.should.equal(testData.rawTx.outputAddress1.address);
+    txData.outputs[1].address.should.equal(testData.rawTx.outputAddress2.address);
+    txData.outputs[2].address.should.equal(testData.rawTx.outputAddress1.address);
+
+    // token assertion
+    const asset1_expectedAssetName = CardanoWasm.AssetName.new(Buffer.from(asset1_assetName, 'hex'));
+    const asset1_expectedPolicyId = CardanoWasm.ScriptHash.from_bytes(Buffer.from(asset1_policyId, 'hex'));
+    txData.outputs[2].should.have.property('multiAssets');
+    (txData.outputs[2].multiAssets as CardanoWasm.MultiAsset)
```
margueriteblair, line 139: Nice!!!

modules/sdk-coin-ada/src/lib/transaction.ts
Vijay-Jagannathan: missed it; will add an OR condition here

modules/sdk-coin-ada/src/lib/transaction.ts
Vijay-Jagannathan: Asset sounds better, will change

modules/sdk-coin-ada/src/lib/transaction.ts
Vijay-Jagannathan: sure we can change that, but any specific reason why ? originally made it the CardanoWasm type because I'd have to extract out the fields anyway, have a new interface and then construct it back

mukeshsp (approved):

modules/sdk-coin-ada/src/lib/transaction.ts
Vijay-Jagannathan: ^can address this in a followup if needed

PR #3998, opened 2023-10-18
chore: add ecdh keychain creation on login

YPAzevedo:
TICKET: CE-2928

### DIFF
--- a/modules/bitgo/test/unit/bitgo.ts
+++ b/modules/bitgo/test/unit/bitgo.ts
@@ -129,7 +129,12 @@ describe('BitGo Prototype Methods', function () {
       bitgo = TestBitGo.decorate(BitGo, { env: 'mock', microservicesUri: 'https://microservices.uri' } as any);
       const scope = nock(BitGoJS.Environments[bitgo.getEnv()].uri)
         .post('/api/auth/v1/session')
-        .reply(200, { user: 'test@bitgo.com', access_token: 'token12356' });
+        .reply(200, {
+          user: {
+            username: 'test@bitgo.com',
+          },
+          access_token: 'token12356',
+        });
 
       await bitgo.authenticate(authenticateRequest);
       scope.isDone().should.be.true();
@@ -139,7 +144,12 @@ describe('BitGo Prototype Methods', function () {
       bitgo = TestBitGo.decorate(BitGo, { env: 'mock' });
       const scope = nock(BitGoJS.Environments[bitgo.getEnv()].uri)
         .post('/api/auth/v1/session')
-        .reply(200, { user: 'test@bitgo.com', access_token: 'token12356' });
+        .reply(200, {
+          user: {
+            username: 'test@bitgo.com',
+          },
+          access_token: 'token12356',
+        });
 
       await bitgo.authenticate(authenticateRequest);
       scope.isDone().should.be.true();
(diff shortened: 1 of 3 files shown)


modules/sdk-api/src/types.ts
```diff
@@ -275,3 +276,18 @@ export interface RegisterPushTokenOptions {
...
+  encryptedECDHXprv?: string;
+  encryptedToken?: string;
+  // Unit timestamp of expiration
+  expires_at: number;
+  // seconds in which the token will expire from issuance
+  expires_in: number;
+  scope: string[];
+  token_type: string;
+  user: unknown;
+}
```
zahin-mohammad, line 293: Nit: It would be nice to have this as two types/interfaces. One for what the API returns and an extension of that for what the SDK annotates.

modules/sdk-api/src/types.ts
```diff
@@ -88,6 +88,7 @@ export interface AuthenticateOptions {
   forceSMS?: boolean;
   extensible?: boolean;
   forceV1Auth?: boolean;
+  ensureEcdhKeychain?: boolean;
```
zahin-mohammad, line 91: I know the other fields dont have a description, but it would be good to have a small comment here on what this field represents? What happens if its true/false, etc.

modules/bitgo/test/unit/bitgo.ts
```diff
@@ -609,4 +609,42 @@ describe('BitGo Prototype Methods', function () {
       );
     });
   });
+
+  describe('authenticate', function () {
+    it('should calculate a new ECDH sharing secret correctly', async function () {
```
bitgoAaron, line 614: copy pasta title? 
```suggestion
    it('should get or create the ecdhKeychain if ensureEcdhKeychain is set', async function () {
```

YPAzevedo: updating

bitgoAaron (dismissed):
nice job

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -748,10 +750,44 @@ export class BitGoAPI implements BitGoBase {
...
+        settings: {
+          ecdhKeychain: newKeychain.xpub,
+        },
+      });
+      userSettings.settings.ecdhKeychain = newKeychain.xpub;
+      return userSettings.settings;
+    }
+
+    return userSettings.settings;
+  };
```
zahin-mohammad, line 785: Whats the reasoning for this format vs  
```
private async ensureUserEcdhKeychainIsCreated(password: string): Promise<any>  {
```

I'd prefer if they where all the same format if there is no benefit of one vs the other.

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -748,10 +750,44 @@ export class BitGoAPI implements BitGoBase {
...
+        input: hdNode.toBase58(),
+      }),
+    });
+  };
+
+  private updateUserSettings(params: any): Promise<any> {
+    return this.put(this.url('/user/settings', 2)).send(params).result();
+  }
+
+  private ensureUserEcdhKeychainIsCreated = async (password: string): Promise<any> => {
```
zahin-mohammad, line 771:
```suggestion
  private ensureUserEcdhKeychainIsCreated = async (loginPassword: string): Promise<any> => {
```

we use password as a name for wallets as well, its nice to be explicit.

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -748,10 +750,44 @@ export class BitGoAPI implements BitGoBase {
     this._token = accessToken;
   }
 
+  public createAccountKeychain = async (password: string): Promise<any> => {
```
zahin-mohammad, line 753: Please add a jsDOC

modules/sdk-api/src/bitgoAPI.ts
zahin-mohammad, line 771: Please add a jsDOC

modules/bitgo/test/unit/bitgo.ts
```diff
@@ -609,4 +609,42 @@ describe('BitGo Prototype Methods', function () {
       );
     });
   });
+
+  describe('authenticate', function () {
+    it('should get or create the ecdhKeychain if ensureEcdhKeychain is set', async function () {
```
zahin-mohammad, line 614: Please make these two separate tests:

`should get ecdhKeychain if already set`

`should create the ecdhKeychain if not already is set`

zahin-mohammad (approved):
small nits/requested changes. 

In general changes lgtm, if you are pressed for time please address in a follow up.

modules/sdk-api/src/types.ts
YPAzevedo: i dont think we add anything extra (not super familiar with all the flow), the user settings we just update the field with the new user settings

pranavjain97 (approved):

zahin-mohammad (dismissed):

bitgoAaron (approved):

PR #4002, opened 2023-10-19
fix(sdk-core): fix dc wallet creation

davidkaplanbitgo:
`isDistributedCustody` is a boolean, not a stirng

TICKET: BTC-553


lcovar (approved):

margueriteblair (approved):

modules/sdk-core/src/bitgo/keychain/keychains.ts
```diff
@@ -197,7 +197,6 @@ export class Keychains implements IKeychains {
         'originalPasscodeEncryptionCode',
         'enterprise',
         'derivedFromParentWithSeed',
-        'isDistributedCustody',
```
mmcshinsky-bitgo, line 200: Does https://developers.bitgo.com/api/v2.key.add need to be updated?

PR #4005, opened 2023-10-20
feat: use psbt for prebuild when wallet is distributedCustody

lcovar:
When the wallet is of type distributed custody want to always use 'psbt' as the transaction format.

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -1442,7 +1442,11 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     };
   }
 
-  async getExtraPrebuildParams(buildParams: ExtraPrebuildParamsOptions): Promise<any> {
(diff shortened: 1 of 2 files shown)


davidkaplanbitgo (approved):

margueriteblair (approved):

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -1442,7 +1442,11 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     };
   }
 
-  async getExtraPrebuildParams(buildParams: ExtraPrebuildParamsOptions): Promise<any> {
+  async getExtraPrebuildParams(buildParams: ExtraPrebuildParamsOptions & { wallet: Wallet }): Promise<any> {
```
mmcshinsky-bitgo, line 1445: Is this new functionality covered in the transaction portion of the developer portal api docs?

mmcshinsky-bitgo, line 1445: Should add `wallet` to the `ExtraPrebuildParamsOptions` type.

PR #4016, opened 2023-10-25
fix(utxo-lib): check prevout script against computed for non-taproot inputs

OttoAllmendinger:
Make sure that we are actually signing the right script.

Taproot cases will follow.

Issue: BTC-428

### DIFF
--- a/modules/utxo-lib/src/bitgo/signature.ts
+++ b/modules/utxo-lib/src/bitgo/signature.ts
@@ -8,6 +8,7 @@ import {
   createOutputScript2of3,
   createOutputScriptP2shP2pk,
   createSpendScriptP2tr,
+  getOutputScript,
   ScriptType,
   ScriptType2Of3,
   scriptType2Of3AsPrevOutType,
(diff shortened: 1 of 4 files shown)


modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -102,6 +103,21 @@ export function getSignatureVerifications<TNumber extends number | bigint>(
     }
   }
 
+  if (
+    parsedScript.scriptType !== 'taprootKeyPathSpend' &&
+    parsedScript.scriptType !== 'taprootScriptPathSpend' &&
```
rushilbg, line 108: Don't we have an array of taproot narrowed types already? Let's add it if not.

```
if (!isTaprootScript(parsedScript))

File: modules/utxo-lib/src/bitgo/signature.ts [hunk 2]
```diff
@@ -102,6 +103,21 @@ export function getSignatureVerifications<TNumber extends number | bigint>(
...
+    prevOutputs
+  ) {
+    const prevoutScript = prevOutputs[inputIndex].script;
```
rushilbg, line 111:
```suggestion
    const prevOutScript = prevOutputs[inputIndex].script;
```

modules/utxo-lib/src/bitgo/signature.ts
```diff
@@ -102,6 +103,21 @@ export function getSignatureVerifications<TNumber extends number | bigint>(
...
+
+    const output = getOutputScript(parsedScript.scriptType, parsedScript.pubScript);
+    if (!prevoutScript.equals(output)) {
+      throw new Error(
+        `prevout script ${prevoutScript.toString('hex')} does not match computed script ${output.toString('hex')}`
+      );
```
rushilbg, line 117: We're gonna be computing the script every time we verify the existing sigs?
Although most likely a non-issue, but off the top of your head, how much overhead is this gonna add in the regular signing flow?

OttoAllmendinger: I think it is very small: this sort of check happens in the psbt flow by default without significant impact

modules/utxo-lib/src/bitgo/signature.ts
OttoAllmendinger: unfortunately that won't narrow down the type of `parsedScript` enough


```
export function isParsedScriptTypeTaproot(
  t: ParsedScriptType2Of3
): t is 'taprootKeyPathSpend' | 'taprootScriptPathSpend' {
  return t === 'taprootKeyPathSpend' || t === 'taprootScriptPathSpend';
}
```

image

rushilbg (approved):

PR #4017, opened 2023-10-25
feat(utxo-bin): add new command convertTx

OttoAllmendinger:
- feat(blockapis): add fetchPrevTxBuffers
- feat(utxo-bin): use strict mode
- feat(utxo-bin): add new command convertTx

### DIFF
--- a/modules/utxo-bin/src/commands.ts
+++ b/modules/utxo-bin/src/commands.ts
@@ -21,7 +21,7 @@ import { AddressParser } from './AddressParser';
 import { BaseHttpClient, CachingHttpClient, HttpClient } from '@bitgo/blockapis';
 import { readStdin } from './readStdin';
 import { parseUnknown } from './parseUnknown';
-import { getParserTxProperties } from './ParserTx';
+import { getParserTxProperties, ParserTx } from './ParserTx';
 import { ScriptParser } from './ScriptParser';
 import { stringToBuffer } from './parseString';
 import {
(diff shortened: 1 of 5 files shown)


modules/utxo-bin/src/commands.ts
```diff
@@ -32,16 +32,115 @@ import {
...
+    if (data) {
+      throw new Error(`conflicting arguments`);
+    }
+    data = argv.data;
+  }
+
+  // strip whitespace
+  if (!data) {
+    throw new Error(`no txdata`);
+  }
```
rushilbg, line 115: Possible to move all this in a separate function `validateArguments` or such? This whole function will become easier to reason about then, it's quite long currently.

modules/utxo-bin/src/convertTransaction.ts
```diff
@@ -0,0 +1,60 @@
...
+): Promise<void> {
+  const { format } = params;
+
+  switch (format) {
+    case 'legacy':
+      if (tx instanceof utxolib.bitgo.UtxoTransaction) {
+        throw new Error(`input is already in legacy format`);
+      }
+      if (tx instanceof utxolib.bitgo.UtxoPsbt) {
+        throw new Error(`TODO`);
```
rushilbg, line 42:
```
throw new Error(`not implemented yet`);

File: modules/utxo-bin/src/convertTransaction.ts [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+
+      if (tx instanceof utxolib.bitgo.UtxoTransaction) {
+        const psbt = await legacyToPsbt(params.httpClient, tx);
+        const { outfile = getOutfile(tx, format) } = params;
+        await fs.writeFile(outfile, psbt.toBase64());
+        console.log(`wrote ${outfile}`);
+        return;
+      }
+
+      break;
```
rushilbg, line 58: Redundant?

OttoAllmendinger: improved this case

modules/utxo-bin/src/commands.ts
OttoAllmendinger: split it up into two methods

modules/utxo-bin/src/convertTransaction.ts
OttoAllmendinger: done

rushilbg (approved):

PR #4034, opened 2023-11-01
feat(sdk-coin-sol): add sol token recovery support

alia-bitgo:
Adding in token recovery support for solana. This has been tested e2e for both cold and hot wallets.

TICKET: WP-930

### DIFF
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -5,7 +5,7 @@
 import BigNumber from 'bignumber.js';
 import * as base58 from 'bs58';
 
-import { BaseCoin as StaticsBaseCoin, CoinFamily, coins } from '@bitgo/statics';
+import { BaseCoin as StaticsBaseCoin, CoinFamily, coins, BaseNetwork } from '@bitgo/statics';
 import * as _ from 'lodash';
 import {
   BaseCoin,
@@ -38,6 +38,8 @@ import {
   MPCUnsignedTx,
   MPCSweepRecoveryOptions,
   MPCTxs,
+  OvcInput,
+  OvcOutput,
 } from '@bitgo/sdk-core';
 import { KeyPair as SolKeyPair, Transaction, TransactionBuilder, TransactionBuilderFactory } from './lib';
 import {
@@ -46,6 +48,7 @@ import {
   isValidAddress,
   isValidPrivateKey,
   isValidPublicKey,
+  getSolTokenFromAddress,
 } from './lib/utils';
 import * as request from 'superagent';
 import { getDerivationPath } from '@bitgo/sdk-lib-mpc';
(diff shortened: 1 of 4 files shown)


modules/sdk-coin-sol/src/lib/constants.ts
```diff
@@ -114,3 +114,188 @@ export const ataInitInstructionIndexes = {
 
 export const nonceAdvanceInstruction = 'AdvanceNonceAccount';
 export const validInstructionData = '0a00000001000000';
+
+export const SOL_MAINNET_TOKENS = {
```
zahin-mohammad, line 118: Shouldn't this live in statics instead? 

https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/coins.ts

zahin-mohammad, line 118: Shouldn't this live in statics instead? 

https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/coins.ts

alia-bitgo: I was just trying to make a structure that was easy to check based on name and/or mint address if it was supported in recovery, like how its done for say tron tokens. seemed easier than dealing with the `solToken` structure in that file

rlutzbitgo (dismissed):

modules/sdk-coin-sol/src/lib/constants.ts
zahin-mohammad: But repeating this large list is not desirable. By adding/using statics, it enables other packages to use this token data in a more generic way without havbing to import individual coin modules.


For example:  UI's can ingest this data from the statics lib instead of having to switch between coin modules to get TRON tokens vs SOL tokens.

zahin-mohammad: Also reduces the risk of mis-configed data where statics defines a token one way but this hard coded list has it configured a seperate way.

zahin-mohammad (changes requested):
Replied to the comment, but I am not convinced that hard coding this structre outside of statics is the best way to list out these tokens, maybe theres a constraint I'm not aware of that makes this harder to do. Lets discuss in slack/focus hour?

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -572,6 +600,38 @@ export class Sol extends BaseCoin {
     };
   }
 
+  protected async getTokenAccountsByOwner(pubKey = ''): Promise<[] | TokenAccount[]> {
```
zahin-mohammad, line 603: Should an empty pubkey be valid for this?

zahin-mohammad (approved):
Minor comment, buy WP owned files lgtm

rlutzbitgo (approved):

PR #4040, opened 2023-11-07
fix: address verification for cosmos like coins

hitansh-madan:
Ticket: WIN-1069

Fix failing address verification for addresses with memoId for cosmos like coins

### DIFF
--- a/modules/abstract-cosmos/src/lib/utils.ts
+++ b/modules/abstract-cosmos/src/lib/utils.ts
@@ -701,6 +701,42 @@ export class CosmosUtils implements BaseUtils {
   }
 
   /**
+   * Checks if a cosmos like Bech32 address matches given regular expression and
+   * validates memoId if present
+   * @param {string} address
+   * @param {RegExp} regExp Regular expression to validate the root address against after trimming the memoId
+   * @returns {boolean} true if address is valid
+   */
+  isValidCosmosLikeAddressWithMemoId(address: string, regExp: RegExp): boolean {
+    if (typeof address !== 'string') return false;
+    const addressArray = address.split('?memoId=');
+    if (
+      ![1, 2].includes(addressArray.length) || // should have at most one occurrence of 'memoId='
+      !regExp.test(addressArray[0]) ||
+      (addressArray[1] && !this.isValidMemoId(addressArray[1]))
+    ) {
+      return false;
+    }
+    return true;
+  }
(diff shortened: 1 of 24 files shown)


modules/abstract-cosmos/src/lib/utils.ts
```diff
@@ -700,6 +700,42 @@ export class CosmosUtils implements BaseUtils {
...
   }
 
+  /**
+   * Checks if a cosmos like Bech32 address matches given regular expression and
+   * validates memoId if present
+   * @param {string} address
+   * @param {RegExp} regExp Regular expression to validate the root address against after trimming the memoId
+   * @returns {boolean} true if address is valid
+   */
+  isValidCosmosLikeAddressWithMemoId(address: string, regExp: RegExp): boolean {
```
Vijay-Jagannathan, line 710: seems to be same/similar to the other ones. can this live in a common place like abstract cosmos util since it isn't coin specific

Vijay-Jagannathan (dismissed):

modules/abstract-cosmos/src/lib/utils.ts
Vijay-Jagannathan: so atom is not part of abstract cosmos yet, so that's why it was separate. Resolving

mukeshsp (approved):

modules/sdk-coin-atom/src/lib/utils.ts
```diff
@@ -82,14 +82,50 @@ export class Utils implements BaseUtils {
...
+   * validates memoId if present
+   * @param {string} address
+   * @param {RegExp} regExp Regular expression to validate the root address against after trimming the memoId
+   * @returns {boolean} true if address is valid
+   */
+  isValidCosmosLikeAddressWithMemoId(address: string, regExp: RegExp): boolean {
+    if (typeof address !== 'string') return false;
+    const addressArray = address.split('?memoId=');
+    if (
+      ![1, 2].includes(addressArray.length) || // should have at most one occurrence of 'memoId='
```
DinshawKothari, line 96: any possibility that `address?memoId=0` is considered as `address` only in UI ?

hitansh-bitgo: Not sure, since that parsing is implemented in UI, and this function is only used for validation

PR #4042, opened 2023-11-08
feat(utxo-lib): allow a custom sequenceNumber to be added in tx creation

rushilbg:
This adds support for signaling RBF transactions.

BTC-591

### DIFF
--- a/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
+++ b/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
@@ -36,6 +36,9 @@ import { parseSignatureScript } from '../parseInput';
 import { checkForInput } from 'bip174/src/lib/utils';
 import { ProprietaryKeySubtype, PSBT_PROPRIETARY_IDENTIFIER } from '../PsbtUtil';
 
+/** Final (non-replaceable) */
+export const TX_INPUT_SEQUENCE_NUMBER_FINAL = 0xffffffff;
+
 export interface WalletUnspent<TNumber extends number | bigint = number> extends Unspent<TNumber> {
   chain: ChainCode;
   index: number;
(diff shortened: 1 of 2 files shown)


rushilbg (commented):
Opening it up for preliminary review while thinking of a best way to update the tests.

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -186,11 +188,12 @@ export function updateReplayProtectionUnspentToPsbt(
   }
 }
 
-function addUnspentToPsbt(psbt: UtxoPsbt, id: string): void {
+function addUnspentToPsbt(psbt: UtxoPsbt, id: string, sequenceNumber: number = DEFAULT_TX_INPUT_SEQUENCE_NUMBER): void {
```
OttoAllmendinger, line 191:
```suggestion
function addUnspentToPsbt(psbt: UtxoPsbt, id: string, { sequenceNumber  = DEFAULT_TX_INPUT_SEQUENCE_NUMBER }: { sequenceNumber?: number } = {}): void {
```

more extensible like this

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -360,11 +363,12 @@ export function addWalletUnspentToPsbt(
   /**
    * @deprecated
    */
-  network: Network = psbt.network
+  network: Network = psbt.network,
+  customParams?: { sequenceNumber?: number }
```
OttoAllmendinger, line 367: This would force passing the deprecated parameter. Let's just get rid of it

```suggestion
  params: { sequenceNumber?: number },
```

OttoAllmendinger (changes requested):

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
rushilbg: SDK: https://github.com/BitGo/BitGoJS/pull/4043
WP: https://github.com/BitGo/bitgo-microservices/pull/37560

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -36,6 +36,8 @@ import { parseSignatureScript } from '../parseInput';
 import { checkForInput } from 'bip174/src/lib/utils';
 import { ProprietaryKeySubtype, PSBT_PROPRIETARY_IDENTIFIER } from '../PsbtUtil';
 
+export const DEFAULT_TX_INPUT_SEQUENCE_NUMBER = 0xffffffff;
```
OttoAllmendinger, line 39:
```suggestion
/** Final (non-replaceable) */
export const TX_INPUT_SEQUENCE_NUMBER_FINAL = 0xffffffff;
```

whether it is default or not is defined by how it is used

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #4050, opened 2023-11-08
feat(utxo-lib): build transaction without nonWitnessUtxo when specified

davidkaplanbitgo:
TICKET: BTC-606

In the instance that we want to doing a bulk sweep, for network and client performance reasons we are substituting the nonWitnessUtxo for p2sh and p2shP2pk inputs with a witnessUtxo. We need the witnessUtxo so that we can half sign the transaction locally with the user key. When we send the half signed to BitGo, the PSBT will be properly populated such that the non-segwit inputs have the nonWitnessUtxo. This means when we send it to BitGo we should remove the witnessUtxo so that it just has the partialSig and redeemScript.

### DIFF
--- a/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
+++ b/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
@@ -163,7 +163,8 @@ export function updateReplayProtectionUnspentToPsbt(
   psbt: UtxoPsbt,
   inputIndex: number,
   u: Unspent<bigint>,
-  redeemScript?: Buffer
+  redeemScript?: Buffer,
+  customParams?: { skipNonWitnessUtxo?: boolean }
 ): void {
   if (!psbtIncludesUnspentAtIndex(psbt, inputIndex, u.id)) {
     throw new Error(`unspent does not correspond to psbt input`);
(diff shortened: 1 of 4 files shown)


lcovar (dismissed):

rushilbg (dismissed):

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -178,12 +179,15 @@ export function updateReplayProtectionUnspentToPsbt(
...
-  if (!isUnspentWithPrevTx(u) && !isZcash) {
+  if (!isUnspentWithPrevTx(u) && !isZcash && !customParams?.skipNonWitnessUtxo) {
     throw new Error('Error, require previous tx to add to PSBT');
   }
   if (isZcash && !input.witnessUtxo) {
     const { script, value } = toPrevOutput(u, psbt.network);
     psbt.updateInput(inputIndex, { witnessUtxo: { script, value } });
+  } else if (customParams?.skipNonWitnessUtxo) {
+    const { script, value } = toPrevOutput(u, psbt.network);
+    psbt.updateInput(inputIndex, { witnessUtxo: { script, value } });
```
saravanan7mani, line 190:
```
if (customParams?.skipNonWitnessUtxo || (isZcash && !input.witnessUtxo)) {
    const { script, value } = toPrevOutput(u, psbt.network);
    psbt.updateInput(inputIndex, { witnessUtxo: { script, value } });
}
```

saravanan7mani (changes requested):

saravanan7mani (approved):

PR #4054, opened 2023-11-09
feat(sdk-coin-ada): increase min amount to 1.5 ADA for token consolid…

Vijay-Jagannathan:
Ticket: WIN-736

### Context

Minimum balance to be sent in a token transaction is ~ 1.15 ADA. Increasing ADA minimum amount in the tx builder to 1.5 for token consolidation to avoid any issues


modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -282,7 +282,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
         // support for multi-asset consolidation
         if (this._multiAssets !== undefined) {
           const totalNumberOfAssets = CardanoWasm.BigNum.from_str(this._multiAssets.length.toString());
-          const minAmountNeededForOneAssetOutput = CardanoWasm.BigNum.from_str('1000000');
+          const minAmountNeededForOneAssetOutput = CardanoWasm.BigNum.from_str('1500000');
```
margueriteblair, line 285: What happens if theres more than one asset?

Vijay-Jagannathan: We construct a separate output for each asset/token in the UTXO, so each output will have 1.5 based on the new minimum amount

gianchandania (approved):

PR #4055, opened 2023-11-10
feat(utxo-lib): accept isReplaceable flag while adding unspents to PSBT

rushilbg:
If this flag is set to true, then a non-final is used for tx input nSequence, otherwise final value is used. The non-final value is what Bitcoin Core uses in these cases.

BTC-591

### DIFF
--- a/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
+++ b/modules/utxo-lib/src/bitgo/wallet/Unspent.ts
@@ -39,6 +39,11 @@ import { ProprietaryKeySubtype, PSBT_PROPRIETARY_IDENTIFIER } from '../PsbtUtil'
 /** Final (non-replaceable) */
 export const TX_INPUT_SEQUENCE_NUMBER_FINAL = 0xffffffff;
 
+/** Non-Final (Replaceable)
+ * Reference: https://github.com/bitcoin/bitcoin/blob/v25.1/src/rpc/rawtransaction_util.cpp#L49
+ * */
+export const MAX_BIP125_RBF_SEQUENCE = 0xffffffff - 2;
+
 export interface WalletUnspent<TNumber extends number | bigint = number> extends Unspent<TNumber> {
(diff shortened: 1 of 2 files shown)


modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -39,6 +39,11 @@ import { ProprietaryKeySubtype, PSBT_PROPRIETARY_IDENTIFIER } from '../PsbtUtil'
 /** Final (non-replaceable) */
 export const TX_INPUT_SEQUENCE_NUMBER_FINAL = 0xffffffff;
 
+/** Non-Final (Replaceable)
+ * Reference: https://github.com/bitcoin/bitcoin/blob/88c3b100f0e83318b9379b17ba7bd2a1c5a51b74/src/rpc/rawtransaction_util.cpp#L49
```
OttoAllmendinger, line 43:
```suggestion
 * Reference: https://github.com/bitcoin/bitcoin/blob/v25.1/src/rpc/rawtransaction_util.cpp#L49
```

modules/utxo-lib/src/bitgo/wallet/Unspent.ts
```diff
@@ -39,6 +39,11 @@ import { ProprietaryKeySubtype, PSBT_PROPRIETARY_IDENTIFIER } from '../PsbtUtil'
...
+ * */
+export const TX_INPUT_SEQUENCE_NUMBER_NON_FINAL = 0xffffffff - 2;
```
OttoAllmendinger, line 45:
```suggestion
export const MAX_BIP125_RBF_SEQUENCE = 0xffffffff - 2;
```

(more precise, there are many nonfinal numbers but only one max bip125 rbf sequence number)

OttoAllmendinger (dismissed):

saravanan7mani (approved):

PR #4056, opened 2023-11-10
Win 1087 add common method in abstract eth

gianchandania:
This PR adds getContractData method in abstract-eth, so that it can be removed from coinSpecific modules
WIN-1087


rushilbg (approved):

OttoAllmendinger (approved):

sachushaji (approved):
Nice!!

PR #4065, opened 2023-11-14
feat(sdk-coin-ada): dynamically modify fee based on asset output for …

Vijay-Jagannathan:
Ticket: WIN-736

### Context

 - While trying ADA token consolidation in test env, we encountered a fee error while broadcasting the tx
   - `Error message: Bad Request Response body: \"transaction submit error ShelleyTxValidationError ShelleyBasedEraBabbage (ApplyTxError [UtxowFailure (UtxoFailure (FromAlonzoUtxoFail (FeeTooSmallUTxO (Coin 171617) (Coin 167217))))])`
 - This was happening because we were not dynamically changing the fee value based on the asset outputs
 - Fee estimation is currently done using a draft transaction. With this PR, we are adding the asset outputs to the draft tx before calculating fee and thereby will have a more accurate fee value when broadcasting


modules/sdk-coin-ada/src/lib/transactionBuilder.ts
```diff
@@ -161,11 +161,70 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+            change = change.checked_add(CardanoWasm.BigNum.from_str(withdrawal.value));
+          });
+        }
+
+        // If totalAmountToSend is 0, its consolidation
+        if (totalAmountToSend.to_str() == '0') {
+          // support for multi-asset consolidation
+          if (this._multiAssets !== undefined) {
+            const totalNumberOfAssets = CardanoWasm.BigNum.from_str(this._multiAssets.length.toString());
+            const minAmountNeededForOneAssetOutput = CardanoWasm.BigNum.from_str('1500000');
```
margueriteblair, line 183: nit: can we put this in a top level variable in case it needs to be changed in the future?

margueriteblair (approved):

PR #4067, opened 2023-11-14
feat(statics): onboard new batch of tokens

evalenti-simtlix:
Onboard new batch of tokens

EA-2134

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -3107,10 +3107,18 @@ export const coins = CoinMap.fromCoins([
   ),
   erc20(
     'ab3e6da9-d9b2-4ff5-a25e-e97b532b9a95',
-    'eurcv',
+    'eurcvv0',
     'EUR Coinvertible',
     18,
     '0xf7790914dc335b20aa19d7c9c9171e14e278a134',
+    UnderlyingAsset.EURCVV0
+  ),
+  erc20(
+    '4a47b0a6-f5a4-4a74-89f0-2475fb92e191',
+    'eurcv',
+    'EUR CoinVertible',
+    18,
+    '0x5f7827fdeb7c20b443265fc2f40845b715385ff2',
     UnderlyingAsset.EURCV
   ),
   erc20(
@@ -10748,6 +10756,14 @@ export const coins = CoinMap.fromCoins([
     6,
     '0x2791bca1f2de4661ed88a30c99a7a9449aa84174',
     UnderlyingAsset['polygon:usdc']
+  ),
+  polygonErc20(
+    '5a17fe9c-3b78-4f12-b419-ed3662f4e985',
+    'polygon:usdcv2',
+    'USD Coin (native)',
+    6,
+    '0x3c499c542cef5e3811e1192ce70d8cc03d5c3359',
+    UnderlyingAsset['polygon:usdcv2']
   ),
   polygonErc20(
     'dee07c7a-96cb-4dbb-906c-b0ad98160cff',
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -285,6 +285,7 @@ export enum UnderlyingAsset {
   EOS = 'eos',
   ERD = 'erd',
   EUR = 'eur',
+  EURCVV0 = 'eurcvv0',
   EURCV = 'eurcv',
   EUROC = 'euroc',
   GBP = 'gbp',
@@ -1281,6 +1282,7 @@ export enum UnderlyingAsset {
 
   // polygon Token ERC-20
   'polygon:usdc' = 'polygon:usdc',
+  'polygon:usdcv2' = 'polygon:usdcv2',
   'polygon:usdt' = 'polygon:usdt',
   'polygon:weth' = 'polygon:weth',
   'polygon:wbtc' = 'polygon:wbtc',


mullapudipruthvik (dismissed):

modules/statics/src/coins.ts
```diff
@@ -3113,6 +3113,14 @@ export const coins = CoinMap.fromCoins([
     '0xf7790914dc335b20aa19d7c9c9171e14e278a134',
     UnderlyingAsset.EURCV
   ),
+  erc20(
+    '4a47b0a6-f5a4-4a74-89f0-2475fb92e191',
+    'eurcvv2',
+    'EUR CoinVertible',
+    18,
+    '0x5F7827FDeb7c20b443265Fc2F40845B715385Ff2',
```
aarvay, line 3121: the address need to be in lower case

aarvay (changes requested):

modules/statics/src/coins.ts
alebusse: Ya, i dont think we use checksum addresses.

alebusse (dismissed):

aarvay (dismissed):

modules/statics/src/coins.ts
```diff
@@ -3113,6 +3113,14 @@ export const coins = CoinMap.fromCoins([
     '0xf7790914dc335b20aa19d7c9c9171e14e278a134',
     UnderlyingAsset.EURCV
   ),
+  erc20(
+    '4a47b0a6-f5a4-4a74-89f0-2475fb92e191',
+    'eurcvv2',
+    'EUR CoinVertible',
+    18,
+    '0x5f7827fdeb7c20b443265fc2f40845b715385ff2',
```
mmcshinsky-bitgo, line 3121: https://etherscan.io/address/0x5f7827fdeb7c20b443265fc2f40845b715385ff2 shows this as `eurcv` not `eurcvv2`. How will a user know this is this specific contract in places like the UI?

modules/statics/src/coins.ts
```diff
@@ -10749,6 +10781,14 @@ export const coins = CoinMap.fromCoins([
     '0x2791bca1f2de4661ed88a30c99a7a9449aa84174',
     UnderlyingAsset['polygon:usdc']
   ),
+  polygonErc20(
+    '5a17fe9c-3b78-4f12-b419-ed3662f4e985',
+    'polygon:usdcv2',
+    'USDC (native)',
```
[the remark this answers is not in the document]
doodles92: @evalenti-simtlix lets change this to "USD Coin (native)" 

@mmcshinsky-bitgo there is a bridged version that we are renaming after we onboard this; our problem w/ bitgo stack is that we can't use the same symbol

modules/statics/src/coins.ts
doodles92: the "v2" is our bitgo platform label which is the latest version (on chain they jsut call it the symbol since the older contract is just deprecated but we need to have it in bitgo since tehcnialyl they are separate tokens)

doodles92: actually @mmcshinsky-bitgo we're going to make following change
change existing EURCV -> EURCVv0 (since no one is using the old token)
EURCVv2 -> EURCV

modules/statics/src/coins.ts
```diff
@@ -7949,6 +7957,30 @@ export const coins = CoinMap.fromCoins([
...
+    '29c91204-eab5-499c-a623-1604c37b3848',
+    'tusdc',
+    'Goerli USDC',
+    18,
+    '0xef7ed2d5636c3dd9dc5e1d9a01fdb362d2927014',
+    UnderlyingAsset.TUSDC,
+    undefined,
+    undefined,
+    undefined,
+    Networks.test.goerli
```
mishkinf-bitgo, line 7970: isn't this asset on Polygon?

modules/statics/src/coins.ts
```diff
@@ -7949,6 +7957,30 @@ export const coins = CoinMap.fromCoins([
     undefined,
     Networks.test.goerli
   ),
+  terc20(
+    '29c91204-eab5-499c-a623-1604c37b3848',
+    'tusdc',
+    'Goerli USDC',
+    18,
```
mishkinf-bitgo, line 7964: Is this the underlying contract in etherscan? If so then it says the token has 6 decimals: https://goerli.etherscan.io/token/0x07865c6e87b9f70255377e024ace6630c1eaa37f no?

modules/statics/src/coins.ts
```diff
@@ -7949,6 +7957,30 @@ export const coins = CoinMap.fromCoins([
...
+    undefined,
+    undefined,
+    Networks.test.goerli
+  ),
+  terc20(
+    '9aed2928-d976-4263-b80b-0e96d6fe6240',
+    'tusdt',
+    'Goerli USDT V2',
+    18,
+    '0x1697152c0e82b1a1e75ad2976f4d1e388ba65dff',
```
mishkinf-bitgo, line 7977: Same for this one. Does this token have only 6 decimals? https://goerli.etherscan.io/token/0xc2c527c0cacf457746bd31b2a698fe89de2b6d49?a=0x1697152c0e82b1a1e75ad2976f4d1e388ba65dff

modules/statics/src/coins.ts
evalenti-simtlix: Is not the shared address shared in the ticket, i think maybe the ticket address was wrong

https://bitgoinc.atlassian.net/browse/EA-2134

evalenti-simtlix: Removed to push the prod tokens

CapnMigraine (approved):

PR #4070, opened 2023-11-14
feat(sdk-core): allow tss signing with unencrypted prv

alebusse:
Modified the getUserPrv method to allow using the provided unencrypted prv and not try to derivate if its a TSS wallet, also fixed the signTransaction use the txRequestId if the txPrebuild is empty for a TSS wallet

WP-990

TICKET: WP-990

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -1595,7 +1595,7 @@ export class Wallet implements IWallet {
    * @return {*}
    */
   async signTransaction(params: WalletSignTransactionOptions = {}): Promise<SignedTransaction | TxRequest> {
-    const { txPrebuild, apiVersion } = params;
+    const { txPrebuild, apiVersion, txRequestId } = params;
 
     if (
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1617,7 +1617,11 @@ export class Wallet implements IWallet {
     }
 
     if (!txPrebuild || typeof txPrebuild !== 'object') {
-      throw new Error('txPrebuild must be an object');
+      if (!txRequestId || this.multisigType() === 'onchain') {
+        throw new Error('txPrebuild must be an object');
```
YPAzevedo, line 1621: this error seems a bit misleading here no? maybe we should add more errors here and a bit more information on the error we throw, so its easier for us to debug in the future

mkottaichamy: +1, looks like copy/paste mistake. Would give a different error message

mkottaichamy (approved):

YPAzevedo (approved):

PR #4075, opened 2023-11-15
feat(statics): prevent usage of TON in the United States

FloBitGo:
TON should not be allowed to be used in the US. This adds a new property `properties` to account coins with an array of `restrictedCountries` in which the coin will not be available.

Ticket: COPS-678


rushilbg (dismissed):

OttoAllmendinger (commented):
do we need the extra level of nesting? Can't we just put the coin properties on the toplevel?

lcovar (approved):

PR #4104, opened 2023-11-29
feat(utxo-lib): add basic utils to utxo-lib

davidkaplanbitgo:
Be able to parse basic information from a psbt.

BTC-637

### DIFF
--- a/modules/utxo-lib/src/bitgo/index.ts
+++ b/modules/utxo-lib/src/bitgo/index.ts
@@ -10,6 +10,7 @@ export * from './dash';
 export * from './parseInput';
 export * from './signature';
 export * from './transaction';
+export * from './transactionAmounts';
 export * from './types';
 export * from './Unspent';
 export * from './UtxoPsbt';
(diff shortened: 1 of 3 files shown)


modules/utxo-lib/src/bitgo/wallet/Psbt.ts
```diff
@@ -584,3 +584,38 @@ export function deleteWitnessUtxoForNonSegwitInputs(psbt: UtxoPsbt): void {
     }
   });
 }
+
+export function getTransactionInfoFromPsbt(psbt: UtxoPsbt): {
```
OttoAllmendinger, line 588:
```suggestion
export function getTransactionAmountsFromPsbt(psbt: UtxoPsbt): {
```

OttoAllmendinger (changes requested):
let's move it into a file `transactionAmounts.ts`

modules/utxo-lib/src/networks.ts
```diff
@@ -35,6 +35,21 @@ const coins = {
   DOGE: 'doge',
 } as const;
 
+const testnetCoins = {
```
OttoAllmendinger, line 38: as described in the `coins` const - we don't actually want or need to define coin names here, we already have it in the `statics` package

OttoAllmendinger (dismissed):

lcovar (approved):

PR #4114, opened 2023-12-01
feat(sdk-coin-sol): add isTokenTransaction field to SOL recovery txns

alia-bitgo:
for display of token recovery transactions in ovc, adding in a flag to check in ovc

TICKET: WP-1087


zahin-mohammad (commented):
Flushing from review Q while this is in draft (changes lgtm, just missing some tests ;) )

lcovar (commented):
flushing queue

PR #4116, opened 2023-12-04
feat: add new coin feature for bulk staking transaction

dpkjnr:
Ticket: WIN-705


modules/statics/src/coins.ts
```diff
@@ -219,6 +219,7 @@ const SUI_FEATURES = [
   CoinFeature.TSS_COLD,
   CoinFeature.STAKING,
   CoinFeature.BULK_TRANSACTION,
+  CoinFeature.BULK_STAKING_TRANSACTION,
```
DinshawKothari, line 222: why enabling it for sui also ?

dpkjnr: It is already enabled for SUI in the UI by hardcoding. Creating this coin feature to remove the hardcoding thats why enabling it for SUI also

DinshawKothari (approved):

PR #4123, opened 2023-12-06
feat(statics): add ERC20 tokens for FTX

Vijay-Jagannathan:
Ticket: WIN-1313

### Context

Onboarding the following ERC20 tokens for FTX, which are part of the token txns [marked in RED ](https://docs.google.com/spreadsheets/d/1HkZNYVF4u8_tOyDc6LvsI7Q9s4qOuCDGH9clXmkmkf8/edit?pli=1#gid=601048272)
```
0x622dffcc4e83c64ba959530a5a5580687a57581b
0x47f33feb70ec153fc22755908a4061777d22c2da
0xc5b3d3231001a776123194cf1290068e8b0c783b
0xdb726152680ece3c9291f1016f1d36f3995f6941
0xb9340fb0f8f93f4857eb3012b4f85db2d632e154
0x5079aca02ed583b8a7235442a6455d3f3289997b
0xaaef88cea01475125522e117bfe45cf32044e238
0x6dea81c8171d0ba574754ef6f8b412f2ed88c54d
```

some of these tokens have names which have already been taken, so we have used a variation to differentiate


modules/statics/src/coins.ts
```diff
@@ -7502,6 +7502,70 @@ export const coins = CoinMap.fromCoins([
...
+    UnderlyingAsset.XX
+  ),
+  erc20(
+    'd340bbfc-1215-418d-8e57-64ffc954e2d7',
+    'litt',
+    'LITT',
+    18,
+    '0xc5b3d3231001a776123194cf1290068e8b0c783b',
+    UnderlyingAsset.LITT
+  ),
```
alebusse, line 7528: Are we sure is LITT and not LIT?
https://etherscan.io/address/0xc5b3d3231001a776123194cf1290068e8b0c783b

modules/statics/src/coins.ts
```diff
@@ -7502,6 +7502,70 @@ export const coins = CoinMap.fromCoins([
...
+    '0xdb726152680ece3c9291f1016f1d36f3995f6941',
+    UnderlyingAsset.MEDIANETWORK
+  ),
+  erc20(
+    '2b2800aa-42f3-4bf4-9f15-35f76aace2a8',
+    'blurr',
+    'BLUR',
+    6,
+    '0xb9340fb0f8f93f4857eb3012b4f85db2d632e154',
+    UnderlyingAsset.BLURR
```
alebusse, line 7543:
```suggestion
    UnderlyingAsset.BLUR
```

modules/statics/src/coins.ts
Vijay-Jagannathan: We already have something called `LIT` - https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/coins.ts#L4100, which is different from this one

For lack of a better name, I added `LITT` but I'm open to suggestions

modules/statics/src/coins.ts
Vijay-Jagannathan: Same thing with this one too - https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/coins.ts#L2001

We already had BLUR already, so had to add it with the extra letter to differentiate

alebusse:
@Vijay-Jagannathan im ok with the names, just confirm with someone from Product since i think thats the name they will have to use and see in any tx i believe

Vijay-Jagannathan:
> @Vijay-Jagannathan im ok with the names, just confirm with someone from Product since i think thats the name they will have to use and see in any tx

makes sense, ill check with daniel if he has any recommendations (also this will be just for FTX transfer drifts though, not supporting via UI, etc)

alebusse (dismissed):

Vijay-Jagannathan:
Have changed the conflicting ones as per daniel's suggestion via slack

ravneet-bitgo (approved):

PR #4128, opened 2023-12-06
feat(sdk-core): add optional StakeOptions fields to support STX staking

sandra-wang-bitgo:
TICKET: EA-2430

STX staking would need FE to pass in `duration` and `subType`.
Duration as in the number of delegation cycles.
SubType is to distinguish delegation vs self-stack.

corresponding change in `staking-service`: https://github.com/BitGo/staking-service/pull/1639

### DIFF
--- a/modules/sdk-core/src/bitgo/staking/iStakingWallet.ts
+++ b/modules/sdk-core/src/bitgo/staking/iStakingWallet.ts
@@ -23,11 +23,22 @@ export interface DelegationRequest {
   validator: string;
 }
 
+/**
+ * Represents the options for staking.
+ * @typedef {Object} StakeOptions
+ * @property {string} [amount] - amount to stake
+ * @property {string} [clientId] - clientId
+ * @property {string} [validator] - chosen validator
+ * @property {string} [duration] - delegation duration: a numeric string, in days or cycles
+ * @property {string} [subType] - coin sepcific staking subtype
+ * @property {DelegationRequest[]} [delegationRequests] - The delegation requests
+ */
 export interface StakeOptions {
   amount?: string;
   clientId?: string;
   validator?: string;
-  durationSeconds?: string;
+  duration?: string;
+  subType?: string;
   delegationRequests?: DelegationRequest[];
 }
 


modules/sdk-core/src/bitgo/staking/iStakingWallet.ts
```diff
@@ -27,7 +27,8 @@ export interface StakeOptions {
   amount?: string;
   clientId?: string;
   validator?: string;
-  durationSeconds?: string;
+  duration?: string;
```
OttoAllmendinger, line 30: what's the format and unit here?

OttoAllmendinger (changes requested):

OttoAllmendinger (dismissed):

evanzbitgo (approved):

lcovar (approved):

modules/sdk-core/src/bitgo/staking/iStakingWallet.ts
```diff
@@ -23,11 +23,22 @@ export interface DelegationRequest {
...
+ * @property {string} [subType] - coin sepcific staking subtype
+ * @property {DelegationRequest[]} [delegationRequests] - The delegation requests
+ */
 export interface StakeOptions {
   amount?: string;
   clientId?: string;
   validator?: string;
-  durationSeconds?: string;
+  duration?: string;
+  subType?: string;
```
zahin-mohammad, line 41: Can we enumerate the different types instead of just a string? When is this expected to be populated?

zahin-mohammad (approved):
not blocking, but it would be great to enumerate valid strings for subType instead of keeping the type as just `string`

modules/sdk-core/src/bitgo/staking/iStakingWallet.ts
OttoAllmendinger: if you go over this again, please use this jsdoc syntax

```
interface I {
 /** foo */
 foo: bar
}
```

OttoAllmendinger (approved):

PR #4130, opened 2023-12-07
feat: add address book methods to wallet class

mmcshinsky-bitgo:
PX-262

TICKET: PX-262


alebusse (approved):
Looks good, but you could add Prime as codeowner of this new address book directory in another PR too?

PR #4135, opened 2023-12-08
feat(root): whitelist apiVersion for buildAccountConsolidations

alebusse:
In order to start migrating TSS EdDSA to full txRequest, we need to allow this param to be passed to use txRequest full, similar on how we do for sendMany

WP-1123

TICKET: WP-1123


OttoAllmendinger (approved):

zahin-mohammad (approved):
lgtm. Let's hold off on merging this in until the backend changes are in place.

PR #4139, opened 2023-12-08
fix(sdk-coin-ton): support non-bouncable address as recipient

ewangbitgo:
Ticket: EA-2457

TICKET: EA-2457

### DIFF
--- a/modules/sdk-coin-ton/src/lib/iface.ts
+++ b/modules/sdk-coin-ton/src/lib/iface.ts
@@ -5,6 +5,7 @@ export interface TxData {
   id: string;
   sender: string;
   destination: string;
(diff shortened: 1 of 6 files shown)


modules/sdk-coin-ton/src/lib/iface.ts
```diff
@@ -5,6 +5,7 @@ export interface TxData {
   id: string;
   sender: string;
   destination: string;
+  destination_alias: string;
```
noel-bitgo, line 8: why not camel case?

ewangbitgo: fixed

noel-bitgo (approved):

PR #4141, opened 2023-12-10
refactor(sdk-coin-atom): use abstract-cosmos

hitansh-madan:
Ticket: WIN-1192


OttoAllmendinger (dismissed):

mullapudipruthvik (dismissed):

modules/sdk-coin-atom/src/lib/index.ts
```diff
@@ -1,12 +1,10 @@
+import * as Constants from './constants';
 import * as Utils from './utils';
-import * as Interface from './iface';
 
+export {
+  CosmosTransaction as Transaction,
+  CosmosTransactionBuilder as TransactionBuilder,
+} from '@bitgo/abstract-cosmos';
 export { KeyPair } from './keyPair';
-export { Transaction } from './transaction';
```
mmcshinsky-bitgo, line 5: Removal of all these exports completely from here is a breaking change since /src/index.ts exposes these to end users.

hitansh-bitgo: Added some of the exports back from abstract-cosmos & added a breaking change footer in the commit message for the ones still being removed completely

modules/sdk-coin-atom/src/lib/constants.ts
```diff
@@ -1,12 +1,5 @@
-export const DEFAULT_SEED_SIZE_BYTES = 16;
-export const sendMsgTypeUrl = '/cosmos.bank.v1beta1.MsgSend';
-export const delegateMsgTypeUrl = '/cosmos.staking.v1beta1.MsgDelegate';
-export const undelegateMsgTypeUrl = '/cosmos.staking.v1beta1.MsgUndelegate';
-export const withdrawDelegatorRewardMsgTypeUrl = '/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward';
 export const validDenoms = ['natom', 'uatom', 'matom', 'atom'];
 export const accountAddressRegex = /^(cosmos)1(['qpzry9x8gf2tvdw0s3jn54khce6mua7l']{38})$/;
 export const validatorAddressRegex = /^(cosmosvaloper)1(['qpzry9x8gf2tvdw0s3jn54khce6mua7l']{38})$/;
-export const contractAddressRegex = /^(cosmos)1(['qpzry9x8gf2tvdw0s3jn54khce6mua7l']+)$/;
```
DinshawKothari, line 9: why removing contractAddressRegex ?

modules/statics/src/base.ts
```diff
@@ -1438,7 +1438,7 @@ export enum KeyCurve {
  * This enum contains the base units for the coins that BitGo supports
  */
 export enum BaseUnit {
-  ATOM = 'uATOM',
+  ATOM = 'uatom',
```
DinshawKothari, line 1441: Will this break anything in WP or we use lowercase there ?

hitansh-bitgo: We were using lowercase everywhere except here

modules/sdk-coin-atom/src/lib/constants.ts
hitansh-bitgo: Contract support was only added in abstract-cosmos & other chains but wasn't added for Atom it seems, so this was unused. I've updated the PR with the contract support changes needed for atom as well.

DinshawKothari (approved):

PR #4144, opened 2023-12-11
feat(sdk-core): provide skipKeychain to wallet share API request

zahin-mohammad:
TICKET: WP-1136

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -58,11 +58,25 @@ describe('V2 Wallet:', function () {
     multisigType: 'onchain',
     type: 'hot',
   };
+  const coldWalletData = {
+    id: '65774419fb4d9690847fbe4b00000000',
+    coin: 'tbtc',
+    keys: ['65774412e54b7516393c9df800000000', '6577442428664ffe791af7ea00000000', '6577442b7317a945756c2fd900000000'],
+    coinSpecific: {},
+    multisigType: 'onchain',
+    type: 'cold',
+  };
   const wallet = new Wallet(bitgo, basecoin, walletData);
+  const coldWallet = new Wallet(bitgo, basecoin, coldWalletData);
   const bgUrl = common.Environments[bitgo.getEnv()].uri;
   const address1 = '0x174cfd823af8ce27ed0afee3fcf3c3ba259116be';
   const address2 = '0x7e85bdc27c050e3905ebf4b8e634d9ad6edd0de6';
   const tbtcHotWalletDefaultParams = { txFormat: 'psbt', addressType: 'p2trMusig2' };
+
+  afterEach(function () {
+    sinon.restore();
+    sinon.reset();
+  });
 
   describe('Wallet transfers', function () {
     it('should search in wallet for a transfer', async function () {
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1507,6 +1507,7 @@ export class Wallet implements IWallet {
       reshare: params.reshare,
       message: params.message,
       disableEmail: params.disableEmail,
+      skipKeychain: (this.type() ?? 'hot') === 'hot' ? !!params.skipKeychain : undefined,
```
zahin-mohammad, line 1510: Given how sensitive wallet sharing is, I want users to explicitly pass this field in as input instead of deriving it based on other input params. This is why the params are passed as is for hot wallets instead of derived from other params such as `options.keychain`.

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -1955,18 +1969,22 @@ describe('V2 Wallet:', function () {
...
-        .get(`/api/v2/tbtc/key/${wallet.keyIds()[2]}`)
+        .get(`/api/v2/tbtc/key/${coldWallet.keyIds()[2]}`)
         .reply(200, {});
 
       const createShareNock = nock(bgUrl)
-        .post(`/api/v2/tbtc/wallet/${wallet.id()}/share`, { user: userId, permissions, keychain: {} })
+        .post(`/api/v2/tbtc/wallet/${coldWallet.id()}/share`, {
+          user: userId,
+          permissions,
+          keychain: {},
```
pranavjain97, line 1983: why are we allowing `keychain` to be `{}` if skipKeychain is not set to true?

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -437,6 +437,11 @@ export interface CreateShareOptions {
   reshare?: boolean;
   message?: string;
   disableEmail?: boolean;
+  /**
+   * If true, allows sharing hot wallets without a keychain.
```
pranavjain97, line 441: based on this statement, it should not allow if its not true

modules/sdk-core/src/bitgo/wallet/wallet.ts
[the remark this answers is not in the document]
zahin-mohammad: Yea, after looking at the API, it can get a little messy if we only have this for hot-wallets. Gonna move this to a draft until I finalize the backend changes.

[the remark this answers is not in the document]
zahin-mohammad: Changed it to be dynamic based on the inputs. The inputs already handle `skipKeychain`, users using this method should have the expected outcome with the api while minimizing integration changes clients need to make.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1500,21 +1500,16 @@ export class Wallet implements IWallet {
...
       }
     }
-
     const options: CreateShareOptions = {
       user: sharing.userId,
       permissions: params.permissions,
       reshare: params.reshare,
       message: params.message,
       disableEmail: params.disableEmail,
+      skipKeychain: Object.keys(sharedKeychain ?? {}).length === 0,
```
pengyuc-bitgo, line 1509: but I thought you want this flag to be explicit? Is it only on the API level?

pengyuc-bitgo (approved):

PR #4149, opened 2023-12-12
fix(statics): don't use . or $ or " " in token names

rlutzbitgo:
TICKET: EA-2376


CapnMigraine (approved):

mmcshinsky-bitgo:
I'm interested to know the logic behind why the symbols need to be removed?

PR #4150, opened 2023-12-13
feat(sdk-core, sdk-coin-eth): add function to transfer nfts

mohammadalfaiyazbitgo:
Ticket: WP-1094

TICKET: WP-1094

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -41,6 +41,7 @@ import {
   VerifyAddressOptions as BaseVerifyAddressOptions,
   VerifyTransactionOptions,
   Wallet,
+  BuildNftTransferDataOptions,
 } from '@bitgo/sdk-core';
 import {
   BaseCoin as StaticsBaseCoin,
@@ -58,6 +59,8 @@ import { SignTypedDataVersion, TypedDataUtils, TypedMessage } from '@metamask/et
 
 import {
   calculateForwarderV1Address,
+  ERC1155TransferBuilder,
+  ERC721TransferBuilder,
   getCommon,
   getProxyInitcode,
   getToken,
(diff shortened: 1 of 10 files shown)


mohammadalfaiyazbitgo:
Transferred NFT with method here: https://goerli.etherscan.io/tx/0xba821690529fd637d368e7fbeff165dc7af60c177ca5a46fc7a984e0320c3304

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2474,4 +2477,38 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
     }
     return Buffer.concat(parts);
   }
+
+  /**
+   * Build the data to transfer an ERC-721 or ERC-1155 token to another address
+   * @param params
+   */
+  buildNftTransferData(params: NFTTransferOptions & { fromAddress: string }): string {
```
margueriteblair, line 2485: I see that this is only implemented for `abstractEthLikeNewCoins` and not `abstractEthLike`. That's intentional right? I'm not super familiar on the ETH set up, just wanna make sure that we're only expecting it to be for these new coins

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2099,6 +2102,79 @@ export class Wallet implements IWallet {
...
+          }
+        }
+
+        const data = this.baseCoin.buildNftTransferData({ ...sendNftOptions, fromAddress: baseAddress });
+        return this.sendMany({
+          ...sendOptions,
+          recipients: [
+            {
+              address: sendNftOptions.tokenContractAddress,
+              amount: '0',
```
margueriteblair, line 2169: Is there a case where we'd have any fees and it doesnt make sense for `amount` to be 0?

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
mohammadalfaiyazbitgo: I believe so, both `eth` & `polygon` extend from `abstractEthLikeNewCoins` which are the only chains we need to implement nfts for atm.

modules/sdk-core/src/bitgo/wallet/wallet.ts
mohammadalfaiyazbitgo: Don't think so, the documentation https://developers.bitgo.com/guides/transact/nfts says to always set the amount to 0.

rlutzbitgo (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2099,6 +2102,91 @@ export class Wallet implements IWallet {
...
+              data: data,
+            },
+          ],
+        });
```
alebusse, line 2156: seems duplicated, a nit for another pr

alebusse (approved):

PR #4154, opened 2023-12-13
feat: introduce sdk-types from developer portal

mmcshinsky-bitgo:
Experimental proposal to use the openapi spec from [https://developers.bitgo.com](https://developers.bitgo.com/api/overview) as a source of truth for types. Much of the SDK interacts with the API layer but miss out on consistent and non-manually written types. Using the published API can also help reduce a lot of the `any` types as well of which users have regularly complained about in the past.

Since the developer portal is the central spec system published for requests and responses, `@bitgo/sdk-types` or `@bitgo/sdk-openapi` can be published alongside as needed `@bitgo/public-types`.

Included is a small example usage of what it would or could look like to replace manual types inside a module.

Open to probably a lot of feedback on peoples thoughts around this.

PX-000000

TICKET: PX-000000


OttoAllmendinger (approved):
lgtm but ci errors

PR #4156, opened 2023-12-15
chore(statics): add new batch of tokens

evalenti-simtlix:
add new batch of tokens

EA-2421

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -338,6 +338,7 @@ export enum UnderlyingAsset {
   AERGO = 'aergo',
   AERGO1 = 'aergo1',
   AGEUR = 'ageur',
+  AGI = 'agi',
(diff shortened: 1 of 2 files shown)


rlutzbitgo (changes requested):

modules/statics/src/base.ts
```diff
@@ -1690,6 +1693,8 @@ export enum UnderlyingAsset {
   'tpolygon:link' = 'tpolygon:link',
   'tpolygon:name' = 'tpolygon:name',
   'tpolygon:opm' = 'tpolygon:opm',
+  'tpolygon:pme' = 'tpolygon:pme',
+
```
rlutzbitgo, line 1697: Where is sol:wec?

rlutzbitgo (approved):

PR #4166, opened 2023-12-20
feat(abstract-eth): add chain id as network identifier

gianchandania:
This PR adds chain id as the network id in the prefix used in generating an operation hash for sending native coins and tokens. The prefixes for coins are removed from the transferBuilder and instead added to the network.ts file because now as the prefix includes chain id, it will depend on the network. This PR adds tests related to transferBuilder for coins arbeth and opeth.

WIN-1446
TICKET: WIN-1446


modules/sdk-coin-rbtc/test/unit/transactionBuilder/send.ts
```diff
@@ -75,62 +56,6 @@ describe('Rbtc send transaction', function () {
...
-
-    const { signature } = decodeTransferData(tx.toJson().data);
-    const operationHash = getOperationHash(tx);
-
-    const { v, r, s } = ethUtil.fromRpcSig(signature);
-    const senderPubKey = ethUtil.ecrecover(Buffer.from(operationHash || ''), v, r, s);
-    const senderAddress = ethUtil.pubToAddress(senderPubKey);
-    const senderKey = new KeyPair({ prv: testData.PRIVATE_KEY_1 });
-    ethUtil.bufferToHex(senderAddress).should.equal(senderKey.getAddress());
-  });
```
gianchandania, line 133: These tests have been moved from here to sdk-coin-eth module because rbtc doesn't have tokens, and the token used here is actually an erc20 token

modules/sdk-coin-eth/test/unit/transferBuilder.ts
```diff
@@ -85,30 +73,6 @@ describe('Eth send multi sig builder', function () {
...
-        .coin('tcusd')
-        .expirationTime(1590078260)
-        .amount('0')
-        .to(toAddress)
-        .contractSequenceId(2)
-        .key(key);
-      const result = builder.signAndBuild();
-      should.equal(result, testData.SEND_TOKEN_AMOUNT_ZERO_DATA);
-    });
-
```
gianchandania, line 111: These tests have been moved to sdk-coin-celo module because the token used here is a celo token

modules/abstract-eth/src/lib/transferBuilder.ts
```diff
@@ -137,11 +143,12 @@ export class TransferBuilder {
 
   protected getOperationData(): (string | Buffer)[][] {
     let operationData;
+
     if (this._tokenContractAddress !== undefined) {
       operationData = [
         ['string', 'address', 'uint', 'address', 'uint', 'uint'],
         [
-          this.getTokenOperationHashPrefix(),
+          this._tokenOperationHashPrefix ?? this.getTokenOperationHashPrefix(),
```
hitansh-bitgo, line 151: Could this check be in `getTokenOperationHashPrefix()` instead?

gianchandania: Basically we need coin and network to derive this value because now as we are using chain id. We are overriding getTokenOperationHashPrefix() method in the coin specific modules like sdk-coin-eth, sdk-coin-polygon etc. So we will have to add this similar logic in all the individual coin's modules. Hence I thought to do it at this one place only. The reason why I added this method call as well in this condition, is to make it backwards compatible

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -1415,6 +1418,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
     });
     const transferBuilder = txBuilder.transfer() as TransferBuilder;
     transferBuilder
+      .coin(this.staticsCoin?.name as string)
```
sachushaji, line 1421: Can't we initialise coin as parameter up initialisation / can we figure out that this coin needs to chain prefixes without having to pass in coin name explicitly like this. The reason being, passing in coinName for some and not for others could be confusing for users. We already have _coin which is the coin statics loaded on initilization, so can't we use that where ever we need to take of chainId logics so that it would mean minimal changes one need to make for places which already makes use of transaction builder

modules/sdk-coin-arbeth/test/unit/transferBuilder.ts
```diff
@@ -0,0 +1,187 @@
...
+      const builder = new TransferBuilder()
+        .expirationTime(1590078260)
+        .coin('tarbeth')
+        .amount(amount)
+        .to(toAddress)
+        .contractSequenceId(2)
+        .key(key)
+        .data('0x');
+      const result = builder.signAndBuild();
+      should.equal(result, testData.SEND_FUNDS_DATA);
```
sachushaji, line 38: What is the difference between the two testcases? does the order of the coin parameter matter?

modules/sdk-coin-arbeth/test/unit/transferBuilder.ts
```diff
@@ -0,0 +1,187 @@
...
+      const builder = new TransferBuilder()
+        .coin('tarbeth:link')
+        .expirationTime(1590078260)
+        .amount(amount)
+        .to(toAddress)
+        .contractSequenceId(0)
+        .key(key);
+      const result = builder.signAndBuild();
+      should.equal(result, testData.SEND_TOKEN_SEQUENCE_ZERO_DATA);
+    });
```
sachushaji, line 89: What is special with zero sequence id so that we need to do an extra test for it?

modules/sdk-coin-arbeth/test/unit/transferBuilder.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    it('should build without a signature set', () => {
+      const builder = new TransferBuilder()
+        .expirationTime(1590078260)
+        .amount(amount)
+        .to(toAddress)
+        .contractSequenceId(2)
+        .data('0x');
+      const result = builder.signAndBuild();
+      should.equal(result, testData.SEND_FUNDS_NO_KEY_DATA);
```
sachushaji, line 111: No need for coin here for this one?

modules/sdk-coin-arbeth/test/unit/transferBuilder.ts
```diff
@@ -0,0 +1,187 @@
...
+      should(() => {
+        new TransferBuilder().amount('-10');
+      }).throw('Invalid amount');
+    });
+
+    it('should fail with an invalid expiration time', () => {
+      should(() => {
+        new TransferBuilder().expirationTime(-1);
+      }).throw('Invalid expiration time');
+    });
```
sachushaji, line 170: These tests are not coinspecific right? can we move them to a more common place?

modules/sdk-coin-eth/test/unit/transferBuilder.ts
sachushaji: Nice!!!

modules/sdk-coin-opeth/test/unit/transferBuilder.ts
```diff
@@ -0,0 +1,187 @@
...
+    it('should fail with an invalid destination address', () => {
+      should(() => {
+        new TransferBuilder().to('invalidaddress');
+      }).throw('Invalid address');
+    });
+
+    it('should fail with an invalid amount: text value', () => {
+      should(() => {
+        new TransferBuilder().amount('invalidamount');
+      }).throw('Invalid amount');
```
sachushaji, line 157: Same as arbeth comments, can we move to a common place?

modules/abstract-eth/src/lib/transferBuilder.ts
sachushaji: Actually, this might not be applicable for those coins because they won't have chain Id changes right so need to keep the override as is anyways? For coins which we have chainId changes, we don't need to overrirde them as the changes here should be sufficient as we are not using coin specific names

gianchandania: Yeah, override is not needed for the new coins, it is just for needed for older coins

modules/sdk-coin-opeth/test/unit/transferBuilder.ts
gianchandania: moved these tests to abstract-eth module

modules/sdk-coin-arbeth/test/unit/transferBuilder.ts
gianchandania: moved these tests to abstract-eth module

modules/sdk-coin-arbeth/test/unit/transferBuilder.ts
gianchandania: My bad, actually this test case was present in sdk-coin-eth because for eth, we don't need to pass coin param. But these 2 are same tests only, removing one

modules/sdk-coin-arbeth/test/unit/transferBuilder.ts
gianchandania: It seems like previously we didn't support sequence id from 0 and this change was introduced later, hence this testcase was added for eth, https://github.com/BitGo/BitGoJS/commit/bdc53412b61076967dd4b325b70fafb87a07dc6d

modules/abstract-eth/src/lib/transferBuilder.ts
gianchandania: Have updated this.

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
gianchandania: As discussed setting this coin is actually required for getting the correct chain id and this will be needed only for new coins which will be using chain id so the change is backwards compatible.

modules/sdk-coin-arbeth/test/unit/transferBuilder.ts
```diff
@@ -0,0 +1,107 @@
+import should from 'should';
+import { KeyPair, TransferBuilder } from '../../src';
+
+import * as testData from '../resources';
+
+describe('Eth send multi sig builder', function () {
```
sachushaji, line 6:
```suggestion
describe('Arbeth send multi sig builder', function () {
```
NIT

modules/sdk-coin-opeth/test/unit/transferBuilder.ts
```diff
@@ -0,0 +1,107 @@
+import should from 'should';
+import { KeyPair, TransferBuilder } from '../../src';
+
+import * as testData from '../resources';
+
+describe('Eth send multi sig builder', function () {
```
sachushaji, line 6:
```suggestion
describe('Opeth send multi sig builder', function () {
```

sachushaji (approved):
LGTM Nice Work!!!

modules/sdk-coin-opeth/test/unit/transferBuilder.ts
gianchandania: will update these in a follow up

PR #4167, opened 2023-12-20
feat(statics): add SOL tokens for FTX transfer drift

Vijay-Jagannathan:
Ticket: WIN-1405

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -1758,6 +1758,8 @@ export enum UnderlyingAsset {
   'sol-perp' = 'sol-perp',
   'sol-woo' = 'sol-woo',
   'sol-weth' = 'sol-weth',
+  'btc-sollet' = 'btc-sollet',
+  'eth-sollet' = 'eth-sollet',
 
   // XRP tokens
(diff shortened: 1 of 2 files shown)


Vijay-Jagannathan:
Please merge if you're the last reviewer, ty!

ajays97 (approved):
Token and token configs (decimals, token address, name) looks good.

modules/statics/src/base.ts
```diff
@@ -1758,6 +1758,8 @@ export enum UnderlyingAsset {
   'sol-perp' = 'sol-perp',
   'sol-woo' = 'sol-woo',
   'sol-weth' = 'sol-weth',
+  'btc-sollet' = 'btc-sollet',
+  'eth-sollet' = 'eth-sollet',
```
ajays97, line 1762: Only these tokens to be onboarded for fixing FTX TDs? @Vijay-Jagannathan

Vijay-Jagannathan: We already added a few recently. This ticket deals only with a specific set of txns, which need these 2 tokens specifically

PR #4169, opened 2023-12-21
WP-1185 fix SDK using proxy

pengyuc-bitgo:
The `superagent-proxy` was removed earlier in this [PR](https://github.com/BitGo/BitGoJS/pull/3921).
And we have to use the proxy-agent now. But that change missed some old usage of the `superagent-proxy`.

This PR also changed the proxy-agent version to 5.0. This is the version that was used under the hood before. The new version requires setting proxy url in the environment variable. It does not quite fit how we use the proxy-agent currently. If we want to upgrade to the latest proxy-agent, it should be done in a separate PR.

This change is tested with the new example codes and local proxy server. Verified that it is using the proxy by shutting down the proxy and saw the connection failure error.

Ticket: WP-1185

### DIFF
--- a/examples/ts/http-proxy/README.md
+++ b/examples/ts/http-proxy/README.md
@@ -1,0 +1,16 @@
+This directory contains a very simple example and test for http proxy setup that allows 
+talking to BitGo's API backend through a http proxy.
+
+This is in contrast to the [BitGo Express module](https://github.com/BitGo/BitGoJS/tree/master/modules/express) which implements the SDK on the
+server that you host in order to provide specific SDK logic before proxying requests 
+to BitGo APIs for your convenience.
+
+This is also different from the [proxy](../proxy) example, which allows developers to use local BitGo SDK methods with a non-BitGo back-end.
+Please take the time to review your use case as you choose between these options.
+
+## Setup + Usage
+- Acquire a test environment account, enterprise and access token
+- Fill in the `TODO` sections with the relevant credentials
+- `npm install`
+- `ts-node server.ts`
+- In a separate shell: `ts-node create-wallet.ts`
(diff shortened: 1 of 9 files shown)


examples/ts/proxy/server.ts
```diff
@@ -1,29 +1,8 @@
...
-const app = express();
-const { createProxyMiddleware } = require('http-proxy-middleware');
+import { createProxy } from 'proxy';
+import { AddressInfo } from 'net';
+const proxy = createProxy();
 
-// This *must* match the environment you are using in the create-wallet script
-const bitgoApi = 'https://app.bitgo-test.com';
-// TODO: replace with your access token
-const secretAccessToken = 'yourAccessToken';
```
zahin-mohammad, line 9: Why are we changing this example? The goal of this example is to show an example of the server injecting the access token opposed to needing it in the frontend.

zahin-mohammad (commented):
Changes lgtm, but I think the examples for create-wallet/server should stay the same. Feel free to sync with @alebusse to see why its having issues (we have clients using similar setups so we should ensure it still works)

examples/ts/proxy/server.ts
pengyuc-bitgo: I have roll-back the changes to this example and added another example for http proxy usage.
Got confused because of the name. The use case are actually different.

pengyuc-bitgo:
> Changes lgtm, but I think the examples for create-wallet/server should stay the same. Feel free to sync with @alebusse to see why its having issues (we have clients using similar setups so we should ensure it still works)

Got confused by the example's name. Added another example for http proxy. Also updated the README.md.

zahin-mohammad (approved):

PR #4176, opened 2024-01-03
feat(bitgo): change default testnet config to holesky

dpkjnr:
Ticket: WIN-710

### DIFF
--- a/modules/sdk-core/src/bitgo/environments.ts
+++ b/modules/sdk-core/src/bitgo/environments.ts
@@ -141,7 +141,7 @@ const testnetBase: EnvironmentTemplate = {
   serverXpub: hardcodedPublicKeys.serverXpub.test,
   hsmXpub: hardcodedPublicKeys.hsmXpub.test,
   btcExplorerBaseUrl: 'https://blockstream.info/testnet/api',
-  etherscanBaseUrl: 'https://api-goerli.etherscan.io',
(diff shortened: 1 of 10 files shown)


modules/sdk-core/src/bitgo/environments.ts
```diff
@@ -141,7 +141,7 @@ const testnetBase: EnvironmentTemplate = {
   serverXpub: hardcodedPublicKeys.serverXpub.test,
   hsmXpub: hardcodedPublicKeys.hsmXpub.test,
   btcExplorerBaseUrl: 'https://blockstream.info/testnet/api',
-  etherscanBaseUrl: 'https://api-goerli.etherscan.io',
+  etherscanBaseUrl: 'https://api-holesky.etherscan.io',
```
gianchandania, line 144: Should we replace or add another url for holesky till the time we depreciate gteth because this is mostly used for recovery purpose. Means once this gets replaced WRW recoveries for gteth will stop working

zahin-mohammad (approved):

modules/sdk-core/src/bitgo/environments.ts
dpkjnr: I think there should not be any usecase of actual funds recovering, it mostly is required to verify whether wrw is working for eth so it should be fine to override. Confirmed it with Daniel

gianchandania (approved):

PR #4189, opened 2024-01-09
feat: whitelist rbf build params

lcovar:
Ticket: BTC-650


rushilbg (commented):
Assuming BuildParamsUTXO is used in the build APIs as well: build, accelerate, initiate, send.
Is there a way to test the API request formed by the SDK? That would be more robust IMO.

lcovar:
> Can we look into moving this codec to https://github.com/BitGo/public-types ?

oh noice, wasn't aware of this. i can do this on a follow up.

PR #4195, opened 2024-01-11
Btc 715 demo

saravanan7mani:


modules/utxo-lib/src/bitgo/Descriptor/Descriptor.ts
```diff
@@ -50,11 +56,12 @@ export interface ParsedPsbtInputWithDescriptor {
 const { expand, Output } = desc.DescriptorsFactory(ecc);
 
 export function isDescriptorSupported(network: Network): boolean {
-  return getMainnet(network) === networks.bitcoin;
+  return true;
+  // return getMainnet(network) === networks.bitcoin;
```
mmcshinsky-bitgo, line 60: Can we remove this comment?

PR #4196, opened 2024-01-16
feat: support webauthn based decryption

zahin-mohammad:
TICKET: WP-1266

### DIFF
--- a/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
+++ b/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
@@ -4,9 +4,11 @@ import { BackupProvider, IWallet } from '../wallet';
 import { BitGoKeyFromOvcShares, OvcToBitGoJSON } from './ovcJsonCodec';
 
 export type KeyType = 'tss' | 'independent' | 'blsdkg';
+
 export type SourceType = 'bitgo' | 'backup' | 'user' | 'cold';
 
 export type WebauthnFmt = 'none' | 'packed' | 'fido-u2f';
+
 export interface WebauthnAuthenticatorInfo {
   credID: string;
   fmt: WebauthnFmt;
@@ -40,7 +42,11 @@ export interface Keychain {
   webauthnDevices?: KeychainWebauthnDevice[];
 }
 
-export type KeychainEncryptedKey = Pick<Keychain, 'encryptedPrv' | 'webauthnDevices'>;
+export type OptionalKeychainEncryptedKey = Pick<Keychain, 'encryptedPrv' | 'webauthnDevices'>;
+
+export type KeychainWithEncryptedPrv = Omit<Keychain, 'encryptedPrv'> & {
+  encryptedPrv: string;
+};
 
 export interface ChangedKeychains {
   [pubkey: string]: string;
--- a/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
+++ b/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
@@ -6,6 +6,22 @@ import { BitGoKeyFromOvcShares, OvcToBitGoJSON } from './ovcJsonCodec';
 export type KeyType = 'tss' | 'independent' | 'blsdkg';
 export type SourceType = 'bitgo' | 'backup' | 'user' | 'cold';
 
+export type WebauthnFmt = 'none' | 'packed' | 'fido-u2f';
+export interface WebauthnAuthenticatorInfo {
+  credID: string;
+  fmt: WebauthnFmt;
+  publicKey: string;
+}
+
+export interface KeychainWebauthnDevice {
+  otpDeviceId: string;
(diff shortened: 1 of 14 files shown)


modules/sdk-core/src/bitgo/keychain/iKeychains.ts
```diff
@@ -20,8 +36,16 @@ export interface Keychain {
   walletHSMGPGPublicKeySigs?: string;
   type: KeyType;
   source?: SourceType;
+  // If the user has opted for password-less wallets, this is expected to be populated
+  webauthnDevices?: KeychainWebauthnDevice[];
```
zahin-mohammad, line 40: Introduced in https://github.com/BitGo/bitgo-microservices/pull/38527

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1357,14 +1357,13 @@ export class Wallet implements IWallet {
...
   /**
-   * Gets the user key chain for this wallet
+   * Gets the user keychain for this wallet
    *
-   * The user key chain is the first keychain of the wallet and usually has the encrypted prv stored on BitGo.
+   * The user keychain is the first keychain of the wallet and usually has the encrypted prv stored on BitGo.
    * Useful when trying to get the users' keychain from the server before decrypting to sign a transaction.
-   * @param params
    */
-  async getEncryptedUserKeychain(params: Record<string, never> = {}): Promise<{ encryptedPrv: string }> {
```
mmcshinsky-bitgo, line 1366: Does this change mean that the following are to be expected in the returned object?
```  id: string;
  pub?: string;
  prv?: string;
  provider?: string;
  encryptedPrv?: string;
  derivationPath?: string;
  derivedFromParentWithSeed?: string;
  commonPub?: string;
  commonKeychain?: string;
  keyShares?: ApiKeyShare[];
  walletHSMGPGPublicKeySigs?: string;
  type: KeyType;
  source?: SourceType;
```
from Keychain?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1408,16 +1407,14 @@ export class Wallet implements IWallet {
...
-
-    let userPrv;
-    try {
-      userPrv = this.bitgo.decrypt({ input: userEncryptedPrv, password: params.walletPassphrase });
-    } catch (e) {
-      throw new Error('error decrypting wallet passphrase');
+    const userKeychain = await this.getEncryptedUserKeychain();
+    if (!params.walletPassphrase) {
+      throw new Error('wallet passphrase was not provided');
+    }
```
alebusse, line 1413:
```suggestion
```
Probably not needed since we check if the `prv` and `walletPassphrase` are undefined, and we return if the `prv` exist, meaning that the other case is that the `walletPassphrase` exists

modules/sdk-core/src/bitgo/wallet/wallet.ts
zahin-mohammad: Yup! It's returning the keychain from the wp api.

zahin-mohammad: Actually not 100% on some of these fields... I don't think the following are valid fields that returned from the api: 
- keyShares
-  walletHSMGPGPublicKeySigs
will double check.

zahin-mohammad: Just double checked, those fields are only returned during the key creation workflows. 

Its a little unfortunate, but the `GET` and `POST` codecs in wp re-use the same response even though this is not valid. The `GET` call will return less data as the `POST` since the `POST` call has some ephemeral data. 

According to the codec, this type for this SDK call is valid, but ik its wrong. Ill create a ticket to address the codec issue, and then a separate ticket to fix the SDK. 

- https://bitgoinc.atlassian.net/browse/WP-1283
- https://bitgoinc.atlassian.net/browse/WP-1284

modules/sdk-core/src/bitgo/wallet/wallet.ts
zahin-mohammad: ah fair, this is redundant.

zahin-mohammad: Ah i remember why I added this now, there isn't enough type narrowing here for typescript to know that `walletPassphrase` is not undefined at this point. To add some better type narrowing requires some refactors that I didn't want to introduce at this point.

alebusse (approved):

davidkaplanbitgo (approved):
btc related files look fine

modules/sdk-core/src/bitgo/wallet/wallet.ts
mmcshinsky-bitgo: Awesome, thanks!

PR #4207, opened 2024-01-21
fix(abstract-utxo): assert locking script before signing

saravanan7mani:
Before signing, locking scripts in a PSBT should be asserted that they match valid BitGo scripts for any of supported 2-of-3 or p2shP2pk.

This change should cover for all signing components like OVC, WRW, etc.

Ticket: BTC-815


rushilbg (commented):
Add a negative test?

rushilbg (approved):

zahin-mohammad (approved):

PR #4212, opened 2024-01-22
feat(sdk-coin-hbar): implement recover method for native hbar

alebusse:
implemented recover method for native hbar

WP-1289

TICKET: WP-1289

Test using an alpha release in WRW
unsigned sweep + ovc : https://hashscan.io/testnet/transaction/1706202635.110903357
non-bitgo recovery : https://hashscan.io/testnet/transaction/1706200972.000778003

### DIFF
--- a/modules/sdk-coin-hbar/src/hbar.ts
+++ b/modules/sdk-coin-hbar/src/hbar.ts
@@ -6,7 +6,6 @@ import {
   BaseCoin,
   BitGoBase,
   KeyPair,
-  MethodNotImplementedError,
   ParsedTransaction,
   ParseTransactionOptions,
   SignedTransaction,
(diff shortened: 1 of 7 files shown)


modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -249,8 +286,134 @@ export class Hbar extends BaseCoin {
...
+        json: txJson,
+      };
+    }
+
+    txBuilder.sign({ key: userPrv });
+    txBuilder.sign({ key: backUp });
+
+    const tx = await txBuilder.build();
+
+    const result = await this.broadcastTransaction(client, tx.toBroadcastFormat());
```
alebusse, line 409: Hbar is very sensitive to the selected nodeId and the broadcast time window, so instead of just returning the broadcasteable tx here, we broadcast the tx.

gianchandania (dismissed):

modules/sdk-coin-hbar/src/hbar.ts
zahin-mohammad: Discussed in focus hour, we will split this out to a separate step and build a new re-usable form in WRW for this.

zahin-mohammad (changes requested):

gianchandania (dismissed):

zahin-mohammad (approved):

margueriteblair (approved):

PR #4217, opened 2024-01-24
fix(sdk-coin-dot): fix baseFactor for tdot

dpkjnr:
Ticket: WIN-1783

### DIFF
--- a/modules/sdk-coin-dot/src/dot.ts
+++ b/modules/sdk-coin-dot/src/dot.ts
@@ -361,7 +361,7 @@ export class Dot extends BaseCoin {
     const partialFee = await this.getFee(destAddr, senderAddr, amount);
     // Polkadot has a concept of existential desposit (ed), it is the minimum amount required by an address to have
     // to keep the account active
-    const existentialDeposit = this.getChain() === 'tdot' ? 10000000000 : 1000000000000;
+    const existentialDeposit = 1 * this.getBaseFactor();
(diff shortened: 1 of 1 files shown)


gianchandania (approved):

modules/sdk-coin-dot/src/dot.ts
```diff
@@ -361,7 +361,7 @@ export class Dot extends BaseCoin {
     const partialFee = await this.getFee(destAddr, senderAddr, amount);
     // Polkadot has a concept of existential desposit (ed), it is the minimum amount required by an address to have
     // to keep the account active
-    const existentialDeposit = this.getChain() === 'tdot' ? 10000000000 : 1000000000000;
+    const existentialDeposit = 1 * this.getBaseFactor();
```
ajays97, line 364: We can ignore multiplying this by 1 since `this.getBaseFactor()` is enough.

ajays97 (changes requested):

modules/sdk-coin-dot/src/dot.ts
ajays97:
```suggestion
    const existentialDeposit = this.getBaseFactor();
```

alebusse:
Nice catch, im might take over with this pr https://github.com/BitGo/BitGoJS/pull/4219

dpkjnr:
fixed in https://github.com/BitGo/BitGoJS/pull/4219

PR #4224, opened 2024-01-25
fix(sdk-core): sign txRequest full in pendingApproval.approve

zahin-mohammad:
Prior to this PR, the `approve` method on the Pending Approval class accepted a `walletPassphrase` but did not sign transactions for pending approvals of type `transactionRequestFull`. This PR adds support for this pending approval type by adding a private `postApprove` function that signs the txRequest after the approval is done.

TICKET: WP-1186

### DIFF
--- a/modules/bitgo/test/v2/unit/pendingApproval.ts
+++ b/modules/bitgo/test/v2/unit/pendingApproval.ts
@@ -6,13 +6,14 @@ import * as nock from 'nock';
 import * as sinon from 'sinon';
 
 import { TestableBG, TestBitGo } from '@bitgo/sdk-test';
-import { BitGo } from '../../../src/bitgo';
+import { BitGo } from '../../../src';
 
 import {
   BaseCoin,
   Environments,
   PendingApproval,
   PendingApprovalData,
+  PendingApprovalInfo,
   RequestTracer,
   SignatureShareType,
   State,
@@ -58,6 +59,7 @@ describe('Pending Approvals:', () => {
 
   beforeEach(function () {
     sandbox = sinon.createSandbox();
+    nock.disableNetConnect();
   });
 
   afterEach(function () {
@@ -65,7 +67,6 @@ describe('Pending Approvals:', () => {
   });
 
   before(async () => {
-    nock.disableNetConnect();
     // create wallet
     bitgo = TestBitGo.decorate(BitGo, { env: 'mock' });
     bitgo.initializeTestVars();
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -1000,6 +1000,7 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
    * @returns {Promise<TxRequest>} fully signed TxRequest object
    */
   async signTxRequest(params: TSSParams): Promise<TxRequest> {
+    this.bitgo.setRequestTracer(params.reqId);
```
zahin-mohammad, line 1003: we never used this param before...

modules/bitgo/test/v2/unit/pendingApproval.ts
```diff
@@ -6,13 +6,14 @@ import * as nock from 'nock';
 import * as sinon from 'sinon';
 
 import { TestableBG, TestBitGo } from '@bitgo/sdk-test';
-import { BitGo } from '../../../src/bitgo';
+import { BitGo } from '../../../src';
```
zahin-mohammad, line 9: Ignore whitespace when viewing this file

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
```diff
@@ -374,4 +300,136 @@ export class PendingApproval implements IPendingApproval {
...
+  }
+
+  /**
+   * Internal helper function to perform any post-approval actions.
+   * If type is 'transactionRequestFull', this will sign the txRequestFull if possible
+   * @param params
+   * @param reqId
+   * @private
+   */
+  private async postApprove(params: ApproveOptions = {}, reqId: IRequestTracer): Promise<void> {
```
YPAzevedo, line 385: this is fine, but i did get a little confused when i was reading the function that uses it, because this reads like we are using a post method to approve

YPAzevedo (approved):

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
zahin-mohammad: we can re-name the pre and post to something like `handlePreApproval` and `handlePostApproval` ?

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
```diff
@@ -374,4 +300,136 @@ export class PendingApproval implements IPendingApproval {
...
+      recipients.length === 0 &&
+      type !== 'consolidate'
+    );
+  }
+
+  /*
+   * Internal helper function to get the serialized transaction which is being approved.
+   * If this PA is of type 'transactionRequest' this function will try to rebuild and resign the transaction
+   * @param {ApproveOptions} params
+   * @param {boolean} canRecreateTransaction -
```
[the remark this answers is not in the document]
zahin-mohammad: Good catch, I removed it at one point, forgot to update the jsDOC. Ill remove this in a follow up PR since its on a private method.

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
YPAzevedo: i think that reads a bit more clear :)

alebusse (approved):

PR #4235, opened 2024-01-29
feat(statics): adds ofcsbc in coins.ts

shivagupta94:
Ticket: PX-2979

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -10124,6 +10124,7 @@ export const coins = CoinMap.fromCoins([
   ofcerc20('67fb4a6f-ba00-41fd-972d-728d2226a3d5', 'ofccho', 'Choise.com', 18, UnderlyingAsset.CHO),
   ofcerc20('af198c54-53d5-42b3-9e69-7fcc5887c0a0', 'ofcdia', 'DIA Token', 18, UnderlyingAsset.DIA),
   ofcerc20('4da7f6a6-9c56-44f7-a9b2-07ae7c669e42', 'ofcldo', 'Lido DAO Token', 18, UnderlyingAsset.LDO),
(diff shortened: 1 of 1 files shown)


modules/statics/src/coins.ts
```diff
@@ -10124,6 +10124,7 @@ export const coins = CoinMap.fromCoins([
   ofcerc20('67fb4a6f-ba00-41fd-972d-728d2226a3d5', 'ofccho', 'Choise.com', 18, UnderlyingAsset.CHO),
   ofcerc20('af198c54-53d5-42b3-9e69-7fcc5887c0a0', 'ofcdia', 'DIA Token', 18, UnderlyingAsset.DIA),
   ofcerc20('4da7f6a6-9c56-44f7-a9b2-07ae7c669e42', 'ofcldo', 'Lido DAO Token', 18, UnderlyingAsset.LDO),
+  ofcerc20('4da7f6a6-9c56-44f7-a9b2-07ae7c669e42', 'ofcsbc', 'Sustainable Bitcoin Certificate', 8, UnderlyingAsset.SBC),
```
manav-c, line 10127: What does the first argument mean here? Seems like it should be different for every coin. Currently it's set to the same value as the coin above this one

Phani024, line 10127: The uuid is same as the uuid for `ofcldo` ^

shivagupta94: Yeah, i missed that

Phani024 (changes requested):
Blocking since the uuid is duplicate

modules/statics/src/coins.ts
shivagupta94: Changed

shivagupta94: Yeah, missed

Phani024 (approved):
LGTM

mandalarnav (approved):

PR #4238, opened 2024-01-29
fix(sdk-core): fix hash for tss ecdsa PA signing

alebusse:
Moved hash selection to EcDSA utils to fix PA signing and avoid the need to pass the hash to the signer

WP-1186

TICKET: WP-1186


mkottaichamy (approved):
lgtm

PR #4249, opened 2024-02-01
fix(abstract-eth): change tx encoding method for arb and op

gianchandania:
This PR changes encoding type used for txdata for arbitrum and optimism
based on the new contracts that will be used for all EVM coins going
forward. This new encoding type is applicable only for native coins. This should not affect other
EVM coins like Eth, Polygon, etc.

WIN-1867
TICKET: WIN-1867


aarvay (dismissed):

modules/abstract-eth/src/lib/transferBuilder.ts
```diff
@@ -93,8 +95,9 @@ export class TransferBuilder {
     throw new InvalidParameterValueError('Invalid expiration time');
   }
 
-  signAndBuild(chainId?: string): string {
+  signAndBuild(chainId?: string, coinUsesNonPackedEncodingForTxData?: boolean): string {
     this._chainId = chainId;
+    this._coinUsesNonPackedEncodingForTxData = coinUsesNonPackedEncodingForTxData;
```
mullapudipruthvik, line 100: I'm not sure why are we initializing the variable here, which is not used in the function itself?
Probably better to have a separate function to set like `expirationTime` 
or you could initialize in the constructor ?

ajays97 (dismissed):
Looks good.

modules/abstract-eth/src/lib/transferBuilder.ts
```diff
@@ -133,7 +136,20 @@ export class TransferBuilder {
    */
   private getOperationHash(): string {
     const operationData = this.getOperationData();
-    return ethUtil.bufferToHex(EthereumAbi.soliditySHA3(...operationData));
+    let operationHash: string;
+    // If the coin uses non-packed encoding for tx data, the operation hash is calculated differently
+    // This new encoding type is applicable only for native coins and not tokens
+    if (this._coinUsesNonPackedEncodingForTxData && this._tokenContractAddress === undefined) {
+      const types: string[] = operationData[0] as string[];
```
mullapudipruthvik, line 143: Why do we have to check for `tokenContractAddress` here, can we somehow check it before initializing the `_coinUsesNonPackedEncodingForTxData` variable

mullapudipruthvik (commented):
LGTM apart from few questions

modules/abstract-eth/src/lib/transferBuilder.ts
gianchandania: The reason why we did it like this is so that there are no changes required wherever transferBuilder is being called. If it is done as a separate method like expirationTime, then we will have to modify the places in SDK/Wallet platform, and it will be a breaking change. The reason why it is initialized in this method is because signAndBuild is being called from transactionBuilder as shown [here](https://github.com/BitGo/BitGoJS/blob/fb84e7ffafaf655b87b71c5019df910760835499/modules/abstract-eth/src/lib/transactionBuilder.ts#L617) and hence we used this method to initialize this param which is going to be used in during building the transaction

mullapudipruthvik (approved):

PR #4251, opened 2024-02-01
fix(sdk-core): add check to validate if PA triggered another condition

alebusse:
Approving a PA can return either the PA doc or an object containing an error and the PA when the
PA triggered another condition and requires an extra approval

CS-2592

TICKET: CS-2592


LeoKotschenreuther (dismissed):

davidkaplanbitgo (dismissed):

alebusse:
Sorry, i reworked the fix since we found the root cause of the issue

davidkaplanbitgo (approved):

PR #4252, opened 2024-02-02
fix(sdk-coin-trx): fix withdrawal issue with trx:sun token

gianchandania:
We were using number type for amount during trx token withdrawals. As trx:sun token has 18 decimals, so we were getting issues when we were trying to convert the amount to BigNumber if the amount is more than 1e^21. This PR changes the type for amount as string to solve the issue

CR-1041
TICKET: CR-1041


sachushaji (approved):
LGTM

PR #4253, opened 2024-02-02
feat: rectify the external/internal recipients handling for RBF

rushilbg:
`verifyTransaction()` needed to be correctly filtering out the external recipients of the original tx to be replaced while verifying the built replacement tx.

BTC-820

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -522,8 +522,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     if (txParams.rbfTxIds) {
       assert(txParams.rbfTxIds.length === 1);
 
-      // TODO: Pass `includeRbf: true` in the `getTransaction` request
-      const txToBeReplaced = await wallet.getTransaction({ txHash: txParams.rbfTxIds[0] });
+      const txToBeReplaced = await wallet.getTransaction({ txHash: txParams.rbfTxIds[0], includeRbf: true });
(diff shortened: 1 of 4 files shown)


modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -516,19 +516,19 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
     const allOutputs = [...explanation.outputs, ...explanation.changeOutputs];
 
     let expectedOutputs;
     if (txParams.rbfTxIds) {
       assert(txParams.rbfTxIds.length === 1);
-      const txToBeReplacedTransfer = await wallet.getTransfer({ id: txParams.rbfTxIds[0] });
+
+      const txToBeReplaced = await wallet.getTransaction({ txHash: txParams.rbfTxIds[0], includeRbf: true });
       // Note: Will work only when there is single transaction output per address
       // TODO: https://bitgoinc.atlassian.net/browse/BTC-826
```
OttoAllmendinger, line 527:
```suggestion

```

OttoAllmendinger (approved):
nit

modules/abstract-utxo/src/abstractUtxoCoin.ts
rushilbg: Oh yeah using transaction doesn't warrant this TODO now.

rushilbg: Raising PR.

PR #4254, opened 2024-02-02
fix(trx): add token list from statics

sachushaji:
tron transaction builder was using token contract
address to identify how the inputs and outputs
needs to be parsed for the transaction.
Changing the logic so that all tokenAddresses
onboarded in statics is considered for
parsing insead of hardcoding the addresses.

Ticket: CR-1041

### DIFF
--- a/modules/sdk-coin-trx/src/lib/utils.ts
+++ b/modules/sdk-coin-trx/src/lib/utils.ts
@@ -4,6 +4,7 @@ import * as tronweb from 'tronweb';
 import { protocol } from '../../resources/protobuf/tron';
 
 import { UtilsError } from '@bitgo/sdk-core';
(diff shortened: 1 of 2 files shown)


victor-bitgo:
Please add in the commit comment the context around the change: why is it being made? The "what "is discoverable from the diff, but the "why" is not, so it is important to convey the "what" in the commit comment. This will make it easier for future readers of the code to understand why changes were made.

sachushaji:
> Please add in the commit comment the context around the change: why is it being made? The "what "is discoverable from the diff, but the "why" is not, so it is important to convey the "what" in the commit comment. This will make it easier for future readers of the code to understand why changes were made.

updated @victor-bitgo

grseetha (approved):

PR #4259, opened 2024-02-06
feat(abstract-eth): add v4 forwarder create txtype

gianchandania:
This PR updates address initialization logic in transaction builder to support new v4 forwarders

WIN-1926
TICKET: WIN-1926


modules/abstract-eth/src/lib/utils.ts
```diff
@@ -672,37 +676,98 @@ export function getV1WalletInitializationData(walletOwners: string[], salt: stri
...
- * Returns the create address method calling data for v1 wallets
+ * Returns the create address method calling data for v1, v2, v4 forwarders
  *
  * @param {string} baseAddress - The address of the wallet contract
  * @param {string} salt - The salt for address initialization transactions
+ * @param {string} feeAddress - The fee address for the enterprise
  * @returns {string} - the createForwarder method encoded
  */
-export function getV1AddressInitializationData(baseAddress: string, salt: string): string {
+export function getV1AddressInitializationData(baseAddress: string, salt: string, feeAddress?: string): string {
```
mullapudipruthvik, line 686: nit: the method name is misleading

modules/abstract-eth/src/lib/utils.ts
```diff
@@ -672,37 +676,98 @@ export function getV1WalletInitializationData(walletOwners: string[], salt: stri
...
+  let createForwarderParams;
+  let createForwarderTypes;
+  if (feeAddress) {
+    createForwarderParams = [baseAddress, feeAddress, saltBuffer];
+    createForwarderTypes = createV4ForwarderTypes;
+  } else {
+    createForwarderParams = [baseAddress, saltBuffer];
+    createForwarderTypes = createV1ForwarderTypes;
+  }
+  return { createForwarderParams, createForwarderTypes };
```
mullapudipruthvik, line 744:
```suggestion
  let createForwarderParams = [baseAddress, saltBuffer];;
  let createForwarderTypes = createV1ForwarderTypes;
  if (feeAddress) {
    createForwarderParams = [baseAddress, feeAddress, saltBuffer];
    createForwarderTypes = createV4ForwarderTypes;
  } 
  return { createForwarderParams, createForwarderTypes };
```

modules/abstract-eth/src/lib/utils.ts
```diff
@@ -711,6 +776,18 @@ export function decodeForwarderCreationData(data: string): ForwarderInitializati
...
+  } else {
+    const [baseAddress, feeAddress, saltBuffer] = getRawDecoded(
+      createV4ForwarderTypes,
+      getBufferedByteCode(v4CreateForwarderMethodId, data)
+    );
+
+    return {
+      baseAddress: addHexPrefix(baseAddress as string),
+      feeAddress: addHexPrefix(feeAddress as string),
+      addressCreationSalt: bufferToHex(saltBuffer as Buffer),
```
mullapudipruthvik, line 790:
```suggestion
      addressCreationSalt: bufferToHex(saltBuffer as Buffer),
      } as const
```

modules/abstract-eth/src/lib/utils.ts
```diff
@@ -672,37 +676,98 @@ export function getV1WalletInitializationData(walletOwners: string[], salt: stri
...
+ * @returns {string} - the createForwarder method encoded
+ *
+ */
+export function getAddressInitDataAllForwarderVersions(
+  forwarderVersion: number,
+  baseAddress: string,
+  salt: string,
+  feeAddress?: string
+): string {
+  if (forwarderVersion === defaultForwarderVersion) {
```
mullapudipruthvik, line 715: Whats the `defaultForwarderVersion` now ? how is it in sync with the one in wp?
should we have it per coin?

gianchandania: Actually this defaultForwarderVersion is not same as WP as it is not needed here ideally. In WP defaultForwarderVersion is per coin basis. This was mainly added when we were onboarding polygon and we had to add logic related to v1/v2 forwarder init. So we had referred to v0 as defaultForwarderVersion

modules/abstract-eth/src/lib/utils.ts
gianchandania: There are couple of instances where naming are misleading in sdk,  just because when they were added it was only for v1, but then later those methods were reused for other wallet/forwarder versions as well. I think we probably should update all of them in one PR as that will be a breaking change

modules/abstract-eth/src/lib/utils.ts
gianchandania: updated

modules/abstract-eth/src/lib/utils.ts
gianchandania: Have created a follow up ticket with all the method/variable names to be renamed. https://bitgoinc.atlassian.net/browse/WIN-1964

modules/abstract-eth/src/lib/utils.ts
mullapudipruthvik: So I have try to create a forwarder using bitgojs without giving any version, it will default to 0 ?

modules/sdk-coin-opeth/test/unit/transactionBuilder/walletInitialization.ts
```diff
@@ -0,0 +1,109 @@
+import should from 'should';
+import { TransactionType } from '@bitgo/sdk-core';
+import { TransactionBuilder } from '../../../src';
+import * as testData from '../../resources';
+import { coins } from '@bitgo/statics';
+
+describe('Arbeth wallet initialization', function () {
```
sachushaji, line 7:
```suggestion
describe('Opeth wallet initialization', function () {
```

modules/abstract-eth/src/lib/utils.ts
gianchandania: I tried by calling sdk createAddress method without giving a forwarderVersion, it seems like it is eventually deciding on which forwarder to create in wallet platform and not sdk when it is calling wallet platform create address API [here](https://github.com/BitGo/BitGoJS/blob/303e0dd0e89e61a87d3c8edcb012012cf6b35f34/modules/sdk-core/src/bitgo/wallet/wallet.ts#L1215)

modules/sdk-coin-opeth/test/unit/transactionBuilder/walletInitialization.ts
gianchandania: updated

modules/abstract-eth/src/lib/utils.ts
mullapudipruthvik: Then I think we should somehow remove `defaultForwarderVersion`  probably in a follow-up ticket

mullapudipruthvik (approved):

PR #4268, opened 2024-02-07
feat(root): multisig eddsa root keys test

alebusse:
test

WP-0000


modules/sdk-coin-hbar/test/unit/hbar.ts
```diff
@@ -939,4 +939,33 @@ describe('Hedera Hashgraph:', function () {
...
+      const derivedPub = await basecoin.deriveKeyWithSeed({ key: rootKeyPair.pub, seed });
+      assert.ok(derivedPub.derivationPath === expectedDerivationPath);
+      assert.ok(derivedPub.key === expectedPub);
+
+      const derivedPrv = await basecoin.deriveKeyWithSeed({ key: rootKeyPair.prv, seed });
+      assert.ok(derivedPrv.derivationPath === expectedDerivationPath);
+      derivedPrv.key.should.equal(expectedPrv);
+
+      const derivedKeychain = new KeyPair({ prv: derivedPrv.key }).getKeys();
+      derivedKeychain.pub.should.equal(expectedPub);
```
alebusse, line 967: Fails here, why the derivedKeychain from prv doesnt generate the same pub as the derivation from a pub.

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -592,4 +604,78 @@ export class Hbar extends BaseCoin {
...
+      const pub = rootKeyPair.getKeys(true).pub;
+
+      // generic code that could be moved out of this function
+      const pk = bigIntFromBufferLE(Buffer.from(pub, 'hex'));
+      const derivedKeychain = hdTree.privateDerive({ sk, pk, chaincode, prefix }, derivationPath);
+      const derivedPrv = bigIntToBufferLE(derivedKeychain.sk).toString('hex');
+      // const derivedPub = bigIntToBufferLE(derivedKeychain.pk).toString('hex');
+
+      // coinspecific code
+      const derivedHbarPrv = new HbarKeyPair({ prv: derivedPrv }).getKeys().prv;
```
islamaminBitGo, line 648: i thought we figured out we need to append a pub to the derivedPrv for Hbar to use this prv as we expect

islamaminBitGo (changes requested):

modules/sdk-coin-hbar/src/hbar.ts
alebusse: This derivedPrv is the "hbar" formatted prv, appending the pub to the doesnt work
Also, appending the pub means that we always used the prv as a seed based on what we saw where the checks expect prv length === 64 and the seed === 32

modules/sdk-coin-hbar/test/unit/hbar.ts
alebusse: @islamaminBitGo  This derivedPrv is the "hbar" formatted prv, appending the pub to the doesnt work.

modules/sdk-coin-hbar/src/hbar.ts
islamaminBitGo: so derivedPrv above seems to me like it's not formatted for any coin yet until line 648. I remember when we looked at the code we found out that when u pass a 32 byte prv to HbarKeyPair it uses as a seed. That's not what we want. We want it to use it as is. So for that u had to append the pub key cause when prv is 64 bytes the first 32 bytes are used as the prv and the last 32 bytes are used as the pub key

modules/sdk-coin-hbar/test/unit/hbar.ts
islamaminBitGo: well the fix would be applied to the hbar.ts as im outlining below not here

modules/sdk-coin-hbar/src/hbar.ts
alebusse: Sorry, editted the msg, at this point we only care about the prv and the return of this is the formatted prv, is the same result if you use prv or prv+pub (my guess cause of the KeyPair implementation)

PR #4271, opened 2024-02-08
fix(bitgo): fix updating the wallet build defaults

davidkaplanbitgo:
TICKET: BTC-900

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -1289,7 +1289,13 @@ export class Wallet implements IWallet {
     common.validateParams(params, [], ['minFeeRate', 'changeAddressType', 'txFormat']);
     return this.bitgo
       .put(this.url())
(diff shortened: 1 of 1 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1289,7 +1289,13 @@ export class Wallet implements IWallet {
     common.validateParams(params, [], ['minFeeRate', 'changeAddressType', 'txFormat']);
     return this.bitgo
       .put(this.url())
-      .send({ minFeeRate: params.minFeeRate, changeAddressType: params.changeAddressType, txFormat: params.txFormat })
+      .send({
+        buildDefaults: {
```
rushilbg, line 1293: Isn't the arg type changed with this? Is it expected?

davidkaplanbitgo: Yes, the old arg type is not what `v2.wallet.update` was expecting

rushilbg (approved):

PR #4280, opened 2024-02-13
feat(root): eddsa root keys

alebusse:
test eddsa root keys

WP-0000

TICKET: WP-0000


modules/sdk-core/src/account-lib/index.ts
```diff
@@ -5,5 +5,6 @@ export * from './mpc';
 export * from './util/crypto';
 // Deprecated
 export * as acountLibCrypto from './util/crypto';
+// Deprecated
```
mmcshinsky-bitgo, line 8:
```suggestion
/**
  * @deprecated
*/
```

mmcshinsky-bitgo, line 8:
```suggestion
```
unneeded comment already covered by the added @deprecated

PR #4287, opened 2024-02-15

evalenti-simtlix:
add new batch of tokens

EA-3112

TICKET: EA-3112


modules/statics/src/coins.ts
```diff
@@ -12173,6 +12211,15 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.SRM,
     SOL_TOKEN_FEATURES
   ),
+  solToken(
+    'a81dd35d-30ef-47c8-85a0-dada2695c6e3',
+    'sol:mplx',
```
mmcshinsky-bitgo, line 12216: This token already exists

evalenti-simtlix: Yes, that was my mistake

mmcshinsky-bitgo (approved):

PR #4303, opened 2024-02-22
fix(statics): update zetachain to zetaevm

evalenti-simtlix:
update zetachain to zetaevm

EA-3126

TICKET: EA-3126

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -1521,7 +1521,7 @@ export enum UnderlyingAsset {
   ZCO = 'zco',
   ZECBEAR = 'zecbear',
   ZECBULL = 'zecbull',
-  ZETACHAIN = 'zetachain',
+  ZETAEVM = 'zetaevm',
   ZIL = 'zil',
   ZIP = 'zip',
   ZIX = 'zix',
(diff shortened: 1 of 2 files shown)


hitansh-bitgo:
Should be marked as breaking change in the commit footer

rushilbg (approved):

modules/statics/src/base.ts
```diff
@@ -1521,7 +1521,7 @@ export enum UnderlyingAsset {
   ZCO = 'zco',
   ZECBEAR = 'zecbear',
   ZECBULL = 'zecbull',
-  ZETACHAIN = 'zetachain',
```
ajays97, line 1524: What are the consequences of removing this ENUM value downstream? Can we please check

evalenti-simtlix: I think the consequences of removing is anyone, because i'm updating this one. But i need to update in Indexer and BGMS also.

alebusse (changes requested):
Mark it as breaking change, i think this is gonna leave every transfer or already saved doc with the old coin name invalid, i dont know if we ever did a name change like this

ewangbitgo (approved):

mmcshinsky-bitgo (changes requested):
Badly formatted commit. BREAKING CHANGE must be in the commit footer.

mmcshinsky-bitgo (approved):

CapnMigraine (approved):

alebusse (approved):

PR #4306, opened 2024-02-26
fix(ecdsa): exclude key derive call

mullapudipruthvik:
Ticket: WIN-1935

TICKET: WIN-1935


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -628,11 +629,16 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
...
     const MPC = new Ecdsa();
-    const signingKey = MPC.keyDerive(
-      userSigningMaterial.pShare,
-      [userSigningMaterial.bitgoNShare, userSigningMaterial.backupNShare],
-      derivationPath
-    );
+    // const signingKey = MPC.keyDerive(
+    //   userSigningMaterial.pShare,
+    //   [userSigningMaterial.bitgoNShare, userSigningMaterial.backupNShare],
+    //   derivationPath
```
mmcshinsky-bitgo, line 635: Removal of the required field `derivationPath` presents a false narrative to users of the sdk on how these functions are used up through `createOfflineKShare` and `UnsignedTransactionTss`. We probably need a better solution than commenting it out.

PR #4308, opened 2024-02-26
feat(statics): add token as available for switzerland

mukeshsp:
Ticket: EA-3157

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -258,6 +258,7 @@ const COSMOS_SIDECHAIN_FEATURES = [
   CoinFeature.BULK_STAKING_TRANSACTION,
 ];
 const INJECTIVE_FEATURES = [...COSMOS_SIDECHAIN_FEATURES, CoinFeature.CUSTODY_BITGO_SWITZERLAND];
+const TOKEN_FEATURES_WITH_SWISS = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.CUSTODY_BITGO_SWITZERLAND];
 const GENERIC_TOKEN_FEATURES = [
   CoinFeature.ACCOUNT_MODEL,
   CoinFeature.REQUIRES_BIG_NUMBER,
@@ -3972,7 +3973,8 @@ export const coins = CoinMap.fromCoins([
     'Injective (ERC20)',
     18,
     '0xe28b3b32b6c345a34ff64674606124dd5aceca30',
-    UnderlyingAsset.INJV2
+    UnderlyingAsset.INJV2,
+    TOKEN_FEATURES_WITH_SWISS
   ),
   erc20(
     '108a64ac-9be8-4bf9-a683-3f30d49708de',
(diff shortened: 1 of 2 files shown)


mukeshsp:
> I have my doubts about this solution, because that implies that each token that we want to be visible to Switzerland, we need to add a feature flag here, in addition to having (or not), the token in the gatekept list (bgms).
> 
> Here I think it should be transparent just as it happens with the rest of public tokens on Bitgo US and the others.

that is also added using the DEFAULT_FEATURES which common for all the tokens

modules/statics/src/coins.ts
```diff
@@ -6390,15 +6392,17 @@ export const coins = CoinMap.fromCoins([
...
   ),
   erc20(
     'e1d609f1-eb3f-4b3f-8ef6-87f434a6de83',
     'usdt',
     'Tether',
     6,
     '0xdac17f958d2ee523a2206206994597c13d831ec7',
-    UnderlyingAsset.USDT
+    UnderlyingAsset.USDT,
+    TOKEN_FEATURES_WITH_SWISS
```
grsind19, line 6405: @mukeshsp  if we add another region, this naming will become irrelevant.
How we handle this ? can you find a better way ?

mukeshsp: In that case, we will add this feature with new variable similar to this case 
https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/coins.ts#L260

evalenti-simtlix:
> > I have my doubts about this solution, because that implies that each token that we want to be visible to Switzerland, we need to add a feature flag here, in addition to having (or not), the token in the gatekept list (bgms).
> > Here I think it should be transparent just as it happens with the rest of public tokens on Bitgo US and the others.
> 
> that is also added using the DEFAULT_FEATURES which common for all the tokens

I mean, it is right for now, but what happens if in a near future Switzerland wants to use all ERC20 tokens? we need to add one by one the flag? I right as temporaly solution, but at least from my side, is not the best way. Also, with that, in each ticket for token onboarding need to specify if it will be used by Switzerland or not.

mukeshsp:
> > > I have my doubts about this solution, because that implies that each token that we want to be visible to Switzerland, we need to add a feature flag here, in addition to having (or not), the token in the gatekept list (bgms).
> > > Here I think it should be transparent just as it happens with the rest of public tokens on Bitgo US and the others.
> > 
> > 
> > that is also added using the DEFAULT_FEATURES which common for all the tokens
> 
> I mean, it is right for now, but what happens if in a near future Switzerland wants to use all ERC20 tokens? we need to add one by one the flag? I right as temporaly solution, but at least from my side, is not the best way. Also, with that, in each ticket for token onboarding need to specify if it will be used by Switzerland or not.

If it is require for all the erc20 in that case we can make it default feature and that will not require addition of one by one.

aarvay (approved):

ravneet-bitgo (approved):

PR #4335, opened 2024-03-06
Win 2260 fix recovery method

gianchandania:
This PR fixes the recovery method for eth and ethlike coins

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -1362,6 +1362,9 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
 
     // get balance of wallet
     const txAmount = await this.queryAddressBalance(params.walletContractAddress);
+    if (new BigNumber(txAmount).isLessThanOrEqualTo(0)) {
+      throw new Error('Wallet does not have enough funds to recover');
+    }
 
     // build recipients object
     const recipients = [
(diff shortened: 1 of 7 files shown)


modules/sdk-coin-arbeth/test/unit/arbeth.ts
```diff
@@ -585,6 +585,38 @@ describe('Arbitrum', function () {
       recipient.amount.should.equal('9999999999999999928');
     });
 
+    it('should generate an unsigned sweep', async function () {
```
sachushaji, line 588: Should throw error should be the test description?

sachushaji (dismissed):

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -1362,6 +1362,9 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
 
     // get balance of wallet
     const txAmount = await this.queryAddressBalance(params.walletContractAddress);
+    if (txAmount.toString(10) === '0') {
```
alebusse, line 1365: I would use bignumber just in case and also check its equal or lesser than 0
```suggestion
    if (new BigNumber(txAmount).lte(0)) {
```

modules/sdk-coin-arbeth/test/unit/arbeth.ts
gianchandania: my bad, updated

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
gianchandania: done

alebusse (approved):

sachushaji (approved):

PR #4348, opened 2024-03-14
feat(statics): add new token MYRC and fix MSN address

evalenti-simtlix:
add new token MYRC and fix MSN address

EA-3328

TICKET: EA-3328

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -4719,6 +4719,14 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.MWT
   ),
   erc20(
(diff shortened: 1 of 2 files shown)


modules/statics/src/coins.ts
```diff
@@ -9151,7 +9159,7 @@ export const coins = CoinMap.fromCoins([
     'msn',
     'meson.network',
     18,
-    '0x6c4d441f41e423b88a0edc001fe15bc4657b3abf',
+    '0xaa247c0d81b83812e1abf8bab078e4540d87e3fb',
```
mullapudipruthvik, line 9162: how will we make sure WP and indexer is in sync with this change?

mullapudipruthvik (approved):

PR #4349, opened 2024-03-14
fix: factor gas and eip1559 params on ethlike token recovery

mohammadalfaiyazbitgo:
Ticket: WP-1552

### DIFF
--- a/modules/sdk-coin-eth/src/erc20Token.ts
+++ b/modules/sdk-coin-eth/src/erc20Token.ts
@@ -159,9 +159,11 @@ export class Erc20Token extends Eth {
     const userKey = params.userKey.replace(/\s/g, '');
     const backupKey = params.backupKey.replace(/\s/g, '');
 
-    // Set new eth tx fees (using default config values from platform)
-    const gasPrice = this.getRecoveryGasPrice();
-    const gasLimit = this.getRecoveryGasLimit();
+    // Set new eth tx fees (default to using platform values if none are provided)
+    const gasPrice = params.eip1559
+      ? new optionalDeps.ethUtil.BN(params.eip1559.maxFeePerGas)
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-eth/src/erc20Token.ts
```diff
@@ -283,7 +285,7 @@ export class Erc20Token extends Eth {
     });
 
     if (isUnsignedSweep) {
-      return this.formatForOfflineVault(txInfo, tx, userKey, backupKey, gasPrice, gasLimit) as any;
+      return this.formatForOfflineVault(txInfo, tx, userKey, backupKey, gasPrice, gasLimit, params.eip1559) as any;
```
mullapudipruthvik, line 288: incase of Non-Bitgo Recovery for cold wallet, it doesn't not interact with WP
1. Build unsigned sweep using WRW
2. Sign using OVC(first signature)
3. Sign using OVC (second signature)

For the above case I feel your change breaks it
Can we have some default values if we don't get it from params?
else we can make these fields mandatory in WRW

mohammadalfaiyazbitgo: I'm not sure which change in the PR you're referring to.
For the highlighted piece, the function typing handles an undefined `eip1559` params.

mohammadalfaiyazbitgo: The other piece is doing the same thing as eth for recovery: https://github.com/BitGo/BitGoJS/blob/7744d831f7d974802501245312787b81f8abae47/modules/abstract-eth/src/abstractEthLikeNewCoins.ts#L1292

mullapudipruthvik: nvm I see setGasPrice method returns some default value

mullapudipruthvik (approved):

PR #4358, opened 2024-03-19
fix(sdk-coin-avaxp): add support for recipient in P > C import tx

CapnMigraine:
TICKET: CR-1060

### DIFF
--- a/modules/sdk-coin-avaxp/test/unit/avaxp.ts
+++ b/modules/sdk-coin-avaxp/test/unit/avaxp.ts
@@ -415,7 +415,24 @@ describe('Avaxp', function () {
     });
   });
 
-  describe('verify transaction', function () {
+  describe('Verify transaction', function () {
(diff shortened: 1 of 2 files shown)


mohammadalfaiyazbitgo (approved):

modules/sdk-coin-avaxp/test/unit/avaxp.ts
```diff
@@ -510,7 +532,7 @@ describe('Avaxp', function () {
       const txParams = {
         recipients: [
           {
-            address: EXPORT_P_2_C_VERIFY.receiveAddress,
+            address: '',
```
noel-bitgo, line 535: this empty string expected?

CapnMigraine: It's expected, `address` is required in `recipients`. 
We could change that in the future

noel-bitgo (approved):

PR #4376, opened 2024-03-27

evalenti-simtlix:
add new batch of tokens

EA-3421

TICKET: EA-3421

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -4879,6 +4879,14 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.OMG
   ),
   erc20(
+    'def47964-e90c-4862-a724-086e439d01de',
+    'omnia',
+    'OMNIA Protocol',
+    18,
+    '0x2e7e487d84b5baba5878a9833fb394bc89633fd7',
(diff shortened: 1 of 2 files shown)


modules/statics/src/coins.ts
```diff
@@ -7536,6 +7544,14 @@ export const coins = CoinMap.fromCoins([
...
     UnderlyingAsset.XTZBULL
   ),
+  erc20(
+    'fec45ebe-80a0-4cfa-8fb3-7c4100517f18',
+    'xusd',
+    'XUSD',
+    6,
+    '0xc08e7e23c235073c6807c2efe7021304cb7c2815',
+    UnderlyingAsset.XUSD
+  ),
```
Vijay-Jagannathan, line 7554: Why are we not adding this ?

gianchandania (approved):

zahin-mohammad (approved):

modules/statics/src/coins.ts
evalenti-simtlix: We are not adding this in the base.ts because we already have this token for other coin (In this case, Solana)

sandra-wang-bitgo (approved):

linyah9912 (approved):

PR #4395, opened 2024-04-04
fix(abstract-eth): fix opeth:op transfers

gianchandania:
WIN-2643
TICKET: WIN-2643

This PR fixes the issue with optimism token transfers. Optimism token has the same contract address on eth and op chain, which is causing issues while transferring the token.


sachushaji (commented):
Can we add tests

gianchandania:
> Can we add tests

added

sachushaji (approved):

PR #4407, opened 2024-04-08
feat(utxo-lib): find psbt outputs of internal and external bitgo wallets

saravanan7mani:
Trustlessly find psbt output indices of internal bitgo wallet
Trustlessly calculate psbt output amount of internal bitgo wallet
Find psbt output indices of external bitgo wallet
Calculate psbt output amount of external bitgo wallet

ref https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#change-detection

TICKET: BTC-1070

### DIFF
--- a/modules/utxo-lib/src/bitgo/wallet/Psbt.ts
+++ b/modules/utxo-lib/src/bitgo/wallet/Psbt.ts
@@ -6,7 +6,13 @@ import { BIP32Interface } from 'bip32';
 import * as bs58check from 'bs58check';
 import { UtxoPsbt } from '../UtxoPsbt';
 import { UtxoTransaction } from '../UtxoTransaction';
-import { createOutputScript2of3, getLeafHash, scriptTypeForChain, toXOnlyPublicKey } from '../outputScripts';
+import {
+  createOutputScript2of3,
+  getLeafHash,
+  ScriptType2Of3,
+  scriptTypeForChain,
+  toXOnlyPublicKey,
+} from '../outputScripts';
 import { DerivedWalletKeys, RootWalletKeys } from './WalletKeys';
 import { toPrevOutputWithPrevTx } from '../Unspent';
 import { createPsbtFromHex, createPsbtFromTransaction } from '../transaction';
@@ -26,6 +32,7 @@ import {
   ParsedScriptType,
   isPlaceholderSignature,
   parseSignatureScript,
+  ParsedScriptType2Of3,
 } from '../parseInput';
 import { parsePsbtMusig2PartialSigs } from '../Musig2';
 import { isTuple, Triple } from '../types';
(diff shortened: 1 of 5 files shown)


modules/utxo-lib/src/bitgo/wallet/Psbt.ts
```diff
@@ -489,6 +497,231 @@ export function addXpubsToPsbt(psbt: UtxoPsbt, rootWalletKeys: RootWalletKeys):
...
+  parsedScriptType,
+  network,
+}: {
+  publicKeys: Buffer[];
+  perm: Triple<number>;
+  scriptPubKey: Buffer;
+  parsedScriptType: ParsedScriptType2Of3;
+  network: Network;
+}): boolean {
+  const orderedPublicKeys: Triple<Buffer> = [publicKeys[perm[0]], publicKeys[perm[1]], publicKeys[perm[2]]];
```
OttoAllmendinger, line 562: instead of passing indices we could pass the buffers themselves

modules/utxo-lib/src/bitgo/wallet/Psbt.ts
```diff
@@ -489,6 +497,231 @@ export function addXpubsToPsbt(psbt: UtxoPsbt, rootWalletKeys: RootWalletKeys):
...
+  network: Network
+): Triple<number> {
+  const permutations: Array<Triple<number>> = [
+    [0, 1, 2],
+    [0, 2, 1],
+    [1, 0, 2],
+    [1, 2, 0],
+    [2, 0, 1],
+    [2, 1, 0],
+  ];
```
OttoAllmendinger, line 591: instead of doing it by index we can do it by buffer

```
const [a, b, c] = publicKeys;
const permutations = [
 [a, b, c],
 [a, c, b],
 ...
]`
```

one less layer of indirection

modules/utxo-lib/src/bitgo/wallet/Psbt.ts
```diff
@@ -489,6 +497,231 @@ export function addXpubsToPsbt(psbt: UtxoPsbt, rootWalletKeys: RootWalletKeys):
   psbt.updateGlobal({ globalXpub: xPubs });
 }
 
+/**
+ * Derives the appropriate BIP32 key pair for a given output.
+ * It uses either tapBip32Derivation or bip32Derivation paths from the output, considering taproot specifics.
+ * @param bip32 - The BIP32Interface object to derive from.
+ * @param output - The specific PSBT output to derive for.
+ * @returns The derived BIP32 key pair, or the original bip32 if no derivation was possible.
```
OttoAllmendinger, line 505: ... or undefined in what case?

OttoAllmendinger (changes requested):
You are adding quite a lot of code and blowing up `Psbt.ts`.

Let's add add a directory `/psbt/` and I guess `/psbt/rootNodes.ts` or something

OttoAllmendinger (changes requested):

modules/utxo-lib/src/bitgo/wallet/psbt/PsbtOutputs.ts
```diff
@@ -0,0 +1,52 @@
+import { isBufferArray, Triple } from '../../types';
```
OttoAllmendinger, line 1:
```suggestion
/**
 * Contains helper methods for determining if a transaction output belongs to a given BitGo wallet
 */
 
import { isBufferArray, Triple } from '../../types';
```

OttoAllmendinger (changes requested):
code looks good but pls add some jsdoc

OttoAllmendinger (approved):

PR #4413, opened 2024-04-09
fix(sdk-coin-zeta): fix zeta tests

alebusse:
fixed zeta tests that had the wrong node url mock

WP-1692

TICKET: WP-1692


mkottaichamy (approved):
lgtm

PR #4414, opened 2024-04-09
feat(bitgo): add PATCH and OPTIONS to redirectRequest

jayfgrollman:
Currently if someone tries to use BitGo Express to do a PATCH or OPTIONS call, we will fail because the code we have does not have a path for either of these HTTP Methods.

This PR adds paths for both methods, which should allow users to be able to call endpoints like EVS's PATCH identity endpoint via BitGo Express.

Ticket: COPS-1288

NOTE: Please merge when fully approved.

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -994,7 +994,7 @@ function handleV2CoinSpecificREST(req: express.Request, res: express.Response, n
 }
 
 /**
- * Redirect a request using the bitgo request functions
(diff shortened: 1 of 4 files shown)


gianchandania (commented):
Tests seem to be failing, please re-request review once build is passed

bitgoAaron (approved):

therealdwright (approved):

sumanmukherjee03 (approved):

alebusse (approved):

PR #4419, opened 2024-04-10
feat: modify accept share method

ravneet-bitgo:
Ticket: PX-3296

## Description
We are modifying the accept share method for go account wallet invitations. If the `keychainOverrideRequired` key is present in the walletShare, we need to create a new keychain and pass the same to the `updateShare` method so we can override the existing key. After successful completion of the same, we need to `reshareWallet` to all the existing spenders.

TDD: [link](https://docs.google.com/document/d/1Dll1Z2OnVi2ofR__4LlqDdEI1aNUPuLXrqnITPQ0sf8/edit#bookmark=id.tr9x4ewh6fhv)

### DIFF
--- a/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
+++ b/modules/sdk-core/src/bitgo/keychain/iKeychains.ts
@@ -181,4 +181,5 @@ export interface IKeychains {
   getKeysForSigning(params?: GetKeysForSigningOptions): Promise<Keychain[]>;
   createMpc(params: CreateMpcOptions): Promise<KeychainsTriplet>;
   createTssBitGoKeyFromOvcShares(ovcOutput: OvcToBitGoJSON): Promise<BitGoKeyFromOvcShares>;
+  createUserKeychain(userPassword: string): Promise<Keychain>;
 }
--- a/modules/sdk-core/src/bitgo/keychain/keychains.ts
+++ b/modules/sdk-core/src/bitgo/keychain/keychains.ts
@@ -3,7 +3,7 @@ import assert from 'assert';
 import * as common from '../../common';
 import { IBaseCoin, KeychainsTriplet, KeyPair } from '../baseCoin';
 import { BitGoBase } from '../bitgoBase';
-import { BlsUtils, RequestTracer, EDDSAUtils, ECDSAUtils, decodeOrElse } from '../utils';
+import { BlsUtils, RequestTracer, EDDSAUtils, ECDSAUtils, decodeOrElse, generateRandomPassword } from '../utils';
 import {
   AddKeychainOptions,
   ApiKeyShare,
@@ -425,6 +425,29 @@ export class Keychains implements IKeychains {
 
     return decodeOrElse(BitGoKeyFromOvcShares.name, BitGoKeyFromOvcShares, output, (errors) => {
       throw new Error(`Error producing the output: ${errors}`);
+    });
+  }
+
+  /**
+   * Create keychain for ofc wallet using the password
+   * @param userPassword
+   * @returns
+   */
+  async createUserKeychain(userPassword: string): Promise<Keychain> {
+    const keychains = this.baseCoin.keychains();
+    const newKeychain = keychains.create();
+    const originalPasscodeEncryptionCode = generateRandomPassword(5);
+
+    const encryptedPrv = this.bitgo.encrypt({
+      password: userPassword,
+      input: newKeychain.prv,
+    });
+
+    return await keychains.add({
+      encryptedPrv,
+      originalPasscodeEncryptionCode,
+      pub: newKeychain.pub,
+      source: 'user',
     });
   }
 }
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -28,6 +28,7 @@ import {
 import { ILightning } from '../lightning';
 import { SerializedNtilde } from '../../account-lib/mpc/tss/ecdsa/types';
 import { IAddressBook } from '../address-book';
+import { WalletUser } from '@bitgo/public-types';
 
 export interface MaximumSpendableOptions {
   minValue?: number | string;
@@ -609,6 +610,7 @@ export interface WalletData {
   tokens?: Record<string, any>[];
   nfts?: { [contractAddressOrToken: string]: NftBalance };
   unsupportedNfts?: { [contractAddress: string]: NftBalance };
+  users?: WalletUser[];
 }
 
 export interface RecoverTokenOptions {
--- a/modules/sdk-core/src/bitgo/wallet/wallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallets.ts
@@ -539,6 +539,44 @@ export class Wallets implements IWallets {
   }
 
   /**
+   * Re-share wallet with existing spenders of the wallet
+   * @param walletId
+   * @param userPassword
+   */
+  async reshareWalletWithSpenders(walletId: string, userPassword: string): Promise<void> {
+    const wallet = await this.get({ id: walletId });
+    if (!wallet?._wallet?.enterprise) {
+      throw new Error('Enterprise not found for the wallet');
+    }
+
+    const enterpriseUsersResponse = await this.bitgo
+      .get(`/api/v2/enterprise/${wallet?._wallet?.enterprise}/user`)
+      .result();
+    // create a map of users for easy lookup - we need the user email id to share the wallet
+    const usersMap = new Map(
+      [...enterpriseUsersResponse?.adminUsers, ...enterpriseUsersResponse?.nonAdminUsers].map((obj) => [obj.id, obj])
+    );
+
+    if (wallet._wallet.users) {
+      for (const user of wallet._wallet.users) {
+        const userObject = usersMap.get(user.user);
+        if (user.permissions.includes('spend') && !user.permissions.includes('admin') && userObject) {
+          const shareParams = {
+            walletId: walletId,
+            user: user.user,
+            permissions: user.permissions.join(','),
+            walletPassphrase: userPassword,
+            email: userObject.email.email,
+            reshare: true,
+            skipKeychain: false,
+          };
+          await wallet.shareWallet(shareParams);
+        }
+      }
+    }
+  }
+
+  /**
    * Accepts a wallet share, adding the wallet to the user's list
    * Needs a user's password to decrypt the shared key
    *
@@ -554,9 +592,31 @@ export class Wallets implements IWallets {
     common.validateParams(params, ['walletShareId'], ['overrideEncryptedPrv', 'userPassword', 'newWalletPassphrase']);
 
     let encryptedPrv = params.overrideEncryptedPrv;
-
-    const walletShare = (await this.getShare({ walletShareId: params.walletShareId })) as any;
-
+    const walletShare = await this.getShare({ walletShareId: params.walletShareId });
+    if (
+      walletShare.keychainOverrideRequired &&
+      walletShare.permissions.indexOf('admin') !== -1 &&
+      walletShare.permissions.indexOf('spend') !== -1
+    ) {
+      if (_.isUndefined(params.userPassword)) {
+        throw new Error('userPassword param must be provided to decrypt shared key');
+      }
+      const walletKeychain = await this.baseCoin.keychains().createUserKeychain(params.userPassword);
+      const response = await this.updateShare({
+        walletShareId: params.walletShareId,
+        state: 'accepted',
+        keyId: walletKeychain.id,
+      });
+      // If the wallet share was accepted successfully (changed=true), reshare the wallet with the spenders
+      if (response.changed && response.state === 'accepted') {
+        try {
+          await this.reshareWalletWithSpenders(walletShare.wallet, params.userPassword);
+        } catch (e) {
+          // Do nothing
+        }
+      }
+      return response;
+    }
     // Return right away if there is no keychain to decrypt, or if explicit encryptedPrv was provided
     if (!walletShare.keychain || !walletShare.keychain.encryptedPrv || encryptedPrv) {
       return this.updateShare({
@@ -606,7 +666,6 @@ export class Wallets implements IWallets {
     if (encryptedPrv) {
       updateParams.encryptedPrv = encryptedPrv;
     }
-
     return this.updateShare(updateParams);
   }
 
--- a/modules/bitgo/test/v2/unit/wallets.ts
+++ b/modules/bitgo/test/v2/unit/wallets.ts
@@ -11,6 +11,7 @@ import {
   ECDSAUtils,
   KeychainsTriplet,
   GenerateWalletOptions,
+  Wallet,
 } from '@bitgo/sdk-core';
 import * as _ from 'lodash';
 import { TestBitGo } from '@bitgo/sdk-test';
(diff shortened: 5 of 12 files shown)


modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -506,10 +508,7 @@ export class Wallets implements IWallets {
   async updateShare(params: UpdateShareOptions = {}): Promise<any> {
     common.validateParams(params, ['walletShareId'], []);
 
-    return await this.bitgo
-      .post(this.baseCoin.url('/walletshare/' + params.walletShareId))
-      .send(params)
-      .result();
+    return await this.bitgo.post(this.baseCoin.url('/walletshare/' + params.walletShareId)).send(params);
```
ravneet-bitgo, line 511: We have removed `.result()` because we need to check the status code of this API later.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,77 @@ export class Wallets implements IWallets {
...
+      }
+    });
+  }
+
+  /**
+   * Generate a random password
+   * @param   {Number} numWords     Number of 32-bit words
+   * @returns {String}          base58 random password
+   */
+  generateRandomPassword(numWords = 5): string {
```
shivagupta94, line 585: Lets ADD TODO comment here, if we can find some better place for this method

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -609,6 +609,7 @@ export interface WalletData {
   tokens?: Record<string, any>[];
   nfts?: { [contractAddressOrToken: string]: NftBalance };
   unsupportedNfts?: { [contractAddress: string]: NftBalance };
+  users?: any[];
```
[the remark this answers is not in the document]
zahin-mohammad: +1, please type this out. Even if it is not fully typed, add what known fields you are using.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,77 @@ export class Wallets implements IWallets {
...
   }
 
+  /**
+   * Re-share wallet with existing spenders of the wallet
+   * @param walletId
+   * @param userPassword
+   */
+  async reshareOfcAccountWithSpenders(walletId: string, userPassword: string): Promise<void> {
+    const wallet = await this.bitgo.coin('ofc').wallets().get({ id: walletId });
+    const enterpriseUsersResponse = (await this.bitgo
```
zahin-mohammad, line 550: Assert `wallet.wallet.enterprise` above.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,77 @@ export class Wallets implements IWallets {
...
+      .get(`/api/v2/enterprise/${wallet?._wallet?.enterprise}/user`)
+      .result()) as any;
```
zahin-mohammad, line 552: Why is this typed as any?

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,77 @@ export class Wallets implements IWallets {
...
+            permissions: user.permissions.join(','),
+            walletPassphrase: userPassword,
+            email: userObject.email.email,
+            coin: wallet.coin,
+            reshare: true,
+          };
+          wallet.shareWallet(shareParams);
+        }
+      } catch (e) {
+        // TODO: gracefully handle this error
```
zahin-mohammad, line 574: Please address this TODO.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,77 @@ export class Wallets implements IWallets {
...
+        console.error(e);
```
zahin-mohammad, line 575: Reove this console.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,77 @@ export class Wallets implements IWallets {
...
+  }
+
+  /**
+   * Create keychain for ofc wallet using the password
+   * @param userPassword
+   * @returns
+   */
+  async createKeychain(userPassword: string): Promise<Keychain> {
+    const sdkCoin = await this.bitgo.coin('ofc');
+    const keychains = sdkCoin.keychains();
```
zahin-mohammad, line 597:
```suggestion
    const keychains = this.baseCoin.keychains();
```

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -554,9 +627,28 @@ export class Wallets implements IWallets {
...
 
     let encryptedPrv = params.overrideEncryptedPrv;
-
     const walletShare = (await this.getShare({ walletShareId: params.walletShareId })) as any;
-
+    if (walletShare.keychainOverrideRequired && walletShare.permissions.indexOf('admin') !== -1) {
+      if (_.isUndefined(params.userPassword)) {
+        throw new Error('userPassword param must be provided to decrypt shared key');
+      }
+      const walletKeychain = await this.createKeychain(params.userPassword);
```
zahin-mohammad, line 635: Why are we creating a new keychain? Shouldn't it just be an update?

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -554,9 +627,28 @@ export class Wallets implements IWallets {
...
+      const response = await this.updateShare({
+        walletShareId: params.walletShareId,
+        state: 'accepted',
+        keyId: walletKeychain.id,
+      });
+      // If the wallet share was accepted successfully (changed=true), reshare the wallet with the spenders
+      if (response.changed && response.state === 'accepted') {
+        try {
+          await this.reshareOfcAccountWithSpenders(walletShare.wallet, params.userPassword);
```
zahin-mohammad, line 644: Is `keychainOverrideRequired` only set for a single wallet share on the OFC wallet? Is there a case where two users have this set to true in their wallet share?

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -554,9 +627,28 @@ export class Wallets implements IWallets {
...
+        } catch (e) {
+          // TODO: gracefully handle this error
+          console.error(e);
```
zahin-mohammad, line 647: Address these TODO's

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,77 @@ export class Wallets implements IWallets {
       .result();
   }
 
+  /**
+   * Re-share wallet with existing spenders of the wallet
+   * @param walletId
+   * @param userPassword
+   */
+  async reshareOfcAccountWithSpenders(walletId: string, userPassword: string): Promise<void> {
```
zahin-mohammad, line 548: Are these methods meant to be exposed? If so they need to updated in the wallet interface in `iWallet.ts`

zahin-mohammad (changes requested):
Please add tests

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: We only need to generate a new keychain when keychainOverrideRequired is set to true, indicating the intentional need to create a new keychain and replace the existing one in the wallet.

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: > Is there a case where two users have this set to true in their wallet share?

Yes, it's possible for two different wallet share requests from different users to have `keychainOverrideRequired` set to true. However, it's important to note that only the Admin APIs, accessible exclusively to the support team, can set this `keychainOverrideRequired` flag.

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: Do you suggest adding a ticket or providing a thorough resolution?

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: Do you suggest adding a ticket or providing a thorough resolution?

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +542,75 @@ export class Wallets implements IWallets {
...
+      }
+    });
+  }
+
+  /**
+   * Generate a random password
+   * @param   {Number} numWords     Number of 32-bit words
+   * @returns {String}          base58 random password
+   */
+  generateRandomPassword(numWords = 5): string {
```
manav-c, line 586: why are you redefining this? it already exists in the SDK - https://github.com/BitGo/BitGoJS/blob/97c0f3ce5aa83bd4d412ba278be931473bc7c923/modules/sdk-api/src/bitgoAPI.ts#L1106

modules/sdk-core/src/bitgo/wallet/wallets.ts
manav-c: https://github.com/BitGo/BitGoJS/pull/4419/files#r1566982186

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +542,75 @@ export class Wallets implements IWallets {
...
+    const keychains = this.baseCoin.keychains();
+    const newKeychain = keychains.create();
+    const originalPasscodeEncryptionCode = this.generateRandomPassword();
+
+    const encryptedPrv = this.bitgo.encrypt({
+      password: userPassword,
+      input: newKeychain.prv,
+    });
+
+    const walletKeychain = await keychains.add({
```
manav-c, line 606: is the keychain also being added as part of the `updateShare` flow? @shivagupta94

modules/sdk-core/src/bitgo/wallet/wallets.ts
ravneet-bitgo: We are aware of this, moving it to a util method and import it at both the places because as of now the sdk-api is not imported in sdk-core (circular dependency issue).

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: > is the keychain also being added as part of the updateShare flow? @shivagupta94
You mean in the wallet?

So, we need to create a keychain on the client side and save it in the backend using `keychains.add` method. Then, when triggering the updateShare method, we should pass the newly generated keychain ID in the payload. This way, we can replace it with the existing key ID.

manav-c: No. To the DB. This step - https://docs.google.com/document/d/1Dll1Z2OnVi2ofR__4LlqDdEI1aNUPuLXrqnITPQ0sf8/edit#bookmark=id.a9tdqvm0ql4f

modules/sdk-core/src/bitgo/utils/ofc.ts
```diff
@@ -0,0 +1,12 @@
+import * as sjcl from '@bitgo/sjcl';
+import * as bs58 from 'bs58';
+
+/**
+ * Generate a random password
+ * @param   {Number} numWords     Number of 32-bit words
+ * @returns {String}          base58 random password
+ */
+export function generateRandomPassword(numWords: number): string {
```
manav-c, line 9: is this just an ofc util?

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -554,9 +617,27 @@ export class Wallets implements IWallets {
...
+        walletShareId: params.walletShareId,
+        state: 'accepted',
+        keyId: walletKeychain.id,
+      });
+      // If the wallet share was accepted successfully (changed=true), reshare the wallet with the spenders
+      if (response.changed && response.state === 'accepted') {
+        try {
+          await this.reshareOfcAccountWithSpenders(walletShare.wallet, params.userPassword);
+        } catch (e) {
+          debug('failed to reshare wallet with spenders', e);
```
manav-c, line 636: we should ideally be very loud in this case and raise alarms because this means current spenders on the wallet cannot sign txns (due to the key being replaced) and manual intervention is required. Someone should be notified.

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: We're using the `keychains.add` function for this purpose. It saves the keychain details in the database and provides us with a reference ID for the same.

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: I agree. Initially, we discussed this, but since we don't have any features available on the client side to capture these errors and push them to the backend or raise any kind of alarm, we have to log them only.

However, we could potentially create a new API that can consume these errors and notify us about such cases, allowing us to take necessary steps.

Alternatively, we could consider throwing the error instead of consuming it. This way, we can inform newly added users that no spender can perform any transaction due to this issue. @manav-c , what are your thoughts on this approach?

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: > I was just asking if you're doing this during updateShare too and you're not

Yes we are adding keychain to DB on client side only

> if updateShare fails for any reason, then this key will be unused (not on any wallet)

I agree, implementing that approach would require setting up a rollback mechanism as well. Since creating a keychain is a prerequisite for this acceptshare flow, ensuring we can revert any erroneous actions would be essential.

modules/sdk-core/src/bitgo/wallet/wallets.ts
manav-c: Yeah I guess we could abort from this flow so the invited user has to accept their invite again

modules/sdk-core/src/bitgo/wallet/wallets.ts
manav-c: > I agree, implementing that approach would require setting up a rollback mechanism as well

You can use mongo transactions in WP for this

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: So ideally, it's the responsibility of the support team to ensure that only one invitation is sent at a time.

manav-c: As a follow up, we can come back and put some logic in the admin API (that sends out the invitation) to abort if there is already an active WalletShare for a wallet with the `keychainOverrideRequired` property set to true.

manav-c: > So ideally, it's the responsibility of the support team to ensure that only one invitation is sent at a time

Ideally, we should not rely on this.

manav-c:
please add tests

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -575,6 +575,11 @@ export interface CrossChainUTXO {
 export type WalletType = 'backing' | 'cold' | 'custodial' | 'custodialPaired' | 'hot' | 'trading';
 export type SubWalletType = 'distributedCustody';
 
+export type WalletUser = {
```
mmcshinsky-bitgo, line 578: Can we pull this from `@bitgo/public-types`. If not, we may want to add it to provide full types.

zahin-mohammad: +1

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -554,9 +615,27 @@ export class Wallets implements IWallets {
...
 
     let encryptedPrv = params.overrideEncryptedPrv;
-
-    const walletShare = (await this.getShare({ walletShareId: params.walletShareId })) as any;
-
+    const walletShare = await this.getShare({ walletShareId: params.walletShareId });
+    if (walletShare.keychainOverrideRequired && walletShare.permissions.indexOf('admin') !== -1) {
+      if (_.isUndefined(params.userPassword)) {
+        throw new Error('userPassword param must be provided to decrypt shared key');
+      }
```
zahin-mohammad, line 622: Throwing here is misleading. Encrypted key data is only shared with people with `spend` permissions, not `admin`. You can technically be an admin and not a spender.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,65 @@ export class Wallets implements IWallets {
       .result();
   }
 
+  /**
+   * Re-share wallet with existing spenders of the wallet
+   * @param walletId
+   * @param userPassword
+   */
+  async reshareOfcAccountWithSpenders(walletId: string, userPassword: string): Promise<void> {
```
zahin-mohammad, line 548:
```suggestion
  async reshareWalletWithSpenders(walletId: string, walletPassword: string): Promise<void> {
```

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,65 @@ export class Wallets implements IWallets {
...
+    const wallet = await this.bitgo.coin('ofc').wallets().get({ id: walletId });
```
zahin-mohammad, line 549:
```suggestion
    const wallet = await this.get({ id: walletId });
```

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,65 @@ export class Wallets implements IWallets {
...
+          user: user.user,
+          permissions: user.permissions.join(','),
+          walletPassphrase: userPassword,
+          email: userObject.email.email,
+          coin: wallet.coin,
+          reshare: true,
+        };
+        wallet.shareWallet(shareParams);
+      }
+    });
```
zahin-mohammad, line 576: Consider using `Promise.all` in conjunction with a `map`. Map all wallet users to a promise, then await them all.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -538,6 +540,65 @@ export class Wallets implements IWallets {
...
+  }
+
+  /**
+   * Create keychain for ofc wallet using the password
+   * @param userPassword
+   * @returns
+   */
+  async createKeychain(userPassword: string): Promise<Keychain> {
```
zahin-mohammad, line 584: This should live in the `keychains` class if its not already there.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -554,9 +615,27 @@ export class Wallets implements IWallets {
...
+      const walletKeychain = await this.createKeychain(params.userPassword);
+      const response = await this.updateShare({
+        walletShareId: params.walletShareId,
+        state: 'accepted',
+        keyId: walletKeychain.id,
+      });
+      // If the wallet share was accepted successfully (changed=true), reshare the wallet with the spenders
+      if (response.changed && response.state === 'accepted') {
+        try {
+          await this.reshareOfcAccountWithSpenders(walletShare.wallet, params.userPassword);
```
zahin-mohammad, line 632:
```suggestion
          await this.reshareWalletWithSpenders(walletShare.wallet, params.userPassword);
```

modules/sdk-core/src/bitgo/wallet/wallets.ts
zahin-mohammad: Please address the concern here. If it's possible to have two admins with the keychain override, then wallet spenders will get invited multiple times when they do not need to. There should be restrictions in place to ensure only one admin at a time is doing this op.

modules/sdk-core/src/bitgo/wallet/wallets.ts
zahin-mohammad: Please throw the error in this case. Debug logs will get swalloed by the client side and should not exist in our production sdk code.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -28,6 +28,8 @@ import {
 } from './iWallets';
 import { Wallet } from './wallet';
 
+const debug = require('debug')('bitgo:v2:wallets');
```
zahin-mohammad, line 31:
```suggestion
```

zahin-mohammad (changes requested):
Have some suggestions/concerns on the implementation here. Please see the review comments.

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: @rajucse fyi^

modules/sdk-core/src/bitgo/wallet/wallets.ts
ravneet-bitgo: This is the userPassword and not walletPassword.

Screenshot 2024-04-18 at 12 36 19 PM

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: @zahin-mohammad 
I had a discussion with the team [here](https://bitgo.slack.com/archives/C06MHA22P08/p1713423281043779) regarding the concern you raised. The team has decided to prioritize picking up the improvement later, and for the time being, we will proceed with the current approach.

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: I had a discussion with the team [here](https://bitgo.slack.com/archives/C06MHA22P08/p1713423281043779) regarding the concern you raised. The team has decided to prioritize picking up the improvement later, and for the time being, we will proceed with the current approach.

shivagupta94:
> Please add tests

Added

modules/sdk-core/src/bitgo/wallet/wallets.ts
zahin-mohammad: > they will have encrypted key data, which holds true for two roles: admin and spenders.

This is not true for the amdin role. There is no gaurentee that the `admin` has the prv material. 

You need to check both admin and spender permissions here, not just admin.

modules/sdk-core/src/bitgo/keychain/keychains.ts
```diff
@@ -427,4 +427,28 @@ export class Keychains implements IKeychains {
       throw new Error(`Error producing the output: ${errors}`);
     });
   }
+
+  /**
+   * Create keychain for ofc wallet using the password
+   * @param userPassword
+   * @returns
+   */
+  async createWalletKeychain(userPassword: string): Promise<Keychain> {
```
zahin-mohammad, line 436:
```suggestion
  async createUserKeychain(userPassword: string): Promise<Keychain> {
```

zahin-mohammad: Its not technically a wallet keychain until you associate the key with the wallet. Keychains are standalone objects.

modules/sdk-core/src/bitgo/keychain/keychains.ts
```diff
@@ -427,4 +427,28 @@ export class Keychains implements IKeychains {
...
+      input: newKeychain.prv,
+    });
+
+    const walletKeychain = await keychains.add({
+      encryptedPrv,
+      originalPasscodeEncryptionCode,
+      pub: newKeychain.pub,
+      source: 'user',
+    });
+    return walletKeychain;
```
zahin-mohammad, line 452:
```suggestion
    return await keychains.add({
      encryptedPrv,
      originalPasscodeEncryptionCode,
      pub: newKeychain.pub,
      source: 'user',
    });
```

modules/sdk-core/src/bitgo/keychain/iKeychains.ts
```diff
@@ -181,4 +181,5 @@ export interface IKeychains {
   getKeysForSigning(params?: GetKeysForSigningOptions): Promise<Keychain[]>;
   createMpc(params: CreateMpcOptions): Promise<KeychainsTriplet>;
   createTssBitGoKeyFromOvcShares(ovcOutput: OvcToBitGoJSON): Promise<BitGoKeyFromOvcShares>;
+  createWalletKeychain(userPassword: string): Promise<Keychain>;
```
zahin-mohammad, line 184:
```suggestion
  createUserKeychain(userPassword: string): Promise<Keychain>;
```

modules/sdk-core/src/bitgo/wallet/wallets.ts
zahin-mohammad: Please remove this.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -554,9 +594,27 @@ export class Wallets implements IWallets {
     common.validateParams(params, ['walletShareId'], ['overrideEncryptedPrv', 'userPassword', 'newWalletPassphrase']);
 
     let encryptedPrv = params.overrideEncryptedPrv;
-
-    const walletShare = (await this.getShare({ walletShareId: params.walletShareId })) as any;
-
+    const walletShare = await this.getShare({ walletShareId: params.walletShareId });
+    if (walletShare.keychainOverrideRequired && walletShare.permissions.indexOf('admin') !== -1) {
```
zahin-mohammad, line 598:
```suggestion
    if (walletShare.keychainOverrideRequired && walletShare.permissions.indexOf('admin') !== -1 && walletShare.permissions.indexOf('spend') !== -1) {
```

modules/sdk-core/src/bitgo/wallet/wallets.ts
zahin-mohammad: This is not acceptible for production code. Please throw and remove the debug logs.

zahin-mohammad (changes requested):

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: Done

modules/sdk-core/src/bitgo/keychain/iKeychains.ts
shivagupta94: Done

modules/sdk-core/src/bitgo/keychain/keychains.ts
shivagupta94: Done

modules/sdk-core/src/bitgo/keychain/keychains.ts
shivagupta94: Done

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: Done

modules/sdk-core/src/bitgo/wallet/wallets.ts
shivagupta94: Done

zahin-mohammad (approved):

PR #4427, opened 2024-04-12
feat: fix getUnspentsForAddresses in BlockchairApi

lcovar:
When we pass an array of addresses we make a single call to the blockchair API with a comma separated list of addresses. The response includes the list of all utxos for every address in the list.

Example response

```
  "utxo": [
    {
      "block_id": 838353,
      "transaction_hash": "19a2e26860981413ceab82052738386d72d530333b3a7e0237518e242c65524d",
      "index": 0,
      "value": 42000,
      "address": "34EaQXdjELqWrTFxJWFgmVmF91kPhdP4us"
    },
    {
      "block_id": 838353,
      "transaction_hash": "19a2e26860981413ceab82052738386d72d530333b3a7e0237518e242c65524d",
      "index": 1,
      "value": 657269,
      "address": "bc1qv669qw5tzxsue5qrl36ygqnaahm98n9t68hdz577cz2r8eldyj3q6d90c4"
    }
  ]

```

However, with the code as it is now, for every address we map all the utxos in the response to the address. In the above example we will have a return value of an array of size 4 instead of the expected size 2. Associated unspents to addresses it doesnt belong to.

Ticket: BTC-1084

### DIFF
--- a/modules/blockapis/src/impl/BlockchairApi.ts
+++ b/modules/blockapis/src/impl/BlockchairApi.ts
@@ -37,6 +37,7 @@ type BlockchairUnspent = {
   script_hex: string;
   spending_transaction_hash: string;
   spending_index: number;
+  address: string;
 };
 
 // https://blockchair.com/api/docs#link_300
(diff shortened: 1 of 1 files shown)


rushilbg (commented):
This is purely logical, would benefit from a unit test.

OttoAllmendinger:
let's refactor the code a bit first so it is less reliant on nocks and add a unit test after that

OttoAllmendinger (approved):

PR #4430, opened 2024-04-12
Combine partial signatures util support

islamaminBitGo:


zahin-mohammad (changes requested):
Discueed in slack, but we'll need another util in sdk-core to produce the fully signed message from 2 broadcast messages (msg4 from user and msg4 from bitgo)

zahin-mohammad (approved):

PR #4434, opened 2024-04-12
feat(statics): add ofc token to represent Erc20 token on Arbitrum chain

subavicky:
- add test cases to test one mainnent and one testnet arb erc20 token

TICKET: PX-3568

### DIFF
--- a/modules/bitgo/test/v2/unit/coins/ofcToken.ts
+++ b/modules/bitgo/test/v2/unit/coins/ofcToken.ts
@@ -284,6 +284,29 @@ describe('OFC:', function () {
     });
   });
 
+  describe('check ofc tokens for arbethErc20', function () {
+    const tokenMain = 'ofcarbeth:link';
+    const tokenTest = 'ofctarbeth:link';
+    describe('for main network', function () {
+      it(`should have the correct values for ${tokenMain}`, function () {
+        const ofcCoin = bitgo.coin(tokenMain);
+        ofcCoin.getChain().should.equal(tokenMain);
+        ofcCoin.getFullName().should.equal('Chainlink Token');
+        ofcCoin.getBaseFactor().should.equal(PRECISION_18);
+        ofcCoin.addressCoin.should.equal('arbeth');
+      });
+    });
+    describe('for test network', function () {
+      it(`should have the correct values for ${tokenTest}`, function () {
+        const ofcCoin = bitgo.coin(tokenTest);
+        ofcCoin.getChain().should.equal(tokenTest);
+        ofcCoin.getFullName().should.equal('Arbitrum Test LINK');
+        ofcCoin.getBaseFactor().should.equal(PRECISION_18);
+        ofcCoin.addressCoin.should.equal('tarbeth');
+      });
+    });
+  });
+
   it('can sign payloads', async function () {
     const inputParams = {
       txPrebuild: {
--- a/modules/bitgo/test/v2/unit/coins/ofcToken.ts
+++ b/modules/bitgo/test/v2/unit/coins/ofcToken.ts
@@ -293,7 +293,6 @@ describe('OFC:', function () {
         ofcCoin.getChain().should.equal(tokenMain);
         ofcCoin.getFullName().should.equal('Chainlink Token');
         ofcCoin.getBaseFactor().should.equal(PRECISION_18);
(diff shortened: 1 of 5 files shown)


modules/statics/src/ofc.ts
```diff
@@ -554,3 +554,105 @@ export function tofcHederaToken(
...
+      prefix,
+      suffix,
+      features,
+      decimalPlaces,
+      isToken,
+      asset,
+      kind,
+      addressCoin,
+      primaryKeyCurve,
+      baseUnit: BaseUnit.ETH,
```
drew-bitgo, line 604: Shouldn't this be arbeth?

modules/statics/src/ofc.ts
```diff
@@ -554,3 +554,105 @@ export function tofcHederaToken(
...
+ * @param decimalPlaces Number of decimal places this coin supports (divisibility exponent)
+ * @param asset Asset which this coin represents. This is the same for both mainnet and testnet variants of a coin.
+ * @param kind Differentiates coins which represent fiat assets from those which represent crypto assets
+ * @param prefix? Optional coin prefix. Defaults to empty string
+ * @param suffix? Optional coin suffix. Defaults to coin name.
+ * @param isToken? Whether or not this account coin is a token of another coin
+ * @param features? Features of this coin. Defaults to the DEFAULT_FEATURES defined in `OfcCoin`
+ * @param primaryKeyCurve The elliptic curve for this chain/token
+ */
+export function ofcArbethErc20(
```
drew-bitgo, line 574:
```suggestion
export function ofcArbethToken(
```

modules/statics/src/ofc.ts
```diff
@@ -554,3 +554,105 @@ export function tofcHederaToken(
...
+ * @param decimalPlaces Number of decimal places this coin supports (divisibility exponent)
+ * @param asset Asset which this coin represents. This is the same for both mainnet and testnet variants of a coin.
+ * @param kind Differentiates coins which represent fiat assets from those which represent crypto assets
+ * @param prefix? Optional coin prefix. Defaults to empty string
+ * @param suffix? Optional coin suffix. Defaults to coin name.
+ * @param isToken? Whether or not this account coin is a token of another coin
+ * @param features? Features of this coin. Defaults to the DEFAULT_FEATURES defined in `OfcCoin`
+ * @param primaryKeyCurve The elliptic curve for this chain/token
+ */
+export function tofcArbethErc20(
```
drew-bitgo, line 625:
```suggestion
export function tofcArbethToken(
```

drew-bitgo (changes requested):

modules/statics/src/ofc.ts
subavicky: The base unit for arbeth is also 'wei'. There is no specific base unit for arbeth. Even the onchain token used the same ETH as base unit.

modules/statics/src/ofc.ts
subavicky: Actually the onchain token is named as arbethErc20, so named it like this to maintain consistency. It might be confusing to name it as token for ofc alone.

modules/statics/src/ofc.ts
drew-bitgo: disregard. I see now that those tokens are called `arbetherc20`. Seems weird

modules/statics/src/ofc.ts
drew-bitgo: disregard. I see now that those tokens are called arbetherc20. Seems weird

modules/bitgo/test/v2/unit/coins/ofcToken.ts
```diff
@@ -284,6 +284,27 @@ describe('OFC:', function () {
     });
   });
 
+  describe('check ofc tokens for arbethErc20', function () {
```
drew-bitgo, line 287: Add a `should.equal` for `addressCoin`, since that is a field that we (Prime) uses. Dont want it changing from underneath us without anyone noticing

subavicky: @drew-bitgo added the check. But have to add addressCoin in the ofcTokenConfig as it wasn't there. Please review.

drew-bitgo (approved):

ajays97 (approved):

PR #4442, opened 2024-04-16
Revert "feat: protect pass by value when sending data out"

lcovar:
This reverts commit 8755bdd5f4174de21f0b7e4c1e7bb74fb9b8e40f.

TICKET: BTC-1084

This breaks WRW, run into `strucutedClone is not defined` error.

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -683,12 +683,12 @@ export class BitGoAPI implements BitGoBase {
    * Caution: contains sensitive data
    */
   toJSON(): BitGoJson {
-    return structuredClone({
+    return {
       user: this._user,
       token: this._token,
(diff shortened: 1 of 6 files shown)


pranavjain97 (approved):

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -683,12 +683,12 @@ export class BitGoAPI implements BitGoBase {
    * Caution: contains sensitive data
    */
   toJSON(): BitGoJson {
-    return structuredClone({
```
margueriteblair, line 686: lol, the culprit

margueriteblair (approved):

modules/sdk-api/src/bitgoAPI.ts
lcovar: 🤷‍♂️

PR #4453, opened 2024-04-18
feat(eos): add ability to do explainTransaction with txHex for eos

margueriteblair:
Ticket: WP-1507

 Becomes a pain to have to pass in `transaction.packed_trx` for all `explainTransaction` implementations with EOS. This PR allows you to pass in a `txHex` parameter, and then under the hood it'll convert the txHex to the packed_trx. Specifically remedies the fact that we cannot use `explainTransaction` for EOS in the AOK signing flow.

### DIFF
--- a/modules/sdk-coin-eos/src/eos.ts
+++ b/modules/sdk-coin-eos/src/eos.ts
@@ -1,6 +1,7 @@
 /**
  * @prettier
  */
+import assert from 'assert';
 import { BigNumber } from 'bignumber.js';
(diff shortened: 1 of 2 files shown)


mullapudipruthvik (approved):
LGTM however is this a breaking change?

margueriteblair:
> LGTM however is this a breaking change?

I don't think so seeing as the required params only became optional? But let me get a second opinion tomorrow.

PR #4454, opened 2024-04-19
feat(abstract-utxo): support trustless change outputs from explaintx

saravanan7mani:
explainTransaction uses change addresses from platform for its change outputs
Now, they are trustlessly deduced from PSBT

TICKET: BTC-1112

### DIFF
--- a/modules/abstract-utxo/src/transaction.ts
+++ b/modules/abstract-utxo/src/transaction.ts
@@ -206,12 +206,26 @@ export function explainPsbt<TNumber extends number | bigint>(
   } catch (e) {
     throw new Error('failed to parse psbt hex');
   }
+  const txOutputs = psbt.txOutputs;
+  function getChangeAddresses() {
+    try {
+      return utxolib.bitgo
+        .findInternalOutputIndices(psbt)
+        .map((i) => utxolib.address.fromOutputScript(txOutputs[i].script, network));
+    } catch (e) {
+      if (e instanceof utxolib.bitgo.ErrorNoMultiSigInputFound) {
+        return [];
+      }
+      throw e;
+    }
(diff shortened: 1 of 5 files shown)


modules/abstract-utxo/src/transaction.ts
```diff
@@ -206,12 +206,26 @@ export function explainPsbt<TNumber extends number | bigint>(
...
     throw new Error('failed to parse psbt hex');
   }
+  const txOutputs = psbt.txOutputs;
+  function getChangeAddresses() {
+    try {
+      return utxolib.bitgo
+        .findInternalOutputIndices(psbt)
+        .map((i) => utxolib.address.fromOutputScript(txOutputs[i].script, network));
+    } catch (e) {
+      if (e instanceof Error && e.message === 'No multi sig input found') {
```
OttoAllmendinger, line 216: instead of relying on error message we should have a dedicated error class

```
class ErrorNoMultiSigInputFound extends Error { ... }

...
if (e instanceof ErrorNoMultiSigInputFound) { ... }
```

but why not propagate the error instead?

OttoAllmendinger (changes requested):

modules/abstract-utxo/src/transaction.ts
saravanan7mani: > why not propagate the error instead?
As per existing test cases, `explainTransaction` should return empty change outputs for a PSBT with no multi-sig inputs (ex: PSBT only with `p2shP2pk`). It makes sense because it's not in `explainTransaction`'s scope to assert for presence of mult-sig inputs in PSBTs.

OttoAllmendinger (approved):

PR #4462, opened 2024-04-22
feat(sdk-lib-mpc): support mpcv1 to mpcv2 retrofit

islamaminBitGo:
Ticket: HSM-342


modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
```diff
@@ -1,22 +1,70 @@
 import { KeygenSession, Keyshare, Message } from '@silencelaboratories/dkls-wasm-ll-node';
-import { DeserializedBroadcastMessage, DeserializedMessages, DkgState } from './types';
+import { DeserializedBroadcastMessage, DeserializedMessages, DkgState, RetrofitData } from './types';
 import { decode } from 'cbor';
+import { encode } from 'cbor-x';
```
dpkjnr, line 4: do we need both cbor and cbor-x?

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
```diff
@@ -1,22 +1,70 @@
...
+    if (retrofitData) {
+      if (!retrofitData.xShare.y || !retrofitData.xShare.chaincode || !retrofitData.xShare.x) {
+        throw Error('xShare must have a public key, private share value, and a chaincode.');
+      }
+      if (n !== 3 || t !== 2) {
+        throw Error('only 2 of 3 retrofit is supported.');
+      }
+      if (retrofitData.bigSiList.length !== 2) {
+        throw Error("bigSiList should contain the other parties' Si's");
+      }
```
dpkjnr, line 34: these conditions can be extracted to a function

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
```diff
@@ -1,22 +1,70 @@
...
+      for (let i = 0; i < n; i++) {
+        if (i === partyIdx) {
+          const secp256k1 = new Secp256k1Curve();
+          big_s_list.push(Array.from(bigIntToBufferBE(secp256k1.basePointMult(BigInt('0x' + retrofitData.xShare.x)))));
+        } else {
+          big_s_list.push(Array.from(Buffer.from(retrofitData.bigSiList[j], 'hex')));
+          j++;
+        }
+      }
+      const x_i_list = [new Array(32).fill(0), new Array(32).fill(0), new Array(32).fill(0)];
```
dpkjnr, line 46: I think we should move this whole retrofit related code out of the constructor and add jsdoc to explain the process.

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
islamaminBitGo: good point. we actually need cbor-x since the encode from cbor does not follow expected behaviour and the encoded cbor is not what the dkls code is expecting. will see if i can change uses of cbor to cbor-x so i can remove cbor.

zahin-mohammad (commented):
flushing from review Q

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
islamaminBitGo: done

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
islamaminBitGo: not sure the validation needs to be in it's own function since it's very minimal and also it is needed there to assert that xShare.y, xShare.chaincode, xShare.x are defined so that the compiler does not complain when we later use them as defined only variables later in the same code block.

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
```diff
@@ -1,22 +1,70 @@
...
+      if (!retrofitData.xShare.y || !retrofitData.xShare.chaincode || !retrofitData.xShare.x) {
+        throw Error('xShare must have a public key, private share value, and a chaincode.');
+      }
+      if (n !== 3 || t !== 2) {
+        throw Error('only 2 of 3 retrofit is supported.');
+      }
+      if (retrofitData.bigSiList.length !== 2) {
+        throw Error("bigSiList should contain the other parties' Si's");
+      }
+      const big_s_list: Array<Array<number>> = [];
```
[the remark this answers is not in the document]
islamaminBitGo: done

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
islamaminBitGo: which part of the process you think needs to be in the jsdoc , this is a key refresh flow do you think we need a brief description of key refresh in the jsdoc?

islamaminBitGo: moved it to a new function and called in initDkg instead of the constructor which does feel more appropriate

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
```diff
@@ -25,6 +31,49 @@ export class Dkg {
     }
   }
 
+  private _createDKLsRetrofitKeyShare() {
+    if (this.retrofitData) {
```
zahin-mohammad, line 35:
```suggestion
  private _createDKLsRetrofitKeyShare() {
    if (!this.retrofitData) return
```

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
```diff
@@ -25,6 +31,49 @@ export class Dkg {
...
+          bigSList.push(
+            Array.from(bigIntToBufferBE(secp256k1.basePointMult(BigInt('0x' + this.retrofitData.xShare.x))))
+          );
+        } else {
+          bigSList.push(Array.from(Buffer.from(this.retrofitData.bigSiList[j], 'hex')));
+          j++;
+        }
+        xiList.push(Array.from(bigIntToBufferBE(BigInt(i + 1), 32)));
+      }
+      const dklsKeyShare = {
```
zahin-mohammad, line 57: Is there a type defined for this? That would ensure you are not missing any fields.

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
```diff
@@ -77,9 +131,10 @@ export class Dkg {
   }
 
   getKeyShare(): Buffer {
-    const keyShareBuff = Buffer.from(this.dkgKeyShare.toBytes());
-    this.dkgKeyShare.free();
-    return keyShareBuff;
+    if (!this.keyShareBuff) {
+      throw Error('Can not get key share, DKG is not complete yet.');
+    }
+    return this.keyShareBuff;
```
zahin-mohammad, line 137: Is removing `this.dkgKeyShare.free();` intentional?

zahin-mohammad (approved):

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
islamaminBitGo: moved it somewhere else so not a removal

PR #4477, opened 2024-04-25
fix: handle 502 after tx broadcast

kamleshmugdiya:
Handle 502 error after transaction broadcast

This commit adds error handling for the scenario where `sendUserToBitgoGShare` succeeds but `getTxRequest` fails with a 502 status code.
In such cases, it throws a specific error message indicating that the transaction was broadcasted successfully, but BitGo failed to update the transaction status.
This change ensures that more informative error message is provided when encountering bad gateway error during the transaction status retrieval.

Ticket: COIN-92

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
@@ -886,7 +886,17 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       SendShareType.SShare,
       userSShare
     );
-    return await getTxRequest(this.bitgo, this.wallet.id(), txRequestObj.txRequestId);
+    try {
+      return await getTxRequest(this.bitgo, this.wallet.id(), txRequestObj.txRequestId);
+    } catch (error) {
+      if (error.status === 502) {
+        throw new Error(
+          'Transaction broadcasted successfully, but BitGo failed to get the transaction status ' + error
+        );
+      } else {
+        throw error;
+      }
+    }
   }
 
   /**
@@ -999,7 +1009,17 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       SendShareType.SShare,
       userSShare
     );
-    return await getTxRequest(this.bitgo, this.wallet.id(), txRequest.txRequestId);
+    try {
+      return await getTxRequest(this.bitgo, this.wallet.id(), txRequest.txRequestId);
+    } catch (error) {
+      if (error.status === 502) {
+        throw new Error(
+          'Transaction broadcasted successfully, but BitGo failed to get the transaction status ' + error
+        );
+      } else {
+        throw error;
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -886,7 +886,17 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       SendShareType.SShare,
       userSShare
     );
-    return await getTxRequest(this.bitgo, this.wallet.id(), txRequestObj.txRequestId);
+    try {
+      return await getTxRequest(this.bitgo, this.wallet.id(), txRequestObj.txRequestId);
+    } catch (error) {
+      if (error.status === 502) {
+        throw new Error(
+          'Transaction broadcasted successfully, but BitGo failed to update the transaction status ' + error
```
Vijay-Jagannathan, line 894: We seem to be doing a get and not update

Vijay-Jagannathan (commented):
Can we add a test to replicate it using existing tests ?

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
gianchandania: +1

pengyuc-bitgo (dismissed):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
kamleshmugdiya: fixed it.

kamleshmugdiya: have added couple of tests.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -629,8 +639,17 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
     );
 
     await sendUserToBitgoGShare(this.bitgo, this.wallet.id(), txRequestId, userToBitGoGShare, apiVersion);
-
-    return await getTxRequest(this.bitgo, this.wallet.id(), txRequestId);
+    try {
+      return await getTxRequest(this.bitgo, this.wallet.id(), txRequestId);
+    } catch (error) {
+      if (error.status === 502) {
```
mullapudipruthvik, line 645: do we have to do the same for other status codes like 504 ?

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -886,7 +886,17 @@ export class EcdsaUtils extends baseTSSUtils<KeyShare> {
       SendShareType.SShare,
       userSShare
     );
-    return await getTxRequest(this.bitgo, this.wallet.id(), txRequestObj.txRequestId);
+    try {
+      return await getTxRequest(this.bitgo, this.wallet.id(), txRequestObj.txRequestId);
+    } catch (error) {
+      if (error.status === 502) {
```
bhavidhingra, line 892: can we use a variable here instead of hard-coded numeric?

zahin-mohammad, line 892: How can we be sure that all 502 errors here have a succesfull broadcast? The error handling here does not seem verbose enough and potentially misleading. What about other error codes? Lets discuss more over slack or in the wallets team focus hour.

zahin-mohammad (changes requested):

PR #4485, opened 2024-04-29
Revert "Revert "feat(sdk-core): create dkls wallets with version 3 as well""

pranavjain97:
Reverts BitGo/BitGoJS#4481


pranavjain97:
Doing this on the client side with user feature flags instead

PR #4497, opened 2024-05-01
fix(sdk-core): route PA flow to use MPCv2

pranavjain97:
TICKET: WP-1802


zahin-mohammad (commented):
Changes look good, but lets get some tests

zahin-mohammad (approved):
Tests will be added in a follow up.

PR #4505, opened 2024-05-03
fix(root): fix keychains updatePassword for tss

alebusse:
instead of using the commonKeychain for tss keys (which is not unique) we will use the key id

WP-1815

TICKET: WP-1815


zahin-mohammad (dismissed):

zahin-mohammad (commented):
lgtm. Reserving approval until Tuesday

zahin-mohammad (approved):

PR #4513, opened 2024-05-07
chore(sdk-coin-trx): support trx token consolidation in recovery

mukeshsp:
Confluence page : https://bitgoinc.atlassian.net/wiki/spaces/ENG/pages/3261235223/Tron+TRX+Recovery

Testing : build BitGoJs locally and bumped in WRW 
Performed token consolidation : https://shasta.tronscan.org/#/transaction/6b712285a3146be4332d1a64c74235366a02ecfe14380892b0f0cdedec74c28a

Performed coin consolidation :
https://shasta.tronscan.org/#/transaction/663c766ea20e252e342d03674c7fa4989368edfb667202f39939ed0e4fbf135b

token sweep from SMC wallet 
https://shasta.tronscan.org/#/transaction/461ed3325adfdd30e58b1b291f5519874e37ebe04396d1ede438cc006dbc0d1c

Ticket: COIN-563

### DIFF
--- a/modules/sdk-coin-trx/src/trx.ts
+++ b/modules/sdk-coin-trx/src/trx.ts
@@ -31,6 +31,7 @@ import { Interface, Utils, WrappedBuilder } from './lib';
 import { getBuilder } from './lib/builder';
 import { TransactionReceipt } from './lib/iface';
 import { isInteger, isUndefined } from 'lodash';
+import { tokenMainnetContractAddresses, tokenTestnetContractAddresses } from './lib/utils';
 
 export const MINIMUM_TRON_MSIG_TRANSACTION_FEE = 1e6;
 export const SAFE_TRON_TRANSACTION_FEE = 2.1 * 1e6; // TRON foundation recommends 2.1 TRX as fees for guaranteed transaction
@@ -795,11 +796,17 @@ export class Trx extends BaseCoin {
         let recoveryAmount = 0;
         // Tokens must be consolidate before the native asset. First construct token txns
         let rawTokenTxn: any | undefined;
-        for (const token of accountInfo.data[0].trc20) {
-          for (const tokenContractAddr of TOKEN_CONTRACT_ADDRESSES) {
-            if (token[tokenContractAddr]) {
-              const amount = token[tokenContractAddr];
-              const tokenContractAddrHex = Utils.getHexAddressFromBase58Address(tokenContractAddr);
+        // token address
+        if (accountInfo.data[0].balance > SAFE_TRON_TOKEN_TRANSACTION_FEE && accountInfo.data[0].trc20[0]) {
+          const tokenDataArray = accountInfo.data[0].trc20;
+          for (const tokenData of tokenDataArray) {
+            const contractAddress = Object.keys(tokenData) as Array<string>;
+            if (
+              tokenMainnetContractAddresses.includes(contractAddress[0]) ||
+              tokenTestnetContractAddresses.includes(contractAddress[0])
+            ) {
+              const amount = tokenData[contractAddress[0]];
+              const tokenContractAddrHex = Utils.getHexAddressFromBase58Address(contractAddress[0]);
               rawTokenTxn = (
                 await this.getTriggerSmartContractTransaction(
                   baseAddrHex,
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-trx/src/trx.ts
```diff
@@ -792,27 +792,70 @@ export class Trx extends BaseCoin {
...
-
-        if (!isUnsignedConsolidations) {
-          const userPrv = this.xprvToCompressedPrv(userKey.toBase58());
-          // receive address only needs to be signed by user key
-          txBuilder.sign({ key: userPrv });
+        let recoveryAmount = 0;
+        // Tokens must be consolidate before the native asset. First construct token txns
+        let rawTokenTxn: any | undefined;
+        for (const token of accountInfo.data[0].trc20) {
+          for (const tokenContractAddr of TOKEN_CONTRACT_ADDRESSES) {
```
gianchandania, line 799: Let's not just do this implementation for this list of tokens, ideally we should do this for all the tokens that we support and that information should be fetched from statics

modules/sdk-coin-trx/src/trx.ts
```diff
@@ -792,27 +792,70 @@ export class Trx extends BaseCoin {
...
+            if (token[tokenContractAddr]) {
+              const amount = token[tokenContractAddr];
+              const tokenContractAddrHex = Utils.getHexAddressFromBase58Address(tokenContractAddr);
+              rawTokenTxn = (
+                await this.getTriggerSmartContractTransaction(
```
gianchandania, line 804: I think we can use Promise.all here

modules/sdk-coin-trx/src/trx.ts
gianchandania, line 799: Also I think we can avoid this nested loop by forming a Set from the tokens present in the statics and checking the Set to find if token is supported or not

modules/sdk-coin-trx/src/trx.ts
```diff
@@ -792,27 +792,70 @@ export class Trx extends BaseCoin {
...
+          // extend the expiry to 1 day
+          txBuilder.extendValidTo(RECOVER_TRANSACTION_EXPIRY);
+          // this tx should be enough to drop into a node
+          if (!isUnsignedConsolidations) {
+            const userPrv = this.xprvToCompressedPrv(userKey.toBase58());
+            // receive address only needs to be signed by user key
+            txBuilder.sign({ key: userPrv });
+          }
+          const tx = await txBuilder.build();
+          txnsBatch.push(this.formatForOfflineVault(tx, SAFE_TRON_TRANSACTION_FEE, recoveryAmount, addressInfo));
```
gianchandania, line 835: As this is token recovery, so I guess will have to pass SAFE_TRON_TOKEN_TRANSACTION_FEE?

modules/sdk-coin-trx/src/trx.ts
```diff
@@ -792,27 +792,70 @@ export class Trx extends BaseCoin {
...
-        const txBuilder = (getBuilder(this.getChain()) as WrappedBuilder).from(buildTx);
-        // Default expiry is 1 minute which is too short for recovery purposes
-        // extend the expiry to 1 day
-        txBuilder.extendValidTo(RECOVER_TRANSACTION_EXPIRY);
-
-        if (!isUnsignedConsolidations) {
-          const userPrv = this.xprvToCompressedPrv(userKey.toBase58());
-          // receive address only needs to be signed by user key
-          txBuilder.sign({ key: userPrv });
+        let recoveryAmount = 0;
```
gianchandania, line 795: Now as we are also recovering Tron Tokens here, we will have to use SAFE_TRON_TOKEN_TRANSACTION_FEE in the condition if tokens are to be recovered

modules/sdk-coin-trx/src/trx.ts
mukeshsp: We just want add support of USDC, if we will going to do for all the supported coin we might encounter with a lot errors.

mukeshsp: I reused the array which already used in other methods

mukeshsp:
verified token consolidation after new changes https://shasta.tronscan.org/#/transaction/c444f88433864507e63aaa7ee4c745564afb326afaed4f35548f9bf5d6fbdea4

modules/sdk-coin-trx/src/trx.ts
mukeshsp: updated !!

mukeshsp: updated

modules/sdk-coin-trx/src/trx.ts
mukeshsp: updated

gianchandania (approved):

PR #4514, opened 2024-05-07

evalenti-simtlix:
add new batch of tokens

EA-3652

TICKET: EA-3652


ajays97 (dismissed):

modules/statics/src/coins.ts
```diff
@@ -9199,6 +9215,14 @@ export const coins = CoinMap.fromCoins([
     '0x8bc8eddd7eddd80ec96320a3500cb84310c3c1f5',
     UnderlyingAsset.OCTAV
   ),
+  erc20(
+    '3f6a9178-8e58-4885-8575-dcf8d4a56e80',
+    'zro',
+    'Layerzero',
+    18,
+    '0x1a44076050125825900e736c501f859c50fe728c',
```
alebusse, line 9223:
```suggestion
    '0x1a44076050125825900e736c501f859c50fe728c',
```
Check if this is the right token address
https://etherscan.io/address/0x32002658029da06fffcfbad58706aa325b3e236e

alebusse (changes requested):
review zro token address

modules/statics/src/coins.ts
evalenti-simtlix: We will skip this token while wait confirmation, because is different than they show in docs.

Also, we added 2 new tokens (LIF3 and L3USD)

alebusse (approved):

PR #4518, opened 2024-05-08
fix(statics): fix display name issue for arbeth tokens

subavicky:
TICKET: PX-3568


ravneet-bitgo (approved):

sachushaji (changes requested):
Just double checking this one, This seems to be breaking change, Kindly make sure to double check at any parts of the system there is any side effects which would cause because of removing ofc like this. Kindly add a breaking changes footer in the commit message, Also for some of the other ofcs, its not done like this, whats the need for arbeth?

subavicky:
> Just double checking this one, This seems to be breaking change, Kindly make sure to double check at any parts of the system there is any side effects which would cause because of removing ofc like this. Kindly add a breaking changes footer in the commit message, Also for some of the other ofcs, its not done like this, whats the need for arbeth?

Actually other ofc tokens are not supported by prime yet on prod. Only eth is supported as of now and the logic is same for [eth](https://github.com/BitGo/BitGoJS/blob/3026acd22dcb1ee3af6c8cb56a73911269743551/modules/statics/src/ofc.ts#L179)  as well. First went ahead without this change, but the display name is coming along with ofc which will be fixed by this change.

zahin-mohammad (approved):

sachushaji (approved):

PR #4524, opened 2024-05-09

evalenti-simtlix:
add new batch of tokens

EA-3687

TICKET: EA-3687


zahin-mohammad (approved):
verified tokens on etherscan, polygonscan, and hederaexplorer

PR #4528, opened 2024-05-13
feat(abstract-eth): refactor eth wrw recovery tss

dpkjnr:
Ticket: WP-1885

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -1,17 +1,10 @@
 /**
  * @prettier
  */
-import debugLib from 'debug';
-import { bip32 } from '@bitgo/utxo-lib';
-import { BigNumber } from 'bignumber.js';
-import { randomBytes } from 'crypto';
-import Keccak from 'keccak';
-import _ from 'lodash';
-import secp256k1 from 'secp256k1';
-import BN from 'bn.js';
 import {
   AddressCoinSpecific,
   BitGoBase,
+  BuildNftTransferDataOptions,
   common,
   ECDSA,
   Ecdsa,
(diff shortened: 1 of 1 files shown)


mullapudipruthvik (commented):
do we have a test for this?, so what we know the refactor didn't break anything

dpkjnr:
> do we have a test for this?, so what we know the refactor didn't break anything

gg18 recovery tests are already there and those are passing. I will add the DKLS specific tests with DKLS recovery signing

mullapudipruthvik (approved):

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -1902,56 +1902,79 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+        signingMaterial.bitgoNShare &&
+        (signingMaterial.userNShare || signingMaterial.backupNShare)
+      );
+    } catch (error) {
+      return false;
+    }
+  }
 
-    // build recipients object
+  private async buildTssRecoveryTxn(baseAddress: string, gasPrice: any, gasLimit: any, params: RecoverOptions) {
```
sachushaji, line 1975: NIT: we could type gasPrice and gaslimit

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -1966,44 +1989,35 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
-        params.replayProtectionOptions
-      );
-    }
+  async validateBalanceAndGetTxAmount(baseAddress: string, gasPrice: BN, gasLimit: BN) {
+    const baseAddressBalance = await this.queryAddressBalance(baseAddress);
 
-    const signableHex = tx.getMessageToSign(false).toString('hex');
-    if (!userKeyCombined || !backupKeyCombined) {
-      throw new Error('Missing key combined shares for user or backup');
+    const totalGasNeeded = gasPrice.mul(gasLimit);
```
sachushaji, line 2009:
```suggestion
    const totalGasCostNeeded = gasPrice.mul(gasLimit);
```
NIT: Might be the correct term

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -1966,44 +1989,35 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+    const weiToGwei = new BN(10 ** 9);
+    if (baseAddressBalance.lt(totalGasNeeded)) {
+      throw new Error(
+        `Backup key address ${baseAddress} has balance ${baseAddressBalance.div(weiToGwei).toString()} Gwei.` +
```
sachushaji, line 2013: There is no concept of a backup key address in GG18, all the public keys are corresponding to the base address. So we could change this error message to say base address and not backup address which could confuse people

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
dpkjnr: will make this change along with the DKLS signing PR

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
dpkjnr: will make this change along with the DKLS signing PR

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
dpkjnr: will make this change along with the DKLS signing PR

PR #4534, opened 2024-05-15
feat: send changeAddressType preferences array 

lcovar:

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -1466,10 +1466,10 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
 
   async getExtraPrebuildParams(buildParams: ExtraPrebuildParamsOptions & { wallet: Wallet }): Promise<{
     txFormat?: 'legacy' | 'psbt';
-    addressType?: ScriptType2Of3;
+    changeAddressType?: ScriptType2Of3[] | ScriptType2Of3;
   }> {
     let txFormat = buildParams.txFormat as 'legacy' | 'psbt' | undefined;
-    let addressType = buildParams.addressType as ScriptType2Of3 | undefined;
+    let changeAddressType = buildParams.changeAddressType as ScriptType2Of3[] | ScriptType2Of3 | undefined;
 
     const walletFlagMusigKp = buildParams.wallet.flag('musigKp') === 'true';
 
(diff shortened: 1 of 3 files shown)


margueriteblair (dismissed):

OttoAllmendinger (dismissed):

OttoAllmendinger (approved):

OttoAllmendinger (changes requested):
see other review

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -1466,10 +1466,10 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
 
   async getExtraPrebuildParams(buildParams: ExtraPrebuildParamsOptions & { wallet: Wallet }): Promise<{
     txFormat?: 'legacy' | 'psbt';
-    addressType?: ScriptType2Of3;
+    changeAddressType?: ScriptType2Of3[] | ScriptType2Of3;
   }> {
     let txFormat = buildParams.txFormat as 'legacy' | 'psbt' | undefined;
-    let addressType = buildParams.addressType as ScriptType2Of3 | undefined;
+    let changeAddressType = buildParams.changeAddressType as ScriptType2Of3[] | ScriptType2Of3 | undefined;
```
OttoAllmendinger, line 1472: I believe this breaks backwards compatability - callers will still be passing addressType here

OttoAllmendinger (changes requested):

modules/abstract-utxo/src/abstractUtxoCoin.ts
lcovar: hmm not sure if i see the problem. If they do we won't be overriding it? On line 1489 we check if addressType is defined, if it is we don't default changeAddressType to the array and it will stay undefined. 

It as intended to be changed to changeAddressType [here](https://github.com/BitGo/BitGoJS/commit/94942559b0be8f9789c7afbb3740924aaab76ab3) but i believe the backend wasn't ready to handle so it was then [reverted](https://github.com/BitGo/BitGoJS/commit/5e45e62d512c2794a517030027b1f9f66b20e048). I believe we're good now.

OttoAllmendinger: A caller that passes `addressType: 'foo'` previously received `addressType: 'foo'` as a result 

with your change, `addressType: 'foo'` now results in `changeAddressType: undefined`

You are doing two things at once in this PR (deprecating a parameter and changing behavior). Better split it up into two PRs.

OttoAllmendinger (changes requested):

modules/abstract-utxo/src/abstractUtxoCoin.ts
OttoAllmendinger: ok apparently the caller will retain `addressType` and it will be sent to WP anyway

OttoAllmendinger (approved):

PR #4535, opened 2024-05-15
fix(statics): update stellar token label

shivpippal:
TICKET: WIN-1601


ravneet-bitgo (approved):

ajays97 (approved):
Interesting why we named this as plain Ethereum. But nice catch!!

PR #4543, opened 2024-05-19
fix: should be able to generate onchain custodial wallet with express

alexhu-bitgo:
Ticket: WP-1508

### DIFF
--- a/modules/bitgo/test/v2/unit/wallets.ts
+++ b/modules/bitgo/test/v2/unit/wallets.ts
@@ -398,6 +398,38 @@ describe('V2 Wallets:', function () {
         scope.done();
       }
     });
+
+    it('should generate custodial onchain wallet without passing m, n, keys, keySignatures', async () => {
+      const params: GenerateWalletOptions = {
+        label: 'test wallet',
+        enterprise: 'myenterprise',
+        type: 'custodial',
+        passphrase: 'secret',
+      };
+
+      const walletNock = nock(bgUrl)
+        .post('/api/v2/tbtc/wallet', function (body) {
+          body.type.should.equal('custodial');
+          should.not.exist(body.m);
+          should.not.exist(body.n);
+          should.not.exist(body.keys);
+          should.not.exist(body.keySignatures);
+          return true;
+        })
+        .reply(200);
+
+      nock(bgUrl)
+        .post('/api/v2/tbtc/key', _.matches({ source: 'bitgo' }))
+        .reply(200, { pub: 'bitgoPub' });
+      nock(bgUrl).post('/api/v2/tbtc/key', _.matches({})).reply(200);
+      nock(bgUrl)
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -463,6 +463,14 @@ export class Wallets implements IWallets {
       walletParams.rootPrivateKey = params.rootPrivateKey;
     }
 
+    // Custodial onchain wallets do not need m, n, keys, or keySignatures
+    if (params.type === 'custodial' && params.multisigType === 'onchain') {
```
zahin-mohammad, line 467: Onchain is typically the default to preserve backwards compatibility. 
```suggestion
    if (params.type === 'custodial' && (params.multisigType ?? 'onchain') === 'onchain') {
```

zahin-mohammad (dismissed):
LGTM, did you test this against staging/test?

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -463,6 +463,14 @@ export class Wallets implements IWallets {
...
+      walletParams.n = undefined;
+      walletParams.m = undefined;
+      walletParams.keys = [];
```
alebusse, line 470:
```suggestion
      walletParams.keys = undefined;
```
maybe

modules/sdk-core/src/bitgo/wallet/wallets.ts
alebusse: i think you dont need the onchain check, if you got this far, is an onchain multisig wallet

modules/sdk-core/src/bitgo/wallet/wallets.ts
alexhu-bitgo: I decided to use `keys = []` because it meant I wouldn't need to make the `keys` field optional in `iBaseCoin`. But I think keeping things consistent here is ideal, so I wouldn't mind making this change.

modules/bitgo/test/v2/unit/wallets.ts
```diff
@@ -398,6 +398,40 @@ describe('V2 Wallets:', function () {
...
+
+    it('should generate custodial onchain wallet without passing m, n, keys, keySignatures', async () => {
+      const params = {
+        label: 'test wallet',
+        enterprise: 'myenterprise',
+        multisigType: 'onchain',
+        type: 'custodial',
+        passphrase: 'secret',
+        userKey: 'xpub123',
+      };
```
alebusse, line 410:
```suggestion
      const params = {
        label: 'test wallet',
        enterprise: 'myenterprise',
        type: 'custodial',
      };
```

alebusse (approved):

PR #4551, opened 2024-05-21
feat: encrypt and return backup key by default

ericli-bitgo:
This PR does 2 things:
 1. it by default returns the private information that is created during `generateWallet` - this is the breaking change
 2. it by default saves the `encryptedPrv` of the backup key in the key document. The old behavior is that this never happened. Note that this only affects hot wallets

TSS keys are already encrypted (and backed up) by default, so this only affects non-TSS keys.

TICKET: BTC-1174

### DIFF
--- a/modules/bitgo/test/v2/unit/keychains.ts
+++ b/modules/bitgo/test/v2/unit/keychains.ts
@@ -600,9 +600,17 @@ describe('V2 Keychains', function () {
     });
   });
 
-  it('should generate backup without encryptedPrv when passphrase not provided', () => {
-    const backup = keychains.createBackup({});
+  it('should generate backup without encryptedPrv when passphrase not provided', async () => {
+    const scope = nock(bgUrl)
+      .post('/api/v2/tltc/key', (body) => {
+        return body.encryptedPrv === undefined && body.prv === undefined;
+      })
+      .reply(200);
+
+    const backup = await keychains.createBackup({});
+    scope.isDone().should.be.true();
     backup.should.not.have.property('encryptedPrv');
+    backup.should.have.property('prv');
   });
 
   it('should generate backup and call endpoint with encryptedPrv when passphrase is provided', async () => {
--- a/modules/bitgo/test/v2/unit/keychains.ts
+++ b/modules/bitgo/test/v2/unit/keychains.ts
@@ -599,4 +599,23 @@ describe('V2 Keychains', function () {
       assert(decodedRes);
     });
   });
+
+  it('should generate backup without encryptedPrv when passphrase not provided', () => {
+    const backup = keychains.createBackup({});
+    backup.should.not.have.property('encryptedPrv');
+  });
+
+  it('should generate backup and call endpoint with encryptedPrv when passphrase is provided', async () => {
+    const scope = nock(bgUrl)
+      .post('/api/v2/tltc/key', (body) => {
+        return body.encryptedPrv !== undefined && body.prv === undefined;
+      })
+      .reply(200);
+
+    const backup = await keychains.createBackup({ passphrase: 't3stSicretly!' });
+    scope.isDone().should.be.true();
+    backup.should.have.property('encryptedPrv');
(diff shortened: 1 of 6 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -616,10 +616,10 @@ export async function handleV2GenerateWallet(req: express.Request) {
   const bitgo = req.bitgo;
   const coin = bitgo.coin(req.params.coin);
   const result = await coin.wallets().generateWallet(req.body);
-  if (req.query.includeKeychains) {
-    return { ...result, wallet: result.wallet.toJSON() };
+  if (req.query.doNotIncludeKeychains) {
```
mmcshinsky-bitgo, line 619: Instead of changed the implementation customers may already be using, set `includeKeychains` to be checked or set to `true`.

mmcshinsky-bitgo (commented):
@davidkaplanbitgo has identified that these actions will be breaking changes per the ticket information. Commit needs to reflect that if it is the case.

modules/sdk-core/src/bitgo/keychain/keychains.ts
```diff
@@ -264,6 +264,9 @@ export class Keychains implements IKeychains {
       // if the provider is undefined, we generate a local key and add the source details
       const key = this.create();
       _.extend(params, key);
+      if (params.passphrase !== undefined) {
+        _.extend(params, { encryptedPrv: this.bitgo.encrypt({ input: key.prv, password: params.passphrase }) });
+      }
```
alebusse, line 269: checking line 273, do we want to return the prv and the encryptedPrv ?

zahin-mohammad: No, the params here get sent to WP via line 272. 

line 273 annotates the response from WP with the prv, so the PRV doesn't actually get sent to WP.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -435,7 +436,7 @@ export class Wallets implements IWallets {
           throw new Error('cannot generate backup keypair without passphrase');
         }
         // No provided backup xpub or address, so default to creating one here
-        return this.baseCoin.keychains().createBackup({ reqId });
+        return this.baseCoin.keychains().createBackup({ reqId, passphrase: params.passphrase });
```
zahin-mohammad, line 439: Is this the only call-site we need to update?

zahin-mohammad (commented):
lgtm so far, but lets add some tests as well

modules/sdk-core/src/bitgo/wallet/wallets.ts
davidkaplanbitgo: yes

davidkaplanbitgo (dismissed):

margueriteblair (commented):
Will we need to update any dev docs about these changes? If so, can we make a follow up ticket?

davidkaplanbitgo:
> Will we need to update any dev docs about these changes? If so, can we make a follow up ticket?

@margueriteblair - AFAIK there is no documentation about the default behavior of key backups for hot wallets. If you think it is important, we can make a follow up ticket for it

davidkaplanbitgo (dismissed):

margueriteblair (dismissed):

modules/bitgo/test/v2/unit/keychains.ts
```diff
@@ -599,4 +599,23 @@ describe('V2 Keychains', function () {
       assert(decodedRes);
     });
   });
+
+  it('should generate backup without encryptedPrv when passphrase not provided', () => {
+    const backup = keychains.createBackup({});
+    backup.should.not.have.property('encryptedPrv');
```
zahin-mohammad, line 605: should `prv` exist? if so, lets add that to the test.

zahin-mohammad (dismissed):

alebusse (approved):

davidkaplanbitgo (approved):

PR #4553, opened 2024-05-21
feat(sdk-core): remove hardcoded coinSpecific forwarderVersions

pranavjain97:
TICKET: WP-1930

MPCv2 wallets uses `forwarderVersion = 4`, and this line of code is setting the wrong version. 

There is no need to set for coinSpecific hardcoded defaults for forwarderVersions.
WP has logic to handle defaults based on the wallet.

**Testing**
- Tested mpcv2 forwarder addresses are still created if no body is passed in. 
- Tested mpcv2 receive addresses are still created if no body is passed in.

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -1154,8 +1154,6 @@ export class Wallet implements IWallet {
         throw new Error('forwarderVersion has to be an integer 0, 1, 2, 3 or 4');
       }
       addressParams.forwarderVersion = forwarderVersion;
-    } else if (this._wallet.multisigType === 'tss' && this.baseCoin.getMPCAlgorithm() === 'ecdsa') {
-      addressParams.forwarderVersion = 3;
     }
 
     if (!_.isUndefined(label)) {


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1155,7 +1155,11 @@ export class Wallet implements IWallet {
       }
       addressParams.forwarderVersion = forwarderVersion;
     } else if (this._wallet.multisigType === 'tss' && this.baseCoin.getMPCAlgorithm() === 'ecdsa') {
-      addressParams.forwarderVersion = 3;
+      if (this._wallet.multisigTypeVersion === 'MPCv2') {
+        addressParams.forwarderVersion = 4;
```
zahin-mohammad, line 1159: Where is the `forwarderVersion` documented? Just want to make sure we are using the correct value here.

pranavjain97: Not sure where it's documented, but we've already tested it working with `forwarderVersion` = 4

zahin-mohammad: @BitGo/ethalt-team  can u confirm the correct forwarder version number for MPCv2 wallets?

dpkjnr: forwarder version 4 is correct

sachushaji: This change is really eth specific, Only eth like coins have forwarder versions. So when it comes to cosmos etc which doesn't use forwarders at all this could cause issues. I think by default, if you don't pass in forwarder version, it should by default create the address with version 4 and it should be the case for all coins where it will create the address according to its defaults if we didn't explicitly specify. If we are able to create other versions of forwarders, Kindly let me know, I can raise a ticket for it

pranavjain97: Removed the coin specific logic completely as discussed offline and will let WP handle the defaults

alebusse (approved):

PR #4556, opened 2024-05-22
feat: remove p2tr from changeAddressType list

lcovar:
Ticket: BTC-1176


margueriteblair (approved):

margueriteblair (commented):
Is there a follow up to remove references of `p2tr` from WP too?

davidkaplanbitgo (approved):

lcovar:
> Is there a follow up to remove references of `p2tr` from WP too?

No, that will stay. We just wanted to remove it from this list preferences array. If a client wanted to use p2tr they can pass changeAddressType: 'p2tr' and that still work.

PR #4564, opened 2024-05-23
feat: use psbt format for hot and custodial wallets 

lcovar:
Ticket: BTC-776

TICKET: BTC-776

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -2,7 +2,7 @@
  * @prettier
  */
 import * as utxolib from '@bitgo/utxo-lib';
-import { bip32, BIP32Interface, bitgo, isTestnet } from '@bitgo/utxo-lib';
+import { bip32, BIP32Interface, bitgo } from '@bitgo/utxo-lib';
 import * as assert from 'assert';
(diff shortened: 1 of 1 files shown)


modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -1471,15 +1471,13 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+    // if the txFormat is not specified, default to psbt for distributed custody, hot and custodial wallets
+    // FIXME(BTC-776): eventually default to psbt for cold wallets as well
     if (
       buildParams.txFormat === undefined &&
       (buildParams.wallet.subType() === 'distributedCustody' ||
-        (isTestnet(this.network) && buildParams.wallet.type() === 'hot') ||
-        // FIXME(BTC-776): default to psbt for all mainnet wallets in the future
-        walletFlagMusigKp)
+        buildParams.wallet.type() === 'hot' ||
+        buildParams.wallet.type() === 'custodial')
```
davidkaplanbitgo, line 1480:
```suggestion
        buildParams.wallet.type() === 'hot')
```

davidkaplanbitgo (changes requested):
Still rolling out custodial - I feel like we should only set it here once we have fully rolled out

davidkaplanbitgo (approved):

PR #4566, opened 2024-05-24
fix(abstract-eth): fix wallet version validation

dpkjnr:
Ticket: WP-1940

Transaction building is failing in wallet-platform due to this validation. Error details: https://grafana.bitgo-test.com/goto/6Cs4XsEIg?orgId=1

### DIFF
--- a/modules/abstract-eth/src/lib/transactionBuilder.ts
+++ b/modules/abstract-eth/src/lib/transactionBuilder.ts
@@ -824,7 +824,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
    * @param {number} version wallet version
    */
   walletVersion(version: number): void {
-    if (version < 0 || version > 4 || version === 3) {
+    if (version < 0 || version > 5 || version === 3) {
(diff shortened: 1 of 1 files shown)


modules/abstract-eth/src/lib/transactionBuilder.ts
```diff
@@ -824,7 +824,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
    * @param {number} version wallet version
    */
   walletVersion(version: number): void {
-    if (version < 0 || version > 4 || version === 3) {
+    if (version < 0 || version > 5 || version === 3) {
```
gianchandania, line 827: It is throwing an error in case of Mpc wallet, i think this should be required only in the case of a wallet contract. Why there is a need of calling this with Mpcv2?

dpkjnr: forwarder deployment is failing, not quite sure why we are passing walletVersion while building deploy forwarder txn

PR #4570, opened 2024-05-28
feat(ada): support passing `unspents` as intent param

hitansh-madan:
Ticket: WIN-1832

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/mpcUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/mpcUtils.ts
@@ -134,7 +134,7 @@ export abstract class MpcUtils {
       return formattedRecipient;
     });
 
-    const baseIntent = {
+    const baseIntent: PopulatedIntent = {
       intentType: params.intentType,
       sequenceId: params.sequenceId,
       comment: params.comment,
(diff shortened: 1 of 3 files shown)


margueriteblair (changes requested):
Please add context for why we're making these changes? Trying to understand why the unspents would need to be passed into the intent; I feel like that doesn't make sense. Also you need tests.

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
```diff
@@ -169,6 +169,10 @@ export abstract class MpcUtils {
       }
     }
 
+    if (baseCoin.getFamily() === 'ada' && params.unspents) {
+      baseIntent.unspents = params.unspents;
+    }
```
zahin-mohammad, line 174: Agree with @margueriteblair , why is the user not supplying intents in the intent directly?

zahin-mohammad (changes requested):

hitansh-bitgo:
@zahin-mohammad @margueriteblair I think I might still need this change after all. To add some context, we recently added a change to let a user specify in the intent which unspents to use for txn building for Ada https://github.com/BitGo/common-interface/pull/54. Now this works fine when using `v2.wallet.txrequest.create`, but we also need to support passing unspents in sendMany calls, and for that I'll need this change,right? Or is there a better way to go about this?

PR #4571, opened 2024-05-28
feat(ton): support non-bounceable address format

ajays97:
Ticket: WIN-2969

Moving from EQ (bounceable) addresses to UQ (non-bounceable) addresses to fix bouncing deposits issues.

BREAKING CHANGE: All new wallets will generate UQ addresses for ton. External applications and clients depending on these kind of addresses should migrate. No change is being done to existing EQ address wallets.

Wallet Platform will use the bounceable flag based on the walletVersion.

### DIFF
--- a/modules/sdk-coin-ton/src/lib/utils.ts
+++ b/modules/sdk-coin-ton/src/lib/utils.ts
@@ -48,7 +48,7 @@ export class Utils implements BaseUtils {
     }
   }
 
-  async getAddressFromPublicKey(publicKey: string): Promise<string> {
+  async getAddressFromPublicKey(publicKey: string, bounceable = true): Promise<string> {
     const tonweb = new TonWeb(new TonWeb.HttpProvider(''));
     const WalletClass = tonweb.wallet.all['v4R2'];
     const wallet = new WalletClass(tonweb.provider, {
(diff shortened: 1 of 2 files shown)


bhavidhingra (dismissed):
Seems like this was tried and reverted earlier by @ewangbitgo.
Check the PRs of this ticket for reference: https://bitgoinc.atlassian.net/browse/EA-2266

ajays97:
We have now decided to introduce UQ addresses based on wallet versions. Newer v1 wallets will have UQ addresses and older addresses will continue to only deal with addresses in the EQ format. In order to support both, I am now introducing a parameter called `bounceable` that can be used by users of the SDK to specify which address format they want to use for creating addresses. This will by default be true to fallback to EQ addresses.
The decision to set this flag will be made in WP APIs based on wallet versions.

msrshahrukh100 (approved):

modules/sdk-coin-ton/src/lib/utils.ts
```diff
@@ -48,15 +48,27 @@ export class Utils implements BaseUtils {
...
       publicKey: TonWeb.utils.hexToBytes(publicKey),
       wc: 0,
     });
     const address = await wallet.getAddress();
-    return address.toString(true, true, true);
+    return address.toString(true, true, bounceable);
+  }
+
+  getAddress(address: string, bounceable = true): string {
+    if (bounceable) {
```
bhavidhingra, line 63: what if we simply do

`return new TonWeb.Address(address).toString(true, true, bounceable);`

PR #4577, opened 2024-05-29

evalenti-simtlix:
add new batch of tokens

EA-3779

TICKET: EA-3779

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -1880,6 +1880,10 @@ export enum UnderlyingAsset {
   'mobile' = 'mobile',
   'wif' = 'wif',
   'natix' = 'natix',
+  'sol:ks' = 'sol:ks',
+
+  // TRX tokens
+  'trx:jst' = 'trx:jst',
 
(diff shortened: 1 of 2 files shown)


alebusse (changes requested):
@evalenti-simtlix we need to change the way we add non erc20 tokens to the underlying asset list, check this pr https://github.com/BitGo/BitGoJS/pull/4561/files.
Lets use the `coinName:tokenName` format because otherwise it bring issues to WP and backing wallets when the underlying asset name is not the same as the asset name.

evalenti-simtlix:
> @evalenti-simtlix we need to change the way we add non erc20 tokens to the underlying asset list, check this pr https://github.com/BitGo/BitGoJS/pull/4561/files. Lets use the `coinName:tokenName` format because otherwise it bring issues to WP and backing wallets when the underlying asset name is not the same as the asset name.

Perfect, i changed it to use the coinName:tokenName format. I used the other one before to kept how we have been working until now. Is good to know for the future.

alebusse (approved):

PR #4585, opened 2024-05-31
fix(sdk-core): call txRequest migration for all apiVersion=full

zahin-mohammad:
This PR updates `sendManyTss` to return the expected return value from `sendMany`. After this PR it should be aligned with the return value from `v2.wallet.send`.

Blocked till June 4th.

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -3312,8 +3312,7 @@ export class Wallet implements IWallet {
       throw new Error('txRequestId missing from signed transaction');
     }
 
-    // TODO: BG-51122 Remove conditional when moved to txRequestFull for everything
-    if (this._wallet.type === 'custodial') {
+    if (apiVersion === 'full') {
       await this.bitgo
         .post(
           this.bitgo.url(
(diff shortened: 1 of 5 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3312,21 +3312,32 @@ export class Wallet implements IWallet {
       throw new Error('txRequestId missing from signed transaction');
     }
 
-    // TODO: BG-51122 Remove conditional when moved to txRequestFull for everything
-    if (this._wallet.type === 'custodial') {
-      await this.bitgo
+    if (onlySupportsTxRequestFull || apiVersion === 'full') {
```
dpkjnr, line 3315: what if apiVersion is not passed for eddsa custodial wallet?

zahin-mohammad: Then “onlySupportsTxRequestFull” should be true.

alexhu-bitgo (approved):

alebusse:
Should this be considered a breaking change? the UI may break.

zahin-mohammad:
> Should this be considered a breaking change? the UI may break.

Kind of? Technically a bug-fix, this method should never had returned a txRequest to begin with.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3312,21 +3312,32 @@ export class Wallet implements IWallet {
...
+      const latestTxRequest = await getTxRequest(this.bitgo, this.id(), signedTransaction.txRequestId);
+      const transfer: { state: string; pendingApproval?: string; txid?: string } = await this.bitgo
```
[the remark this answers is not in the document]
zahin-mohammad: Potentially, yes. The same as we do in the multisig flow.

PR #4594, opened 2024-06-03
fix(sdk-core): pad alpha and mu shares of MPCV1

islamaminBitGo:
TICKET: HSM-395

BitGo's side of the protocol is expecting 1536 hex length alpha and mu shares.


alebusse (commented):
Any chance you could make a constant and maybe add a comment for those magic numbers?

alebusse (approved):

zhongxishen (approved):

PR #4595, opened 2024-06-03
fix(sdk-coin-sol): added rpc-websockets with a fixed version to sol sdk

alebusse:
This will prevent from bumping the lib rpc-websockets version since it has a breaking change and the solana library doesnt have a locked version of it

WP-2005

TICKET: WP-2005

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,8 @@
     "json5": "^2.2.2",
     "ua-parser-js": ">0.7.30 <0.8.0",
     "socks": "2.7.3",
-    "web3-utils": "4.2.1"
+    "web3-utils": "4.2.1",
+    "rpc-websockets": "7.11.0"
   },
   "workspaces": [
     "modules/*"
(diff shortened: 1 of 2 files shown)


pengyuc-bitgo (approved):

hitansh-bitgo:
@alebusse do we need the same change in account-lib/package.json?

alebusse:
> @alebusse do we need the same change in account-lib/package.json?

No, the root package.json should work for other modules, why?

hitansh-bitgo:
> No, the root package.json should work for other modules, why?

account-lib has this as a dependency - `"@solana/web3.js": "1.91.6"`. When I install the latest account-lib in a project the generated package-lock still uses "7.11.1". 
I was trying to bump deps in WP but it still gave me error for this dependency, so I thought this might be the culprit. I'll try doing a `git clean` & bumping again.

PR #4597, opened 2024-06-04
fix(account-lib): pin `rpc-websockets` version

hitansh-bitgo:
Ticket: WP-2005


hitansh-bitgo:
closing in favour of #4598

PR #4601, opened 2024-06-05
fix(sdk-mpc-lib): replace @noble/secp256k1 with @noble/curves

alebusse:
replaced @noble/secp256k1 with @noble/curves and refactored  the ecdsa curve functions

WP-2008

TICKET: WP-2008


alebusse:
moved to https://github.com/BitGo/BitGoJS/pull/4602

PR #4605, opened 2024-06-06
feat(sdk-core): propagate reqId for prebuildTransaction downstream methods

bdesoky:
TICKET: WP-1862

## Description

Added `reqId` as a prameter to methods and fucntions that are called downstrea of `prebuildTransaction`. I've also updated `prebuildTransaction` to return the request tracer `reqId`.

## Type of change

# How Has This Been Tested?

I have added unit tests where possible to make sure that the api's `setRequestTracer` method is being called properly. I have also tested that `reqId` is returned from `prebuildTransaction`

### DIFF
--- a/modules/bitgo/test/v2/unit/tss/eddsa.ts
+++ b/modules/bitgo/test/v2/unit/tss/eddsa.ts
@@ -19,6 +19,7 @@ import {
   CommitmentShareRecord,
   CommitmentType,
   SignatureShareType,
+  RequestTracer,
 } from '@bitgo/sdk-core';
 import * as openpgp from 'openpgp';
 import * as should from 'should';
@@ -26,6 +27,7 @@ import * as _ from 'lodash';
 import { TestBitGo } from '@bitgo/sdk-test';
 import { BitGo } from '../../../../src/bitgo';
 import { nockGetTxRequest, nockSendSignatureShare } from './helpers';
+import * as sinon from 'sinon';
 import nock = require('nock');
 
 describe('test tss helper functions', function () {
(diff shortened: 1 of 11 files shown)


modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
```diff
@@ -210,7 +212,8 @@ export async function offerUserToBitgoRShare(
   vssProof?: string,
   privateShareProof?: string,
   userPublicGpgKey?: string,
-  publicShare?: string
+  publicShare?: string,
+  reqId?: IRequestTracer
```
margueriteblair, line 216: Do we not have to add the `reqId` param to `ecdsa.ts` as well?

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -518,18 +521,23 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
       txRequestId,
       rShare,
       encryptedSignerShare.share,
-      apiVersion
+      apiVersion,
+      undefined,
```
margueriteblair, line 525: Maybe there's a way to clean these up so they're not all undefined? Maybe we can put `reqId` into a `params` object {} so we don't have to pass in 4 undefined's before?

modules/bitgo/test/v2/unit/tss/eddsa.ts
```diff
@@ -536,6 +538,39 @@ describe('test tss helper functions', function () {
           'signerShare'
         ).should.be.rejectedWith('Invalid RShare, is not from User to Bitgo');
       });
+
+      it('should call setRequestTracer', async function () {
+        const signatureShare = {
```
margueriteblair, line 543: tests look solid

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
bdesoky: Yea I also don't like all those `undefined`, initially I thought about just having a `params` object with all of the parameters, but that would be breaking. If we put `reqId` into a `params` object, wouldn't we still need to use `undefined` for the previous parameters? Or do you mean putting all the optional parameters into `params`?

margueriteblair: Yeah, you're right that would be breaking...hmmmmm. I was thinking passing in all those optionals as a param, but in that case for now this is fine

modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts
bdesoky: Yes I think I missed those, will add them now

margueriteblair (approved):

PR #4607, opened 2024-06-06
fix(abstract-eth): use correct derivation path in wrw recover function

dpkjnr:
Ticket: WP-1937


rushilbg (commented):
Mark it draft until the CI is green?

margueriteblair (approved):

PR #4615, opened 2024-06-10
feat(sdk-core): added propagation of reqId to sendMany

bdesoky:
TICKET: WP-1861

## Description

Propagated `reqId` in `wallet.sendMany`. This also includes methods that are downstream of `sendMany`, and any methods that are called within `sednMany`

## Type of change

# How Has This Been Tested?

I have added a unit test that ensures that `setRequestTracer` is called with the `reqId` passed in the parameters of `sendMany`


margueriteblair (approved):
looks good to me!

PR #4624, opened 2024-06-12
DKLS | Add recreateMpc method

YPAzevedo:
TICKET: WP-1851


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -86,6 +94,7 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
       userGpgKey.publicKey,
       backupGpgKey.publicKey,
       round1Messages
+      // Yago: Pass tag that indicates this is a retrofit on the first round
```
YPAzevedo, line 97: we will need the `bitgoBigSi` here and a tag that dictates this is retrofit for round 1

zahin-mohammad (approved):

PR #4625, opened 2024-06-13
feat: add `recover` function for v1 btc wallets

rushilbg:
Given encrypted user and backup keys, wallet passphrase, unspents list, this function returns a full signed transaction ready for broadcast to the bitcoin network.

BTC-1235


examples/ts/btc/legacySafe/config.ts
```diff
@@ -1,14 +1,26 @@
...
-bitgo.authenticateWithAccessToken({ accessToken });
+const accessToken = '';
+const walletId = '';
+const walletPassphrase = '';
+const userKey = '';
+const backupKey = '';
+
+const sdk = new BitGoAPI({ env });
+sdk.register('tbtc', Tbtc.createInstance);
+sdk.authenticateWithAccessToken({ accessToken });
```
lcovar, line 15: Is this still needed? other than this, lgtm

lcovar (approved):

alebusse (approved):

examples/ts/btc/legacySafe/config.ts
rushilbg: The `register` would be required, the latter might not be. Looking into this and raising another PR if needed.

PR #4628, opened 2024-06-13
feat(root): implement MPCv2 SMC wallet creation utils

alebusse:
implemented MPCv2 SMC wallet creation utils

WP-1860

TICKET: WP-1860

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/SMC/types.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/SMC/types.ts
@@ -1,0 +1,172 @@
+import { DklsTypes } from '@bitgo/sdk-lib-mpc';
+
+export enum KeyCreationMPCv2States {
+  WaitingForOVC1Round1Data = -1,
+  WaitingForOVC2Round1Data,
+  WaitingForBitgoRound1Data,
+  WaitingForOVC1Round2Data,
+  WaitingForOVC2Round2Data,
+  WaitingForBitgoRound2Data,
+  WaitingForOVC1Round3aData,
+  WaitingForOVC2Round3Data,
+  WaitingForOVC1Round3bData,
+  WaitingForBitgoRound3Data,
+  WaitingForOVC1GenerateKey,
+  WaitingForOVC2GenerateKey,
+  KeyGenerationComplete,
+}
+interface baseRoundPayload {
+  tssVersion: string;
+  walletType: string;
+  coin: string;
+  state: KeyCreationMPCv2States;
+}
+
+export const OVC_ONE = 1 as const;
+export const OVC_TWO = 2 as const;
+
+export interface OVC1ToOVC2Round1Payload extends baseRoundPayload {
+  ovc: {
+    [OVC_ONE]: {
+      gpgPubKey: string;
+      ovcMsg1: DklsTypes.AuthBroadcastMessage;
+    };
+  };
+}
+
+export interface OVC2ToBitgoRound1Payload extends baseRoundPayload {
+  ovc: {
+    [key in typeof OVC_ONE | typeof OVC_TWO]: { gpgPubKey: string; ovcMsg1: DklsTypes.AuthBroadcastMessage };
+  };
+}
+
+export interface PlatformRound1Payload {
+  walletGpgPubKeySigs: string;
+  sessionId: string;
+  bitgoMsg1: DklsTypes.AuthBroadcastMessage;
+  ovc: {
+    [key in typeof OVC_ONE | typeof OVC_TWO]: {
+      bitgoToOvcMsg2: DklsTypes.AuthEncP2PMessage;
+    };
+  };
+}
+export interface BitgoToOVC1Round1Response {
+  wallet: OVC2ToBitgoRound1Payload & {
+    platform: PlatformRound1Payload;
+  };
+}
+
+export interface OVC1ToOVC2Round2Payload extends baseRoundPayload {
+  ovc: {
+    [OVC_ONE]: {
+      gpgPubKey: string;
+      ovcMsg1: DklsTypes.AuthBroadcastMessage;
+      ovcToBitgoMsg2: DklsTypes.AuthEncP2PMessage;
+      ovcToOvcMsg2: DklsTypes.AuthEncP2PMessage;
+    };
+    [OVC_TWO]: {
+      gpgPubKey: string;
+      ovcMsg1: DklsTypes.AuthBroadcastMessage;
+    };
+  };
+  platform: PlatformRound1Payload;
+}
+
+export interface OVC2ToBitgoRound2Payload extends baseRoundPayload {
+  ovc: {
+    [key in typeof OVC_ONE | typeof OVC_TWO]: {
+      gpgPubKey: string;
+      ovcMsg1: DklsTypes.AuthBroadcastMessage;
+      ovcToBitgoMsg2: DklsTypes.AuthEncP2PMessage;
+      ovcToOvcMsg2: DklsTypes.AuthEncP2PMessage;
+    };
+  };
+  platform: PlatformRound1Payload;
+}
+
+export interface PlatformRound2Payload {
+  sessionId: string;
+  bitgoMsg1: DklsTypes.AuthBroadcastMessage;
+  bitgoCommitment2: string;
+  ovc: {
+    [key in typeof OVC_ONE | typeof OVC_TWO]: {
+      bitgoToOvcMsg2: DklsTypes.AuthEncP2PMessage;
+      bitgoToOvcMsg3: DklsTypes.AuthEncP2PMessage;
+    };
+  };
+}
+
+export interface BitgoToOVC1Round2Response {
+  wallet: OVC2ToBitgoRound2Payload & {
+    platform: PlatformRound2Payload;
+  };
+}
+
+export interface OVC1ToOVC2Round3Payload extends baseRoundPayload {
+  ovc: {
+    [OVC_ONE]: {
+      gpgPubKey: string;
+      ovcMsg1: DklsTypes.AuthBroadcastMessage;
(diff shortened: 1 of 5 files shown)


OttoAllmendinger (commented):
flush - insufficient context

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -1168,4 +1179,140 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
     return { signatureShareRound3 };
   }
   // #endregion
+
+  // #region self-managed cold utils
+
+  public async keyGenSMCRound1(
```
pranavjain97, line 1185:
```suggestion
  public async keyGenOVCRound1(
```

Since this is so specific to the OVC format, i'd prefer if we rename it to OVC so it can be searched for easily.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -1168,4 +1179,140 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
...
+        platform: {
+          ...payload.platform,
+          commonKeychain: result.commonKeychain,
+          bitgoMsg4: this.formatBitgoBroadcastMessage(result.bitgoMsg4),
+        },
+      },
+    };
+  }
+
+  public async keyGenSMCAddOthersKeys(
```
pranavjain97, line 1296:
```suggestion
  public async keyGenOVCAddUserAndBackupKeys(
```

pranavjain97 (changes requested):
Because all this is only needed/used by the OVC and specific to its formatting, I would honestly prefer if we created a separate utils sub directory like `ovcUtils` under `utils/tss/ecdsa` and moved all this code under that so that it doesn't populate/distract from the core `ecdsaMPCv2` keygen logic. (Same for the types)

pengyuc-bitgo:
Agree with @pranavjain97 on extracting the OVC-specific codes out of those files.

alebusse:
> Because all this is only needed/used by the OVC and specific to its formatting, I would honestly prefer if we created a separate utils sub directory like `ovcUtils` under `utils/tss/ecdsa` and moved all this code under that so that it doesn't populate/distract from the core `ecdsaMPCv2` keygen logic. (Same for the types)

These are client utils, like the UI, not for OVC.

alebusse:
> Agree with @pranavjain97 on extracting the OVC-specific codes out of those files.

The types will be extracted once we put them in the public-types library and will refactor this to use the codecs also.
But these files are for MPCv2 utils, so are these new methods.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/typesMPCv2.ts
```diff
@@ -29,3 +30,174 @@ export const generateMPCv2KeyRequestResponse = t.union([
...
+  WaitingForBitgoRound2Data,
+  WaitingForOVC1Round3aData,
+  WaitingForOVC2Round3Data,
+  WaitingForOVC1Round3bData,
+  WaitingForBitgoRound3Data,
+  WaitingForOVC1GenerateKey,
+  WaitingForOVC2GenerateKey,
+  KeyGenerationComplete,
+}
+interface baseRoundPayload {
```
zahin-mohammad, line 51: We should make these all codecs and extract types from the codec.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
zahin-mohammad: I think the naming can be updated as well, `OtherKeys` is a little vague. (i.e. if I where reading the call site of this function, I have to read the function body to know what it's doing).

zahin-mohammad: Also how is this method intended to be used?

zahin-mohammad (changes requested):
lgtm. Let's remove the lockfile change.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/typesMPCv2.ts
zahin-mohammad: Looks like this was already discussed.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
alebusse: This is not really necessary but an easy way for the clients to create the user and backup keys docs once the OVCs generation is done, then they can be used to generate a wallet.

alebusse:
@zahin-mohammad renamed keyGenSMCAddOthersKeys to keyGenSMCAddOthersKeys, this method is for creating the user and backup keys once they generation is done in the OVCs, after this you should have the 3 keys ids to be able to create the wallet.
Im thinking that maybe the input for this method should be the bitgoKey id and the commonKeychain, so i can validate it matches before creating the User and Backup, that will be more useful too.

pranavjain97:
> > Agree with @pranavjain97 on extracting the OVC-specific codes out of those files.
> 
> The types will be extracted once we put them in the public-types library and will refactor this to use the codecs also. But these files are for MPCv2 utils, so are these new methods.

Right, which is why i think its fine to have them as a subdirectory under `utils/tss/ecdsa`. Any client wanting to use ovc/smc related code can go direcrtly there and doesn't need to worry about the other methods there.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/SMC/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+  private MPCv2Utils: EcdsaMPCv2Utils;
+
+  constructor(bitgo: BitGoBase, private baseCoin: IBaseCoin) {
+    this.MPCv2Utils = new EcdsaMPCv2Utils(bitgo, baseCoin);
+  }
+
+  public async keyGenRound1(enterprise: string, payload: OVC2ToBitgoRound1Payload): Promise<BitgoToOVC1Round1Response> {
+    assert(
+      payload.state === KeyCreationMPCv2States.WaitingForBitgoRound1Data,
+      'Invalid state for round 1, got: ' + payload.state
```
mohammadalfaiyazbitgo, line 27:
```suggestion
      `Invalid state for round 1, got: ${payload.state}`
```

modules/sdk-core/src/bitgo/utils/tss/ecdsa/SMC/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+    );
+    const ovc1 = payload.ovc[1];
+    const ovc2 = payload.ovc[2];
```
mohammadalfaiyazbitgo, line 30: nit: can we use the const for party (user/backup/bitgo) instead of `1/2`

modules/sdk-core/src/bitgo/utils/tss/ecdsa/SMC/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+          },
+        },
+      },
+    };
+  }
+
+  public async keyGenRound2(enterprise: string, payload: OVC2ToBitgoRound2Payload): Promise<BitgoToOVC1Round2Response> {
+    assert(
+      payload.state === KeyCreationMPCv2States.WaitingForBitgoRound2Data,
+      'Invalid state for round 2, got: ' + payload.state
```
mohammadalfaiyazbitgo, line 64: nit: same formatting as above.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/SMC/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+    );
+    const ovc1 = payload.ovc[1];
+    const ovc2 = payload.ovc[2];
```
mohammadalfaiyazbitgo, line 67: similar point.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/SMC/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+        coin: payload.coin,
+        ovc: payload.ovc,
+        state: KeyCreationMPCv2States.WaitingForOVC1Round3aData,
+        platform: {
+          ...payload.platform,
+          sessionId: result.sessionId,
+          bitgoCommitment2: result.bitgoCommitment2,
+          ovc: {
+            1: { ...payload.platform.ovc[1], bitgoToOvcMsg3: this.MPCv2Utils.formatP2PMessage(result.bitgoToUserMsg3) },
+            2: {
```
mohammadalfaiyazbitgo, line 85: same point about using const for 1/2

pranavjain97 (approved):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/SMC/types.ts
```diff
@@ -0,0 +1,172 @@
+import { DklsTypes } from '@bitgo/sdk-lib-mpc';
+
+export enum KeyCreationMPCv2States {
+  WaitingForOVC1Round1Data = -1,
```
zahin-mohammad, line 4: Why does it start at -1?

zahin-mohammad (approved):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/SMC/types.ts
alebusse: Thats the "initial" state in OVC

PR #4630, opened 2024-06-14
feat: add btc public signet network and coin

ericli-bitgo:
TICKET: BTC-1250


modules/utxo-lib/src/networks.ts
```diff
@@ -116,6 +117,15 @@ export const networks: Record<NetworkName, Network> = {
     wif: 0xef,
     coin: coins.BTC,
   },
+  bitcoinPublicSignet: {
```
OttoAllmendinger, line 120: please add jsdoc here what exact network is meant

modules/utxo-lib/src/networks.ts
```diff
@@ -116,6 +117,15 @@ export const networks: Record<NetworkName, Network> = {
...
+    messagePrefix: '\x18Bitcoin Signed Message:\n',
+    bech32: 'tb',
+    bip32: getDefaultBip32Testnet(),
+    pubKeyHash: 0x6f,
+    scriptHash: 0xc4,
+    wif: 0xef,
+    coin: coins.BTC,
```
OttoAllmendinger, line 127: please add source for the constants as a review comment

OttoAllmendinger (changes requested):

modules/utxo-lib/src/networks.ts
ericli-bitgo: don't have a source, but I believe the public signet constants are the same as testnet - please correct me if I am wrong

ericli-bitgo: nvm, I think I do need to change some of the constants - esp. the header (`0xecc7daa2`)

ericli-bitgo: maybe not in this specific code block though

modules/utxo-lib/src/networks.ts
ericli-bitgo, line 120: Source for constants: https://github.com/bitcoin/bitcoin/blob/2c79abc7ad4850e9e3ba32a04c530155cda7f980/src/kernel/chainparams.cpp#L387

modules/utxo-lib/src/networks.ts
```diff
@@ -116,6 +117,18 @@ export const networks: Record<NetworkName, Network> = {
     wif: 0xef,
     coin: coins.BTC,
   },
+  /**
+   * Additional testnet for BTC that restricts mining to trusted entities (BIP 325)
```
OttoAllmendinger, line 121:
```suggestion
   * Additional testnet for BTC that restricts mining to trusted entities (BIP 325)
   * Address encoding is the same as for testnet.
```

modules/utxo-lib/test/integration_local_rpc/fixtures/bitcoinPublicSignet/v1/deposit_p2pkh.json
```diff
@@ -0,0 +1,104 @@
+{
```
OttoAllmendinger, line 1: I assume these are all derived via the btc regtest node anyway

please change the test code so that we use the same fixtures for signet and testnet here

OttoAllmendinger (changes requested):

modules/utxo-lib/test/integration_local_rpc/fixtures/bitcoinPublicSignet/v1/deposit_p2pkh.json
ericli-bitgo: I tried, but testnet addresses are e.g., `bcrt1q3vyfqm32xqwnmv0up47n9r00pch2qmz8se0ngw` using the fixtures, while public signet addresses are `t1q...`

ericli-bitgo: If the prefixes matter, I can do some digging to see why they're different

modules/statics/src/utxo.ts
```diff
@@ -199,6 +199,15 @@ export const utxoCoins: Readonly<BaseCoin>[] = [
     BaseUnit.BTC,
     BTC_FEATURES
   ),
+  utxo(
+    'a0d53dc9-2dcc-4ebb-a2d4-51983fe20da6',
```
OttoAllmendinger, line 203: where do these come from by the way?

modules/utxo-lib/src/networks.ts
OttoAllmendinger: pls add this link into jsdoc https://github.com/bitcoin/bitcoin/blob/v27.1/src/kernel/chainparams.cpp#L287-L290

modules/utxo-lib/test/integration_local_rpc/fixtures/bitcoinPublicSignet/v1/deposit_p2pkh.json
OttoAllmendinger: Oh I see, the `testnet` fixtures are actually `regtest`

OttoAllmendinger (changes requested):
jsdoc change pls

modules/statics/src/utxo.ts
ericli-bitgo: these were randomly generated

lcovar (approved):

OttoAllmendinger (approved):

PR #4636, opened 2024-06-17
feat(abstract-utxo): add internal/external info when known

davidkaplanbitgo:
In the case that we have a PSBT, we do not want to query WP for all of the outputs to see if they are internal and external since we have that information already.

TICKET: BTC-1268

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -112,11 +112,24 @@ export interface VerifyAddressOptions extends BaseVerifyAddressOptions {
   index: number;
 }
 
-export interface Output {
+export interface BaseOutput {
   address: string;
   amount: string | number;
+  // Even though this external flag is redundant with the chain property, it is necessary for backwards compatibility
+  // with legacy transaction format.
   external?: boolean;
+}
+
+export interface WalletOutput extends BaseOutput {
   needsCustomChangeKeySignatureVerification?: boolean;
+  chain: number;
+  index: number;
+}
+
+export type Output = BaseOutput | WalletOutput;
+
+export function isWalletOutput(output: Output): output is WalletOutput {
+  return (output as WalletOutput).chain !== undefined && (output as WalletOutput).index !== undefined;
 }
 
 export interface TransactionExplanation extends BaseTransactionExplanation<string, string> {
@@ -148,6 +161,11 @@ export interface ExplainTransactionOptions<TNumber extends number | bigint = num
   txInfo?: TransactionInfo<TNumber>;
   feeInfo?: string;
   pubs?: Triple<string>;
+}
+
+export interface DecoratedExplainTransactionOptions<TNumber extends number | bigint = number>
+  extends ExplainTransactionOptions<TNumber> {
+  changeInfo?: { address: string; chain: number; index: number }[];
 }
 
 export type UtxoNetwork = utxolib.Network;
@@ -592,7 +610,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     );
 
     const needsCustomChangeKeySignatureVerification = allOutputDetails.some(
-      (output) => output.needsCustomChangeKeySignatureVerification
+      (output) => (output as WalletOutput)?.needsCustomChangeKeySignatureVerification
     );
 
     const changeOutputs = _.filter(allOutputDetails, { external: false });
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -33,7 +33,6 @@ import {
   IBaseCoin,
   InvalidAddressDerivationPropertyError,
   InvalidAddressError,
-  InvalidAddressVerificationObjectPropertyError,
   IRequestTracer,
   isTriple,
   ITransactionExplanation as BaseTransactionExplanation,
@@ -967,7 +966,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
    * @throws {UnexpectedAddressError}
    */
   async isWalletAddress(params: VerifyAddressOptions): Promise<boolean> {
-    const { address, addressType, keychains, coinSpecific, chain, index } = params;
+    const { address, addressType, keychains, chain, index } = params;
 
     if (!this.isValidAddress(address)) {
       throw new InvalidAddressError(`invalid address: ${address}`);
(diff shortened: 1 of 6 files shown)


modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -117,6 +117,8 @@ export interface Output {
   amount: string | number;
   external?: boolean;
```
OttoAllmendinger, line 118: this seems to be redundant with `chain` and `index` being set

elsewhere we split this up into two types

```
export interface BaseOutput {
  address: string;
  amount: string | number;
}

export interface WalletOutput extends Output {
  needsCustomChangeKeySignatureVerification?: boolean;
  chain: number;
  index: number;
}

type Output = BaseOutput | WalletOutput;

function isWalletOutput(o: Output): o is WalletOutput {
  return o.chain !== undefined;
}
```

modules/abstract-utxo/src/parseOutput.ts
```diff
@@ -222,6 +222,29 @@ export async function parseOutput({
   let currentAddressType: string | undefined = undefined;
   const RECIPIENT_THRESHOLD = 1000;
   try {
+    // In the case of PSBTs, we can already determine the internal/external status of the output addresses
+    // based on the derivation information being included in the PSBT. We can short circuit GET v2.wallet.address
+    // and save on network requests. Since we have the derivation information already, we can still verify the address
+    if (currentOutput.external !== undefined) {
```
OttoAllmendinger, line 228: you can then use `isWalletOutput(o)` here

modules/utxo-lib/src/bitgo/wallet/psbt/PsbtOutputs.ts
```diff
@@ -79,6 +79,26 @@ export function findInternalOutputIndices(psbt: UtxoPsbt): number[] {
   return findWalletOutputIndices(psbt, rootNodes);
 }
 
+export function getChainAndIndexFromInternalOutput(output: PsbtOutput): { chain: number; index: number } | undefined {
```
OttoAllmendinger, line 82: it's unclear how we are asserting that the output is _internal_ here - as I understand even external outputs can have derivation info.

Can we pass in the wallet xpubs as an argument maybe?

OttoAllmendinger (changes requested):

modules/abstract-utxo/src/abstractUtxoCoin.ts
davidkaplanbitgo: Agreed in principle, but we still need it because we still support legacy format. If it is a PSBT we can explicitly set `external = true` and skip the check. Otherwise if it was just PSBTs I would agree with you

modules/abstract-utxo/src/parseOutput.ts
davidkaplanbitgo: not really, if `external === true` we want to go in here as well

OttoAllmendinger: gotcha

modules/abstract-utxo/src/transaction.ts
```diff
@@ -120,6 +130,12 @@ function explainCommon<TNumber extends number | bigint>(
     outputs.push({
       address: currentAddress,
       amount: currentAmount.toString(),
+      // If changeInfo has a length greater than zero, it means that the change information
```
OttoAllmendinger, line 133: in JS empty objects are truthy, if you want to identify an empty object you can use `Object.keys(changeInfo).length`

modules/abstract-utxo/src/transaction.ts
```diff
@@ -207,21 +223,57 @@ export function explainPsbt<TNumber extends number | bigint>(
...
+        throw new Error(
+          'Unable to get a single chain and index on the output because there are different paths for different keys'
+        );
+      }
+    });
+    const derivationIndices = paths[0].split('/');
+    return {
+      chain: Number(derivationIndices[2]),
+      index: Number(derivationIndices[3]),
+    };
```
OttoAllmendinger, line 251: please add a function `getChainAndIndexFromPath(s: string) { chain: number, index: number }` that has more verification:

* check that path has right number of components
* check that chain and index are positive integers

pls add unit tests

OttoAllmendinger (dismissed):

modules/abstract-utxo/src/transaction.ts
davidkaplanbitgo: I need to update this comment - only when changeInfo is undefined do we want to skip this. 

In the case that changeInfo is an empty array - it means that the outputs have been checked internally that they are internal/external but none of them are. This is also the way that we can say that outs are external, not just internal. 

I will update the Jsdoc and this comment to say this, because it isn't obvious

modules/abstract-utxo/src/parseOutput.ts
```diff
@@ -222,6 +222,29 @@ export async function parseOutput({
...
+      // In the case that we have a custom change wallet, we need to verify the address against the custom change keys
+      // and not the wallet keys. This check is done in the handleVerifyAddressError function if this error is thrown.
+      if (customChange !== undefined) {
+        throw new UnexpectedAddressError('`address validation failure');
+      }
+      // If it is an internal address, we can skip the network request and just verify the address locally with the
+      // derivation information we have. Otherwise, if the address is external, which is the only remaining case, we
+      // can just return the current output as is without contacting the server.
+      if (isWalletOutput(currentOutput)) {
+        await coin.isWalletAddress({
```
OttoAllmendinger, line 238: The name suggests that this returns a boolean - if that's not the case please add a comment. Otherwise check the result.

lcovar (approved):

modules/utxo-lib/src/bitgo/parseInput.ts
```diff
@@ -698,3 +698,20 @@ export function parsePubScript(
...
+export function getChainAndIndexFromPath(path: string): { chain: number; index: number } {
+  const parts = path.split('/');
+  if (parts.length <= 2) {
+    throw new Error(`invalid path "${path}"`);
+  }
+  const chain = Number(parts[parts.length - 2]);
+  const index = Number(parts[parts.length - 1]);
+  if (isNaN(chain) || isNaN(index)) {
+    throw new Error(`Could not parse chain and index into numbers from path ${path}`);
+  }
```
OttoAllmendinger, line 711: use `Number.isInteger()` here

PR #4638, opened 2024-06-18
feat(sdk-coin-xlm): root key support for xlm

dwei-bitgo:
TICKET: WP-1418

### DIFF
--- a/modules/sdk-coin-xlm/src/lib/keyPair.ts
+++ b/modules/sdk-coin-xlm/src/lib/keyPair.ts
@@ -1,0 +1,49 @@
+import * as stellar from 'stellar-sdk';
+import { DefaultKeys, Ed25519KeyPair, KeyPairOptions, NotImplementedError } from '@bitgo/sdk-core';
+
+export class KeyPair extends Ed25519KeyPair {
+  /**
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   * @param { KeyPairOptions } source Either a master seed, a private key, or a public key
+   */
+  constructor(source?: KeyPairOptions) {
+    super(source);
+  }
+
+  getKeys(raw = false): DefaultKeys {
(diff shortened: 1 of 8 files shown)


modules/sdk-coin-xlm/src/lib/keyPair.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (this.keyPair.prv) {
+      const privateKey = stellar.StrKey.encodeEd25519SecretSeed(Buffer.from(this.keyPair.prv, 'hex'));
+      result.prv = raw ? stellar.StrKey.decodeEd25519SecretSeed(privateKey).toString('hex') : privateKey;
+    }
+
+    return result;
+  }
+
+  /** @inheritdoc */
+  getAddress(): string {
```
dwei-bitgo, line 29: Not sure if this is necessary for XLM

alebusse: I think its just the pub key in XLM format, so just return the pub

modules/sdk-coin-xlm/src/xlm.ts
```diff
@@ -246,7 +259,7 @@ export class Xlm extends BaseCoin {
    * @returns is it valid?
    */
   isValidPub(pub: string): boolean {
-    return stellar.StrKey.isValidEd25519PublicKey(pub);
+    return stellar.StrKey.isValidEd25519PublicKey(pub) || isValidEd25519PublicKey(pub);
```
dpkjnr, line 262: why both? it looks confusing

dwei-bitgo: the stellar validation method only permits the stellar specific G-prefix version of public key, it fails to validate raw root keys in hex form, that is what the second method is there for, will add a code comment to clarify

alebusse (approved):

CapnMigraine (approved):

PR #4655, opened 2024-06-24
fix(sdk-core): default to MPCv2 for ecdsa

alebusse:
removed the condition for walletVersion 3 since we need to default every ecdsa wallet to MPCv2 (cosmos and evm)

[WP-2089](https://bitgoinc.atlassian.net/browse/WP-2089)

TICKET: [WP-2089](https://bitgoinc.atlassian.net/browse/WP-2089)

[WP-2089]: https://bitgoinc.atlassian.net/browse/WP-2089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WP-2089]: https://bitgoinc.atlassian.net/browse/WP-2089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallets.ts
@@ -789,7 +789,7 @@ export class Wallets implements IWallets {
     originalPasscodeEncryptionCode,
     backupProvider,
   }: GenerateMpcWalletOptions): Promise<WalletWithKeychains> {
-    if (multisigType === 'tss' && this.baseCoin.getMPCAlgorithm() === 'ecdsa' && walletVersion === 3) {
+    if (multisigType === 'tss' && this.baseCoin.getMPCAlgorithm() === 'ecdsa') {
       const tssSettings: TssSettings = await this.bitgo
         .get(this.bitgo.microservicesUrl('/api/v2/tss/settings'))
         .result();


bdesoky (approved):

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -789,7 +789,7 @@ export class Wallets implements IWallets {
     originalPasscodeEncryptionCode,
     backupProvider,
   }: GenerateMpcWalletOptions): Promise<WalletWithKeychains> {
-    if (multisigType === 'tss' && this.baseCoin.getMPCAlgorithm() === 'ecdsa' && walletVersion === 3) {
+    if (multisigType === 'tss' && this.baseCoin.getMPCAlgorithm() === 'ecdsa') {
```
mohammadalfaiyazbitgo, line 792: in this case why not remove `walletVersion` from the parameters and determine the wallet version based on the tss Settings? Otherwise you're leaving a hole if a user passes in `undefined or 2` and they don't have mpcv2 enabled.

alebusse: because if MPCv2 is disabled, we do nothing and let the flow continues with whatever walletVersion was passed in

mohammadalfaiyazbitgo: is wallet version 5 blocked if the enterprise/user doesn't have the correct flags?

alebusse: yes, depends on the user flag or the env flag, only if one of those is enabled it works

mohammadalfaiyazbitgo (approved):

PR #4666, opened 2024-06-26
fix: find change address instead of using index

lcovar:
Ticket: BTC-0000


rushilbg (approved):
Would be helpful to know where this caused an issue.

davidkaplanbitgo (approved):

lcovar:
> Would be helpful to know where this caused an issue.

The code had an array of change addresses and it was assuming that the index of the change address in the tx.outs was the was found in the same index of the change address array.

so if tx.out[1].address was change it would check changeInfo[1] but this wasn't always true. It would only be true if changeInfo contained all addresses, not only change. 

Using .find instead checks if the change address is present at any index in the changeInfo array

PR #4685, opened 2024-07-03
feat(abstract-utxo): allow non-segwit signing by looking at txHex

davidkaplanbitgo:
TICKET: BTC-000

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -1534,7 +1534,14 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   }
 
   async presignTransaction(params: PresignTransactionOptions): Promise<any> {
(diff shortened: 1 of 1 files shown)


modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -1534,7 +1534,12 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   }
 
   async presignTransaction(params: PresignTransactionOptions): Promise<any> {
-    if (params.bulk && params.allowNonSegwitSigningWithoutPrevTx === undefined) {
```
davidkaplanbitgo, line 1537: It does not make sense to keep the check for `params.bulk` because it doesn't really make sense in retrospect

OttoAllmendinger (dismissed):

lcovar (approved):

PR #4689, opened 2024-07-03
feat(root) implement external signer for mpcv2

alebusse:
added base utils for MPCv2 external signer
added MPCv2 base code to support external signer
Implemented singing using external signer for MPCv2
Use the hashed signableTx as Adata to verify that the encrypted session matches with the tx during
the second and third DSG rounds in external signer, this would prevent re using the stored session
for a different tx

WP-2136

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -4,6 +4,7 @@
 import {
   EddsaUtils,
   EcdsaUtils,
+  EcdsaMPCv2Utils,
   CustomGShareGeneratingFunction,
   CustomRShareGeneratingFunction,
   UnsupportedCoinError,
@@ -24,6 +25,10 @@ import {
   CreateNetworkConnectionParams,
   GetNetworkPartnersResponse,
   encryptRsaWithAesGcm,
+  Wallet,
+  CustomMPCv2Round1GeneratingFunction,
+  CustomMPCv2Round2GeneratingFunction,
+  CustomMPCv2Round3GeneratingFunction,
 } from '@bitgo/sdk-core';
 import { BitGo, BitGoOptions, Coin, CustomSigningFunction, SignedTransaction, SignedTransactionRequest } from 'bitgo';
 import * as bodyParser from 'body-parser';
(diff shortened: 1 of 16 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -828,7 +855,7 @@ function createSendParams(req: express.Request) {
   }
 }
 
-function createTSSSendParams(req: express.Request) {
+function createTSSSendParams(req: express.Request, wallet?: Wallet) {
```
zahin-mohammad, line 858: shouldn't wallet be required? Are there routes that aren't passing it in?

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -107,6 +110,39 @@ export interface CustomGShareGeneratingFunction {
     bitgoToUserCommitment: CommitmentShareRecord;
   }): Promise<GShare>;
 }
+
+export interface CustomMPCv2Round1GeneratingFunction {
```
zahin-mohammad, line 114: Nit: signing and key gen both have multiple rounds, would be clearer if we specified signing here.
```suggestion
export interface CustomMPCv2SigningRound1GeneratingFunction {
```

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -724,5 +718,306 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
...
+    externalSignerMPCv2Round3Generator: CustomMPCv2Round3GeneratingFunction,
+    requestType: RequestType = RequestType.tx
+  ): Promise<TxRequest> {
+    const { txRequest, reqId } = params;
+    let txRequestResolved: TxRequest;
+
+    assert(
+      requestType === RequestType.tx,
+      'Only transaction signing is supported for external signer, got: ' + requestType
+    );
```
zahin-mohammad, line 801: Lets add a ticket to support message signing with external signer. Don't want to let it fall through the cracks even though it is not needed atm.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3152,6 +3161,54 @@ export class Wallet implements IWallet {
...
+    if (!params.customMPCv2Round3GenerationFunction) {
+      throw new Error('Generator function for MPCv2 Round 3 share required to sign transactions with External Signer.');
+    }
+
+    try {
+      assert(this.tssUtils, 'tssUtils must be defined');
+      const signedTxRequest = await this.tssUtils.signEcdsaMPCv2TssUsingExternalSigner(
+        {
+          txRequest: txRequestId,
+          prv: '',
```
zahin-mohammad, line 3199: Nit: we should allow undefined `prv` if the custom signing functions are passed. Seems weird to pass an empty string.

modules/sdk-core/src/bitgo/utils/tss/baseTSSUtils.ts
```diff
@@ -154,6 +157,26 @@ export default class BaseTssUtils<KeyShare> extends MpcUtils implements ITssUtil
...
+   * @returns {Promise<TxRequest>} - a signed tx request
+   */
+  signEcdsaMPCv2TssUsingExternalSigner(
+    params: TSSParams | TSSParamsForMessage,
+    externalSignerMPCv2Round1Generator: CustomMPCv2Round1GeneratingFunction,
+    externalSignerMPCv2Round2Generator: CustomMPCv2Round2GeneratingFunction,
+    externalSignerMPCv2Round3Generator: CustomMPCv2Round3GeneratingFunction,
+    requestType?: RequestType
+  ): Promise<TxRequest> {
+    throw new Error('Method not implemented.');
```
zahin-mohammad, line 177: This still needs to be implemented in this PR right? Or will this be addressed in a different PR?

alebusse: Nop, the base utils use the non implemented errors incase you used the wrong util (like eddsa)

modules/express/src/clientRoutes.ts
alebusse: its not required for Eddsa since we dont check the multisigTypeVersion, but i can make it required

modules/sdk-core/src/bitgo/wallet/wallet.ts
alebusse: Ya, its mostly because its using the wrong type. i will fix it

zahin-mohammad (dismissed):

zahin-mohammad (approved):

islamaminBitGo (approved):

PR #4690, opened 2024-07-03
feat(sdk-lib-mpc): implement session get and restore for dsg

alebusse:
implemented new methods for DSG class to get current session and restore session that are required for async signing used in external signer

WP-2166

TICKET: WP-2166

### DIFF
--- a/modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dsg.ts
+++ b/modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dsg.ts
@@ -49,6 +49,41 @@ export class Dsg {
     }
   }
 
+  /**
+   * Returns the current DSG session as a base64 string.
+   * @returns {string} - base64 string of the current DSG session
+   */
+  getSession(): string {
+    return Buffer.from(this.dsgSessionBytes).toString('base64');
+  }
+
+  /**
+   * Sets the DSG session from a base64 string.
+   * @param {string} session - base64 string of the DSG session
+   */
+  setSession(session: string): void {
+    this.dsgSession = undefined;
+    const sessionBytes = new Uint8Array(Buffer.from(session, 'base64'));
+    const round = decode(sessionBytes).round;
+    switch (true) {
+      case round === 'WaitMsg1':
+        this.dsgState = DsgState.Round1;
+        break;
+      case round === 'WaitMsg2':
+        this.dsgState = DsgState.Round2;
+        break;
+      case round === 'WaitMsg3':
+        this.dsgState = DsgState.Round3;
+        break;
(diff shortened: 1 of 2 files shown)


modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dsg.ts
```diff
@@ -65,6 +99,8 @@ export class Dsg {
     try {
       const payload = this.dsgSession.createFirstMessage().payload;
       this._deserializeState();
+      this.dsgSessionBytes = this.dsgSession.toBytes();
+      this.dsgSession = undefined;
```
alebusse, line 103: adding a similar behaviour as the finally code block in `handleIncomingMessages`

modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dsg.ts
```diff
@@ -49,6 +49,40 @@ export class Dsg {
     }
   }
 
+  /**
+   * Returns the current DSG session as a hex string.
+   * @returns {string} - hex string of the current DSG session
+   */
+  getSession(): string {
+    return Buffer.from(this.dsgSessionBytes).toString('hex');
```
islamaminBitGo, line 57: user base64 so the string is more compact

zahin-mohammad (approved):
lgtm

modules/sdk-lib-mpc/test/unit/tss/ecdsa/dklsDsg.ts
```diff
@@ -255,4 +264,149 @@ describe('DKLS Dsg 2x3', function () {
...
+        broadcastMessages: [party1Round1Message],
+      });
+    }
+
+    // round 3
+    let party1Round3Messages: DeserializedMessages,
+      party2Round3Messages: DeserializedMessages,
+      party1Round3Session: string;
+    {
+      const party1Round3DSG = new DklsDsg.Dsg(
```
islamaminBitGo, line 324: instead of having to construct a Dsg object from scratch everytime and calling setSession, i think we should support constructing a Dsg object using a function like fromSessionBytes(buffer)

alebusse: Ya i was thinking the same but since there are not exposed methods to get `keyShareBytes`, `messageHash`, `derivationPath` and `partyIdx` from the `dsgSession` (sometime they are not there yet), i was thinking in making the session payload to include all this, but that would mean that the `key` would leave external signer.

islamaminBitGo (approved):

PR #4694, opened 2024-07-04
feat: ignore additional external outputs

lcovar:
Update the validation of paygo amount in output only in case of absence of `allowExternalChangeOutputs`, and check for the presence of the external change output in case this flag is passed in.

TICKET: BTC-1313


rushilbg:
Closing this in favor of https://github.com/BitGo/BitGoJS/pull/4700
because Github is not able to process this one.

Screenshot 2024-07-05 at 7 01 08 PM

PR #4702, opened 2024-07-05
feat(sdk-coin-xlm): root key support for XLM recovery and refactor XLM utils

dwei-bitgo:
TICKET: [WP-2094](https://bitgoinc.atlassian.net/browse/WP-2094)

[WP-2094]: https://bitgoinc.atlassian.net/browse/WP-2094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-coin-xlm/src/lib/utils.ts
+++ b/modules/sdk-coin-xlm/src/lib/utils.ts
@@ -9,9 +9,17 @@ import * as stellar from 'stellar-sdk';
  */
 export function createStellarKeypairFromPub(pub: string): stellar.Keypair {
   if (pub.startsWith('G')) {
+    if (!isValidStellarPublicKey(pub)) {
+      throw new Error('Invalid Stellar public key');
+    }
     return stellar.Keypair.fromPublicKey(pub);
   }
-  return stellar.Keypair.fromPublicKey(encodePublicKey(Buffer.from(pub, 'hex')));
+
+  const encodedPub = encodePublicKey(Buffer.from(pub, 'hex'));
+  if (!isValidStellarPublicKey(encodedPub)) {
+    throw new Error('Invalid root public key');
+  }
+  return stellar.Keypair.fromPublicKey(encodedPub);
 }
 
 /**
@@ -22,9 +30,17 @@ export function createStellarKeypairFromPub(pub: string): stellar.Keypair {
  */
 export function createStellarKeypairFromPrv(prv: string): stellar.Keypair {
   if (prv.startsWith('S')) {
+    if (!isValidStellarPrivateKey(prv)) {
+      throw new Error('Invalid Stellar private key');
+    }
     return stellar.Keypair.fromSecret(prv);
   }
-  return stellar.Keypair.fromSecret(encodePrivateKey(Buffer.from(prv.slice(0, 64), 'hex')));
+
+  const encodedPrv = encodePrivateKey(Buffer.from(prv.slice(0, 64), 'hex'));
+  if (!isValidStellarPrivateKey(encodedPrv)) {
+    throw new Error('Invalid root private key');
+  }
+  return stellar.Keypair.fromSecret(encodedPrv);
 }
 
 /**
--- a/modules/sdk-coin-xlm/src/lib/utils.ts
+++ b/modules/sdk-coin-xlm/src/lib/utils.ts
@@ -28,6 +28,18 @@ export function createStellarKeypairFromPrv(prv: string): stellar.Keypair {
 }
 
 /**
+ * @deprecated Use isValidStellarPublicKey instead
+ *
+ * Validates a Stellar public key
+ * Stellar public keys are prefixed with 'G'
+ * @param pub A Stellar public key to validate
+ * @returns Whether the input is a valid Stellar public key
+ */
+export function isValidStellarPub(pub: string): boolean {
+  return stellar.StrKey.isValidEd25519PublicKey(pub);
+}
+
+/**
  * Validates a Stellar public key
  * Stellar public keys are prefixed with 'G'
  * @param pub A Stellar public key to validate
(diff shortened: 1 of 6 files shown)


modules/sdk-coin-xlm/test/unit/xlm.ts
```diff
@@ -314,12 +314,12 @@ describe('XLM:', function () {
...
       pub: '37c6be1c677873e42a90f3d0ed3b0e4f3ba8591f285769d8c922f7024ae0c009',
-      prv: 'c034ca796a145e79acfac9bc9d97e4d3aaa42bfbe648b46bb1f3da794d7765fb',
+      prv: 'c034ca796a145e79acfac9bc9d97e4d3aaa42bfbe648b46bb1f3da794d7765fb37c6be1c677873e42a90f3d0ed3b0e4f3ba8591f285769d8c922f7024ae0c009',
     };
     // This key pair is the decoded version of the backupKeychain above
     const backupRootKeychain = {
       pub: 'b63de5ad57cc75b256483bc1c1de82689497eaf27a930896e7733f418acf9eca',
-      prv: '35a98c286af2c36283adcd47e92cc221b1cd24eeded852a21abfbbc65e98eb8a',
+      prv: '35a98c286af2c36283adcd47e92cc221b1cd24eeded852a21abfbbc65e98eb8ab63de5ad57cc75b256483bc1c1de82689497eaf27a930896e7733f418acf9eca',
     };
```
zahin-mohammad, line 323: Instead of changing these, would it make sense to create a new keypair in the new format? That way we still have tests for the old format.

modules/sdk-coin-xlm/src/lib/utils.ts
```diff
@@ -1,22 +1,48 @@
...
+ * Private key can be either a raw root private key (hex string) or a Stellar private key (prefixed with 'S').
+ * @param secret Either hex root private key or Stellar private key
+ * @returns Stellar keypair from the provided prv
+ */
+export function createStellarKeypairFromPrv(prv: string): stellar.Keypair {
+  if (prv.startsWith('S')) {
+    return stellar.Keypair.fromSecret(prv);
+  }
+  return stellar.Keypair.fromSecret(encodePrivateKey(Buffer.from(prv.slice(0, 64), 'hex')));
+}
```
zahin-mohammad, line 28: Lets add tests for these utils.

modules/sdk-coin-xlm/src/lib/utils.ts
```diff
@@ -1,22 +1,48 @@
...
+
 /**
  * Validates a Stellar public key
  * Stellar public keys are prefixed with 'G'
  * @param pub A Stellar public key to validate
  * @returns Whether the input is a valid Stellar public key
  */
-export function isValidStellarPub(pub: string): boolean {
+export function isValidStellarPublicKey(pub: string): boolean {
```
zahin-mohammad, line 36: This is a breaking change. Keep the existing function, and mark it as `@deprecated` in the jsdocs, and create a new function for what you need. In the jsdocs mention the new function clients should use. 

This applies to all the other function name changes.

zahin-mohammad (changes requested):

modules/sdk-coin-xlm/src/lib/utils.ts
dwei-bitgo: There are some tests for these, they are in the `utils.ts` test file at the bottom

modules/sdk-coin-xlm/test/unit/xlm.ts
dwei-bitgo: Hmm, I changed these as I realized that root prvs are represented in `prv + pub` format, so I think that the original format of these test keychains actually would not occur

modules/sdk-coin-xlm/src/lib/utils.ts
dwei-bitgo: Updated that

modules/sdk-coin-xlm/test/unit/xlm.ts
zahin-mohammad: It could be that we used to support this format but then changed it later, which would mean some old wallets would have that old format. I am not actually sure, but if the tests passed with the old format, I'd expect that it should still work? 

Are the old formats still valid for the tests they are used for or do they error out?

modules/sdk-coin-xlm/test/unit/utils.ts
```diff
@@ -91,8 +91,46 @@ describe('Stellar Utils', () => {
...
       const decodedBackupPrv = Utils.decodePrivateKey(backupKeychain.prv).toString('hex');
 
-      decodedPrv.should.equal(rootKeychain.prv);
-      decodedBackupPrv.should.equal(backupRootKeychain.prv);
+      decodedPrv.should.equal(rootKeychain.prv.slice(0, 64));
+      decodedBackupPrv.should.equal(backupRootKeychain.prv.slice(0, 64));
+    });
+  });
+
+  describe('create Stellar keypairs', () => {
```
zahin-mohammad, line 99: Can you add non-happy path tests as well? What errors do we expect for invalid input, what if input is empty, etc.

zahin-mohammad (dismissed):
lgtm. You can add the extra tests either in this PR or another.

modules/sdk-coin-xlm/test/unit/xlm.ts
dwei-bitgo: So these tests and this test data I added in the initial root key PR for XLM here: https://github.com/BitGo/BitGoJS/pull/4638

Before this, no XLM wallet was able to be created using root keys. I reason I updated these keys was because I realized that root keys generated in the OVC, specifically the private key part, are given in an encrypted version of the `prv + pub` 64-byte format, rather than that old 32-byte `prv` only one.

The old format does still pass the tests, since in the code the place they are used is calling a `slice(0, 64)` to extract the first 32-bytes of the hex string anyway.

modules/sdk-coin-xlm/test/unit/utils.ts
dwei-bitgo: Added these tests into this PR. Also refactored these methods to have better error handling

margueriteblair (approved):

PR #4707, opened 2024-07-08
feat(sdk-coin-sol): add sol close ATA recovery support along with tok…

DinshawKothari:
…en recovery

Ticket: WIN-2648

### DIFF
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -149,6 +149,8 @@ export interface SolRecoveryOptions extends MPCRecoveryOptions {
   };
   tokenContractAddress?: string;
   closeAtaAddress?: string;
+  // destination address where token should be sent before closing the ATA address
+  recoveryDestinationAtaAddress?: string;
 }
 
 export interface SolConsolidationRecoveryOptions extends MPCConsolidationRecoveryOptions {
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-sol/src/sol.ts
```diff
@@ -1043,15 +1041,63 @@ export class Sol extends BaseCoin {
...
+      txBuilder = factory
+        .getTokenTransferBuilder()
+        .nonce(blockhash)
+        .sender(bs58EncodedPublicKey)
+        .fee({ amount: feePerSignature })
+        .associatedTokenAccountRent(rentExemptAmount.toString())
+        .feePayer(bs58EncodedPublicKey);
+
+      const network = this.getNetwork();
+      const token = getSolTokenFromAddress(tokenInfo.info.mint, network);
```
mukeshsp, line 1077: what if there are multiple tokens ?

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -1043,15 +1041,63 @@ export class Sol extends BaseCoin {
...
+    const tokenInfo = await this.getTokenAccountInfo(params.closeAtaAddress);
+    const tokenBalance = Number(tokenInfo.info.tokenAmount.amount);
+
+    if (tokenBalance > 0) {
+      // closeATA address has some token balance, it needs to be withdrawn before closing ATA
+      console.log(
+        `closeATA address ${params.closeAtaAddress} has token balance ${tokenBalance}, it needs to be withdrawn before closing ATA address`
+      );
+
+      if (!params.recoveryDestinationAtaAddress || !this.isValidAddress(params.recoveryDestinationAtaAddress)) {
```
mukeshsp, line 1062: FYI : recoveryDestinationAtaAddress should be belong to same token account

modules/sdk-coin-sol/src/sol.ts
DinshawKothari: There won't be multiple tokens, we are not dealing with receive address here, but we are only interested in closeATA address, and a ATA address will always belong to just one token.

modules/sdk-coin-sol/src/sol.ts
DinshawKothari: Yes correct, we will provide this detail in coin sandbox script clearly.

mukeshsp (approved):

PR #4711, opened 2024-07-09
fix(key-card): broken keycard img scaling based on fixed size

mtexeira-simtlix:
TICKET: WP-2122

KeyCardImage rendered on pdf (the one with the coin logo) was using as dimensions 303 x 40 for every coin but some images comes with different sizes so the solution is to re-scalate based on the available size and checking if one of the dimensions still exceeds after the dimensions change.


mtexeira-simtlix:
Added a DO NOT MERGE label because i'm doing an alpha release first in order to test the changes with multiple wallets.

zahin-mohammad (approved):
lgtm

alebusse (approved):

PR #4716, opened 2024-07-10
feat(sdk-api): add adata param to encryption methods

alebusse:
added adata param to encryption methods

WP-2177

TICKET: WP-2177

### DIFF
--- a/modules/sdk-api/test/unit/encrypt.ts
+++ b/modules/sdk-api/test/unit/encrypt.ts
@@ -1,0 +1,71 @@
+import * as assert from 'assert';
+import { randomBytes } from 'crypto';
+
+import { decrypt, encrypt } from '../../src';
+
+describe('encryption methods tests', () => {
+  describe('encrypt', () => {
+    it('encrypts the plaintext with the given password', () => {
+      const password = 'myPassword';
+      const plaintext = 'Hello, World!';
+      const ciphertext = encrypt(password, plaintext);
+
+      assert(ciphertext !== plaintext, 'ciphertext should not be equal to plaintext');
+    });
+
+    it('returns a different ciphertext for the same plaintext and password', () => {
+      const password = 'myPassword';
+      const plaintext = 'Hello, World!';
+      const ciphertext1 = encrypt(password, plaintext);
+      const ciphertext2 = encrypt(password, plaintext);
+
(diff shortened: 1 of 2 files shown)


modules/sdk-api/test/unit/encrypt.ts
```diff
@@ -0,0 +1,71 @@
...
+    it('returns a different ciphertext for the same plaintext and password', () => {
+      const password = 'myPassword';
+      const plaintext = 'Hello, World!';
+      const ciphertext1 = encrypt(password, plaintext);
+      const ciphertext2 = encrypt(password, plaintext);
+
+      assert(ciphertext1 !== ciphertext2, 'ciphertexts should not be equal');
+    });
+
+    it('throws an error if the salt length is not 8 bytes', () => {
```
margueriteblair, line 25: Do we have docs somewhere that says it needs to be 8 bytes?

modules/sdk-api/test/unit/encrypt.ts
```diff
@@ -0,0 +1,71 @@
...
+      const ciphertext = encrypt(password, plaintext);
+      const wrongPassword = 'wrongPassword';
+
+      assert.throws(() => decrypt(wrongPassword, ciphertext), 'sjcl exception: cc: invalid aes key');
+    });
+
+    it('decrypts the ciphertext with the given password and adata', () => {
+      const password = 'myPassword';
+      const plaintext = 'Hello, World!';
+      const adata = 'additional data';
```
margueriteblair, line 64: What exactly does the adata do here? It's just additional plaintext data that doesn't get encrypted?

margueriteblair (approved):
changes seem fine, just had some q's

modules/sdk-api/test/unit/encrypt.ts
alebusse: Its a requirement we added, nothing new in my PR, was already there.

modules/sdk-api/test/unit/encrypt.ts
alebusse: its like an extra value that its required to decrypt the ciphertext, its visible but you could keep it secret same as the password. Could be a value that its generated from something else like a walletId, so if you try to decrypt with the right password but the wrong walletId wont work

islamaminBitGo (approved):

PR #4719, opened 2024-07-12
feat: remove bounceable address as default for ton

ajays97:
Ticket: WIN-3114

We were by default converting recipient addresses to bounceable format (EQ) for ton transactions. This might result in bounce backs if the receiver is an uninitialized wallet. Since with Ton v1 wallets we are switching to UQ address formats (non bouncable), our transactions should also not enforce bounceable format.

### DIFF
--- a/modules/sdk-coin-ton/src/lib/transaction.ts
+++ b/modules/sdk-coin-ton/src/lib/transaction.ts
@@ -17,6 +17,7 @@ const WALLET_ID = 698983191;
 
 export class Transaction extends BaseTransaction {
   public recipient: Recipient;
+  public bounceable: boolean;
   public message: string;
   seqno: number;
   expireTime: number;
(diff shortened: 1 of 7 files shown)


shivpippal (dismissed):

modules/sdk-coin-ton/src/lib/transaction.ts
```diff
@@ -101,12 +105,7 @@ export class Transaction extends BaseTransaction {
       }
     }
 
-    const orderHeader = TonWeb.Contract.createInternalMessageHeader(
-      new TonWeb.Address(address),
-      new BN(amount),
-      true,
-      false
-    );
+    const orderHeader = TonWeb.Contract.createInternalMessageHeader(new TonWeb.Address(address), new BN(amount));
```
ewangbitgo, line 108: can we make bouncable a input parameter for builder because for staking transaction, we need to set bounceable to false.

ewangbitgo (approved):

PR #4736, opened 2024-07-17
feat: add bounce versions of to and from address

ajays97:
Ticket: WIN-3114

### DIFF
--- a/modules/sdk-coin-ton/src/lib/transaction.ts
+++ b/modules/sdk-coin-ton/src/lib/transaction.ts
@@ -18,6 +18,8 @@ const WALLET_ID = 698983191;
 export class Transaction extends BaseTransaction {
   public recipient: Recipient;
   public bounceable: boolean;
+  public fromAddressBounceable: boolean;
+  public toAddressBounceable: boolean;
   public message: string;
   seqno: number;
   expireTime: number;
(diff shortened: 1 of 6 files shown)


modules/sdk-coin-ton/src/lib/transaction.ts
```diff
@@ -28,7 +30,9 @@ export class Transaction extends BaseTransaction {
 
   constructor(coinConfig: Readonly<CoinConfig>) {
     super(coinConfig);
-    this.bounceable = false;
+    this.bounceable = true;
```
margueriteblair, line 33: will there be instances of types of transactions where bouncable is false?

ajays97: Yes, we want to be able to build transactions with bounceable as false for our new v1 wallets that we are planning to roll out for TON. For older wallets, bounceable remains true by default, and for newer wallets, we will be setting this flag as needed.

margueriteblair (dismissed):

modules/sdk-coin-ton/src/ton.ts
```diff
@@ -205,6 +221,21 @@ export class Ton extends BaseCoin {
     try {
       const factory = new TransactionBuilderFactory(coins.get(this.getChain()));
       const transactionBuilder = factory.from(Buffer.from(params.txHex, 'hex').toString('base64'));
+
+      const { bounceable, toAddressBounceable, fromAddressBounceable } = params;
+
+      if (typeof bounceable === 'boolean') {
```
sreerajs, line 227: I guess this is not needed as bounce flag would be decoded from hex

sreerajs (approved):
Lgtm

PR #4737, opened 2024-07-17
feat(sdk-core): create lightning wallet document

saravanan7mani:
lightning wallet creation is a multi stage process. creating wallet document is stage one

Ticket: BTC-1341


modules/sdk-core/src/bitgo/keychain/iKeychains.ts
```diff
@@ -105,6 +113,7 @@ export interface AddKeychainOptions {
   isDistributedCustody?: boolean;
   // indicates if the key is MPCv2 or not
   isMPCv2?: boolean;
+  coinSpecific?: { [coinName: string]: AddLightningKeychainCoinSpecific };
```
OttoAllmendinger, line 116: IMO this is accurate - the caller should check the type

otherwise this will end up a a supertype with all coins adding their CoinSpecific
 
```suggestion
  coinSpecific?: { [coinName: string]: unknown };
```

OttoAllmendinger (changes requested):

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -168,6 +169,65 @@ export class Wallets implements IWallets {
...
+
+    const lightningParams = _.pick(params, ['label', 'passphrase', 'enterprise', 'passcodeEncryptionCode']);
+    const { label, passphrase, enterprise, passcodeEncryptionCode } = lightningParams;
+
+    assert(label && typeof label === 'string', 'label parameter is required and must be a string');
+    assert(passphrase && typeof passphrase === 'string', 'passphrase parameter is required and must be a string');
+    assert(enterprise && typeof enterprise === 'string', 'enterprise parameter is required and must be a string');
+    assert(
+      passcodeEncryptionCode && typeof passcodeEncryptionCode === 'string',
+      'passcodeEncryptionCode parameter is required and must be a string'
```
zahin-mohammad, line 184: Why do we need all these type assertions still? 

AFAIK a lot of the existing assertions like this where just carry over from the JS sdk.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -168,6 +169,65 @@ export class Wallets implements IWallets {
...
+      nodeAuthKeychain: keychainPromises[2](),
+    });
+
+    const walletParams: SupplementGenerateWalletOptions = {
+      label,
+      m: 1,
+      n: 1,
+      type: 'hot',
+      enterprise,
+      keys: [userKeychain.id],
```
zahin-mohammad, line 215: For lightning, what is the userKeychain used for?

zahin-mohammad: Also have you run this function against staging/test? 

Im surprised wallet platform works with just 1 keyId as input instead of 3.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -168,6 +169,65 @@ export class Wallets implements IWallets {
...
+    const newWallet = await this.bitgo.post(this.baseCoin.url('/wallet')).send(walletParams).result();
+    const wallet = new Wallet(this.bitgo, this.baseCoin, newWallet);
+    const result: LightningWalletWithKeychains = {
+      wallet,
+      userKeychain,
+      userAuthKeychain,
+      nodeAuthKeychain,
+    };
+
+    return result;
```
zahin-mohammad, line 228: Nit: having top level discriminator for the response will make this easy to differentiate the `generateWallet` responses if we have more types we need to add to the union.

modules/sdk-core/src/bitgo/wallet/wallets.ts
saravanan7mani: Lightning wallet creation is primarily will be done by Express users.

`generateWallet` is also called from BitGo Express which does not pass do any type assertion from its HTTP request. 

So TS type safety is unreliable in that case. Thats why run time assertion is done.

modules/sdk-core/src/bitgo/wallet/wallets.ts
saravanan7mani: We recently added support for single user key wallet support for lightning in WP.

saravanan7mani: > Also have you run this function against staging/test?
yes, worked

modules/sdk-core/src/bitgo/wallet/wallets.ts
zahin-mohammad: If the intention is to add type safety for express users, then create a codec for the request that can be used by the express route. 

The type extracted from the codec can be used downstream. Having this parsing is redundant downstream in the sdk, it only really makes sense upstream in the express controllers.

saravanan7mani: Makes sense. Done.

modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -1,9 +1,18 @@
...
 import { Wallet } from './wallet';
 
 export interface WalletWithKeychains extends KeychainsTriplet {
+  walletType: 'WalletWithKeychains';
+  wallet: IWallet;
+  warning?: string;
+}
+
+export interface LightningWalletWithKeychains extends LightningKeychainsTriplet {
+  walletType: 'LightningWalletWithKeychains';
```
OttoAllmendinger, line 15: I don't think we have to add this discriminator

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -168,6 +180,58 @@ export class Wallets implements IWallets {
...
+      coinSpecific: { [this.baseCoin.getChain()]: { keys: [userAuthKeychain.id, nodeAuthKeychain.id] } },
+    };
+
+    const newWallet = await this.bitgo.post(this.baseCoin.url('/wallet')).send(walletParams).result();
+    const wallet = new Wallet(this.bitgo, this.baseCoin, newWallet);
+    return {
+      wallet,
+      userKeychain,
+      userAuthKeychain,
+      nodeAuthKeychain,
```
OttoAllmendinger, line 230:
```suggestion
      userKeychain,
      userAuthKeychain,
      nodeAuthKeychain,
```

these only exist for the LN case so I don't think the `walletType` is required

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #4749, opened 2024-07-20
Update noble_ecc.ts

Ebenezer000:
Proposed changes fix errors in brandonblack package imports


OttoAllmendinger:
where do you see the errors?

PR #4757, opened 2024-07-24
feat(sdk-core): list wallet accepting enterprise id

pengyuc-bitgo:
Allow setting enterprise id when listing wallets.
If the enterprise allows users to see other wallets, those wallets shall be included in the result.

Add the same parameter to the enterprise coinWallets to support pagination and to be consistent with usage.

Ticket: WP-2295

### DIFF
--- a/modules/sdk-core/src/bitgo/enterprise/enterprise.ts
+++ b/modules/sdk-core/src/bitgo/enterprise/enterprise.ts
@@ -6,7 +6,7 @@ import { IBaseCoin } from '../baseCoin';
 import { BitGoBase } from '../bitgoBase';
 import { EnterpriseData, EnterpriseFeatureFlag, IEnterprise } from '../enterprise';
 import { getFirstPendingTransaction } from '../internal';
-import { Wallet } from '../wallet';
+import { ListWalletOptions, Wallet } from '../wallet';
 import { BitGoProofSignatures, EcdsaUtils, SerializedNtildeWithVerifiers } from '../utils/tss/ecdsa';
 import { EcdsaTypes } from '@bitgo/sdk-lib-mpc';
 import { verifyEcdhSignature } from '../ecdh';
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -65,40 +65,11 @@ export class Wallets implements IWallets {
...
-      }
-      queryObject.limit = params.limit;
-    }
-
-    if (params.allTokens) {
-      if (!_.isBoolean(params.allTokens)) {
-        throw new Error('invalid allTokens argument, expecting boolean');
-      }
-      queryObject.allTokens = params.allTokens;
-    }
```
pengyuc-bitgo, line 99: Are these type checks still necessary now that we have typed? 
I think they are probably from the old codes before we use typescripts, right?

mohammadalfaiyazbitgo: If the list route is now typed then yeah, this should be unnecessary 🧹 Though worth a check to see if the current parsing matches what's happening here.

mohammadalfaiyazbitgo (dismissed):
LGTM, just  one point to verify

alebusse (dismissed):

modules/sdk-core/src/bitgo/enterprise/enterprise.ts
```diff
@@ -57,12 +57,18 @@ export class Enterprise implements IEnterprise {
...
+  async coinWallets(params: ListWalletOptions = {}): Promise<Wallet[]> {
+    if (params.skip && params.prevId) {
+      throw new Error('cannot specify both skip and prevId');
+    }
+
+    const walletData = (await this.bitgo
+      .get(this.baseCoin.url('/wallet/enterprise/' + this.id))
+      .query(params)
+      .result()) as any;
+    return walletData.wallets.map((w) => {
```
pengyuc-bitgo, line 69: To avoid changing the function return type and cause a breaking change. I changed the return to just a list of wallets.

alebusse (approved):

PR #4759, opened 2024-07-24
fix(sdk-coin-sui): return all input objects in `toJson` result

hitansh-madan:
Ticket: WIN-3242

- Adds `inputObjects[]` to TxData for sui transactions. `inputObjects[]` is a list of Sui token objects used in building a token transaction. The native sui objects used as gas payment are still returned as part of `gasData`
- Fixes `toJson()` result when number of native sui objects is greater than 255 (only supported for native transfers & staking) . In this case the we parse the objects from the transaction and append it to `gasData.payment[]`


ajays97 (approved):
LGTM!

PR #4764, opened 2024-07-24
fix(sdk-lib-mpc): lazy import dkls wasm

YPAzevedo:
Fixes: https://bitgoinc.atlassian.net/browse/WP-2338

### DIFF
--- a/modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
+++ b/modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
@@ -1,8 +1,14 @@
-import { KeygenSession, Keyshare, Message } from '@silencelaboratories/dkls-wasm-ll-node';
-import { DeserializedBroadcastMessage, DeserializedMessages, DkgState, ReducedKeyShare, RetrofitData } from './types';
+import type { KeygenSession, Keyshare, Message } from '@silencelaboratories/dkls-wasm-ll-node';
 import { decode, encode } from 'cbor-x';
-import { bigIntToBufferBE } from '../../util';
 import { Secp256k1Curve } from '../../curves';
+import { bigIntToBufferBE } from '../../util';
+import { DeserializedBroadcastMessage, DeserializedMessages, DkgState, ReducedKeyShare, RetrofitData } from './types';
+
+type NodeWasmer = typeof import('@silencelaboratories/dkls-wasm-ll-node');
+type WebWasmer = typeof import('@silencelaboratories/dkls-wasm-ll-web');
+type BundlerWasmer = typeof import('@silencelaboratories/dkls-wasm-ll-bundler');
+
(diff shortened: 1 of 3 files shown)


modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
```diff
@@ -111,9 +135,10 @@ export class Dkg {
     }
     if (this.dklsKeyShareRetrofitObject) {
       this.dkgSession = this.seed
-        ? KeygenSession.initKeyRotation(this.dklsKeyShareRetrofitObject, new Uint8Array(this.seed))
-        : KeygenSession.initKeyRotation(this.dklsKeyShareRetrofitObject);
+        ? this.getDklsWasm().KeygenSession.initKeyRotation(this.dklsKeyShareRetrofitObject, new Uint8Array(this.seed))
+        : this.getDklsWasm().KeygenSession.initKeyRotation(this.dklsKeyShareRetrofitObject);
     } else {
+      const { KeygenSession } = this.getDklsWasm();
```
islamaminBitGo, line 141: can we put this before the if/else block so it can be used within both if and else

YPAzevedo:
tested in the bg23 app, created a wallet, and verified my changes were present. tested on alpha release

islamaminBitGo (dismissed):

alebusse (approved):

islamaminBitGo (approved):

zahin-mohammad (approved):

bitgoAaron (commented):
did you need to make any changes to webpack config to get this working? If so, please update in `webpack/bitgojs.config.js`.

Some documentation or tests would be ideal for a change like this, so any other potential users of this package will have some guidance on how to properly integrate in their web applications as well.

YPAzevedo:
> did you need to make any changes to webpack config to get this working? If so, please update in `webpack/bitgojs.config.js`.
> 
> Some documentation or tests would be ideal for a change like this, so any other potential users of this package will have some guidance on how to properly integrate in their web applications as well.

No bundler changes on BitGoJs - Will create a follow up ticket for any instructions necessary, but im not sure users will be using this DKG API directly -- what do you think @islamaminBitGo

islamaminBitGo:
> > did you need to make any changes to webpack config to get this working? If so, please update in `webpack/bitgojs.config.js`.
> > Some documentation or tests would be ideal for a change like this, so any other potential users of this package will have some guidance on how to properly integrate in their web applications as well.
> 
> No bundler changes on BitGoJs - Will create a follow up ticket for any instructions necessary, but im not sure users will be using this DKG API directly -- what do you think @islamaminBitGo

yes, this DKG API is not intended for external customers, however i do agree since we've added this dependency injection capability it's probably worth it we document how to use it for internal stakeholders. Let's create a ticket and do this as a fast follow-up.

PR #4765, opened 2024-07-25
fix(sdk-core): support backup party id in mpcv2 signing utils

islamaminBitGo:
TICKET: HSM-442


mohammadalfaiyazbitgo (commented):
any unit tests you can add for this?

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -17,7 +17,8 @@ import assert from 'assert';
 
 export async function getSignatureShareRoundOne(
   round1Message: DklsTypes.DeserializedBroadcastMessage,
-  userGpgKey: openpgp.SerializedKeyPair<string>
+  userGpgKey: openpgp.SerializedKeyPair<string>,
+  partyId: 0 | 1 = 0
```
mohammadalfaiyazbitgo, line 21: any reason you're using `0 | 1` instead of `boolean` ?

islamaminBitGo:
> any unit tests you can add for this?

there is no value of any additional tests in my opinion, the user signing is tested so adding a backup signing test would be a clone of that

pranavjain97 (dismissed):
Changes look good. Can you add a test with the backup id case here `tssUtils/ecdsaMPCv2/signTxRequest.ts`?

pranavjain97:
> > any unit tests you can add for this?
> 
> there is no value of any additional tests in my opinion, the user signing is tested so adding a backup signing test would be a clone of that

It's more so that nobody removes that logic from this code if there is test for it

alebusse:
do we want to change the top method signRequestBase too https://github.com/BitGo/BitGoJS/blob/a481acddd572873aa9d3303f97832dbe88289de8/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts#L698?

islamaminBitGo:
> do we want to change the top method signRequestBase too
> 
> https://github.com/BitGo/BitGoJS/blob/a481acddd572873aa9d3303f97832dbe88289de8/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts#L698
> 
> ?

i feel this might be out of scope for this PR since this is meant only for the utils that other apps are using rather than the hot wallet regular flow itself. Nevertheless, yes, I think we can do this but in a follow up ticket + PR

islamaminBitGo:
actuall

> > do we want to change the top method signRequestBase too
> > https://github.com/BitGo/BitGoJS/blob/a481acddd572873aa9d3303f97832dbe88289de8/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts#L698
> > 
> > ?
> 
> i feel this might be out of scope for this PR since this is meant only for the utils that other apps could be using (such as WRW) rather than the hot wallet regular flow itself. Nevertheless, yes, I think we can do this but in a follow up ticket + PR

actually on second thought, it's not a big change so i think it's ok to include it in this ticket. will apply this change as well

islamaminBitGo:
> Changes look good. Can you add a test with the backup id case here `tssUtils/ecdsaMPCv2/signTxRequest.ts`?

done

islamaminBitGo:
> do we want to change the top method signRequestBase too
> 
> https://github.com/BitGo/BitGoJS/blob/a481acddd572873aa9d3303f97832dbe88289de8/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts#L698
> 
> ?

done

mohammadalfaiyazbitgo (approved):

PR #4771, opened 2024-08-01
chore: fix token name to eliminate confusion in the UI

sandra-wang-bitgo:
TICKET: CE-4303

When a user creates an access token, they can choose to set a spending limit on the asset token. In the coin dropdown, USDT is displayed with the name only. Note that there are two USDT options here but representing USDT and eos:USDT. This can be confusing for users. The same case applies to `algo:usdt`. 

The fix in this PR is to update the name to accurately reflect the actual asset token.
Screenshot 2024-07-30 at 3 14 48 PM

image


modules/statics/src/coins.ts
```diff
@@ -13301,7 +13301,7 @@ export const coins = CoinMap.fromCoins([
     'dd48a295-4f59-4a36-bc40-801998b9ff90',
     'talgo:USDt-180447',
     undefined,
-    'USDt',
+    'Algorand USDt',
```
aarvay, line 13304:
```suggestion
    'Algorand USDT',
```

aarvay (commented):
Should we also prefix "Test" for testnet tokens?

modules/statics/src/coins.ts
sandra-wang-bitgo: ah yes, you beat to it

aarvay (approved):

alebusse (approved):

bitgoAaron (approved):

CapnMigraine (approved):

PR #4775, opened 2024-08-05
chore(examples): generate unsigned hex for fee estimate

MohammedRyaan786:
TICKET: COIN-1379

Generate unsigned hex which will be used
in api endpoint for fee estimate

### DIFF
--- a/examples/ts/sui/get-transaction-hex.ts
+++ b/examples/ts/sui/get-transaction-hex.ts
@@ -21,29 +21,32 @@ const accessToken = '';
 // TODO: set a walletId
 const walletId = '';
 
-// TODO: set your passphrase for your wallet here
-const passphrase = '';
-
 async function buildPayTx() {
-  bitgo.authenticateWithAccessToken({ accessToken });
-  const newWallet = await bitgo.coin(coin).wallets().getWallet({ id: walletId });
-  const recipients = [
-    {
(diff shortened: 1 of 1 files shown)


examples/ts/sui/get-transaction-hex.ts
```diff
@@ -21,29 +21,33 @@ const accessToken = '';
...
+      {
+        address: 'tsui-address', // TODO: set a tsui receive address
+        amount: '10000', // TODO: set an amount for the transaction
+      },
+    ];
+
+    const transaction = await wallet.prebuildTransaction({
+      recipients,
+      type: 'transfer',
+    });
```
alebusse, line 38: without the preview flag, its a txRequest creation
```suggestion
    const transaction = await wallet.prebuildTransaction({
      recipients,
      type: 'transfer',
      preview: true,
    });
```

alebusse: maybe it needs to be a different example file, this one seems to be used to convert from base64 to hex

MohammedRyaan786: Hi @alebusse, this example file was created by me, it is used to generate a hex which is required to be passed in the api for fee estimate for sui. The conversion is done as prebuildTransaction returns tx in a format which sui doesn't expect in its sui_dryRunTransactionBlock endpoint

alebusse (approved):

PR #4779, opened 2024-08-06
fix(sdk-lib-mpc): fix time warp issues with pgp verifications

islamaminBitGo:
TICKET: HSM-448


alebusse (approved):

zahin-mohammad:
> Can you write some testcases as followup which will fail if this change was not made, Approving since we need this to unblock

It's kind of hard to spoof; I guess we can create a signature using a comp with its clock set to like 100 years from now?

zahin-mohammad (approved):

islamaminBitGo:
> > Can you write some testcases as followup which will fail if this change was not made, Approving since we need this to unblock
> 
> It's kind of hard to spoof; I guess we can create a signature using a comp with its clock set to like 100 years from now?

i think (have to check) u can actually set the time for a signature to "Infinity"

PR #4783, opened 2024-08-06
feat(express): lightning wallet creation step two

saravanan7mani:
initialize signer lnd and update platform.

Ticket: BTC-1356


modules/express/src/clientRoutes.ts
```diff
@@ -1658,3 +1659,12 @@ export function setupSigningRoutes(app: express.Application, config: Config): vo
     promiseWrapper(handleV2OFCSignPayloadInExtSigningMode)
   );
 }
+
+export function setupLightningRoutes(app: express.Application, config: Config): void {
+  app.post(
+    '/api/v2/:coin/initlightningwallet/:id',
```
lcovar, line 1665: :nit: personally thing this is a bit cleaner of a route. On the handler we can check if it's a lightning coin, if it's not we can throw a not supported error. 

```suggestion
    '/api/v2/:coin/:walletid/init',
```

PR #4796, opened 2024-08-08
fix(sdk-core): update ADATA to include the derivationPath

alebusse:
updated ADATA to include the derivation path

WP-2386

TICKET: WP-2386


islamaminBitGo (approved):
LGTM

PR #4800, opened 2024-08-08
feat(sdk-core): add create and verify message signature 

lzjluzijie:
Ticket: BTC-1401


modules/sdk-core/src/bitgo/lightning/signature.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @param {Signable} message - The message to sign.
+ * @param {string} prv - The private key in BIP32 format.
+ * @param {utxolib.Network} network - The network to use for signing.
+ * @returns {string} - Returns the signature in hexadecimal format.
+ */
+export function createMessageSignature(
+  message: Signable,
+  prv: string,
+  network: utxolib.Network = utxolib.networks.bitcoin
```
lzjluzijie, line 37: should we use testnet?

saravanan7mani (approved):

PR #4804, opened 2024-08-08
fix(root): bump public-types and fix SMC MPCv2 format

alebusse:
Bumped public-types lib and fixed SMC MPCv2 response formats to remove the wallet wrapper

WP-2468

TICKET: WP-2468

### DIFF
--- a/modules/bitgo/package.json
+++ b/modules/bitgo/package.json
@@ -116,7 +116,7 @@
     "superagent": "^9.0.1"
   },
   "devDependencies": {
-    "@bitgo/public-types": "2.33.3",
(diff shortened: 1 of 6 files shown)


sachushaji (approved):
Chnages LGTM, but would this break some things in WP?, I would suggest to merge this only when you have the WP changes also in place if it break things, otherwise it will block sdk bumps

alebusse:
> Chnages LGTM, but would this break some things in WP?, I would suggest to merge this only when you have the WP changes also in place if it break things, otherwise it will block sdk bumps

This wont affect WP.

zahin-mohammad (approved):

PR #4809, opened 2024-08-09
fix(sdk-core): add back state checks for MPCv2 SMC utils

alebusse:
added back state checks for MPCv2 SMC utils

WP-2471

TICKET: WP-2471


zahin-mohammad (dismissed):
lgtm, if we can add a small test that would be nice.

margueriteblair (dismissed):

pranavjain97 (approved):

PR #4834, opened 2024-08-19
feat(statics): add separate coinfeature for tss stuck txn management

dpkjnr:
Ticket: WP-2155


saravanan7mani (approved):

modules/statics/src/base.ts
```diff
@@ -275,9 +275,14 @@ export enum CoinFeature {
   MPCV2 = 'mpcv2',
 
   /**
-   * This coin supports acceleration or nonce filling txn for stuck transactions
+   * This coin supports acceleration or nonce filling txn for stuck transactions for tss wallet
    */
-  STUCK_TRANSACTION_MANAGEMENT = 'stuck-transaction-management',
```
zahin-mohammad, line 280: Instead of removing this value I think it would be better to: 
- mark it as deprecated
- update usage internally to check either `STUCK_TRANSACTION_MANAGEMENT` or `STUCK_TRANSACTION_MANAGEMENT_ONCHAIN`


In later releases:
- remove `STUCK_TRANSACTION_MANAGEMENT`


this way its not a breaking change, at least not immediately.

dpkjnr: I added this feature last week and hasn't been used anywhere yet

sachushaji (approved):

PR #4838, opened 2024-08-20
feat(sdk-coin-sui): create broadcastable MPC transaction

bhavidhingra:
TICKET: [WIN-3293](https://bitgoinc.atlassian.net/browse/WIN-3293)

[WIN-3293]: https://bitgoinc.atlassian.net/browse/WIN-3293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
@@ -3,7 +3,7 @@ import { IRequestTracer } from '../../../api';
 import { KeychainsTriplet, ParsedTransaction } from '../../baseCoin';
 import { ApiKeyShare, Keychain } from '../../keychain';
 import { ApiVersion, Memo, WalletType } from '../../wallet';
-import { EDDSA, GShare, SignShare, Signature } from '../../../account-lib/mpc/tss';
+import { EDDSA, GShare, Signature, SignShare } from '../../../account-lib/mpc/tss';
 import { KeyShare } from './ecdsa';
 import { EcdsaTypes } from '@bitgo/sdk-lib-mpc';
 import { TssEcdsaStep1ReturnMessage, TssEcdsaStep2ReturnMessage, TxRequestChallengeResponse } from '../../tss/types';
@@ -445,6 +445,10 @@ export interface MPCTx {
     commonKeychain?: string;
     lastScanIndex?: number;
   };
+  // the amount recovered using WRW
+  recoveryAmount?: string;
+  // the transaction signature used for broadcasting the transaction
+  signature?: string;
 }
 
 export interface MPCRecoveryOptions {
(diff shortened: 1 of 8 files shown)


.gitcommitscopes
```diff
@@ -0,0 +1 @@
+sdk-coin-sui
```
zahin-mohammad, line 1: curious on what this file does

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -455,6 +457,8 @@ export interface MPCRecoveryOptions {
   walletPassphrase?: string;
   seed?: string;
   index?: number;
+  startingScanIndex?: number;
+  scan?: number;
```
zahin-mohammad, line 461: Please add jsdocs for these three fields. How is index different from scan?

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -445,6 +445,8 @@ export interface MPCTx {
     commonKeychain?: string;
     lastScanIndex?: number;
   };
+  recoveryAmount?: string;
+  signature?: string;
```
zahin-mohammad, line 449: jsdoc herw would also be useful

.gitcommitscopes
bhavidhingra: remove it

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
bhavidhingra: `index` and `startingScanIndex` both serve the same purpose, some WRW flows used `index` while others use `startingScanIndex`.

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
bhavidhingra: done

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -445,6 +445,10 @@ export interface MPCTx {
     commonKeychain?: string;
     lastScanIndex?: number;
   };
+  // the amount recovered using WRW
+  recoveryAmount?: string;
+  // the transaction signature used for broadcasting the transaction
+  signature?: string;
```
zahin-mohammad, line 451: How is this different from `serializedTx`? Is the idea to deprecate that field? 
If you look at the sol recovery implementation, the `serializedTx` field is used to return the signed tx ready for broadcast

bhavidhingra: To broadcast a Sui transaction, we need both `serializedTx` and `signature`.

zahin-mohammad (approved):

hitansh-madan (approved):

PR #4842, opened 2024-08-21
feat(sdk-core): add encryptedWalletPassphrase to generateWallet method

alebusse:
Added a new field in the generateWallet response that its the wallet passphrase encrypted with the passcode, this data can be used to do recover the original password, same as the box D on a keycard. Its only available for hot wallets and if passphrase and the passcode is provided

WP-2525

TICKET: WP-2525

### DIFF
--- a/modules/bitgo/test/v2/unit/wallets.ts
+++ b/modules/bitgo/test/v2/unit/wallets.ts
@@ -429,9 +429,47 @@ describe('V2 Wallets:', function () {
         .post('/api/v2/tbtc/key', _.matches({ source: 'backup' }))
         .reply(200, { pub: 'backupPub' });
 
-      await wallets.generateWallet(params);
+      const response = await wallets.generateWallet(params);
 
       walletNock.isDone().should.be.true();
+
+      assert.ok(response.encryptedWalletPassphrase === undefined);
+      assert.ok(response.wallet);
+    });
+
+    it('should generate hot onchain wallet', async () => {
+      const params: GenerateWalletOptions = {
+        label: 'test wallet',
+        passphrase: 'multisig password',
+        enterprise: 'enterprise',
+        passcodeEncryptionCode: 'originalPasscodeEncryptionCode',
+      };
+
+      const walletNock = nock(bgUrl)
+        .post('/api/v2/tbtc/wallet', function (body) {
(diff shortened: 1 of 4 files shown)


modules/bitgo/test/v2/unit/wallets.ts
```diff
@@ -490,15 +528,64 @@ describe('V2 Wallets:', function () {
 
       const wallets = new Wallets(bitgo, tsol);
 
-      await wallets.generateWallet({
+      const params = {
         label: 'tss wallet',
         passphrase: 'tss password',
-        multisigType: 'tss',
+        multisigType: 'tss' as any,
```
sachushaji, line 534: Can't we avoid this any type casting?, multisigType should consider tss as valid

sachushaji (dismissed):

modules/bitgo/test/v2/unit/wallets.ts
OttoAllmendinger: pretty sure `as const` will work here

OttoAllmendinger (dismissed):

margueriteblair (commented):
can approve once the casting to `any` comments are addressed

margueriteblair (dismissed):

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -1008,4 +1024,14 @@ export class Wallets implements IWallets {
...
+
+  /**
+   * Returns the encrypted wallet password
+   * @param {string} passphrase - the passphrase to encrypt
+   * @param {string}passcodeEncryptionCode - the passcode encryption code
+   * @returns {string} - the encrypted wallet password
+   */
+  private getEncryptedWalletPassphrase(passphrase: string, passcodeEncryptionCode: string): string {
+    return this.bitgo.encrypt({ input: passphrase, password: passcodeEncryptionCode });
+  }
```
zahin-mohammad, line 1036: Do we really need this? We can just directly call `this.bitgo.encrypt`

zahin-mohammad (dismissed):

zahin-mohammad (approved):

margueriteblair (approved):

PR #4848, opened 2024-08-23
feat(sdk-core): improve wallet.signTransaction to get the user key

alebusse:
In order to allow the signTssTx express endpoint to sign tx without external signer, we need to modify the signTransaction method to get the key if the walletPassphrase is provided

WP-2535

TICKET: WP-2523


margueriteblair (approved):
I've tested that it works to sign a TSS tx request in this new flow in express!

PR #4861, opened 2024-08-28
feat(sdk-coin-sol): Automatic ATA creation during consolidation

baltiyal:
feat: Automatic ATA creation during consolidation

TICKET: [COIN-1521](https://bitgoinc.atlassian.net/browse/COIN-1521)

[COIN-1521]: https://bitgoinc.atlassian.net/browse/COIN-1521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/sdk-coin-sol/src/lib/transferBuilderV2.ts
```diff
@@ -98,7 +131,7 @@ export class TransferBuilderV2 extends TransactionBuilder {
               fromAddress: this._sender,
               toAddress: sendParams.address,
               amount: sendParams.amount,
-              tokenName: sendParams.tokenName,
+              tokenName: coin.name,
```
Vijay-Jagannathan, line 134: nit: why did this change ? seems like both are the same

Vijay-Jagannathan (approved):
lgtm

modules/sdk-coin-sol/src/lib/transferBuilderV2.ts
baltiyal: yes, both are same, in this block (can see the below return) I was using the `tokenName: coin.name`, so referential similar changed it.

PR #4873, opened 2024-09-02
feat(utxo-lib): add convertExtendedKeyNetwork function

saravanan7mani:
To convert extended key from one network to another

Ticket: BTC-1467

### DIFF
--- a/modules/utxo-lib/src/bitgo/keyutil.ts
+++ b/modules/utxo-lib/src/bitgo/keyutil.ts
@@ -1,6 +1,7 @@
 import { ECPairInterface } from 'ecpair';
+import * as bs58check from 'bs58check';
 import { Network } from '../networks';
-import { ECPair } from '../noble_ecc';
+import { bip32, ECPair } from '../noble_ecc';
 
 /**
  * Create an ECPair from the raw private key bytes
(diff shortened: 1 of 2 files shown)


modules/utxo-lib/src/bitgo/keyutil.ts
```diff
@@ -36,3 +37,29 @@ export function privateKeyBufferFromECPair(ecPair: ECPairInterface): Buffer {
...
+  try {
+    const decodedData = bs58check.decode(extendedKey);
+    const hdNode = bip32.fromBase58(extendedKey, fromNetwork);
+    const targetVersionBytes = hdNode.isNeutered() ? targetNetwork.bip32.public : targetNetwork.bip32.private;
+    const versionBuffer = Buffer.alloc(4);
+    versionBuffer.writeUInt32BE(targetVersionBytes, 0);
+    versionBuffer.copy(decodedData, 0, 0, 4);
+    return bs58check.encode(decodedData);
+  } catch (error) {
+    throw new Error(`Failed to convert extended key from ${fromNetwork} to ${targetNetwork}: ${error.message}`);
```
OttoAllmendinger, line 63: networks themselves cannot be stringified - you can use `getNetworkName`

the `catch`  will hide the exception details btw (line number) so consider removing it

OttoAllmendinger (approved):

PR #4874, opened 2024-09-02
feat(sdk-core): add utils for signer lnd init

saravanan7mani:
Create watch only init data.
Add ip caveat to macaroon.
Signer macaroon permissions.
Get utxo lib network for coin name.

Ticket: BTC-1468


modules/sdk-core/package.json
```diff
@@ -71,7 +71,9 @@
     "strip-hex-prefix": "^1.0.0",
     "superagent": "^9.0.1",
     "tweetnacl": "^1.0.3",
-    "uuid": "^8.3.2"
+    "uuid": "^8.3.2",
+    "macaroon": "^3.0.4",
```
sachushaji, line 75: Seems not a very well maintained library, Last publish is 6 years ago.

sachushaji: Any other possible alternatives?

saravanan7mani: Unfortunately no. Other alternative libs also has similar maintenance. This has most downloads. Writing macaroon functions are very low level.

BTW, reason for low maintenance- macaroon tokens are not popular and its like JWT, nothing much to improve.

sachushaji (approved):
package changes lgtm, please get review from btc team for utxo changes

PR #4877, opened 2024-09-03
feat: bump typescript to 4.9

OttoAllmendinger:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-8.html

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html

Issue: BTC-1472

### DIFF
--- a/modules/deser-lib/src/cbor.ts
+++ b/modules/deser-lib/src/cbor.ts
@@ -27,7 +27,7 @@ function getType(value: unknown): string {
     );
   }
   if (typeof value === 'string') {
-    if (value.startsWith('0x')) {
+    if ((value as string).startsWith('0x')) {
       return 'bytes';
     }
     return 'string';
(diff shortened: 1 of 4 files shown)


vincentyangBitGo (approved):
lgtm

modules/deser-lib/src/cbor.ts
```diff
@@ -27,7 +27,7 @@ function getType(value: unknown): string {
     );
   }
   if (typeof value === 'string') {
-    if (value.startsWith('0x')) {
+    if ((value as string).startsWith('0x')) {
```
zahin-mohammad, line 30: the if condition on the line above is no longer a valid type guard?

zahin-mohammad (approved):

ravneet-bitgo (approved):

modules/deser-lib/src/cbor.ts
OttoAllmendinger: It is - but it actually constrains to `never` here.

I believe a type guard higher up is to blame.

PR #4879, opened 2024-09-03
fix(sdk-core): refactoring generate custodial multisig to not need pw

margueriteblair:
For custodial multisig, we've been passing in the password parameter, which is not needed. The flow needs to be updated so that when we don't pass in the password, there's no error thrown.

Ticket: WP-2535

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallets.ts
@@ -12,7 +12,7 @@ import * as common from '../../common';
 import { IBaseCoin, KeychainsTriplet, KeyPair, SupplementGenerateWalletOptions } from '../baseCoin';
 import { BitGoBase } from '../bitgoBase';
 import { getSharedSecret } from '../ecdh';
-import { AddKeychainOptions, Keychain } from '../keychain';
+import { AddKeychainOptions, Keychain, KeyIndices } from '../keychain';
 import { decodeOrElse, promiseProps, RequestTracer } from '../utils';
 import {
   AcceptShareOptions,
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -447,138 +448,158 @@ export class Wallets implements IWallets {
...
+      walletParams.n = undefined;
+      walletParams.m = undefined;
+      walletParams.keys = undefined;
+      walletParams.keySignatures = undefined;
+
+      const newWallet = await this.bitgo.post(this.baseCoin.url('/wallet')).send(walletParams).result();
+
+      const result: WalletWithKeychains = {
+        wallet: new Wallet(this.bitgo, this.baseCoin, newWallet),
+        userKeychain: {
```
margueriteblair, line 461: Need to remove

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -447,138 +447,156 @@ export class Wallets implements IWallets {
...
-            seed: params.coldDerivationSeed,
-          });
-          derivationPath = derivation.derivationPath;
-          userKeychain.pub = derivation.key;
-          userKeychain.derivedFromParentWithSeed = params.coldDerivationSeed;
-        }
-      } else {
-        if (!canEncrypt) {
-          throw new Error('cannot generate user keypair without passphrase');
+    if (params.type === 'custodial' && (params.multisigType ?? 'onchain') === 'onchain') {
```
zahin-mohammad, line 450: Nit: creating two private helper functions might help with readability

zahin-mohammad (approved):

PR #4880, opened 2024-09-03
feat(express): add config for lightning signer at express

saravanan7mani:
Config file to contain url and tls cert of lnd signer

Ticket: BTC-1473

### DIFF
--- a/modules/express/src/config.ts
+++ b/modules/express/src/config.ts
@@ -38,6 +38,7 @@ export interface Config {
   externalSignerUrl?: string;
   signerMode?: boolean;
   signerFileSystemPath?: string;
+  lightningSignerFileSystemPath?: string;
 }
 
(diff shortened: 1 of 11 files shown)


modules/express/src/config.ts
```diff
@@ -102,7 +105,7 @@ export const DefaultConfig: Config = {
  * @param url
  * @return {string}
  */
-function _forceSecureUrl(url: string): string {
+export function _forceSecureUrl(url: string): string {
```
OttoAllmendinger, line 108: the leading `_` usually means "private"
```suggestion
export function forceSecureUrl(url: string): string {
```

OttoAllmendinger (changes requested):

OttoAllmendinger (dismissed):

OttoAllmendinger (approved):

margueriteblair (approved):

PR #4881, opened 2024-09-03
fix(bitgo): update ada pledge example

theMobiusStrip:
Latest cardano-cli transaction witness generates a different CBOR format.

EA-4175


ryandengBG (approved):
LGTM

PR #4883, opened 2024-09-03
feat: remove GG18 signing during ECDSA TSS recovery

zahin-mohammad:
This PR does a gg18 -> dkls retrofit when performing a recovery tx.

TICKET: WP-2954


modules/bitgo/test/v2/fixtures/tss/recoveryFixtures.ts
```diff
@@ -1,5 +1,4 @@
 export const ethLikeGG18Keycard = {
-  senderAddress: '0xf3243334a491b6558f3f2c791afecc6a5eb955fa',
```
zahin-mohammad, line 2: Hardcoding these addresses hid a bug in the actual recover code.

PR #4884, opened 2024-09-04
feat(sdk-core): add bulkShareWallet method

SoumalyaBh:
Bulk wallet share sdk

Ticket: CS-3699

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -485,6 +485,52 @@ export interface ShareWalletOptions {
   disableEmail?: boolean;
 }
 
+export interface BulkCreateShareOption {
+  user: string;
+  permissions: string[];
+  keychain: BulkWalletShareKeychain;
+}
+
+export interface BulkWalletShareOptions {
+  walletPassphrase: string;
+  keyShareOptions: Array<{
+    userId: string;
+    pubKey: string;
+    path: string;
+    permissions: string[];
+  }>;
+}
+
+export type WalletShareState = 'active' | 'accepted' | 'canceled' | 'rejected' | 'pendingapproval';
+
+export interface BulkWalletShareKeychain {
+  pub: string;
+  encryptedPrv: string;
+  fromPubKey: string;
+  toPubKey: string;
+  path: string;
+}
+
+export interface WalletShare {
+  id: string;
+  coin: string;
+  wallet: string;
+  walletLabel?: string;
+  fromUser: string;
+  toUser: string;
+  permissions: string[];
+  state?: WalletShareState;
+  enterprise?: string;
+  message?: string;
+  pendingApprovalId?: string;
+  keychainOverrideRequired?: boolean;
+  isUMSInitiated?: boolean;
+  keychain?: BulkWalletShareKeychain;
+}
+
+export interface CreateBulkWalletShareListResponse {
+  shares: WalletShare[];
+}
 export interface RemoveUserOptions {
   userId?: string;
 }
@@ -753,6 +799,8 @@ export interface IWallet {
   getPrv(params?: GetPrvOptions): Promise<any>;
   createShare(params?: CreateShareOptions): Promise<any>;
   shareWallet(params?: ShareWalletOptions): Promise<any>;
+  createBulkKeyShares(params?: BulkCreateShareOption[]): Promise<CreateBulkWalletShareListResponse>;
+  createBulkWalletShare(params?: BulkWalletShareOptions): Promise<CreateBulkWalletShareListResponse>;
   removeUser(params?: RemoveUserOptions): Promise<any>;
   prebuildTransaction(params?: PrebuildTransactionOptions): Promise<PrebuildTransactionResult>;
   signTransaction(params?: WalletSignTransactionOptions): Promise<SignedTransaction>;
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -42,6 +42,8 @@ import {
   CreateAddressOptions,
   CreatePolicyRuleOptions,
   CreateShareOptions,
+  BulkCreateShareOption,
+  BulkWalletShareOptions,
   CrossChainUTXO,
   DeployForwardersOptions,
   DownloadKeycardOptions,
@@ -91,6 +93,7 @@ import {
   WalletSignTransactionOptions,
   WalletSignTypedDataOptions,
   WalletType,
+  CreateBulkWalletShareListResponse,
 } from './iWallet';
 import { StakingWallet } from '../staking';
 import { Lightning } from '../lightning/custodial';
@@ -1464,13 +1467,124 @@ export class Wallet implements IWallet {
   }
 
   /**
+   * Shares a wallet with multiple users by creating bulk wallet shares.
+   *
+   * @async
+   * @param {BulkWalletShareOptions} params - The options for sharing wallets in bulk.
+   * @param {Array<ShareOption>} params.shareOptions - An array of share option objects containing user and permissions information.
+   * @param {Object} [params.shareOptions[].keychain] - The keychain object used to share the wallet.
+   * @param {string} [params.shareOptions[].keychain.toPubKey] - The recipient's public key.
+   * @param {string} [params.shareOptions[].keychain.path] - The derivation path of the keychain.
+   * @param {string} params.shareOptions[].user - The user to share the wallet with.
+   * @param {string} params.shareOptions[].permissions - The permissions granted to the user.
+   * @param {string} [params.walletPassphrase] - The wallet passphrase used to decrypt the keychain.
+   * @throws {Error} If `shareOptions` is empty, or if required keychain parameters (`toPubKey` and `path`) are missing when needed.
+   * @throws {Error} If unable to decrypt the user keychain.
+   * @returns {Promise<CreateBulkWalletShareListResponse>} A promise that resolves with the response of the bulk wallet share creation.
+   */
+  async createBulkWalletShare(params: BulkWalletShareOptions): Promise<CreateBulkWalletShareListResponse> {
+    if (!params.keyShareOptions || Object.keys(params.keyShareOptions).length === 0) {
+      throw new Error('shareOptions cannot be empty');
+    }
+    const bulkCreateShareOptions: BulkCreateShareOption[] = [];
+
+    for (const shareOption of params.keyShareOptions) {
+      common.validateParams(shareOption, ['userId', 'pubKey', 'path'], []);
+
+      const needsKeychain = shareOption.permissions && shareOption.permissions.includes('spend');
+      let sharedKeychain;
+
+      if (needsKeychain) {
+        try {
+          const keychain = await this.getEncryptedUserKeychain();
+
+          if (keychain.encryptedPrv) {
+            const userPrv = decryptKeychainPrivateKey(this.bitgo, keychain, params.walletPassphrase);
+            if (!userPrv) {
+              throw new Error('Unable to decrypt user keychain.');
+            }
+
+            const ecdhkey = makeRandomKey();
+            assert(shareOption.pubKey, 'pubKey must be defined for sharing');
+
+            const secret = getSharedSecret(ecdhkey, Buffer.from(shareOption.pubKey, 'hex')).toString('hex');
+            const newEncryptedPrv = this.bitgo.encrypt({ password: secret, input: userPrv });
+
+            let pub = keychain.pub ?? keychain.commonPub;
+            if (keychain.commonKeychain) {
+              pub =
+                this.baseCoin.getMPCAlgorithm() === 'eddsa'
+                  ? EddsaUtils.getPublicKeyFromCommonKeychain(keychain.commonKeychain)
+                  : EcdsaUtils.getPublicKeyFromCommonKeychain(keychain.commonKeychain);
+            }
+
+            sharedKeychain = {
+              pub,
+              encryptedPrv: newEncryptedPrv,
+              fromPubKey: ecdhkey.publicKey.toString('hex'),
+              toPubKey: shareOption.pubKey,
+              path: shareOption.path,
+            };
+          }
+        } catch (e) {
+          if (e.message === 'No encrypted keychains on this wallet.') {
+            sharedKeychain = {};
+            // ignore this error because this looks like a cold wallet
+          } else {
+            throw e;
+          }
+        }
+        const keychain = Object.keys(sharedKeychain ?? {}).length === 0 ? undefined : sharedKeychain;
+        if (keychain) {
+          bulkCreateShareOptions.push({
+            user: shareOption.userId,
+            permissions: shareOption.permissions,
+            keychain: keychain,
+          });
+        }
+      }
+    }
+    return await this.createBulkKeyShares(bulkCreateShareOptions);
+  }
+
+  /**
+   * Creates bulk wallet share entries for specified share options.
+   * Filters out share options that do not contain valid keychain information or have missing keychain fields.
+   * If all share options are invalid or empty, it throws an error.
+   * Sends a POST request to create the wallet shares for valid share options.
+   *
+   * @async
+   * @param {BulkCreateShareOption[]} [params=[]] - The array of share options to process.
+   *   Keychain entries must include the following fields: `pub`, `encryptedPrv`, `fromPubKey`, `toPubKey`, and `path`.
+   * @returns {Promise<CreateBulkWalletShareListResponse>} A promise resolving to the result of the wallet shares creation.
+   * @throws {Error} Throws an error if no valid share options are provided.
+   */
+  async createBulkKeyShares(params: BulkCreateShareOption[] = []): Promise<CreateBulkWalletShareListResponse> {
+    params = params.filter((shareOption) => {
+      try {
+        common.validateParams(shareOption.keychain, ['pub', 'encryptedPrv', 'fromPubKey', 'toPubKey', 'path'], []);
+        return true;
+      } catch (e) {
+        // Exclude share options with invalid keychain
+        return false;
+      }
+    });
+
+    if (!params || Object.keys(params).length === 0) {
+      throw new Error('shareOptions cannot be empty');
+    }
+
+    const url = this.bitgo.url(`/wallet/${this._wallet.id}/walletshares`, 2);
+    return this.bitgo.post(url).send({ shareOptions: params }).result();
+  }
+
+  /**
    * Share this wallet with another BitGo user.
    * @param params
    * @returns {*}
    */
   async shareWallet(params: ShareWalletOptions = {}): Promise<any> {
     common.validateParams(params, ['email', 'permissions'], ['walletPassphrase', 'message']);
-
     if (params.reshare !== undefined && !_.isBoolean(params.reshare)) {
       throw new Error('Expected reshare to be a boolean.');
     }
--- a/modules/bitgo/test/v2/unit/wallets.ts
+++ b/modules/bitgo/test/v2/unit/wallets.ts
@@ -6,6 +6,7 @@ import * as nock from 'nock';
 import * as sinon from 'sinon';
 import * as should from 'should';
 import * as _ from 'lodash';
+
 import { TestBitGo } from '@bitgo/sdk-test';
 import {
   BlsUtils,
@@ -17,10 +18,14 @@ import {
   GenerateWalletOptions,
   Wallet,
   isWalletWithKeychains,
+  BulkWalletShareOptions,
+  OptionalKeychainEncryptedKey,
+  KeychainWithEncryptedPrv,
 } from '@bitgo/sdk-core';
 import { BitGo } from '../../../src';
 import { afterEach } from 'mocha';
 import { TssSettings } from '@bitgo/public-types';
+import * as moduleBitgo from '@bitgo/sdk-core';
 
 describe('V2 Wallets:', function () {
   const bitgo = TestBitGo.decorate(BitGo, { env: 'mock' });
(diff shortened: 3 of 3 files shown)


Silvrbckw (approved):
merge

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -753,6 +782,8 @@ export interface IWallet {
   getPrv(params?: GetPrvOptions): Promise<any>;
   createShare(params?: CreateShareOptions): Promise<any>;
   shareWallet(params?: ShareWalletOptions): Promise<any>;
+  createBulkShares(params?: BulkCreateShareOptions): Promise<any>;
+  bulkShareWallet(params?: BulkShareWalletOptions): Promise<any>;
```
mohammadalfaiyazbitgo, line 786: add a type for the response

mohammadalfaiyazbitgo (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1465,111 @@ export class Wallet implements IWallet {
...
+      throw new Error('shareOptions cannot be empty');
+    }
+    params.shareOptions.forEach((shareOption) => {
+      if (shareOption.keychain) {
+        const { pub, encryptedPrv, fromPubKey, toPubKey, path } = shareOption.keychain;
+        if (!pub || !encryptedPrv || !fromPubKey || !toPubKey || !path) {
+          throw new Error('requires keychain parameters - pub, encryptedPrv, fromPubKey, toPubKey, path');
+        }
+      }
+    });
```
mohammadalfaiyazbitgo, line 1557: you can use discriminated union types for your params so you wouldn't need to have this check i think.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1465,111 @@ export class Wallet implements IWallet {
...
+    // URL format - `https://bitgo.fakeurl/api/v2/wallet/${this._wallet.id}/walletshares`;
+    const url = `${this.baseCoin.url('/wallet/').replace(this.baseCoin.getChain() + '/', '')}${
+      this._wallet.id
+    }/walletshares`;
```
mohammadalfaiyazbitgo, line 1561: just use `bitGo.url` if you're not using the `coin` url. Or extend the `baseCoin.url` method to replace `coin` with `chain`.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1465,111 @@ export class Wallet implements IWallet {
...
+          keychain: Object.keys(sharedKeychain ?? {}).length === 0 ? undefined : sharedKeychain,
+        });
+      }
+    }
+    return await this.createBulkShares(bulkCreateShareOptions);
+  }
+
+  /**
+   * Send an encrypted wallet share to BitGo.
+   * @param params
```
mohammadalfaiyazbitgo, line 1544: ?

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -485,6 +485,35 @@ export interface ShareWalletOptions {
...
+      pub: string;
+      encryptedPrv: string;
+      fromPubKey: string;
+      toPubKey: string;
+      path: string;
+    };
+  }>;
+}
+
+export interface BulkShareWalletOptions {
```
pranavjain97, line 502:
```suggestion
export interface BulkWalletShareOptions {
```

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -485,6 +485,35 @@ export interface ShareWalletOptions {
...
+    permissions: string;
+    keychain: {
+      pub: string;
+      encryptedPrv: string;
+      fromPubKey: string;
+      toPubKey: string;
+      path: string;
+    };
+  }>;
+}
```
pranavjain97, line 500:
```suggestion
export interface BulkKeyShareOption {
    user: string;
    permissions: string;
    keychain: {
      pub: string;
      encryptedPrv: string;
      fromPubKey: string;
      toPubKey: string;
      path: string;
    };
}
```
I dont see why it needs to be nested under `shareOptions`. Just use an array type where its being used

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -485,6 +485,35 @@ export interface ShareWalletOptions {
...
+    permissions: string;
+    keychain: {
+      pub: string;
+      encryptedPrv?: string;
+      fromPubKey?: string;
+      toPubKey?: string;
+      path?: string;
+    };
+  }>;
+}
```
pranavjain97, line 515: why is this being retyped?? 
```suggestion
  shareOptions: BulkKeyShareOption[]
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1465,111 @@ export class Wallet implements IWallet {
...
+   */
+  async createBulkShares(params: BulkCreateShareOptions = { shareOptions: [] }): Promise<any> {
```
pranavjain97, line 1546:
```suggestion
  async createBulkKeyShares(params: BulkCreateShareOptions = { shareOptions: [] }): Promise<any> {
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1465,111 @@ export class Wallet implements IWallet {
     return this.bitgo.post(this.url('/share')).send(params).result();
   }
 
+  // Send wallet shares belonging to the same wallet in bulk
+  async bulkShareWallet(params: BulkShareWalletOptions): Promise<BulkCreateShareOptions> {
```
pranavjain97, line 1469:
```suggestion
  async bulkWalletShare(params: BulkShareWalletOptions): Promise<BulkCreateShareOptions> {
```

pranavjain97 (changes requested):

modules/sdk-core/src/bitgo/wallet/iWallet.ts
SoumalyaBh: The shareOptions array has been included to keep in sync with the input format for the createBulkWalletShares endpoint - https://github.com/BitGo/wallet-platform-types/pull/503/files#diff-9c2bf6507afd36913d5840997df501241100ca0db7ee94d1a5cee763d93cfbb0

modules/sdk-core/src/bitgo/wallet/wallet.ts
SoumalyaBh: The bitgo.url gives the following
```
this.bitgo.url("/hello")
'https://bitgo.fakeurl/api/v1/hello'
```
 While I need a 
` https://bitgo.fakeurl/api/v2/wallet/${this._wallet.id}/walletshares`
I'm using the baseCoin URL and replacing the coin with space(removes the /coin/ part) And adding the walletId

modules/sdk-core/src/bitgo/wallet/wallet.ts
SoumalyaBh: Actually I was following this format - https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/wallet.ts#L1452-L1460
Its also easier for the caller to understand the error if the caller has missing keychain parameter.
Should I still modify the code to use `discriminated union types`??

modules/sdk-core/src/bitgo/wallet/wallet.ts
SoumalyaBh: removed

modules/sdk-core/src/bitgo/wallet/wallet.ts
sdoshibitgo: this.bitgo.url("/hello", 2) will give you v2 url.

modules/sdk-core/src/bitgo/wallet/wallet.ts
mohammadalfaiyazbitgo: yes if it's not a huge refactor, because if you're using a good union, the caller would get an error during compilation and not run time.

modules/sdk-core/src/bitgo/wallet/wallet.ts
mohammadalfaiyazbitgo: no you shouldn't remove them, you should add move documentation on the parameters instead of `params params`

mohammadalfaiyazbitgo: and the jsdoc description doesn't state this a /bulk/ key share method.

modules/sdk-core/src/bitgo/wallet/iWallet.ts
pranavjain97: Then please lets change the createBulkWalletShares endpoint because it doesnt make any sense to nest it.

modules/sdk-core/src/bitgo/wallet/iWallet.ts
pranavjain97: Can you explain why they are optional in one and required in another? I see them being required in all key share operations.

pranavjain97 (commented):
Left some comments

modules/sdk-core/src/bitgo/wallet/iWallet.ts
SoumalyaBh: sure, 
The bulk version of this method has been kept in sync with the non bulk version implemented here
CreateShareOptions - https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/iWallet.ts#L453-L473
Moreover, In the method `bulkWalletShare`, we are generating new encryptedPrv, as well as pub and creating a `sharedKeychain`.

While in the `createBulkKeyShares` method we need all the parameters in the keychain.

modules/sdk-core/src/bitgo/wallet/wallet.ts
SoumalyaBh: Hi @mohammadalfaiyazbitgo 
Modified, the distributed union types is being enforced by - https://github.com/BitGo/BitGoJS/pull/4884/files#diff-d998f4b5c3a58f8f1ac3c8075b65be2eef631b7e951bb9af3fa81ccf4af52642R492
We can remove the 
`if (!pub || !encryptedPrv || !fromPubKey || !toPubKey || !path) {
          throw new Error('requires keychain parameters - pub, encryptedPrv, fromPubKey, toPubKey, path');
        }` it will give compile time error - `No match for request`. 

However, by using this check, we can provide the caller with better error logging. 
Should I still remove the check?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,138 @@ export class Wallet implements IWallet {
...
+    }
+    params.shareOptions.forEach((shareOption) => {
+      if (shareOption.keychain) {
+        const { pub, encryptedPrv, fromPubKey, toPubKey, path } = shareOption.keychain;
+        if (!pub || !encryptedPrv || !fromPubKey || !toPubKey || !path) {
+          throw new Error('requires keychain parameters - pub, encryptedPrv, fromPubKey, toPubKey, path');
+        }
+      }
+    });
+    // URL format - `https://bitgo.fakeurl/api/v2/wallet/${this._wallet.id}/walletshares`;
```
sdoshibitgo, line 1588: this comment doesn't add much value

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,138 @@ export class Wallet implements IWallet {
...
+
+  async createBulkKeyShares(
+    params: BulkCreateShareOptions = { shareOptions: [] }
+  ): Promise<CreateBulkWalletShareListResponse> {
+    if (!params.shareOptions || Object.keys(params.shareOptions).length === 0) {
+      throw new Error('shareOptions cannot be empty');
+    }
+    params.shareOptions.forEach((shareOption) => {
+      if (shareOption.keychain) {
+        const { pub, encryptedPrv, fromPubKey, toPubKey, path } = shareOption.keychain;
```
sdoshibitgo, line 1582: we can use common.validateParams on this

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,138 @@ export class Wallet implements IWallet {
...
+   */
+
+  async createBulkKeyShares(
+    params: BulkCreateShareOptions = { shareOptions: [] }
+  ): Promise<CreateBulkWalletShareListResponse> {
+    if (!params.shareOptions || Object.keys(params.shareOptions).length === 0) {
+      throw new Error('shareOptions cannot be empty');
+    }
+    params.shareOptions.forEach((shareOption) => {
+      if (shareOption.keychain) {
```
sdoshibitgo, line 1581: isn't this mandatory in bulk share? We should throw error if keychain is not found

SoumalyaBh: not throwing an error/ breaking the call for a missing keychain. However filtering out the shareoptions without keychain/ invalid keychain

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,146 @@ export class Wallet implements IWallet {
...
+    if (!params.shareOptions || Object.keys(params.shareOptions).length === 0) {
+      throw new Error('shareOptions cannot be empty');
+    }
+    const filteredShareOptions = params.shareOptions.filter((shareOption) => {
+      if (shareOption.keychain) {
+        // Validate that all keychain fields are present
+        try {
+          const { pub, encryptedPrv, fromPubKey, toPubKey, path } = shareOption.keychain;
+          common.validateParams(shareOption.keychain, ['pub', 'encryptedPrv', 'fromPubKey', 'toPubKey', 'path'], []);
+          return pub && encryptedPrv && fromPubKey && toPubKey && path;
```
sdoshibitgo, line 1586: we can return true here as validateParams will throw error if any of the attribute is not found

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,146 @@ export class Wallet implements IWallet {
...
+        } catch (e) {
+          // Exclude share options with invalid keychain
+          return false;
+        }
+      }
+      // Exclude share options that don't have a keychain
+      return false;
+    });
+    params.shareOptions = filteredShareOptions;
```
sdoshibitgo, line 1595: we should check if filteredShareOptions has any elements or not before making http call

SoumalyaBh: The bulkCreateShare api will be handling the empty call

sdoshibitgo: yes but why to make api call for empty body?

sdoshibitgo (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,141 @@ export class Wallet implements IWallet {
...
+   * @returns {Promise<CreateBulkWalletShareListResponse>} A promise that resolves with the response of the bulk wallet share creation.
+   */
+  async bulkWalletShare(params: BulkWalletShareOptions): Promise<CreateBulkWalletShareListResponse> {
+    if (!params.shareOptions || Object.keys(params.shareOptions).length === 0) {
+      throw new Error('shareOptions cannot be empty');
+    }
+    const bulkCreateShareOptions: BulkCreateShareOptions = { shareOptions: [] };
+
+    for (let index = 0; index < params.shareOptions.length; index++) {
+      const shareOption = params.shareOptions[index];
```
mohammadalfaiyazbitgo, line 1492: nit
```suggestion
    for (const shareOption of params.shareOptions) {
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,141 @@ export class Wallet implements IWallet {
...
+
+      common.validateParams(shareOption, ['userId', 'pubKey', 'path'], []);
+
+      const needsKeychain = shareOption.permissions && shareOption.permissions.indexOf('spend') !== -1; // Remove !shareOption.keychain
```
mohammadalfaiyazbitgo, line 1496: nit
```suggestion
      const needsKeychain = shareOption.permissions && shareOption.permissions.includes('spend'); // Remove !shareOption.keychain
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,141 @@ export class Wallet implements IWallet {
...
+              pub,
+              encryptedPrv: newEncryptedPrv,
+              fromPubKey: eckey.publicKey.toString('hex'),
+              toPubKey: sharing.pubKey,
+              path: sharing.path,
+            };
+          }
+        } catch (e) {
+          if (e.message === 'No encrypted keychains on this wallet.') {
+            sharedKeychain = {};
```
mohammadalfaiyazbitgo, line 1536: why are you not re-throwing this error?

modules/sdk-core/src/bitgo/wallet/wallet.ts
mohammadalfaiyazbitgo, line 1492: I think you could parallelize this, if you loop over and implement this sharing as a promise.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,141 @@ export class Wallet implements IWallet {
...
+   * @param {BulkCreateShareOptions} [params={ shareOptions: [] }] - The options for creating bulk key shares.
+   * @param {Array<ShareOption>} params.shareOptions - An array of share option objects containing keychain information.
+   * @param {Object} params.shareOptions[].keychain - The keychain object for each share option.
+   * @param {string} params.shareOptions[].keychain.pub - The public key of the keychain.
+   * @param {string} params.shareOptions[].keychain.encryptedPrv - The encrypted private key of the keychain.
+   * @param {string} params.shareOptions[].keychain.fromPubKey - The sender's public key.
+   * @param {string} params.shareOptions[].keychain.toPubKey - The recipient's public key.
+   * @param {string} params.shareOptions[].keychain.path - The derivation path of the keychain.
+   * @throws {Error} If `shareOptions` is empty or any required keychain parameter is missing.
+   * @returns {Promise<any>} A promise that resolves with the result of the bulk key shares creation request.
```
mohammadalfaiyazbitgo, line 1564: this return is incorrect.

modules/sdk-core/src/bitgo/wallet/wallet.ts
SoumalyaBh: This is the reason - https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/wallet.ts#L1527-L1531

modules/sdk-core/src/bitgo/wallet/iWallet.ts
pranavjain97: The link isnt working. Either way I'm not sure I follow. Why do u need an unecessary nesting to accomodate additonal properties in the future? Can you give me an example? Whoever uses the interface can still use it as an array

SoumalyaBh: I think this will be a good example - https://github.com/BitGo/BitGoJS/pull/4884/files#diff-d998f4b5c3a58f8f1ac3c8075b65be2eef631b7e951bb9af3fa81ccf4af52642R496-R504 
This is the reason I preferred to keep the nesting. For ex, here we are adding a walletPassphrase along with the array. The nesting will help in adding  additional properties to the same interface.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,133 @@ export class Wallet implements IWallet {
...
+   * @param {string} [params.walletPassphrase] - The wallet passphrase used to decrypt the keychain.
+   * @throws {Error} If `shareOptions` is empty, or if required keychain parameters (`toPubKey` and `path`) are missing when needed.
+   * @throws {Error} If unable to decrypt the user keychain.
+   * @returns {Promise<CreateBulkWalletShareListResponse>} A promise that resolves with the response of the bulk wallet share creation.
+   */
+  async bulkWalletShare(params: BulkWalletShareOptions): Promise<CreateBulkWalletShareListResponse> {
+    if (!params.shareOptions || Object.keys(params.shareOptions).length === 0) {
+      throw new Error('shareOptions cannot be empty');
+    }
+    // const bulkCreateShareOptions: BulkCreateShareOptions = { shareOptions: [] };
```
mohammadalfaiyazbitgo, line 1489: remove unused code.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,133 @@ export class Wallet implements IWallet {
...
+          common.validateParams(shareOption.keychain, ['pub', 'encryptedPrv', 'fromPubKey', 'toPubKey', 'path'], []);
+          return true;
+        } else {
+          // Exclude share options that don't have a keychain
+          return false;
+        }
+      } catch (e) {
+        // Exclude share options with invalid keychain
+        return false;
+      }
```
mohammadalfaiyazbitgo, line 1577: this confuses me, why make `shareOption.keychain` optional if it's required for this function to actually do anything? Am I reading this wrong, if the `keychain` doesn't exist then this won't go through to creating a wallet share, but then why make it optional?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,133 @@ export class Wallet implements IWallet {
...
+            sharedKeychain = {};
+            // ignore this error because this looks like a cold wallet
+          } else {
+            throw e;
+          }
+        }
+        bulkCreateShareOptions.push({
+          user: shareOption.userId,
+          permissions: shareOption.permissions,
+          keychain: Object.keys(sharedKeychain ?? {}).length === 0 ? undefined : sharedKeychain,
```
mohammadalfaiyazbitgo, line 1544: you could make your `createBulkKeyShares` method cleaner by making `keychain` required and not pushing a share option if keychain doesn't exist.

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -485,6 +485,55 @@ export interface ShareWalletOptions {
...
+  state?: WalletShareState;
+  enterprise?: string;
+  message?: string;
+  pendingApprovalId?: string;
+  keychainOverrideRequired?: boolean;
+  isUMSInitiated?: boolean;
+  keychain?: BulkWalletShareKeychain;
+}
+
+// Define the CreateBulkWalletShareListResponse type
```
mohammadalfaiyazbitgo, line 533: these comments are redundant, remove them.

modules/sdk-core/src/bitgo/wallet/wallet.ts
mohammadalfaiyazbitgo: not a fan of handling errors like this, but no alternative comes to mind at the moment.

mohammadalfaiyazbitgo (changes requested):
few more comments on the changes.

mohammadalfaiyazbitgo (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,129 @@ export class Wallet implements IWallet {
...
+    if (!params.shareOptions || Object.keys(params.shareOptions).length === 0) {
+      throw new Error('shareOptions cannot be empty');
+    }
+    const bulkCreateShareOptions: BulkCreateShareOption[] = [];
+
+    for (const shareOption of params.shareOptions) {
+      common.validateParams(shareOption, ['userId', 'pubKey', 'path'], []);
+
+      const needsKeychain = shareOption.permissions && shareOption.permissions.includes('spend');
+      let sharedKeychain: any = undefined;
```
pranavjain97, line 1495:
```suggestion
      let sharedKeychain;
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,129 @@ export class Wallet implements IWallet {
...
+
+          if (!keychain.encryptedPrv) {
+            sharedKeychain = undefined;
+          } else {
+            const userPrv = decryptKeychainPrivateKey(this.bitgo, keychain, params.walletPassphrase);
+            if (!userPrv) {
+              throw new Error('Unable to decrypt user keychain.');
+            }
+
+            const eckey = makeRandomKey();
```
pranavjain97, line 1509:
```suggestion
            const ecdhkey = makeRandomKey();
```

pranavjain97: Are you sure this is how the ecdh key is generated? 🤔 I beleive we have some standardized methods we use to generate it

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,129 @@ export class Wallet implements IWallet {
...
+            const sharing = shareOption;
```
pranavjain97, line 1510: why is this being re-assigned?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,129 @@ export class Wallet implements IWallet {
...
+    const filteredShareOptions = params.filter((shareOption) => {
+      try {
+        common.validateParams(shareOption.keychain, ['pub', 'encryptedPrv', 'fromPubKey', 'toPubKey', 'path'], []);
+        return true;
+      } catch (e) {
+        // Exclude share options with invalid keychain
+        return false;
+      }
+    });
+    params = filteredShareOptions;
```
pranavjain97, line 1575: not sure why is it re-assigned here

pranavjain97 (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1463,14 +1466,125 @@ export class Wallet implements IWallet {
...
+              fromPubKey: ecdhkey.publicKey.toString('hex'),
+              toPubKey: shareOption.pubKey,
+              path: shareOption.path,
+            };
+          }
+        } catch (e) {
+          if (e.message === 'No encrypted keychains on this wallet.') {
+            sharedKeychain = {};
+            // ignore this error because this looks like a cold wallet
+          } else {
```
pranavjain97, line 1533: Isn't this the same code as this? https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/wallet.ts#L1507

Let's not copy paste. Preferably move it to a helper and re-use

pranavjain97 (approved):
Expecting the de-duplication of walletShare logic in a follow-up.

PR #4904, opened 2024-09-11
feat(bitgo): add BitGo Korea entity

pranishnepal:
## What
Currently, "BiGo Sister Trust 1" represents BitGo Korea entity. This commit adds "BitGo Korea" as a new entity since that would be safer than changing the existing.

## Issue Number
TICKET: COPS-2076

## Type of change

## Testing
- Unit Tests 
- `yarn run unit-test --scope @bitgo/statics`


modules/statics/src/base.ts
```diff
@@ -213,9 +213,14 @@ export enum CoinFeature {
    */
   CUSTODY_BITGO_SINGAPORE = 'custody-bitgo-singapore',
   /*
-   * This coin supports custody in BitGo Sister Trust 1 entities
+   * This coin supports custody in BitGo Sister Trust 1 entities.
+   * This used to represent BitGo Korea previously. Going forward, use BitGo Korea instead.
```
aarvay, line 217:
```suggestion
   * @deprecated This used to represent BitGo Korea previously. Going forward, use `CUSTODY_BITGO_KOREA` instead.
```

pranishnepal: updated, thanks!

yarn.lock
abhishekdhakla, line 1: I think we don't need this one!

abhishekdhakla (changes requested):

yarn.lock
pranishnepal: i thought about it, but didn't see it in `.gitignore` so i let it commit. Shall i add it to `.gitignore,` then?

pranishnepal: removed + added to `.gitignore`

abhishekdhakla (approved):

alebusse (approved):

lzjluzijie (approved):

PR #4906, opened 2024-09-16
feat: Add utility function for get sharing keys AP

SoumalyaBh:
Ticket: CSI-107

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -17,7 +17,7 @@ import {
 import { makeRandomKey } from '../bitcoin';
 import { BitGoBase } from '../bitgoBase';
 import { getSharedSecret } from '../ecdh';
-import { AddressGenerationError, MethodNotImplementedError } from '../errors';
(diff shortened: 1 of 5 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1578,6 +1559,61 @@ export class Wallet implements IWallet {
...
+        sharedKeychain = {
+          pub,
+          encryptedPrv: newEncryptedPrv,
+          fromPubKey: eckey.publicKey.toString('hex'),
+          toPubKey: pubkey,
+          path: path,
+        };
+      }
+    } catch (e) {
+      if (e.message === 'No encrypted keychains on this wallet.') {
```
pengyuc-bitgo, line 1606: I think this is old code but we really shouldn't be doing production logics based on the error messages. 
It would be fine in the tests to verify the error messages are as expected but not in the production code to base the logics on.
Please add an error class of `MissingEncryptedKeychainError extends Error` and do the logic with `instanceof`
```suggestion
      if (e instanceof MissingEncryptedKeychainError) {
```

pengyuc-bitgo (approved):

PR #4907, opened 2024-09-16
fix: add wallet type check for sign transaction

margueriteblair:
Ticket: WP-2664

TICKET: WP-2664


zahin-mohammad (approved):
lgtm; but please add a test as a follow up

PR #4908, opened 2024-09-16
chore: refactor dkg initialization to seperately between user and backup

mohammadalfaiyazbitgo:
Ticket: WP-2292

TICKET: WP-2292

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
@@ -388,6 +388,31 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
   }
 
   /**
+   * Converts a User or Backup MPCv1 SigningMaterial to RetrofitData needed by MPCv2 DKG.
+   *
+   * @param decryptedKeyshare - MPCv1 decrypted signing material for user or backup as a json.stringify string and bitgo's Big Si.
+   * @param partyId - The party ID of the MPCv1 keyshare.
+   * @returns The retrofit data needed to start an MPCv2 DKG session.
+   */
+  getKeyDataForRetrofit(
+    decryptedKeyshare: string,
+    partyId: MPCv2PartiesEnum.BACKUP | MPCv2PartiesEnum.USER
+  ): DklsTypes.RetrofitData {
+    const mpc = new Ecdsa();
+    const xiList = [
+      Array.from(bigIntToBufferBE(BigInt(1), 32)),
+      Array.from(bigIntToBufferBE(BigInt(2), 32)),
+      Array.from(bigIntToBufferBE(BigInt(3), 32)),
+    ];
+    return this.getMpcV2RetrofitDataFromMpcV1Key({
+      mpcv1PartyKeyShare: decryptedKeyshare,
+      mpcv1PartyIndex: partyId === MPCv2PartiesEnum.USER ? 1 : 2,
+      xiList,
+      mpc,
+    });
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -387,6 +387,62 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
...
+    seed,
+  }: {
+    m: number;
+    n: number;
+    decryptedKeyShare?: string;
+    partyId: MPCv2PartiesEnum.USER | MPCv2PartiesEnum.BACKUP;
+    seed?: Buffer;
+  }): DklsDkg.Dkg {
+    if (decryptedKeyShare) {
+      const retrofitData = this.getKeyDataForRetrofit(decryptedKeyShare!, partyId);
```
islamaminBitGo, line 439: why is the ! needed here? i thought typescript would detect that, within this scope, decryptedKeyShare would not be undefined

islamaminBitGo (changes requested):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -387,6 +387,62 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
...
+  }: {
+    m: number;
+    n: number;
+    decryptedKeyShare?: string;
+    partyId: MPCv2PartiesEnum.USER | MPCv2PartiesEnum.BACKUP;
+    seed?: Buffer;
+  }): DklsDkg.Dkg {
+    if (decryptedKeyShare) {
+      const retrofitData = this.getKeyDataForRetrofit(decryptedKeyShare, partyId);
+      return new DklsDkg.Dkg(n, m, partyId, seed, retrofitData);
```
islamaminBitGo, line 440: i guess you would also want to dependency inject a dkls wasm since im assuming this helper is to be called from applications that will require the bundler wasm rather than the default node wasm

mohammadalfaiyazbitgo: hmm good point, in that case I think it would be better to just have the caller for DKG to init the session themselves and manage how the wasm should be loaded. In which case this wouldn't be needed.

islamaminBitGo (approved):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -387,6 +387,31 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
...
   }
 
+  /**
+   * Converts a User or Backup MPCv1 SigningMaterial to RetrofitData needed by MPCv2 DKG.
+   *
+   * @param decryptedKeyshare - MPCv1 decrypted signing material for user or backup as a json.stringify string and bitgo's Big Si.
+   * @param partyId - The party ID of the MPCv1 keyshare.
+   * @returns The retrofit data needed to start an MPCv2 DKG session.
+   */
+  getKeyDataForRetrofit(
```
zahin-mohammad, line 397: Nit: can you refactor the call sites of `getMpcV2RetrofitDataFromMpcV1Keys` to use this new fn instead?

Seems a little redundant to have both, we can use this as a more generic util function.

zahin-mohammad (approved):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
mohammadalfaiyazbitgo: yeah I'll line up a refactor pr

PR #4910, opened 2024-09-18
refactor(sdk-core): decouple request sender from MPCv2 codes

pengyuc-bitgo:
Allow overriding the key-gen request sender in the MPCv2 key generation codes. This will allow other
work flow such as custodial key retrofitting to reuse most of the key generation codes in the sdk.

The interface of all APIs are not changed. Should all be backward compatible. 
New test is not needed because the original test in [TSS Ecdsa MPCv2 Utils](https://github.com/BitGo/BitGoJS/blob/c6f673d8fe810e586a99df5be3e8f7a1c1977816/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsaMPCv2/createKeychains.ts#L31) covers these codes.

Ticket: WP-2617


pranavjain97 (commented):
Looks fine to me. Just test E2E with existing SMC for regression

margueriteblair (approved):

PR #4931, opened 2024-09-24
feat: add psbt signing for v1 wallets

OttoAllmendinger:
This commit adds optional PSBT signing to the v1 wallet API.

Call the `signTransaction` method with a `psbt: string` (hex-encoded) argument
to co-sign a PSBT.

The method will return the signed PSBT as a hex-encoded string.

Issue: BTC-1351


OttoAllmendinger:
updating CODEOWNERS

saravanan7mani (dismissed):

saravanan7mani (approved):

PR #4933, opened 2024-09-24
feat(sdk-api): authenticate with passkey params

alvin-dai-bitgo:
Ticket: WP-2592

Add a new method `authenticateWithPasskey`

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -47,7 +47,6 @@ import {
   AddAccessTokenOptions,
   AddAccessTokenResponse,
   AuthenticateOptions,
-  AuthenticateWithPasskeyOptions,
   AuthenticateWithAuthCodeOptions,
   BitGoAPIOptions,
   BitGoJson,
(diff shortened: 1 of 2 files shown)


modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -772,6 +774,58 @@ export class BitGoAPI implements BitGoBase {
...
+  preprocessAuthenticationPasskeyParams(params: AuthenticateWithPasskeyOptions): ProcessedAuthenticationPasskeyOptions {
+    if (!_.isString(params.username)) {
+      throw new Error('expected string username');
+    }
+
+    if (!_.isString(params.credId)) {
+      throw new Error('expected string credId');
+    }
+
+    if (!_.isObject(params.response)) {
```
pranavjain97, line 789: dont think it needs to be nested under response

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -772,6 +774,58 @@ export class BitGoAPI implements BitGoBase {
...
+
+    const processedParams: ProcessedAuthenticationPasskeyOptions = {
+      username: params.username,
+      credId: params.credId,
+      response: params.response,
+    }
+
+    if (params.otp) {
+      processedParams.otp = params.otp;
+    }
```
pranavjain97, line 813: what is this for? the current "response" is the replacement of the otp data

pranavjain97 (changes requested):

pranavjain97 (dismissed):

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -772,6 +773,32 @@ export class BitGoAPI implements BitGoBase {
...
   }
 
+  /**
+   * Validate the passkey response is in the expected format
+   * Should be as is returned from navigator.credentials.get()
+   */
+  validateWebauthnResponse(params: AuthenticateWithPasskeyOptions): void {
+    if (!_.isString(params.username)) {
+      throw new Error('expected string username');
+    }
```
pranavjain97, line 783: We'll actually extract the username from the webauthn response's userHandle in a follow-up. That way - UI can send data as it.

alebusse (approved):

PR #4935, opened 2024-09-24
fix(sdk-coin-hbar): hbar address validation

ajays97:
Ticket: PX-5154

HBAR address validation should pass for addresses that contain memoId.


sachushaji (approved):
LGTM

PR #4944, opened 2024-09-25
chore(avaxc) : wrw support for evm ccr 

nrjsuthar:
chore(avaxc) : wrw support for evm ccr 

TICKET: COIN-1708

### DIFF
--- a/modules/sdk-coin-avaxc/src/avaxc.ts
+++ b/modules/sdk-coin-avaxc/src/avaxc.ts
@@ -62,11 +62,12 @@ import {
   VerifyAvaxcTransactionOptions,
 } from './iface';
 import { AvaxpLib } from '@bitgo/sdk-coin-avaxp';
+import { AvaxcRecovery } from './avaxcRecovery';
 
 export class AvaxC extends BaseCoin {
   static hopTransactionSalt = 'bitgoHopAddressRequestSalt';
-
   protected readonly _staticsCoin: Readonly<StaticsBaseCoin>;
+  private ethLikeAvaxc: any;
 
   protected constructor(bitgo: BitGoBase, staticsCoin?: Readonly<StaticsBaseCoin>) {
     super(bitgo);
@@ -74,8 +75,8 @@ export class AvaxC extends BaseCoin {
     if (!staticsCoin) {
       throw new Error('missing required constructor parameter staticsCoin');
     }
-
     this._staticsCoin = staticsCoin;
+    this.ethLikeAvaxc = AvaxcRecovery.createInstance(bitgo, staticsCoin);
   }
 
   static createInstance(bitgo: BitGoBase, staticsCoin?: Readonly<StaticsBaseCoin>): BaseCoin {
@@ -577,6 +578,9 @@ export class AvaxC extends BaseCoin {
     if (_.isUndefined(params.recoveryDestination) || !this.isValidAddress(params.recoveryDestination)) {
       throw new Error('invalid recoveryDestination');
     }
+
+    // COIN-1708 : Recover method to be invoked from WRW
+    if (params.bitgoFeeAddress) return this.ethLikeAvaxc.recoverEthLikeforEvmBasedRecovery(params);
 
     // TODO (BG-56531): add support for krs
     const isUnsignedSweep = getIsUnsignedSweep(params);
--- a/modules/sdk-coin-avaxc/package.json
+++ b/modules/sdk-coin-avaxc/package.json
@@ -52,7 +52,9 @@
     "keccak": "^3.0.3",
     "lodash": "^4.17.14",
     "secp256k1": "5.0.0",
-    "superagent": "^9.0.1"
+    "superagent": "^9.0.1",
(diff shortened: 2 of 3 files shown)


kamleshmugdiya:
tests are failing

modules/sdk-coin-avaxc/src/EthLikeAvaxc.ts
```diff
@@ -0,0 +1,350 @@
...
+  async getGasPriceFromExternalAPI(): Promise<BN> {
+    try {
+      // COIN -1708 : hardcoded for half signing
+      const gasPrice = new BN(250000);
+      console.log(` Got hardcoded gas price: ${gasPrice}`);
+      return gasPrice;
+    } catch (e) {
+      throw new Error('Failed to get gas price');
+    }
+  }
```
kamleshmugdiya, line 338: same

kamleshmugdiya (changes requested):
this defeats the purpose of inheritance and code reusability.

nrjsuthar:
@kamleshmugdiya 
This PR https://github.com/BitGo/BitGoJS/pull/4925 had the change you mentioned, where the relationship exists. The changes in the current PR were because of the Gas and Limit, wherein all the parent methods are copied. The ideal change would be to extend avaxc class with abstractEthLikeNewCoin (mentioned PR), but removed as no one was sure of the change.

nrjsuthar:
Also, the gas price and limit from the explorer were not required here for the half sign.

modules/sdk-coin-avaxc/src/EthLikeAvaxc.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    const gasPriceMax = ethGasConfigs.maximumGasPrice;
+    const gasPriceMin = ethGasConfigs.minimumGasPrice;
+    if (userGasPrice < gasPriceMin || userGasPrice > gasPriceMax) {
+      throw new Error(`Gas price must be between ${gasPriceMin} and ${gasPriceMax}`);
+    }
+    return userGasPrice;
+  }
+
+  async recoveryBlockchainExplorerQuery(query: Record<string, any>): Promise<any> {
```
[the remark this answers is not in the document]
nrjsuthar: ok

kamleshmugdiya:
> @kamleshmugdiya This PR #4925 had the change you mentioned, where the relationship exists. The changes in the current PR were because of the Gas and Limit, wherein all the parent methods are copied. The ideal change would be to extend avaxc class with abstractEthLikeNewCoin (mentioned PR), but removed as no one was sure of the change.

I don't understand this, if the function has the same impl in AbstractEthLikeNewCoins, what is the need to override with the same code in the new class?

nrjsuthar:
> > @kamleshmugdiya This PR #4925 had the change you mentioned, where the relationship exists. The changes in the current PR were because of the Gas and Limit, wherein all the parent methods are copied. The ideal change would be to extend avaxc class with abstractEthLikeNewCoin (mentioned PR), but removed as no one was sure of the change.
> 
> I don't understand this, if the function has the same impl in AbstractEthLikeNewCoins, what is the need to override with the same code in the new class?

This was based on testing, for each of the issues, the method has been copied. 

Example:

A -> B -> C is the method flow and A has to be changed, now i we need to change the method A over here, we will need to override all its parents, the sub method will not be able to be called.

kamleshmugdiya:
> > > @kamleshmugdiya This PR #4925 had the change you mentioned, where the relationship exists. The changes in the current PR were because of the Gas and Limit, wherein all the parent methods are copied. The ideal change would be to extend avaxc class with abstractEthLikeNewCoin (mentioned PR), but removed as no one was sure of the change.
> > 
> > 
> > I don't understand this, if the function has the same impl in AbstractEthLikeNewCoins, what is the need to override with the same code in the new class?
> 
> This was based on testing, for each of the issues, the method has been copied.
> 
> Example:
> 
> A -> B -> C is the method flow and A has to be changed, now i we need to change the method A over here, we will need to override all its parents, the sub method will not be able to be called.

what testing?
and the methods that I pointed out, do they have any difference in implementation vs the AbstractEthLikeNewCoins?
they all looked the same to me.

modules/sdk-coin-avaxc/src/avaxc.ts
```diff
@@ -62,20 +62,21 @@ import {
   VerifyAvaxcTransactionOptions,
 } from './iface';
 import { AvaxpLib } from '@bitgo/sdk-coin-avaxp';
+import { EthLikeAvaxc } from './EthLikeAvaxc';
 
 export class AvaxC extends BaseCoin {
   static hopTransactionSalt = 'bitgoHopAddressRequestSalt';
-
   protected readonly _staticsCoin: Readonly<StaticsBaseCoin>;
+  private ethLikeAvaxc: any;
```
kamleshmugdiya, line 70: 1. Why is composition relationship required?
2. can you name this better EthLikeAvaxc?
just by name, I am not able to understand what is the diff between EthLikeAvaxc and AvaxC.

kamleshmugdiya:
pipeline failing again

PR #4947, opened 2024-09-26
build(sdk-coin-xrp): update ripple libs

hitansh-madan:
Ticket: WIN-3541

### DIFF
--- a/modules/sdk-coin-xrp/test/unit/xrp.ts
+++ b/modules/sdk-coin-xrp/test/unit/xrp.ts
@@ -3,7 +3,7 @@ import 'should';
 import { TestBitGo, TestBitGoAPI } from '@bitgo/sdk-test';
 import { BitGoAPI } from '@bitgo/sdk-api';
 import { Txrp } from '../../src/txrp';
-const ripple = require('../../src/ripple');
+import ripple from '../../src/ripple';
 
 import * as nock from 'nock';
 import assert from 'assert';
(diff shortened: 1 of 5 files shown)


modules/sdk-coin-xrp/test/unit/xrp.ts
```diff
@@ -91,7 +91,7 @@ describe('XRP:', function () {
       txHex:
         '{"TransactionType":"Payment","Account":"rBSpCz8PafXTJHppDcNnex7dYnbe3tSuFG","Destination":"rfjub8A4dpSD5nnszUFTsLprxu1W398jwc","DestinationTag":0,"Amount":"253481","Flags":2147483648,"LastLedgerSequence":1626225,"Fee":"45","Sequence":7}',
     });
-    unsignedExplanation.id.should.equal('CB36F366F1AC25FCDB38A19F17384ED3509D9B7F063520034597852FB10A1B45');
+    unsignedExplanation.id.should.equal('37486621138DFB0C55FEF45FD275B565254464651A04CB02EE371F8C4A84D8CA');
```
mullapudipruthvik, line 94: why are the values different ?
is this a breaking change ?

alebusse (dismissed):

modules/sdk-coin-xrp/test/unit/xrp.ts
hitansh-bitgo: `ripple-lib` 1.x only had the `computeBinaryTransactionHash` function that we used for both signed & unsigned txns, but `xrpl` 2.x has two separate functions to compute hashes for signed and unsigned transactions. 
The `hashSignedTx` returns the same hash as `computeBinaryTransactionHash` since both the functions are same but `hashTx`(used for getting hash of unsigned Tx) uses a different HashPrefix than `computeBinaryTransactionHash` & `hashSignedTx`, hence the change in these hashes.

It most likely wouldn't break anything on WP but technically still is a breaking change so I've updated commit message

DinshawKothari (approved):

PR #4978, opened 2024-10-02
chore(sdk-api): enable esModuleInterop

OttoAllmendinger:
Requires some changes to the imports in the codebase to use the default import
syntax.

Issue: BTC-1351

### DIFF
--- a/modules/sdk-api/src/api.ts
+++ b/modules/sdk-api/src/api.ts
@@ -2,12 +2,12 @@
  * @prettier
  */
 import Debug from 'debug';
-import * as eol from 'eol';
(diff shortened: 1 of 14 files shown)


davidkaplanbitgo (approved):

lzjluzijie (approved):

zahin-mohammad (changes requested):
The changes look good to me but, 
- lacking context on why we need this change 
- the linked ticket seems irrelevant to this change (maybe it is, its just not obvious to me)

OttoAllmendinger:
@zahin-mohammad fair point, opened #4991 with fixed issue number and improved description

PR #5010, opened 2024-10-09
feat(statics): add token and coins as available for Frankfurt and Sin…

baltiyal:
TICKET: COIN-1900

Ungating tokens and coins for Frankfurt and Singapore.
Also, removing Tomoe Token, as someone added it back while resolving conflicts.
(PR for Tomoe Replacement with Vic : https://github.com/BitGo/BitGoJS/pull/5005/files#diff-7212ff935627ebe65f0968af71f5f24a8eba5ff455ae548bbfcaff4ea691a337L7030)

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -205,6 +205,7 @@ const SOL_FEATURES = [
   CoinFeature.CUSTODY_BITGO_GERMANY,
   CoinFeature.CUSTODY_BITGO_FRANKFURT,
 ];
+const TSOL_FEATURES = [...SOL_FEATURES, CoinFeature.CUSTODY_BITGO_SINGAPORE];
 const SOL_TOKEN_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.TSS, CoinFeature.TSS_COLD];
 const STX_FEATURES = [
   ...AccountCoin.DEFAULT_FEATURES,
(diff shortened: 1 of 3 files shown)


modules/statics/src/coins.ts
```diff
@@ -338,6 +350,11 @@ const BERA_FEATURES = [
   CoinFeature.EVM_WALLET,
   CoinFeature.USES_NON_PACKED_ENCODING_FOR_TXDATA,
 ];
+const MOCA_FEATURES = [...ETH_FEATURES, CoinFeature.CUSTODY_BITGO_FRANKFURT];
+const CWBTC_FEATURES = [...ETH_FEATURES, CoinFeature.CUSTODY_BITGO_FRANKFURT];
+const TKX_FEATURES = [...ETH_FEATURES, CoinFeature.CUSTODY_BITGO_FRANKFURT];
+const MANA_FEATURES = [...ETH_FEATURES, CoinFeature.CUSTODY_BITGO_FRANKFURT];
+const ARBETH_ARB_FEATURES = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.CUSTODY_BITGO_FRANKFURT];
```
gianchandania, line 357: Do something like [this](https://github.com/BitGo/BitGoJS/blob/e06be4e59f36a9ae56da097e3a92197e328a6842/modules/statics/src/coins.ts#L312)
Just confirm if it will be DefaultFeatures or EthFeatures for the eth tokens.

modules/statics/src/coins.ts
```diff
@@ -205,6 +205,17 @@ const SOL_FEATURES = [
...
+  ...AccountCoin.DEFAULT_FEATURES,
+  CoinFeature.TSS,
+  CoinFeature.TSS_COLD,
+  CoinFeature.REQUIRES_RESERVE,
+  CoinFeature.SUPPORTS_TOKENS,
+  CoinFeature.STAKING,
+  CoinFeature.CUSTODY_BITGO_GERMANY,
+  CoinFeature.CUSTODY_BITGO_FRANKFURT,
+  CoinFeature.CUSTODY_BITGO_SINGAPORE,
+];
```
gianchandania, line 218: const TSOL_FEATURES = [...SOL_FEATURES, CoinFeature.CUSTODY_BITGO_SINGAPORE]

gianchandania (changes requested):
Tests seem to be failing as well

OttoAllmendinger (commented):
flush: failing tests

modules/statics/src/coins.ts
baltiyal: This way its extendable for ungating for other Trusts as well. For example, MANA needs to be ungated after sometime for CUSTODY_BITGO_SINGAPORE as well.

baltiyal: Done, have confirmed its EthFeatures for eth tokens, as EthFeatures is inheriting from DefaultFeatures.

modules/statics/src/coins.ts
baltiyal: Done

PR #5019, opened 2024-10-11
BTC-1553: fix isValidAddress returning false for valid cashaddr bch addresses

kcheng-bitgo:
TICKET: BTC-1553
Previously, `isValidAddress` verified the address by converting it to script and back. However, the previous implementation converts the script back to the legacy address and compares it against the original address. Thus, if the original was cashaddr format, it would fail. Fix this by generating addresses in all formats from the script and check if the original address is included.

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -415,10 +415,15 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       throw new Error('deprecated');
     }
 
-    const formats = param && param.anyFormat ? undefined : ['default' as const];
+    const formats = utxolib.addressFormat.addressFormats.filter((format) =>
+      utxolib.addressFormat.isSupportedAddressFormat(format, this.network)
+    );
     try {
       const script = utxolib.addressFormat.toOutputScriptTryFormats(address, this.network, formats);
-      return address === utxolib.address.fromOutputScript(script, this.network);
+      const genAddresses = formats.map((format) =>
+        utxolib.addressFormat.fromOutputScriptWithFormat(script, format, this.network)
+      );
+      return genAddresses.includes(address);
     } catch (e) {
       return false;
     }
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -415,24 +415,19 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       throw new Error('deprecated');
     }
 
-    // Default to anyFormat to true - if anyFormat is false then only check the default. Otherwise, check
-    // all formats
-    const anyFormat = (param as { anyFormat: boolean | undefined })?.anyFormat;
-    const formats =
-      anyFormat === undefined || anyFormat
-        ? utxolib.addressFormat.addressFormats.filter((format) =>
-            utxolib.addressFormat.isSupportedAddressFormat(format, this.network)
-          )
-        : ['default' as const];
-
+    // By default, allow all address formats.
+    // At the time of writing, the only additional address format is bch cashaddr.
+    const anyFormat = (param as { anyFormat: boolean } | undefined)?.anyFormat ?? true;
     try {
-      // To make sure that we are preserving `fromOutputScript(script, network) === address`, we need to do the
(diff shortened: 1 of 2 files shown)


davidkaplanbitgo (dismissed):

davidkaplanbitgo (dismissed):

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -415,10 +415,15 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
       throw new Error('deprecated');
     }
 
-    const formats = param && param.anyFormat ? undefined : ['default' as const];
```
OttoAllmendinger, line 418: So previously the `anyFormat` flag controlled this behavior. What you want to do here is make `anyFormat = true` the default.

OttoAllmendinger (changes requested):
the commit title / PR title should state what it does: with your change we return `true`

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -415,10 +415,15 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+    const formats = utxolib.addressFormat.addressFormats.filter((format) =>
+      utxolib.addressFormat.isSupportedAddressFormat(format, this.network)
+    );
     try {
       const script = utxolib.addressFormat.toOutputScriptTryFormats(address, this.network, formats);
-      return address === utxolib.address.fromOutputScript(script, this.network);
```
OttoAllmendinger, line 421: you can just remove this line - `toOutputScriptsTryFormats` will fail if it cannot parse the address with any of the formats

davidkaplanbitgo (dismissed):

pengyuc-bitgo (dismissed):

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -415,10 +415,24 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+    const anyFormat = (param as { anyFormat: boolean | undefined })?.anyFormat;
+    const formats =
+      anyFormat === undefined || anyFormat
+        ? utxolib.addressFormat.addressFormats.filter((format) =>
+            utxolib.addressFormat.isSupportedAddressFormat(format, this.network)
+          )
+        : ['default' as const];
+
     try {
+      // To make sure that we are preserving `fromOutputScript(script, network) === address`, we need to do the
```
OttoAllmendinger, line 429:
```suggestion
      // To make sure that we are preserving `fromOutputScriptWithFormat(script, format, network) === address`, we need to do the
```

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -415,10 +415,24 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
...
+      // circular check below. An example of something that this protects against is uppercase Bech32.
       const script = utxolib.addressFormat.toOutputScriptTryFormats(address, this.network, formats);
-      return address === utxolib.address.fromOutputScript(script, this.network);
+      const genAddresses = formats.map((format) =>
+        utxolib.addressFormat.fromOutputScriptWithFormat(script, format, this.network)
+      );
+      return genAddresses.includes(address);
```
OttoAllmendinger, line 435:
```
const [format, script] = utxolib.addressFormat.toOutputScriptAndFormat(address, network);
if (!anyFormat && format !== 'default') {
  return false;
}
return address === utxolib.addressFormat.fromOutputScriptWithFormat(script, format, network);
```

OttoAllmendinger (changes requested):
consider this implementation

```
// By default, allow all address formats.
// At the time of writing, the only additional address format is bch cashaddr.
const anyFormat = params.anyFormat ?? true;
try  {
  // Find out if the address is valid for any format. Tries all supported formats by default.
  // Throws if address cannot be decoded with any format.
  const [format, script] = utxolib.addressFormat.toOutputScriptAndFormat(address, network);
  // unless anyFormat is set, only 'default' is allowed.
  if (!anyFormat && format !== 'default') {
    return false;
  }
  // make sure that address is in normal representation for given format.
  return address === utxolib.addressFormat.fromOutputScriptWithFormat(script, format, network);
} catch (e) {
  return false
}
```

OttoAllmendinger (approved):

PR #5020, opened 2024-10-14
feat(statics): add deprecated feature to gteth

ranger2255:
Ticket: COIN-622

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -345,6 +345,8 @@ const BERA_FEATURES = [
   CoinFeature.BULK_TRANSACTION,
 ];
 const ETH_FEATURES_WITH_FRANKFURT = [...ETH_FEATURES, CoinFeature.CUSTODY_BITGO_FRANKFURT];
+const ZETA_FEATURES = [...COSMOS_SIDECHAIN_FEATURES, CoinFeature.CUSTODY_BITGO_SINGAPORE];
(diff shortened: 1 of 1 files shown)


modules/statics/src/base.ts
```diff
@@ -292,6 +292,10 @@ export enum CoinFeature {
    * This coin is onboarded on etheruem rollup chain
    */
   ETH_ROLLUP_CHAIN = 'eth-rollup-chain',
+  /**
+   * This coin is hidden
+   */
+  HIDDEN = 'hidden',
```
dpkjnr, line 298: Name is very ambiguous, when and where is it hidden?

dpkjnr (commented):
You haven't added hidden coin feature to any coin, will it be done in upcoming PRs?

ranger2255:
> You haven't added hidden coin feature to any coin, will it be done in upcoming PRs?

yes thats correct

modules/statics/src/base.ts
ranger2255: feature to be used from UI to hide a coin/token

margueriteblair: I'd agree and say name should be more explicit, or at least in the comment you can add `This coin is hidden from the UI`

ranger2255: As discussed with @dpkjnr, we ll go with the existing DEPRECATED coin feature to hide gteth from gas tank; 
steps:
1. create an alpha version with deprecated feature for gteth
2. use this version to fix tests in wp
3. attach evidences from (2) here and merge this PR
4. create a PR for wp with test fixes

PR #5025, opened 2024-10-14
fix: replace getFees rpc call with getFeeForMessage

hrishikeshjain:
TICKET: COIN-1879

`getFees` rpc method is deprecated used `getFeeForMessage` instead
https://www.quicknode.com/docs/solana/getFeeForMessage


Vijay-Jagannathan (approved):
lgtm, test the recovery flow E2E to make sure there's no regression

PR #5030, opened 2024-10-16
SDK changes for rune coin and statics

at31416:
Ticket: COIN-2037, COIN-2038

PR #5053, opened 2024-10-23
chore(static): update statics for rune

at31416:
Ticket: COIN-2037


modules/statics/src/coins.ts
```diff
@@ -1006,6 +1006,26 @@ export const coins = CoinMap.fromCoins([
     BaseUnit.TCOREUM,
     COSMOS_SIDECHAIN_FEATURES
   ),
+  account(
+    '9fa0f191-4eed-4030-864a-d14bbd98c8af',
+    'rune-native',
```
at31416, line 1011: calling this as `rune-native` because `rune` is already taken by another eth token.

mukeshsp (approved):

modules/statics/src/coins.ts
zahin-mohammad: aren't eth tokens in the format of `eth:tokenName` ?

modules/statics/test/unit/coins.ts
```diff
@@ -368,9 +368,11 @@ coins.forEach((coin, coinName) => {
     });
 
     if (!coin.isToken && coin.family !== CoinFamily.FIAT) {
-      it(`has expected network type`, function () {
-        coin.network.type.should.eql(coin.name === coin.family ? NetworkType.MAINNET : NetworkType.TESTNET);
-      });
+      if (coin.family !== CoinFamily.THOR) {
```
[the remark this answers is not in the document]
alebusse: So the family is Thor and the coin name is rune-native, maybe instead of ignoring the case, add a test for this case

modules/statics/src/coins.ts
```diff
@@ -1006,6 +1006,26 @@ export const coins = CoinMap.fromCoins([
...
+    'Rune',
+    Networks.main.rune,
+    8,
+    UnderlyingAsset.RUNE,
+    BaseUnit.RUNE,
+    COSMOS_SIDECHAIN_FEATURES
```
alebusse, line 1017: i dont think this is a cosmos like coin.

modules/statics/src/coins.ts
```diff
@@ -1006,6 +1006,26 @@ export const coins = CoinMap.fromCoins([
...
     COSMOS_SIDECHAIN_FEATURES
   ),
+  account(
+    '9fa0f191-4eed-4030-864a-d14bbd98c8af',
+    'rune-native',
+    'Rune',
+    Networks.main.rune,
+    8,
+    UnderlyingAsset.RUNE,
+    BaseUnit.RUNE,
```
alebusse, line 1016: i think the base unit is called `tor`
https://docs.thorchain.org/how-it-works/thorchain-name-service#fees

alebusse (changes requested):
left some comments

modules/statics/src/coins.ts
alebusse: eth tokens are always in `tokenName` format

mohitkh7:
Changes merged in PR: https://github.com/BitGo/BitGoJS/pull/5030

modules/statics/test/unit/coins.ts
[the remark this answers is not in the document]
bhavidhingra: This test fails for the coins where the coinName ("thorchain:rune") doesn't match with the class name (Rune). That's why it was required to exclude it from the test.

modules/statics/src/coins.ts
at31416: It is a cosmos like chain, confirmed with product team also with the foundation.

modules/statics/src/coins.ts
at31416: base unit is mentioned as tor in documents, but throchain apis always return the amount in `denom: rune`. Confirmed it with foundation team of thorchain

PR #5054, opened 2024-10-23
feat(abstract-cosmos): add message type for rune

at31416:
Also made registry protected as sdk rune needs to override a method which uses registry varaible.

Ticket: COIN-2038


mukeshsp (approved):

at31416:
changes merged as part of PR: https://github.com/BitGo/BitGoJS/pull/5030

PR #5059, opened 2024-10-24
feat(statics): adding ofchterc18dp, ofchterc6dp, ofctusds in coins.ts

Doddanna17:
Ticket: USDS-45

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -14452,6 +14452,49 @@ export const coins = CoinMap.fromCoins([
     undefined,
     'gteth'
   ),
+  tofcerc20(
+    '4144a64f-eacd-4df1-a482-72e9c0d976ff',
+    'ofchterc18dp',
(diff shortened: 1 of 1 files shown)


MohammedRyaan786 (approved):

modules/statics/src/coins.ts
```diff
@@ -14452,6 +14452,49 @@ export const coins = CoinMap.fromCoins([
...
+    undefined,
+    undefined,
+    undefined,
+    undefined,
+    'hteth'
+  ),
+  tofcerc20(
+    '5f83eaf5-9ba2-4aee-8d6a-b97bf2669edb',
+    'ofctusds',
+    'Test ERC Token 6 decimals TUSDS',
```
Phani024, line 14486: This should just be `Holesky Testnet USD Standard`

Doddanna17: Noted, will update

PR #5071, opened 2024-10-28
fix(sdk-core): pass correct walletVersion to wallet platform

dpkjnr:
Ticket: COIN-2021

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -2389,7 +2389,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
       );
     }
 
-    if (impliedForwarderVersion === 0 || impliedForwarderVersion === 3) {
+    if (impliedForwarderVersion === 0 || impliedForwarderVersion === 3 || impliedForwarderVersion === 6) {
       return true;
     } else {
       const ethNetwork = this.getNetwork();
--- a/modules/bitgo/test/v2/unit/wallets.ts
+++ b/modules/bitgo/test/v2/unit/wallets.ts
@@ -944,6 +944,120 @@ describe('V2 Wallets:', function () {
       });
     });
 
+    it(`should create a new hteth TSS MPCv2 wallet with walletVersion 6`, async function () {
+      const testCoin = bitgo.coin('hteth');
+      const stubbedKeychainsTriplet: KeychainsTriplet = {
+        userKeychain: {
+          id: '1',
+          commonKeychain: 'userPub',
+          type: 'tss',
+          source: 'user',
+        },
+        backupKeychain: {
+          id: '2',
+          commonKeychain: 'userPub',
+          type: 'tss',
+          source: 'backup',
+        },
+        bitgoKeychain: {
+          id: '3',
+          commonKeychain: 'userPub',
+          type: 'tss',
+          source: 'bitgo',
+        },
+      };
+      const stubCreateKeychains = sandbox
+        .stub(ECDSAUtils.EcdsaMPCv2Utils.prototype, 'createKeychains')
+        .resolves(stubbedKeychainsTriplet);
+
+      const walletNock = nock('https://bitgo.fakeurl')
+        .post(`/api/v2/hteth/wallet`, (body) => {
+          body.walletVersion.should.equal(6);
+          return true;
+        })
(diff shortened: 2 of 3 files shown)


zahin-mohammad (changes requested):
please add a test

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -982,7 +982,7 @@ export class Wallets implements IWallets {
         .result();
       const multisigTypeVersion =
         tssSettings.coinSettings[this.baseCoin.getFamily()]?.walletCreationSettings?.multiSigTypeVersion;
-      if (this.baseCoin.isEVM() && multisigTypeVersion === 'MPCv2') {
+      if (!walletVersion && this.baseCoin.isEVM() && multisigTypeVersion === 'MPCv2') {
         walletVersion = 5;
       }
```
pengyuc-bitgo, line 987: The codes before force the `walletVersion` to be 5. Now we will allow this `walletVersion` to be set by the user. If I understand it correctly, the walletVersion here can only be 5 or 6. Should we check this here? Or do you plan to validate that later in the backend?

dpkjnr: backend has the validation, and overriding user input doesn't look correct to me. Similar code is there in the UI also to decide the wallet version

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -982,8 +982,11 @@ export class Wallets implements IWallets {
         .result();
       const multisigTypeVersion =
         tssSettings.coinSettings[this.baseCoin.getFamily()]?.walletCreationSettings?.multiSigTypeVersion;
+
       if (this.baseCoin.isEVM() && multisigTypeVersion === 'MPCv2') {
-        walletVersion = 5;
+        if (!walletVersion || (walletVersion !== 5 && walletVersion !== 6)) {
+          walletVersion = 5;
+        }
```
mohammadalfaiyazbitgo, line 989: can we encapsulate this logic in a function, like `getWalletVersionForMPCv2` ? Just reading this snippet is confusing for someone without context.

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2389,7 +2389,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
       );
     }
 
-    if (impliedForwarderVersion === 0 || impliedForwarderVersion === 3) {
+    if (impliedForwarderVersion === 0 || impliedForwarderVersion === 3 || impliedForwarderVersion === 6) {
```
mohammadalfaiyazbitgo, line 2392: can you create a const or enums for `evm walletVersions` that support forwarder addresses. This functionality is not well documented and makes it hard for others to parse through and understand.

mohammadalfaiyazbitgo (changes requested):

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
dpkjnr: The forwarder version is smart contract-based, unlike versions 3 and 5, which use MPC-derived addresses. Creating an enum just for versions 3 and 5 seems excessive to me, especially since we're completely deprecating version 3.

mullapudipruthvik (approved):

modules/bitgo/test/v2/unit/wallets.ts
```diff
@@ -944,6 +944,120 @@ describe('V2 Wallets:', function () {
...
+
+      assert.ok(response.encryptedWalletPassphrase);
+      assert.ok(response.wallet);
+      assert.equal(
+        bitgo.decrypt({ input: response.encryptedWalletPassphrase, password: params.passcodeEncryptionCode }),
+        params.passphrase
+      );
+    });
+
+    it(`should create a new MPCv2 wallet with version 5 if walletVersion passed is not 5 or 6`, async function () {
```
zahin-mohammad, line 1004:
```suggestion
    it(`should create a new MPCv2 wallet with version 5 if walletVersion passed is 3`, async function () {
```

zahin-mohammad (approved):
lgtm

pengyuc-bitgo (approved):

mohammadalfaiyazbitgo (approved):

PR #5079, opened 2024-10-29
Parse and verify CoreDao OP_RETURN

davidkaplanbitgo:
Best to review commit-by-commit

[fix(utxo-coredao): clarifications from the coredao team](https://github.com/BitGo/BitGoJS/commit/56157b34e6802895489928d8e7a87f8f5c8129ed): Got clarifications from the CoreDao team and needed to make fixes

[feat(utxo-coredao): parse and verify op-return](https://github.com/BitGo/BitGoJS/commit/f32729df1d1f832811d7f537f0848400fbc1d9ec)

[refactor(utxo-coredao): rename from transaction to opReturn](https://github.com/BitGo/BitGoJS/commit/5b7c4d8f6d301f5d053e422cefd9c0829c40032f)


modules/utxo-coredao/src/opReturn.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+export function decodeTimelock(buffer: Buffer): number {
+  if (buffer.length !== 4) {
+    throw new Error('Invalid timelock buffer length');
+  }
+  return buffer.readUInt32LE();
+}
+
+export function encodeOpReturnLength(length: number): Buffer {
```
OttoAllmendinger, line 24: I'm almost positive that

```
bitcoin.script.compile([OPS.OP_RETURN, buffer]);
```

does the right thing and performs these calculations for you - please take a look


see here for instance: https://github.com/bitcoinjs/bitcoinjs-lib/blob/32e08aa57f6a023e995d8c4f0c9fbdc5f11d1fa0/ts_src/payments/embed.ts#L38

you could use `payments.p2data()` as well maybe

OttoAllmendinger (changes requested):

modules/utxo-coredao/src/opReturn.ts
OttoAllmendinger: in fact you should be able to use `payments.p2data()` to encode and decode `OP_RETURN` outputs

modules/utxo-coredao/src/opReturn.ts
```diff
@@ -0,0 +1,214 @@
...
+  ...rest
+}: OpReturnParams): Buffer {
+  /**
+   * As of v2, this is the construction of the OP_RETURN:
+   * Source: https://docs.coredao.org/docs/Learn/products/btc-staking/design#op_return-output
+   *
+   * The OP_RETURN output should contain all staking information in order, and be composed in the following format:
+   *
+   * OP_RETURN: identifier 0x6a
+   * LENGTH: which represents the total byte length after the OP_RETURN opcode. Note that all data has to be pushed with its appropriate size byte(s). [1]
```
OttoAllmendinger, line 60:
```suggestion
```

modules/utxo-coredao/src/opReturn.ts
```diff
@@ -0,0 +1,214 @@
...
+    return { ...baseParams, redeemScript: Buffer.from(dataBuffer.subarray(offset)) };
+  }
+}
+
+/**
+ * Check that the params are valid and that they make the given script
+ * @param script
+ * @param params
+ */
+export function verifyCoreDaoOpReturnOutputScript(script: Buffer, params: OpReturnParams): boolean {
```
OttoAllmendinger, line 206: this hides the exception details which can be crucial

it looks like this is just a testutil at the moment so just move it there

if you think it is worth in the main code please change it to

```suggestion
export assertStandardCoreDaoOpReturnOutputScript(script: Buffer, params: OpReturnParams): void {
```

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #5082, opened 2024-10-30
feat: update public-types

zahin-mohammad:
This pr updates public-types and io-ts. 

The latest version of public-types is only compatible with the bitgo fork of io-ts.

I had to also set a resolution to a polkadot lib to fix some compilation issues. It's not directly related to this PR, but gets caught here because the `package updates` CI workflow only runs when you change a `package.json` file.

TICKET: [WP-2957](https://bitgoinc.atlassian.net/browse/WP-2957)

[WP-2957]: https://bitgoinc.atlassian.net/browse/WP-2957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.gitignore
```diff
@@ -11,3 +11,4 @@ lab/
 .vscode/
 .env
 .yarn
+modules/**/dist/
```
zahin-mohammad, line 14: Suprised that each coin module had to define this explicitly.... 

Only an issue now because the new coredao module doesn't have a gitignore, so I keep getting a `dist` in my un-staged files.

package.json
```diff
@@ -81,7 +81,8 @@
     "ua-parser-js": ">0.7.30 <0.8.0",
     "secp256k1": "5.0.1",
     "socks": "2.7.3",
-    "web3-utils": "4.2.1"
+    "web3-utils": "4.2.1",
+    "@polkadot/api": "14.1.1"
```
zahin-mohammad, line 85: https://github.com/paritytech/txwrapper-core/blob/main/packages/txwrapper-core/package.json

has a dep on `"@polkadot/api": "^14.0.1",` 

14.1.1 is fine, but 14.2.1 starts breaking stuff. 

Simply upgrading our stack to 14.2.1 didn't fix things. Maybe it will but the coins team will need to investigate the changes that need to be made.

package.json
```diff
@@ -121,5 +122,6 @@
   },
   "dependencies": {
     "terser": "^5.14.2"
-  }
+  },
+  "packageManager": "yarn@1.22.22"
```
zahin-mohammad, line 126: I got this from the version printed out in the `CI` workflow.

.yarnrc
```diff
@@ -0,0 +1 @@
+registry "https://registry.npmjs.org"
```
zahin-mohammad, line 1: Prevents huge yarn lock diffs for the url changing.

package.json
alebusse: its usually some combination between all the libs that belong to polkadot, they use ^ internally and do breaking changes with minor versions, causing any "compatible" lib upgrades not really compatible

alebusse (approved):

davidkaplanbitgo (approved):

noel-bitgo (approved):

PR #5084, opened 2024-10-31
feat: retrofit during GG18 user and backup recovery

zahin-mohammad:
This PR removes the need for openssl when recovering from GG18 wallets

For any GG18 wallet recovering with a user and backup key, we will instead perform a 2of2 retrofit to MPCv2, and sign with MPCv2.

TICKET: WP-2959

### DIFF
--- a/modules/abstract-cosmos/src/lib/iface.ts
+++ b/modules/abstract-cosmos/src/lib/iface.ts
@@ -30,7 +30,6 @@ export interface SendMessage {
 export interface RecoveryOptions {
   userKey?: string; // Box A
   backupKey?: string; // Box B
-  bitgoKey: string; // Box C
   recoveryDestination: string;
   krsProvider?: string;
   walletPassphrase?: string;
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
@@ -1261,22 +1261,131 @@ export function isGG18SigningMaterial(keyShare: string, walletPassphrase: string
 }
 
 /**
+ * Get the MPC v2 recovery key shares from the provided user and backup key shares.
+ * @param encryptedUserKey encrypted gg18 or MPCv2 user key
+ * @param encryptedBackupKey encrypted gg18 or MPCv2 backup key
+ * @param walletPassphrase password for user and backup key
+ * @returns MPC v2 recovery key shares
+ */
+export async function getMpcV2RecoveryKeyShares(
+  encryptedUserKey: string,
+  encryptedBackupKey: string,
+  walletPassphrase?: string
+): Promise<{
+  userKeyShare: Buffer;
+  backupKeyShare: Buffer;
+  commonKeyChain: string;
+}> {
+  if (isGG18SigningMaterial(encryptedUserKey, walletPassphrase)) {
+    return getMpcV2RecoveryKeySharesFromGG18(encryptedUserKey, encryptedBackupKey, walletPassphrase);
+  }
+  return getMpcV2RecoveryKeySharesFromReducedKey(encryptedUserKey, encryptedBackupKey, walletPassphrase);
+}
+
+/**
+ * Signs a message hash using MPC v2 recovery key shares.
+ *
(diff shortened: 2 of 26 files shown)


modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -1340,3 +1324,143 @@ export async function signRecoveryMpcV2(
...
+  const MPC = new Ecdsa();
+
+  const userKeyCombined = MPC.keyCombine(userSigningMaterial.pShare, [
+    userSigningMaterial.bitgoNShare,
+    userSigningMaterial.backupNShare,
+  ]);
+  const backupKeyCombined = MPC.keyCombine(backupSigningMaterial.pShare, [
+    backupSigningMaterial.userNShare,
+    backupSigningMaterial.bitgoNShare,
+  ]);
```
zahin-mohammad, line 1456: Note: no derivation path is applied here.

modules/abstract-cosmos/src/lib/iface.ts
```diff
@@ -30,7 +30,6 @@ export interface SendMessage {
 export interface RecoveryOptions {
   userKey?: string; // Box A
   backupKey?: string; // Box B
-  bitgoKey: string; // Box C
```
zahin-mohammad, line 33: This was supplied just to get the common keychain....

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -1340,3 +1324,143 @@ export async function signRecoveryMpcV2(
...
+  let userPrv;
+  try {
+    backupPrv = sjcl.decrypt(walletPassphrase, encryptedGG18BackupKey);
+    userPrv = sjcl.decrypt(walletPassphrase, encryptedGG18UserKey);
+  } catch (e) {
+    throw new Error(`Error decrypting backup keychain: ${e.message}`);
+  }
+
+  const userSigningMaterial = JSON.parse(userPrv) as ECDSAMethodTypes.SigningMaterial;
+  const backupSigningMaterial = JSON.parse(backupPrv) as ECDSAMethodTypes.SigningMaterial;
```
mohammadalfaiyazbitgo, line 1437: can you not use io-ts to safely parse these?

zahin-mohammad: good point, we could. I'd rather do this in a separate PR though, this PR is primarily moving around the same functions without changing their behaviour too much.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
islamaminBitGo: signRecoveryMpcv2 signs for m/0 all the time, which makes this work. However, I think that function's name should probably change to reflect that behaviour (signRecoveryMpcv2ForBaseAddress) or a parameter could be added to allow different derivation paths to be passed with a default of m/0.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -1261,46 +1261,25 @@ export function isGG18SigningMaterial(keyShare: string, walletPassphrase: string
...
 /**
- * Retrieves the MPC v2 recovery key shares from the provided user and backup key shares.
- *
- * @param {string} userPublicOrPrivateKeyShare
- * @param {string} backupPrivateOrPublicKeyShare
- * @param {string} [walletPassphrase] - The passphrase used to decrypt the key shares
- * @returns {Promise<{ userKeyShare: KeyShare, backupKeyShare: KeyShare, commonKeyChain: string }>}
- *
- * @async
+ * Get the MPC v2 recovery key shares from the provided user and backup key shares.
```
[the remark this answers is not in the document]
islamaminBitGo: LGTM, just a nit.

noel-bitgo (approved):

PR #5085, opened 2024-10-31
feat(sdk-core): update prebuildSelfManagedStakingTransaction

noel-bitgo:
- for TSS transactions we need to delete shares (similar thing done in the build function)
- we only care about the expanded params if they exists

TICKET: SC-498

### DIFF
--- a/modules/sdk-core/src/bitgo/staking/stakingWallet.ts
+++ b/modules/sdk-core/src/bitgo/staking/stakingWallet.ts
@@ -218,11 +218,17 @@ export class StakingWallet implements IStakingWallet {
 
   /**
    * Create prebuilt staking transaction.
+   *
+   * for transactions with tx request id (TSS transactions), we need to delete signature shares before creating prebuild transaction
+   * we only need to get transaction build params if they exist to pre build
+   *
    * @param transaction
    */
   async prebuildSelfManagedStakingTransaction(transaction: StakingTransaction): Promise<PrebuildTransactionResult> {
-    const builtStakingTransaction = await this.build(transaction);
-    const buildParams = builtStakingTransaction.transaction.buildParams;
+    if (transaction.txRequestId) {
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/staking/stakingWallet.ts
```diff
@@ -221,16 +221,20 @@ export class StakingWallet implements IStakingWallet {
    * @param transaction
    */
   async prebuildSelfManagedStakingTransaction(transaction: StakingTransaction): Promise<PrebuildTransactionResult> {
-    const builtStakingTransaction = await this.build(transaction);
-    const buildParams = builtStakingTransaction.transaction.buildParams;
+    let buildParams: PrebuildTransactionOptions | undefined;
+    if ((this.wallet.baseCoin.supportsTss() && this.wallet.baseCoin.getFamily() !== 'eth') || this.isEthTss) {
```
zahin-mohammad, line 225: The description lists `eth like` but we are only checking eth here. Can we generalize this and make it less coin specific?

zahin-mohammad: Also a comment on why we need this differentiation in the code would be helpful

zahin-mohammad (changes requested):
Please add some test(s)

modules/sdk-core/src/bitgo/staking/stakingWallet.ts
noel-bitgo: yeah i meant to write multisig instead of eth like in commit message.

we are checking eth specifically because we have eth multisig and eth tss, so we need 2 separate checks as far as i understand.

noel-bitgo: changed the logic

noel-bitgo:
> Please add some test(s)

done

zahin-mohammad (approved):

PR #5103, opened 2024-11-06
Add support for commonkeychain in TSS recovery for eth-like coins

simhonc:
## Description

Link to Jira: https://bitgoinc.atlassian.net/browse/WP-2971

This is to prevent the error "Unsupported public key" when trying to create an unsigned sweep for a TSS SMC wallet (eth-like), and to allow the user to go ahead with creating their unsigned transaction. 

Type of change: Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

These SDK changes have been tested in wallet-recovery-wizard by modifying in node_modules. TODO: use the SDK beta version from this PR to test. 

# Checklist:

### DIFF
--- a/modules/abstract-eth/package.json
+++ b/modules/abstract-eth/package.json
@@ -59,6 +59,7 @@
   },
   "devDependencies": {
     "@bitgo/sdk-api": "^1.56.1",
-    "@bitgo/sdk-test": "^8.0.48"
+    "@bitgo/sdk-test": "^8.0.48",
+    "@bitgo/sdk-lib-mpc": "^10.1.0"
   }
 }
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -68,6 +68,7 @@ import {
   TransactionBuilder,
   TransferBuilder,
 } from './lib';
+import { getDerivationPath } from '@bitgo/sdk-lib-mpc';
 
 /**
  * The prebuilt hop transaction returned from the HSM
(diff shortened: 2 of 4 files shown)


modules/bitgo/test/v2/lib/recovery-nocks.ts
```diff
@@ -263,6 +263,30 @@ const nockEthData: any[] = [
       result: '2200000000000000000',
     },
   },
+  {
+    params: {
+      module: 'account',
+      action: 'txlist',
+      address: '0xd18ee37daeab8468020dd65fa7f6f6b9ac741584',
```
simhonc, line 270: This address doesn't match up with the wallet base address or the recipient address... is this correct? I got this just from seeing the error message ("no match for request with URL "https://api-holesky.etherscan.io/api?module=account&action=txlist&address=0xd18ee37daeab8468020dd65fa7f6f6b9ac741584"), but I don't know if this is the proper way to do it...


In addition, even with this change, it shows me the error: 
```
Error: Nock: No match for request {
  "method": "GET",
  "url": "https://api-holesky.etherscan.io/api?module=account&action=txlist&address=0xd18ee37daeab8468020dd65fa7f6f6b9ac741584",
  "headers": {
    "accept-encoding": "gzip, deflate"
  }
}
```

which I thought adding this would be the proper way to fix, but it doesn't... is this the proper way to make a new nock, or is there something else I need to add?

simhonc: resolved, turns out that 1. for TSS the wallet address is retrieved from the commonkeychain itself, and 2. i had created the nock at the wrong place.

PR #5106, opened 2024-11-07
fix(sdk-core): choose correct default apiVersion for tss

zahin-mohammad:
TICKET: WP-3038

### DIFF
--- a/modules/bitgo/test/v2/unit/coins/utxo/prebuildAndSign.ts
+++ b/modules/bitgo/test/v2/unit/coins/utxo/prebuildAndSign.ts
@@ -96,17 +96,23 @@ function run(coin: AbstractUtxoCoin, inputScripts: ScriptType[], txFormat: TxFor
     feeMultiplier?: number;
     selfSend?: boolean;
     nockOutputAddresses?: boolean;
+    txFormat?: string;
   }): nock.Scope[] {
     const nocks: nock.Scope[] = [];
 
     // Nock the prebuild route (/tx/build, blockheight)
+    const expected_params = {
+      recipients: [params.recipient],
+      rbfTxIds: params.rbfTxIds,
+      feeMultiplier: params.feeMultiplier,
+      changeAddressType: ['p2trMusig2', 'p2wsh', 'p2shP2wsh', 'p2sh', 'p2tr'],
+    };
+    if (params.txFormat) {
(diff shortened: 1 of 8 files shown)


modules/bitgo/test/v2/unit/coins/utxo/prebuildAndSign.ts
```diff
@@ -212,6 +218,7 @@ function run(coin: AbstractUtxoCoin, inputScripts: ScriptType[], txFormat: TxFor
 
     [true, false].forEach((useWebauthn) => {
       it(`should succeed with ${useWebauthn ? 'webauthn encryptedPrv' : 'encryptedPrv'}`, async function () {
+        const txCoins = ['tzec', 'zec', 'ltc', 'bcha', 'doge', 'dash', 'btg', 'bch'];
```
margueriteblair, line 221: Should this not include `btc`?

modules/bitgo/test/v2/unit/coins/utxo/prebuildAndSign.ts
```diff
@@ -257,6 +265,7 @@ function run(coin: AbstractUtxoCoin, inputScripts: ScriptType[], txFormat: TxFor
       it(`should be able to build, sign, & verify a replacement transaction with selfSend: ${selfSend}`, async function () {
         const rbfTxIds = ['tx-to-be-replaced'],
           feeMultiplier = 1.5;
+        const txCoins = ['tzec', 'zec', 'ltc', 'bcha', 'doge', 'dash', 'btg', 'bch'];
```
margueriteblair, line 268: If we're gonna reuse this `txCoins`, can we make it global?

margueriteblair (commented):
Most tests and changes look good - just a few nitty q's

OttoAllmendinger (approved):
not sure why the utxo change was necessary

alebusse:
> not sure why the utxo change was necessary

Thanks for reviewing @OttoAllmendinger , the test change was necessary because the default wallet type is now 'hot' and it changes the prebuild params for utxos here
https://github.com/BitGo/BitGoJS/blob/98b8f31f9933cbf05716a1b83fb20b8e67925e3f/modules/abstract-utxo/src/abstractUtxoCoin.ts#L1084 and the nocks didnt match after that change.

alebusse (approved):

PR #5125, opened 2024-11-12
feat(statics): add new batch of tokens COIN-2227

ravibitgo:
Ticket: [COIN-2227](https://bitgoinc.atlassian.net/browse/COIN-2227)

[COIN-2227]: https://bitgoinc.atlassian.net/browse/COIN-2227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -22,6 +22,7 @@ export enum CoinKind {
 export enum CoinFamily {
   ADA = 'ada',
   ALGO = 'algo',
+  APT = 'apt',
   ARBETH = 'arbeth',
   ATOM = 'atom',
   AVAXC = 'avaxc',
@@ -2351,6 +2352,7 @@ export enum KeyCurve {
  */
 export enum BaseUnit {
   ATOM = 'uatom',
+  APT = 'octa',
   ETH = 'wei',
   BTC = 'satoshi',
   BSC = 'jager',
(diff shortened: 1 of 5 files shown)


modules/statics/src/base.ts
```diff
@@ -761,6 +761,10 @@ export enum UnderlyingAsset {
   EQO = 'eqo',
   ESE = 'ese',
   ETA = 'eta',
+  'eth:cusdo' = 'eth:cusdo',
+  'eth:edu' = 'eth:edu',
+  'eth:tlos' = 'eth:tlos',
+  'eth:usdg' = 'eth:usdg',
```
alebusse, line 767: I think for eth is usually like this
```suggestion
  'cusdo' = 'cusdo',
  'edu' = 'edu',
  'tlos' = 'tlos',
  'usdg' = 'usdg',
```

ravibitgo: There's a collision between pTokens TLOS & the new Telos TLOS token, that's why for eth type tokens too from now on we will be following `eth:<token_symbol>`. It is mentioned in the ticket [COIN-2227](https://bitgoinc.atlassian.net/browse/COIN-2227)

ravibitgo: We are checking this on testnet first, since this is the first time we're using `eth:` prefix for eth type tokens

PR #5115, opened 2024-11-10
feat(sdk-coin-apt): new coin generation, apt skeleton

baltiyal:
New Aptos SDK Addition : sdk-coin-apt
Addition of Skeleton Code for Apt Coin to support further implementation
TICKET: COIN-2220

PR #5129, opened 2024-11-13
Add Apt Coin to Statics

baltiyal:
Ticket : COIN-2257

### DIFF
--- a/modules/bitgo/test/browser/browser.spec.ts
+++ b/modules/bitgo/test/browser/browser.spec.ts
@@ -33,8 +33,6 @@ describe('Coins', () => {
       XrpToken: 1,
       Rune: 1,
       Trune: 1,
-      Apt: 1,
-      Tapt: 1,
     };
     Object.keys(BitGoJS.Coin)
       .filter((coinName) => !excludedKeys[coinName])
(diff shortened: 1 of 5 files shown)


modules/bitgo/test/browser/browser.spec.ts
```diff
@@ -33,8 +33,6 @@ describe('Coins', () => {
       XrpToken: 1,
       Rune: 1,
       Trune: 1,
-      Apt: 1,
```
at31416, line 36: you will be adding these at the time implementation of apt sdk?

ananth99 (approved):

modules/bitgo/test/browser/browser.spec.ts
baltiyal: Added this in this PR: https://github.com/BitGo/BitGoJS/pull/5115/files#diff-f8b260fd3dd9567a3a5c8d55b54d988b7428d1d7f1fb7c9f7e5ede728a4a94faR36
As the browser test case was failing without statics, so had to put this coin in excluded list.

bhavidhingra (approved):

PR #5155, opened 2024-11-20
fix: validate addresses created matches the requested format

lcovar:
Ticket: BTC-1632

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -29,6 +29,7 @@ import {
   AddressTypeChainMismatchError,
   BaseCoin,
   BitGoBase,
+  CreateAddressFormat,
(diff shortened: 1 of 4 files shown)


margueriteblair (approved):

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -1220,7 +1223,7 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
     );
 
     return {
-      address,
+      address: this.canonicalAddress(address, params.format),
```
pranavjain97, line 1226: what happens if `params.format` is `undefined` here?

lcovar: It will not do any address format conversion.

davidkaplanbitgo (approved):

PR #5159, opened 2024-11-21
fix(sdk-coin-etc): Unsigned sweep error message

sampras-saha:
Previously error key was not read from the rpc response object,
resulting in showing undefined to the user.

Now it is properly reading **error** key from the response object and
displaying proper error message to the user.

TICKET: COIN-2221

### DIFF
--- a/modules/sdk-coin-etc/src/etc.ts
+++ b/modules/sdk-coin-etc/src/etc.ts
@@ -301,10 +301,18 @@ export class Etc extends AbstractEthLikeCoin {
       params: [address, 'latest'],
       id: 1,
     });
-    // throw if the result does not exist or the result is not a valid number
-    if (!result || !result.result || isNaN(result.result)) {
-      throw new Error(`Could not obtain address balance for ${address} from etc.network, got: ${result.result}`);
+
+    // throw if the result object does not exist
+    if (!result) {
+      throw new Error(`Could not obtain address balance for ${address} from etc.network, got: Empty object response`);
+    } else if (result.error) {
+      // throw if result.error exists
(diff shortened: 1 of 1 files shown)


mukeshsp (dismissed):

modules/utxo-bin/package.json
```diff
@@ -29,7 +29,7 @@
     "@bitgo/blockapis": "^1.10.6",
     "@bitgo/statics": "^50.8.0",
     "@bitgo/utxo-lib": "^11.0.1",
-    "@bitgo/unspents": "^0.47.13",
+    "@bitgo/unspents": "^0.47.14",
     "@bitgo/wasm-miniscript": "^1.8.0",
```
mukeshsp, line 33: Please remove conflict

modules/utxo-bin/package.json
```diff
@@ -29,7 +29,7 @@
     "@bitgo/blockapis": "^1.10.6",
     "@bitgo/statics": "^50.8.0",
     "@bitgo/utxo-lib": "^11.0.1",
-    "@bitgo/unspents": "^0.47.13",
+    "@bitgo/unspents": "^0.47.14",
```
venkateshv1266, line 32: If you are updating the dependency, please add `package-lock.json` file as well.

sampras-saha: As discussed with @venkateshv1266, the `package-lock.json` is not required as it is an [untracked file](https://github.com/BitGo/BitGoJS/blob/1fb05dfecd51bc2916fa2d13111afde81ed89935/.gitignore)

modules/utxo-bin/package.json
sampras-saha: Resolved

venkateshv1266 (approved):

PR #5187, opened 2024-11-29
fix: typecheck some tests in `modules/bitgo/`

OttoAllmendinger:
- **move failing tests**
  

- **refactor(bitgo): move broken tests to test/legacy**
  For unclear reasons, the `test/` folder is often excluded from `tsconfig.json`,
  which leads to tests that do not compile correctly.
  
  This commit quarantines these existing tests in `modules/bitgo` to
  `test/legacy`, allowing a fresh start in `test/` with type-checked tests.
  
  Issue: BTC-1450
  

- **feat(bitgo): add README to test/legacy**
  Issue: BTC-1450
  

- **fix(bitgo): include test files in default `tsconfig.json`**
  We should in fact build the tests.
  
  They will not be packaged due to the corresponding lines in `.npmignore`.
  
  Issue: BTC-1450
  

- **refactor(bitgo): move utxo tests to test/utxo**
  The utxo tests do in fact typecheck, so we can move them out of `test/legacy/`
  
  Issue: BTC-1450


modules/bitgo/test/tsconfig.json
OttoAllmendinger, line 1: this file was pretty pointless - I believe it provided IDE hints which were ignored

PR #5232, opened 2024-12-09
fix: use max spendable amount to sweep balance

hrishikeshjain:
TICKET: COIN-2432

Currently `/sweep` api endpoint throws insufficent funds error because while building transaction we check for amount + fee and amount that we were sending was spendable balanceso it will always fail . So updated it to maxSpendable balance which returns amount subtrcating fee.

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -937,19 +937,23 @@ export class Wallet implements IWallet {
           'cannot sweep when unconfirmed funds exist on the wallet, please wait until all inbound transactions confirm'
         );
       }
-
-      const value = this.spendableBalanceString();
-      if (_.isUndefined(value) || value === '0') {
+      const value = await this.bitgo.get(this.url('/maximumSpendable')).result();
+      const maximumSpendable = new BigNumber(value.maximumSpendable);
+      if (value !== undefined || maximumSpendable.isZero()) {
         throw new Error('no funds to sweep');
       }
-      (params as any).recipients = [
-        {
-          address: params.address,
-          amount: value,
-        },
-      ];
 
-      return this.sendMany(params);
+      const sendManyParams: SendManyOptions = {
+        ...params,
+        recipients: [
+          {
+            address: params.address || '', // Ensure address is always a string
+            amount: maximumSpendable.toString(),
+          },
+        ],
+      };
+
+      return this.sendMany(sendManyParams);
     }
     // the following flow works for all UTXO coins
 
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -1790,6 +1790,41 @@ describe('V2 Wallet:', function () {
     });
   });
 
+  describe('sweep wallet', function () {
+    let basecoin;
+    let wallet;
+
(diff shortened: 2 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -938,14 +938,15 @@ export class Wallet implements IWallet {
         );
       }
 
-      const value = this.spendableBalanceString();
-      if (_.isUndefined(value) || value === '0') {
+      const value = await this.maximumSpendable();
```
gianchandania, line 941: Add tests for the change

hrishikeshjain: It looks like we cant write test case for sweep functionality, @gianchandania any idea if theres a way?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -938,14 +938,15 @@ export class Wallet implements IWallet {
...
+      const maximumSpendable = new BigNumber(value.maximumSpendable).toString();
+      if (_.isUndefined(value) || maximumSpendable === '0') {
         throw new Error('no funds to sweep');
       }
       (params as any).recipients = [
         {
           address: params.address,
-          amount: value,
+          amount: maximumSpendable,
```
mohammadalfaiyazbitgo, line 949: you don't need `lodash` to do a simple `!== undefined` check. Also why convert `maximumSpendable` to a string to perform a numerical check? Why not https://mikemcl.github.io/bignumber.js/#isPos ?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -938,14 +938,15 @@ export class Wallet implements IWallet {
...
       }
 
-      const value = this.spendableBalanceString();
-      if (_.isUndefined(value) || value === '0') {
+      const value = await this.maximumSpendable();
+      const maximumSpendable = new BigNumber(value.maximumSpendable).toString();
+      if (_.isUndefined(value) || maximumSpendable === '0') {
         throw new Error('no funds to sweep');
       }
       (params as any).recipients = [
```
mohammadalfaiyazbitgo, line 946: nit: if you can, could you please refactor this to avoid `any` ?

mohammadalfaiyazbitgo (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
mohammadalfaiyazbitgo: > It looks like we cant write test case for sweep functionality

Can you elaborate on why?

modules/sdk-core/src/bitgo/wallet/wallet.ts
hrishikeshjain: updated 👍

modules/sdk-core/src/bitgo/wallet/wallet.ts
hrishikeshjain: Added tests 😄

hrishikeshjain:
@mohammadalfaiyazbitgo can I get a re-review please

mohammadalfaiyazbitgo (approved):

PR #5234, opened 2024-12-10
Win 3948 optimized onboarding of tokens

rv-bitgo:
## A slightly better approach to managing coins.ts file.

coins.ts has already ~20,000 lines of code config which is bound to only grow with time if we keep adding new tokens/coins to it on every onboarding. This comes with the pro of easily putting all tokens in one place but comes at a cost of easily messing up the entire flow by a single mistake.

I propose any new onboarded coins be put into smaller files that can be unique to their assets (erc20, sol and so on.).
Please check this out and let me know what do you guys think of this structure and any potential issues that might come with this approach.


parasgarg-bitgo (dismissed):

lcovar (dismissed):

modules/statics/src/coins/erc20Coins.ts
```diff
@@ -0,0 +1,44 @@
+import { UnderlyingAsset } from '../base';
```
zahin-mohammad, line 1: Is this meant for all erc20's or just eth erc20's?

rv-bitgo: Intention-wise, we will be using this file for any eth erc20 tokens. Any already onboarded tokens can be migrated to this structure in future depending on resource availability.

modules/statics/src/coins/erc20Coins.ts
```diff
@@ -0,0 +1,44 @@
...
+    'Kava.io',
+    6,
+    '0x0c356b7fd36a5357e5a017ef11887ba100c9ab76',
+    UnderlyingAsset['eth:kava']
+  ),
+  erc20(
+    '0b22028a-fa33-47e3-9b4d-b32173b53ab6',
+    'eth:iq',
+    'Everipedia IQ',
+    8,
```
yogeshwar-bitgo, line 16: https://etherscan.io/token/0x579cea1889991f68acc35ff5c3dd0621ff29b0c9
It should be 18 decimals as per link

rv-bitgo: Good catch. Thank you for noticing this. 🙇

mohd-kashif (approved):

modules/statics/src/coins/erc20Coins.ts
parasgarg-bitgo: If we want to keep this file for only eth erc20 tokens, please rename it to something like `ethErc20Coins`

rv-bitgo: Since ERC-20 tokens are built on the Ethereum blockchain, i don't think we need to explicitly mention eth. erc20 should be self-explanatory. 
Please update me if my understanding is wrong about all erc20 tokens being eth tokens.

rv-bitgo: Once this PR is merged and the entire flow is tested on prod and we have taken care of concerns like how would the validation be done(easier rn to just go through one file) and do we need to update the onboarding docs etc, we can focus on migrating the existing ones.
I'll take the responsibility to create a tech-debt ticket in future once.

PR #5238, opened 2024-12-10
fix(abstract-eth): fix issue with ethlike signTx

gianchandania:
In case of signing a custodial tx with contract data, we were getting transaction data does not match original because halfSigned tx was not having data field. This PR fixes the same. Taking recipients from params similar to [eth](https://github.com/BitGo/BitGoJS/blob/acaeff7bef27d6dd48075dbd1b27ecee526bb851/modules/sdk-coin-eth/src/eth.ts#L454)

COIN-2506
TICKET: COIN-2506


noel-bitgo (dismissed):
Any test that would validate the output?

gianchandania:
> Any test that would validate the output?

added the test

noel-bitgo (approved):

PR #5270, opened 2024-12-13
refactor(abstract-utxo): move signTransaction to its own file

OttoAllmendinger:
Issue: BTC-1450

### DIFF
--- a/modules/abstract-utxo/src/abstractUtxoCoin.ts
+++ b/modules/abstract-utxo/src/abstractUtxoCoin.ts
@@ -2,8 +2,7 @@ import assert from 'assert';
 import { randomBytes } from 'crypto';
 import _ from 'lodash';
 import * as utxolib from '@bitgo/utxo-lib';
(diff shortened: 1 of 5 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2147,10 +2147,7 @@ export class Wallet implements IWallet {
     const signingParams = {
       ...params,
       txPrebuild,
-      wallet: {
-        // this is the version of the multisig address at wallet creation time
-        addressVersion: this._wallet.coinSpecific.addressVersion,
```
OttoAllmendinger, line 2152: this was used here, but the current `algo` code doesn't use it anymore https://github.com/BitGo/BitGoJS/commit/7a02a19745035ec6851a1eaf478249c59be86068#diff-8a72caea304a885f750f24d950ccf6a0e5a8372d9f5c941d1856154d029ce6caR140

davidkaplanbitgo (approved):

PR #5304, opened 2024-12-23
feat(statics): add new batch of tokens, COIN-2524

akshay-bitgo:
Ticket: COIN-2524

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -1792,6 +1792,10 @@ export enum UnderlyingAsset {
   'eth:iris' = 'eth:iris',
   'eth:hard' = 'eth:hard',
   'eth:hegic' = 'eth:hegic',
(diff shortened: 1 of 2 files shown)


yogeshwar-bitgo (approved):
LGTM

rv-bitgo:
Looks good but please refer to #5234 once. Idea is not to append raw info to coins.ts anymore. There is a dedicated file `modules/statics/src/coins/erc20Coins.ts` where erc20 tokens info could be put. It's already being appened to the coins.ts array.

akshay-bitgo:
> Looks good but please refer to #5234 once. Idea is not to append raw info to coins.ts anymore. There is a dedicated file `modules/statics/src/coins/erc20Coins.ts` where erc20 tokens info could be put. It's already being appened to the coins.ts array.

I think it would be better if we fully refactor this, otherwise this is creating confusion.

PR #5305, opened 2024-12-23
feat(sdk-coin-bera): add TSS Suport

adarsh312:
Ticket: WIN-4073


dpkjnr (commented):
there should be some tests around DKLS and walletVersion 5, can you add bera also there. This coinfeature also needs to be added `CoinFeature.MPCV2`

dpkjnr:
> there should be some tests around DKLS and walletVersion 5, can you add bera also there. This coinfeature also needs to be added `CoinFeature.MPCV2`

this is one test https://github.com/BitGo/BitGoJS/blob/5360e06bca19430324e48e30ce5d0e4df1d79125/modules/bitgo/test/v2/unit/wallets.ts#L904

modules/statics/src/coins.ts
```diff
@@ -386,7 +386,8 @@ const ZKETH_FEATURES = [
 ];
 const BERA_FEATURES = [
   ...ETH_FEATURES,
-  CoinFeature.MULTISIG_COLD,
+  CoinFeature.TSS,
```
ajays97, line 389: Does this mean that bera started out as multisig and we are switching to TSS before rolling it out?

adarsh312: yes, we are switching to TSS. this decision is made before moving to mainnet

dpkjnr (approved):

PR #5309, opened 2024-12-23
fix(abstract-eth): fix verifyCoin method to validate native coin

venkateshv1266:
TICKET: COIN-2577

Withdraw transaction using the following script worked with this fix. 
Screenshot 2024-12-23 at 7 34 12 PM

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -2230,7 +2230,8 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
    * @returns {boolean}
    */
   verifyCoin(txPrebuild: TransactionPrebuild): boolean {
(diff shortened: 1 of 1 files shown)


modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2230,7 +2230,8 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
    * @returns {boolean}
    */
   verifyCoin(txPrebuild: TransactionPrebuild): boolean {
-    return txPrebuild.coin === this.getChain();
```
venkateshv1266, line 2233: For eg:
```
txPrebuild.coin = tarbeth
this.getChain() = tarbeth:link
```

hence this validation always fails in case of token withdrawal.

gianchandania (approved):

baltiyal (approved):

PR #5315, opened 2024-12-25
fix(abstract-utxo): verify transaction works for max recipient

davidkaplanbitgo:
TICKET: BTC-000

### DIFF
--- a/modules/abstract-utxo/src/transaction/fixedScript/parseTransaction.ts
+++ b/modules/abstract-utxo/src/transaction/fixedScript/parseTransaction.ts
@@ -74,6 +74,16 @@ export async function parseTransaction<TNumber extends bigint | number>(
 
   const allOutputs = [...explanation.outputs, ...explanation.changeOutputs];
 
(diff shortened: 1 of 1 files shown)


modules/abstract-utxo/src/transaction/fixedScript/parseTransaction.ts
```diff
@@ -74,6 +74,16 @@ export async function parseTransaction<TNumber extends bigint | number>(
 
   const allOutputs = [...explanation.outputs, ...explanation.changeOutputs];
 
+  /** Get the amount on the transaction. This amount is being derived explicitly from the txHex
+   * and not by any API metadata, so this is safe to do */
+  function getExplanationAmountForAddress(address: string): number | string {
+    const recipient = explanation.outputs.find((output) => output.address === address);
```
OttoAllmendinger, line 80: there can be more than one output with an address - this only returns the first one

OttoAllmendinger (changes requested):

PR #5316, opened 2024-12-27
feat(sdk-core): enable token enablements for multisig backing wallets

avalon22022002:
Ticket: GNA-10
scripts used :
[package.json](https://github.com/user-attachments/files/18259301/package.json)
[main.txt](https://github.com/user-attachments/files/18259327/main.txt)

- backing wallets are custodial wallets so flow should be same for both of them
- this change enables the token enablement of multi-sig backing wallets [ example : xrp ]

- output prior to this change
- image

- output after this change
- was able to send token enablement transactions  
- image

- steps to run 
     - npm install
     - rename the file main.txt as  main.js
     - run node main.js


avalon22022002:
reopened the same with a new ticket and branch
https://github.com/BitGo/BitGoJS/pull/5317

PR #5321, opened 2024-12-30
feat(statics): add go account support for coredao

avalon22022002:
Ticket: GNA-431


lcovar (approved):
adds `GO` account support for coreDAO

PR #5337, opened 2025-01-07
fix(abstract-eth): skip derived address validation for v5 forwarder

dpkjnr:
Ticket: COIN-2694

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -2197,7 +2197,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
       );
     }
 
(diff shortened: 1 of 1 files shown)


modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2197,7 +2197,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
       );
     }
 
-    if (impliedForwarderVersion === 0 || impliedForwarderVersion === 3 || impliedForwarderVersion === 6) {
+    if (impliedForwarderVersion === 0 || impliedForwarderVersion === 3 || impliedForwarderVersion === 5) {
```
mullapudipruthvik, line 2200: what is the difference between 5 and 6 ?
also can you add a testcase

mullapudipruthvik (approved):

PR #5338, opened 2025-01-07
feat(statics): ungate assets for frankfurt trust

yogeshwar-bitgo:
Ticket: WIN-4115


Vijay-Jagannathan (commented):
flushing

zahin-mohammad (commented):
PR is in draft and failing CI.

modules/statics/src/coins.ts
```diff
@@ -11340,7 +11382,7 @@ export const coins = CoinMap.fromCoins([
     18,
     '0xae78736cd615f374d3085123a210448e74fc6393',
     UnderlyingAsset['RETH-ROCKET'],
-    RETH_ROCKET_FEATURES
+    [...RETH_ROCKET_FEATURES, CoinFeature.CUSTODY_BITGO_FRANKFURT]
```
akshay-bitgo, line 11385: nit: I see that in some coins we create feature and then use it. 
You might want to consider using it. 
const RETH-ROCKET_FEATURES = [
  ...RETH_ROCKET_FEATURES, CoinFeature.CUSTODY_BITGO_FRANKFURT
];

akshay-bitgo (approved):
minor comments, else LGTM

modules/statics/src/coins.ts
yogeshwar-bitgo: Its good when we have to use it in multiple places. I thought since its single usage, have directly defined.

PR #5342, opened 2025-01-07
chore: onboard GoUSD token with eth prefix WIN-4114

rv-bitgo:
WIN-4114


modules/statics/src/base.ts
```diff
@@ -1533,7 +1533,6 @@ export enum UnderlyingAsset {
   TGBP = 'tgbp',
   TUSDS = 'tusds',
   TGOUSD = 'tgousd',
-  GOUSD = 'gousd',
```
sachushaji, line 1536: why do we need to change the base to eth: gousd?, only name needs to be changed?

modules/statics/src/coins.ts
```diff
@@ -4542,7 +4542,7 @@ export const coins = CoinMap.fromCoins([
     'goUSD',
     6,
     '0xf1f6b8cc3a0d544a40f1d29909396378370e6938',
-    UnderlyingAsset.GOUSD
+    UnderlyingAsset['eth:gousd']
```
sachushaji, line 4545: no need to change the token name with eth prefixed?

rv-bitgo: Just following conventions from previous PRs.

sachushaji (approved):
LGTM

PR #5347, opened 2025-01-08
feat(sdk-coin-icp): added ICP to statics

mohd-kashif:
TICKET: WIN-4236

PR #5361, opened 2025-01-10
feat(statics): removed MPCV2 feature from aptos

mohd-kashif:
TICKET: WIN-4236

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -426,7 +426,6 @@ const APT_FEATURES = [
   ...AccountCoin.DEFAULT_FEATURES,
   CoinFeature.TSS,
   CoinFeature.TSS_COLD,
-  CoinFeature.MPCV2,
   CoinFeature.STAKING,
   CoinFeature.BULK_TRANSACTION,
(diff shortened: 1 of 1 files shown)


modules/statics/src/coins.ts
```diff
@@ -426,7 +426,6 @@ const APT_FEATURES = [
   ...AccountCoin.DEFAULT_FEATURES,
   CoinFeature.TSS,
   CoinFeature.TSS_COLD,
-  CoinFeature.MPCV2,
```
mukeshsp, line 429: Why are you removing the mpcv2 flag ?

mohd-kashif: this was added by mistake while resolving merge conflicts for [PR](https://github.com/BitGo/BitGoJS/pull/5347)

bhavidhingra: it's ok @mohd-kashif , we would've eventually added it for APT. Let it stay and close this PR.

mohd-kashif: Thanks

PR #5368, opened 2025-01-14
feat(sdk-core): add WalletBalance codecs

ericli-bitgo:
TICKET: BTC-1692


davidkaplanbitgo (changes requested):
can you export the codecs as well so we can use them for decoding responses?

davidkaplanbitgo (approved):

PR #5373, opened 2025-01-15
feat(sdk-core): add skipReceiveAddress optional parameter IWallets

kisslove-dewangan:
TICKET: WP-1271

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallets.ts
@@ -166,11 +166,7 @@ export interface ListWalletOptions extends PaginationOptions {
   skip?: number;
   getbalances?: boolean;
   allTokens?: boolean;
-  /**
-   * Optional property for receiving an address.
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -166,6 +166,7 @@ export interface ListWalletOptions extends PaginationOptions {
   skip?: number;
   getbalances?: boolean;
   allTokens?: boolean;
+  skipReceiveAddress?: boolean;
```
zahin-mohammad, line 169: Please add a jsdoc for this. 

Can you also create a test showing the api gets called with this new field?

zahin-mohammad (dismissed):

modules/sdk-core/src/bitgo/wallet/iWallets.ts
kisslove-dewangan: added test case and jsdoc

pengyuc-bitgo (approved):

PR #5378, opened 2025-01-15
feat(sdk-core): return all fields in WalletBalance codecs

ericli-bitgo:
BTC-1692

TICKET: BTC-1692

### DIFF
--- a/modules/sdk-core/src/bitgo/lightning/codecs.ts
+++ b/modules/sdk-core/src/bitgo/lightning/codecs.ts
@@ -128,6 +128,8 @@ export const LndWalletBalance = t.strict(
     totalBalance: t.string,
     confirmedBalance: t.string,
     unconfirmedBalance: t.string,
+    lockedBalance: t.string,
+    reservedBalanceAnchorChan: t.string,
   },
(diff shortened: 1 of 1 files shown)


modules/sdk-core/src/bitgo/lightning/codecs.ts
```diff
@@ -128,6 +128,8 @@ export const LndWalletBalance = t.strict(
     totalBalance: t.string,
     confirmedBalance: t.string,
     unconfirmedBalance: t.string,
+    lockedBalance: t.string,
+    reservedBalanceAnchorChan: t.string,
```
davidkaplanbitgo, line 132:
```suggestion
    reservedBalanceAnchorChain: t.string,
```

davidkaplanbitgo (changes requested):

modules/sdk-core/src/bitgo/lightning/codecs.ts
davidkaplanbitgo: JK - I was wrong

davidkaplanbitgo (approved):

PR #5403, opened 2025-01-21
feat(sdk-coin-sol): fetch prio fees from WP

Vijay-Jagannathan:
Ticket: CR-1224

Allowing WP to set the prio fees for token transactions (this can be extended for other types as well since this implementation lives in the transaction builder and therefore other txn type builders in WP can also set this value if needed along with corresponding change in the SDK builder to add the necessary instruction)


gianchandania (approved):
lgtm

PR #5409, opened 2025-01-21
feat(sdk-coin-sol): ignore zero value prio fee

Vijay-Jagannathan:
Ticket: CR-1224

With this change, unless WP passes a prio fee > 0, we will not add a set compute price instruction. This is to ensure that we can dynamically change the prio fee value using an ENV variable in WP and if not, set the prio fee as 0


gianchandania (approved):
lgtm

PR #5429, opened 2025-01-26
fix: replace enums with string literal types

bitgoAaron:
- SignatureShareType, CommitmentType, and EncryptedSignerShareType converted to const objects with type exports

Ticket: [CE-5160](https://bitgoinc.atlassian.net/browse/CE-5160)

[CE-5160]: https://bitgoinc.atlassian.net/browse/CE-5160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
@@ -360,11 +360,13 @@ export type CreateKeychainParamsBase = {
 
 export type CreateBitGoKeychainParamsBase = Omit<CreateKeychainParamsBase, 'bitgoKeychain'>;
 
-export enum SignatureShareType {
-  USER = 'user',
-  BACKUP = 'backup',
-  BITGO = 'bitgo',
-}
+export const SignatureShareType = {
+  USER: 'user',
+  BACKUP: 'backup',
+  BITGO: 'bitgo',
+} as const;
+
+export type SignatureShareType = (typeof SignatureShareType)[keyof typeof SignatureShareType];
 
 interface ShareBaseRecord {
   from: SignatureShareType;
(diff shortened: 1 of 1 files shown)


margueriteblair (dismissed):

kaustubhbitgo:
If bundle size, performance is the issue, then instead of introducing a breaking change to replace enums with string literals we can opt for any of the following method.

1. `const enum` - Erases enum type completely and provides only type definition. If used then `tsc` replaces it with string literals.
```typescript
// file.ts
export const enum CommitmentType {
  COMMITMENT = 'commitment',
  DECOMMITMENT = 'decommitment',
}

// example.ts
console.log(CommitmentType.COMMITMENT)

// example.js (after tsc build)
console.log('commitment')
```

2. `as const` - Provides readonly type definitions, less impact on bundle size since it is a direct object mapping instead of function shenanigans that enums do.
```typescript
// file.ts
export const CommitmentType = {
  COMMITMENT: 'commitment',
  DECOMMITMENT: 'decommitment',
} as const;

// example.ts
console.log(CommitmentType.COMMITMENT) /* same in compiled js with less overhead */
```

Have you explored any of these methods?

bitgoAaron:
> If bundle size, performance is the issue, then instead of introducing a breaking change to replace enums with string literals we can opt for any of the following method.
> 
> 1. `const enum` - Erases enum type completely and provides only type definition. If used then `tsc` replaces it with string literals.
> 
> ```ts
> // file.ts
> export const enum CommitmentType {
>   COMMITMENT = 'commitment',
>   DECOMMITMENT = 'decommitment',
> }
> 
> // example.ts
> console.log(CommitmentType.COMMITMENT)
> 
> // example.js (after tsc build)
> console.log('commitment')
> ```
> 
> 2. `as const` - Provides readonly type definitions, less impact on bundle size since it is a direct object mapping instead of function shenanigans that enums do.
> 
> ```ts
> // file.ts
> export const CommitmentType = {
>   COMMITMENT: 'commitment',
>   DECOMMITMENT: 'decommitment',
> } as const;
> 
> // example.ts
> console.log(CommitmentType.COMMITMENT) /* same in compiled js with less overhead */
> ```
> 
> Have you explored any of these methods?

good call @kaustubhbitgo , went with your suggestion & included a type export which will satisfy our needs tree shaking the UI.

margueriteblair (approved):

PR #5433, opened 2025-01-26
feat(sdk-coin-apt): set transaction type in transactionBuilder

bhavidhingra:
TICKET: COIN-2904


bhavidhingra:
Not required because TransferBuilder will already have a type.

PR #5437, opened 2025-01-27
feat(sdk-coin-apt): change fungibleAddress to assetId

baltiyal:
Ticket: COIN-2892

This change is to make code more extendable to further types of tokens supported for aptos i.e. non fungible, legacy token and legacy coin

### DIFF
--- a/modules/sdk-coin-apt/src/aptToken.ts
+++ b/modules/sdk-coin-apt/src/aptToken.ts
@@ -36,8 +36,8 @@ export class AptToken extends Apt {
     return this.tokenConfig.network;
   }
 
-  get fungibleAssestAddress(): string {
-    return this.tokenConfig.fungibleAssetAddress;
+  get assetId(): string {
+    return this.tokenConfig.assetId;
   }
 
   get decimalPlaces(): number {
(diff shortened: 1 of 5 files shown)


sachushaji (commented):
Seems to be a breaking change, can you add the BREAKING change footer in the commit message. Interesting that this change didn't require any test case changes, can you add one as well

modules/sdk-coin-apt/src/aptToken.ts
```diff
@@ -36,8 +36,8 @@ export class AptToken extends Apt {
     return this.tokenConfig.network;
   }
 
-  get fungibleAssestAddress(): string {
-    return this.tokenConfig.fungibleAssetAddress;
+  get assetId(): string {
```
bhavidhingra, line 39: add assertion for this in `modules/sdk-coin-apt/test/unit/aptToken.ts`

bhavidhingra:
> Seems to be a breaking change, can you add the BREAKING change footer in the commit message. Interesting that this change didn't require any test case changes, can you add one as well

APT onboarding is still in staging env stage. This will not cause any breaking change.

modules/sdk-coin-apt/src/aptToken.ts
baltiyal: Added.

margueriteblair (approved):

pengyuc-bitgo (approved):

bhavidhingra (approved):

PR #5450, opened 2025-01-29
feat(sdk-core): generalize getSignatureShareRoundX mpcv2 helpers

islamaminBitGo:
TICKET: HSM-661

### DIFF
--- a/modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts
+++ b/modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts
@@ -11,6 +11,18 @@ import {
 } from '@bitgo/public-types';
 import assert from 'assert';
 
+function partyIdToSignatureShareType(partyId: 0 | 1 | 2): SignatureShareType {
+  assert(partyId === 0 || partyId === 1 || partyId === 2, 'Invalid partyId for MPCv2 signing');
+  switch (partyId) {
+    case 0:
+      return SignatureShareType.USER;
+    case 1:
+      return SignatureShareType.BACKUP;
+    case 2:
+      return SignatureShareType.BITGO;
+  }
+}
+
 /**
  Helpers in this take care of all interaction with WP API's
 **/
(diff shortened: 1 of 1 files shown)


modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -11,14 +11,26 @@ import {
 } from '@bitgo/public-types';
 import assert from 'assert';
 
+function partyIdToSignatureShareType(partyId: 0 | 1 | 2): SignatureShareType {
+  assert(partyId == 0 || partyId == 1 || partyId == 2, 'Invalid partyId for MPCv2 signing');
```
mohammadalfaiyazbitgo: why do you need to assert if you've typed the params to `0|1|2`?

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -11,14 +11,26 @@ import {
...
+  if (partyId == 0) {
+    return SignatureShareType.USER;
+  } else if (partyId == 1) {
+    return SignatureShareType.BACKUP;
+  } else {
+    return SignatureShareType.BITGO;
+  }
+}
```
mohammadalfaiyazbitgo: nit: switch case.

mohammadalfaiyazbitgo (commented):
LGTM, just a few nits.

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts
islamaminBitGo: to protect from a friendly neighbourhood coder who would do something like `partyIdToSignatureShareType(5 as 0 | 1 | 2_ `

islamaminBitGo: typescript is not strictly typed enough for me to not worry about these assertions completely.

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts
islamaminBitGo: done

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts
mohammadalfaiyazbitgo: then it becomes the caller's problem lol, but fair enough.

mohammadalfaiyazbitgo (approved):

modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts
islamaminBitGo: that's a valid point, i could see valid points both ways

PR #5464, opened 2025-01-31
fix(sdk-coin-sol): set priority fee in initBuilder

Vijay-Jagannathan:
Ticket: COIN-2932

We currently only set the priority fee from the tokenTransfer and the consolidation builder. But in WP, we also create the builder using the raw transaction object, which requires some necessary fields to be set again. Due to this, the priority fee variable was not set, so the tokenTransfer/consolidation builder (in SDK) did not attach any set priority fee instruction even though WP was setting the priority fee. This PR amends this flow to set the priority fee variable even while creating the builder through the raw txn data

### DIFF
--- a/modules/sdk-coin-sol/src/lib/transactionBuilder.ts
+++ b/modules/sdk-coin-sol/src/lib/transactionBuilder.ts
@@ -76,6 +76,10 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     this.feePayer(txData.feePayer as string);
     this.nonce(txData.nonce, txData.durableNonce);
     this._instructionsData = instructionParamsFactory(tx.type, tx.solTransaction.instructions);
+    // Parse priority fee instruction data
+    const filteredPriorityFeeInstructionsData = txData.instructionsData.filter(
+      (data) => data.type === InstructionBuilderTypes.SetPriorityFee
+    );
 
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-sol/src/lib/transactionBuilder.ts
```diff
@@ -76,6 +76,15 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
     this._instructionsData = instructionParamsFactory(tx.type, tx.solTransaction.instructions);
+    // Set priority fee if it exists
+    const filteredPriorityFeeInstructionsData = txData.instructionsData.filter(
+      (data) => data.type === InstructionBuilderTypes.SetPriorityFee
+    );
+
+    if (filteredPriorityFeeInstructionsData.length > 0) {
+      const priorityFeeInstructionsData = filteredPriorityFeeInstructionsData[0];
+      this._priorityFee = Number(priorityFeeInstructionsData.params.fee);
+    }
```
gianchandania, line 87:
```
if (instruction.type === InstructionBuilderTypes.SetPriorityFee) {
        this.setPriorityFee()
      }
```

We are already iterating over the instructions in below for loop, lets add this condition there itself. Also we can use the setter instead of setting the value directly

dialsquare1886 (approved):

PR #5477, opened 2025-02-04
feat(sdk-core): add invoice codecs

ericli-bitgo:
BTC-0


saravanan7mani (approved):

saravanan7mani (changes requested):
Only signer node operational APIs of lightning (initWallet, signerMacaroon, etc) are specific to Express. All other regular client APIs (create invoice, create payment, etc) will be available in wallets. So `sdk-core` is the right place for these codecs.

ericli-bitgo:
oops I thought I was in sdk-core - accidentally put it in the wrong file

ericli-bitgo:
Should be in the right place now

saravanan7mani (approved):

PR #5492, opened 2025-02-06
refactor(sdk-core): don't use implied types and use string instead

ericli-bitgo:
It was difficult to reason about these types

BTC-0

TICKET: BTC-0


davidkaplanbitgo (approved):

OttoAllmendinger (changes requested):
what was the issue here? the original code looks correct

ericli-bitgo:
When implementing it on WP, it was hard to tell when it was a string or a bigint because the TS type of this is `string | bigint`.

PR #5495, opened 2025-02-06
chore(root): add eslint rule to restrict sinon usage

kev-daniell:
Ticket: DX-1013

TICKET: DX-1013

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -67,6 +67,27 @@
       "rules": {
         "no-undef": "off"
       }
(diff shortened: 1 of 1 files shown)


zahin-mohammad (commented):
What's the alternative if we are moving away from sinon? 

Let's take this chat offline with the rest of the wallets team

zahin-mohammad (dismissed):

zahin-mohammad (approved):

PR #5491, opened 2025-02-06
chore(statics): add statics fields to common fields

rakeshgsharma:
TICKET: WIN-4536

## Description

We want to merge coin specific fields to common fields which are used across other coins. This ensures there's uniformity and eventually help building a proper structure for coin statics.

`packageId::module::symbol` -> `contractAddress`
`tokenId`-> `contractAddress`
`contractName`-> `contractAddress`
`tokenAddress`-> `contractAddress`
`issuerAddress::currencyCode`-> `contractAddress`

After we release this, we will be making necessary changes required in our systems to use the standard fields and then eventually remove these non-standard coin specific fields permanently.

## Issue Number

WIN-4536

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

This is non breaking change, and only adds extra fields which are already present.

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My code compiles correctly for both Node and Browser environments
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My commits follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) and I have properly described any BREAKING CHANGES
- [x] The ticket or github issue was included in the commit message as a reference
- [x] I have made corresponding changes to the documentation and on any new/updated functions and/or methods - [jsdoc](https://jsdoc.app/)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #5508, opened 2025-02-10
chore(statics): use contractAddress to form issuerAddress and currenc…

rakeshgsharma:
## Description

We want to generate `issuerAddress` and `currencyCode` from existing param `contractAddress`. `contractAddress` is formed using `issuerAddress::currencyCode`.  Eventually, we will be removing `issuerAddress` and `currencyCode` being passed from SDK statics.

## Issue Number

WIN-4541

## Type of change

# How Has This Been Tested?

We're simply changing how a variable gets its value. There's no new variable being created, hence if existing test cases are passing, we are good to go.

# Checklist:

### DIFF
--- a/modules/statics/src/account.ts
+++ b/modules/statics/src/account.ts
@@ -451,8 +451,8 @@ export class XrpCoin extends AccountCoinToken {
     }
 
     this.domain = options.domain as string;
-    this.currencyCode = options.currencyCode;
-    this.issuerAddress = options.issuerAddress;
+    this.issuerAddress = options.contractAddress.split('::')[0];
+    this.currencyCode = options.contractAddress.split('::')[1];
     this.contractAddress = options.contractAddress;
   }
 }
--- a/modules/sdk-coin-xrp/src/xrpToken.ts
+++ b/modules/sdk-coin-xrp/src/xrpToken.ts
@@ -44,6 +44,10 @@ export class XrpToken extends Xrp {
     return this.tokenConfig.currencyCode;
   }
 
+  get contractAddress(): string {
+    return this.tokenConfig.contractAddress;
+  }
+
   get domain(): string {
     return this.tokenConfig.domain || '';
   }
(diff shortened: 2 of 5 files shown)


modules/statics/src/account.ts
```diff
@@ -451,8 +451,8 @@ export class XrpCoin extends AccountCoinToken {
     }
 
     this.domain = options.domain as string;
-    this.currencyCode = options.currencyCode;
-    this.issuerAddress = options.issuerAddress;
+    this.issuerAddress = options.contractAddress.split('::')[0];
```
zahin-mohammad, line 454: Coul you add a small test for this? Just to verify all address's have `::` set and no on introduces a bad contractaddress

zahin-mohammad (changes requested):

modules/statics/src/account.ts
sachushaji: +1, I might be wrong but I didn't see contract address defined in XrpToken object?

sachushaji: +1, I might be wrong but I didn't see contract address defined in XrpToken object?

rakeshgsharma: @sachushaji - xrpToken has contractAddress here -https://github.com/BitGo/BitGoJS/blob/447a985265ec4f484fac1ef7f3e1db8b937863d6/modules/statics/src/account.ts#L2208

@zahin-mohammad - test case was already added in a previous PR here - https://github.com/BitGo/BitGoJS/pull/5491/files#diff-5f212436d31cca0d1a2757b536444b73fd3ab25286cb351b8bd8f2e808730860R780

rakeshgsharma: test case - 
https://github.com/BitGo/BitGoJS/blob/master/modules/statics/test/unit/coins.ts#L780:L790

sachushaji: can you add a test which should have broke unless the change introduced by you wasn't added

rakeshgsharma: done

DinshawKothari (approved):

zahin-mohammad (approved):

PR #5509, opened 2025-02-10
feat(sdk-coin-sui): walrus unstaking tx support

benkelcher-bitgo:
TICKET: SC-1098

### DIFF
--- a/modules/sdk-coin-sui/src/lib/iface.ts
+++ b/modules/sdk-coin-sui/src/lib/iface.ts
@@ -20,6 +20,8 @@ export enum SuiTransactionType {
   CustomTx = 'CustomTx',
   TokenTransfer = 'TokenTransfer',
   WalrusStakeWithPool = 'WalrusStakeWithPool',
+  WalrusRequestWithdrawStake = 'WalrusRequestWithdrawStake',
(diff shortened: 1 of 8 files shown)


modules/sdk-coin-sui/src/lib/iface.ts
```diff
@@ -119,13 +134,13 @@ export enum MethodNames {
    *
    * @see https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/docs/sui_system.md#function-request_add_stake
    */
-  RequestAddStake = 'request_add_stake',
+  RequestAddStake = '::sui_system::request_add_stake',
```
noel-bitgo, line 137: why ? 🤔

benkelcher-bitgo: we use these to infer BitGo-internal transaction type based on a deserialized SUI transaction. we were getting problems since `request_withdraw_stake` was used for both SUI and WALRUS (ie. the same postfix), so when deserializing a WALRUS unstaking transaction it would be incorrectly labelled as a SUI unstaking transaction. adding more to the method name like this fixes that

not strictly necessary for `request_add_stake` but changed it for consistency

ewangbitgo (approved):

PR #5510, opened 2025-02-10
Adding Staking Builder for Tao

abhijit0943:


modules/sdk-coin-tao/test/unit/transactionBuilder/stakingBuilder.ts
```diff
@@ -0,0 +1,247 @@
...
+      should.deepEqual(txJson.sender, sender.address);
+      should.deepEqual(txJson.blockNumber, 3933);
+      should.deepEqual(txJson.referenceBlock, referenceBlock);
+      should.deepEqual(txJson.genesisHash, '0xe143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e');
+      should.deepEqual(txJson.specVersion, 9430);
+      should.deepEqual(txJson.nonce, 200);
+      should.deepEqual(txJson.tip, 0);
+      should.deepEqual(txJson.transactionVersion, 22);
+      should.deepEqual(txJson.chainName, 'Westend');
+      should.deepEqual(txJson.eraPeriod, 64);
```
benkelcher-bitgo, line 80: one interesting thing to check could be if it is able to re-build the transaction to the same transaction bytes - i think that this is something that happens during our internal transaction signing flow. could add something like this to the test:
```typescript
const rawTx = tx.toBroadcastFormat();
const rebuilder = factory.from(rawTx);
rebuilder.sender({ ... }).validity({ ... }); // not sure if this part is needed
const rebuiltTx = await rebuilder.build();
assert(rebuiltTx.toBroadcastFormat() === rawTx);
```

modules/sdk-coin-tao/src/lib/stakingBuilder.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (this._addToStake) {
+      tx = methods.staking.bondExtra({ maxAdditional: this._amount }, baseTxInfo.baseTxInfo, baseTxInfo.options);
+    } else {
+      tx = methods.staking.bond(
+        { controller: this._controller, value: this._amount, payee: this._payee },
+        baseTxInfo.baseTxInfo,
+        baseTxInfo.options
+      );
+    }
+    if (this._validityWindow && (tx as any).signature && baseTxInfo.options.registry) {
```
noel-bitgo, line 53: isn't this the DOT specific methods?

PR #5558, opened 2025-02-14
fix(sdk-lib-mpc): fix time warp during decrypting comms in MPCv2

islamaminBitGo:
TICKET: HSM-706

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
@@ -132,13 +132,13 @@ export async function verifySharedDataProof(
   const senderPubKey = await pgp.readKey({ armoredKey: senderPubKeyArm });
   const signedKey = await pgp.readKey({ armoredKey: keyWithNotation });
   if (
-    !(await verifyPrimaryUserWrapper(signedKey, senderPubKey, true).then((values) =>
+    !(await verifyPrimaryUserWrapper(signedKey, senderPubKey, false).then((values) =>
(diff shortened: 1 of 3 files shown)


islamaminBitGo:
Found a way to test time warps. Will add tests and then open the PR again

margueriteblair (dismissed):

alebusse (commented):
flushing from review queue until is ready

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
```diff
@@ -132,13 +132,13 @@ export async function verifySharedDataProof(
...
   if (
-    !(await verifyPrimaryUserWrapper(signedKey, senderPubKey, true).then((values) =>
+    !(await verifyPrimaryUserWrapper(signedKey, senderPubKey, false).then((values) =>
       _.some(values, (value) => value.valid)
     ))
   ) {
     return false;
   }
-  const primaryUser = await signedKey.getPrimaryUser();
+  const primaryUser = await signedKey.getPrimaryUser(null as unknown as undefined);
```
zahin-mohammad, line 141: what a line... 

we should create a type def for this to override their invalid types. can do this in a follow up.

zahin-mohammad (approved):

alebusse (approved):

modules/sdk-core/src/bitgo/utils/opengpgUtils.ts
islamaminBitGo: yeah agreed, almost puked while writing this

PR #5565, opened 2025-02-17
fix: enable tokens for cold wallet

abhishekagrawal080:
TICKET: WIN-4643


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3698,7 +3698,7 @@ export class Wallet implements IWallet {
     // Doing a sanity check for password here to avoid doing further work if we know it's wrong
     // we ignore this check with if customSigningFunction is provided
     //  which means that the user is handling the signing in external signing mode
-    if (keychains[0].encryptedPrv && !customSigningFunction && walletPassphrase) {
+    if (!customSigningFunction && keychains?.[0]?.encryptedPrv && walletPassphrase) {
```
mullapudipruthvik, line 3701: can you add a test for this?

Issue #4157, opened 2023-12-15
Security issue: replace non-secure cryptography


paulmillr:
You're using elliptic, which has long been unmaintained, and has a few cases where it produces invalid outputs, which means in blockchain context "users lose money".

It is advised to replace it with **audited** `@noble/curves`.

paulmillr:
Do you guys still care about security?

Elliptic just had a vulnerability which leaked private keys from signatures https://github.com/indutny/elliptic/security/advisories/GHSA-vjh7-7g9h-fjfh

OttoAllmendinger:
thank you for pointing this out

https://github.com/BitGo/BitGoJS/pull/5577

PR #5566, opened 2025-02-17
fix(root): bump elliptic to 6.6.1

alebusse:
bumped elliptic to 6.6.1

WP-3700

TICKET: WP-3700


package.json
```diff
@@ -82,7 +82,8 @@
     "secp256k1": "5.0.1",
     "socks": "2.7.3",
     "web3-utils": "4.2.1",
-    "@polkadot/api": "14.1.1"
+    "@polkadot/api": "14.1.1",
+    "elliptic": "^6.6.1"
```
alebusse, line 86: not a solution but a work around, many coins libs still use the an older version

mohd-kashif (approved):

zahin-mohammad (approved):

paulmillr:
Elliptic should not be upgraded, it should be REPLACED. I've opened an issue 1.2 years ago: #4157.

Do you even care about security? CVE leaked users private keys from signatures. There is another elliptic CVE upcoming.

@mohd-kashif @zahin-mohammad @alebusse

OttoAllmendinger:
moving elliptic to devDeps for utxo-lib in this PR: https://github.com/BitGo/BitGoJS/pull/5569

OttoAllmendinger (approved):

PR #5567, opened 2025-02-18
fix(apt): correction 64 bit LE to two buffer of 32 bits

baltiyal:
TICKET: COIN-3157

Same change has been made for Apt Native : https://github.com/BitGo/BitGoJS/pull/5497
Have moved that implementation to utils and implemented for Native APT and Fungible Asset.


sachushaji (commented):
Flushing from queue since draft, Write UTs also when you undraft

baltiyal:
> Flushing from queue since draft, Write UTs also when you undraft

Added UT for the change.

sachushaji (approved):

PR #5575, opened 2025-02-18
feat(tao): staking builder example

noel-bitgo:

### DIFF
--- a/modules/sdk-coin-tao/src/lib/iface.ts
+++ b/modules/sdk-coin-tao/src/lib/iface.ts
@@ -1,0 +1,38 @@
+import { BaseTxInfo, TypeRegistry } from '@substrate/txwrapper-core/lib/types';
+import { Args } from "@substrate/txwrapper-core/lib/types/method";
+
+export { HexString } from '@polkadot/util/types';
+
+/**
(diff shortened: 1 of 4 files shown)


modules/sdk-coin-tao/src/lib/iface.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export interface TxMethod {
+  args: AddStakeArgs;
+  name: MethodNames;
+  pallet: string;
+}
+
+export interface AddStakeArgs extends Args {
+  amount: string;
+  hotkey: string;
```
noel-bitgo, line 36: should add `netuid` too

modules/sdk-coin-tao/src/lib/iface.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Decoded TxMethod from a transaction hex
+ */
+export interface TxMethod {
+  args: AddStakeArgs;
+  name: MethodNames;
+  pallet: string;
+}
+
+export interface AddStakeArgs extends Args {
+  amount_staked: string;
```
noel-bitgo, line 35: check if `amount_staked` or `amountStaked` i remember for DOT there was weird behaviour with naming convention when building tx

PR #5578, opened 2025-02-18
feat(express): walletId should come from the path

davidkaplanbitgo:
BREAKING CHANGE: path for lightning is different

TICKET: BTC-1846

### DIFF
--- a/modules/express/src/lightning/lightningSignerRoutes.ts
+++ b/modules/express/src/lightning/lightningSignerRoutes.ts
@@ -58,7 +58,7 @@ function getMacaroonRootKey(passphrase: string, nodeAuthEncryptedPrv: string, de
  * Handle the request to initialise remote signer LND for a wallet.
  */
 export async function handleInitLightningWallet(req: express.Request): Promise<unknown> {
-  const { walletId, passphrase, signerTlsKey, signerTlsCert, signerIP, expressIP } = decodeOrElse(
+  const { passphrase, signerTlsKey, signerTlsCert, signerIP, expressIP } = decodeOrElse(
     InitLightningWalletRequest.name,
     InitLightningWalletRequest,
     req.body,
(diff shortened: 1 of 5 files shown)


OttoAllmendinger (dismissed):

modules/express/src/lightning/lightningSignerRoutes.ts
```diff
@@ -68,16 +68,16 @@ export async function handleInitLightningWallet(req: express.Request): Promise<u
     }
   );
 
-  const lndSignerClient = await LndSignerClient.create(walletId, req.config);
-
   const bitgo = req.bitgo;
+  const walletId = req.params.id;
```
saravanan7mani, line 72:
```
const walletId = req.params.id;
  if (typeof walletId !== 'string') {
    throw new Error(`Invalid wallet id: ${walletId}`);
  }
```

modules/express/src/lightning/lightningSignerRoutes.ts
```diff
@@ -125,7 +128,8 @@ export async function handleInitLightningWallet(req: express.Request): Promise<u
  * Handle the request to create a signer macaroon from remote signer LND for a wallet.
  */
 export async function handleCreateSignerMacaroon(req: express.Request): Promise<unknown> {
-  const { walletId, passphrase, watchOnlyIP } = decodeOrElse(
+  const walletId = req.params.id;
```
saravanan7mani, line 131:
```
const walletId = req.params.id;
  if (typeof walletId !== 'string') {
    throw new Error(`Invalid wallet id: ${walletId}`);
  }
```

modules/express/src/lightning/lightningSignerRoutes.ts
```diff
@@ -218,6 +228,7 @@ export async function handleUnlockLightningWallet(req: express.Request): Promise
     throw new Error(`Invalid coin to unlock lightning wallet: ${coinName}`);
   }
 
+  const walletId = req.params.id;
```
saravanan7mani, line 231:
```
const walletId = req.params.id;
  if (typeof walletId !== 'string') {
    throw new Error(`Invalid wallet id: ${walletId}`);
  }
```

saravanan7mani (changes requested):

saravanan7mani (approved):

PR #5585, opened 2025-02-19
chore(abstract-eth): add a setter method in TransferBuilder

venkateshv1266:
TICKET: COIN-2630

This is required for CCR service to decode and validate the signer.

### DIFF
--- a/modules/abstract-eth/src/lib/transferBuilder.ts
+++ b/modules/abstract-eth/src/lib/transferBuilder.ts
@@ -106,6 +106,11 @@ export class TransferBuilder {
     return this;
   }
 
+  setCoinUsesNonPackedEncodingForTxData(isCoinUsesNonPackedEncodingForTxData: boolean): TransferBuilder {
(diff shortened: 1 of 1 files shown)


modules/abstract-eth/src/lib/transferBuilder.ts
```diff
@@ -166,7 +166,7 @@ export class TransferBuilder {
     return operationHash;
   }
 
-  protected getOperationData(): (string | number | Buffer)[][] {
+  public getOperationData(): (string | number | Buffer)[][] {
```
mullapudipruthvik, line 169: why do you require this, I see signature is not returned here ?

venkateshv1266: Once we get the operationData I can use this to calculate operationHash and then decode signer in the ccr-service itself. 

I can't directly use getOperationHash of abstract-eth, because it would return wrong and operationHash since the prefix is undefined.

mullapudipruthvik (approved):

PR #5591, opened 2025-02-19
fix: address validation for cosmos like coins

DinshawKothari:
Ticket: COIN-3178

### DIFF
--- a/modules/abstract-cosmos/src/lib/utils.ts
+++ b/modules/abstract-cosmos/src/lib/utils.ts
@@ -825,13 +825,21 @@ export class CosmosUtils implements BaseUtils {
    * @returns true if memo id is valid
    */
   isValidMemoId(memoId: string): boolean {
-    let memoIdNumber: BigNumber;
-    try {
-      memoIdNumber = new BigNumber(memoId);
-    } catch (e) {
+    // Allow alphanumeric memo IDs (including uppercase and lowercase letters)
+    const alphanumericRegex = /^[0-9a-zA-Z]+$/;
+
+    // Check if the memoId is alphanumeric
+    if (!alphanumericRegex.test(memoId)) {
       return false;
     }
-    return memoIdNumber.gte(0) && memoIdNumber.isInteger();
+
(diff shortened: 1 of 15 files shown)


modules/abstract-cosmos/src/lib/utils.ts
```diff
@@ -825,13 +825,21 @@ export class CosmosUtils implements BaseUtils {
...
+    const alphanumericRegex = /^[0-9a-zA-Z]+$/;
+
+    // Check if the memoId is alphanumeric
+    if (!alphanumericRegex.test(memoId)) {
       return false;
     }
-    return memoIdNumber.gte(0) && memoIdNumber.isInteger();
+
+    // If the memoId is purely numeric, ensure it is a positive integer
+    if (/^\d+$/.test(memoId)) {
```
at31416, line 837: any test cases for this? if it's urgent to merge we can create a todo ticket to add test case for this specific check.

modules/sdk-coin-baby/test/resources/baby.ts
```diff
@@ -170,7 +170,7 @@ export const address = {
   validatorAddress4: 'bbnvaloper1rlrjlm6rec6h2nvzakyj39d5fv7p7vgs3kq472',
   noMemoIdAddress: 'bbn1274ep8pnrlej5vgmtwppysynzcd4fhxc3ku0t3',
   validMemoIdAddress: 'bbn1274ep8pnrlej5vgmtwppysynzcd4fhxc3ku0t3?memoId=2',
-  invalidMemoIdAddress: 'bbn1274ep8pnrlej5vgmtwppysynzcd4fhxc3ku0t3?memoId=xyz',
```
at31416, line 173: we can add a test case where memoId is passed as a special character for `invalidMemoIdAddress` rather than removing this.

DinshawKothari: This is for CR, so not adding more testcases here. Probably we will take this as a todo.

dpkjnr (approved):

PR #5599, opened 2025-02-20
feat(sdk-coin-apt): recommend min gas tank balance for next 1000 txns.

bhavidhingra:
TICKET: COIN-3185

### DIFF
--- a/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
+++ b/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
@@ -599,4 +599,20 @@ export abstract class BaseCoin implements IBaseCoin {
   setCoinSpecificFieldsInIntent(intent: PopulatedIntent, params: PrebuildTransactionWithIntentOptions): void {
     return;
   }
+
+  /**
+   * Gets the multiplying factor of fee estimate to alert for low fee address balance
+   * i.e. alert threshold balance = factor * fee estimate
+   */
+  getGasTankLowBalanceAlertThresholdFactor(): number {
+    return 2;
+  }
+
+  /**
+   * Gets the multiplying factor of fee estimate for recommending the fee address min. balance
+   * i.e. recommended min. balance = factor * fee estimate
+   */
+  getGasTankMinBalanceRecommendationFactor(): number {
+    return 10;
+  }
 }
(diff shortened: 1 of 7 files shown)


modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
```diff
@@ -599,4 +599,20 @@ export abstract class BaseCoin implements IBaseCoin {
...
+  getGasTankLowBalanceAlertThresholdFactor(): number {
+    return 2;
+  }
+
+  /**
+   * Gets the multiplying factor of fee estimate for recommending the fee address min. balance
+   * i.e. recommended min. balance = factor * fee estimate
+   */
+  getGasTankMinBalanceRecommendationFactor(): number {
+    return 10;
```
at31416, line 616: this will be used for email notification?

at31416: `if getGasTankLowBalanceAlertThresholdFactor() * fee estimate > gas tank balance` condition should be reversed?

bhavidhingra: oh yes :)

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
```diff
@@ -599,4 +599,20 @@ export abstract class BaseCoin implements IBaseCoin {
   setCoinSpecificFieldsInIntent(intent: PopulatedIntent, params: PrebuildTransactionWithIntentOptions): void {
     return;
   }
+
+  /**
+   * Gets the multiplying factor of fee estimate to alert for low fee address balance
+   * i.e. alert threshold balance = factor * fee estimate
+   */
+  getGasTankLowBalanceAlertThresholdFactor(): number {
```
at31416, line 607: should these methods be abstract? and only coins which needs these can implement according to their specific number.

bhavidhingra: won't every coin have to implement this method if we make it `abstract`?

at31416: multiplier for all the coins won't be same, so that's why I was thinking should we make it abstract?

at31416 (approved):

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
bhavidhingra: Currently in the UI, these values are the same for all the eth like coins. i.e. 2 and 10 respectively.

PR #5614, opened 2025-02-22
feat(abstract-substrate): add missing methods for TAO stake/unstake

Vijay-Jagannathan:
Ticket: SC-1044

Adding some methods that are to be consumed in the respective staking/unstaking builder PRs

### DIFF
--- a/modules/abstract-substrate/src/lib/transaction.ts
+++ b/modules/abstract-substrate/src/lib/transaction.ts
@@ -151,8 +151,8 @@ export class Transaction extends BaseTransaction {
       tip: decodedTx.tip ? Number(decodedTx.tip) : 0,
     };
 
+    const txMethod = decodedTx.method.args;
     if (this.type === TransactionType.Send) {
-      const txMethod = decodedTx.method.args;
       if (utils.isTransfer(txMethod)) {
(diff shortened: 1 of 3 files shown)


hitansh-madan (approved):

modules/abstract-substrate/src/lib/transaction.ts
```diff
@@ -168,6 +168,26 @@ export class Transaction extends BaseTransaction {
       } else {
         throw new ParseTransactionError(`Serializing unknown Transfer type parameters`);
       }
+    } else if (this.type === TransactionType.StakingActivate) {
+      if (utils.isAddStake(txMethod)) {
```
hitansh-madan, line 172: nit:
```suggestion
    } else if (this.type === TransactionType.StakingActivate && utils.isAddStake(txMethod))
```

Vijay-Jagannathan: kept it this way since I'm not sure if there's a separate method for adding extra stake to the same validator hotkey address. I saw that dot has bond and bondExtra, the latter does the addition of new stake to the same validator.

Once we confirm the above, can refactor these accordingly

PR #5615, opened 2025-02-22
feat(statics): add GasTankAccountCoin

bhavidhingra:
TICKET: COIN-3185

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -12,6 +12,7 @@ import {
   erc20CompatibleAccountCoin,
   erc721,
   fiat,
+  gasTankAccount,
   hederaCoin,
   hederaToken,
   nonstandardToken,
@@ -1156,7 +1157,7 @@ export const coins = CoinMap.fromCoins([
     BaseUnit.ETH,
     EVM_FEATURES
   ),
-  account(
+  gasTankAccount(
     '75a71e9c-e3a0-4852-8e4b-9613ffed2a4c',
     'apt',
     'Aptos',
@@ -1165,9 +1166,11 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.APT,
     BaseUnit.APT,
     APT_FEATURES,
-    KeyCurve.Ed25519
+    KeyCurve.Ed25519,
+    400,
+    1000
   ),
-  account(
+  gasTankAccount(
     '7aca10bf-79dd-428b-aeb6-54f03f9aec0f',
     'tapt',
     'Testnet Aptos',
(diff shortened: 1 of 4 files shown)


modules/statics/src/coins.ts
```diff
@@ -1176,7 +1179,9 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset.APT,
     BaseUnit.APT,
     APT_FEATURES,
-    KeyCurve.Ed25519
+    KeyCurve.Ed25519,
+    400,
+    1000
```
Vijay-Jagannathan, line 1184: trying to understand this better: would this be something that we'll need to change often/have the ability to change quickly if needed ?

if yes, wouldn't we be better off having this in WP instead of statics ? say we do have it in statics, we need two changes to update it

additionally, not sure whether we have this model for any other coin where the gastank thresholds, etc are part of the SDK. Ideally statics shouldn't have to serve something custom/dynamic such as gas tank threshold

modules/statics/src/coins.ts
```diff
@@ -1156,7 +1157,7 @@ export const coins = CoinMap.fromCoins([
     BaseUnit.ETH,
     EVM_FEATURES
   ),
-  account(
+  gasTankAccount(
```
at31416, line 1160: some other coins also have gas tank logic, so do we have to change the way they are initialised also?

modules/statics/src/coins.ts
bhavidhingra: The reason I've added it in statics is to have it controlled from a single place. The threshold and recommended min balance is required both in WP and UI. Having the configs at multiple places can created inconsistencies in the long term.

modules/statics/src/coins.ts
bhavidhingra: we don't have to change any coin right now. It can be done one coin at a time for the coins that need these threshold and recommended min. balance values in WP/UI.

modules/statics/src/coins.ts
Vijay-Jagannathan: ah, makes sense then if we need to update at multiple places (do add context in ticket/PR description), but I haven't seen the usage of gas tank threshold in the UI separately.

Is this only for Aptos or some specific set of coins ?

Vijay-Jagannathan (approved):

modules/statics/src/coins.ts
bhavidhingra: It is already in place for eth like coins for displaying a banner on the Wallets Page.
Please refer to this PR for clarity over their usage and the current threshold for eth-like coins: https://github.com/BitGo/bitgo-ui/pull/6645

PR #5617, opened 2025-02-24
chore:add non bitgo recovery-sgb,flr,wemix

prithvishet2503:
TICKET: WIN-4675

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -237,6 +237,10 @@ export class BitGoAPI implements BitGoBase {
       'bscscanApiToken',
       'coredaoExplorerApiToken',
       'oasExplorerApiToken',
+      'sgbExplorerApiToken',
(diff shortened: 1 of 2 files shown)


parasgarg-bitgo (changes requested):

parasgarg-bitgo (approved):

at31416:
Can you check do we need to add node rpc urls also? ref:https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/environments.ts#L119

zahin-mohammad (approved):

parasgarg-bitgo:
> Can you check do we need to add node rpc urls also? ref:https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/environments.ts#L119

We don't need rpc URL for Non-BitGo recoveries for Eth-like coins. 
We use the explorer apis to get address nonce and balance. Fee params are provided by the client on WRW form.

PR #5627, opened 2025-02-24
fix(sdk-core): ed tss bug fix

islamaminBitGo:
TICKET: HSM-735

### DIFF
--- a/modules/sdk-core/src/account-lib/mpc/tss/eddsa/eddsa.ts
+++ b/modules/sdk-core/src/account-lib/mpc/tss/eddsa/eddsa.ts
@@ -136,8 +136,7 @@ export default class Eddsa {
             uShare.i
           );
         } catch (err) {
-          // TODO(BG-61036): Fix Verification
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/account-lib/mpc/tss/eddsa/eddsa.ts
```diff
@@ -136,8 +136,7 @@ export default class Eddsa {
             uShare.i
           );
         } catch (err) {
-          // TODO(BG-61036): Fix Verification
-          // throw new Error(`Could not verify share from participant ${share.j}. Verification error: ${err}`);
```
zahin-mohammad, line 140: Why did we comment this out before?

zhongxishen (approved):
Approving since it's kinda important.  Would like to see a test on failed vss verification.

islamaminBitGo:
> Approving since it's kinda important. Would like to see a test on failed vss verification.

https://bitgoinc.atlassian.net/browse/HSM-736

zahin-mohammad (approved):

PR #5634, opened 2025-02-25
feat(statics): add staking feature to babylon

sbplat:
TICKET: SC-1213

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -815,7 +815,7 @@ export const coins = CoinMap.fromCoins([
     6,
     UnderlyingAsset.BABY,
     BaseUnit.BABY,
-    COSMOS_SIDECHAIN_FEATURES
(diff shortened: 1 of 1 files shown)


modules/statics/src/coins.ts
```diff
@@ -815,7 +815,7 @@ export const coins = CoinMap.fromCoins([
     6,
     UnderlyingAsset.BABY,
     BaseUnit.BABY,
-    COSMOS_SIDECHAIN_FEATURES
+    [...COSMOS_SIDECHAIN_FEATURES, CoinFeature.STAKING]
```
pengyuc-bitgo, line 818: but isn't this already in the `COSMOS_SIDECHAIN_FEATURES` ?
https://github.com/BitGo/BitGoJS/blob/7e71064752d314b821b621c362b26303c210cfd1/modules/statics/src/coinFeatures.ts#L273

PR #5657, opened 2025-02-28
feat(sdk-coin-tao): add netuid validation

Vijay-Jagannathan:
Ticket: SC-1044

Only root subnet (0) is supported as of today

### DIFF
--- a/modules/abstract-substrate/src/lib/transactionBuilder.ts
+++ b/modules/abstract-substrate/src/lib/transactionBuilder.ts
@@ -256,6 +256,17 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     }
   }
 
+  /**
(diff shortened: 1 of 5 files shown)


modules/abstract-substrate/src/lib/transactionBuilder.ts
```diff
@@ -256,6 +256,17 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     }
   }
 
+  /**
+   * Helper method to check whether netuid is 0 (only root network is supported now)
+   * @param {string} netduid of the subnet
+   */
+  public validateNetuid(netduid: string) {
```
margueriteblair, line 263: We only check if it's 0? What if it's some other type of string value?

PR #5662, opened 2025-02-28
feat(statics): add list of gatekept coins and tokens

veetragjain:
TICKET: WIN-4743


modules/statics/src/gatekeep.ts
```diff
@@ -0,0 +1,1789 @@
...
+  'ofceth:telos',
+  'ofceth:cusdo',
+  'ofceth:vice',
+  'ofceth:god',
+  'ofceth:sky',
+  'ofceth:move',
+  'ofceth:fuel',
+  'ofcbera',
+  'ofcoas',
+  'bera:ibera',
```
gianchandania, line 1788: gatekeep file in WP is only supposed to have tokens. If we have to hide any coins that we are onboarding, they need to go in hiddenCoins in production.js file. Mentioning because the commit message says gatekept coins as well, but this file doesn't have coins generally. so any plans on adding hiddenCoins to the statics?

veetragjain: I've seen some coins also being added to this list as well. Anyway the idea here is that for token onboardings, we just bump statics and update no other file on wp so we just need to ensure that all 'hidden and gatekept' tokens are present in this file.

gianchandania (approved):

PR #5688, opened 2025-03-04
feat(sdk-api): update sdk to support Non-JSON response

xintongli-bitgo:
Ticket: DE-1849

The `/api/v2/admin/reports/${reportId}/document` wallet platform api that downloads a report does not a return a json response. Instead, it directly returns the report data with a `Content-Type` of either 'application/pdf' or 'text/csv'. When calling this API using the BitGo SDK, the response can’t be decoded properly because it is missing a `body` field.

To address this, we are updating the API SDK to support this type of response.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -28,6 +28,7 @@ module.exports = {
         'CR-',
         'CS-',
         'CSI-',
+        'DE-',
         'DES-',
         'DO-',
         'DOS-',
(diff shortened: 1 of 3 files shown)


FloBitGo (dismissed):
I think we should distinctly handle `text` responses, but otherwise I think it's fine.

FloBitGo (approved):

PR #5695, opened 2025-03-05
fix(sdk-core): multisig type should not be a required type

lokesh-bitgo:
Ticket: WP-3785

### DIFF
--- a/modules/bitgo/test/v2/unit/wallets.ts
+++ b/modules/bitgo/test/v2/unit/wallets.ts
@@ -25,6 +25,7 @@ import {
   BulkWalletShareOptions,
   KeychainWithEncryptedPrv,
   WalletWithKeychains,
+  multisigTypes,
 } from '@bitgo/sdk-core';
 import { BitGo } from '../../../src';
 import { afterEach } from 'mocha';
@@ -184,6 +185,7 @@ describe('V2 Wallets:', function () {
   });
 
   describe('Generate wallet:', function () {
+    const sandbox = sinon.createSandbox();
     it('should validate parameters', async function () {
       let params = {};
       await wallets.generateWallet(params).should.be.rejectedWith('Missing parameter: label');
@@ -475,6 +477,42 @@ describe('V2 Wallets:', function () {
         params.passphrase
       );
     });
+
+    it('should generate hot onchain wallet without passing multisig type', async () => {
+      const params: GenerateWalletOptions = {
+        label: 'test wallet',
+        passphrase: 'multisig password',
+        enterprise: 'enterprise',
+        passcodeEncryptionCode: 'originalPasscodeEncryptionCode',
+      };
+
+      const walletNock = nock(bgUrl)
+        .post('/api/v2/tbtc/wallet/add', function (body) {
+          body.type.should.equal('hot');
+          return true;
+        })
+        .reply(200);
+
+      nock(bgUrl)
+        .post('/api/v2/tbtc/key', _.matches({ source: 'bitgo' }))
+        .reply(200, { pub: 'bitgoPub' });
+      nock(bgUrl).post('/api/v2/tbtc/key', _.matches({})).reply(200);
+      nock(bgUrl)
+        .post('/api/v2/tbtc/key', _.matches({ source: 'backup' }))
+        .reply(200, { pub: 'backupPub' });
+
+      const generateWalletSpy = sandbox.spy(wallets, 'generateWallet');
+      const response = await wallets.generateWallet(params);
+      walletNock.isDone().should.be.true();
+      sinon.assert.calledOnce(generateWalletSpy);
+      assert.equal(generateWalletSpy.firstCall?.args[0]?.multisigType, multisigTypes.onchain);
+      assert.ok(response.encryptedWalletPassphrase);
+      assert.ok(response.wallet);
+      assert.equal(
+        bitgo.decrypt({ input: response.encryptedWalletPassphrase, password: params.passcodeEncryptionCode }),
+        params.passphrase
+      );
+    });
   });
 
   describe('Generate TSS wallet:', function () {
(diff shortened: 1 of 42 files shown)


pranavjain97 (changes requested):
Changes look good. Please also add a test that captures a req with no `multisigType` param and ensure that `multisigType` is being set correctly to the API call for a tss and a non-tss coin.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -605,6 +617,20 @@ export class Wallets implements IWallets {
...
+   * @param baseCoin
+   * @returns {string | undefined}
+   */
+  private getDefaultMultisigType(baseCoin: IBaseCoin): MultisigType | undefined {
+    if (baseCoin.supportsMultisig()) {
+      return multisigTypes.onchain;
+    } else if (baseCoin.supportsTss()) {
+      return multisigTypes.tss;
+    }
+    return undefined;
```
kaustubhbitgo, line 631: there is an additional method `supportsBlsDkg()` which should set the type to `blsdkg`.

lokesh-bitgo: Had discussion on this with Zahin. It's not require please check slack thread have tagged you there.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -243,6 +250,11 @@ export class Wallets implements IWallets {
   async generateWallet(
     params: GenerateWalletOptions = {}
   ): Promise<WalletWithKeychains | LightningWalletWithKeychains> {
+    // Assign the default multiSig type value based on the coin
+    if (!params.multisigType) {
+      params.multisigType = this.getDefaultMultisigType(this.baseCoin);
```
zahin-mohammad, line 255: Nit: this could also be a coin method right?

zahin-mohammad (dismissed):

modules/sdk-core/src/bitgo/wallet/wallets.ts
lokesh-bitgo: Yeah, It should.

modules/sdk-core/src/bitgo/wallet/wallets.ts
mullapudipruthvik, line 631: this method should be in baseCoin cause for coins which supports both multisig and tss 
this method always returns `onchain` 
there is no option to default to `tss`

modules/bitgo/test/v2/unit/wallets.ts
```diff
@@ -552,6 +552,52 @@ describe('V2 Wallets:', function () {
...
+      const params = {
+        label: 'tss wallet',
+        passphrase: 'tss password',
+        enterprise: 'enterprise',
+        passcodeEncryptionCode: 'originalPasscodeEncryptionCode',
+      };
+
+      const response = await wallets.generateWallet(params);
+
+      walletNock.isDone().should.be.true();
```
pranavjain97, line 591: Your wallet nock needs to have the expected multisigType and make sure the API is called with that value. This is currently not testing anything

pranavjain97 (changes requested):

modules/bitgo/test/v2/unit/wallets.ts
lokesh-bitgo: Added test case.

pranavjain97 (dismissed):

modules/sdk-coin-bsc/src/bsc.ts
```diff
@@ -26,6 +26,16 @@ export class Bsc extends AbstractEthLikeNewCoins {
...
   }
 
+  /** {@inheritDoc} */
+  supportsMultisig(): boolean {
+    return true;
+  }
+
+  /** inherited doc */
+  getDefaultMultisigType(): MultisigType {
+    return multisigTypes.onchain;
```
mullapudipruthvik, line 36: I think for bsc its `tss`

lokesh-bitgo: bsc supports both MultiSig and TSS. So for this we are considering the default value `onchain`.

mullapudipruthvik: I don't think bsc supports multisig 
https://github.com/BitGo/bitgo-microservices/blob/eb856019f52d2d27ddb28b713496bc498d307825/packages/wallet-platform/app/controllers/api/v2/coins/bsc.ts#L109

mullapudipruthvik: am i missing something ?

lokesh-bitgo: @mullapudipruthvik 
Can you please check at the line no 51 in the same link you have shared.

mullapudipruthvik: that looks wrong, right now we explicitly pass multisigType as `tss` for bsc wallet creation 
passing it as `multisig` will break wallet creation

lokesh-bitgo: As I referred the [above link](https://github.com/BitGo/bitgo-microservices/blob/eb856019f52d2d27ddb28b713496bc498d307825/packages/wallet-platform/app/controllers/api/v2/coins/bsc.ts#L52), It supported both multisig and tss. Let me confirm on this.

lokesh-bitgo: Have discussed this with the team and proceeding with the necessary changes.

lokesh-bitgo: @mullapudipruthvik 
Have pushed the changes please review it.

modules/sdk-coin-bsc/src/bsc.ts
```diff
@@ -26,6 +26,16 @@ export class Bsc extends AbstractEthLikeNewCoins {
     return true;
   }
 
+  /** {@inheritDoc} */
+  supportsMultisig(): boolean {
+    return true;
```
sachushaji, line 31: BSC doesn't support multisig

modules/sdk-core/src/coins/ofc.ts
```diff
@@ -58,6 +58,11 @@ export class Ofc extends BaseCoin {
     return 'Offchain';
   }
 
+  /** {@inheritDoc } **/
+  supportsMultisig(): boolean {
```
sachushaji, line 62: I don't know this a good practice to use it for ofc because, ofc is a abstraction and can support multiple chains under it which can or not support multisig, Insted of defaulting like this we should bring in chain specific context on which token the ofc is representing and should return the value appropriately

modules/sdk-coin-bsc/src/bsc.ts
sachushaji: BSC doesn't support multisig

sachushaji: I think the line in wp is actually wrong, we were planning to introduce it at one point but I think we dropped it

mullapudipruthvik (commented):
please squash your commits

modules/sdk-coin-bsc/src/bsc.ts
mullapudipruthvik, line 31:
```suggestion
```

modules/sdk-coin-bsc/src/bsc.ts
lokesh-bitgo: Removed.

modules/sdk-coin-bsc/src/bsc.ts
lokesh-bitgo: Removed.

modules/sdk-core/src/coins/ofc.ts
lokesh-bitgo: Removed for now.

modules/sdk-coin-bsc/src/bsc.ts
lokesh-bitgo: Done.

mullapudipruthvik (approved):

kisslove-dewangan (approved):

lcovar (changes requested):
This is something that should live in Statics instead of another interface method. The interface will never stop growing.

kaustubhbitgo (approved):

lcovar (approved):

abhishekagrawal080 (approved):
LGTM

PR #5699, opened 2025-03-05
fix(sdk-coin-icp): correct buffer handling for sender principal and principal conversion

mohd-kashif:


DinshawKothari:
Check commit lint please.

mukeshsp (approved):

PR #5703, opened 2025-03-05
chore: make it look extra cute 😎

lcovar:
Remove the usage of virtual sizes in backup recovery and stick to only dimensions class. .

Ticket: BTC-1844

TICKET: BTC-1844


davidkaplanbitgo (approved):
😸

PR #5704, opened 2025-03-05
fix: default to api version full if none specified

mohammadalfaiyazbitgo:
Ticket: WP-3872

TICKET: WP-3872

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -2898,6 +2898,10 @@ export class Wallet implements IWallet {
       throw new Error(`${this.baseCoin.getFullName()} does not allow account consolidations.`);
     }
 
+    const apiVersion =
+      params.apiVersion ??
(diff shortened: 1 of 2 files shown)


zahin-mohammad (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2898,6 +2898,10 @@ export class Wallet implements IWallet {
       throw new Error(`${this.baseCoin.getFullName()} does not allow account consolidations.`);
     }
 
+    const apiVersion =
+      params.apiVersion ??
+      (this.tssUtils && this.tssUtils.supportedTxRequestVersions().includes('full') ? 'full' : undefined);
+
```
pengyuc-bitgo, line 2904: Just a question:
Does the `buildAccountConsolidations` not call the v2.walllet.tx.send if the apiVersion is `full`?

pengyuc-bitgo: So, its to pass the apiVersion the user intended instead of falling to the default behavior mistakenly.

pengyuc-bitgo (approved):

PR #5718, opened 2025-03-07
feat(sdk-coin-icp): enhance address validation and add memo ID support

mohd-kashif:
TICKET: WIN-4759

### DIFF
--- a/modules/sdk-coin-icp/src/lib/utils.ts
+++ b/modules/sdk-coin-icp/src/lib/utils.ts
@@ -37,13 +37,31 @@ export class Utils implements BaseUtils {
   }
 
   /**
-   * Checks if the provided address is a valid hexadecimal string.
+   * Checks if the provided address is a valid ICP address.
    *
    * @param {string} address - The address to validate.
-   * @returns {boolean} - Returns `true` if the address is a valid 64-character hexadecimal string, otherwise `false`.
+   * @returns {boolean} - Returns `true` if the address is valid, otherwise `false`.
    */
   isValidAddress(address: string): boolean {
-    return this.isValidHash(address);
+    const rootAddress = this.validateMemoAndReturnRootAddress(address);
+    return rootAddress !== undefined && this.isValidHash(rootAddress);
+  }
+
+  /**
+   * Validates the memo ID in the address and returns the root address.
+   *
+   * @param {string} address - The address to validate and extract the root address from.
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-icp/src/lib/utils.ts
```diff
@@ -268,32 +286,51 @@ export class Utils implements BaseUtils {
...
+   * @param {string} fee - The fee to validate.
+   * @throws {BuildTransactionError} - If the fee is zero or invalid.
+   */
+  validateFee(fee: string): boolean {
+    const feeValue = new BigNumber(fee);
+    if (feeValue.isZero()) {
+      throw new BuildTransactionError('Fee cannot be zero');
     }
-    if (fee !== this.gasData()) {
+    if (!feeValue.isEqualTo(this.gasData())) {
```
DinshawKothari, line 300: Why fee should be always equal to gasData ? It is possible to have more fee also right.

modules/sdk-coin-icp/src/lib/utils.ts
```diff
@@ -268,32 +286,51 @@ export class Utils implements BaseUtils {
...
-    if (Number(memo) < 0 || Number(memo) === null || Number(memo) === undefined || Number.isNaN(Number(memo))) {
+  /**
+   * Validates the provided memo.
+   *
+   * @param {number | BigInt} memo - The memo to validate.
+   * @returns {boolean} - Returns `true` if the memo is valid.
+   * @throws {BuildTransactionError} - If the memo is invalid.
+   */
+  validateMemo(memo: number | BigInt): boolean {
+    const memoNumber = Number(memo);
```
at31416, line 322: Memo will always be a number in ICP? asking this because recently we had a CR where we had to allow alphanumeric as memoIds for cosmos chain coins. https://github.com/BitGo/BitGoJS/pull/5591/files

mohd-kashif: as of now, memo for ICP is number because there some concurrency logic for memo combined with `created_time` at ICP side
@DinshawKothari

DinshawKothari (approved):

modules/sdk-coin-icp/src/lib/utils.ts
DinshawKothari: @at31416  ICP only supports numeric memo at the chain level itself. So there is no concern here.

PR #5723, opened 2025-03-08
feat(): send lightning payment after pending approval

lcovar:
Add handling to submit lightning payments after they are approved in the pending approval flow. For lightning transactions, we need to send the payment request after approval rather than recreating and signing like other TSS transactions.

Ticket: BTC-1898


saravanan7mani (approved):

margueriteblair (commented):
No need to test these changes?

lcovar:
> No need to test these changes?

will follow up with tests

PR #5734, opened 2025-03-11
fix: pass isBatch in send API

kamleshmugdiya:
TICKET: COIN-3387


modules/sdk-coin-polygon/test/unit/polygon.ts
```diff
@@ -221,6 +222,94 @@ describe('Polygon', function () {
       halfSignedRawTx.halfSigned.eip1559.maxFeePerGas.should.equal('7593123');
       halfSignedRawTx.halfSigned.eip1559.maxPriorityFeePerGas.should.equal('150');
     });
+
+    it('should include isBatch field in halfSigned when txPrebuild has isBatch=true', async function () {
```
shivam-bitgo, line 226: nit- can we change the test name such that it says when it's a batch transaction in place of isBatch=true

kamleshmugdiya: remaned to should include isBatch field in halfSigned when signing a native batch transaction

shivam-bitgo (approved):

PR #5749, opened 2025-03-12
feat(sdk-coin-stx): modified explain transaction to accommodate token transfer transactions

ravibitgo:
Ticket: [COIN-3291](https://bitgoinc.atlassian.net/browse/COIN-3291)

[COIN-3291]: https://bitgoinc.atlassian.net/browse/COIN-3291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-coin-stx/src/stx.ts
+++ b/modules/sdk-coin-stx/src/stx.ts
@@ -9,6 +9,7 @@ import {
   VerifyTransactionOptions,
 } from '@bitgo/sdk-core';
 import { BaseCoin as StaticsBaseCoin, CoinFamily, coins } from '@bitgo/statics';
+import { cvToString, cvToValue } from '@stacks/transactions';
 import { ExplainTransactionOptions, StxSignTransactionOptions, StxTransactionExplanation } from './types';
 import { StxLib } from '.';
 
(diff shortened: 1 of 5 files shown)


modules/sdk-coin-stx/src/stx.ts
```diff
@@ -182,24 +183,41 @@ export class Stx extends BaseCoin {
     const txJson = tx.toJson();
 
     if (tx.type === TransactionType.Send) {
-      const outputs: TransactionRecipient[] = [
-        {
+      // check if it is a token transaction or native coin transaction
+      let transactionRecipient: TransactionRecipient;
+      let outputAmount: string;
+      let memo = '';
```
gianchandania, line 189: if memo was not passed for a native tx, was it empty string or null in the final output?

ravibitgo: it is an optional property in `StxTransactionExplanation`, so if in native tx memo is not passed then it will be null, I'll make it same

gianchandania (approved):

PR #5755, opened 2025-03-12
fix: ensure user params are used for tx verification

mohammadalfaiyazbitgo:
Ticket: WP-3966

TICKET: WP-3966

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -2147,7 +2147,7 @@ export class Wallet implements IWallet {
 
     try {
       await this.baseCoin.verifyTransaction({
-        txParams: txPrebuild.buildParams || params,
+        txParams: { ...txPrebuild.buildParams, ...params },
         txPrebuild,
         wallet: this,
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2126,7 +2126,7 @@ export class Wallet implements IWallet {
 
     try {
       await this.baseCoin.verifyTransaction({
-        txParams: txPrebuild.buildParams || params,
+        txParams: { ...txPrebuild.buildParams, ...params },
```
mullapudipruthvik, line 2129: do we want to override buildParams with params?

mullapudipruthvik (commented):
please fix your tests

modules/sdk-core/src/bitgo/wallet/wallet.ts
mohammadalfaiyazbitgo: yes because they come from platform, and we want to validate the transaction that platform builds against the params the user specified.

zahin-mohammad (commented):
flush

zahin-mohammad (approved):

PR #5764, opened 2025-03-13
chore: returning transfer object with coinspecific object

007harshmahajan:
returning transfer object with coinspecific object
TICKET: BTC-1919

### DIFF
--- a/modules/abstract-lightning/src/wallet/lightning.ts
+++ b/modules/abstract-lightning/src/wallet/lightning.ts
@@ -33,6 +33,7 @@ export type PayInvoiceResponse = {
   pendingApproval?: PendingApprovalData;
   // Absent if there's a pending approval
   paymentStatus?: LndCreatePaymentResponse;
+  transfer?: any;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


modules/abstract-lightning/src/wallet/lightning.ts
```diff
@@ -254,13 +255,18 @@ export class LightningWallet implements ILightningWallet {
     );
 
     const coinSpecific = transactionRequestSend.transactions?.[0]?.unsignedTx?.coinSpecific;
+    let transfer;
```
saravanan7mani, line 258: add Transfer type if it exists.

modules/abstract-lightning/src/wallet/lightning.ts
```diff
@@ -33,6 +33,7 @@ export type PayInvoiceResponse = {
   pendingApproval?: PendingApprovalData;
   // Absent if there's a pending approval
   paymentStatus?: LndCreatePaymentResponse;
+  transfer?: any;
```
007harshmahajan, line 36: need to change transfer type here

PR #5766, opened 2025-03-13
chore: remove unnecessary express endpoints and use subType for wallet

lcovar:
After conversation with the custody operations team, it makes more sense to have a hot wallet type than a custodial wallet type for lightning - these are the changes to reflect what has been done on the rest of our infrastructure.

Additionally, we remove some of the express endpoints that are not required and can be sent directly to the backend

Ticket: BTC-1922


saravanan7mani (dismissed):

davidkaplanbitgo (dismissed):

saravanan7mani (dismissed):

saravanan7mani (dismissed):

saravanan7mani (approved):

modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -80,9 +81,7 @@ export const GenerateLightningWalletOptionsCodec = t.strict(
     passphrase: t.string,
     enterprise: t.string,
     passcodeEncryptionCode: t.string,
-    // custodial - bitgo controls the node private key
-    // hot - client controls the node private key by running remote signer node
-    type: t.union([t.literal('custodial'), t.literal('hot')]),
+    subType: t.union([t.literal('lightningCustody'), t.literal('lightningSelfCustody')]),
```
margueriteblair, line 84: A bit confused, for my own clarification, are we removing `type` entirely? Or now we'll have both `type` and `subType` field?

davidkaplanbitgo: user is going to only input the `subType` because lighting can only be hot. When we make the request to wallet platform it will also have `type: 'hot'`

lcovar: no need for type. we are going to always use hot do client only needs to pass subType

margueriteblair (approved):

PR #5780, opened 2025-03-17
feat(fiat): add aed and sgd

drew-bitgo:
TICKET: FIAT-176

Add United Arab Emirates Dirham and Singapore Dollar


sandra-wang-bitgo:
flushing

modules/sdk-core/src/coins/fiataed.ts
```diff
@@ -0,0 +1,76 @@
...
+  MethodNotImplementedError,
+  ParsedTransaction,
+  ParseTransactionOptions,
+  SignedTransaction,
+  SignTransactionOptions,
+  VerifyAddressOptions,
+  VerifyTransactionOptions,
+} from '../';
+
+export class FiatAED extends BaseCoin {
```
zahin-mohammad, line 17: Would it be easier to just create a new package for the fiat coins, `abstract-fiat`? And then each of these fiat coins can extend a common FiatBaseCoin class.

zahin-mohammad (approved):

PR #5784, opened 2025-03-17
feat(sdk-coin-icp): implement signing functionality and improve transaction validation

mohd-kashif:
TICKET: WIN-4880

### DIFF
--- a/modules/sdk-coin-icp/src/lib/transferBuilder.ts
+++ b/modules/sdk-coin-icp/src/lib/transferBuilder.ts
@@ -1,6 +1,6 @@
 import { BaseCoin as CoinConfig } from '@bitgo/statics';
 import { TransactionBuilder } from './transactionBuilder';
-import { BaseTransaction, BuildTransactionError, MethodNotImplementedError } from '@bitgo/sdk-core';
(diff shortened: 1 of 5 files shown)


modules/sdk-coin-icp/src/lib/transferBuilder.ts
```diff
@@ -123,7 +123,9 @@ export class TransferBuilder extends TransactionBuilder {
   }
 
   /** @inheritdoc */
-  protected signImplementation(): BaseTransaction {
-    throw new MethodNotImplementedError();
+  protected signImplementation(key: BaseKey): BaseTransaction {
```
DinshawKothari, line 126: where is this method getting consumed ?

DinshawKothari (approved):

modules/sdk-coin-icp/src/lib/transferBuilder.ts
mohd-kashif: we call the sign() method of baseTransactionBuilder from where the signImplementation is called

```sign(key: BaseKey): void {
    this.validateKey(key);
    if (!this.transaction.canSign(key)) {
      throw new SigningError('Private key cannot sign the transaction');
    }

    this.transaction = this.signImplementation(key);
  }```

PR #5799, opened 2025-03-19
feat(ci): move dependency audit step to code quality matrix

OttoAllmendinger:
Move the dependency audit step from its own workflow into the 
code quality matrix to streamline our CI process and reduce 
overall build time.

Issue: BTC-0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,9 +59,6 @@ jobs:
       - name: Check In-Repo Package Versions
         run: yarn run check-versions
 
-      - name: Dependency Audit
-        run: yarn run improved-yarn-audit --min-severity high
-
       - name: build packages
         env:
           # Workaround for https://github.com/nodejs/node/issues/51555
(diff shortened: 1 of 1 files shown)


ericli-bitgo (dismissed):

andrew-scott-fischer (commented):
Nothing is wrong with these changes, but it's strange to create a matrix which runs one of five steps exclusively depending on its key. Each of these matrix jobs is either better defined as a separate job rather than one branch of a matrix or as a single job with five sequential steps (this will reduce the cost of this whole matrix by whatever the build time is times 4).

What is the long bar in BitGoJS CI? If it's not this matrix, I would unroll it and just have one job with each of these five steps run sequentially. 

If it is this step, I would make each of these matrix jobs its own step so there's more meta data associated with each step.

zahin-mohammad (approved):

zahin-mohammad:
> Nothing is wrong with these changes, but it's strange to create a matrix which runs one of five steps exclusively depending on its key. Each of these matrix jobs is either better defined as a separate job rather than one branch of a matrix or as a single job with five sequential steps (this will reduce the cost of this whole matrix by whatever the build time is times 4).
> 
> What is the long bar in BitGoJS CI? If it's not this matrix, I would unroll it and just have one job with each of these five steps run sequentially.
> 
> If it is this step, I would make each of these matrix jobs its own step so there's more meta data associated with each step.

There is definitely room for improvement here

PR #5801, opened 2025-03-19
chore: add security exclusion for jspdf

mohammadalfaiyazbitgo:
Ticket: WP-4055

TICKET: WP-4055

### DIFF
--- a/.iyarc
+++ b/.iyarc
@@ -1,3 +1,4 @@
 GHSA-3h5v-q93c-6h6q
 GHSA-8hc4-vh64-cxmj
 GHSA-9wv6-86v2-598j
+GHSA-w532-jxjh-hjhj


andrew-scott-fischer (changes requested):
From [WP-4055](https://bitgoinc.atlassian.net/browse/WP-4055)

> adding exclusion for https://github.com/advisories/GHSA-w532-jxjh-hjhjCan't find link  , we don’t use jspdf in platform to generate PDFs. PDF generation is done on the client with the sdk. 

If we don't use it, why not remove it as a dependency? Ignoring a [CVE with an 8.7 score ](https://github.com/advisories/GHSA-w532-jxjh-hjhj)is not a great practice.

mohammadalfaiyazbitgo:
> From [WP-4055](https://bitgoinc.atlassian.net/browse/WP-4055)
> 
> > adding exclusion for https://github.com/advisories/GHSA-w532-jxjh-hjhjCan't find link  , we don’t use jspdf in platform to generate PDFs. PDF generation is done on the client with the sdk.
> 
> If we don't use it, why not remove it as a dependency? Ignoring a [CVE with an 8.7 score ](https://github.com/advisories/GHSA-w532-jxjh-hjhj)is not a great practice.

@andrew-scott-fischer  The SDK uses it to generate Key cards, but it's on the client end.

andrew-scott-fischer:
ok, but then you should update the dependency, even if it's a transitive one.

PR #5810, opened 2025-03-20
fix(sdk-coin-sui): detect sui mainnet

sbplat:
TICKET: SC-1399


benkelcher-bitgo (approved):

modules/sdk-coin-sui/src/lib/walrusStakingBuilder.ts
```diff
@@ -33,7 +33,7 @@ export class WalrusStakingBuilder extends TransactionBuilder<WalrusStakingProgra
     this._transaction = new StakingTransaction(_coinConfig);
 
     // TODO improve mainnet vs. testnet configuration
```
islamaminBitGo, line 35: was this TODO addressed via this PR , or are there further improvements to be made?

sbplat: I guess we can refactor this logic into the Sui utils in another pr

islamaminBitGo: please create a ticket and update the TODO comment to include that ticket so we dont forget to do this down the line

PR #5817, opened 2025-03-21
feat(sdk-core): add GoStakingWallet

noel-bitgo:
- initial support for staking out of BitGo go accounts
- stake and unstake requests
- sign
- get requests

SC-1389

TICKET: SC-1389

### DIFF
--- a/modules/sdk-core/src/bitgo/staking/goStakingWallet.ts
+++ b/modules/sdk-core/src/bitgo/staking/goStakingWallet.ts
@@ -1,0 +1,119 @@
+/**
+ * @prettier
+ */
+import { BitGoBase } from '../bitgoBase';
+import { IWallet, PrebuildTransactionResult } from '../wallet';
+import {
+  BaseGoStakeOptions,
+  FrontTransferSendRequest,
+  GoStakeFinalizeOptions,
+  GoStakeOptions,
+  GoStakingRequest,
+  IGoStakingWallet,
+  UnsignedGoStakingRequest,
+} from './iGoStakingWallet';
+import assert from 'assert';
+
+export class GoStakingWallet implements IGoStakingWallet {
+  private readonly bitgo: BitGoBase;
+
+  public wallet: IWallet;
+
+  constructor(wallet: IWallet) {
+    this.wallet = wallet;
+    this.bitgo = wallet.bitgo;
+  }
+
+  get accountId(): string {
+    return this.wallet.id();
+  }
+
+  async stake(options: GoStakeOptions): Promise<GoStakingRequest> {
+    // call preview
+    const preview = await this.previewStake({
+      amount: options.amount,
+      clientId: options.clientId,
+    } as BaseGoStakeOptions);
+
+    // sign the transaction
+    const halfSignedTransaction = (await this.wallet.prebuildAndSignTransaction({
+      walletPassphrase: options.walletPassphrase,
+      prebuildTx: {
+        payload: preview.payload,
+      } as PrebuildTransactionResult,
+    })) as FrontTransferSendRequest;
+
+    // call finalize to submit the go staking request to go staking service
+    assert(halfSignedTransaction.halfSigned?.payload, 'missing payload in half signed transaction');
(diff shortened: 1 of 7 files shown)


modules/sdk-core/src/bitgo/staking/goStakingWallet.ts
```diff
@@ -0,0 +1,136 @@
...
+    // NOTE: using this.wallet.prebuildAndSignTransaction did not work as expected because on signing step the coin was OFC instead of OFC token
+    // setup ofc coin to sign the transaction
+    const halfSignedTransaction = (await this.bitgo.coin(`ofc${coin}`).signTransaction({
+      txPrebuild: {
+        payload: preview.payload,
+      },
+      prv: this.wallet.getUserPrv({
+        keychain: keychain[0],
+        walletPassphrase: options.walletPassphrase,
+      } as GetUserPrvOptions),
```
bitgoAaron, line 52: don't think the `as GetUserPrvOptions` is necessary, best to avoid unnecessary type casting like this

bitgoAaron (changes requested):
remove 'coin' as function params from all functions and use this.wallet.baseCoin instead

modules/sdk-core/src/bitgo/staking/goStakingWallet.ts
```diff
@@ -0,0 +1,136 @@
...
+    } as BaseGoStakeOptions);
+
+    // get keychain from wallet
+    const keychain = await this.wallet.baseCoin.keychains().getKeysForSigning({
+      wallet: this.wallet,
+    });
+
+    // NOTE: using this.wallet.prebuildAndSignTransaction did not work as expected because on signing step the coin was OFC instead of OFC token
+    // setup ofc coin to sign the transaction
+    const halfSignedTransaction = (await this.bitgo.coin(`ofc${coin}`).signTransaction({
```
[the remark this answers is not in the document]
noel-bitgo: i am having issues in the test doing `this.wallet.baseCoin.signTransaction`

initiating the wallet as OFC, has basecoin as OFC not OfcToken, so signing throws an error.

Do you have an example of how to initiate a go account wallet that can sign with ofc token ?

modules/sdk-core/src/bitgo/staking/goStakingWallet.ts
noel-bitgo: updated

modules/sdk-core/src/bitgo/staking/goStakingWallet.ts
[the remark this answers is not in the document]
noel-bitgo: fixed it

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2721,6 +2721,13 @@ export class Wallet implements IWallet {
     return new StakingWallet(this, isEthTss);
   }
 
+  /**
+   * Create a go staking wallet from this wallet
+   */
+  toGoStakingWallet(): GoStakingWallet {
+    return new GoStakingWallet(this);
```
ryandengBG, line 2728: could we do something similar to

ryandengBG:
```
  toTradingAccount(): TradingAccount {
    if (this.baseCoin.getFamily() !== 'ofc') {
      throw new Error('Can only convert an Offchain (OFC) wallet to a trading account');
    }
    return new TradingAccount(this._wallet.enterprise, this, this.bitgo);
  }
```

to only allow go account wallets to be convertable to go staking wallet

ryandengBG (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
noel-bitgo: let me check if family would be ofc always

ryandengBG (approved):

bitgoAaron (approved):

PR #5825, opened 2025-03-24
feat(express): add includeBalance option when fetching lightning wallets

lcovar:
Add includeBalance=false flag when fetching lightning wallets in express routes to optimize performance by skipping balance calculation when it's not needed. This improves response times for lightning wallet operations that don't require balance information.

Ticket: BTC-1946

### DIFF
--- a/modules/express/src/lightning/lightningSignerRoutes.ts
+++ b/modules/express/src/lightning/lightningSignerRoutes.ts
@@ -84,7 +84,7 @@ export async function handleInitLightningWallet(req: express.Request): Promise<u
     }
   );
 
-  const wallet = await coin.wallets().get({ id: walletId });
+  const wallet = await coin.wallets().get({ id: walletId, includeBalance: false });
   if (wallet.subType() !== 'lightningSelfCustody') {
     throw new ApiResponseError(`not a self custodial lighting wallet ${walletId}`, 400);
   }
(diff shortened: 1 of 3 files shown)


ericli-bitgo (dismissed):

modules/express/src/lightning/lightningSignerRoutes.ts
```diff
@@ -84,7 +84,7 @@ export async function handleInitLightningWallet(req: express.Request): Promise<u
     }
   );
 
-  const wallet = await coin.wallets().get({ id: walletId });
+  const wallet = await coin.wallets().get({ id: walletId, includeBalance: false });
```
saravanan7mani, line 87: `handleCreateSignerMacaroon` too

lcovar: I think pay should have it, if we can't get the balance for some reason then that means we cant connect to the node. So i think this is valid.

OttoAllmendinger (approved):
maybe `false` should be the default

kaustubhbitgo (approved):

lokesh-bitgo (approved):

lcovar:
> maybe `false` should be the default

I agree but unsure if clients rely on this at the moment. Might be a big breaking change.

PR #5809, opened 2025-03-20
feat(sdk-coin-apt): send many: addition of recipients (single recipient)

baltiyal:
TICKET: COIN-3327

This is part of PRs for send many support for Aptos.

- Added a new `recipients` array, currently containing a single recipient.
- Replaced internal (protected) `recipient` field with `recipients`.
- Ensured backward compatibility for existing recipient-based flows.
- Future PRs will extend `recipients` to support batch transactions with multiple recipients.

PR #5826, opened 2025-03-25
Revert "Revert "feat(sdk-coin-apt): send many: addition of recipients (single recipient)""

baltiyal:
Reverts BitGo/BitGoJS#5821

**Background**
Aptos withdrawals on the testnet were failing during broadcasting with the following error:
`Invalid transaction: Type: InvariantViolation Code: UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`

Since we couldn't immediately identify the root cause, we initially reverted related PRs, including #5809, as it introduced changes to the transaction structure and was a potential suspect.

**Root Cause**
Upon further investigation, we discovered that the issue was not caused by the changes in #5809. 
Instead, it stemmed from our **Aptos testnet fullnode running an outdated version**, which prevented it from broadcasting transactions properly.

**Resolution**
Now that we have identified the actual cause, we are reverting the previous revert, as #5821 had no impact on the failure.

### DIFF
--- a/modules/sdk-coin-apt/src/lib/iface.ts
+++ b/modules/sdk-coin-apt/src/lib/iface.ts
@@ -16,6 +16,7 @@ export interface TxData {
   id: string;
   sender: string;
   recipient: TransactionRecipient;
+  recipients: TransactionRecipient[];
   sequenceNumber: number;
   maxGasAmount: number;
   gasUnitPrice: number;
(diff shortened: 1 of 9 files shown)


yogeshwar-bitgo (approved):
LGTM

abhishekagrawal080 (approved):

at31416 (approved):
please add some description for a wider audience why we decided to revert and now reverting the revert back.

PR #5835, opened 2025-03-25
fix(sdk-core): validate unsigned staking transaction before sign

mohammadalfaiyazbitgo:
Ticket: WP-3967

TICKET: WP-3967


noel-bitgo (commented):
So basically if there is no tx request id we validate?

mohammadalfaiyazbitgo:
> So basically if there is no tx request id we validate?

Yes, I have a follow up for tx requests here: https://bitgoinc.atlassian.net/browse/SC-1435

noel-bitgo (approved):

PR #5839, opened 2025-03-26
feat(statics): get formatted tokens from custom coin map

veetragjain:
This PR adds a method to get formatted tokens list from a coin map, which if not passed, defaults to `coins` coin map.

Ticket: WIN-4944


modules/statics/src/tokenConfig.ts
```diff
@@ -761,6 +786,10 @@ const verifyTokens = function (tokens: BaseTokenConfig[]) {
   return verifiedTokens;
 };
 
+export const tokens = getFormattedTokens();
+
+export const formattedAlgoTokens = getFormattedAlgoTokens();
```
mullapudipruthvik, line 791: why do we need to export this here ?
can't you call the same function in algo sdk ?

veetragjain: I did not add this here newly, it was present in the file [before.](https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/tokenConfig.ts#L482). Just refactored it to use the new method.

veetragjain: All the other coin sdk use [token](https://github.com/BitGo/BitGoJS/blob/master/modules/statics/src/tokenConfig.ts#L603) object to get formatted tokens, but algo use formattedAlgoTokens directly. Refactoring algo sdk to use the token object that is used by all other coin sdk is out of the scope of current PR. Will have a follow up PR to refactor algo sdk.

mullapudipruthvik: FYI we are unnecessarily exports things which is not required

mullapudipruthvik (approved):

PR #5840, opened 2025-03-26
fix(abstract-eth): return eip 191 encoded message as a hex string

sbplat:
Ticket: SC-1424

BREAKING CHANGE: messageEncoded is now encoded as a hex string


margueriteblair (approved):
lgtm

PR #5842, opened 2025-03-27
feat: add approveToken function for erc20 sendMany

kamleshmugdiya:
TICKET: COIN-2785

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -31,6 +31,7 @@ import {
 import { SerializedNtilde } from '../../account-lib/mpc/tss/ecdsa/types';
 import { IAddressBook } from '../address-book';
 import { WalletUser } from '@bitgo/public-types';
+import { SubmitTransactionResponse } from '../inscriptionBuilder';
 
 export interface MaximumSpendableOptions {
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -899,4 +899,5 @@ export interface IWallet {
   fetchCrossChainUTXOs(params: FetchCrossChainUTXOsOptions): Promise<CrossChainUTXO[]>;
   getChallengesForEcdsaSigning(): Promise<WalletEcdsaChallenges>;
   getNftBalances(): Promise<NftBalance[]>;
+  approveToken(walletPassphrase: string, tokenName: string): Promise<any>;
```
margueriteblair, line 902: There's no more strict return type we can use here other than `any` ?

margueriteblair (dismissed):
Test please

modules/sdk-core/src/bitgo/wallet/iWallet.ts
kamleshmugdiya: using existing `SubmitTransactionResponse`.
also added test and an example script.

ewangbitgo (approved):

zahin-mohammad (approved):

PR #5845, opened 2025-03-28
fix: fund forwarders script

bhavidhingra:
The fund forwarders script did not work as it is in the BitGoJS examples.
It worked with theses modifications.
API reference: [v2.wallet.fundforwarders](https://developers.bitgo.com/api/v2.wallet.fundforwarders)

TICKET: COIN-3571

### DIFF
--- a/examples/ts/apt/fund-forwarders.ts
+++ b/examples/ts/apt/fund-forwarders.ts
@@ -1,0 +1,37 @@
+/**
+ * Send funds from a gas tank to a BitGo wallet's receive address.
+ *
+ * Copyright 2025, BitGo, Inc.  All Rights Reserved.
+ */
+import { BitGoAPI } from '@bitgo/sdk-api';
+import { Tapt } from '@bitgo/sdk-coin-apt';
+import { FundForwarderParams } from '@bitgo/sdk-core';
+
+const bitgo = new BitGoAPI({
+  accessToken: '<access token>',
+  env: 'test',
+});
+
+const coin = 'tapt';
+bitgo.register(coin, Tapt.createInstance);
+
+const walletId = '<wallet id>';
+
+async function main() {
+  bitgo.unlock({ otp: '000000' });
+  const walletInstance = await bitgo.coin(coin).wallets().get({ id: walletId });
+
+  const fundForwarderParams: FundForwarderParams = {
+    forwarders: [
+      {
+        forwarderAddress: '<forwarder address>',
+        amount: '<amount>' // optional field
+      },
+    ],
+  }
+
+  const response = await walletInstance.fundForwarders(fundForwarderParams);
+  console.log('Response', response);
+}
+
+main().catch((e) => console.error(e));
--- a/examples/ts/eth/fund-forwarder.ts
+++ b/examples/ts/eth/fund-forwarder.ts
@@ -1,3 +1,5 @@
+import { FundForwarderParams } from "@bitgo/sdk-core";
+
 /**
  * Send funds from a fee address to a forwarder.
  *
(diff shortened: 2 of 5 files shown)


examples/ts/eth/fund-forwarder.ts
```diff
@@ -28,12 +30,16 @@ async function fundForwarder() {
   bitgo.authenticateWithAccessToken({ accessToken });
   const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
 
-  const fundForwarderOptions = {
-    forwarderAddress: forwarderAddress,
-    amount: amount,
-  };
+  const fundForwarderParams: FundForwarderParams = {
+    forwarders: [
```
sachushaji, line 34: Was this a recent change in the API?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3624,18 +3624,17 @@ export class Wallet implements IWallet {
...
    * @param {Object} params - parameters object
-   * @param {String} params.forwarderAddress - Address of the forwarder to send funds to.
-   * @param {String} params.amount - Amount to send the forwarder (optional). If not given, defaults to sending an estimate of the amount needed for a fund recovery
+   * @param {List} params.forwarders - list of fund forwarder options
    * @returns {*}
    */
-  public async fundForwarder(params: FundForwardersOptions): Promise<any> {
-    if (_.isUndefined(params.forwarderAddress)) {
-      throw new Error('forwarder address required');
+  public async fundForwarders(params: FundForwarderParams): Promise<any> {
```
sachushaji, line 3633: This could be a breaking change and could break existing client integration, Was the change made in API?, I would suggest if its a recent change in API, fix the breaking change from API

bhavidhingra: This API doesn't exist in express server. Most probably this implementation is not being used. May be I'll not modify it and add a new method instead.

examples/ts/apt/fund-forwarders.ts
```diff
@@ -0,0 +1,37 @@
+/**
```
pranavjain97, line 1: This file shoudn't need a WP review. Please update the CODEOWNERS file accordingly

pranavjain97 (dismissed):
Please send me the PR to add any coin specific example folders to the coins team. I'll approve it. We're trying to get these upto date so teams can move faster and not unblocked on WP.

pranavjain97 (commented):
Apart from that, wallet changes LGTM

modules/sdk-core/src/bitgo/wallet/wallet.ts
bhavidhingra: Added a new method instead of modifying the existing one.

examples/ts/apt/fund-forwarders.ts
bhavidhingra: I've mapped the `examples` directory to `BitGo/ethalt-team`.

sijuu (approved):
Approving CODEOWNERS

modules/sdk-core/src/bitgo/wallet/wallet.ts
sachushaji:
```suggestion
  public async fundForwarders(params: FundForwarderParams): Promise<WalletData> {
```

sachushaji: we can type this one

pranavjain97 (approved):

PR #5846, opened 2025-03-28
feat(abstract-eth): add extra methods in transfer builder

venkateshv1266:
TICKET: COIN-3248

- `getSignatureData()` -> used for building unsigned txn for a custodial hot wallet. The output of this method will be passed to HSM for user signature.
- `setSignature()` -> used to set the hsm signature.


modules/abstract-eth/src/lib/transferBuilder.ts
```diff
@@ -111,6 +112,11 @@ export class TransferBuilder {
     return this;
   }
 
+  setSignature(signature: string): TransferBuilder {
+    this._signature = signature;
```
zhongxishen, line 116: Since `getSignature()` is defined as 
```
  protected getSignature(): string {
    if (this._signKey) {
      this._signature = this.ethSignMsgHash();
    }
    return this._signature!;
  }
```
Does it make sense to set the `_signKey` to a falsy value, if signature is directly set?

venkateshv1266: `_signKey` will be set only when we use `key()` method on the `transfer` object. So by default it would be `undefined`. So I don't think its required to set a falsy value.

modules/abstract-eth/src/lib/transferBuilder.ts
```diff
@@ -289,4 +295,17 @@ export class TransferBuilder {
       this._tokenContractAddress = transferData.tokenContractAddress;
     }
   }
+
+  public buildUnsignedTransaction(): Buffer<ArrayBuffer> {
```
mullapudipruthvik, line 299:
```suggestion
  public getSignatureData(): Buffer<ArrayBuffer> {
```
this is not an unsigned txn

mullapudipruthvik (commented):
please add a testcase in `sdk-coin-ethlike`

modules/abstract-eth/src/lib/transferBuilder.ts
zhongxishen: My concern is that it's possible to call `key()` on a transfer builder, and later on call `setSignature()` on the same builder for whatever reason.  Now, calling `signAndBuild()` on the builder may not get you what you'd expect, and this subtle cause may be a little tricky to debug.

What I was trying to say is, setting `_signKey` to `undefined` whenever `_signature` is set explicitly makes the API more 'ergonomic' and harder to misuse.

modules/abstract-eth/src/lib/transferBuilder.ts
venkateshv1266: Added Tests

modules/abstract-eth/src/lib/transferBuilder.ts
venkateshv1266: Got it. Makes sense

modules/abstract-eth/src/lib/transferBuilder.ts
```diff
@@ -111,6 +113,12 @@ export class TransferBuilder {
     return this;
   }
 
+  setSignature(signature: string): TransferBuilder {
+    this._signKey = undefined;
```
mullapudipruthvik, line 117:
```suggestion
    this._signKey = null;
```

venkateshv1266: done

mullapudipruthvik (approved):

PR #5849, opened 2025-03-28
refactor: remove unused bls and eth2 codebase

kisslove-dewangan:
Ticket: WP-3961


zahin-mohammad (commented):
We should also deprecate `https://www.npmjs.com/package/@bitgo/sdk-coin-eth2`

kisslove-dewangan:
> We should also deprecate `https://www.npmjs.com/package/@bitgo/sdk-coin-eth2`

As Andy mentioned, we don't need to unpublish now, and these changes have removed `sdk-coin-eth2` dependency/usages from the codebase. Let me know if anything else I need to take care of.

sijuu (dismissed):
Approving CODEOWNERS file change

sachushaji (commented):
I would suggest to do this only after we have removed this dep in downstream services like WP utilising this library. Other wise this will block a lot of folks

zahin-mohammad:
> I would suggest to do this only after we have removed this dep in downstream services like WP utilising this library. Other wise this will block a lot of folks

It shouldn't block people, consumers (including WP) would still have access to the existing published versions of the library. 

Only potential issue is just excluding this package when doing bumps, which is not hard to do.

zahin-mohammad (dismissed):
Lgtm.  Given this is not a product that is used at all, it should be safe to remove the backend impl before the sdk.

lokesh-bitgo (dismissed):

kaustubhbitgo (dismissed):
@kisslove-dewangan Instead of explicitly checking for non-nulls like `if (BigNumber) { ... }` use default export rather than named export when importing `BigNumber`. This should fix your undefined issue.
```diff
- import { BigNumber } from 'bignumber.js'; 
+ import BigNumber from 'bignumber.js';
```

kisslove-dewangan:
> @kisslove-dewangan Instead of explicitly checking for non-nulls like `if (BigNumber) { ... }` use default export rather than named export when importing `BigNumber`. This should fix your undefined issue.
> 
> ```diff
> - import { BigNumber } from 'bignumber.js'; 
> + import BigNumber from 'bignumber.js';
> ```

any idea why this issue was not happening in earlier PRs? now this is happening across codebase where bignumber is used

kaustubhbitgo:
> any idea why this issue was not happening in earlier PRs? now this is happening across codebase where bignumber is used

I just checked, this is happening in some of other currently opened PRs most likely due to `Remove yarn.lock` step & installing fresh dependencies.

The author of `bignumber.js` has published a new version `9.2.0` which introduced this breaking change https://github.com/MikeMcl/bignumber.js/commit/a543ab5444c7dd6e7ea1759949a4f5ca74fe8f98

kisslove-dewangan:
> @kisslove-dewangan Instead of explicitly checking for non-nulls like `if (BigNumber) { ... }` use default export rather than named export when importing `BigNumber`. This should fix your undefined issue.
> 
> ```diff
> - import { BigNumber } from 'bignumber.js'; 
> + import BigNumber from 'bignumber.js';
> ```

if we do this, then it won't be using below code:
```
declare module 'bignumber.js' {
  export declare class BigNumber implements BigNumber.Instance {
    toFormat(
      decimalPlaces: number | null,
      roundingMode: BigNumber.RoundingMode | null,
      format?: BigNumber.Format
    ): string;
  }
}
```

which is used in very few places so how can we handle this?

modules/bitgo/CHANGELOG.md
```diff
@@ -2533,7 +2533,6 @@ BTC-820
 - **bitgo:** add lightning and lnurl examples ([8894bac](https://github.com/BitGo/BitGoJS/commit/8894bac0fd47eb9841e545011dc25d516dd72e9c))
 - **bitgo:** adding example for tx-build ([c967e26](https://github.com/BitGo/BitGoJS/commit/c967e2630bb7768278572361c61667c0452af1ba))
 - **bitgo:** expose Ethw in core bitgo module ([183cda4](https://github.com/BitGo/BitGoJS/commit/183cda433f8c683722843e2c30bf46101a1cd677))
-- **bitgo:** support chaincodes on BLS-DKG keychains creation ([bfaa380](https://github.com/BitGo/BitGoJS/commit/bfaa380551d2fe90e041975b392d4398c781074a))
```
zahin-mohammad, line 2536: Don't manually update the changelog or remove things from it.

zahin-mohammad (changes requested):
```
refactor: remove unused bls and eth2 codebase
TICKET: [WP-3961](https://bitgoinc.atlassian.net/browse/WP-3961)
```


These are breaking changes, please update your commit to include the `BREAKING CHANGE` tag. This will trigger a new major version during the next release.

modules/bitgo/CHANGELOG.md
kisslove-dewangan: done

modules/sdk-core/CHANGELOG.md
```diff
@@ -3189,7 +3189,6 @@ Ticket: BG-68940
 - add message signing support for polygon ([ab2bac1](https://github.com/BitGo/BitGoJS/commit/ab2bac13dad55ce8571d014796298aa52a24a5f2))
 - add u value proof during tss eddssa key creation ([79d2c91](https://github.com/BitGo/BitGoJS/commit/79d2c91ea5b101f8cad9b107b9e4426939333c5f))
 - adding support for message signing ([01c6303](https://github.com/BitGo/BitGoJS/commit/01c63032d067e6ba5aef78804ea747b5e62709fe))
-- **bitgo:** support chaincodes on BLS-DKG keychains creation ([bfaa380](https://github.com/BitGo/BitGoJS/commit/bfaa380551d2fe90e041975b392d4398c781074a))
```
mullapudipruthvik, line 3192: lets not change this

mullapudipruthvik (dismissed):
non blocking

mullapudipruthvik (approved):
tests doesn't seem to run, rebase and see if it works

margueriteblair (approved):
same as mullapudi, lgtm, but rebase to see if tests run

zahin-mohammad (approved):

sijuu (approved):
Approving change in CODEOWNERS

PR #5855, opened 2025-03-28
feat(sdk-api): add dynamic headers in sdk requests

jakeyangbitgo:
TICKET: [ME-178](https://bitgoinc.atlassian.net/browse/ME-178)

Provide a way to dynamically set headers in sdk requests

[ME-178]: https://bitgoinc.atlassian.net/browse/ME-178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -395,8 +395,8 @@ export class BitGoAPI implements BitGoBase {
 
       req.set('BitGo-Auth-Version', this._authVersion === 3 ? '3.0' : '2.0');
 
+      const data = serializeRequestData(req);
       if (this._token) {
-        const data = serializeRequestData(req);
         setRequestQueryString(req);
 
         const requestProperties = this.calculateRequestHeaders({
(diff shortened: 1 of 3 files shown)


zahin-mohammad (commented):
can you update the jsdoc for this new param

pengyuc-bitgo (dismissed):

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -414,6 +416,14 @@ export class BitGoAPI implements BitGoBase {
         req.set('HMAC', requestProperties.hmac);
       }
 
+      if (this.getAdditionalHeadersCb) {
+        const data = serializeRequestData(req);
+        const additionalHeaders = this.getAdditionalHeadersCb(method, url, data);
+        for (const { key, value } of additionalHeaders) {
+          req.set(key, value);
+        }
+      }
```
bitgoAaron, line 425: move this up a few lines into the same block where we check for token, and re-use the data prop so we don't have to serialize the payload twice

bitgoAaron (changes requested):
looks great, just move the callback check up a few lines to reduce unnecessary serialization calls

bitgoAaron (approved):

PR #5861, opened 2025-03-31
feat(sdk-coin-icp): reduce the payload to 1

adarsh312:
Ticket: WIN-4983


modules/sdk-coin-icp/src/lib/signedTransactionBuilder.ts
```diff
@@ -48,33 +41,20 @@ export class SignedTransactionBuilder {
       for (const ingressExpiry of unsignedTransaction.ingress_expiries) {
```
DinshawKothari, line 41: As discussed, Can we please remove this for loop altogether. As unsignedTransaction.ingress_expiries will only have one ingress_expiry in it. Let's use it with index 0. Along with this please add a comment too mentioning that we only expect to create one ingress_expiry and one update payload with respect to this.

modules/sdk-coin-icp/src/lib/unsignedTransactionBuilder.ts
```diff
@@ -72,15 +72,6 @@ export class UnsignedTransactionBuilder {
         signature_type: SignatureType.ECDSA,
       };
       payloads.push(transactionPayload);
-
```
DinshawKothari, line 75: Same comment as above for the for loop here as well.

modules/sdk-coin-icp/src/lib/iface.ts
```diff
@@ -161,7 +160,6 @@ export interface ReadStateEnvelope {
```
DinshawKothari, line 160: We should remove ReadState and related interface also as we have removed it's usage.

modules/sdk-coin-icp/src/lib/unsignedTransactionBuilder.ts
adarsh312: have fixed it

modules/sdk-coin-icp/src/lib/signedTransactionBuilder.ts
adarsh312: i have made the code changes, have fetched from the first record in the array

modules/sdk-coin-icp/src/lib/signedTransactionBuilder.ts
```diff
@@ -45,38 +38,28 @@ export class SignedTransactionBuilder {
...
+      requestEnvelopes.push({
+        update: updateEnvelope,
+      });
 
-        requestEnvelopes.push({
-          update: updateEnvelope,
-          read_state: readStateEnvelope,
-        });
-      }
       envelopes.push([reqType, requestEnvelopes]);
```
mohd-kashif, line 63: do we still need the envelopes ?

adarsh312: we had a discussion on this, we decided to preserve it for future use cases.
lets say we want to add more states

modules/sdk-coin-icp/src/lib/iface.ts
DinshawKothari: This is still pending. Please do a cleanup in icp-module related to this change.

modules/sdk-coin-icp/src/lib/transactionBuilder.ts
```diff
@@ -23,8 +23,12 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     return this._transaction.signaturePayload;
   }
 
-  public unsignedTransaction(): string {
```
DinshawKothari, line 26: why this function is being removed ?

DinshawKothari: We still have its usage and transfer builder and other things also rely on this.

modules/sdk-coin-icp/src/lib/transactionBuilder.ts
```diff
@@ -23,8 +23,12 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
-    return this._transaction.payloadsData.unsigned_transaction;
+  public payloadData(): PayloadsData {
```
DinshawKothari, line 26: why is this method required ?

modules/sdk-coin-icp/src/lib/transactionBuilder.ts
```diff
@@ -23,8 +23,12 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+    return this._transaction.payloadsData;
+  }
+
+  public icpTransaction(): IcpTransaction {
```
DinshawKothari, line 30: why is this method required ?

modules/sdk-coin-icp/src/lib/signedTransactionBuilder.ts
DinshawKothari: I am okay if we want the envelop implementation to be intact with the new changes also. But how are we making sure that this signedTransaction that we use in the combine step is in broadcastable form for the fullnode and rosetta.
Please make relevant changes.

modules/sdk-coin-icp/test/unit/transactionBuilder/transactionBuilder.ts
```diff
@@ -51,10 +51,9 @@ describe('ICP Transaction Builder', async () => {
...
     should.equal(icpTransaction.operations[2].amount.value, '-10000');
     should.equal(icpTransaction.public_keys[0].hex_bytes, testData.accounts.account1.publicKey);
-    txBuilder.unsignedTransaction().should.be.a.String();
     payloadsData.payloads.should.be.an.Array();
-    payloadsData.payloads.length.should.equal(2);
-    should.equal(txBuilder.unsignedTransaction(), testData.payloadsData.unsigned_transaction);
+    payloadsData.payloads.length.should.equal(1);
+    should.deepEqual(txBuilder.payloadData(), testData.payloadsData);
     should.deepEqual(payloadsData.payloads, testData.payloadsData.payloads);
   });
```
DinshawKothari, line 59: It is recommended to have an integration test that builds an unsigned transaction, signs it using the SDK, generates the final broadcastable transaction, and submits it directly to the IC ledger instead of Rosetta. This will help identify any existing gaps more effectively.

modules/sdk-coin-icp/src/lib/transactionBuilder.ts
adarsh312: we need it for we don't need unsignedTransaction if we have payloadData() which is the parent data

modules/sdk-coin-icp/src/lib/transactionBuilder.ts
adarsh312: i don't see its usage in SDK, can you help me understand this ?

modules/sdk-coin-icp/src/lib/signedTransactionBuilder.ts
mohd-kashif: now the signed txn can only broadcasted to the fullnode and not to the rosetta

modules/sdk-coin-icp/src/lib/iface.ts
```diff
@@ -132,13 +131,7 @@ export interface CborUnsignedTransaction {
...
 }
 
-export interface ReadState {
-  sender: Uint8Array;
-  paths: Array<[Buffer, Buffer]>;
-  ingress_expiry: bigint;
-}
-
-export interface UpdateEnvelope {
+export interface Envelope {
```
DinshawKothari, line 134: why are we changing the name of this envelop ?
I would still suggest to keep this UpdateEnvelope as this gives clear idea that the following is used for creating update kind of envelop. In future there can be more types of envelops also.

modules/sdk-coin-icp/src/lib/iface.ts
```diff
@@ -132,13 +131,7 @@ export interface CborUnsignedTransaction {
...
   content: {
     request_type: RequestType;
```
DinshawKothari, line 136: As we know UpdateEnvelope will always have call type of request_type. So let's use that directly here rather than assigning it at the creation of UpdateEnvelope.

modules/sdk-coin-icp/src/lib/iface.ts
adarsh312: this is completed

modules/sdk-coin-icp/src/lib/iface.ts
```diff
@@ -151,17 +144,8 @@ export interface UpdateEnvelope {
...
-    request_type: RequestType;
-  };
-  sender_pubkey: Uint8Array;
-  sender_sig: Uint8Array;
-}
-
 export interface RequestEnvelope {
   update: UpdateEnvelope;
-  read_state: ReadStateEnvelope;
 }
```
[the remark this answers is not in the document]
DinshawKothari: For this reason only, I was referring to an integration test where we mock the broadcast methods also and do not just broadcast manually. This will ensure that no older code reference is used which points to rosetta submit step.

modules/sdk-coin-icp/src/lib/iface.ts
```diff
@@ -151,17 +144,8 @@ export interface UpdateEnvelope {
...
-export interface ReadStateEnvelope {
-  content: ReadState & {
-    request_type: RequestType;
-  };
-  sender_pubkey: Uint8Array;
-  sender_sig: Uint8Array;
-}
-
 export interface RequestEnvelope {
-  update: UpdateEnvelope;
```
DinshawKothari, line 163: In UpdateEnvelope, I see you have same details as HttpCanisterUpdate and request_type in the content. Please make use of HttpCanisterUpdate instead of duplicating it again. This will also help you at a below code where you casting signedTransaction as UpdateEnvelope and then as HttpCanisterUpdate in parseSignedTransaction()

modules/sdk-coin-icp/src/lib/transaction.ts
```diff
@@ -288,10 +289,7 @@ export class Transaction extends BaseTransaction {
 
   async parseSignedTransaction(rawTransaction: string): Promise<ParsedTransaction> {
     const signedTransaction = this._utils.cborDecode(this._utils.blobFromHex(rawTransaction));
-    const signedTransactionTyped = signedTransaction as { requests: any[] };
-    const envelopes = signedTransactionTyped.requests[0][1];
-    const updates = envelopes.map((envelope) => envelope.update);
-    const httpCanisterUpdate = updates[0].content as HttpCanisterUpdate;
+    const httpCanisterUpdate = (signedTransaction as Envelope).content as HttpCanisterUpdate;
```
DinshawKothari, line 292: Given a suggestion in above comment for type casting changes here.

modules/sdk-coin-icp/src/lib/signedTransactionBuilder.ts
```diff
@@ -30,11 +23,12 @@ export class SignedTransactionBuilder {
...
     ) as CborUnsignedTransaction;
-    assert(combineRequest.signatures.length === unsignedTransaction.ingress_expiries.length * 2);
+    // at present we expect only one request type and one signature
+    assert(combineRequest.signatures.length === 1);
     assert(unsignedTransaction.updates.length === 1);
     const envelopes = this.getEnvelopes(unsignedTransaction, signatureMap);
-    const envelopRequests = { requests: envelopes };
-    const signedTransaction = utils.cborEncode(envelopRequests);
+    assert(envelopes.length === 1);
+    const signedTransaction = utils.cborEncode(envelopes[0][1][0].update);
```
DinshawKothari, line 31: We can use `as UpdateEnvelope` to ensure what are we cbor encoding here.

DinshawKothari: I would suggest to break this `envelopes[0][1][0].update` in different constants to explain better that we aim to UpdateEnvelope from envelopes, rather than directly using it. Also we can add types at whatever steps necessary here.

modules/sdk-coin-icp/src/lib/iface.ts
adarsh312: this has been discussed offline

modules/sdk-coin-icp/src/lib/signedTransactionBuilder.ts
DinshawKothari: @mohd-kashif Yes, its a typo for my comment.
my bad, I just edited my comment and added a "not" there.
Thanks for pointing out.

DinshawKothari (approved):

PR #5883, opened 2025-04-04
fix(ofc): correct ofc decimal places for stx tokens

at31416:
Correcting the decimal places for stx tokens as per this doc: https://docs.google.com/document/d/1pWXoEBJP04q9KhZhZWyvR0fo42dJpeoouvb_gBSVPVU/edit?tab=t.0

decimal places for onchain stx tokens for this batch is already correct

Ticket: COIN-3601


MohammedRyaan786 (approved):
LGTM

PR #5884, opened 2025-04-04
feat(sdk-ams): add sdk support for ams

veetragjain:
This PR includes - 

- New coin factory and global coin factory that are initialized using a custom coin map(which defaults to statics coin map).
- New API class that extends the current BitGo sdk api, with methods that converts the ams response into a coinmap(Will also have ams endpoints exposed in the sdk once they are public and have methods to push a single coin or token into the coin map and coin factory in a follow up PR). 

Ticket: WIN-4974


modules/sdk-ams/src/bitgoV2.ts
```diff
@@ -0,0 +1,38 @@
+import { BitGo } from 'bitgo';
+import { BaseCoin } from '@bitgo/sdk-core';
+import { BitGoAPIOptions } from '@bitgo/sdk-api';
+import { createTokenMapUsingConfigDetails, AmsTokenConfig } from '@bitgo/statics';
+import { GlobalCoinFactoryV2 } from './globalCoinFactoryV2';
+import { CoinFactoryV2 } from './coinFactoryV2';
+
+export class BitGoV2 extends BitGo {
```
mukeshsp, line 8: Can you add uses of this class ?

mukeshsp (approved):

OttoAllmendinger (changes requested):
please respond to my TDD comments first

sijuu (approved):

zahin-mohammad (changes requested):
stale pr

PR #5916, opened 2025-04-09
feat(sdk-api): add decryptKeys method for bulk wallet key decryption

ayush-9:
TICKET: CSI-384


aarvay (commented):
I think this would be better implemented inside sdk-core/wallets

ayush-9:
> I think this would be better implemented inside sdk-core/wallets

UI uses BitGoAPI, where BitGoAPI.wallets() has been deprecated

aarvay (approved):

PR #5928, opened 2025-04-10
chore(sdk-coin-tao): skip tao recover unit test flow

nvrakesh06:
TICKET: WIN-4315

Action Item: https://bitgoinc.atlassian.net/browse/WIN-5173

In the recover test flow, there's a call being made to Tao's WebSocket endpoint, but the connection is failing with the error:
 API-WS: disconnected from wss://test.finney.opentensor.ai/: 1006:: Abnormal Closure
 
 Skipping the recover test flow for now.


hitansh-madan (approved):
Yeah, we need to add mocks for all these network calls

PR #5938, opened 2025-04-10
feat(express): add support to change keychain password

joeybaibg:
- adds a new route to change keychain passphrase
- method: POST
- path: '/api/v2/:coin/keychain/:id/changepassword'

TICKET: CAAS-41

API Docs: https://github.com/BitGo/bitgo-microservices/pull/47405

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -2,33 +2,33 @@
  * @prettier
  */
 import {
-  EddsaUtils,
-  EcdsaUtils,
-  EcdsaMPCv2Utils,
-  CustomGShareGeneratingFunction,
-  CustomRShareGeneratingFunction,
-  UnsupportedCoinError,
-  GShare,
-  SignShare,
-  CustomCommitmentGeneratingFunction,
   CommitmentShareRecord,
-  EncryptedSignerShareRecord,
-  CustomPaillierModulusGetterFunction,
-  CustomKShareGeneratingFunction,
-  CustomMuDeltaShareGeneratingFunction,
-  CustomSShareGeneratingFunction,
-  TssEcdsaStep1ReturnMessage,
-  TssEcdsaStep2ReturnMessage,
-  SShare,
-  ShareType,
-  MPCType,
   CreateNetworkConnectionParams,
-  GetNetworkPartnersResponse,
-  encryptRsaWithAesGcm,
-  Wallet,
+  CustomCommitmentGeneratingFunction,
+  CustomGShareGeneratingFunction,
+  CustomKShareGeneratingFunction,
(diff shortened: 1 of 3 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -1025,6 +1025,38 @@ async function handleWalletUpdate(req: express.Request): Promise<unknown> {
...
+  }
+
+  const wallet = await coin.wallets().get({ id: req.params.id, reqId });
+  const keychain = await wallet.getEncryptedUserKeychain();
+
+  const updatedKeychain = await wallet.baseCoin.keychains().updateSingleKeychainPassword({
+    keychain,
+    oldPassword,
+    newPassword,
+  });
```
zahin-mohammad, line 1053: The user might store their backup key with BitGo as well. You should change that password as well.

joeybaibg: @zahin-mohammad this is good to know. 

I modeled this after the implementation in the UI, which is only changing the User key pw: https://github.com/BitGo/bitgo-ui/blob/develop/apps/app/src/containers/WalletSettings/WalletPassword/WalletPasswordDialog/WalletPasswordDialog.tsx#L59-L70

Since this is to unblock Fold, are you okay with limiting this to the User key and explaining the details in documentation?

modules/express/encryptedPrivKeys.json
```diff
@@ -1,3 +1 @@
-{
-  "61f039aad587c2000745c687373e0fa9": "{\"iv\":\"Ca36HxntBXcUttPW2MX0UA==\",\"v\":1,\"iter\":10000,\"ks\":256,\"ts\":64,\"mode\":\"ccm\",\"adata\":\"\",\"cipher\":\"aes\",\"salt\":\"HE1f6uG1A5Y=\",\"ct\":\"UaoGmQBoRXj9cVBonH4rqyiMHHUbeJ6eiUkS7ne7yHWLKUQmdOfor0Ob/ut9fYykhJpYmbOy2YxPJuU64Rg0DWucyOgedf9EGHct/QqLf+IikwPRBu3invi218nyDCltEZF2AbDanpb0u2y52mMTETKg3CbcqbE=\"}"
-}
+{}
```
zahin-mohammad, line 1: Was this change intentional?

zahin-mohammad (dismissed):
lgtm. Please create a follow up ticket to add the api-docs for this if you didn't already (I believe they are currently still part of the platform api spec)

modules/express/encryptedPrivKeys.json
joeybaibg: @zahin-mohammad not intentional. i reverted

zahin-mohammad (approved):

PR #5951, opened 2025-04-14
chore: add BULK_TRANSACTION to topeth and hteth tokens

kamleshmugdiya:
TICKET: COIN-3747

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -1023,7 +1023,7 @@ export const coins = CoinMap.fromCoins([
     18,
     UnderlyingAsset.OPETH,
     BaseUnit.ETH,
-    OPETH_FEATURES
+    [...OPETH_FEATURES, CoinFeature.BULK_TRANSACTION]
   ),
   account(
     '53b1e350-f907-45ec-abf7-11d132547055',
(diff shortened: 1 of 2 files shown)


shivam-bitgo (dismissed):

modules/statics/src/coins.ts
```diff
@@ -1023,7 +1023,7 @@ export const coins = CoinMap.fromCoins([
     18,
     UnderlyingAsset.OPETH,
     BaseUnit.ETH,
-    OPETH_FEATURES
+    [...OPETH_FEATURES, CoinFeature.BULK_TRANSACTION]
```
gianchandania, line 1026: The ticket is for tokens sendMany, why are we adding Bulk transaction to opeth native coin?

kamleshmugdiya: right now, topeth doesn't support native batching.
UI uses this coin feature to show the csv upload section and token approval flow.
as we have deployed [batcher contract](https://optimistic.etherscan.io/address/0x3E1e5d78e44f15593B3B61ED278f12C27F0fF33e#code) on opeth, we can add support for both native and token batching.

forgot to add this in the ticket, will update it.

gianchandania (approved):

PR #5983, opened 2025-04-17
feat(statics): add coin feature for erc20 batching

kamleshmugdiya:
TICKET: COIN-3843

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -289,6 +289,11 @@ export enum CoinFeature {
   BULK_TRANSACTION = 'bulk-transaction',
 
   /**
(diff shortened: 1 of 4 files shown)


sachushaji (approved):
Are we enabling to prod also now or just test

kamleshmugdiya:
> Are we enabling to prod also now or just test

we have already enabled erc20 sendMany for eth and opt on prod.
we used BULK_TRANSACATION feature in the UI to show csv upload and token approval flow for now.
but it will have issue, atleast for now, where coin only native batch, eg. polygon.
we are showing token batching in UI. so we are introducing new coin feature.

PR #5997, opened 2025-04-21
feat(sdk-coin-icp): replace protoDefinition with staticProtoDefinition for protobuf parsing

mohd-kashif:
TICKET: WIN-5259


modules/sdk-coin-icp/src/lib/utils.ts
```diff
@@ -604,7 +604,7 @@ export class Utils implements BaseUtils {
   }
 
   async fromArgs(arg: Uint8Array): Promise<SendArgs> {
-    const root = protobuf.parse(protoDefinition).root;
+    const root = protobuf.Root.fromJSON(require(path.resolve(__dirname, './staticProtoDefinition.json')));
```
DinshawKothari, line 607: Have you explored how other coins are using protobuf ?
I see for cosmos, tron they have used it but in a different way.

mohd-kashif: I dont see these coins having the same use case as ICP

DinshawKothari: In our case, we are creating a compiled json file for protobuf. Similarly I see protobuf files in resources in abstract-cosmos. Are we doing something completely different or on similar lines ?

DinshawKothari: We can take it up later as optimisation.

DinshawKothari (approved):

PR #6006, opened 2025-04-21
feat: added withdraw route for lightning on chain withdrawal

parvbitgo067:
Withdraw route calls build and send transaction in wallet platform. Test cases added.

### DIFF
--- a/modules/abstract-lightning/src/wallet/lightning.ts
+++ b/modules/abstract-lightning/src/wallet/lightning.ts
@@ -24,6 +24,8 @@ import {
   TransactionQuery,
   PaymentInfo,
   PaymentQuery,
+  LightningOnchainWithdrawParams,
+  LightningOnchainWithdrawResponse,
 } from '../codecs';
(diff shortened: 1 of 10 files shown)


modules/abstract-lightning/src/wallet/lightning.ts
```diff
@@ -301,6 +312,49 @@ export class LightningWallet implements ILightningWallet {
...
+
+    const transactionRequestSend = await commonTssMethods.sendTxRequest(
+      this.wallet.bitgo,
+      this.wallet.id(),
+      transactionRequestCreate.txRequestId,
+      RequestType.tx,
+      reqId
+    );
+
+    return {
```
saravanan7mani, line 352: We have to return `transferId`, we can do it in follow up PR.

saravanan7mani (approved):

lokesh-bitgo (approved):
Approved based on CODEOWNERS

PR #6009, opened 2025-04-22
fix: fix the currency code for the xrp tokens if more than 3 characters

prithvishet2503:
Ticket: WIN-5270


MohammedRyaan786 (approved):
LGTM

PR #6011, opened 2025-04-22
WIN-5232 | implementation of default memo as 0 for ICP

mohd-kashif:
TICKET: WIN-5232


modules/sdk-coin-icp/src/lib/transaction.ts
```diff
@@ -109,10 +110,8 @@ export class Transaction extends BaseTransaction {
             senderPublicKeyHex: senderPublicKeyHex,
             transactionType: transactionType,
             expiryTime: Number(parsedTx.metadata.created_at_time + (MAX_INGRESS_TTL - PERMITTED_DRIFT)),
+            memo: parsedTx.metadata.memo || DEFAULT_MEMO,
```
DinshawKothari, line 113: There should not be a case where `parsedTx.metadata.memo` does not exist. Also memo is no more optional in Metadata interface. So we do not need to add  DEFAULT_MEMO condition here.

modules/sdk-coin-icp/src/lib/transaction.ts
```diff
@@ -149,7 +148,7 @@ export class Transaction extends BaseTransaction {
           sender: this._icpTransactionData.senderAddress,
           senderPublicKey: this._icpTransactionData.senderPublicKeyHex,
           recipient: this._icpTransactionData.receiverAddress,
-          memo: this._icpTransactionData.memo,
+          memo: DEFAULT_MEMO,
```
DinshawKothari, line 151: why are we using DEFAULT_MEMO here ?

modules/sdk-coin-icp/src/lib/transaction.ts
```diff
@@ -284,12 +283,10 @@ export class Transaction extends BaseTransaction {
       operations: [senderOperation, receiverOperation, feeOperation],
       metadata: {
         created_at_time: args.createdAtTime.timestampNanos,
+        memo: Number(args.memo.memo) || DEFAULT_MEMO,
```
DinshawKothari, line 286: Same here. We need not default everywhere. If we handle it correctly in txnBuilder creation, it will be automatically propagated to the rest of the code.

modules/sdk-coin-icp/src/lib/transactionBuilder.ts
```diff
@@ -137,6 +137,7 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
     this._receiverId = icpTransactionData.receiverAddress;
     this._publicKey = icpTransactionData.senderPublicKeyHex;
     this._amount = icpTransactionData.amount;
+    this._memo = DEFAULT_MEMO;
```
DinshawKothari, line 140: Same here.
```suggestion
    this._memo = icpTransactionData.memo ?? DEFAULT_MEMO;
```

PR #6012, opened 2025-04-22
feat(statics): added somnia config

yogeshwar-bitgo:
TICKET: WIN-5025


mullapudipruthvik (approved):

modules/statics/src/networks.ts
```diff
@@ -1257,6 +1257,23 @@ class WorldTestnet extends Testnet implements EthereumNetwork {
   nativeCoinOperationHashPrefix = '4801';
 }
 
+class Somnia extends Mainnet implements EthereumNetwork {
+  name = 'Somnia';
+  family = CoinFamily.STT;
+  explorerUrl = 'https://shannon-explorer.somnia.network/tx/'; //TODO: WIN-5278: change it with mainnet explorer, when its available
```
adarsh312, line 1263: if we don't have mainnet explorer how are these configs added ?

PR #6022, opened 2025-04-23
feat: handle rejected promises in bulk wallet share updates

ayush-9:
Ticket: CSI-419

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallets.ts
@@ -1030,14 +1030,24 @@ export class Wallets implements IWallets {
     );
 
     // Extract successful updates
-    const successfulUpdates = settledUpdates
-      .filter(
-        (result): result is PromiseFulfilledResult<BulkUpdateWalletShareOptionsRequest[]> =>
-          result.status === 'fulfilled'
-      )
-      .flatMap((result) => result.value);
+    const successfulUpdates = settledUpdates.flatMap((result) => (result.status === 'fulfilled' ? result.value : []));
 
-    // Send updates to the server
+    // Extract failed updates - only from rejected promises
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -1037,7 +1037,20 @@ export class Wallets implements IWallets {
...
+      .map(({ result, index }) => {
+        // Since we've filtered for rejected promises, we can safely cast to PromiseRejectedResult
+        const rejectedResult = result as PromiseRejectedResult;
+        return {
+          walletShareId: resolvedShares[index].walletShareId,
+          reason: rejectedResult.reason?.message || String(rejectedResult.reason),
+        };
+      });
+
+    // Send successful updates to the server
```
sdoshibitgo, line 1053:
```suggestion
    // Extract failed updates - only from rejected promises
    const successfulUpdates = settledUpdates
  .flatMap((result) => (result.status === 'fulfilled' ? (result as PromiseFulfilledResult<BulkUpdateWalletShareOptionsRequest[]>).value : []));

const failedUpdates = settledUpdates
  .reduce((acc, result, index) => {
    if (result.status === 'rejected') {
      const rejectedResult = result as PromiseRejectedResult;
      acc.push({
        walletShareId: resolvedShares[index].walletShareId,
        reason: rejectedResult.reason?.message || String(rejectedResult.reason),
      });
    }
    return acc;
  }, []);

    // Send successful updates to the server
```

ayush-9: i feel no need to assert TS will figure it out, but yes let me use reduce

hrishikeshjain (approved):

sdoshibitgo (approved):

PR #6037, opened 2025-04-25
chore: ignore GHSA-cpj6-fhp6-mr6j

bitgoAaron:
- this is only affecting react-router-dom, only used by web-demo, it can safely be ignored since we cannot upgrade while still supporting Node 18

Ticket: CE-00000

### DIFF
--- a/.iyarc
+++ b/.iyarc
@@ -2,3 +2,5 @@ GHSA-3h5v-q93c-6h6q
 GHSA-8hc4-vh64-cxmj
 GHSA-9wv6-86v2-598j
 GHSA-3gc7-fjrx-p6mg
+GHSA-cpj6-fhp6-mr6j
+GHSA-f46r-rw29-r322


at31416 (approved):

mohammadalfaiyazbitgo (approved):

andrew-scott-fischer:
@bitgoAaron Should we stop supporting node18? What do we need to do to tell clients we aren't supporting node18 any longer?

bitgoAaron:
> @bitgoAaron Should we stop supporting node18? What do we need to do to tell clients we aren't supporting node18 any longer?

I can't imagine why we would continue to support it, given it's EOL next week. Let's kill it

andrew-scott-fischer:
> > @bitgoAaron Should we stop supporting node18? What do we need to do to tell clients we aren't supporting node18 any longer?
> 
> I can't imagine why we would continue to support it, given it's EOL next week. Let's kill it

Is there anything we need to communicate to customers? By what strategy should we kill it? should we just bump the `engines` property to exclude node18? I'm not sure if there's something we contractually need to do to tell customers of this, or just make a note of it in the BitGoJS release notes?

PR #6041, opened 2025-04-25
feat(statics): update ICP explorer URLs for mainnet and testnet

mohd-kashif:
TICKET: WIN-4242

### DIFF
--- a/modules/statics/src/networks.ts
+++ b/modules/statics/src/networks.ts
@@ -172,18 +172,17 @@ class AptTestnet extends Testnet implements AccountNetwork {
 class Icp extends Mainnet implements AccountNetwork {
   name = 'Internet Computer';
   family = CoinFamily.ICP;
-  explorerUrl = 'https://www.icpexplorer.org/#/tx/'; //TODO(WIN-4242): update to the test rosetta explorer url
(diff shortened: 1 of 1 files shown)


modules/statics/src/networks.ts
```diff
@@ -172,18 +172,17 @@ class AptTestnet extends Testnet implements AccountNetwork {
 class Icp extends Mainnet implements AccountNetwork {
   name = 'Internet Computer';
   family = CoinFamily.ICP;
-  explorerUrl = 'https://www.icpexplorer.org/#/tx/'; //TODO(WIN-4242): update to the test rosetta explorer url
-  accountExplorerUrl = 'https://www.icpexplorer.org/#/account/?network=mainnet';
-  blockExplorerUrl = 'https://www.icpexplorer.org/#/block/?network=mainnet';
```
mohd-kashif, line 177: there is not block explorer for ICP as each block contains only one transaction

abhishekagrawal080 (approved):

DinshawKothari (approved):
lgtm

modules/statics/src/networks.ts
DinshawKothari: May be just check if we use 'blockExplorerUrl' in UI or sdk somewhere.

zahin-mohammad (approved):
lgtm

PR #6043, opened 2025-04-25
feat: list wallet addresses sorted by balance

bhavidhingra:
TICKET: COIN-3909

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -37,6 +37,7 @@ import {
 } from '../utils';
 import {
   AccelerateTransactionOptions,
+  AddressesByBalanceOptions,
   AddressesOptions,
   BuildConsolidationTransactionOptions,
   BuildTokenEnablementOptions,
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1132,6 +1133,38 @@ export class Wallet implements IWallet {
...
+    }
+    if (!_.isNumber(query.page)) {
+      throw new Error('invalid page argument, expecting number');
+    }
+    if (!_.isNumber(params.limit)) {
+      throw new Error('invalid limit argument, expecting number');
+    }
+    if (params.limit < 1 || params.limit > 500) {
+      throw new Error('limit argument must be between 1 and 500');
+    }
```
hrishikeshjain, line 1160: should we be checking if the basecoin supports list addresses by sorted balances or does the api takes care of it?

at31416: API would throw the `unsupported coin` error from IMS. I guess that should be fine? 
ref: https://github.com/BitGo/indexerdb-microservice/blob/master/src/app/controllers/balances.js#L300

hrishikeshjain (approved):

at31416 (approved):

margueriteblair (approved):

PR #6060, opened 2025-04-30
feat(statics): add method to insert single token in coin map

veetragjain:
Ticket: WIN-5359

### DIFF
--- a/modules/statics/src/map.ts
+++ b/modules/statics/src/map.ts
@@ -18,33 +18,37 @@ export class CoinMap {
   }
 
   static fromCoins(coins: Readonly<BaseCoin>[]): CoinMap {
-    return coins.reduce((coinMap, coin) => {
-      if (coinMap.has(coin.name)) {
-        throw new DuplicateCoinDefinitionError(coin.name);
-      }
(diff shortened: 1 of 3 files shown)


modules/statics/src/map.ts
```diff
@@ -18,33 +18,37 @@ export class CoinMap {
...
+    if (this._map.has(coin.name)) {
+      throw new DuplicateCoinDefinitionError(coin.name);
+    }
+    if (this._coinByIds.has(coin.id)) {
+      throw new DuplicateCoinIdDefinitionError(coin.id);
+    }
+    const alias = coin.alias;
+    if (alias && this._coinByAliases.has(alias)) {
+      throw new DuplicateCoinDefinitionError(alias);
+    }
```
ajays97, line 38: Since the behaviour doesn't change if we don't throw these errors, can we get rid of them?

ajays97 (approved):

modules/statics/src/map.ts
veetragjain: We need the errors so that we don't insert tokens with duplicate fields. Each token is supposed to have a distinct name, id and alias.

PR #6088, opened 2025-05-06
refactor(sdk-core): making supportsMultiSig as coin feature

lokesh-bitgo:
Ticket: WP-4028

Refactored [this PR](https://github.com/BitGo/BitGoJS/pull/5695) by making supportsMultiSig() as coin feature.

### DIFF
--- a/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
+++ b/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
@@ -2,11 +2,11 @@
  * @prettier
  */
 import * as crypto from 'crypto';
+import { Hash } from 'crypto';
+import * as utxolib from '@bitgo/utxo-lib';
 import { bip32 } from '@bitgo/utxo-lib';
 import { BigNumber } from 'bignumber.js';
-
-import * as utxolib from '@bitgo/utxo-lib';
-import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import { BaseCoin as StaticsBaseCoin, CoinFeature } from '@bitgo/statics';
 
 import { InitiateRecoveryOptions } from '../recovery';
 import { signMessage } from '../bip32util';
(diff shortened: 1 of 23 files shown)


OttoAllmendinger (dismissed):

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
```diff
@@ -185,14 +185,6 @@ export abstract class BaseCoin implements IBaseCoin {
     return false;
   }
 
-  /**
-   * Flag indicating if this coin supports MultiSig wallets.
-   * @return {boolean} True if MultiSig wallets can be created for this coin
-   */
-  supportsMultisig(): boolean {
-    return false;
-  }
```
zahin-mohammad, line 194: To avoid making a breaking change can you keep this function and just mark at as deprecated?

By default it can return the value from statics.

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
```diff
@@ -493,7 +493,6 @@ export interface IBaseCoin {
   allowsAccountConsolidations(): boolean;
   getTokenEnablementConfig(): TokenEnablementConfig;
   supportsTss(): boolean;
-  supportsMultisig(): boolean;
```
zahin-mohammad, line 496: Please keep the method for backwards compatibility.

zahin-mohammad (changes requested):

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
lokesh-bitgo: Hi Zahin,

This function is currently unused. Do we still need to keep it and mark it as deprecated?

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
lokesh-bitgo: Same, this function is currently not used by anyone.

zahin-mohammad (changes requested):
Discussed on slack. Still need to keep the method but mark as deprecated.

modules/statics/src/base.ts
```diff
@@ -363,6 +363,11 @@ export enum CoinFeature {
    * This coins is an EVM compatible coin and should use common EVM functionality
    */
   SHARED_EVM_SIGNING = 'shared-evm-signing',
+
+  /**
+   * This coin supports multisig wallets
```
abhishekagrawal080, line 368: Please change This coin supports multisig wallets this to This coin supports multisig

zahin-mohammad (approved):
lgtm

OttoAllmendinger (approved):

dpkjnr (approved):

PR #6089, opened 2025-05-06
fix: cosmos batch rpc to rest urls and dorado testnet for fetch

baltiyal:
TICKET: COIN-4012, COIN-4013


MohammedRyaan786 (approved):
lgtm

PR #6108, opened 2025-05-09
fix(sdk-coin-trx): handling string type for scanning factor

abhishekagrawal080:
TICKET: WIN-5440

### DIFF
--- a/modules/sdk-coin-trx/src/trx.ts
+++ b/modules/sdk-coin-trx/src/trx.ts
@@ -586,10 +586,15 @@ export class Trx extends BaseCoin {
     } else if (!isInteger(startIdx) || startIdx < 0) {
       throw new Error('Invalid starting index to scan for addresses');
     }
+
     let numIteration = params.scan;
     if (isUndefined(numIteration)) {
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-trx/src/trx.ts
```diff
@@ -586,10 +586,15 @@ export class Trx extends BaseCoin {
     } else if (!isInteger(startIdx) || startIdx < 0) {
       throw new Error('Invalid starting index to scan for addresses');
     }
+
     let numIteration = params.scan;
     if (isUndefined(numIteration)) {
       numIteration = 20;
-    } else if (!isInteger(numIteration) || numIteration <= 0) {
+    } else if (typeof numIteration === 'string') {
```
nvrakesh06, line 593: You can also add a `!isInteger(numIteration)` check for non-numeric strings.
Otherwise, parseInt() won’t throw an error—it will silently return NaN, which can break the flow later.

nvrakesh06 (approved):
lgtm

shashankms288 (approved):

PR #6118, opened 2025-05-13
feat(sdk-coin-icp): integrate ic0 library for account balance retrieval

mohd-kashif:
TIcket: WIN-5479


modules/sdk-coin-icp/src/icp.ts
```diff
@@ -268,8 +234,7 @@ export class Icp extends BaseCoin {
       const decodedResponse = utils.cborDecode(response.data) as PublicNodeSubmitResponse;
 
       if (decodedResponse.status === 'replied') {
-        const txnId = this.extractTransactionId(decodedResponse);
-        return { txId: txnId };
+        return {};
```
DinshawKothari, line 237: why returning empty object here? We should know what recovery txn hash we generated.

modules/sdk-coin-icp/src/icp.ts
```diff
@@ -286,37 +251,60 @@ export class Icp extends BaseCoin {
...
+  protected async getBalanceFromPrincipal(principalId: string): Promise<string> {
     try {
-      const payload = {
-        account_identifier: {
-          address: address,
-        },
+      const agent = this.createAgent();
+      const ic = replica(agent, { local: true });
+
+      const ledger = ic(Principal.fromUint8Array(LEDGER_CANISTER_ID).toText());
```
DinshawKothari, line 280: I think these steps will be common for any rpc call that we directly make to public fullnode for this canister. For eg v3/submit api and balance fetch api. If yes, then please make a generic function and pass params.

modules/sdk-coin-icp/src/icp.ts
```diff
@@ -430,10 +418,15 @@ export class Icp extends BaseCoin {
     txBuilder.transaction.addSignature(signatures);
     txBuilder.combine();
     const broadcastableTxn = txBuilder.transaction.toBroadcastFormat();
-    const result = await this.broadcastTransaction({ serializedSignedTransaction: broadcastableTxn });
-    if (!result.txId) {
-      throw new Error('Transaction failed to broadcast');
-    }
-    return result.txId;
+    await this.broadcastTransaction({ serializedSignedTransaction: broadcastableTxn });
```
DinshawKothari, line 421: If broadcasting fails, you are still returning recoveredTransaction. Please do a try catch.

modules/sdk-coin-icp/src/lib/iface.ts
```diff
@@ -185,6 +185,14 @@ export interface RecoveryOptions {
   memo?: number | BigInt;
 }
 
+export interface RecoveryTransaction {
```
DinshawKothari, line 188: Is this a standard that other coins also follow for recovery ? As far as I remember they only return txId.

modules/sdk-coin-icp/src/icp.ts
mohd-kashif: `broadcastTransaction()` returns `BaseBroadcastTransactionResult` which have the `txId` but we dont have a way to get the `txId` from the response and we cant change the signature of `broadcastTransaction()` as its a inherited function
so I have returned a `{}` response and set the `txId` in the `recover()` as we have the builder object in the `recover()`

modules/sdk-coin-icp/src/icp.ts
mohd-kashif: v3/submit and balance fetch has different process to make the request therefore they have different methods 
maybe other methods like feeEstimates or derive might have same process as the balance fetch

modules/sdk-coin-icp/src/icp.ts
mohd-kashif: added a try catch

modules/sdk-coin-icp/src/lib/iface.ts
mohd-kashif: i checked XRP, TRX and TON coins, they are having similar return fields

modules/sdk-coin-icp/src/icp.ts
DinshawKothari: cool, we need to just handle it properly from the caller then.

DinshawKothari: nit: add a comment to explain/why status == 'replied' is treated as a success response.

modules/sdk-coin-icp/src/icp.ts
DinshawKothari: as discussed over call, please add the todo ticket to move the ic-agent calls to a separate file as a follow up.

modules/sdk-coin-icp/src/lib/iface.ts
DinshawKothari: I think the need here is only for id and tx. As discussed offline, pls make the necessary change.

mohd-kashif: there is only id and tx

modules/sdk-coin-icp/src/icp.ts
mohd-kashif: added

modules/sdk-coin-icp/src/icp.ts
mohd-kashif: added

DinshawKothari (approved):

PR #6201, opened 2025-05-28
fix(sdk-coin-sol): fix ata creation for sol 2022 token

abhishekagrawal080:
TICKET: WIN-5670

### DIFF
--- a/modules/sdk-coin-sol/src/lib/utils.ts
+++ b/modules/sdk-coin-sol/src/lib/utils.ts
@@ -12,6 +12,7 @@ import {
   decodeCloseAccountInstruction,
   getAssociatedTokenAddress,
   TOKEN_PROGRAM_ID,
+  TOKEN_2022_PROGRAM_ID,
 } from '@solana/spl-token';
 import {
   Keypair,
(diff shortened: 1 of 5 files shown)


Ranjna-G (dismissed):
LGTM

shashankms288 (dismissed):
LGTM

modules/statics/src/account.ts
```diff
@@ -1610,6 +1613,7 @@ export function solToken(
   contractAddress: string,
   asset: UnderlyingAsset,
   features: CoinFeature[] = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.REQUIRES_RESERVE],
+  programId = '',
```
veetragjain, line 1616: Why not default it to spl token program id, instead of having an empty string.

abhishekagrawal080: Done

modules/statics/src/base.ts
```diff
@@ -2703,6 +2703,11 @@ export enum KeyCurve {
   Ed25519 = 'ed25519',
 }
 
+export enum ProgramID {
+  TokenProgramId = 'TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA',
+  Token2022ProgramId = 'TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb',
+}
+
```
veetragjain, line 2710: Nit: This can reside in account.ts file, base.ts have enums and types that are required by baseCoin class. Can you please also add a description for this enum.

modules/sdk-coin-sol/src/lib/utils.ts
```diff
@@ -475,6 +476,29 @@ export function getSolTokenFromAddress(tokenAddress: string, network: BaseNetwor
...
+  const tokens = coins.filter((coin) => {
+    if (coin instanceof SolCoin) {
+      return coin.tokenAddress.toLowerCase() === tokenAddress.toLowerCase();
+    }
+    return false;
+  });
+  const tokensArray = tokens.map((token) => token);
+  if (tokensArray.length >= 1) {
+    // there should never be two tokens with the same contract address, so we assert that here
+    assert(tokensArray.length === 1);
```
veetragjain, line 496: Please verify if we have any testnet and mainnet token with same contract address. This check will fail in that case.

veetragjain (dismissed):

modules/sdk-coin-sol/src/lib/utils.ts
abhishekagrawal080: I have checked there is no token with same address in testnet and mainnet

veetragjain (approved):

PR #6206, opened 2025-05-28
feat(statics): onboard new batch of tokens

ravibitgo:
Ticket: [WIN-5669](https://bitgoinc.atlassian.net/browse/WIN-5669)

[WIN-5669]: https://bitgoinc.atlassian.net/browse/WIN-5669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/statics/src/gatekeep.ts
+++ b/modules/statics/src/gatekeep.ts
@@ -1824,4 +1824,10 @@ export const gatekeep = [
   'ofcsol:rock',
   'sol:dood',
   'ofcsol:dood',
+  'xrp:aau',
+  'ofcxrp:aau',
+  'hbar:bonzo',
+  'ofchbar:bonzo',
+  'eth:ctrl',
+  'ofceth:ctrl',
 ];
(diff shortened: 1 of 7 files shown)


pranavjain97 (commented):
Can you please update the codeowners of these files? This is not for WP to review and would prefer coins to TAL.

modules/statics/src/gatekeep.ts
```diff
@@ -1824,4 +1824,10 @@ export const gatekeep = [
   'ofcsol:rock',
   'sol:dood',
   'ofcsol:dood',
+  'xrp:aau',
+  'ofcxrp:aau',
+  'hbar:bonzo',
+  'ofchbar:bonzo',
+  'eth:ctrl',
+  'ofceth:ctrl',
```
DinshawKothari, line 1832: How about 'sol:chillguy', we don't need to gatekeep this ?

ravibitgo: As per the [ticket](https://bitgoinc.atlassian.net/browse/WIN-5669), that token should be ungated.

at31416 (approved):

modules/statics/src/gatekeep.ts
mukeshsp: please update gatekeep in ams too, will dm you

mukeshsp (approved):

PR #6246, opened 2025-06-05
feat(statics): adding EVM_COIN coin feature

parasgarg-bitgo:
Ticket: WIN-4729

Adding a new coin feature EVM_COIN to simplify check to verify if a coin is ETH-Like

### DIFF
--- a/modules/statics/src/coinFeatures.ts
+++ b/modules/statics/src/coinFeatures.ts
@@ -25,6 +25,7 @@ export const ETH_FEATURES = [
   CoinFeature.SUPPORTS_TOKENS,
   CoinFeature.ENTERPRISE_PAYS_FEES,
   CoinFeature.WALLET_CONNECT_DEFI,
+  CoinFeature.EVM_COIN,
 ];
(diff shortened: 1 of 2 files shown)


modules/statics/src/coinFeatures.ts
```diff
@@ -31,6 +31,7 @@ export const ETH_FEATURES_WITH_STAKING = [...ETH_FEATURES, CoinFeature.STAKING];
 export const ETH_FEATURES_WITH_STAKING_AND_MMI = [...ETH_FEATURES_WITH_STAKING, CoinFeature.METAMASK_INSTITUTIONAL];
 export const ETC_FEATURES = [
   ...ETH_FEATURES,
+  CoinFeature.EVM_COIN,
```
gianchandania: In this file, you have not added this feature for celo and bera. Also why not add it to ETH_FEATURES. It is getting used in all the evm coins.

parasgarg-bitgo: updated

gianchandania (approved):

PR #6256, opened 2025-06-06
feat(bitgo): add method to register single token from ams in coin factory

veetragjain:
There are methods implemented to initiate coin factory using list of token metadata. This PR adds method to register a single token into coin factory.

Ticket: WIN-5057

### DIFF
--- a/modules/bitgo/src/bitgo.ts
+++ b/modules/bitgo/src/bitgo.ts
@@ -8,8 +8,13 @@ import * as _ from 'lodash';
 
 import { BaseCoin, CoinFactory, common } from '@bitgo/sdk-core';
 import { BitGoAPI, BitGoAPIOptions } from '@bitgo/sdk-api';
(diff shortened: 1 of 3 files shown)


modules/bitgo/src/bitgo.ts
```diff
@@ -74,6 +83,27 @@ export class BitGo extends BitGoAPI {
     return GlobalCoinFactory.getInstance(this, coinName);
   }
 
+  /**
+   * Register a token in the coin factory
+   * @param tokenConfig - The token metadata from AMS
+   */
+  registerToken(tokenConfig: TrimmedAmsTokenConfig): void {
+    if (!this._useAms) {
```
ajays97, line 91: How do we set this to true?

veetragjain: Consumer sets this variable to true when initializing `bitgo` class.

sreerajs (approved):

pranavjain97 (approved):

abhishekagrawal080 (approved):

lokesh-bitgo (approved):

PR #6258, opened 2025-06-08
feat(sdk-coin-vet): add transaction builder logic for vechain

at31416:
This pull request introduces significant updates to the VeChain (`VET`) SDK module, including dependency additions, interface enhancements, and substantial refactoring of the transaction handling logic. These changes aim to improve the SDK's functionality, maintainability, and support for VeChain-specific features.

### Dependency Updates:
* Added `@vechain/sdk-core` and `@vechain/sdk-network` as new dependencies in `package.json` to support VeChain-specific transaction and network capabilities.

### Interface and Type Enhancements:
* Introduced `VetTransactionExplanation` interface in `iface.ts` to extend the base transaction explanation with VeChain-specific fields like `sender` and transaction `type`.
* Added `ExplainTransactionOptions` and `VetParseTransactionOptions` in `types.ts` to define options for transaction explanation and parsing.

### Refactoring of Transaction Handling:
* Reorganized transaction-related files into subdirectories (`transaction/transaction`, `transactionBuilder/transactionBuilder`, etc.) for better structure and modularity.
* Replaced the old `Transaction` and `TransactionBuilder` implementations with new classes that leverage `@vechain/sdk-core` for VeChain-specific functionality, including support for transaction clauses, sender signatures, and fee payer signatures. [[1]](diffhunk://#diff-7f14aff2d52d69d24082bf421d06406f3ba12b2fc6414a5d4c1f263bc5589599L1-L16) [[2]](diffhunk://#diff-e4d243d7c0082d6f2dfd009037c36220aac337a5323fb02e9c013fa5d3d0f5c0L1-L99) [[3]](diffhunk://#diff-4f951085c09181d139985c396627bdbe357e491fa849f37a1c94b3b90ddbb99eR1-R192)
* Added a `TransferBuilder` class to handle native VET transfers, ensuring validation of transaction clauses and recipient details. [[1]](diffhunk://#diff-34d35852cdf4902ebe499a7689247383ca40056698f6cbd8a6e386762e9018a4R1-R46) [[2]](diffhunk://#diff-d821ba7069b154ab201e31aa86c9e80d742175e23c6bfe69aec98ff198e25486L1-L13)

### TransactionBuilderFactory Enhancements:
* Updated `TransactionBuilderFactory` to support deserialization of signed transactions and initialization of builders with VeChain-specific logic.

### VeChain Coin Implementation (`Vet` class):
* Implemented `verifyTransaction`, `parseTransaction`, and `explainTransaction` methods to validate, parse, and explain transactions, including checks for recipient and output consistency. [[1]](diffhunk://#diff-5f82a96e30f8c83f2c585480d46fc9adcbfba9ae1a60855427c9fe09cff3a1e8L82-R121) [[2]](diffhunk://#diff-5f82a96e30f8c83f2c585480d46fc9adcbfba9ae1a60855427c9fe09cff3a1e8L94-R165)
* Added support for rebuilding transactions using the new transaction builder factory.

These changes collectively enhance the SDK's capabilities for working with VeChain transactions while improving code organization and maintainability.Ticket: COIN-4379


modules/sdk-coin-vet/src/lib/transactionBuilder/transactionBuilder.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  protected abstract get transactionType(): TransactionType;
+
+  getRebuiltTransaction(): Transaction {
+    return this._transaction;
+  }
```
ajays97: Will this be updated later on? This seems to be the same as `get transaction()` ? So, is this needed?

modules/sdk-coin-vet/src/lib/transactionBuilder/transferBuilder.ts
```diff
@@ -0,0 +1,46 @@
...
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+    this.transaction = new Transaction(_coinConfig);
+  }
+
+  protected get transactionType(): TransactionType {
+    return TransactionType.Send;
+  }
+
+  protected isValidTransactionClauses(clauses: TransactionClause[]): boolean {
```
ajays97, line 20: Will `boolean` be useful for the caller of this method?

For eg., if a transaction is for native value transfer and data is not "0x", then the caller wouldn't know what to do next?

modules/sdk-coin-vet/src/vet.ts
```diff
@@ -19,6 +21,10 @@ import utils from './lib/utils';
 import { bip32 } from '@bitgo/secp256k1';
 import { randomBytes } from 'crypto';
 import { KeyPair as EthKeyPair } from '@bitgo/abstract-eth';
+import { TransactionBuilderFactory } from './lib';
+import { ExplainTransactionOptions, VetParseTransactionOptions } from './lib/types';
+// import { AptTransactionExplanation } from "@bitgo/sdk-coin-apt/dist/src/lib/iface";
```
ajays97: Can we remove this if not being used

modules/sdk-coin-vet/src/lib/utils.ts
```diff
@@ -41,6 +42,27 @@ export class Utils implements BaseUtils {
     const regex = new RegExp(`^(0x|0X)[a-fA-F0-9]{${length}}$`);
     return regex.test(value);
   }
+
+  deserializeSignedTransaction(rawTransaction: string): Transaction {
+    const txBytes = HexUInt.of(rawTransaction).bytes;
+    try {
+      return Transaction.decode(txBytes, false);
+    } catch (err) {
+      if (err.message?.includes('Expected 9 items, but got 10')) {
```
ajays97, line 51: Can we not rely on the error message content? This method expects a signed txn, so, we have to be always calling `Transaction.decode(txBytes, true)` right?

modules/sdk-coin-vet/src/lib/utils.ts
```diff
@@ -41,6 +42,27 @@ export class Utils implements BaseUtils {
     const regex = new RegExp(`^(0x|0X)[a-fA-F0-9]{${length}}$`);
     return regex.test(value);
   }
+
+  deserializeSignedTransaction(rawTransaction: string): Transaction {
```
ajays97: Why is this parameter called `rawTransaction`, aren't we expecting signed txn to this?

modules/sdk-coin-vet/src/vet.ts
```diff
@@ -136,16 +198,26 @@ export class Vet extends BaseCoin {
...
   }
 
-  protected getTxBuilderFactory() {
-    throw new Error('Method not implemented.');
+  protected getTxBuilderFactory(): TransactionBuilderFactory {
+    return new TransactionBuilderFactory(this._staticsCoin);
+  }
+
+  protected getTransactionBuilder(): TransactionBuilderFactory {
+    return this.getTxBuilderFactory();
```
ajays97: Can both of these be clubbed into one function? Since we are not really passing any parameter to `getTxBuilderFactory` method.

modules/sdk-coin-vet/src/vet.ts
```diff
@@ -136,16 +198,26 @@ export class Vet extends BaseCoin {
...
   }
 
   protected async rebuildTransaction(txHex: string): Promise<BaseTransaction> {
-    throw new Error('Method not implemented');
+    const txBuilderFactory = this.getTxBuilderFactory();
+    try {
+      const txBuilder = txBuilderFactory.from(txHex);
+      return txBuilder.getRebuiltTransaction();
```
ajays97: Can we simply use `getTransaction` instead. We don't really need a separate function for just returning `this.transaction` unless we plan to update the logic later on?

ajays97 (changes requested):

modules/sdk-coin-vet/src/lib/transactionBuilder/transferBuilder.ts
at31416: This method is specific for native transfer, all the native transactions has `0x` as `data`.

modules/sdk-coin-vet/src/lib/transactionBuilder/transactionBuilder.ts
at31416: removed this method and removed protected from transaction getter.

modules/sdk-coin-vet/src/vet.ts
at31416: yes, removed. thanks for pointing.

modules/sdk-coin-vet/src/lib/utils.ts
at31416: renamed to serializedTransaction. the method is also renamed to deserializeTransaction as it handles both signed/unsigned serialized transaction.

modules/sdk-coin-vet/src/vet.ts
at31416: yes. done.

modules/sdk-coin-vet/src/vet.ts
at31416: updated.

modules/sdk-coin-vet/src/lib/utils.ts
at31416: this method is responsible for deserializing signed/unsigned serialized transaction (renamed the method as well now). the error which is thrown doesn't have a error code but a methodName `RLP.unpackData()` which we cannot rely on, therefore used the error message instead.

ajays97 (approved):

modules/sdk-coin-vet/src/lib/transactionBuilder/transferBuilder.ts
ajays97: Yes, what I mean is, instead of boolean, would this be more helpful if we throw an error message instead? Since we could be returning `false` in multiple scenarios.

PR #6260, opened 2025-06-09
fix: creating transfer before sending tx

parvbitgo067:
Currently we are creating transfer during sendTx. This flow has 2 problems: 
- Flow different from the regular 2 of 3 wallets.
- After sendTx there is no proper way to get the transfer to return back to the client especially in withdraw flow.
So we will be creating the transfer in BitGoJS before sending the sendTx and removing the logic of creating in sendTx from WP.

TICKET: BTC-2146

### DIFF
--- a/modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
+++ b/modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
@@ -435,6 +435,17 @@ export class PendingApproval implements IPendingApproval {
             assert(this._pendingApproval.txRequestId, 'Missing txRequestId');
             // this.populateWallet is called before this so we should be good here
             assert(this.wallet?.id(), 'Missing wallet id');
+            // todo: add test case for this
+            // add new transfer before sending tx
+            await this.bitgo
+              .post(
+                this.bitgo.url(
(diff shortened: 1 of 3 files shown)


davidkaplanbitgo (dismissed):

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
```diff
@@ -435,6 +435,17 @@ export class PendingApproval implements IPendingApproval {
             assert(this._pendingApproval.txRequestId, 'Missing txRequestId');
             // this.populateWallet is called before this so we should be good here
             assert(this.wallet?.id(), 'Missing wallet id');
+            // add new transfer before sending tx
+            // todo:(current) : test case for this???
```
mohammadalfaiyazbitgo: is this still valid?

mohammadalfaiyazbitgo (changes requested):

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
parvbitgo067: Will do this in follow up PR

saravanan7mani (approved):

mohammadalfaiyazbitgo (approved):

PR #6265, opened 2025-06-09
feat(root): address ws vulnerability

zahin-mohammad:
The ws vulnerability does not affect bitgojs, however it
is a cause of alarm when viewing the audit report for this repo. 

Unfortunately the vulnerability comes from a deep nested dependency with celo,
where even upgrading to the latest celo deps will not fix the issue. 

This PR creates a resolution for "ws" to address the audit.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -20002,11 +20002,6 @@ uint8array-tools@0.0.7:
   resolved "https://registry.npmjs.org/uint8array-tools/-/uint8array-tools-0.0.7.tgz#a7a2bb5d8836eae2fade68c771454e6a438b390d"
   integrity sha512-vrrNZJiusLWoFWBqz5Y5KMCgP9W9hnjZHzZiZRT8oNAkq3d5Z5Oe76jAvVVSRh4U8GGR90N2X1dWtrhvx6L8UQ==
 
-ultron@~1.1.0:
-  version "1.1.1"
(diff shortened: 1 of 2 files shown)


pranavjain97 (approved):

margueriteblair (approved):

lokesh-bitgo (approved):

yarn.lock
```diff
@@ -21100,10 +21095,12 @@ write-pkg@^4.0.0:
     type-fest "^0.4.1"
     write-json-file "^3.2.0"
 
-ws@7.4.6:
-  version "7.4.6"
-  resolved "https://registry.npmjs.org/ws/-/ws-7.4.6.tgz#5654ca8ecdeee47c33a9a4bf6d28e2be2980377c"
-  integrity sha512-YmhHDO4MzaDLB+M9ym/mDA5z0naX8j7SIlT8f8z+I0VtzsRbekxEutHSme7NPS2qE8StCYQNUnfWdXta/Yu85A==
+ws@7.4.6, ws@8.8.0, ws@^3.0.0, ws@^5.2.4:
+  version "5.2.4"
```
lokesh-bitgo, line 21099: Are we locking the `ws` version to `5.2.4`? Some dependent packages require `ws@8.8.0`, but this change forces them to use `5.2.4`, which might not support features introduced in the newer versions. Could this lead to runtime issues or missing functionality?

zahin-mohammad: Good point, we probably shouldn't force the resolution for higher versions. I'll investigate what nested dep caused this.

PR #6270, opened 2025-06-10
feat(sdk-coin-near): added verify transaction for near token & storage deposit transactions

ravibitgo:
Ticket: [COIN-4150](https://bitgoinc.atlassian.net/browse/COIN-4150)

[COIN-4150]: https://bitgoinc.atlassian.net/browse/COIN-4150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/sdk-coin-near/src/lib/transaction.ts
```diff
@@ -522,9 +522,9 @@ export class Transaction extends BaseTransaction {
         }
       }
     }
-    if (hasFtTransfer && hasStorageDeposit) {
+    if (hasFtTransfer) {
       return totalTokenAmount.toString();
-    } else if (!hasFtTransfer && hasStorageDeposit) {
+    } else if (hasStorageDeposit) {
       return totalNearDeposit.toString();
```
gianchandania, line 528: won't there be a case, when a single tx is having both?

modules/sdk-coin-near/src/lib/transaction.ts
```diff
@@ -573,10 +580,20 @@ export class Transaction extends BaseTransaction {
...
+              ? (functionCall.args['account_id'] as string)
+              : json.signerId;
+          // in storage deposit, the outer receiver id will be contract address of the token
+          const tokenName = utils.findTokenNameFromContractAddress(json.receiverId);
+          const output: ITransactionRecipient = {
+            address: receiverId,
             amount: functionCall.deposit,
-          });
+          };
+          if (tokenName) {
```
gianchandania, line 593: In storage deposit, will we have to add tokenName?

modules/sdk-coin-near/src/near.ts
```diff
@@ -755,16 +755,28 @@ export class Near extends BaseCoin {
 
     // users do not input recipients for consolidation requests as they are generated by the server
     if (txParams.recipients !== undefined) {
-      const filteredRecipients = txParams.recipients?.map((recipient) => _.pick(recipient, ['address', 'amount']));
-      const filteredOutputs = explainedTx.outputs.map((output) => _.pick(output, ['address', 'amount']));
+      const filteredRecipients = txParams.recipients?.map((recipient) => {
+        if (txParams.type !== 'enabletoken') {
```
gianchandania, line 773: trying to understand why enabletoken is handled in verifyTransaction of near class and not nep141token class?

modules/sdk-coin-near/src/near.ts
```diff
@@ -755,16 +755,28 @@ export class Near extends BaseCoin {
...
 
       if (!_.isEqual(filteredOutputs, filteredRecipients)) {
         throw new Error('Tx outputs does not match with expected txParams recipients');
       }
       for (const recipients of txParams.recipients) {
-        totalAmount = totalAmount.plus(recipients.amount);
+        totalAmount = txParams.type !== 'enabletoken' ? totalAmount.plus(recipients.amount) : BigNumber(0);
       }
-      if (!totalAmount.isEqualTo(explainedTx.outputAmount)) {
+      if (!totalAmount.isEqualTo(explainedTx.outputAmount) && txParams.type !== 'enabletoken') {
```
gianchandania, line 793: don't we need to check anything related to amount in case of enable token

modules/sdk-coin-near/src/lib/transaction.ts
ravibitgo: When both storage deposit & ft transfers are in a single tx, from the input perspective it will still be ft transfer. So we are handling only that.

modules/sdk-coin-near/src/lib/transaction.ts
ravibitgo: Yes, it is `enableToken`  flow

modules/sdk-coin-near/src/near.ts
ravibitgo: In this, the currency involved it the native near so keeping it in the near class.

modules/sdk-coin-near/src/near.ts
ravibitgo: Since we started keeping the storage deposit amount in the sdk, will add a check for that amount.

gianchandania (dismissed):

gianchandania (approved):

PR #6272, opened 2025-06-10
remove statics dependency from sol module 

mukeshsp:
Refer to the ticket description for details.
Ticket : WIN-5679


modules/sdk-coin-sol/src/lib/transaction.ts
```diff
@@ -291,7 +303,8 @@ export class Transaction extends BaseTransaction {
     const instructionParams = instructionParamsFactory(
       this.type,
       this._solTransaction.instructions,
-      this._coinConfig.name
+      this._coinConfig.name,
+      this._instructionsData
```
veetragjain: We may need to pass the token contract address in outputs, instructionData might not be available when building from a raw tx.

mukeshsp: Could you share the example ?

veetragjain: Try to build a transaction from raw txHex of any unsupported token transaction, it will throw an error.

modules/sdk-coin-sol/src/lib/instructionParamsFactory.ts
```diff
@@ -840,6 +869,26 @@ function findTokenName(mintAddress: string): string {
...
+        instruction.type === InstructionBuilderTypes.TokenTransfer &&
+        instruction.params.tokenAddress === mintAddress
+      ) {
+        token = instruction.params.tokenName;
+      }
+    });
+  }
+
+  if (!token && _useTokenAddressTokenName) {
+    token = mintAddress;
```
veetragjain, line 889: Why not pass tokenName undefined and token address as a separate fields and the populate token name and token address both in outputs.

mukeshsp: Since the approach was suggested by you, we can't pass the token name as undefined—these instructions are used in multiple places, which could lead to errors due to the undefined value.

veetragjain (approved):

abhishekagrawal080 (approved):

PR #6278, opened 2025-06-10
feat(soneium): update multisig type for soneium and add new features

mohd-kashif:
Ticket: WIN-5732

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -362,6 +362,11 @@ export enum CoinFeature {
   TSS_SUPPORT_GATED = 'tss-support-gated',
 
   /**
+   * This coin is gated for Multisig Support
+   */
+  MULTISIG_SUPPORT_GATED = 'multisig-support-gated',
+
+  /**
    * This coins is an EVM compatible coin and should use common EVM functionality
    */
   SHARED_EVM_SIGNING = 'shared-evm-signing',
--- a/modules/statics/src/coinFeatures.ts
+++ b/modules/statics/src/coinFeatures.ts
@@ -510,6 +510,20 @@ export const TAO_FEATURES = [
   CoinFeature.STAKING,
 ];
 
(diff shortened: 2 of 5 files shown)


modules/statics/src/base.ts
```diff
@@ -361,6 +361,11 @@ export enum CoinFeature {
    */
   TSS_SUPPORT_GATED = 'tss-support-gated',
 
+  /**
+   * This coin is gated for TSS Support
+   */
+  MULTISIG_SUPPORT_GATED = 'multisig-support-gated',
```
sachushaji, line 367: If the flag is meant to gate tss, then it should be named like that? or did you mean multisig support gated?

sachushaji: Whats the reason for gating multisig support?

mohd-kashif: @sachushaji the flag is meant to gate multisig support and the naming is like the TSS flag `TSS_SUPPORT_GATED`
the reason is to gate this feature until all the development (WP and HSM changes) is complete and then remove this flag at the end

sachushaji: The comment of the flag is not right?  should say gate multisig support

mohd-kashif: corrected, thanks

modules/sdk-coin-soneium/src/soneium.ts
```diff
@@ -38,7 +38,7 @@ export class Soneium extends AbstractEthLikeNewCoins {
 
   /** inherited doc */
   getDefaultMultisigType(): MultisigType {
-    return multisigTypes.tss;
+    return multisigTypes.onchain;
```
sachushaji: Should we keep tss until multsig is there in prod?

mohd-kashif: done

sachushaji (approved):

PR #6280, opened 2025-06-10
refactor(sdk-coin-eos): use token symbol while verifying transaction

veetragjain:
Ticket: WIN-5533

### DIFF
--- a/modules/sdk-coin-eos/src/eos.ts
+++ b/modules/sdk-coin-eos/src/eos.ts
@@ -3,6 +3,7 @@
  */
 import { BigNumber } from 'bignumber.js';
 import { bip32, BIP32Interface } from '@bitgo/secp256k1';
+import { coins, EosCoin } from '@bitgo/statics';
 import { createHash, randomBytes } from 'crypto';
 import { Api, ApiInterfaces, JsonRpc, RpcInterfaces } from 'eosjs';
 import * as ecc from 'eosjs-ecc';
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-eos/src/eos.ts
```diff
@@ -1123,7 +1124,7 @@ export class Eos extends BaseCoin {
       }
 
       if (txPrebuild.coin === 'eos' || txPrebuild.coin === 'teos') {
-        const expectedSymbol = _.isNil(txPrebuild.token) ? 'EOS' : txPrebuild.token.split(':')[1];
+        const expectedSymbol = _.isNil(txPrebuild.token) ? 'EOS' : (coins.get(txPrebuild.token) as EosCoin).symbol;
```
mukeshsp, line 1127: Can we pass tokenSymbol as a parameter instead of fetching it from statics?
As you know, we're working on removing the statics dependency from the module as part of the AMS integration. So, let's avoid using statics for any new integrations.

mukeshsp (changes requested):

modules/sdk-coin-eos/src/eos.ts
veetragjain: It cannot be passed down, it can only be fetched from statics coin map in current state. Moreover this function is not used to build transactions on wp, it's used on client side to verify the build obtained from wp.

mukeshsp (approved):

PR #6293, opened 2025-06-13
feat(sdk-coin-near): added storage deposit amount to token config

ravibitgo:
Ticket: [COIN-4495](https://bitgoinc.atlassian.net/browse/COIN-4495)

[COIN-4495]: https://bitgoinc.atlassian.net/browse/COIN-4495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/statics/src/coins/nep141Tokens.ts
+++ b/modules/statics/src/coins/nep141Tokens.ts
@@ -9,6 +9,7 @@ export const nep141Tokens = [
     'USD Coin',
     6,
     '17208628f84f5d6ad33f0da3bbbeb27ffcb398eac501a31bd6ad2011e36133a1',
+    '1250000000000000000000',
     UnderlyingAsset['near:usdc'],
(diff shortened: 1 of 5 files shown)


modules/statics/src/coins/nep141Tokens.ts
```diff
@@ -9,6 +9,7 @@ export const nep141Tokens = [
     'USD Coin',
     6,
     '17208628f84f5d6ad33f0da3bbbeb27ffcb398eac501a31bd6ad2011e36133a1',
+    '1250000000000000000000',
```
gianchandania, line 12: As mentioned here in NEAR docs, https://docs.near.org/primitives/ft#registering-a-user
this amount seems to be same for all tokens. Hence no need to add separately for each token

PR #6300, opened 2025-06-16
feat(sdk-api): add method to create request for public AMS endpoints

veetragjain:
The AMS endpoints are public and do not need any HMAC validations, this PR adds a method to create and execute requests for public AMS endpoints. A follow up PR will be created in `bitgo` package to use this method and fetch asset details.

Ticket: WIN-5839

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -2102,4 +2102,27 @@ export class BitGoAPI implements BitGoBase {
     // use defaultConstants as the backup for keys that are not set in this._constants
     return _.merge({}, defaultConstants(this.getEnv()), BitGoAPI._constants[this.getEnv()]);
   }
+
+  /**
+   * Execute an asset request which does not need HMAC validation
+   * @param url The URL for the asset request
+   * @returns {Promise<any>} The response body
+   */
+  protected async executeAssetRequest(url: string): Promise<any> {
+    const req = this.getAgentRequest('get', url);
+    req.set('BitGo-SDK-Version', this._version);
+    if (this._customProxyAgent) {
+      req.agent(this._customProxyAgent);
+    }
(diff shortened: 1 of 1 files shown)


modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -2102,4 +2102,25 @@ export class BitGoAPI implements BitGoBase {
...
     return _.merge({}, defaultConstants(this.getEnv()), BitGoAPI._constants[this.getEnv()]);
   }
+
+  /**
+   * Execute an asset request which does not need HMAC validation
+   * @param url The URL for the asset request
+   * @returns {Promise<any>} The response body
+   */
+  protected async executeAssetRequest(url: string): Promise<any> {
+    const req = this.getAgentRequest('get', url);
```
Doddanna17, line 2112: Hi @veetragjain ,
The current implementation directly uses the url input without validation, which can make the app vulnerable to SSRF (Server-Side Request Forgery).

Could you please consider adding the below checks & configs as part of code.

1. URL validation – allow only trusted domains.
2. SSRF protection – block internal/private IPs like 127.0.0.1, 169.254.x.x, etc.
3. Request timeout – to prevent hanging or resource exhaustion.

Doddanna17 (changes requested):
Requesting for validation changes.

modules/sdk-api/src/bitgoAPI.ts
veetragjain: This is a protected function, it will only be used by other functions in this or inheriting classes. The `url` function of this class only allows trusted domains and that is how I intend to build and pass the url to this function.

Doddanna17: Okay.

veetragjain: Added timeout, same as that we have in `requestPatch` function.

PR #6311, opened 2025-06-17
fix(sdk-coin-sui): fix contract address for sui wusdc token

rakeshgsharma:
TICKET: WIN-5935


balakrishna507 (approved):
this is in line with what suivision has too: https://suivision.xyz/coin/0x5d4b302506645c37ff133b98c4b50a5ae14841659738d6d733d59d0d217a93bf::coin::COIN

we have gone through a few other tokens mysten has, and there shouldn't be any more.

PR #6316, opened 2025-06-18
feat: adding new batch of tokens

Doddanna17:
feat: adding new batch of tokens

TICKET: COIN-4565


yash-bitgo (changes requested):
lets fix the failing tests

yash-bitgo (approved):
lgtm

PR #6317, opened 2025-06-18
fix(sdk-coin-sol): handling sol 2022 token recoveries

abhishekagrawal080:
TICKET: WIN-5929

### DIFF
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -58,6 +58,8 @@ import {
   isValidPublicKey,
   validateRawTransaction,
 } from './lib/utils';
+import { TOKEN_2022_PROGRAM_ID, TOKEN_PROGRAM_ID } from '@solana/spl-token';
+
 export const DEFAULT_SCAN_FACTOR = 20; // default number of receive addresses to scan for funds
 
 export interface TransactionFee {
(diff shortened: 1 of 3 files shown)


modules/sdk-coin-sol/src/sol.ts
```diff
@@ -154,6 +155,7 @@ export interface SolRecoveryOptions extends MPCRecoveryOptions {
   closeAtaAddress?: string;
   // destination address where token should be sent before closing the ATA address
   recoveryDestinationAtaAddress?: string;
+  isSol2022Token?: boolean; // whether the token is a Sol 2022 token
```
mukeshsp: we can take packageID as optional param instead of taking flag

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -154,6 +156,7 @@ export interface SolRecoveryOptions extends MPCRecoveryOptions {
   closeAtaAddress?: string;
   // destination address where token should be sent before closing the ATA address
   recoveryDestinationAtaAddress?: string;
+  programId?: string; // whether the token is a Sol 2022 token
```
mukeshsp:
```suggestion
  programId?: string; // programId of the token
```

abhishekagrawal080: done

modules/sdk-coin-sol/src/sol.ts
abhishekagrawal080: done

mukeshsp (approved):

PR #6321, opened 2025-06-19
feat(bitgo): use AMS endpoints to register single token

veetragjain:
The register token function was introduced in a recent release, it's not used by clients or any other internal services. This PR refactors the register token function to get token details from AMS and register it in coin factory. 

Ticket: WIN-5839


modules/bitgo/src/v2/coinFactory.ts
```diff
@@ -482,6 +482,329 @@ export function registerCoinConstructors(coinFactory: CoinFactory, coinMap: Coin
...
+      return Bch.createInstance;
+    case 'bcha':
+      return Bcha.createInstance;
+    case 'bera':
+      return Bera.createInstance;
+    case 'bld':
+      return Bld.createInstance;
+    case 'bsc':
+      return Bsc.createInstance;
+    case 'bsv':
```
pritam-gembali, line 517: Do we have support for Bitcoin Satoshi Vision?

veetragjain: I'm not sure about that, added it here because it's [present](https://github.com/BitGo/BitGoJS/blob/master/modules/bitgo/src/v2/coinFactory.ts#L225) in global coin factory.

nvrakesh06 (approved):

PR #6332, opened 2025-06-22
fix(sdk-coin-trx): fees on freeze, vote transactions

benkelcher-bitgo:
SC-2215
- **TODO** create and link ticket
- adds `fee` as a required field on the freeze and vote transactions used in the TRX staking flow
- context: we need to require that these staking txns (and in particular the vote txn) use TRX to pay their gas fees instead of BANDWIDTH, so that our transactions generate transfers (otherwise these txns get stuck in signed on the WP side)
  - see [this thread](https://bitgo.slack.com/archives/C0931KN7F40/p1750548973880639?thread_ts=1750542863.741069&cid=C0931KN7F40)
  - longer-term it would be nice to be able to use BANDWIDTH to pay the fees

### DIFF
--- a/modules/sdk-coin-trx/src/lib/freezeBalanceTxBuilder.ts
+++ b/modules/sdk-coin-trx/src/lib/freezeBalanceTxBuilder.ts
@@ -1,9 +1,16 @@
 import { createHash } from 'crypto';
-import { TransactionType, BaseKey, ExtendTransactionError, BuildTransactionError, SigningError } from '@bitgo/sdk-core';
-import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import {
+  TransactionType,
+  BaseKey,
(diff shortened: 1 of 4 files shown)


evanzbitgo:
Change LGTM. We might just do some tests

evanzbitgo:
> > Change LGTM. We might just do some tests
> 
> @evanzbitgo I don't think we might need this since fee limit is used only for contract call txns to not burn TRX beyond a limit incase there's a high energy consumption. In our case, no energy is used + additionally, we will always burn atleast 1 TRX since its a multi-sig txn. have added more details in the slack channel

I just read the conversation there https://bitgo.slack.com/archives/C0931KN7F40/p1750564616420339?thread_ts=1750560353.377609&cid=C0931KN7F40 yeah, then we might not need this. But we should test

PR #6355, opened 2025-06-25
chore: dynamic import

ppongbitgo:


OttoAllmendinger (changes requested):
`Descriptor` is an actual class, so I don't think this works

PR #6365, opened 2025-06-27
feat(express): add name field with default name on error

starfy84:
Ticket: DX-1197

This PR ensures that the name field is populated when there's an error. Additionally, if the error does not have a name, the name will be set to "BitGoExpressError" by default.


Taseen08 (approved):
LGTM

PR #6380, opened 2025-07-01
refactor(utxo-core): trimMessagePrefix fix function

christopher-fong:
TICKET: BTC-2235
Fixed the trimMessagePrefix function as it was trimming the message wrong and leaving a space in the beginning

### DIFF
--- a/modules/utxo-core/src/paygo/trimMessagePrefix.ts
+++ b/modules/utxo-core/src/paygo/trimMessagePrefix.ts
@@ -2,24 +2,24 @@ import * as utxolib from '@bitgo/utxo-lib';
 
 /** We receive a proof in the form:
  * 0x18Bitcoin Signed Message:\n<varint_length><ENTROPY><ADDRESS><UUID>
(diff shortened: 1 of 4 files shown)


modules/utxo-core/test/paygo/psbt/payGoAddressProof.ts
```diff
@@ -131,3 +134,37 @@ describe('getPaygoAddressProofIndex', () => {
     );
   });
 });
+
+// Small test to verify, leave in this test file
```
OttoAllmendinger:
```suggestion
```

modules/utxo-core/src/paygo/trimMessagePrefix.ts
```diff
@@ -11,15 +11,16 @@ import * as utxolib from '@bitgo/utxo-lib';
  */
 export function trimMessagePrefix(proof: Buffer): Buffer {
```
OttoAllmendinger, line 11: change the jsdoc so that it describes what this func does (the reason why it does that is secondary)

OttoAllmendinger (changes requested):

PR #6382, opened 2025-07-02
feat: add icp staking flow

Gautam2305:
Ticket: SC-1970


Gautam2305:
flushing - not required

PR #6390, opened 2025-07-03
feat(sdk-coin-trx): add support for reward claim

Vijay-Jagannathan:
Ticket: SC-2333

Add support for reward claim txn

### DIFF
--- a/modules/sdk-coin-trx/src/lib/iface.ts
+++ b/modules/sdk-coin-trx/src/lib/iface.ts
@@ -50,6 +50,7 @@ export interface RawData {
     | VoteWitnessContract[]
     | UnfreezeBalanceV2Contract[]
     | WithdrawExpireUnfreezeContract[]
+    | WithdrawBalanceContract[]
     | ResourceManagementContract[];
 }
 
@@ -208,6 +209,12 @@ export interface WithdrawExpireUnfreezeValue {
  * Withdraw expire unfreeze contract interface
  */
 export interface WithdrawExpireUnfreezeContract {
+  parameter: WithdrawExpireUnfreezeValue;
+  type?: string;
+}
+
+export interface WithdrawBalanceContract {
+  // same as withdraw expire unfreeze
   parameter: WithdrawExpireUnfreezeValue;
   type?: string;
 }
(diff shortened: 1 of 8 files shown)


modules/sdk-coin-trx/src/lib/iface.ts
```diff
@@ -212,6 +213,11 @@ export interface WithdrawExpireUnfreezeContract {
   type?: string;
 }
 
+export interface WithdrawBalanceContract {
+  parameter: WithdrawExpireUnfreezeValue;
```
Vijay-Jagannathan, line 218: we're using `WithdrawExpireUnfreezeValue` since its the sam; didn't want to create a duplicate type

modules/sdk-coin-trx/src/lib/utils.ts
```diff
@@ -564,6 +568,40 @@ export function decodeUnfreezeBalanceV2Contract(base64: string): UnfreezeBalance
   ];
 }
 
+/**
+ * Deserialize the segment of the txHex corresponding with withdraw balance contract
+ *
+ * @param {string} base64 - The base64 encoded contract data
+ * @returns {WithdrawExpireUnfreezeContractParameter[]} - Array containing the decoded withdraw contract
+ */
+export function decodeWithdrawBalanceContract(base64: string): WithdrawExpireUnfreezeContractParameter[] {
```
Vijay-Jagannathan, line 578: same here, `WithdrawExpireUnfreezeContractParameter` is same as the type for withdraw balance too

modules/sdk-coin-trx/src/lib/iface.ts
gianchandania: let's add a comment in the code as well maybe for clarification then, otherwise it might be confusing for someone looking at the code

modules/sdk-coin-trx/src/lib/utils.ts
gianchandania: same as above, either 

we can add a comment to clarify, or 
we can use an alias like export type WithdrawBalanceContractParameter = WithdrawExpireUnfreezeContractParameter to avoid confusion. same for above as well

Vijay-Jagannathan: adding

gianchandania (approved):

PR #6399, opened 2025-07-07
feat(babylonlabs-io-btc-staking-ts): update vendor script

OttoAllmendinger:
Enhancing the vendor-github-repo script with cherry-pick capability.

This makes it easier to vendor the dependency for a new version.

This change also contains some minor changes to the vendored dependency, staying on `v1.0.3` but
tweaking the README.

Reference: BTC-2143


modules/babylonlabs-io-btc-staking-ts/src/index.ts
```diff
@@ -1,13 +1,21 @@
...
+export * from "./utils/utxo/getScriptType";
+export {
+  getBabylonParamByBtcHeight,
+  getBabylonParamByVersion,
+} from "./utils/staking/param";
+export * from "./staking/manager";
+
+// BitGo-specific exports
+export * from "./utils/babylon";
+export * from "./utils/staking";
```
kamleshmugdiya: nit: Please add a newline at the end of the file.

scripts/vendor-github-repo.ts
```diff
@@ -1,28 +1,61 @@
...
-      }
-    });
+export function getDescription(source: GithubSource): string {
+  if ('tag' in source) {
+    return `${source.org}/${source.repo}#${source.tag}`;
+  } else if ('ref' in source) {
+    return `${source.org}/${source.repo}#${source.ref}`;
+  } else {
+    throw new Error('Invalid GithubSource, must have either tag or ref');
+  }
```
kamleshmugdiya, line 18: nit: can be simplified to avoid if/else if structure.
```suggestion
  const tagOrRef = source.tag ?? source.ref;

  if (tagOrRef) {
    return `${source.org}/${source.repo}#${tagOrRef}`;
  }

  throw new Error('Invalid GithubSource, must have either tag or ref');
```

OttoAllmendinger: the type signature only allows either `tag` or `ref`, not both properties, so this won't work

kamleshmugdiya: you're right.
my bad.

kamleshmugdiya (approved):

davidkaplanbitgo (approved):
lgtm

PR #6406, opened 2025-07-07
fix(bitgo): register ofc constructor without using AMS api

veetragjain:
Ticket: WIN-6095


modules/bitgo/src/bitgo.ts
```diff
@@ -111,6 +112,11 @@ export class BitGo extends BitGoAPI {
       return;
     }
 
+    //ofc is not present in statics coin map
+    if (tokenName === 'ofc') {
```
feng-bitgo, line 116: Is this supposed to cover the ofc version of coins as well eg. ofcsol ofceth

veetragjain: ofcsol and all such other tokens are present in statics coin map, this covers only ofc token.

feng-bitgo (approved):

PR #6413, opened 2025-07-08
refactor(utxo-core): Pay Go Functions prefix

christopher-fong:
TICKET: BTC-2251

This pr is to add automatic parsing on our PayGo proof, so that if the prefix `\x18Bitcoin Signed Message:\n` is in/not in our proof and will automatically parse accordingly into its parts `{ entropy, address, uuid }`.


davidkaplanbitgo (approved):
nice dude

PR #6415, opened 2025-07-09
chore(utxo-lib): change mock UTXO to P2PKH from P2WPKH

davidkaplanbitgo:
Update the mock previous transaction to use P2PKH rather than P2WPKH for better compatibility (doge, dash) with testing needs. Since some chains dont support segwit, we cannot use this function for mocking transactions for these coins. Better to just use the simplest input and sign and finalize the transaction without requiring a `prevTx`, which I assume is the primary reason for going with `p2wpkh` in the beginning.

Issue: BTC-2259


modules/utxo-lib/test/bitgo/psbt/ZcashPsbt.ts
```diff
@@ -13,7 +13,7 @@ const rootWalletKeys = getDefaultWalletKeys();
 describe('Zcash PSBT', function () {
   let psbt: utxolib.bitgo.ZcashPsbt;
   before(async function () {
-    const unspents = mockUnspents(rootWalletKeys, ['p2sh'], BigInt('10000000000000000'), network);
+    const unspents = mockUnspents(rootWalletKeys, ['p2sh'], BigInt('1000000000000000'), network);
```
davidkaplanbitgo, line 16: This was throwing an error, i am not sure why it is happening now and not before. The error is that it was not allowing us to add 100m as an input to a transaction on a network that has a max supply of 21m

OttoAllmendinger (dismissed):

margueriteblair (approved):
jesus dude

christopher-fong (approved):

PR #6423, opened 2025-07-09
feat(sdk-coin-sui): fee payer signing support

abhishekagrawal080:
This change allow sponsor/fee address signing in case of sui, currently sui fee amount is paid from sender address so there was no need set fee address and fee signature, but for new sponsored transaction we will provide fee adress and fee signature as well

TICKET: WIN-6028

### DIFF
--- a/modules/sdk-coin-sui/src/lib/transaction.ts
+++ b/modules/sdk-coin-sui/src/lib/transaction.ts
@@ -24,6 +24,8 @@ export abstract class Transaction<T> extends BaseTransaction {
   protected _suiTransaction: SuiTransaction<T>;
   protected _signature: Signature;
   private _serializedSig: Uint8Array;
+  protected _feePayerSignature: Signature;
+  private _serializedFeePayerSig: Uint8Array;
 
   protected constructor(_coinConfig: Readonly<CoinConfig>) {
     super(_coinConfig);
@@ -51,12 +53,22 @@ export abstract class Transaction<T> extends BaseTransaction {
     this.serialize();
   }
 
+  addFeePayerSignature(publicKey: BasePublicKey, signature: Buffer): void {
+    this._signatures.push(signature.toString('hex'));
+    this._feePayerSignature = { publicKey, signature };
+    this.serialize();
+  }
+
   get suiSignature(): Signature {
     return this._signature;
   }
 
   get serializedSig(): Uint8Array {
     return this._serializedSig;
+  }
+
+  get serializedFeePayerSig(): Uint8Array {
+    return this._serializedFeePayerSig;
   }
 
   setSerializedSig(publicKey: BasePublicKey, signature: Buffer): void {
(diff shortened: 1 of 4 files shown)


modules/sdk-coin-sui/test/unit/transactionBuilder/feePayer.ts
```diff
@@ -0,0 +1,205 @@
...
+      should.doesNotThrow(() => Buffer.from(rawTx, 'base64'));
+    });
+
+    it('should be able to add sponsor after setting gasData', async function () {
+      const feePayerPrv = testData.privateKeys.prvKey3;
+      const feePayerKeyPair = new KeyPair({ prv: feePayerPrv });
+      const feePayerAddress = feePayerKeyPair.getAddress();
+
+      const txBuilder = factory.getTransferBuilder();
+      txBuilder.type(SuiTransactionType.Transfer);
```
balakrishna507: are we going to stick with this transaction type, or create a new one?

krupanand-bitgo:
Please update ticket and PR descriptions for better clarity and easier review.

modules/sdk-coin-sui/test/unit/transactionBuilder/feePayer.ts
abhishekagrawal080: we will create new one in another ticket

abhishekagrawal080:
> Please update ticket and PR descriptions for better clarity and easier review.

sure will add more details

modules/sdk-coin-sui/src/lib/transaction.ts
```diff
@@ -87,16 +111,56 @@ export abstract class Transaction<T> extends BaseTransaction {
...
+    };
+
+    if (this._suiTransaction.gasData?.sponsor) {
+      if (!this._serializedFeePayerSig) {
+        throw new InvalidTransactionError('Sponsored transaction must have fee payer signature');
+      }
+      result['sponsorSignature'] = toB64(this._serializedFeePayerSig);
+    }
+
+    return JSON.stringify(result);
```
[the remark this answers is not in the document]
balakrishna507: @abhishekagrawal080 isn't this a problem? `toBroadcastFormat()` should return a b64 or hex string of the transaction data right? not json?

modules/sdk-coin-sui/test/unit/transactionBuilder/feePayer.ts
balakrishna507: actually, on second thought, do we really need a new one? if the wallet has a gas tank we'll set the sponsor on the gas objects, and if it doesn't we won't, right?

krupanand-bitgo (dismissed):
Lgtm

modules/sdk-coin-sui/test/unit/transactionBuilder/feePayer.ts
abhishekagrawal080: Thats true if the wallet has a gas tank we'll set the sponsor on the gas objects, and if it doesn't we won't, right? thats why i am thinking if we need then only we will create new type

veetragjain (approved):
LGTM

krupanand-bitgo (approved):

PR #6428, opened 2025-07-09
fix(sdk-coin-vet): fix issue in fromRawTransaction for vechain

gianchandania:
COIN-4742
TICKET: COIN-4742


margueriteblair (commented):
So for my own understanding, each type of txn on vechain needs will have specific inputs/outputs that need to be loaded up? I guess for me it's not exactly clear what the inputs and outputs are in each method

margueriteblair (dismissed):
My comment is really just for my own understanding, tests seem good to me

at31416 (dismissed):
LGTM!

gianchandania:
> So for my own understanding, each type of txn on vechain needs will have specific inputs/outputs that need to be loaded up? I guess for me it's not exactly clear what the inputs and outputs are in each method

Yeah, actually the name of the method was bit ambiguous with respect to types of transaction other than Send. As Address init and flush token transactions don't have any inputs and outputs. Hence I have removed the loadInputsAndOutputs method from other 2 types.

at31416 (approved):
LGTM

PR #6438, opened 2025-07-11
chore: fix underlyingAsset for test token and coins

mukeshsp:
The `underlyingAsset` should be set to the corresponding testnet asset if the coin belongs to a testnet. This PR refactors such tokens to ensure consistency.

Ticket: WIN-6149


manas-at-bitgo (approved):

abhishekagrawal080 (approved):

modules/statics/src/base.ts
bitgoAaron, line 1: this is a potentially breaking change, several of our user interfaces were previously built under the assumption that the UnderlyingAsset provided an environment agnostic way of defining a coin. We should first do an alpha release and vet this change before merging to master

bitgoAaron (changes requested):
1. please use conventional commit messages, this type of change is not a `chore`, it should be a feature or fix.
2. please include BREAKING CHANGE, as we are now fundamentally re-defining what UnderlyingAsset is used for
3. please do an alpha release & verify this change in bitgo-ui before proceeding with merges to master

lcovar (changes requested):
agree with aaron

PR #6446, opened 2025-07-14
feat(account-lib): return hex broadcastable message string

bhavidhingra:
1. In baseMessage.ts, the toBroadcastString() method now converts the JSON string to a hexadecimal format instead of returning the plain JSON.
2. In baseMessageBuilderFactory.ts, the fromBroadcastString() method was updated to handle the new hex format by first converting it back to a string.
3. Tests were updated in multiple files to account for this change:
    - Converting hex back to string before parsing
    - Updating test descriptions and expectations

TICKET: COIN-4755

### DIFF
--- a/examples/ts/build-message.ts
+++ b/examples/ts/build-message.ts
@@ -8,8 +8,9 @@
  */
 
 import {BitGoAPI} from '@bitgo/sdk-api';
+import {MessageStandardType} from "@bitgo/sdk-core";
+import {MIDNIGHT_TNC_HASH} from "@bitgo/account-lib/dist/src/utils";
 import {Hteth} from "@bitgo/sdk-coin-eth";
-import {MessageStandardType} from "@bitgo/sdk-core"; // Replace with your given coin (e.g. Ltc, Tltc)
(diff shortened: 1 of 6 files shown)


examples/ts/build-message.ts
```diff
@@ -28,9 +29,12 @@ async function main() {
   const wallet = await bitgo.coin(coin).wallets().get({ id });
   console.log(`Wallet label: ${wallet.label()}`);
 
+  const adaTestnetDestinationAddress = 'addr_test1vz7xs7ceu4xx9n5xn57lfe86vrwddqpp77vjwq5ptlkh49cqy3wur';
```
margueriteblair, line 32: Why is this called `adaTestnetDestinationAddress` if this is for account lib?

bhavidhingra: the whitelisted message format requires an ADA destination address.
https://github.com/BitGo/BitGoJS/blob/master/modules/account-lib/src/utils/messages/index.ts#L15

hrishikeshjain (approved):

PR #6448, opened 2025-07-14
feat(sdk-core): handle TSS EDDSA message signing

bhavidhingra:
Enables BitGo's TSS implementation to sign not just transactions but also arbitrary messages using EDDSA cryptography, which is important for various blockchain operations like signing proofs and attestations.

1. In modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts:
    - Added a new method signTxRequestForMessage to support message signing using TSS EDDSA
    - Modified the existing signTxRequest to use a common base method (signRequestBase) that can handle both transaction and message signing
  2. In modules/sdk-core/src/bitgo/tss/eddsa/eddsa.ts:
    - Updated several functions to support message signing, including:
      - offerUserToBitgoRShare
      - getBitgoToUserRShare
      - sendUserToBitgoGShare
    - Added support for a new request type parameter (RequestType.message)
  3. In modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts:
    - Enhanced the TxRequest type to include additional message-related fields:
        - messageBroadcastable
        - messageStandardType
  4. In the test files:
    - Added new test cases in modules/bitgo/test/v2/unit/internal/tssUtils/eddsa.ts for message signing
    - Updated utility functions in modules/bitgo/test/v2/unit/internal/tssUtils/common.ts to handle message signing

TICKET: COIN-4820

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
@@ -355,9 +355,12 @@ export type TxRequest = {
     signatureShares: SignatureShareRecord[];
     messageRaw: string;
     messageEncoded?: string;
+    messageBroadcastable?: string;
+    messageStandardType?: MessageStandardType;
     derivationPath: string;
     combineSigShare?: string;
     txHash?: string;
+    commitmentShares?: CommitmentShareRecord[];
   }[];
   apiVersion?: TxRequestVersion;
   latest: boolean;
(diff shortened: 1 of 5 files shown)


bhavidhingra:
@claude explain the changes

hrishikeshjain (dismissed):

Taseen08 (commented):
looks good, just needs some test coverage

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -355,6 +355,8 @@ export type TxRequest = {
     signatureShares: SignatureShareRecord[];
     messageRaw: string;
     messageEncoded?: string;
+    messageBroadcastable?: string;
```
islamaminBitGo, line 358: i think we need to add commitmentShares here as well as we do in transactions. messages[idx].commitmentShares is being appended to in platform.

bhavidhingra: I've added it, but it is not being used anywhere same as the `commitmentShares` of the `transactions` list.

hrishikeshjain (approved):

PR #6449, opened 2025-07-14
feat(statics): add hypeevm chain config

DinshawKothari:
Ticket: WIN-5800


modules/statics/src/coins.ts
```diff
@@ -1333,6 +1333,26 @@ export const coins = CoinMap.fromCoins([
...
+    BaseUnit.ETH,
+    [...EVM_FEATURES, CoinFeature.SHARED_EVM_SIGNING, CoinFeature.SHARED_EVM_SDK]
+  ),
+  account(
+    'e0500947-1105-404c-af52-765b1cb2a7c1',
+    'thypeevm',
+    'Hyperliquid EVM Testnet',
+    Networks.test.hypeevm,
+    18,
+    UnderlyingAsset.HYPEEVM,
```
mukeshsp: please define **thypeevm** as UnderlyingAsset and use it

DinshawKothari: Hey, sorry we define the same asset name and not different for test and main. So using HYPEEVM is correct for testnet as well.

DinshawKothari: As per our discussion, added THYPEEVM.

mukeshsp (approved):

PR #6453, opened 2025-07-15
feat: added new error NeedsUserSignupError

sdoshibitgo:
this error tells that user needs to signup if pubkey is not found which comes when user does not have ecdhKey

TICKET: CSI-702

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -23,6 +23,7 @@ import {
   IncorrectPasswordError,
   MethodNotImplementedError,
   MissingEncryptedKeychainError,
+  NeedUserSignupError,
 } from '../errors';
 import * as internal from '../internal/internal';
 import { drawKeycard } from '../internal';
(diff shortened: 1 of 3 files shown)


ayush-9 (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1602,7 +1603,14 @@ export class Wallet implements IWallet {
     const bulkCreateShareOptions: BulkCreateShareOption[] = [];
 
     for (const shareOption of params.keyShareOptions) {
-      common.validateParams(shareOption, ['userId', 'pubKey', 'path'], []);
+      try {
+        common.validateParams(shareOption, ['userId', 'pubKey', 'path'], []);
+      } catch (e) {
+        if (!('pubkey' in shareOption) || !shareOption.pubkey) {
```
Vijay-Jagannathan: do we need both OR conditions ? can we just use `!shareOption.pubkey`

sdoshibitgo: yes will work. removed.

FloBitGo (approved):

kaustubhbitgo (approved):

lokesh-bitgo (approved):

PR #6468, opened 2025-07-16
feat: optional walletLabel key card

bitgoAaron:
- the new UI will allow users to create keycards before specifying a wallet label, so we need to make this field optional & not render the label if its not specified

Ticket: CE-6422

### DIFF
--- a/modules/key-card/src/drawKeycard.ts
+++ b/modules/key-card/src/drawKeycard.ts
@@ -149,12 +149,20 @@ export async function drawKeycard({
   const date = new Date().toDateString();
   y = moveDown(y, margin);
   doc.setFontSize(font.body).setTextColor(color.gray);
-  const title = curve ? KeyCurveName[curve] + ' key:' : 'wallet named:';
-  doc.text('Created on ' + date + ' for ' + title, left(0), y);
-  // copy
-  y = moveDown(y, 25);
-  doc.setFontSize(font.subheader).setTextColor(color.black);
-  doc.text(walletLabel, left(0), y);
+
+  if (curve || walletLabel) {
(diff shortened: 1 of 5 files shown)


modules/key-card/src/drawKeycard.ts
```diff
@@ -149,12 +149,20 @@ export async function drawKeycard({
...
   doc.setFontSize(font.body).setTextColor(color.gray);
-  const title = curve ? KeyCurveName[curve] + ' key:' : 'wallet named:';
-  doc.text('Created on ' + date + ' for ' + title, left(0), y);
-  // copy
-  y = moveDown(y, 25);
-  doc.setFontSize(font.subheader).setTextColor(color.black);
-  doc.text(walletLabel, left(0), y);
+
+  if (curve || walletLabel) {
+    const title = curve ? KeyCurveName[curve] + ' key:' : 'wallet named:';
```
sachushaji, line 154: Will it makes sense to use wallet id instead of curve for the title?  

Actually I don't see a walletID in the keycard which I guess is important though, how can somebody identify the wallet with just user key and wallet label is something which can change down the line

sachushaji (approved):

modules/key-card/src/drawKeycard.ts
bitgoAaron: we generate the keycard _before_ we create the wallet, and in the new UI we generate the card before we even prompt for a wallet name. Client can name the file anything they want at anytime

jamesmili (approved):

Samyakk123 (approved):

mohammadalfaiyazbitgo (approved):

PR #6511, opened 2025-07-22
feat(root): enable passing apiKey for recovery on eth likes

mohammadalfaiyazbitgo:
Ticket: WP-5348

### DIFF
--- a/examples/ts/eth/recover-eth.ts
+++ b/examples/ts/eth/recover-eth.ts
@@ -1,0 +1,47 @@
+/**
+ * Recover ETH from a multi-sig wallet
+ *
+ * This script demonstrates how to generate an unsigned recovery transaction for a
+ * Multi-Sig ETH wallet.
+ */
+
+import { BitGo } from 'bitgo';
+import { Hteth } from '@bitgo/sdk-coin-eth';
+
+// Set up BitGo object
+const bitgo = new BitGo({ env: 'test' }); // change to 'prod' for mainnet
+bitgo.register('hteth', Hteth.createInstance);
+
+async function recoverEth() {
+  try {
+    // Use the recovery with API key for output
+    const recovery = (await (bitgo.coin('hteth') as Hteth).recover({
+      userKey: 'user-public-key',
+      backupKey: 'backup-public-key',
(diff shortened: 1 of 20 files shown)


margueriteblair (commented):
Remove `TICKET: [WP-000000](https://bitgoinc.atlassian.net/browse/WP-000000)` from PR desc?

margueriteblair (dismissed):

pranavjain97 (dismissed):

margueriteblair (dismissed):

pranavjain97 (dismissed):

examples/ts/eth/recover-eth.ts
```diff
@@ -0,0 +1,47 @@
...
+async function recoverEth() {
+  try {
+    // Use the recovery with API key for output
+    const recovery = (await (bitgo.coin('hteth') as Hteth).recover({
+      userKey: 'user-public-key',
+      backupKey: 'backup-public-key',
+      walletContractAddress: 'Address-of-your-multisig-wallet',
+      recoveryDestination: 'Address-To-Recover-Funds-To',
+      isUnsignedSweep: true,
+      apiKey: 'Add Your Etherscan ApiKey here',
```
mullapudipruthvik, line 24: may I know why are we adding `apikey` here instead of  doing`new BitGoAPI({ env: environment, etherscanApiToken: apiKey })`;

mohammadalfaiyazbitgo: The motivation is to make recoveries have a similar interface across coins. If you look at UTXO recoveries, https://github.com/BitGo/BitGoJS/blob/7f0604a692b77aa1acdfd864e38647feceb8b659/modules/abstract-utxo/src/recovery/crossChainRecovery.ts#L397 they expose an API key param as well. Which to me makes sense as you only need the api key at time of recovery, which may not be when the bitgo object is initiated, so this flow is more flexible when it comes to usage.

zahin-mohammad: This is relevant for applications like WRW

margueriteblair (approved):

Taseen08 (approved):

PR #6517, opened 2025-07-23
feat: add script for fetching TRX balance that can be consolidated

Vijay-Jagannathan:
Ticket: COIN-00000

### DIFF
--- a/examples/ts/trx/fetch-addresses-with-trx-balance-for-consolidation.ts
+++ b/examples/ts/trx/fetch-addresses-with-trx-balance-for-consolidation.ts
@@ -1,0 +1,51 @@
+/**
+ * Copyright 2024, BitGo, Inc.  All Rights Reserved.
+ */
+import { BitGoAPI } from '@bitgo/sdk-api';
+import { Trx } from '@bitgo/sdk-coin-trx';
+import * as fs from 'fs';
+require('dotenv').config({ path: '../../.env' });
+
+const bitgo = new BitGoAPI({
+  accessToken: '',
+  env: 'prod',
+});
+
+const coin = 'trx';
+bitgo.register(coin, Trx.createInstance);
+
+const walletId = '';
+
+async function main() {
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  let prevId = undefined;
+  let index = 1;
+
+  do {
+    const addresses = await wallet.addresses({ includeBalances: true, prevId });
+    prevId = addresses.nextBatchPrevId;
+
+    for (const { address, balance, needsConsolidation, tokenConsolidationState } of addresses.addresses) {
+      const { tokens, spendableBalanceString } = balance;
+      const usdtBalance = tokens['trx:usdt']?.spendableBalanceString;
+
+      console.log(`Address ${index}: ${address}`);
+      console.log('USDT token balance: ', usdtBalance || 'Does not have USDT balance');
+      console.log('TRX balance: ', spendableBalanceString);
+
+      if (usdtBalance > 0 && spendableBalanceString > 36000000) {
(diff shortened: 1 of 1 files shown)


examples/ts/trx/fetch-addresses-with-trx-balance-for-consolidation.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright 2024, BitGo, Inc.  All Rights Reserved.
```
gianchandania, line 2: nit: 2025

examples/ts/trx/fetch-addresses-with-trx-balance-for-consolidation.ts
```diff
@@ -0,0 +1,51 @@
...
+    for (const { address, balance, needsConsolidation, tokenConsolidationState } of addresses.addresses) {
+      const { tokens, spendableBalanceString } = balance;
+      const usdtBalance = tokens['trx:usdt']?.spendableBalanceString;
+
+      console.log(`Address ${index}: ${address}`);
+      console.log('USDT token balance: ', usdtBalance || 'Does not have USDT balance');
+      console.log('TRX balance: ', spendableBalanceString);
+
+      if (usdtBalance > 0 && spendableBalanceString > 36000000) {
+        const data = `${address}, USDT balance: ${usdtBalance}, TRX balance: ${spendableBalanceString}, V1: ${needsConsolidation}, V2: ${tokenConsolidationState['trx']}\n`;
```
gianchandania, line 37: V2: ${tokenConsolidationState['trx']
why console v2 flag related to trx and not token, trying to understand the reason

examples/ts/trx/fetch-addresses-with-trx-balance-for-consolidation.ts
```diff
@@ -0,0 +1,51 @@
...
+        fs.appendFileSync('addresses-with-usdt-balance-and-trx-greater-than-36.txt', data);
+      }
+
+      if ((!usdtBalance || usdtBalance <= 0) && spendableBalanceString > 1000000) {
```
gianchandania, line 41: what list of addresses are you trying to get here? Should we check addresses having no tokens and spendable greater than 1000000?

examples/ts/trx/fetch-addresses-with-trx-balance-for-consolidation.ts
Vijay-Jagannathan: we're only printing the flag states for TRX because we're finding the TRX balance that can be consolidated

if incase there's some TRX that can be consolidated but V2 flag is false, we need to reset, hence TRX

examples/ts/trx/fetch-addresses-with-trx-balance-for-consolidation.ts
Vijay-Jagannathan: this is for addresses which don't have USDT (this code is specific to a client usecase who had only USDT and TRX, hence we only have these 2 conditions, but can be adapted to other tokens or multiple tokens as well) but TRX

we want this because, we've already checked the addresses which have both USDT and TRX prior

examples/ts/trx/fetch-addresses-with-trx-balance-for-consolidation.ts
gianchandania: Ok, my only concern is that if client is having some other supported token, then this won't be able to fetch the correct addresses as more amount will be blocked

gianchandania (approved):

examples/ts/trx/fetch-addresses-with-trx-balance-for-consolidation.ts
Vijay-Jagannathan: yep, this was written for a specific case, but we can tweak this to work on any token; can do that as a followup

PR #6522, opened 2025-07-23
feat(sdk-coin-cosmos): add cosmos shared coin functionalities

baltiyal:
TICKET: COIN-4881

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8692,7 +8692,7 @@ component-emitter@1.2.1:
   resolved "https://registry.npmjs.org/component-emitter/-/component-emitter-1.2.1.tgz#137918d6d78283f7df7a6b7c5a63e140e69425e6"
   integrity sha512-jPatnhd33viNplKjqXKRkGU345p263OIWzDL2wH3LGIGp5Kojo+uXizHmOADRvhGFFTnJqX3jBAKP6vvmSDKcA==
 
-component-emitter@^1.2.0, component-emitter@^1.3.0, component-emitter@~1.3.0:
+component-emitter@^1.2.0, component-emitter@^1.3.0, component-emitter@^1.3.1, component-emitter@~1.3.0:
   version "1.3.1"
   resolved "https://registry.npmjs.org/component-emitter/-/component-emitter-1.3.1.tgz#ef1d5796f7d93f135ee6fb684340b26403c97d17"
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-cosmos/src/lib/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+    );
+  }
+
+  /** @inheritdoc */
+  isValidContractAddress(address: string): boolean {
+    return this.isValidBech32AddressMatchingRegex(
+      address,
+      new RegExp(`^${this._network.addressPrefix}1['qpzry9x8gf2tvdw0s3jn54khce6mua7l]+$`)
+    );
+  }
```
MohammedRyaan786: do we need this function?

modules/sdk-coin-cosmos/src/cosmosSharedCoin.ts
```diff
@@ -1,16 +1,90 @@
...
+    return utils.isValidAddress(address) || utils.isValidValidatorAddress(address);
+  }
+
+  /** @inheritDoc **/
+  getDenomination(): string {
+    return this._network.denom;
+  }
+
+  /** @inheritDoc **/
+  getGasAmountDetails(): { gasAmount: string; gasLimit: number } {
```
MohammedRyaan786: nit: we can have an interface defining these properties

modules/sdk-coin-cosmos/src/lib/utils.ts
baltiyal: This is for validation of contract address, it is present for even existing cosmos coins.

modules/sdk-coin-cosmos/src/cosmosSharedCoin.ts
baltiyal: thanks, I'll do that.

yarn.lock
```diff
@@ -6801,14 +6801,6 @@ ansi-styles@^6.0.0, ansi-styles@^6.1.0:
   resolved "https://registry.npmjs.org/ansi-styles/-/ansi-styles-6.2.1.tgz#0e62320cf99c21afff3b3012192546aacbfb05c5"
```
bhavidhingra: how come so many deletions? is it normal?

baltiyal: tbh, not sure.
Cleaned the present `yarn-lock` and made it again using `yarn install` and got similar or same result

PR #6523, opened 2025-07-23
feat: add cosmos chain token standard support

MohammedRyaan786:
TICKET: COIN-4915

### DIFF
--- a/modules/abstract-cosmos/src/cosmosToken.ts
+++ b/modules/abstract-cosmos/src/cosmosToken.ts
@@ -1,0 +1,63 @@
+import { BitGoBase, CoinConstructor, NamedCoinConstructor } from '@bitgo/sdk-core';
+import { CosmosTokenConfig, coins, tokens } from '@bitgo/statics';
+import { CosmosCoin } from './cosmosCoin';
+
+export class CosmosToken extends CosmosCoin {
+  public readonly tokenConfig: CosmosTokenConfig;
+
+  constructor(bitgo: BitGoBase, tokenConfig: CosmosTokenConfig) {
+    const staticsCoin = coins.get(tokenConfig.coin);
+    super(bitgo, staticsCoin);
+    this.tokenConfig = tokenConfig;
+  }
+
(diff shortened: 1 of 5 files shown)


modules/abstract-cosmos/src/cosmosToken.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  getBaseChain(): string {
+    return this.coin;
+  }
+
+  getFullName(): string {
+    if (this.tokenConfig.coin) {
+      // Eg - returns "Atom Token", "Osmo Token"
+      return `${this.tokenConfig.coin.charAt(0).toUpperCase()}${this.tokenConfig.coin.slice(1)} Token`;
+    } else {
```
baltiyal: I think this else should be removed.

modules/abstract-cosmos/src/register.ts
```diff
@@ -0,0 +1,8 @@
+import { CosmosToken } from './cosmosToken';
+import { BitGoBase } from '@bitgo/sdk-core';
+
+export const register = (sdk: BitGoBase): void => {
```
baltiyal, line 4: could you explain where registration will happen?
Coz I can't see this PR too: https://github.com/BitGo/BitGoJS/pull/6525

MohammedRyaan786: Register is only for the tokens, not for a particular coin since its cosmos. It's similar to how we register tokens on other chains

modules/abstract-cosmos/src/cosmosToken.ts
```diff
@@ -0,0 +1,63 @@
+import { CosmosCoin } from './cosmosCoin';
+import { BitGoBase, CoinConstructor, NamedCoinConstructor } from '@bitgo/sdk-core';
+import { CosmosTokenConfig, coins, tokens } from '@bitgo/statics';
```
ravibitgo, line 2: external imports should be on top

MohammedRyaan786: corrected it

baltiyal (approved):
looks good

PR #6529, opened 2025-07-23
fix: run test in /dist to avoid strip-only mode for node 24

mhahn2003:
TICKET: WP-5182

### DIFF
--- a/modules/utxo-core/package.json
+++ b/modules/utxo-core/package.json
@@ -27,7 +27,7 @@
     "lint": "eslint --quiet .",
     "prepare": "npm run build",
     "test": "npm run unit-test",
-    "unit-test": "mocha --recursive test/"
+    "unit-test": "mocha --recursive \"dist/test/**/*.js\""
   },
   "author": "BitGo SDK Team <sdkteam@bitgo.com>",
   "license": "MIT",
@@ -45,6 +45,7 @@
   "publishConfig": {
     "access": "public"
   },
+  "type": "commonjs",
   "nyc": {
     "extension": [
       ".ts"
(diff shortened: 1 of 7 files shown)


modules/utxo-core/package.json
```diff
@@ -20,14 +20,14 @@
     "./testutil/descriptor": "./dist/src/testutil/descriptor/index.js"
   },
   "scripts": {
-    "build": "yarn tsc --build --incremental --verbose .",
+    "build": "yarn tsc --build --incremental --verbose . && cpx \"test/descriptor/psbt/fixtures/**/*\" dist/test/descriptor/psbt/fixtures",
```
zahin-mohammad: This will cause our deployed bundle to include these fixture files, which doesn't seem right

modules/utxo-staking/package.json
```diff
@@ -8,14 +8,14 @@
     "dist/src"
   ],
   "scripts": {
-    "build": "yarn tsc --build --incremental --verbose .",
+    "build": "yarn tsc --build --incremental --verbose . && cpx \"test/fixtures/**/*\" dist/test/fixtures",
```
zahin-mohammad: Same here

modules/utxo-core/package.json
mhahn2003: Without copying the fixture files I get this error
```
  1) createPsbt Wsh2Of3
       creates psbt with expected properties:
     Error: fixture directory /Users/michaelhahn/bitgo/BitGoJS/modules/utxo-core/dist/test/descriptor/psbt/fixtures not found, please create it first
      at getFixture (src/testutil/fixtures.utils.ts:69:13)
      at async assertEqualsFixture (test/descriptor/psbt/psbt.ts:50:30)
      at async Context.<anonymous> (test/descriptor/psbt/psbt.ts:134:7)
 ```

mhahn2003: I'm guessing the issue is most likely relative paths now that the tests are running in `/dist/test` instead of `/test`, I've had to resolve other similar issues to this, I'll look for a fix

lcovar (approved):

PR #6541, opened 2025-07-24
feat(statics): use const instead of enum

mhahn2003:
BREAKING CHANGE: enums used in `statics/src/base` are changed to const, both the object and type are still exported but cannot rely on them being enums

TICKET: CAAS-325


mullapudipruthvik (commented):
can you explain why are you doing this change?
and what is the client expected to change on their end ?

OttoAllmendinger (changes requested):
I'm still not convinced we need this. I'm sure we can run this typescript code with node24.

PR #6547, opened 2025-07-25
feat: update hoodi testnet contracts

sachushaji:
this change updates the hteth usd1
and gousd contracts to use hoodi
testnet addresses.

Ticket: TMS-1191

### DIFF
--- a/modules/statics/src/coins/erc20Coins.ts
+++ b/modules/statics/src/coins/erc20Coins.ts
@@ -11438,26 +11438,26 @@ export const erc20Coins = [
   terc20(
     '17377290-40ff-47a3-8346-ba03885f7cc3',
     'tgousd',
-    'Holesky Testnet GoUSD',
+    'Hoodi Testnet GoUSD',
     6,
-    '0x51cef49e20d11a2735d09e3bb68d79ec490e3579',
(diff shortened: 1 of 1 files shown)


modules/statics/src/coins/erc20Coins.ts
```diff
@@ -11440,7 +11440,7 @@ export const erc20Coins = [
     'tgousd',
     'Holesky Testnet GoUSD',
     6,
-    '0x51cef49e20d11a2735d09e3bb68d79ec490e3579',
+    '0x32275b636171ba98fec28e86ed5f0289bd42aeb9',
```
shivam-bitgo, line 11443: we need to update the network as well from holesky to hoodi?

kamleshmugdiya: +1

sachushaji: Good Catch, Updated

mukeshsp (approved):

shivam-bitgo (approved):
Let's merge it on wednesday when we are planning to merge other changes for hoodi on test as if we merge this today then these tokens will stop working in staging/test if sdk is bumped in other repos?

Phani024 (approved):
LGTM

PR #6550, opened 2025-07-25
feat(abstract-eth): add recover consolidation for eth

abhi-bitgo:
Adding the recover consolidation functionality for `eth` and `hteth` which would be used for unsigned consolidation in WRW 

ticket: WIN-5700

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -19,6 +19,7 @@ import {
   IWallet,
   KeyPair,
   MPCSweepRecoveryOptions,
+  MPCSweepTxs,
   MPCTx,
   MPCTxs,
   ParsedTransaction,
@@ -57,7 +58,7 @@ import { BigNumber } from 'bignumber.js';
 import BN from 'bn.js';
 import { randomBytes } from 'crypto';
 import debugLib from 'debug';
-import { addHexPrefix, bufArrToArr, stripHexPrefix } from 'ethereumjs-util';
+import { addHexPrefix, bufArrToArr, stripHexPrefix, bufferToHex, setLengthLeft, toBuffer } from 'ethereumjs-util';
 import Keccak from 'keccak';
 import _ from 'lodash';
 import secp256k1 from 'secp256k1';
@@ -70,6 +71,7 @@ import {
   ERC721TransferBuilder,
   getBufferedByteCode,
   getCommon,
+  getCreateForwarderParamsAndTypes,
   getProxyInitcode,
   getRawDecoded,
   getToken,
@@ -224,6 +226,11 @@ export type UnsignedSweepTxMPCv2 = {
   }[];
 };
 
+export type UnsignedBuilConsolidation = {
+  transactions: MPCSweepTxs[] | UnsignedSweepTxMPCv2[] | RecoveryInfo[] | OfflineVaultTxInfo[];
+  lastScanIndex: number;
+};
+
 export type RecoverOptionsWithBytes = {
   isTss: true;
   /**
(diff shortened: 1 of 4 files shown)


modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -1192,6 +1221,160 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+        );
+        possibleConsolidationAddresses.push(forwarderAddress);
+      } catch (e) {
+        console.log(`Failed to generate forwarder address: ${e.message}`);
+      }
+    }
+
+    if (params.userKey) {
+      try {
+        const derivedAddress = this.deriveAddressFromPublicKey(params.userKey, index);
```
mullapudipruthvik, line 1289: derivedAddress is returned for every `getConsolidationAddress` call, why is that ?

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -1192,6 +1221,160 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+    baseAddress: string,
+    feeAddress: string,
+    forwarderFactoryAddress: string,
+    forwarderImplementationAddress: string,
+    index: number
+  ): string {
+    const salt = addHexPrefix(index.toString(16));
+    const saltBuffer = setLengthLeft(toBuffer(salt), 32);
+
+    const { createForwarderParams, createForwarderTypes } = getCreateForwarderParamsAndTypes(
```
mullapudipruthvik, line 1239: this is specific for v4 wallet, can you not make it generic ?

mullapudipruthvik (changes requested):
also please add test cases

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
abhi-bitgo: We don't have any means to differentiate between v6 and v5 wallet thus we try to derive address every time.

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
abhi-bitgo: Since we are not sure which version is being used, we have kept this in a try statement in order to handle the failure

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
mullapudipruthvik: in that case this would execute for v6 wallet as well right ?
`        const forwarderAddress = this.generateForwarderAddress(
                                                     params.walletContractAddress,
          params.bitgoFeeAddress,
          forwarderFactoryAddress,
          forwarderImplementationAddress,
          index
        );
        possibleConsolidationAddresses.push(forwarderAddress);`

modules/sdk-coin-eth/test/unit/eth.ts
```diff
@@ -1053,6 +1055,113 @@ describe('ETH:', function () {
...
+      const baseUrl = common.Environments.test.etherscanBaseUrl as string;
+      let bitgo: TestBitGoAPI;
+      let basecoin: Hteth;
+
+      before(function () {
+        bitgo = TestBitGo.decorate(BitGoAPI, { env: 'test' });
+        basecoin = bitgo.coin('hteth') as Hteth;
+      });
+
+      it('should generate an unsigned consolidation', async function () {
```
mullapudipruthvik: add a test case for v5 and v6 separately

abhi-bitgo: We don't need to write separate test cases since all the values are provided and based on the which wallet version we are assuming, we are using the relevant values and generating address for consolidation

abhi-bitgo: > in that case this would execute for v6 wallet as well right ?
>         const forwarderAddress = this.generateForwarderAddress( params.walletContractAddress, params.bitgoFeeAddress, forwarderFactoryAddress, forwarderImplementationAddress, index ); possibleConsolidationAddresses.push(forwarderAddress);

yes, this will run in both the cases since we have no means of verifying which wallet version is this

mullapudipruthvik: why are you sure whether this would throw an error for V6 wallet?
better to add a testcase to verify

abhi-bitgo: Added the separate test cases

mullapudipruthvik (approved):

PR #6579, opened 2025-07-29
feat: onboard new ofc testnet hoodi coin

manav-c:
Ticket: GO-1488. This is being done so that we can essentially "reset" the underlying data for HTETH without actually wiping anything. Essentially the old data will have ofchteth on it and after migration, HTETH will point to "ofchooditeth" giving us a clean slate.

### DIFF
--- a/modules/statics/src/coins/ofcCoins.ts
+++ b/modules/statics/src/coins/ofcCoins.ts
@@ -270,6 +270,14 @@ export const ofcCoins = [
     CoinKind.CRYPTO
   ),
   tofc(
+    '97f83379-a536-4977-a5e8-5a4dae5a3db7',
+    'ofchooditeth',
+    'Test Hoodi Ether',
+    18,
+    UnderlyingAsset.ETH,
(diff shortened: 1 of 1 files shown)


mukeshsp (approved):

modules/statics/src/coins/ofcCoins.ts
```diff
@@ -269,6 +269,14 @@ export const ofcCoins = [
     UnderlyingAsset.ETH,
     CoinKind.CRYPTO
   ),
+  tofc(
+    '97f83379-a536-4977-a5e8-5a4dae5a3db7',
+    'ofchooditeth',
```
shivam-bitgo, line 274: any reason why we are not keeping `ofchteth` for hoodi?

Phani024: +1. Why hooditeth instead of hteth?

manav-c: look at PR description

PR #6583, opened 2025-07-30
feat: add spl token operations builder for sol

raj-bitgo:
- Add mint/burn transaction builder for sol
- Could support more instructions in the future

TICKET: TMS-1218

### DIFF
--- a/modules/sdk-coin-sol/src/lib/iface.ts
+++ b/modules/sdk-coin-sol/src/lib/iface.ts
@@ -40,8 +40,7 @@ export type InstructionParams =
   | StakingAuthorize
   | StakingDelegate
   | MintTo
-  | Burn
-  | SplTokenOps;
+  | Burn;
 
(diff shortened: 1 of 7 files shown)


modules/sdk-coin-sol/src/lib/iface.ts
```diff
@@ -106,6 +107,28 @@ export interface Burn {
...
+  tokenName?: string;
+  decimalPlaces?: number;
+  programId?: string;
+  // For mint operations
+  destinationAddress?: string;
+  // For burn operations
+  accountAddress?: string;
+  // Authority address for both mint and burn
+  authorityAddress: string;
+}
```
sachushaji: Instead of having a combined SplTokenOperation, can we use a union of supported SPL operations,


```suggestion
export type SplTokenOperation = Mint | Burn
```

PR #6597, opened 2025-07-31
feat: add custom instruction tx builder for sol

raj-bitgo:
- added custom instruction for sol transactions

TICKET: TMS-1232


modules/sdk-coin-sol/src/lib/constants.ts
```diff
@@ -49,6 +49,7 @@ export enum InstructionBuilderTypes {
   SetPriorityFee = 'SetPriorityFee',
   MintTo = 'MintTo',
   Burn = 'Burn',
+  CustomInstruction = 'CustomInstruction',
```
sachushaji, line 52: Lets remove Burn / minto, we can follow up

sachushaji (approved):

PR #6606, opened 2025-08-01
feat: added custom tx builder for sol

raj-bitgo:
- added custom instruction builder for sol
- use new intent type solInstruction to build the tx in WP

TICKET: TMS-1218

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -2780,6 +2780,285 @@ describe('V2 Wallet:', function () {
         });
       });
 
+      it('should build a custom instruction transaction', async function () {
+        const recipients = [
+          {
+            address: '6DadkZcx9JZgeQUDbHh12cmqCpaqehmVxv6sGy49jrah',
+            amount: '1000',
+          },
+        ];
+        const solInstructions = [
+          {
+            programId: '11111111111111111111111111111112',
+            keys: [
+              {
+                pubkey: '6DadkZcx9JZgeQUDbHh12cmqCpaqehmVxv6sGy49jrah',
+                isSigner: true,
+                isWritable: true,
+              },
+            ],
+            data: 'SGVsbG8gV29ybGQ=',
+          },
+        ];
+
+        const prebuildTxWithIntent = sandbox.stub(TssUtils.prototype, 'prebuildTxWithIntent');
+        prebuildTxWithIntent.resolves(txRequest);
+        prebuildTxWithIntent.calledOnceWithExactly({
+          reqId,
+          intentType: 'customTx',
+          solInstructions,
+          recipients,
+        });
+
+        const txPrebuild = await tssSolWallet.prebuildTransaction({
+          reqId,
+          recipients,
+          type: 'customTx',
+          solInstructions,
+        });
+
+        txPrebuild.should.deepEqual({
+          walletId: tssSolWallet.id(),
+          wallet: tssSolWallet,
+          txRequestId: 'id',
+          txHex: 'ababcdcd',
+          buildParams: {
+            recipients,
+            type: 'customTx',
+            solInstructions,
+          },
+          feeInfo: {
+            fee: 5000,
+            feeString: '5000',
+          },
+        });
+      });
+
+      it('should build a custom instruction transaction with multiple instructions', async function () {
+        const recipients = [];
+        const solInstructions = [
+          {
+            programId: '11111111111111111111111111111112',
+            keys: [
+              {
+                pubkey: '6DadkZcx9JZgeQUDbHh12cmqCpaqehmVxv6sGy49jrah',
+                isSigner: true,
+                isWritable: true,
+              },
+            ],
+            data: 'SGVsbG8gV29ybGQ=',
+          },
+          {
+            programId: 'ComputeBudget111111111111111111111111111111',
+            keys: [],
+            data: 'AwAA6AMAAAAA',
+          },
+        ];
+
+        const prebuildTxWithIntent = sandbox.stub(TssUtils.prototype, 'prebuildTxWithIntent');
+        prebuildTxWithIntent.resolves(txRequest);
+        prebuildTxWithIntent.calledOnceWithExactly({
+          reqId,
+          intentType: 'customTx',
+          solInstructions,
+          recipients,
+        });
+
+        const txPrebuild = await tssSolWallet.prebuildTransaction({
+          reqId,
+          recipients,
+          type: 'customTx',
+          solInstructions,
+        });
+
+        txPrebuild.should.deepEqual({
+          walletId: tssSolWallet.id(),
+          wallet: tssSolWallet,
+          txRequestId: 'id',
+          txHex: 'ababcdcd',
+          buildParams: {
+            recipients,
+            type: 'customTx',
+            solInstructions,
+          },
+          feeInfo: {
+            fee: 5000,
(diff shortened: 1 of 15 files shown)


zahin-mohammad (changes requested):
Some concerns here, moving this to slack

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -5100,4 +5379,349 @@ describe('V2 Wallet:', function () {
...
+        const prebuildTxWithIntent = sandbox.stub(TssUtils.prototype, 'prebuildTxWithIntent');
+        prebuildTxWithIntent.resolves(txRequest);
+        prebuildTxWithIntent.calledOnceWithExactly({
+          reqId,
+          intentType: 'solInstruction',
+          solInstructions: [],
+          recipients: testRecipients,
+        });
+
+        const txPrebuild = await tssSolWallet.prebuildTransaction({
```
sachushaji: We should throw an error for this case?

modules/sdk-coin-sol/src/lib/solInstructionFactory.ts
```diff
@@ -551,15 +551,55 @@ function burnInstruction(data: Burn): TransactionInstruction[] {
...
+      return decoded === str;
+    } catch {
+      return false;
+    }
+  };
+
+  // Check if the data is valid hex
+  const isValidHex = (str: string): boolean => {
+    return /^[0-9A-Fa-f]*$/.test(str) && str.length % 2 === 0;
+  };
```
sachushaji: We can move this utilities in a dedicated util file?

modules/bitgo/test/v2/unit/wallet.ts
raj-bitgo: Yeah, good catch. Added an assertion check

modules/sdk-coin-sol/src/lib/solInstructionFactory.ts
raj-bitgo: moved

sachushaji (dismissed):

modules/sdk-coin-sol/src/lib/utils.ts
```diff
@@ -147,6 +147,29 @@ export function isValidMemo(memo: string): boolean {
...
+  }
+}
+
+/**
+ * Checks if a string is valid hexadecimal data
+ * @param str - The string to validate
+ * @returns True if the string is valid hex, false otherwise
+ */
+export function isValidHex(str: string): boolean {
+  return /^[0-9A-Fa-f]*$/.test(str) && str.length % 2 === 0;
```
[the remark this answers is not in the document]
Phani024: 👍

modules/sdk-core/src/bitgo/wallet/BuildParams.ts
```diff
@@ -104,6 +104,8 @@ export const BuildParams = t.exact(
       // param to set emergency flag on a custodial transaction.
       // This transaction should be performed in less than 1 hour or it will fail.
       emergency: t.unknown,
+      // Solana custom instructions for transaction building
+      solInstructions: t.unknown,
```
zahin-mohammad, line 108: Why is this unknown? You clearly have a type defined for this.

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -257,6 +270,19 @@ export interface PopulatedIntent extends PopulatedIntentBase {
...
+   * Each instruction should contain programId, keys, and data fields.
+   */
+  solInstructions?: {
+    programId: string;
+    keys: Array<{
+      pubkey: string;
+      isSigner: boolean;
+      isWritable: boolean;
+    }>;
+    data: string;
```
zahin-mohammad: This pr repeats this type in-line at least 3 different times. Can we define this somewhere?

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
```diff
@@ -117,7 +117,16 @@ export abstract class MpcUtils {
   populateIntent(baseCoin: IBaseCoin, params: PrebuildTransactionWithIntentOptions): PopulatedIntent {
     const chain = this.baseCoin.getChain();
 
-    if (!['acceleration', 'fillNonce', 'transferToken', 'tokenApproval'].includes(params.intentType)) {
+    if (params.intentType === 'solInstruction') {
+      assert(
+        params.solInstructions && params.solInstructions.length > 0,
+        `'solInstructions' is a required parameter for solInstruction intent`
+      );
+    }
```
zahin-mohammad: Why is the intent type `solInstruction`? I think it would be better to do something like 
`if (params.intentType === 'customTx' && chain === 'sol')`

The concept of a custom tx is not unique to SOL, [other chains have this as well](https://github.com/BitGo/public-types/blob/392e88bb9421afcb2868b9697d968f64f163711f/src/schema/transactionRequest/intents/intentType.ts).

zahin-mohammad (changes requested):

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
raj-bitgo: Hmm, I might have to add this chain check at other places as well then. Even in WP, does that work?

modules/sdk-core/src/bitgo/wallet/BuildParams.ts
raj-bitgo: all the variables defined in this type are `unknown`.

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
raj-bitgo: updated

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
raj-bitgo: updated

modules/sdk-coin-sol/src/lib/solInstructionFactory.ts
```diff
@@ -593,15 +593,40 @@ function burnInstruction(data: Burn): TransactionInstruction[] {
...
+  // Convert string data to Buffer
+  let dataBuffer: Buffer;
+
+  if (isValidBase64(params.data)) {
+    dataBuffer = Buffer.from(params.data, 'base64');
+  } else if (isValidHex(params.data)) {
+    dataBuffer = Buffer.from(params.data, 'hex');
+  } else {
+    // Fallback to UTF-8
+    dataBuffer = Buffer.from(params.data, 'utf8');
```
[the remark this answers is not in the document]
zahin-mohammad: @raj-bitgo please address this comment as a won't fix with a reason or with a code change to address the issue.

zahin-mohammad (approved):

modules/sdk-coin-sol/src/lib/solInstructionFactory.ts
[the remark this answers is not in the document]
raj-bitgo: We need to check if there's a standard to accept the format in for sol instructions. Will address this in a follow up rather getting the type in an argument as suggested by copilot. Right now since it'll be gated behind a feature flag and will be for internal use, keeping it the same.

dpkjnr (approved):
LGTM

PR #6615, opened 2025-08-04
feat: onboarding BitMart Token (BMX)

rohitsaw115:
TICKET: WIN-6409


prithvishet2503 (approved):
LGTM

modules/statics/src/coins/erc20Coins.ts
```diff
@@ -11024,12 +11024,12 @@ export const erc20Coins = [
     UnderlyingAsset['eth:vvs']
   ),
   erc20(
-    '395f00e4-aaf8-4a4b-8fa1-54dfd7543951',
```
ankur-mazumder, line 11027: we don't need to change the id here

PR #6618, opened 2025-08-04
feat(sdk-coin-ton): add jetton skeleton

Chinmay-Sharma1:
Ticket: WIN-3198

TICKET: WIN-3198


modules/statics/src/account.ts
```diff
@@ -630,6 +634,17 @@ export class Nep141Token extends AccountCoinToken {
   }
 }
 
+export class JettonToken extends AccountCoinToken {
+  public contractAddress: string;
```
bhavidhingra: please check whether `contractAddress` is the terminology used in Ton?

modules/sdk-coin-ton/src/jettonToken.ts
```diff
@@ -0,0 +1,105 @@
+import BigNumber from 'bignumber.js';
+import { BitGoBase, CoinConstructor, NamedCoinConstructor, VerifyTransactionOptions } from '@bitgo/sdk-core';
+import { coins, JettonTokenConfig, NetworkType, tokens } from '@bitgo/statics';
+
+import { Transaction } from './lib';
+import { Ton } from './ton';
+
+export class JettonToken extends Ton {
```
bhavidhingra: should we name it JettonToken or TonToken?

modules/statics/src/account.ts
Chinmay-Sharma1: We don't use contractAddress in ton, We use Jetton Master contract. I will replace the contractAddress with the Jetton master.

modules/sdk-coin-ton/src/jettonToken.ts
Chinmay-Sharma1: I think TonToken would be the appropriate name, I will make the necessary changes, thanks.

modules/sdk-coin-ton/src/tonToken.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  getFullName(): string {
+    return 'TON Token';
+  }
+
+  getBaseFactor(): number {
+    return Math.pow(10, this.tokenConfig.decimalPlaces);
+  }
+
+  async verifyTransaction(params: VerifyTransactionOptions): Promise<boolean> {
```
bhavidhingra, line 64: how will the verification work for different sub-transactions of a jetton transaction?

PR #6624, opened 2025-08-05
fix(express)!: generated xprv is permissioned only for owner

mohammadalfaiyazbitgo:
TICKET: WP-4201


modules/express/src/fetchEncryptedPrivKeys.ts
```diff
@@ -94,12 +94,19 @@ export async function fetchKeys(ids: WalletIds, token: string, accessToken?: str
...
   const data = JSON.stringify(output, null, '\t');
   const fileName = 'encryptedPrivKeys.json';
-  writeFile(fileName, data, (err) => {
-    if (err) {
-      throw err;
+  writeFile(
+    fileName,
+    data,
+    {
+      mode: 0o600,
```
0xPrabh, line 102: Could you also write comment on this? What this mode will do.

0xPrabh: By reading this code, devs will not be able to understand

prabh-codes (changes requested):

modules/express/src/fetchEncryptedPrivKeys.ts
mohammadalfaiyazbitgo: it's a standard permissions format, but will update.

zahin-mohammad (approved):

prabh-codes (approved):
LGTM

mohammadalfaiyazbitgo:
> Before merging please test express with this new change.

validated script still generates the appropriate json and be used to send out a transaction.

PR #6644, opened 2025-08-06
feat(root): use additonal headers on constants fetch

mohammadalfaiyazbitgo:
- Added code in `fetchConstants()` to check if `getAdditionalHeadersCb` is set
- When available, retrieves additional headers using the callback with appropriate parameters
- Applies any returned headers to the request before execution

TICKET: WP-000000


bitgoAaron (approved):
changes are fine. Dream state would be passing constants to the SDK ctor so we could avoid the fetch call entirely, but not a requirement

PR #6665, opened 2025-08-08
feat: automate ERC20 automation

rv-bitgo:
WIN-6598

TICKET: WIN-6598


modules/bitgo/src/v2/coinFactory.ts
```diff
@@ -532,6 +533,19 @@ export function registerCoinConstructors(coinFactory: CoinFactory, coinMap: Coin
   VetToken.createTokenConstructors().forEach(({ name, coinConstructor }) =>
     coinFactory.register(name, coinConstructor)
   );
+
+  // Generic ERC20 token registration for coins with SUPPORTS_ERC20 feature
+  coins
+    .filter((coin) => coin.features.includes(CoinFeature.SUPPORTS_ERC20))
+    .forEach((coin) => {
+      const coinNames = {
+        [coin.network.type]: coin.name,
```
rv-bitgo: would only populate for Mainnet

rv-bitgo: Updated.

modules/bitgo/src/v2/coinFactory.ts
```diff
@@ -532,6 +533,21 @@ export function registerCoinConstructors(coinFactory: CoinFactory, coinMap: Coin
...
   );
+
+  // Generic ERC20 token registration for coins with IS_ERC20_TOKEN feature
+  coins
+    .filter((coin) => coin.features.includes(CoinFeature.IS_ERC20_TOKEN))
+    .forEach((coin) => {
+      //TODO: Decide if it should stay here, maybe move to network or coin itself.
+      const coinNames = {
+        Mainnet: `${coin.network.name}`,
+        Testnet: `t${coin.network.name}`,
```
parasgarg-bitgo: Should this be network name or coin name?

modules/sdk-coin-evm/src/register.ts
```diff
@@ -20,6 +21,20 @@ export const register = (coinFamily: string, sdk: BitGoBase): void => {
...
+      .filter((coin) => coin.family === coinFamily && coin.isToken)
+      .forEach((coin) => {
+        //TODO: Decide if it should stay here, maybe move to network or coin itself.
+        const coinNames = {
+          Mainnet: `${coin.network.name}`,
+          Testnet: `t${coin.network.name}`,
+        };
+
+        EthLikeErc20Token.createTokenConstructors(coinNames).forEach(({ name, coinConstructor }) => {
+          sdk.register(name, coinConstructor);
```
parasgarg-bitgo: This'll end up calling createTokenConstructors multiple time. One for each token.
I don't think that's the expected behavior

modules/sdk-coin-evm/src/register.ts
```diff
@@ -20,6 +21,20 @@ export const register = (coinFamily: string, sdk: BitGoBase): void => {
       .forEach((coin) => {
         sdk.register(coin.name, EvmCoin.createInstance);
       });
-    //TODO: add token registration after EVM Token Optimisation is implemented
+  }
+  if (coins.get(coinFamily).features.includes(CoinFeature.IS_ERC20_TOKEN)) {
```
parasgarg-bitgo: If we want to add the feature to coin, then let's rework the feature name. 
Current feature name suggests the coin is token as well.

modules/statics/src/account.ts
```diff
@@ -431,6 +431,15 @@ export class AdaCoin extends AccountCoinToken {
   }
 }
 
+export class EthLikeERC20Token extends ContractAddressDefinedToken {
+  public readonly coinNames: { Mainnet: string; Testnet: string };
+
+  constructor(options: Erc20ConstructorOptions & { coinNames: { Mainnet: string; Testnet: string } }) {
```
parasgarg-bitgo: Why take this as constructor params. Evaluate if we can derive this from the Network

modules/bitgo/src/v2/coinFactory.ts
rv-bitgo: As discussed, would fetch this from coin.name now.

modules/sdk-coin-evm/src/register.ts
rv-bitgo: Idea was to use this flag with tokens instead of parent coin, hence the name `IS_ERC20_TOKEN`. I've now updated it to `SUPPORTS_ERC20`.

modules/sdk-coin-evm/src/register.ts
rv-bitgo: A valid catch. 👍 
Since it was implemented on tokens instead of parent coin, the loop would call all tokens and register all tokens for each individual tokens.
Since we've moved the approach to parent coin instead of tokens, this would be fine now.

modules/statics/src/account.ts
rv-bitgo: Removed from constructors and static properties. It'd be fetched from coin name now. As discussed, we would not make this part of the network property as it's the same as coin.name in all of our use cases.

modules/sdk-coin-evm/src/register.ts
```diff
@@ -14,12 +15,26 @@ export const registerAll = (sdk: BitGoBase): void => {
...
+  if (coinFeatures.includes(CoinFeature.SUPPORTS_ERC20)) {
+    coins
+      .filter((coin) => coin.family === coinFamily && !coin.isToken)
+      .forEach((coin) => {
+        const coinNames = {
+          Mainnet: `${coin.name}`,
+          Testnet: `${coin.name}`,
+        };
+
+        EthLikeErc20Token.createTokenConstructors(coinNames).forEach(({ name, coinConstructor }) => {
```
parasgarg-bitgo: Can this not be moved inside the previous loop?

rv-bitgo: Good idea. Lets put them in one loop.

parasgarg-bitgo (dismissed):

yash-bitgo (commented):
tests are failing

rv-bitgo:
> tests are failing

Fixed the browser test. 22.x is failing for all PRs(most likely due to mismatch versions).

parasgarg-bitgo (approved):

PR #6671, opened 2025-08-11
feat(express): migrate user.login to typed routes

kaustubhbitgo:
Ticket: WP-5396

### DIFF
--- a/modules/express/src/expressApp.ts
+++ b/modules/express/src/expressApp.ts
@@ -11,6 +11,7 @@ import * as morgan from 'morgan';
 import * as fs from 'fs';
 import type { Request as StaticRequest } from 'express-serve-static-core';
 import * as timeout from 'connect-timeout';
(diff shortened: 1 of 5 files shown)


modules/express/src/expressApp.ts
```diff
@@ -318,6 +319,8 @@ export function app(cfg: Config): express.Application {
   checkPreconditions(cfg);
   debug('preconditions satisfied');
 
+  app.use(bodyParser.json({ limit: '20mb' }));
```
adityasoni-bitgo, line 322: Is there any specific reason as to why we are setting this upper limit?

kaustubhbitgo: to enable passing body in express, it is a middleware so need to be enable before api-ts middlewares.

adityasoni-bitgo (approved):
LGTM!

PR #6677, opened 2025-08-11
chore: add new feature flag `SUPPORTS_ERC20` for chains that supports_erc20_token

rv-bitgo:
WIN-6598

TICKET: WIN-6598

Adding this flag for all chains that supports ERC20 tokens. This flag would be used to identify the coins that are supporting ERC20 tokens in BITGO system.

This PR is part of ERC20 automation EPIC and this flag's usage can be seen in this [PR](https://github.com/BitGo/BitGoJS/pull/6665/files)

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -398,6 +398,11 @@ export enum CoinFeature {
   SHARED_EVM_SDK = 'shared-evm-sdk',
 
   /**
+   * This coin supports erc20 tokens
+   */
+  SUPPORTS_ERC20 = 'supports-erc20-token',
+
+  /**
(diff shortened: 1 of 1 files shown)


modules/statics/src/base.ts
```diff
@@ -397,6 +397,11 @@ export enum CoinFeature {
    */
   SHARED_EVM_SDK = 'shared-evm-sdk',
 
+  /**
+   * This coin supports erc20 tokens
+   */
+  IS_ERC20_TOKEN = 'is-erc20-token',
```
mukeshsp: Why do we require this flag?

rv-bitgo: Updated the description

PR #6687, opened 2025-08-12
feat(statics): add RESTRICTED coin feature for internal tokens

saarujansathees578:
TICKET: SC-2604

This PR adds a new CoinFeature (RESTRICTED) which, when enabled, avoids exposing certain coins/tokens to the user. This is necessary for erc721:unsteth for LIDO, as it prevents UI and API users from withdrawing or seeing an unnecessary NFT, which can disrupt the unstaking flow.

### DIFF
--- a/modules/statics/src/base.ts
+++ b/modules/statics/src/base.ts
@@ -451,6 +451,11 @@ export enum CoinFeature {
    * This coin is an EVM compatible coin and should use common EVM logic in WP
    */
   EVM_COMPATIBLE_WP = 'evm-compatible-wp',
+
+  /**
+   * This token is internal and shouldn't be exposed to users
+   */
+  RESTRICTED = 'restricted',
 }
 
 /**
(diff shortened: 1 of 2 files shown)


saarujansathees578:
The failing test is occuring for every PR; doesn't block this from getting merged

modules/statics/src/base.ts
```diff
@@ -451,6 +451,11 @@ export enum CoinFeature {
    * This coin is an EVM compatible coin and should use common EVM logic in WP
    */
   EVM_COMPATIBLE_WP = 'evm-compatible-wp',
+
+  /**
+   * This nft shouldn't be shown when fetching balances
+   */
+  HIDDEN_NFT = 'hidden-nft',
```
theMobiusStrip: Let's call it `RESTRICTED`, and also apply to fungible tokens, not just NFT

theMobiusStrip (approved):

joeyangelino (approved):

PR #6688, opened 2025-08-12
feat(sdk-core): add retry logic for bulk accept share requests

aarvay:
## Changes Made
- Added `bulkAcceptShareRequestWithRetry()` private method
- Implements batching with retry logic for 413 errors
- Modified `bulkAcceptShareRequest()` to use the new retry method

## Details
The solution automatically handles payload size limits by:
1. Starting with optimal batch size (min(totalItems, 20))
2. On 413 error, reduces batch size by half and retries
3. Continues until successful or batch size reaches 1
4. Maintains original error if even single items fail

### DIFF
--- a/modules/bitgo/test/v2/unit/wallets.ts
+++ b/modules/bitgo/test/v2/unit/wallets.ts
@@ -22,6 +22,7 @@ import {
   makeRandomKey,
   getSharedSecret,
   BulkWalletShareOptions,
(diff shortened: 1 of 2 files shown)


sdoshibitgo (dismissed):

sdoshibitgo (approved):

kisslove-dewangan (approved):
lgtm

PR #6692, opened 2025-08-12
feat(utxo-core): add bip32derivation fields to BIP322 message signing

davidkaplanbitgo:
Add BIP32 derivation info to Psbt message signing for chain and index.

Issue: BTC-2384

### DIFF
--- a/modules/utxo-core/src/bip322/toSign.ts
+++ b/modules/utxo-core/src/bip322/toSign.ts
@@ -65,14 +65,20 @@ export function buildToSignPsbtForChainAndIndex(
   if (isTaprootChain(chain)) {
     throw new Error('BIP322 is not supported for Taproot script types.');
   }
-  const output = bitgo.outputScripts.createOutputScript2of3(
-    rootWalletKeys.deriveForChainAndIndex(chain, index).publicKeys,
-    bitgo.scriptTypeForChain(chain)
-  );
+  const walletKeys = rootWalletKeys.deriveForChainAndIndex(chain, index);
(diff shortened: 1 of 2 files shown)


modules/utxo-core/src/bip322/toSign.ts
```diff
@@ -65,14 +65,20 @@ export function buildToSignPsbtForChainAndIndex(
...
     scriptPubKey: output.scriptPubKey,
     redeemScript: output.redeemScript,
     witnessScript: output.witnessScript,
   });
+
+  psbt.updateInput(
+    0,
+    bitgo.getPsbtBip32DerivationOutputUpdate(rootWalletKeys, walletKeys, bitgo.scriptTypeForChain(chain))
+  );
+  bitgo.addXpubsToPsbt(psbt as bitgo.UtxoPsbt, rootWalletKeys);
```
islamaminBitGo: im noticing that buildToSignPsbt is doing this as bitgo.UtxoPsbt cast as well, https://github.com/BitGo/BitGoJS/blob/2641f250bb5ec7e8d0cbd12d146bc88d8a161df9/modules/utxo-core/src/bip322/toSign.ts#L26

is it possible to replace that line in toSign.ts with new UtxoPsbt instead of new Psbt? would that break anything?

davidkaplanbitgo: nothing really requires this to be a UtxoPsbt, which is more specific to 2of3, so I am keeping it this way. Maybe it is best to re-type those functions itself. Will do that

ppongbitgo (approved):

PR #6700, opened 2025-08-13
chore: removed stt

rohitsaw115:
TICKET: WIN-6703


modules/bitgo/src/v2/coins/index.ts
```diff
@@ -128,7 +128,7 @@ export { Rune, Trune };
 export { Sgb, Tsgb };
 export { Sol, Tsol };
 export { Soneium, Tsoneium, SoneiumToken };
-export { Stt, Tstt };
```
rv-bitgo, line 131: We wouldn't require Tstt either.

modules/bitgo/src/v2/coins/index.ts
```diff
@@ -54,7 +54,7 @@ import { Rbtc, Trbtc } from '@bitgo/sdk-coin-rbtc';
 import { Rune, Trune } from '@bitgo/sdk-coin-rune';
 import { Sei, Tsei } from '@bitgo/sdk-coin-sei';
 import { Soneium, Tsoneium, SoneiumToken } from '@bitgo/sdk-coin-soneium';
-import { Stt, Tstt } from '@bitgo/sdk-coin-stt';
```
rv-bitgo, line 57: don't require Tstt either i believe but please confirm if it's the expected update.

rv-bitgo (changes requested):
Please verify if we want to keep tstt functions, tests and related code.

rv-bitgo:
Also, requesting more descriptive commit messages or ticket description if needed.

modules/bitgo/src/v2/coins/index.ts
rohitsaw115: It is required for Testnet.

modules/bitgo/src/v2/coins/index.ts
rohitsaw115: It is required for Testnet.

rohitsaw115:
> Lets add a TODO ticket to add it back if required in future.

This is not required because STT has been replaced with SOMI in mainnet.

rv-bitgo (commented):
Looks good to merge but please be ready to monitor the repos which were touched during stt onboarding as they might still be trying to access stt from SDK.
A better idea would be to pro-actively check these following repos before merging this PR.

Repo | STT Reference checked | Need updates? | Remarks
-- | -- | -- | --
UI | ✅ | Yes | Will update after bitgo-beta release.
Indexer | ✅ | Yes | I think not required, indexer does not depend on sdk. although some stt reference are there.
IMS | ✅ | Yes | Will update during sdk version bump.
BGMS | ✅ | Yes | Will update during sdk version bump.
BGA | ✅ | Yes | Will update during sdk version bump.
OVC | ✅ | No | 
Dev portal | ✅ | No | 
AMS | ✅ | No | 
prime-microservices | ✅ | Yes | Will update during sdk version bump.
WRW | ✅ | No |

rv-bitgo (approved):

PR #6702, opened 2025-08-13
fix: use same feature for ofc tokens

mukeshsp:
We observed that separate features are being maintained for OFC tokens, but they should use the same features as their corresponding native coin or token. I confirmed with the Go team (with Manav) that currently, the OFC feature from the SDK isn’t being used—instead, they have a hardcoded mapping in the Prime service.

**Problem:**
Currently, the mapping of OFC tokens with Trust is based on hardcoded data in the Prime service. This means that whenever a new token needs to be onboarded, the hardcoded file has to be updated, which slows down the onboarding process.

**Solution:**
Instead of using hardcoded mappings, we should leverage the feature flags defined for OFC tokens. To enable this, we first need to correct the feature flags for OFC tokens.

Ticket: WIN-6707


sampras-saha (changes requested):
Test cases are failing. Please check

mukeshsp:
> Test cases are failing. Please check

that is flaky test which is not related to my change

sampras-saha (dismissed):
LGTM

modules/statics/test/unit/coins.ts
```diff
@@ -862,15 +862,14 @@ coins.forEach((coin, coinName) => {
...
+        const coinSupportsCustody =
+          coin.family !== CoinFamily.LNBTC &&
+          coin.family !== CoinFamily.CELO &&
+          coin.name !== 'ofccelo' &&
+          coin.name !== 'ofctcelo';
         coin.features.includes(CoinFeature.CUSTODY).should.eql(coinSupportsCustody);
         coin.features.includes(CoinFeature.CUSTODY_BITGO_TRUST).should.eql(coinSupportsCustody);
       });
-
-      it('should return false for all non-SD coin feature', () => {
```
gianchandania, line 870: why this was removed?

modules/statics/src/ofc.ts
```diff
@@ -90,6 +119,10 @@ export function ofcToken(
   isToken = true,
   kind: CoinKind = CoinKind.CRYPTO
 ): Readonly<OfcCoin> {
+  const filteredFeatures = getFilteredFeatures(suffix);
```
gianchandania, line 122: nit: Can add a method like and then invoke this in all the places

```
function applyFilteredFeatures(suffix: string, features: CoinFeature[]): CoinFeature[] {
  const filteredFeatures = getFilteredFeatures(suffix);
  return filteredFeatures.length > 0 ? filteredFeatures : features;
}
```

modules/statics/test/unit/coins.ts
mukeshsp: Since the OFC token uses the same feature flag as the native coin, it already includes the features related to `CUSTODY_BITGO_NEW_YORK` and `CUSTODY_BITGO_SWITZERLAND`. Therefore, this test case is no longer needed.

modules/statics/src/ofc.ts
mukeshsp: We are planing to remove the default feature so this is not require.

gianchandania (dismissed):

gianchandania (approved):

modules/statics/src/allCoinsAndTokens.ts
FloBitGo, line 1: Seems like this should rather be named `someCoinsAndTokens` :wink: 
What's the logic that determines which ones are defined here and which ones in `coins.ts`?

FloBitGo (approved):

FloBitGo (changes requested):
Please fix the failing tests

modules/statics/src/allCoinsAndTokens.ts
mukeshsp: all the native coins and token will be in [allCoinsAndTokens.ts] and ofc tokens are in ofc.ts file

mukeshsp:
> Please fix the failing tests

this is a flaky test which is failing for many PRs and we are looking into it parallelly

FloBitGo (approved):
Code is OK, test failure is unrelated and will get fixed in another PR

PR #6708, opened 2025-08-14
ref(wp): typed router implementation for acceptShare

rahulhegde-bitgo:
migrating acceptShare to typed router

Ticket: WP-5409


modules/express/src/typedRoutes/api/common/acceptShare.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export const PostAcceptShare = httpRoute({
+  path: '/api/v1/walletshare/:shareId/acceptShare',
+  method: 'POST',
+  request: httpRequest({
+    params: AcceptShareRequestParams,
+    body: AcceptShareRequestBody,
+  }),
+  response: {
+    200: t.UnknownRecord,
```
rahulhegde-bitgo, line 29: unknownRecord here is because of how function is defined, in the interface its return type is promise of any hence its very difficult to get the return type due to wallet specific implementation

kaustubhbitgo (approved):

modules/express/src/clientRoutes.ts
```diff
@@ -1598,12 +1598,7 @@ export function setupAPIRoutes(app: express.Application, config: Config): void {
   );
 
   app.post('/api/v1/wallet/:id/simpleshare', parseBody, prepareBitGo(config), promiseWrapper(handleShareWallet));
-  app.post(
-    '/api/v1/walletshare/:shareId/acceptShare',
-    parseBody,
-    prepareBitGo(config),
-    promiseWrapper(handleAcceptShare)
-  );
+  router.post('express.v1.wallet.acceptShare', [prepareBitGo(config), typedPromiseWrapper(handleAcceptShare)]);
```
pranavjain97, line 1601: why are we changing the name from `v1.wallet.acceptShare` -> `express.v1.wallet.acceptShare`? what is the "express" for?

pranavjain97: ah its for this https://developers.bitgo-dev.com/api/express.wallet.acceptshare, just double check it doesnt break the api docs after the api name change

rahulhegde-bitgo: Yes thats why I started with express, nice was looking for docs to verify the changes would be helpful will check this

PR #6713, opened 2025-08-14
feat(utxo-core): implement function to extract BIP322 message by index

davidkaplanbitgo:
Add `getBip322ProofMessageAtIndex` to retrieve the BIP322 message from a specific input index in a PSBT. The function returns the message as a Buffer or undefined if not found.

Issue: BTC-2375


christopher-fong (approved):
lgtm

PR #6714, opened 2025-08-14
feat(abstract-eth): support eip712 signing

benkelcher-bitgo:
Ticket: SC-2622
- add support for eip712 (`eth_signTypedData_v4`) to new message signing system
- use metamask's implementation of the standard


modules/abstract-eth/src/lib/messages/eip712/eip712Message.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(options: MessageOptions) {
+    super({
+      ...options,
+      type: MessageStandardType.EIP712,
+    });
+  }
+
+  async getSignablePayload(): Promise<string | Buffer> {
+    /*this.signablePayload = TypedDataUtils.eip712Hash(JSON.parse(this.payload), SignTypedDataVersion.V4);
+    return this.signablePayload;*/
```
haritkapadia: Missed debug comment

modules/abstract-eth/src/lib/messages/eip712/eip712Message.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    const typedData = JSON.parse(this.payload);
```
haritkapadia: This is untyped data, so I don't think it's a great var name.
https://metamask.github.io/eth-sig-util/latest/variables/TypedDataUtils.html even calls it `data`, not `typedData`.

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
```diff
@@ -0,0 +1,86 @@
...
+
+      const broadcastFormat = await message.toBroadcastFormat();
+
+      const expectedSerializedSignatures = serializeSignatures([fixtures.eip712.signature]);
+      broadcastFormat.type.should.equal(MessageStandardType.EIP712);
+      JSON.parse(broadcastFormat.payload).should.deepEqual(fixtures.messages.helloBob.message);
+      broadcastFormat.serializedSignatures?.should.deepEqual(expectedSerializedSignatures);
+      broadcastFormat.metadata!.should.deepEqual(fixtures.eip712.metadata);
+      broadcastFormat.signablePayload!.should.equal(
+        'GQHyzuN1+kK0IUOAQCX8RJ3q/VDMAxyiV+CxlKZQqRIJD8UsDuXYQmRHGAYpCj8sTOz8VJBia/kS0B8kDXonSzce'
```
haritkapadia: This should probably go in fixtures.

haritkapadia (changes requested):
I think as a follow up, we should identify which tests common to eip191 and eip712 can be extracted to a different file. For example, `'should reconstruct a message from broadcast format'` is applicable to every message type.

modules/abstract-eth/src/lib/messages/eip712/eip712Message.ts
benkelcher-bitgo: fixed, thanks

modules/abstract-eth/src/lib/messages/eip712/eip712Message.ts
benkelcher-bitgo: good call, will change

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
benkelcher-bitgo: for sure, and i will make the same change for the eip191 test as well

benkelcher-bitgo:
> I think as a follow up, we should identify which tests common to eip191 and eip712 can be extracted to a different file. For example, `'should reconstruct a message from broadcast format'` is applicable to every message type.

i don't disagree - but no need to do this as a follow-up, this PR is not urgent we might as well do it now. will make that change

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
```diff
@@ -0,0 +1,47 @@
...
+import { eip712Fixtures as fixtures } from '../fixtures';
+import { MessageStandardType } from '@bitgo/sdk-core';
+import { testBroadcastFormat, testMaintainingSignersAndSignatures } from '../abstractMessagesTests';
+
+describe('EIP712 Message', () => {
+  const sandbox = sinon.createSandbox();
+
+  afterEach(() => {
+    sandbox.restore();
+  });
```
haritkapadia: Is `sinon` used anywhere?

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  describe('signable payload generation', () => {
```
haritkapadia: Can this also be in the abstract test? BaseMessage.toBroadcastFormat already calls `getSignablePayload`.

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
```diff
@@ -0,0 +1,47 @@
...
+    for (const testCase of Object.keys(fixtures.messages)) {
```
haritkapadia: You can do `for (const x in obj) {}` to iterate over keys. You can also do `Object.entries(obj).map(([k, v]) => {})` for keys and values.

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  const validMessage = new EIP712Message({
+    coinConfig: fixtures.coin,
+    payload: JSON.stringify(fixtures.messages.simple.payload),
+    signers: [fixtures.signing.signer],
+    signatures: [fixtures.signing.signature],
+  });
+
+  it('should initialize with the correct type', () => {
```
haritkapadia: Can this also be in the abstract test? `getType` is implemented there.

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
```diff
@@ -0,0 +1,47 @@
...
+    validMessage.getType().should.equal(MessageStandardType.EIP712);
+  });
+
+  testMaintainingSignersAndSignatures(MessageStandardType.EIP712, { ...fixtures.signing }, validMessage);
+  testBroadcastFormat(
+    MessageStandardType.EIP712,
+    { ...fixtures.signing, signablePayload: fixtures.messages.simple.base64 },
+    validMessage
+  );
```
haritkapadia: Ideally we would run all of the abstract tests with one function call. This way, subclassers don't have to think about which tests to call.

However, I can see how that would be messy (a giant params object) or inconvenient (you don't know all the params yet).

Try having one function for tests and see if "messy" or "inconvenient". If it is, then just respond here.

haritkapadia: Also, let's put the abstract tests at the top to make it obvious we're executing other tests.

modules/abstract-eth/test/unit/messages/eip712/eip712MessageBuilder.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  describe('constructor', () => {
+    it('should initialize with the correct message type', () => {
+      // Test the builder's private property indirectly through build()
+      const builder = new Eip712MessageBuilder(fixtures.coin);
+      builder.should.be.instanceof(Eip712MessageBuilder);
```
haritkapadia: I think this is always true. Maybe you're thinking of `builder.type`? See my other comment:

> Can this also be in the abstract test? getType is implemented there.

modules/abstract-eth/test/unit/messages/eip712/eip712MessageBuilder.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      const message = await builder.build();
+      JSON.parse(message.getPayload()).should.deepEqual(fixtures.messages.simple.payload);
+      message.should.be.instanceof(EIP712Message);
+      message.getType().should.equal(MessageStandardType.EIP712);
+      message.getMetadata()?.should.have.property('customData', 'test data');
+      message.getMetadata()?.should.have.property('encoding', 'utf8');
+    });
+
+    it('should throw an error when building without setting payload', async () => {
```
haritkapadia: Can this also be in the abstract test? The payload requirement is checked in `BaseMessageBuilder.build`.

modules/abstract-eth/test/unit/messages/abstractMessageBuilderTests.ts
```diff
@@ -0,0 +1,75 @@
...
+    payload,
+    signature,
+    signer,
+  }: {
+    broadcastMessage: BroadcastableMessage;
+    payload: string;
+    signature: Signature;
+    signer: string;
+  },
+  additionalChecks: (msg: IMessage) => void
```
haritkapadia: I don't think we should allow arbitrary additional checks. I see you use it to test the type of the constructed message object. You can make this function accept the class as an argument.

Maybe this will help? https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-construct-signatures

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
```diff
@@ -0,0 +1,96 @@
...
+        {
+          publicKey: { pub: 'pub2' },
+          signature: Buffer.from('replaced-signature'),
+        },
+      ]);
+      message.getSigners().should.deepEqual(['replaced-signer']);
+
+      // Reset signatures before returning
+      message.setSignatures([signature]);
+      message.setSigners([signer]);
```
haritkapadia: Haha, it's unfortunate we have to do that. How did the old test handle this?

Can we maybe turn `message: BaseMessage` into `message: () => BaseMessage`, constructing a new message each time? If so, some JSDoc to explain how to use this parameter would be important.

haritkapadia (changes requested):

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
benkelcher-bitgo: done - some of these tests do overlap a bit, but they test different things so i think it's probably alright

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
benkelcher-bitgo: done

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
benkelcher-bitgo: done this, it's not too messy

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
benkelcher-bitgo: yeah, passing the constructor around allows us to avoid this. fixed

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
haritkapadia: What's the test called? I don't see it in `abstractMessagesTests.ts`.

benkelcher-bitgo: we do it as part of the payload generation test:
```ts
message.getType().should.equal(messageType)
```

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
```diff
@@ -0,0 +1,178 @@
+import { BaseMessage, MessageMetadata, MessageStandardType, serializeSignatures, Signature } from '@bitgo/sdk-core';
+import { BaseCoin } from '@bitgo/statics';
+
+type MessageCtor = new (...args: any[]) => BaseMessage;
```
haritkapadia: Let's change `(...args: any[])` to a stricter type. Can you reuse types from BaseMessage?

```
        const message = new messageCtor({
          payload,
          coinConfig,
          metadata,
        });
```

```
      const message = new messageCtor({
        coinConfig,
        payload,
        signatures: [signature],
        signers: [signer],
      });
```

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
```diff
@@ -0,0 +1,178 @@
...
+    messages,
+    detailedMessage,
+  }: {
+    messages: { [_: string]: Message };
+    detailedMessage: DetailedMessage;
+  }
+): void {
+  testSignablePayloadGeneration(messageType, messageCtor, { messages });
+  testMaintainingSignersAndSignatures(messageType, messageCtor, { ...detailedMessage });
+  testBroadcastFormat(messageType, messageCtor, { ...detailedMessage });
```
haritkapadia: You shouldn't need to shallow copy the input data.

```suggestion
  testMaintainingSignersAndSignatures(messageType, messageCtor, detailedMessage);
  testBroadcastFormat(messageType, messageCtor, detailedMessage);
```

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+function testSignablePayloadGeneration(
+  messageType: MessageStandardType,
+  messageCtor: MessageCtor,
+  {
+    messages,
+  }: {
+    messages: { [_: string]: Message };
```
haritkapadia:
```suggestion
    messages: Record<string, Message>;
```

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
```diff
@@ -0,0 +1,178 @@
...
+    payload,
+    signature,
+    signer,
+  }: {
+    payload: string;
+    signature: Signature;
+    signer: string;
+  }
+): void {
+  const coinConfig = { name: 'eth' } as BaseCoin;
```
haritkapadia: Can you avoid the `as` type assertion? I think there's a way to generate a `coinConfig` with `coins.get("eth")`. If that doesn't work, try:

```suggestion
  const coinConfig: BaseCoin = { name: 'eth' };
```

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
```diff
@@ -0,0 +1,178 @@
+import { BaseMessage, MessageMetadata, MessageStandardType, serializeSignatures, Signature } from '@bitgo/sdk-core';
+import { BaseCoin } from '@bitgo/statics';
+
+type MessageCtor = new (...args: any[]) => BaseMessage;
+type Message = { payload: string; hex: string; metadata?: MessageMetadata };
+type DetailedMessage = Message & { base64: string; signature: Signature; signer: string };
```
haritkapadia: Is there any relation between `Message` and the arguments to `messageCtor`? It's unclear whether `hex` will be used to construct the message, or if the constructed message should broadcast to `hex`.

Same with `DetailedMessage`: `signature` and `signer` are for construction, `base64` is not.

See if you can show this relationship.

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export function testEthMessageSigning(
+  messageType: MessageStandardType,
+  messageCtor: MessageCtor,
+  {
+    messages,
+    detailedMessage,
+  }: {
+    messages: { [_: string]: Message };
+    detailedMessage: DetailedMessage;
```
haritkapadia: What the difference is between `messages` and `detailedMessage`? Can you explain that difference in the argument names?

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
```diff
@@ -0,0 +1,178 @@
...
+      it(`should generate the correct signable payload for message '${key}'`, async () => {
+        const message = new messageCtor({
+          payload,
+          coinConfig,
+          metadata,
+        });
+        const signablePayload = await message.getSignablePayload();
+
+        message.getType().should.equal(messageType);
+        message.getMetadata()?.should.deepEqual(metadata || {});
```
haritkapadia: I don't think this handles `getMetadata() === undefined && metadata !== undefined`.

haritkapadia (changes requested):
1. The comments I left for `abstractMessagesTests.ts` also apply to `abstractMessageBuilderTests.ts`.
2. Should we change the name of `abstractMessageBuilderTests.ts` to `baseMessageBuilderTests.ts`? Also `abstractMessagesTests.ts` to `baseMessageTests.ts`.

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
haritkapadia: Hmmmm, I think we should preserve all of the existing tests.

modules/abstract-eth/test/unit/messages/eip712/eip712MessageBuilder.ts
haritkapadia: Keeping this unresolved regarding preserving tests: https://github.com/BitGo/BitGoJS/pull/6714#discussion_r2305233271

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
benkelcher-bitgo: we can use `MessageOptions` directly :+1:

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
benkelcher-bitgo: we can split the types into `MessageConstructionData` and `MessageValidationData`, and then use the union of those two as the type for the params.

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
benkelcher-bitgo: `detailedMessage` includes signing data and base64 broadcast format payload for validation. i will change the name to `messageWithSigningData`, and modify the typing for the validation to make this distinction more clear

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
benkelcher-bitgo: good call - something like this should work
```ts
if (metadata) {
          message.getMetadata()?.should.not.be.empty();
          message.getMetadata()?.should.deepEqual(metadata);
        } else {
          message.getMetadata()?.should.be.empty();
        }
```
(`getMetadata()` is never actually undefined for EVM, if there is no metadata passed it's just an empty object)

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
benkelcher-bitgo: sure, i can split this off

modules/abstract-eth/test/unit/messages/eip712/eip712Message.ts
benkelcher-bitgo: removed

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
haritkapadia: `?.` terminates early. I think `should.strictEqual(a, b)` should work.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

```ts
import "should"

function f(x: boolean): string | undefined {
  if (x) return "Hi"
  return undefined
}

f(true)?.should.equal("Hi")
f(false)?.should.equal("Hi") // should fail, does not
f(true)?.should.equal("Bye") // should be unreachable, reaches
```

modules/abstract-eth/test/unit/messages/abstractEthMessagesTests.ts
```diff
@@ -0,0 +1,182 @@
...
+type MessageValidationData = { hex: string };
+type MessageValidationDataWithBase64 = MessageValidationData & { base64: string };
+
+const coinConfig = coins.get('eth');
+
+export function testEthMessageSigning(
+  messageType: MessageStandardType,
+  messageCtor: MessageCtor,
+  messages: Record<string, MessageConstructionData & MessageValidationData>,
+  messageWithSigningData: MessageConstructionDataWithSigning & MessageValidationDataWithBase64
```
haritkapadia: Can you pass data without taking the type intersection (`&`)? For example:

```ts
type MessageSignablePayloadAssertion = {
  constructionData: {
    coinConfig: BaseCoin;
    payload: string;
    metadata?: MessageMetadata;
  };
  expected: {
    signablePayloadHex: string;
  };
};
type EquivalentMessageData = {
  constructionData: {
    coinConfig: BaseCoin;
    payload: string;
    signature: Signature;
    signer: string;
  };
  broadcastFormat: {
    base64: string;
  };
};
```

haritkapadia (changes requested):
1. I think you should choose the abstract test function parameters, then shape the data from `fixtures` as required.
2. Put shared types in a new file so you can deduplicate them.
3. `MessageValidationData.hex` is only used to validate `message.getSignablePayload()`. Better names can be given based on this information.

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
haritkapadia: Related: https://github.com/BitGo/BitGoJS/pull/6714#discussion_r2308231625

modules/abstract-eth/test/unit/messages/abstractMessagesTests.ts
haritkapadia: > getMetadata() is never actually undefined for EVM

Since the type says it's nullable, we should handle the null case.

modules/abstract-eth/test/unit/messages/abstractEthMessagesTests.ts
benkelcher-bitgo: updated this - now we are using intersections in the type definitions to clarify the distinction between test inputs and outputs, but the testing function itself only takes in a `TestConfig` object.

benkelcher-bitgo:
> 1. I think you should choose the abstract test function parameters, then shape the data from `fixtures` as required.
> 2. Put shared types in a new file so you can deduplicate them.
> 3. `MessageValidationData.hex` is only used to validate `message.getSignablePayload()`. Better names can be given based on this information.

1. Fair point, this has been done - fixtures are cleaner now and have been separated for EIP191 and EIP712. This does cause a small amount of duplication (the signature information) but I think that the clarity we get by separating the two is worth that small duplication.
2. Done.
3. Better names have been given across the board. The hex and base64 params in particular were just different formats of the same piece of data, so the naming has been updated to make this clear.

modules/abstract-eth/test/unit/messages/abstractEthMessageBuilderTests.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      const builtMessage = await builder.build();
+      builtMessage.getType().should.equal(messageType);
+      builtMessage.getPayload().should.equal(payload);
+
+      if (!metadata) {
+        should.deepEqual(builtMessage.getMetadata(), {});
+      } else {
+        should.deepEqual(builtMessage.getMetadata(), metadata);
+      }
```
haritkapadia: Nit: switch if bodies

```suggestion
      if (metadata) {
        should.deepEqual(builtMessage.getMetadata(), metadata);
      } else {
        should.deepEqual(builtMessage.getMetadata(), {});
      }
```

modules/abstract-eth/test/unit/messages/abstractEthMessageBuilderTests.ts
```diff
@@ -0,0 +1,126 @@
...
+      const builder = factory.getMessageBuilder(messageType);
+      builder.setPayload(payload);
+      builder.setMetadata({ encoding: 'hex', customData: 'test data' });
+
+      const message = await builder.build();
+      should.notEqual(message.getMetadata(), {});
+      if (message.getMetadata()) {
+        message.getMetadata()?.should.have.property('encoding', 'utf8');
+        message.getMetadata()?.should.have.property('customData', 'test data');
+      }
```
haritkapadia: Let's remove `?.`.

```suggestion
      const metadata = message.getMetadata();
      should.not.be.undefined(metadata);
      should.not.deepEqual(metadata, {});
      metadata.should.have.property('encoding', 'utf8');
      metadata.should.have.property('customData', 'test data');
```
Note:
```
> {} === {}
false
> Object.is({}, {})
false
```

modules/abstract-eth/test/unit/messages/abstractEthMessagesTests.ts
```diff
@@ -0,0 +1,143 @@
...
+        });
+
+        const signablePayload = await message.getSignablePayload();
+        signablePayload.toString('hex').should.equal(expected.expectedSignableHex);
+
+        if (!input.metadata) {
+          should.deepEqual(message.getMetadata(), {});
+        } else {
+          should.deepEqual(message.getMetadata(), input.metadata);
+        }
```
haritkapadia: Nit: switch if bodies

haritkapadia (changes requested):
The test structure looks much better now 👍

modules/abstract-eth/test/unit/messages/abstractEthMessageBuilderTests.ts
benkelcher-bitgo: becuase `metadata` might be undefined we need to use this syntax for checking it instead if we want to avoid `?.` and `!.`. also, `not.deepEqual({})` is extraneous because we are checking if it has properties on it anyway (if it passes the properties checks, we know it is not `{}`)
```typescript
should(metadata).not.be.undefined();
// should(metadata).not.deepEqual({}); remove this one
should(metadata).have.property('encoding', 'utf8');
should(metadata).have.property('customData', 'test data');
```

haritkapadia (approved):
Looks good!

kisslove-dewangan (approved):
lgtm based on codeowners

PR #6717, opened 2025-08-14
fix: support node 22 in CI

zahin-mohammad:
Example failure: https://github.com/BitGo/BitGoJS/actions/runs/16976597735/job/48127026773?pr=6716


margueriteblair (dismissed):

modules/sdk-api/test/unit/v1/wallet.ts
```diff
@@ -12,7 +12,7 @@ import * as _ from 'lodash';
 import { common } from '@bitgo/sdk-core';
 import * as utxolib from '@bitgo/utxo-lib';
 import * as should from 'should';
-import nock from 'nock';
+const nock = require('nock');
```
zahin-mohammad, line 15:
```
The error shows that nock is exported with export = syntax and can't be default-imported without the esModuleInterop flag.
```

OttoAllmendinger (approved):

PR #6718, opened 2025-08-15
feat(utxo-core): add verification func for to_sign txs

davidkaplanbitgo:
Add function to verify BIP322 to_sign transaction structure and that inputs reference the correct to_spend transactions for given addresses and messages. Includes test coverage for verifying multiple messages.

TICKET: BTC-2375

### DIFF
--- a/modules/utxo-core/src/bip322/verify.ts
+++ b/modules/utxo-core/src/bip322/verify.ts
@@ -1,0 +1,53 @@
+import assert from 'assert';
+
+import * as utxolib from '@bitgo/utxo-lib';
+
+import { buildToSpendTransaction } from './toSpend';
+
+export type VerificationInfo = {
+  message: string;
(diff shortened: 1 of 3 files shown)


modules/utxo-core/src/bip322/verify.ts
```diff
@@ -0,0 +1,53 @@
...
+  address: string;
+};
+
+/**
+ *
+ * @param toSignTx
+ * @param verificationInfo
+ * @param network
+ */
+export function verifyFixedScriptToSignTxWithInfo(
```
OttoAllmendinger, line 18: per convention, `void` funcs that throw are prefixed with `assert`

```suggestion
export function assertValidFixedScriptToSignTxWithInfo(
```

OttoAllmendinger (changes requested):

PR #6724, opened 2025-08-18
feat(express): typed-router for pendingApprovals

rahulhegde-bitgo:
moving pendingApprovals api to typed router

Ticket: WP-5410

### DIFF
--- a/modules/express/src/typedRoutes/api/v1/pendingApproval.ts
+++ b/modules/express/src/typedRoutes/api/v1/pendingApproval.ts
@@ -1,0 +1,35 @@
+import * as t from 'io-ts';
+import { httpRoute, httpRequest, optional } from '@api-ts/io-ts-http';
+import { BitgoExpressError } from '../../schemas/error';
+
+export const pendingApprovalRequestParams = {
+  id: t.string,
+};
+
+export const pendingApprovalRequestBody = {
+  walletPassphrase: optional(t.string),
+  otp: optional(t.string),
+  tx: optional(t.string),
+  xprv: optional(t.string),
+  previewPendingTxs: optional(t.boolean),
+  pendingApprovalId: optional(t.string),
+};
+
+/**
+ * Pending approval request
+ *
+ * @operationId express.v1.pendingapprovals
+ */
+
+export const PutPendingApproval = httpRoute({
+  path: '/api/v1/pendingapprovals/:id/express',
+  method: 'PUT',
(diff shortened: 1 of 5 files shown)


modules/express/src/typedRoutes/api/v1/pendingApproval.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const PutPendingApproval = httpRoute({
+  path: '/api/v1/pendingapprovals/:id/express',
+  method: 'PUT',
+  request: httpRequest({
+    params: pendingApprovalRequestParams,
+    body: pendingApprovalRequestBody,
+  }),
+  response: {
+    200: t.UnknownRecord,
```
rahulhegde-bitgo, line 32: keeping it like this since there are no docs for this and has many paths which difficiult to infer the types from

modules/express/test/unit/typedRoutes/pendingApproval.ts
```diff
@@ -0,0 +1,207 @@
+import * as assert from 'assert';
+import * as t from 'io-ts';
+import {
+  pendingApprovalRequestParams,
+  pendingApprovalRequestBody,
+  PutPendingApproval,
+} from '../../../src/typedRoutes/api/v1/pendingApproval';
```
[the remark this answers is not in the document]
rahulhegde-bitgo: Since this is pretty straight forward we can ignore this

modules/express/src/typedRoutes/api/v1/pendingApproval.ts
```diff
@@ -0,0 +1,35 @@
...
+  id: t.string,
+};
+
+export const pendingApprovalRequestBody = {
+  walletPassphrase: optional(t.string),
+  otp: optional(t.string),
+  tx: optional(t.string),
+  xprv: optional(t.string),
+  previewPendingTxs: optional(t.boolean),
+  pendingApprovalId: optional(t.string),
```
kaustubhbitgo, line 15: does the bitgojs has comments for this type? If yes can you please add it here? in `/** comment-here */` this will help in auto generating api docs

rahulhegde-bitgo: Actually it doesn't have any comments in the interface so we can skip this maybe?

kaustubhbitgo (approved):
lgtm

PR #6742, opened 2025-08-19
fix: make VetNFTCollection extend NFTCollectionIdDefinedToken

at31416:
Ticket: COIN-5253


ravibitgo (dismissed):

modules/statics/src/coins.ts
```diff
@@ -4241,22 +4240,6 @@ export const coins = CoinMap.fromCoins([
     UnderlyingAsset['tapt:beta3loanbook'],
     APT_FEATURES
   ),
-  vetNFTCollection(
```
at31416, line 4244: Will take up onboarding these NFTs once WP for VetNFTCollection class is ready, else it will block beta-bumps in BGMS.

MohammedRyaan786 (approved):

PR #6748, opened 2025-08-19
feat(sdk-coin-sol): add support to recover unsupported spl tokens

venkateshv1266:
TICKET: COIN-5327

This PR allows unsupported token recovery using **Non-Bitgo Recovery** form from WRW.


nvjsr (approved):

nvjsr (commented):
Can we have test case for this

PR #6786, opened 2025-08-22
feat: enable smart contract calls aptos

shivam-bitgo:
Ticket: COIN-5352

This pull request adds support for custom Aptos transactions that invoke arbitrary entry functions, enabling users to interact with any Aptos smart contract or system function through the BitGo SDK. It introduces new types, transaction classes, and builder logic to handle these custom transactions, while also updating the transaction factory and utility logic to recognize and process them.

### Custom Transaction Support

* Added the `CustomTransaction` class in `customTransaction.ts` to represent transactions that call arbitrary Aptos entry functions, including support for module/function names, type arguments, function arguments, and optional ABI validation.
* Introduced the `CustomTransactionBuilder` in `customTransactionBuilder.ts` to construct and validate custom transactions, ensuring required parameters are set and integrating ABI type checking.
* Defined the `CustomTransactionParams` interface in `iface.ts`, documenting all fields needed for custom entry function calls and optional ABI validation.

### SDK Integration

* Updated `index.ts` to export `CustomTransaction` and `CustomTransactionBuilder`, making them available for SDK consumers.
* Modified `transactionBuilderFactory.ts` to recognize and handle custom transactions, including deserialization, builder instantiation, and fallback logic for unknown transaction types. [[1]](diffhunk://#diff-10277f68441387da0666485787f449c51f6848665fdb1e68ae9e737b3ea9fd3dR13-R14) [[2]](diffhunk://#diff-10277f68441387da0666485787f449c51f6848665fdb1e68ae9e737b3ea9fd3dR40-R43) [[3]](diffhunk://#diff-10277f68441387da0666485787f449c51f6848665fdb1e68ae9e737b3ea9fd3dR54-R59) [[4]](diffhunk://#diff-10277f68441387da0666485787f449c51f6848665fdb1e68ae9e737b3ea9fd3dR77-R86)

### Utility Improvements

* Changed `utils.ts` so that unknown function calls are now treated as custom transactions, ensuring the SDK can process any valid Aptos entry function.

### Type and ABI Integration

* Imported `EntryFunctionABI` from the Aptos SDK to support ABI-based validation and type safety for custom transactions.


baltiyal (commented):
Please change the structure to extend the existing transaction, and builder, so that we have least code redundancy and easier maintainability.

modules/sdk-coin-apt/test/unit/transactionBuilder/customTransactionBuilder.ts
```diff
@@ -0,0 +1,317 @@
+import { coins } from '@bitgo/statics';
+import { TransactionBuilderFactory, CustomTransaction } from '../../../src';
+import * as testData from '../../resources/apt';
+import { TransactionType } from '@bitgo/sdk-core';
+import should from 'should';
+
+describe('Apt Custom Transaction Builder', () => {
```
baltiyal, line 7: Add test cases for smart contracts via the newer structure.
As discussed, minting and burning of tokens would be needed.
So add test cases for those transactions.

shivam-bitgo: added tests for mint and burn

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
```diff
@@ -0,0 +1,112 @@
...
+  InputGenerateTransactionPayloadData,
+  TransactionPayload,
+  TransactionPayloadEntryFunction,
+} from '@aptos-labs/ts-sdk';
+import { CustomTransactionParams } from '../iface';
+
+/**
+ * Transaction class for custom Aptos transactions with entry function payloads.
+ */
+export class CustomTransaction extends Transaction {
```
baltiyal, line 17: As already discussed, can't we extend `FungibleAssetTransfer` instead?

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
```diff
@@ -0,0 +1,112 @@
...
+    super(coinConfig);
+    this._type = TransactionType.CustomTx;
+  }
+
+  /**
+   * Set the custom transaction parameters
+   *
+   * @param {CustomTransactionParams} params - Custom transaction parameters
+   */
+  setCustomTransactionParams(params: CustomTransactionParams): void {
```
baltiyal, line 32: Why setters are public in this file?

modules/sdk-coin-apt/test/unit/transactionBuilder/customTransactionBuilder.ts
baltiyal: Instead of mint and burn for AptosCoin, could you try for fungible assests (i.e. tokens) as that is the sole objective for adding this functionality.

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+    return `${this._moduleName}::${this._functionName}`;
+  }
+
+  /**
+   * Parse a transaction payload to extract the custom transaction data
+   *
+   * @param {TransactionPayload} payload - The transaction payload to parse
+   */
+  protected parseTransactionPayload(payload: TransactionPayload): void {
```
baltiyal, line 70: In this function, there is not much of the validation happening, we will be allowing any kind of transaction to pass through.
Prima facie, it might look we are enabling for all kind of transaction and we are bettering it. But, it is not just building, even indexing of the those transaction needs to be defined.
Thus, we would be needing a certain types of transactions which we will be allowing for building or indexing and validate them properly.

modules/sdk-coin-apt/test/unit/transactionBuilder/customTransactionBuilder.ts
```diff
@@ -0,0 +1,380 @@
...
+  const factory = new TransactionBuilderFactory(coins.get('tapt'));
+
+  describe('Custom Entry Function Call', () => {
+    describe('Succeed', () => {
+      it('should build a custom transaction with transfer_coins function', async function () {
```
kamleshmugdiya, line 12: duplicated test `should build a custom transaction with transfer_coins function`.

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Generate the transaction payload data for the custom transaction
+   *
+   * @returns {InputGenerateTransactionPayloadData} The transaction payload data
+   */
+  protected getTransactionPayloadData(): InputGenerateTransactionPayloadData {
+    return {
+      function: this.fullFunctionName as `${string}::${string}::${string}`,
+      typeArguments: this._typeArguments,
+      functionArguments: this._functionArguments,
```
baltiyal, line 109: check the functionality of multiple recipients too here.
Add a test case may be for that too.

modules/sdk-coin-apt/src/lib/transactionBuilder/customTransactionBuilder.ts
```diff
@@ -0,0 +1,122 @@
...
+    return this;
+  }
+
+  /**
+   * Check if a transaction payload is valid for this builder
+   *
+   * @param {TransactionPayload} payload - The transaction payload to check
+   * @returns {boolean} True if the payload is valid for this builder
+   */
+  protected isValidTransactionPayload(payload: TransactionPayload): boolean {
```
baltiyal, line 73: Needs more validation.

modules/sdk-coin-apt/src/lib/iface.ts
```diff
@@ -34,3 +35,89 @@ export interface RecipientsValidationResult {
...
+  /**
+   * Function arguments in the order expected by the entry function (optional)
+   *
+   * Arguments are automatically serialized based on their JavaScript types.
+   * Use appropriate types: strings for addresses, numbers/BigInts for amounts.
+   *
+   * @example [recipient_address, amount] for transfers
+   * @example [domain_name, duration] for domain registration
+   */
+  functionArguments?: any[];
```
kamleshmugdiya: is it possible to avoid `any`?
and use a type.

kamleshmugdiya: if not, try to use `unknown` instead, which provides safer type safety than `any`.

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  /**
+   * Generate the transaction payload data for the custom transaction
+   *
+   * @returns {InputGenerateTransactionPayloadData} The transaction payload data
+   */
+  protected getTransactionPayloadData(): InputGenerateTransactionPayloadData {
+    return {
+      function: this.fullFunctionName as `${string}::${string}::${string}`,
```
kamleshmugdiya: `${string}::${string}::${string}` this is just compile time check.
this won't check value at runtime.

add validation for the input `this.fullFunctionName` to check the format.

modules/sdk-coin-apt/test/unit/transactionBuilder/customTransactionBuilder.ts
shivam-bitgo: fungible_asset::mint is not an entry function, so it can't be called directly from transactions or simulations. Unlike managed_coin, it requires a custom Move module to wrap the mint/burn logic and manage capabilities

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
shivam-bitgo: in FungibleAssetTransfer we have:
  - Hardcoded for specific functions: Only 0x1::primary_fungible_store::transfer or batch transfers
  - Fixed argument structure: Always expects exactly 3 arguments (assetId, recipient, amount)
  - Fixed validation: isValidPayload() enforces strict argument count and type constraints

If we want to extend it then we would need to make more modifications and I think it's better not to extend it as we can have more functionalities apart from just mint/burn using this custom tx 

cc- @sachushaji

modules/sdk-coin-apt/src/lib/transactionBuilder/customTransactionBuilder.ts
shivam-bitgo: added more validations

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
shivam-bitgo: done

modules/sdk-coin-apt/src/lib/iface.ts
shivam-bitgo: addressed

modules/sdk-coin-apt/test/unit/transactionBuilder/customTransactionBuilder.ts
shivam-bitgo: removed

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -181,6 +181,17 @@ export interface PrebuildTransactionOptions {
     }[];
     data: string;
   }[];
+  /**
+   * Custom transaction parameters for Aptos entry function calls.
+   * Used with the customTx intent type for Aptos smart contract interactions.
+   */
+  customTransactionParams?: {
```
raj-bitgo: I think we should name this specific to aptos, customTransactionParams sounds generic which this isn't

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3410,7 +3410,10 @@ export class Wallet implements IWallet {
             sequenceId: params.sequenceId,
             comment: params.comment,
             solInstructions: params.solInstructions,
+            customTransactionParams: params.customTransactionParams,
```
raj-bitgo: can you check if you can add any validations inside `prebuildTxWithIntent` for aptos

baltiyal (dismissed):
As discussed, constraint to allow functions or modules specific to mint and burn and other needed types can be included later if they are needed. This should be reflected in validations and checks to enable only specific types of modules and functions to be allowed in transaction building. Take reference from existing fungible, digital and transfer handling for aptos.

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -219,6 +219,17 @@ export interface PrebuildTransactionWithIntentOptions extends IntentOptionsBase
    * Each instruction should contain programId, keys, and data fields.
    */
   solInstructions?: SolInstruction[];
+  /**
+   * Custom transaction parameters for Aptos entry function calls.
+   * Used with the customTx intent type for Aptos smart contract interactions.
+   */
+  customTransactionParams?: {
```
baltiyal: Aligned with @raj-bitgo it should mention it is related to aptos.

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
kaustubhbitgo: I believe this is transaction payload, validation for sender / receiver addresses should be happening somewhere already.

kaustubhbitgo (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
shivam-bitgo: Added

modules/sdk-core/src/bitgo/wallet/iWallet.ts
shivam-bitgo: Changed

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
shivam-bitgo: Done

raj-bitgo (dismissed):

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
```diff
@@ -0,0 +1,210 @@
...
+   */
+  get recipient(): any {
+    // For custom transactions, return a placeholder recipient if no recipients exist
+    if (this._recipients.length === 0) {
+      return {
+        address: '', // Empty address for custom transactions
+        amount: '0',
+      };
+    }
+    return this._recipients[0];
```
raj-bitgo, line 184: We're always returning the first recipient? Can there be multiple recipients?

shivam-bitgo: that is handled via recipients method, I had to override this method as in normal tx there is an assert the recipient should not be empty but in our case since we are passing recipient as function arg we can have recipient empty as well.

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -181,6 +181,17 @@ export interface PrebuildTransactionOptions {
     }[];
     data: string;
   }[];
+  /**
+   * Custom transaction parameters for Aptos entry function calls.
+   * Used with the customTx intent type for Aptos smart contract interactions.
+   */
+  aptosCustomTransactionParams?: {
```
baltiyal, line 188: `aptosCustomTransactionParams` itself being optional? What will be the usecase for that?

baltiyal: Ohh, you are using here : https://github.com/BitGo/BitGoJS/pull/6786/files#diff-61564054eb5fde698be0bf79948ebc38ac638f59e9608c12ac6a7e8f1c9d0c1aR116 as unknown

modules/sdk-coin-apt/test/unit/transactionBuilder/customTransactionBuilder.ts
```diff
@@ -0,0 +1,430 @@
...
+        // Valid payload should pass validation
+        txBuilder.sender(testData.sender2.address);
+        txBuilder.gasData({
+          maxGasAmount: 200000,
+          gasUnitPrice: 100,
+        });
+        txBuilder.sequenceNumber(14);
+        txBuilder.expirationTime(1736246155);
+        txBuilder.customTransaction({
+          moduleName: '0x1::managed_coin',
```
baltiyal, line 262: nit: could add test case for token deployed on address other than `0x1` as done in sandboxing. If you are confirmed that `0x1::managed_coin` will be module then this is not required. But, if different address, module & function would be there, then test case for that make sense.

modules/sdk-coin-apt/src/lib/transactionBuilderFactory.ts
```diff
@@ -44,7 +51,12 @@ export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
 
   getTransactionTypeFromSignedTxn(signedTxn: SignedTransaction): TransactionType {
     const rawTxn = signedTxn.raw_txn;
-    return utils.getTransactionTypeFromTransactionPayload(rawTxn.payload);
+    try {
+      return utils.getTransactionTypeFromTransactionPayload(rawTxn.payload);
+    } catch (e) {
```
baltiyal: Should not be returning `TransactionType.CustomTx` here, as it could be a cas of invalid payload. Also, already you are returning in default - `TransactionType.CustomTx`

modules/sdk-coin-apt/src/lib/transactionBuilder/customTransactionBuilder.ts
```diff
@@ -0,0 +1,143 @@
...
+        console.error('Invalid transaction payload: not an entry function');
+        return false;
+      }
+
+      // Create a temporary CustomTransaction to leverage existing validation
+      const tempTransaction = new CustomTransaction(this._coinConfig);
+
+      try {
+        // Use the transaction's built-in parsing which includes validation
+        tempTransaction['parseTransactionPayload'](payload);
```
baltiyal: What does this do?
Are we calling `parseTransactionPayload` method for CustomTransaction?
Isn't there a better way to implement this rather than creating a temporary transaction?

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+  }
+
+  /**
+   * Override the deprecated recipient getter to handle custom transactions gracefully
+   * Custom transactions may not have traditional recipients
+   *
+   * @deprecated - use `recipients()`
+   */
+  get recipient(): any {
```
baltiyal, line 176: could we use `recipients` as its deprecated?

shivam-bitgo: this is for toJson() i had to override

modules/sdk-coin-apt/src/lib/transactionBuilder/customTransactionBuilder.ts
shivam-bitgo: addressed

modules/sdk-coin-apt/src/lib/transactionBuilderFactory.ts
shivam-bitgo: removed redundant check

modules/sdk-coin-apt/test/unit/transactionBuilder/customTransactionBuilder.ts
shivam-bitgo: will do it in followup PR

kaustubhbitgo (approved):

venkateshv1266 (approved):
LGTM

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
baltiyal: Use `recipients` instead, as already mentioned `recipient` is deprecated and would be removed later.
Ticket for reference : https://bitgoinc.atlassian.net/browse/COIN-3798

modules/sdk-coin-apt/src/lib/transaction/customTransaction.ts
baltiyal: In future, you will need to change this functionality to support multiple recipients (sendMany)

modules/sdk-coin-apt/src/lib/transactionBuilder/customTransactionBuilder.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+    try {
+      const entryFunction = payload.entryFunction;
+      // Validate module address format
+      const moduleAddress = entryFunction.module_name.address.toString();
+      if (!moduleAddress.startsWith('0x')) {
+        return false;
+      }
+      // Validate module and function identifiers
+      const moduleIdentifier = entryFunction.module_name.name.identifier;
```
baltiyal, line 86: Could create utility methods to remove the redundant implementation of validation of module and function.

modules/sdk-coin-apt/src/lib/transactionBuilder/customTransactionBuilder.ts
baltiyal: Removed this, but add redundant logic for validation of function and module.

modules/sdk-coin-apt/src/lib/transactionBuilder/customTransactionBuilder.ts
shivam-bitgo: will address this in followup PR

baltiyal (approved):

raj-bitgo (approved):

PR #6787, opened 2025-08-22
fix: fix wallet proofs class url (change wallet to wallets)

yash-joshi-bitgo:
### Description:
Fix WalletProof class URL (replace "wallet" with "wallets").

### Type of Change: Bug Fix

Ticket: BTC-2404


zahin-mohammad (approved):

0xPrabh (approved):
LGTM

PR #6792, opened 2025-08-25
fix: contract abi for vechain staking

Gautam2305:
Ticket: SC-2596


Vijay-Jagannathan (approved):
lgtm, what about unstake ?

PR #6795, opened 2025-08-25
feat: add ofc statics for kaia, sonic, hypeevm, seievm, apechain

Venkat-Annavazzala:
Ticket: WIN-6915

### DIFF
--- a/modules/statics/src/coins/ofcCoins.ts
+++ b/modules/statics/src/coins/ofcCoins.ts
@@ -156,6 +156,58 @@ export const ofcCoins = [
     UnderlyingAsset.WEMIX,
     CoinKind.CRYPTO
   ),
+  ofc('5405ad60-40c2-4789-b6b3-79265f167669', 'ofckaia', 'Kaia', 18, UnderlyingAsset.KAIA, CoinKind.CRYPTO),
+  tofc('1d41e93a-7de8-4e2e-a8bd-adc22cbe83e6', 'ofctkaia', 'Kaia Testnet', 18, UnderlyingAsset.KAIA, CoinKind.CRYPTO),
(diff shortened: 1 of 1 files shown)


modules/statics/src/coins/ofcCoins.ts
```diff
@@ -156,6 +156,58 @@ export const ofcCoins = [
     UnderlyingAsset.WEMIX,
     CoinKind.CRYPTO
   ),
+  ofc('5405ad60-40c2-4789-b6b3-79265f167669', 'ofckaia', 'Kaia', 18, UnderlyingAsset.KAIA, CoinKind.CRYPTO),
+  tofc('1d41e93a-7de8-4e2e-a8bd-adc22cbe83e6', 'ofctkaia', 'Kaia Testnet', 18, UnderlyingAsset.KAIA, CoinKind.CRYPTO),
```
gianchandania, line 160: Generally the naming convention that we follow for Testnet coins is 

Testnet <Chain/Coin name>

So ideally it should be Testnet Kaia and similarly for others in this PR

Venkat-Annavazzala: I understand the convention and agree we should follow it. The coins in this PR are already being used in production, and renaming them would require changes in multiple places (e.g., mapTokenNetwork.ts, allCoinsAndTokens.ts, tkaia.ts, networks.ts). I’m happy to accommodate the change, but I’m a bit concerned it might break something unintentionally. Could you confirm if it’s safe to update these now, or would you prefer we keep the existing names for consistency and follow the convention strictly for new additions going forward?

gianchandania: ideally changing token names should not cause an issue afaik. Issue is caused when token symbol is changed. But as the coins are already in prod with the given names and you are just adding ofc versions I will approve this.

gianchandania (approved):

PR #6813, opened 2025-08-26
feat(utxo-core): add bip322 proof checker

davidkaplanbitgo:
BTC-2408

TICKET: BTC-2408


modules/utxo-core/src/bip322/utils.ts
```diff
@@ -33,3 +33,46 @@ export function getBip322ProofMessageAtIndex(psbt: utxolib.Psbt, inputIndex: num
   }
   return Buffer.from(proprietaryKeyVals[0].value);
 }
+
+/**
+ * Checks if a BIP322 proof.
```
OttoAllmendinger:
```
checks if the transaction contains a BIP322 proof

does not check if the proof is valid
```

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #6844, opened 2025-08-29
feat(sdk-core): add walletUtil module with UTXO message proof

davidkaplanbitgo:
Add utility for bulk signing BIP322 messages with a focus on Midnight claims. Includes MidnightMessageProvider for retrieving claim addresses and a generic message provider interface.

TICKET: BTC-2425

### DIFF
--- a/modules/sdk-core/src/bitgo/walletUtil/utxoMessageProof.ts
+++ b/modules/sdk-core/src/bitgo/walletUtil/utxoMessageProof.ts
@@ -1,0 +1,128 @@
+import * as utxolib from '@bitgo/utxo-lib';
+
+import { IWallet } from '../wallet/iWallet';
+import { Environments } from '../environments';
+
+const NUM_MESSAGES_PER_TRANSACTION = 2;
+const NUM_MESSAGES_PER_QUERY = 1000;
+
+type MessageInfo = {
+  message: string;
+  address: string;
+};
+
+type Claim = {
+  originWalletId: string;
+  status: string;
+  originAddress?: string;
+  allocationAmount: string;
+};
+
+type BulkMessageResponse = {
+  success: boolean;
+  numMessages: number;
+  transactions: Record<string, unknown>[];
+};
+
+export interface IMessageProvider {
+  /**
+   * Returns the messages and addresses that we want to sign. We call this function multiple times until there are no more
+   * messages. If there are no more messages, an empty array is returned. Note that we only return messages in sets of 200.
+   */
+  getMessagesAndAddressesToSign(): Promise<MessageInfo[]>;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/walletUtil/utxoMessageProof.ts
```diff
@@ -0,0 +1,102 @@
...
+export async function bulkSignBip322MidnightMessages(
+  wallet: IWallet,
+  message: string,
+  walletPassphrase?: string
+): Promise<{ success: boolean; numMessagesSigned: number }> {
+  const provider = new MidnightMessageProvider(wallet.bitgo, wallet.coin(), wallet.id(), message);
+  return bulkSignBip322MessagesWithProvider(provider, wallet);
+}
+
+export async function bulkSignBip322MessagesWithProvider(
```
bhavidhingra: This function should be agnostic of where the payload is coming from. It should just take that as an input and generate the transaction for signing.

modules/sdk-core/src/bitgo/walletUtil/utxoMessageProof.ts
```diff
@@ -0,0 +1,102 @@
...
+  constructor(private bitgo: BitGoBase, private coinName: string, private walletId: string, private message: string) {
+    this.messageCache = [];
+    this.network = utxolib.networks[coinName];
+    this.midnightClaimUrl = `https://app.bitgo${
+      utxolib.isTestnet(this.network) ? '-test' : ''
+    }.com/api/airdrop-claim/v1/midnight/claims/${coinName}/${walletId}`;
+  }
+
+  async getMessagesAndAddressesToSign(): Promise<MessageInfo[]> {
+    if (this.messageCache.length > 0) {
```
bhavidhingra, line 40: why not have this in a while loop iteration rather than recursion?

lcovar (approved):

margueriteblair (approved):

modules/sdk-core/src/bitgo/walletUtil/utxoMessageProof.ts
```diff
@@ -0,0 +1,128 @@
...
+      };
+    });
+    const toReturn = this.messageCache.splice(0, NUM_MESSAGES_PER_TRANSACTION);
+    return toReturn;
+  }
+}
+
+export async function bulkSignBip322MidnightMessages(
+  wallet: IWallet,
+  message: string,
```
bhavidhingra, line 95: message will different for every address of the wallet.
Message = "STAR {claim.allocationAmount} to {destinationAddress} {midnight TnC}"

The second param of this function should be `destinationAddress`.

PR #6845, opened 2025-08-29
fix(sdk-core): send staking tx signatures based on wallet config

mohammadalfaiyazbitgo:
- staking signatures for multi-sig wallets on coins that support both multisig & tss
- the send logic used the coin configuration to determine if calling 'send' would be required, but this does not work on coin/wallet pairs where both mpc and multisig are supported
- fix logic to use wallet configuration instead of coin configuration in these cases

TICKET: WP-000000


kisslove-dewangan (approved):
lgtm

noel-bitgo (approved):

PR #6846, opened 2025-08-29
feat(sdk-coin-sol): verification on transaction type

cpatino-intive:
TICKET: WP-5744


modules/sdk-coin-sol/test/unit/sol.ts
```diff
@@ -591,6 +591,55 @@ describe('SOL:', function () {
...
+
+      const validTransaction = await basecoin.verifyTransaction({
+        txParams,
+        txPrebuild,
+        memo,
+        durableNonce,
+        wallet: walletObj,
+      } as any);
+      validTransaction.should.equal(true);
+    });
```
mohammadalfaiyazbitgo: add a unit test with a an actual transaction that you get from wallet platform as well.

mohammadalfaiyazbitgo (changes requested):
can you also do the transaction type check for token enablement and use that in the token enablement flow?

mukeshsp (commented):
Please remove conflicts

PR #6847, opened 2025-08-29
feat(sdk-coin-hbar): verification on transaction type

cpatino-intive:
TICKET: WP-5746

### DIFF
--- a/modules/sdk-coin-hbar/src/hbar.ts
+++ b/modules/sdk-coin-hbar/src/hbar.ts
@@ -24,6 +24,7 @@ import {
   MultisigType,
   multisigTypes,
   AuditDecryptedKeyParams,
+  TransactionType,
 } from '@bitgo/sdk-core';
 import { BigNumber } from 'bignumber.js';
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -208,6 +209,23 @@ export class Hbar extends BaseCoin {
...
+   * @param txParamsType - The type from txParams
+   */
+  private validateTransactionTypeMatch(transactionType: TransactionType | undefined, txParamsType?: string): void {
+    if (transactionType === undefined || !txParamsType) {
+      return;
+    }
+
+    const transactionTypeStr = TransactionType[transactionType];
+
+    if (transactionTypeStr === 'Send' && txParamsType !== 'transfer') {
```
mullapudipruthvik, line 224: I don't understand, why we have `Send` as internal state but we expect user to send `transfer`

mullapudipruthvik (approved):

PR #6849, opened 2025-08-29
feat: add sequenceId and comment to custom transaction parameters

sachushaji:
add sequenceID and comment to
customTx

Ticket: TMS-1509


Phani024 (approved):
LGTM

PR #6860, opened 2025-09-02
fix: export bip322 types and utility functions

bhavidhingra:
This pull request makes a small change to the exports in the `modules/abstract-utxo/src/transaction/index.ts` file, adding support for exporting everything from the `bip322` module.

TICKET: COIN-5426

### DIFF
--- a/modules/abstract-utxo/src/transaction/index.ts
+++ b/modules/abstract-utxo/src/transaction/index.ts
@@ -3,3 +3,4 @@ export { explainTx } from './explainTransaction';
 export { parseTransaction } from './parseTransaction';
 export { verifyTransaction } from './verifyTransaction';
 export * from './fetchInputs';
+export * as bip322 from './bip322';


modules/abstract-utxo/src/transaction/index.ts
```diff
@@ -3,3 +3,4 @@ export { explainTx } from './explainTransaction';
 export { parseTransaction } from './parseTransaction';
 export { verifyTransaction } from './verifyTransaction';
 export * from './fetchInputs';
+export * from './bip322';
```
OttoAllmendinger:
```suggestion
export * as bip322 from './bip322';
```

OttoAllmendinger (changes requested):

modules/abstract-utxo/src/transaction/index.ts
bhavidhingra: done

OttoAllmendinger (approved):

davidkaplanbitgo (approved):

PR #6864, opened 2025-09-03
fix: use publicKey for cold wallet recovery

MohammedRyaan786:
TICKET: COIN-5418

In case of cold wallets, we should derive public key from BitGo key which is passed along with the params.


abhi-bitgo (approved):

borkds (approved):
Looks good

PR #6872, opened 2025-09-03
feat(sdk-core): EVM keyring wallet and address creation changes

nvjsr:
TICKET: COIN-5193

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -1251,6 +1251,8 @@ export class Wallet implements IWallet {
       baseAddress,
       allowSkipVerifyAddress = true,
       onToken,
+      referenceCoin,
+      referenceAddress,
     } = params;
 
     if (!_.isUndefined(chain)) {
@@ -1323,6 +1325,30 @@ export class Wallet implements IWallet {
       if (!_.isString(onToken)) {
         throw new Error('onToken has to be a string');
       }
+    }
+
+    // Validate EVM keyring params, referenceAddress is required and referenceCoin is optional for EVM keyring
+    if (!_.isUndefined(referenceAddress)) {
+      if (!_.isString(referenceAddress)) {
+        throw new Error('referenceAddress has to be a string');
+      }
+      if (!this.baseCoin.isEVM()) {
+        throw new Error('referenceAddress is only supported for EVM chains');
+      }
+      if (!this.baseCoin.isValidAddress(referenceAddress)) {
+        throw new Error('referenceAddress must be a valid address');
+      }
+      addressParams.referenceAddress = referenceAddress;
+
+      if (!_.isUndefined(referenceCoin)) {
+        if (!_.isString(referenceCoin)) {
+          throw new Error('referenceCoin has to be a string');
+        }
+        addressParams.referenceCoin = referenceCoin;
+      }
+    } else if (!_.isUndefined(referenceCoin)) {
+      // referenceCoin cannot be used without referenceAddress
+      throw new Error('referenceAddress is required when using referenceCoin for EVM keyring');
     }
 
     // get keychains for address verification
--- a/modules/sdk-core/src/bitgo/wallet/wallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallets.ts
@@ -101,8 +101,21 @@ export class Wallets implements IWallets {
       throw new Error('missing required string parameter label');
     }
 
-    // no need to pass keys for (single) custodial wallets
-    if (params.type !== 'custodial') {
+    // Validate referenceWalletId parameter
+    if (params.referenceWalletId) {
+      if (!_.isString(params.referenceWalletId)) {
+        throw new Error('invalid referenceWalletId argument, expecting string');
+      }
+      if (!this.baseCoin.isEVM()) {
+        throw new Error('referenceWalletId is only supported for EVM chains');
+      }
+    }
+
+    // For wallets with referenceWalletId, skip multisig validation as configuration is inherited
+    if (params.referenceWalletId) {
+      // Skip all multisig validation - configuration will be inherited from reference wallet
+    } else if (params.type !== 'custodial') {
+      // Standard validation for non-custodial wallets without referenceWalletId
       if (Array.isArray(params.keys) === false || !_.isNumber(params.m) || !_.isNumber(params.n)) {
         throw new Error('invalid argument');
       }
@@ -272,9 +285,10 @@ export class Wallets implements IWallets {
       throw new Error('missing required string parameter label');
     }
 
-    const { type = 'hot', label, passphrase, enterprise, isDistributedCustody } = params;
+    const { type = 'hot', label, passphrase, enterprise, isDistributedCustody, referenceWalletId } = params;
     const isTss = params.multisigType === 'tss' && this.baseCoin.supportsTss();
     const canEncrypt = !!passphrase && typeof passphrase === 'string';
+    const isEVMWithReference = this.baseCoin.isEVM() && referenceWalletId;
 
     const walletParams: SupplementGenerateWalletOptions = {
       label: label,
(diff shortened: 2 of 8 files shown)


modules/sdk-core/test/unit/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -21,7 +21,8 @@ import { loadWebAssembly } from '@bitgo/sdk-opensslbytes';
 
 const openSSLBytes = loadWebAssembly().buffer;
 
-describe('ecdsa tss', function () {
+describe('ecdsa tss', function (this: Mocha.Context) {
+  this.timeout(60000);
```
nvjsr, line 25: Added the timeout since it was failing with error `Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called;` It was successfully running locally but failing during PR build.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1325,6 +1327,36 @@ export class Wallet implements IWallet {
       }
     }
 
+    // Validate and handle referenceCoin for EVM keyring
+    if (!_.isUndefined(referenceCoin)) {
```
pritam-gembali: combine validations and check for all

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -101,8 +101,24 @@ export class Wallets implements IWallets {
...
-    if (params.type !== 'custodial') {
+    // Validate referenceWalletId parameter
+    if (params.referenceWalletId) {
+      if (!_.isString(params.referenceWalletId)) {
+        throw new Error('invalid referenceWalletId argument, expecting string');
+      }
+      if (!this.baseCoin.isEVM()) {
+        throw new Error('referenceWalletId is only supported for EVM chains');
+      }
+      if (!params.enterprise) {
```
pritam-gembali: this is not needed

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1325,6 +1327,36 @@ export class Wallet implements IWallet {
...
+      if (!this.baseCoin.isValidAddress(referenceAddress)) {
+        throw new Error('referenceAddress must be a valid address');
+      }
+      addressParams.referenceAddress = referenceAddress;
+    }
+
+    // For EVM keyring, both referenceCoin and referenceAddress should be provided together
+    if ((referenceCoin && !referenceAddress) || (!referenceCoin && referenceAddress)) {
+      throw new Error('For EVM keyring, both referenceCoin and referenceAddress must be provided together');
+    }
```
shivam-bitgo:
```suggestion
   if (!!referenceCoin !== !!referenceAddress) {
  throw new Error('For EVM keyring, both referenceCoin and referenceAddress must be provided together');
}
```

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -297,15 +319,63 @@ export class Wallets implements IWallets {
...
 
+    // Validate referenceWalletId for EVM keyring
+    if (!_.isUndefined(referenceWalletId)) {
+      if (!_.isString(referenceWalletId)) {
+        throw new Error('invalid referenceWalletId argument, expecting string');
+      }
+      if (!this.baseCoin.isEVM()) {
+        throw new Error('referenceWalletId is only supported for EVM chains');
+      }
+      if (!enterprise) {
```
pritam-gembali: we don't need this

modules/sdk-core/src/bitgo/wallet/wallet.ts
shivam-bitgo: also can we move this up for early return?

modules/sdk-core/src/bitgo/wallet/wallet.ts
nvjsr: Added the check for referenceAddress only and referenceCoin is optional if present then will have validation and add in params

modules/sdk-core/src/bitgo/wallet/wallets.ts
nvjsr: removed

modules/sdk-core/src/bitgo/wallet/wallet.ts
nvjsr: this is not needed now, as only referenceCoin is optional it should not be present always with referenceAddress

modules/sdk-core/src/bitgo/wallet/wallets.ts
nvjsr: removed

pritam-gembali (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1325,6 +1327,30 @@ export class Wallet implements IWallet {
...
+        if (!_.isString(referenceCoin)) {
+          throw new Error('referenceCoin has to be a string');
+        }
+        addressParams.referenceCoin = referenceCoin;
+      }
+    } else if (!_.isUndefined(referenceCoin)) {
+      // referenceCoin cannot be used without referenceAddress
+      throw new Error('referenceAddress is required when using referenceCoin for EVM keyring');
+    }
+
```
pranavjain97, line 1353: Can you instead do a `if(_.isUndefined(referenceCoin) && !_.isUndefined(referenceAddress)) { throw error }`
That's more readable.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1325,6 +1327,30 @@ export class Wallet implements IWallet {
       }
     }
 
+    // Validate EVM keyring params, referenceAddress is required and referenceCoin is optional for EVM keyring
```
pranavjain97, line 1330: if `referenceCoin` by itself isn't valid, it should't be an independent option. Let's instead have `referenceAddressEvmKeyRing` and have the address and coin as nested fields.

shivam-bitgo (approved):

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -101,8 +101,21 @@ export class Wallets implements IWallets {
...
+      if (!this.baseCoin.isEVM()) {
+        throw new Error('referenceWalletId is only supported for EVM chains');
+      }
+    }
+
+    // For wallets with referenceWalletId, skip multisig validation as configuration is inherited
+    if (params.referenceWalletId) {
+      // Skip all multisig validation - configuration will be inherited from reference wallet
+    } else if (params.type !== 'custodial') {
+      // Standard validation for non-custodial wallets without referenceWalletId
```
pranavjain97, line 118: We shoudn't be having a `referenceId` as a top level param that is EVM based only. Can we atleast name it `evmKeyRingReferenceWalletId`? Coin-specific logic shoudn't be leaking in generic add wallet methods, can we create a coin-specific util in eth-lib that can take care of this validation?

pranavjain97 (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
nvjsr: Adding changes in followup PR for same

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -297,15 +316,59 @@ export class Wallets implements IWallets {
...
+      const result: WalletWithKeychains = {
+        wallet: new Wallet(this.bitgo, this.baseCoin, newWallet),
+        userKeychain: userKey,
+        backupKeychain: backupKey,
+        bitgoKeychain: bitgoKey,
+        responseType: 'WalletWithKeychains',
+      };
+
+      return result;
+    }
```
pranavjain97, line 370: This should also be an EVM lib util i.e something like `createEvmKeyRingWallet`.

pranavjain97 (commented):
Have we updated the express dev docs anywhere for the new fields added for wallet/address creation?

PR #6873, opened 2025-09-03
feat(dot): integrate the Westend AssetHub chain for tdot

arvind-bg:
Assets on Westend have been migrated to Westend AssetHub. As a result, transaction broadcasts are failing in the lower environments.
This change integrates the Westend Assethub chain to be used for tdot. It updates RPC and metadata information. 
Test cases failing because of the inability to parse Westend (not Westend AssetHub) transactions have been skipped. They will be fixed in future commits. Post release, the corresponding transactions will be replaced by Westend AssetHub transactions to fix the tests.

Ticket: WIN-7059


modules/sdk-coin-dot/test/unit/transaction.ts
ajays97, line 1: Can this be tracked in a ticket in the backlog?

modules/statics/src/networks.ts
```diff
@@ -521,15 +521,27 @@ class Polkadot extends Mainnet implements DotNetwork {
...
+// class PolkadotTestnet extends Testnet implements DotNetwork {
+//   name = 'Westend';
+//   family = CoinFamily.DOT;
+//   explorerUrl = 'https://westend.subscan.io/extrinsic/';
+//   specName = 'westend' as PolkadotSpecNameType;
+//   genesisHash = '0xe143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e';
+//   specVersion = 9430;
+//   chainName = 'Westend';
+//   txVersion = 22;
+// }
```
ajays97: Do we need these commented, why?

modules/sdk-coin-dot/test/unit/transaction.ts
arvind-bg: Yes, https://bitgoinc.atlassian.net/browse/WIN-7065

modules/statics/src/networks.ts
arvind-bg: Leaving it as-is causes lint issues. I wanted to retain it for the time being just in case we find it useful. We can remove it when we do this exercise for prod.

ajays97 (dismissed):
Looks good to me

hitansh-madan (dismissed):

modules/statics/src/networks.ts
mmcshinsky-bitgo: We should remove these as we can always pull them from git history anyway instead of littering the code with commented out code.

modules/sdk-coin-dot/test/unit/transactionBuilder/withdrawUnstakedBuilder.ts
```diff
@@ -77,7 +77,7 @@ describe('Dot WithdrawUnstaked Builder', () => {
       should.deepEqual(txJson.eraPeriod, 64);
     });
 
-    it('should build from raw signed tx', async () => {
+    xit('should build from raw signed tx', async () => {
```
mmcshinsky-bitgo, line 80: Shouldn't we be fixing these tests as part of this PR effort, not skipping them for later?

mmcshinsky-bitgo: If transactions are failing as part of this integration, what are the ramifications of usage in apps. Are we going to break staging and test by not fixing it now?

arvind-bg: Staging and test are broken currently, and have been in a broken state for months. The assets on Westend (our previous stage/test chain) moved to AssetHub, so all calls need to go to AssetHub instead of Westend.
We released support for AssetHub in Indexer recently. So, indexing of transactions and blocks is working.
However, our SDK is only talking to Westend currently. So, withdrawals are not working at all. This change is meant to fix that.
These tests use transactions from Westend. We will not be able to parse them with the RpcMetadata from Assethub (which is the change this PR makes). We will have to replace them with transactions from Assethub.
We do not have any internal transactions to use since we have not been able to build any until now. This change should fix that. Dot has many transaction types and we have tests for many of them. However, the explorer is not suited to filter by transaction types. Otherwise, we could have filtered via the explorer and use them in our test suites.
Once, we are able to verify that this change is working, we should be able to build and broadcast transactions across transaction types,  verify they succeed on chain, and then use them as part of our test suite as well.

modules/statics/src/networks.ts
arvind-bg: This has been removed.

akarath (approved):
lgtm

Ranjna-G (approved):

hitansh-madan (approved):

kisslove-dewangan (approved):
lgtm based on codeowners

PR #6883, opened 2025-09-04
fix(sdk-coin-sol): create broadcast tx for 2022 token

abhishekagrawal080:
TICKET: WIN-6819

### DIFF
--- a/modules/sdk-coin-sol/src/lib/tokenTransferBuilder.ts
+++ b/modules/sdk-coin-sol/src/lib/tokenTransferBuilder.ts
@@ -162,14 +162,14 @@ export class TokenTransferBuilder extends TransactionBuilder {
         let tokenAddress: string;
         let tokenName: string;
         let programId: string | undefined;
-        if (recipient.tokenAddress && recipient.programId) {
-          tokenName = recipient.tokenName;
-          tokenAddress = recipient.tokenAddress;
-          programId = recipient.programId;
-        } else if (coin) {
+        if (coin) {
           tokenName = coin.name;
           tokenAddress = coin.tokenAddress;
           programId = coin.programId;
+        } else if (recipient.tokenAddress && recipient.programId) {
+          tokenName = recipient.tokenName;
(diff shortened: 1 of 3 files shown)


abhi-bitgo (dismissed):

modules/sdk-coin-sol/src/lib/tokenTransferBuilder.ts
```diff
@@ -162,11 +161,7 @@ export class TokenTransferBuilder extends TransactionBuilder {
         let tokenAddress: string;
         let tokenName: string;
         let programId: string | undefined;
-        if (recipient.tokenAddress && recipient.programId) {
-          tokenName = recipient.tokenName;
-          tokenAddress = recipient.tokenAddress;
-          programId = recipient.programId;
```
mukeshsp, line 168: Why are we removing the uses of option param? this was added intentionally

abhishekagrawal080: recipient comes from _createATA from transactionBuilder that is getting set from createAssociatedTokenAccount as well from that passes incorrect programId check line number 65 in this file

modules/sdk-coin-sol/src/lib/tokenTransferBuilder.ts
```diff
@@ -162,11 +161,7 @@ export class TokenTransferBuilder extends TransactionBuilder {
         let tokenAddress: string;
         let tokenName: string;
         let programId: string | undefined;
-        if (recipient.tokenAddress && recipient.programId) {
```
Ranjna-G, line 165: Would you please brief on how removing this condition will help with the issue ? (not sweeping tokens to the destination address without ATA) ?

mukeshsp (approved):

veetragjain (approved):
Lgtm

nvrakesh06 (approved):

krupanand-bitgo (approved):

PR #6888, opened 2025-09-04
fix(sdk-core): override test timeout of before()

zhongxishen:
TICKET: HSM-1084

### DIFF
--- a/modules/sdk-core/test/unit/account-lib/mpc/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/test/unit/account-lib/mpc/tss/ecdsa/ecdsa.ts
@@ -21,13 +21,14 @@ import { loadWebAssembly } from '@bitgo/sdk-opensslbytes';
 
 const openSSLBytes = loadWebAssembly().buffer;
 
-describe('ecdsa tss', function (this: Mocha.Context) {
-  this.timeout(60000);
+describe('ecdsa tss', function () {
   const ecdsa = new Ecdsa();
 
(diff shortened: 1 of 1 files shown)


modules/sdk-core/test/unit/account-lib/mpc/tss/ecdsa/ecdsa.ts
```diff
@@ -21,13 +21,14 @@ import { loadWebAssembly } from '@bitgo/sdk-opensslbytes';
...
-describe('ecdsa tss', function (this: Mocha.Context) {
-  this.timeout(60000);
+describe('ecdsa tss', function () {
   const ecdsa = new Ecdsa();
 
   let signCombine1: SignCombineRT, signCombine2: SignCombineRT;
 
   before('generate key and sign phase 1 to 4', async function () {
+    // In some execution environments (e.g. CI), generateNtilde below may take longer at times.
+    this.timeout('40s');
```
zhongxishen, line 31: refining a fix that has been in place above

- move timeout override from test suite to the `before` hook since it's the part that may take a long time.
- use more human readable form for the timeout duration.
- repeated running unit test suites in CI, there has been no failure triggered by timeout here with 40 seconds specified (there was for 30 seconds), so it seems to be an appropriate value for the current CI environment.

islamaminBitGo (approved):
🕐

shivam-bitgo (approved):
lgtm

kamleshmugdiya (approved):

PR #6889, opened 2025-09-04
fix(sdk-coin-ada): verify consolidation txHex

mohammadalfaiyazbitgo:
- verify unsigned txHex consolidates funds to base address

TICKET: WP-5730

### DIFF
--- a/modules/sdk-coin-ada/src/ada.ts
+++ b/modules/sdk-coin-ada/src/ada.ts
@@ -122,7 +122,7 @@ export class Ada extends BaseCoin {
   async verifyTransaction(params: VerifyTransactionOptions): Promise<boolean> {
     try {
       const coinConfig = coins.get(this.getChain());
-      const { txPrebuild: txPrebuild, txParams: txParams } = params;
+      const { txPrebuild: txPrebuild, txParams: txParams, verification, wallet } = params;
       const transaction = new Transaction(coinConfig);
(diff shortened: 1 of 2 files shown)


mukeshsp (approved):

modules/sdk-coin-ada/src/ada.ts
```diff
@@ -122,7 +122,7 @@ export class Ada extends BaseCoin {
   async verifyTransaction(params: VerifyTransactionOptions): Promise<boolean> {
     try {
       const coinConfig = coins.get(this.getChain());
-      const { txPrebuild: txPrebuild, txParams: txParams } = params;
+      const { txPrebuild: txPrebuild, txParams: txParams, verification, wallet } = params;
```
kamleshmugdiya, line 125: do we need to pass entire wallet object?
maybe we can pass base/root address with `consolidationToBaseAddress` in `VerificationOptions`.

mohammadalfaiyazbitgo: `wallet` is one of the required  parameters of the method https://github.com/BitGo/BitGoJS/blob/be4a440cbc37bc28d2550158021ea9a18e1e9e6f/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts#L182

kamleshmugdiya (approved):

PR #6899, opened 2025-09-05
feat(sdk-coin-sol): prevent blind signing sol token enablement

mtexeira-simtlix:
TICKET: WP-5744

Contents:
- Blind signing guard for token enablement on type and outputs.
- 3 new Unit tests
- Script showcasing how blind signing happens (in the ticket comments).


modules/sdk-coin-sol/test/unit/sol.ts
```diff
@@ -3215,4 +3215,52 @@ describe('SOL:', function () {
...
+            txPrebuild: tamperedTxPrebuild,
+            wallet,
+          } as unknown as SolVerifyTransactionOptions)
+      ),
+        { message: 'Tx type "Send" does not match expected txParams type "enabletoken"' };
+    });
+
+    it('should throw an error when outputs is not empty', async function () {
+      const { txParams, txPrebuildRaw, tamperedTxHexWithOutputs, walletData } = testData.enableTokenFixtures;
+      const tamperedTxPrebuild = { ...txPrebuildRaw, txHex: tamperedTxHexWithOutputs };
```
mtexeira-simtlix, line 3297: I'm having some issues to generate a txHex that comes with some outputs, I'll try again tomorrow but if you have any advice it's gonna be really useful, thanks @mohammadalfaiyazbitgo

mtexeira-simtlix: Ignore that, I reused a transfer tx and removed the outputs by hand in one test and left those in another so we could check for both. Any other way of inject the outputs and then prebuild keep stripping the outputs array.

TL;DR: solved, we've a test for that. Thanks!

modules/sdk-coin-sol/test/unit/sol.ts
```diff
@@ -3215,4 +3215,52 @@
...
+      // tamperedTxPrebuild has the type Send instead of 'enabletoken'
+      await assert.rejects(
+        async () =>
+          await basecoin.verifyTransaction({
+            txParams,
+            txPrebuild: tamperedTxPrebuild,
+            wallet,
+          } as unknown as SolVerifyTransactionOptions)
+      ),
+        { message: 'Tx type "Send" does not match expected txParams type "enabletoken"' };
```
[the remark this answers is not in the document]
mtexeira-simtlix: Solved.

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -40,14 +45,10 @@ import {
...
-  multisigTypes,
-  AuditDecryptedKeyParams,
-  PopulatedIntent,
-  PrebuildTransactionWithIntentOptions,
 } from '@bitgo/sdk-core';
 import { auditEddsaPrivateKey, getDerivationPath } from '@bitgo/sdk-lib-mpc';
 import { BaseNetwork, CoinFamily, coins, BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import { TOKEN_2022_PROGRAM_ID, TOKEN_PROGRAM_ID } from '@solana/spl-token';
 import * as _ from 'lodash';
 import * as request from 'superagent';
```
mohammadalfaiyazbitgo: nit: let's move external deps to the top of the import list.

modules/sdk-coin-sol/test/unit/sol.ts
```diff
@@ -3215,4 +3216,97 @@ describe('SOL:', function () {
...
+    let explainStub: sinon.SinonStub;
+    beforeEach(() => {
+      originalExplain = Transaction.prototype.explainTransaction;
+    });
+
+    afterEach(() => {
+      // Restore the stub to avoid affecting other tests in cascade as explainTransaction
+      // is used in multiple places
+      if (explainStub) explainStub.restore();
+    });
```
mohammadalfaiyazbitgo, line 3231: let's not stub this.

mohammadalfaiyazbitgo (changes requested):

modules/sdk-coin-sol/test/unit/sol.ts
mohammadalfaiyazbitgo: please write tests with real hex values.

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -261,6 +289,9 @@ export class Sol extends BaseCoin {
     transaction.fromRawTransaction(rawTxBase64);
     const explainedTx = transaction.explainTransaction();
 
+    this.verifyNoOutputs(txParams.type, explainedTx);
+    this.verifyTxType(txParams.type, explainedTx.type);
```
mohammadalfaiyazbitgo, line 293: let's wrap this around a verification option, https://github.com/BitGo/BitGoJS/blob/a5144105e413769e2509d8e2904636f4d59b2d3d/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts#L165
which would default to true if unspecified.

mohammadalfaiyazbitgo (changes requested):
Also consider verifying that the request token is the one being enabled in the txHex.

modules/sdk-coin-sol/test/unit/sol.ts
mtexeira-simtlix: I tried multiple things and explain keeps removing the outputs so I opted for this way, do you have any other way to force outputs inside the txHex?

https://github.com/BitGo/BitGoJS/pull/6899#discussion_r2331248239

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: Mmmm not sure about that tbh, so far I'm starting on XRP and there're things to verify that are not concerns to SOL.

I could try to work the way up in a follow up when we get both XRP and SOL merged to at least extract things in common or see what's an edge case and avoid delaying the delivering of the security features implementation.

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: Sure! will change that although pretty sure the code formatter will move it back haha.

modules/sdk-coin-sol/test/unit/sol.ts
mtexeira-simtlix: One way of having outputs is to set the tx type as a Send, but if you do that the verify that will be triggered is for the type instead of the outputs. 
With the stub the thing that matters is that we're forcing an outputs array when it shouldn't be one and verify catchs it later as expected, doesn't matter how the attacker add it, verify always catchs the extra.

mtexeira-simtlix:
> Also consider verifying that the request token is the one being enabled in the txHex.

@mohammadalfaiyazbitgo txHex doesn't comes with the token, how are we gonna check that?
This is the payload I shared with you some days ago, doesn't matter if you're choosing the ATA or an extra generated address:

`{
    "displayOrder": [
        "id",
        "type",
        "blockhash",
        "durableNonce",
        "outputAmount",
        "changeAmount",
        "outputs",
        "changeOutputs",
        "fee",
        "memo"
    ],
    "id": "UNAVAILABLE",
    "type": "AssociatedTokenAccountInitialization",
    "changeOutputs": [],
    "changeAmount": "0",
    "outputAmount": "0",
    "outputs": [],
    "fee": {
        "fee": "UNAVAILABLE"
    },
    "blockhash": "GNU8DULvY5HRFU9F1FC2F5upB3bsJyQjukSvdpEh8aSH",
    "durableNonce": {
        "walletNonceAddress": "38WoASL5WQoEEBi7hLLUa5w5RAQC71cwgkpmptE1ZBPj",
        "authWalletAddress": "2vbA4Rq4Yf1oRD35SmuiRnQhYxQFSxpkxBJXCmiqYfKk"
    }
}`

That's the explainedTx payload logged directly from the UI to give you an idea.

I could check if there is a "token" or "coin" or something prop added to the object but I may need the name of the prop to check although if should come as empty, may be easier to check that "if enabletoken then txPrebuildHex coin / token should be undefined in case that there is a coin / token there".

Does that makes sense?

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: betrayed by code formatter, forgot to hit "save without formatting" will submit the fix in a minute, apologies.

mtexeira-simtlix: image
Done.

modules/sdk-coin-sol/test/unit/sol.ts
mtexeira-simtlix: (forgot to clarify, this is for only 1 of 3-4 tests, all the others are with proper real hexes)

mtexeira-simtlix: Addressed on https://github.com/BitGo/BitGoJS/pull/6966 
(no stubs and tests with real hex values)

PR #6900, opened 2025-09-08
feat(statics): add GATENet (GATE) ERC20 token

nayandas190:
TICKET: WIN-6923

### DIFF
--- a/modules/statics/src/coins/erc20Coins.ts
+++ b/modules/statics/src/coins/erc20Coins.ts
@@ -7796,7 +7796,7 @@ export const erc20Coins = [
   erc20(
     '8c3142c8-0cd9-42f2-8849-3aca3da1a214',
     'gate',
-    'GATE',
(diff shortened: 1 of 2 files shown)


modules/statics/src/coins/erc20Coins.ts
```diff
@@ -7795,11 +7795,11 @@ export const erc20Coins = [
   ),
   erc20(
     '8c3142c8-0cd9-42f2-8849-3aca3da1a214',
-    'gate',
```
at31416, line 7798: I don't think we can directly rename an asset which was already onboarded before.

nayandas190: Thanks! Corrected now.

yogeshwar-bitgo (approved):
LGTM

prithvishet2503 (approved):

PR #6916, opened 2025-09-09
feat: upgrade the express version

abhi-bitgo:
ticket: DX-1559


abhi-bitgo:
Also, resolves the [DX-1561](https://bitgoinc.atlassian.net/browse/DX-1561) issue

PR #6919, opened 2025-09-09
test: Token naming convention for polygon

0xPrabh:
Ticket: [COIN-5570]

[COIN-5570]: https://bitgoinc.atlassian.net/browse/COIN-5570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/statics/test/unit/tokenNamingConvention.ts
+++ b/modules/statics/test/unit/tokenNamingConvention.ts
@@ -3,10 +3,11 @@ import { cosmosTokens } from '../../src/coins/cosmosTokens';
 import { avaxTokens } from '../../src/coins/avaxTokens';
 import { bscTokens } from '../../src/coins/bscTokens';
 import { nep141Tokens } from '../../src/coins/nep141Tokens';
+import { polygonTokens } from '../../src/coins/polygonTokens';
 import { NetworkType } from '../../src/networks';
 
 describe('Token Naming Convention Tests', function () {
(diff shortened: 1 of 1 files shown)


ravibitgo (approved):

modules/statics/test/unit/tokenNamingConvention.ts
```diff
@@ -319,8 +321,15 @@ describe('Token Naming Convention Tests', function () {
...
   it('should ensure all token names are lowercase', function () {
+    // List of tokens that are exceptions to the lowercase rule
+    const lowercaseExceptions: string[] = ['tpolygon:BitGoTest'];
+
     allTokens.forEach((token) => {
       const tokenName = token.name;
+      // Skip known exceptions
+      if (lowercaseExceptions.includes(tokenName)) {
+        return;
+      }
```
bitgoAaron, line 332: can you provide context here and why the original lowercase invariant was applied, and why this token should be excluded from that check? Not seeing much in the commit or JIRA ticket explaining why we should allow the exception here

0xPrabh: I am just adding checks for the new tokens we are onboarding. For tokens that are already onboarded, we shouldn't change their token names(thus adding into knownException list at multiple places). Ideally, we should follow the convention of using lowercase only for token names from now on, and this test case will help enforce that.

prithvishet2503 (approved):

PR #6931, opened 2025-09-10
feat(sdk-coin-ada): Add fingerprint as the contract address

Ranjna-G:
Ticket: WIN-7162

### DIFF
--- a/modules/statics/src/coins/adaTokens.ts
+++ b/modules/statics/src/coins/adaTokens.ts
@@ -8,10 +8,10 @@ export const adaTokens = [
     'a7678172-84b3-4c7e-ac46-9875d23a1cb7',
     'tada:water',
     'Test ADA Token',
-    6,
+    0, // Tokens are not divisible - https://cardano-ledger.readthedocs.io/en/latest/explanations/token-bundles.html
     '2533cca6eb42076e144e9f2772c390dece9fce173bc38c72294b3924',
     'WATER',
-    '5741544552',
+    'asset1n69xf60d0760xvn8v2ffd5frvsm0cl2r8hfjf6',
     UnderlyingAsset['tada:water'],
     ADA_TOKEN_FEATURES,
     Networks.test.ada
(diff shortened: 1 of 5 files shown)


abhishekagrawal080 (dismissed):
lgtm

abhi-bitgo (dismissed):

modules/statics/src/coins/adaTokens.ts
```diff
@@ -6,13 +6,12 @@ import { Networks } from '../networks';
 export const adaTokens = [
   tadaToken(
     'a7678172-84b3-4c7e-ac46-9875d23a1cb7',
-    'tada:water',
+    'tada:asset1n69xf60d0760xvn8v2ffd5frvsm0cl2r8hfjf6',
     'Test ADA Token',
-    6,
+    0,
```
mukeshsp: nit: please add the source for this in the comment

mukeshsp (dismissed):

modules/statics/src/coins/adaTokens.ts
```diff
@@ -8,10 +8,10 @@ export const adaTokens = [
     'a7678172-84b3-4c7e-ac46-9875d23a1cb7',
     'tada:water',
     'Test ADA Token',
-    6,
+    0, // Tokens are not divisible - https://cardano-ledger.readthedocs.io/en/latest/explanations/token-bundles.html
```
mukeshsp, line 11: This will cause the issue, we use the smallest unit to calculate the total amount.

mukeshsp: lets use 1 instead of zero

mukeshsp (approved):

modules/statics/src/coins/adaTokens.ts
Ranjna-G: Tokens are represented as integers in ada

abhishekagrawal080 (approved):

PR #6942, opened 2025-09-10
feat(express): migrate signpayload to typed routes

danielzhao122:
Typed router for signPayload
Added UT: "express.ofc.signPayload" in test/unit/typedRoutes/decode.ts
TICKET: WP-5443


modules/express/src/typedRoutes/api/v2/ofcSignPayload.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @property walletId – non-empty string identifying the OFC wallet
+ * @property payload  – JSON value to sign
+ * @property walletPassphrase – optional passphrase used to decrypt the user key (string)
+ */
+export const PostOfcSignPayload = httpRoute({
+  path: '/api/v2/ofc/signPayload',
+  method: 'POST',
+  request: httpRequest({ body: OfcSignPayloadBody }),
+  response: {
+    200: t.type({ payload: t.string, signature: t.string }),
```
zahin-mohammad: Please define the response codec as a variable and add some jsdoc's regarding their fields. 

The codecs need to be their own variables due to how the code generation works.

zahin-mohammad: Also to ensure this works as we expect it to, lets find an existing test for `signPayload` and use the codec to decode the response

modules/express/src/typedRoutes/api/v2/ofcSignPayload.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Json, NonEmptyString } from 'io-ts-types';
+import { httpRoute, httpRequest, optional } from '@api-ts/io-ts-http';
+
+/**
+ * Sign an arbitrary payload using an OFC trading account key.
+ * @operationId express.ofc.signPayload
+ * POST /api/v2/ofc/signPayload
+ */
+export const OfcSignPayloadBody = {
+  walletId: NonEmptyString,
```
zahin-mohammad, line 11: Missing jsdocs. In this case, since there is no platform.yaml to reference, lets use the jsdocs from the underlying code.

modules/express/src/clientRoutes.ts
```diff
@@ -1649,8 +1642,8 @@ export function setupAPIRoutes(app: express.Application, config: Config): void {
     promiseWrapper(handleV2AcceptWalletShare)
   );
 
-  // sign arbitrary payloads w/ trading account key
-  app.post(`/api/v2/ofc/signPayload`, parseBody, prepareBitGo(config), promiseWrapper(handleV2OFCSignPayload));
+  // sign arbitrary payloads w/ trading account key (migrated to typed route express.ofc.signPayload)
+  // app.post(`/api/v2/ofc/signPayload`, parseBody, prepareBitGo(config), promiseWrapper(handleV2OFCSignPayload));
```
zahin-mohammad:
```suggestion
```

zahin-mohammad (changes requested):
Looks like there is some CI errors as well

Taseen08 (dismissed):

modules/express/package.json
```diff
@@ -51,6 +51,7 @@
     "dotenv": "^16.0.0",
     "express": "4.21.2",
     "io-ts": "npm:@bitgo-forks/io-ts@2.1.4",
+    "io-ts-types": "^0.5.19",
```
zahin-mohammad, line 54: did you re-generate the lock file?

modules/express/src/typedRoutes/api/v2/ofcSignPayload.ts
```diff
@@ -0,0 +1,41 @@
+import * as t from 'io-ts';
+import { Json, NonEmptyString } from 'io-ts-types';
+import { httpRoute, httpRequest, optional } from '@api-ts/io-ts-http';
+
+/**
+ * Sign an arbitrary payload using an OFC trading account key.
+ * @operationId express.ofc.signPayload
```
zahin-mohammad: The operation id jsdoc should be part of the docs for `PostOfcSignPayload`

zahin-mohammad (changes requested):

modules/express/package.json
danielzhao122: I added this to use the `JSON` type from `io-ts-types`. Should I change the type and remove this import?

Taseen08: Should be fine I believe, since other modules are importing it too.

zahin-mohammad (commented):
merge conflicts. re-request a review when ready

PR #6958, opened 2025-09-11
Wp 5782/fix near token enablement validation

SimonVutovB:
## Summary
Added token enablement validation and tests

## Changes
- Inside, `verifyTransaction` in near.ts of sdk-coin-near, validate the txHex is a valid token enablement transaction for the specified token and does not have additional transactions embedded.

TICKET: WP-5782


modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1068,43 @@ export class Near extends BaseCoin {
...
+    explainedTx: TransactionExplanation,
+    txParams: VerifyTransactionOptions['txParams'],
+    txHex: string
+  ): void {
+    // Parse transaction from hex to validate actual transaction matches what user sees
+    const coinConfig = coins.get(this.getChain());
+    const freshTx = new Transaction(coinConfig);
+    freshTx.fromRawTransaction(txHex);
+    const freshTxData = freshTx.toJson();
+    const originalTxData = transaction.toJson();
```
mtexeira-simtlix: Hey @SimonVutovB you already have the freshTxData on the explainedTx. https://github.com/BitGo/BitGoJS/blob/b81a0d0f7162add92dbbaebe404e48dedf349e45/modules/sdk-coin-near/src/near.ts#L1001

since you're already passing down the explainedTx on validateTokenEnablementTransaction, just delete this part of the code and use the explainedTx instead of freshTxData.

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1068,43 @@ export class Near extends BaseCoin {
...
+
+    // Verify key transaction fields match to prevent tampering
+    if (
+      freshTxData.signerId !== originalTxData.signerId ||
+      freshTxData.receiverId !== originalTxData.receiverId ||
+      freshTxData.publicKey !== originalTxData.publicKey ||
+      freshTxData.actions.length !== originalTxData.actions.length
```
mtexeira-simtlix: Could you move every check that's unrelated to their own validateFunctions please?

I dunno about near but as an example if this applies:
```suggestion
     this.validateSigner(userParams, explainedTx);
     this.validateReceivers(userParams, explainedTx);
     ....
```

mtexeira-simtlix: In the prev example, userParams is the same as originalTxData, and explainedTx is the thing that I explained in the prev comment (the freshTxData that's already decoded).

mtexeira-simtlix: An example for validateSigner (same pattern for the others):
```suggestion
     validateSigner(userParams, explainedTx): void{
     // do nothing in case that this isn't a token enablement
     if (userParams.type !== 'enablementToken') return; // check the type on my PRs i always forgot if it's enabletoken, enablementToken, etc
     
     if (userParams.signerId !== explainedTx.signerId) throw new Error(`Error on token enablements: signers are not the same, expected ${userParams.signerId} but got ${explainedTx.signerId}`)
     
```

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1068,43 @@ export class Near extends BaseCoin {
...
+
+    // Validate addresses match between parameters and explained transaction
+    if (txParams.recipients && explainedTx.outputs) {
+      const expectedAddresses = txParams.recipients.map((r) => r.address);
+      const explainedAddresses = explainedTx.outputs.map((o) => o.address);
+
+      for (const addr of expectedAddresses) {
+        if (!explainedAddresses.includes(addr)) {
+          throw new Error(`Address mismatch: ${addr}`);
+        }
```
mtexeira-simtlix: Please move this snippet to a different validate/verifyFunction as the comments above, also, you can check if outputs matchs recipients on a single loop (O(n)):
```suggestion
   // remember to check that you're on a token enablement, except if this was checked for the general case or if this is always checked for other tx types. 
   if (txParams.recipients.length !== explainedTx.outputs.length) throw new Error('Error on token enablements: output doesn't match');
   
   const diffAddrs = explainedAddresses.filter((addr, index) => addr !== txParams.recipients[index].address);
   
   if (diffAddrs.length > 0) throw new Error(`Error on token enablements: address doesn't match for ${CONCAT your addresses code here}`);
```

mtexeira-simtlix (changes requested):
A few changes requested but you got a general first idea, good job! let me know if you need further clarification on any of the mentioned points @SimonVutovB 

Thanks!

modules/sdk-coin-near/src/near.ts
SimonVutovB: Hi @mtexeira-simtlix , I appreciate the suggestion, I think explainedTx (TransactionExplanation) doesn't contain the raw transaction fields like signerId, publicKey, and actions - it only has the processed display data like outputs and type. Let me know if I'm missing something, or if there is a better way of doing it.

mtexeira-simtlix: I'm summoning @mohammadalfaiyazbitgo for this one, it seems that we may need to add more data to the explained version of the tx here as we did in Solana.

@SimonVutovB we could talk about this during tomorrow's sync, remind us if we forgot please. Thanks!

SimonVutovB: Ok Sounds good, thank you.

mtexeira-simtlix:
@SimonVutovB for this issue that you're getting on the CI/CD steps, just rebase master since your dep versions are out of date:

image

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1068,83 @@ export class Near extends BaseCoin {
...
+    explainedTx: TransactionExplanation,
+    txParams: VerifyTransactionOptions['txParams'],
+    txHex: string
+  ): void {
+    // Parse transaction from hex to validate actual transaction matches what user sees
+    const coinConfig = coins.get(this.getChain());
+    const freshTx = new Transaction(coinConfig);
+    freshTx.fromRawTransaction(txHex);
+    const freshTxData = freshTx.toJson();
+    const originalTxData = transaction.toJson();
```
mohammadalfaiyazbitgo: not sure why you need to regenerate the transaction data, isn't that what Line #1000 is doing?

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1068,83 @@ export class Near extends BaseCoin {
...
+
+  //Validates that the public key matches between original and fresh transaction
+  private validatePublicKey(originalTxData: any, freshTxData: any): void {
+    if (originalTxData.publicKey !== freshTxData.publicKey) {
+      throw new Error(
+        `Error on token enablements: public keys are not the same, expected ${originalTxData.publicKey} but got ${freshTxData.publicKey}`
+      );
+    }
+  }
+
```
mohammadalfaiyazbitgo: these methods should compare `txParams` with `transaction`.

modules/sdk-coin-near/src/near.ts
```diff
@@ -1003,17 +1003,8 @@ export class Near extends BaseCoin {
...
-        if (tokenName) {
-          const nepToken = nearUtils.getTokenInstanceFromTokenName(tokenName);
-          if (nepToken) {
-            explainedTx.outputs.forEach((output) => {
-              if (output.amount !== nepToken.storageDepositAmount) {
-                throw new Error('Storage deposit amount not matching!');
-              }
-            });
-          }
-        }
```
mohammadalfaiyazbitgo, line 1015: why was this removed?

mohammadalfaiyazbitgo (changes requested):

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
```diff
@@ -0,0 +1,160 @@
...
+    const verifyOptions: VerifyTransactionOptions = {
+      txParams, // What the user thinks they're signing
+      txPrebuild, // The actual transaction hex from BitGo
+      wallet: { id: 'test-wallet' } as any, // Mock wallet object
+    };
+
+    // This should NOT throw an error - the transaction should be valid
+    // The verifyTransaction method will call validateTokenEnablementTransaction internally
+    await basecoin.verifyTransaction(verifyOptions);
+  });
```
mohammadalfaiyazbitgo, line 75: could you also verify that a transaction sent from wallet platform is verified as well

mohammadalfaiyazbitgo (changes requested):
add a unit test to show `sendTokenEnablements` throws an error when you mock a response with a spoofed TxHex. Otherwise lgtm.

mohammadalfaiyazbitgo (changes requested):
please add a test case showcasing `sendAccountConsolidations` fails when a spoofed txHex is returned in the response.

mohammadalfaiyazbitgo (changes requested):
Can we add a test that shows a spoofed tx from platform will be rejected, similar to HBAR's?

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
```diff
@@ -0,0 +1,504 @@
...
+// Import core types for transaction verification
+import { VerifyTransactionOptions, common } from '@bitgo/sdk-core';
+// Import NEAR-specific types and classes
+import { TransactionPrebuild, Near } from '../../src/near';
+import { TNear as TNearCoin } from '../../src/tnear';
+// Import test data containing sample transactions and account info
+import * as testData from '../resources/near';
+// Import testing utilities
+import nock from 'nock';
+import assert from 'assert';
```
balakrishna507: nit: do we need these comments?

balakrishna507 (commented):
i'll also request a review from the coin owners.

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1079,72 @@ export class Near extends BaseCoin {
       }
     }
   }
+
+  // Validates that the transaction matches what the user expects
+  private validateTokenEnablementTransaction(
+    transaction: Transaction,
+    explainedTx: TransactionExplanation,
+    txParams: VerifyTransactionOptions['txParams']
```
arvind-bg: Can we import and use the TransactionParams type directly?

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1079,72 @@ export class Near extends BaseCoin {
...
+    // Validate each aspect of the transaction separately
+    this.validateSigner(txParams, transactionData);
+    this.validateReceiver(txParams, transactionData);
+    this.validatePublicKey(txParams, transactionData);
+    this.validateActions(txParams, transactionData);
+    this.validateAddresses(txParams, explainedTx);
+  }
+
+  // Validates that the signer ID matches between txParams and transaction
+  private validateSigner(txParams: VerifyTransactionOptions['txParams'], transactionData: any): void {
```
arvind-bg: txParams appears unused. Can we remove it?

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1079,72 @@ export class Near extends BaseCoin {
...
+  private validateActions(txParams: VerifyTransactionOptions['txParams'], transactionData: any): void {
+    // For token enablement, we validate that the transaction has the expected actions
+    if (!transactionData.actions || transactionData.actions.length === 0) {
+      throw new Error('Error on token enablements: missing actions in transaction');
+    }
+  }
+
+  //Validates that addresses match between parameters and explained transaction
+  private validateAddresses(txParams: VerifyTransactionOptions['txParams'], explainedTx: TransactionExplanation): void {
+    if (!txParams.recipients || !explainedTx.outputs) {
```
arvind-bg: Unaware of the context here - if !txParams.recipients existed but !explainedTx.outputs didn't, should that throw an error?

arvind-bg (changes requested):
Clarifications + minor changes requested

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1079,72 @@ export class Near extends BaseCoin {
...
+  private validateSigner(txParams: VerifyTransactionOptions['txParams'], transactionData: any): void {
+    // For token enablement, we validate that the transaction signer matches the expected wallet
+    // This is a basic validation - in practice, the wallet address should match the signer
+    if (!transactionData.signerId) {
+      throw new Error('Error on token enablements: missing signer ID in transaction');
+    }
+  }
+
+  // Validates that the receiver ID matches between txParams and transaction
+  private validateReceiver(txParams: VerifyTransactionOptions['txParams'], transactionData: any): void {
```
arvind-bg: Can we use the actual type instead of any here?

arvind-bg (changes requested):

arvind-bg (dismissed):
lgtm

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1081,75 @@ export class Near extends BaseCoin {
...
+    txParams: TransactionParams
+  ): void {
+    const transactionData = transaction.toJson();
+
+    // Validate each aspect of the transaction separately
+    this.validateSigner(transactionData);
+    this.validateReceiver(transactionData);
+    this.validatePublicKey(transactionData);
+    this.validateActions(transactionData);
+    this.validateAddresses(txParams, explainedTx);
```
mtexeira-simtlix: We're missing a validate method that checks the tx type.
In txParams, type is 'enabletoken' and when that's the type, in transactionData we should observe another kind of type (in my experience it's coin dependant).
For reference, check the verifyTxType on this SOL pr:

image

https://github.com/BitGo/BitGoJS/pull/6966/files#diff-b1c96ac782eb4397efa8aca769f141cdbaa3b181ca8c1a613efc2e2b3cded801R347

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
```diff
@@ -0,0 +1,498 @@
...
+    // Prepare the verification options that would be passed to BitGo
+    const verifyOptions: VerifyTransactionOptions = {
+      txParams, // What the user thinks they're signing
+      txPrebuild, // The actual transaction hex from BitGo
+      wallet: { id: 'test-wallet' } as any, // Mock wallet object
+    };
+
+    // This should NOT throw an error - the transaction should be valid
+    // The verifyTransaction method will call validateTokenEnablementTransaction internally
+    await basecoin.verifyTransaction(verifyOptions);
```
mtexeira-simtlix, line 74: nit: mind doing some comments cleanup on the code? Please let the ones that adds some info that's not implicit in the code to increase readability.

i.e:
` // Setup that runs once before all tests
  before(function () {`
 That's not case specific, it's a generic behaviour of before.
 
 ` // Register the TNEAR (testnet NEAR) coin with BitGo
    bitgo.safeRegister('tnear', TNearCoin.createInstance);`
Same as the prev one, it's self-explanatory and could be removed.

But this one:
`// Create transaction prebuild using a real storage deposit transaction`
adds good context because no one is able to decrypt the txHex by looking at it, so it helps to know what's the hidden context and we could maintain it.

Thanks Simon!

mtexeira-simtlix (changes requested):
Not sure if there is an exception on Near but cannot find in the PR a validation/verify method for token name in txParams vs the decrypted prebuild tx, same as for the tx type in params ('enabletoken') that should match with some prebuild tx type on decoded tx (it depends on the coin, XLM uses 'changeTrust' and i don't recall it right now but SOL uses something else, and also XRP).

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
```diff
@@ -0,0 +1,498 @@
...
+        };
+
+        // This should fail because the spoofed hex doesn't represent a valid NEAR transaction
+        await basecoin.verifyTransaction(verifyOptions);
+      },
+      (error: any) => {
+        // The error should indicate that the transaction is invalid
+        // This could be various validation errors depending on what the spoofed hex contains
+        return error.message.includes('unable to build transaction from raw');
+      }
```
mohammadalfaiyazbitgo: This should be checking an assert is thrown in `sendTokenEnablements` given just the nock with the spoofed transaction.

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1081,75 @@ export class Near extends BaseCoin {
...
+    explainedTx: TransactionExplanation,
+    txParams: TransactionParams
+  ): void {
+    const transactionData = transaction.toJson();
+
+    // Validate each aspect of the transaction separately
+    this.validateSigner(transactionData);
+    this.validateReceiver(transactionData);
+    this.validatePublicKey(transactionData);
+    this.validateActions(transactionData);
```
mohammadalfaiyazbitgo: none of these need to verify against `txParams: TransactionParams` ?

mohammadalfaiyazbitgo (changes requested):

modules/sdk-coin-near/src/near.ts
SimonVutovB: I've implemented a validateTxType that checks the txParams.type and explainedTx.type.

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
SimonVutovB: I've pushed changes addressing the comments. Thanks for the help Max!

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
SimonVutovB: Thanks, I've pushed changes fixing this.

modules/sdk-coin-near/src/near.ts
SimonVutovB: Yes, they do. I’ve pushed changes implementing that. It looks like I had accidentally removed the txParams in a commit, but it’s fixed now.

modules/sdk-coin-near/src/near.ts
```diff
@@ -1014,6 +1016,8 @@ export class Near extends BaseCoin {
             });
           }
         }
+        // Validate that this is a proper token enablement transaction
+        this.validateTokenEnablementTransaction(transaction, explainedTx, txParams);
```
mtexeira-simtlix: (as we discussed during sync, swap this couple of lines:
image

)

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1093,207 @@ export class Near extends BaseCoin {
...
+      recipient.amount !== action.functionCall.deposit
+    ) {
+      throw new Error(
+        `Error on token enablements: user specified amount '${recipient.amount}' does not match storage deposit '${action.functionCall.deposit}'`
+      );
+    }
+  }
+
+  private validateTxType(txParams: TransactionParams, explainedTx: TransactionExplanation): void {
+    if (txParams.type === 'enabletoken') {
```
mtexeira-simtlix: validateTxType is already called inside the validateTokenEnablementTransaction func, and that function is called inside another block that checks for txParams.type as 'enabletoken'.

You can remove this line safely.

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1093,207 @@ export class Near extends BaseCoin {
...
+    explainedTx: TransactionExplanation,
+    txParams: TransactionParams
+  ): void {
+    this.validateTxType(txParams, explainedTx);
+    this.validateSigner(transaction);
+    this.validateRawReceiver(transaction, txParams);
+    this.validatePublicKey(transaction);
+    this.validateRawActions(transaction, txParams);
+    this.validateBeneficiary(explainedTx, txParams);
+    this.validateTokenOutput(explainedTx, txParams);
```
mtexeira-simtlix, line 1099: This one is more like a nit for optimization:
The idea is to remove the internal transaction.toJson() calls that I'll assume could be heavy ones and just pass the parsed tx data down
```suggestion
    const transactionData = transaction.toJson();
    this.validateTxType(txParams, explainedTx);
    this.validateSigner(transactionData);
    this.validateRawReceiver(transactionData, txParams);
    this.validatePublicKey(transactionData);
    this.validateRawActions(transactionData, txParams);
    this.validateBeneficiary(explainedTx, txParams);
    this.validateTokenOutput(explainedTx, txParams);
```

modules/sdk-coin-near/src/near.ts
```diff
@@ -1077,4 +1093,207 @@ export class Near extends BaseCoin {
...
+      // ed25519 keys are 32 bytes = 44 base58 chars
+      throw new Error('Error on token enablements: invalid ed25519 public key format');
+    }
+
+    // Validate it's actually valid base58
+    try {
+      const decoded = nearAPI.utils.serialize.base_decode(base58Part);
+      if (decoded.length !== 32) {
+        throw new Error('Error on token enablements: invalid ed25519 public key length');
+      }
```
mtexeira-simtlix: This throw got masked by the catch below as you throw inside the same try as the serialize.base_decode call.
Something like this could do the job!

```suggestion
      let decoded;
      try{
         decoded = nearAPI.utils.serialize.base_decode(base58Part);
      } catch {
         throw new Error('Error on token enablements: invalid base58 encoding in public key');
      }
      
      if (!decoded || decoded.length !== 32){
         throw new Error('Error on token enablements: invalid ed25519 public key length');
      }
```

mtexeira-simtlix (changes requested):
Thanks for the changes Simon! It's looking better.
I left a nit and a couple of requested changes, feel free to ask about these!
Cheers.

modules/sdk-coin-near/src/near.ts
mtexeira-simtlix: verified, thanks!

modules/sdk-coin-near/src/near.ts
mtexeira-simtlix: done, thanks!

mtexeira-simtlix (dismissed):
lgtm, thanks!

mohammadalfaiyazbitgo (dismissed):

modules/sdk-coin-near/src/near.ts
```diff
@@ -1001,6 +1003,10 @@ export class Near extends BaseCoin {
     const explainedTx = transaction.explainTransaction();
 
     // users do not input recipients for consolidation requests as they are generated by the server
+    if (txParams.type === 'enabletoken') {
```
mohammadalfaiyazbitgo:
```suggestion
    if (txParams.type === 'enabletoken' && verification.verifyTokenEnablement) {
```

mohammadalfaiyazbitgo (commented):
LGTM, missing one last change.

mtexeira-simtlix (dismissed):

mohammadalfaiyazbitgo (dismissed):

modules/sdk-coin-near/src/near.ts
```diff
@@ -1032,6 +1038,18 @@ export class Near extends BaseCoin {
       });
 
       if (!_.isEqual(filteredOutputs, filteredRecipients)) {
+        // For enabletoken, provide more specific error messages for address mismatches
+        if (txParams.type === 'enabletoken') {
```
FloBitGo:
```suggestion
        if (txParams.type === 'enabletoken' && params.verification?.verifyTokenEnablement) {
```
Not sure if we always want to throw this new error or only if the param is passed?

FloBitGo (dismissed):

modules/sdk-coin-near/src/near.ts
SimonVutovB: Yeah you're right, good catch. Just pushed change to add the check for the verifyTokenEnablement flag.

mohammadalfaiyazbitgo (dismissed):

mukeshsp (dismissed):

mukeshsp (commented):
Please squash all the commits into a single commit.

lerna.json
```diff
@@ -1,6 +1,7 @@
 {
   "version": "independent",
   "npmClient": "yarn",
+  "packages": ["modules/*"],
```
mohammadalfaiyazbitgo: is this needed?

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
```diff
@@ -0,0 +1,488 @@
...
+          name: 'tnear:tnep24dp',
+        },
+      ],
+      walletPassphrase: 'test', // Required for hot wallet signing
+    });
+
+    // The result should contain failures due to the spoofed transaction hex
+    result.success.should.have.length(0);
+    result.failure.should.have.length(1);
+    result.failure[0].message.should.containEql('unable to build transaction from raw');
```
mohammadalfaiyazbitgo: I'm not sure how this is testing that the spoofed transaction does not match the intended transaction?

mohammadalfaiyazbitgo (changes requested):

modules/sdk-coin-near/src/near.ts
```diff
@@ -1055,4 +1074,216 @@ export class Near extends BaseCoin {
...
+
+  private validateBeneficiary(explainedTx: TransactionExplanation, txParams: TransactionParams): void {
+    if (!explainedTx.outputs || explainedTx.outputs.length === 0) {
+      throw new Error('Error on token enablements: transaction has no outputs to validate beneficiary');
+    }
+
+    const output = explainedTx.outputs[0];
+    const recipient = txParams.recipients?.[0];
+
+    if (!recipient?.address) {
```
mtexeira-simtlix, line 1130: Could you confirm if near token enablements supports a single recipient? because if that's the case then you may need to check for outputs.length === 1 and if it isn't then you may need to iterate over the outputs or recipients and match one with the other (also you can preliminary compare outputs.length with recipients.length and they should match).

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
mtexeira-simtlix: Yeah I agree with @mohammadalfaiyazbitgo , I don't remember how the internal flow goes exactly but if you mock the build results then you're basically shutting down the internal verifies somewhere.

@SimonVutovB mind trying an approach similar to what I did in this other tests? It's also a little shorter and you may throw an error msg that should be the same as one of your token enablement error msg's on the current PR:
image


For ref, this is on modules/sdk-coin-xlm/test/unit/xlm.ts , look for the test name on that file please.

lerna.json
SimonVutovB: I deleted it and all checks pass, so I guess it was not needed.

modules/sdk-coin-near/src/near.ts
```diff
@@ -1055,4 +1074,229 @@ export class Near extends BaseCoin {
...
+      throw new Error('No base address found on wallet');
+    }
+    const baseAddress = baseAddresses[0];
+
+    for (const output of explainedTx.outputs) {
+      if (output.address !== baseAddress.address) {
+        throw new Error('tx outputs does not match with expected address');
+      }
+    }
+  }
```
mohammadalfaiyazbitgo: let's remove this since it was reverted.

mohammadalfaiyazbitgo (changes requested):

modules/sdk-coin-near/src/near.ts
SimonVutovB: done

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
```diff
@@ -0,0 +1,424 @@
...
+          },
+        ],
+        coin: 'tnear',
+        type: 'enabletoken',
+        feeInfo: {
+          size: 1000,
+          fee: 1160407,
+          feeRate: 1160407,
+        },
+      });
```
mohammadalfaiyazbitgo, line 422: is this the actual call made when a transaction is built? We actually should be going through here: https://github.com/BitGo/BitGoJS/blob/af32d868ac2f944c59fc60d173e42df1d5e42494/modules/sdk-core/src/bitgo/wallet/wallet.ts#L3329

mohammadalfaiyazbitgo (commented):
I think changes are getting there, I tested this locally and transactions are being verified but we need to write a more comprehensive unit test.

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
SimonVutovB: I added tests that use buildTokenEnablements.

mohammadalfaiyazbitgo (dismissed):

modules/sdk-coin-near/src/near.ts
```diff
@@ -1055,4 +1073,212 @@ export class Near extends BaseCoin {
     }
     auditEddsaPrivateKey(prv, publicKey ?? '');
   }
+
+  private validateTokenEnablementTransaction(
```
ravibitgo, line 1087: JSDoc

modules/sdk-coin-near/src/near.ts
```diff
@@ -1055,4 +1073,212 @@ export class Near extends BaseCoin {
...
+    }
+
+    if (recipient?.tokenName && recipient.tokenName !== output.tokenName) {
+      throw new Error(
+        `Error on token enablements: token mismatch - user expects '${recipient.tokenName}', transaction has '${output.tokenName}'`
+      );
+    }
+  }
+
+  private validatePublicKey(transactionData: TxData): void {
```
ravibitgo, line 1187: this class has `isValidPublicKey` method, can we not use that?

modules/sdk-coin-near/src/near.ts
```diff
@@ -1055,4 +1073,212 @@ export class Near extends BaseCoin {
...
+
+    const action = transactionData.actions[0];
+
+    // Must be a functionCall action (not transfer)
+    if (!action.functionCall) {
+      throw new Error('Error on token enablements: action must be a function call');
+    }
+
+    // Must be storage_deposit method (NEAR spec requirement)
+    if (action.functionCall.methodName !== 'storage_deposit') {
```
ravibitgo, line 1218: can we use the `STORAGE_DEPOSIT` from the constants?

modules/sdk-coin-near/test/unit/tokenEnablementValidation.ts
```diff
@@ -0,0 +1,674 @@
+import 'should';
+import { TestBitGo, TestBitGoAPI } from '@bitgo/sdk-test';
+import { BitGoAPI } from '@bitgo/sdk-api';
+import { VerifyTransactionOptions, common, Wallet, TransactionType } from '@bitgo/sdk-core';
+import { TransactionPrebuild, Near } from '../../src/near';
+import { TNear as TNearCoin } from '../../src/tnear';
+import * as testData from '../resources/near';
+import nock from 'nock';
```
ravibitgo: external imports should be on top

SimonVutovB: Just pushed changes addressing these comments. Thanks for the suggestions Ravi!

mukeshsp (approved):

mohammadalfaiyazbitgo (approved):

MohammedRyaan786 (approved):

nvjsr (approved):

PR #6962, opened 2025-09-11
fix: bump up follow-redirects version

koralkulacoglu:
TICKET: DX-1554

Screenshot 2025-09-11 at 2 44 30 PM

NOTE: we actually use version 1.15.11 not ^1.15.6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "body-parser": "1.20.3",
     "degenerator": "5.0.0",
     "eventsource": "2.0.2",
-    "follow-redirects": "1.15.4",
+    "follow-redirects": "^1.15.6",
     "io-ts": "npm:@bitgo-forks/io-ts@2.1.4",
     "isbinaryfile": "5.0.0",
     "minimist": "1.2.6",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
     "body-parser": "1.20.3",
     "degenerator": "5.0.0",
     "eventsource": "2.0.2",
-    "follow-redirects": "1.15.4",
+    "follow-redirects": "^1.15.6",
```
zahin-mohammad: Please remove the carrot. 
```suggestion
    "follow-redirects": "1.15.6",
```

also why not just directly go to 1.15.11?

https://www.npmjs.com/package/follow-redirects/v/1.15.11

zahin-mohammad (changes requested):

package.json
koralkulacoglu: Should I use the latest version (that's patched) instead then?

koralkulacoglu: Oh sorry I didn't see your second comment (wifi went out), yes that's what I was thinking of doing.

zahin-mohammad (approved):

margueriteblair (approved):

PR #6966, opened 2025-09-11
feat(sdk-coin-sol): blind signing token enablements guards

mtexeira-simtlix:
Ticket: WP-5744

Description: 
- Token enablements blind signing guards for solana.


modules/sdk-coin-sol/src/sol.ts
```diff
@@ -261,6 +344,15 @@ export class Sol extends BaseCoin {
     transaction.fromRawTransaction(rawTxBase64);
     const explainedTx = transaction.explainTransaction();
 
+    this.verifyTxType(txParams, explainedTx.type);
```
mtexeira-simtlix: To have the verifyTxType outside of the if block on next line (L348) is something intentional, the verifyType is generalized, non specific to enabletoken although we don't have any other cases to check by now.

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -261,6 +344,15 @@ export class Sol extends BaseCoin {
...
+    if (txParams.type === 'enabletoken') {
+      const outputs = this.throwIfMissingOutputsOrReturn(explainedTx);
+      const enableTokensConfig = this.throwIfMissingEnableTokenConfigOrReturn(txParams);
+
```
mtexeira-simtlix: @mohammadalfaiyazbitgo based on your reviews on the XRP PR, instead of putting these checks inside every verify, I just moved that out and put them in a upper level than the this.verifyXYZ calls to avoid checking if the outputs or some configs are there or throwing multiple times.

In any case, putting those back doesn't change the algorithmic time as both are plain checks.

mtexeira-simtlix: Code changed a little since my last push, outputs was replaced by tokenEnablementsPrebuild (regarding to our slack conversation).

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -261,6 +351,16 @@ export class Sol extends BaseCoin {
...
     const explainedTx = transaction.explainTransaction();
 
+    this.verifyTxType(txParams, explainedTx.type);
+    if (txParams.type === 'enabletoken') {
+      const tokenEnablementsPrebuild = this.throwIfMissingTokenEnablementsOrReturn(explainedTx);
+      const enableTokensConfig = this.throwIfMissingEnableTokenConfigOrReturn(txParams);
+
+      this.verifyTokenName(tokenEnablementsPrebuild, enableTokensConfig);
+      await this.verifyTokenAddress(tokenEnablementsPrebuild, enableTokensConfig);
+      this.verifyNoOutputs(explainedTx);
```
mohammadalfaiyazbitgo: Can someone from Coins/Eco-System team confirm an `enableToken` will never have additional outputs in the tx?

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -233,6 +237,92 @@ export class Sol extends BaseCoin {
...
+  throwIfMissingTokenEnablementsOrReturn(explanation: TransactionExplanation): ITokenEnablement[] {
+    if (!explanation.tokenEnablements || explanation.tokenEnablements.length === 0)
+      throw new Error('Missing tx token enablements data on token enablement tx prebuild');
+    return explanation.tokenEnablements;
+  }
+
+  throwIfMissingEnableTokenConfigOrReturn(txParams: TransactionParams): TokenEnablement {
+    if (!txParams.enableTokens || txParams.enableTokens.length === 0) throw new Error('Missing enable token config');
+    if (txParams.enableTokens.length > 1)
+      throw new Error('Multiple token enablement not supported in a single transaction');
```
mohammadalfaiyazbitgo: where is this requirement coming from?

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -233,6 +237,92 @@ export class Sol extends BaseCoin {
...
+      } catch {
+        throw new Error(`Unable to derive ATA for token address: ${expectedTokenAddress}`);
+      }
+      if (
+        !tokenMintAddress ||
+        tokenMintAddress.tokenAddress === undefined ||
+        tokenMintAddress.programId === undefined
+      ) {
+        throw new Error(`Unable to get token mint address for ${expectedTokenName}`);
+      }
```
mohammadalfaiyazbitgo: we shouldn't throw in this case, there maybe cases where a client wants to enable a token that's not configured in the SDK yet. In these cases we should just skip the this verification with a warning log.

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -261,6 +351,16 @@ export class Sol extends BaseCoin {
     transaction.fromRawTransaction(rawTxBase64);
     const explainedTx = transaction.explainTransaction();
 
+    this.verifyTxType(txParams, explainedTx.type);
+    if (txParams.type === 'enabletoken') {
```
mohammadalfaiyazbitgo: let's also add an option to https://github.com/BitGo/BitGoJS/blob/076f4ccf6bae5031c9bcd3bf49dde2c297767fec/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts#L176 called `verifyTokenEnablement` with default to true, and perform this check if `txParams.type === enableToken && verifyTokenEnablement === true`

mohammadalfaiyazbitgo (changes requested):
over all lgtm, just a few comments to address.

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: @mohammadalfaiyazbitgo I forgot to tag you in this one, my bad. Please check the comment above and if you agree please mark this conversation as Resolved. Thanks!

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: I found @venkateshv1266 on a confluence doc redacted by Coins team so decided to ask him to ask on his team channel as a favor since I cannot see/access any other channels that aren't the ones I'm already added.

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: I'm not so sure about that as in the original code from which I got the call:

image

which means that tokenAddress & programId should come with something which is not possible if tokenMintAddress comes as null/undefined.

In other words if we left it pass as a warning that's supposed to crash later (both in the new code of this PR on L304 ata = await getAssociated...Address and in the original code that i posted above).

For some reason in the original code they assumed that tokenMintAddress will always comes as something populated despite that getSolTokenFromTokenName(...) could return undefined hence the `!` to force the non null of that field :/

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: We talk with Venkatesh and all questions goes to the same notify channel as the eth-alt team, I thought that coins india had a specific channel for their team.

I'll ask about this in a few. Thanks guys.

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -351,8 +346,8 @@ export class Sol extends BaseCoin {
     transaction.fromRawTransaction(rawTxBase64);
     const explainedTx = transaction.explainTransaction();
 
-    this.verifyTxType(txParams, explainedTx.type);
-    if (txParams.type === 'enabletoken') {
+    if (txParams.type === 'enabletoken' && verificationOptions?.verifyTokenEnablement) {
+      this.verifyTxType(txParams.type, explainedTx.type);
```
mtexeira-simtlix, line 347: @mohammadalfaiyazbitgo turned into a check on token enablements so it gets affected by the verifyTokenEnablement verify prop.

Also, goes first as a tx of type "Send" won't contain things that are tested on L351 & L352 ( the lines below) and if I don't put things in this order it's basically impossible to write the tests with real payloads as it will fail by throwIfMissingTokenEnablementsOrReturn(...) instead of the invalid/unexpected tx type.

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: Done! to avoid the burden of checking all the files from scratch, I put this on a new commit, please give it a check when you can, thanks!
https://github.com/BitGo/BitGoJS/pull/6966/commits/65054798a57ac86fd9779ad5d0f145e0e9936e23

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: https://bitgo.slack.com/archives/C053T7939M0/p1758205471956299

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: Good catch, thanks! let me double check if I made this change for some reason in particular and get back to you, if not, I'll change the code to support multiple entries on enableTokens.

mtexeira-simtlix: Yeah found nothing about supporting a single token enablement per tx, perhaps got confused as this was my first contact with blind signing + tokens enablements combined into the SDK, will change this in a moment. Thanks and kudos!

mtexeira-simtlix: @mohammadalfaiyazbitgo removed the single check and adapted the code to support an array instead of the single entry that I wrongly assumed:
https://github.com/BitGo/BitGoJS/pull/6966/commits/42ddfffabd121b067344589371190aa1858e3380

modules/sdk-coin-sol/test/unit/sol.ts
```diff
@@ -3334,5 +3335,22 @@ describe('SOL:', function () {
         }
       );
     });
+
+    it('should fail sendTokenEnablement call on spoofed data', async function () {
```
mtexeira-simtlix, line 3339: @mohammadalfaiyazbitgo the new test requested, I could add tests for the other cases that we verify but it's not worth it as I already added the same kind of unit tests at the base of this PR.

modules/sdk-coin-sol/src/sol.ts
mtexeira-simtlix: Forget this, no longer applies, Resolving the conversation as I moved the method inside the enabletoken type check.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3231,6 +3231,8 @@ export class Wallet implements IWallet {
    */
   public async sendTokenEnablement(params: PrebuildAndSignTransactionOptions = {}): Promise<any> {
     const teConfig = this.baseCoin.getTokenEnablementConfig();
+    if (params.verification && params.verification.verifyTokenEnablement === undefined)
+      params.verification.verifyTokenEnablement = true;
```
mohammadalfaiyazbitgo: default to true even if `params.verification` is undefined.

mohammadalfaiyazbitgo (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
mtexeira-simtlix: done!

prithvishet2503 (approved):

PR #6975, opened 2025-09-12
fix: use new instances of wallet in parallel calls

bhavidhingra:
This pull request updates the logic for bulk signing account-based messages, specifically improving how messages are processed in parallel and how wallet instances are managed. The main change is that each message is now processed with a new `Wallet` instance, which helps avoid potential side effects from reusing the same wallet object. Additionally, the code now gathers all messages before processing, ensuring that all batches are handled consistently.

**Parallel message processing improvements:**

* All messages to be processed are now gathered upfront by flattening all batches, rather than processing one batch at a time.
* Each message is processed in parallel using a new `Wallet` instance constructed with the relevant parameters, ensuring isolation between message signings.
* The `Wallet` class is now imported alongside `IWallet` to support instantiating new wallet objects for each message.

TICKET: COIN-5426


zahin-mohammad (commented):
Can you add a test for these changes?

hrishikeshjain (approved):

ravibitgo (approved):
lgtm

PR #6984, opened 2025-09-13
feat: link security policy for publishing

udayshanmugam:
Create a symbolic link to the security policy so that it will be included in the published BitGoJS

Ticket: DX-1885

### DIFF
--- a/modules/bitgo/SECURITY.md
+++ b/modules/bitgo/SECURITY.md
@@ -1,0 +1,28 @@
+# Security Policy
+
+The BitGo team and the wider BitGo community take the security of the `BitGoJS` library seriously. We are committed to a secure and transparent vulnerability disclosure process. This policy outlines how to responsibly report security vulnerabilities in the `BitGoJS` repository and its associated ecosystem.
+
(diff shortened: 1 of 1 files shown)


modules/bitgo/SECURITY.md
```diff
@@ -0,0 +1 @@
+../../SECURITY.md
```
zahin-mohammad: Does this actually work when publishing NPM modules?

zahin-mohammad (commented):
Not sure symlinking actually bundles the files with the published sdk, and also this change only affects 1 module.

udayshanmugam:
> Not sure symlinking actually bundles the files with the published sdk, and also this change only affects 1 module.

Replaced symlink with a copy of the file.
The policy at the root of the repo will work for developers who visit our repo.
However the policy at the repo root does not get packaged into BitGoJS which is published to the npm registry and hence the copy.

udayshanmugam:
As per https://bitgoinc.atlassian.net/browse/DX-1956 , the security policy is in effect and this PR is not needed

PR #6989, opened 2025-09-15
fix(utxo-staking): update btc staking params

benkelcher-bitgo:
Ticket: SC-3117


lcovar (approved):

benkelcher-bitgo:
btc-babylon staking params are now fetched dynamically by `staking-service` and passed into SDK to mitigate the need for updating this every time the foundation changes their params.

PR #6941, opened 2025-09-10
feat: blind signing guards for xrp token enablements

mtexeira-simtlix:
TICKET: WP-5820

Description: 
Blind signing guards for xrp token enablements including token name, issuer, account addr and others.

PR #7003, opened 2025-09-16
feat(sdk-coin-hbar): add token enablement transaction verification

SimonVutovB:
Add verifyTokenEnablementTransaction method to validate token associate transactions with comprehensive checks for amount, account ID, token name, and transaction type. Integrates with existing verifyTransaction flow for enhanced token enablement validation.

- Add organized validation functions for transaction structure, amount, account ID, token name, and transaction type

TICKET: WP-5746

### DIFF
--- a/modules/sdk-coin-hbar/src/hbar.ts
+++ b/modules/sdk-coin-hbar/src/hbar.ts
@@ -28,7 +28,7 @@ import {
 import { BigNumber } from 'bignumber.js';
 import * as stellar from 'stellar-sdk';
 import { SeedValidator } from './seedValidator';
-import { KeyPair as HbarKeyPair, TransactionBuilderFactory, Transaction } from './lib';
+import { KeyPair as HbarKeyPair, TransactionBuilderFactory, Transaction, Recipient as HederaRecipient } from './lib';
 import * as Utils from './lib/utils';
 import * as _ from 'lodash';
 import {
@@ -39,6 +39,31 @@ import {
   Hbar as HbarUnit,
 } from '@hashgraph/sdk';
 import { PUBLIC_KEY_PREFIX } from './lib/keyPair';
+
+// Hedera-specific transaction data interface for raw transaction validation
+interface HederaRawTransactionData {
+  id?: string;
+  from?: string;
+  fee?: number;
+  startTime?: string;
+  validDuration?: string;
+  node?: string;
+  memo?: string;
+  amount?: string;
+  instructionsData?: {
+    type?: string;
+    accountId?: string;
+    params?: {
+      accountId?: string;
+      tokenNames?: string[];
+      recipients?: Array<{
+        address: string;
+        amount: string;
+        tokenName?: string;
+      }>;
+    };
+  };
+}
 export interface HbarSignTransactionOptions extends SignTransactionOptions {
   txPrebuild: TransactionPrebuild;
   prv: string;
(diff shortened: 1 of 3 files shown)


mukeshsp (dismissed):

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -245,6 +405,23 @@ export class Hbar extends BaseCoin {
       throw new Error('missing required tx params property recipients');
     }
 
+    // for enabletoken, use verifyTokenEnablementTransaction and return immediately
+    if (txParams.type === 'enabletoken') {
```
mohammadalfaiyazbitgo: let's cherry pick the these changes as well for `wallet.ts`, `modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts` from this pr   https://github.com/BitGo/BitGoJS/pull/6941/files and use the same pattern here to only verify if `enableToken` && `verifyTokenEnablement === true`

mohammadalfaiyazbitgo (changes requested):

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -224,6 +224,166 @@ export class Hbar extends BaseCoin {
...
+    if (Array.isArray(obj.transfers) && obj.transfers.length > 0) return true;
+    if (Array.isArray(obj.tokenTransferLists) && obj.tokenTransferLists.some((t: any) => this.hasAnyTransfers(t)))
+      return true;
+
+    // Recursively check all nested objects
+    return Object.values(obj).some((value: any) => this.hasAnyTransfers(value));
+  }
+
+  // Validate account ID matches in both explained and raw transaction data with normalization
+  private validateAccountIdMatches(ex: TransactionExplanation, raw: any, expectedAccountId: string): void {
```
mohammadalfaiyazbitgo: can we avoid `any` type for `raw`

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -224,6 +224,166 @@ export class Hbar extends BaseCoin {
...
+    const assocAcct = raw.instructionsData?.accountId ?? raw.instructionsData?.owner ?? raw.accountId;
+    if (assocAcct && !Utils.isSameBaseAddress(assocAcct, expectedAccountId)) {
+      throw new Error(`Associate account ${assocAcct} does not match expected ${expectedAccountId}`);
+    }
+  }
+
+  // Validate token enablement target with preference for tokenId over tokenName
+  private validateTokenEnablementTarget(
+    ex: TransactionExplanation,
+    raw: any,
```
mohammadalfaiyazbitgo: same here.

modules/sdk-coin-hbar/test/unit/hbar.ts
```diff
@@ -501,6 +501,91 @@ describe('Hedera Hashgraph:', function () {
     });
   });
 
+  describe('Verify Token Enablement Transaction:', () => {
```
mohammadalfaiyazbitgo, line 516: let's add a test that uses a valid transaction response from wallet platform with a spoofed txHex, 
the test should have the structure of 
```
await assert.rejects(
        wallet.sendTokenEnablements(...)
)...
```
so we know a spoofed tx hex will throw an error on this flow.

mohammadalfaiyazbitgo (changes requested):

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -39,6 +39,43 @@ import {
...
+        address: string;
+        amount: string;
+        tokenName?: string;
+      }>;
+    };
+  };
+  // These are any[] because they're only used to verify existence, not to access contents
+  instructions?: any[];
+  innerInstructions?: any[];
+  scheduledTransactionBody?: any;
```
mohammadalfaiyazbitgo: nit: prefer `unknown`

mohammadalfaiyazbitgo (dismissed):

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -224,10 +260,178 @@ export class Hbar extends BaseCoin {
...
+    if (Array.isArray(obj.accountAmounts) && obj.accountAmounts.length > 0) return true;
+    if (Array.isArray(obj.tokenTransfers) && obj.tokenTransfers.length > 0) return true;
+    if (Array.isArray(obj.nftTransfers) && obj.nftTransfers.length > 0) return true;
+    if (Array.isArray(obj.transfers) && obj.transfers.length > 0) return true;
+    if (Array.isArray(obj.tokenTransferLists) && obj.tokenTransferLists.some((t: any) => this.hasAnyTransfers(t)))
+      return true;
+
+    // Recursively check all nested objects
+    return Object.values(obj).some((value: any) => this.hasAnyTransfers(value));
+  }
```
balakrishna507: can this happen? transfers containing transfers that contain even more transfers? do we support such transactions on our sdk?

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -224,10 +260,178 @@ export class Hbar extends BaseCoin {
...
+    } catch (error) {
+      throw new Error(`Invalid token enablement transaction: ${error.message}`);
+    }
+  }
+
+  // Strict validation: allow 0 outputs or exactly 1 output with amount 0
+  private validateTxStructureStrict(ex: TransactionExplanation): void {
+    if (!ex.outputs || ex.outputs.length === 0) {
+      return; // acceptable for pure associate in some explainers
+    }
```
balakrishna507, line 293: can we add a comment about when or for which tokens this might happen?

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -245,6 +449,23 @@ export class Hbar extends BaseCoin {
...
+      const r0 = txParams.recipients[0];
+      const expectedToken: { tokenId?: string; tokenName?: string } = {};
+
+      // Use tokenName from recipient (tokenId not available in current API)
+      if (r0.tokenName) {
+        expectedToken.tokenName = r0.tokenName;
+        // Note: tokenName validation is less secure than tokenId, but tokenId is not available in ITransactionRecipient
+      } else {
+        throw new Error('Token enablement requires tokenName in recipient');
+      }
```
balakrishna507: `txParams` does not use token ids? wouldn't this be a problem, or is there there something that disambiguates between tokens that share the same name?

balakrishna507 (commented):
requesting a review from the other coin owner.

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -224,10 +260,178 @@ export class Hbar extends BaseCoin {
...
+   * @param txHex - The transaction hex to verify
+   * @param expectedToken - Object containing tokenId (preferred) or tokenName
+   * @param expectedAccountId - The expected account ID that will enable the token
+   * @returns Promise<boolean> - True if valid token enablement transaction
+   */
+  async verifyTokenEnablementTransaction(
+    txHex: string,
+    expectedToken: { tokenId?: string; tokenName?: string },
+    expectedAccountId: string
+  ): Promise<boolean> {
```
arvind-bg: I think this need not return anything. We don't return false and only throw an exception + the return value is not being used.

modules/sdk-coin-hbar/src/hbar.ts
SimonVutovB: Good point, in practice our SDK doesn’t support nested transfers within transfers (as seen in tokenTransferBuilder.ts and other scripts), so the recursive scan was unnecessary. I updated the validation to do a simpler direct check for recipients and amounts instead.

modules/sdk-coin-hbar/src/hbar.ts
SimonVutovB: I fixed it to use token IDs.

modules/sdk-coin-hbar/src/hbar.ts
SimonVutovB: Token association transactions in Hedera should have exactly 1 output with amount '0' for each token being associated. I changed it to throw an error with 0 outputs. Thank you for bringing it up.

modules/sdk-coin-hbar/test/unit/hbar.ts
mohammadalfaiyazbitgo: also another test verifying a valid token enablement tx won't cause an error.

mohammadalfaiyazbitgo (changes requested):

modules/sdk-coin-hbar/test/unit/hbar.ts
SimonVutovB: Tests on 685-725 and 726-774 are the two tests that address this.

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -224,10 +249,195 @@ export class Hbar extends BaseCoin {
...
+    const out0 = ex.outputs[0];
+    if (out0.amount !== '0') {
+      throw new Error(`Expected output amount '0', got ${out0.amount}`);
+    }
+  }
+
+  private validateNoTransfers(raw: HederaRawTransactionData): void {
+    if (raw.instructionsData?.params?.recipients?.length && raw.instructionsData.params.recipients.length > 0) {
+      const hasNonZeroTransfers = raw.instructionsData.params.recipients.some(
+        (recipient: any) => recipient.amount && recipient.amount !== '0'
```
mohammadalfaiyazbitgo: please address the typing in a follow up.

modules/sdk-coin-hbar/src/hbar.ts
```diff
@@ -224,10 +249,195 @@ export class Hbar extends BaseCoin {
...
+    }
+  }
+
+  private validateAssociateInstructionOnly(raw: HederaRawTransactionData): void {
+    const t = String(raw.instructionsData?.type || '').toLowerCase();
+
+    if (t === 'contractexecute' || t === 'contractcall' || t === 'precompile') {
+      throw new Error(`Contract-based token association not allowed for blind enablement; got ${t}`);
+    }
+
```
mohammadalfaiyazbitgo: please update the variable naming here.

mohammadalfaiyazbitgo (dismissed):
generally LGTM, some comments to follow up on.

mohammadalfaiyazbitgo (approved):

mukeshsp (commented):
Please squash all the commits into a single commit.

mukeshsp (approved):

kamleshmugdiya (approved):

PR #7006, opened 2025-09-16
feat(sdk-coin-xlm): blind signing guards for token enablements

mtexeira-simtlix:
Ticket: WP-5819

Description:
- Blind signing preventive checks for token enablements on XLM.
- Removed casts and changed for guards/properties checks.
- Unit tests


modules/sdk-coin-xlm/src/xlm.ts
```diff
@@ -1067,6 +1062,95 @@ export class Xlm extends BaseCoin {
     });
   }
 
+  getRecipientOrThrow(txParams: TransactionParams): ITransactionRecipient {
+    if (!txParams.recipients || txParams.recipients.length === 0)
+      throw new Error('Missing recipients on token enablement');
+    if (txParams.recipients.length > 1) throw new Error('Multiple recipients not supported on token enablement');
```
mohammadalfaiyazbitgo, line 1072: is this a valid check for XLM?

modules/sdk-coin-xlm/src/xlm.ts
```diff
@@ -1067,6 +1062,95 @@ export class Xlm extends BaseCoin {
...
+    return issuer;
+  }
+
+  // first to verify
+  verifyTxType(txParams: TransactionParams, operations: stellar.Operation[]): void {
+    operations.forEach((operation) => {
+      if (!operation.type) throw new Error('Missing operation type on token enablement');
+      if (operation.type !== 'changeTrust')
+        throw new Error(`Invalid operation on token enablement: expected changeTrust, got ${operation.type}`);
+    });
```
mohammadalfaiyazbitgo, line 1103: why not use `getTrustlineOperationLimitOrThrow` here?

modules/sdk-coin-xlm/src/xlm.ts
```diff
@@ -1067,6 +1062,95 @@ export class Xlm extends BaseCoin {
...
+  }
+
+  verifyAssetType(txParams: TransactionParams, operations: stellar.Operation[]): void {
+    operations.forEach((operation) => {
+      const line = this.getTrustlineOperationLineOrThrow(operation);
+      const assetType = line.getAssetType();
+      if (assetType === 'liquidity_pool_shares')
+        throw new Error(`Invalid asset type on token enablement: got ${assetType}`);
+    });
```
mohammadalfaiyazbitgo, line 1113: could we use `isOperationLineOfAssetType` here?

mohammadalfaiyazbitgo (changes requested):

modules/sdk-coin-xlm/src/xlm.ts
mtexeira-simtlix: Oh totally, I double checked since I thought that I made the same mistake than in solana but look at this piece of code at the start of verifyTransaction (before we even assets if it's a trusline op, a token enablement, etc):

image

modules/sdk-coin-xlm/src/xlm.ts
mtexeira-simtlix: We can for sure! But I wanted to maintain the checks (the verify functions) as specific as possible, if we do that, we're mixing checking if the op comes with limits for which we've a specific check later.

Please let me know if you still want the change, thanks @mohammadalfaiyazbitgo !

modules/sdk-coin-xlm/src/xlm.ts
mtexeira-simtlix: Yup we can, there is an issue however and it's that we're changing 2 code lines for other 2 code lines since we still need to call line.getAssetType() in order to get the wrong asset name to display in the error:
image

One solution is to change isOperationLineOfAssetType to return the assetType but that brings a different problem as isOperationLineOfAssetType functions like a type guard in another func:

image
as line could be stellar.Asset | stellar.LiquidityPoolAsset and we need to assert that we've a stellar.Asset in order to call getTokenNameFromStellarAsset.

I could change the code to something like this tho, let me know if that's alright  please and I'll send a new commit:

image
(the highlighted lines are the new ones)

modules/sdk-coin-xlm/src/xlm.ts
mohammadalfaiyazbitgo: I'd prefer if we reused function where possible since the check here is identical no?

modules/sdk-coin-xlm/src/xlm.ts
mohammadalfaiyazbitgo: I think I prefer the above changes, it's easier to read IMO.

modules/sdk-coin-xlm/src/xlm.ts
mtexeira-simtlix: As we discussed, we're gonna refactor to reuse the most checks that we can.

mohammadalfaiyazbitgo (approved):

arvind-bg (approved):

PR #7008, opened 2025-09-17
feat: removing deprecated testnet kovan

singhlovepreetbitgo9:
## Description

we call this api https://app.bitgo-test.com/api/v1/assets/list/testnet for the testnets and we are getting this below response where we receive testnets and the order of matching prefix is 
const prefixes = ['t', 'g', 'ht', 'gt'];  https://github.com/BitGo/dev-portal/blob/29a8c65a5945e59544bc75c489bda6fd4a1e584d/app/components/modules/coinsTable/SupportedCoinsTable.tsx#L60

As we have “teth” in response its being matched first and gets render in the table

### DIFF
--- a/modules/statics/src/allCoinsAndTokens.ts
+++ b/modules/statics/src/allCoinsAndTokens.ts
@@ -84,7 +84,6 @@ import {
   DOT_FEATURES,
   EOS_FEATURES,
   ETC_FEATURES,
-  ETH_FEATURES,
   ETH_FEATURES_WITH_STAKING_AND_MMI,
   EVM_FEATURES,
   EVM_NON_EIP1559_FEATURES,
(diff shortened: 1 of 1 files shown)


mukeshsp (approved):

mukeshsp (commented):
Please fix the failed testcases

modules/statics/src/allCoinsAndTokens.ts
```diff
@@ -305,16 +304,6 @@ export const allCoinsAndTokens = [
       CoinFeature.CUSTODY_BULK_TRANSACTION,
     ]
   ), // we should probably refactor this into a eth() method
-  account(
```
shivam-bitgo, line 308: are we keeping goerli?

PR #7020, opened 2025-09-18
feat(sdk-core): add transaction validation for (consolidate) unspent management

danielpeng1:
#### Description:
- Reordered + refactored transaction prebuild validation to occur after keychain retrieval, ensuring `verifyTransaction` has access to required keychain data for proper spoofing protection (for both consolidate and fanout)
- Added unit test for consolidate unspent management spoofability protection in `test/btc.ts` (for UTXO coins)
- Verifies that platform responses are validated against user parameters and covers BUILD_SIGN_SEND flow

Ticket: WP-5945

### DIFF
--- a/modules/sdk-coin-btc/test/unit/btc.ts
+++ b/modules/sdk-coin-btc/test/unit/btc.ts
@@ -106,4 +106,74 @@ describe('BTC:', () => {
       );
     });
   });
+
+  describe('Unspent management spoofability (BUILD_SIGN_SEND)', () => {
+    let coin: Tbtc;
+    let bitgoTest: TestBitGoAPI;
+    before(() => {
+      bitgoTest = TestBitGo.decorate(BitGoAPI, { env: 'test' });
+      bitgoTest.safeRegister('tbtc', Tbtc.createInstance);
+      bitgoTest.initializeTestVars();
+      coin = bitgoTest.coin('tbtc') as Tbtc;
+    });
+
+    it('should detect hex spoofing in BUILD_SIGN_SEND', async (): Promise<void> => {
+      const { getDefaultWalletKeys, toKeychainObjects } = require('../../../bitgo/test/v2/unit/coins/utxo/util/keychains');
+      const rootWalletKey = getDefaultWalletKeys();
+      const keysObj = toKeychainObjects(rootWalletKey, 'pass');
+
+      const { Wallet } = await import('@bitgo/sdk-core');
+      const wallet = new Wallet(bitgoTest, coin, {
+        id: '5b34252f1bf349930e34020a',
+        coin: 'tbtc',
+        keys: keysObj.map((k) => k.id),
+      });
+
+      const originalPsbt = utxolib.testutil.constructPsbt(
+        [{ scriptType: 'p2wsh' as const, value: BigInt(10000) }],
+        [{ address: 'tb1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3q0sl5k7', value: BigInt(9000) }],
+        coin.network,
+        rootWalletKey,
+        'unsigned' as const
+      );
+      utxolib.bitgo.addXpubsToPsbt(originalPsbt, rootWalletKey);
+
+      const spoofedPsbt = utxolib.testutil.constructPsbt(
+        [{ scriptType: 'p2wsh' as const, value: BigInt(10000) }],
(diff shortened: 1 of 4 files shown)


kamleshmugdiya (changes requested):
can do this in parallel using Promise.all

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -735,6 +735,18 @@ export class Wallet implements IWallet {
...
+    const txPrebuilds = Array.isArray(buildResponse) ? buildResponse : [buildResponse];
+    for (const txPrebuild of txPrebuilds) {
+      await this.baseCoin.verifyTransaction({
+        txParams: params,
+        txPrebuild,
+        wallet: this,
+        verification: params.verification ?? {},
+        reqId,
+      });
+    }
```
kamleshmugdiya: can do this in parallel using `Promise.all`

```suggestion
 await Promise.all(
  txPrebuilds.map((txPrebuild) =>
    this.baseCoin.verifyTransaction({
      txParams: params,
      txPrebuild,
      wallet: this,
      verification: params.verification ?? {},
      reqId,
    })
  )
);
```

kamleshmugdiya (changes requested):
can do this in parallel using `Promise.all`

modules/sdk-core/src/bitgo/wallet/wallet.ts
danielpeng1: Thanks, using this now

nvjsr (dismissed):
LGTM

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -739,6 +739,29 @@ export class Wallet implements IWallet {
       .keychains()
       .getKeysForSigning({ wallet: this, reqId })) as unknown as Keychain[];
 
+    // Validate that the platform-built transaction matches user parameters
+    const txPrebuilds = Array.isArray(buildResponse) ? buildResponse : [buildResponse];
+    await Promise.all(
+      txPrebuilds
+        .filter((txPrebuild) => txPrebuild.txHex)
```
mohammadalfaiyazbitgo: we shouldn't filter out `prebuilds` without the `txHex`

mohammadalfaiyazbitgo: those prebuilds should fail

mohammadalfaiyazbitgo (changes requested):

modules/sdk-core/src/bitgo/wallet/wallet.ts
danielpeng1: Got it, I removed the filter and added a stub for `verifyTransaction` in an `unspents` unit test that was failing (looks fine since it's already stubbing `signTransaction` + just testing some param handling, not validation).

mohammadalfaiyazbitgo (dismissed):

modules/sdk-coin-btc/test/unit/btc.ts
```diff
@@ -8,6 +8,8 @@ import { Tbtc } from '../../src';
 import { BitGoAPI } from '@bitgo/sdk-api';
 import * as utxolib from '@bitgo/utxo-lib';
 
+const { getDefaultWalletKeys, toKeychainObjects } = require('../../../bitgo/test/v2/unit/coins/utxo/util/keychains');
```
OttoAllmendinger: why `require` instead of import?

OttoAllmendinger (changes requested):

modules/sdk-coin-btc/test/unit/btc.ts
danielpeng1: Seems like `import` won't work unless I add the `bitgo` module as a project reference in `tsconfig.json` to allow cross-module imports.

OttoAllmendinger: you probably want to move these testutils out of test/v2/unit/coins if you want to use them here

danielpeng1: I made a small refactor to the tests to not require these.

mohammadalfaiyazbitgo (approved):

lokesh-bitgo (approved):
Looks good

OttoAllmendinger (approved):

kamleshmugdiya (approved):

PR #7024, opened 2025-09-18
fix(express): signPayload API to handle stringified payload as req

parvaparikh-30:
Ticket: GNA-2162


modules/express/test/unit/clientRoutes/signPayload.ts
```diff
@@ -59,6 +60,27 @@ describe('Sign an arbitrary payload with trading account key', function () {
       query: {},
     } as unknown as Request;
     await handleV2OFCSignPayload(req).should.be.resolvedWith(expectedResponse);
+    // console.log("should return a signed payload with type as object - test passed");
+  });
+  it('should return a signed payload with type as json string', async function () {
+    // TODO(GO-1015): unskip test
+    return;
```
seanchen0818: Why are we skipping this test?

modules/express/test/unit/clientRoutes/signPayload.ts
```diff
@@ -59,6 +60,27 @@ describe('Sign an arbitrary payload with trading account key', function () {
...
+    const req = {
+      bitgo: bitGoStub,
+      body: {
+        payload: stringifiedPayload,
+        walletId,
+      },
+      query: {},
+    } as unknown as Request;
+    await handleV2OFCSignPayload(req).should.be.resolvedWith(expectedResponse);
+    // console.log("should return a signed payload with type as json string - test passed");
```
seanchen0818: Please remove

modules/express/test/unit/clientRoutes/signPayload.ts
parvaparikh-30: https://bitgoinc.atlassian.net/browse/GO-1015
Its causing some problem in CI, it was already defined so kept it as it is
But ran locally, which ran perfectly, you can find it in description

modules/express/src/clientRoutes.ts
```diff
@@ -630,7 +639,7 @@ export async function handleV2OFCSignPayload(req: express.Request): Promise<{ pa
 
   const walletPassphrase = bodyWalletPassphrase || getWalletPwFromEnv(wallet.id());
   const tradingAccount = wallet.toTradingAccount();
-  const stringifiedPayload = JSON.stringify(req.body.payload);
+  const stringifiedPayload = isJsonString(req.body.payload) ? req.body.payload : JSON.stringify(req.body.payload);
```
zahin-mohammad, line 634: As a follow up, please consider migrating this api to an api-ts typed route! This way you can define your input codec and have your input validation done automatically!

zahin-mohammad: io-ts and io-ts-types already have codecs for json and json strings.

modules/express/test/unit/clientRoutes/signPayload.ts
zahin-mohammad: > Its causing some problem in CI,
> But ran locally, which ran perfectly,

This is not an acceptable approach. Please fix the failing test.

zahin-mohammad: If the test failure is consistent, that points to this PR as the cause.

zahin-mohammad (changes requested):

modules/express/test/unit/clientRoutes/signPayload.ts
parvaparikh-30: Thanks Zahin for the feedback, it was from different team so kept the same
But I will look into the issue & rectify it

modules/express/src/clientRoutes.ts
```diff
@@ -589,6 +589,15 @@ export async function handleV2OFCSignPayloadInExtSigningMode(
   }
 }
 
+const isJsonString = (str: any): boolean => {
```
mmcshinsky-bitgo: Can we move one-off functions like this out to a util file? clientRoutes is huge as it is already 😭

modules/express/src/clientRoutes.ts
stevenharrisBG: Yeah, that was my original thoughts is that the payload strings could be decoded using same codec as the payload in the request bodies (super safe and well typed).

modules/express/test/unit/clientRoutes/signPayload.ts
parvaparikh-30: Resolved the issue

PR #7028, opened 2025-09-19
feat: verify psbt withdraw lightning

parvbitgo067:
- parse funded psbt
- parse outputs of the psbt
- verify the recipients and corresponding amounts are coorect
- verify the change address

Reference: https://github.com/lightningnetwork/lnd/blob/master/docs/remote-signing.md#required-accounts

TICKET: BTC-2470

### DIFF
--- a/modules/abstract-lightning/src/lightning/parseWithdrawPsbt.ts
+++ b/modules/abstract-lightning/src/lightning/parseWithdrawPsbt.ts
@@ -1,0 +1,145 @@
+import * as utxolib from '@bitgo/utxo-lib';
+import { Psbt } from '@bitgo/utxo-lib';
+import { WatchOnlyAccount, WithdrawBaseOutputUTXO } from '../codecs';
+import { LightningOnchainRecipient } from '@bitgo/public-types';
+import { Bip32Derivation } from 'bip174/src/lib/interfaces';
+
+function parseDerivationPath(derivationPath: string): {
+  purpose: number;
+  change: number;
+  addressIndex: number;
+} {
+  const pathSegments = derivationPath.split('/');
+  const purpose = Number(pathSegments[1].replace(/'/g, ''));
+  const change = Number(pathSegments[pathSegments.length - 2]);
+  const addressIndex = Number(pathSegments[pathSegments.length - 1]);
+  return { purpose, change, addressIndex };
+}
+
+function parsePsbtOutputs(psbt: Psbt, network: utxolib.Network): WithdrawBaseOutputUTXO<bigint>[] {
+  const parsedOutputs: WithdrawBaseOutputUTXO<bigint>[] = [];
+  let bip32Derivation: Bip32Derivation | undefined;
+
+  for (let i = 0; i < psbt.data.outputs.length; i++) {
+    const output = psbt.data.outputs[i];
+    const txOutput = psbt.txOutputs[i];
+
+    let address = '';
+    const value = txOutput.value;
+    let isChange = false;
+
+    if (output.bip32Derivation && output.bip32Derivation.length > 0) {
+      isChange = true;
+      bip32Derivation = output.bip32Derivation[0];
+    }
+    if (txOutput.script) {
+      address = utxolib.address.fromOutputScript(txOutput.script, network);
+    }
+    const valueBigInt = BigInt(value);
+
+    parsedOutputs.push({
+      address,
+      value: valueBigInt,
+      change: isChange,
+      bip32Derivation,
+    });
+  }
+
+  return parsedOutputs;
+}
+
+function verifyChangeAddress(
+  output: WithdrawBaseOutputUTXO<bigint>,
+  accounts: WatchOnlyAccount[],
+  network: utxolib.Network
+): void {
+  if (!output.bip32Derivation || !output.bip32Derivation.path) {
+    throw new Error(`bip32Derivation path not found for change address`);
+  }
+  // derivation path example: m/84'/0'/0'/1/0
+  const { purpose, change, addressIndex } = parseDerivationPath(output.bip32Derivation.path);
+
+  // Find the corresponding account using the purpose
+  const account = accounts.find((acc) => acc.purpose === purpose);
+  if (!account) {
+    throw new Error(`Account not found for purpose: ${purpose}`);
+  }
+
+  // Create a BIP32 node from the xpub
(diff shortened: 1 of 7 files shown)


modules/abstract-lightning/src/lightning/parseWithdrawPsbt.ts
```diff
@@ -0,0 +1,167 @@
...
+          continue;
+      }
+      if (address) {
+        return address;
+      }
+    } catch (e) {
+      // Do nothing, just try the next type
+    }
+  }
+  return '';
```
OttoAllmendinger: `return undefined` would be better

modules/abstract-lightning/src/wallet/lightning.ts
```diff
@@ -364,6 +370,32 @@ export class LightningWallet implements ILightningWallet {
...
+
+    try {
+      validatePsbtForWithdraw(
+        transactionRequestCreate.transactions[0].unsignedTx.serializedTxHex,
+        network,
+        params.recipients,
+        watchOnlyAccountDetails.accounts
+      );
+    } catch (err: any) {
+      throw new Error(`error validating withdraw psbt: ${err.message}`);
```
OttoAllmendinger: you are dropping the stack here, add log the error with the full stack as well here some way

modules/abstract-lightning/src/wallet/lightning.ts
```diff
@@ -364,6 +370,32 @@ export class LightningWallet implements ILightningWallet {
...
+    if (!walletData.coinSpecific.watchOnlyAccounts) {
+      throw new Error(`wallet is missing watch only accounts`);
+    }
+
+    const watchOnlyAccountDetails = decodeOrElse(
+      WatchOnly.name,
+      WatchOnly,
+      walletData.coinSpecific.watchOnlyAccounts,
+      (errors) => {
+        throw new Error(`invalid watch only accounts`);
```
OttoAllmendinger: isn't this also dropping the error details?

modules/abstract-lightning/src/lightning/parseWithdrawPsbt.ts
```diff
@@ -0,0 +1,167 @@
+import * as utxolib from '@bitgo/utxo-lib';
+import { Psbt } from '@bitgo/utxo-lib';
+import { WatchOnlyAccount, WithdrawBaseOutputUTXO } from '../codecs';
+import { LightningOnchainRecipient } from '@bitgo/public-types';
+import { Bip32Derivation } from 'bip174/src/lib/interfaces';
+
+const paymentTypes = ['p2wpkh', 'p2sh', 'p2tr', 'p2pkh', 'p2wsh'];
+
+function parsePaymentAddress(script: Buffer, network: utxolib.Network) {
```
OttoAllmendinger: please add a jsdoc and test for this

modules/abstract-lightning/src/lightning/parseWithdrawPsbt.ts
```diff
@@ -0,0 +1,167 @@
...
+          break;
+        case 'p2wpkh':
+          address = utxolib.payments.p2wpkh({ output: script, network }).address;
+          break;
+        case 'p2wsh':
+          address = utxolib.payments.p2wsh({ output: script, network }).address;
+          break;
+        case 'p2sh':
+          address = utxolib.payments.p2sh({ output: script, network }).address;
+          break;
```
OttoAllmendinger: to my understanding this creates a p2wsh around `script` here - so for instances for a 2of3 `p2wsh` will always succeed and we can't really reach the p2sh case

I don't really understand what this function does

modules/abstract-lightning/src/lightning/parseWithdrawPsbt.ts
```diff
@@ -0,0 +1,167 @@
...
+  if (!account) {
+    throw new Error(`Account not found for purpose: ${purpose}`);
+  }
+
+  // Create a BIP32 node from the xpub
+  const xpubNode = utxolib.bip32.fromBase58(account.xpub, network);
+
+  // Extract the change and address_index from the path (e.g., m/84'/0'/0'/1/0 -> 1 and 0)
+  const change = Number(pathSegments[pathSegments.length - 2]);
+  const addressIndex = Number(pathSegments[pathSegments.length - 1]);
```
OttoAllmendinger: we're reading a lot of structure into these derivation paths, it probably makes sense to have a `parseDerivationPath(path: string): { purpose: number; isChange: boolean; addressIndex: number } ` with tests

modules/abstract-lightning/src/lightning/parseWithdrawPsbt.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  // Derive the public key from the xpub using the change and address index
+  const derivedPubkey = xpubNode.derive(change).derive(addressIndex).publicKey;
+
+  if (derivedPubkey.toString('hex') !== output.bip32Derivation.pubkey.toString('hex')) {
+    throw new Error(`Derived pubkey does not match for address: ${output.address}`);
```
OttoAllmendinger: the pubkey mismatch would also be interesting

modules/abstract-lightning/src/lightning/parseWithdrawPsbt.ts
```diff
@@ -0,0 +1,167 @@
...
+      break;
+    case 86: // P2TR (Taproot)
+      derivedAddress = utxolib.payments.p2tr({
+        pubkey: derivedPubkey,
+        network,
+      }).address;
+      break;
+    default:
+      throw new Error(`Unsupported purpose: ${purpose}`);
+  }
```
OttoAllmendinger, line 108: this would also be a nice function that we could write tests for

OttoAllmendinger (changes requested):

modules/abstract-lightning/src/lightning/parseWithdrawPsbt.ts
parvbitgo067: I am trying to get the address for each of the paymentType possible. We don't initially know what type it is right? It will try to get the address for each of the type. I actually got the idea from this code: https://github.com/BitGo/BitGoJS/blob/cecb4d46208c2b7018fcd2eef79cd52649572df5/modules/utxo-bin/src/ScriptParser.ts#L88-L96

OttoAllmendinger (approved):

zahin-mohammad (approved):

PR #7051, opened 2025-09-22
fix: replaced vulnerable request with axios

yanxue-22:
TICKET: DX-1558

The tests in the "bitgo" module used the request package, which is deprecated and vulnerable to GHSA-p8p7-x288-28g6. This PR attempts to replace the request with a more modern HTTP client library, Axios. Axios is taken as an alternative to a first party HTTP client like fetch because it has a cleaner migration from request-promise. 

Another PR under this ticket will address the request package being transitively depended on. 

Note: the unit-tests failing happen in modules/bitgo were happening prior to this commit.


therealdwright (commented):
please squash the commits and fix the typo in the commit message.

PR #7052, opened 2025-09-22
refactor: migrated generateWallet to typed routes

danielzhao122:
added keychain codecs to a general codec file, as I assume that these codecs could be reused (but not sure if they will). refactored the UT that covered generateWallet to ensure that encoding and decoding works e2e.

Ticket: WP-5415


modules/express/src/clientRoutes.ts
```diff
@@ -646,7 +646,7 @@ export async function handleV2OFCSignPayload(req: express.Request): Promise<{ pa
  * handle new wallet creation
  * @param req
  */
-export async function handleV2GenerateWallet(req: express.Request) {
+export async function handleV2GenerateWallet(req: ExpressApiRouteRequest<'express.wallet.generate', 'post'>) {
   const bitgo = req.bitgo;
   const coin = bitgo.coin(req.params.coin);
   const result = await coin.wallets().generateWallet(req.body);
```
zahin-mohammad: shouldn't this be req.decoded?

modules/express/src/typedRoutes/api/v2/generate.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+export const GenerateWalletBody = {
+  /** Wallet label */
+  label: t.string,
+  /** Enterprise id. This is required for Ethereum wallets since they can only be created as part of an enterprise */
+  enterprise: t.string,
+  /** If absent, BitGo uses the default wallet type for the asset */
+  multisigType: optional(t.union([t.literal('onchain'), t.literal('tss'), t.literal('blsdkg')])),
+  /** The type of wallet, defined by key management and signing protocols. 'hot' and 'cold' are both self-managed wallets. If absent, defaults to 'hot' */
+  type: optional(t.union([t.literal('hot'), t.literal('cold'), t.literal('custodial')])),
```
zahin-mohammad: Can you move these out to their own codecs (in different files under /schema)

modules/express/src/typedRoutes/schemas/keychain.ts
```diff
@@ -0,0 +1,41 @@
+import * as t from 'io-ts';
```
zahin-mohammad, line 1: Can you define these codecs in 

`modules/express/src/typedRoutes/schema`

modules/express/test/unit/clientRoutes/generateWallet.ts
```diff
@@ -1,72 +1,213 @@
...
+
     const reqIncludeKeychains = {
-      bitgo: stubBitgo,
+      bitgo,
       params: {
         coin,
       },
       query: {
-        includeKeychains: true,
+        includeKeychains: 'true',
```
zahin-mohammad: boolean doesn't work anymore?

modules/express/src/typedRoutes/api/v2/generateWallet.ts
```diff
@@ -0,0 +1,114 @@
...
+  coin: t.string,
+};
+
+/**
+ * Query parameters for wallet generation.
+ * @property includeKeychains - Include user, backup and bitgo keychains along with generated wallet
+ */
+export const GenerateWalletV2Query = {
+  /** Include user, backup and bitgo keychains along with generated wallet */
+  includeKeychains: optional(t.string),
```
zahin-mohammad: https://gcanti.github.io/io-ts-types/modules/BooleanFromString.ts.html

maybe we use this
```suggestion
  includeKeychains: optional(t.union([t.string.pipe(BooleanFromString), t.boolean])),
```

danielzhao122: I don't think t.boolean is necessary because query params are strings. So we only have to consider 'true' and 'false.

modules/express/test/unit/clientRoutes/generateWallet.ts
danielzhao122: Only booleans here. Because the codec should decode into boolean values.

modules/express/src/clientRoutes.ts
```diff
@@ -646,11 +646,11 @@ export async function handleV2OFCSignPayload(req: express.Request): Promise<{ pa
...
  * @param req
  */
-export async function handleV2GenerateWallet(req: express.Request) {
+export async function handleV2GenerateWallet(req: ExpressApiRouteRequest<'express.wallet.generate', 'post'>) {
   const bitgo = req.bitgo;
-  const coin = bitgo.coin(req.params.coin);
+  const coin = bitgo.coin(req.decoded.coin);
   const result = await coin.wallets().generateWallet(req.body);
-  if (req.query.includeKeychains === 'false') {
+  if ((req.decoded.includeKeychains as any) === false) {
```
zahin-mohammad, line 644: Why are we casting to any here?

danielzhao122: This is because `includeKeychains` is type `BooleanFromString` which has output type string so statically it is typed as a string but during runtime it is actually a boolean. I think this is because of how we setup the decoder.

Also `includeKeychains` is a parameter in the request and `httpRequest` requires that `params` are of type string

modules/express/src/typedRoutes/api/v2/generateWallet.ts
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Request body for wallet generation.
+ */
+export const GenerateWalletBody = {
+  /** Wallet label */
+  label: t.string,
+  /** Enterprise id. This is required for Ethereum wallets since they can only be created as part of an enterprise */
+  enterprise: t.string,
+  /** If absent, BitGo uses the default wallet type for the asset */
+  multisigType: multisigType,
```
zahin-mohammad: The command and the codec here seem to conflict. The codec makes this a required field. The comment says this should be optional.

modules/express/src/typedRoutes/api/v2/generateWallet.ts
```diff
@@ -0,0 +1,115 @@
...
+  /** The type of wallet, defined by key management and signing protocols. 'hot' and 'cold' are both self-managed wallets. If absent, defaults to 'hot' */
+  type: walletType,
```
zahin-mohammad: Same here.

modules/express/src/typedRoutes/api/v2/generateWallet.ts
```diff
@@ -0,0 +1,115 @@
...
+import { multisigType, walletType } from '../../schemas/wallet';
+
+/**
+ * Request body for wallet generation.
+ */
+export const GenerateWalletBody = {
+  /** Wallet label */
+  label: t.string,
+  /** Enterprise id. This is required for Ethereum wallets since they can only be created as part of an enterprise */
+  enterprise: t.string,
```
zahin-mohammad: Enterprise is optional in `GenerateWalletOptions`

zahin-mohammad: I think this might actually be required, but to avoid introducing a breaking change for this change we should map the codec to what `GenerateWalletOptions` defines.

zahin-mohammad (changes requested):

modules/express/src/clientRoutes.ts
ericcrosson-bitgo: Has this uncovered a bug in `GenerateWalletOptions`? 

When I remove the `as any` there are two compiler errors shown, but both errors (on `disableTransactionNotifications` and `gasPrice`) seem actually inconsistent with other uses and types in the code base. This patch, aligning `GenerateWalletOptions` with other types, fixes all compilation errors. 


```patch
diff --git i/modules/express/src/clientRoutes.ts w/modules/express/src/clientRoutes.ts
index 00b89566a..adfc0d262 100755
--- i/modules/express/src/clientRoutes.ts
+++ w/modules/express/src/clientRoutes.ts
@@ -640,7 +640,7 @@ export async function handleV2OFCSignPayload(
 export async function handleV2GenerateWallet(req: ExpressApiRouteRequest<'express.wallet.generate', 'post'>) {
   const bitgo = req.bitgo;
   const coin = bitgo.coin(req.decoded.coin);
-  const result = await coin.wallets().generateWallet(req.decoded as any);
+  const result = await coin.wallets().generateWallet(req.decoded);
   if ((req.decoded.includeKeychains as any) === false) {
     return result.wallet.toJSON();
   }
diff --git i/modules/sdk-core/src/bitgo/wallet/iWallets.ts w/modules/sdk-core/src/bitgo/wallet/iWallets.ts
index a1a3ea929..958aa8a3e 100644
--- i/modules/sdk-core/src/bitgo/wallet/iWallets.ts
+++ w/modules/sdk-core/src/bitgo/wallet/iWallets.ts
@@ -51,8 +51,8 @@ export interface GenerateWalletOptions {
   backupXpubProvider?: string;
   passcodeEncryptionCode?: string;
   enterprise?: string;
-  disableTransactionNotifications?: string;
-  gasPrice?: string;
+  disableTransactionNotifications?: boolean;
+  gasPrice?: number;
   eip1559?: {
     maxFeePerGas: string;
     maxPriorityFeePerGas: string;
```

zahin-mohammad (commented):
Merge conflicts

zahin-mohammad (commented):
unresolved comment https://github.com/BitGo/BitGoJS/pull/7052/files#r2446122581

zahin-mohammad (approved):

PR #7055, opened 2025-09-23
Dx 1558 replace vulnerable request package with axios

yanxue-22:
TICKET: [DX-1558](https://bitgoinc.atlassian.net/browse/DX-1558)

The tests in the "bitgo" module used the request package, which is deprecated and vulnerable to https://github.com/advisories/GHSA-p8p7-x288-28g6. This PR attempts to replace the request with a more modern HTTP client library, Axios. Axios is taken as an alternative to a first party HTTP client like fetch because it has a cleaner migration from request-promise.

`npm ls request` was ran to confirm that this PR resolved all direct dependencies on the vulnerable request package.

We can have more confidence it works because this change was to the unit tests for modules/bitgo. If the changes were to be broken, the Unit tests in the CI checks would have also failed.

Another PR under this ticket will address the request package being transitively depended on, and after this point, the security vulnerabilities for the request package will disappear.  

[DX-1558]: https://bitgoinc.atlassian.net/browse/DX-1558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/bitgo/package.json
+++ b/modules/bitgo/package.json
@@ -154,9 +154,7 @@
     "keccak": "3.0.3",
     "libsodium-wrappers-sumo": "^0.7.9",
     "puppeteer": "2.1.1",
-    "q": "^1.1.2",
-    "request": "^2.88.0",
-    "request-promise": "^4.2.2"
+    "q": "^1.1.2"
   },
   "optionalDependencies": {
     "@ethereumjs/common": "^2.6.5",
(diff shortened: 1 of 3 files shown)


shenbenson (approved):

starfy84 (approved):
LGTM

pranavjain97 (approved):

PR #7066, opened 2025-09-23
DX-1549-Resolved-BigInt-Buffer-Vulnerability

tanjeemh:
**Mitigate GHSA-3gc7-fjrx-p6mg (high-severity buffer overflow in `bigint-buffer@<=1.1.5`) impacting Solana paths in BitGoJS.**

**What Changed:**
- Replaced transitive `bigint-buffer` with `@trufflesuite/bigint-buffer@1.1.10` via alias, ensuring all Solana dependencies resolve to the safe implementation.
- Ensured Solana deps resolve through versions that pick up the alias (e.g., `@solana/web3.js@1.95.8`, `@solana/spl-token@0.3.1` -> `@solana/buffer-layout-utils@0.2.0`).
- Added a lightweight runtime guard for BigInt<->Buffer conversions in `@bitgo/sdk-coin-sol` (`bigint-buffer-guard.ts`) to validate input lengths and reduce misuse risk. It's purpose is to enforce input type/length checks around BigInt buffer operations; imported it so validations run at module load
- Removed any audit suppression related to this advisory so future scans surface regressions.

**Motivation / Context:**
`bigint-buffer@<=1.1.5` has a buffer overflow in `toBigIntLE()`; the original package appears unmaintained. We migrate to a maintained alternative and add validation to protect Solana flows.

**Dependencies:**
Alias: `bigint-buffer@npm:@trufflesuite/bigint-buffer@1.1.10`
Solana libs as above; no code changes required outside the guard and dependency updates.

**Type of Change:**
 - Bug fix (non-breaking change which fixes a security issue)
 
**How Has This Been Tested?**
Refer to comment in JIRA Ticket: DX-1549, for all test results.

Ticket: DX-1549


modules/sdk-coin-sol/src/bigint-buffer-guard.ts
```diff
@@ -0,0 +1,28 @@
+try {
```
zahin-mohammad, line 4: can you add some docs here on why this file exists. It's not straightforward unless someone searches through the git blame.

zahin-mohammad (dismissed):

modules/sdk-coin-sol/src/bigint-buffer-guard.ts
tanjeemh: > * Added a lightweight runtime guard for BigInt<->Buffer conversions in `@bitgo/sdk-coin-sol` (`bigint-buffer-guard.ts`) to validate input lengths and reduce misuse risk.

Yep! It's in the description above^

mukeshsp (dismissed):

modules/sdk-coin-sol/src/bigint-buffer-guard.ts
zahin-mohammad: Right, but I meant like a comment in the code. 

Having important context outside the codebase makes the context hard to find (dig through a series of git blames).

zahin-mohammad (approved):

mukeshsp (approved):

PR #7077, opened 2025-09-24
fix: Remove Validation of coin/subnet while building move stake transaction

raksha-r7:
Ticket: SC-3247
Description: The necessity of passing the coin name/ the check the subnet and coin association is not required for tao move stake functionality.

Move stake transactions are about redistributing existing stake between hotkeys/subnets, not transferring different coin types. The transaction operates within the same TAO ecosystem.

### DIFF
--- a/modules/sdk-coin-tao/src/lib/moveStakeTransaction.ts
+++ b/modules/sdk-coin-tao/src/lib/moveStakeTransaction.ts
@@ -1,5 +1,5 @@
 import { Interface as SubstrateInterface, Transaction as SubstrateTransaction } from '@bitgo/abstract-substrate';
-import { InvalidTransactionError, TransactionRecipient } from '@bitgo/sdk-core';
+import { InvalidTransactionError } from '@bitgo/sdk-core';
 import { decode } from '@substrate/txwrapper-polkadot';
 import { MoveStakeTxData } from './iface';
 import utils from './utils';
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-tao/src/lib/moveStakeTransaction.ts
```diff
@@ -42,13 +42,11 @@ export class MoveStakeTransaction extends SubstrateTransaction {
...
       value: txMethod.alphaAmount,
-      coin: utils.getTaoTokenBySubnetId(txMethod.originNetuid).name,
     });
 
     this._outputs.push({
       address: txMethod.destinationHotkey,
       value: txMethod.alphaAmount,
-      coin: utils.getTaoTokenBySubnetId(txMethod.destinationNetuid).name,
     });
   }
```
hitansh-madan, line 52: since the inputs/outputs technically represent funds moved by the tx, only stripping off the coin would still suggest that native tao is being moved, which might be handled incorrectly by any service consuming these values. We should ideally push nothing in these arrays given that the funds aren't leaving that coldkey anyways.

modules/sdk-coin-tao/src/lib/moveStakeTransaction.ts
```diff
@@ -42,13 +42,11 @@ export class MoveStakeTransaction extends SubstrateTransaction {
...
       address: txMethod.originHotkey,
       value: txMethod.alphaAmount,
-      coin: utils.getTaoTokenBySubnetId(txMethod.originNetuid).name,
     });
 
     this._outputs.push({
       address: txMethod.destinationHotkey,
       value: txMethod.alphaAmount,
-      coin: utils.getTaoTokenBySubnetId(txMethod.destinationNetuid).name,
     });
```
hitansh-madan, line 50:
```suggestion
```

hitansh-madan (changes requested):

hitansh-madan (approved):

kamleshmugdiya (approved):

PR #7084, opened 2025-09-24
feature(express): migrated ofcSignPayload to typed routes

danielzhao122:
TICKET: WP-5443


modules/express/src/typedRoutes/api/v2/ofcSignPayload.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Sign an arbitrary payload using an OFC trading account key.
+ */
+export const OfcSignPayloadBody = {
+  /** The ID of the OFC wallet to sign the payload with. */
+  walletId: NonEmptyString,
+  /** The payload to sign. Will always decode into a JSON object. */
+  payload: t.union([Json, t.string.pipe(JsonFromString)]),
+  /** The passphrase to decrypt the user key. */
+  walletPassphrase: optional(t.string),
+};
```
zahin-mohammad:
```suggestion
} as const;
```

modules/express/src/typedRoutes/api/v2/ofcSignPayload.ts
```diff
@@ -0,0 +1,45 @@
...
+import { httpRoute, httpRequest, optional } from '@api-ts/io-ts-http';
+import { BitgoExpressError } from '../../schemas/error';
+
+/**
+ * Sign an arbitrary payload using an OFC trading account key.
+ */
+export const OfcSignPayloadBody = {
+  /** The ID of the OFC wallet to sign the payload with. */
+  walletId: NonEmptyString,
+  /** The payload to sign. Will always decode into a JSON object. */
```
zahin-mohammad:
```suggestion
  /** The payload to sign. The input can either be a stringified JSON, or a JSON object. */
```

These will eventually go into out api docs, lets write them form the perspective of an integrator for this api.

modules/express/test/unit/clientRoutes/signPayload.ts
```diff
@@ -68,10 +74,38 @@ describe('Sign an arbitrary payload with trading account key', function () {
...
+      },
+    } as unknown as ExpressApiRouteRequest<'express.ofc.signPayload', 'post'>;
+
+    const result = await handleV2OFCSignPayload(req);
+    decodeOrElse('OfcSignPayloadResponse200', OfcSignPayloadResponse[200], result, (_) => {
+      throw new Error(`Response did not match expected codec`);
+    });
+    result.should.eql(expected);
+    OfcSignPayloadResponse[200].is(result).should.be.true();
+  });
```
zahin-mohammad, line 114: Should we instead add the decode piece to the test above?

modules/express/src/clientRoutes.ts
```diff
@@ -612,14 +611,6 @@ export async function handleV2OFCSignPayload(req: express.Request): Promise<{ pa
...
   }
 
-  if (!payload) {
-    throw new ApiResponseError('Missing required field: payload', 400);
-  }
-
-  if (!walletId) {
-    throw new ApiResponseError('Missing required field: walletId', 400);
-  }
-
```
pranavjain97, line 622: Have you checked what the new `400` error message looks like with io-ts? Their default error msg's can be bit weird

pranavjain97 (approved):

modules/express/test/unit/clientRoutes/signPayload.ts
danielzhao122: I added it to the two tests above. 
I don't think it hurts to have it here because it ensures that the expected response matches the response codec.

modules/express/src/clientRoutes.ts
danielzhao122: Currently we return a BitGoExpressError with the default io-ts error message. Here is an example of a 400 error: 
`"Invalid value 12345 supplied to : httpRequest/body: Exact<({ address: string } & Partial<{ supportOldScriptHashVersion: (boolean | undefined) }>)>/0: { address: string }/address: string"`

modules/express/src/typedRoutes/api/v2/ofcSignPayload.ts
```diff
@@ -0,0 +1,45 @@
...
+  /** The string form of the JSON payload that was signed. */
+  payload: t.string,
+  /** Hex-encoded signature generated by the trading account key. */
+  signature: t.string,
+});
+
+/**
+ * Response for signing an arbitrary payload with an OFC wallet key.
+ */
+export const OfcSignPayloadResponse = {
```
pranavjain97, line 28: There can be potential unhandled 5xx errors, maybe worth adding that. Not sure how other APIs do that

pranavjain97 (dismissed):
lgtm, non-blocking

pranavjain97 (approved):

PR #7087, opened 2025-09-24
feat: upgrade to lerna 6

zahin-mohammad:
TICKET: WP-6047


nx.json
```diff
@@ -0,0 +1,59 @@
+{
```
zahin-mohammad, line 1: Running unit-test twice locally goes from  ~500s to ~2s

Once this is in a good spot locally, we can integrate a remote cache with CI

PR #7096, opened 2025-09-25
fix(sdk-coin-avaxp): update utxo selection to also check STAKEABLE_LOCK_OUT

noel-bitgo:
noticed that some utxo will be of type 22 with locktim 0 those are eligible to be used for transactions

TICKET: SC-3257

### DIFF
--- a/modules/sdk-coin-avaxp/src/lib/atomicTransactionBuilder.ts
+++ b/modules/sdk-coin-avaxp/src/lib/atomicTransactionBuilder.ts
@@ -11,7 +11,7 @@ import {
 } from 'avalanche/dist/apis/platformvm';
 import { Credential } from 'avalanche/dist/common';
 import { BuildTransactionError } from '@bitgo/sdk-core';
-import { SECP256K1_Transfer_Output } from './iface';
+import { SECP256K1_STAKEABLE_LOCK_OUT, SECP256K1_Transfer_Output } from './iface';
 
 /**
  * Cross-chain transactions (export and import) are atomic operations.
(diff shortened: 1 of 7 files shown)


modules/sdk-core/src/bitgo/staking/stakingWallet.ts
```diff
@@ -458,18 +459,6 @@ export class StakingWallet implements IStakingWallet {
...
-        !isEqual((explainedTransaction as any).aptosCustomTransactionParams, buildParams.aptosCustomTransactionParams)
-      ) {
-        mismatchErrors.push(
-          `Aptos custom transaction parameters mismatch. Expected: ${JSON.stringify(
-            buildParams.aptosCustomTransactionParams
-          )}, Got: ${JSON.stringify((explainedTransaction as any).aptosCustomTransactionParams)}`
-        );
-      }
-    }
-
```
balakrishna507: is this related? this isn't part of the log removals we're doing also as part of this pr, is it?

balakrishna507 (dismissed):

modules/sdk-core/src/bitgo/staking/stakingWallet.ts
noel-bitgo: this should not have been here in the first place, because we do not have aptos staking yet

CapnMigraine (dismissed):

zahin-mohammad (commented):
Added as a reviewer but don't own any of the files here. Im guessing the changes we where needed in have been removed.

CapnMigraine (dismissed):

theMobiusStrip (dismissed):

benkelcher-bitgo (approved):

feng-bitgo (approved):

PR #7097, opened 2025-09-25
feat(sdk-api): Add support for passing constants to the SDK as an optional parameter

danielpeng1:
Added optional clientConstants parameter to the BitGoAPI constructor, allowing users to provide/pass constants that override server-fetched values to the SDK.
- Notably, modified `fetchConstants()` to merge server constants with client constants (client overrides take precedence).
- Unit test in `test/unit/bitgoAPI.ts` passing in a constant under this flow.

Ticket: WP-5538

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -55,6 +55,7 @@ import {
   CalculateRequestHeadersOptions,
   CalculateRequestHmacOptions,
   ChangePasswordOptions,
+  Constants,
   DeprecatedVerifyAddressOptions,
   EstimateFeeOptions,
   ExtendTokenOptions,
@@ -275,6 +276,10 @@ export class BitGoAPI implements BitGoBase {
     this._baseApiUrlV2 = this._baseUrl + '/api/v2';
     this._baseApiUrlV3 = this._baseUrl + '/api/v3';
     this._token = params.accessToken;
+
+    const clientConstants = params.clientConstants;
+    this._initializeClientConstants(clientConstants);
+
     this._userAgent = params.userAgent || 'BitGoJS-api/' + this.version();
     this._reqId = undefined;
     this._refreshToken = params.refreshToken;
@@ -303,17 +308,34 @@ export class BitGoAPI implements BitGoBase {
 
     this._customProxyAgent = params.customProxyAgent;
 
-    // capture outer stack so we have useful debug information if fetch constants fails
-    const e = new Error();
+    // Only fetch constants from constructor if clientConstants was not provided
+    if (!clientConstants) {
+      // capture outer stack so we have useful debug information if fetch constants fails
+      const e = new Error();
 
-    // Kick off first load of constants
-    this.fetchConstants().catch((err) => {
-      if (err) {
-        // make sure an error does not terminate the entire script
-        console.error('failed to fetch initial client constants from BitGo');
-        debug(e.stack);
+      // Kick off first load of constants
+      this.fetchConstants().catch((err) => {
+        if (err) {
+          // make sure an error does not terminate the entire script
+          console.error('failed to fetch initial client constants from BitGo');
+          debug(e.stack);
+        }
+      });
+    }
+  }
+
+  /**
+   * Initialize client constants if provided.
+   * @param clientConstants - The client constants from params
+   * @private
+   */
+  private _initializeClientConstants(clientConstants: any): void {
+    if (clientConstants) {
+      if (!BitGoAPI._constants) {
+        BitGoAPI._constants = {};
       }
-    });
+      BitGoAPI._constants[this.env] = 'constants' in clientConstants ? clientConstants.constants : clientConstants;
+    }
   }
 
   /**
@@ -531,17 +553,15 @@ export class BitGoAPI implements BitGoBase {
    * but are unlikely to change during the lifetime of a BitGo object,
    * so they can safely cached.
    */
-  async fetchConstants(): Promise<any> {
+  async fetchConstants(): Promise<Constants> {
     const env = this.getEnv();
 
-    if (!BitGoAPI._constants) {
-      BitGoAPI._constants = {};
-    }
-    if (!BitGoAPI._constantsExpire) {
-      BitGoAPI._constantsExpire = {};
-    }
-
-    if (BitGoAPI._constants[env] && BitGoAPI._constantsExpire[env] && new Date() < BitGoAPI._constantsExpire[env]) {
+    // Check if we have cached constants that haven't expired
+    if (
+      BitGoAPI._constants &&
+      BitGoAPI._constants[env] &&
+      (!BitGoAPI._constantsExpire || !BitGoAPI._constantsExpire[env] || new Date() < BitGoAPI._constantsExpire[env])
+    ) {
       return BitGoAPI._constants[env];
     }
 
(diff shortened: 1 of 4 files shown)


modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -307,13 +321,35 @@ export class BitGoAPI implements BitGoBase {
...
+  ): { constants: Record<string, any>; ttl?: number } | undefined {
+    if (!clientConstants) {
+      return undefined;
+    }
+
+    if ('constants' in (clientConstants as any)) {
+      const constantsContainer = clientConstants as { constants: Record<string, any>; ttl?: number };
+      return {
+        constants: constantsContainer.constants,
+        ttl: typeof constantsContainer.ttl === 'number' ? constantsContainer.ttl : undefined,
```
mohammadalfaiyazbitgo: are we using this for anything?

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -105,6 +105,7 @@ export class BitGoAPI implements BitGoBase {
   protected _blockchain?: any;
   protected _travelRule?: any;
   protected _pendingApprovals?: any;
+  protected _clientConstants?: Record<string, any>;
```
mohammadalfaiyazbitgo: instead of a new field, when a user passes in a constants option, let's set the `_constants` field

modules/sdk-api/test/unit/bitgoAPI.ts
```diff
@@ -305,4 +305,17 @@ describe('Constructor', function () {
       result.should.containDeep(['wallet-id-2', 'wallet-id-4']);
     });
   });
+
```
mohammadalfaiyazbitgo, line 309: can we also add test to show that the constants are refreshed with new values once `if (cachedConstants && (!cacheExpiry || new Date() < cacheExpiry)) {` these conditions are met.

mohammadalfaiyazbitgo (changes requested):

modules/sdk-api/src/bitgoAPI.ts
danielpeng1: Removed and made logic more concise.

modules/sdk-api/src/bitgoAPI.ts
danielpeng1: Does this mean removing `protected static _constants: any;` (and `protected static _constantsExpire: any;`) in favour of a new `protected _constants?: Record<string, any>;`, and then updating `get/fetchConstants` implementations and wallet test mocks?

Edit: continuing with `protected static _constants: any;` and `protected static _constantsExpire: any;`

modules/sdk-api/test/unit/bitgoAPI.ts
danielpeng1: added a few tests covering this:
- BitGoAPI._constants[env] exists but new Date() < BitGoAPI._constantsExpire[env] is false (cache expired)
- when all parts of the condition are true
- when BitGoAPI._constantsExpire[env] is undefined, so !BitGoAPI._constantsExpire[env] is true

mohammadalfaiyazbitgo (dismissed):

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -302,18 +312,6 @@ export class BitGoAPI implements BitGoBase {
     }
 
     this._customProxyAgent = params.customProxyAgent;
-
```
bitgoAaron, line 305: this might be too risky to change at this point, we should only do the initial fetchConstants from the constructor if clientConstants is not provided to the SDK. Let's revert this change, and avoid introducing a breaking change.

modules/sdk-api/src/types.ts
```diff
@@ -2,6 +2,8 @@ import { EnvironmentName, IRequestTracer, V1Network } from '@bitgo/sdk-core';
 import { ECPairInterface } from '@bitgo/utxo-lib';
 import { type Agent } from 'http';
 
+export type Constants = Record<string, any>;
```
bitgoAaron, line 5: we can do a lot better than this. Let's see about getting https://github.com/BitGo/public-types/pull/243 in and then referencing the types from there instead

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -275,6 +275,16 @@ export class BitGoAPI implements BitGoBase {
...
+
+    const providedConstants = params.clientConstants;
+    if (providedConstants) {
+      if (!BitGoAPI._constants) {
+        BitGoAPI._constants = {};
+      }
+      BitGoAPI._constants[this.env] =
+        'constants' in providedConstants ? providedConstants.constants : providedConstants;
+    }
+
```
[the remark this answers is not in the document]
pranavjain97: I like this suggestion. The constructor is already too big to have more if blocks.
```
const clientConstants = params.clientConstants;
this._initializeClientConstants(clientConstants);
```

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -560,13 +572,23 @@ export class BitGoAPI implements BitGoBase {
...
+    BitGoAPI._constants[env] = _.merge({}, constants);
+
+    // Use TTL from server response if provided, otherwise default to 5 minutes
+    const ttlSeconds = result.body.ttl && typeof result.body.ttl === 'number' ? result.body.ttl : 300;
+    BitGoAPI._constantsExpire[env] = new Date(Date.now() + ttlSeconds * 1000);
 
-    return BitGoAPI._constants[env];
+    return _.merge({}, constants);
```
pranavjain97: Unless needed, can we ship this change separately? It could potentially have some unexpected downstream behavior - and we should limit the scope of this PR and test them separately.

pranavjain97 (changes requested):
Also, lets squash the commits please.

modules/sdk-api/src/bitgoAPI.ts
danielpeng1: Reverted with the condition you specified.

modules/sdk-api/src/bitgoAPI.ts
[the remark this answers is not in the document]
danielpeng1: Done 👍.

modules/sdk-api/src/bitgoAPI.ts
danielpeng1: Sounds good, reverted.

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -560,13 +556,23 @@ export class BitGoAPI implements BitGoBase {
...
+    }
+    if (!BitGoAPI._constantsExpire) {
+      BitGoAPI._constantsExpire = {};
     }
 
-    return BitGoAPI._constants[env];
+    BitGoAPI._constants[env] = _.merge({}, constants);
+
+    // Use TTL from server response if provided, otherwise default to 5 minutes
+    const ttlSeconds = result.body.ttl && typeof result.body.ttl === 'number' ? result.body.ttl : 300;
```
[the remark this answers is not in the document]
danielpeng1: Outdated from https://github.com/BitGo/BitGoJS/pull/7097#discussion_r2396519683

bitgoAaron (approved):

mohammadalfaiyazbitgo (approved):

pranavjain97 (approved):
lesgo!

davidkaplanbitgo (approved):

PR #7107, opened 2025-09-26
fix(express): remove hardcoded dockerfile labels

louib:
Removing the hardcoded labels from the Dockerfile will remove the need to create a new commit when publishing a new version of `bitgo/express`, which in turn will make the release process more streamlined. 

I'm also removing explicit references to `podman` from the build scripts since `docker` is used to produce the official releases of `bitgo/express`. The CI job that builds the docker image on PRs continues to use `podman` since it is proving to be significantly faster than `docker` (10 minutes instead of `docker` which errors out after 15 minutes). Unless we want to use a bigger runner, this is the best compromise I can think of.

I was able to verify that the labels are still populated correctly:
```
            "Labels": {
                "created": "2025-09-26T20:17:27Z",
                "git_hash": "69e6314ac0a27b9d09dc117c9ee78bf5f6bf4bc0",
                "version": "15.1.0"
            }
```

Ticket: VL-3498


.github/workflows/ci.yml
```diff
@@ -271,4 +271,4 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
       - name: Build BitGoJS Express Docker Image
-        run: podman build .
+        run: ./scripts/build-docker-express.sh
```
zahin-mohammad: Should we enforce that there is no diff in the dockerfile as well? That way devs are forced to commit changes to the dockerfile when adding or removing coins

zahin-mohammad (dismissed):

.github/workflows/ci.yml
zahin-mohammad: (can resolve this comment if we want to do this in a separate pr)

louib: > Should we enforce that there is no diff in the dockerfile as well? 

yes! I'll add a new CI job specifically for that in the next PR.

zahin-mohammad (approved):

therealdwright (approved):

.github/workflows/ci.yml
louib: Here it is -> https://github.com/BitGo/BitGoJS/pull/7110

PR #7109, opened 2025-09-26
feat(sdk-coin-sol): support versioned sol transactions with customTx intent

derekchen745:
Ticket: SC-3231

We currently only support rebuilding Solana custom transactions in the legacy format. This PR adds support for Solana’s versioned transactions (with Address Lookup Tables) to enable WalletConnect compatibility, since most DApps rely on versioned transactions (Jupiter, Adrena, Kamino, etc)


modules/sdk-core/src/bitgo/wallet/BuildParams.ts
```diff
@@ -112,6 +112,8 @@ export const BuildParams = t.exact(
       emergency: t.unknown,
       // Solana custom instructions for transaction building
       solInstructions: t.unknown,
+      // Solana versioned transaction data for transaction building
+      versionedTransactionData: t.unknown,
```
benkelcher-bitgo: we should maybe call this `solVersionedTransactionData` for clarity (seems other coin-specific fields are marked with the coin name)

benkelcher-bitgo:
@claude

modules/bitgo/package.json
```diff
@@ -26,7 +26,7 @@
     "webpack-dev": "cross-env NODE_ENV=development webpack",
     "webpack-prod": "NODE_OPTIONS=--max-old-space-size=4096 cross-env NODE_ENV=production webpack",
     "test": "npm run coverage",
-    "unit-test": "mocha 'test/v2/unit/**/*.ts' 'test/unit/**/*.ts'",
+    "unit-test": "NODE_OPTIONS=--max-old-space-size=8192 mocha 'test/v2/unit/**/*.ts' 'test/unit/**/*.ts'",
```
derekchen745, line 29: The bitgo module unit-tests were failing with `FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory`. I tried optimizing and reviewing my changes in `sdk-core`, but it doesn’t seem like they should be causing the memory issue, so I bumped the memory to unblock for now.

modules/sdk-coin-sol/test/unit/transactionBuilder/customInstructionBuilder.ts
```diff
@@ -230,4 +237,251 @@ describe('Sol Custom Instruction Builder', () => {
...
+      const tx = await txBuilder.build();
+
+      // Verify transaction properties
+      tx.should.be.ok();
+      tx.type.should.equal(31); // TransactionType.CustomTx enum value
+
+      // Should be able to get signable payload
+      const payload = tx.signablePayload;
+      payload.should.be.instanceOf(Buffer);
+      payload.length.should.be.greaterThan(0);
```
haritkapadia, line 424: Can you put the actual payload?
```suggestion
      const payload = tx.signablePayload;
      payload.should.be.instanceOf(Buffer);
      payload.length.should.be.greaterThan(0);
      payload.toString("hex").should.equal(testData.thePayload);
```

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -228,6 +251,11 @@ export interface PrebuildTransactionWithIntentOptions extends IntentOptionsBase
    * Each instruction should contain programId, keys, and data fields.
    */
   solInstructions?: SolInstruction[];
+  /**
+   * Deconstructed Versioned Transaction data for Solana customTx intent type.
+   * Contains compiled instructions, Address Lookup Tables, and static account keys.
+   */
+  solVersionedTransactionData?: SolVersionedTransactionData;
```
haritkapadia, line 258: It's weird that there's sol instructions in the base type... can we address refactoring this in a follow-up?

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
```diff
@@ -118,9 +118,15 @@ export abstract class MpcUtils {
...
+        params.solVersionedTransactionData &&
+        params.solVersionedTransactionData.versionedInstructions &&
+        params.solVersionedTransactionData.versionedInstructions.length > 0;
+
       assert(
-        params.solInstructions && params.solInstructions.length > 0,
-        `'solInstructions' is a required parameter for customTx intent`
+        hasSolInstructions || hasVersionedData,
+        `'solInstructions' or 'solVersionedTransactionData' is required for customTx intent`
       );
```
haritkapadia, line 130: Is this a user error? If so, throw Error instead of asserting.

Also, should this check be in `modules/sdk-coin-sol/src/sol.ts: setCoinSpecificFieldsInIntent()` instead?

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -181,6 +181,28 @@ export interface PrebuildTransactionOptions {
...
+  /**
+   * Solana versioned transaction data for building transactions with Address Lookup Tables.
+   * Contains compiled instructions, address lookup tables, static account keys, and message header.
+   */
+  solVersionedTransactionData?: {
+    versionedInstructions: {
+      programIdIndex: number;
+      accountKeyIndexes: number[];
+      data: string;
+    }[];
```
haritkapadia, line 193: Should this be `SolVersionedInstruction[]`?

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -47,6 +47,29 @@ export interface SolInstruction {
   data: string;
 }
 
+export interface SolVersionedInstruction {
+  programIdIndex: number;
+  accountKeyIndexes: number[];
+  data: string;
```
haritkapadia, line 53: Can you rename this to `dataHex` or similar?

modules/sdk-coin-sol/src/lib/transactionBuilder.ts
```diff
@@ -179,6 +205,85 @@ export abstract class TransactionBuilder extends BaseTransactionBuilder {
...
+  private buildFromVersionedData(data: {
+    versionedInstructions: Array<{ programIdIndex: number; accountKeyIndexes: number[]; data: string }>;
+    addressLookupTables: Array<{ accountKey: string; writableIndexes: number[]; readonlyIndexes: number[] }>;
+    staticAccountKeys: string[];
+    messageHeader: {
+      numRequiredSignatures: number;
+      numReadonlySignedAccounts: number;
+      numReadonlyUnsignedAccounts: number;
+    };
+  }): VersionedTransaction {
```
haritkapadia: Can you use the named types? For example, `versionedInstructions: SolVersionedInstruction[]`. Even better if you can replace the entire thing with `(data: ???)`

modules/sdk-coin-sol/src/lib/transaction.ts
```diff
@@ -178,6 +276,16 @@ export class Transaction extends BaseTransaction {
...
   /** @inheritdoc */
   toBroadcastFormat(): string {
+    if (this._versionedTransaction) {
+      try {
+        // VersionedTransaction.serialize() doesn't need requireAllSignatures parameter
+        // It automatically handles whatever signatures are present
+        return Buffer.from(this._versionedTransaction.serialize()).toString('base64');
+      } catch (e) {
+        throw e;
+      }
```
haritkapadia: Unnecessary try-catch?

modules/sdk-coin-sol/src/lib/transaction.ts
```diff
@@ -154,6 +230,28 @@ export class Transaction extends BaseTransaction {
...
+      const signers: Signer[] = [];
+      for (const kp of keyPairs) {
+        const keys = kp.getKeys(true);
+        if (!keys.prv) {
+          throw new SigningError('Missing private key');
+        }
+        signers.push({ publicKey: new PublicKey(keys.pub), secretKey: keys.prv as Uint8Array });
+      }
+      try {
+        this._versionedTransaction.sign(signers);
```
haritkapadia: Is this logic is a duplicate of `solTransaction`? If so, please deduplicate.

modules/sdk-coin-sol/src/lib/transaction.ts
```diff
@@ -154,6 +230,28 @@ export class Transaction extends BaseTransaction {
...
+      } catch (e) {
+        throw e;
+      }
```
haritkapadia: Unnecessary try-catch?

modules/sdk-coin-sol/src/lib/transaction.ts
```diff
@@ -109,9 +134,19 @@ export class Transaction extends BaseTransaction {
...
 
-    for (const solSignature of this._solTransaction.signatures) {
-      if (solSignature.signature) {
-        signatures.push(base58.encode(solSignature.signature));
+    if (this._versionedTransaction) {
+      // Handle VersionedTransaction signatures
+      for (const sig of this._versionedTransaction.signatures) {
+        if (sig && sig.length > 0) {
+          signatures.push(base58.encode(sig));
+        }
```
haritkapadia: Is it necessary to check `sig.length > 0`? If so, leave a comment explaining why.

modules/sdk-coin-sol/src/lib/transaction.ts
```diff
@@ -76,12 +86,27 @@ export class Transaction extends BaseTransaction {
...
   /** @inheritDoc */
   get signablePayload(): Buffer {
+    if (this._versionedTransaction) {
+      return Buffer.from(this._versionedTransaction.message.serialize());
+    }
     return this._solTransaction.serializeMessage();
   }
 
   /** @inheritDoc **/
   get id(): string {
```
haritkapadia, line 96: Nit: can you refactor this to:

```
if (this.versionedTransaction?.signatures && ...) {
  return base58.encode(...)
}

if (this._solTransaction.signature(...) { ... }

return UNAVAILABLE_TEXT
```

modules/sdk-coin-sol/src/lib/customInstructionBuilder.ts
```diff
@@ -41,13 +44,23 @@ export class CustomInstructionBuilder extends TransactionBuilder {
...
-  addCustomInstruction(instruction: SolInstruction): this {
+  addCustomInstruction(instruction: SolInstruction | SolVersionedInstruction): this {
     this.validateInstruction(instruction);
-    const customInstruction: CustomInstruction = {
-      type: InstructionBuilderTypes.CustomInstruction,
-      params: instruction,
-    };
-    this._customInstructions.push(customInstruction);
+
+    if ('programId' in instruction) {
```
haritkapadia: Can you find a better way to determine if an instruction is a `SolInstruction` or `SolVersionedInstruction`? E.g. an `instructionType` field in `SolVersionedInstruction`.

haritkapadia (changes requested):

modules/sdk-coin-sol/src/lib/customInstructionBuilder.ts
derekchen745: `programId` and `programIdIndex` are mutually exclusive between the two types, so this check is a simple and reliable way to distinguish them without adding extra fields

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
derekchen745: This field is actually base58 (not hex). The code treating it as hex was incorrect. I’ve updated both the implementation and the related tests to use base58 consistently. Also keeping it named `data` is consistent with the `solInstructions` naming

modules/sdk-core/src/bitgo/wallet/iWallet.ts
derekchen745: The current pattern in iWallet.ts redeclares related types inline (like solInstructions and aptosCustomTransactionParams), so I kept it consistent here

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
derekchen745: this is internal validation of the data pipeline and is for passing custom instructions using the customTx intent in the wallet layer which is why its in the sdk-core module

modules/sdk-coin-sol/src/lib/transaction.ts
derekchen745: no that check is not necessary. We do however need to check for placeholder signatures so I've replaced it with `sig.some(b => b !== 0)`

modules/sdk-coin-sol/src/lib/transaction.ts
derekchen745: good catch

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
derekchen745: sure, just following the existing structure atm (e.g `solInstructions`, `aptosCustomTransactionParams`), will make a ticket to address as a followup

modules/sdk-coin-sol/src/lib/customInstructionBuilder.ts
haritkapadia: Yes, but when someone decides to add one more instruction type, they will run into issues.

You can use an enum:
```
enum SolInstructionType {
  VERSIONED
}

interface SolVersionedInstruction {
  type: SolInstructionType.VERSIONED;
}
```

or alternatively a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates):
```
isVersionedBlahBlahBlah(instruction: SolInstruction | SolVersionedInstruction): instruction is SolVersionedInstruction {
  return 'programIdIndex' in instruction
}
```

modules/sdk-coin-sol/src/lib/transaction.ts
haritkapadia: Can you leave a comment that "0" means "placeholder signature"?

modules/sdk-coin-sol/src/lib/customInstructionBuilder.ts
derekchen745: added the type predicate check

haritkapadia (approved):

feng-bitgo (approved):

kaustubhbitgo (approved):

sbplat (approved):

benkelcher-bitgo (approved):

starfy84 (approved):

bitgopatmcl (approved):

PR #7111, opened 2025-09-27
fix: correct methods and abi for vechain

Gautam2305:
Ticket: SC-3296


modules/sdk-coin-vet/src/lib/constants.ts
```diff
@@ -3,12 +3,12 @@ export const VET_ADDRESS_LENGTH = 40;
...
+export const EXIT_DELEGATION_METHOD_ID = '0x69e79b7d';
+export const BURN_NFT_METHOD_ID = '0x2e17de78';
 export const TRANSFER_NFT_METHOD_ID = '0x23b872dd';
-export const CLAIM_BASE_REWARDS_METHOD_ID = '0x037402d3';
-export const CLAIM_STAKING_REWARDS_METHOD_ID = '0xeb2767fa';
+export const CLAIM_BASE_REWARDS_METHOD_ID = '0x858d50e8'; // claimVetGeneratedVtho(uint256)
+export const CLAIM_STAKING_REWARDS_METHOD_ID = '0x0962ef79'; // claimRewards(uint256)
 
 export const STARGATE_NFT_ADDRESS = '0x1856c533ac2d94340aaa8544d35a5c1d4a21dee7';
 export const STARGATE_DELEGATION_ADDRESS = '0x4cb1c9ef05b529c093371264fab2c93cc6cddb0e';
```
Vijay-Jagannathan, line 14: these are mainnet values I believe 

tried searching for the testnet contract addresses https://docs.stargate.vechain.org/for-developers/contracts#testnet but couldn't find them in SDK/WP/Staking service. Only found them in tests, can you please confirm if we are setting the right contract addresses for the respective environments ?

modules/sdk-coin-vet/src/lib/transactionBuilder/stakingBuilder.ts
```diff
@@ -88,7 +88,29 @@ export class StakingBuilder extends TransactionBuilder {
...
+  /**
+   * Sets the autorenew flag for this staking tx.
+   *
+   * @param {boolean} autorenew - Whether to enable autorenew
+   * @returns {StakingBuilder} This transaction builder
+   */
+  autorenew(autorenew: boolean): this {
+    this.stakingTransaction.autorenew = autorenew;
+    return this;
+  }
```
Vijay-Jagannathan: nit: we don't need a setter if we are defaulting to `true` for `stakeAndDelegate`

modules/sdk-coin-vet/src/lib/transactionBuilder/stakingBuilder.ts
```diff
@@ -127,7 +149,22 @@ export class StakingBuilder extends TransactionBuilder {
...
+          throw new Error('Invalid character');
+        }
+      } catch (e) {
+        throw new Error('Invalid character');
+      }
+    }
+
     assert(transaction.stakingContractABI, 'Staking contract ABI is required');
+    assert(transaction.levelId !== undefined && transaction.levelId !== null, 'Level ID is required');
+    assert(transaction.autorenew !== undefined && transaction.autorenew !== null, 'Autorenew flag is required');
```
Vijay-Jagannathan: nit: can use != null

modules/sdk-coin-vet/src/lib/utils.ts
```diff
@@ -125,22 +125,45 @@ export class Utils implements BaseUtils {
...
-   * @param {string} stakingAmount - The amount to stake in wei
+   * @param {number} levelId - The level ID for staking
+   * @param {boolean} autorenew - Whether to enable autorenew
    * @returns {string} - The encoded transaction data
    */
-  getStakingData(stakingAmount: string): string {
-    const methodName = 'stake';
-    const types = ['uint256'];
-    const params = [new BN(stakingAmount)];
+  getStakingData(levelId: number, autorenew = true): string {
```
Vijay-Jagannathan: this method is present in the respective builders for the other txn types, let's move it there for consistency

modules/sdk-coin-vet/src/lib/transaction/burnNftTransaction.ts
```diff
@@ -53,7 +53,7 @@ export class BurnNftTransaction extends Transaction {
    * @returns {string} The encoded transaction data as a hex string
    */
   private getBurnNftData(): string {
```
Vijay-Jagannathan, line 58: same method is there in the builder also, lets keep this method either the builder/txn (ideally wherever most appropriate since it'll only needed by one of them; double check this across all 3 txn types)

modules/sdk-coin-vet/test/transactionBuilder/stakingTransaction.ts
Vijay-Jagannathan, line 1: what about unstake tests, no changes needed ?

Vijay-Jagannathan (changes requested):
please double check once

modules/sdk-coin-vet/src/lib/constants.ts
Gautam2305: yes these are the mainnet values

modules/sdk-coin-vet/src/lib/utils.ts
Gautam2305: done

modules/sdk-coin-vet/src/lib/transaction/burnNftTransaction.ts
Gautam2305: yepp removed now

modules/sdk-coin-vet/test/transactionBuilder/stakingTransaction.ts
Gautam2305: nope no changes are needed there

modules/sdk-coin-vet/src/lib/transaction/claimRewards.ts
```diff
@@ -83,42 +80,26 @@ export class ClaimRewardsTransaction extends Transaction {
   }
 
   /**
-   * Get the delegation contract address to use for claims
-   * Uses the address from claimRewardsData if provided, otherwise falls back to default
-   */
-  private getDelegationAddress(): string {
-    return this._claimRewardsData.delegationContractAddress || STARGATE_DELEGATION_ADDRESS;
-  }
-
```
Gautam2305, line 92: moved these functions to utils

modules/sdk-coin-vet/src/lib/transactionBuilder/stakingBuilder.ts
```diff
@@ -81,14 +81,26 @@ export class StakingBuilder extends TransactionBuilder {
    * @param {string} address - The staking contract address
    * @returns {StakingBuilder} This transaction builder
    */
-  stakingContractAddress(address: string): this {
-    this.validateAddress({ address });
-    this.stakingTransaction.stakingContractAddress = address;
+  stakingContractAddress(address?: string): this {
+    const contractAddress = address || utils.getDefaultStakingAddress(this._coinConfig);
```
Vijay-Jagannathan: this would work only if the intent being passed from staking service -> WP has the correct address. Otherwise we might be using the incorrect address.

Make sure to use the correct value in staking service

modules/sdk-coin-vet/src/lib/utils.ts
```diff
@@ -125,20 +130,25 @@ export class Utils implements BaseUtils {
...
-   * @param {string} stakingAmount - The amount to stake in wei
-   * @returns {string} - The encoded transaction data
+   * @param {string} data - The encoded transaction data
+   * @returns {object} - Object containing levelId and autorenew
    */
-  getStakingData(stakingAmount: string): string {
-    const methodName = 'stake';
-    const types = ['uint256'];
-    const params = [new BN(stakingAmount)];
+  decodeStakingData(data: string): { levelId: number; autorenew: boolean } {
```
Vijay-Jagannathan, line 138: qq why are we decoding stakingData alone and not the other ones ?

modules/sdk-coin-vet/src/lib/transactionBuilder/burnNftBuilder.ts
```diff
@@ -95,17 +93,31 @@ export class BurnNftBuilder extends TransactionBuilder {
...
+   * If not provided, uses the network-appropriate default Stargate NFT address.
    *
    * @param {string} address - The NFT contract address
    * @returns {BurnNftBuilder} This transaction builder
    */
-  nftContract(address: string = STARGATE_NFT_ADDRESS): this {
-    this.validateAddress({ address });
-    this.burnNftTransaction.contract = address;
+  nftContract(address?: string): this {
+    const contractAddress = address || this.getDefaultNftAddress();
```
Vijay-Jagannathan: same thing as above, the correct address needs to be called from WP (ig staking service passes this in the intent)

modules/sdk-coin-vet/src/lib/transactionBuilder/burnNftBuilder.ts
```diff
@@ -95,17 +93,31 @@ export class BurnNftBuilder extends TransactionBuilder {
...
+   */
+  private getDefaultNftAddress(): string {
+    const isTestnet = this._coinConfig.network.type === 'testnet';
+
+    if (isTestnet) {
+      return STARGATE_NFT_ADDRESS_TESTNET;
+    }
+
+    return STARGATE_NFT_ADDRESS;
+  }
```
Vijay-Jagannathan: lets be consistent with the method that we call to avoid confusion:

there's another method called 
```
getDefaultStakingAddress(coinConfig: Readonly<CoinConfig>): string {
    const isTestnet = coinConfig.network.type === 'testnet';
    return isTestnet ? STARGATE_NFT_ADDRESS_TESTNET : STARGATE_NFT_ADDRESS;
  }
```

modules/sdk-coin-vet/src/lib/transaction/stakingTransaction.ts
```diff
@@ -6,15 +6,19 @@ import { VetTransactionData } from '../iface';
 import EthereumAbi from 'ethereumjs-abi';
 import utils from '../utils';
 import BigNumber from 'bignumber.js';
+import { addHexPrefix } from 'ethereumjs-util';
 
 export class StakingTransaction extends Transaction {
   private _stakingContractAddress: string;
+  private _levelId: number;
+  private _autorenew: boolean;
```
Vijay-Jagannathan: here's where we can default it to true so that its explicit
```
private _autorenew: boolean = true;
```

setters can exist in the SDK, but let's not call it from WP

modules/sdk-coin-vet/src/lib/transaction/exitDelegation.ts
Vijay-Jagannathan, line 1: in buildClauses: to is set as this._contract, can you confirm if its being set to the correct address ?

ideally some validation to check if the address being set is the correct one in our config

modules/sdk-coin-vet/src/lib/transaction/claimRewards.ts
```diff
@@ -254,41 +226,35 @@ export class ClaimRewardsTransaction extends Transaction {
...
 
     this._claimRewardsData = {
-      validatorAddress,
-      delegatorAddress,
+      tokenId,
       delegationContractAddress:
-        delegationContractAddress !== STARGATE_DELEGATION_ADDRESS ? delegationContractAddress : undefined,
+        delegationContractAddress && !utils.isDelegationContractAddress(delegationContractAddress)
+          ? delegationContractAddress
+          : undefined,
```
Vijay-Jagannathan, line 236: why are we passing delegationContractAddress alone and not the NFT contract address too ?

where is claimRewardsData consumed and does it need the delegationContractAddress in any way?

modules/sdk-coin-vet/src/lib/transaction/burnNftTransaction.ts
Vijay-Jagannathan, line 1: in buildClauses, please confirm if the `to` address which is pointing to `this._contract` is indeed the correct contract address that we want (should be the stargate NFT address)

modules/sdk-coin-vet/src/lib/iface.ts
```diff
@@ -28,6 +28,9 @@ export interface VetTransactionData {
   tokenId?: string; // Added for unstaking and burn NFT transactions
   stakingContractAddress?: string;
   amountToStake?: string;
+  levelId?: number; // Level ID for stakeAndDelegate method
+  nftTokenId?: number; // NFT Token ID for stakeAndDelegate method (same as levelId)
```
Vijay-Jagannathan: these are different right ? for example, this is level id: https://docs.stargate.vechain.org/overview/nft-tiers; whereas the nftTokenId can keep changing for each NFT within the same tier

Dawn NFT 1 and Dawn NFT 2 of an address will have same levelId, but different nftTokenIds, correct ?

if yes, please correct this

Vijay-Jagannathan (commented):
please double confirm comments below + additionally try one more thing:

in a unit test in SDK, try building the same txn data in both coins sandbox (till the unsigned building) and compare the data, it should be same

modules/sdk-coin-vet/src/lib/transaction/claimRewards.ts
Gautam2305: done

Vijay-Jagannathan (approved):
lgtm

joshisakshi (approved):

modules/sdk-coin-vet/src/lib/transaction/burnNftTransaction.ts
Gautam2305: done

PR #7112, opened 2025-09-28
fix(sdk-core): fix TransactionType check

noel-bitgo:
string to enum checks

TICKET: SC-3303


modules/sdk-core/src/bitgo/staking/stakingWallet.ts
```diff
@@ -458,18 +459,6 @@ export class StakingWallet implements IStakingWallet {
       }
     }
 
-    if (buildParams?.aptosCustomTransactionParams) {
```
mukeshsp, line 461: Was this deleted intentionally? This is related to Aptos, but your ticket is related to Avax

noel-bitgo: There is no aptos staking

modules/bitgo/test/v2/fixtures/staking/topethStakingFixtures.ts
```diff
@@ -1,97 +1,3 @@
-export const newStakingRequest = {
```
benkelcher-bitgo, line 1: these mocks were unused previously?

noel-bitgo: correct, just cleanup

modules/sdk-core/src/bitgo/staking/stakingWallet.ts
```diff
@@ -442,7 +443,11 @@ export class StakingWallet implements IStakingWallet {
       );
     }
 
-    if (buildParams?.type && (explainedTransaction as any).type !== buildParams.type) {
+    if (
+      buildParams?.type &&
+      (explainedTransaction as any).type !== undefined &&
```
haritkapadia, line 448: With correct downcasting, I think the compiler would have caught this issue of comparing a string to an enum.

We can downcast from `ITransactionExplanation` with either https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions
or https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates.

This can be a follow-up.

noel-bitgo: yeah i was looking into typing this correctly, but this part about the explained transaction is all over the place. didn't want to make this a big change

balakrishna507 (approved):

haritkapadia (approved):

modules/sdk-core/src/bitgo/staking/stakingWallet.ts
haritkapadia: https://bitgoinc.atlassian.net/browse/SC-3311

PR #7136, opened 2025-09-30
feat(express): migrate createAddress to typed routes

danielzhao122:
TICKET: WP-5418


zahin-mohammad (dismissed):

modules/express/test/unit/clientRoutes/createAddress.ts
```diff
@@ -30,14 +31,22 @@ describe('Create Address', () => {
       bitgo: bitgoStub,
       params: {
         coin: 'tbtc',
-        id: '23423423423423',
+        walletId: '23423423423423',
```
pranavjain97: does it not work with `id` anymore? that would be a breaking change

danielzhao122: Good point. I will revert back to `id`

zahin-mohammad (approved):

PR #7145, opened 2025-10-01
feat: upgrade to lerna 9

zahin-mohammad:
The previous upgrade was reverted due to publishing issues. This PR is purely just the lerna version bump, caching will be handled separately. 

Updates the "files" field in each package.json if it is not already set, required to ensure publishing works as expected. Lerna v7+ only publishes ROOT DIRECTORY files by default, NOT subdirectories, hence why we had issues with publishing the beta packages previously.

Example working publish https://www.npmjs.com/package/@bitgo-beta/statics/v/10.0.1-alpha.406?activeTab=code

https://github.com/lerna/lerna/releases/tag/v9.0.0

Lerna 9 enables us to support OIDC

TICKET: WP-6096

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,3 +8,4 @@ CHANGELOG.md
 flake.lock
 
 modules/babylonlabs-io-btc-staking-ts/
+/.nx/workspace-data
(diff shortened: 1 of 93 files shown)


pranavjain97 (dismissed):

pranavjain97 (approved):

vmccarty (approved):
LGTM

PR #7153, opened 2025-10-02
Babylon cherry picks 2.5.7

benkelcher-bitgo:
- vendor v2.5.7 of `babylonlabs-io/btc-staking-ts`
- apply same cherry-picks as previous
- add one more commit to update protobuf version to `1.7.2` in `utxo-staking`, remove `build/` directory from the vendored module, and update the delegation message generation function in `utxo-staking` to block Babylon's new Expand Stake feature (which we do not support)


OttoAllmendinger (dismissed):

scripts/vendor-github-repo.ts
```diff
@@ -91,7 +91,7 @@ async function fetchArchive(lib: GithubSource, outfile: string): Promise<void> {
   await fs.writeFile(outfile, Buffer.from(await result.arrayBuffer()));
 }
 
-function pipe(cmd: ResultPromise): ResultPromise {
+function pipe(cmd: any): any {
```
Marzooqa: Why are we using `any` here?

benkelcher-bitgo: oh good catch - i had a tricky time getting the vendoring script to work locally and had to do this to make it happy. i think the issue was local, this was the fastest way to get it to work but i hadn't intended to check this change in. will fix

benkelcher-bitgo: took a closer look - actually moved away from `ResultPromise` because the return type changed. Found the correct one, it should be `ExecaChildProcess` now.

OttoAllmendinger (approved):

mukeshsp (approved):

nvjsr (approved):

kaustubhbitgo (approved):

Marzooqa (approved):

PR #7154, opened 2025-10-03
fix(sdk-coin-iota): update transaction and block hash validation

parasgarg-bitgo:
Ticket: WIN-7458

TICKET: WIN-7458

### DIFF
--- a/modules/sdk-coin-iota/src/lib/constants.ts
+++ b/modules/sdk-coin-iota/src/lib/constants.ts
@@ -1,7 +1,7 @@
 export const IOTA_ADDRESS_LENGTH = 64;
-export const IOTA_TRANSACTION_ID_LENGTH = 64;
-export const IOTA_BLOCK_ID_LENGTH = 64;
-export const IOTA_SIGNATURE_LENGTH = 128;
+export const IOTA_TRANSACTION_DIGEST_LENGTH = 32;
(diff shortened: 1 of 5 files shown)


mullapudipruthvik (approved):

modules/sdk-coin-iota/src/lib/constants.ts
```diff
@@ -1,7 +1,7 @@
 export const IOTA_ADDRESS_LENGTH = 64;
-export const IOTA_TRANSACTION_ID_LENGTH = 64;
-export const IOTA_BLOCK_ID_LENGTH = 64;
-export const IOTA_SIGNATURE_LENGTH = 128;
+export const IOTA_TRANSACTION_DIGEST_LENGTH = 32;
+export const IOTA_BLOCK_DIGEST_LENGTH = 32;
+export const IOTA_SIGNATURE_LENGTH = 64;
```
hitansh-madan, line 4: Maybe it's better to call it something like `IOTA_TRANSACTION_DIGEST_BYTES_LENGTH` so that it's more explicit & less confusing

hitansh-madan (approved):

modules/sdk-coin-iota/src/lib/constants.ts
parasgarg-bitgo: Will address this along with changes for tx building

PR #7163, opened 2025-10-03
feat(sdk-core): add psbt-lite as a transaction format option

davidkaplanbitgo:
Add 'psbt-lite' as a valid option for txFormat in PrebuildTransactionOptions interface.

Issue: BTC-2272

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -167,7 +167,7 @@ export interface PrebuildTransactionOptions {
    * This comment applies to UTXO coins. It's asking which transaction format to use:
    * the legacy format defined by bitcoinjs-lib, or the 'psbt' format, which follows the BIP-174.
    */
-  txFormat?: 'legacy' | 'psbt';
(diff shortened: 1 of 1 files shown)


parvbitgo067 (approved):

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -167,7 +167,7 @@ export interface PrebuildTransactionOptions {
    * This comment applies to UTXO coins. It's asking which transaction format to use:
    * the legacy format defined by bitcoinjs-lib, or the 'psbt' format, which follows the BIP-174.
```
margueriteblair, line 168: so nitty - but can you add `psbt-lite` to the jsdoc description? and describe what the difference between lite and full would be? can be done in a follow up

margueriteblair (approved):

modules/sdk-core/src/bitgo/wallet/iWallet.ts
davidkaplanbitgo: Will do in follow up

PR #7164, opened 2025-10-04
feat(sdk-core): add verification options for signing txHex

mohammadalfaiyazbitgo:
TICKET: WP-6188

Tested with BTC & ETH Hot Wallets.

### DIFF
--- a/modules/bitgo/test/v2/unit/signTransactionVerification.ts
+++ b/modules/bitgo/test/v2/unit/signTransactionVerification.ts
@@ -1,0 +1,154 @@
+import nock = require('nock');
+import * as sinon from 'sinon';
+import * as assert from 'assert';
+import 'should';
+
+import { BitGoAPI } from '@bitgo/sdk-api';
+import { TestBitGo } from '@bitgo/sdk-test';
+import { Tbtc } from '@bitgo/sdk-coin-btc';
+import { common, BaseCoin, BitGoBase, Wallet, WalletSignTransactionOptions } from '@bitgo/sdk-core';
+
+describe('Wallet signTransaction with verifyTxParams', function () {
+  let wallet: Wallet;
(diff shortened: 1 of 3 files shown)


modules/bitgo/test/v2/unit/signTransactionVerification.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    const txPrebuild = {
+      txHex: realTxHex,
+      walletId: '6840948b17e91662b782d55bbf988c4e',
+    };
+
+    // Verification parameters with wrong expected recipient
+    const verifyTxParams = {
+      txParams: {
+        recipients: [
```
margueriteblair, line 63: And it will trip an error for anything in this `verifyTxParams` that is does not match the txHex? So in this case it's the recipient, but could also be amount too, right?

margueriteblair (commented):
This seems fine to me, do we have plans for E2E testing in a lower env?

margueriteblair (dismissed):

Marzooqa (dismissed):

OttoAllmendinger (changes requested):
please don't add magic binary blobs in tests

this is testing very utxo specific code and belongs in `abstract-utxo`

however this particular func already is extensively tested. Here you can just setup a sinon spy and ensure it gets called.

modules/bitgo/test/v2/unit/signTransactionVerification.ts
mohammadalfaiyazbitgo: yes, if `recipients` is defined it will attempt to match that all recipients are in the resulting txHex with the appropriate amount, without additional recipients.

OttoAllmendinger (approved):

FloBitGo (approved):

Marzooqa (approved):

PR #7168, opened 2025-10-06
chore: add rebase token feature for vtho testnet

at31416:
Adding rebase token feature to use the same implementation for `vtho` as for the rebase tokens.
VTHO balance automatically increases for any address holding VET on vechain, therefore we can use the same implementation as rebase tokens which handles automatic updates of token balances by syncing on chain balances periodically.

Ticket: COIN-5905


ravibitgo (approved):
lgtm

baltiyal (approved):

PR #7175, opened 2025-10-06
feat(sdk-coin-canton): added wallet initialization builder class

ravibitgo:
Ticket: [COIN-5917](https://bitgoinc.atlassian.net/browse/COIN-5917)

[COIN-5917]: https://bitgoinc.atlassian.net/browse/COIN-5917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/sdk-coin-canton/src/lib/walletInitBuilder.ts
```diff
@@ -0,0 +1,261 @@
...
+   * @throws Error if the hint is empty or more than 5 characters long.
+   */
+  partyHint(hint: string): this {
+    const trimmedHint = hint.trim();
+    if (!trimmedHint) {
+      throw new Error('partyHint cannot be empty');
+    }
+    if (trimmedHint.length > 5) {
+      throw new Error('partyHint must be less than 6 characters long');
+    }
```
balakrishna507, line 141: our party hints on devnet were longer weren't they? the test case also has this party, and i'm guessing this is the hint?
```
test-ravi-party-01
```

modules/sdk-coin-canton/src/lib/walletInitialization/walletInitTransaction.ts
```diff
@@ -0,0 +1,67 @@
...
+  get signablePayload(): Buffer {
+    if (!this._preparedParty) {
+      throw new InvalidTransactionError('Empty transaction data');
+    }
+    return Buffer.from(this._preparedParty.multiHash);
+  }
+
+  fromRawTransaction(rawTx: string): void {
+    try {
+      const decoded: PreparedParty = JSON.parse(Buffer.from(rawTx, 'base64').toString('utf8'));
```
balakrishna507, line 59: is the input a `PreparedTransactionRawData` like the one in the test?

modules/sdk-coin-canton/src/lib/walletInitBuilder.ts
ravibitgo: There is no such limit on this but this won't be a user input, we will be giving the party hint, in canton SDK they take the first 5 characters of the `fingerprint` if hint is not provided so thinking of keeping the same logic

modules/sdk-coin-canton/src/lib/walletInitialization/walletInitTransaction.ts
ravibitgo: Yes, since there is no deserializer, I'm thinking of keeping this as the base64 encoded preparedTransaction

balakrishna507 (approved):

modules/sdk-coin-canton/src/lib/walletInitBuilder.ts
balakrishna507: if the 5 character limit only comes up if we don't specify anything, we should never face this situation if we just use our wallet id as the hint right?

balakrishna507 (approved):

baltiyal (approved):
lgtm

PR #7177, opened 2025-10-06
Fix Docker build disk space exhaustion in CI

andrew-scott-fischer:
Fixes disk space exhaustion during Docker builds in GitHub Actions CI by using BuildKit cache mounts for the yarn cache directory.

## Changes
- Modified Dockerfile to use `--mount=type=cache` for yarn cache
- Yarn cache is now excluded from Docker image layers
- Cache persists between builds on the same runner

## Benefits
- Fixes "no space left on device" errors in CI
- Reduces final image size (yarn cache not included)
- Maintains or improves build performance (persistent cache)

Ticket: VL-3537


mullapudipruthvik (commented):
flush: re-request when ready

PR #7180, opened 2025-10-07
feat: change the netUID data type from string to number for Tao Staking

Doddanna17:
feat: change the netUID data type from string to number for Tao Staking
TICKET: SC-3399


Vijay-Jagannathan (approved):
non-blocking, but whats the issue with having it as string, consumer can take care ?

at31416 (approved):
lgtm

PR #7212, opened 2025-10-10
feat: disallow EVM_COMPATIBLE_WP feature for OFC coin

lcovar:
Ticket: BTC-0

TICKET: BTC-0

### DIFF
--- a/modules/statics/src/ofc.ts
+++ b/modules/statics/src/ofc.ts
@@ -35,7 +35,9 @@ const DISALLOWED_FEATURES = [
   CoinFeature.CHILD_PAYS_FOR_PARENT,
   CoinFeature.PAYGO,
   CoinFeature.SUPPORTS_TOKENS,
+  // OFC inherits features from the base asset but the features are not always applicable to OFC
   CoinFeature.SHARED_EVM_SIGNING,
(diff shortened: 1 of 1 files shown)


modules/statics/src/ofc.ts
```diff
@@ -36,6 +36,7 @@ const DISALLOWED_FEATURES = [
   CoinFeature.PAYGO,
   CoinFeature.SUPPORTS_TOKENS,
   CoinFeature.SHARED_EVM_SIGNING,
+  CoinFeature.EVM_COMPATIBLE_WP,
```
mukeshsp, line 40: Can you add the description ? Why are we adding as disallowed.

zahin-mohammad: Because OFC is not evm compatible...
OFC by default inherits features from their base asset.

But this is not always applicable.

zahin-mohammad (dismissed):

modules/statics/src/ofc.ts
lcovar: added a comment

mukeshsp (approved):

at31416 (approved):

zahin-mohammad (approved):

PR #7225, opened 2025-10-14
feat(express): add supertest for createAddress

kaustubhbitgo:
Ticket: WP-6312

### DIFF
--- a/modules/express/test/unit/typedRoutes/createAddress.ts
+++ b/modules/express/test/unit/typedRoutes/createAddress.ts
@@ -6,6 +6,12 @@ import {
   PostCreateAddress,
 } from '../../../src/typedRoutes/api/v2/createAddress';
 import { assertDecode } from './common';
+import 'should';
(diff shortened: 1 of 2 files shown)


modules/express/test/unit/typedRoutes/createAddress.ts
```diff
@@ -26,6 +33,144 @@ function getCodecParamNames(paramsCodec: Record<string, any>): string[] {
...
+
+    before(function () {
+      const args: any = {
+        debug: false,
+        env: 'test',
+        logfile: '/dev/null',
+      };
+
+      const app = expressApp(args);
+      agent = request.agent(app);
```
kisslove-dewangan: nit: can we put agent init logic in some base util file? if it's a recurring logic

kaustubhbitgo: Yup will do

PR #7234, opened 2025-10-14
chore(statics): remove rebase feature from vtho token

at31416:
his feature was introduced in https://bitgoinc.atlassian.net/browse/COIN-5905
 
this is was only used in indexer to put this token metadata in RebaseTokensDataFactory so that while parsing the entries we can identify an address that it should be added to rebaseAddress collection.

but with the new implementation where we are filtering out address when the entry is having coin to be vet/ tvet we don't need token to be explicitly marked as a rebase token.

Ticket: COIN-6055


baltiyal (approved):
lgtm

ravibitgo (approved):

PR #7239, opened 2025-10-14
feat!(sdk-coin-celo): remove unused files and functions

ppongbitgo:


modules/sdk-coin-celo/src/lib/utils.ts
```diff
@@ -1,56 +1,8 @@
...
 import EthereumCommon from '@ethereumjs/common';
-import { recoverTransaction } from '@celo/wallet-base';
-import * as ethers from 'ethers';
-import BigNumber from 'bignumber.js';
-import { ETHTransactionType, LegacyTxData } from '@bitgo/sdk-coin-eth';
-import { InvalidTransactionError, ParseTransactionError } from '@bitgo/sdk-core';
+import { InvalidTransactionError } from '@bitgo/sdk-core';
 import { mainnetCommon, testnetCommon } from './resources';
 
-/**
```
zahin-mohammad, line 10: These removals are technically breaking changes. They are exported functions for this package.

Lets update the commit so that this is reflected in a major version bump. Might need to verify internal usage as well.

ppongbitgo: done - adding BREAKING CHANGE: in commit

ppongbitgo: this seemed to only be used in a test previously https://github.com/BitGo/BitGoJS/commit/f23f5a7fd7c7bf1b03e66090f70bcfe697dbd5a0#diff-5025e3f5924558532cb227a8195d523527fc41041ca22b3525505cf24948fe06R70

zahin-mohammad (approved):

theMobiusStrip (approved):

PR #7256, opened 2025-10-15
feat(sdk-core): add automatic signature cleanup for Express TSS signing

dpkjnr:
Fixes Express API gap where re-signing partially signed Full TxRequests would fail. The /signtxtss route now automatically detects and cleans up partial signature shares before attempting to sign.

Changes:
- Add ensureCleanSigSharesAndSignTransaction() to Wallet class
- Update Express handleV2SignTSSWalletTx() to use new method
- Add comprehensive unit tests for signature cleanup logic
- Check signature shares presence instead of transaction state

Benefits:
- Fixes re-signing failures for partially signed transactions
- Backward compatible with existing signing flows
- No API changes required
- Keeps tssUtils properly encapsulated

Ticket: COIN-5989

### DIFF
--- a/modules/express/test/unit/typedRoutes/walletTxSignTSS.ts
+++ b/modules/express/test/unit/typedRoutes/walletTxSignTSS.ts
@@ -46,9 +46,9 @@ describe('WalletTxSignTSS codec tests', function () {
         apiVersion: 'lite' as const,
       };
 
-      // Create mock wallet with signTransaction method
+      // Create mock wallet with ensureCleanSigSharesAndSignTransaction method
       const mockWallet = {
-        signTransaction: sinon.stub().resolves(mockFullySignedResponse),
+        ensureCleanSigSharesAndSignTransaction: sinon.stub().resolves(mockFullySignedResponse),
       };
 
       // Stub the wallets().get() chain
@@ -85,7 +85,7 @@ describe('WalletTxSignTSS codec tests', function () {
       assert.strictEqual(coinStub.calledOnceWith(coin), true);
       assert.strictEqual(mockCoin.wallets.calledOnce, true);
       assert.strictEqual(walletsGetStub.calledOnceWith({ id: walletId }), true);
-      assert.strictEqual(mockWallet.signTransaction.calledOnce, true);
+      assert.strictEqual(mockWallet.ensureCleanSigSharesAndSignTransaction.calledOnce, true);
     });
 
     it('should successfully sign a half-signed TSS wallet transaction', async function () {
@@ -106,9 +106,9 @@ describe('WalletTxSignTSS codec tests', function () {
         },
       };
 
-      // Create mock wallet with signTransaction method
+      // Create mock wallet with ensureCleanSigSharesAndSignTransaction method
       const mockWallet = {
-        signTransaction: sinon.stub().resolves(mockHalfSignedResponse),
+        ensureCleanSigSharesAndSignTransaction: sinon.stub().resolves(mockHalfSignedResponse),
       };
 
       // Stub the wallets().get() chain
@@ -146,7 +146,7 @@ describe('WalletTxSignTSS codec tests', function () {
       assert.strictEqual(coinStub.calledOnceWith(coin), true);
       assert.strictEqual(mockCoin.wallets.calledOnce, true);
       assert.strictEqual(walletsGetStub.calledOnceWith({ id: walletId }), true);
-      assert.strictEqual(mockWallet.signTransaction.calledOnce, true);
+      assert.strictEqual(mockWallet.ensureCleanSigSharesAndSignTransaction.calledOnce, true);
     });
 
     it('should successfully sign a half-signed UTXO transaction', async function () {
@@ -165,9 +165,9 @@ describe('WalletTxSignTSS codec tests', function () {
           '0100000001c7dad3d9607a23c45a6c1c5ad7bce02acff71a0f21eb4a72a59d0c0e19402d0f00000000484730440220abc123def456...',
       };
 
-      // Create mock wallet with signTransaction method
+      // Create mock wallet with ensureCleanSigSharesAndSignTransaction method
       const mockWallet = {
-        signTransaction: sinon.stub().resolves(mockHalfSignedUtxoResponse),
+        ensureCleanSigSharesAndSignTransaction: sinon.stub().resolves(mockHalfSignedUtxoResponse),
       };
 
       // Stub the wallets().get() chain
@@ -204,7 +204,7 @@ describe('WalletTxSignTSS codec tests', function () {
       assert.strictEqual(coinStub.calledOnceWith(coin), true);
       assert.strictEqual(mockCoin.wallets.calledOnce, true);
       assert.strictEqual(walletsGetStub.calledOnceWith({ id: walletId }), true);
-      assert.strictEqual(mockWallet.signTransaction.calledOnce, true);
+      assert.strictEqual(mockWallet.ensureCleanSigSharesAndSignTransaction.calledOnce, true);
     });
 
     it('should successfully return a TSS transaction request ID', async function () {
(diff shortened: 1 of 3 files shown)


nvjsr (dismissed):

nvjsr:
@claude review

kamleshmugdiya (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3566,6 +3566,45 @@ export class Wallet implements IWallet {
...
+   *
+   * @param params - signing options
+   * @returns signed transaction
+   */
+  public async ensureCleanSigSharesAndSignTransaction(
+    params: WalletSignTransactionOptions = {}
+  ): Promise<SignedTransaction | TxRequest> {
+    const txRequestId = params.txRequestId || params.txPrebuild?.txRequestId;
+
+    if (txRequestId && this.tssUtils && this.multisigType() === 'tss') {
```
sachinBitgo, line 3684: can we also separate the cleaning part in different method. It think it will make the method ensureCleanSigSharesAndSignTransaction more readable.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3566,6 +3566,48 @@ export class Wallet implements IWallet {
...
+        (txRequest.transactions ?? []).length > 0 ||
+        (txRequest.messages ?? []).length > 0;
+
+      if (isFull) {
+        // Check if there are any partial signature shares present
+        // Signature shares array is populated during signing and cleared when complete
+        let isPartiallySigned = false;
+
+        if (txRequest.transactions && txRequest.transactions.length > 0) {
+          isPartiallySigned = txRequest.transactions.some((tx) => (tx.signatureShares ?? []).length > 0);
```
lokesh-bitgo: Are we not considering any transaction state for this?

modules/express/test/unit/typedRoutes/walletTxSignTSS.ts
```diff
@@ -46,9 +46,9 @@ describe('WalletTxSignTSS codec tests', function () {
         apiVersion: 'lite' as const,
       };
 
-      // Create mock wallet with signTransaction method
+      // Create mock wallet with ensureCleanSigSharesAndSignTransaction method
       const mockWallet = {
-        signTransaction: sinon.stub().resolves(mockFullySignedResponse),
+        ensureCleanSigSharesAndSignTransaction: sinon.stub().resolves(mockFullySignedResponse),
```
lokesh-bitgo, line 51: The current tests target signTransaction. Please add tests that perform cleanup of any existing transaction and then sign the transaction, to validate the business logic in ensureCleanSigSharesAndSignTransaction.

modules/sdk-core/src/bitgo/wallet/wallet.ts
dpkjnr: I saw something similar in wallet-platform also, would there be any case where sigShares are there but state is postSign or something where we should not delete sig shares?

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -3765,6 +3765,123 @@ describe('V2 Wallet:', function () {
       });
     });
 
+    describe('ensureCleanSigSharesAndSignTransaction', function () {
```
lokesh-bitgo: I think these tests should go in `test/unit/typedRoutes/walletTxSignTss.ts`, since we are calling `ensureCleanSigSharesAndSignTransaction` from the handler `handleV2SignTSSWalletTx`.

dpkjnr: Added those before your PR was merged, let me move these to walletTxSignTss

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3566,6 +3566,48 @@ export class Wallet implements IWallet {
...
+        } else if (txRequest.messages && txRequest.messages.length > 0) {
+          isPartiallySigned = txRequest.messages.some((msg) => (msg.signatureShares ?? []).length > 0);
+        }
+
+        if (isPartiallySigned) {
+          await this.tssUtils.deleteSignatureShares(txRequestId);
+        }
```
zahin-mohammad: Can't we always just rebuild/delete sig shares before signing? Or are we avoiding this for latency reasons?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3566,6 +3566,48 @@ export class Wallet implements IWallet {
...
+    const txRequestId = params.txRequestId || params.txPrebuild?.txRequestId;
+
+    if (txRequestId && this.tssUtils && this.multisigType() === 'tss') {
+      const txRequest = await this.tssUtils.getTxRequest(txRequestId);
+
+      // Check if txRequest is Full (not Lite)
+      const isFull =
+        txRequest.apiVersion === 'full' ||
+        (txRequest.transactions ?? []).length > 0 ||
+        (txRequest.messages ?? []).length > 0;
```
zahin-mohammad:
```suggestion
      const isFull =
        txRequest.apiVersion === 'full' &&
        ((txRequest.transactions ?? []).length > 0 ||
        (txRequest.messages ?? []).length > 0);
```
or 
```suggestion
      const isFull = txRequest.apiVersion === 'full';
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
zahin-mohammad: The state could be `readyToCombineShares` which means it has been fully signed, but has not been broadcasted yet (the end state of v2.wallet.txrequest.sign is `readyToCombineShares`, and this is validated in the send api)

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3566,6 +3566,45 @@ export class Wallet implements IWallet {
...
+
+      if (isFull) {
+        // Check if there are any partial signature shares present
+        // Signature shares array is populated during signing and cleared when complete
+        let isPartiallySigned = false;
+
+        if (txRequest.transactions && txRequest.transactions.length > 0) {
+          isPartiallySigned = txRequest.transactions.some((tx) => (tx.signatureShares ?? []).length > 0);
+        } else if (txRequest.messages && txRequest.messages.length > 0) {
+          isPartiallySigned = txRequest.messages.some((msg) => (msg.signatureShares ?? []).length > 0);
```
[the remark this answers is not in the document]
zahin-mohammad: +1, 

can do something like 

```
const signabeResources: { signatureShares: [] } = txRequest.transactions?.length ?  txRequest.transactions : txRequest.messages ?? [];
```

modules/sdk-core/src/bitgo/wallet/wallet.ts
dpkjnr: that looks cleaner but we would make additional delete sigshares API call even in case of fresh transaction. If it looks fine to your team then I can make it default for txRequest full

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3620,6 +3566,36 @@ export class Wallet implements IWallet {
...
+   * @param params - signing options
+   * @returns signed transaction
+   */
+  public async ensureCleanSigSharesAndSignTransaction(
+    params: WalletSignTransactionOptions = {}
+  ): Promise<SignedTransaction | TxRequest> {
+    const txRequestId = params.txRequestId || params.txPrebuild?.txRequestId;
+
+    if (txRequestId && this.tssUtils && this.multisigType() === 'tss') {
+      const txRequest = await this.tssUtils.getTxRequest(txRequestId);
```
kaustubhbitgo, line 3685: Would it make sense to move this whole logic to backend via a `params.rebuild` flag in signTransaction call? This way we can abstract the logic for multiple wallet types (future requirement) and control the implementation details.

dpkjnr: I don't think it will be possible, in case of TSS we do first step in browser before calling sign API in the platform. Rebuild has to happen before signing starts.

kaustubhbitgo (approved):

PR #7263, opened 2025-10-16
test(express): added supertest for fanoutUnspents

lokesh-bitgo:
Ticket: WP-6326

### DIFF
--- a/modules/express/test/unit/typedRoutes/fanoutUnspents.ts
+++ b/modules/express/test/unit/typedRoutes/fanoutUnspents.ts
@@ -7,8 +7,614 @@ import {
   PutFanoutUnspents,
 } from '../../../src/typedRoutes/api/v1/fanoutUnspents';
 import { assertDecode } from './common';
+import 'should';
+import 'should-http';
+import 'should-sinon';
+import * as sinon from 'sinon';
+import { BitGo } from 'bitgo';
+import { setupAgent } from '../../lib/testutil';
 
 describe('FanoutUnspents codec tests', function () {
+  describe('fanoutUnspents', function () {
+    const agent = setupAgent();
+    const walletId = '68c02f96aa757d9212bd1a536f123456';
+
+    const mockFanoutResponse = {
+      status: 'accepted',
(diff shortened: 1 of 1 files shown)


modules/express/test/unit/typedRoutes/fanoutUnspents.ts
```diff
@@ -7,8 +7,602 @@ import {
...
+        const mockWallets = { get: walletsGetStub };
+        sinon.stub(BitGo.prototype, 'wallets').returns(mockWallets as any);
+
+        const result = await agent
+          .put(`/api/v1/wallet/${walletId}/fanoutunspents`)
+          .set('Authorization', 'Bearer test_access_token_12345')
+          .set('Content-Type', 'application/json')
+          .send(requestBody);
+
+        assert.ok(result.status === 200 || result.status >= 400);
```
danielzhao122: We should probably only assert if it fails or succeeds. Doing this renders the test pointless

modules/express/test/unit/typedRoutes/fanoutUnspents.ts
```diff
@@ -7,8 +7,602 @@ import {
...
```
danielzhao122: likewise here

modules/express/test/unit/typedRoutes/fanoutUnspents.ts
```diff
@@ -7,8 +7,602 @@ import {
...
```
danielzhao122: same thing here

modules/express/test/unit/typedRoutes/fanoutUnspents.ts
```diff
@@ -7,8 +7,602 @@ import {
...
```
danielzhao122: same thing here

Marzooqa (dismissed):

modules/express/test/unit/typedRoutes/fanoutUnspents.ts
lokesh-bitgo: Addressed

danielzhao122 (approved):

PR #7265, opened 2025-10-16
feat(sdk-coin-sol): big endian support for verifyTransaction

pranavjain97:
TICKET: WP-6284

Fix s390x endianness issue in Solana transaction verification

## Problem

On s390x (big-endian) architecture, `@solana/web3.js` incorrectly reads Solana's little-endian u64 amounts as big-endian during transaction parsing in `Transaction.fromRawTransaction()`. The library uses Node.js Buffer methods that are architecture-dependent, causing byte order misinterpretation.

This results in corrupted amounts when `explainTransaction()` is called:
- Small amounts (e.g., 10,000 lamports) → appear as huge values (~1.16×10^18 lamports)
- Large amounts (e.g., ~5×10^17 lamports) → appear as tiny values (7 lamports)

Both cases cause `verifyTransaction()` to fail when comparing parsed transaction amounts against user-provided recipients.

## Root Cause

- Solana stores amounts as u64 in **little-endian** format
- `@solana/web3.js` parser uses Buffer reads that respect architecture endianness
- On s390x, Buffer reads are **big-endian** by default
- **Result**: Parser consistently reads all amounts with wrong byte order

## Solution

Added `getAmountBasedOnEndianness()` helper function that:
1. Detects s390x architecture via `process.arch`
2. Byte-swaps all amounts to undo the parser's incorrect endianness interpretation
3. Returns amounts unchanged on other architectures

Applied in `verifyTransaction()` when:
- Comparing transaction outputs against user-provided recipients
- Summing output amounts for total validation

### DIFF
--- a/modules/sdk-coin-sol/test/unit/sol.ts
+++ b/modules/sdk-coin-sol/test/unit/sol.ts
@@ -24,7 +24,7 @@ import {
 } from '@bitgo/sdk-core';
 import { TestBitGo, TestBitGoAPI } from '@bitgo/sdk-test';
 import { coins } from '@bitgo/statics';
-import { KeyPair, Sol, SolVerifyTransactionOptions, Tsol } from '../../src';
+import { getAmountBasedOnEndianness, KeyPair, Sol, SolVerifyTransactionOptions, Tsol } from '../../src';
 import { Transaction } from '../../src/lib';
 import { AtaInit, InstructionParams, TokenTransfer } from '../../src/lib/iface';
 import { getAssociatedTokenAccountAddress } from '../../src/lib/utils';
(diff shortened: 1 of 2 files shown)


margueriteblair (commented):
No tests needed?

pranavjain97:
> No tests needed?

yeah cant really have tests for this because the behavior woudnt be like s390x

margueriteblair (commented):
not even some unit test where you take data that matches `s390x` ? i get it'd be hard to test E2E, but even tests for this function?

margueriteblair (approved):

Vijay-Jagannathan (approved):

modules/sdk-coin-sol/test/unit/sol.ts
```diff
@@ -601,6 +601,93 @@ describe('SOL:', function () {
...
+      Object.defineProperty(process, 'arch', {
+        value: 's390x',
+        writable: true,
+        configurable: true,
+      });
+
+      // Standard amount 300,000 (0x493E0) swaps to large value
+      const result = getAmountBasedOnEndianness('300000');
+      // Verify it's different (swapped)
+      result.should.not.equal('300000');
```
Vijay-Jagannathan, line 674: should we have another assert which also verifies what this value is (since its deterministic) apart from asserting that it is not equal to the original value

pranavjain97: this test is actually not needed as we verify against the exact values in the tests above, but yea we can add it

PR #7273, opened 2025-10-17
fix: change the netUID data type from number to string for Tao Staking

raksha-r7:
Ticket: SC-3399
Resusing UI components for staking drawer which expects the value to be string.


Doddanna17 (approved):
LGTM

abhijit0943 (approved):

PR #7280, opened 2025-10-20
test(express): added supertest for wallet signtransaction

lokesh-bitgo:
Ticket: WP-6337

### DIFF
--- a/modules/express/test/unit/typedRoutes/signTransaction.ts
+++ b/modules/express/test/unit/typedRoutes/signTransaction.ts
@@ -6,6 +6,12 @@ import {
   PostSignTransaction,
 } from '../../../src/typedRoutes/api/v1/signTransaction';
 import { assertDecode } from './common';
+import 'should';
+import 'should-http';
+import 'should-sinon';
+import * as sinon from 'sinon';
+import { BitGo } from 'bitgo';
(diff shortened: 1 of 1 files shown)


modules/express/test/unit/typedRoutes/signTransaction.ts
```diff
@@ -335,4 +341,883 @@ describe('SignTransaction codec tests', function () {
...
+      sinon.stub(BitGo.prototype, 'wallets').returns(mockWallets as any);
+
+      const result = await agent
+        .post(`/api/v1/wallet/${walletId}/signtransaction`)
+        .set('Authorization', 'Bearer test_access_token_12345')
+        .set('Content-Type', 'application/json')
+        .send(requestBody);
+
+      // Should handle additional properties in unspents
+      assert.ok(result.status === 200 || result.status >= 400);
```
Marzooqa, line 1084: can we have one status code?

lokesh-bitgo: Yes, we’ll have only one status code here. I still need to address a few additional points, which I’ll cover in the upcoming follow-up PRs.

Marzooqa (approved):

PR #7286, opened 2025-10-20
feat(sdk-coin-canton): added verify transaction method handling for wallet init

ravibitgo:
Ticket: [COIN-6128](https://bitgoinc.atlassian.net/browse/COIN-6128)

[COIN-6128]: https://bitgoinc.atlassian.net/browse/COIN-6128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-coin-canton/src/canton.ts
+++ b/modules/sdk-coin-canton/src/canton.ts
@@ -10,11 +10,13 @@ import {
   ParseTransactionOptions,
   SignedTransaction,
   SignTransactionOptions,
-  VerifyAddressOptions,
+  TransactionType,
+  TssVerifyAddressOptions,
   VerifyTransactionOptions,
 } from '@bitgo/sdk-core';
 import { auditEddsaPrivateKey } from '@bitgo/sdk-lib-mpc';
-import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import { BaseCoin as StaticsBaseCoin, coins } from '@bitgo/statics';
+import { TransactionBuilderFactory } from './lib';
 import { KeyPair as CantonKeyPair } from './lib/keyPair';
 import utils from './lib/utils';
 
@@ -70,12 +72,29 @@ export class Canton extends BaseCoin {
   }
 
   /** @inheritDoc */
-  verifyTransaction(params: VerifyTransactionOptions): Promise<boolean> {
-    throw new Error('Method not implemented.');
+  async verifyTransaction(params: VerifyTransactionOptions): Promise<boolean> {
+    const coinConfig = coins.get(this.getChain());
+    // extract `txParams` when verifying other transaction types
+    const { txPrebuild: txPrebuild } = params;
+    const rawTx = txPrebuild.txHex;
+    if (!rawTx) {
+      throw new Error('missing required tx prebuild property txHex');
+    }
+    const txBuilder = new TransactionBuilderFactory(coinConfig).from(rawTx);
+    const transaction = txBuilder.transaction;
+    switch (transaction.type) {
+      case TransactionType.WalletInitialization: {
+        // there is no input for this type of transaction, so always return true
+        return true;
+      }
+      default: {
+        throw new Error(`unknown transaction type, ${transaction.type}`);
+      }
+    }
   }
 
   /** @inheritDoc */
-  isWalletAddress(params: VerifyAddressOptions): Promise<boolean> {
+  isWalletAddress(params: TssVerifyAddressOptions): Promise<boolean> {
     throw new Error('Method not implemented.');
   }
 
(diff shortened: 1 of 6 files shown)


mohd-kashif (dismissed):

modules/sdk-coin-canton/src/lib/utils.ts
```diff
@@ -12,7 +12,10 @@ import { RecordField } from './resourcesInterface';
 export class Utils implements BaseUtils {
   /** @inheritdoc */
   isValidAddress(address: string): boolean {
```
at31416, line 14: would it help to add basic tests around this?

modules/sdk-coin-canton/src/canton.ts
```diff
@@ -70,12 +72,29 @@ export class Canton extends BaseCoin {
...
+    // extract `txParams` when verifying other transaction types
+    const { txPrebuild: txPrebuild } = params;
+    const rawTx = txPrebuild.txHex;
+    if (!rawTx) {
+      throw new Error('missing required tx prebuild property txHex');
+    }
+    const txBuilder = new TransactionBuilderFactory(coinConfig).from(rawTx);
+    const transaction = txBuilder.transaction;
+    switch (transaction.type) {
+      case TransactionType.WalletInitialization: {
```
at31416, line 86: are there no tests for this because we have only written the verifyTransaction only for `WalletInitialization` type of transactions and will add in future?

modules/sdk-coin-canton/src/lib/utils.ts
ravibitgo: will add

modules/sdk-coin-canton/src/canton.ts
ravibitgo: yes, right now for wallet init there is no check here since no input from user, will add tests once we implement this for other builders

at31416 (approved):
lgtm!

modules/sdk-coin-canton/src/lib/utils.ts
```diff
@@ -12,7 +12,10 @@ import { RecordField } from './resourcesInterface';
...
-    throw new Error('Method not implemented.');
+    if (!address || address.trim() === '') return false;
+    const [partyHint, fingerprint] = address.trim().split('::');
+    if (!partyHint || !fingerprint) return false;
+    return partyHint.length === 5 && this.isValidCantonHex(fingerprint);
```
balakrishna507, line 18: does it make sense to just leave out the first part of the check? afaiu every canton party/address just needs a hint and a fingerprint, and the hint can be anything right? we use the first five characters of the fingerprint by default and we might even enforce this somewhere on our platforms (bgms, etc.). but can we leave this out of the sdk and allow party hints of otherwise sensible sizes? we don't have security concerns right? just in case we need to allow longer hints we won't need an sdk change. our clients might also want something more than five characters on occasion?

```suggestion
    return partyHint.length <= MAX_PARTY_HINT_LENGTH && this.isValidCantonHex(fingerprint);
```

modules/sdk-coin-canton/src/canton.ts
```diff
@@ -70,12 +72,29 @@ export class Canton extends BaseCoin {
...
+        // there is no input for this type of transaction, so always return true
```
balakrishna507, line 87: if the fingerprint in the initialization (party allocation) transaction doesn't match the public key that signed it, it isn't valid and there's no harm for the client, right? so not even the fingerprint is an input?

modules/sdk-coin-canton/src/canton.ts
```diff
@@ -70,12 +72,29 @@ export class Canton extends BaseCoin {
...
-    throw new Error('Method not implemented.');
+  async verifyTransaction(params: VerifyTransactionOptions): Promise<boolean> {
+    const coinConfig = coins.get(this.getChain());
+    // extract `txParams` when verifying other transaction types
+    const { txPrebuild: txPrebuild } = params;
+    const rawTx = txPrebuild.txHex;
+    if (!rawTx) {
+      throw new Error('missing required tx prebuild property txHex');
+    }
+    const txBuilder = new TransactionBuilderFactory(coinConfig).from(rawTx);
```
balakrishna507, line 83: does this end up calling our new address validation procedure?

balakrishna507 (approved):

modules/sdk-coin-canton/src/lib/utils.ts
ravibitgo: In the wallet init builder we have the check for party hint when setting it, we would need sdk changes if we plan to change the length in future or give an option for users to input.

modules/sdk-coin-canton/src/canton.ts
ravibitgo: nothing is the input from the user, hence returning `true`.
the check you mentioned we will be doing when we get the signature, key pair has a `isValidSignature` method which checks if the generated signature is generated by the expected public key or not

modules/sdk-coin-canton/src/canton.ts
ravibitgo: you mean `isValidAddress`? - if so the answer is no, this is the factory which gives you the required builder from the raw transaction data

PR #7287, opened 2025-10-20
feat(sdk-coin-sol): add recentBlockhash to custom versioned solana tx flow

derekchen745:
Ticket: SC-3567


benkelcher-bitgo:
can you confirm that there is a test covering the scenario where the `recentBlockhash` is not explicitly provided in the data (ie., a backwards compatible case)?

derekchen745:
> can you confirm that there is a test covering the scenario where the `recentBlockhash` is not explicitly provided in the data (ie., a backwards compatible case)?

yup can confirm

benkelcher-bitgo (approved):

sbplat (approved):

PR #7290, opened 2025-10-21
feat: added createAccount intent for canton wallet init

ravibitgo:
Ticket: [COIN-6101](https://bitgoinc.atlassian.net/browse/COIN-6101)

[COIN-6101]: https://bitgoinc.atlassian.net/browse/COIN-6101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -1247,6 +1247,87 @@ describe('V2 Wallet:', function () {
     });
   });
 
+  describe('Canton tests: ', () => {
+    let cantonWallet: Wallet;
+    const cantonBitgo = TestBitGo.decorate(BitGo, { env: 'mock' });
+    cantonBitgo.initializeTestVars();
+    const walletData = {
+      id: '598f606cd8fc24710d2ebadb1d9459bb',
+      coinSpecific: {
+        baseAddress: '12205::12205b4e3537a95126d90604592344d8ad3c3ddccda4f79901954280ee19c576714d',
+        pendingChainInitialization: true,
+        lastChainIndex: { 0: 0 },
+      },
+      coin: 'tcanton',
+      keys: [
+        '598f606cd8fc24710d2ebad89dce86c2',
+        '598f606cc8e43aef09fcb785221d9dd2',
+        '5935d59cf660764331bafcade1855fd7',
+      ],
+      multisigType: 'tss',
+    };
+
+    before(async function () {
+      cantonWallet = new Wallet(bitgo, bitgo.coin('tcanton'), walletData);
+      nock(bgUrl).get(`/api/v2/${cantonWallet.coin()}/key/${cantonWallet.keyIds()[0]}`).times(3).reply(200, {
+        id: '598f606cd8fc24710d2ebad89dce86c2',
+        pub: '5f8WmC2uW9SAk7LMX2r4G1Bx8MMwx8sdgpotyHGodiZo',
+        source: 'user',
+        encryptedPrv:
+          '{"iv":"hNK3rg82P1T94MaueXFAbA==","v":1,"iter":10000,"ks":256,"ts":64,"mode":"ccm","adata":"","cipher":"aes","salt":"cV4wU4EzPjs=","ct":"9VZX99Ztsb6p75Cxl2lrcXBplmssIAQ9k7ZA81vdDYG4N5dZ36BQNWVfDoelj9O31XyJ+Xri0XKIWUzl0KKLfUERplmtNoOCn5ifJcZwCrOxpHZQe3AJ700o8Wmsrk5H"}',
+        coinSpecific: {},
+      });
+
+      nock(bgUrl).get(`/api/v2/${cantonWallet.coin()}/key/${cantonWallet.keyIds()[1]}`).times(2).reply(200, {
+        id: '598f606cc8e43aef09fcb785221d9dd2',
+        pub: 'G1s43JTzNZzqhUn4aNpwgcc6wb9FUsZQD5JjffG6isyd',
+        encryptedPrv:
+          '{"iv":"UFrt/QlIUR1XeQafPBaAlw==","v":1,"iter":10000,"ks":256,"ts":64,"mode":"ccm","adata":"","cipher":"aes","salt":"7VPBYaJXPm8=","ct":"ajFKv2y8yaIBXQ39sAbBWcnbiEEzbjS4AoQtp5cXYqjeDRxt3aCxemPm22pnkJaCijFjJrMHbkmsNhNYzHg5aHFukN+nEAVssyNwHbzlhSnm8/BVN50yAdAAtWreh8cp"}',
+        source: 'backup',
+        coinSpecific: {},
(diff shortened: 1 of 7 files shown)


modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
```diff
@@ -381,6 +381,14 @@ export abstract class BaseCoin implements IBaseCoin {
...
   }
 
+  /**
+   * Check whether a coin requires wallet initialization
+   * @returns {boolean}
+   */
+  requiresWalletInitialization(): boolean {
+    return false;
+  }
+
```
balakrishna507, line 391: can we change this name to something like `requiresExplicitWalletInitialization`, or `requiresWalletInitializationTransaction`, or anything else that signals that this is an action the user needs to perform and sign? this is to differentiate it from other coins whose wallets also require initialization but no transaction for this.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3319,6 +3320,61 @@ export class Wallet implements IWallet {
...
+  public async sendWalletInitialization(params: PrebuildAndSignTransactionOptions = {}): Promise<WalletInitResult> {
+    if (!this.baseCoin.requiresWalletInitialization()) {
+      throw new Error(`Wallet initialization is not required for ${this.baseCoin.getFullName()}`);
+    }
+    if (params.reqId) {
+      this.bitgo.setRequestTracer(params.reqId);
+    }
+    const buildParams: PrebuildTransactionOptions = _.pick(params, this.prebuildWhitelistedParams());
+    if (!buildParams.type) {
+      buildParams.type = 'createAccount';
```
balakrishna507, line 3341: this is not a new transaction build type, right?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3319,6 +3320,61 @@ export class Wallet implements IWallet {
...
+    if (typeof params.prebuildTx === 'string' || params.prebuildTx?.buildParams?.type !== 'createAccount') {
+      throw new Error('Invalid build of wallet init');
+    }
+    const successfulTxs: PrebuildTransactionResult[] = [];
+    const failedTxs = new Array<Error>();
+    try {
+      if (this._wallet.multisigType === 'tss') {
+        const sendTx = await this.sendManyTxRequests(unsignedBuildWithOptions);
+        successfulTxs.push(sendTx);
+      } else {
```
balakrishna507: wouldn't canton wallet types always be `tss`?

modules/sdk-core/src/bitgo/wallet/wallet.ts
ravibitgo: no, an existing intent type

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
ravibitgo: updated

modules/sdk-core/src/bitgo/wallet/wallet.ts
ravibitgo: yes, updated

modules/sdk-coin-canton/.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 .idea/
 dist/
+.DS_Store
```
hrishikeshjain, line 4: is this intended change?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3319,6 +3320,48 @@ export class Wallet implements IWallet {
     };
   }
 
+  /**
+   * The chain canton, requires the wallet to be initialized by sending out a transaction
+   * to be onboarded onto the validator.
```
hrishikeshjain, line 3325: is new line intended here?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3439,6 +3482,16 @@ export class Wallet implements IWallet {
           params.preview
         );
         break;
+      case 'createAccount':
+        txRequest = await this.tssUtils!.prebuildTxWithIntent(
```
hrishikeshjain, line 3486: are we sure `tssUtils` won't be undefined?

hrishikeshjain (dismissed):
minor nits

modules/sdk-coin-canton/.gitignore
balakrishna507: this should be okay, right? we'd never want these in our repo and sometimes when we open the sdk repo's directories using finder these would be created?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3319,6 +3320,48 @@ export class Wallet implements IWallet {
...
+    const buildParams: PrebuildTransactionOptions = _.pick(params, this.prebuildWhitelistedParams());
+    if (!buildParams.type) {
+      buildParams.type = 'createAccount';
+    }
+    const prebuildTx = await this.prebuildTransaction(buildParams);
+    const unsignedBuildWithOptions: PrebuildAndSignTransactionOptions = {
+      ...params,
+      prebuildTx,
+    };
+    if (typeof params.prebuildTx === 'string' || params.prebuildTx?.buildParams?.type !== 'createAccount') {
```
balakrishna507, line 3348: do we need the `typeof` check?

```suggestion
    if (params.prebuildTx?.buildParams?.type !== 'createAccount') {
```

balakrishna507 (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
ravibitgo: it won't be since base coin here will be `canton` which is a eddsa tss coin, I'll remove the non-null assertion in a follow-up

modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -1247,6 +1247,87 @@ describe('V2 Wallet:', function () {
...
+      nock.cleanAll();
+    });
+
+    it('Should build wallet initialization transactions correctly', async function () {
+      const txRequestNock = nock(bgUrl)
+        .post(`/api/v2/wallet/${cantonWallet.id()}/txrequests`)
+        .reply((url, body) => {
+          const bodyParams = body as any;
+          bodyParams.intent.intentType.should.equal('createAccount');
+          bodyParams.intent.recipients.length.should.equal(0);
```
balakrishna507, line 1310: usually we use matchers. i'm not sure what `isDone()` would return if exceptions are thrown while processing `reply()`.

balakrishna507 (approved):

hrishikeshjain (approved):

PR #7300, opened 2025-10-22
feat(abstract-utxo): move and reorganize UTXO coin implementations and tests

OttoAllmendinger:
Move UTXO coin implementations from coin-specific modules into the
abstract-utxo module. Each coin implementation now resides in its own
directory under src/impl/ and is re-exported from the main module.

The original structure did not provide much benefit and makes testing much more difficult than it should be.

This change:
- Consolidates UTXO coin implementations in abstract-utxo
- Allows coin-specific modules to simply re-export implementations
- Simplifies maintenance and keeps the code DRY
- Moves test files from individual UTXO coin modules into abstract-utxo
- Reorganizes test directory structure with separate unit and integration
  test directories
- Adds sdk-test and mocha as devDependencies to abstract-utxo

This changes allows us to move the utxo tests from `bitgo/test/v2/unit/coins/utxo` to
`abstract-utxo` in subsequent PRs.

BTC-2652


ppongbitgo (approved):

modules/sdk-coin-bch/package.json
```diff
@@ -11,9 +11,7 @@
     "clean": "rm -r ./dist",
     "lint": "eslint --quiet .",
     "prepare": "npm run build",
-    "test": "npm run coverage",
-    "coverage": "nyc -- npm run unit-test",
-    "unit-test": "mocha"
+    "unit-test": "echo 'test in abstract-utxo'"
```
zahin-mohammad, line 14: Won't this be an issue for CI? We run `yarn run unit-test-changed`. So if a PR changes individual coins, but does not change abstract-utxo, then these tests will not run.

zahin-mohammad:
```
lerna run unit-test --since $(git merge-base HEAD~ origin/${GITHUB_REPO_BRANCH:-master}) --stream"
````

OttoAllmendinger: There is no `test/` directory for these coins anymore, the tests moved to `abstract-utxo/test/impl`

If a PR changes an individual coin, it will happen in `abstract-utxo/src/impl` and hence kick off the tests.

zahin-mohammad (approved):

PR #7322, opened 2025-10-23
fix: handle vet token recovery in coin class

at31416:
This is because wrw calls coin class' recover function instead of token class recover function

Ticket: COIN-6042


MohammedRyaan786 (approved):
LGTM

hrishikeshjain (approved):

PR #7348, opened 2025-10-26
feat(sdk-coin-iota): transaction builder for iota

parasgarg-bitgo:
- Implements transaction builder infrastructure for IOTA coin with transfer, validation, and broadcasting capabilities

TICKET: WIN-6890


modules/sdk-coin-iota/src/lib/utils.ts
```diff
@@ -31,7 +32,11 @@ export class Utils implements BaseUtils {
 
   /** @inheritdoc */
   isValidSignature(signature: string): boolean {
-    return fromBase64(signature).length === IOTA_SIGNATURE_LENGTH;
+    try {
+      return fromBase64(signature).length === IOTA_SIGNATURE_LENGTH;
```
mohd-kashif, line 36: can we have a regex against which the signature can be validated ?

mohd-kashif (approved):

modules/sdk-coin-iota/src/lib/transferTransaction.ts
```diff
@@ -0,0 +1,212 @@
...
+      if (this.paymentObjects.length > 1) {
+        let idx = 1;
+        while (idx < this.paymentObjects.length) {
+          [firstInputObj] = this._iotaTransaction.mergeCoins(
+            firstInputObj,
+            this.paymentObjects
+              .slice(idx, idx + MAX_INPUT_OBJECTS)
+              .map((input) => this._iotaTransaction.object(IotaInputs.ObjectRef(input)))
+          );
+          idx += MAX_INPUT_OBJECTS;
```
hitansh-madan, line 138: Please validate that this works as expected. Given this is similar to SUI, based on [iota's config](https://github.com/iotaledger/iota/blob/9ba7f807dfda3651ec9d106f495e787930b2ec98/crates/iota-protocol-config/src/lib.rs#L1575C13-L1575C26) we might need to use `max_arguments` instead of `MAX_INPUT_OBJECTS`. 

Also, not that important but it might be worth testing if we can even create tx with 2048 inputs because `MAX_INPUT_OBJECTS` represents the theoretical max number of inputs for a tx, but in practice it might be lower since the tx size is limited to 128 KiB.

hitansh-madan (approved):
Need to validate limits for edge cases, but changes generally look fine

modules/sdk-coin-iota/src/lib/transferTransaction.ts
parasgarg-bitgo: Iota also has a separate definition of [max input objects](https://github.com/iotaledger/iota/blob/9ba7f807dfda3651ec9d106f495e787930b2ec98/crates/iota-protocol-config/src/lib.rs#L1569).
`max_arguments` are relevant when we are making a Move call (equivalent to contract method call).

I'll look into creating a tx with theoretical max input size.

parasgarg-bitgo: The `max_arguments` limit is for Move package calls (equivalent to Contract method call).
There is a separate limit for [max_input_objects](https://github.com/iotaledger/iota/blob/9ba7f807dfda3651ec9d106f495e787930b2ec98/crates/iota-protocol-config/src/lib.rs#L1569)

I'll look into creating a tx with theoretical max number of objects

modules/sdk-coin-iota/src/lib/transferTransaction.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  /**
+   * @inheritDoc
+   */
+  protected explainTransactionImplementation(
+    json: TxData,
+    explanationResult: TransactionExplanation
+  ): TransactionExplanation {
+    const outputs: TransactionRecipient[] = this.recipients.map((recipient) => recipient);
```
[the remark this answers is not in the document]
parasgarg-bitgo: Will update in later revisions

modules/sdk-coin-iota/src/lib/utils.ts
parasgarg-bitgo: Since we are already doing the conversion from Base64, it'll validate the signature string character. Additionally, Iota does not have a pattern for valid signatures so we don't need to validate agains a regex.

PR #7356, opened 2025-10-27
feat(sdk-coin-dot): configure metadata for polkadot assethub

arvind-bg:
Ticket: WIN-7697


modules/sdk-core/src/bitgo/environments.ts
```diff
@@ -194,7 +194,7 @@ const mainnetBase: EnvironmentTemplate = {
   cronosNodeUrl: 'https://cronos-pos-rest.publicnode.com/',
   initiaNodeUrl: 'https://rest.initia.xyz/',
   asiNodeUrl: 'https://rest-fetchhub.fetch.ai/',
-  dotNodeUrls: ['wss://rpc.polkadot.io'],
+  dotNodeUrls: ['wss://asset-hub-polkadot-rpc.n.dwellir.com'],
```
zahin-mohammad, line 197: Where did you get this link? Seems to differ from https://www.dwellir.com/networks/asset-hub

arvind-bg: It is available in their home page here https://www.dwellir.com/
I was able to connect to wss://asset-hub-polkadot-rpc.n.dwellir.com and query their metadata. I was not able to do so with the one listed in the endpoint you shared. it could be because it is priced.

noel-bitgo (approved):

Ranjna-G (approved):
LGTM

PR #7364, opened 2025-10-28
feat(express): migrated fanoutunspents as type route

lokesh-bitgo:
Ticket: WP-5431

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -812,10 +812,10 @@ export async function handleV2ConsolidateAccount(req: express.Request) {
  * handle wallet fanout unspents
  * @param req
  */
-async function handleV2FanOutUnspents(req: express.Request) {
+async function handleV2FanOutUnspents(req: ExpressApiRouteRequest<'express.v2.wallet.fanoutunspents', 'post'>) {
   const bitgo = req.bitgo;
-  const coin = bitgo.coin(req.params.coin);
-  const wallet = await coin.wallets().get({ id: req.params.id });
+  const coin = bitgo.coin(req.decoded.coin);
+  const wallet = await coin.wallets().get({ id: req.decoded.id });
   return wallet.fanoutUnspents(createSendParams(req));
 }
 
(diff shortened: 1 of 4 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -812,7 +812,7 @@ export async function handleV2ConsolidateAccount(req: express.Request) {
  * handle wallet fanout unspents
  * @param req
  */
-async function handleV2FanOutUnspents(req: express.Request) {
+async function handleV2FanOutUnspents(req: ExpressApiRouteRequest<'express.v2.wallet.fanoutunspents', 'post'>) {
   const bitgo = req.bitgo;
   const coin = bitgo.coin(req.params.coin);
   const wallet = await coin.wallets().get({ id: req.params.id });
```
zahin-mohammad: Don't we need to access the params via the `decoded` object?

modules/express/src/typedRoutes/api/v2/fanoutUnspents.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Request parameters for fanning out unspents in a wallet (v2)
+ */
+export const FanoutUnspentsRequestParams = {
+  /** The coin identifier (e.g., 'btc', 'tbtc') */
+  coin: t.string,
+  /** The ID of the wallet */
+  id: t.string,
+};
```
zahin-mohammad:
```suggestion
} as const;
```

modules/express/src/typedRoutes/api/v2/fanoutUnspents.ts
```diff
@@ -0,0 +1,119 @@
...
+  maxFeePercentage: optional(t.number),
+  /** Estimate fees to aim for first confirmation within this number of blocks */
+  feeTxConfirmTarget: optional(t.number),
+  /** Comment to attach to the transaction */
+  comment: optional(t.string),
+  /** One-time password for 2FA */
+  otp: optional(t.string),
+  /** Target address for the fanout outputs */
+  targetAddress: optional(t.string),
+};
```
zahin-mohammad:
```suggestion
} as const;
```

modules/express/src/clientRoutes.ts
lokesh-bitgo: Yes, we need to add decoded here.

PR #7398, opened 2025-11-03
feat(express): migrated sendmany as type route

lokesh-bitgo:
Ticket: WP-5427

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -935,11 +935,11 @@ async function handleV2SendOne(req: express.Request) {
  * handle send many
  * @param req
  */
-async function handleV2SendMany(req: express.Request) {
+async function handleV2SendMany(req: ExpressApiRouteRequest<'express.v2.wallet.sendmany', 'post'>) {
   const bitgo = req.bitgo;
-  const coin = bitgo.coin(req.params.coin);
+  const coin = bitgo.coin(req.decoded.coin);
   const reqId = new RequestTracer();
-  const wallet = await coin.wallets().get({ id: req.params.id, reqId });
+  const wallet = await coin.wallets().get({ id: req.decoded.id, reqId });
   req.body.reqId = reqId;
   let result;
   try {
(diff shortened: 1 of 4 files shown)


rahulhegde-bitgo (dismissed):
LGTM

danielzhao122 (dismissed):

modules/express/src/clientRoutes.ts
```diff
@@ -935,11 +935,11 @@ async function handleV2SendOne(req: express.Request) {
  * handle send many
  * @param req
  */
-async function handleV2SendMany(req: express.Request) {
+async function handleV2SendMany(req: ExpressApiRouteRequest<'express.v2.wallet.sendmany', 'post'>) {
```
zahin-mohammad, line 938: While this works, we lose the type saftey of api-ts when using things like `createTSSSendParams` and `createSendParams`. Can we have a follow up to fix these? 

As long as we have those two utilities, the codec and the actual request body we are acting on won't be aligned.

modules/express/src/typedRoutes/api/v2/sendmany.ts
```diff
@@ -0,0 +1,475 @@
...
+ */
+const Recipient = t.intersection([RecipientParams, RecipientParamsOptional]);
+
+/**
+ * Request body for sending to multiple recipients (v2)
+ *
+ * This endpoint supports the full set of parameters available in the BitGo SDK
+ * for building, signing, and sending transactions to multiple recipients.
+ */
+export const SendManyRequestBody = {
```
zahin-mohammad, line 109: How can _everything_ be optional?

zahin-mohammad: These codecs will eventually generate our api-docs, so it will be an issue if we advertise params as required, then change to optional, then back to required.

modules/express/src/typedRoutes/api/v2/sendmany.ts
```diff
@@ -0,0 +1,475 @@
...
+  /** Pending approval object or ID (if transaction requires approval) */
+  pendingApproval: t.unknown,
+});
+
+/**
+ * TSS transaction response object (for TSS wallets using txRequest flow)
+ */
+const SendManyTssResponse = t.type({
+  /** Transaction request object */
+  txRequest: t.unknown,
```
zahin-mohammad: This will look like a breaking change in our docs. 

Please cross-reference https://developers.bitgo.com/reference/expresswalletsendmany to ensure we aren't introducing breaking changes, and if we are, we should only be introducing them to match what the code is currently doing.

lokesh-bitgo: Adding the proper TxRequest type codec here from public-types.

modules/express/src/typedRoutes/api/v2/sendmany.ts
lokesh-bitgo: As per code implementation, all fields are optional - https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-core/src/bitgo/wallet/iWallet.ts#L90

modules/express/src/clientRoutes.ts
lokesh-bitgo: I have created a follow-up ticket for this - https://bitgoinc.atlassian.net/browse/WP-6651, will take it up.

modules/express/src/typedRoutes/api/v2/sendmany.ts
zahin-mohammad: :(

PR #7401, opened 2025-11-03
feat(sei): enabling staking flag in testnet

Doddanna17:
feat(sei): enabling staking flag in testnet
TICKET: SC-3783


abhijit0943 (approved):

modules/statics/src/allCoinsAndTokens.ts
```diff
@@ -813,7 +813,7 @@ export const allCoinsAndTokens = [
     6,
     UnderlyingAsset.SEI,
     BaseUnit.SEI,
-    SEI_FEATURES
+    [...SEI_FEATURES, CoinFeature.STAKING]
```
prithvishet2503: to enable staking feature for SEI OR SEI EVM 
in the ticket it seems its supposed to be SEI EVM right

PR #7426, opened 2025-11-04
fix: validate 32-byte message requirement when shouldHash=false

orentysor:
When shouldHash=false, the message must be exactly 32 bytes (a hash). Messages longer than 32 bytes caused calculation errors (messages larger than the curve order when converted to bigint) or verification failures because secp256k1.recoverPublicKey() expects exactly 32 bytes.

Adds validation in sign() and verify() methods to enforce this requirement and adds tests to verify the fix.

Refs: HSM-129


zhongxishen (approved):
LGTM

PR #7447, opened 2025-11-07
fix(express): fanoutUnspentsV2 type codec

lokesh-bitgo:
Ticket: WP-6658

### DIFF
--- a/modules/express/src/typedRoutes/api/v2/fanoutUnspents.ts
+++ b/modules/express/src/typedRoutes/api/v2/fanoutUnspents.ts
@@ -54,6 +54,8 @@ export const FanoutUnspentsRequestBody = {
   otp: optional(t.string),
   /** Target address for the fanout outputs */
   targetAddress: optional(t.string),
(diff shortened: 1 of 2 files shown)


modules/express/src/typedRoutes/api/v2/fanoutUnspents.ts
```diff
@@ -54,6 +54,8 @@ export const FanoutUnspentsRequestBody = {
   otp: optional(t.string),
   /** Target address for the fanout outputs */
   targetAddress: optional(t.string),
+  /** Transaction format type (e.g., 'legacy', 'psbt', 'psbt-lite') - controls output format */
+  txFormat: optional(t.union([t.literal('legacy'), t.literal('psbt'), t.literal('psbt-lite')])),
```
Marzooqa, line 58: Can we have this as like an enum?

lokesh-bitgo: I’ll take this up during the refactoring task. Some codecs are shared across multiple APIs, so I’ll move them to a common files and handle these cases in the follow-up PRs.

Marzooqa (approved):

PR #7448, opened 2025-11-07
fix(express): consolidateUnspentsV2 type codec

lokesh-bitgo:
Ticket: WP-6658

### DIFF
--- a/modules/express/src/typedRoutes/api/v2/consolidateunspents.ts
+++ b/modules/express/src/typedRoutes/api/v2/consolidateunspents.ts
@@ -52,6 +52,8 @@ export const ConsolidateUnspentsRequestBody = {
   otp: optional(t.string),
   /** Target address for the consolidation outputs */
   targetAddress: optional(t.string),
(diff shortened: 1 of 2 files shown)


modules/express/src/typedRoutes/api/v2/consolidateunspents.ts
```diff
@@ -52,6 +52,8 @@ export const ConsolidateUnspentsRequestBody = {
   otp: optional(t.string),
   /** Target address for the consolidation outputs */
   targetAddress: optional(t.string),
+  /** Transaction format type (e.g., 'legacy', 'psbt', 'psbt-lite') - controls output format */
+  txFormat: optional(t.union([t.literal('legacy'), t.literal('psbt'), t.literal('psbt-lite')])),
```
Marzooqa, line 56: Can we have this as an enum?

Marzooqa (approved):

modules/express/src/typedRoutes/api/v2/consolidateunspents.ts
lokesh-bitgo: Will do it in refactoring task.

PR #7450, opened 2025-11-07
docs: clarified default multisigtype dx-2273

vmccarty:
Updated the description of the `multisigType` to clarify the defaults. Also updated the endpoint description while I was at it, since it was... weird.

Resolves DX-2273.

### DIFF
--- a/modules/express/src/typedRoutes/api/v2/generateWallet.ts
+++ b/modules/express/src/typedRoutes/api/v2/generateWallet.ts
@@ -13,7 +13,7 @@ export const GenerateWalletBody = {
   label: t.string,
   /** Enterprise id. Required for Ethereum wallets since they can only be created as part of an enterprise. Optional for other coins. */
   enterprise: optional(t.string),
-  /** If absent, BitGo uses the default wallet type for the asset */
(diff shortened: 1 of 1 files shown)


modules/express/src/typedRoutes/api/v2/generateWallet.ts
```diff
@@ -87,19 +87,17 @@ export const GenerateWalletV2Query = {
...
+ * 2. Creates the backup keychain locally on your machine.
+ * 3. Uploads the encrypted user keychain and public backup keychain to BitGo.
+ * 4. Creates the BitGo key (and the backup key if you pass `backupXpubProvider`) on the service.
+ * 5. Creates the wallet on BitGo with the 3 public keys above.
  *
- * ⓘ This endpoint should be called through BitGo Express if used without the SDK, such as when using cURL.
+ * ⓘ Many account-based assets, including Ethereum, require you to [Fund Gas Tanks](https://developers.bitgo.com/docs/get-started-gas-tanks#/) to initialize a new wallets on chain. Ensure your gas tank has a sufficient balance to cover this cost before generating a new wallet.
  *
  * @operationId express.wallet.generate
  */
```
danielzhao122, line 104: Hey seems like this was missed in the migration, could you add a `express` tag here
Thanks!

vmccarty: Good catch. Added it.

danielzhao122 (approved):

PR #7453, opened 2025-11-07
feat(sdk-coin-sol): support SetComputeUnitLimit instructions

derekchen745:
Ticket: SC-3899


modules/sdk-coin-sol/src/lib/utils.ts
```diff
@@ -403,6 +403,9 @@ export function getInstructionType(instruction: TransactionInstruction): ValidIn
         );
       }
     case COMPUTE_BUDGET:
+      if (instruction.data.length >= 1 && instruction.data[0] === 2) {
+        return 'SetComputeUnitLimit';
+      }
       return 'SetPriorityFee';
```
haritkapadia, line 410: Use the official decoder if you can: https://github.com/solana-foundation/solana-web3.js/blob/c26c13bf841821d9bbf83bb476c567fe29d13821/src/programs/compute-budget.ts#L180

Just like here: https://github.com/BitGo/BitGoJS/blob/579dcb22db336e8fcb1825791da5e1d0fc1ec787/modules/sdk-coin-sol/src/lib/utils.ts#L395

haritkapadia (changes requested):
Everything else looks good

modules/sdk-coin-sol/src/lib/utils.ts
derekchen745: good catch, updated!

haritkapadia (approved):

PR #7462, opened 2025-11-10
feat(sdk-coin-ada): add unspents and seq id in transfer token intent

Ranjna-G:
Ticket: WIN-7804

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -3448,6 +3448,8 @@ export class Wallet implements IWallet {
             recipients: params.recipients || [],
             nonce: params.nonce,
             feeOptions,
+            unspents: params.unspents,
+            sequenceId: params.sequenceId,
           },
           apiVersion,
           params.preview
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -21,6 +21,7 @@ import {
   MessageStandardType,
   MessageTypes,
   PopulatedIntent,
+  PrebuildTransactionOptions,
   PrebuildTransactionWithIntentOptions,
   RequestTracer,
   SendManyOptions,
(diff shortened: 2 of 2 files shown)


abhi-bitgo (dismissed):
lgtm

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3440,18 +3440,23 @@ export class Wallet implements IWallet {
...
+          isTss: params.isTss,
+          intentType: 'transferToken',
+          recipients: params.recipients || [],
+          nonce: params.nonce,
+          feeOptions,
+        };
+        const tokenName = params.recipients?.[0]?.tokenName;
+        if (tokenName) {
+          const coin = this.bitgo.coin(tokenName);
+          if (coin && typeof coin.getCoinSpecificBuildParams === 'function') {
```
kamleshmugdiya: I think this check is not enough.
since all coin implementations should now have this method via inheritance from BaseCoin.

can you check if how `coin.getCoinSpecificBuildParams === 'function')` when coin is non-ada token?

kamleshmugdiya: maybe the check is redundant.
you need not check it, since the default impl is empty.

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
```diff
@@ -626,4 +627,5 @@ export interface IBaseCoin {
    */
   assertIsValidKey({ publicKey, encryptedPrv, walletPassphrase, multiSigType }: AuditKeyParams): void;
   requiresWalletInitializationTransaction(): boolean;
+  getCoinSpecificBuildParams(params: PrebuildTransactionOptions): Record<string, unknown>;
```
kamleshmugdiya: IMHO, this will introduces an interface bloat issue.
all coins now carry a method that only ADA token needs.

Ranjna-G: Yes but adding a coin specific check is also not ideal. What other solution do you propose ? We are adding methods in this interface eg: requiresWalletInitializationTransaction, setCoinSpecificFieldsInIntent, etc that are implemented only by few coins

kamleshmugdiya: there is already a func `setCoinSpecificFieldsInIntent` in `BaseCoin`
you can override this in ADA token class instead of introducing a new func.

kamleshmugdiya (approved):
lgtm

PR #7480, opened 2025-11-11
feat: add audit api spec to pr workflow

danielzhao122:
This PR adds automated API specification auditing as a GitHub Actions workflow that runs on every pull request. Previously, we only audited the API spec during releases, which meant API specification errors could be merged into the codebase and only discovered later in the release process.

Since the actual `audit-api-spec` workflow from dev-portal requires a vault token, we've implemented our own workflow that mimics the actual audit process used in dev-portal without the need of the token.
The new workflow:
- Generates the API specification from TypeScript routes using `@api-ts/openapi-generator@v5`
- Audits the generated spec using `vacuum v0.18.1` (same version as dev-portal)
- Enforces API quality rules via the same `ruleset.yaml `used in `dev-portal`

Test run that shows the `audit-api-specs` job fails when it detects errors: https://github.com/BitGo/BitGoJS/actions/runs/19271735416/job/55101534330?pr=7480

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -368,3 +368,86 @@ jobs:
             git diff
             exit 1
           fi
+
+  audit-api-spec:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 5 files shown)


zahin-mohammad (approved):

adityasoni-bitgo (approved):
Lgtm

PR #7493, opened 2025-11-12
feat: add FBTC token support in ERC20 and OFC ERC20 definitions

shubham-damkondwar:
Ticket: WIN-7847


modules/statics/src/base.ts
```diff
@@ -53,6 +53,7 @@ export enum CoinFamily {
   ETHW = 'ethw',
   ETC = 'etc',
   EOS = 'eos',
+  FBTC = 'fbtc',
```
ArunBala-Bitgo: Underlying asset should be eth:fbtc

prithvishet2503: eth:fbtc pls change

prithvishet2503 (approved):

PR #7504, opened 2025-11-12
feat: use wasm-utxo for address generation on testnets

lcovar:
Use wasm-utxo's fixedScriptWallet.address for testnets instead of utxo-lib's address derivation. This improves compatibility with the native implementation.

Co-authored-by: llm-git <llm-git@ttll.de>

Ticket: BTC-2720

TICKET: BTC-2720

### DIFF
--- a/modules/abstract-utxo/src/address/fixedScript.ts
+++ b/modules/abstract-utxo/src/address/fixedScript.ts
@@ -14,6 +14,7 @@ import {
 import * as utxolib from '@bitgo/utxo-lib';
 import { bitgo } from '@bitgo/utxo-lib';
 import { bip32 } from '@bitgo/secp256k1';
+import * as wasmUtxo from '@bitgo/wasm-utxo';
 
 type ScriptType2Of3 = bitgo.outputScripts.ScriptType2Of3;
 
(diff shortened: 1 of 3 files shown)


mukeshsp (dismissed):

modules/utxo-lib/src/bitgo/wallet/WalletScripts.ts
OttoAllmendinger, line 1: actually we want to leave utxo-lib unchanged

wasm-utxo is going to be the the replacement for utxo-lib, not an addition

modules/abstract-utxo/src/address/fixedScript.ts
OttoAllmendinger, line 82: we are doing a conversion to the `format` here, I believe you can pass `format` as the last parameter in `address()` above

https://github.com/BitGo/BitGoWASM/blob/master/packages/wasm-utxo/js/fixedScriptWallet.ts#L44

it looks like there is missing test coverage, please add a test that shows that the `format` param is considered

OttoAllmendinger (changes requested):

modules/abstract-utxo/src/address/fixedScript.ts
```diff
@@ -56,6 +57,19 @@ export function generateAddressWithChainAndIndex(
   index: number,
   format: CreateAddressFormat | undefined
 ): string {
+  if (utxolib.isTestnet(network)) {
+    // Convert CreateAddressFormat to AddressFormat for wasm-utxo
+    // 'base58' -> 'default', 'cashaddr' -> 'cashaddr'
+    const wasmFormat = format === 'base58' ? 'default' : format;
```
OttoAllmendinger, line 63: oops my bad

OttoAllmendinger (approved):

PR #7510, opened 2025-11-13
fix(express): pendingApprovalV1 type codec

lokesh-bitgo:
Ticket: WP-6717

### DIFF
--- a/modules/express/src/typedRoutes/api/v1/pendingApproval.ts
+++ b/modules/express/src/typedRoutes/api/v1/pendingApproval.ts
@@ -8,6 +8,8 @@ export const pendingApprovalRequestParams = {
 };
 
 export const pendingApprovalRequestBody = {
+  /** State of the approval: 'approved' to approve, omit or 'rejected' to reject (defaults to rejection) */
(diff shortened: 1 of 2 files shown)


Taseen08 (commented):
(duplicate comment)

modules/express/src/typedRoutes/api/v1/pendingApproval.ts
```diff
@@ -8,6 +8,8 @@ export const pendingApprovalRequestParams = {
 };
 
 export const pendingApprovalRequestBody = {
+  /** State of the approval: 'approved' to approve, omit or 'rejected' to reject (defaults to rejection) */
```
Taseen08, line 11: nit: would it make more sense if we say: `'approved' to approve, 'rejected' to reject (defaults to rejection)`
Some might think `omit` is a valid state to pass. I think `defaults to rejection` already covers that providing empty state value leads to rejection

lokesh-bitgo: Will update this in the follow-up API doc tickets.

PR #7511, opened 2025-11-13
fix(account-lib): remove message whitelist for cold and custody wallets

bhavidhingra:
TICKET: COIN-6527

This pull request simplifies the message verification logic in the `modules/account-lib/src/index.ts` file by removing redundant validation and unused imports. The most important changes are:

Message verification logic:

* The `verifyMessage` function now only checks the validity of the encoded payload using `verifyEncodedPayload`, removing the additional call to `validateAgainstMessageTemplates`. This streamlines the verification process and avoids redundant checks.

Code cleanup:

* The unused `validateAgainstMessageTemplates` import was removed from the file, as it is no longer needed after the above change.

### DIFF
--- a/modules/account-lib/src/index.ts
+++ b/modules/account-lib/src/index.ts
@@ -206,7 +206,7 @@ export { CosmosSharedCoin };
 import * as Canton from '@bitgo/sdk-coin-canton';
 export { Canton };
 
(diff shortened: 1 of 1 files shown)


akarath (approved):
Approved

PR #7513, opened 2025-11-13
fix(sdk-core): include coredao family for acceleration

bhavidhingra:
TICKET: COIN-6534

This pull request adds support for the `coredao` coin family to the intent population logic and its associated tests. The main changes ensure that `coredao` is treated similarly to other EVM-compatible chains like `eth`, `polygon`, and `bsc` when creating transaction intents.

**Support for coredao in intent population:**

* Updated the conditional in `MpcUtils.populateIntent` to include `coredao` as a supported coin family for intent creation, aligning its behavior with other EVM-compatible families (`eth`, `polygon`, `bsc`).

**Testing enhancements:**

* Added a new unit test in `wallet.ts` to validate that intent population for `coredao` acceleration works correctly, ensuring the correct structure and values in the resulting intent.

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/mpcUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/mpcUtils.ts
@@ -191,7 +191,7 @@ export abstract class MpcUtils {
       isTestTransaction: params.isTestTransaction,
     };
 
-    if (baseCoin.getFamily() === 'eth' || baseCoin.getFamily() === 'polygon' || baseCoin.getFamily() === 'bsc') {
+    if (['eth', 'polygon', 'bsc', 'coredao'].includes(baseCoin.getFamily())) {
       switch (params.intentType) {
         case 'payment':
         case 'transferToken':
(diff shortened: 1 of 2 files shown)


akarath (approved):
lgtm

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
```diff
@@ -191,7 +191,7 @@ export abstract class MpcUtils {
       isTestTransaction: params.isTestTransaction,
     };
 
-    if (baseCoin.getFamily() === 'eth' || baseCoin.getFamily() === 'polygon' || baseCoin.getFamily() === 'bsc') {
+    if (['eth', 'polygon', 'bsc', 'coredao'].includes(baseCoin.getFamily())) {
```
akarath, line 194: with this, we might not receive some fields for polygon like 
```
        token: params.tokenName,
        enableTokens: params.enableTokens,
```
We checked that it shouldn't be an issue

PR #7526, opened 2025-11-14
feat(sdk-core): exclude keychains for multi-user-key

pranavjain97:
TICKET: WP-6780

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -1594,12 +1594,26 @@ export class Wallet implements IWallet {
     return userPrv;
   }
 
+  private isMultiUserKeyWallet(): boolean {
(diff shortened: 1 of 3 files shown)


mohammadalfaiyazbitgo (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1594,13 +1594,27 @@ export class Wallet implements IWallet {
...
+  }
+
   /**
    * Send an encrypted wallet share to BitGo.
    * @param params
    */
   async createShare(params: CreateShareOptions = {}): Promise<any> {
     common.validateParams(params, ['user', 'permissions'], []);
 
+    const isMultiUserKeyWallet = this.isMultiUserKeyWallet();
```
zahin-mohammad, line 1608: We also need to update `shareWallet`. I created a follow up ticket for this.

zahin-mohammad (approved):

PR #7529, opened 2025-11-14
feat(sdk-core): update wallet sharing for multi-user-key

zahin-mohammad:
Updates wallet sharing to handle multi-user-key OFC wallets. When wallets support `multi-user-key`, wallet shares do not require to share the keychain with the receiver.

WP-6790

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -2072,9 +2072,11 @@ describe('V2 Wallet:', function () {
 
     describe('OFC Multi-User-Key Wallet Sharing', function () {
       const userId = '123';
+      const email = 'shareto@sdktest.com';
       const permissions = 'view,spend';
       let ofcWallet: Wallet;
       let ofcMultiUserKeyWallet: Wallet;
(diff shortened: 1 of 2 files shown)


modules/bitgo/test/v2/unit/wallet.ts
```diff
@@ -6254,7 +6432,7 @@ describe('V2 Wallet:', function () {
         });
       });
 
-      it('should validate aptosCustomTransactionParams structure for smart contract calls', async function () {
+      xit('should validate aptosCustomTransactionParams structure for smart contract calls', async function () {
```
margueriteblair: is the reason for skipping related to these changes? or there's a flake?

margueriteblair (commented):
changes look good, i'd just fix the failing ci

modules/bitgo/test/v2/unit/wallet.ts
zahin-mohammad: Forgot to uncomment this. This test either takes way too long or times out for me locally. I'll have to debug it separately.

mohammadalfaiyazbitgo (approved):

PR #7530, opened 2025-11-14
feat(sdk-core): update wallet share acceptance for multi-user-key

zahin-mohammad:
## Summary

Adds support for accepting wallet shares when `userMultiKeyRotationRequired` is true. When this flag is set, users must provide a password to generate a new keychain, and both the public key and encrypted private key are sent to accept the share.

## Changes

- **`acceptShare()` and `processAcceptShare()`**: Handle `userMultiKeyRotationRequired` by requiring `userPassword`, generating a new keychain, encrypting the private key, and sending both `pub` and `encryptedPrv` to accept the share
- **Type definitions**: Added `userMultiKeyRotationRequired?: boolean` to `WalletShare` interface and `pub?: string` to `UpdateShareOptions` interfaces
- **Testing**: Added comprehensive test coverage for error handling, successful acceptance, and bulk operations

## Key Difference

Unlike `keychainOverrideRequired`, when `userMultiKeyRotationRequired` is true, the wallet is **NOT** reshared with spenders after acceptance. The user generates their own keychain rather than using a provided one.

**Ticket**: WP-6769

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -663,7 +663,7 @@ export interface WalletShare {
    */
   keychainOverrideRequired?: boolean;
   /**
-   * If true, the wallet share requires the user to provide to generate a new key-pair.
+   * If true, the wallet share requires the user to to generate a new key-pair.
    * In addition to the encrypted private key, the user must provide the public key.
    * */
   userMultiKeyRotationRequired?: boolean;
--- a/modules/sdk-core/src/bitgo/wallet/iWallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallet.ts
@@ -657,7 +657,16 @@ export interface WalletShare {
   enterprise?: string;
   message?: string;
   pendingApprovalId?: string;
+  /**
+   * If true, the wallet share requires the user to provide a new keychain.
+   * When accepted, the wallet will be re-shared with all other spenders.
+   */
   keychainOverrideRequired?: boolean;
+  /**
+   * If true, the wallet share requires the user to provide to generate a new key-pair.
+   * In addition to the encrypted private key, the user must provide the public key.
+   * */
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -657,7 +657,16 @@ export interface WalletShare {
   enterprise?: string;
   message?: string;
   pendingApprovalId?: string;
+  /**
+   * If true, the wallet share requires the user to provide a new keychain.
+   * When accepted, the wallet will be re-shared with all other spenders.
+   */
   keychainOverrideRequired?: boolean;
+  /**
+   * If true, the wallet share requires the user to provide to generate a new key-pair.
```
Chent03: _If true, the wallet share requires the user to generate a new key pair._

mohammadalfaiyazbitgo (dismissed):

modules/sdk-core/src/bitgo/wallet/iWallet.ts
zahin-mohammad: Good catch, fixed the typo.

mohammadalfaiyazbitgo (approved):

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -1028,6 +1055,26 @@ export class Wallets implements IWallets {
     });
     const newWalletPassphrase = params.newWalletPassphrase || params.userLoginPassword;
     const keysForWalletShares = walletShares.flatMap((walletShare) => {
+      // Handle userMultiKeyRotationRequired case - these shares don't have keychains
+      if (walletShare.userMultiKeyRotationRequired) {
+        if (!params.userLoginPassword) {
```
ayush-9, line 1060: here params.userLoginPassword is checked but at line 1065 newWalletPassphrase is used which can be params.newWalletPassphrase || params.userLoginPassword;

PR #7536, opened 2025-11-17
feat(sdk-coin-canton): added memoId check for isValidAddress

borkds:
Ticket: COIN-6566


modules/sdk-coin-canton/src/lib/utils.ts
```diff
@@ -56,6 +58,15 @@ export class Utils implements BaseUtils {
     return regex.test(value);
   }
 
+  /**
+   * Method to validate the input is a valid memo id
+   * @param {String} value the memo id value
+   * @returns {Boolean} true if valid
+   */
+  isValidMemoId(value: string): boolean {
+    return /^[0-9]+$/.test(value);
```
ravibitgo: we might need to cover special characters too here

borkds: we allow any memos into the system? Should I not validate anything at all?

modules/sdk-coin-canton/test/resources.ts
```diff
@@ -84,6 +84,8 @@ export const CANTON_ADDRESSES = {
   INVALID_FINGERPRINT: '12205::12205b4e3537a95126d9060459234gd8ad3c3ddccda4f79901954280ee19c576714d',
   MISSING_PARTY_HINT: '::12205b4e3537a95126d9060459234gd8ad3c3ddccda4f79901954280ee19c576714d',
   MISSING_FINGERPRINT: '12205::',
+  INVALID_MEMO_ID: '1220a::1220a43d89dc7d8f85316116aac093667f769fce55411aef6846ccb933b2e1a3b598?memoId=xyz',
```
ravibitgo: this is not invalid

ravibitgo (approved):
looks good

PR #7556, opened 2025-11-19
feat(express): migrated pendingApprovalV2 to type route

lokesh-bitgo:
Ticket: WP-5437

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -664,11 +664,13 @@ export async function handleV2CreateAddress(req: ExpressApiRouteRequest<'express
  * handle v2 approve transaction
  * @param req
  */
-async function handleV2PendingApproval(req: express.Request): Promise<any> {
+async function handleV2PendingApproval(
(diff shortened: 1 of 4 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -664,11 +664,13 @@ export async function handleV2CreateAddress(req: ExpressApiRouteRequest<'express
  * handle v2 approve transaction
  * @param req
  */
-async function handleV2PendingApproval(req: express.Request): Promise<any> {
+async function handleV2PendingApproval(
+  req: ExpressApiRouteRequest<'express.v2.pendingapprovals', 'put'>
+): Promise<any> {
```
Taseen08, line 669: nit: we can possibly leave out the `Promise<any>` as TS will auto infer it from what the function returns - that's how rest of the async handlers are doing it. Suggesting this to avoid using `any`

lokesh-bitgo: Yeah sure, will do in follow up PR

PR #7561, opened 2025-11-19
feat: wallet ownership validation for eth-like, ada, apt, vet, icp

danielzhao122:
Fix `AbstractEthLikeNewCoins.isWalletAddress` forwarder version bypass
   - Remove unconditional `return true` for versions 0, 3, 5
   - Implement validation for these versions

Fix partial implementations for `ada`, `apt`, `icp`, `vet`
   - add wallet address verification for these coins, previously they were only format checks  

ticket: WP-6461

### DIFF
--- a/modules/sdk-coin-icp/src/icp.ts
+++ b/modules/sdk-coin-icp/src/icp.ts
@@ -8,6 +8,7 @@ import {
   Ecdsa,
   ECDSAUtils,
   Environments,
+  InvalidAddressError,
   KeyPair,
   MPCAlgorithm,
   MultisigType,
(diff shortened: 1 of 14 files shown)


modules/sdk-coin-icp/src/icp.ts
```diff
@@ -141,7 +144,20 @@ export class Icp extends BaseCoin {
   }
 
   async isWalletAddress(params: TssVerifyAddressOptions): Promise<boolean> {
-    return this.isValidAddress(params.address);
+    const { address } = params;
```
zahin-mohammad, line 147: Looks like there might be a difference between wallet version 1 and other wallet versions. Double check if the address verification is the same between these 2.

zahin-mohammad (commented):
PR will be split by coin.

PR #7563, opened 2025-11-19
feat: add wallet address verification for eth and eth like

danielzhao122:
Fix `AbstractEthLikeNewCoins.isWalletAddress` forwarder version bypass
- Remove unconditional return true for versions 0, 3, 5
- Implement validation for these versions

Fix `forwarder version` bug in address verification
- forwarder v4 requires feeAddress to calculate address
- wallet v4 and v5 requires `walletV4ForwarderFactoryAddress` and `walletV4ForwarderImplementationAddress`

TICKET: WP-6461


modules/sdk-coin-eth/test/unit/eth.ts
```diff
@@ -662,182 +664,250 @@ describe('ETH:', function () {
   });
 
   describe('Address Verification', function () {
-    it('should verify an address generated using forwarder version 0', async function () {
-      const coin = bitgo.coin('teth') as Teth;
+    describe('isWalletAddress', function () {
+      it('should verify an address generated using forwarder version 0', async function () {
```
zahin-mohammad: Doesn't forwarder version 0 always return true?

modules/sdk-coin-eth/test/unit/eth.ts
```diff
@@ -662,182 +664,250 @@ describe('ETH:', function () {
   });
 
   describe('Address Verification', function () {
-    it('should verify an address generated using forwarder version 0', async function () {
-      const coin = bitgo.coin('teth') as Teth;
+    describe('isWalletAddress', function () {
```
zahin-mohammad, line 667: Are we missing tests for wallet version 5 and 6?

modules/sdk-coin-eth/test/unit/eth.ts
danielzhao122: yeah this test is kind of useless. I kept it for completeness.
Will delete it.

modules/sdk-coin-eth/test/unit/eth.ts
danielzhao122: Version 6 test is on line 828.
Will add version 5 test

zahin-mohammad (dismissed):
Looks good to me, lets also get an ETH-ALT review for the same

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2732,23 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
     return {};
   }
 
+  getFactoryAndImplContractAddresses(walletVersion: number | undefined): {
```
mullapudipruthvik:
```suggestion
  getForwarderFactoryAndImplContractAddresses(walletVersion: number | undefined): {
```

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2732,23 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+    forwarderFactoryAddress: string;
+    forwarderImplementationAddress: string;
+  } {
+    const ethNetwork = this.getNetwork();
+    if (walletVersion && (walletVersion === 5 || walletVersion === 4)) {
```
mullapudipruthvik: what about walletVersion 2 ?

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2736,45 +2760,57 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
    * @throws {UnexpectedAddressError}
    * @returns {boolean} True iff address is a wallet address
    */
-  async isWalletAddress(params: VerifyEthAddressOptions): Promise<boolean> {
+  async isWalletAddress(params: VerifyEthAddressOptions | TssVerifyEthAddressOptions): Promise<boolean> {
```
mullapudipruthvik, line 2960: is this called only for receive address or can base address be called as well ?

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2732,23 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
   }
 
+  getFactoryAndImplContractAddresses(walletVersion: number | undefined): {
+    forwarderFactoryAddress: string;
+    forwarderImplementationAddress: string;
+  } {
+    const ethNetwork = this.getNetwork();
+    if (walletVersion && (walletVersion === 5 || walletVersion === 4)) {
+      return {
+        forwarderFactoryAddress: ethNetwork?.walletV4ForwarderFactoryAddress as string,
```
mullapudipruthvik: walletV4ForwarderFactoryAddress is not defined for eth like coins

danielzhao122: So we should first check if `walletV4ForwarderFactoryAddress` is defined for the coin and if not fall back on `forwarderFactoryAddress`?

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
danielzhao122: walletVersion 2 does not use the v4 factories.

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
danielzhao122: this is only for receive address

zahin-mohammad: What happens if the base address is passed into this function?

danielzhao122: It will probably fail, should this covered in the function?

danielzhao122: This is probably something we want. will add coverage for this 👍

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2766,178 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+        };
+      case 4:
+      case 5:
+        if (!ethNetwork?.walletV4ForwarderFactoryAddress || !ethNetwork?.walletV4ForwarderImplementationAddress) {
+          throw new Error(`Wallet v${walletVersion} factory addresses not configured for this network`);
+        }
+        return {
+          walletFactoryAddress: ethNetwork.walletV4ForwarderFactoryAddress,
+          walletImplementationAddress: ethNetwork.walletV4ForwarderImplementationAddress,
+        };
```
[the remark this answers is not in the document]
danielzhao122: No wallet version 4 and 5 both use the V4 address

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2766,178 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+  ): { expectedAddress: string; actualAddress: string } {
+    const { address, coinSpecific, baseAddress } = params;
+
+    const { forwarderFactoryAddress, forwarderImplementationAddress } =
+      this.getForwarderFactoryAddressesAndForwarderImplementationAddress(params.walletVersion);
+    const initcode = getProxyInitcode(forwarderImplementationAddress);
+    const saltBuffer = this.createSaltBuffer(coinSpecific.salt || '');
+
+    const { createForwarderParams, createForwarderTypes } =
+      forwarderVersion === 4
```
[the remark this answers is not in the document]
danielzhao122: Only forward version 4 requires the feeAddress, this is a typo in the PR description.

abhishekagrawal080 (dismissed):
looks good to me

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -401,7 +405,44 @@ export interface EthConsolidationRecoveryOptions {
...
+export interface KeychainWithEthAddress {
+  ethAddress: string;
+  pub: string;
+}
+
+/**
+ * BIP32 wallet base address verification options
+ * Supports V1, V2, and V4 wallets that use ethAddress-based derivation
+ */
+export interface VerifyBip32BaseAddressOptions extends VerifyEthAddressOptions {
```
dpkjnr: why bip32? these are contract based addresses

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2766,178 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+   * @param {number | undefined} walletVersion - The wallet version
+   * @returns {object} Factory and implementation addresses for forwarders
+   */
+  getForwarderFactoryAddressesAndForwarderImplementationAddress(walletVersion: number | undefined): {
+    forwarderFactoryAddress: string;
+    forwarderImplementationAddress: string;
+  } {
+    const ethNetwork = this.getNetwork();
+
+    switch (walletVersion) {
```
dpkjnr: what about walletVersion 1? and I would suggest to take forwarderVersion as a param instead of walletVersion

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2766,178 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+  /**
+   * Get wallet base address factory and implementation addresses.
+   * This is used for base address verification for V1, V2, V4, and V5 wallets.
+   * The base address is the main wallet contract deployed via CREATE2.
+   *
+   * @param {number} walletVersion - The wallet version (1, 2, 4, or 5)
+   * @returns {object} Factory and implementation addresses for the wallet base address
+   * @throws {Error} if wallet version addresses are not configured
+   */
+  getWalletBaseAddressFactoryAddressesAndImplementationAddress(walletVersion: number): {
```
dpkjnr: it can simply be getWalletFactoryAddressesAndImplementationAddress, similar naming is followed most of the places

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
dpkjnr: just saw that default branch is taking care of walletVersion 1, better keep it explicit case instead of default, that looks more readable to me

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2766,178 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+
+  /**
+   * Verify BIP32 wallet base address (V1, V2, V4).
+   * These wallets use a wallet factory to deploy base addresses with CREATE2.
+   * The address is derived from the keychains' ethAddresses and a salt.
+   *
+   * @param {VerifyBip32BaseAddressOptions} params - Verification parameters
+   * @returns {object} Expected and actual addresses for comparison
+   */
+  private verifyBip32BaseAddress(params: VerifyBip32BaseAddressOptions): {
```
dpkjnr: name is misleading, it says verify but we are just returning actual and expected addresses

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
danielzhao122: I think this is bad naming on my part.
For V1, V2, V4 we use the `ethAddress`'s and `CREATE2` to derive `baseAddress`

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2736,19 +2949,46 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+    // TSS/MPC wallet address verification (V3, V5, V6)
+    // V5 base addresses use TSS, but V5 forwarders use the regular forwarder verification
+    const isTssWalletVersion = params.walletVersion === 3 || params.walletVersion === 5 || params.walletVersion === 6;
+    const shouldUseTssVerification =
+      isTssVerifyAddressOptions(params) && isTssWalletVersion && (params.walletVersion !== 5 || isVerifyingBaseAddress);
+
+    if (shouldUseTssVerification) {
+      if (isVerifyingBaseAddress) {
+        const index = typeof params.index === 'string' ? parseInt(params.index, 10) : params.index;
+        if (index !== 0) {
```
dpkjnr, line 2986: make sure this is valid for existing data and will it be called for custodial wallets also?

danielzhao122: No custodial wallets are out of scope for this pr

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2766,182 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+          forwarderImplementationAddress: ethNetwork.walletV4ForwarderImplementationAddress,
+        };
+      default:
+        if (!ethNetwork?.forwarderFactoryAddress || !ethNetwork?.forwarderImplementationAddress) {
+          throw new Error('Forwarder factory addresses not configured for this network');
+        }
+        return {
+          forwarderFactoryAddress: ethNetwork.forwarderFactoryAddress,
+          forwarderImplementationAddress: ethNetwork.forwarderImplementationAddress,
+        };
```
zahin-mohammad: Why is v1 not handled explicitly as the first case? If we add a 6th forwarder version in the future, the reason this fails will not be very straightforward.

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2725,6 +2766,182 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
+          walletImplementationAddress: ethNetwork.walletV4ForwarderImplementationAddress,
+        };
+      default:
+        if (!ethNetwork?.walletFactoryAddress || !ethNetwork?.walletImplementationAddress) {
+          throw new Error('Wallet v1 factory addresses not configured for this network');
+        }
+        return {
+          walletFactoryAddress: ethNetwork.walletFactoryAddress,
+          walletImplementationAddress: ethNetwork.walletImplementationAddress,
+        };
```
zahin-mohammad: The default case should throw an error. V1 should be handled explicitly.

zahin-mohammad (dismissed):
I think in general this lgtm, we can address some of the comments as a follow up.

zahin-mohammad (approved):

jamesmili (approved):

PR #7580, opened 2025-11-20
feat: add tx explanation for TxIntentMismatch errors

danielpeng1:
Adding tx explanation for TxIntentMismatch errors for UTXO and ETH-like coins.

- Implemented `getTxExplanation` method in both `AbstractEthLikeNewCoins` and `AbstractUtxoCoin` to generate detailed transaction explanations for error reporting.
  - Created `txExplanation.ts` to use in `abstractUtxoCoin.ts` and both `verifyTransaction`s
- Updated error handling in tx verification methods to include tx explanations in `TxIntentMismatchError` and its subclasses.
- Added unit tests to validate the inclusion of tx explanations in error instances.

Ticket: WP-6608


modules/abstract-utxo/src/transaction/txExplanation.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (!txPrebuild.txHex) {
+    return undefined;
+  }
+
+  try {
+    const explanation = await coin.explainTransaction({
+      txHex: txPrebuild.txHex,
+      txInfo: txPrebuild.txInfo,
+    });
+    return JSON.stringify(explanation, null, 2);
```
OttoAllmendinger: this will not work with bigint values

modules/abstract-utxo/src/abstractUtxoCoin.ts
```diff
@@ -616,11 +618,19 @@ export abstract class AbstractUtxoCoin extends BaseCoin {
   async verifyTransaction<TNumber extends number | bigint = number>(
     params: VerifyTransactionOptions<TNumber>
   ): Promise<boolean> {
+    const txExplanation = await getTxExplanation(this, params.txPrebuild);
```
OttoAllmendinger: this is pretty wasteful if we only use it in the `catch` block - move it there

modules/abstract-utxo/src/transaction/descriptor/verifyTransaction.ts
```diff
@@ -75,13 +76,16 @@ export async function verifyTransaction<TNumber extends number | bigint>(
   params: VerifyTransactionOptions<TNumber>,
   descriptorMap: DescriptorMap
 ): Promise<boolean> {
+  const txExplanation = await getTxExplanation(coin, params.txPrebuild);
```
OttoAllmendinger: same

OttoAllmendinger (changes requested):

modules/abstract-utxo/src/transaction/txExplanation.ts
danielpeng1: will do a bigint-safe JSON serialization. Can I create a test file modules/abstract-utxo/test/unit/txExplanation.ts to test this

OttoAllmendinger: why do we need to a return a `string` here anyway? we already have `src/transaction/explainTransaction.ts` already as well, so why are we adding another file here?

why can't the caller of this func (whoever that is) deal with the string conversion?

modules/sdk-core/src/bitgo/errors.ts
```diff
@@ -256,11 +256,13 @@ export interface ContractDataPayload {
  * @property {string | IRequestTracer | undefined} id - Transaction ID or request tracer for tracking
  * @property {TransactionParams[]} txParams - Array of transaction parameters that were analyzed
  * @property {string | undefined} txHex - The raw transaction in hexadecimal format
+ * @property {string | undefined} txExplanation - Stringified transaction explanation
```
OttoAllmendinger, line 260: better make this `unknown`

modules/sdk-core/src/bitgo/errors.ts
```diff
@@ -269,17 +271,20 @@ export class TxIntentMismatchError extends BitGoJsError {
...
    * @param {string | undefined} txHex - Raw transaction hex string
+   * @param {string | undefined} txExplanation - Stringified transaction explanation
    */
   public constructor(
     message: string,
     id: string | IRequestTracer | undefined,
     txParams: TransactionParams[],
-    txHex: string | undefined
+    txHex: string | undefined,
+    txExplanation?: string
```
OttoAllmendinger: `unknown` here

modules/sdk-core/src/bitgo/errors.ts
```diff
@@ -269,17 +271,20 @@ export class TxIntentMismatchError extends BitGoJsError {
...
   ) {
     super(message);
     this.id = id;
     this.txParams = txParams;
     this.txHex = txHex;
+    this.txExplanation = txExplanation;
```
OttoAllmendinger: do some sort of safe stringification here (`JSON.stringify` with bigint substitution)

OttoAllmendinger (changes requested):
better deal with the string explanation -> conversion in the sdk-core `errors`

modules/abstract-utxo/src/transaction/txExplanation.ts
danielpeng1: Is the alternative to return the object (`explanation`)? Then the parent error class `TxIntentMismatchError`'s constructor would handle the `JSON.stringify` for example?

regardless I can move this function to `src/transaction/explainTransaction.ts`

OttoAllmendinger: >Is the alternative to return the object (explanation)? Then the parent error class TxIntentMismatchError's constructor would handle the JSON.stringify for example?

yes

> regardless I can move this function to src/transaction/explainTransaction.ts


I don't think we need a new function

danielpeng1: but then we would need to duplicate/in-line this try catch logic in each call site? In `descriptor/verifyTransaction.ts`, `fixedScript/verifyTransaction.ts`, `abstractUtxoCoin.ts`

modules/sdk-core/src/bitgo/errors.ts
```diff
@@ -256,11 +256,13 @@ export interface ContractDataPayload {
  * @property {string | IRequestTracer | undefined} id - Transaction ID or request tracer for tracking
  * @property {TransactionParams[]} txParams - Array of transaction parameters that were analyzed
  * @property {string | undefined} txHex - The raw transaction in hexadecimal format
+ * @property {string | undefined} txExplanation - Stringified transaction explanation
  */
 export class TxIntentMismatchError extends BitGoJsError {
```
OttoAllmendinger, line 262: add a static function here 

```
static tryGetTxExplanation(coin: AbstractCoin, tx): Promise<string | undefined> {
   try {
      return bigintSafeStringify(coin.getExplanation(tx));
   } catch (e) {
     return String(e); // with tx etc
   }
 }
```

modules/abstract-utxo/src/transaction/txExplanation.ts
OttoAllmendinger: how about https://github.com/BitGo/BitGoJS/pull/7580/files/418878dd16a114c8dae0b233019abc5ad4bdf41f#r2546538479

danielpeng1: Isn't this is returning a string again though? (instead of the `explanation` object)

OttoAllmendinger: yes but at least it is relegated to the error class, so I'm not as concerned about the interface

you can implement it returning `unknown` or `string | undefined`, your call

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2811,9 +2839,18 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
   async verifyTssTransaction(params: VerifyEthTransactionOptions): Promise<boolean> {
     const { txParams, txPrebuild, wallet } = params;
 
+    const txExplanation = await this.getTxExplanation(txPrebuild);
```
mullapudipruthvik: can we have this inside `throwRecipientMismatch` function ?
so that its only called when required

mullapudipruthvik: and probably have the string conversion inside, instead of a new function

modules/abstract-utxo/src/transaction/index.ts
```diff
@@ -3,5 +3,6 @@ export * from './recipient';
 export { explainTx } from './explainTransaction';
 export { parseTransaction } from './parseTransaction';
 export { verifyTransaction } from './verifyTransaction';
+export { getTxExplanation } from './txExplanation';
```
mohammadalfaiyazbitgo: this would fail lint no?

modules/sdk-core/src/bitgo/errors.ts
mohammadalfaiyazbitgo: @danielpeng1 let's go with Otto's suggestion, but not as a static function on the `error` class. Unfortunately there's too many discrepancies between coins on how to generate a `transaction explanation`. So let's create a helper function to generate the error and the explanation, that way we can encapsulate the coin specific logic of creating the explanation outside of the class, so in cases where the code is already generating the explanation we can pass it directly to the error class. 

Below is a general overview for reviewer's context
```ts
class GenerateTxIntentErrorFactory() { 
 private static tryGetTxExplanation({coin: string, txExplanationParams: unknown} | {coin: 'btc', txExplanationParams: UtxoExplainTransactionOptions| | {coin: 'eth', txExplanationParams: AbstractEthExplanationParams...etc) {
 ... return coin.explainTransaction(...txExplanationParams);


createTxIntentError(....): {
  let explanation;
switch(coin.getFamily()) {
 explanation = coin.getExplanation(...)
 default: explanation = `unable to generate explanation for coin: ${coin}`
}
return TxIntentError(..., explanation)
}

```

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
danielpeng1: If I do this `throwRecipientMismatch` is now `async` with `getTxExplanation` inside it, and all calls to `throwRecipientMismatch` in `verifyTssTransaction` and `verifyTransaction` now use `await`, is that okay?

modules/sdk-core/src/bitgo/errors.ts
```diff
@@ -269,17 +272,58 @@ export class TxIntentMismatchError extends BitGoJsError {
...
     this.txHex = txHex;
+    this.txExplanation = txExplanation ? this.safeStringify(txExplanation) : undefined;
+  }
+
+  /**
+   * Safely stringify a value with BigInt support
+   * @param value - Value to stringify
+   * @returns JSON string with BigInts converted to strings
+   */
+  private safeStringify(value: unknown): string {
```
OttoAllmendinger: could be a static method as well

OttoAllmendinger (dismissed):

modules/sdk-core/src/bitgo/errors.ts
danielpeng1: thanks, rebased and did this

OttoAllmendinger (approved):

modules/sdk-core/src/bitgo/errors.ts
danielpeng1: this would get a bit ugly with each new coin. I'm leveraging the `IBaseCoin` interface to try getting the `txExplanation` typed with `unknown` instead.

mohammadalfaiyazbitgo (approved):

alia-bitgo (approved):

baltiyal (approved):

PR #7597, opened 2025-11-22
fix(sdk-coin-vet): preserve precision for large value transactions

kamleshmugdiya:
Ticket: COIN-0000

This PR fixes a critical issue in VET transaction processing where large numeric values lost precision due to JavaScript's Number type limitations.
When handling transactions with values exceeding ~15-17 significant digits, the values were converted to scientific notation (e.g., `2.332734928448876e+22` instead of `23327349284488761096413`).

The issue was in Transaction.fromDeserializedSignedTransaction() which used Number(clause.value).toString(), converting large values to scientific notation. This caused transaction output parsing to fail when comparing the expected amount with the actual amount.

  The fix:
  - Replaces `Number(clause.value).toString()` with `new BigNumber(clause.value).toFixed()`.
  - Ensures exact precision is maintained throughout transaction processing.
  - Adds test to verify precision handling of large numbers.

This change ensures accurate value representation in VET transactions, particularly important for exchanges and applications handling large value transfers.

### DIFF
--- a/modules/sdk-coin-vet/src/lib/transaction/transaction.ts
+++ b/modules/sdk-coin-vet/src/lib/transaction/transaction.ts
@@ -292,7 +292,7 @@ export class Transaction extends BaseTransaction {
       // Set recipients from clauses
       this.recipients = body.clauses.map((clause) => ({
         address: (clause.to || '0x0').toString().toLowerCase(),
-        amount: Number(clause.value).toString(),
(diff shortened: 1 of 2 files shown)


prithvishet2503 (approved):
LGTM

Vijay-Jagannathan (approved):
good catch, we should never be using Number for amount

PR #7598, opened 2025-11-24
fix: explorer url of sonic testnet

ArunBala-Bitgo:
Ticket: WIN-7690


prithvishet2503 (changes requested):
https://docs.soniclabs.com/sonic/build-on-sonic/getting-started#sonic-testnet
blaze testnet is depricated 
we are migrating it

PR #7638, opened 2025-11-26
Wp 6461 fix address verification during creation

danielzhao122:
Was not passing in wallet address to verification, this caused the wallet to always assume that the wallet was onchain and thus use the wrong verification method.

TICKET: WP-6461


pranavjain97 (dismissed):

Vijay-Jagannathan (approved):
LGTM

PR #7647, opened 2025-11-27
feat(sdk-coin-flrp): add recoverySignature method for public key recovery from signature

mohd-kashif:
Ticket: WIN-8068


parasgarg-bitgo (approved):
LGTM

PR #7674, opened 2025-12-02
feat: address verification for vet

danielzhao122:
Address verification for `vet`
We use TSS verification since only v6 wallets are supported for `vet`

TICKET: WP-7050

### DIFF
--- a/modules/sdk-coin-vet/src/vet.ts
+++ b/modules/sdk-coin-vet/src/vet.ts
@@ -20,14 +20,17 @@ import {
   SignedTransaction,
   SignTransactionOptions,
   TokenTransferRecipientParams,
+  TssVerifyAddressOptions,
   VerifyAddressOptions,
   VerifyTransactionOptions,
+  isTssVerifyAddressOptions,
   TokenType,
   Ecdsa,
   ECDSAUtils,
   Environments,
   BaseBroadcastTransactionOptions,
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-vet/src/vet.ts
```diff
@@ -153,12 +162,50 @@ export class Vet extends BaseCoin {
...
-    const { address: newAddress } = params;
+  /**
+   * Verify that an address belongs to this wallet.
+   *
+   * @param {VerifyAddressOptions | TssVerifyAddressOptions} params - Verification parameters
+   * @returns {Promise<boolean>} True if address belongs to wallet
+   * @throws {InvalidAddressError} If address format is invalid
+   * @throws {Error} If invalid wallet version or missing parameters
+   */
+  async isWalletAddress(params: VerifyAddressOptions | TssVerifyAddressOptions): Promise<boolean> {
```
sachinBitgo: Could the **TssVerifyAddressOptions** interface be merged into the existing **VerifyAddressOptions** interface? This would allow us to have a single, common interface for address verification options.

If consolidation isn't feasible, I think params should be of **TssVerifyVetAddressOptions** as this interface extending **TssVerifyAddressOptions**.

sachinBitgo (approved):
LGTM

PR #7686, opened 2025-12-03
feat: address verification for icp

danielzhao122:
Address verification for icp
unit tests for wallet v1 and v2

TICKET: WP-7080


modules/sdk-coin-icp/src/icp.ts
```diff
@@ -141,8 +143,81 @@ export class Icp extends BaseCoin {
...
+   */
+  async isWalletAddress(params: TssVerifyIcpAddressOptions): Promise<boolean> {
+    const { address, rootAddress, walletVersion } = params;
+
+    if (!this.isValidAddress(address)) {
+      throw new InvalidAddressError(`invalid address: ${address}`);
+    }
+
+    if (walletVersion === 1) {
+      return this.verifyMemoBasedAddress(address, rootAddress);
```
dpkjnr: even in case of memo based address, root address still needs to be verified against they common keychain

modules/sdk-coin-icp/src/icp.ts
```diff
@@ -141,8 +144,62 @@ export class Icp extends BaseCoin {
...
+      if (!extractedRootAddress || extractedRootAddress === address) {
+        throw new Error('memoId is required for wallet version 1 addresses');
+      }
+      if (extractedRootAddress.toLowerCase() !== rootAddress.toLowerCase()) {
+        throw new UnexpectedAddressError(
+          `address validation failure: expected ${rootAddress} but got ${extractedRootAddress}`
+        );
+      }
+      addressToVerify = rootAddress;
+    } else if (rootAddress && address.toLowerCase() === rootAddress.toLowerCase() && parsedIndex !== 0) {
```
dpkjnr: do we really need this branch? verifyMPCWalletAddress will anyway fail if index is not correct

dpkjnr (approved):

PR #7701, opened 2025-12-05
fix(express): support legacy EIP1559 transaction in type valid…

baltiyal:
TICKET: WIN-8042

### DIFF
--- a/modules/express/src/typedRoutes/api/v2/coinSignTx.ts
+++ b/modules/express/src/typedRoutes/api/v2/coinSignTx.ts
@@ -14,13 +14,19 @@ export const CoinSignTxParams = {
 /**
  * EIP1559 transaction parameters for Ethereum
  * Reference: modules/abstract-eth/src/abstractEthLikeNewCoins.ts:116-119
- * Note: Both fields are REQUIRED when EIP1559 object is provided
+ *
+ * Note: Changed to t.partial() to support multiple use cases:
+ * 1. Full EIP1559 transactions: { maxFeePerGas, maxPriorityFeePerGas }
+ * 2. Legacy token transactions: { isEip1559: false } (SDK marker for non-EIP1559)
+ * 3. Legacy base coin transactions: undefined (field omitted entirely)
  */
-export const EIP1559 = t.type({
-  /** Maximum priority fee per gas (REQUIRED) */
+export const EIP1559 = t.partial({
+  /** Maximum priority fee per gas */
   maxPriorityFeePerGas: t.union([t.string, t.number]),
-  /** Maximum fee per gas (REQUIRED) */
+  /** Maximum fee per gas */
   maxFeePerGas: t.union([t.string, t.number]),
+  /** Flag indicating whether transaction uses EIP1559 fee model */
+  isEip1559: t.boolean,
 });
 
 /**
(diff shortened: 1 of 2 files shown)


modules/express/src/typedRoutes/api/v2/coinSignTx.ts
```diff
@@ -14,13 +14,19 @@ export const CoinSignTxParams = {
...
- * Note: Both fields are REQUIRED when EIP1559 object is provided
+ *
+ * Note: Changed to t.partial() to support multiple use cases:
+ * 1. Full EIP1559 transactions: { maxFeePerGas, maxPriorityFeePerGas }
+ * 2. Legacy token transactions: { isEip1559: false } (SDK marker for non-EIP1559)
+ * 3. Legacy base coin transactions: undefined (field omitted entirely)
  */
-export const EIP1559 = t.type({
-  /** Maximum priority fee per gas (REQUIRED) */
+export const EIP1559 = t.partial({
```
lokesh-bitgo, line 23: Could you please share the code reference for this?

I have taken the reference from here - https://github.com/BitGo/BitGoJS/blob/master/modules/abstract-eth/src/abstractEthLikeNewCoins.ts#L119

lokesh-bitgo, line 23: For AVAXC, I have found code reference here - https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-coin-avaxc/src/iface.ts#L48

But I can't find the fields are optional here and also the additional field `isEip1559: t.boolean` I am not able to find it. Could you please share the code reference for your changes ?

baltiyal: This was how we send **EIP1559 fee** for **Avaxc Tokens**, nothing is changed here. Before making a hard type check here, we should ensure that the **EIP1559** params passed by WP build response should match the type check.
-  https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/v2/coins/tokens/avaxcErc20Token.ts#L241
- https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/v2/coins/tokens/avaxcErc20Token.ts#L440

lokesh-bitgo: okay got it

lokesh-bitgo (approved):
Had a discussion with @baltiyal. The type codec for EIP1559 defined in the current code is correct and aligned with the [AVAXC](https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-coin-avaxc/src/iface.ts#L48) type codec.

However, the AVAXC token behavior is slightly different — after the transaction is built at [BGMS](https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/v2/coins/tokens/avaxcErc20Token.ts#L241), it sends a **new field isEip1559 inside the EIP1559 type and does not pass the other two fields**. This is why it was not working for AVAXC tokens.

Since we have introduced type codecs for all Express endpoints, any coin specific request/response changes made on the BGMS side need to be reflected on the Express side as well to avoid impacts to clients.

cc: @zahin-mohammad  @ranga-r2

modules/express/src/typedRoutes/api/v2/coinSignTx.ts
baltiyal: As discussed @lokesh-bitgo, sharing the txPrebuild.eip1559 behavior from local debugging log data for wallet 6618f448f6c53303d38130bc60e9efe6 :

Token (AVAXC:USDC) `prebuild → txPrebuild.eip1559 = { "isEip1559": false }`

Coin  (AVAX) `prebuild → txPrebuild.eip1559 = undefined`

Vijay-Jagannathan (approved):
LGTM

PR #7707, opened 2025-12-08
feat: ERC721 token support for EVM coins

rohitsaw115:
ticket: win-8154

Changes:-
- Optimization for erc721 token standard.
- With new changes, only feature flag is required to support erc721 standard for new coins.
- erc721 token test token.


ArunBala-Bitgo (dismissed):
Could we add test cov for the new changes

modules/statics/src/tokenConfig.ts
```diff
@@ -1088,6 +1095,43 @@ export const getFormattedEthLikeTokenConfig = (customCoinMap = coins): EthLikeTo
...
+    }
+    return acc;
+  }, []);
+
+const getEthLikeErc721Tokens = (network: 'Mainnet' | 'Testnet'): EthLikeTokenMap => {
+  const networkTokens = getFormattedEthLikeERC721Tokens().filter((token) => token.network === network);
+  const ethLikeTokenMap = {} as EthLikeTokenMap;
+
+  coins.forEach((coin) => {
+    if (coin instanceof AccountCoin && coin.features.includes(CoinFeature.SUPPORTS_ERC721)) {
```
ArunBala-Bitgo: Can we pass this as a param to the getEthLikeConfig?

ArunBala-Bitgo (dismissed):
Looks good to me, Please make sure that the test token onboarded works in lower environments before releasing in prod

rohitsaw115:
@claude please review this

modules/sdk-coin-evm/src/register.ts
```diff
@@ -35,5 +36,17 @@ export const register = (coinFamily: string, sdk: BitGoBase): void => {
           sdk.register(name, coinConstructor);
         });
       }
+
+      // Handle SUPPORTS_ERC721 registration
+      if (coinFeatures.includes(CoinFeature.SUPPORTS_ERC721)) {
+        const coinNames = {
+          Mainnet: `${coin.name}`,
+          Testnet: `t${coin.name}`,
```
ArunBala-Bitgo: Note: if there are exception from this standard (tcoin) we need to add a seperate logic to handle that

modules/statics/test/unit/coins.ts
```diff
@@ -1299,4 +1299,30 @@ describe('create token map using config details', () => {
       }
     }
   });
+
+  it('should create ERC721 tokens for all coins supporting ERC721 using createToken', () => {
+    // Get all ERC721 token configs from allCoinsAndTokens that support ERC721
+    const erc721TokenConfigs = allCoinsAndTokens
```
ArunBala-Bitgo, line 1305: Can we onboard a testnet token on this PR to verify this testcase, because right now since we have no erc721 tokens this would be empty

rohitsaw115: We have testnet token (terc721:hbarevmtoken) to verify this test case.

modules/sdk-coin-evm/src/register.ts
rohitsaw115: Change registration similar to erc20 where mainnet and testnet user same coin.name.

ArunBala-Bitgo (approved):

ArunBala-Bitgo (approved):

PR #7712, opened 2025-12-08
feat(sdk-coin-near): address verification for near

danielzhao122:
Implement `isWalletAddress` for `near` to cryptographically verify that an address belongs to a TSS/MPC wallet using `HD verification`

### DIFF
--- a/modules/sdk-coin-near/src/near.ts
+++ b/modules/sdk-coin-near/src/near.ts
@@ -1004,17 +1004,6 @@ export class Near extends BaseCoin {
    * @throws {Error} If invalid parameters or root address verification with wrong index
    */
   async isWalletAddress(params: TssVerifyNearAddressOptions): Promise<boolean> {
-    const { address, rootAddress } = params;
-
-    // Root address verification requires index 0
-    const isVerifyingRootAddress = rootAddress && address === rootAddress;
-    if (isVerifyingRootAddress) {
-      const index = typeof params.index === 'string' ? parseInt(params.index, 10) : params.index;
-      if (index !== 0) {
-        throw new Error(`Root address verification requires index 0, but got index ${params.index}.`);
-      }
-    }
-
     const result = await verifyEddsaTssWalletAddress(
       params,
       (address) => this.isValidAddress(address),
@@ -1022,7 +1011,7 @@ export class Near extends BaseCoin {
     );
 
     if (!result) {
-      throw new UnexpectedAddressError(`address validation failure: address ${params.address} is not a wallet address`);
+      throw new UnexpectedAddressError(`address validation failure`);
     }
 
     return true;
--- a/modules/sdk-coin-near/src/near.ts
+++ b/modules/sdk-coin-near/src/near.ts
@@ -18,8 +18,8 @@ import {
   EDDSAMethods,
   EDDSAMethodTypes,
   Environments,
+  InvalidAddressError,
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-near/src/near.ts
```diff
@@ -984,8 +995,41 @@ export class Near extends BaseCoin {
...
+  /**
+   * Verify if an address belongs to a NEAR wallet using EdDSA TSS MPC derivation.
+   * For NEAR, the address is the public key directly (implicit accounts).
+   *
+   * @param {TssVerifyAddressOptions} params - Verification parameters
+   * @returns {Promise<boolean>} True if address belongs to wallet
+   * @throws {InvalidAddressError} If address format is invalid or doesn't match derived address
+   * @throws {Error} If invalid parameters
+   */
+  async isWalletAddress(params: TssVerifyNearAddressOptions): Promise<boolean> {
```
dpkjnr, line 1006: can it be done similar to solana?

```
  async isWalletAddress(params: TssVerifyAddressOptions): Promise<boolean> {
    return verifyEddsaTssWalletAddress(
      params,
      (address) => this.isValidAddress(address),
      (publicKey) => this.getAddressFromPublicKey(publicKey)
    );
  }
```

modules/sdk-coin-near/src/near.ts
```diff
@@ -984,8 +994,38 @@ export class Near extends BaseCoin {
...
+
+    // Root address verification requires index 0
+    const isVerifyingRootAddress = rootAddress && address === rootAddress;
+    if (isVerifyingRootAddress) {
+      const index = typeof params.index === 'string' ? parseInt(params.index, 10) : params.index;
+      if (index !== 0) {
+        throw new Error(`Root address verification requires index 0, but got index ${params.index}.`);
+      }
+    }
+
```
dpkjnr: lets avoid these, keep implementation very simple, similar to Sol

modules/sdk-coin-near/src/near.ts
```diff
@@ -984,8 +994,38 @@ export class Near extends BaseCoin {
...
+      params,
+      (address) => this.isValidAddress(address),
+      (publicKey) => publicKey
+    );
+
+    if (!result) {
+      throw new UnexpectedAddressError(`address validation failure: address ${params.address} is not a wallet address`);
+    }
+
+    return true;
```
dpkjnr, line 1017: this too

danielzhao122: I think we want to throw a error if the validation fails.
That is what I have been doing with other coins as well.
I will update `sol` to do this as well

dpkjnr: A function returning boolean should return true or false rather than throwing an error for expected validation failures. Throwing here forces callers to wrap every call in try-catch instead of using a simple if condition:

```
// Current (awkward)
try {
  await coin.isWalletAddress(params);
} catch { /* invalid */ }

// Expected (clean)
if (await coin.isWalletAddress(params)) { /* valid */ }
```

Errors should be reserved for truly unexpected scenarios (internal failures, malformed state), not for "address is invalid"—which is an expected outcome for a validation function.
I can approve this PR, but let's standardize this behavior across all EdDSA coins (can be done in a follow-up PR).

dpkjnr (approved):

PR #7714, opened 2025-12-09
feat(statics): onboard eth:sofid token

damodarnaik699:
Ticket: SCAAS-2020

### DIFF
--- a/modules/statics/src/coins/erc20Coins.ts
+++ b/modules/statics/src/coins/erc20Coins.ts
@@ -10945,6 +10945,15 @@ export const erc20Coins = [
     [...ACCOUNT_COIN_DEFAULT_FEATURES, CoinFeature.STABLECOIN]
   ),
   erc20(
+    '4400af90-30d8-43fb-98b6-ca6791c94aa2',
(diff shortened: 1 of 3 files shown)


sachushaji (approved):

modules/statics/src/coins/erc20Coins.ts
```diff
@@ -12601,7 +12610,7 @@ export const erc20Coins = [
   terc20(
     'a0660bd4-c508-4704-ac2b-c7ead1f6ad43',
     'hteth:stgsofid',
-    'Test SoFi USD',
+    'Test SoFiUSD',
```
raj-bitgo, line 12613: should this be named staging?

raj-bitgo (approved):

modules/statics/src/coins/erc20Coins.ts
damodarnaik699: we kept as Test only for both test and staging so it should be fine

Phani024 (approved):
LGTM

vipulchaursiya908:
LGTM

PR #7723, opened 2025-12-10
fix(sdk-coin-vet): add prefix for validator address

Gautam2305:
Ticket: SC-4444


modules/sdk-coin-vet/test/resources/vet.ts
```diff
@@ -215,3 +215,8 @@ export const SERIALIZED_SIGNED_FLUSH_TOKEN_TX =
 
 export const VALID_NFT_CONTRACT_DATA =
   '0x23b872dd0000000000000000000000007ca00e3bc8a836026c2917c6c7c6d049e52099dd000000000000000000000000e59f1cea4e0fef511e3d0f4eec44adf19c4cbeec00000000000000000000000000000000000000000000000000000000000004d2';
+
+export const DELEGATE_CLAUSE_DATA =
+  '0x08bbb8240000000000000000000000000000000000000000000000000000000000003d45000000000000000000000000ae99cb89767a09d53e589a40cb4016974aba4b94';
+export const DELEGATE_TOKEN_ID = '15685';
+export const DELEGATE_VALIDATOR = '0xae99cb89767a09d53e589a40cb4016974aba4b94';
```
Vijay-Jagannathan, line 222: what if our decoder is not working correctly and we're expecting the wrong address in line 222; additionally, we have to make sure the signed hex origin also has the sender address correctly

instead let's build from raw hex (raw hex will be hardcoded and we can take this from sandbox ig) in the last test of the delegate txn builder (this should have some validator address etc), and we can assert that this raw hex has so and so sender, validator address, etc

Gautam2305: done

abhijit0943 (approved):
lgtm

Doddanna17 (approved):
lgtm

raksha-r7 (approved):

PR #7744, opened 2025-12-15
feat: addres verification for trx

danielzhao122:
Address verification for `trx`

TICKET: WP-7083

### DIFF
--- a/modules/sdk-coin-trx/src/trx.ts
+++ b/modules/sdk-coin-trx/src/trx.ts
@@ -14,6 +14,7 @@ import {
   getIsKrsRecovery,
   getIsUnsignedSweep,
   KeyPair,
+  KeyIndices,
   MethodNotImplementedError,
   ParsedTransaction,
   ParseTransactionOptions,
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-trx/src/trx.ts
```diff
@@ -246,7 +266,74 @@ export class Trx extends BaseCoin {
...
+    }
+
+    const chain = Number(params.chain ?? params.coinSpecific?.chain ?? 0);
+
+    if (!this.isValidAddress(address)) {
+      throw new Error(`Invalid address: ${address}`);
+    }
+
+    const userPub = keychains && keychains.length > 0 ? keychains[0].pub : undefined;
+    const bitgoPub = keychains && keychains.length > 2 ? keychains[2].pub : undefined;
```
dpkjnr: Use already defined constants for key indices

```
import { KeyIndices } from '@bitgo/sdk-core';

const userPub = keychains && keychains.length > 0 ? keychains[KeyIndices.USER].pub : undefined;
const bitgoPub = keychains && keychains.length > 2 ? keychains[KeyIndices.BITGO].pub : undefined;
```

Also, it would be better if userPub and bitgoPub are defined in the code blocks where validation is happening, instead of assigning undefined if not needed

dpkjnr (approved):

PR #7750, opened 2025-12-16
fix: test cases for parseWithdrawPsbt

parvbitgo067:
TICKET: BTC-2545

### DIFF
--- a/modules/abstract-lightning/test/unit/lightning/createPsbt.ts
+++ b/modules/abstract-lightning/test/unit/lightning/createPsbt.ts
@@ -1,0 +1,175 @@
+import * as utxolib from '@bitgo/utxo-lib';
+
+export interface PsbtCreationOptions {
+  network: utxolib.Network;
+  inputValue?: number;
+  outputValue?: number;
+  outputAddress?: string;
+  changeValue?: number;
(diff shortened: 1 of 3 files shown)


modules/abstract-lightning/test/unit/lightning/createPsbt.ts
```diff
@@ -0,0 +1,176 @@
...
+  changePurpose: number;
+}
+
+/**
+ * Creates a random PSBT for testing purposes with customizable options.
+ * This helper function generates a PSBT with a fake input and configurable outputs.
+ *
+ * @param options - Configuration options for PSBT creation
+ * @returns A constructed PSBT instance and the master key used
+ */
```
OttoAllmendinger, line 28: we don't really want "random", it can lead to flaky tests

that being said, you are not using a lot of randomness anyway (just the keys)

rename to `createTestPsbt` and use fixed keys instead

you can get an ECPair instance with utxolib.bip32.fromSeed(...).privateKey

this makes the output deterministic which use good for testing

OttoAllmendinger (changes requested):

OttoAllmendinger (approved):

PR #7766, opened 2025-12-17
feat: PSBTify cross-chain recovery transactions

lcovar:
Replace legacy transaction builder with PSBT for cross-chain recovery, aligning with the pattern used in backupKeyRecovery.ts.

Changes:
- Use UtxoPsbt for both signed and unsigned cross-chain recovery
- Add signAndVerifyPsbt for signing with user key
- Make txInfo optional in response interfaces (PSBT is self-contained)
- Remove unused legacy functions (getTxInfo, old createSweepTransaction)

Test updates:
- Fix nested it() blocks that were never executing
- Add 142 fixture files for cross-chain recovery combinations

Ticket: BTC-2890

TICKET: BTC-2890


modules/abstract-utxo/src/recovery/crossChainRecovery.ts
```diff
@@ -28,17 +27,17 @@ type FeeInfo = {
...
+  feeInfo?: FeeInfo;
   address: string;
   coin: string;
 }
 
 export interface CrossChainRecoverySigned<TNumber extends number | bigint = number> {
   version: 1 | 2;
   txHex: string;
-  txInfo: TransactionInfo<TNumber>;
+  txInfo?: TransactionInfo<TNumber>;
```
OttoAllmendinger, line 40: I think we might to fully get rid of these

OttoAllmendinger (approved):

PR #7780, opened 2025-12-19
feat(statics): updates contract address for gousd test and prod

amishpatel1994:
Ticket: SCAAS-2223

GoUSD is going to be using the newly deployed smart contract that is generalized for all stablecoins. This PR updates the contract address for test and production env GoUSD stablecoin.


Phani024 (commented):
This would create issues with the existing goUSD balances/transfers I guess, although it's just a few transfers. Do we need to check what needs to be done there?

sachushaji (approved):

PR #7781, opened 2025-12-19
fix: celo unstake methods collide with weth unwrap

derekchen745:
Ticket: SC-4630

### DIFF
--- a/modules/abstract-eth/src/lib/utils.ts
+++ b/modules/abstract-eth/src/lib/utils.ts
@@ -664,7 +664,7 @@ export function decodeFlushTokensData(data: string, to?: string): FlushTokensDat
  * @param {string} data The data to classify the transaction with
  * @returns {TransactionType} The classified transaction type
  */
-export function classifyTransaction(data: string): TransactionType {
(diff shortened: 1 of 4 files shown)


modules/abstract-eth/src/lib/utils.ts
```diff
@@ -664,22 +664,41 @@ export function decodeFlushTokensData(data: string, to?: string): FlushTokensDat
...
+
+  if (transactionType !== undefined) {
+    if (!isCeloStaking && CELO_STAKING_METHOD_IDS.has(methodId)) {
+      transactionType = TransactionType.ContractCall;
+    }
+  }
+
   if (transactionType === undefined) {
     transactionType = TransactionType.ContractCall;
   }
```
benkelcher-bitgo: this could be simplified to something like
```typescript
if (!isCeloStaking && CELO_STAKING_METHOD_IDS.has(methodId) || transactionType === undefined) {
  transactionType = TransactionType.ContractCall;
}
```

although ideally we shouldn't need CELO-specific logic here at all

PR #7785, opened 2025-12-22
feat(sol): add transaction size benchmark tool

nvrakesh06:
Ticket: WIN-8401

Add benchmark tool to empirically determine Solana transaction size limits for token transfers.

PR #7789, opened 2025-12-22
fix(sdk-coin-sol): add early validation for transaction size limits

nvrakesh06:
TICKET: WIN-8401

Add transaction size validation to prevent buffer overflow errors when sending to too many recipients


nvrakesh06:
Ref:
https://github.com/BitGo/BitGoJS/pull/7785
https://github.com/BitGo/BitGoJS/pull/7805

nvrakesh06:
@at31416  good catch, you’re right, I completely missed that case. If we start explicitly handling every edge case here, it’ll quickly get messy and hard to maintain.

I’m leaning towards letting the Solana SDK surface the error as-is (since it’s the source of truth). Do you think we should keep it that way, or wrap it in a higher-level BitGo error with a cleaner message/context?

PR #7795, opened 2025-12-23
fix: fix config for ofc erc20 tokens

ArunBala-Bitgo:
Ticket: WIN-8435

We should pass the correct address coin for ofc erc20 tokens 

Previously for all ofcerc20 tokens the default value`eth` was picked and used

export function ofcerc20(
  id: string,
  name: string,
  fullName: string,
  decimalPlaces: number,
  asset: UnderlyingAsset,
  kind: CoinKind = CoinKind.CRYPTO,
  features: CoinFeature[] = OfcCoin.DEFAULT_FEATURES,
  prefix = '',
  suffix: string = name.replace(/^ofc/, '').toUpperCase(),
  network: OfcNetwork = Networks.main.ofc,
  isToken = true,
  addressCoin = 'eth',  <--- **HERE WE ARE USING DEFAULT VALUE AS ETH** 
  primaryKeyCurve: KeyCurve = KeyCurve.Secp256k1
)

### DIFF
--- a/modules/bitgo/test/v2/unit/coins/ofcErc20Tokens.ts
+++ b/modules/bitgo/test/v2/unit/coins/ofcErc20Tokens.ts
@@ -1,0 +1,236 @@
+import 'should';
+import { TestBitGo } from '@bitgo/sdk-test';
+import { BitGo } from '../../../../src/bitgo';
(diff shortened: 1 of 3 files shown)


Venkat-Annavazzala (approved):
LGTM

PR #7798, opened 2025-12-23
feat(sdk-coin-ton): add ton whales withdrawal builder

feng-bitgo:
TICKET: SC-4504

Adds the withdrawal builder for TON Whales staking. There are no breaking changes as it only adds new functionality.


modules/sdk-coin-ton/src/lib/transaction.ts
```diff
@@ -138,6 +139,14 @@ export class Transaction extends BaseTransaction {
             payloadCell.bits.writeUint(parseInt(TON_WHALES_DEPOSIT_OPCODE, 10), 32);
             payloadCell.bits.writeUint(parseInt(queryId, 16), 64);
             payloadCell.bits.writeCoins(TonWeb.utils.toNano('1'));
+          } else if (payload.length > 26 && payload.substring(0, 10) === TON_WHALES_WITHDRAW_OPCODE) {
```
haritkapadia:
```suggestion
          } else if (payload.length >= 26 && payload.substring(0, 10) === TON_WHALES_WITHDRAW_OPCODE) {
```

Should it be `>=`?

modules/sdk-coin-ton/src/lib/transaction.ts
```diff
@@ -138,6 +139,14 @@ export class Transaction extends BaseTransaction {
...
+            const queryId = payload.substring(10, 26);
+            const amountStr = payload.substring(26);
+
+            payloadCell.bits.writeUint(parseInt(TON_WHALES_WITHDRAW_OPCODE, 10), 32);
+            payloadCell.bits.writeUint(new BN(queryId, 16), 64);
+            payloadCell.bits.writeCoins(TonWeb.utils.toNano('1')); // Hardcode 1 TON Gas
```
haritkapadia: > `// Hardcode 1 TON Gas`

Can you explain in a comment why we do this? Is it because protocol says so? Do we overwrite this later?

modules/sdk-core/src/account-lib/baseCoin/enum.ts
```diff
@@ -122,7 +122,7 @@ export enum TransactionType {
 
   // ton whales
   TonWhalesDeposit,
-  TonWhalesWithdraw,
+  TonWhalesWithdrawal,
```
haritkapadia, line 125: What happened here haha

haritkapadia (changes requested):
Overall looks good

modules/sdk-core/src/account-lib/baseCoin/enum.ts
feng-bitgo: I renamed it to be consistent with the file names but it wasn't used before so it's actually fine 👍

modules/sdk-coin-ton/src/lib/transaction.ts
feng-bitgo: It's just a gas limit that needs to be added to the message for TON Whales

feng-bitgo: but i added a comment

modules/sdk-coin-ton/src/lib/transaction.ts
feng-bitgo: sure

haritkapadia (approved):
Thanks!

PR #7804, opened 2025-12-25
feat: token onboarding correction

bhuvanr159:
TICKET: COIN-6893

Updated the token symbol from 'HYPE:hwHYPE' to 'HYPEEVM:hwHYPE' to accurately reflect that the token's network is HyperEVM. 
This chain is onboarded as 'HYPEEVM', not 'HYPE'.

### DIFF
--- a/modules/statics/src/allCoinsAndTokens.ts
+++ b/modules/statics/src/allCoinsAndTokens.ts
@@ -2965,11 +2965,11 @@ export const allCoinsAndTokens = [
   // hypeevm mainnet tokens
   erc20Token(
     '062b64cd-5569-41e9-a144-478976fd576f',
-    'hype:hwhype',
+    'hypeevm:hwhype',
     'Hyperwave HYPE',
     18,
     '0x4de03ca1f02591b717495cfa19913ad56a2f5858',
-    UnderlyingAsset['hype:hwhype'],
+    UnderlyingAsset['hypeevm:hwhype'],
     Networks.main.hypeevm
   ),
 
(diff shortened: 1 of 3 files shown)


modules/statics/src/allCoinsAndTokens.ts
```diff
@@ -2965,11 +2965,11 @@ export const allCoinsAndTokens = [
   // hypeevm mainnet tokens
   erc20Token(
     '062b64cd-5569-41e9-a144-478976fd576f',
-    'hype:hwhype',
+    'hypeevm:hwhype',
```
ajays97, line 2968: Should it be hyperevm instead?

ajays97 (approved):

modules/statics/src/allCoinsAndTokens.ts
bhuvanr159: but for testnet tokens, it is written as 'thypeevm:token-name', so followed the same.

ajays97: Understood. Just wanted to point out that the network is named HyperEVM. So this would be the right time to make edits

damodarnaik699 (commented):
A better description for this change in commit message will be good as the ticket is telling it's name as `hype:hwhype`

can you please add the description for this change

modules/statics/src/allCoinsAndTokens.ts
bhuvanr159: I re-verified with @abisheksingh2024, so it should be kept HYPEEVM itself.

PR #7830, opened 2026-01-02
fix: add xdc token contructor

at31416:
This pull request adds support for XDC-based tokens to the SDK by introducing a new `XdcToken` class, updating the registration logic to include XDC tokens, and providing unit tests to verify the new functionality.

**XDC Token Support:**

* Added a new `XdcToken` class in `xdcToken.ts` that extends `EthLikeToken` to support XDC-based tokens, including methods for transaction building, explorer queries, and token metadata.
* Updated the `register` function in `register.ts` to automatically register all available XDC token coin constructors with the SDK.

**Testing:**

* Added unit tests in `xdcToken.ts` to verify that XDC tokens are correctly registered and that their properties are set as expected.
Ticket: COIN-7060

### DIFF
--- a/modules/sdk-coin-xdc/src/register.ts
+++ b/modules/sdk-coin-xdc/src/register.ts
@@ -1,8 +1,15 @@
 import { BitGoBase } from '@bitgo/sdk-core';
+import { EthLikeErc20Token } from '@bitgo/sdk-coin-evm';
 import { Xdc } from './xdc';
 import { Txdc } from './txdc';
 
 export const register = (sdk: BitGoBase): void => {
   sdk.register('xdc', Xdc.createInstance);
   sdk.register('txdc', Txdc.createInstance);
+  EthLikeErc20Token.createTokenConstructors({
+    Mainnet: 'xdc',
+    Testnet: 'txdc',
+  }).forEach(({ name, coinConstructor }) => {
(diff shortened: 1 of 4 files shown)


modules/sdk-coin-xdc/src/register.ts
```diff
@@ -1,8 +1,12 @@
 import { BitGoBase } from '@bitgo/sdk-core';
 import { Xdc } from './xdc';
 import { Txdc } from './txdc';
+import { XdcToken } from './xdcToken';
```
ArunBala-Bitgo: Could we use `EthLikeErc20Token` here from **@bitgo/sdk-coin-evm** instead of XdcToken? @at31416

ArunBala-Bitgo (dismissed):

modules/sdk-coin-xdc/src/register.ts
ArunBala-Bitgo: Could we use EthLikeErc20Token from sdk-coin-evm here instead of XdcToken?

at31416: done.

ArunBala-Bitgo (approved):
LGTM

ravibitgo (approved):

PR #7844, opened 2026-01-06
fix(key-card): upgrade jspdf to 4.0.0 to resolve critical security bug/vulnerability

nvrakesh06:
TICKET: WIN-0000

Upgrades jspdf from ^3.0.2 to ^4.0.0 to address GHSA-f8cm-6447-x5h2, a critical path traversal vulnerability (CVSS 9.2) that could allow local file inclusion in the node.js build.


at31416 (approved):
lgtm

PR #7848, opened 2026-01-06
fix: override verify tss txn for xdc token

ArunBala-Bitgo:
Ticket: COIN-7060

**Why is this change necessary?**

XDC only supports legacy transactions and during validating the tss transaction we need to skip certain validations that aren’t relevant to this chain.  Consequently, we’re overriding them.

However, we overlooked overriding this in the xdcToken class during onboarding which is causing problems when initiating withdrawals.

BscTokens also uses legacy transactions and employs a similar approach.

reference: [bscToken.ts](https://github.com/BitGo/BitGoJS/blob/master/modules/sdk-coin-bsc/src/bscToken.ts)

### DIFF
--- a/modules/statics/src/account.ts
+++ b/modules/statics/src/account.ts
@@ -2888,7 +2888,7 @@ export function xdcErc20(
   decimalPlaces: number,
   contractAddress: string,
   asset: UnderlyingAsset,
-  features: CoinFeature[] = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.EIP1559],
+  features: CoinFeature[] = AccountCoin.DEFAULT_FEATURES,
   prefix = '',
   suffix: string = name.toUpperCase(),
   network: AccountNetwork = Networks.main.xdc,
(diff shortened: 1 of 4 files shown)


modules/statics/src/allCoinsAndTokens.ts
```diff
@@ -2996,31 +2996,35 @@ export const allCoinsAndTokens = [
     'Law Block Token',
     18,
     '0x05940b2df33d6371201e7ae099ced4c363855dfe',
-    UnderlyingAsset['xdc:lbt']
+    UnderlyingAsset['xdc:lbt'],
+    XDC_FEATURES
```
rohitsaw115: This will override the DEFAULT_FEATURES of xdcErc20 constructor. is this intended?

modules/statics/src/account.ts
```diff
@@ -2888,7 +2888,7 @@ export function xdcErc20(
   decimalPlaces: number,
   contractAddress: string,
   asset: UnderlyingAsset,
-  features: CoinFeature[] = [...AccountCoin.DEFAULT_FEATURES, CoinFeature.EIP1559],
```
rohitsaw115, line 2891: LGTM, If CoinFeature.EIP1559 is not required for xdc tokens

rohitsaw115 (dismissed):

prithvishet2503 (approved):

PR #7849, opened 2026-01-06
fix(wallet-key-share): Optimize createBulkWalletShare to Fetch Keychain Once and Prevent Rate Limiting

abk404:
Ticket: CSI-1548

## Type of change

## Summary

Optimizes `createBulkWalletShare` to fetch the wallet keychain only once instead of once per user, drastically reducing API calls and preventing rate limiting (HTTP 429) errors during bulk wallet sharing.

## Problem

When sharing a wallet with multiple users via `createBulkWalletShare`, the SDK was fetching the wallet keychain (`v2.key.get`) for **each user** in the share list. 

For example, sharing 26 wallets with 15 users each resulted in **390 API calls** (26 × 15), causing rate limiting errors and failures with the message: "Failed to process wallet shares after retrying with single requests." 
Check the support ticket : CSI-1569

## Solution

Refactored the code to:

1. **Extract common logic** into two new helper methods:
   - `getDecryptedKeychainForSharing()` - Fetches and decrypts the keychain once
   - `encryptPrvForUser()` - Encrypts the private key for a specific user (no API calls)

2. **Optimize `createBulkWalletShare()`** to fetch the keychain once before the loop, then reuse it for all users

### DIFF
--- a/modules/bitgo/test/v2/unit/wallets.ts
+++ b/modules/bitgo/test/v2/unit/wallets.ts
@@ -3708,6 +3708,99 @@ describe('V2 Wallets:', function () {
         assert.equal(error.message, 'Password shared is incorrect for this wallet');
       }
     });
+
+    it('should fetch keychain only ONCE when sharing with multiple users', async () => {
+      const walletPassphrase = 'bitgo1234';
+      const pub = 'Zo1ggzTUKMY5bYnDvT5mtVeZxzf2FaLTbKkmvGUhUQk';
+      const prv1 = 'xprv1';
+
+      // Create multiple users to share with
+      const users = [
+        {
+          userId: 'user1@example.com',
+          permissions: ['view', 'spend'],
+          pubKey: '02705a6d33a2459feb537e7abe36aaad8c11532cdbffa3a2e4e58868467d51f532',
+          path: 'm/999999/1/1',
+        },
+        {
+          userId: 'user2@example.com',
+          permissions: ['view', 'spend'],
+          pubKey: '03705a6d33a2459feb537e7abe36aaad8c11532cdbffa3a2e4e58868467d51f533',
+          path: 'm/999999/1/2',
+        },
+        {
+          userId: 'user3@example.com',
+          permissions: ['view', 'spend'],
+          pubKey: '02815a6d33a2459feb537e7abe36aaad8c11532cdbffa3a2e4e58868467d51f534',
+          path: 'm/999999/1/3',
+        },
+      ];
+
+      const params: BulkWalletShareOptions = {
+        walletPassphrase,
+        keyShareOptions: users,
+      };
+
+      const getDecryptedKeychainStub = sinon.stub(wallet, 'getDecryptedKeychainForSharing').resolves({
+        prv: prv1,
+        pub,
+      });
+
+      const encryptPrvForUserStub = sinon
+        .stub(wallet, 'encryptPrvForUser')
+        .callsFake((prv, pubKey, userPubKey, path) => {
+          return {
+            pub: pubKey,
+            encryptedPrv: 'dummyEncryptedPrv',
+            fromPubKey: 'dummyFromPubKey',
+            toPubKey: userPubKey,
+            path: path,
+          };
+        });
+
+      sinon.stub(wallet, 'createBulkKeyShares').resolves({
+        shares: users.map((user) => ({
+          id: user.userId,
+          coin: walletData.coin,
+          wallet: walletData.id,
+          fromUser: 'fromUser',
+          toUser: user.userId,
+          permissions: user.permissions,
+          keychain: {
+            pub: 'dummyPub',
+            encryptedPrv: 'dummyEncryptedPrv',
+            fromPubKey: 'dummyFromPubKey',
+            toPubKey: user.pubKey,
+            path: user.path,
+          },
+        })),
+      });
+
+      await wallet.createBulkWalletShare(params);
+
+      // Verify keychain is fetched only ONCE, not once per user
+      assert.strictEqual(
+        getDecryptedKeychainStub.callCount,
+        1,
+        'getDecryptedKeychainForSharing should be called exactly once, not once per user'
+      );
+
+      // Verify encryption happens for EACH user (once per user)
+      assert.strictEqual(
+        encryptPrvForUserStub.callCount,
+        users.length,
+        `encryptPrvForUser should be called once per user (${users.length} times)`
+      );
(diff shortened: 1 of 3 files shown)


modules/bitgo/test/v2/unit/wallets.ts
```diff
@@ -3708,6 +3708,71 @@ describe('V2 Wallets:', function () {
...
+          permissions: user.permissions,
+          keychain: {
+            pub: 'dummyPub',
+            encryptedPrv: 'dummyEncryptedPrv',
+            fromPubKey: 'dummyFromPubKey',
+            toPubKey: user.pubKey,
+            path: user.path,
+          },
+        })),
+      });
```
sdoshibitgo: **Expand test coverage:** Also verify `encryptPrvForUser` is called once per user to ensure encryption happens for each recipient.

abk404: Added

sdoshibitgo (approved):

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -645,6 +645,11 @@ export interface BulkWalletShareOptions {
 
 export type WalletShareState = 'active' | 'accepted' | 'canceled' | 'rejected' | 'pendingapproval';
 
+export interface DecryptedKeychainData {
+  prv: string;
+  pub: string;
+}
```
zahin-mohammad, line 651: Are we sure we don't already have this type defined somewhere?

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1776,62 +1782,107 @@ export class Wallet implements IWallet {
...
-        keychain.prv = userPrv;
-        const eckey = makeRandomKey();
-        const secret = getSharedSecret(eckey, Buffer.from(pubkey, 'hex')).toString('hex');
-        const newEncryptedPrv = this.bitgo.encrypt({
-          password: secret,
-          input: keychain.prv,
-        });
+    // Only one of pub/commonPub/commonKeychain should be present in the keychain
+    let pub = keychain.pub ?? keychain.commonPub;
+    if (keychain.commonKeychain) {
```
zahin-mohammad, line 1813:
```suggestion
    // Only one of pub/commonKeychain should be present in the keychain
    let pub = keychain.pub;
```

commonPub is deprecated field, we shouldn't have any active wallets that are using this field.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1776,62 +1782,107 @@ export class Wallet implements IWallet {
...
+      pub =
+        this.baseCoin.getMPCAlgorithm() === 'eddsa'
+          ? EddsaUtils.getPublicKeyFromCommonKeychain(keychain.commonKeychain)
+          : EcdsaUtils.getPublicKeyFromCommonKeychain(keychain.commonKeychain);
```
zahin-mohammad, line 1817: Why is this necessary??

zahin-mohammad: Will ping on slack.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -1776,62 +1782,107 @@ export class Wallet implements IWallet {
...
+
+    const keychain: BulkWalletShareKeychain = {
+      pub,
+      encryptedPrv: newEncryptedPrv,
+      fromPubKey: eckey.publicKey.toString('hex'),
+      toPubKey: userPubkey,
+      path: path,
+    };
+
+    assert(keychain.pub, 'pub must be defined for sharing');
```
zahin-mohammad, line 1850:
```suggestion
    const keychain: BulkWalletShareKeychain = {
      encryptedPrv: newEncryptedPrv,
      fromPubKey: eckey.publicKey.toString('hex'),
      toPubKey: userPubkey,
      path: path,
    };
```

This shouldn't be necessary.

modules/sdk-core/src/bitgo/wallet/iWallet.ts
abk404: No , we don't have this type defined somewhere else. 
This was not required before because the `prepareSharedKeychain` was doing the operations in the same function. 

Since, I broke down that function into 3 different function , we need this new interface for having the return typed for the  new function. 

There is a `Keychain` type defined which has these two fields(along with ton of other fields), but there also these fields are marked as optional .
i can do  `Required<Pick<Keychain, 'prv' | 'pub'>>` but the current approcah of defining a new interface seems more readable and explicit. 

This also follows the current pattern where a new type `BulkWalletShareKeychain` was created for the specific usecase instead of usign the `Keychain` type.

modules/sdk-core/src/bitgo/wallet/wallet.ts
abk404: The `commonPub` fallback was pre-existing code from the original bulk sharing implementation. This PR is a refactoring that moves this logic from `prepareSharedKeychain()` to `getDecryptedKeychainForSharing()` without changing any behaviour. 

If it's a deprecated field , I can remove this, but just thinking if it will break some other flows in downstream services  which calls the `/key` endpoint and still expect the `commonPub` field.

modules/sdk-core/src/bitgo/wallet/wallet.ts
abk404: This was also there before in the codebase I refactored the code to pull out this pub key creation logic into a separate function. 
The `/walletshare/{id}` responses  still include the `pub` , so I am a little skeptical if removing the field might break some other services consuming this endpoint. Also pub is used when accepting shares. 

If we want to remove pub or simplify this extraction logic, I'd suggest doing it in a separate cleanup PR to avoid scope creep and ensure proper testing.

modules/sdk-core/src/bitgo/wallet/wallet.ts
abk404: Same as above comment.

zahin-mohammad (approved):

modules/sdk-core/src/bitgo/wallet/wallet.ts
abk404: Created [this ticket](https://bitgoinc.atlassian.net/browse/CSI-1597) to handle the cleanup task.

PR #7853, opened 2026-01-06
Use wasm-utxo dimensions for recovery

lcovar:


modules/abstract-utxo/src/recovery/crossChainRecovery.ts
```diff
@@ -406,10 +407,10 @@ function createSweepTransactionWasm<TNumber extends number | bigint = number>(
...
   addWalletInputsToWasmPsbt(wasmPsbt, unspentsBigint, walletKeys);
 
-  // Convert to utxolib PSBT temporarily for dimension calculation
-  const tempPsbt = wasmPsbtToUtxolibPsbt(wasmPsbt, network);
-  const vsize = Dimensions.fromPsbt(tempPsbt)
-    .plus(Dimensions.fromOutput({ script: utxolib.address.toOutputScript(targetAddress, network) }))
+  // Calculate dimensions using wasm-utxo Dimensions
+  const targetOutputScript = utxolib.address.toOutputScript(targetAddress, network);
+  const vsize = fixedScriptWallet.Dimensions.fromPsbt(wasmPsbt)
+    .plus(fixedScriptWallet.Dimensions.fromOutput(new Uint8Array(targetOutputScript)))
```
OttoAllmendinger, line 413: there is a `fromOutput(address, coinName)` as well that allows us to get rid of utxolib completely

modules/abstract-utxo/src/recovery/psbt.ts
```diff
@@ -249,18 +249,15 @@ function createBackupKeyRecoveryPsbtWasm(
...
   addWalletInputsToWasmPsbt(wasmPsbt, unspents, rootWalletKeys);
 
-  // Convert to utxolib PSBT temporarily for dimension calculation
-  const tempPsbt = wasmPsbtToUtxolibPsbt(wasmPsbt, network);
-  let dimensions = Dimensions.fromPsbt(tempPsbt).plus(
-    Dimensions.fromOutput({ script: utxolib.address.toOutputScript(recoveryDestination, network) })
+  // Calculate dimensions using wasm-utxo Dimensions
+  const recoveryOutputScript = utxolib.address.toOutputScript(recoveryDestination, network);
+  let dimensions = fixedScriptWallet.Dimensions.fromPsbt(wasmPsbt).plus(
+    fixedScriptWallet.Dimensions.fromOutput(new Uint8Array(recoveryOutputScript))
```
OttoAllmendinger, line 255: same

modules/abstract-utxo/src/recovery/crossChainRecovery.ts
OttoAllmendinger: we don't need `new Uint8Array`

OttoAllmendinger: let's fix it in a follow-up

davidkaplanbitgo (approved):

PR #7878, opened 2026-01-09
fix(sdk-core): update go staking request interface

noel-bitgo:
to match actual API response

TICKET: SC-4862


feng-bitgo (approved):
lgtm

PR #7884, opened 2026-01-12
feat(abstract-utxo): implement BIP-322 signature verification with wasm-utxo

OttoAllmendinger:
This PR adds support for verifying BIP-322 message signatures using 
the wasm-utxo implementation instead of utxo-core. The implementation:

- Adds verification capabilities for BIP-322 message signatures created 
  with wasm-utxo
- Implements test cases for signature validation across both p2shP2wsh 
  and p2wsh script types
- Adds support for verifying multiple inputs in a single PSBT
- Simplifies the verification process by removing redundant nonWitnessUtxo 
  checks, since the txid check in assertTxInput already ensures correct 
  references

BTC-2916

### DIFF
--- a/modules/utxo-core/src/bip322/verify.ts
+++ b/modules/utxo-core/src/bip322/verify.ts
@@ -113,10 +113,8 @@ export function assertBip322PsbtProof(psbt: utxolib.bitgo.UtxoPsbt, messageInfo:
     // Check that the metadata in the PSBT matches the messageInfo, then check the input data
     const info = messageInfo[inputIndex];
 
-    // Check that the to_spend transaction is encoded in the nonWitnessUtxo
-    assert.ok(input.nonWitnessUtxo, `PSBT input ${inputIndex} is missing nonWitnessUtxo`);
-    const toSpendTx = buildToSpendTransaction(prevOuts[inputIndex].script, info.message);
-    assert.deepStrictEqual(input.nonWitnessUtxo.toString('hex'), toSpendTx.toHex());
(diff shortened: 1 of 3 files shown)


modules/utxo-core/src/bip322/verify.ts
```diff
@@ -113,10 +113,8 @@ export function assertBip322PsbtProof(psbt: utxolib.bitgo.UtxoPsbt, messageInfo:
     // Check that the metadata in the PSBT matches the messageInfo, then check the input data
     const info = messageInfo[inputIndex];
 
-    // Check that the to_spend transaction is encoded in the nonWitnessUtxo
-    assert.ok(input.nonWitnessUtxo, `PSBT input ${inputIndex} is missing nonWitnessUtxo`);
-    const toSpendTx = buildToSpendTransaction(prevOuts[inputIndex].script, info.message);
-    assert.deepStrictEqual(input.nonWitnessUtxo.toString('hex'), toSpendTx.toHex());
+    // Note: nonWitnessUtxo is not required - the txid check in assertTxInput verifies
+    // the input references the correct to_spend transaction (computed from message + scriptPubKey)
```
OttoAllmendinger, line 117: @davidkaplanbitgo 👁️

davidkaplanbitgo: Originally it was done like this so that the validation script run by midnight labs could re-derive all of the information needed during proofing. It is not required per BIP322

davidkaplanbitgo (approved):

PR #7888, opened 2026-01-13
feat: address resolution function for hbarevm

rohitsaw115:
ticket: win-8570

### DIFF
--- a/modules/sdk-coin-evm/src/lib/utils.ts
+++ b/modules/sdk-coin-evm/src/lib/utils.ts
@@ -272,3 +272,47 @@ export function validateHederaAccountId(address: string): { valid: boolean; erro
     error: null,
   };
 }
+
+/**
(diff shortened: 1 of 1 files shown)


modules/sdk-coin-evm/src/lib/utils.ts
```diff
@@ -272,3 +272,47 @@ export function validateHederaAccountId(address: string): { valid: boolean; erro
     error: null,
   };
 }
+
+/**
+ * Convert Hedera Account ID (e.g., 0.0.12345) to EVM address
+ */
+export async function resolveHederaAccountIdToEvmAddress(
```
ArunBala-Bitgo, line 279: Do we have testcase for this method?

rohitsaw115: moved this from ui, should have integration tests in bitgo-ui. will move relevents tests here later.

ArunBala-Bitgo (approved):

PR #7902, opened 2026-01-14
fix: pass head sha to find merge base

starfy84:
This commit passes the head sha when checking out the PR so that we can find the correct merge base commit. With the previous logic, the merge base was being incorrectly identified as a commit newer than the head sha. This resulted in breaking changes being flagged on PRs that don't even make express changes.

TICKET: DX-2778


ericcrosson-bitgo (approved):
LGTM

PR #7914, opened 2026-01-15
fix(sdk): change password should work with ofc multi-user key

alextse-bg:
TICKET: WP-7461

### DIFF
--- a/modules/sdk-core/src/bitgo/keychain/keychains.ts
+++ b/modules/sdk-core/src/bitgo/keychain/keychains.ts
@@ -119,7 +119,11 @@ export class Keychains implements IKeychains {
             newPassword: params.newPassword,
           });
           if (updatedKeychain.encryptedPrv) {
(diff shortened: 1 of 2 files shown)


deepesh-bitgo (approved):

modules/sdk-core/src/bitgo/keychain/keychains.ts
```diff
@@ -119,7 +119,11 @@ export class Keychains implements IKeychains {
             newPassword: params.newPassword,
           });
           if (updatedKeychain.encryptedPrv) {
-            const changedKeyIdentifier = updatedKeychain.type === 'tss' ? updatedKeychain.id : updatedKeychain.pub;
+            // Both TSS and multi-user-ofc keys have multiple public keys in their key document and thus need to use objectID
```
zahin-mohammad, line 122:
```suggestion
            // TSS and multi-user-ofc keys cannot uniquely identify key documents via the pub (either the pub is common among the 3 keys, or the root pub is a magic value). 
```

zahin-mohammad (approved):

PR #7930, opened 2026-01-19
fix(sdk-coin-ada): use derivedFromParentWithSeed in address verification

dpkjnr:
Address verification was failing for wallets created with existing keys because isWalletAddress was not using the derivedFromParentWithSeed property to compute the correct derivation path. This fix extracts the seed from the keychain and passes it to getAdaAddressAndAccountId.

TICKET: COIN-7254


MohammedRyaan786 (approved):
LGTM

PR #7940, opened 2026-01-20
fix(sdk-coin-ada): token verifytransaction

akarath:
JIRA: COIN-7292

### DIFF
--- a/modules/sdk-coin-ada/src/adaToken.ts
+++ b/modules/sdk-coin-ada/src/adaToken.ts
@@ -1,6 +1,15 @@
 import { Ada } from './ada';
-import { BitGoBase, CoinConstructor, NamedCoinConstructor } from '@bitgo/sdk-core';
+import {
+  BitGoBase,
+  CoinConstructor,
+  NamedCoinConstructor,
+  VerifyTransactionOptions,
+  NodeEnvironmentError,
+} from '@bitgo/sdk-core';
 import { coins, tokens, AdaTokenConfig } from '@bitgo/statics';
+import { Transaction } from './lib';
+import * as CardanoWasm from '@emurgo/cardano-serialization-lib-nodejs';
+import assert from 'assert';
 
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-ada/src/adaToken.ts
```diff
@@ -85,4 +94,50 @@ export class AdaToken extends Ada {
...
+            const multiAssets = output.multiAssets as CardanoWasm.MultiAsset;
+            const tokenQty = multiAssets.get_asset(policyScriptHash, assetName);
+            return tokenQty && tokenQty.to_str() === recipient.amount;
+          });
+
+          if (!found) {
+            throw new Error('cannot find recipient in expected output');
+          }
+        }
+      }
```
Ranjna-G, line 156: Can we please add the consolidation IF case as well so that we do not have to remember this touchpoint in single asset consolidation of tokens?

modules/sdk-coin-ada/src/adaToken.ts
```diff
@@ -85,4 +94,50 @@ export class AdaToken extends Ada {
...
+      const coinConfig = coins.get(this.getBaseChain());
+      const { txPrebuild, txParams } = params;
+      const transaction = new Transaction(coinConfig);
+      assert(txPrebuild.txHex, new Error('missing required tx prebuild property txHex'));
+
+      transaction.fromRawTransaction(txPrebuild.txHex);
+      const txJson = transaction.toJson();
+
+      if (txParams.recipients !== undefined) {
+        const policyScriptHash = CardanoWasm.ScriptHash.from_hex(this.tokenConfig.policyId);
```
Ranjna-G: This policyId might have the encoded asset name as the suffix whereas the output.assets might have just the policy id, can we ensure this case is handled ?

Ranjna-G (approved):
looks good

PR #7941, opened 2026-01-20
feat(statics): change kava to new cosmos structure

abhijeet8986:
TICKET: COIN-7315


modules/statics/src/networks.ts
```diff
@@ -975,16 +975,28 @@ class InjectiveTestnet extends Testnet implements AccountNetwork {
   explorerUrl = 'https://testnet.explorer.injective.network/transaction/';
 }
 
-class Kava extends Mainnet implements AccountNetwork {
+class Kava extends Mainnet implements CosmosNetwork {
   name = 'Kava';
   family = CoinFamily.KAVA;
   explorerUrl = 'https://www.mintscan.io/kava/transactions/';
+  addressPrefix = 'kava';
```
baltiyal, line 982: Could you provide the source of these values?

abhijeet8986: The values come from coin sandboxing report for kava, here is the PR for the same: [coin-sandbox-pr](https://github.com/BitGo/coins-sandbox/pull/692)

baltiyal: As a general practice, first get approval and merge the sandbox pr before raising the statics PR.

modules/statics/src/networks.ts
```diff
@@ -975,16 +975,28 @@ class InjectiveTestnet extends Testnet implements AccountNetwork {
   explorerUrl = 'https://testnet.explorer.injective.network/transaction/';
 }
 
-class Kava extends Mainnet implements AccountNetwork {
+class Kava extends Mainnet implements CosmosNetwork {
   name = 'Kava';
   family = CoinFamily.KAVA;
   explorerUrl = 'https://www.mintscan.io/kava/transactions/';
```
baltiyal: The explorer endpoints will also change.

abhijeet8986: Kava explorer URLs are as follows:
Mainnet: https://www.mintscan.io/kava/tx/ — for Cosmos-native Kava transactions
Testnet: https://www.mintscan.io/kava-testnet/tx/ — for testnet transactions

Changed to above

modules/statics/src/networks.ts
baltiyal: @abhijeet8986 As discussed, if we onboard Kava as a Cosmos coin while keeping KavaEVM as is, let’s avoid changing the existing EVM code and instead add new, separate code for this.

alextse-bg (commented):
not sure why the wallets team is ping on this as it doesn't seems like we own anything is statics. Deferring to eth-alt

baltiyal (approved):
lgtm

PR #7957, opened 2026-01-21
feat: add explainSolTransaction using @bitgo/wasm-solana

lcovar:
## Summary

Add a standalone `explainSolTransaction()` function to `sdk-coin-sol` that uses `@bitgo/wasm-solana` to explain Solana transactions without depending on `@solana/web3.js`.

This is a thin ~100-line adapter that replaces the need for the legacy 800+ line `Transaction` class explain logic. The heavy lifting (parsing, instruction combining, type derivation) happens in `@bitgo/wasm-solana` — this adapter only handles:
- Token name resolution (mint address → `sol:usdc` via `@bitgo/statics`)
- bigint → string conversion at the serialization boundary
- `StakingAuthorize` → `StakingAuthorizeRaw` type mapping for downstream compatibility
- Mapping to `TransactionExplanation` interface

## Changes

- **`explainTransactionWasm.ts`** (new): Standalone adapter that calls `@bitgo/wasm-solana`'s `explainTransaction()`, resolves token names, maps staking authorize fields, and returns `TransactionExplanation` including `inputs` and `feePayer`
- **`transaction.ts`**: Route tsol through WASM-based explain (validates against production traffic before replacing legacy for all networks)
- **`sol.ts`**: Add `explainTransactionWithWasm()` method at coin class level
- **`iface.ts`**: Add `ataOwnerMap` and `stakingAuthorize` to `TransactionExplanation` interface
- **`instructionParamsFactory.ts`**: Add Jito stake pool operations (`StakePoolDepositSol`, `StakePoolWithdrawStake`)
- **`jitoWasmVerification.ts`**: New test file verifying Jito liquid staking explain output
- **`transaction.ts` tests**: Updated assertions for WASM parity (fee defaults to '0' instead of 'UNAVAILABLE', ataOwnerMap, inputs, feePayer, token name via statics)
- **`sol.ts` tests**: Updated explain assertions to include `inputs` and `feePayer` returned by WASM path
- **`webpack config`**: Add WASM asset handling for `@bitgo/wasm-solana`
- **`package.json`**: Add `@bitgo/wasm-solana` dependency

## Key decisions

- **tsol only**: WASM path is gated to testnet (`tsol`) to validate against production traffic before replacing the legacy path for mainnet
- **Fee behavior**: Returns `fee: '0'` and `feeRate: undefined` when fee info not provided (instead of `'UNAVAILABLE'`). Downstream already converts UNAVAILABLE → '0', so this is a no-op change
- **StakingAuthorizeRaw**: WASM can decode both raw and non-raw authorize instructions natively, but downstream expects the `StakingAuthorizeRaw` type name, so the adapter maps it
- **Token names**: Resolved via `@bitgo/statics`. Tokens not in statics return the raw mint address
- **inputs/feePayer**: Now included in the return so consumers (e.g. wallet-platform) can use the SDK as a single source of truth without calling wasm-solana directly
- **Test updates**: 26 test assertions updated to include `inputs` and `feePayer` — these fields were already on the `TransactionExplanation` interface but weren't populated by the WASM path until now

## Depends on

- [BitGoWASM PR #166](https://github.com/BitGo/BitGoWASM/pull/166) — `explainTransaction()` with `StakingAuthorizeInfo`, optional `lamportsPerSignature` (defaults to 5000n)

## Test Plan

- [x] All existing sdk-coin-sol tests pass with updated assertions
- [x] New Jito WASM verification tests pass
- [x] Parity verified against legacy `Transaction` class output for all transaction types

TICKET: BTC-3025

### DIFF
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -57,7 +57,13 @@ import {
 } from '@bitgo/sdk-core';
 import { auditEddsaPrivateKey, getDerivationPath } from '@bitgo/sdk-lib-mpc';
 import { BaseNetwork, CoinFamily, coins, SolCoin, BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import {
+  explainTransaction as wasmExplainTransaction,
+  type ExplainedTransaction as WasmExplainedTransaction,
+} from '@bitgo/wasm-solana';
 import { KeyPair as SolKeyPair, Transaction, TransactionBuilder, TransactionBuilderFactory } from './lib';
+import { UNAVAILABLE_TEXT } from './lib/constants';
+import { TransactionExplanation as SolLibTransactionExplanation } from './lib/iface';
 import {
   getAssociatedTokenAccountAddress,
   getSolTokenFromAddress,
@@ -67,6 +73,7 @@ import {
   isValidPublicKey,
   validateRawTransaction,
 } from './lib/utils';
+import { findTokenName } from './lib/instructionParamsFactory';
 
 export const DEFAULT_SCAN_FACTOR = 20; // default number of receive addresses to scan for funds
 
@@ -80,6 +87,7 @@ export interface ExplainTransactionOptions {
   txBase64: string;
   feeInfo: TransactionFee;
   tokenAccountRentExemptAmount?: string;
+  coinName?: string;
 }
 
 export interface TxInfo {
@@ -696,6 +704,7 @@ export class Sol extends BaseCoin {
   }
 
   async parseTransaction(params: SolParseTransactionOptions): Promise<SolParsedTransaction> {
+    // explainTransaction now uses WASM for testnet automatically
     const transactionExplanation = await this.explainTransaction({
       txBase64: params.txBase64,
       feeInfo: params.feeInfo,
@@ -741,9 +750,16 @@ export class Sol extends BaseCoin {
 
   /**
    * Explain a Solana transaction from txBase64
+   * Uses WASM-based parsing for testnet, with fallback to legacy builder approach.
    * @param params
    */
   async explainTransaction(params: ExplainTransactionOptions): Promise<SolTransactionExplanation> {
+    // Use WASM-based parsing for testnet (simpler, faster, no @solana/web3.js rebuild)
+    if (this.getChain() === 'tsol') {
+      return this.explainTransactionWithWasm(params) as SolTransactionExplanation;
+    }
+
+    // Legacy approach for mainnet (until WASM is fully validated)
     const factory = this.getBuilder();
     let rebuiltTransaction;
 
@@ -765,6 +781,14 @@ export class Sol extends BaseCoin {
     const explainedTransaction = (rebuiltTransaction as BaseTransaction).explainTransaction();
 
     return explainedTransaction as SolTransactionExplanation;
+  }
+
+  /**
+   * Explain a Solana transaction using WASM parsing (bypasses @solana/web3.js rebuild).
+   * Delegates to standalone explainSolTransaction().
+   */
+  explainTransactionWithWasm(params: ExplainTransactionOptions): SolLibTransactionExplanation {
+    return explainSolTransaction({ ...params, coinName: params.coinName ?? this.getChain() });
   }
 
   /** @inheritDoc */
@@ -1744,4 +1768,64 @@ export class Sol extends BaseCoin {
       intent.solVersionedTransactionData = params.solVersionedTransactionData;
     }
   }
+}
+
+/**
+ * Standalone WASM-based transaction explanation — no class instance needed.
+ * Thin adapter over @bitgo/wasm-solana's explainTransaction that resolves
+ * token names via @bitgo/statics and maps to BitGoJS TransactionExplanation.
+ */
+export function explainSolTransaction(
+  params: ExplainTransactionOptions & { coinName: string }
+): SolLibTransactionExplanation {
+  const txBytes = Buffer.from(params.txBase64, 'base64');
+  const explained: WasmExplainedTransaction = wasmExplainTransaction(txBytes, {
+    lamportsPerSignature: BigInt(params.feeInfo?.fee || '0'),
+    tokenAccountRentExemptAmount: params.tokenAccountRentExemptAmount
+      ? BigInt(params.tokenAccountRentExemptAmount)
+      : undefined,
+  });
+
+  // Resolve token mint addresses → human-readable names (e.g. "tsol:usdc")
+  // Convert bigint amounts to strings at this serialization boundary.
+  const outputs = explained.outputs.map((o) => ({
+    address: o.address,
+    amount: String(o.amount),
+    ...(o.tokenName ? { tokenName: findTokenName(o.tokenName, undefined, true) } : {}),
+  }));
+
+  // Build tokenEnablements with resolved token names
+  const tokenEnablements: ITokenEnablement[] = explained.tokenEnablements.map((te) => ({
+    address: te.address,
+    tokenName: findTokenName(te.mintAddress, undefined, true),
+    tokenAddress: te.mintAddress,
+  }));
+
+  return {
+    displayOrder: [
+      'id',
+      'type',
+      'blockhash',
+      'durableNonce',
+      'outputAmount',
+      'changeAmount',
+      'outputs',
+      'changeOutputs',
+      'tokenEnablements',
+      'fee',
+      'memo',
+    ],
+    id: explained.id ?? UNAVAILABLE_TEXT,
+    type: explained.type,
+    changeOutputs: [],
+    changeAmount: '0',
+    outputAmount: String(explained.outputAmount),
+    outputs,
+    fee: { fee: String(explained.fee), feeRate: Number(params.feeInfo?.fee || '0') },
+    memo: explained.memo,
+    blockhash: explained.blockhash,
+    durableNonce: explained.durableNonce,
+    tokenEnablements,
+    ataOwnerMap: explained.ataOwnerMap,
+  };
 }
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -58,11 +58,12 @@ import {
 import { auditEddsaPrivateKey, getDerivationPath } from '@bitgo/sdk-lib-mpc';
 import { BaseNetwork, CoinFamily, coins, SolCoin, BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
 import {
-  explainTransaction as wasmExplainTransaction,
-  type ExplainedTransaction as WasmExplainedTransaction,
-} from '@bitgo/wasm-solana';
-import { KeyPair as SolKeyPair, Transaction, TransactionBuilder, TransactionBuilderFactory } from './lib';
-import { UNAVAILABLE_TEXT } from './lib/constants';
+  KeyPair as SolKeyPair,
+  Transaction,
+  TransactionBuilder,
+  TransactionBuilderFactory,
+  explainSolTransaction,
+} from './lib';
 import { TransactionExplanation as SolLibTransactionExplanation } from './lib/iface';
 import {
   getAssociatedTokenAccountAddress,
@@ -73,7 +74,6 @@ import {
   isValidPublicKey,
   validateRawTransaction,
 } from './lib/utils';
-import { findTokenName } from './lib/instructionParamsFactory';
 
 export const DEFAULT_SCAN_FACTOR = 20; // default number of receive addresses to scan for funds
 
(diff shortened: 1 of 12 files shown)


modules/sdk-coin-sol/src/lib/wasm/transaction.ts
```diff
@@ -0,0 +1,377 @@
...
+import { InstructionBuilderTypes, UNAVAILABLE_TEXT } from '../constants';
+import { DurableNonceParams, InstructionParams, TxData, TransactionExplanation } from '../iface';
+
+/**
+ * Solana transaction using WASM for all parsing operations.
+ *
+ * Key differences from legacy Transaction:
+ * - No @solana/web3.js dependency
+ * - No conditional code paths
+ * - ~150 lines instead of 800+
```
zahin-mohammad:
```suggestion
```

Is this necessary?

modules/sdk-coin-sol/src/lib/wasm/transaction.ts
```diff
@@ -0,0 +1,377 @@
...
+      'outputAmount',
+      'changeAmount',
+      'outputs',
+      'changeOutputs',
+      'fee',
+      'memo',
+    ];
+
+    const outputs: { address: string; amount: string; memo?: string }[] = [];
+    let outputAmount = '0';
```
OttoAllmendinger:
```suggestion
    let outputAmount = 0n;
```

modules/sdk-coin-sol/src/lib/wasm/transaction.ts
```diff
@@ -0,0 +1,377 @@
...
+    let memo: string | undefined;
+
+    for (const instr of this._instructionsData) {
+      switch (instr.type) {
+        case InstructionBuilderTypes.Transfer:
+          outputs.push({
+            address: instr.params.toAddress,
+            amount: instr.params.amount,
+          });
+          outputAmount = (BigInt(outputAmount) + BigInt(instr.params.amount)).toString();
```
OttoAllmendinger: let's avoid the internal bigint casts - why is instr.params.amount not a bigint already?

modules/sdk-coin-sol/src/lib/wasmInstructionMapper.ts
```diff
@@ -0,0 +1,499 @@
...
+// Individual Instruction Mapping
+// =============================================================================
+
+function mapTransfer(instr: WasmTransferParams): InstructionParams {
+  return {
+    type: InstructionBuilderTypes.Transfer,
+    params: {
+      fromAddress: instr.fromAddress,
+      toAddress: instr.toAddress,
+      amount: instr.amount.toString(),
```
OttoAllmendinger: sigh I guess because we have to satisfy this type

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -766,6 +785,169 @@ export class Sol extends BaseCoin {
...
+    // Use centralized combining utility - single source of truth for all combining logic
+    const { instructions: combinedInstructions, transactionType } = combineWasmInstructionsFromBytes(txBytes);
+
+    // Extract memo from parsed transaction
+    const memo = parsed.instructionsData.find((i) => i.type === 'Memo')?.memo;
+
+    // Derive outputs and tokenEnablements from combined instructions
+    const outputs: TransactionRecipient[] = [];
+    const tokenEnablements: ITokenEnablement[] = [];
+    let outputAmount = new BigNumber(0);
```
OttoAllmendinger: yet another way to sum up things... use bigint instead

OttoAllmendinger (changes requested):
we can move towards bigint a little bit more

where external interfaces force us to `string` keep it

modules/sdk-coin-sol/src/lib/wasm/transaction.ts
```diff
@@ -0,0 +1,377 @@
...
+    super(coinConfig);
+  }
+
+  // =============================================================================
+  // Core Properties
+  // =============================================================================
+
+  /** Transaction ID (first signature, base58 encoded) */
+  get id(): string {
+    if (!this._wasmTransaction) return UNAVAILABLE_TEXT;
```
OttoAllmendinger: I'm guessing the interface forces us to do this ugliness?

lcovar: Unfortunately yes, base class forces the return type of `string` while _id can be `string | undefined` so it's a bit silly. We have to return a string, i guess we can return an empty string. 

Seeing what it would take to update the base class to accept `undefined` as a return type. It would cause some breaking changes though but lets see. 🤖

OttoAllmendinger: Fixing the interface could spiral out of control very quickly, so let's leave it. I just need to know what is hard to change and what is easy to change. You can use the `override` keyword in typescript to signal that a method or getter implements a base method.

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -767,6 +786,169 @@ export class Sol extends BaseCoin {
...
+
+        case InstructionBuilderTypes.CreateAssociatedTokenAccount:
+          tokenEnablements.push({
+            address: instr.params.ataAddress,
+            tokenName: findTokenName(instr.params.mintAddress, undefined, true),
+            tokenAddress: instr.params.mintAddress,
+          });
+          break;
+      }
+    }
```
OttoAllmendinger: this whole thing gives me a deja vue 

@claude can we consolidate here

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -767,6 +786,169 @@ export class Sol extends BaseCoin {
...
+
+    // Calculate fee: lamportsPerSignature * numSignatures + (rent * numATAs)
+    const lamportsPerSignature = parseInt(params.feeInfo?.fee || '0', 10);
+    const rentPerAta = parseInt(params.tokenAccountRentExemptAmount || '0', 10);
```
OttoAllmendinger: @claude suggest a inner func that has a nicer signature with already parsed args

don't we want `bigint` for amounts anyway?

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -767,6 +786,169 @@ export class Sol extends BaseCoin {
...
+    const signatureFee = lamportsPerSignature * parsed.numSignatures;
+    const rentFee = rentPerAta * tokenEnablements.length;
+    const totalFee = (signatureFee + rentFee).toString();
+
+    // Get transaction id from first signature (base58 encoded) or UNAVAILABLE
+    let txId = 'UNAVAILABLE';
```
OttoAllmendinger: this needs to be a higher level const

modules/sdk-coin-sol/src/sol.ts
```diff
@@ -767,6 +786,169 @@ export class Sol extends BaseCoin {
...
+        return 'StakingAuthorize';
+      case TransactionType.AssociatedTokenAccountInitialization:
+        return 'AssociatedTokenAccountInitialization';
+      case TransactionType.CloseAssociatedTokenAccount:
+        return 'CloseAssociatedTokenAccount';
+      case TransactionType.CustomTx:
+        return 'CustomTx';
+      default:
+        return 'Send';
+    }
```
OttoAllmendinger: this should not be a coin method

OttoAllmendinger (changes requested):
counting on you claude

modules/sdk-coin-sol/src/sol.ts
OttoAllmendinger: *🤖 Claude here.*

Yes, this output extraction loop is duplicated in three places:
1. `Transaction.explainTransaction()` (transaction.ts:525-587) using `BigNumber`
2. `WasmTransaction.explainTransaction()` (wasm/transaction.ts:206-240) using `bigint`
3. `Sol.explainTransactionWithWasm()` (sol.ts:810-860) using `bigint`

Proposal: extract a shared free function in a new `lib/explainUtil.ts`:

```typescript
interface ExplainedOutputs {
  outputs: TransactionRecipient[];
  tokenEnablements: ITokenEnablement[];
  outputAmount: bigint;
  memo?: string;
}

function extractOutputsFromInstructions(instructions: InstructionParams[]): ExplainedOutputs {
  const outputs: TransactionRecipient[] = [];
  const tokenEnablements: ITokenEnablement[] = [];
  let outputAmount = 0n;
  let memo: string | undefined;

  for (const instr of instructions) {
    switch (instr.type) {
      case InstructionBuilderTypes.Transfer:
        outputs.push({ address: instr.params.toAddress, amount: instr.params.amount });
        outputAmount += BigInt(instr.params.amount);
        break;
      case InstructionBuilderTypes.TokenTransfer:
        outputs.push({
          address: instr.params.toAddress,
          amount: instr.params.amount,
          tokenName: instr.params.tokenName,
        });
        break;
      case InstructionBuilderTypes.CreateNonceAccount:
        outputs.push({ address: instr.params.nonceAddress, amount: instr.params.amount });
        outputAmount += BigInt(instr.params.amount);
        break;
      case InstructionBuilderTypes.StakingActivate:
        outputs.push({ address: instr.params.stakingAddress, amount: instr.params.amount });
        outputAmount += BigInt(instr.params.amount);
        break;
      case InstructionBuilderTypes.StakingWithdraw:
        outputs.push({ address: instr.params.fromAddress, amount: instr.params.amount });
        outputAmount += BigInt(instr.params.amount);
        break;
      case InstructionBuilderTypes.CreateAssociatedTokenAccount:
        tokenEnablements.push({
          address: instr.params.ataAddress,
          tokenName: instr.params.tokenName,
          tokenAddress: instr.params.mintAddress,
        });
        break;
      case InstructionBuilderTypes.Memo:
        memo = instr.params.memo;
        break;
    }
  }
  return { outputs, tokenEnablements, outputAmount, memo };
}
```

All three call sites then become thin wrappers that call this, eliminating the duplication. The legacy `Transaction.explainTransaction()` can convert the `bigint` result to `BigNumber` at the boundary if needed (or migrate to `bigint` too).

modules/sdk-coin-sol/src/sol.ts
OttoAllmendinger: *🤖 Claude here.*

Yes, `bigint` for amounts. The `ExplainTransactionOptions.feeInfo.fee` and `tokenAccountRentExemptAmount` come in as `string` from the external interface (`TransactionFee.fee` is `string` in sdk-core), so we should convert exactly once at the boundary.

Proposal — parse external args at the top, then pass typed values into a pure inner function:

```typescript
interface WasmExplainArgs {
  combinedInstructions: InstructionParams[];
  transactionType: TransactionType;
  memo?: string;
  nonce: string;
  durableNonce?: DurableNonceParams;
  numSignatures: number;
  signatures: Uint8Array[];
  lamportsPerSignature: bigint;
  rentPerAta: bigint;
}

function buildWasmExplanation(args: WasmExplainArgs): SolLibTransactionExplanation {
  const { outputs, tokenEnablements, outputAmount } = extractOutputsFromInstructions(args.combinedInstructions);
  const signatureFee = args.lamportsPerSignature * BigInt(args.numSignatures);
  const rentFee = args.rentPerAta * BigInt(tokenEnablements.length);
  const totalFee = (signatureFee + rentFee).toString();
  const txId = getTransactionId(args.signatures);

  return {
    displayOrder: [ ... ],
    id: txId,
    type: TransactionType[args.transactionType],
    changeOutputs: [],
    changeAmount: '0',
    outputAmount: outputAmount.toString(),
    outputs,
    fee: { fee: totalFee, feeRate: Number(args.lamportsPerSignature) },
    memo: args.memo,
    blockhash: args.nonce,
    durableNonce: args.durableNonce,
    tokenEnablements,
  };
}
```

Then `explainTransactionWithWasm` becomes a thin adapter:

```typescript
explainTransactionWithWasm(params: ExplainTransactionOptions): SolLibTransactionExplanation {
  const txBytes = Buffer.from(params.txBase64, 'base64');
  const wasmTx = WasmTransaction.fromBytes(txBytes);
  const parsed = wasmParseTransaction(txBytes);
  const { instructions, transactionType } = combineWasmInstructionsFromBytes(txBytes);
  const memo = parsed.instructionsData.find((i) => i.type === 'Memo')?.memo;

  return buildWasmExplanation({
    combinedInstructions: instructions,
    transactionType,
    memo,
    nonce: parsed.nonce,
    durableNonce: parsed.durableNonce,
    numSignatures: parsed.numSignatures,
    signatures: wasmTx.signatures,
    lamportsPerSignature: BigInt(params.feeInfo?.fee || '0'),
    rentPerAta: BigInt(params.tokenAccountRentExemptAmount || '0'),
  });
}
```

String → bigint conversion happens once at the boundary, all arithmetic is native `bigint`, and `.toString()` only at the final return where the external interface requires `string`.

modules/sdk-coin-sol/src/sol.ts
OttoAllmendinger: *🤖 Claude here.*

Agreed — should use the existing `UNAVAILABLE_TEXT` constant from `lib/constants.ts` instead of a string literal. Same constant is already used in `WasmTransaction` for the same purpose.

modules/sdk-coin-sol/src/sol.ts
OttoAllmendinger: *🤖 Claude here.*

Agreed. `TransactionType` is a numeric enum, so TypeScript's reverse mapping already gives us string names. The entire `mapTransactionTypeToString` method can be replaced with:

```typescript
TransactionType[transactionType]
```

This is exactly what the legacy `Transaction` class does at transaction.ts:645:
```typescript
type: TransactionType[this.type].toString(),
```

No switch statement or coin method needed. This should be a standalone expression, not a method on `Sol`.

OttoAllmendinger (dismissed):

benkelcher-bitgo (approved):

margueriteblair (approved):

zahin-mohammad (approved):

modules/sdk-coin-sol/src/lib/wasm/transaction.ts
lcovar: That's the goal, no, or at least less js deps for blockchain related functions.

OttoAllmendinger (approved):

mmcshinsky-bitgo (approved):

PR #7960, opened 2026-01-21
fix: address validation for SMC wallets

mrdanish26:
TICKET: WP-7507

**Problem**
Address validation failed for Self-Managed Custodial (SMC) TSS wallets on EdDSA-based coins (e.g., Solana). The client-side verification used m/{index} while the wallet derivation used m/{prefix}/{index}, causing a mismatch.

**Fix**
Extract the derivation prefix from the User keychain's derivedFromParentWithSeed for SMC (cold) TSS wallets
Pass the prefix to address verification to construct the correct derivation path
Use m/{prefix}/{index} when a prefix is present, otherwise m/{index}

**Changes**
* wallet.ts: Extract derivationPrefix from User keychain for SMC TSS wallets
* iBaseCoin.ts: Add optional derivationPrefix to TssVerifyAddressOptions and VerifyAddressOptions
* addressVerification.ts: Use derivationPrefix when constructing the derivation path
* Tests: Add unit tests for derivation prefix extraction and path construction

**Testing**
* Added unit tests for SMC wallet prefix extraction
* Added unit tests for derivation path construction with/without prefix

**Impact**
* Fixes address validation for SMC TSS wallets
* Backward compatible (prefix is optional)
* No impact on custodial or non-TSS wallets

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/addressVerification.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/addressVerification.ts
@@ -1,7 +1,6 @@
 import { Ecdsa } from '../../../account-lib/mpc';
 import { TssVerifyAddressOptions } from '../../baseCoin/iBaseCoin';
 import { InvalidAddressError } from '../../errors';
-import { EDDSAMethods } from '../../tss';
 
 /**
  * Extracts and validates the commonKeychain from keychains array.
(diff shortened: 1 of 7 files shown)


Marzooqa:
Changes the commit `2730531955bbfc9a98b526c9acb91e1641b4b824`:
Move derivation prefix computation inside `verifyMPCWalletAddress` so callers only need to pass `derivedFromParentWithSeed`. This enables any endpoint / method to verify SMC wallet addresses without computing the prefix themselves.

modules/sdk-core/src/bitgo/utils/tss/addressVerification.ts
```diff
@@ -65,15 +66,21 @@ export async function verifyMPCWalletAddress(
...
   }
 
   const MPC = params.keyCurve === 'secp256k1' ? new Ecdsa() : await EDDSAMethods.getInitializedMpcInstance();
   const commonKeychain = extractCommonKeychain(keychains);
-  const derivedPublicKey = MPC.deriveUnhardened(commonKeychain, 'm/' + index);
+
+  // Compute derivation path:
+  // - For SMC wallets with derivedFromParentWithSeed, compute prefix and use: {prefix}/{index}
+  // - For other wallets, use simple path: m/{index}
+  const prefix = derivedFromParentWithSeed ? getDerivationPath(derivedFromParentWithSeed.toString()) : undefined;
```
kaustubhbitgo, line 81: We should also add test cases for this function. Right now all tests verify only the create address verification data. We don't have to mock anything let's add test cases that actually generates an address.

Marzooqa: For the `getDerivationPath` method, we have existing tests ([Link](https://github.com/BitGo/BitGoJS/blob/d8ab3df38c7f0dc445117f68340cd3f17dfc9a68/modules/sdk-lib-mpc/test/unit/util.ts#L85))

The `verifyMPCWalletAddress` method actually generates the address and checks if the address returned from wp is the same. So we have added tests here ([link](https://github.com/BitGo/BitGoJS/pull/7960/changes#diff-7b9984a75455ee51dd091d91d8ea1e08fe94c45d0ae0315e3d5747eb2db6d374R3482-R3495))

ranga-r2 (approved):

parasgarg-bitgo (approved):

PR #7981, opened 2026-01-28
feat(express): add derivedFromParentWithSeed for isWalletAddress

kaustubhbitgo:
Ticket: WP-7522


adityasoni-bitgo (approved):

Marzooqa (commented):
Can you add validation to make sure this field is present when walletType is "cold" ie, SMC wallets and isn't required otherwise?

PR #7982, opened 2026-01-28
feat(ada): Token sponsorship transaction sandboxing script

krupanand-bitgo:
ADA token sponsorship transactions sandboxing script
Ticket: WIN-8328

### DIFF
--- a/examples/ts/ada/adaTokenSponsorshipTransaction.ts
+++ b/examples/ts/ada/adaTokenSponsorshipTransaction.ts
@@ -1,0 +1,464 @@
+/**
+ * ADA Token Sponsorship Transaction Script (using BitGo SDK)
+ *
+ * Builds and broadcasts a sponsored token transfer where:
+ * - Sender sends a specific amount of a token to recipient
+ * - Remaining tokens and ADA go back to sender as change
+ * - Fee is paid by a separate sponsor address
+ * - Fee change goes back to sponsor address
+ * - Both sender and sponsor sign the transaction
+ *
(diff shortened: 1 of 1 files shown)


examples/ts/ada/adaTokenSponsorshipTransaction.ts
```diff
@@ -0,0 +1,464 @@
...
+    '',
+  sponsorAddress:
+    '',
+  // Recipient (receives token)
+  recipientAddress:
+    '',
+  // Token to transfer
+  tokenToSend: {
+    policyId: '', // Will be auto-detected from sender's UTXOs
+    assetName: '5741544552', // "WATER" in hex (uppercase)
```
mukeshsp, line 40: Can we add a reference for this? That way, if we need to perform a transaction for another token, we can use the reference to retrieve the token hex.

examples/ts/ada/adaTokenSponsorshipTransaction.ts
```diff
@@ -0,0 +1,464 @@
...
+/**
+ * Build and sign a sponsored token transaction using BitGo SDK
+ *
+ * Transaction structure:
+ * - Inputs: [sender UTXOs with token] + [sponsor UTXOs]
+ * - Outputs: [token to recipient] + [token change to sender] + [ADA change to sender] + [sponsor fee change]
+ */
+async function buildTokenSponsorshipTransaction() {
+  // Initialize BitGo SDK factory
+  const factory = new TransactionBuilderFactory(coins.get('tada'));
```
mukeshsp, line 64: Can we define the coin name as a constant so it’s easy to switch between testnet and mainnet?

PR #7984, opened 2026-01-28
feat(statics): add dYdX L1 chain support

abhijeet8986:
Add dYdX as a Cosmos-based L1 blockchain using the shared Cosmos SDK pattern, similar to Kava. This enables offline operations including address generation, transaction building, and validation.

TICKET: COIN-7422

Sandbox-PR: [Link](https://github.com/BitGo/coins-sandbox/pull/709)


modules/statics/src/allCoinsAndTokens.ts
```diff
@@ -1037,6 +1037,26 @@ export const allCoinsAndTokens = [
     BaseUnit.KAVA,
     [...COSMOS_SIDECHAIN_FEATURES, CoinFeature.SHARED_COSMOS_SDK, CoinFeature.SHARED_COSMOS_WP]
   ),
+  account(
+    'f5a8c3d1-7b2e-4f9a-8c6d-1e3b5a7c9d2f',
+    'dydxcosmos',
```
baltiyal, line 1042: do you have product confirmation on the symbol and name for the chain?
Bcoz I checked the ticket and their is no details and product sign-off for that.

modules/statics/test/unit/coins.ts
```diff
@@ -878,7 +878,7 @@ coins.forEach((coin, coinName) => {
     });
 
     if (!coin.isToken && coin.family !== CoinFamily.FIAT) {
-      if (coin.family !== CoinFamily.THOR) {
+      if (coin.family !== CoinFamily.THOR && coin.family !== CoinFamily.DYDX) {
```
baltiyal, line 881: Why a specific condition addition for dydx here?

modules/statics/src/allCoinsAndTokens.ts
abhijeet8986: Here's the ticket for symbol: [COIN-7282](https://bitgoinc.atlassian.net/browse/COIN-7282?selectedIssue=COIN-7409)

modules/statics/test/unit/coins.ts
abhijeet8986: The exception is needed because the dYdX Cosmos chain uses dydxcosmos/tdydxcosmos as coin names (per the asset specification) to avoid a naming conflict with the existing ERC20 token dydx already in the codebase.
The test expects coin.name === coin.family for mainnet coins. Since dydxcosmos !== dydx (the family), the test would incorrectly expect the mainnet coin to be TESTNET.
This follows the same pattern as CoinFamily.THOR which also has this exception for similar naming reasons.

baltiyal: @at31416 / @bhavidhingra could you please check this condition.
I feel this is not similar condition as thor and this check is not needed.
If there is some handling needed that would not be here.

bhavidhingra: @baltiyal it looks fine because it seems that for mainnet the name and family are not the same for dydx.

bhavidhingra (approved):

PR #8035, opened 2026-02-03
feat: integrate user change password batching flow

kisslove-dewangan:
TICKET: WP-7566

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -1932,9 +1932,15 @@ export class BitGoAPI implements BitGoBase {
       password: this.calculateHMAC(user.username, newPassword),
     };
 
+    // Calculate payload size in KB
+    const payloadSizeBytes = JSON.stringify(updatePasswordParams).length;
+    const payloadSizeKB = Math.ceil(payloadSizeBytes / 1024);
+
     // Check if batching flow is enabled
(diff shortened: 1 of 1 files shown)


modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -1932,6 +1932,30 @@ export class BitGoAPI implements BitGoBase {
       password: this.calculateHMAC(user.username, newPassword),
     };
 
+    // Check if batching flow is enabled
+    try {
+      const batchingFlowCheck = await this.get(this.url('/user/checkBatchingPasswordFlow', 2)).result();
```
vibhavgo: This endpoint (`'/user/checkBatchingPasswordFlow'`) expects a `payloadSize` parameter based on the updatePasswordParams
Suggested fix:

```
    // Calculate payload size in KB
    const payloadSizeBytes = JSON.stringify(updatePasswordParams).length;
    const payloadSizeKB = Math.ceil(payloadSizeBytes / 1024);

    // Check if batching flow is enabled
    try {
      const batchingFlowCheck = await this.get(this.url('/user/checkBatchingPasswordFlow', 2))
        .query({ payloadSize: payloadSizeKB.toString() })
        .result();
```

kaustubhbitgo (approved):
I assume we will add tests later, approving for now.

PR #8051, opened 2026-02-04
Fix consolidate

AS195-sketch:
Issue Ticket: [WIN-7796](https://bitgoinc.atlassian.net/browse/WIN-7796)

**Issue:** insufficient funds error was thrown despite addresses having sufficient funds.

- occured as the check for balance was compared with 10^7 lovelace (10 ADA) and not 10^6 lovelace (1 ADA)
- error thrown even if a single address had insufficient (balance - fees) even if wallet had sufficient funds.
- "missing wallet passphrase" error thrown as wallet passphrase is always excluded in input parameters

**Fixes:**

- changed check from 10^7 to 10^6 lovelace
- skipped addresses with insufficient (balance - fees) to continue the consolidation
- refactored isUnsignedSweep logic to depend only on the presence of walletPassphrase as checks for user and backup key are already handled

[WIN-7796]: https://bitgoinc.atlassian.net/browse/WIN-7796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/sdk-coin-ada/CHANGELOG.md
+++ b/modules/sdk-coin-ada/CHANGELOG.md
@@ -5,221 +5,119 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 # [4.21.0](https://github.com/BitGo/BitGoJS/compare/@bitgo/sdk-coin-ada@4.20.8...@bitgo/sdk-coin-ada@4.21.0) (2026-01-22)
 
-
 ### Features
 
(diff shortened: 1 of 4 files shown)


mullapudipruthvik (commented):
please have a commit message with more details
1. issue
2. how are you solving it
also change log would be auto generated afaik

mullapudipruthvik (approved):

PR #8059, opened 2026-02-04
feat(abstract-cosmos,sdk-coin-hash): support group proposals

sreerajs:
Adding support for Figure Markets to integrate bitgo wallets to their tokenized equity exchange

- Add support for base64-encoded pre-encoded messages in ContractCallBuilder
- Add group proposal message support (MsgSubmitProposal)
- Add ProposalCompiled protobuf definitions
- Add ContractCallBuilder tests with pre-encoded messages

The exchange module innermsg is wrapped inside a GroupProposal, hence we do not fully support decoding the messages to identify what asset is being transferred. This will be added in followup PRs. This is to exihit bitgo's ability to sign contract interactions

Ticket: WIN-8842


modules/abstract-cosmos/src/lib/protobuf-init.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Initialize and patch protobuf modules to ensure proper namespace binding
+ */
+export function initializeProtobufModules(): { MsgCompiled: any; ProposalCompiled: any; unifiedRoot: any } {
+  const MsgCompiled = require('../../resources/MsgCompiled');
+
+  // Store cosmos.base before loading ProposalCompiled (both modules share the same $root object)
+  const cosmosBase = MsgCompiled.cosmos?.base;
+  const ProposalCompiled = require('../../resources/ProposalCompiled');
```
veetragjain, line 9: Any reason for having dynamic imports in this function?

veetragjain (approved):

modules/abstract-cosmos/src/lib/protobuf-init.ts
sreerajs: @veetragjain to control loading order and prevent namespace collision
Both modules share the same protobuf $root object, and ProposalCompiled overwrites MsgCompiled.cosmos.base

PR #8063, opened 2026-02-05
feat(flrp): add TxBuilder for P-chain staking

joshisakshi:
TICKET: SC-5233

Implement transaction builder for AddPermissionlessDelegator transactions on Flare P-chain, enabling stake delegation to validators.

Key changes:
- Add PermissionlessDelegatorTxBuilder with full build and initBuilder support
- Use post-Etna API (pvm.e.newAddPermissionlessDelegatorTx) exclusively
- Require feeState parameter (consistent with Import/Export builders)
- Add AddPermissionlessDelegatorTx to SerializedTx type union
- Add FlrpStakingOptions interface for delegation parameters
- Add transaction type verification in flrp.ts
- Comprehensive unit tests for builder validation

Note: Rewards always go to the C-chain address derived from the delegator's public key though it does not affect onchain reward distribution and rewardAddress param is kept for rewards tracking


modules/sdk-coin-flrp/src/lib/permissionlessDelegatorTxBuilder.ts
```diff
@@ -0,0 +1,314 @@
...
+    this._endTime = validator.endTime.value();
+    this._stakeAmount = validator.weight.value();
+
+    // Extract from addresses from stake outputs
+    // In delegation transactions, stake outputs contain the owner addresses
+    const stakeOutputs = delegatorTx.stake;
+    if (stakeOutputs.length > 0) {
+      const firstStakeOutput = stakeOutputs[0];
+      // StakeableLockOut has outputOwners directly accessible
+      const output = firstStakeOutput.output as unknown as {
```
Gautam2305: is there a type here that can be used instead of `unknown`?

modules/sdk-coin-flrp/src/flrp.ts
```diff
@@ -111,6 +111,10 @@ export class Flrp extends BaseCoin {
           this.validateImportTx(explainedTx.inputs, params.txParams);
         }
         break;
+      case TransactionType.AddPermissionlessDelegator:
+        // Validate delegation transaction
+        this.validateDelegationTx(params.txParams);
+        break;
```
Gautam2305, line 117: looking at how `export` and `import` transactions are validated - do we need to check `explainedTx` alongside `txParams` here?

modules/sdk-coin-flrp/src/flrp.ts
```diff
@@ -166,6 +170,40 @@ export class Flrp extends BaseCoin {
...
+    if (!durationSeconds) {
+      throw new Error('Delegation transaction requires durationSeconds');
+    }
+
+    if (!rewardAddress) {
+      throw new Error('Delegation transaction requires rewardAddress');
+    }
+
+    // Validate nodeID format
+    if (!nodeID.startsWith('NodeID-')) {
```
Gautam2305: can we use the `validateNodeID` method used in `permissionlessDelegatorTxBuilder` ? if yes, can you make that a util method please?

modules/sdk-coin-flrp/src/lib/permissionlessDelegatorTxBuilder.ts
joshisakshi: updated

modules/sdk-coin-flrp/src/flrp.ts
joshisakshi: done

modules/sdk-coin-flrp/src/flrp.ts
joshisakshi: Yes, addressed. The validateDelegationTx method now accepts both txParams and explainedTx

modules/sdk-coin-flrp/src/lib/permissionlessDelegatorTxBuilder.ts
```diff
@@ -0,0 +1,335 @@
...
+  protected get transactionType(): TransactionType {
+    return TransactionType.AddPermissionlessDelegator;
+  }
+
+  // Validation methods
+  validateLocktime(locktime: bigint): void {
+    if (locktime < BigInt(0)) {
+      throw new BuildTransactionError('Invalid transaction: locktime must be 0 or higher');
+    }
+  }
```
Vijay-Jagannathan: whats the upper limit ?

modules/sdk-coin-flrp/src/lib/permissionlessDelegatorTxBuilder.ts
```diff
@@ -0,0 +1,335 @@
...
+   *
+   * NOTE: On Flare P-chain, this parameter is accepted by the API but does NOT
+   * control where rewards are actually sent. Rewards always go to the C-chain
+   * address derived from the delegator's public key.
+   * @param address The C-chain reward address (for tracking)
+   */
+  rewardAddresses(address: string | string[]): this {
+    const rewardAddresses = address instanceof Array ? address : [address];
+    this.transaction._rewardAddresses = rewardAddresses.map(utils.parseAddress);
+    return this;
```
Vijay-Jagannathan: why is it `rewardAddresses` and not `rewardAddress` if its always tied to a single C chain reward address

modules/sdk-coin-flrp/src/lib/utils.ts
```diff
@@ -127,6 +127,51 @@ export class Utils implements BaseUtils {
...
+      return false;
+    }
+  }
+
+  /**
+   * Validates a NodeID format and throws if invalid
+   * @param {string} nodeID - NodeID to validate (e.g., "NodeID-xxx")
+   * @throws {InvalidTransactionError} if nodeID is invalid
+   */
+  validateNodeID(nodeID: string): void {
```
Vijay-Jagannathan: saw the same method in another place (maybe for a different usecase, in which case these should be named differently) 

or if it can be consolidated, make use of a single method

modules/sdk-coin-flrp/src/lib/utils.ts
```diff
@@ -127,6 +127,51 @@ export class Utils implements BaseUtils {
     return /^(0x){0,1}([0-9a-f])+$/i.test(str);
   }
 
+  /**
+   * Validates a NodeID format
+   * @param {string} nodeID - NodeID to validate (e.g., "NodeID-xxx")
+   * @returns {boolean} - true if valid, false otherwise
+   */
+  isValidNodeID(nodeID: string): boolean {
```
Vijay-Jagannathan, line 139: this method also does the same thing as the below one

Vijay-Jagannathan (commented):
few comments

modules/sdk-coin-flrp/src/lib/permissionlessDelegatorTxBuilder.ts
joshisakshi: Renamed to `rewardAddress`

modules/sdk-coin-flrp/src/lib/permissionlessDelegatorTxBuilder.ts
joshisakshi: Removed locktime entirely from the delegation builder. Unlike AVAXP which creates SECPOwnerOutput with locktime for reward UTXOs, Flare does not have reward UTXOs on P-chain but FlareJS API handles locktime internally with default 0.

modules/sdk-coin-flrp/src/lib/utils.ts
joshisakshi: Consolidated as follows:
- Removed utils.validateNodeID() but kept utils.isValidNodeID() 
- Builder methods (permissionlessDelegatorTxBuilder.validateNodeID()) now call utils.isValidNodeID() internally and throw BuildTransactionError with specific messages

modules/sdk-coin-flrp/src/lib/utils.ts
```diff
@@ -127,6 +127,26 @@ export class Utils implements BaseUtils {
...
+    if (!nodeID) {
+      return false;
+    }
+    if (!nodeID.startsWith('NodeID-')) {
+      return false;
+    }
+    try {
+      const decoded = bs58.decode(nodeID.slice(7));
+      return decoded.length === 24;
```
Vijay-Jagannathan: nit, non-blocking: these (`NodeId-`, `24`) can be constants

Vijay-Jagannathan (dismissed):
lgtm

abhijit0943 (changes requested):
Hey Sakshi, nice work on the builder — the overall structure looks solid. One issue though:

`transaction.ts` wasn't updated in this PR. The `setTransactionType()` method (line 379) only allows `AddPermissionlessValidator` in its whitelist, so when the delegation builder calls `setTransactionType(AddPermissionlessDelegator)` during `buildImplementation()`, it'll throw "Transaction type is not supported". The `outputs`, `changeOutputs`, `inputs`, and `explainTransaction` getters also need cases for the delegation tx type.

The reason the existing tests don't catch this is that every `build()` test expects a rejection before it ever gets to `setTransactionType` (missing nodeID, missing UTXOs, etc.). There's no test that actually builds a complete valid delegation transaction end-to-end.

Would be good to:
1. Update `transaction.ts` to support `AddPermissionlessDelegator` in `setTransactionType` and the output/input/explain getters
2. Add the delegation type to the `Tx` union in `iface.ts` (it was added to `SerializedTx` but not `Tx`)
3. Add an end-to-end build test with all required params, and ideally an `initBuilder` round-trip test (build -> serialize -> `factory.from()` -> verify)

abhijit0943 (approved):
Looks good — all the issues from the previous review are addressed. transaction.ts is updated, Tx type union is fixed, and the round-trip tests with real tx data cover the deserialization path. Ship it.

PR #8065, opened 2026-02-05
feat: capture raw body bytes for v4 hmac calculation

MohammedRyaan786:
TICKET: CAAS-659

This pull request improves how HTTP request bodies are handled in the Express app, specifically to support v4 HMAC authentication that requires the exact raw request bytes. The main changes add logic to preserve the raw request body and ensure it is sent unchanged when proxying requests, which is important for signature verification.

**Enhancements for request body handling and HMAC authentication:**

* The Express app now stores the raw request body buffer on each request before JSON parsing, enabling downstream code to access the exact bytes received from the client. This is necessary for v4 HMAC authentication, which relies on the unmodified request body.
* A new `sendRequestBody` helper function is introduced in `clientRoutes.ts` to send the raw request body buffer when available, falling back to the parsed body for backward compatibility.
* The `redirectRequest` function is updated to use `sendRequestBody` for all HTTP methods that send a body, ensuring the raw bytes are forwarded when possible.

Current PR is related to this [BGMS PR](https://github.com/BitGo/bitgo-microservices/pull/55329)

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -1193,6 +1193,38 @@ async function handleNetworkV1EnterpriseClientConnections(
 }
 
 /**
+ * Helper to send request body, using raw bytes when available.
+ *
+ * For v4 HMAC authentication, we need to send the exact bytes that were
+ * received from the client to ensure the HMAC signature matches.
+ * The rawBodyBuffer is captured by body-parser's verify callback before
+ * JSON parsing, preserving exact whitespace, key ordering, etc.
+ *
+ * For v2/v3, sending the raw string also works because serializeRequestData
+ * now properly returns strings as-is for HMAC calculation.
+ *
+ * @param request - The superagent request object
+ * @param req - The Express request containing body and rawBodyBuffer
+ * @returns The request with body attached
+ */
+function sendRequestBody(request: ReturnType<BitGo['post']>, req: express.Request) {
+  if (req.rawBodyBuffer) {
+    // Preserve original Content-Type header from client
+    const contentTypeHeader = req.headers['content-type'];
+    if (contentTypeHeader) {
+      request.set('Content-Type', Array.isArray(contentTypeHeader) ? contentTypeHeader[0] : contentTypeHeader);
+    }
+    // Send raw body as UTF-8 string to preserve exact bytes for HMAC.
+    // JSON is always UTF-8 (RFC 8259), so this is lossless for JSON bodies.
+    // serializeRequestData will return this string as-is for HMAC calculation.
+    return request.send(req.rawBodyBuffer.toString('utf8'));
+  }
+
+  // Fall back to parsed body for backward compatibility (e.g., non-JSON bodies)
+  return request.send(req.body);
+}
+
+/**
  * Redirect a request using the bitgo request functions.
  * @param bitgo
  * @param method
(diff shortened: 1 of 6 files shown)


MohammedRyaan786:
@claude please review this PR

lokesh-bitgo:
@claude review this PR with a primary focus on the bug, and confirm whether these changes impact any other flows.

modules/express/src/clientRoutes.ts
```diff
@@ -1192,6 +1192,27 @@ async function handleNetworkV1EnterpriseClientConnections(
...
+ * @param req - The Express request containing body and rawBodyBuffer
+ * @returns The request with body attached
+ */
+function sendRequestBody(request: ReturnType<BitGo['post']>, req: express.Request) {
+  if (req.rawBodyBuffer) {
+    const contentTypeHeader = req.headers['content-type'];
+    if (contentTypeHeader) {
+      request.set('Content-Type', Array.isArray(contentTypeHeader) ? contentTypeHeader[0] : contentTypeHeader);
+    }
+    return request.send(req.rawBodyBuffer.toString('utf8'));
```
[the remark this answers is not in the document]
MohammedRyaan786: Then serializeRequestData function checks:
`if (typeof data !== 'string') {  // Buffer is NOT a string → TRUE  serialize = superagent.serialize['application/json'];  data = serialize(data);  // Converts Buffer to {"type":"Buffer","data":[123,34,...]}}`
Result: The body becomes {"type": @copilot @copilot "Buffer","data":[123,34,97,100,...]} leading to HMAC mismatch

modules/sdk-api/src/api.ts
```diff
@@ -121,28 +121,61 @@ function createResponseErrorString(res: superagent.Response): string {
...
+
+  if (data === undefined || data === null) {
+    return undefined;
+  }
+
+  // Serialize object data based on Content-Type
+  let contentType = req.get('Content-Type');
+  // Parse out just the content type from the header (ignore the charset)
+  if (contentType) {
+    contentType = contentType.split(';')[0].trim();
```
pranavjain97, line 156: This catch just re-throws, safe to remove the try/catch entirely.

modules/sdk-api/src/api.ts
```diff
@@ -121,28 +121,61 @@ function createResponseErrorString(res: superagent.Response): string {
...
-    if (contentType) {
-      contentType = contentType.split(';')[0];
-    }
-    let serialize = superagent.serialize[contentType];
-    if (!serialize && /[\/+]json\b/.test(contentType)) {
-      serialize = superagent.serialize['application/json'];
+  const data: string | Buffer | Record<string, unknown> | undefined = (req as any)._data;
+
+  if (typeof data === 'string') {
+    return data;
```
pranavjain97, line 134: Mutating `_data` from Buffer to string means if `serializeRequestData` runs twice (retries?), the second call gets a string instead of a Buffer, different HMAC. Worth a guard or a comment confirming single-call assumption.

MohammedRyaan786: The function is actually already idempotent and safe for retries, so the HMAC will be identical on every call:

First call: _data is a Buffer → converts to UTF-8 string, mutates _data to that string, returns it → HMAC calculated on '{"address":"tb1qtest"}'

Second call (retry): _data is now a string → hits the early return → returns the same string as-is → HMAC calculated on '{"address":"tb1qtest"}'

Both calls return the exact same string, so the HMAC is identical.

The mutation itself is also required. Without it, superagent serializes a Buffer as {"type":"Buffer","data":[123,...]} instead of the actual content, which would cause an HMAC mismatch with what's sent over the wire.
I've added a comment explaining both the retry safety and why the mutation is necessary.

Thanks for catching it!

pranavjain97 (approved):
🚢

PR #8072, opened 2026-02-06
feat(sdk-coin-sol): upgrade @solana/spl-token from 0.3.1 to 0.4.9

lcovar:
## Why this upgrade is needed

The current @solana/spl-token@0.3.1 has a broken ESM build - the .mjs files have incorrect import paths that reference directories instead of files (e.g., `./memoTransfer.mjs` instead of `./memoTransfer/index.mjs`).

This blocks webpack browser bundles from using ESM resolution for @bitgo/sdk-coin-sol, which is required for @bitgo/wasm-solana support (WASM modules need ESM for proper async initialization in browsers).

## What changed in 0.4.x

1. Fixed ESM imports - uses .js files with "type": "module" and correct paths like `./memoTransfer/index.js`

2. Removed SYSVAR_RENT_PUBKEY from Associated Token Account instructions. This is a valid Solana protocol change - the rent sysvar was deprecated because the Solana runtime now provides rent information automatically. Transactions are now more efficient (6 accounts instead of 7).

## Test fixture updates

27 test fixtures in test/resources/sol.ts were regenerated to reflect the new transaction serialization. The transactions are functionally equivalent and valid on-chain.

Ticket: BTC-3009


Vijay-Jagannathan (commented):
im assuming that bumping this change in WP could potentially cause some test failures ? anything else that could affect ?

Vijay-Jagannathan (approved):
lgtm

therealdwright (approved):

therealdwright (commented):
Sign your commit please.

PR #8089, opened 2026-02-09
feat(sdk-coin-iota): add recovery and consolidation support

shubham-damkondwar:
This covers the main changes: adding recover, recoverConsolidations, and sweep methods to the IOTA coin module, along with related constants, utility functions, and environment configuration.

TICKET: WIN-8627

### DIFF
--- a/modules/sdk-coin-iota/src/iota.ts
+++ b/modules/sdk-coin-iota/src/iota.ts
@@ -2,25 +2,36 @@ import {
   AuditDecryptedKeyParams,
   BaseCoin,
   BitGoBase,
+  EDDSAMethods,
+  EDDSAMethodTypes,
+  Environments,
   KeyPair,
-  ParsedTransaction,
-  SignTransactionOptions,
-  SignedTransaction,
-  VerifyTransactionOptions,
+  MPCAlgorithm,
+  MPCConsolidationRecoveryOptions,
+  MPCRecoveryOptions,
+  MPCSweepTxs,
+  MPCTx,
+  MPCTxs,
+  MPCType,
+  MPCUnsignedTx,
   MultisigType,
   multisigTypes,
-  MPCAlgorithm,
-  TssVerifyAddressOptions,
-  MPCType,
+  ParsedTransaction,
   PopulatedIntent,
   PrebuildTransactionWithIntentOptions,
+  RecoveryTxRequest,
+  SignedTransaction,
+  SignTransactionOptions,
   TransactionRecipient,
+  TransactionType,
+  TssVerifyAddressOptions,
   verifyEddsaTssWalletAddress,
+  VerifyTransactionOptions,
 } from '@bitgo/sdk-core';
 import { BaseCoin as StaticsBaseCoin, CoinFamily, coins } from '@bitgo/statics';
 import utils from './lib/utils';
-import { KeyPair as IotaKeyPair, Transaction, TransactionBuilderFactory } from './lib';
-import { auditEddsaPrivateKey } from '@bitgo/sdk-lib-mpc';
+import { KeyPair as IotaKeyPair, Transaction, TransactionBuilder, TransactionBuilderFactory } from './lib';
+import { auditEddsaPrivateKey, getDerivationPath } from '@bitgo/sdk-lib-mpc';
 import BigNumber from 'bignumber.js';
 import * as _ from 'lodash';
 import {
@@ -28,8 +39,24 @@ import {
   IotaParseTransactionOptions,
   TransactionExplanation,
   TransferTxData,
+  TransactionObjectInput,
 } from './lib/iface';
 import { TransferTransaction } from './lib/transferTransaction';
+import {
+  DEFAULT_GAS_OVERHEAD,
+  DEFAULT_SCAN_FACTOR,
+  MAX_GAS_BUDGET,
+  MAX_GAS_OBJECTS,
+  MAX_OBJECT_LIMIT,
+} from './lib/constants';
+
+export interface IotaRecoveryOptions extends MPCRecoveryOptions {
+  fullnodeRpcUrl?: string; // Override default RPC URL
+}
+
+interface IotaObjectWithBalance extends TransactionObjectInput {
+  balance: string;
+}
 
 /**
  * IOTA coin implementation.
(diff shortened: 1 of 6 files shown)


modules/sdk-coin-iota/src/iota.ts
```diff
@@ -253,6 +281,714 @@ export class Iota extends BaseCoin {
...
+    );
+  }
+
+  /**
+   * Checks whether the address has enough native IOTA balance to cover gas fees.
+   */
+  private async hasSufficientBalanceForGas(senderAddress: string, params: IotaRecoveryOptions): Promise<boolean> {
+    try {
+      const balance = await this.getBalance(senderAddress, params.fullnodeRpcUrl);
+      return balance > BigInt(MAX_GAS_BUDGET);
```
parasgarg-bitgo: MAX_GAS_BUDGET has a very high value. Instead get the exact gasEstimates - refer the flow for Consolidation tx building - [source](https://github.com/BitGo/indexerdb-microservice/blob/master/src/app/controllers/iota/consolidateHandler.ts#L94)

modules/sdk-coin-iota/src/iota.ts
```diff
@@ -253,6 +281,714 @@ export class Iota extends BaseCoin {
...
+      } catch (e) {
+        continue;
+      }
+
+      if (ownedObjects.length === 0) {
+        continue;
+      }
+
+      // Cap objects to prevent oversized transactions (IOTA max tx size = 128 KiB)
+      if (ownedObjects.length > MAX_OBJECT_LIMIT) {
```
parasgarg-bitgo: Since recovery is self-sponsored txn, the objects should be capped to MAX_GAS_OBJECTS

modules/sdk-coin-iota/src/lib/constants.ts
```diff
@@ -88,3 +88,29 @@ export const MAX_GAS_PRICE = 100000;
...
+ * We use 1280 as a safe limit, keeping room for recipients, gas data, etc.
+ * (IOTA protocol max_input_objects = 2048, max_tx_size_bytes = 131072)
+ */
+export const MAX_OBJECT_LIMIT = 1280;
+
+/**
+ * Maximum number of token objects to include in a token recovery transaction.
+ * Kept lower than MAX_OBJECT_LIMIT since token transactions also need gas objects.
+ */
+export const TOKEN_OBJECT_LIMIT = 512;
```
parasgarg-bitgo: Gas Objects have a separate limit in IOTA. No need to reduce the number of token objects to account for gas

modules/sdk-coin-iota/src/lib/utils.ts
```diff
@@ -173,6 +173,26 @@ export class Utils implements BaseUtils {
     const iotaPublicKey = new Ed25519PublicKey(Buffer.from(publicKey, 'hex'));
     return iotaPublicKey.toIotaAddress();
   }
+
+  // ========================================
+  // Recovery Validation Methods
+  // ========================================
+
+  validateNonNegativeNumber(defaultVal: number, errorMsg: string, inputVal?: number): number {
```
parasgarg-bitgo: nit: consider renaming to something like `getSafeNumber`

modules/sdk-coin-iota/src/iota.ts
```diff
@@ -253,6 +281,714 @@ export class Iota extends BaseCoin {
...
+   *
+   * @param params - Consolidation recovery parameters
+   * @param params.bitgoKey - The commonKeychain (combined TSS public key)
+   * @param params.startingScanIndex - Starting address index to scan (default: 1)
+   * @param params.endingScanIndex - Ending address index to scan (default: startingScanIndex + 20)
+   * @param params.walletPassphrase - Optional passphrase for signing (omit for unsigned transactions)
+   * @returns MPCTxs (signed) or MPCSweepTxs (unsigned) containing all consolidation transactions
+   * @throws Error if no addresses with funds are found in the scan range
+   */
+  async recoverConsolidations(params: MPCConsolidationRecoveryOptions): Promise<MPCTxs | MPCSweepTxs> {
```
parasgarg-bitgo, line 454: `recover()` is already doing everything that `recoverConsolidation()` is doing. Why do we need a separate method for recover consolidation?

modules/sdk-coin-iota/src/iota.ts
```diff
@@ -253,6 +281,714 @@ export class Iota extends BaseCoin {
...
+        if (BigInt(balance) > 0n) {
+          allObjects.push({ objectId, version: version.toString(), digest, balance });
+        }
+      }
+    }
+
+    if (pageCount >= MAX_PAGES) {
+      console.warn(`fetchOwnedObjects: Hit max page limit (${MAX_PAGES}) for ${address}`);
+    }
+    // Fallback: try without filter if no coins found
```
parasgarg-bitgo: Why do we need this fallback?

modules/sdk-coin-iota/src/iota.ts
```diff
@@ -253,6 +281,714 @@ export class Iota extends BaseCoin {
...
+
+    // Build temp transaction for estimation
+    const factory = this.getTxBuilderFactory();
+    const tempBuilder = factory.getTransferBuilder();
+    const estimationAmount = totalBalance > 0n ? '1' : '0';
+
+    tempBuilder
+      .sender(senderAddress)
+      .recipients([{ address: params.recoveryDestination, amount: estimationAmount }])
+      .gasData({ gasBudget: MAX_GAS_BUDGET, gasPrice, gasPaymentObjects: gasObjects });
```
parasgarg-bitgo: We don't need to set gasData in builder to estimate gas usage. The tx builder internally uses the default values if no gasData is provided.

modules/sdk-coin-iota/src/iota.ts
shubham-damkondwar: `recover() `stops at the first address with funds and sweeps to external address. 
while `recoverConsolidations()` iterates the entire scan range, collects all funded addresses, and sweeps them back to the base address (index 0)
recoverConsolidations() internally uses recover() only

modules/sdk-coin-iota/src/iota.ts
```diff
@@ -253,6 +280,666 @@ export class Iota extends BaseCoin {
...
+   * Builds funds recovery transaction(s) without BitGo
+   *
+   * @param {IotaRecoveryOptions} params parameters needed to construct and
+   * (maybe) sign the transaction
+   *
+   * @returns {MPCTx | MPCSweepTxs} array of the serialized transaction hex strings and indices
+   * of the addresses being swept
+   */
+  async recover(params: IotaRecoveryOptions): Promise<MPCTxs | MPCSweepTxs> {
+    if (!params.bitgoKey) {
```
dpkjnr, line 293: why do we need bitgoKey? is it the public key for wallet? if yes then can we rename the param to avoid confusion?

modules/sdk-coin-iota/src/iota.ts
```diff
@@ -253,6 +280,666 @@ export class Iota extends BaseCoin {
...
+    while (hasNextPage && pageCount < MAX_PAGES) {
+      pageCount++;
+      const query = {
+        filter: { StructType: normalizedCoinType },
+        options: { showContent: true, showType: true },
+      };
+
+      const response = await this.makeRpcCall(url, 'iotax_getOwnedObjects', [address, query, cursor, 50]);
+      const { data, nextCursor, hasNextPage: more } = response;
+      if (nextCursor === cursor) {
```
dpkjnr: what happens when there is only one page? what will nextCursor have in that case, if it will be null then we might not process the data

modules/sdk-coin-iota/src/iota.ts
```diff
@@ -253,6 +280,666 @@ export class Iota extends BaseCoin {
...
+          serializedTx,
+          signature: fullSignatureBase64,
+          coin: tokenCoin,
+        },
+      ],
+      lastScanIndex: idx,
+    };
+  }
+
+  private async signRecoveryTransaction(
```
dpkjnr, line 802: most of this function is duplicate of other coin classes, lets move it to some common util

alextse-bg (dismissed):
wallets-owned changes lgtm

modules/sdk-coin-iota/src/iota.ts
shubham-damkondwar: Yes it is BitGo Public Key.

modules/sdk-coin-iota/src/iota.ts
shubham-damkondwar: Thanks for pointing out. Fixed.

modules/sdk-coin-iota/src/iota.ts
shubham-damkondwar: We are using this same param for all other coins as well. Should I go ahead and rename it across all coins?

modules/sdk-coin-iota/src/iota.ts
shubham-damkondwar: I will make this change in a separate PR as it will make changes across 6-7 coins. I have created a separate ticket for that: https://bitgoinc.atlassian.net/browse/WIN-8974

dpkjnr (approved):

PR #8090, opened 2026-02-10
feat(statics): add testnet tokens

nvrakesh06:
TICKET: WIN-8887

### DIFF
--- a/modules/statics/src/tokenConfig.ts
+++ b/modules/statics/src/tokenConfig.ts
@@ -38,6 +38,7 @@ import {
   JettonToken,
   AccountCoin,
   CantonToken,
+  Tip20Token,
 } from './account';
 import { CoinFamily, CoinKind, BaseCoin, CoinFeature } from './base';
 import { coins } from './coins';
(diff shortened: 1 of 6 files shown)


modules/statics/src/tokenConfig.ts
```diff
@@ -1108,17 +1109,26 @@ const getFormattedJettonTokens = (customCoinMap = coins) =>
     return acc;
   }, []);
 
+function getTip20TokenConfig(coin: Tip20Token): Tip20TokenConfig {
+  return {
+    type: coin.name,
+    coin: coin.network.type === NetworkType.MAINNET ? 'tempo' : 'ttempo',
```
shubham-damkondwar, line 1121: neither the function name nor the file suggest us that this is tempo specific. Better we move it to tempo specific file.

nvrakesh06: TIP-20 is Tempo’s native token standard. While it’s currently tied to Tempo, we’re using the standard name here so that if other chains adopt TIP-20 in the future, they can be supported dynamically by switching the base coin.

MohammedRyaan786 (approved):

ArunBala-Bitgo (approved):

kamleshmugdiya (approved):

PR #8115, opened 2026-02-12
feat(sdk-coin-irys): add commitment transaction builder

abhijit0943:
Add the IrysCommitmentTransactionBuilder for building STAKE and PLEDGE commitment transactions. These are custom non-EVM transactions using 7-field RLP encoding with keccak256 prehash for HSM signing.

Includes:
- Module scaffolding (package.json, tsconfig, mocharc)
- Shared interface contract (iface.ts)
- Base58 and hex address utility functions (utils.ts)
- Commitment transaction builder with RLP encoding, anchor fetching, prehash computation, and broadcast payload creation

Ticket: SC-5392

### DIFF
--- a/modules/sdk-coin-irys/src/lib/commitmentTransactionBuilder.ts
+++ b/modules/sdk-coin-irys/src/lib/commitmentTransactionBuilder.ts
@@ -1,0 +1,253 @@
+import { RLP } from '@ethereumjs/rlp';
+import { arrayify, keccak256 } from 'ethers/lib/utils';
+import request from 'superagent';
+import {
+  CommitmentType,
+  CommitmentTypeId,
+  CommitmentTransactionFields,
+  CommitmentTransactionBuildResult,
+  EncodedSignedCommitmentTransaction,
+  EncodedCommitmentType,
+  AnchorInfo,
+  COMMITMENT_TX_VERSION,
+} from './iface';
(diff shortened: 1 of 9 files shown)


modules/sdk-coin-irys/src/lib/commitmentTransactionBuilder.ts
```diff
@@ -0,0 +1,253 @@
...
+ *   builder.setCommitmentType({ type: CommitmentTypeId.PLEDGE, pledgeCount: 0n });
+ */
+export class IrysCommitmentTransactionBuilder {
+  private _irysApiUrl: string;
+  private _chainId: bigint;
+  private _commitmentType?: CommitmentType;
+  private _fee?: bigint;
+  private _value?: bigint;
+  private _signer?: Uint8Array; // 20 bytes
+  private _anchor?: Uint8Array; // 32 bytes (set during build, or manually for testing)
```
raksha-r7, line 40: Shall we call this as `nonce` for simplicity and then send this field as `anchor` while building?

abhijit0943: I don't prefer we do that since this is not a sequential counter like nonce and it would be better to keep it as per chain semantics

raksha-r7 (approved):
lgtm

PR #8147, opened 2026-02-16
feat(statics): add Hemi EVM chain support

manojkumar138:
Ticket: CECHO-115

### DIFF
--- a/modules/statics/src/allCoinsAndTokens.ts
+++ b/modules/statics/src/allCoinsAndTokens.ts
@@ -2285,6 +2285,42 @@ export const allCoinsAndTokens = [
       CoinFeature.EVM_UNSIGNED_SWEEP_RECOVERY,
     ]
   ),
+  account(
+    '68d22683-a8f2-47b3-8446-92e02a1963ae',
+    'hemieth',
+    'Hemi Ethereum',
+    Networks.main.hemieth,
+    18,
+    UnderlyingAsset.HEMIETH,
+    BaseUnit.ETH,
+    [
+      ...EVM_FEATURES,
+      CoinFeature.SHARED_EVM_SIGNING,
+      CoinFeature.SHARED_EVM_SDK,
+      CoinFeature.EVM_COMPATIBLE_IMS,
+      CoinFeature.EVM_COMPATIBLE_UI,
+      CoinFeature.EVM_COMPATIBLE_WP,
+      CoinFeature.SUPPORTS_ERC20,
+    ]
+  ),
+  account(
+    'df0007d9-974b-4099-bf37-9dc59399c058',
+    'themieth',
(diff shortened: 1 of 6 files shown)


modules/statics/src/allCoinsAndTokens.ts
```diff
@@ -2249,6 +2253,52 @@ export const allCoinsAndTokens = [
...
+      CoinFeature.EVM_COMPATIBLE_IMS,
+      CoinFeature.EVM_COMPATIBLE_UI,
+      CoinFeature.EVM_COMPATIBLE_WP,
+      CoinFeature.EVM_NON_BITGO_RECOVERY,
+      CoinFeature.EVM_UNSIGNED_SWEEP_RECOVERY,
+      CoinFeature.SUPPORTS_ERC20,
+    ],
+    KeyCurve.Secp256k1,
+    '',
+    'Hemi'
```
parasgarg-bitgo: Why do we need suffix for hemi?

modules/bitgo/test/v2/unit/keychains.ts
```diff
@@ -103,6 +103,7 @@ describe('V2 Keychains', function () {
         n.asset !== UnderlyingAsset.FLUENTETH &&
         n.asset !== UnderlyingAsset.MANTLE &&
         n.asset !== UnderlyingAsset.JOVAYETH &&
+        n.asset !== UnderlyingAsset.HEMI &&
```
parasgarg-bitgo: We don't need to add new coins to this, as the implementation is already available in the common evm module

modules/statics/src/allCoinsAndTokens.ts
```diff
@@ -2249,6 +2253,52 @@ export const allCoinsAndTokens = [
...
+    BaseUnit.ETH,
+    [
+      ...EVM_FEATURES,
+      CoinFeature.SHARED_EVM_SIGNING,
+      CoinFeature.SHARED_EVM_SDK,
+      CoinFeature.EVM_COMPATIBLE_IMS,
+      CoinFeature.EVM_COMPATIBLE_UI,
+      CoinFeature.EVM_COMPATIBLE_WP,
+      CoinFeature.EVM_NON_BITGO_RECOVERY,
+      CoinFeature.EVM_UNSIGNED_SWEEP_RECOVERY,
```
parasgarg-bitgo: Let's add these CoinFeatures later - when we are ready for adding recovery support.
As things stand, no one will be able to bump SDK in WRW without corresponding config changes for the new coin in WRW

modules/statics/src/allCoinsAndTokens.ts
manojkumar138: Removed them now.

parasgarg-bitgo (approved):
LGTM

PR #8171, opened 2026-02-18
fix(express): restore backward compatibility for empty eip1559 object

Marzooqa:
## Summary                                                                                 
  - Restores backward compatibility for clients sending `eip1559: {}` 
  - Adds validation to reject partial eip1559 objects (only one of                           
  maxFeePerGas/maxPriorityFeePerGas)                                                         
  - Applies to both `/sendcoins` and `/sendmany` endpoints                                   
                                                                                             
  ## Changes                                                                                 
  - Updated `EIP1559Params` schema to use `t.partial` for flexible input validation          
  - Added controller-level validation with clear error messages:                             
    - `eip1559 missing maxPriorityFeePerGas`                                                 
    - `eip1559 missing maxFeePerGas`                
  - Empty eip1559: {} is normalized to undefined.                                  
  - Added tests for backward compatibility and partial object rejection

TICKET: WP-7882

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -937,6 +937,23 @@ async function handleV2SendOne(req: ExpressApiRouteRequest<'express.v2.wallet.se
   const wallet = await coin.wallets().get({ id: req.decoded.id, reqId });
   req.body.reqId = reqId;
 
+  // Validate eip1559: reject partial objects, normalize empty objects to undefined
+  if (req.body.eip1559) {
+    const { maxFeePerGas, maxPriorityFeePerGas } = req.body.eip1559;
+    const hasMax = maxFeePerGas !== undefined;
+    const hasPriority = maxPriorityFeePerGas !== undefined;
+    if (hasMax && !hasPriority) {
+      throw new ApiResponseError('eip1559 missing maxPriorityFeePerGas', 400);
+    }
+    if (hasPriority && !hasMax) {
+      throw new ApiResponseError('eip1559 missing maxFeePerGas', 400);
+    }
+    // Normalize empty object to undefined for server-side auto-estimation
+    if (!hasMax && !hasPriority) {
+      req.body.eip1559 = undefined;
+    }
+  }
+
   let result;
   try {
     result = await wallet.send(createSendParams(req));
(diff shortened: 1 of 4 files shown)


modules/express/src/clientRoutes.ts
```diff
@@ -937,6 +937,19 @@ async function handleV2SendOne(req: ExpressApiRouteRequest<'express.v2.wallet.se
   const wallet = await coin.wallets().get({ id: req.decoded.id, reqId });
   req.body.reqId = reqId;
 
+  // Validate eip1559: must have both fields or neither
+  if (req.body.eip1559) {
+    const { maxFeePerGas, maxPriorityFeePerGas } = req.body.eip1559;
+    const hasMax = maxFeePerGas !== undefined;
+    const hasPriority = maxPriorityFeePerGas !== undefined;
+    if (hasMax && !hasPriority) {
+      throw new ApiResponseError('eip1559 missing maxPriorityFeePerGas', 400);
```
lokesh-bitgo, line 946: Have you confirmed whether both fields are required to be passed here?

Marzooqa: Yes. The server-side auto-estimation only triggers when both fields are missing. If only one field is provided, the server won't auto-estimate, it will fail downstream. Better to reject early with a clear error.

modules/express/src/typedRoutes/api/v2/sendmany.ts
```diff
@@ -16,12 +16,17 @@ export const SendManyRequestParams = {
...
  * EIP-1559 fee parameters for Ethereum transactions
- * When eip1559 object is present, both fields are REQUIRED
+ *
+ * Accepts:
+ * - Empty object {} - triggers automatic fee estimation (backward compatible)
+ * - Object with both maxFeePerGas AND maxPriorityFeePerGas - uses provided values
+ *
+ * Note: Partial objects (only one field) pass validation but backend handles them
  */
-export const EIP1559Params = t.type({
```
alextse-bg, line 21: nit:

```
export const EIP1559Params = t.union([
  t.type({}),
  t.type({
    /** Maximum priority fee per gas (in wei) - REQUIRED */
    maxPriorityFeePerGas: t.union([t.number, t.string]),
    /** Maximum fee per gas (in wei) - REQUIRED */
    maxFeePerGas: t.union([t.number, t.string]),
  })
]);
```

alextse-bg (commented):
some nits

lokesh-bitgo (approved):

modules/express/src/typedRoutes/api/v2/sendmany.ts
Marzooqa: `t.type({})` accepts any object, including partials. So added the validation on the controller.

alextse-bg (approved):

PR #8186, opened 2026-02-20
fix: tada withdrawal failure caused due to strict conditions

dgm003:
TICKET: CSHLD-34

### DIFF
--- a/modules/sdk-coin-ada/src/ada.ts
+++ b/modules/sdk-coin-ada/src/ada.ts
@@ -139,7 +139,10 @@ export class Ada extends BaseCoin {
         for (const recipient of txParams.recipients) {
           let find = false;
           for (const output of explainedTx.outputs) {
-            if (recipient.address === output.address && recipient.amount === output.amount) {
+            if (
+              recipient.address === output.address &&
+              (recipient.amount === 'max' || BigInt(output.amount) >= BigInt(recipient.amount))
+            ) {
               find = true;
             }
           }
--- a/modules/sdk-coin-ada/src/adaToken.ts
+++ b/modules/sdk-coin-ada/src/adaToken.ts
@@ -137,7 +137,7 @@ export class AdaToken extends Ada {
             }
             const multiAssets = output.multiAssets as CardanoWasm.MultiAsset;
             const tokenQty = multiAssets.get_asset(policyScriptHash, assetName);
-            return tokenQty && tokenQty.to_str() === recipient.amount;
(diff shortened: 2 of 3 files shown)


modules/sdk-coin-ada/src/ada.ts
```diff
@@ -139,7 +139,10 @@ export class Ada extends BaseCoin {
         for (const recipient of txParams.recipients) {
           let find = false;
           for (const output of explainedTx.outputs) {
-            if (recipient.address === output.address && recipient.amount === output.amount) {
+            if (
+              recipient.address === output.address &&
+              (recipient.amount === 'max' || BigInt(output.amount) >= BigInt(recipient.amount))
```
Ranjna-G, line 144: recipient.amount is the withdrawal amount. Max is a flag that we use internally to find out if it is a max withdrawal. amount can never be **max**. Please remove this condition

Ranjna-G, line 144: `BigInt(output.amount) >= BigInt(recipient.amount)` this condition will allow withdrawals exceeding the expected amount and will not flag any buggy behaviour in build. As the issue is with the fee address deposits from the same enterprise as the fee address, let's check if the certs from the tx is 2 (one for sender and the other for gas tank), if so, `BigInt(output.amount) >= BigInt(recipient.amount)` can be validated

modules/sdk-coin-ada/src/adaToken.ts
```diff
@@ -137,7 +137,7 @@ export class AdaToken extends Ada {
             }
             const multiAssets = output.multiAssets as CardanoWasm.MultiAsset;
             const tokenQty = multiAssets.get_asset(policyScriptHash, assetName);
-            return tokenQty && tokenQty.to_str() === recipient.amount;
+            return tokenQty && (recipient.amount === 'max' || BigInt(tokenQty.to_str()) >= BigInt(recipient.amount));
```
Ranjna-G, line 140: Given the issue is with the fee address deposits which does not include tokens, this change is not required

modules/sdk-coin-ada/test/unit/ada.ts
```diff
@@ -249,6 +249,97 @@ describe('ADA', function () {
...
+        ],
+      };
+
+      await basecoin
+        .verifyTransaction({ txParams, txPrebuild, verification: {} })
+        .should.be.rejectedWith('cannot find recipient in expected output');
+    });
+
+    it('should succeed when recipient amount is max', async () => {
+      const txPrebuild = newTxPrebuild();
```
Ranjna-G, line 293: Please fix/add test cases with new changes

PR #8199, opened 2026-02-25
feat(stnear): adding support for metapool withdraw_all txn builder

aditya-bitgo:
ticket: SC-5580

### DIFF
--- a/modules/sdk-coin-near/src/lib/constants.ts
+++ b/modules/sdk-coin-near/src/lib/constants.ts
@@ -7,6 +7,7 @@ export const StakingContractMethodNames = {
   DepositAndStake: 'deposit_and_stake',
   Unstake: 'unstake',
   Withdraw: 'withdraw',
+  WithdrawAll: 'withdraw_all',
 } as const;
 
 export const FT_TRANSFER = 'ft_transfer';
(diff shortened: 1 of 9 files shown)


Doddanna17 (commented):
`MetaPoolWithdrawBuilder` duplicates significant portions of `StakingWithdrawBuilder` (constructor, initBuilder, gas, buildImplementation). Consider extending the existing builder with a `validateArgs` hook -- see inline suggestion.

`explainStakingWithdrawTransaction` returns `outputAmount: '0'` for `withdraw_all` since the amount is resolved on-chain. Acceptable trade-off, but downstream consumers should be aware.

Factory routing, transaction type mapping, and input/output handling are all correct. `LIQUID_STAKING` addition to stnear tokens is appropriate.

Tests are comprehensive: signed/unsigned builds, round-trips, routing verification, error case. LGTM.

Doddanna17 (commented):
Hey @aditya-bitgo, the refactoring looks clean -- the 17-line MetaPoolWithdrawBuilder extending StakingWithdrawBuilder with the `validateArgs` hook is exactly the right pattern.

One small thing: `MetaPoolWithdrawBuilder` inherits the public `amount()` method from `StakingWithdrawBuilder`. Since `validateArgs` is a no-op here, calling `.amount('...')` on a MetaPoolWithdrawBuilder would silently set args on a `withdraw_all` call without any error. Nothing calls it today, but an override that throws would prevent accidental misuse:

```typescript
public amount(_amount: string): this {
  throw new BuildTransactionError('amount is not applicable for withdraw_all');
}
```

Rest looks solid. LGTM with that minor addition.

Doddanna17 (approved):
All feedback addressed -- amount() override with throw, validateArgs parameter signature, and test for the rejection case. Clean inheritance pattern, comprehensive tests. LGTM.

PR #8224, opened 2026-03-03
fix: deriveTransactionType defaults to CustomTx for non-transfer transactions

lcovar:
## Summary

- `deriveTransactionType()` (introduced in #8206) fell through to `Send` for any transaction that didn't match a known staking or ATA pattern. This misclassified custom transactions (e.g. boilerplate-only Memo+NonceAdvance txs) as `Send`.
- Replaced the implicit `Send` fallback with an explicit check for `Transfer`/`TokenTransfer` instructions. Everything else defaults to `CustomTx`, matching the legacy `getTransactionType()` behavior.
- Added unit test covering the boilerplate-only classification case.

## Test plan

- [x] New unit test: boilerplate-only tx (NonceAdvance + Memo) classified as `CustomTx`
- [x] Full sdk-coin-sol test suite: 546 passing, 0 regressions


lcovar:
existing tests cover sendtx, added a new tests specifically for custom tx via the wasm path

mukeshsp (approved):

PR #8226, opened 2026-03-03
feat: add support of custodial/cold wallet

HimanshuS2212:
Ticket: CGARD-395

TICKET: CGARD-395


modules/sdk-core/src/bitgo/evm/evmUtils.ts
```diff
@@ -13,6 +13,8 @@ export interface CreateEvmKeyRingWalletParams {
   evmKeyRingReferenceWalletId: string;
   bitgo: BitGoBase;
   baseCoin: IBaseCoin;
+  /** Enterprise ID - required for cold/custodial wallets */
+  enterprise?: string;
```
nvjsr, line 17: why enterprise is needed for cold and custodial support ?

PR #8227, opened 2026-03-03
fix(ada):token recovery build

dgm003:
Ticket: CSHLD-133

## Description

Recovery transactions for ADA wallets holding native tokens were dropping
the tokens entirely, producing only a single ADA output instead of the
expected token output + ADA remainder output.

**Root cause:** 'recover()' function in 'ada.ts' called 'changeAddress()' without
passing the token asset list from the UTXO set, so the transaction builder
had no knowledge of tokens. Additionally, the 'buildImplementation()' gate
in 'transactionBuilder.ts' did not route token recovery into the refactored
build path ('processTokenBuild') which handles token outputs correctly.

**Fix:**
- Aggregate 'asset_list' from all UTxOs in 'recover()' into a
  fingerprint-keyed map and pass it as the 3rd argument to 'changeAddress()'
- Call 'txBuilder.isTokenTransaction()' when the aggregated asset list is
  non-empty, so the builder routes through 'processTokenBuild()' →
  'addChangeOutput()' which produces both a token output (1.5 ADA + tokens)
  and an ADA remainder output
  
## Issue Number

CSHLD-133

## Type of change

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Added two unit tests in `test/unit/ada.ts`:
- **ADA-only recovery** — asserts 1 output, correct ADA amount, no multiAssets
- **ADA + token recovery** — asserts 2 outputs: one token output (1.5 ADA +
  correct token quantity/policy) and one ADA remainder output (totalBalance - minADA - fee)

Ran locally with:
npx mocha
All the 160 test cases passed.

### DIFF
--- a/modules/sdk-coin-ada/src/ada.ts
+++ b/modules/sdk-coin-ada/src/ada.ts
@@ -393,10 +393,32 @@ export class Ada extends BaseCoin {
       throw new Error('Did not find address with funds to recover.');
     }
 
+    // Aggregate token assets from all UTxOs into a fingerprint-keyed map for the builder
+    const aggregatedAssetList: Record<string, { policy_id: string; asset_name: string; quantity: string }> = {};
+    for (const utxo of utxoSet) {
+      for (const asset of utxo.asset_list ?? []) {
+        const fp = asset.fingerprint as string;
+        if (aggregatedAssetList[fp]) {
+          aggregatedAssetList[fp].quantity = (
+            BigInt(aggregatedAssetList[fp].quantity) + BigInt(asset.quantity)
+          ).toString();
+        } else {
+          aggregatedAssetList[fp] = {
+            policy_id: asset.policy_id,
+            asset_name: asset.encoded_asset_name ?? asset.asset_name,
+            quantity: asset.quantity,
+          };
+        }
+      }
+    }
+
     // first build the unsigned txn
     const tipAbsSlot = await this.getChainTipInfo();
     const txBuilder = this.getBuilder().getTransferBuilder();
-    txBuilder.changeAddress(params.recoveryDestination, balance.toString());
+    txBuilder.changeAddress(params.recoveryDestination, balance.toString(), aggregatedAssetList);
+    if (Object.keys(aggregatedAssetList).length > 0) {
+      txBuilder.isTokenTransaction();
+    }
     for (const utxo of utxoSet) {
       txBuilder.input({ transaction_id: utxo.tx_hash, transaction_index: utxo.tx_index });
     }
--- a/modules/sdk-coin-ada/test/unit/ada.ts
+++ b/modules/sdk-coin-ada/test/unit/ada.ts
@@ -21,6 +21,7 @@ import {
   wrwUser,
 } from '../resources';
 import * as _ from 'lodash';
+import * as CardanoWasm from '@emurgo/cardano-serialization-lib-nodejs';
 import { Ada, KeyPair, Tada } from '../../src';
 import { Transaction } from '../../src/lib';
 import { TransactionType } from '../../../sdk-core/src/account-lib/baseCoin/enum';
@@ -584,9 +585,10 @@ describe('ADA', function () {
   describe('Recover Transactions:', () => {
     const destAddr = address.address2;
     const sandBox = sinon.createSandbox();
+    let callBack: sinon.SinonStub;
 
     beforeEach(function () {
-      const callBack = sandBox.stub(Ada.prototype, 'getDataFromNode' as keyof Ada);
+      callBack = sandBox.stub(Ada.prototype, 'getDataFromNode' as keyof Ada);
       callBack
         .withArgs('address_info', {
           _addresses: [wrwUser.walletAddress0],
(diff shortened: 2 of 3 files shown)


modules/sdk-coin-ada/src/ada.ts
```diff
@@ -393,10 +393,32 @@ export class Ada extends BaseCoin {
...
+      for (const asset of utxo.asset_list ?? []) {
+        const fp = asset.fingerprint as string;
+        if (aggregatedAssetList[fp]) {
+          aggregatedAssetList[fp].quantity = (
+            BigInt(aggregatedAssetList[fp].quantity) + BigInt(asset.quantity)
+          ).toString();
+        } else {
+          aggregatedAssetList[fp] = {
+            policy_id: asset.policy_id,
+            asset_name: asset.encoded_asset_name,
```
Ranjna-G: `encoded_asset_name` in utxo asset_list is introduced with token onboarding, can we add the backward compatibility logic as well? Please refer paymentHandler.ts in ims

modules/sdk-coin-ada/test/unit/ada.ts
```diff
@@ -737,6 +738,98 @@ describe('ADA', function () {
...
+    const tokenPolicyId = '2533cca6eb42076e144e9f2772c390dece9fce173bc38c72294b3924';
+    const tokenEncodedAssetName = '5741544552';
+    const tokenQuantity = '111';
+    const minADAForToken = 1500000;
+
+    afterEach(function () {
+      sandBox.restore();
+    });
+
+    it('should recover ADA only UTXOs with no asset list in outputs (signed)', async function () {
```
Ranjna-G: There is already a test for ADA only recovery, please remove this

modules/sdk-coin-ada/test/unit/ada.ts
```diff
@@ -737,6 +738,98 @@ describe('ADA', function () {
...
+      should.deepEqual(txJson.inputs[1].transaction_id, testnetUTXO.UTXO_2.tx_hash);
+
+      txJson.outputs.length.should.equal(1);
+      should.deepEqual(txJson.outputs[0].address, destAddr);
+      should.deepEqual(Number(txJson.outputs[0].amount) + fee, testnetUTXO.UTXO_1.value + testnetUTXO.UTXO_2.value);
+      should.not.exist(txJson.outputs[0].multiAssets);
+    });
+
+    it('should recover ADA plus token UTXOs - token and ADA both appear in outputs (signed)', async function () {
+      const callBack = sandBox.stub(Ada.prototype, 'getDataFromNode' as keyof Ada);
```
Ranjna-G: These stubs are already present in `beforeEach`, please remove

modules/sdk-coin-ada/test/unit/ada.ts
```diff
@@ -737,6 +738,98 @@ describe('ADA', function () {
...
+      should.deepEqual(txJson.inputs[0].transaction_id, testnetUTXO.UTXO_1.tx_hash);
+      should.deepEqual(txJson.inputs[1].transaction_id, testnetUTXO.UTXO_2.tx_hash);
+
+      txJson.outputs.length.should.equal(1);
+      should.deepEqual(txJson.outputs[0].address, destAddr);
+      should.deepEqual(Number(txJson.outputs[0].amount) + fee, testnetUTXO.UTXO_1.value + testnetUTXO.UTXO_2.value);
+      should.not.exist(txJson.outputs[0].multiAssets);
+    });
+
+    it('should recover ADA plus token UTXOs - token and ADA both appear in outputs (signed)', async function () {
```
Ranjna-G: Please move this test case to the exisiting `Recover Transactions:` test

modules/sdk-coin-ada/test/unit/ada.ts
```diff
@@ -671,6 +673,56 @@ describe('ADA', function () {
...
+        .to_str()
+        .should.equal(tokenQuantity);
+
+      const adaOutput = txJson.outputs.find((o) => o.multiAssets === undefined);
+      should.exist(adaOutput);
+      should.deepEqual(adaOutput!.address, destAddr);
+      const fee = Number(tx.explainTransaction().fee.fee);
+      const totalBalance = testnetUTXO.UTXO_1.value + testnetUTXO.UTXO_TOKEN.value;
+      should.deepEqual(Number(adaOutput!.amount), totalBalance - minADAForToken - fee);
+    });
```
abhishekagrawal080, line 725: can we add cold wallet recovery test cases as well? as recover method is common for both

abhishekagrawal080 (approved):

modules/sdk-coin-ada/test/unit/ada.ts
dgm003: There is another ticket related to it.
Will make these suggested changes in that ticket  👍

modules/sdk-coin-ada/test/unit/ada.ts
dgm003: Made the required changes 👍

modules/sdk-coin-ada/src/ada.ts
dgm003: Made the suggested changes 👍

PR #8229, opened 2026-03-03
feat: remove multisig support for zksyncera

ArunBala-Bitgo:
Ticket: CECHO-284

As per the requirements we don't need to add support for multisig wallets for this chain, hence removing the feature from zksyncera features.


nvjsr (approved):
lgtm

PR #8231, opened 2026-03-03
feat(logger): enhance sanitize to handle  Date, BigInt, and string types and add the test file.Reduce the depth from 50 to 25.

rishikeshdadam136:
TICKET-8110

The sanitize() function in @bitgo/logger handles the below ones correctly and added the test files.

- Date objects.
- BigInt values. 
- Strings — bearer v2 tokens passed as top-level strings are not redacted.
-Reduce the depth.

### DIFF
--- a/modules/logger/src/sanitizeLog.ts
+++ b/modules/logger/src/sanitizeLog.ts
@@ -15,7 +15,7 @@ const SENSITIVE_KEYS = new Set([
   '_token',
 ]);
 
-const BEARER_V2_PATTERN = /^v2x[a-f0-9]{32,}$/i;
+const SENSITIVE_PREFIXES = ['v2x', 'xprv'];
 
 /**
  * Checks if a key is sensitive (case-insensitive)
@@ -25,10 +25,30 @@ function isSensitiveKey(key: string): boolean {
 }
 
 /**
- * Checks if a value matches the bearer v2 token pattern
+ * Checks if a string value is sensitive based on known prefixes.
+ * Unlike isSensitiveKey (which checks property names), this identifies
+ * sensitive data by recognizable content patterns — useful when there
+ * is no key context (e.g. top-level strings, array elements).
  */
-function isBearerV2Token(value: unknown): boolean {
-  return typeof value === 'string' && BEARER_V2_PATTERN.test(value);
+function isSensitiveStringValue(s: string): boolean {
+  return SENSITIVE_PREFIXES.some((prefix) => s.startsWith(prefix));
+}
+
+export function getErrorData(error: unknown): unknown {
+  if (!(error && error instanceof Error)) {
+    return error;
+  }
+
+  const errorData: Record<string, unknown> = {
+    name: error.name,
+  };
+
+  for (const key of Object.getOwnPropertyNames(error)) {
+    const value = (error as unknown as Record<string, unknown>)[key];
+    errorData[key] = value instanceof Error ? getErrorData(value) : value;
+  }
+
+  return errorData;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


rishikeshdadam136:
@copilot Review it as 10x engineer.

modules/logger/src/sanitizeLog.ts
```diff
@@ -31,21 +31,45 @@ function isBearerV2Token(value: unknown): boolean {
...
     return obj;
   }
 
+  // Handle BigInt (JSON.stringify(1n) throws TypeError)
+  if (typeof obj === 'bigint') {
+    return obj.toString();
+  }
+
+  if (typeof obj === 'string') {
+    return isBearerV2Token(obj) ? '<REMOVED>' : obj;
```
OttoAllmendinger: why only bearerV2? shouldn't we remove all sensitive fields?

OttoAllmendinger (changes requested):

modules/logger/src/sanitizeLog.ts
rishikeshdadam136: Because it has a recognizable format (v2x + hex) that can be detected from the string content alone. For other sensitive values (passwords, private keys, etc.), we can only identify them when they appear under a sensitive key name, which is already handled in the object iteration loop.
if u suggest me any particular format like bearer i will add that also.
Thank you.

OttoAllmendinger: I see

`xprv` is a pattern that we detect as a prefix as well

OttoAllmendinger: add a function `isSensitiveStringValue(s: string)` that does the `v2x` and `xprv` prefix checks

OttoAllmendinger (changes requested):

modules/logger/src/sanitizeLog.ts
rishikeshdadam136: Will look into it 
Very Thanks

OttoAllmendinger (dismissed):

modules/logger/src/sanitizeLog.ts
```diff
@@ -62,16 +89,21 @@ export function sanitize(obj: unknown, seen = new WeakSet<Record<string, unknown
     return obj.map((item) => sanitize(item, seen, depth + 1));
   }
 
+  // Handle Date objects
+  if (obj instanceof Date) {
+    return obj.toISOString();
```
mrdanish26:
```suggestion
    return isNaN(obj.getTime()) ? '[Invalid Date]' : obj.toISOString();
```

modules/logger/src/sanitizeLog.ts
```diff
@@ -15,7 +15,7 @@ const SENSITIVE_KEYS = new Set([
   '_token',
 ]);
 
-const BEARER_V2_PATTERN = /^v2x[a-f0-9]{32,}$/i;
+const SENSITIVE_PREFIXES = ['v2x', 'xprv'];
```
mrdanish26, line 18: The old implementation used a case-insensitive regex , so tokens like V2Xaabb... were caught. The new startsWith('v2x') check is case-sensitive, so mixed-case tokens will slip through unredacted?

rishikeshdadam136: Yes will slip through but normally the tokens in the logging statements will be in this format(v2xaabb..) only right,i mean no capital alphabet.

modules/logger/src/sanitizeLog.ts
rishikeshdadam136: Will look into it.

OttoAllmendinger (approved):

kaustubhbitgo (approved):

mrdanish26 (approved):

dpkjnr (approved):

PR #8233, opened 2026-03-03
feat(sdk-core): pass buildParams in /tx/send for UTXO and account-based coins

derranW26:
## Summary

- Preserve server-validated `buildParams` from `/tx/build` response through the prebuild→sign→send pipeline
- Forward `buildParams` fields (e.g. `recipients`, `feeRate`, `rbfTxIds`, `maxFee`, `feeMultiplier`, `nonce`) in the `/tx/send` request body
- Enables platform intent verification by matching what was built against what is being sent
- Changes are **coin-agnostic** — applies to both UTXO coins (BTC, LTC, BCH, DOGE, ZEC) and account-based coins (ETH, Polygon, BSC, etc.)

### `modules/sdk-core/src/bitgo/wallet/wallet.ts`

1. **`prebuildTransaction()`**: Capture `buildResponse.buildParams` before `Object.assign` overwrites it. Instead of deleting `buildParams`, preserve server-validated params (falling back to request params if server doesn't return them).

2. **`prebuildAndSignTransaction()`**: After signing, attach `buildParams` from the prebuild to the signed transaction result so they flow through to `sendMany()`.

3. **`sendMany()`**: Destructure `buildParams` from the signed transaction and spread its fields into the final send params. User `selectParams` take precedence over server `buildParams` for any overlapping fields.

### `modules/bitgo/test/v2/unit/wallet.ts`

**UTXO coin tests (WP-8015):**
- `prebuildTransaction` preserves server-returned `buildParams` on the result
- `prebuildTransaction` falls back to request params when server doesn't return `buildParams`
- `sendMany` includes `recipients` from server `buildParams` in `/tx/send` request
- `sendMany` includes RBF fields (`rbfTxIds`, `maxFee`, `feeMultiplier`) in `/tx/send` request
- `sendMany` works when server returns no `buildParams` (backwards compat)
- `sendMany` lets explicit user params override server `buildParams`

**Account-based coin tests (WP-8014):**
- ETH `/tx/send` has top-level `recipients` alongside `halfSigned` with server `buildParams`
- ETH `/tx/send` has top-level `recipients` via `selectParams` even without server `buildParams`
- ETH `/tx/send` includes `nonce` and other fields from server `buildParams` at top level

### `modules/sdk-coin-eth/test/unit/ethWallet.ts`

- Updated ETH hop transaction test to expect `buildParams` on prebuild result (no longer deleted)

## Test plan

- [ ] Existing wallet unit tests pass (no regressions)
- [ ] New UTXO tests verify buildParams flow
- [ ] New account-based coin tests verify `recipients` at top level alongside `halfSigned`
- [ ] Verify `/tx/send` request body contains `recipients` for both UTXO and account-based sendMany calls

WP-8015, WP-8014

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -1848,81 +1848,10 @@ describe('V2 Wallet:', function () {
 
       response.isDone().should.be.true();
     });
-
-    it('should preserve server-returned buildParams on prebuild result', async function () {
-      const serverBuildParams = {
-        recipients: [{ address: 'addr1', amount: '5000' }],
-        feeRate: 10000,
-      };
-      const params = { offlineVerification: true };
-      nock(bgUrl)
-        .post(`/api/v2/${wallet.coin()}/wallet/${wallet.id()}/tx/build`, tbtcHotWalletDefaultParams)
-        .query(params)
-        .reply(200, { buildParams: serverBuildParams });
-      const blockHeight = 100;
-      sinon.stub(basecoin, 'getLatestBlockHeight').resolves(blockHeight);
-      sinon.stub(basecoin, 'postProcessPrebuild').callsFake((params) => Promise.resolve(params));
-      const txPrebuild = await wallet.prebuildTransaction({ ...params, reqId: reqId });
-      should.exist(txPrebuild.buildParams);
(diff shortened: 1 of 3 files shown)


KaustubhPatange (approved):

kaustubhbitgo (dismissed):

OttoAllmendinger (changes requested):
see slack

PR #8250, opened 2026-03-05
feat: add tempo pathusd and usdc tokens

nayandas190:
Ticket: CECHO-303

- add tempo pathusd and usdc tip20 tokens


ArunBala-Bitgo (approved):
LGTM

PR #8262, opened 2026-03-07
feat: hmac authentication strategy and response verification

bitgoAaron:
- Updated  function to be asynchronous, allowing for better handling of HMAC verification.
- Introduced  for default HMAC handling and added support for custom strategies.
- Integrated  for browser compatibility, enabling HMAC signing and verification using the Web Crypto API.
- Enhanced  to utilize the new HMAC strategies for request signing and response verification.
- Added unit tests for the new HMAC strategies and their integration with the BitGoAPI.
- Updated web demo to include a new component for WebCrypto authentication.

Ticket: CE-10122

### DIFF
--- a/modules/sdk-api/src/api.ts
+++ b/modules/sdk-api/src/api.ts
@@ -11,7 +11,7 @@ import querystring from 'querystring';
 
 import { ApiResponseError, BitGoRequest } from '@bitgo/sdk-core';
 
-import { AuthVersion, VerifyResponseOptions } from './types';
+import { AuthVersion, VerifyResponseInfo, VerifyResponseOptions } from './types';
 import { BitGoAPI } from './bitgoAPI';
 
 const debug = Debug('bitgo:api');
@@ -214,44 +214,23 @@ export function setRequestQueryString(req: superagent.SuperAgentRequest): void {
 }
 
 /**
- * Verify that the response received from the server is signed correctly.
- * Right now, it is very permissive with the timestamp variance.
+ * Validate a completed verification response and throw a descriptive `ApiResponseError` if it
+ * indicates the response is invalid or outside the acceptable time window.
  */
-export function verifyResponse(
+function assertVerificationResponse(
   bitgo: BitGoAPI,
   token: string | undefined,
-  method: VerifyResponseOptions['method'],
   req: superagent.SuperAgentRequest,
   response: superagent.Response,
-  authVersion: AuthVersion
-): superagent.Response {
-  // we can't verify the response if we're not authenticated
-  if (!req.isV2Authenticated || !req.authenticationToken) {
-    return response;
-  }
-
-  const verificationResponse = bitgo.verifyResponse({
-    url: req.url,
-    hmac: response.header.hmac,
-    statusCode: response.status,
-    text: response.text,
-    timestamp: response.header.timestamp,
-    token: req.authenticationToken,
-    method,
-    authVersion,
-  });
-
+  verificationResponse: VerifyResponseInfo
+): void {
   if (!verificationResponse.isValid) {
-    // calculate the HMAC
-    const receivedHmac = response.header.hmac;
-    const expectedHmac = verificationResponse.expectedHmac;
-    const signatureSubject = verificationResponse.signatureSubject;
     // Log only the first 10 characters of the token to ensure the full token isn't logged.
     const partialBitgoToken = token ? token.substring(0, 10) : '';
     const errorDetails = {
-      expectedHmac,
-      receivedHmac,
-      hmacInput: signatureSubject,
+      expectedHmac: verificationResponse.expectedHmac,
+      receivedHmac: response.header.hmac,
+      hmacInput: verificationResponse.signatureSubject,
       requestToken: req.authenticationToken,
       bitgoToken: partialBitgoToken,
     };
@@ -271,5 +250,62 @@ export function verifyResponse(
       errorDetails
     );
   }
+}
+
+/**
+ * Verify that the response received from the server is signed correctly.
+ * Right now, it is very permissive with the timestamp variance.
+ */
+export function verifyResponse(
+  bitgo: BitGoAPI,
+  token: string | undefined,
+  method: VerifyResponseOptions['method'],
+  req: superagent.SuperAgentRequest,
+  response: superagent.Response,
+  authVersion: AuthVersion
+): superagent.Response {
+  if (!req.isV2Authenticated || !req.authenticationToken) {
+    return response;
+  }
+
+  const verificationResponse = bitgo.verifyResponse({
+    url: req.url,
+    hmac: response.header.hmac,
+    statusCode: response.status,
+    text: response.text,
+    timestamp: response.header.timestamp,
+    token: req.authenticationToken,
+    method,
+    authVersion,
+  });
+
+  assertVerificationResponse(bitgo, token, req, response, verificationResponse);
+  return response;
+}
+
+export async function verifyResponseAsync(
+  bitgo: BitGoAPI,
+  token: string | undefined,
+  method: VerifyResponseOptions['method'],
+  req: superagent.SuperAgentRequest,
+  response: superagent.Response,
+  authVersion: AuthVersion
+): Promise<superagent.Response> {
+  if (!req.isV2Authenticated || !req.authenticationToken) {
+    return response;
+  }
+
+  const verificationResponse = await bitgo.verifyResponseAsync({
+    url: req.url,
+    hmac: response.header.hmac,
+    statusCode: response.status,
+    text: response.text,
+    timestamp: response.header.timestamp,
+    token: req.authenticationToken,
+    method,
+    authVersion,
+  });
+
+  assertVerificationResponse(bitgo, token, req, response, verificationResponse);
   return response;
 }
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -378,6 +378,84 @@ export class BitGoAPI implements BitGoBase {
   }
 
   /**
+   * Signs and sends a v2-authenticated request, then verifies the response HMAC.
+   * Extracted from the req.then override in requestPatch to keep that method readable.
+   */
+  private async _sendRequestWithHmac({
+    req,
+    method,
+    url,
+    data,
+    strategyAuthenticated,
+    onfulfilled,
+    originalThen,
+  }: {
+    req: superagent.SuperAgentRequest;
+    method: RequestMethods;
+    url: string;
+    data: string | undefined;
+    strategyAuthenticated: boolean;
+    onfulfilled: ((response: superagent.Response) => any) | null | undefined;
+    originalThen: (onfulfilled: any, onrejected?: any) => Promise<any>;
+  }): Promise<any> {
+    if (this._token || strategyAuthenticated) {
+      setRequestQueryString(req);
+
+      const requestProperties = await this._hmacAuthStrategy.calculateRequestHeaders({
+        url: req.url,
+        token: this._token ?? '',
+        method,
+        text: data || '',
+        authVersion: this._authVersion,
+      });
+      req.set('Auth-Timestamp', requestProperties.timestamp.toString());
+
+      req.set('Authorization', 'Bearer ' + requestProperties.tokenHash);
+      debug(
+        'sending v2 %s request to %s with token %s',
+        method,
+        url,
+        this._token?.substr(0, 8) ?? '(strategy-managed)'
+      );
+
+      req.set('HMAC', requestProperties.hmac);
+    }
+
+    if (this.getAdditionalHeadersCb) {
+      const additionalHeaders = this.getAdditionalHeadersCb(method, url, data);
+      for (const { key, value } of additionalHeaders) {
+        req.set(key, value);
+      }
+    }
+
+    /**
+     * Verify the response before calling the original onfulfilled handler,
+     * and make sure onrejected is called if a verification error is encountered
+     */
+    const newOnFulfilled = onfulfilled
+      ? async (response: superagent.Response) => {
+          // HMAC verification is only allowed to be skipped in certain environments.
+          // This is checked in the constructor, but checking it again at request time
+          // will help prevent against tampering of this property after the object is created
+          if (!this._hmacVerification && !common.Environments[this.getEnv()].hmacVerificationEnforced) {
+            return onfulfilled(response);
+          }
+
+          const verifiedResponse = await verifyResponseAsync(
+            this,
+            this._token,
+            method,
+            req,
+            response,
+            this._authVersion
+          );
+          return onfulfilled(verifiedResponse);
+        }
+      : null;
+    return originalThen(newOnFulfilled);
+  }
+
+  /**
    * This is a patching function which can apply our authorization
    * headers to any outbound request.
    * @param method
@@ -446,65 +524,15 @@ export class BitGoAPI implements BitGoBase {
 
       const data = serializeRequestData(req);
 
-      const sendWithHmac = (async () => {
-        if (this._token || strategyAuthenticated) {
-          setRequestQueryString(req);
-
-          const requestProperties = await this._hmacAuthStrategy.calculateRequestHeaders({
-            url: req.url,
-            token: this._token ?? '',
-            method,
-            text: data || '',
-            authVersion: this._authVersion,
-          });
-          req.set('Auth-Timestamp', requestProperties.timestamp.toString());
-
-          req.set('Authorization', 'Bearer ' + requestProperties.tokenHash);
-          debug(
-            'sending v2 %s request to %s with token %s',
-            method,
-            url,
-            this._token?.substr(0, 8) ?? '(strategy-managed)'
-          );
-
-          req.set('HMAC', requestProperties.hmac);
-        }
-
-        if (this.getAdditionalHeadersCb) {
-          const additionalHeaders = this.getAdditionalHeadersCb(method, url, data);
-          for (const { key, value } of additionalHeaders) {
-            req.set(key, value);
-          }
-        }
-
-        /**
-         * Verify the response before calling the original onfulfilled handler,
-         * and make sure onrejected is called if a verification error is encountered
-         */
-        const newOnFulfilled = onfulfilled
-          ? async (response: superagent.Response) => {
-              // HMAC verification is only allowed to be skipped in certain environments.
-              // This is checked in the constructor, but checking it again at request time
-              // will help prevent against tampering of this property after the object is created
-              if (!this._hmacVerification && !common.Environments[this.getEnv()].hmacVerificationEnforced) {
-                return onfulfilled(response);
-              }
-
-              const verifiedResponse = await verifyResponseAsync(
-                this,
-                this._token,
-                method,
-                req,
-                response,
-                this._authVersion
-              );
-              return onfulfilled(verifiedResponse);
-            }
-          : null;
-        return originalThen(newOnFulfilled);
-      })();
-
-      return sendWithHmac.catch(onrejected);
+      return this._sendRequestWithHmac({
+        req,
+        method,
+        url,
+        data,
+        strategyAuthenticated,
+        onfulfilled,
+        originalThen,
+      }).catch(onrejected);
     };
     return toBitgoRequest(req);
   }
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -412,7 +412,8 @@ export class BitGoAPI implements BitGoBase {
 
       req.set('Authorization', 'Bearer ' + requestProperties.tokenHash);
       debug(
-        'sending v2 %s request to %s with token %s',
+        'sending v%d %s request to %s with token %s',
+        this._authVersion,
         method,
         url,
         this._token?.substr(0, 8) ?? '(strategy-managed)'
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -23,6 +23,7 @@ import {
   sanitizeLegacyPath,
 } from '@bitgo/sdk-core';
 import * as sdkHmac from '@bitgo/sdk-hmac';
+import { DefaultHmacAuthStrategy, type IHmacAuthStrategy } from '@bitgo/sdk-hmac';
 import * as utxolib from '@bitgo/utxo-lib';
 import { bip32, ECPairInterface } from '@bitgo/utxo-lib';
 import * as bitcoinMessage from 'bitcoinjs-message';
@@ -37,7 +38,7 @@ import {
   serializeRequestData,
   setRequestQueryString,
   toBitgoRequest,
-  verifyResponse,
+  verifyResponseAsync,
 } from './api';
 import { decrypt, encrypt } from './encrypt';
 import { verifyAddress } from './v1/verifyAddress';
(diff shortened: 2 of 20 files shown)


modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+    const expectedHmac = await webCryptoHmacSign(this.cryptoKey, subject);
+
+    const now = Date.now();
+    const backwardValidityWindow = 1000 * 60 * 5;
+    const forwardValidityWindow = 1000 * 60;
+    const isInResponseValidityWindow =
+      params.timestamp >= now - backwardValidityWindow && params.timestamp <= now + forwardValidityWindow;
+
+    return {
+      isValid: expectedHmac === params.hmac,
```
mattreid1: In `hmacv4.ts` we use `crypto.timingSafeEqual`. While this is unlikely to be exploited, probably best to use that here too for timing attack prevention

modules/web-demo/src/components/WebCryptoAuth/index.tsx
```diff
@@ -0,0 +1,467 @@
...
+      const restored = await strategy.restoreToken();
+      if (restored) {
+        appendLog(
+          'CryptoSigning restored (CryptoKey + tokenHash). No raw token involved.',
+        );
+      } else {
+        appendLog('No stored CryptoSigning found in IndexedDB.');
+      }
+
+      const options: Record<string, unknown> = {
```
mattreid1:
```suggestion
      const options: BitGoAPIOptions = {
```

I know this is a demo app, but might be worth using a type safe param here so if there's a breaking change, we'd get type errors as another signal something broke

modules/sdk-hmac/src/webCryptoStrategy.ts
mattreid1: In `hmacv4.ts` we use `crypto.timingSafeEqual`. While this is unlikely to be an issue, probably best to use that here too for timing attack prevention. Might need a polyfill for this though.

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+}): string {
+  let method = params.method;
+  if (method === 'del') {
+    method = 'delete';
+  }
+
+  const queryPath = extractQueryPath(params.urlPath);
+
+  let prefixedText: string;
+  if (params.statusCode !== undefined && isFinite(params.statusCode) && Number.isInteger(params.statusCode)) {
```
mattreid1: Linting suggests `Number.isFinite` instead of `isFinite`

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+  CryptoSigning,
+  IHmacAuthStrategy,
+  ITokenStore,
+  CalculateRequestHeadersOptions,
+  RequestHeaders,
+  VerifyResponseOptions,
+  VerifyResponseInfo,
+} from './types';
+
+function arrayBufToHex(buffer: ArrayBuffer): string {
```
mattreid1, line 20: This is unbounded, might it be worth putting a max size on this?

Also in Node.js you could do (about 100x faster):
```ts
function arrayBufToHexNode(buffer: ArrayBuffer): string {
  return Buffer.from(buffer).toString('hex');
}
```

Maybe could have different implementations/code paths for browser and Node.js?

---

I ran a quick benchmark with `new ArrayBuffer(1024) // 1KB`:
```
arrayBufToHex: 0.155ms
arrayBufToHexNode: 0.02ms
```

and `new ArrayBuffer(1024 * 1024 * 10) // 10MB`:
```
arrayBufToHex: 487.858ms
arrayBufToHexNode: 4.55ms
```

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+      timestamp: params.timestamp,
+      method: params.method,
+      statusCode: params.statusCode,
+      authVersion: params.authVersion,
+    });
+
+    const expectedHmac = await webCryptoHmacSign(this.cryptoKey, subject);
+
+    const now = Date.now();
+    const backwardValidityWindow = 1000 * 60 * 5;
```
mattreid1, line 170: Any reason for 5 minutes instead of just 1 minute? Not sure how long our timeouts are, but iirc HMAC is done right at the end, just before the request is sent down the wire, so we should only really need to take network latency and user clock drift into account plus some buffer

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+  RequestHeaders,
+  VerifyResponseOptions,
+  VerifyResponseInfo,
+} from './types';
+
+function arrayBufToHex(buffer: ArrayBuffer): string {
+  const bytes = new Uint8Array(buffer);
+  const hexParts: string[] = new Array(bytes.length);
+  for (let i = 0; i < bytes.length; i++) {
+    hexParts[i] = bytes[i].toString(16).padStart(2, '0');
```
mattreid1, line 24: A lookup table would be quicker for buffers greater than a few hundred bytes

```ts
const hexTable = Array.from({ length: 256 }, (_, i) => i.toString(16).padStart(2, '0'));

for (let i = 0; i < bytes.length; i++) {
  hexChars.push(hexTable[bytes[i]]);
}
```

mattreid1 (changes requested):

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: if it were a general purpose function or exported, sure. In this case though, this is only used as an internal util for converting specific buffers to hex strings, and the specific buffers are always fixed size:

the outputs of:
`crypto.subtle.sign('HMAC', ...) ` and `crypto.subtle.digest('SHA-256', ...)` are the only inputs to this function, and both of those are always 32 bytes. 

The bigger issue though is that would require depending on `Buffer`, which is not something we want to introduce. One of the design constraints of this file was to be 0 dependency (including NodeJS natives like Buffer, url, crypto etc). So, all that to say I think we're good with this as is.

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: 5 min follows the same logic that other hmac validity checks follow, I don't think we need to refine that as part of this change.

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: sure, but in practice this will never make a difference since the buffer is always 32 bytes

mattreid1 (dismissed):

modules/sdk-api/src/api.ts
```diff
@@ -214,44 +214,23 @@ export function setRequestQueryString(req: superagent.SuperAgentRequest): void {
 }
 
 /**
- * Verify that the response received from the server is signed correctly.
- * Right now, it is very permissive with the timestamp variance.
+ * Validate a completed verification response and throw a descriptive `ApiResponseError` if it
+ * indicates the response is invalid or outside the acceptable time window.
  */
-export function verifyResponse(
+function assertVerificationResponse(
```
OttoAllmendinger, line 220: 👍 good naming pattern

modules/sdk-api/src/api.ts
```diff
@@ -271,5 +250,62 @@ export function verifyResponse(
...
+  bitgo: BitGoAPI,
+  token: string | undefined,
+  method: VerifyResponseOptions['method'],
+  req: superagent.SuperAgentRequest,
+  response: superagent.Response,
+  authVersion: AuthVersion
+): superagent.Response {
+  if (!req.isV2Authenticated || !req.authenticationToken) {
+    return response;
+  }
```
OttoAllmendinger, line 269: can we maybe move this to the caller? it's a bit unclear in this context here why we would want to skip these reqs

modules/sdk-api/src/api.ts
```diff
@@ -271,5 +250,62 @@ export function verifyResponse(
...
+  bitgo: BitGoAPI,
+  token: string | undefined,
+  method: VerifyResponseOptions['method'],
+  req: superagent.SuperAgentRequest,
+  response: superagent.Response,
+  authVersion: AuthVersion
+): Promise<superagent.Response> {
+  if (!req.isV2Authenticated || !req.authenticationToken) {
+    return response;
+  }
```
OttoAllmendinger, line 296: same

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -439,51 +445,66 @@ export class BitGoAPI implements BitGoBase {
...
-        req.set('Auth-Timestamp', requestProperties.timestamp.toString());
-
-        // we're not sending the actual token, but only its hash
-        req.set('Authorization', 'Bearer ' + requestProperties.tokenHash);
-        debug('sending v2 %s request to %s with token %s', method, url, this._token?.substr(0, 8));
 
-        // set the HMAC
-        req.set('HMAC', requestProperties.hmac);
-      }
+      const sendWithHmac = (async () => {
```
OttoAllmendinger: clarity: this deserves to be a method

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+  text: string;
+  timestamp: number;
+  method: string;
+  statusCode?: number;
+  authVersion: AuthVersion;
+}): string {
+  let method = params.method;
+  if (method === 'del') {
+    method = 'delete';
+  }
```
OttoAllmendinger: Looks like we need to normalize the method here, maybe we should go further and introduce a more strict type. Unclear what the callers are but maybe we need to lowercase for instance.

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+
+  const queryPath = extractQueryPath(params.urlPath);
+
+  let prefixedText: string;
+  if (params.statusCode !== undefined && isFinite(params.statusCode) && Number.isInteger(params.statusCode)) {
+    prefixedText =
+      params.authVersion === 3
+        ? [method.toUpperCase(), params.timestamp, queryPath, params.statusCode].join('|')
```
OttoAllmendinger: ... or normalize to uppercase I guess

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+    }
+  }
+}
+
+/**
+ * Build the HMAC subject string for v2/v3 request or response signing.
+ * Browser-compatible equivalent of `calculateHMACSubject` from hmac.ts,
+ * producing identical output for string inputs.
+ */
+function buildHmacSubject(params: {
```
OttoAllmendinger: `export` would allow some nice unit tests for this

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+      return url.search.length > 0 ? url.pathname + url.search : url.pathname;
+    } catch {
+      return urlPath;
+    }
+  }
+}
+
+/**
+ * Build the HMAC subject string for v2/v3 request or response signing.
+ * Browser-compatible equivalent of `calculateHMACSubject` from hmac.ts,
```
OttoAllmendinger: `hmac.ts` probably referencing something from `sdk-api` I guess...

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+
+async function webCryptoSha256Hex(data: string): Promise<string> {
+  const encoded = new TextEncoder().encode(data);
+  const hash = await crypto.subtle.digest('SHA-256', encoded);
+  return arrayBufToHex(hash);
+}
+
+// ---------------------------------------------------------------------------
+// IndexedDB Token Store
+// ---------------------------------------------------------------------------
```
OttoAllmendinger: Could be a separate file and separate PR

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+  try {
+    await new Promise<void>((resolve, reject) => {
+      const tx = db.transaction(CRYPTO_STORE_NAME, 'readwrite');
+      tx.objectStore(CRYPTO_STORE_NAME).delete(CRYPTO_RECORD_KEY);
+      tx.oncomplete = () => resolve();
+      tx.onerror = () => reject(tx.error);
+    });
+  } finally {
+    db.close();
+  }
```
OttoAllmendinger: the early return, try/catch/finally pattern could be a `withDb` wrapper

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+        db.createObjectStore(CRYPTO_STORE_NAME);
+      }
+    };
+    request.onsuccess = () => resolve(request.result);
+    request.onerror = () => reject(request.error);
+  });
+}
+
+async function persistCryptoSigning(signing: CryptoSigning): Promise<void> {
+  if (!hasIndexedDB()) return;
```
OttoAllmendinger: it would be a little nicer to have a `class CryptoStore` with a `db: IDBDatabase` property with some methods

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+   * const headers = await strategy.getAuthHeaders({ url, method: 'GET' });
+   * const response = await fetch(url, { headers });
+   * ```
+   */
+  async getAuthHeaders(params: { url: string; method: string; body?: string }): Promise<Record<string, string>> {
+    const requestHeaders = await this.calculateRequestHeaders({
+      url: params.url,
+      token: '',
+      method: params.method as CalculateRequestHeadersOptions['method'],
+      text: params.body ?? '',
```
OttoAllmendinger: we often seem to pass empty strings to `token` and `text` - better make them optional params instead and normalize in the impl

modules/sdk-hmac/test/webCryptoStrategy.ts
```diff
@@ -0,0 +1,302 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import { WebCryptoHmacStrategy } from '../src/webCryptoStrategy';
+import * as hmac from '../src/hmac';
+import type { CryptoSigning, ITokenStore } from '../src/types';
+
+const MOCK_TIMESTAMP = 1672531200000;
+
+/**
+ * In-memory token store for testing (IndexedDB is not available in Node.js).
```
OttoAllmendinger, line 10: it may be worthwhile pulling in `fake-indexeddb` as a dev dep

modules/sdk-hmac/src/webCryptoStrategy.ts
```diff
@@ -0,0 +1,364 @@
...
+}
+
+/**
+ * Build the HMAC subject string for v2/v3 request or response signing.
+ * Browser-compatible equivalent of `calculateHMACSubject` from hmac.ts,
+ * producing identical output for string inputs.
+ */
+function buildHmacSubject(params: {
+  urlPath: string;
+  text: string;
```
OttoAllmendinger: this seems actually optional

modules/sdk-api/src/api.ts
bitgoAaron: not without introducing a breaking change as part of this. This function is publicly exported, and requiring the client to suddenly validate that would require a breaking change. Everything else this PR is implementing is entirely opt in & transparent to existing clients, so in short - no.

modules/sdk-api/src/api.ts
bitgoAaron: see above.

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: this has already been done (see https://github.com/BitGo/BitGoJS/pull/7282). The type is strictly defined, and must include `del` in order to support the original implementation where we used superagent method names instead of proper HTTP verbs.

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: Realized here that we are duplicating the logic used for constructing the hmac subject. Changing this to import the same function from `./hmac` to implement subject generation.

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: see above, removing this in favor of existing util.

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: agreed. I'll pull this out into a separate file & tests to make things a bit easier to reason about.

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: we do indeed have that, `IndexedDbTokenStore` and it is the default class used by `WebCryptoHmacStrategy` for its `tokenStore`, but can be overridden by the client as long as they implement the interface.

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: fixing the types here, WebCryptoHmacStrategy does not need the token will extend a type that omits this.
In this class, `text` is optional and defaults to empty string. I don't think we should make any other changes to existing types & contracts as part of this (making `CalculateHmacSubjectOptions` 's `text` optional).

modules/sdk-hmac/test/webCryptoStrategy.ts
bitgoAaron: not necessary, sufficient tests implemented with a couple functions to mock indexedDB's surface. BitGoJS already has enough deps

modules/sdk-hmac/src/webCryptoStrategy.ts
bitgoAaron: adhering to types shared by core util functions here.

mohammadalfaiyazbitgo (dismissed):
LGTM

OttoAllmendinger (dismissed):

alextse-bg (dismissed):

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -374,6 +377,84 @@ export class BitGoAPI implements BitGoBase {
...
+    originalThen,
+  }: {
+    req: superagent.SuperAgentRequest;
+    method: RequestMethods;
+    url: string;
+    data: string | undefined;
+    strategyAuthenticated: boolean;
+    onfulfilled: ((response: superagent.Response) => any) | null | undefined;
+    originalThen: (onfulfilled: any, onrejected?: any) => Promise<any>;
+  }): Promise<any> {
```
margueriteblair, line 400: nitty - do we need all these `any` types here?

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -374,6 +377,84 @@ export class BitGoAPI implements BitGoBase {
...
+        token: this._token ?? '',
+        method,
+        text: data || '',
+        authVersion: this._authVersion,
+      });
+      req.set('Auth-Timestamp', requestProperties.timestamp.toString());
+
+      req.set('Authorization', 'Bearer ' + requestProperties.tokenHash);
+      debug(
+        'sending v2 %s request to %s with token %s',
```
margueriteblair: in this debug log, will it always be v2?

modules/sdk-api/test/unit/hmacStrategy.ts
```diff
@@ -0,0 +1,175 @@
...
+import type {
+  IHmacAuthStrategy,
+  CalculateRequestHeadersOptions,
+  RequestHeaders,
+  VerifyResponseOptions,
+  VerifyResponseInfo,
+} from '@bitgo/sdk-hmac';
+import assert from 'node:assert';
+
+const TEST_TOKEN = 'v2x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdefab';
```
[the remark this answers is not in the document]
margueriteblair: am i crazy? is there a reason we'd keep this?

modules/sdk-api/test/unit/hmacStrategy.ts
```diff
@@ -0,0 +1,175 @@
...
+          }
+        );
+
+      try {
+        await bitgo.authenticate({
+          username: 'test@test.com',
+          password: 'mypassword',
+        });
+      } catch {
+        // Authentication may fail for various reasons in this test context,
```
margueriteblair: uh, we want this to catch nothing?

modules/sdk-hmac/test/defaultStrategy.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  beforeEach(() => {
+    strategy = new DefaultHmacAuthStrategy();
+  });
+
+  describe('calculateRequestHeaders', () => {
+    it('should produce the same result as the sync calculateRequestHeaders', async () => {
+      const params = {
+        url: 'https://app.bitgo.com/api/v2/wallet',
+        token: 'v2x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdefab',
```
margueriteblair, line 28: eh?

modules/sdk-hmac/test/defaultStrategy.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      expect(asyncResult.hmac).to.equal(syncResult.hmac);
+      expect(asyncResult.timestamp).to.equal(syncResult.timestamp);
+      expect(asyncResult.tokenHash).to.equal(syncResult.tokenHash);
+    });
+
+    it('should produce correct headers for v3 auth with a body', async () => {
+      const params = {
+        url: 'https://app.bitgo.com/api/v2/wallet/send',
+        token: 'v2x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdefab',
```
margueriteblair, line 45: I'm sure this is some AI generated token or something? But still don't think this would be fun to commit to a public repo?

mattreid1 (dismissed):

modules/sdk-api/src/bitgoAPI.ts
bitgoAaron: This repo unfortunately uses `any` rather prolifically, and is out of scope of this refactor to change that. Doing so would likely result in breaking changes to downstream clients.

modules/sdk-api/src/bitgoAPI.ts
bitgoAaron: good catch ~ this is has been like this forever, might as well fix it while we're here.

modules/sdk-api/test/unit/hmacStrategy.ts
bitgoAaron: yeah, good catch. This was just the LLM being overly cautious getting the tests to pass, not needed.

modules/sdk-hmac/test/defaultStrategy.ts
bitgoAaron: it's literally `v2x123456abcdefg`... i think its fine.

mohammadalfaiyazbitgo (approved):

mattreid1 (approved):

margueriteblair (approved):

PR #8215, opened 2026-03-02
fix: update serialize-javascript to 7.0.3

yashvanthbl137-crypto:
### Summary
Updates `serialize-javascript` from `6.0.2` to `7.0.3` to resolve a Remote Code Execution (RCE) vulnerability.

### Vulnerability Details
- **CVE**: GHSA-5c6j-r48x-rmvq
- **Severity**: HIGH
- **Attack Vector**: RCE via malicious `RegExp.flags` and `Date.prototype.toISOString()` methods

### Changes
- Added Yarn resolution `**/serialize-javascript: 7.0.3` to force version across all transitive dependencies
- Updated `yarn.lock` to reflect the resolution

Failed Job: https://github.com/BitGo/BitGoJS/actions/runs/22571247763/job/65379334392

Ticket: CGARD-518

PR #8279, opened 2026-03-11
chore: remove serialize-javascript security exception

yashvanthbl137-crypto:
Removes the serialize-javascript security exception (GHSA-5c6j-r48x-rmvq) from `.iyarc` since it was upgraded to 7.0.3 in #8215 and the vulnerability is now fixed.

Verified locally that `yarn audit-high` passes with no issues.

Fixes CGARD-518


louib (approved):

bhargavirao24:
Good to see the exclusions being removed. Thanks!

PR #8286, opened 2026-03-12
fix: coinName fix in token config

rohitsaw115:
ticket: cecho-125

Replace hardcoded coin.name.split(':')[0] base coin derivation in getEthLikeTokenConfig and getEthLikeERC721TokenConfig with coins.coinNameFromChainId(network.chainId), falling back to the split logic for backward compatibility
Fixes token registration for coins like hbarevm:tmt where the token name prefix doesn't match the expected testnet base coin name (e.g., hbarevm:tmt is a testnet token but name.split(':')[0] returns hbarevm instead of thbarevm)


nayandas190 (approved):
lgtm

HimanshuS2212 (approved):

PR #8291, opened 2026-03-13
feat(statics): add DynamicNetwork and DynamicCoin for AMS-discovered chains

shubham-damkondwar:
Ticket: CECHO-439

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -30,10 +30,10 @@ import {
   erc721Token,
 } from './account';
 import { ofcToken } from './ofc';
-import { BaseCoin, CoinFeature } from './base';
+import { BaseCoin, CoinFeature, DynamicCoin } from './base';
 import { AmsTokenConfig, TrimmedAmsTokenConfig } from './tokenConfig';
 import { CoinMap } from './map';
-import { Networks, NetworkType } from './networks';
+import { BaseNetwork, getNetwork, getNetworksMap, NetworkType } from './networks';
 import { networkFeatureMapForTokens } from './networkFeatureMapForTokens';
 import { ofcErc20Coins, tOfcErc20Coins } from './coins/ofcErc20Coins';
 import { ofcCoins } from './coins/ofcCoins';
@@ -74,6 +74,14 @@ allCoinsAndTokens.forEach((coin) => {
 });
 
 export function createToken(token: AmsTokenConfig): Readonly<BaseCoin> | undefined {
+  if (!token.isToken) {
+    try {
+      return buildDynamicCoin(token);
+    } catch (error) {
+      console.warn(`Failed to build dynamic coin for ${token.name} (${token.id}):`, error);
+      return undefined;
+    }
+  }
   const initializerMap: Record<string, unknown> = {
     algo: algoToken,
     apt: aptToken,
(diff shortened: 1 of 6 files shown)


modules/statics/src/coins.ts
```diff
@@ -352,6 +355,33 @@ export function createToken(token: AmsTokenConfig): Readonly<BaseCoin> | undefin
   }
 }
 
+/**
+ * Build a real DynamicCoin + DynamicNetwork instance for AMS-discovered base chains
+ * whose family is not yet registered in the SDK's initializerMap.
+ * Called from createToken() as a fallback when no initializer exists and isToken is false.
+ */
+function buildDynamicCoin(token: AmsTokenConfig): Readonly<BaseCoin> {
+  const network = Object.freeze(new DynamicNetwork(token.network as DynamicNetworkOptions));
```
manas-at-bitgo: We need to fetch this network from a network factory which returns networks ingested by ams as well.

manas-at-bitgo (changes requested):

manas-at-bitgo:
@claude review this PR. We are trying to add methods and utilities to dynamically inject network from a centralised service like we do for tokens in coinmap. Think of factory pattern for network enum/objects.

manas-at-bitgo (changes requested):
### Must Fix

**1. `token.features` can be `undefined` — runtime crash**

`AmsTokenConfig.features` is `string[] | undefined`. In `buildDynamicCoin`, `token.features as string[]` passes `undefined` through to `BaseCoin.validateOptions()`, which iterates with `for...of` → `TypeError: options.features is not iterable`.

```diff
- features: token.features as string[],
+ features: (token.features ?? []) as string[],
```

**2. Missing `denom` propagation**

`AmsTokenConfig.denom` is never passed to the `DynamicCoin` constructor. Cosmos-family chains (and others using denoms) will silently lose this field.

```diff
  new DynamicCoin({
    ...
    alias: token.alias,
+   denom: token.denom,
  })
```

### Should Fix

**3. EVM-centric `baseUnit` default**

`baseUnit` defaults to `'wei'`. For non-EVM chains (UTXO, Cosmos, Substrate), this is incorrect. Consider requiring AMS to always provide `baseUnit`, or use a family-aware default.

**4. No unit tests**

Zero test coverage for `DynamicCoin`, `DynamicNetwork`, `getNetwork()`, `buildDynamicCoin()`. At minimum:
- `DynamicCoin` construction with valid and edge-case AMS payloads
- `getNetwork()` with JSON string, plain name, and invalid input
- `buildDynamicCoin()` with `features: undefined` (the bug above)

**5. `getNetwork()` resolution order may shadow local statics**

JSON parse runs before `getNetworkByName()`. If AMS sends `{"name": "bitcoin", "type": "mainnet", "family": "btc"}`, a new `DynamicNetwork` is created instead of returning the existing `Networks.main.bitcoin` singleton. Code comparing network references (`coin.network === Networks.main.bitcoin`) would break.

```suggestion
// Check local statics first, then fall back to JSON parse
const networkObj = getNetworkByName(network);
if (networkObj) return networkObj;
// ... then try JSON parse
```

**6. DynamicNetwork instances aren't cached**

Every `getNetwork()` call with the same JSON string creates a new instance. If multiple AMS-discovered tokens share a network, they'll hold different objects. Consider a `Map<string, DynamicNetwork>` cache keyed on `name+type`.

manas-at-bitgo (dismissed):
lgtm

nvrakesh06 (approved):

PR #8323, opened 2026-03-19
chore: add go account signing script

alia-bitgo:
TICKET: CAAS-1008


Taseen08 (approved):
looks good - adds example script for go account sign tx

CapnMigraine (approved):

PR #8337, opened 2026-03-23
feat(canton): return extended payload from signablePayload 

derranW26:
Ticket: WP-8187


sachushaji:
@claude

ravibitgo (dismissed):
What is the release plan for this? WP changes are also needed for this, else signing will break

sachushaji:
@claude

sachushaji:
@claude

PR #8339, opened 2026-03-24
fix: sdk hmac strategy init & clear

bitgoAaron:
handle hmac strategy token init in
authenticate methods, add new clearAsync
to handle clearing the strategy

Ticket: CE-10122

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -1046,6 +1046,14 @@ export class BitGoAPI implements BitGoBase {
         response.body.access_token = this._token;
       }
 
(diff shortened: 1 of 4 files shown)


margueriteblair (approved):
lgtm!

ecastro-intive (approved):
lgtm.

one small note. We might want to call `await sdk.clearAsync();` when clearing token on [WebCryptoAuth](https://github.com/BitGo/BitGoJS/blob/d7698adbf5e40dc36f49239eab913f38f083effa/modules/web-demo/src/components/WebCryptoAuth/index.tsx) to wipe memory session as well, otherwise an "already logged in" error shows

alextse-bg (approved):

PR #8342, opened 2026-03-24
feat: add registerNetwork function to dynamically register new networks

shubham-damkondwar:
CECHO-549

TICKET: CECHO-549


mukeshsp (approved):

manas-at-bitgo (commented):
Any reason to have this as a seperate PR and not include it with #8291?

PR #8352, opened 2026-03-26
feat: added signing support for TRX MPC

bhuvanr159:
TICKET: CHALO-347

### DIFF
--- a/modules/sdk-coin-trx/src/trx.ts
+++ b/modules/sdk-coin-trx/src/trx.ts
@@ -36,7 +36,7 @@ import {
   isTssVerifyAddressOptions,
 } from '@bitgo/sdk-core';
 import { auditEcdsaPrivateKey } from '@bitgo/sdk-lib-mpc';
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-trx/src/trx.ts
```diff
@@ -396,6 +400,50 @@ export class Trx extends BaseCoin {
...
 
+  /**
+   * Verify a TSS transaction where txHex is the signableHex (raw_data_hex)
+   * rather than a full transaction JSON.
+   *
+   * @param signableHex The raw_data_hex of the transaction
+   * @param txParams The original transaction parameters
+   * @returns true if the transaction is valid
+   */
+  private verifyTssTransaction(signableHex: string, txParams: any): boolean {
```
bhavidhingra: Do we need to add this function separately?
most of the transaction verification done in this function is done for multisig also.

bhavidhingra (dismissed):

bhavidhingra (approved):

PR #8367, opened 2026-03-28
feat(sdk-api): add registerToken method to BitGoAPI

manas-at-bitgo:
## Summary

- Adds `registerToken(coinConfig, coinConstructor)` method to `BitGoAPI` in `modules/sdk-api/src/bitgoAPI.ts`
- Delegates directly to `GlobalCoinFactory.registerToken` — idempotent, does not affect existing statics coins
- Enables runtime registration of AMS-sourced tokens without requiring statics library updates or releases
- Uses `Parameters<CoinFactory['registerToken']>[0]` for the coin config type (no new direct dep on `@bitgo/statics`)

## Test plan

- [x] Unit test: `sdk.coin(name)` resolves correctly after `registerToken`
- [x] Unit test: calling `registerToken` twice for the same token does not throw (idempotency)
- [x] Unit test: existing statics coins are unaffected after `registerToken` calls
- [x] Full unit-test suite passes (114 tests)

Closes CSHLD-89

### DIFF
--- a/modules/bitgo/src/bitgo.ts
+++ b/modules/bitgo/src/bitgo.ts
@@ -100,12 +100,12 @@ export class BitGo extends BitGoAPI {
   }
 
   /**
-   * Register a token in the coin factory
-   * @param tokenConfig - The token metadata from AMS
+   * Register a token in the coin factory by name, fetching metadata from AMS if needed.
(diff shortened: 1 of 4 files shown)


alextse-bg (commented):
CI is failing

PR #8369, opened 2026-03-28
feat(root): remove node engine upper bound

andrew-scott-fischer:
## Summary
- Removes the `<25` upper bound from the `engines.node` field across all 90 module package.json files
- Node 25 users currently get silently downgraded to `bitgo@4.49.2`, which depends on `bitgo-utxo-lib@1.3.1`, which references the deleted `BitGo/blake2b` git repo — breaking `npm install bitgo` entirely
- Changing to `">=20"` prevents this from recurring on future Node releases

TICKET: CR-1399

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -128,4 +128,4 @@ This will generate the necessary boilerplate for a new coin implementation.
 
 ## Node.js Version Support
 
-BitGoJS supports Node.js versions >=20 and <25, with NPM >=3.10.10.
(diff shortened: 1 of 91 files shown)


louib (approved):

pranavjain97 (approved):

lcovar (approved):

pranavjain97 (commented):
I think we should actually keep the upper bound i.e <26 for now. In future node versions it might then silently fail at runtime

PR #8381, opened 2026-03-31
fix(sdk-coin-sol): fix decimalPlaces falsy check in solInstructionsFactory

ArunBala-Bitgo:
Solana unsupported SPL NFTs with 0 decimal places fail during pending approval because decimalPlaces is checked with a truthy comparison. In JavaScript, 0 is falsy, so the condition `data.params.decimalPlaces` evaluates to false even when decimalPlaces is explicitly set to 0. This causes the code to skip the sendParams/data.params branch, fall through to the token registry lookup (which fails for unsupported tokens), and throw "Invalid token name".

Replace truthy checks with `!=null` in this location:
- solInstructionFactory.ts tokenTransferInstruction()

TICKET: CECHO-606

### DIFF
--- a/modules/sdk-coin-sol/src/lib/solInstructionFactory.ts
+++ b/modules/sdk-coin-sol/src/lib/solInstructionFactory.ts
@@ -196,7 +196,7 @@ function tokenTransferInstruction(data: TokenTransfer): TransactionInstruction[]
   let tokenAddress: string;
   let programId: string | undefined;
   let decimalPlaces: number;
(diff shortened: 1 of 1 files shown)


mullapudipruthvik (commented):
### Code review

No issues found. Checked for: CLAUDE.md compliance, bugs, error handling, security, bot-specific failure modes, database operations, code comment accuracy.

:white_check_mark: **This PR looks safe to approve** — small, trivial change with no logic modifications.

mullapudipruthvik (approved):

PR #8416, opened 2026-04-07
fix: handle serializedTxHex format in TSS verifyTransaction

bhuvanr159:
TICKET: CHALO-380

## Summary
- **Issue:** TRX MPC wallet withdrawals were failing with `UtilsError: Number of contracts is greater than 1.`
- **Root cause:** `verifyTransaction` is called twice during the TSS signing flow:
  1. From `prebuildAndSignTransaction` (wallet.ts:2480) — `txHex` is `serializedTxHex`, a full JSON string (`{"txID":"...","raw_data":{...},"raw_data_hex":"..."}`)
  2. From ECDSA signing flow (ecdsa.ts:761) — `txHex` is `signableHex`, raw protobuf hex bytes

  The TSS branch only handled case 2 (protobuf hex). When case 1 passed a JSON string, the protobuf decoder produced garbage data and threw the error.
- **Fix:** Detect the format by attempting `JSON.parse(txHex)` — if it succeeds, extract `raw_data_hex` from the parsed JSON; if it fails, treat `txHex` as raw protobuf hex. This mirrors the pattern used by the builder's own `getTxReceipt()` method in `wrappedBuilder.ts`.

## Test plan
- [x] Added unit tests for `serializedTxHex` (JSON string) path — validates success and amount-mismatch failure
- [x] Existing unit tests for `signableHex` (protobuf) path still pass
- [x] All 115 TRX unit tests pass (`nx run @bitgo/sdk-coin-trx:unit-test`)
- [x] TypeScript compilation clean (`tsc --noEmit`)
- [ ] Manual test: TRX MPC wallet withdrawal on testnet

### DIFF
--- a/modules/sdk-coin-trx/src/trx.ts
+++ b/modules/sdk-coin-trx/src/trx.ts
@@ -379,10 +379,21 @@ export class Trx extends BaseCoin {
     }
 
     if (walletType === 'tss') {
-      // For TSS wallets, txHex is the signableHex (raw_data_hex protobuf bytes),
-      // not a full transaction JSON. Decode it directly via protobuf.
-      // Note: decodeTransaction already validates exactly 1 contract exists.
-      const decodedTx = Utils.decodeTransaction(txPrebuild.txHex);
+      // For TSS wallets, verifyTransaction is called from two places:
+      // 1. prebuildAndSignTransaction (wallet.ts) — txHex is serializedTxHex, a full JSON string
+      //    containing { txID, raw_data, raw_data_hex }.
+      // 2. ECDSA signing flow (ecdsa.ts) — txHex is signableHex, the raw protobuf bytes (raw_data_hex).
+      // We need to extract the raw_data_hex in case 1 before decoding.
+      let rawDataHex: string;
+      try {
+        // serializedTxHex: full JSON string — extract the raw_data_hex field
+        rawDataHex = JSON.parse(txPrebuild.txHex).raw_data_hex;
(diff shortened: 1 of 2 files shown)


modules/sdk-coin-trx/src/trx.ts
```diff
@@ -379,10 +379,20 @@ export class Trx extends BaseCoin {
...
-      // not a full transaction JSON. Decode it directly via protobuf.
-      // Note: decodeTransaction already validates exactly 1 contract exists.
-      const decodedTx = Utils.decodeTransaction(txPrebuild.txHex);
+      // For TSS wallets, verifyTransaction is called from two places:
+      // 1. prebuildAndSignTransaction (wallet.ts) — txHex is serializedTxHex, a full JSON string
+      //    containing { txID, raw_data, raw_data_hex }.
+      // 2. ECDSA signing flow (ecdsa.ts) — txHex is signableHex, the raw protobuf bytes (raw_data_hex).
+      // We need to extract the raw_data_hex in case 1 before decoding.
+      let rawDataHex: string;
+      if (txPrebuild.txHex.startsWith('{')) {
```
bhavidhingra: rather than this if check, should we do it like this: https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/base/custodialHot/signCustodialHotTransaction.ts#L134?
So JSON.parse by default, if it fails, use txHex directly.

bhavidhingra (approved):

PR #8422, opened 2026-04-07
feat: add AmsNetworkConfig and add enrichBaseChainFeatures method

shubham-damkondwar:
TICKET: CECHO-634

### DIFF
--- a/modules/statics/src/tokenConfig.ts
+++ b/modules/statics/src/tokenConfig.ts
@@ -42,7 +42,7 @@ import {
 } from './account';
 import { CoinFamily, CoinKind, BaseCoin, CoinFeature } from './base';
 import { coins } from './coins';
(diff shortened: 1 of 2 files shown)


modules/statics/src/tokenConfig.ts
```diff
@@ -289,6 +289,12 @@ export interface AmsTokenConfig {
   policyId?: string;
 }
 
+export interface AmsNetworkConfig extends DynamicNetworkOptions {
+  features: string[];
```
manas-at-bitgo: can we use CoinFeature type here?

manas-at-bitgo (approved):

modules/statics/src/tokenConfig.ts
shubham-damkondwar: done

PR #8462, opened 2026-04-09
fix(tss): require txParams with recipients for TSS tx signing

mrdanish26:
## Summary

- Fixes WAL-375: MPCv2 (and ECDSA TSS) `signTxRequest()` silently defaulted `txParams` to `{ recipients: [] }` when the caller omitted it, allowing a compromised BitGo API to swap `signableHex` to redirect funds without client detection
- Adds an early guard in `signRequestBase()` that throws when `txParams.recipients` is absent or empty for `RequestType.tx`, ensuring `verifyTransaction()` always receives explicit caller-supplied params
- Propagates optional `txParams` through `recreateTxRequest()` so the pending-approval re-sign path keeps working; extracts recipients from `pendingApproval.info.transactionRequest`

## Test plan

- [x] Existing success tests for `signTxRequest` in `ecdsaMPCv2/signTxRequest.ts` and `ecdsa.ts` updated to pass `txParams` with recipients
- [x] New negative tests verify rejection when `txParams` is missing
- [x] New negative tests verify rejection when `txParams.recipients` is an empty array
- [x] Run `yarn unit-test` in `modules/bitgo`

### DIFF
--- a/modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
+++ b/modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
@@ -246,7 +246,9 @@ export class PendingApproval implements IPendingApproval {
     }
 
     const decryptedPrv = await this.wallet.getPrv({ walletPassphrase });
-    const txRequest = await this.tssUtils!.recreateTxRequest(txRequestId, decryptedPrv, reqId);
+    const pendingApprovalRecipients = this._pendingApproval.info?.transactionRequest?.recipients;
+    const txParams = pendingApprovalRecipients?.length ? { recipients: pendingApprovalRecipients } : undefined;
+    const txRequest = await this.tssUtils!.recreateTxRequest(txRequestId, decryptedPrv, reqId, txParams);
     if (txRequest.apiVersion === 'lite') {
       if (!txRequest.unsignedTxs || txRequest.unsignedTxs.length === 0) {
         throw new Error('Unexpected error, no transactions found in txRequest.');
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
@@ -47,6 +47,7 @@ import {
 } from '../../../tss/types';
 import { BaseEcdsaUtils } from './base';
 import { IRequestTracer } from '../../../../api';
+import { InvalidTransactionError } from '../../../errors';
 
 const encryptNShare = ECDSAMethods.encryptNShare;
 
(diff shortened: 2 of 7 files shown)


sachushaji:
@claude

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -745,21 +745,25 @@ export class EcdsaUtils extends BaseEcdsaUtils {
       const unsignedTx =
         txRequest.apiVersion === 'full' ? txRequest.transactions![0].unsignedTx : txRequest.unsignedTxs[0];
 
+      if (!params.txParams?.recipients?.length) {
+        throw new Error('txParams with recipients is required for TSS transaction signing');
```
pranavjain97: this should be an `Invalid` and let's make the error more UI friendly

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -736,21 +736,25 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
       const unsignedTx =
         txRequest.apiVersion === 'full' ? txRequest.transactions![0].unsignedTx : txRequest.unsignedTxs[0];
 
+      if (!params.txParams?.recipients?.length) {
+        throw new Error('txParams with recipients is required for TSS transaction signing');
+      }
```
pranavjain97, line 744: same

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
```diff
@@ -246,7 +246,9 @@ export class PendingApproval implements IPendingApproval {
     }
 
     const decryptedPrv = await this.wallet.getPrv({ walletPassphrase });
-    const txRequest = await this.tssUtils!.recreateTxRequest(txRequestId, decryptedPrv, reqId);
+    const pendingApprovalRecipients = this._pendingApproval.info?.transactionRequest?.recipients;
+    const txParams = pendingApprovalRecipients?.length ? { recipients: pendingApprovalRecipients } : undefined;
+    const txRequest = await this.tssUtils!.recreateTxRequest(txRequestId, decryptedPrv, reqId, txParams);
```
pranavjain97, line 251: whats the behavior here if `txParams` is undefined?

pranavjain97 (changes requested):

modules/sdk-core/src/bitgo/pendingApproval/pendingApproval.ts
mrdanish26: If `txParams` is undefined, `recreateTxRequest` still calls signTxRequest with txParams omitted. For ECDSA transaction signing, the SDK then throws error because recipients are required for verification—same guard as when callers omit `txParams` elsewhere. It’s only “optional” at the recreateTxRequest call site when the pending approval has no transactionRequest.recipients; downstream ECDSA tx signing rejects that instead of defaulting to empty recipients.

pranavjain97 (approved):

zhongxishen (approved):
lgtm

PR #8466, opened 2026-04-09
fix(sdk-lib-mpc): re-derive dkgState from WASM bytes in restoreSession

mohammadalfaiyazbitgo:
## Summary

- `Dkg.restoreSession()` now calls `_deserializeState()` after `_restoreSession()` to always derive `dkgState` from the actual WASM bytes, ignoring the caller-supplied value in `DkgSessionData`
- Adds a `Complete` state override when `keyShareBuff` is present, since the WASM bytes still reflect `WaitMsg4` after `keyshare()` frees the session
- Adds a regression test asserting that a tampered `dkgState` in session data is ignored on restore

## Test plan

- [ ] `yarn run unit-test --scope @bitgo/sdk-lib-mpc -- -- --grep "restoreSession"` — new tampered-state test passes
- [ ] `yarn run unit-test --scope @bitgo/sdk-lib-mpc` — all existing DKG round-trip tests continue to pass

WAL-396

### DIFF
--- a/modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
+++ b/modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
@@ -119,7 +119,8 @@ export class Dkg {
         this.dkgState = DkgState.Round3;
         break;
       case 'WaitMsg4':
-        this.dkgState = DkgState.Round4;
+        // keyShareBuff present means keyshare() already ran and freed the session; bytes are frozen at WaitMsg4.
+        this.dkgState = this.keyShareBuff ? DkgState.Complete : DkgState.Round4;
         break;
       case 'Ended':
         this.dkgState = DkgState.Complete;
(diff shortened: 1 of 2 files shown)


modules/sdk-lib-mpc/src/tss/ecdsa-dkls/dkg.ts
```diff
@@ -350,6 +349,15 @@ export class Dkg {
...
     dkg._restoreSession();
+    // Re-derive state from WASM bytes rather than trusting the caller-supplied dkgState.
+    // This prevents a tampered or corrupted dkgState from causing handleIncomingMessages()
+    // to take the wrong branch (e.g. skipping chain code commitment or calling keyshare() prematurely).
+    dkg._deserializeState();
+    if (sessionData.keyShareBuff) {
+      // After keyshare() is called the WASM session is freed and dkgSessionBytes still reflect WaitMsg4.
+      // Presence of keyShareBuff is the reliable signal that the protocol completed.
+      dkg.dkgState = DkgState.Complete;
+    }
```
zahin-mohammad: Shouldn't this be handled by `_deserializeState`

mohammadalfaiyazbitgo: updated

pranavjain97 (approved):

PR #8467, opened 2026-04-09
fix(sdk-lib-mpc): handle WaitMsg4 round in _deserializeState

zahin-mohammad:
## Summary
- Fix `_deserializeState()` missing `WaitMsg4` case in DKLS DSG — at Round 4 the CBOR-decoded round is an object `{ WaitMsg4: { r: ... } }`, not a string, causing `InvalidState` to be thrown
- Convert the `switch` to `if/else if` to support the mixed string/object round type, consistent with how `setSession()` already handles it
- Add test verifying `_deserializeState()` correctly handles the `WaitMsg4` round state

## Test plan
- [x] Existing DKLS DSG tests pass (83 passing)
- [x] New test exercises `_deserializeState()` at WaitMsg4 state and verifies `DsgState.Round4` is set without throwing

Ticket: WAL-384

### DIFF
--- a/modules/sdk-lib-mpc/test/unit/tss/ecdsa/dklsDsg.ts
+++ b/modules/sdk-lib-mpc/test/unit/tss/ecdsa/dklsDsg.ts
@@ -1,4 +1,4 @@
-import { DklsDsg, DklsUtils } from '../../../../src/tss/ecdsa-dkls';
+import { DklsDsg, DklsTypes, DklsUtils } from '../../../../src/tss/ecdsa-dkls';
 import * as fs from 'fs';
 import * as crypto from 'crypto';
 import should from 'should';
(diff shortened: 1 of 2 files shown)


modules/sdk-lib-mpc/test/unit/tss/ecdsa/dklsDsg.ts
```diff
@@ -409,4 +409,44 @@ describe('DKLS Dsg 2x3', function () {
...
+      vector.derivationPath,
+      crypto.createHash('sha256').update(Buffer.from(vector.msgToSign, 'hex')).digest()
+    );
+    await restoredParty.setSession(session);
+
+    // Restore the WASM session and call _deserializeState directly.
+    // Before the fix, this would throw "Invalid State: [object Object]".
+    (restoredParty as any)._restoreSession();
+    (restoredParty as any)._deserializeState();
+    (restoredParty as any).dsgState.should.equal(DklsTypes.DsgState.Round4);
```
zahin-mohammad, line 450: `as any` needed since these are private methods.

mrdanish26 (approved):

mohammadalfaiyazbitgo (approved):

PR #8482, opened 2026-04-10
feat(argon2): vendor hash-wasm v4.12.0 as @bitgo/argon2

pranavjain97:
Vendor the argon2 subset of hash-wasm as a new @bitgo/argon2 module, following the same pattern as @bitgo/sjcl. The pre-built UMD bundle (~29KB) contains argon2 and blake2b WASM binaries embedded as base64 with zero runtime dependencies.

Exports: argon2id, argon2i, argon2d, argon2Verify with full TypeScript type definitions. Includes verify-vendor.sh script for reproducible re-vendoring from npm.

This module is a prerequisite for migrating BitGoJS encryption from SJCL (PBKDF2-SHA256 @ 10k iterations, ~1ms per guess on modern hardware) to Argon2id, which is memory-hard and GPU-resistant.

TICKET: WCN-29


sachushaji:
@claude

modules/argon2/argon2.umd.min.js
zahin-mohammad, line 1: **Supply chain risk: opaque WASM binaries**

This single minified line contains two base64-encoded WASM binaries (argon2 + blake2b). No reviewer can meaningfully audit what the WASM does by reading the diff — it's a trust-on-faith inclusion of compiled cryptographic code.

For a library handling **password hashing in a financial SDK**, this is a significant audit gap. Consider:
- Checking in the decoded `.wasm` files alongside this bundle so the binary hashes can be independently verified against a source build.
- Documenting the exact git commit/tag from `github.com/Daninet/hash-wasm` that produced v4.12.0, not just the npm version string.

modules/argon2/scripts/verify-vendor.sh
zahin-mohammad, line 1: **Supply chain: verification anchored to npm, not source**

This script verifies the vendored file against the **npm tarball**, but npm publication is itself a supply chain vector — a compromised publish would pass this check. For cryptographic code, the verification bar should be higher:

1. Pin the **expected SHA256 hash** in a committed file (e.g., `SHA256SUMS`) rather than computing it on the fly against whatever npm currently serves.
2. Ideally verify against a **reproducible build from source** (pinned git tag + deterministic WASM compilation), not just an npm checksum.
3. Pin the specific git commit from `github.com/Daninet/hash-wasm` that corresponds to v4.12.0 so auditors can trace the full provenance chain.

modules/argon2/test/argon2.test.ts
zahin-mohammad, line 1: **Auditability: no known-answer test (KAT) vectors**

These tests verify behavioral properties (determinism, different outputs for different inputs, correct prefixes) but never assert that a specific input produces a **specific known-correct output**. Without KATs from RFC 9106 or the Argon2 reference implementation, a subtly broken or backdoored WASM binary would pass every test here.

Please add at least 3–4 test cases that assert exact hex output for fixed inputs against published RFC 9106 / reference test vectors. This is the minimum bar for verifying the WASM actually implements argon2 correctly.

modules/argon2/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@bitgo/argon2",
+  "version": "1.0.0",
+  "private": true,
```
zahin-mohammad: **Release risk: `"private": true` blocks npm publish**

If any downstream `@bitgo/*` package declares `@bitgo/argon2` as a dependency, SDK consumers installing from npm won't be able to resolve it. Is the intent to keep this private and only consume it via workspace linking? If so, that should be documented. If this is meant to be published, remove the flag.

modules/argon2/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/BitGo/BitGoJS.git",
+    "directory": "modules/argon2"
+  },
+  "publishConfig": {
+    "access": "public"
+  }
+}
```
zahin-mohammad, line 27: **Release risk: missing standard monorepo scripts and ESM support**

Several gaps vs. other modules in the monorepo:

1. **No scripts** — no `build`, `test`, `lint`, `clean`. CI pipelines that iterate workspace packages expecting these will skip or break on this module.
2. **No `"module"` or `"exports"` field** — only `"main"` is set. Given the documented [dual ESM/CJS build pattern](../../docs/esm.md) and that this targets browser use (WASM), the lack of ESM support is a gap for tree-shaking and modern bundlers.
3. **Not added to root `package.json` workspaces** — only `tsconfig.packages.json` was updated. `yarn install` won't link this package, and workspace-level commands won't discover it.

modules/argon2/index.d.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * @bitgo/argon2 - Vendored from hash-wasm v4.12.0
+ * https://github.com/Daninet/hash-wasm
+ * MIT License - Copyright (c) 2020 Dani Biro
+ */
+
+export type ITypedArray = Uint8Array | Uint16Array | Uint32Array;
+export type IDataType = string | Buffer | ITypedArray;
```
zahin-mohammad: **Release risk: `Buffer` type without `@types/node` dependency**

This references the `Buffer` type, which requires `@types/node`. In browser-only TypeScript projects consuming this package, this will cause a compilation error. Consider using `ArrayBuffer | ArrayBufferView` or adding a conditional type that doesn't depend on Node globals.

modules/argon2/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "allowJs": true,
+    "declaration": false,
+    "declarationMap": false,
+    "outDir": "./dist",
+    "rootDir": "./"
+  },
+  "include": [],
```
zahin-mohammad: **Auditability: hand-written types not validated against JS exports**

`"include": []` means nothing is compiled — the hand-written `index.d.ts` is never checked against the actual JS bundle's exports. Type drift is inevitable as the vendored version changes. Consider at least a smoke-test `.ts` file in `include` that imports and calls each export to catch type/export mismatches at build time.

zahin-mohammad (changes requested):
## Review: Supply Chain, Auditability & Release Risks

The vendoring approach (following `@bitgo/sjcl` precedent) is sound in principle, and the motivation — migrating from PBKDF2-SHA256 to Argon2id — is the right move. However, there are gaps that should be addressed before merging, given this is cryptographic code in a financial SDK.

### Supply Chain Risk (HIGH)
- The vendored bundle contains **opaque base64-encoded WASM binaries** that cannot be reviewed from the diff. Trust is anchored entirely to an npm tarball from a single maintainer.
- `verify-vendor.sh` verifies against npm, not against a reproducible source build. A compromised npm publish would pass verification. Pin the expected SHA256 and trace provenance to a specific git commit.

### Auditability (HIGH)
- **No RFC 9106 known-answer test vectors.** The test suite verifies behavioral properties but never asserts exact output against published reference vectors. A subtly broken or backdoored WASM implementation would pass all current tests. This is the single most important gap to close.
- Hand-written `index.d.ts` with `"include": []` in tsconfig means types are never validated against the JS exports.

### Release Integration (MEDIUM)
- `"private": true` will block npm publish if downstream packages depend on this.
- Missing standard monorepo scripts (`build`, `test`, `lint`, `clean`), no ESM support, not added to root workspaces.
- `Buffer` type in `index.d.ts` without `@types/node` dependency breaks browser-only TS consumers.

See inline comments for specifics on each file.

modules/argon2/package.json
pranavjain97: yeah this was temp until the package was added as trusted publisher on npm

modules/argon2/package.json
pranavjain97: 1. Scripts: Added `test` and verify in the latest push. build/lint/clean are not applicable -- this is a pre-built vendored file with no compilation step.
2. ESM: The package ships a single UMD bundle that works in both CJS and browser contexts. There's no tree-shaking benefit since it's one minified file, so an "exports" field adds no value here.
3. Workspaces: Already covered by the "modules/*" glob in root package.json -- yarn install resolves it as a workspace link and other packages can import from it.

modules/argon2/argon2.umd.min.js
pranavjain97: Git commit (373b796) and SHA256 of the vendored file are documented in PROVENANCE.md. CI verifies the pinned hash on every PR. KAT tests pin exact hex outputs for all three variants, cross-validated against argon2-cffi (C reference impl) -- a corrupted WASM would fail these.

Reproducible builds from source are possible but require setting up the hash-wasm build toolchain. Since we vendor once and never auto-update from npm, the practical risk is limited to the initial vendor, which is locked down. Happy to pursue if appsec requires it.

zahin-mohammad (approved):

PR #8495, opened 2026-04-13
feat(sdk-api): add v2 encrypt/decrypt using Argon2id

pranavjain97:
Add encryptV2() and decryptV2() alongside existing v1 SJCL functions. V2 uses Argon2id (m=64MiB, t=3, p=4) for KDF and WebCrypto AES-256-GCM for symmetric encryption. Self-describing JSON envelope stores all parameters for forward compatibility.

Existing v1 encrypt/decrypt is untouched. No call site changes.

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -40,7 +40,7 @@ import {
   toBitgoRequest,
   verifyResponseAsync,
 } from './api';
-import { decrypt, encrypt } from './encrypt';
+import { decrypt, decryptAsync, encrypt } from './encrypt';
 import { verifyAddress } from './v1/verifyAddress';
 import {
   AccessTokenOptions,
@@ -729,6 +729,29 @@ export class BitGoAPI implements BitGoBase {
     } catch (error) {
       if (error.message.includes("ccm: tag doesn't match")) {
         error.message = 'password error - ' + error.message;
+      }
+      throw error;
+    }
+  }
+
+  /**
+   * Async decrypt that auto-detects v1 (SJCL) or v2 (Argon2id).
+   * Migration path from sync decrypt() -- use this before the breaking release.
+   */
+  async decryptAsync(params: DecryptOptions): Promise<string> {
+    params = params || {};
+    common.validateParams(params, ['input', 'password'], []);
+    if (!params.password) {
+      throw new Error(`cannot decrypt without password`);
+    }
+    try {
+      return await decryptAsync(params.password, params.input);
+    } catch (error) {
+      if (
+        error.message.includes("ccm: tag doesn't match") ||
+        error.message.includes('The operation failed for an operation-specific reason')
+      ) {
+        throw new Error('incorrect password');
       }
       throw error;
     }
--- a/modules/sdk-api/src/encrypt.ts
+++ b/modules/sdk-api/src/encrypt.ts
@@ -1,5 +1,7 @@
+import { base64String, boundedInt, decodeWithCodec } from '@bitgo/sdk-core';
 import * as sjcl from '@bitgo/sjcl';
 import { randomBytes } from 'crypto';
+import * as t from 'io-ts';
 
 /** Default Argon2id parameters per RFC 9106 second recommendation
  * @see https://www.rfc-editor.org/rfc/rfc9106#section-4
@@ -26,15 +28,17 @@ const ARGON2_MAX = {
 /** AES-256-GCM IV length in bytes */
 const GCM_IV_LENGTH = 12;
 
-export interface V2Envelope {
-  v: 2;
-  m: number;
-  t: number;
-  p: number;
-  salt: string;
-  iv: string;
-  ct: string;
-}
+const V2EnvelopeCodec = t.type({
+  v: t.literal(2),
+  m: boundedInt(1, ARGON2_MAX.memorySize, 'memorySize'),
+  t: boundedInt(1, ARGON2_MAX.iterations, 'iterations'),
+  p: boundedInt(1, ARGON2_MAX.parallelism, 'parallelism'),
+  salt: base64String,
+  iv: base64String,
+  ct: base64String,
+});
+
+export type V2Envelope = t.TypeOf<typeof V2EnvelopeCodec>;
 
 /**
  * convert a 4 element Uint8Array to a 4 byte Number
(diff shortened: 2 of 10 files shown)


modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -734,6 +734,29 @@ export class BitGoAPI implements BitGoBase {
...
+   * Migration path from sync decrypt() -- use this before the breaking release.
+   */
+  async decryptAsync(params: DecryptOptions): Promise<string> {
+    params = params || {};
+    common.validateParams(params, ['input', 'password'], []);
+    if (!params.password) {
+      throw new Error(`cannot decrypt without password`);
+    }
+    try {
+      return await decryptAsync(params.password, params.input);
```
zahin-mohammad, line 748: This doesn't cause a naming conflict? To import `decryptAsync` and then define it in the file again?

zahin-mohammad: Oh this is a class function. That's why.

modules/sdk-api/src/encrypt.ts
```diff
@@ -1,6 +1,29 @@
+import { argon2id } from '@bitgo/argon2';
 import * as sjcl from '@bitgo/sjcl';
 import { randomBytes } from 'crypto';
 
+/** Default Argon2id parameters per RFC 9106 second recommendation */
```
zahin-mohammad: Can you link this?

modules/sdk-api/src/encrypt.ts
```diff
@@ -60,3 +83,124 @@ export function encrypt(
...
+/**
+ * Async decrypt that auto-detects v1 (SJCL) or v2 (Argon2id + AES-256-GCM)
+ * from the JSON envelope's `v` field.
+ *
+ * This is the migration path from sync `decrypt()`. Clients should move to
+ * `await decryptAsync()` before the breaking release that makes `decrypt()` async.
+ */
+export async function decryptAsync(password: string, ciphertext: string): Promise<string> {
+  try {
+    const envelope = JSON.parse(ciphertext);
```
zahin-mohammad, line 114: worth io-ts decoding?

modules/sdk-api/src/encrypt.ts
```diff
@@ -60,3 +83,124 @@ export function encrypt(
...
+    if (envelope.v === 2) {
+      return await decryptV2(password, ciphertext);
+    }
+  } catch {
+    // Not valid JSON or no v field -- fall through to v1
```
zahin-mohammad: What if it is a valid json, and is v2 and still fails.

zahin-mohammad: i.e. wrong password

modules/sdk-core/src/bitgo/bitgoBase.ts
```diff
@@ -15,6 +15,7 @@ export interface BitGoBase {
   wallets(): any; // TODO - define v1 wallets type
   coin(coinName: string): IBaseCoin; // need to change it to BaseCoin once it's moved to @bitgo/sdk-core
   decrypt(params: DecryptOptions): string;
```
zahin-mohammad, line 17: are you going to mark this method as deprecated in a future pr?

pranavjain97: yeah after the 60d window once we flip the default to v2

modules/sdk-api/src/encrypt.ts
pranavjain97: fixed

modules/sdk-api/src/encrypt.ts
pranavjain97: not really, its an internal format we produce, not external input

modules/sdk-api/src/encrypt.ts
```diff
@@ -60,3 +84,130 @@ export function encrypt(
...
+  return sjcl.decrypt(password, ciphertext);
+}
+
+/**
+ * Derive a 256-bit key from a password using Argon2id.
+ */
+async function deriveKeyV2(
+  password: string,
+  salt: Uint8Array,
+  params: { memorySize: number; iterations: number; parallelism: number }
```
mrdanish26, line 133: m, t, and p are read from the parsed JSON envelope and passed straight into Argon2 with no caps. Because ciphertext is often untrusted, an attacker could set huge parameters and force excessive CPU/memory work on every decrypt attempt. Can we enforce maximum allowed values here?

zahin-mohammad: The structure and valid values could easily be encoded in an io-ts codec.

pranavjain97: good catch, added io-ts codec for this

mrdanish26 (approved):

modules/sdk-api/src/encrypt.ts
zahin-mohammad: now that you have the codec, can you use it here.

zahin-mohammad: I think you can do `JsonFromString.pipe(V2EnvelopeCodec)` instead of JSON.parse.

modules/sdk-api/src/encrypt.ts
```diff
@@ -60,3 +99,128 @@ export function encrypt(
...
+ * `await decryptAsync()` before the breaking release that makes `decrypt()` async.
+ */
+export async function decryptAsync(password: string, ciphertext: string): Promise<string> {
+  let isV2 = false;
+  try {
+    // Only peeking at the v field to route; this is an internal format we produce, not external input.
+    const envelope = JSON.parse(ciphertext);
+    isV2 = envelope.v === 2;
+  } catch {
+    // Not valid JSON -- fall through to v1
```
zahin-mohammad, line 117: Isn't v1 also supposed to be valid JSON? Why are you falling through instead of throwing?

modules/sdk-api/src/encrypt.ts
```diff
@@ -60,3 +99,128 @@ export function encrypt(
...
+
+/**
+ * Derive a 256-bit key from a password using Argon2id.
+ */
+async function deriveKeyV2(
+  password: string,
+  salt: Uint8Array,
+  params: { memorySize: number; iterations: number; parallelism: number }
+): Promise<CryptoKey> {
+  const { argon2id } = await import('@bitgo/argon2');
```
zahin-mohammad, line 135: Why the dynamic import?

modules/sdk-api/src/encrypt.ts
```diff
@@ -60,3 +99,128 @@ export function encrypt(
...
+ */
+export async function decryptV2(password: string, ciphertext: string): Promise<string> {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(ciphertext);
+  } catch {
+    throw new Error('v2 decrypt: invalid JSON envelope');
+  }
+
+  const envelope = decodeWithCodec(V2EnvelopeCodec, parsed, 'v2 decrypt: invalid envelope');
```
zahin-mohammad, line 211: You can likely do both the JSON parsing and the codec decoding all in io-ts 

`JsonFromString.pipe(V2EnvelopeCodec)` or something similiar.

zahin-mohammad (approved):
minor nits, can address in a follow up.

modules/sdk-api/src/encrypt.ts
pranavjain97: the codecs are in sdk-core, and we already decode downstream there. here it only checks the `v` so not worth adding a new dep

modules/sdk-api/src/encrypt.ts
pranavjain97: yeah currently its just acting as a router for v1/v2 targeted error messages. can throw early in a followup

modules/sdk-api/src/encrypt.ts
pranavjain97: Doesnt add much value, i'll make this static in followup

modules/sdk-api/src/encrypt.ts
pranavjain97: sdk-api doesn't have io-ts-types as a dependency

pranavjain97 (commented):
will address comments in followup, non blocking

PR #8508, opened 2026-04-14
feat(sdk-api): hkdf caching encryption for multi call operations

pranavjain97:
## Summary

  Adds v2 encryption (Argon2id + AES-256-GCM) alongside existing v1 SJCL, with an HKDF session caching layer for multi-call operations.

  - **encryptV2 / decryptV2**: Argon2id (m=64MiB, t=3, p=4) KDF + AES-256-GCM. Self-describing JSON envelope with all parameters.
  - **decryptAsync**: Auto-detects v1/v2 from envelope `v` field. Non-breaking migration path from sync `decrypt()`.
  - **io-ts codec**: V2EnvelopeCodec validates structure and caps Argon2id parameters to prevent DoS.
  - **createEncryptionSession**: Runs Argon2id once, derives per-call AES keys via HKDF (<1ms each). For multi-encrypt flows like MPC signing.

  Existing v1 encrypt/decrypt is untouched. Call site usages in the wallet flows will be done in a separate PR while being backwards compatible with sjcl.

### DIFF
--- a/modules/sdk-api/src/encrypt.ts
+++ b/modules/sdk-api/src/encrypt.ts
@@ -56,11 +56,10 @@ export function decrypt(password: string, ciphertext: string): string {
 export async function decryptAsync(password: string, ciphertext: string): Promise<string> {
   let isV2 = false;
   try {
-    // Peek at v field only to route -- internal format we produce, not external input.
     const envelope = JSON.parse(ciphertext);
     isV2 = envelope.v === 2;
   } catch {
-    // Not valid JSON -- fall through to v1.
+    throw new Error('decrypt: ciphertext is not valid JSON');
   }
   if (isV2) {
     // Do not catch: wrong password on v2 must not silently fall through to v1.
--- a/modules/sdk-api/src/encrypt.ts
+++ b/modules/sdk-api/src/encrypt.ts
@@ -1,44 +1,7 @@
-import { base64String, boundedInt, decodeWithCodec } from '@bitgo/sdk-core';
 import * as sjcl from '@bitgo/sjcl';
 import { randomBytes } from 'crypto';
-import * as t from 'io-ts';
 
-/** Default Argon2id parameters per RFC 9106 second recommendation
- * @see https://www.rfc-editor.org/rfc/rfc9106#section-4
- */
-const ARGON2_DEFAULTS = {
-  memorySize: 65536, // 64 MiB in KiB
-  iterations: 3,
-  parallelism: 4,
-  hashLength: 32, // 256-bit key
-  saltLength: 16, // 128-bit salt
-} as const;
-
-/** Maximum allowed Argon2id parameters to prevent DoS via crafted envelopes.
- * memorySize: 256 MiB (4x default) -- caps memory allocation on untrusted input.
- * iterations: 16 -- caps CPU time.
- * parallelism: 16 -- caps thread count.
- */
-const ARGON2_MAX = {
-  memorySize: 262144,
-  iterations: 16,
-  parallelism: 16,
-} as const;
-
-/** AES-256-GCM IV length in bytes */
-const GCM_IV_LENGTH = 12;
-
-const V2EnvelopeCodec = t.type({
-  v: t.literal(2),
-  m: boundedInt(1, ARGON2_MAX.memorySize, 'memorySize'),
-  t: boundedInt(1, ARGON2_MAX.iterations, 'iterations'),
-  p: boundedInt(1, ARGON2_MAX.parallelism, 'parallelism'),
-  salt: base64String,
-  iv: base64String,
-  ct: base64String,
-});
-
-export type V2Envelope = t.TypeOf<typeof V2EnvelopeCodec>;
+import { decryptV2 } from './encryptV2';
 
 /**
  * convert a 4 element Uint8Array to a 4 byte Number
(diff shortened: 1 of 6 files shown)


modules/sdk-api/src/encrypt.ts
```diff
@@ -88,139 +38,32 @@ export function encrypt(
...
  */
 export async function decryptAsync(password: string, ciphertext: string): Promise<string> {
   let isV2 = false;
   try {
-    // Only peeking at the v field to route; this is an internal format we produce, not external input.
     const envelope = JSON.parse(ciphertext);
     isV2 = envelope.v === 2;
   } catch {
-    // Not valid JSON -- fall through to v1
+    throw new Error('decrypt: ciphertext is not valid JSON');
```
udayshanmugam, line 62: This is a hard error if all call sites do not pass JSON when compared to the previous implementation which was falling back to v1.
If all call sites will be passing JSON, this is not an issue

modules/sdk-api/src/encryptV2.ts
```diff
@@ -0,0 +1,188 @@
...
+    { name: 'HKDF', hash: 'SHA-256', salt: hkdfSalt, info: HKDF_INFO },
+    hkdfKey,
+    { name: 'AES-GCM', length: 256 },
+    false,
+    [usage]
+  );
+}
+
+export async function aesGcmEncrypt(key: CryptoKey, iv: Uint8Array, plaintext: string): Promise<Uint8Array> {
+  const ct = await crypto.subtle.encrypt({ name: 'AES-GCM', iv }, key, new TextEncoder().encode(plaintext));
```
udayshanmugam, line 104: This defaults to a 128 bit tagLength which is correct.
Explicitly set the tagLength to 128 bits in encrypt and decrypt. The code will rejects data that have a truncated tag

modules/sdk-api/src/encryptionSession.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+    const iv = new Uint8Array(Buffer.from(envelope.iv, 'base64'));
+    const ct = new Uint8Array(Buffer.from(envelope.ct, 'base64'));
+    const hkdfSalt = new Uint8Array(Buffer.from(envelope.hkdfSalt, 'base64'));
+    const aesKey = await hkdfDeriveAesKey(key, hkdfSalt, 'decrypt');
+    return aesGcmDecrypt(aesKey, iv, ct);
+  }
+
+  destroy(): void {
+    this.hkdfKey = null;
```
udayshanmugam, line 67: Also set the argon2SaltB64 to null
this.argon2SaltB64 = null;  // requires removing readonly

modules/sdk-api/src/encryptV2.ts
```diff
@@ -0,0 +1,188 @@
...
+  hashLength: 32, // 256-bit key
+  saltLength: 16, // 128-bit salt
+} as const;
+
+/** Maximum allowed Argon2id parameters to prevent DoS via crafted envelopes.
+ * memorySize: 256 MiB (4x default) -- caps memory allocation on untrusted input.
+ * iterations: 16 -- caps CPU time.
+ * parallelism: 16 -- caps thread count.
+ */
+const ARGON2_MAX = {
```
Logicwax, line 22: why not a ARGON2_MIN (to be used in boundedInt)?

modules/sdk-api/src/encrypt.ts
pranavjain97: All internal call sites pass JSON (both v1 SJCL envelopes and v2 envelopes are JSON). The hard error is intentional if something non-JSON reaches decryptAsync, it indicates a bug in the caller rather than a format mismatch.

modules/sdk-api/src/encryptV2.ts
pranavjain97: The lower bound is already enforced via boundedInt(1, ARGON2_MAX.memorySize, 'memorySize')

modules/sdk-api/src/encryptV2.ts
```diff
@@ -0,0 +1,192 @@
...
+]);
+
+export type V2Envelope = t.TypeOf<typeof V2EnvelopeCodec>;
+
+// Crypto helpers
+
+async function argon2Hash(
+  password: string,
+  salt: Uint8Array,
+  params: { memorySize: number; iterations: number; parallelism: number }
```
mohammadalfaiyazbitgo, line 62: nit: extract to named type

modules/sdk-api/src/encryptV2.ts
```diff
@@ -0,0 +1,192 @@
+import { argon2id } from '@bitgo/argon2';
+import { base64String, boundedInt, decodeWithCodec } from '@bitgo/sdk-core';
+import { randomBytes } from 'crypto';
```
mohammadalfaiyazbitgo, line 3: this won't work on browsers no?

mohammadalfaiyazbitgo (changes requested):

modules/sdk-api/src/encryptV2.ts
pranavjain97: hmm, i think these should be browser compatible. which import are u concerned about? 🤔

pranavjain97: Webpack replaces import { randomBytes } from 'crypto' with crypto-browserify at bundle time, which uses the browser-native crypto.getRandomValues() under the hood.

modules/sdk-api/src/encryptV2.ts
```diff
@@ -0,0 +1,192 @@
+import { argon2id } from '@bitgo/argon2';
```
mohammadalfaiyazbitgo, line 1: this was refactored from a dynamic import to a static one. 
Can we revert back to a dynamic one to reduce bundle size?

pranavjain97: i added it back since argon2 is ~29kb which is negligible compared to our bundle size, especially since this will eventually become the default encryption

mohammadalfaiyazbitgo (approved):

modules/sdk-api/src/encryptV2.ts
pranavjain97: will do in one of the followups

PR #8523, opened 2026-04-16
feat: implement EdDSA MPS DKG key gen orchestration

Marzooqa:
### Summary
Implements EddsaMPCv2Utils in sdk-core — the SDK orchestrator for EdDSA DKG using the MPS protocol.

### Changes
**sdk-lib-mpc — PGP comms layer**

- New commsLayer.ts with detachSignMpsMessage and verifyMpsMessage, exported as MPSComms.
- Adds MPSSignedMessage type to types.ts. 

**sdk-core — Key generation**

- New `EddsaMPCv2Utils` with a 2-round createKeychains flow: generates `Ed25519/X25519 GPG` keys for user and backup, runs `EddsaMPSDkg.DKG` for each party, PGP-signs and exchanges messages over POST `/mpc/generatekey` `(MPS-R1, MPS-R2)`, verifies BitGo's signed responses, cross-checks the derived commonPublicKey, then persists all three keychains.
- New `MPSKeyGenSenderForEnterprise` sender with `type: MPCv2`, `keyCurve: EdDSA`.

**bitgo — Unit tests**

- 3 tests: full DKG round-trip, per-participant keychain creation, and rejection on invalid BitGo PGP signature.

TICKET: WCI-5

### DIFF
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
@@ -1,0 +1,425 @@
+import * as assert from 'assert';
+import nock = require('nock');
+import * as openpgp from 'openpgp';
+
+import { EddsaMPCv2KeyGenRound1Request, EddsaMPCv2KeyGenRound2Request } from '@bitgo/public-types';
+import { TestableBG, TestBitGo } from '@bitgo/sdk-test';
+import { AddKeychainOptions, BaseCoin, common, ECDSAUtils, EDDSAUtils, Keychain, Wallet } from '@bitgo/sdk-core';
+import { EddsaMPSDkg, MPSComms, MPSTypes } from '@bitgo/sdk-lib-mpc';
+import { BitGo, BitgoGPGPublicKey } from '../../../../../../src';
+
+const MPCv2PartiesEnum = ECDSAUtils.MPCv2PartiesEnum;
+
+describe('TSS EdDSA MPCv2 Utils:', async function () {
+  const coinName = 'sol';
+  const walletId = '5b34252f1bf349930e34020a00000000';
+  const enterpriseId = '6449153a6f6bc20006d66771cdbe15d3';
+
+  let bgUrl: string;
+  let tssUtils: EDDSAUtils.EddsaMPCv2Utils;
+  let wallet: Wallet;
+  let bitgo: TestableBG & BitGo;
+  let baseCoin: BaseCoin;
+
+  let bitgoGpgKeyPair: openpgp.SerializedKeyPair<string> & { revocationCertificate: string };
+  let bitgoPrvKeyObj: openpgp.PrivateKey;
+  let constants: { mpc: { bitgoPublicKey: string; bitgoMPCv2PublicKey: string } };
+
+  beforeEach(async function () {
+    nock.cleanAll();
+    await nockGetBitgoPublicKeyBasedOnFeatureFlags(coinName, enterpriseId, bitgoGpgKeyPair);
+    nock(bgUrl).get('/api/v1/client/constants').times(10).reply(200, { ttl: 3600, constants });
+  });
+
+  before(async function () {
+    openpgp.config.rejectCurves = new Set();
+
+    bitgoGpgKeyPair = await openpgp.generateKey({
+      userIDs: [{ name: 'bitgo', email: 'bitgo@test.com' }],
+      curve: 'ed25519',
+      format: 'armored',
+    });
+
+    bitgoPrvKeyObj = await openpgp.readPrivateKey({ armoredKey: bitgoGpgKeyPair.privateKey });
+
+    constants = {
+      mpc: {
+        bitgoPublicKey: bitgoGpgKeyPair.publicKey,
+        bitgoMPCv2PublicKey: bitgoGpgKeyPair.publicKey,
+      },
+    };
+
+    bitgo = TestBitGo.decorate(BitGo, { env: 'mock' });
+    bitgo.initializeTestVars();
+    baseCoin = bitgo.coin(coinName);
+    bgUrl = common.Environments[bitgo.getEnv()].uri;
+
+    const walletData = {
+      id: walletId,
+      enterprise: enterpriseId,
+      coin: coinName,
+      coinSpecific: {},
+      multisigType: 'tss',
+    };
+    wallet = new Wallet(bitgo, baseCoin, walletData);
+    tssUtils = new EDDSAUtils.EddsaMPCv2Utils(bitgo, baseCoin, wallet);
+  });
+
+  after(function () {
+    nock.cleanAll();
+  });
+
+  describe('TSS key chains', async function () {
+    it('should generate TSS MPS keys', async function () {
+      const bitgoSession = new EddsaMPSDkg.DKG(3, 2, 2);
+      const bitgoState: { msg2?: MPSTypes.DeserializedMessage } = {};
+      const round1Nock = await nockMPSKeyGenRound1(bitgoSession, bitgoState, 1);
+      const round2Nock = await nockMPSKeyGenRound2(bitgoSession, bitgoState, 1);
+      const addKeyNock = await nockAddKeyChain(coinName, 3);
+
+      const params = {
+        passphrase: 'test',
+        enterprise: enterpriseId,
+        originalPasscodeEncryptionCode: '123456',
+      };
+
+      const { userKeychain, backupKeychain, bitgoKeychain } = await tssUtils.createKeychains(params);
+
+      assert.ok(round1Nock.isDone());
+      assert.ok(round2Nock.isDone());
+      assert.ok(addKeyNock.isDone());
+
+      assert.ok(userKeychain);
+      assert.equal(userKeychain.source, 'user');
+      assert.ok(userKeychain.commonKeychain);
+      assert.ok(userKeychain.encryptedPrv);
+      assert.ok(bitgo.decrypt({ input: userKeychain.encryptedPrv, password: params.passphrase }));
+
+      assert.ok(backupKeychain);
+      assert.equal(backupKeychain.source, 'backup');
+      assert.ok(backupKeychain.commonKeychain);
+      assert.ok(backupKeychain.encryptedPrv);
+      assert.ok(bitgo.decrypt({ input: backupKeychain.encryptedPrv, password: params.passphrase }));
+
+      assert.ok(bitgoKeychain);
+      assert.equal(bitgoKeychain.source, 'bitgo');
+      assert.ok(bitgoKeychain.commonKeychain);
+
+      assert.equal(userKeychain.commonKeychain, backupKeychain.commonKeychain);
+      assert.equal(userKeychain.commonKeychain, bitgoKeychain.commonKeychain);
+    });
+
+    it('should create TSS key chains', async function () {
+      const fakeCommonKeychain = 'a'.repeat(64);
+
+      const nockPromises = [
+        nockKeychain({
+          coin: coinName,
+          keyChain: { id: '1', pub: '1', type: 'tss', source: 'user', reducedEncryptedPrv: '' },
+          source: 'user',
+        }),
+        nockKeychain({
+          coin: coinName,
+          keyChain: { id: '2', pub: '2', type: 'tss', source: 'backup', reducedEncryptedPrv: '' },
+          source: 'backup',
+        }),
+        nockKeychain({
+          coin: coinName,
+          keyChain: { id: '3', pub: '3', type: 'tss', source: 'bitgo', reducedEncryptedPrv: '' },
+          source: 'bitgo',
+        }),
+      ];
+      const [nockedUserKeychain, nockedBackupKeychain, nockedBitGoKeychain] = await Promise.all(nockPromises);
+
+      const [userKeychain, backupKeychain, bitgoKeychain] = await Promise.all([
+        tssUtils.createParticipantKeychain(
+          MPCv2PartiesEnum.USER,
+          fakeCommonKeychain,
+          Buffer.from('userPrivate'),
+          Buffer.from('userReduced'),
+          'passphrase'
+        ),
+        tssUtils.createParticipantKeychain(
+          MPCv2PartiesEnum.BACKUP,
+          fakeCommonKeychain,
+          Buffer.from('backupPrivate'),
+          Buffer.from('backupReduced'),
+          'passphrase'
+        ),
+        tssUtils.createParticipantKeychain(MPCv2PartiesEnum.BITGO, fakeCommonKeychain),
+      ]);
+
+      assert.ok(userKeychain);
+      assert.equal(bitgoKeychain.source, 'bitgo');
+
+      assert.equal(userKeychain.id, nockedUserKeychain.id);
+      assert.equal(backupKeychain.id, nockedBackupKeychain.id);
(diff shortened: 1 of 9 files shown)


modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,298 @@
...
+    ).reverse();
+
+    // Get the BitGo public key based on user/enterprise feature flags
+    // If it doesn't work, use the default public key from the constants
+    const bitgoPublicGpgKeyArmored = (
+      (await this.getBitgoGpgPubkeyBasedOnFeatureFlags(params.enterprise, true)) ?? this.bitgoMPCv2PublicGpgKey
+    ).armor();
+
+    if (envRequiresBitgoPubGpgKeyConfig(this.bitgo.getEnv())) {
+      assert(bitgoPublicGpgKeyArmored, 'BitGo GPG public key is required');
```
zahin-mohammad: This assertion is weaker than the ECDSA equivalent in `ecdsaMPCv2.ts:74` — there we call `assert(isBitgoMpcPubKey(bitgoPublicGpgKey, 'mpcv2'), 'Invalid BitGo GPG public key')`. Here we only assert the armored key is truthy, which will never fail because the `??` fallback always yields a key. In prod/test envs this should verify the key matches one of the known hardcoded BitGo MPC pubkeys. Please restore `isBitgoMpcPubKey(bitgoPublicGpgKeyArmored, 'mpcv2')` (or add an `'mpcv2-eddsa'` variant if the key differs).

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,298 @@
...
+const MPS_ROUND_2 = 'MPS-R2';
+
+export class EddsaMPCv2Utils extends BaseEddsaUtils {
+  /** @inheritdoc */
+  async createKeychains(params: {
+    passphrase: string;
+    enterprise: string;
+    originalPasscodeEncryptionCode?: string;
+  }): Promise<KeychainsTriplet> {
+    const userKey = await pgp.generateKey({
```
zahin-mohammad: ECDSA uses the shared `generateGPGKeyPair('secp256k1')` helper from `../../opengpgUtils`. Please route through `generateGPGKeyPair('ed25519')` (extending it if needed) rather than inlining `pgp.generateKey` with a hard-coded userID and subkey layout. Keeps GPG key setup consistent across TSS flows and avoids divergence if we later add defaults (expiry, key flags, etc.).

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,298 @@
...
+      format: 'object',
+    });
+    const userGpgKey = userKey.privateKey;
+    const userGpgPublicKey = await userKey.publicKey.armor();
+    const userPk = Buffer.from(
+      ((await userGpgKey.getEncryptionKey()).keyPacket.publicParams as { Q: Uint8Array }).Q
+    ).subarray(1);
+    const userSk = Buffer.from(
+      ((await userGpgKey.getEncryptionKey()).keyPacket as unknown as { privateParams: { d: Uint8Array } }).privateParams
+        .d
```
zahin-mohammad: Reaching into `keyPacket.publicParams.Q` / `privateParams.d` with `as unknown as {...}` casts leaks openpgp internals into sdk-core. Compare to ECDSA, where key material stays inside `DklsComms` (sdk-lib-mpc). Please push these primitives down into `sdk-lib-mpc` — e.g. `MPSComms.extractEd25519KeyPair(privateKey)` returning `{ sk, pk }` — so sdk-core only deals with Buffers. That also encapsulates the little-endian scalar reverse (easy to get wrong) in one place.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,298 @@
...
+import { EddsaMPSDkg, MPSComms, MPSTypes } from '@bitgo/sdk-lib-mpc';
+import { KeychainsTriplet } from '../../../baseCoin';
+import { AddKeychainOptions, Keychain, KeyType } from '../../../keychain';
+import { envRequiresBitgoPubGpgKeyConfig } from '../../../tss/bitgoPubKeys';
+import { MPCv2PartiesEnum } from '../ecdsa/typesMPCv2';
+import { BaseEddsaUtils } from './base';
+import { MPSKeyGenSenderForEnterprise } from './eddsaMPCv2KeyGenSender';
+
+/** Round identifiers sent in the `round` field of each API request */
+const MPS_ROUND_1 = 'MPS-R1';
```
zahin-mohammad: ECDSA uses `MPCv2KeyGenState` from `@bitgo/public-types` as the typed round enum. String literals `'MPS-R1' / 'MPS-R2'` are fine as a stopgap, but please add an `MPSKeyGenState` (or similar) enum to public-types so the server schema and SDK stay in lockstep. Server-side parser errors on typos will be much easier to catch.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,298 @@
...
+    return {
+      userKeychain,
+      backupKeychain,
+      bitgoKeychain,
+    };
+  }
+
+  // #region keychain utils
+  async createParticipantKeychain(
+    participantIndex: MPCv2PartiesEnum,
```
zahin-mohammad: Please type `participantIndex` as `MPCv2PartyFromStringOrNumber` for parity with `EcdsaMPCv2Utils.createParticipantKeychain`. The strict enum here means callers passing the string form (which the io-ts union permits in ECDSA) would need a cast. Keeping the same signature shape makes the EdDSA util a drop-in for callers already wired against ECDSA.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2KeyGenSender.ts
```diff
@@ -0,0 +1,30 @@
...
+  bitgo: BitGoBase,
+  enterprise: string
+): {
+  round1: EddsaMPCv2KeyGenSendFn<EddsaMPCv2KeyGenRound1Request, EddsaMPCv2KeyGenRound1Response>;
+  round2: EddsaMPCv2KeyGenSendFn<EddsaMPCv2KeyGenRound2Request, EddsaMPCv2KeyGenRound2Response>;
+} {
+  function send<Req, Res>(round: string, payload: Req): Promise<Res> {
+    return bitgo
+      .post(bitgo.url('/mpc/generatekey', 2))
+      .send({ enterprise, type: KeyGenTypeEnum.MPCv2, keyCurve: 'EdDSA', round, payload })
```
zahin-mohammad: Two things on this line:

1. `keyCurve: 'EdDSA'` is a hardcoded string. Please add a `KeyCurveEnum` (or equivalent) to `@bitgo/public-types` and import it — same pattern as `KeyGenTypeEnum.MPCv2`.
2. The sender shape `{ round1, round2 }` differs from ECDSA's generic `KeyGenSenderForEnterprise<T>(bitgo, ent) → sendFn`. Consider matching the ECDSA shape (single generic sender function) so that downstream consumers (custom senders, OVC, test harnesses) have a uniform extension point across curves.

modules/sdk-core/src/bitgo/utils/tss/eddsa/index.ts
```diff
@@ -13,3 +13,5 @@ export {
   SignatureShareType,
   TxRequest,
 } from '../baseTypes';
+
+export { EddsaMPCv2Utils } from './eddsaMPCv2';
```
zahin-mohammad: ECDSA's index uses `export *` for `ecdsaMPCv2` and `ecdsaMPCv2KeyGenSender`, which exposes the sender + type helpers. Here only the class is re-exported. Please also re-export `EddsaMPCv2KeyGenSendFn`, `MPSKeyGenSenderForEnterprise`, and any MPCv2 types so consumers don't have to import from deep paths.

zahin-mohammad (commented):
Reviewed the EdDSA MPS implementation against the ECDSA DKLS conventions. Behavior looks correct; leaving inline notes on drifts we should close before merging. Retrofit is expected to come later — not flagged.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
```diff
@@ -0,0 +1,311 @@
...
+        .reply(200, {
+          sessionId: 'bad-session',
+          bitgoMsg1: {
+            message: Buffer.from('garbage').toString('base64'),
+            signature: '-----BEGIN PGP SIGNATURE-----\nFAKE\n-----END PGP SIGNATURE-----',
+          },
+        });
+
+      await assert.rejects(tssUtils.createKeychains({ passphrase: 'test', enterprise: enterpriseId }));
+    });
```
zahin-mohammad, line 192: This is the only failure-path test. Please add symmetric coverage for the other invariants asserted in `eddsaMPCv2.ts`:

1. **Round-2 invalid BitGo signature** — mirrors this R1 test but targets `verifyMpsMessage(bitgoMsg2, ...)` at `eddsaMPCv2.ts:143`. Currently uncovered.
2. **Session ID mismatch** — `assert.equal(sessionId, sessionIdRound2)` at `eddsaMPCv2.ts:134`. Nock R2 to return a different `sessionId` and assert `createKeychains` rejects.
3. **commonPublicKey mismatch** — asserts at `eddsaMPCv2.ts:153-154` ("User/Backup computed public key does not match BitGo common public key"). Nock R2 to return a mutated `commonPublicKey` and assert rejection.

These are cheap to add and cover the core integrity checks that protect users from a malicious or buggy BitGo response.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
```diff
@@ -0,0 +1,311 @@
...
+        const backupPubKeyObj = await openpgp.readKey({ armoredKey: backupGpgPublicKey });
+
+        const userPk = Buffer.from(
+          ((await userPubKeyObj.getEncryptionKey()).keyPacket.publicParams as { Q: Uint8Array }).Q
+        ).subarray(1);
+        const backupPk = Buffer.from(
+          ((await backupPubKeyObj.getEncryptionKey()).keyPacket.publicParams as { Q: Uint8Array }).Q
+        ).subarray(1);
+        const bitgoSk = Buffer.from(
+          ((await bitgoPrvKeyObj.getDecryptionKeys())[0].keyPacket.privateParams as { d: Uint8Array }).d
```
zahin-mohammad: This nock handler mirrors production's openpgp-internals extraction (`keyPacket.privateParams.d`, `publicParams.Q`, the `.reverse()` for little-endian scalar). That makes the test a change-detector for the same code path rather than an independent verification.

Once the extraction moves into `sdk-lib-mpc` (per the main-code review comment), please have the nock handler call the same `MPSComms` helper so the test exercises the helper too. Today, a bug in the extraction logic would pass tests because prod + test do it identically.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
```diff
@@ -0,0 +1,311 @@
...
+      userIDs: [{ name: 'bitgo', email: 'bitgo@test.com' }],
+      curve: 'ed25519',
+      format: 'armored',
+    });
+
+    bitgoPrvKeyObj = await openpgp.readPrivateKey({ armoredKey: bitgoGpgKeyPair.privateKey });
+
+    constants = {
+      mpc: {
+        bitgoPublicKey: bitgoGpgKeyPair.publicKey,
```
zahin-mohammad, line 47: `bitgoPublicKey` and `bitgoMPCv2PublicKey` are both set to the same generated key, so there's no coverage for the feature-flag fallback branch at `eddsaMPCv2.ts:63-65` (`getBitgoGpgPubkeyBasedOnFeatureFlags(...) ?? this.bitgoMPCv2PublicGpgKey`). Consider adding a test where `getBitgoGpgPubkeyBasedOnFeatureFlags` returns `undefined` to confirm the fallback to `bitgoMPCv2PublicGpgKey` works and DKG still succeeds.

Relatedly, once the `isBitgoMpcPubKey` assertion is restored (see main-code review), please add a negative test where the BitGo pubkey doesn't match the hardcoded set — `createKeychains` should reject in prod/test envs.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
```diff
@@ -0,0 +1,311 @@
...
+          Buffer.from('backupReduced'),
+          'passphrase'
+        ),
+        tssUtils.createParticipantKeychain(MPCv2PartiesEnum.BITGO, fakeCommonKeychain),
+      ]);
+
+      assert.ok(userKeychain);
+      assert.equal(bitgoKeychain.source, 'bitgo');
+
+      assert.equal(userKeychain.id, nockedUserKeychain.id);
```
zahin-mohammad, line 155: The `createParticipantKeychain` tests use `'a'.repeat(64)` as a fake common keychain. Worth adding one assertion that the encrypted `reducedEncryptedPrv` for user/backup can be decrypted with the passphrase and round-trips to the original reduced private material — catches regressions in the browser-safe `btoa(String.fromCharCode.apply(null, ...))` encoding path, which is easy to break and not otherwise exercised.

zahin-mohammad (commented):
Test coverage notes — the happy path and one failure branch are covered, but several explicit invariants in the production code are untested. Inline suggestions below.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2KeyGenSender.ts
Marzooqa: I'll add `KeyCurveEnum` from `@bitgo/public-types` as a follow-up here https://github.com/BitGo/BitGoJS/pull/8536

Marzooqa: Scratch that, added and bumped the `public-types `version containing `KeyCurveEnum`

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
zahin-mohammad: Outstanding — the other review items are all addressed, but the fallback-branch test (where `getBitgoGpgPubkeyBasedOnFeatureFlags` returns `undefined` and we fall through to `this.bitgoMPCv2PublicGpgKey`) isn't covered yet. Fine to do as a follow-up. Approved.

zahin-mohammad (approved):
All review feedback addressed apart from the feature-flag fallback test, which can ship as a follow-up.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
Marzooqa: yeah I'll take it as a follow-up

PR #8528, opened 2026-04-16
docs(express): update descriptions coinsigntx

rishikeshdadam136:
Ticket: WCI-40

Reference - https://bitgo.slack.com/archives/C057BHBRG4B/p1764020095794829?thread_ts=1764018507.602879&cid=C057BHBRG4B

### DIFF
--- a/modules/express/src/typedRoutes/api/v2/coinSignTx.ts
+++ b/modules/express/src/typedRoutes/api/v2/coinSignTx.ts
@@ -7,7 +7,7 @@ import { BitgoExpressError } from '../../schemas/error';
  * Request parameters for signing a transaction
  */
 export const CoinSignTxParams = {
-  /** The coin type */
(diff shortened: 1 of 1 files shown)


modules/express/src/typedRoutes/api/v2/coinSignTx.ts
```diff
@@ -394,34 +394,12 @@ export const CoinSignTxResponse = {
 };
 
 /**
- * Sign a transaction for a specific coin
+ * This route is for users who would like to maintain their own keys, or otherwise
```
lokesh-bitgo: I think It should be in one line

rishikeshdadam136: Done

alextse-bg (approved):

lokesh-bitgo (approved):

PR #8539, opened 2026-04-16
fix(tss): verify recipients before signing

mrdanish26:
TICKET: WAL-375

## Description

Ensures `verifyTransaction` always receives recipient details before any signing material is sent to BitGo, preventing a compromised API from silently substituting `signableHex` with attacker-controlled bytes.

A `resolveEffectiveTxParams()` guard is added to both ECDSA and MPCv2 `signRequestBase()` paths. It resolves recipients from three sources in priority order:

1. **`txParams.recipients`** — normal case.
2. **`txRequest.intent.recipients`** — fallback for smart-contract interactions (e.g. Tempo `supplyController`) where native ETH amount = 0 so `buildParams` is empty but intent carries the recipients.
  3. **`NO_RECIPIENT_TX_TYPES`** exemption — types that legitimately carry no recipients; `verifyTransaction` handles validation internally for these. Throws `InvalidTransactionError` if none of the above apply.

A database query against `marts.wallet_platform.bitgo2__transaction_request` identified all intent types where `INTENT:recipients IS NULL OR ARRAY_SIZE(INTENT:recipients) = 0`. Two ECDSA-path types with significant
  record counts were missing from the exemption list:

  - enableToken (17,765 records) — TSS wallets call buildTokenEnablements() which does not populate buildParams.recipients
  - customTx (7,082 records) — DeFi/WalletConnect smart contract interactions; already exempted at populateIntent() server-side but inconsistently enforced at signing

Both have been added to NO_RECIPIENT_TX_TYPES and the abstractEthLikeNewCoins.ts bypass list.

 ## NO_RECIPIENT_TX_TYPES (8 types)

  | Type | Reason |
  |---|---|
  | `acceleration`, `fillNonce`, `transferToken`, `tokenApproval`, `consolidate`, `bridgeFunds` | Pre-existing exemptions |
  | `enableToken` | TSS wallets do not populate recipients for token enablement — confirmed via prod DB  |
  | `customTx` | DeFi/WalletConnect smart contract interactions — already exempted at `populateIntent()` server-side; must be consistent at signing  |

  Both `recipientUtils.ts` and the inline bypass list in `abstractEthLikeNewCoins.ts verifyTssTransaction` are kept in sync.

A previous PR (#8462) hard-threw on any empty recipients, breaking cases 2 and 3. It was reverted in #8538. This PR supersedes it.

 ## Testing

  **New unit tests — `tss/recipientUtils.ts` (16 tests):**
  - All 8 exempt types pass without recipients
  - `txParams.recipients` takes priority over intent
  - Intent fallback when `txParams` is empty or undefined
  - Throws for non-exempt types with no recipients

  **Updated integration tests — `ecdsaMPCv2/signTxRequest.ts` (13 tests):**
  - Replaced 4 false-confidence `.should.not.be.rejectedWith()` assertions with nock-backed `isDone()` assertions that prove signing completes end-to-end
  - Added acceptance tests for `enableToken` and `customTx` exemptions
  - Existing reject tests unchanged

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/base.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/base.ts
@@ -1,61 +1,12 @@
 import { secp256k1 } from '@noble/curves/secp256k1';
 
-import { IBaseCoin, TransactionParams } from '../../../baseCoin';
+import { IBaseCoin } from '../../../baseCoin';
 import baseTSSUtils from '../baseTSSUtils';
 import { KeyShare } from './types';
-import { BackupGpgKey, PopulatedIntent, TxRequest } from '../baseTypes';
+import { BackupGpgKey } from '../baseTypes';
 import { generateGPGKeyPair } from '../../opengpgUtils';
 import { BitGoBase } from '../../../bitgoBase';
 import { IWallet } from '../../../wallet';
-import { InvalidTransactionError } from '../../../errors';
-
-/**
- * Transaction types that legitimately carry no explicit recipients.
- * verifyTransaction handles no-recipient validation for these internally.
- * Mirrors the bypass list in abstractEthLikeNewCoins.ts verifyTssTransaction.
- */
-export const NO_RECIPIENT_TX_TYPES = new Set([
-  'acceleration',
-  'fillNonce',
-  'transferToken',
-  'tokenApproval',
-  'consolidate',
-  'bridgeFunds',
-]);
-
-/**
- * Resolves the effective txParams for TSS signing recipient verification.
- *
- * For smart contract interactions, recipients live in txRequest.intent.recipients
- * (native amount = 0, so buildParams is empty). Falls back to intent recipients
- * mapped to ITransactionRecipient shape when txParams.recipients is absent.
- *
- * Throws InvalidTransactionError if no recipients can be resolved and the
- * transaction type is not a known no-recipient type.
- */
-export function resolveEffectiveTxParams(
-  txRequest: TxRequest,
-  txParams: TransactionParams | undefined
-): TransactionParams {
-  const intentRecipients = (txRequest.intent as PopulatedIntent)?.recipients?.map((intentRecipient) => ({
-    address: intentRecipient.address.address,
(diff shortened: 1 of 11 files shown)


sachushaji:
@claude

modules/sdk-core/src/bitgo/utils/tss/ecdsa/base.ts
```diff
@@ -1,12 +1,61 @@
...
+  };
+
+  if (!effectiveTxParams.recipients?.length && !NO_RECIPIENT_TX_TYPES.has(effectiveTxParams.type ?? '')) {
+    throw new InvalidTransactionError(
+      'Recipient details are required to verify this transaction before signing. Pass txParams with at least one recipient.'
+    );
+  }
+
+  return effectiveTxParams;
+}
```
mohammadalfaiyazbitgo, line 58: what about `eddsa` ?

mohammadalfaiyazbitgo (changes requested):

modules/sdk-core/src/bitgo/utils/tss/ecdsa/base.ts
mrdanish26: `EdDSA` has the same vulnerability but I am deferring it rather than flying blind, applying the same guard naively breaks legitimate no-recipient flows across multiple EdDSA coins. The current exemption list was built for ECDSA/ETH only; EdDSA needs a much broader set (e.g. SO staking/closeAssociatedTokenAccount/customTx, ALGO keyreg, DOT ADDRESS_INIT, XLM trustlines, plus the full staking intent enum the server bypasses but the client guard has no knowledge of).

Will track in a follow-up ticket with proper per-coin coverage and tests: https://linear.app/bitgo/issue/WCN-196/security-issue-extend-wal-375-recipient-guard-to-eddsa-tss-signing

mohammadalfaiyazbitgo (dismissed):

zhongxishen (dismissed):

feng-bitgo (approved):

mohammadalfaiyazbitgo (approved):

PR #8553, opened 2026-04-17
refactor(express): splits decrpt into v1 and v2

rishikeshdadam136:
Ticket: WCI-184


modules/express/src/typedRoutes/api/v1/decrypt.ts
```diff
@@ -3,18 +3,22 @@ import { httpRoute, httpRequest, optional } from '@api-ts/io-ts-http';
...
 
 /**
- * Decrypt
+ * Decrypt (v1)
  *
- * @operationId express.decrypt
+ * Decrypts an encrypted string using the provided password.
+ *
+ * @operationId express.v1.decrypt
  * @tag express
```
lokesh-bitgo: Express

lokesh-bitgo (commented):
The changes look good. Have you verified and tested on your machine?

rishikeshdadam136:
> The changes look good. Have you verified and tested on your machine?

yes tested both v1 and v2 in local

pranavjain97 (dismissed):

lokesh-bitgo (dismissed):

modules/express/src/typedRoutes/api/v1/decrypt.ts
```diff
@@ -3,18 +3,22 @@ import { httpRoute, httpRequest, optional } from '@api-ts/io-ts-http';
...
 /**
- * Decrypt
+ * Decrypt (v1)
  *
- * @operationId express.decrypt
- * @tag express
+ * Decrypts an encrypted string using the provided password.
+ *
+ * @operationId express.v1.decrypt
+ * @tag Express
```
vmccarty:
```suggestion
 * Decrypt message (v1)
*
 * Decrypts an encrypted string using the provided password.
 *
 * @operationId express.v1.decrypt
 * @tag Express
 * @private
```

This needs to be marked internal. If we have v1 and v2 endpoints, we only publicly document the v2 endpoint.

modules/express/src/typedRoutes/api/v2/decrypt.ts
```diff
@@ -0,0 +1,26 @@
...
+import { DecryptRequestBody } from '../v1/decrypt';
+
+/**
+ * Decrypt
+ *
+ * Decrypt a ciphertext generated by encrypt route with provided password
+ *
+ * @operationId express.decrypt
+ * @tag Express
+ */
```
vmccarty, line 14:
```suggestion
/**
 * Decrypt message
 *
 * Decrypt a ciphertext generated by encrypt route with provided password
 *
 * @operationId express.decrypt
 * @tag Express
 * @public
 */
```

vmccarty (changes requested):
The "breaking" change is okay, but the endpoints need to be marked as either private or public. Not sure why CI didn't flag it here.

modules/express/src/typedRoutes/api/v1/decrypt.ts
vmccarty: We should also update the summaries to specify exactly what they're decrypting (a message).

vmccarty (approved):
PRLGTM 👍

rishikeshdadam136:
Resolved the merge conflicts and force pushed again

pranavjain97 (approved):

PR #8554, opened 2026-04-17
feat(hbar): add staking support via AccountUpdateBuilder

aditya-bitgo:
ticket: SI-361

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -58,6 +58,7 @@ module.exports = {
         'PX-',
         'QA-',
         'RA-',
+        'SI-',
         'SO-',
         'SC-',
         'ST-',
--- a/modules/sdk-coin-hbar/src/lib/accountUpdateBuilder.ts
+++ b/modules/sdk-coin-hbar/src/lib/accountUpdateBuilder.ts
@@ -1,0 +1,107 @@
+import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import Long from 'long';
+import { proto } from '@hashgraph/proto';
+import { BaseKey, BuildTransactionError, SigningError, TransactionType } from '@bitgo/sdk-core';
+import { Transaction } from './transaction';
+import { TransactionBuilder } from './transactionBuilder';
+import { buildHederaAccountID, isValidAddress, stringifyAccountId } from './utils';
+import { DEFAULT_SIGNER_NUMBER } from './constants';
+
+export class AccountUpdateBuilder extends TransactionBuilder {
+  private readonly _txBodyData: proto.CryptoUpdateTransactionBody;
+  private _accountId: string;
+  private _stakedNodeId?: Long;
+  private _declineStakingReward?: boolean;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+    this._txBodyData = new proto.CryptoUpdateTransactionBody();
+    this._txBody.cryptoUpdateAccount = this._txBodyData;
+  }
+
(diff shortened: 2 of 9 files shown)


aditya-bitgo:
@claude review this PR

Doddanna17 (dismissed):
**Missing unit tests.** Every builder in this SDK has a test file under `test/unit/transactionBuilder/` — `tokenAssociateBuilder.ts`, `transferBuilder.ts`, etc. No `accountUpdateBuilder.ts` was added. The CI unit-test passes because existing tests weren't broken. At minimum need: happy path stake/unstake, `validateMandatoryFields` throwing when `stakedNodeId` is missing, `signImplementation` max-signer check, and `initBuilder` round-trip from a serialized tx.

Doddanna17 (commented):
Added unit tests for `AccountUpdateBuilder` in `modules/sdk-coin-hbar/test/unit/transactionBuilder/accountUpdateBuilder.ts` covering stake, unstake, decline reward, round-trip serialization, and error paths (missing stakedNodeId, invalid account ID, max signers exceeded, invalid key). All 12 tests pass.

Doddanna17 (commented):
A few issues to address before merging:

`validateMandatoryFields()` doesn't call `super.validateMandatoryFields()` — base-class validations (signer count, fee, etc.) are silently bypassed. Add `super.validateMandatoryFields()` at the top.

`stakedNodeId` setter has no lower-bound guard — `-2`, `-99`, etc. pass validation and fail only at the network layer. Only `-1` is the valid Hedera unstake sentinel (confirmed via coins-sandbox `unstake.js` and `unstakeMultisig.js`). Add: `if (nodeId < -1) throw new BuildTransactionError('Invalid stakedNodeId: only -1 is a valid sentinel for unstaking')`.

`declineStakingReward` proto semantics: omitting `BoolValue` vs setting `{ value: false }` have different on-chain meanings. A test covering the "unset" case would prevent regression here.

Missing test: `stakedNodeId(-1)` + `declineStakingReward(true)` combined — the full unstake path is untested.

Doddanna17 (changes requested):
A few issues to address before merging:

validateMandatoryFields() doesn't call super.validateMandatoryFields() — base-class validations (signer count, fee, etc.) are silently bypassed. Add super.validateMandatoryFields() at the top.

stakedNodeId setter has no lower-bound guard — -2, -99, etc. pass validation and fail only at the network layer. Only -1 is the valid Hedera unstake sentinel (confirmed via coins-sandbox unstake.js and unstakeMultisig.js). Add: if (nodeId < -1) throw new BuildTransactionError('Invalid stakedNodeId: only -1 is a valid sentinel for unstaking').

declineStakingReward proto semantics: omitting BoolValue vs setting { value: false } have different [on-chain](https://bitgoinc.atlassian.net/browse/ON-chain) meanings. A test covering the "unset" case would prevent regression here.

Missing test: stakedNodeId(-1) + declineStakingReward(true) combined — the full unstake path is untested.

aditya-bitgo:
@Doddanna17 super.validateMandatoryFields() is already being called at line 66. Rest of the changes i;ve addressed.

Doddanna17 (approved):
All four issues from the previous review are addressed:
- `super.validateMandatoryFields()` called after stakedNodeId guard
- Lower-bound validation on `stakedNodeId` (`< -1`) present
- "Unset" declineReward test verifies proto omission behavior
- Combined `stakedNodeId(-1)` + `declineStakingReward(true)` test added

LGTM.

kaustubhbitgo (approved):

PR #8557, opened 2026-04-17
feat: add round domain separator to adata

zhongxishen:
TICKET: HSM-1513

This is what was left to be done from https://linear.app/bitgo/issue/HSM-36/unauthenticated-dsg-session-restore-in-dsgsetsession-enables-session#comment-fd042a25.

Specifically: 
"What remains valid
The absence of a round number in adata is a legitimate, though low-priority, gap. At the validateAdata() level, a Round 1 ciphertext is indistinguishable from a Round 2 ciphertext for the same transaction. We recommend addressing this as a hardening improvement, consistent with the approach taken in the HSM."

The approach is similar to what's done in hsm-firmware https://github.com/BitGo/hsm-firmware/blob/develop/src/bg/dkls23.c#L40-L41.


zahin-mohammad (commented):
can you add some context for this PR (what the problem is and how this fixes that) 🙏🏾

zhongxishen:
> can you add some context for this PR (what the problem is and how this fixes that) 🙏🏾

Added to the description of the PR.

zahin-mohammad (dismissed):

zahin-mohammad (commented):
sorry, missing one piece, can we add a test for this?

zhongxishen:
> sorry, missing one piece, can we add a test for this?

There were no tests for ECDSA MPCv2 for module `sdk-core`.  They're now added.

zahin-mohammad (approved):

PR #8560, opened 2026-04-19
refactor(express): split encrypt into v1 and v2

rishikeshdadam136:
Ticket: WCI-185

### DIFF
--- a/modules/express/src/typedRoutes/api/v1/encrypt.ts
+++ b/modules/express/src/typedRoutes/api/v1/encrypt.ts
@@ -1,0 +1,34 @@
+import * as t from 'io-ts';
+import { httpRoute, httpRequest, optional } from '@api-ts/io-ts-http';
+import { BitgoExpressError } from '../../schemas/error';
+
+export const EncryptRequestBody = {
+  /** Plaintext message which should be encrypted */
+  input: t.string,
+  /** Password which should be used to encrypt message */
+  password: optional(t.string),
+  adata: optional(t.string),
+};
+
+/**
+ * Encrypt message (v1)
+ *
+ * Symmetrically encrypt an arbitrary message with provided password
+ *
+ * @operationId express.v1.encrypt
+ * @tag Express
+ * @private
+ */
+export const PostV1Encrypt = httpRoute({
+  path: '/api/v1/encrypt',
+  method: 'POST',
+  request: httpRequest({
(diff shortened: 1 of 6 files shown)


lokesh-bitgo (commented):
Changes looks good. Have you tested both endpoint locally ?

lokesh-bitgo (commented):
Changes looks good. Have you tested this locally ?

rishikeshdadam136:
> Changes looks good. Have you tested both endpoint locally ?

Yes,tested locally

lokesh-bitgo (dismissed):

modules/express/src/typedRoutes/api/v1/encrypt.ts
```diff
@@ -3,19 +3,21 @@ import { httpRoute, httpRequest, optional } from '@api-ts/io-ts-http';
...
 };
 
 /**
- * Encrypt
+ * Encrypt (v1)
  *
- * @operationId express.encrypt
- * @tag express
+ * @operationId express.v1.encrypt
+ * @tag Express
```
vmccarty:
```suggestion
 * Encrypt message (v1)
*
 * @operationId express.v1.encrypt
 * @tag Express
 * @private
```

modules/express/src/typedRoutes/api/v2/encrypt.ts
```diff
@@ -0,0 +1,26 @@
...
+import { BitgoExpressError } from '../../schemas/error';
+import { EncryptRequestBody } from '../v1/encrypt';
+
+/**
+ * Encrypt
+ *
+ * Symmetrically encrypt an arbitrary message with provided password
+ *
+ * @operationId express.encrypt
+ * @tag Express
```
vmccarty:
```suggestion
 * Encrypt message
 *
 * Symmetrically encrypt an arbitrary message with provided password
 *
 * @operationId express.encrypt
 * @tag Express
 * @public
```

vmccarty (changes requested):
Same deal as the other PR. The breaking change (separating v1 and v2 is okay), but for some reason CI didn't flag the missing `@private`/`@public` schema. Also, we should update the summaries to specify that these encrypt messages.

vmccarty (approved):
PRLGTM 👍

pranavjain97 (approved):

PR #8563, opened 2026-04-19
refactor(express): split pingExpress into v1 and v2

rishikeshdadam136:
Ticket: WCI-188


modules/express/src/clientRoutes.ts
```diff
@@ -83,7 +83,7 @@ function handlePing(
   return req.bitgo.ping();
 }
 
-function handlePingExpress(req: ExpressApiRouteRequest<'express.pingExpress', 'get'>) {
+function handlePingExpress(req: ExpressApiRouteRequest<'express.pingexpress', 'get'>) {
```
mrdanish26: `handlePingExpress` is wired to both `express.v1.pingexpress and express.pingexpress`, but the parameter type only mentions express.pingexpress. Update to `ExpressApiRouteRequest<'express.v1.pingexpress' | 'express.pingexpress', 'get'>`  so the type matches both registrations.

modules/express/src/clientRoutes.ts
mrdanish26: Update comments so it matches the split

modules/express/src/typedRoutes/api/v1/pingExpress.ts
mrdanish26, line 1: Add a minimal test that both GET `/api/v1/pingexpress` and GET `/api/v2/pingexpress` resolve and return the expected payload

mrdanish26 (changes requested):

modules/express/src/clientRoutes.ts
rishikeshdadam136: Added it

lokesh-bitgo (dismissed):
Looks good

modules/express/src/typedRoutes/api/v2/pingExpress.ts
```diff
@@ -3,13 +3,16 @@ import { httpRoute, httpRequest } from '@api-ts/io-ts-http';
...
+ * Ping bitgo express to ensure that it is still running. Unlike /ping, this does not try connecting to bitgo.com.
+ *
+ * @operationId express.pingexpress
+ * @tag Express
+ * @public
  */
-export const GetPingExpress = httpRoute({
-  path: '/api/v[12]/pingexpress',
+export const GetV2PingExpress = httpRoute({
+  path: '/api/v2/pingexpress',
```
vmccarty, line 15:
```suggestion
 * Ping BitGo Express
*
 * Ping BitGo Express to ensure that it's still running. Unlike [Ping](/reference/expressping), this endpoint doesn't connect to bitgo.com.
 *
 * @operationId express.pingexpress
 * @tag Express
 * @public
*/
export const GetV2PingExpress = httpRoute({
  path: '/api/v2/pingexpress',
```
I'm still not sure how does this endpoint differs from `/api/v2/ping` that we currently [instruct](https://developers.bitgo.com/docs/get-started-express-install#3-verify-express-is-running) clients to use. How does it tell if Express is running?

rishikeshdadam136: **/api/v2/ping** checks if both Express and BitGo servers are reachable — if BitGo is down, it fails even though Express is fine.

**/api/v2/pingexpress** only checks if Express itself is running — it never contacts BitGo, so it always returns 200 as long as the Express process is alive.

So when **pingexpress** returns 200 **{ "status": "express server is ok!" }**, that itself is the proof — the fact that the request was received and responded to means Express is alive and handling requests.

mrdanish26 (approved):

modules/express/src/typedRoutes/api/v2/pingExpress.ts
vmccarty: Noted. Thanks for updating the desctiption!

vmccarty (approved):
LGTM

PR #8571, opened 2026-04-20
refactor(express): update operationId walletSigntxV2

rishikeshdadam136:
Ticket: WCI-50


lokesh-bitgo (approved):
Looks good

PR #8579, opened 2026-04-20
fix: changing the contract address for usd1 on tempo

damodarnaik699:
Ticket: SCAAS-8989

### DIFF
--- a/modules/statics/src/allCoinsAndTokens.ts
+++ b/modules/statics/src/allCoinsAndTokens.ts
@@ -3299,8 +3299,8 @@ export const allCoinsAndTokens = [
     'fa9e874b-e1c0-4c40-ab4c-3cfba7b2ca8b',
     'tempo:usd1',
     'USD1',
-    18,
-    '0x111111d2bf19e43c34263401e0cad979ed1cdb61',
+    6,
+    '0x20c000000000000000000000111111111e910f0f',
     UnderlyingAsset['tempo:usd1'],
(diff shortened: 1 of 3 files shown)


nayandas190 (approved):
LGTM

modules/statics/src/allCoinsAndTokens.ts
```diff
@@ -3299,8 +3299,8 @@ export const allCoinsAndTokens = [
     'fa9e874b-e1c0-4c40-ab4c-3cfba7b2ca8b',
     'tempo:usd1',
     'USD1',
-    18,
-    '0x111111d2bf19e43c34263401e0cad979ed1cdb61',
+    6,
+    '0x20c000000000000000000000111111111e910f0f',
```
Phani024, line 3303: Where can I find the details about why the contract and the decimals are changing?

sachushaji (approved):

modules/statics/src/allCoinsAndTokens.ts
damodarnaik699: so initially I have deployed as the usd1 contract like ( governorContract then TokenContract how we used to do for the EVM chains ) but later understood that for tempo token we need to deploy the contract like tip20 not like erc20 so here is how we deploy the contracts: https://explore.testnet.tempo.xyz/address/0x20Fc000000000000000000000000000000000000?tab=interact

so when you deploy the contract with the above **createToken** then by default the decimal is 6 and you get the contract address in the tip20 formate rather then erc20 formate

PR #8586, opened 2026-04-21
refactor(express): split login endpoint into V1 and V2

rishikeshdadam136:
Ticket: WCI-186

### DIFF
--- a/modules/express/src/typedRoutes/api/v2/login.ts
+++ b/modules/express/src/typedRoutes/api/v2/login.ts
@@ -1,0 +1,37 @@
+import * as t from 'io-ts';
+import { httpRoute, httpRequest, optional } from '@api-ts/io-ts-http';
+import { BitgoExpressError } from '../../schemas/error';
+import { LoginRequest } from '../v1/login';
+
+/**
+ * Login (Express)
+ *
+ * Authenticate a user and retrieve their session details using BitGo Express.
+ *
+ * @operationId express.login
+ * @tag Express
+ * @public
+ */
+export const PostV2Login = httpRoute({
+  path: '/api/v2/user/login',
+  method: 'POST',
+  request: httpRequest({
+    body: LoginRequest,
+  }),
+  response: {
+    200: t.type({
+      email: t.string,
+      password: t.string,
+      forceSMS: t.boolean,
+      otp: optional(t.string),
+      trust: optional(t.number),
+      extensible: optional(t.boolean),
+      extensionAddress: optional(t.string),
+      forceV1Auth: optional(t.boolean),
+      forReset2FA: optional(t.boolean),
+      initialHash: optional(t.string),
+      fingerprintHash: optional(t.string),
+    }),
+    404: BitgoExpressError,
+  },
+});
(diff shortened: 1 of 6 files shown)


rishikeshdadam136:
Tested in local for both v1 and v2

mrdanish26 (commented):
CI failure, flushing the queue

modules/express/src/typedRoutes/api/v2/login.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @public
+ */
+export const PostV2Login = httpRoute({
+  path: '/api/v2/user/login',
+  method: 'POST',
+  request: httpRequest({
+    body: LoginRequest,
+  }),
+  response: {
+    200: t.type({
```
lokesh-bitgo, line 22: Nit: We can create common type for this

lokesh-bitgo (dismissed):
Overall changes looks good

vmccarty (dismissed):
LGTM

rishikeshdadam136:
Resolved merge conflicts and force pushed again.

mrdanish26 (dismissed):

modules/express/src/typedRoutes/api/v2/login.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Login
+ *
+ * Authenticate a user and retrieve their session details.
+ *
+ * @operationId express.login
+ * @tag Express
+ * @public
+ */
+export const PostV2Login = httpRoute({
+  path: '/api/v2/user/login',
```
melizafranco:
```suggestion
 * Login (Express)
 *
 * Authenticate a user and retrieve their session details using BitGo Express.
 *
 * @operationId express.login
 * @tag Express
 * @public
 */
export const PostV2Login = httpRoute({
  path: '/api/v2/user/login/express',
```
Make this endpoint path and summary unique from the REST [login endpoint](https://developers.bitgo.com/reference/v2userlogin) to avoid client confusion.

melizafranco (dismissed):
Duplicate or ambiguous entries in an OpenAPI spec create confusion in generated docs, client SDKs, and tools like Postman. Each operation should have a unique operationId, a distinct path, and a summary that makes clear what it is and how it differs from the similar one.

lokesh-bitgo:
> Duplicate or ambiguous entries in an OpenAPI spec create confusion in generated docs, client SDKs, and tools like Postman. Each operation should have a unique operationId, a distinct path, and a summary that makes clear what it is and how it differs from the similar one.

@melizafranco, I understand your point. However, the scope of this change is different. This is an existing login endpoint that previously combined both V1 and V2, which caused the API docs to fail to compile. You can find the details here: https://bitgo.slack.com/archives/C057BHBRG4B/p1764020095794829?thread_ts=1764018507.602879&cid=C057BHBRG4B.

Based on the feedback, we are splitting the endpoint into separate V1 and V2 endpoints with unique operationIds and paths, without changing the behavior. We will mark the V1 endpoint as private and the V2 endpoint as public, and only the V2 endpoint will be used by clients.

Regarding your feedback, the operationId is already unique and the paths are already distinct (V1 and V2). Since the behavior of both endpoints remains the same, the summary is also the same. As we are only splitting the endpoint without changing any existing behavior, I think the changes look fine. Please let me know if you have any concern here.

vmccarty (approved):
LGTM

mrdanish26 (approved):

melizafranco (approved):
LGTM

PR #8589, opened 2026-04-21
feat(sdk-coin-eth): add registerWithCoinMap for dynamic token registration

manas-at-bitgo:
## Summary
- Add `registerWithCoinMap` in `sdk-coin-eth` as a drop-in replacement for `register()` that supports dynamic ERC20 token registration from a consumer-provided `CoinMap`
- Calls `register(sdk)` internally for hardcoded base coins and tokens, then registers dynamic ERC20 tokens and adds them to the global coin map via `GlobalCoinFactory.registerToken()`
- No public API changes to `BitGoBase` or `BitGoAPI`

## Motivation
Enables dynamic token onboarding: consumers build a `CoinMap` from AMS (which drifts ahead of what's baked into `@bitgo/statics`) and pass it to `registerWithCoinMap`. New tokens become available via `bitgo.coin()` without an SDK version bump.

TICKET: CSHLD-24

## Test plan
- [x] `register()` registers base ETH coins + ERC20/ERC721 constructors
- [x] `registerWithCoinMap()` delegates to `register()` then adds dynamic tokens to the global coin map
- [x] `registerWithCoinMap()` registers constructors by both type name and contract address
- [x] Empty ERC20 coin map results in no `registerToken` calls
- [x] Unit tests pass: `npx mocha --require tsx modules/sdk-coin-eth/test/unit/register.ts`

### DIFF
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -22,6 +22,7 @@ import {
   makeRandomKey,
   sanitizeLegacyPath,
 } from '@bitgo/sdk-core';
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
 import * as sdkHmac from '@bitgo/sdk-hmac';
 import { DefaultHmacAuthStrategy, type IHmacAuthStrategy } from '@bitgo/sdk-hmac';
 import * as utxolib from '@bitgo/utxo-lib';
@@ -1580,6 +1581,11 @@ export class BitGoAPI implements BitGoBase {
    */
   public register(name: string, coin: CoinConstructor): void {
     GlobalCoinFactory.register(name, coin);
+  }
+
+  public registerWithBaseCoin(name: string, coin: CoinConstructor, baseCoin: Readonly<StaticsBaseCoin>): void {
+    GlobalCoinFactory.register(name, coin);
+    GlobalCoinFactory.registerToken(baseCoin, coin);
   }
 
   /**
--- a/modules/sdk-coin-eth/src/register.ts
+++ b/modules/sdk-coin-eth/src/register.ts
@@ -1,4 +1,4 @@
-import { BitGoBase } from '@bitgo/sdk-core';
+import { BitGoBase, GlobalCoinFactory } from '@bitgo/sdk-core';
 import { Erc20Token } from './erc20Token';
 import { Eth } from './eth';
 import { Gteth } from './gteth';
@@ -21,7 +21,19 @@ export const register = (sdk: BitGoBase): void => {
 };
 
 export const registerWithCoinMap = (sdk: BitGoBase, coinMap: CoinMap): void => {
+  sdk.register('eth', Eth.createInstance);
+  sdk.register('gteth', Gteth.createInstance);
+  sdk.register('teth', Teth.createInstance);
+  sdk.register('hteth', Hteth.createInstance);
+  Erc721Token.createTokenConstructors().forEach(({ name, coinConstructor }) => {
+    sdk.register(name, coinConstructor);
+  });
+
+  // Registration for ERC20 tokens from the coin map (includes both hardcoded and dynamic tokens from AMS).
   Erc20Token.createTokenConstructors(getFormattedErc20Tokens(coinMap)).forEach(({ name, coinConstructor }) => {
     sdk.register(name, coinConstructor);
+    if (coinMap.has(name)) {
+      GlobalCoinFactory.registerToken(coinMap.get(name), coinConstructor);
+    }
   });
 };
--- a/modules/sdk-coin-eth/src/register.ts
+++ b/modules/sdk-coin-eth/src/register.ts
@@ -21,7 +21,16 @@ export const register = (sdk: BitGoBase): void => {
 };
 
 export const registerWithCoinMap = (sdk: BitGoBase, coinMap: CoinMap): void => {
-  Erc20Token.createTokenConstructors(getFormattedErc20Tokens(coinMap)).forEach(({ name, coinConstructor }) => {
+  sdk.register('eth', Eth.createInstance);
+  sdk.register('gteth', Gteth.createInstance);
+  sdk.register('teth', Teth.createInstance);
+  sdk.register('hteth', Hteth.createInstance);
+  Erc721Token.createTokenConstructors().forEach(({ name, coinConstructor }) => {
     sdk.register(name, coinConstructor);
+  });
+
+  // Registration for dynamic ERC20 tokens that are not hardcoded in the SDK, but are present in the coin map generated using AMS.
+  Erc20Token.createTokenConstructors(getFormattedErc20Tokens(coinMap)).forEach(({ name, coinConstructor }) => {
+    sdk.registerWithBaseCoin(name, coinConstructor, coinMap.get(name));
   });
 };
--- a/modules/sdk-core/src/bitgo/bitgoBase.ts
+++ b/modules/sdk-core/src/bitgo/bitgoBase.ts
@@ -6,6 +6,7 @@ import {
   GetSharingKeyOptions,
   IRequestTracer,
 } from '../api';
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
 import { IBaseCoin } from './baseCoin';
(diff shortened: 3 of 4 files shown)


modules/sdk-coin-eth/src/register.ts
```diff
@@ -21,7 +21,16 @@ export const register = (sdk: BitGoBase): void => {
...
+  sdk.register('hteth', Hteth.createInstance);
+  Erc721Token.createTokenConstructors().forEach(({ name, coinConstructor }) => {
     sdk.register(name, coinConstructor);
   });
+
+  // Registration for dynamic ERC20 tokens that are not hardcoded in the SDK, but are present in the coin map generated using AMS.
+  Erc20Token.createTokenConstructors(getFormattedErc20Tokens(coinMap)).forEach(({ name, coinConstructor }) => {
+    sdk.registerWithBaseCoin(name, coinConstructor, coinMap.get(name));
+  });
 };
```
pranavjain97, line 39: `CoinMap.get()` throws `CoinNotDefinedError` if the key is missing. if `createTokenConstructors` produces a name not in the coinMap, this will break the whole registration loop. consider a `has()` check first so one bad token doesn't take down all registrations

modules/sdk-coin-eth/src/register.ts
```diff
@@ -21,7 +21,16 @@ export const register = (sdk: BitGoBase): void => {
...
 
 export const registerWithCoinMap = (sdk: BitGoBase, coinMap: CoinMap): void => {
-  Erc20Token.createTokenConstructors(getFormattedErc20Tokens(coinMap)).forEach(({ name, coinConstructor }) => {
+  sdk.register('eth', Eth.createInstance);
```
pranavjain97, line 24: `registerWithCoinMap` now also registers base coins and ERC721 tokens, which it didn't before. any existing caller that expected base coins to already be registered separately could hit duplicate registration. is the behavior change intentional?

pranavjain97 (changes requested):
the approach works but creates duplication that will drift. registerWithCoinMap copies the 4 base coin + ERC721 registrations from register(), and registerWithBaseCoin on BitGoBase is a public interface change for what's really an internal concern. simpler: have registerWithCoinMap call register(sdk) internally, then just add the dynamic token part with GlobalCoinFactory.registerToken() directly. no interface change needed.

also, no tests added for registerWithBaseCoin or the updated registerWithCoinMap.

modules/sdk-coin-eth/src/register.ts
manas-at-bitgo: Good catch. createTokenConstructors returns two entries per token — one by type name (eth:usdt) and one by contract address (0x0c35...). coinMap.get() would throw on contract addresses. Separated into two loops now: first registers constructors for all entries (type name + contract address) via sdk.register(), second iterates over the formatted token configs directly and calls registerWithBaseCoin using token.type which is always a valid coinMap key.

modules/sdk-coin-eth/src/register.ts
manas-at-bitgo: registerWithCoinMap is a drop-in replacement for register, not meant to be used alongside it. So no double-registration risk. That said, refactored it to call register(sdk) internally so there's no duplication to drift.

manas-at-bitgo:
@pranavjain97 
- registerWithCoinMap now calls register(sdk) internally — no more duplication
- Simplified registerWithBaseCoin — dropped the redundant name param since registerToken already registers by baseCoin.name. It's still a public method on BitGoAPI since consumers need it to add tokens to the global coin map at runtime, which is the whole point of dynamic onboarding — can't do that with GlobalCoinFactory directly from outside sdk-api
- Added unit tests covering register(), registerWithCoinMap(), and the empty coin map case

veetragjain (dismissed):

modules/sdk-coin-eth/src/register.ts
```diff
@@ -21,7 +21,16 @@ export const register = (sdk: BitGoBase): void => {
...
 export const registerWithCoinMap = (sdk: BitGoBase, coinMap: CoinMap): void => {
-  Erc20Token.createTokenConstructors(getFormattedErc20Tokens(coinMap)).forEach(({ name, coinConstructor }) => {
+  register(sdk);
+
+  // Registration for dynamic ERC20 tokens that are not hardcoded in the SDK, but are present in the coin map generated using AMS.
+  const formattedTokens = getFormattedErc20Tokens(coinMap);
+  Erc20Token.createTokenConstructors(formattedTokens).forEach(({ name, coinConstructor }) => {
+    // Register constructor for both type names and contract addresses
     sdk.register(name, coinConstructor);
   });
```
veetragjain: Will this not throw an error if coin is not present in the coin map maintained by the coin factory?

modules/sdk-core/src/bitgo/bitgoBase.ts
```diff
@@ -35,4 +36,5 @@ export interface BitGoBase {
   setRequestTracer(reqTracer: IRequestTracer): void;
   url(path: string, version?: number): string;
   register(name: string, coin: CoinConstructor): void;
+  registerWithBaseCoin(coin: CoinConstructor, baseCoin: Readonly<StaticsBaseCoin>): void;
```
veetragjain: Base Coin can get confusing, there are two types of base coin, one used for coin map and one used for coin factory, we can rename it to `registerWithStaticsCoin` or something similar

modules/sdk-coin-eth/src/register.ts
pranavjain97, line 30: registering a token that already exists in `@bitgo/statics` will throw `DuplicateCoinDefinitionError` because `registerToken` calls `coinMap.addCoin` which rejects duplicates. need to either pre-filter tokens not already in the factory, or make `registerToken` skip existing entries

modules/sdk-coin-eth/src/register.ts
```diff
@@ -21,7 +21,16 @@ export const register = (sdk: BitGoBase): void => {
 };
 
 export const registerWithCoinMap = (sdk: BitGoBase, coinMap: CoinMap): void => {
-  Erc20Token.createTokenConstructors(getFormattedErc20Tokens(coinMap)).forEach(({ name, coinConstructor }) => {
+  register(sdk);
+
+  // Registration for dynamic ERC20 tokens that are not hardcoded in the SDK, but are present in the coin map generated using AMS.
+  const formattedTokens = getFormattedErc20Tokens(coinMap);
+  Erc20Token.createTokenConstructors(formattedTokens).forEach(({ name, coinConstructor }) => {
+    // Register constructor for both type names and contract addresses
```
pranavjain97: `createTokenConstructor` is called twice per token (once in `createTokenConstructors`, again here), creating duplicate closures. reuse from the first pass

modules/sdk-core/src/bitgo/bitgoBase.ts
pranavjain97: +1 to veetragjain: `registerWithBaseCoin` is ambiguous, `registerToken` would be clearer

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -1582,6 +1583,10 @@ export class BitGoAPI implements BitGoBase {
     GlobalCoinFactory.register(name, coin);
   }
 
+  public registerWithBaseCoin(coin: CoinConstructor, baseCoin: Readonly<StaticsBaseCoin>): void {
```
pranavjain97: if the only caller is `registerWithCoinMap`, does this need to be on `BitGoBase`? could call `GlobalCoinFactory.registerToken` directly

pranavjain97 (changes requested):
duplicate token registration will throw, and constructors are created twice per token

modules/sdk-coin-eth/src/register.ts
manas-at-bitgo: No, `sdk.register` just adds to the factory. It doesn't require the coin to already exist.

manas-at-bitgo: looking at `registerToken` (line 62-69), it already has a `has()` check and skips `addCoin` if the coin already exists. So duplicates are handled.

veetragjain (approved):

pranavjain97 (approved):

PR #8592, opened 2026-04-21
feat: opt-in v2 Argon2+HKDF encryption for multisig and MPC flows

pranavjain97:
## Summary

Wire v2 encryption (Argon2id + AES-256-GCM + HKDF session caching) into wallet creation and signing call sites across multisig, DKLS MPCv2, and EdDSA flows.

- Opt-in via `encryptionVersion: 2` on wallet/key creation
- Signing auto-detects v1/v2 from the envelope
- Default stays v1 -- zero breaking changes
- `decryptKeychainPrivateKey` made async to support v1/v2 auto-detection in signing paths

## Live Node.JS Testing (testnet)

All flows tested end-to-end on testnet with real transactions:

| Operation | v1 (SJCL/PBKDF2) | v2 (Argon2id+HKDF) | Delta |
|-----------|-------------------|---------------------|-------|
| EdDSA wallet creation (tsol) | 6.02s | 5.53s | -8% |
| ECDSA MPCv2 wallet creation (hteth) | 8.57s | 7.92s | -8% |
| Multisig wallet creation (tbtc) | 2.53s | 2.43s | -4% |
| EdDSA signing (tsol) | 11.64s | 11.06s | -5% |
| ECDSA MPCv2 signing (hteth) | 10.23s | 11.02s | +8% |

v2 wallet creation is consistently faster due to HKDF session caching. Signing is roughly equivalent -- network round trips dominate.

### Works (this PR):
  - Wallet creation (multisig, DKLS, EdDSA)
  - DKLS offline signing (all 3 rounds with adata + domain separators)
  - EdDSA signing (commitment + R-share)
  - getUserPrvAsync / decryptKeychainPrivateKeyAsync at all call sites
  - EVM hop transactions, inscription builder, verifyKey

###  Known gaps (tracked, not this PR):
  - Wallet sharing
  - Password change
  - Recovery
  - GG18 ECDSA signing

## Test plan

- [x] sdk-api unit tests (162 passing)
- [x] sdk-core unit tests (163 passing)
- [x] Codec validation tests for Lightning/GoAccount with encryptionVersion
- [x] E2e tests for DKLS and EdDSA keygen with v2 envelopes
- [x] Live testnet: EdDSA TSS wallet creation + signing (tsol)
- [x] Live testnet: ECDSA MPCv2 wallet creation + signing (hteth)
- [x] Live testnet: Multisig wallet creation + decrypt verification (tbtc)
- [ ] CI

TICKET: WCN-32

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -2554,7 +2554,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
     const walletPassphrase = buildParams.walletPassphrase;
 
     const userKeychain = await this.keychains().get({ id: wallet.keyIds()[0] });
-    const userPrv = wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const userPrv = await wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
     const userPrvBuffer = bip32.fromBase58(userPrv).privateKey;
     if (!userPrvBuffer) {
       throw new Error('invalid userPrv');
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -2554,7 +2554,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
     const walletPassphrase = buildParams.walletPassphrase;
 
     const userKeychain = await this.keychains().get({ id: wallet.keyIds()[0] });
-    const userPrv = await wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const userPrv = await wallet.getUserPrvAsync({ keychain: userKeychain, walletPassphrase });
     const userPrvBuffer = bip32.fromBase58(userPrv).privateKey;
     if (!userPrvBuffer) {
       throw new Error('invalid userPrv');
--- a/modules/abstract-utxo/src/impl/btc/inscriptionBuilder.ts
+++ b/modules/abstract-utxo/src/impl/btc/inscriptionBuilder.ts
@@ -302,7 +302,7 @@ export class InscriptionBuilder implements IInscriptionBuilder {
     txPrebuild: PrebuildTransactionResult
   ): Promise<SubmitTransactionResponse> {
     const userKeychain = await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[KeyIndices.USER] });
-    const prv = this.wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const prv = await this.wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
 
     const halfSigned = (await this.wallet.signTransaction({ prv, txPrebuild })) as HalfSignedUtxoTransaction;
     return this.wallet.submitTransaction({ halfSigned });
--- a/modules/abstract-utxo/src/impl/btc/inscriptionBuilder.ts
+++ b/modules/abstract-utxo/src/impl/btc/inscriptionBuilder.ts
@@ -262,7 +262,7 @@ export class InscriptionBuilder implements IInscriptionBuilder {
     inscriptionData: Buffer
   ): Promise<SubmitTransactionResponse> {
     const userKeychain = await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[KeyIndices.USER] });
-    const xprv = await this.wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const xprv = await this.wallet.getUserPrvAsync({ keychain: userKeychain, walletPassphrase });
 
     const halfSignedCommitTransaction = (await this.wallet.signTransaction({
       prv: xprv,
@@ -302,7 +302,7 @@ export class InscriptionBuilder implements IInscriptionBuilder {
     txPrebuild: PrebuildTransactionResult
   ): Promise<SubmitTransactionResponse> {
     const userKeychain = await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[KeyIndices.USER] });
-    const prv = await this.wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const prv = await this.wallet.getUserPrvAsync({ keychain: userKeychain, walletPassphrase });
 
     const halfSigned = (await this.wallet.signTransaction({ prv, txPrebuild })) as HalfSignedUtxoTransaction;
     return this.wallet.submitTransaction({ halfSigned });
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsaMPCv2/signTxRequest.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsaMPCv2/signTxRequest.ts
@@ -269,7 +269,6 @@ describe('signTxRequest:', function () {
       const userShare = fs.readFileSync(shareFiles[vector.party1]);
       const userPrvBase64 = Buffer.from(userShare).toString('base64');
 
-      // Encrypt the prv with v2 to trigger the v2 path
       const encryptedPrv = await bitgo.encryptAsync({
         input: userPrvBase64,
         password: walletPassphrase,
@@ -277,7 +276,6 @@ describe('signTxRequest:', function () {
       });
       JSON.parse(encryptedPrv).v.should.equal(2);
 
-      // Round 1: encrypt session + GPG key with v2 + adata (purely local, no server call)
       const round1Result = await tssUtils.createOfflineRound1Share({
         txRequest,
         prv: userPrvBase64,
@@ -285,18 +283,17 @@ describe('signTxRequest:', function () {
         encryptedPrv,
       });
 
-      // Verify round 1 output has v2 envelopes with adata
       const r1SessionEnvelope = JSON.parse(round1Result.encryptedRound1Session);
       r1SessionEnvelope.v.should.equal(2);
       r1SessionEnvelope.should.have.property('adata');
       r1SessionEnvelope.should.have.property('hkdfSalt');
+      r1SessionEnvelope.adata.should.containEql('DKLS23_SIGNING_ROUND1_STATE');
 
       const r1GpgEnvelope = JSON.parse(round1Result.encryptedUserGpgPrvKey);
       r1GpgEnvelope.v.should.equal(2);
       r1GpgEnvelope.should.have.property('adata');
-      r1SessionEnvelope.adata.should.equal(r1GpgEnvelope.adata);
+      r1GpgEnvelope.adata.should.containEql('DKLS23_SIGNING_USER_GPG_KEY');
 
-      // Nock BitGo round 1 response and submit
       await nockTxRequestResponseSignatureShareRoundOne(bitgoParty, txRequest, bitgoGpgKey);
       const transactions = getRoute('ecdsa');
       const round1TxRequestResponse = await bitgo
@@ -307,7 +304,6 @@ describe('signTxRequest:', function () {
         })
         .result();
 
-      // Merge server response with original txRequest (server only returns signatureShares)
       const round1TxReq: TxRequest = {
         ...txRequest,
         transactions: [
@@ -318,7 +314,6 @@ describe('signTxRequest:', function () {
         ],
       };
 
-      // Round 2: decrypt v2 round 1 session (validates adata), encrypt round 2 session
       const round2Result = await tssUtils.createOfflineRound2Share({
         txRequest: round1TxReq,
         prv: userPrvBase64,
@@ -328,13 +323,11 @@ describe('signTxRequest:', function () {
         encryptedRound1Session: round1Result.encryptedRound1Session,
       });
 
-      // Verify round 2 output has v2 envelope with adata
       const r2Envelope = JSON.parse(round2Result.encryptedRound2Session);
       r2Envelope.v.should.equal(2);
       r2Envelope.should.have.property('adata');
-      r2Envelope.adata.should.equal(r1SessionEnvelope.adata);
+      r2Envelope.adata.should.containEql('DKLS23_SIGNING_ROUND2_STATE');
 
-      // Nock BitGo round 2 response and submit
       await nockTxRequestResponseSignatureShareRoundTwo(bitgoParty, txRequest, bitgoGpgKey);
       const round2TxRequestResponse = await bitgo
         .post(bitgo.url(`/wallet/${txRequest.walletId}/txrequests/${txRequest.txRequestId + transactions}/sign`, 2))
@@ -354,7 +347,6 @@ describe('signTxRequest:', function () {
         ],
       };
 
-      // Round 3: decrypt v2 round 2 session (validates adata), produce final signature share
       const round3Result = await tssUtils.createOfflineRound3Share({
         txRequest: round2TxReq,
         prv: userPrvBase64,
@@ -367,15 +359,29 @@ describe('signTxRequest:', function () {
       round3Result.should.have.property('signatureShareRound3');
     });
 
-    it('validateAdata rejects v2 envelopes with mismatched adata', async function () {
+    it('validateAdata accepts v2 envelopes with matching adata and domain separator', async function () {
+      const adata = 'txhash:m/0/1';
+      const domainSep = 'DKLS23_SIGNING_ROUND1_STATE';
       const ct = await bitgo.encryptAsync({
         input: 'test-data',
         password: 'testpass',
         encryptionVersion: 2,
-        adata: 'context-A',
+        adata: `${domainSep}:${adata}`,
       });
 
-      (() => (tssUtils as any).validateAdata('context-B', ct)).should.throw(/Adata does not match/);
+      (tssUtils as any).validateAdata(adata, ct, domainSep);
+    });
+
+    it('validateAdata rejects v2 envelopes with mismatched adata', async function () {
+      const domainSep = 'DKLS23_SIGNING_ROUND1_STATE';
+      const ct = await bitgo.encryptAsync({
+        input: 'test-data',
+        password: 'testpass',
+        encryptionVersion: 2,
+        adata: `${domainSep}:context-A`,
+      });
+
+      (() => (tssUtils as any).validateAdata('context-B', ct, domainSep)).should.throw(/Adata does not match/);
     });
   });
 
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsaMPCv2/signTxRequest.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/ecdsaMPCv2/signTxRequest.ts
@@ -263,6 +263,122 @@ describe('signTxRequest:', function () {
     nockPromises[2].isDone().should.be.true();
   });
 
+  describe('v2 encryption (offline rounds with adata)', function () {
+    it('e2e: 3-round offline signing with v2 encrypted keys preserves adata context binding', async function () {
+      const walletPassphrase = 'testpassphrase';
+      const userShare = fs.readFileSync(shareFiles[vector.party1]);
+      const userPrvBase64 = Buffer.from(userShare).toString('base64');
+
+      // Encrypt the prv with v2 to trigger the v2 path
+      const encryptedPrv = await bitgo.encryptAsync({
+        input: userPrvBase64,
+        password: walletPassphrase,
+        encryptionVersion: 2,
+      });
+      JSON.parse(encryptedPrv).v.should.equal(2);
+
+      // Round 1: encrypt session + GPG key with v2 + adata (purely local, no server call)
+      const round1Result = await tssUtils.createOfflineRound1Share({
+        txRequest,
+        prv: userPrvBase64,
+        walletPassphrase,
+        encryptedPrv,
+      });
+
+      // Verify round 1 output has v2 envelopes with adata
+      const r1SessionEnvelope = JSON.parse(round1Result.encryptedRound1Session);
+      r1SessionEnvelope.v.should.equal(2);
+      r1SessionEnvelope.should.have.property('adata');
+      r1SessionEnvelope.should.have.property('hkdfSalt');
+
+      const r1GpgEnvelope = JSON.parse(round1Result.encryptedUserGpgPrvKey);
+      r1GpgEnvelope.v.should.equal(2);
+      r1GpgEnvelope.should.have.property('adata');
+      r1SessionEnvelope.adata.should.equal(r1GpgEnvelope.adata);
+
+      // Nock BitGo round 1 response and submit
+      await nockTxRequestResponseSignatureShareRoundOne(bitgoParty, txRequest, bitgoGpgKey);
+      const transactions = getRoute('ecdsa');
+      const round1TxRequestResponse = await bitgo
+        .post(bitgo.url(`/wallet/${txRequest.walletId}/txrequests/${txRequest.txRequestId + transactions}/sign`, 2))
+        .send({
+          signatureShares: [round1Result.signatureShareRound1],
+          signerGpgPublicKey: round1Result.userGpgPubKey,
+        })
+        .result();
+
+      // Merge server response with original txRequest (server only returns signatureShares)
+      const round1TxReq: TxRequest = {
+        ...txRequest,
+        transactions: [
+          {
+            ...txRequest.transactions![0],
+            signatureShares: round1TxRequestResponse.transactions[0].signatureShares,
+          },
+        ],
+      };
+
+      // Round 2: decrypt v2 round 1 session (validates adata), encrypt round 2 session
+      const round2Result = await tssUtils.createOfflineRound2Share({
+        txRequest: round1TxReq,
+        prv: userPrvBase64,
+        walletPassphrase,
+        bitgoPublicGpgKey: bitgoGpgKey.publicKey,
+        encryptedUserGpgPrvKey: round1Result.encryptedUserGpgPrvKey,
+        encryptedRound1Session: round1Result.encryptedRound1Session,
+      });
+
+      // Verify round 2 output has v2 envelope with adata
+      const r2Envelope = JSON.parse(round2Result.encryptedRound2Session);
+      r2Envelope.v.should.equal(2);
+      r2Envelope.should.have.property('adata');
+      r2Envelope.adata.should.equal(r1SessionEnvelope.adata);
+
+      // Nock BitGo round 2 response and submit
+      await nockTxRequestResponseSignatureShareRoundTwo(bitgoParty, txRequest, bitgoGpgKey);
+      const round2TxRequestResponse = await bitgo
+        .post(bitgo.url(`/wallet/${txRequest.walletId}/txrequests/${txRequest.txRequestId + transactions}/sign`, 2))
+        .send({
+          signatureShares: [round2Result.signatureShareRound2],
+          signerGpgPublicKey: round1Result.userGpgPubKey,
+        })
+        .result();
+
+      const round2TxReq: TxRequest = {
+        ...txRequest,
+        transactions: [
+          {
+            ...txRequest.transactions![0],
+            signatureShares: round2TxRequestResponse.transactions[0].signatureShares,
+          },
+        ],
+      };
+
+      // Round 3: decrypt v2 round 2 session (validates adata), produce final signature share
+      const round3Result = await tssUtils.createOfflineRound3Share({
+        txRequest: round2TxReq,
+        prv: userPrvBase64,
+        walletPassphrase,
+        bitgoPublicGpgKey: bitgoGpgKey.publicKey,
+        encryptedUserGpgPrvKey: round1Result.encryptedUserGpgPrvKey,
+        encryptedRound2Session: round2Result.encryptedRound2Session,
+      });
+
+      round3Result.should.have.property('signatureShareRound3');
+    });
+
+    it('validateAdata rejects v2 envelopes with mismatched adata', async function () {
+      const ct = await bitgo.encryptAsync({
+        input: 'test-data',
+        password: 'testpass',
+        encryptionVersion: 2,
+        adata: 'context-A',
+      });
+
+      (() => (tssUtils as any).validateAdata('context-B', ct)).should.throw(/Adata does not match/);
+    });
+  });
+
   it('fails to signs a txRequest for a dkls hot wallet after receiving over 3 429 errors', async function () {
     const nockPromises = [
       await nockTxRequestResponseSignatureShareRoundOne(bitgoParty, txRequest, bitgoGpgKey),
--- a/modules/sdk-api/src/bitgoAPI.ts
+++ b/modules/sdk-api/src/bitgoAPI.ts
@@ -41,6 +41,8 @@ import {
   verifyResponseAsync,
 } from './api';
 import { decrypt, decryptAsync, encrypt } from './encrypt';
+import { createEncryptionSession } from './encryptionSession';
+import { encryptV2 } from './encryptV2';
 import { verifyAddress } from './v1/verifyAddress';
 import {
   AccessTokenOptions,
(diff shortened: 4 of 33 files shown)


pranavjain97:
Wallet sharing to be done as a separate ticket.

modules/sdk-core/src/api/types.ts
```diff
@@ -21,6 +21,7 @@ export interface EncryptOptions {
   input: string;
   password?: string;
   adata?: string;
+  encryptionVersion?: 2;
```
zahin-mohammad: **Major — use a named alias.** A bare literal `2` makes `encryptionVersion: 1` a type error, even though v1 is the current default — so callers forwarding from a generic config (`{ encryptionVersion: cfg.version }`) need a cast. Suggest:

```ts
export type EncryptionVersion = 1 | 2;

export interface EncryptOptions {
  input: string;
  password?: string;
  adata?: string;
  encryptionVersion?: EncryptionVersion;
}
```

This prepares for v3 without a hunt-and-replace across the 8+ option types where the field appears (`GenerateWalletOptions`, `GenerateMpcWalletOptions`, `CreateMpcOptions`, `CreateBackupOptions`, `CreateKeychainParamsBase`, etc.). Also: `adata` should get a JSDoc note that it is silently dropped when `encryptionVersion === 2` — see the inline comment on `BitGoAPI.encryptAsync`.

modules/sdk-core/src/bitgo/bitgoBase.ts
```diff
@@ -19,6 +19,12 @@ export interface BitGoBase {
   decryptKeys(params: DecryptKeysOptions): string[];
   del(url: string): BitGoRequest;
   encrypt(params: EncryptOptions): string;
+  encryptAsync(params: EncryptOptions): Promise<string>;
+  createEncryptionSession(password: string): Promise<{
+    encrypt(plaintext: string): Promise<string>;
+    decrypt(ciphertext: string): Promise<string>;
+    destroy(): void;
```
zahin-mohammad: **Major — extract a named `IEncryptionSession`.** The anonymous inline return type is repeated 4+ times across the PR (`ecdsaMPCv2.ts` `addUserKeychain`/`addBackupKeychain` parameter types, `eddsa.ts` session usage, etc.). `EncryptionSession` already exists as a class in `modules/sdk-api/src/encryptionSession.ts`. The structural problem is that `sdk-core` can't import from `sdk-api` (circular). Fix: declare an `IEncryptionSession` interface in `modules/sdk-core/src/api/types.ts`, export it from `sdk-core/src/index.ts`, and have both this interface and the `EncryptionSession` class in `sdk-api` reference it.

Without this, no external SDK consumer can write `let session: ???` with a proper type import.

modules/sdk-api/src/bitgoAPI.ts
```diff
@@ -715,6 +717,29 @@ export class BitGoAPI implements BitGoBase {
...
+   * Async encrypt that dispatches to v1 (SJCL) or v2 (Argon2id + AES-256-GCM)
+   * based on `encryptionVersion`.
+   */
+  async encryptAsync(params: EncryptOptions): Promise<string> {
+    common.validateParams(params, ['input', 'password'], []);
+    if (!params.password) {
+      throw new Error('cannot encrypt without password');
+    }
+    if (params.encryptionVersion === 2) {
+      return encryptV2(params.password, params.input);
```
zahin-mohammad: **Major — `params.adata` is silently dropped on the v2 path.** v1 forwards `adata` to `sjcl.encrypt`; v2 calls `encryptV2(password, input)` with no AAD. Two consequences:

1. Callers passing `adata` get false confidence that AAD is bound — it is not.
2. In `ecdsaMPCv2.ts`, the v1 path uses `adata = ${hashBuffer}:${derivationPath}` to bind the encrypted round-session to the transaction context (`validateAdata` enforces the binding). The v2 path has no equivalent (see inline comment on `ecdsaMPCv2.ts` line 1242).

v2's AES-GCM is self-authenticating for the *ciphertext* — but doesn't replicate the *transaction-context binding* unless AAD is threaded through `encryptV2` envelopes. Fix: add an `additionalData?: string` option to `encryptV2`/`aesGcmEncrypt`, store it in the v2 envelope, and verify in `decryptV2`. Or, at minimum, throw if `params.adata` is set and `encryptionVersion === 2` so the silent drop becomes a loud error.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts
```diff
@@ -106,6 +106,7 @@ export class EcdsaUtils extends BaseEcdsaUtils {
     passphrase: string;
     enterprise?: string | undefined;
     originalPasscodeEncryptionCode?: string | undefined;
+    encryptionVersion?: 2;
```
zahin-mohammad: **Major — accepted but never read.** `EcdsaUtils.createKeychains` (non-MPCv2 ECDSA) declares `encryptionVersion?: 2` in its param type, but the function body never references it — `createParticipantKeychain` (line 299) still calls `this.bitgo.encrypt(...)` synchronously. Result: a caller that passes `encryptionVersion: 2` to a non-MPCv2 ECDSA wallet gets v1 silently.

Fix: either thread `encryptionVersion` through `createParticipantKeychain` (mirroring the MPCv2 pattern), or remove the field from the param type if non-MPCv2 ECDSA is not in scope for v2 in this PR.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -184,10 +196,19 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
       originalPasscodeEncryptionCode,
     };
     if (passphrase !== undefined) {
-      userKeychainParams.encryptedPrv = this.bitgo.encrypt({
-        input: JSON.stringify(userSigningMaterial),
-        password: passphrase,
-      });
+      if (encryptionVersion === 2) {
+        const session = await this.bitgo.createEncryptionSession(passphrase);
```
zahin-mohammad: **Major — separate Argon2id session per keychain.** `createUserKeychain` (line 200) and `createBackupKeychain` (line 295) each create their own `createEncryptionSession` for one encryption then immediately `destroy()`. Since both run in parallel via `Promise.all` (line 397/407 of `createKeychains`), Argon2id is computed twice unnecessarily.

`EcdsaMPCv2Utils.createKeychains` does this correctly: a single shared session created in the parent function and threaded into both `addUserKeychain` and `addBackupKeychain`. Please mirror that pattern here — create one session in `createKeychains`, pass it into both children, `destroy()` it in `finally`. Saves one full Argon2id derivation per keygen.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -1173,9 +1238,14 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
       bitgoGpgKey
     );
 
-    const round1Session = this.bitgo.decrypt({ input: encryptedRound1Session, password: walletPassphrase });
+    let round1Session: string;
+    if (useV2) {
+      round1Session = await this.bitgo.decryptAsync({ input: encryptedRound1Session, password: walletPassphrase });
+    } else {
+      round1Session = this.bitgo.decrypt({ input: encryptedRound1Session, password: walletPassphrase });
+      this.validateAdata(adata, encryptedRound1Session);
```
zahin-mohammad: **Major / defense-in-depth — v2 path skips transaction-context binding.** The v1 path (`this.validateAdata(adata, encryptedRound1Session)` on line 1246) binds the encrypted round-session to `${hashBuffer}:${derivationPath}`, preventing replay across different transaction signing contexts. The v2 branch above (line 1242) has no equivalent.

GCM's authentication tag covers the ciphertext, but does not bind the session to a specific transaction — that property was provided by `adata` in v1. AAD support needs to be threaded through `encryptV2` / `EncryptionSession` envelopes (see inline on `BitGoAPI.encryptAsync`), or the v2 round-session needs a different binding mechanism, or the relaxation needs to be documented as accepted security trade-off (with rationale for why round-session replay is not a concern for ephemeral signing keys).

This is also relevant for line 1337 (round-3 path).

Note: Zhongxi's HSM-1513 work (commits `e3bc07cbf6` / `d531ed2802` on a separate branch) adds round domain separators to v1 `adata`. v2 should get the equivalent before this lands.

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -1070,7 +1070,7 @@ export interface IWallet {
   removeUser(params?: RemoveUserOptions): Promise<any>;
   prebuildTransaction(params?: PrebuildTransactionOptions): Promise<PrebuildTransactionResult>;
   signTransaction(params?: WalletSignTransactionOptions): Promise<SignedTransaction>;
-  getUserPrv(params?: GetUserPrvOptions): string;
+  getUserPrv(params?: GetUserPrvOptions): Promise<string>;
```
zahin-mohammad: **Blocker — breaking-change commit marker missing.** This signature change (`string` → `Promise<string>`) is a public-API break for `@bitgo/sdk-core` consumers. `IWallet` is exported from the package barrel; downstream JS / non-strict-TS callers who do `const prv = wallet.getUserPrv(opts); signTransaction({ prv })` will silently pass a `Promise<string>` to the signer.

`lerna publish --conventional-commits` (Angular preset) only triggers a major bump on `feat!:` / `fix!:` / `BREAKING CHANGE:` footer. None of the 7 commits on this PR carry that marker, so `sdk-core` will publish under a minor bump. The repo's `Check breaking changes` CI job is OpenAPI-only and does not catch SDK type-signature changes.

Fix: add a follow-up commit (or rewrite `36c78cd`) with body:

```
BREAKING CHANGE: IWallet.getUserPrv now returns Promise<string> instead of string.
Callers must await the result.
```

Same applies to the `decryptKeychainPrivateKey` export — see inline on `decryptKeychain.ts`.

modules/sdk-core/src/bitgo/keychain/decryptKeychain.ts
```diff
@@ -17,19 +17,20 @@ function maybeDecrypt(bitgo: BitGoBase, input: string, password: string): string
...
  *
  * @param bitgo
  * @param keychain
  * @param password
  */
-export function decryptKeychainPrivateKey(
+export async function decryptKeychainPrivateKey(
   bitgo: BitGoBase,
   keychain: OptionalKeychainEncryptedKey,
   password: string
```
zahin-mohammad, line 64: **Blocker — same as `IWallet.getUserPrv`.** This is a named public export of `@bitgo/sdk-core`. The async signature change requires a `BREAKING CHANGE:` footer so lerna bumps the package to a major version on publish. Without it the published package will be a minor bump and downstream callers in untyped JS / non-strict TS will silently get a `Promise` where they expect a `string | undefined`.

Fix: roll into the same `BREAKING CHANGE:` footer as `getUserPrv` (or split into a separate footer). Suggest:

```
BREAKING CHANGE:
- IWallet.getUserPrv now returns Promise<string> (was string)
- decryptKeychainPrivateKey now returns Promise<string | undefined> (was string | undefined)
All callers must await the result.
```

Also note: `modules/bitgo/test/unit/decryptKeychain.ts` (5 sync `decryptKeychainPrivateKey` calls at lines 18, 36, 55, 74, 78) was not updated — see the summary for details. Please add `await` and make the test functions async so the file actually validates the new behavior.

zahin-mohammad (changes requested):
## Review Summary

After multiple validation rounds with parallel reviewers (architect, crypto, call-site tracer, test-gap, devil's-advocate, consensus-checker), **REQUEST CHANGES** with two blockers and a focused set of inline comments.

---

### Blockers

**B1. No breaking-change marker on commits.** None of the 7 commits on this PR carry `feat!:`, `fix!:`, or a `BREAKING CHANGE:` footer. `lerna publish --conventional-commits` (Angular preset) requires one of those for a major bump. Result: `@bitgo/sdk-core` will publish a *minor* bump despite `IWallet.getUserPrv` and `decryptKeychainPrivateKey` changing from sync `string` returns to `Promise<string>`. Downstream JS / non-strict-TS consumers will silently get a Promise where they expect a string. The repo's `Check breaking changes` CI job only diffs OpenAPI specs from `modules/express` — it does **not** guard SDK type signatures, so this is a code-review responsibility entirely. **Fix:** add a `BREAKING CHANGE:` footer to commit `36c78cd` (or a follow-up commit) describing the async return-type change so lerna bumps `sdk-core` to a major version on publish.

**B2. Sync v1 `decrypt` fan-out callers not converted, no follow-up tickets linked.** A v2-encrypted wallet that flows into any of these sites hard-throws with a confusing SJCL tag-mismatch error. Blast radius is bounded today (v2 is opt-in), but every client that opts into v2 will trip at least one of these in normal usage:

- `modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts:647` — `assertIsValidKey` calls `sjcl.decrypt` directly
- `modules/abstract-utxo/src/recovery/crossChainRecovery.ts:316` — sync `decrypt(passphrase, encryptedPrv)`
- `modules/sdk-core/src/bitgo/recovery/initiate.ts:112` — `bitgo.decrypt` (affects EOS/TRX/STX/XRP/UTXO backup-key recovery)
- `modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts:567, 601` — non-MPCv2 ECDSA offline signing rounds
- `modules/sdk-core/src/bitgo/keychain/keychains.ts:170` — `updateSingleKeychainPassword`
- `modules/sdk-core/src/bitgo/keychain/keychains.ts:354-364` — `recreateMpc` (three sync decrypts)
- `modules/sdk-core/src/bitgo/trading/tradingAccount.ts:72` — OFC payload signing
- `modules/sdk-core/src/bitgo/wallet/wallets.ts` (acceptShare / reshareWalletWithSpenders) — explicitly deferred per the PR comment

**Fix options (any one):** (a) convert each path to `decryptAsync` in this PR; (b) link follow-up tickets in the PR description for each path; (c) add a v2-opt-in guard at wallet creation that fails fast on these callers with a clear "v2 wallet operations require WCN-XX" error rather than letting SJCL surface a misleading tag-mismatch.

---

### Major findings

See inline comments below.

---

### Test coverage gap (not blocking, but please address)

The `useV2` dispatch branches in `createOfflineRound{1,2,3}Share` (MPCv2) and `getUserToBitgoCommitment` / `decryptRShare` (EdDSA) currently have **zero unit-test coverage**. CI is green on the v2 dispatch only because no test exercises the branch — the live testnet runs in the PR description don't substitute for unit-level branch coverage. Please add at least one v2 signing-path test per coin family.

Separately: `modules/bitgo/test/unit/decryptKeychain.ts` (5 sync call sites) was not updated for the new async signature. The test runner uses `tsx` (no type-check), so this is not a CI break — but the assertions now run against a `Promise` object. Either the file silently fails to load and is skipped, or it asserts against a Promise and that's masked by the `cdn=true` reporter option. Either way the file no longer validates the function. Please add `await` and make the test functions async.

---

### What was retracted from earlier review rounds

For the record:
- "CI will fail to compile" — wrong; `tsx` strips types. The decryptKeychain test file is a silent no-op, not a build break.
- "MPCv2 signing creates 3× Argon2id per signing" — wrong; each round is a separate HTTP request lifecycle, so cross-round session sharing is structurally impossible without server-side session storage.
- The v2 `adata` drop is documented design intent per commit `22bab61c` (AES-GCM is self-authenticating). The narrower remaining concern (cross-round replay binding) is captured in the inline comment on `ecdsaMPCv2.ts` below.

---

### Nits

- `isV2Envelope` is duplicated verbatim in `ecdsaMPCv2.ts` and `eddsa.ts` — extract to a shared util.
- `decryptAsync` falls through to v1 SJCL on unknown future versions (`v: 3+`). Should fail closed with `Error('unknown envelope version')`.
- No JSDoc on `EncryptOptions.adata` warning that it's silently dropped when `encryptionVersion: 2`.

pranavjain97:
> ## Review Summary
> After multiple validation rounds with parallel reviewers (architect, crypto, call-site tracer, test-gap, devil's-advocate, consensus-checker), **REQUEST CHANGES** with two blockers and a focused set of inline comments.
> 
> ### Blockers
> **B1. No breaking-change marker on commits.** None of the 7 commits on this PR carry `feat!:`, `fix!:`, or a `BREAKING CHANGE:` footer. `lerna publish --conventional-commits` (Angular preset) requires one of those for a major bump. Result: `@bitgo/sdk-core` will publish a _minor_ bump despite `IWallet.getUserPrv` and `decryptKeychainPrivateKey` changing from sync `string` returns to `Promise<string>`. Downstream JS / non-strict-TS consumers will silently get a Promise where they expect a string. The repo's `Check breaking changes` CI job only diffs OpenAPI specs from `modules/express` — it does **not** guard SDK type signatures, so this is a code-review responsibility entirely. **Fix:** add a `BREAKING CHANGE:` footer to commit `36c78cd` (or a follow-up commit) describing the async return-type change so lerna bumps `sdk-core` to a major version on publish.
> 
> **B2. Sync v1 `decrypt` fan-out callers not converted, no follow-up tickets linked.** A v2-encrypted wallet that flows into any of these sites hard-throws with a confusing SJCL tag-mismatch error. Blast radius is bounded today (v2 is opt-in), but every client that opts into v2 will trip at least one of these in normal usage:
> 
> * `modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts:647` — `assertIsValidKey` calls `sjcl.decrypt` directly
> * `modules/abstract-utxo/src/recovery/crossChainRecovery.ts:316` — sync `decrypt(passphrase, encryptedPrv)`
> * `modules/sdk-core/src/bitgo/recovery/initiate.ts:112` — `bitgo.decrypt` (affects EOS/TRX/STX/XRP/UTXO backup-key recovery)
> * `modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts:567, 601` — non-MPCv2 ECDSA offline signing rounds
> * `modules/sdk-core/src/bitgo/keychain/keychains.ts:170` — `updateSingleKeychainPassword`
> * `modules/sdk-core/src/bitgo/keychain/keychains.ts:354-364` — `recreateMpc` (three sync decrypts)
> * `modules/sdk-core/src/bitgo/trading/tradingAccount.ts:72` — OFC payload signing
> * `modules/sdk-core/src/bitgo/wallet/wallets.ts` (acceptShare / reshareWalletWithSpenders) — explicitly deferred per the PR comment
> 
> **Fix options (any one):** (a) convert each path to `decryptAsync` in this PR; (b) link follow-up tickets in the PR description for each path; (c) add a v2-opt-in guard at wallet creation that fails fast on these callers with a clear "v2 wallet operations require [WCN-XX](https://linear.app/bitgo/issue/WCN-XX)" error rather than letting SJCL surface a misleading tag-mismatch.
> 
> ### Major findings
> See inline comments below.
> 
> ### Test coverage gap (not blocking, but please address)
> The `useV2` dispatch branches in `createOfflineRound{1,2,3}Share` (MPCv2) and `getUserToBitgoCommitment` / `decryptRShare` (EdDSA) currently have **zero unit-test coverage**. CI is green on the v2 dispatch only because no test exercises the branch — the live testnet runs in the PR description don't substitute for unit-level branch coverage. Please add at least one v2 signing-path test per coin family.
> 
> Separately: `modules/bitgo/test/unit/decryptKeychain.ts` (5 sync call sites) was not updated for the new async signature. The test runner uses `tsx` (no type-check), so this is not a CI break — but the assertions now run against a `Promise` object. Either the file silently fails to load and is skipped, or it asserts against a Promise and that's masked by the `cdn=true` reporter option. Either way the file no longer validates the function. Please add `await` and make the test functions async.
> 
> ### What was retracted from earlier review rounds
> For the record:
> 
> * "CI will fail to compile" — wrong; `tsx` strips types. The decryptKeychain test file is a silent no-op, not a build break.
> * "MPCv2 signing creates 3× Argon2id per signing" — wrong; each round is a separate HTTP request lifecycle, so cross-round session sharing is structurally impossible without server-side session storage.
> * The v2 `adata` drop is documented design intent per commit `22bab61c` (AES-GCM is self-authenticating). The narrower remaining concern (cross-round replay binding) is captured in the inline comment on `ecdsaMPCv2.ts` below.
> 
> ### Nits
> * `isV2Envelope` is duplicated verbatim in `ecdsaMPCv2.ts` and `eddsa.ts` — extract to a shared util.
> * `decryptAsync` falls through to v1 SJCL on unknown future versions (`v: 3+`). Should fail closed with `Error('unknown envelope version')`.
> * No JSDoc on `EncryptOptions.adata` warning that it's silently dropped when `encryptionVersion: 2`.

Great findings.  Addressed all review comments across 3 commits:

  - Added EncryptionVersion type alias and IEncryptionSession interface for external consumers
  - Restored sync getUserPrv/decryptKeychainPrivateKey, added async variants -- no breaking change
  - Shared encryption session in EdDSA keygen (saves one Argon2id derivation), mirrors ecdsaMPCv2
  - Threaded AAD through the entire v2 encryption stack -- adata is stored in the v2 envelope and bound via GCM additionalData. DKLS signing passes adata to session.encrypt() and calls validateAdata() on v2 decrypt, same as v1.
  - Removed encryptionVersion from GG18 ECDSA (not in scope, tracked in WCN-283)
  - Extracted isV2Envelope to shared util, fail closed on unknown envelope version
  - Sync decrypt fan-out callers already tracked in WCN-175/269/281/282/283/284
  - Added e2e 3-round v2 signing test, AAD tamper detection tests, and fixed decryptKeychain test suite

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -2554,7 +2554,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
     const walletPassphrase = buildParams.walletPassphrase;
 
     const userKeychain = await this.keychains().get({ id: wallet.keyIds()[0] });
-    const userPrv = wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const userPrv = await wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
```
zahin-mohammad: **Blocker -- regression.** `getUserPrv` was reverted to sync in `62d00bcd33`, but the `await` added here in `083ebcf3` was not reverted. This still calls the sync method, which on a v2-encrypted wallet routes through SJCL and hard-throws with `password error - ccm: tag doesn't match`. The `await` makes this look correct but the runtime call is sync.

Fix:
```diff
-    const userPrv = await wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const userPrv = await wallet.getUserPrvAsync({ keychain: userKeychain, walletPassphrase });
```

This is `createHopTransactionParams` -- every EVM hop transaction on a v2 wallet would fail.

modules/abstract-utxo/src/impl/btc/inscriptionBuilder.ts
```diff
@@ -302,7 +302,7 @@ export class InscriptionBuilder implements IInscriptionBuilder {
     txPrebuild: PrebuildTransactionResult
   ): Promise<SubmitTransactionResponse> {
     const userKeychain = await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[KeyIndices.USER] });
-    const prv = this.wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const prv = await this.wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
```
zahin-mohammad: **Blocker -- same regression as `abstractEthLikeNewCoins.ts:2557`.** This now calls sync `getUserPrv` with a leftover `await`; v2 wallets hard-throw at SJCL.

```diff
-    const prv = await this.wallet.getUserPrv({ keychain: userKeychain, walletPassphrase });
+    const prv = await this.wallet.getUserPrvAsync({ keychain: userKeychain, walletPassphrase });
```

While you're here, line 265 (`signAndSendCommit`) has the same pattern (pre-existing -- the `await` was already there before this PR, but the function was always sync). Same `getUserPrv` → `getUserPrvAsync` fix applies. Both inscription flows would fail on v2 wallets without it.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -441,7 +462,17 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
 
     const encryptedSignerShare = this.createUserToBitgoEncryptedSignerShare(userToBitgoEncryptedSignerShare);
     const stringifiedRShare = JSON.stringify(userSignShare);
-    const encryptedRShare = this.bitgo.encrypt({ input: stringifiedRShare, password: params.walletPassphrase });
+    let encryptedRShare: string;
+    if (params.encryptedPrv && isV2Envelope(params.encryptedPrv)) {
```
zahin-mohammad, line 466: **Major -- no unit test for the EdDSA v2 signing dispatch.** The v2 dispatch branches here (`createCommitmentShareFromTxRequest` line 466 and `createRShareFromTxRequest` line 489) are gated by `isV2Envelope(...)` and never exercised by any test. The only EdDSA v2 test in `tssUtils/eddsa.ts:581` covers keygen.

MPCv2 has an e2e v2 round-trip test (`signTxRequest.ts:270-372`) -- please add the equivalent for EdDSA. At minimum:
1. Encrypt a prv via `encryptAsync({encryptionVersion: 2})`.
2. Call `createCommitmentShareFromTxRequest` with the v2 `encryptedPrv` -- assert `encryptedUserToBitgoRShare.share` is a v2 envelope.
3. Round-trip through `createRShareFromTxRequest` -- assert it decrypts correctly.

Without this, a regression in either dispatch branch will pass CI.

modules/bitgo/test/v2/unit/internal/tssUtils/ecdsaMPCv2/signTxRequest.ts
```diff
@@ -267,6 +267,122 @@ describe('signTxRequest:', function () {
...
+
+    it('validateAdata rejects v2 envelopes with mismatched adata', async function () {
+      const ct = await bitgo.encryptAsync({
+        input: 'test-data',
+        password: 'testpass',
+        encryptionVersion: 2,
+        adata: 'context-A',
+      });
+
+      (() => (tssUtils as any).validateAdata('context-B', ct)).should.throw(/Adata does not match/);
```
zahin-mohammad: **Minor -- test calls `validateAdata` with 2 args, signature requires 3.** Per the round-domain-separator commit `e7dfa8e865`, the signature is `validateAdata(adata, cyphertext, roundDomainSeparator)`. Here `roundDomainSeparator` is `undefined`.

The test passes by accident: when `roundDomainSeparator` is `undefined`, the function compares `envelope.adata` (`"context-A"`) against `"undefined:context-B"` and `"context-B"` -- both fail, so it throws as expected. But this means the test wouldn't catch a regression in the round-domain binding logic.

Pass an explicit `roundDomainSeparator` so the test exercises the intended check:

```ts
(() => (tssUtils as any).validateAdata('context-B', ct, EcdsaMPCv2Utils.DKLS23_SIGNING_ROUND1_STATE)).should.throw(/Adata does not match/);
```

Ideally also add a positive case asserting `validateAdata` accepts a correctly-bound envelope.

modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -105,20 +107,26 @@ export const GenerateLightningWalletOptionsCodec = t.intersection(
     }),
     t.partial({
       lightningProvider: t.union([t.literal('amboss'), t.literal('voltage')]),
+      encryptionVersion: t.literal(2),
```
zahin-mohammad, line 111: **Minor -- codec narrower than declared TS type.** `EncryptionVersion = 1 | 2` (`api/types.ts:20`), but this codec uses `t.literal(2)`. A caller forwarding a generic `cfg.version: EncryptionVersion` will pass type-check but fail codec validation when `version === 1`.

Either widen the codec to allow both:
```ts
encryptionVersion: t.union([t.literal(1), t.literal(2)])
```

or add a comment explaining that the codec deliberately rejects `1` (since v1 is the implicit default and explicitly passing it through the wire is unexpected). Same applies to line 127.

zahin-mohammad (changes requested):
## Re-review summary

Most of the prior round's blockers and majors are resolved -- thanks for the careful refactor. Three items remain before merge:

### Resolved ✅
- **B1** (breaking-change marker on `getUserPrv` / `decryptKeychainPrivateKey`): solved differently and better -- sync versions kept, `*Async` variants added. No public API break.
- **AAD threading**: `aesGcmEncrypt`/`Decrypt`, `EncryptionSession.encrypt(plaintext, adata?)`, and `encryptV2({adata})` now bind AAD via the GCM tag, with `validateAdata` enforcing context binding on v2 round-session decryption (`ecdsaMPCv2.ts:1267, 1366`). Round domain separators applied.
- `EncryptionVersion = 1 | 2` type alias and `IEncryptionSession` interface extracted to `sdk-core/src/api/types.ts`.
- `isV2Envelope` extracted to shared `baseTypes.ts`.
- `decryptAsync` fails closed on unknown envelope versions.
- EdDSA single shared session in `createKeychains` -- saves one Argon2id derivation per keygen.
- `decryptKeychain.ts` test file now has both sync and async test suites.
- MPCv2 e2e v2 signing test exercises `createOfflineRound{1,2,3}` with adata round-trip.

### Remaining

**B3 (blocker, must fix).** Three call sites still `await` the sync `getUserPrv` (left over from the async-revert in `62d00bcd33`). On v1 wallets these are no-ops; on v2 wallets they hard-throw with a misleading SJCL tag-mismatch. See inline comments on:
- `modules/abstract-eth/src/abstractEthLikeNewCoins.ts:2557`
- `modules/abstract-utxo/src/impl/btc/inscriptionBuilder.ts:305` (line 265 in the same file is a pre-existing version of the same issue and should also be flipped)

Fix is one token per site: `getUserPrv` → `getUserPrvAsync`.

**M1 (acknowledgement required).** The original B2 sync-decrypt fan-out is unaddressed and there are no follow-up tickets linked in the PR description. v2-encrypted wallets that flow through any of these will surface the same misleading SJCL tag-mismatch error:

- `sdk-core/src/bitgo/baseCoin/baseCoin.ts:647` -- `assertIsValidKey` (`sjcl.decrypt` direct)
- `sdk-core/src/bitgo/keychain/keychains.ts:170` -- `updateSingleKeychainPassword`
- `sdk-core/src/bitgo/keychain/keychains.ts:361, 366, 371` -- `recreateMpc` (3× sync decrypt)
- `sdk-core/src/bitgo/wallet/wallets.ts:972, 1016, 1027, 1092, 1127, 1256, 1376, 1440` -- share/reshare flows
- `sdk-core/src/bitgo/recovery/initiate.ts:112` -- recovery path
- `sdk-core/src/bitgo/utils/tss/ecdsa/ecdsa.ts:567, 601` -- non-MPCv2 ECDSA offline signing rounds
- `sdk-core/src/bitgo/trading/tradingAccount.ts:72` -- OFC payload signing
- `abstract-utxo/src/recovery/crossChainRecovery.ts:316` -- sync `decrypt(passphrase, encryptedPrv)`

Given v2 is opt-in and not yet flipped to default, **I'm OK with this not being fixed in this PR** -- but please file follow-up tickets and link them in the PR description so we don't lose track. Could you confirm the plan here?

**M2 (must add).** EdDSA v2 signing dispatch has no unit-test coverage. The MPCv2 e2e test is great, but the equivalent v2 branches in EdDSA are uncovered. See inline on `eddsa.ts:466`.

### Minors (nice-to-have)
- Test arity bug on `validateAdata` -- inline.
- Codec/type alignment on `encryptionVersion` -- inline.

### Security and back-compat summary
- Crypto design is sound after the AAD-threading commit. v2 envelope adata is bound by GCM tag; `validateAdata` provides context binding; cross-round/cross-tx replay prevented to v1 parity. Argon2 parameters bounded -- DoS via crafted envelopes capped.
- Public API is additive only (no breaking changes). Conventional `feat:` is correct.

Will re-approve once B3 is fixed, M1 is acknowledged with linked tickets, and M2 has at least one test.

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
pranavjain97: fixed all instances

pranavjain97:
@zahin-mohammad all comments addressed. 

>Given v2 is opt-in and not yet flipped to default, I'm OK with this not being fixed in this PR -- but please file follow-up tickets and link them in the PR description so we don't lose track. Could you confirm the plan here?

We have all the tickets for that as part of the migration project. Updated the PR desc to document the known gaps in v2.

zahin-mohammad (dismissed):

basel-bitgo (dismissed):

davidkaplanbitgo (approved):

basel-bitgo (approved):

zahin-mohammad (approved):

PR #8599, opened 2026-04-22
feat(sdk-lib-mpc): add EdDSA DSG MPS class

vibhavgo:
## Summary

This PR introduces the `DSG` (Distributed Sign Generation) class for EdDSA MPCv2 in `sdk-lib-mpc`, wrapping the `@bitgo/wasm-mps` WASM bindings for the 2-of-3 threshold Multi-Party Schnorr signing protocol. The class follows the same design pattern as the existing `DKG` class — explicit round-state management with session export/restore — so state can be persisted to a database between HTTP round-trips.

**Ticket**: [WCI-164](https://linear.app/bitgo/issue/WCI-164)

## Changes

- **`dsg.ts`** — New `DSG` class with 4-round state machine (Init → WaitMsg1 → WaitMsg2 → WaitMsg3 → Complete), wrapping `ed25519_dsg_round{0..3}_process`. `getSession()` / `restoreSession()` serialise state to/from JSON for DB persistence.
- **`types.ts`** — Adds `DsgState` enum mirroring `DkgState`.
- **`index.ts`** — Exports `DSG` under the `EddsaMPSDsg` namespace.
- **`test/unit/tss/eddsa/dsg.ts`** — Tests covering: full 2-of-3 signing across all
    party-pair combos, session restore continuity, message serialization, derivation-path
    differentiation, and error-handling guards.
- **`test/unit/tss/eddsa/util.ts`** — Adds `runEdDsaDSG` test helper.

## Test Plan

- [ ] `yarn unit-test --scope @bitgo/sdk-lib-mpc` — all tests pass
- [ ] All three 2-of-3 party pairs produce identical signatures verifying under DKG pubkey
- [ ] Session export/restore produces same signature as an uninterrupted session

CLOSES TICKET: WCI-164

### DIFF
--- a/modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
+++ b/modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
@@ -1,0 +1,298 @@
+import assert from 'assert';
+import {
+  ed25519_dsg_round0_process,
+  ed25519_dsg_round1_process,
+  ed25519_dsg_round2_process,
+  ed25519_dsg_round3_process,
+} from '@bitgo/wasm-mps';
+import { DeserializedMessage, DeserializedMessages, DsgState } from './types';
+
+/**
+ * EdDSA Distributed Sign Generation (DSG) implementation using @bitgo/wasm-mps.
+ *
+ * State is explicit: each WASM round function returns
+ * `{ msg, state }` bytes; the state bytes are stored between rounds and passed to the
+ * next round function (this is what a server would persist to a database between API
+ * rounds).
+ *
+ * The protocol is hard-coded 2-of-3: each signing party communicates with exactly one
+ * counterpart. `handleIncomingMessages` accepts both messages (own + counterpart), and
+ * filters own out internally.
+ *
+ * @example
+ * ```typescript
+ * const dsg = new DSG(0);  // partyIdx 0
+ * dsg.initDsg(keyShare, message, 'm', 2);  // counterpart is party 2
+ * const msg1 = dsg.getFirstMessage();
+ * const msg2 = dsg.handleIncomingMessages([msg1, peerMsg1]);  // emits SignMsg2
+ * const msg3 = dsg.handleIncomingMessages([msg2[0], peerMsg2]);  // emits SignMsg3
+ * dsg.handleIncomingMessages([msg3[0], peerMsg3]);  // completes DSG
+ * const signature = dsg.getSignature();  // 64-byte Ed25519 signature
+ * ```
+ */
+export class DSG {
+  protected partyIdx: number;
+  protected otherPartyIdx: number | null = null;
+
+  /** Opaque bincode-serialised Keyshare from a prior DKG */
+  private keyShare: Buffer | null = null;
+  /** Raw message bytes to sign (Ed25519 hashes internally; no prehashing required) */
+  private message: Buffer | null = null;
+  /** BIP-32-style derivation path, e.g. "m" or "m/0/1". Folded in via Keyshare::derive_with_offset */
+  private derivationPath: string | null = null;
+
+  /** Serialised round state bytes returned by the previous round function */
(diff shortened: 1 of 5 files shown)


Marzooqa (dismissed):

johnoliverdriscoll (dismissed):

johnoliverdriscoll (commented):
Why is the message signature not verified in this library? Where does it get verified?

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
```diff
@@ -0,0 +1,239 @@
...
+    this.dsgState = DsgState.Init;
+  }
+
+  /**
+   * Runs round 0 of the DSG protocol. Returns this party's broadcast message
+   * (a `SignMsg1` containing the commitment to `R_i`). Stores the round state
+   * bytes internally for the next round.
+   */
+  getFirstMessage(): DeserializedMessage {
+    if (this.dsgState !== DsgState.Init) {
```
pranavjain97, line 96: please always avoid non-null assertions. use `assert()` before the try block to narrow the types instead

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
```diff
@@ -0,0 +1,239 @@
...
+      return [];
+    }
+
+    throw Error('Unexpected DSG state');
+  }
+
+  /**
+   * Returns the final 64-byte Ed25519 signature produced by round 3.
+   * Only available once the protocol reaches `Complete`.
+   */
```
pranavjain97, line 201: `getSession` permits export at `Init` state before `getFirstMessage`, but `dsgStateBytes` is null at that point. should `Init` be blocked here?

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
```diff
@@ -0,0 +1,239 @@
...
+      throw Error('DSG session is complete. Exporting the session is not allowed.');
+    }
+    if (this.dsgState === DsgState.Uninitialized) {
+      throw Error('DSG session not initialized');
+    }
+    return JSON.stringify({
+      dsgStateBytes: this.dsgStateBytes?.toString('base64') ?? null,
+      dsgRound: this.dsgState,
+      keyShare: this.keyShare?.toString('base64') ?? null,
+      message: this.message?.toString('base64') ?? null,
```
pranavjain97, line 228: `restoreSession` does no validation. a tampered session string can set `dsgState` to an arbitrary value or `partyIdx` out of range, bypassing `initDsg` guards

vibhavgo:
> Why is the message signature not verified in this library? Where does it get verified?

Message signature will be verified before each `handleIncomingMessages()`
Ref: https://github.com/BitGo/BitGoJS/blob/4cd21945db87b582e2812cffebe3b546317124d4/modules/sdk-lib-mpc/src/tss/eddsa-mps/commsLayer.ts#L27

johnoliverdriscoll (approved):

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
vibhavgo: Will do

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
vibhavgo: Blocked `Init`

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
vibhavgo: Agreed
Will do

pranavjain97 (approved):

PR #8603, opened 2026-04-22
refactor(express): rename verifyAddress to verifyCoinAddress in v2 ty…

rishikeshdadam136:
Ticket: WCI-227

Changing  the name for the file /Users/rishikeshdadam/BitGoJS/modules/express/src/typedRoutes/api/v2/verifyAddress.ts into verifyCoinAddress inorder to get rid of naming conflict when we are splitting the verifyAddress common endpoint into v1 and v2.


lokesh-bitgo (approved):
Looks good

PR #8623, opened 2026-04-23
fix: add EdDSA MPCv2 chain code support for address derivation

Marzooqa:
- Bump @bitgo/wasm-mps to 1.7.0 which exposes chaincode on the Share struct
- Add shareChaincode field and getCommonKeychain() to DKG class in sdk-lib-mpc, returning pubkey+chaincode (128 hex chars) matching Eddsa.deriveUnhardened expectations
- Use getCommonKeychain() in EddsaMPCv2Utils.createKeychains and pass full keychain to user/backup/bitgo keychains instead of just the 32-byte public key
- Rename commonPublicKey to commonPublicKeychain in round 2 response handling to align with updated @bitgo/public-types and HSM response field name
- Remove 64-char MPCv2 special case from getPublicKeyFromCommonKeychain; all EdDSA keychains are now uniformly 128 hex chars (pubkey + chaincode)
- Update tests accordingly

TICKET: WCI-230

### DIFF
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
@@ -201,7 +201,7 @@ describe('TSS EdDSA MPCv2 Utils:', async function () {
         .once()
         .reply(200, {
           sessionId: 'test-session-id',
-          commonPublicKey: 'a'.repeat(64),
+          commonPublicKeychain: 'a'.repeat(128),
           bitgoMsg2: {
             message: Buffer.from('garbage').toString('base64'),
             signature: '-----BEGIN PGP SIGNATURE-----\nFAKE\n-----END PGP SIGNATURE-----',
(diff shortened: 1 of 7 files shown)


modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
```diff
@@ -201,7 +201,7 @@ describe('TSS EdDSA MPCv2 Utils:', async function () {
         .once()
         .reply(200, {
           sessionId: 'test-session-id',
-          commonPublicKey: 'a'.repeat(64),
+          commonPublicKeychain: 'a'.repeat(128),
```
zahin-mohammad, line 204: This PR didn't change what gets sent to the server, so how come this mock was changed? Was it just ignored before?

zahin-mohammad (approved):

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
Marzooqa: We updated what gets returned from the server hence the change here, plus for consistency.

PR #8624, opened 2026-04-23
feat: implement BitGo signing in SDK

alextse-bg:
### Commit 1: allowing trading wallet transaction signing on Trading Account Objects

make wallet passphrase optional when signing OFC transactions.
if not present, the SDK attempts to sign using the wallet's BitGo key instead.

### Commit 2: allowing trading wallet transaction signing on Wallet and Coins object

The following are all of the currently valid methods to create a signature on an OFC wallet's `payload` string

1.  `ofcToken.signMessage({prv}, message)`: encrypts the message locally using prv
2. `ofcToken.signTransaction(params)`: signs a half signed transaction by calling the above method
3. `wallet.baseCoin.signMessage`: see above
4. `wallet.baseCoin.signTransaction`: see above
5. `wallet.signTransaction(params)`: signs a half signed transaction by getting the prv through a wallet passphrase then calling `this.baseCoin.signTransaction`
6. `wallet.prebuildAndSignTransaction(params)`: builds and sign a transaction by calling `wallet.signTransaction`
7. all other wallet methods that calls `wallet.prebuildAndSignTransaction` (e.g. sendMany)
8. `wallet.toTradingAccount().signPayload`: signs a half signed transaction using the wallet passphrase

Changes in commit 1 address path 8 already.

For paths that creates the signature using methods of wallet object (i.e. 5-7), all of them eventually calls `wallet.signTransaction`, which pass itself `this` as an argument to `wallet.baseCoin.signTransaction` (see [here](https://github.com/BitGo/BitGoJS/blob/1b0a24b8d2522f3e70f614fb1bac49e824ab4e2e/modules/sdk-core/src/bitgo/wallet/wallet.ts#L2257)), allowing us to sign via BitGo key if we add the implementation to `ofcToken`

As for `ofcToken.signMessage`, add overloads to the method to allow SDK user to pass in the wallet object instead, which creates the signature via the BitGo key.

Note that the `walletPassphrase` is already an optional parameter when calling wallet level methods.

Ticket: WCN-217

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -403,6 +403,15 @@ function getWalletPwFromEnv(walletId: string): string {
   return walletPw;
 }
 
+/**
+ * Returns the wallet passphrase from the environment, or undefined if not set.
+ * Unlike getWalletPwFromEnv, this does not throw when the env variable is absent.
+ * Use this when the passphrase is optional (e.g. KMS-backed wallets).
+ */
+function findWalletPwFromEnv(walletId: string): string | undefined {
+  return process.env[`WALLET_${walletId}_PASSPHRASE`];
+}
+
 async function getEncryptedPrivKey(path: string, walletId: string): Promise<string> {
   const privKeyFile = await fs.readFile(path, { encoding: 'utf8' });
   const encryptedPrivKey = JSON.parse(privKeyFile);
@@ -629,7 +638,9 @@ export async function handleV2OFCSignPayload(
     throw new ApiResponseError(`Could not find OFC wallet ${walletId}`, 404);
   }
 
-  const walletPassphrase = bodyWalletPassphrase || getWalletPwFromEnv(wallet.id());
+  // Prefer the passphrase from the request body; fall back to the env var.
+  // If neither is present, pass undefined — signPayload() routes to KMS internally.
+  const walletPassphrase = bodyWalletPassphrase ?? findWalletPwFromEnv(wallet.id());
   const tradingAccount = wallet.toTradingAccount();
   const stringifiedPayload = typeof payload === 'string' ? payload : JSON.stringify(payload);
   const signature = await tradingAccount.signPayload({
--- a/modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
+++ b/modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
@@ -2,7 +2,7 @@ import { ITradingNetwork } from './network';
 
 export interface SignPayloadParameters {
   payload: string | Record<string, unknown>;
-  walletPassphrase: string;
+  walletPassphrase?: string;
 }
 
 export interface ITradingAccount {
--- a/modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
+++ b/modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
@@ -1,5 +1,10 @@
 import { ITradingNetwork } from './network';
 
+/**
+ * Parameters for the signing a payload from the trading account
+ * @param payload - The payload to sign
+ * @param walletPassphrase - The passphrase of the wallet that will be used to decrypt the user key and sign the payload. If not provided, the BitGo key will be used.
+ */
 export interface SignPayloadParameters {
   payload: string | Record<string, unknown>;
   walletPassphrase?: string;
--- a/modules/sdk-core/src/bitgo/trading/tradingAccount.ts
+++ b/modules/sdk-core/src/bitgo/trading/tradingAccount.ts
@@ -23,13 +23,47 @@ export class TradingAccount implements ITradingAccount {
   }
 
   /**
-   * Signs an arbitrary payload with the user key on this trading account
+   * Signs an arbitrary payload. Use the user key if passphrase is provided, or the BitGo key if not.
    * @param params
    * @param params.payload arbitrary payload object (string | Record<string, unknown>)
    * @param params.walletPassphrase passphrase on this trading account, used to unlock the account user key
    * @returns hex-encoded signature of the payload
    */
   async signPayload(params: SignPayloadParameters): Promise<string> {
+    // if no passphrase is provided, attempt to sign using the wallet's bitgo key remotely
+    if (!params.walletPassphrase) {
+      return this.signPayloadByBitGoKey(params);
+    }
+    // if a passphrase is provided, we must be trying to sign using the user private key - decrypt and sign locally
+    return this.signPayloadByUserKey(params);
+  }
+
+  /**
+   * Signs the payload of a trading account via the trading account BitGo key stored in a remote KMS
+   * @param params
+   * @private
+   */
+  private async signPayloadByBitGoKey(params: Omit<SignPayloadParameters, 'walletPassphrase'>): Promise<string> {
+    const walletData = this.wallet.toJSON();
+    if (walletData.userKeySigningRequired) {
+      throw new Error('Wallet must use user key to sign ofc transaction, please provide the wallet passphrase');
+    }
+    if (walletData.keys.length < 2) {
+      throw new Error('Wallet does not support BitGo signing');
+    }
+
+    const url = this.wallet.url('/tx/sign');
+    const { signature } = await this.wallet.bitgo.post(url).send(params.payload).result();
+
+    return signature;
+  }
+
+  /**
+   * Signs the payload of a trading account locally by fetching the user's encrypted private key and decrypt using passphrase
+   * @param params
+   * @private
+   */
+  private async signPayloadByUserKey(params: SignPayloadParameters): Promise<string> {
     const key = (await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[0] })) as any;
     const prv = this.wallet.bitgo.decrypt({
       input: key.encryptedPrv,
--- a/modules/sdk-core/src/bitgo/trading/tradingAccount.ts
+++ b/modules/sdk-core/src/bitgo/trading/tradingAccount.ts
@@ -39,17 +39,21 @@ export class TradingAccount implements ITradingAccount {
   }
 
   /**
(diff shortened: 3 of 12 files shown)


modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
```diff
@@ -2,7 +2,7 @@ import { ITradingNetwork } from './network';
 
 export interface SignPayloadParameters {
   payload: string | Record<string, unknown>;
-  walletPassphrase: string;
+  walletPassphrase?: string;
```
zahin-mohammad, line 10: Nit: add jsdoc describing why this is optional.

modules/sdk-core/src/bitgo/trading/tradingAccount.ts
```diff
@@ -23,13 +23,47 @@ export class TradingAccount implements ITradingAccount {
...
    * @param params
    * @param params.payload arbitrary payload object (string | Record<string, unknown>)
    * @param params.walletPassphrase passphrase on this trading account, used to unlock the account user key
    * @returns hex-encoded signature of the payload
    */
   async signPayload(params: SignPayloadParameters): Promise<string> {
+    // if no passphrase is provided, attempt to sign using the wallet's bitgo key remotely
+    if (!params.walletPassphrase) {
+      return this.signPayloadByBitGoKey(params);
+    }
```
zahin-mohammad, line 36: First check the wallet setting. 

If there is no walletPassphrase and `wallet.userKeySigningRequired` is set to true, then throw an error.

zahin-mohammad: The backend will check it anyways, but best to do the validation here as well.

modules/sdk-core/src/bitgo/trading/tradingAccount.ts
```diff
@@ -23,13 +23,47 @@ export class TradingAccount implements ITradingAccount {
...
+    // if a passphrase is provided, we must be trying to sign using the user private key - decrypt and sign locally
+    return this.signPayloadByUserKey(params);
+  }
+
+  /**
+   * Signs the payload of a trading account via the trading account BitGo key stored in a remote KMS
```
zahin-mohammad: >    * Signs the payload of a trading account via the BitGo key

modules/sdk-core/src/bitgo/trading/tradingAccount.ts
```diff
@@ -23,13 +23,47 @@ export class TradingAccount implements ITradingAccount {
...
+   * @param params
+   * @private
+   */
+  private async signPayloadByBitGoKey(params: Omit<SignPayloadParameters, 'walletPassphrase'>): Promise<string> {
+    const walletData = this.wallet.toJSON();
+    if (walletData.userKeySigningRequired) {
+      throw new Error('Wallet must use user key to sign ofc transaction, please provide the wallet passphrase');
```
zahin-mohammad:
```
throw new Error('Wallet must use user key to sign ofc transaction, please provide the wallet passphrase or visit your wallet settings page to configure one.');

File: modules/sdk-core/src/bitgo/trading/tradingAccount.ts [hunk 5]
```diff
@@ -23,13 +23,47 @@ export class TradingAccount implements ITradingAccount {
...
+    }
+    if (walletData.keys.length < 2) {
+      throw new Error('Wallet does not support BitGo signing');
+    }
```
zahin-mohammad, line 57:
```
throw new Error('Wallet does not support BitGo signing. Please reach out to support@bitgo.com to resolve this issue.');
```

modules/sdk-core/src/coins/ofc.ts
```diff
@@ -104,6 +105,26 @@ export class Ofc extends BaseCoin {
...
+  async signMessage(wallet: Wallet, message: string): Promise<Buffer>;
+  /**
+   * Signs a message using the private key
+   * @param key - uses the private key to sign the message
+   * @param message
+   */
+  async signMessage(key: { prv: string }, message: string): Promise<Buffer>;
+  async signMessage(keyOrWallet: { prv: string } | Wallet, message: string): Promise<Buffer> {
+    if (!(keyOrWallet instanceof Wallet)) {
+      return super.signMessage(keyOrWallet as { prv: string }, message);
```
zahin-mohammad: do you still need the cast here? I thought the type would be narrowed by the time you are inside this if block.

modules/sdk-core/src/coins/ofc.ts
```diff
@@ -104,6 +105,26 @@ export class Ofc extends BaseCoin {
...
+    }
+    const signatureHexString = await (keyOrWallet as Wallet).toTradingAccount().signPayload({ payload: message });
```
zahin-mohammad: same here, I'm surprised you still need a cast after type narrowing.

modules/express/src/clientRoutes.ts
```diff
@@ -403,6 +403,15 @@ function getWalletPwFromEnv(walletId: string): string {
   return walletPw;
 }
 
+/**
+ * Returns the wallet passphrase from the environment, or undefined if not set.
+ * Unlike getWalletPwFromEnv, this does not throw when the env variable is absent.
```
zahin-mohammad, line 408: Why not just update `getWalletPwFromEnv` to make the throwing behaviour configurable?

zahin-mohammad (dismissed):
lgtm, just a few suggestions/comments.

modules/sdk-core/src/bitgo/trading/tradingAccount.ts
alextse-bg: there is a check in line 48

modules/express/src/clientRoutes.ts
alextse-bg: @mrdanish26

alextse-bg:
### Commit 3
Make wallet passphrase optional for preapreAllocation

Ticket: WCN-216

modules/express/src/clientRoutes.ts
mrdanish26: I went with a separate function to keep the return types unambiguous, `getWalletPwFromEnv` always returns string so all existing callers remain unchanged, while `findWalletPwFromEnv` explicitly returns `string | undefined` making it clear at the call site that the caller must handle the missing case.

zahin-mohammad (approved):

modules/sdk-core/src/coins/ofcToken.ts
```diff
@@ -107,15 +109,25 @@ export class OfcToken extends Ofc {
...
   async signTransaction(params: SignTransactionOptions): Promise<SignedTransaction> {
     const txPrebuild = params.txPrebuild;
     const payload = txPrebuild.payload;
-    const signatureBuffer = (await this.signMessage(params, payload)) as any;
-    const signature: string = signatureBuffer.toString('hex');
+
+    let signature: string;
+    if (params.wallet) {
+      signature = await params.wallet.toTradingAccount().signPayload({ payload, walletPassphrase: params.prv });
+    } else if (params.prv) {
```
pranavjain97, line 124: `params.prv` arriving here via `wallet.signTransaction` is already a decrypted key (from `getUserPrv`), not a passphrase. passing it as `walletPassphrase` to `signPayload` will cause `signPayloadByUserKey` to use it as a decryption password against `encryptedPrv`, which won't work

modules/sdk-core/src/bitgo/wallet/iWallet.ts
```diff
@@ -909,6 +909,7 @@ export interface WalletData {
   evmKeyRingReferenceWalletId?: string;
   isParent?: boolean;
   enabledChildChains?: string[];
+  userKeySigningRequired?: string;
```
pranavjain97, line 912: `userKeySigningRequired` is typed `string` but the guard treats it as boolean. `"false"` from the API would incorrectly block BitGo signing. should be `boolean`

modules/sdk-core/src/bitgo/trading/tradingAccount.ts
```diff
@@ -23,13 +23,51 @@ export class TradingAccount implements ITradingAccount {
...
+    const walletData = this.wallet.toJSON();
+    if (walletData.userKeySigningRequired) {
+      throw new Error(
+        'Wallet must use user key to sign ofc transaction, please provide the wallet passphrase or visit your wallet settings page to configure one.'
+      );
+    }
+    if (walletData.keys.length < 2) {
+      throw new Error(
+        'Wallet does not support BitGo signing. Please reach out to support@bitgo.com to resolve this issue.'
+      );
```
pranavjain97, line 56: when `params.payload` is a string, `.send(string)` sends a plain text body, not JSON. the user-key path stringifies first, this should match

modules/express/src/clientRoutes.ts
```diff
@@ -403,6 +403,15 @@ function getWalletPwFromEnv(walletId: string): string {
...
+ * Use this when the passphrase is optional (e.g. KMS-backed wallets).
+ */
+function findWalletPwFromEnv(walletId: string): string | undefined {
+  return process.env[`WALLET_${walletId}_PASSPHRASE`];
+}
+
 async function getEncryptedPrivKey(path: string, walletId: string): Promise<string> {
   const privKeyFile = await fs.readFile(path, { encoding: 'utf8' });
   const encryptedPrivKey = JSON.parse(privKeyFile);
```
pranavjain97, line 417: +1 to zahin's thread: collapse `findWalletPwFromEnv` and `getWalletPwFromEnv` into one function

pranavjain97 (changes requested):
prv-as-walletPassphrase in ofcToken.signTransaction breaks the user-key signing path from wallet.signTransaction

modules/express/src/clientRoutes.ts
mrdanish26: Will update it in a follow up PR

PR #8630, opened 2026-04-24
Fix/wcn 196 eddsa recipient guard

mrdanish26:
## Summary

  Follow-up to WAL-375 (#8539). Applies the same recipient verification guard to the **EdDSA** TSS signing path.

  ### The vulnerability (same root cause as WAL-375)

  `EddsaUtils.signRequestBase()` did not call `verifyTransaction` with the actual recipients, allowing a compromised API server to substitute a malicious `signableHex` — the transaction presented to the user for signing could differ from what was actually broadcast.

  ### Fix

  Added `resolveEffectiveTxParams` + `verifyTransaction` call inside `signRequestBase()` for `RequestType.tx`, immediately after `unsignedTx` is resolved and before key derivation begins:

  ```typescript
  const effectiveTxParams = resolveEffectiveTxParams(txRequestResolved, params.txParams);
  await this.baseCoin.verifyTransaction({
    txPrebuild: { txHex: unsignedTx.signableHex },
    txParams: effectiveTxParams,
    wallet: this.wallet,
    walletType: this.wallet.multisigType(),
  });

  EdDSA-specific NO_RECIPIENT_TX_TYPES expansion

  EdDSA coins have many more transaction types that legitimately carry no recipients. The shared NO_RECIPIENT_TX_TYPES set in recipientUtils.ts was expanded from 8 → 44 types:

  ┌───────────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                   Category                    │                                                                                                          Types                                                                                                           │
  ├───────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Staking (SOL, ADA, NEAR, DOT, CSPR, SUI, APT) │ stakingActivate, stakingDeactivate, stakingWithdraw, stakingClaim, stakingDelegate, stakingUnlock, stakingUnvote, stakingPledge, stakingAuthorize, stakingAuthorizeRaw, stakingLock, stakingAdd, addStake, withdrawStake │
  ├───────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Account / wallet init                         │ walletInitialization, addressInitialization, createAccount, accountUpdate                                                                                                                                                │
  ├───────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Token operations                              │ trustline, closeAssociatedTokenAccount, associatedTokenAccountInitialization, tokenTransfer, sendToken, sendNFT                                                                                                          │
  ├───────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ NEAR                                          │ storageDeposit                                                                                                                                                                                                           │
  ├───────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ TON                                           │ singleNominatorWithdraw, tonWhalesDeposit, tonWhalesWithdrawal, tonWhalesVestingDeposit, tonWhalesVestingWithdrawal                                                                                                      │
  ├───────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ADA                                           │ voteDelegation                                                                                                                                                                                                           │
  ├───────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Canton 2-step transfers                       │ transferAccept, transferReject, transferAcknowledge, transferOfferWithdrawn, oneStepPreApproval                                                                                                                          │
  └───────────────────────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

  intent.intentType fallback

  EdDSA coins populate txRequest.intent.intentType but often don't set txParams.type. The guard now falls back to intent.intentType when txParams.type is absent — this is required for EdDSA staking and wallet init flows to pass the guard without breaking.

  Test plan

  - recipientUtils unit tests — 20 passing (ECDSA types, EdDSA staking types, Canton types, intentType fallback)
  - EdDSA signTxRequest tests — 9 passing (rejection for payment without recipients, acceptance for stakingActivate and walletInitialization)
  - ECDSA MPCv2 signTxRequest tests — 13 passing (no regression on shared recipientUtils.ts)

  Known edge cases

  - XLM: verifyTransaction makes 2 keychain API calls per sign — ~100–300ms latency, pre-existing behaviour not introduced here
  - HBAR / DOT: strict recipient checks in verifyTransaction for some types — mitigated by comprehensive exempt list; raise a follow-up if new types surface
  ```

TICKET: WCN-196

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
@@ -43,6 +43,7 @@ import { IRequestTracer } from '../../../../api';
 import { getBitgoMpcGpgPubKey } from '../../../tss/bitgoPubKeys';
 import { EnvironmentName } from '../../../environments';
 import { readKey } from 'openpgp';
+import { resolveEffectiveTxParams } from '../recipientUtils';
 
 /**
  * Utility functions for TSS work flows.
@@ -610,6 +611,14 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
       );
       unsignedTx =
         apiVersion === 'full' ? txRequestResolved.transactions![0].unsignedTx : txRequestResolved.unsignedTxs[0];
+
+      const effectiveTxParams = resolveEffectiveTxParams(txRequestResolved, params.txParams);
+      await this.baseCoin.verifyTransaction({
+        txPrebuild: { txHex: unsignedTx.signableHex },
+        txParams: effectiveTxParams,
+        wallet: this.wallet,
+        walletType: this.wallet.multisigType(),
+      });
     } else if (requestType === RequestType.message) {
       assert(txRequestResolved.messages?.length, 'Unable to find messages in txRequest for message signing');
       const message = txRequestResolved.messages[0];
--- a/modules/sdk-core/test/unit/bitgo/utils/tss/recipientUtils.ts
+++ b/modules/sdk-core/test/unit/bitgo/utils/tss/recipientUtils.ts
@@ -23,9 +23,9 @@ function makeTxRequest(
 
 describe('recipientUtils', function () {
   describe('NO_RECIPIENT_TX_TYPES', function () {
-    it('contains exactly the 8 expected exempted types', function () {
+    it('contains all ECDSA exempted types', function () {
       const { NO_RECIPIENT_TX_TYPES } = getModule();
-      const expected = [
+      const ecdsaTypes = [
         'acceleration',
         'fillNonce',
         'transferToken',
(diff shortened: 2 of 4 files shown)


modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
```diff
@@ -610,6 +611,14 @@ export class EddsaUtils extends baseTSSUtils<KeyShare> {
       );
       unsignedTx =
```
pranishnepal, line 612: why do we only consider the first item in the array - is that the only one that _matters_? 
```
 txRequestResolved.transactions![0].unsignedTx : txRequestResolved.unsignedTxs[0];
```

modules/sdk-core/test/unit/bitgo/utils/tss/recipientUtils.ts
```diff
@@ -98,7 +122,22 @@ describe('recipientUtils', function () {
...
 
-    const NO_RECIPIENT_TYPES = [
+    it('allows empty recipients when txParams.type is a no-recipient type', function () {
+      const { resolveEffectiveTxParams } = getModule();
+      const txRequest = makeTxRequest();
+      const result = resolveEffectiveTxParams(txRequest, { type: 'stakingActivate' });
+      result.type.should.equal('stakingActivate');
+    });
+
+    it('allows empty recipients when intent.intentType is a no-recipient type (EdDSA fallback)', function () {
```
pranishnepal, line 132: maybe we can add a test to also validate that the request is rejected when intentType is some unknown type? Or, is that already covered?

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsa.ts
mrdanish26: Each `signTxRequest` call handles a single transaction, the array structure exists in the type for historical/batch reasons, but in practice only index 0 is ever populated per signing round. The same pattern is used identically in the `ECDSA` path (ecdsa.ts and ecdsaMPCv2.ts)

modules/sdk-core/test/unit/bitgo/utils/tss/recipientUtils.ts
mrdanish26: This is already covered. The test `throws when txParams is undefined and intent has no recipients` uses intentType: 'payment' with no txParams. Since payment isn't in the exempt list, the guard throws. That's the rejection path for an unknown intent type.

pranishnepal (approved):

zahin-mohammad (commented):
in general looks good, but please also get a review from @mohammadalfaiyazbitgo

PR #8636, opened 2026-04-27
fix(statics): handle undefined coin.network

shubham-damkondwar:
CECHO-634

TICKET: CECHO-634

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -492,7 +492,7 @@ export function enrichBaseChainFeatures(
     Object.entries(coinTokenMap).map(([key, coins]) => [
       key,
       coins.map((coin) => {
-        if (coin.isToken || coin.additionalFeatures !== undefined) return coin;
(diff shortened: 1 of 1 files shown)


modules/statics/src/coins.ts
```diff
@@ -492,7 +492,7 @@ export function enrichBaseChainFeatures(
     Object.entries(coinTokenMap).map(([key, coins]) => [
       key,
       coins.map((coin) => {
-        if (coin.isToken || coin.additionalFeatures !== undefined) return coin;
+        if (coin.isToken || coin.additionalFeatures !== undefined || coin.network === undefined) return coin;
```
mukeshsp, line 495: Do we have any coins where network in **undefined** ?

shubham-damkondwar: For registering dynamic EVM coins from AMS in WP, we first fetch the network, I have kept network fetch  behind a flipt flag for testing. So there is just a minute possibility if this flag is off. Just adding the check for safety.

nvrakesh06 (approved):

PR #8658, opened 2026-04-30
fix(sdk-core): add EVM staking types to TSS signing bypass list

mrdanish26:
TICKET: WAL-756

## Summary

  **1. WAL-756 — BNB/BSC staking broken**
  Staking wallets call `signTransaction` without `txParams`. After WAL-375, `resolveEffectiveTxParams()` throws `InvalidTransactionError` because `effectiveTxParams.type` is undefined and staking intentType strings were not in `NO_RECIPIENT_TX_TYPES`.

  **2. XDC/BSC token enable/approve**
Coin-specific `verifyTssTransaction` overrides in `xdcToken.ts`, `bscToken.ts`, `xdc.ts`, `bsc.ts`, and `evmCoin.ts` had stale hardcoded bypass lists and never updated. When the parent class added `tokenApproval`, `enableToken`, `bridgeFunds`, etc., these overrides were left behind — causing `"missing txParams"` errors on app.bitgo-test.com when enabling tokens on XDC.

  ## Solution

  **Staking fix (`recipientUtils.ts`):**
  - Detect staking intents generically via `stakingRequestId` presence on
    the intent (a required field on `BaseStakeIntent` in `@bitgo/public-types`)
    rather than enumerating staking intentType strings that drift over time
  - Add `intent.intentType` fallback so `txType` propagates to downstream
    `verifyTssTransaction` callers when `txParams.type` is absent

  **Centralization (all coin modules):**
  - Export `NO_RECIPIENT_TX_TYPES` from sdk-core public API (`tss/index.ts`)
    as the single source of truth
  - Replace all stale inline bypass arrays with `NO_RECIPIENT_TX_TYPES.has()`
    in: `abstractEthLikeNewCoins.ts`, `bsc.ts`, `bscToken.ts`, `xdc.ts`,
    `xdcToken.ts`, `evmCoin.ts`
  - Existing `verifyTssTransaction` overrides preserved

  ## Testing

  - [ ] Unit tests updated 
  - [ ] `staking-bnb-hot` E2E test in bitgo-retail passes after SDK bump

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -43,6 +43,7 @@ import {
   verifyMPCWalletAddress,
   TssVerifyAddressOptions,
   isTssVerifyAddressOptions,
+  NO_RECIPIENT_TX_TYPES,
 } from '@bitgo/sdk-core';
 import { getDerivationPath } from '@bitgo/sdk-lib-mpc';
 import { bip32 } from '@bitgo/secp256k1';
@@ -3108,17 +3109,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
       !(
         txParams.prebuildTx?.consolidateId ||
         txPrebuild?.consolidateId ||
-        (txParams.type &&
-          [
-            'acceleration',
-            'fillNonce',
-            'transferToken',
-            'tokenApproval',
-            'consolidate',
-            'bridgeFunds',
-            'enableToken',
-            'customTx',
-          ].includes(txParams.type))
+        (txParams.type && NO_RECIPIENT_TX_TYPES.has(txParams.type))
       )
     ) {
       throw new Error('missing txParams');
--- a/modules/sdk-coin-bsc/src/bsc.ts
+++ b/modules/sdk-coin-bsc/src/bsc.ts
@@ -1,4 +1,12 @@
-import { BaseCoin, BitGoBase, common, MPCAlgorithm, MultisigType, multisigTypes } from '@bitgo/sdk-core';
+import {
+  BaseCoin,
+  BitGoBase,
+  common,
+  MPCAlgorithm,
+  MultisigType,
+  multisigTypes,
+  NO_RECIPIENT_TX_TYPES,
+} from '@bitgo/sdk-core';
 import { BaseCoin as StaticsBaseCoin, coins } from '@bitgo/statics';
 import {
   AbstractEthLikeNewCoins,
@@ -68,10 +76,7 @@ export class Bsc extends AbstractEthLikeNewCoins {
     const { txParams, txPrebuild, wallet } = params;
     if (
       !txParams?.recipients &&
-      !(
-        txParams.prebuildTx?.consolidateId ||
-        (txParams.type && ['acceleration', 'fillNonce', 'transferToken', 'tokenApproval'].includes(txParams.type))
-      )
+      !(txParams.prebuildTx?.consolidateId || (txParams.type && NO_RECIPIENT_TX_TYPES.has(txParams.type)))
     ) {
       throw new Error(`missing txParams`);
     }
--- a/modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
@@ -4,7 +4,8 @@ import { PopulatedIntent, TxRequest } from './baseTypes';
 
 /**
  * Transaction types that legitimately carry no explicit recipients.
- * verifyTransaction handles no-recipient validation for these internally.
+ * These are non-staking ECDSA types where verifyTransaction handles
+ * no-recipient validation internally.
  * Mirrors the bypass list in abstractEthLikeNewCoins.ts verifyTssTransaction.
  */
 export const NO_RECIPIENT_TX_TYPES = new Set([
@@ -15,7 +16,7 @@ export const NO_RECIPIENT_TX_TYPES = new Set([
   'consolidate',
   'bridgeFunds',
   'enableToken',
-  'customTx', // DeFi/WalletConnect smart contract interactions have no traditional recipients
+  'customTx',
 ]);
 
 /**
@@ -25,8 +26,13 @@ export const NO_RECIPIENT_TX_TYPES = new Set([
  * (native amount = 0, so buildParams is empty). Falls back to intent recipients
  * mapped to ITransactionRecipient shape when txParams.recipients is absent.
  *
+ * Staking intents (BSC delegate/undelegate, CELO stake/unstake, etc.) are
+ * identified generically by the presence of `stakingRequestId` on the intent —
+ * a required field on BaseStakeIntent in @bitgo/public-types. These intents
+ * have no txParams recipients by design; validation is done at the coin layer.
+ *
  * Throws InvalidTransactionError if no recipients can be resolved and the
- * transaction type is not a known no-recipient type.
+ * transaction is not a known no-recipient type.
  */
 export function resolveEffectiveTxParams(
   txRequest: TxRequest,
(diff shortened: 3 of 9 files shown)


modules/sdk-coin-bsc/src/bsc.ts
```diff
@@ -54,34 +50,4 @@ export class Bsc extends AbstractEthLikeNewCoins {
     const explorerUrl = common.Environments[this.bitgo.getEnv()].bscscanBaseUrl;
     return await recoveryBlockchainExplorerQuery(query, explorerUrl as string, apiToken);
   }
-
```
zahin-mohammad, line 57: Context for this removal?

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -3118,6 +3118,14 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
             'enableToken',
             'customTx',
+            // EVM staking (BSC/BNB, ETH, CELO) — mirrors NO_RECIPIENT_TX_TYPES in recipientUtils.ts
+            'stakingLock',
+            'stakingVote',
+            'stakingActivate',
+            'stakingUnvote',
+            'stakingUnlock',
+            'stakingWithdraw',
+            'stakingDeactivate',
```
zahin-mohammad: Not great that we are duplicating this same list in multiple areas/modules. Can we de-duplicate as a follow up?

modules/sdk-coin-bsc/src/bsc.ts
mrdanish26: The original bsc.ts had its own `verifyTssTransaction` override that duplicated the parent's logic but only knew about 3 types:
// OLD bsc.ts — only 3/15 types: ['acceleration', 'fillNonce', 'transferToken'].includes(txParams.type)

This override shadowed the parent (AbstractEthLikeNewCoins), so even though the parent now has the correct 15-type list, BSC never saw it,  the subclass's override ran instead.

Removing it means BSC falls through to the parent, which has the complete correct list. It's the same pattern 21 of the other 23 EVM coins already use — no override at all.

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
mrdanish26: Will fix this in a follow up

zahin-mohammad (dismissed):

lokesh-bitgo:
@claude Please do a thorough review of this PR.

Focus especially on:
- any bugs or incorrect logic in the changes
- whether these changes could impact existing flows or cause regressions
- whether the implementation is aligned with the requirement/design
- whether the changes are correct and complete
- missing edge cases, validation gaps, and test coverage gaps

Please call out concrete issues clearly and separate them by severity if possible.

ashutoshkumar-6 (commented):
Overall the direction of centralizing `NO_RECIPIENT_TX_TYPES` and the staking detection via `stakingRequestId` is good. However, there is a critical gap: the fix in `resolveEffectiveTxParams` prevents `InvalidTransactionError` from throwing, but the downstream `verifyTssTransaction` guards in **every affected coin** (`bsc.ts`, `bscToken.ts`, `xdc.ts`, `xdcToken.ts`, `evmCoin.ts`, and the parent `abstractEthLikeNewCoins.ts`) will still throw `'missing txParams'` for staking types — see inline comments for details.

modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
```diff
@@ -43,7 +49,21 @@ export function resolveEffectiveTxParams(
...
+  if (!effectiveTxParams.type && txType) {
+    effectiveTxParams.type = txType;
+  }
+
+  // All staking intents (BSC delegate/undelegate, CELO stake/unstake, etc.) carry
+  // stakingRequestId as a required field on BaseStakeIntent (@bitgo/public-types).
+  // Use its presence as a generic staking signal — no need to enumerate every intentType.
+  const isStakingIntent = !!(txRequest.intent as any)?.stakingRequestId;
+
+  if (!effectiveTxParams.recipients?.length && !isStakingIntent && !NO_RECIPIENT_TX_TYPES.has(txType)) {
```
ashutoshkumar-6, line 66: **[Critical]** This fix prevents `InvalidTransactionError` here, but the staking bypass does **not** propagate to `verifyTssTransaction`.

The full call chain after this function returns for a staking intent (`stakingRequestId` present, no recipients) is:
```
resolveEffectiveTxParams returns { type: 'delegate' }  ← fixed here
  → baseCoin.verifyTransaction({ txParams: { type: 'delegate' } })
  → verifyTssTransaction: !txParams?.recipients && !NO_RECIPIENT_TX_TYPES.has('delegate')
  → throws 'missing txParams'  ← NOT fixed
```

`'delegate'`, `'undelegate'`, `'stake'`, `'unstake'` are **not** in `NO_RECIPIENT_TX_TYPES`, so every `verifyTssTransaction` override (`bsc.ts:79`, `bscToken.ts:57`, `xdc.ts:79`, `xdcToken.ts:81`, `evmCoin.ts:119`, `abstractEthLikeNewCoins.ts:3112`) will still throw for BSC staking.

The fix moves the error from `InvalidTransactionError` (here) to `Error('missing txParams')` (in `verifyTssTransaction`) but does not eliminate it. The staking flow remains broken end-to-end.

modules/sdk-coin-bsc/src/bsc.ts
```diff
@@ -68,10 +76,7 @@ export class Bsc extends AbstractEthLikeNewCoins {
     const { txParams, txPrebuild, wallet } = params;
     if (
       !txParams?.recipients &&
-      !(
-        txParams.prebuildTx?.consolidateId ||
-        (txParams.type && ['acceleration', 'fillNonce', 'transferToken', 'tokenApproval'].includes(txParams.type))
-      )
+      !(txParams.prebuildTx?.consolidateId || (txParams.type && NO_RECIPIENT_TX_TYPES.has(txParams.type)))
```
ashutoshkumar-6, line 79: **[Critical]** This guard still throws for BSC staking transactions.

After `resolveEffectiveTxParams` returns `effectiveTxParams = { type: 'delegate' }` (no recipients), this evaluates as:
- `!txParams?.recipients` → `true`
- `txParams.prebuildTx?.consolidateId` → `undefined` (no prebuildTx on staking tx)
- `NO_RECIPIENT_TX_TYPES.has('delegate')` → `false`
- Result: **throws `'missing txParams'`**

The staking types propagated via `intent.intentType` are not in `NO_RECIPIENT_TX_TYPES`. Same issue exists in `bscToken.ts`, `xdc.ts`, `xdcToken.ts`, and `evmCoin.ts:119`.

Possible fixes:
1. Check for a staking signal here too (e.g. set a flag on `effectiveTxParams` like `effectiveTxParams.isStakingIntent = true` in `resolveEffectiveTxParams`, then check it here), or
2. Pass staking types through `NO_RECIPIENT_TX_TYPES` with a comment, or
3. Remove this override entirely and let the parent handle it (as the commit message `467ff30d` says was intended — but the override is still present in the PR)

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -3108,17 +3109,7 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
...
-            'acceleration',
-            'fillNonce',
-            'transferToken',
-            'tokenApproval',
-            'consolidate',
-            'bridgeFunds',
-            'enableToken',
-            'customTx',
-          ].includes(txParams.type))
+        (txParams.type && NO_RECIPIENT_TX_TYPES.has(txParams.type))
```
ashutoshkumar-6, line 3112: **[Critical]** Same issue as in `bsc.ts`: `NO_RECIPIENT_TX_TYPES` does not contain staking types (`delegate`, `undelegate`, `stake`, `unstake`, etc.). When a staking transaction arrives here with `txParams = { type: 'delegate' }` (returned by `resolveEffectiveTxParams`), this check throws `'missing txParams'`.

Note that this parent class is also what `bsc.ts` would fall through to if the override were removed (as the commit message for `467ff30d` states). Either way, the staking bypass needs to be present at this layer too.

modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
```diff
@@ -43,7 +49,21 @@ export function resolveEffectiveTxParams(
...
+
+  // Propagate the resolved type so downstream callers (e.g. verifyTssTransaction) can use it.
+  if (!effectiveTxParams.type && txType) {
+    effectiveTxParams.type = txType;
+  }
+
+  // All staking intents (BSC delegate/undelegate, CELO stake/unstake, etc.) carry
+  // stakingRequestId as a required field on BaseStakeIntent (@bitgo/public-types).
+  // Use its presence as a generic staking signal — no need to enumerate every intentType.
+  const isStakingIntent = !!(txRequest.intent as any)?.stakingRequestId;
```
ashutoshkumar-6, line 64: **[High]** `as any` cast is unsafe. `PopulatedIntent` in `baseTypes.ts` does not define `stakingRequestId`, so TypeScript cannot verify this field exists at compile time. If `BaseStakeIntent` in `@bitgo/public-types` renames or removes `stakingRequestId`, this silently returns `false` at runtime with no compile-time error.

Consider importing `BaseStakeIntent` from `@bitgo/public-types` and writing a proper type guard:
```typescript
function isBaseStakeIntent(intent: unknown): intent is BaseStakeIntent {
  return typeof (intent as any)?.stakingRequestId === 'string';
}
const isStakingIntent = isBaseStakeIntent(txRequest.intent);
```
Or extend the `PopulatedIntent` union in `baseTypes.ts` to include a staking variant.

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
mohammadalfaiyazbitgo: @mrdanish26 let's also add the `stakingRequestId`, it should be in the `prebuildTx` object. @noel-bitgo if you can confirm.

mohammadalfaiyazbitgo (commented):
please address conflicts

lokesh-bitgo (commented):
> please address conflicts

+1

PR #8666, opened 2026-04-30
feat(sdk-core): added OFC BitGo signing on trading accounts object

alextse-bg:
TICKET: WCN-217

### DIFF
--- a/modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
+++ b/modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
@@ -1,8 +1,19 @@
 import { ITradingNetwork } from './network';
 
+/**
+ * Parameters for the signing a payload from the trading account.
+ * If both walletPassphrase and prv are not provided, the BitGo key will be used.
+ *
+ * @note If wallet has userKeySigningRequired set to true, then attempting to sign with BitGo key will throw.
+ *
+ * @param payload - The payload to sign
+ * @param walletPassphrase - The passphrase of the wallet that will be used to decrypt the user key and sign the payload.
+ * @param prv - The decrypted user key prv used to sign the payload
+ */
 export interface SignPayloadParameters {
   payload: string | Record<string, unknown>;
-  walletPassphrase: string;
+  walletPassphrase?: string;
+  prv?: string;
 }
 
 export interface ITradingAccount {
--- a/modules/sdk-core/src/bitgo/trading/network/network.ts
+++ b/modules/sdk-core/src/bitgo/trading/network/network.ts
@@ -109,7 +109,7 @@ export class TradingNetwork implements ITradingNetwork {
 
   /**
    * Prepare an allocation for submission
-   * @param {string} walletPassphrase ofc wallet passphrase
+   * @param {string} walletPassphrase ofc wallet passphrase - required only when signing via user key
    * @param {string} connectionId connection to whom to make the allocation or deallocation
    * @param {string=} clientExternalId one time generated uuid v4
    * @param {string} currency currency for which the allocation should be made. e.g. btc / tbtc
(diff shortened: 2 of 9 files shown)


alextse-bg:
reimplements the reverted PR: https://github.com/BitGo/BitGoJS/pull/8624

Taken out the changes to ofcToken/ofc.ts. Changes to these files has no impact on current code path (since we are signing with the user key right now, `prv` is always passed in). Removing those changes to keep things simple. If there is a need to make the SDK more flexible we can re-introduce the changes.

alextse-bg:
@claude

modules/sdk-core/src/bitgo/trading/network/network.ts
```diff
@@ -130,10 +130,7 @@ export class TradingNetwork implements ITradingNetwork {
     }
 
     const payload = JSON.stringify(body);
-
-    const prv = await this.wallet.getPrv({ walletPassphrase });
-    const signedBuffer: Buffer = await this.wallet.baseCoin.signMessage({ prv }, payload);
-    const signature = signedBuffer.toString('hex');
+    const signature = await this.wallet.toTradingAccount().signPayload({ payload, walletPassphrase });
```
mrdanish26, line 133: Should `prepareAllocation` fail fast when walletPassphrase / prv is missing but the wallet requires user signing (if that flag is knowable without extra round trips), instead of building the payload and failing at sign time?

modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
```diff
@@ -1,8 +1,16 @@
 import { ITradingNetwork } from './network';
 
+/**
+ * Parameters for the signing a payload from the trading account.
+ * If both walletPassphrase and prv is not provided the BitGo key will be used
```
zahin-mohammad: We should note that this is only valid when they have the wallet setting enabled.

modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
```diff
@@ -1,8 +1,16 @@
...
+ * @param payload - The payload to sign
+ * @param walletPassphrase - The passphrase of the wallet that will be used to decrypt the user key and sign the payload.
+ * @param prv - The decrypted user key prv used to sign the payload
+ */
 export interface SignPayloadParameters {
   payload: string | Record<string, unknown>;
-  walletPassphrase: string;
+  walletPassphrase?: string;
+  prv?: string;
```
zahin-mohammad, line 16: why is this needed?

modules/sdk-core/src/bitgo/trading/tradingAccount.ts
```diff
@@ -23,18 +23,72 @@ export class TradingAccount implements ITradingAccount {
...
+      throw new Error(
+        'Must provide either prv or walletPassphrase to sign payload using user key. Please provide the wallet passphrase or visit your wallet settings page to configure one.'
+      );
+    }
+
+    let prv: string;
+    if (params.prv) {
+      prv = params.prv;
+    } else {
+      const key = (await this.wallet.baseCoin.keychains().get({ id: this.wallet.keyIds()[0] })) as any;
```
zahin-mohammad: why are we casting as any here?

zahin-mohammad (commented):
In general lgtm, but please address the `as any`

modules/sdk-core/src/bitgo/trading/iTradingAccount.ts
alextse-bg: that was what caused the original bug, with the wallets and ofcToken changes, but now I removed it this is no longer need.

modules/sdk-core/src/bitgo/trading/tradingAccount.ts
alextse-bg: not sure why, this is old code. Fixed it.

modules/sdk-core/src/bitgo/trading/network/network.ts
alextse-bg: build time shouldn't cost too long since the only costly operations here is stringify. I can add a fail fast if it is needed.

mrdanish26 (approved):

zahin-mohammad (approved):

alextse-bg:
@claude

PR #8677, opened 2026-05-04
fix(passkey-crypto): change deriveEnterpriseSalt output encoding from base64 to base64url

derranW26:
Ticket: WCN-410

### DIFF
--- a/modules/passkey-crypto/src/deriveEnterpriseSalt.ts
+++ b/modules/passkey-crypto/src/deriveEnterpriseSalt.ts
@@ -1,4 +1,4 @@
-import { createHmac } from 'crypto';
+import * as sjcl from '@bitgo/sjcl';
 
 /**
  * Derives an enterprise-scoped PRF salt to prevent cross-enterprise key reuse.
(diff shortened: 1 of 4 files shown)


modules/passkey-crypto/src/deriveEnterpriseSalt.ts
```diff
@@ -26,5 +26,5 @@ export function deriveEnterpriseSalt(baseSalt: string, enterpriseId: string): st
   const hmacInstance = new misc.hmac(keyBits, hash.sha256);
   const resultBits = hmacInstance.mac(dataBits);
 
-  return codec.base64.fromBits(resultBits);
+  return codec.base64url.fromBits(resultBits);
```
mrdanish26: Any code that decodes this value with a standard base64 decoder will get wrong bytes or fail and it must use base64url decoding now. Have we checked if this affects any callers?

derranW26: This is for the passkey-crypto package so I have checked for all callers -> we will continue this after brandon's changes get merged (he is apart of the shift to use base64url)

PR #8687, opened 2026-05-05
feat(sdk-core): add EdDSA MPCv2 DSG helpers and DKG key-share util

Marzooqa:
- Add getSignatureShareRoundOne/Two/Three helpers in sdk-core for building
  PGP-signed MPS broadcast messages per signing round
- Add verifyBitGoMessageRoundOne/Two helpers for verifying peer PGP signatures
  and deserialising incoming MPS messages
- Parameterise partyId (default: USER=0) and peerPartyId (type: MPCv2PartiesEnum,
  default: BITGO=2) to support non-user signers without hardcoding
- Wire eddsaMpcV2 type in sendSignatureShareV2 (common.ts)
- Add generateEdDsaDKGKeyShares to sdk-lib-mpc MPSUtil with split EdDsaDKGPartySeed
  (encKey / dkgSeed) to eliminate seed dual-use; add length assertions
- Replace duplicate test-util copy with a re-export from the production source
- Add unit tests covering all helpers including BACKUP party path, tampered-message
  rejection for both round-1 and round-2 verify, and deterministic-seed behaviour

TICKET: WCI-153

### DIFF
--- a/modules/sdk-core/src/bitgo/tss/eddsa/eddsaMPCv2.ts
+++ b/modules/sdk-core/src/bitgo/tss/eddsa/eddsaMPCv2.ts
@@ -1,0 +1,126 @@
+import assert from 'assert';
+import * as openpgp from 'openpgp';
+import { MPSComms, MPSTypes } from '@bitgo/sdk-lib-mpc';
+import {
+  EddsaMPCv2SignatureShareRound1Input,
+  EddsaMPCv2SignatureShareRound1Output,
+  EddsaMPCv2SignatureShareRound2Input,
+  EddsaMPCv2SignatureShareRound2Output,
+  EddsaMPCv2SignatureShareRound3Input,
+} from '@bitgo/public-types';
+import { SignatureShareRecord, SignatureShareType } from '../../utils/tss/baseTypes';
+import { MPCv2PartiesEnum } from '../../utils/tss/ecdsa/typesMPCv2';
+
+function partyIdToSignatureShareType(partyId: 0 | 1 | 2): SignatureShareType {
+  assert(partyId === 0 || partyId === 1 || partyId === 2, 'Invalid partyId for EdDSA MPCv2 signing');
+  switch (partyId) {
+    case 0:
+      return SignatureShareType.USER;
+    case 1:
+      return SignatureShareType.BACKUP;
+    case 2:
+      return SignatureShareType.BITGO;
+  }
+}
+
+/**
+ * Builds the round-1 signature share record.
+ *
+ * PGP-signs the WASM round-0 broadcast message with the signer's ephemeral key and
+ * wraps it into a SignatureShareRecord ready for `sendSignatureShareV2`.
+ */
+export async function getSignatureShareRoundOne(
+  userMsg1: MPSTypes.DeserializedMessage,
+  userGpgPrivKey: openpgp.PrivateKey,
+  partyId: 0 | 1 = 0,
+  otherSignerPartyId: 0 | 1 | 2 = 2
+): Promise<SignatureShareRecord> {
+  const signedMsg1 = await MPSComms.detachSignMpsMessage(Buffer.from(userMsg1.payload), userGpgPrivKey);
+  const share: EddsaMPCv2SignatureShareRound1Input = {
+    type: 'round1Input',
+    data: { msg1: signedMsg1 },
+  };
+  return {
+    from: partyIdToSignatureShareType(partyId),
+    to: partyIdToSignatureShareType(otherSignerPartyId),
+    share: JSON.stringify(share),
+  };
+}
+
+/**
+ * Verifies the peer's round-1 PGP signature and returns the raw deserialized
+ * message ready for `DSG.handleIncomingMessages`.
+ */
+export async function verifyBitGoMessageRoundOne(
+  parsedRound1Output: EddsaMPCv2SignatureShareRound1Output,
+  peerGpgKey: openpgp.Key,
+  peerPartyId: MPCv2PartiesEnum = MPCv2PartiesEnum.BITGO
+): Promise<MPSTypes.DeserializedMessage> {
+  const rawBytes = await MPSComms.verifyMpsMessage(parsedRound1Output.data.msg1, peerGpgKey);
+  return {
+    from: peerPartyId,
+    payload: new Uint8Array(rawBytes),
+  };
+}
+
+/**
+ * Builds the round-2 signature share record.
+ */
+export async function getSignatureShareRoundTwo(
+  userMsg2: MPSTypes.DeserializedMessage,
+  userGpgPrivKey: openpgp.PrivateKey,
+  partyId: 0 | 1 = 0,
+  otherSignerPartyId: 0 | 1 | 2 = 2
+): Promise<SignatureShareRecord> {
+  const signedMsg2 = await MPSComms.detachSignMpsMessage(Buffer.from(userMsg2.payload), userGpgPrivKey);
+  const share: EddsaMPCv2SignatureShareRound2Input = {
+    type: 'round2Input',
+    data: { msg2: signedMsg2 },
+  };
+  return {
+    from: partyIdToSignatureShareType(partyId),
+    to: partyIdToSignatureShareType(otherSignerPartyId),
+    share: JSON.stringify(share),
+  };
+}
+
+/**
+ * Verifies the peer's round-2 PGP signature and returns the raw deserialized
+ * message ready for `DSG.handleIncomingMessages`.
+ */
+export async function verifyBitGoMessageRoundTwo(
+  parsedRound2Output: EddsaMPCv2SignatureShareRound2Output,
+  peerGpgKey: openpgp.Key,
+  peerPartyId: MPCv2PartiesEnum = MPCv2PartiesEnum.BITGO
+): Promise<MPSTypes.DeserializedMessage> {
+  const rawBytes = await MPSComms.verifyMpsMessage(parsedRound2Output.data.msg2, peerGpgKey);
+  return {
+    from: peerPartyId,
+    payload: new Uint8Array(rawBytes),
+  };
+}
+
+/**
+ * Builds the round-3 signature share record (final signer message).
+ *
+ * There is no corresponding `verifyBitGoMessageRoundThree` because Wallet Platform
+ * finalises the signing server-side after receiving round 3; the client obtains the
+ * signed transaction via `sendTxRequest`.
+ */
+export async function getSignatureShareRoundThree(
+  userMsg3: MPSTypes.DeserializedMessage,
+  userGpgPrivKey: openpgp.PrivateKey,
+  partyId: 0 | 1 = 0,
+  otherSignerPartyId: 0 | 1 | 2 = 2
+): Promise<SignatureShareRecord> {
+  const signedMsg3 = await MPSComms.detachSignMpsMessage(Buffer.from(userMsg3.payload), userGpgPrivKey);
+  const share: EddsaMPCv2SignatureShareRound3Input = {
+    type: 'round3Input',
+    data: { msg3: signedMsg3 },
+  };
+  return {
+    from: partyIdToSignatureShareType(partyId),
+    to: partyIdToSignatureShareType(otherSignerPartyId),
+    share: JSON.stringify(share),
+  };
+}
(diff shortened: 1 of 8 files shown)


modules/sdk-lib-mpc/src/tss/eddsa-mps/util.ts
```diff
@@ -7,3 +11,51 @@ export function concatBytes(chunks: Uint8Array[]): Uint8Array {
   const buffers = chunks.map((chunk) => Buffer.from(chunk));
   return new Uint8Array(Buffer.concat(buffers));
 }
+
+function generateX25519Keypair(seed?: Buffer): { privKey: Buffer; pubKey: Buffer } {
+  const privKey = seed ? seed.subarray(0, 32) : crypto.randomBytes(32);
+  const pubKey = Buffer.from(x25519.getPublicKey(privKey));
+  return { privKey: Buffer.from(privKey), pubKey };
+}
```
zahin-mohammad, line 20: **Seed dual-use — please address before merge.** The same `seed` bytes are used here as the X25519 private key and below at L45-47 as the DKG round-0 seed (`getFirstMessage(seedX)`). The original test-util version of this function had a comment that explicitly flagged this ("Seeds are used as X25519 private keys AND as DKG round0 seeds for full determinism"); promoting to production source dropped the warning.

Since `MPSUtil.generateEdDsaDKGKeyShares` is exported via the package barrel, this is now public API. Either restore the explicit "AND" comment so callers know what they're opting into, or split inputs (e.g. accept `encKey`/`dkgSeed` per party, or HKDF-derive both from one seed). I'd lean toward splitting — once callers depend on the current deterministic mapping, changing it is a breaking change.

Also, no length check: `seed.subarray(0, 32)` on a < 32-byte buffer produces a shorter view, then `x25519.getPublicKey` throws `invalid scalar, expected 32 or 32 bytes, got N`. A friendlier `assert(!seed || seed.length >= 32, ...)` up-front would surface the error at the right layer.

modules/sdk-lib-mpc/src/tss/eddsa-mps/util.ts
```diff
@@ -7,3 +11,51 @@ export function concatBytes(chunks: Uint8Array[]): Uint8Array {
...
+
+/**
+ * Generates EdDSA MPS DKG key shares for all 3 parties (user, backup, BitGo).
+ * Optional seeds produce deterministic output; omit for random key material.
+ *
+ * Mirrors `DklsUtils.generateDKGKeyShares` for ECDSA DKLS.
+ */
+export async function generateEdDsaDKGKeyShares(
```
zahin-mohammad, line 38: **Duplicate of the existing test helper.** This function body is near-verbatim from `modules/sdk-lib-mpc/test/unit/tss/eddsa/util.ts` (only comments changed). The pre-existing `test/unit/tss/eddsa/dkg.ts:5` and `dsg.ts:4` still `import { generateEdDsaDKGKeyShares } from './util'`, so they keep using the test-util copy — the two implementations will silently diverge.

Suggested fix (internal-only, follow-up safe): replace the body of the test util with a re-export so existing tests still resolve via `./util`:

```ts
export { generateEdDsaDKGKeyShares } from '../../../src/tss/eddsa-mps/util';
```

Leave `runEdDsaDSG` in the test util — it's test-only orchestration.

modules/sdk-core/src/bitgo/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+}
+
+/**
+ * Builds the round-1 signature share record.
+ *
+ * PGP-signs the WASM round-0 broadcast message with the signer's ephemeral key and
+ * wraps it into a SignatureShareRecord ready for `sendSignatureShareV2`.
+ */
+export async function getEddsaSignatureShareRound1(
```
zahin-mohammad: **Naming inconsistent with ECDSA equivalents.** Compare `modules/sdk-core/src/bitgo/tss/ecdsa/ecdsaMPCv2.ts:30` — `getSignatureShareRoundOne/Two/Three` (English numerals, no `Ecdsa` prefix). Here we have `getEddsaSignatureShareRound1/2/3` (digits + redundant prefix since the file already lives under `tss/eddsa/`). Same for `verifyBitGoEddsaMessageRound1/2` vs ECDSA's `verifyBitGoMessagesAndSignaturesRoundOne/Two`.

Follow-up-safe today (these aren't barrel-exported), but #8697 adds deep-import callers — easier to pick names now than to ripple a rename later. Suggest `getSignatureShareRoundOne/Two/Three` + `verifyBitGoMessageRoundOne/Two`.

modules/sdk-core/src/bitgo/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,126 @@
...
+  };
+}
+
+/**
+ * Verifies the peer's round-1 PGP signature and returns the raw deserialized
+ * message ready for `DSG.handleIncomingMessages`.
+ */
+export async function verifyBitGoEddsaMessageRound1(
+  parsedRound1Output: EddsaMPCv2SignatureShareRound1Output,
+  bitgoGpgKey: openpgp.Key,
```
zahin-mohammad: Nit: parameter is named `bitgoGpgKey` but the function accepts any `peerPartyId: 0 | 1 | 2` — when called with USER or BACKUP as the peer, the name lies. `peerGpgKey` would match the contract. Same comment applies to `verifyBitGoEddsaMessageRound2` at L93. Follow-up safe.

modules/sdk-core/src/bitgo/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,126 @@
...
+  peerPartyId: 0 | 1 | 2 = 2
+): Promise<MPSTypes.DeserializedMessage> {
+  const rawBytes = await MPSComms.verifyMpsMessage(parsedRound1Output.data.msg1, bitgoGpgKey);
+  return {
+    from: peerPartyId as MPCv2PartiesEnum,
```
zahin-mohammad: Nit: `peerPartyId as MPCv2PartiesEnum` is an unsafe cast that's only needed because the parameter type is the literal union `0 | 1 | 2`. Typing the parameter as `MPCv2PartiesEnum` directly removes the cast and makes the contract clearer at the call site. Follow-up safe.

modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,206 @@
...
+    const bitgoSignedMsg1 = await MPSComms.detachSignMpsMessage(Buffer.from(bitgoMsg1.payload), bitgoGpgPrivKey);
+    const [bitgoMsg2] = bitgoDsg.handleIncomingMessages([bitgoMsg1, userMsg1]);
+    const bitgoSignedMsg2 = await MPSComms.detachSignMpsMessage(Buffer.from(bitgoMsg2.payload), bitgoGpgPrivKey);
+
+    const round2Output: EddsaMPCv2SignatureShareRound2Output = {
+      type: 'round2Output',
+      data: { msg2: bitgoSignedMsg2 },
+    };
+
+    void bitgoSignedMsg1;
```
zahin-mohammad: Nit: looks like leftover dead code from a refactor — `bitgoSignedMsg1` is computed above and then explicitly discarded with `void`. Safe to remove both the assignment and this line. Follow-up safe.

modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,206 @@
...
+    const parsed = decodeWithCodec(
+      EddsaMPCv2SignatureShareRound3Input,
+      JSON.parse(share.share),
+      'EddsaMPCv2SignatureShareRound3Input'
+    );
+    assert.strictEqual(parsed.type, 'round3Input');
+    assert.ok(parsed.data.msg3.message, 'msg3.message should be set');
+    assert.ok(parsed.data.msg3.signature, 'msg3.signature should be set');
+  });
+});
```
zahin-mohammad, line 325: Test coverage gaps worth filing as follow-ups (none blocking):
- Round-2 verify has no tampered-signature negative test (round 1 has one at L94-108)
- BACKUP path (`partyId=1`) of `getEddsaSignatureShareRound{1,2,3}` is never exercised
- Deterministic-seed branch of `generateEdDsaDKGKeyShares` is untested — the test calls it without seeds, so that code path isn't verified
- No direct assertion that all 3 DKG parties agree on the public key (currently only used as setup)

zahin-mohammad (commented):
Built and tested locally — `tsc --noEmit` is clean in both `sdk-lib-mpc` and `sdk-core`, and the new sdk-core EdDSA MPCv2 test suite passes (6/6). Confirmed #8697 is the follow-up that wires `EddsaMPCv2Utils.signRequestBase` into these helpers.

One semantics concern that I think is worth resolving before merge because it freezes on the public surface — `MPSUtil.generateEdDsaDKGKeyShares` is barrel-exported, so the seed contract becomes part of the SDK API. The rest are follow-up-safe: helper renames (still internal until #8697 lands deep-import callers), test-util consolidation (pure refactor), and a few test/cleanup nits.

See inline comments.

zahin-mohammad (approved):
LGTM — all 7 prior comments are resolved in this revision:

- ✅ Seed dual-use split into `EdDsaDKGPartySeed { encKey, dkgSeed }` with `>= 32 byte` asserts (the only blocking item, since `MPSUtil.generateEdDsaDKGKeyShares` is barrel-exported)
- ✅ Test-util duplicate replaced with a re-export from the production source
- ✅ Helpers renamed to match ECDSA convention (`getSignatureShareRoundOne/Two/Three`, `verifyBitGoMessageRoundOne/Two`)
- ✅ `bitgoGpgKey` → `peerGpgKey`; `peerPartyId` typed as `MPCv2PartiesEnum` directly (no more `as` cast)
- ✅ Dead `void bitgoSignedMsg1` line removed
- ✅ Test coverage filled: round-2 tampered-signature, BACKUP path for all three rounds, deterministic split-seed test with 3-party public-key agreement assertion

### Residual nits — non-blocking, address in a follow-up

1. **`verifyBitGoMessage*` name vs parameterized peer.** The functions accept `peerPartyId: MPCv2PartiesEnum`, but the name still says `BitGo`. Default of `BITGO` keeps it mostly accurate, but if a USER/BACKUP peer ever flows through, the name lies. Consider `verifyPeerMessageRoundOne/Two` if you want the public surface to match the contract.
2. **Type inconsistency in `eddsaMPCv2.ts`.** Signing helpers use literal `partyId: 0 | 1` and `otherSignerPartyId: 0 | 1 | 2`, while verify helpers use `MPCv2PartiesEnum`. Unifying on `MPCv2PartiesEnum` everywhere — optionally with a `SignerPartyId = USER | BACKUP` alias to enforce non-BITGO origination — would make the contract uniform across the file.
3. **Redundant runtime assert in `partyIdToSignatureShareType`.** The literal `0 | 1 | 2` parameter type already prevents invalid values at compile time, so the `assert` is dead defensive code. Harmless; remove if you want to keep the file tight.

None of these block merge — happy to approve as-is.

modules/sdk-core/src/bitgo/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+ * Verifies the peer's round-1 PGP signature and returns the raw deserialized
+ * message ready for `DSG.handleIncomingMessages`.
+ */
+export async function verifyBitGoMessageRoundOne(
+  parsedRound1Output: EddsaMPCv2SignatureShareRound1Output,
+  peerGpgKey: openpgp.Key,
+  peerPartyId: MPCv2PartiesEnum = MPCv2PartiesEnum.BITGO
```
pranavjain97, line 57: function is named `verifyBitGoMessageRound*` but accepts a `peerPartyId` param -- should this be `verifyPeerMessageRound*`, or is it BitGo-only and the param should be removed?

modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -0,0 +1,325 @@
...
+  getSignatureShareRoundTwo,
+  getSignatureShareRoundThree,
+  verifyBitGoMessageRoundOne,
+  verifyBitGoMessageRoundTwo,
+} from '../../../../../../src/bitgo/tss/eddsa/eddsaMPCv2';
+import { decodeWithCodec } from '../../../../../../src/bitgo/utils/codecs';
+import { generateGPGKeyPair } from '../../../../../../src/bitgo/utils/opengpgUtils';
+import { MPCv2PartiesEnum } from '../../../../../../src/bitgo/utils/tss/ecdsa/typesMPCv2';
+
+describe('EdDSA MPS DSG helper functions', async () => {
```
pranavjain97, line 23: mocha `describe` callbacks shouldn't be `async` -- can cause tests to silently not run in some versions

PR #8697, opened 2026-05-05
feat(sdk-core): add EdDSA MPCv2 full 3-round signing orchestration

Marzooqa:
- Add signTxRequest, signTxRequestForMessage, and signRequestBase to EddsaMPCv2Utils implementing WASM round 0 + API rounds 1 and 2
- Verify PGP signatures on BitGo round1Output and round2Output
- Return latestTxRequest after round 2 (round 3 + send deferred to next PR)
- Add signTxRequest unit tests using nock with real BitGo DSG sessions

TICKET: WCI-154

### DIFF
--- a/modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/signTxRequest.ts
+++ b/modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/signTxRequest.ts
@@ -1,0 +1,604 @@
+import {
+  BaseCoin,
+  BitgoGPGPublicKey,
+  common,
+  ECDSAUtils,
+  EDDSAUtils,
+  RequestTracer,
+  RequestType,
+  SignatureShareRecord,
+  SignatureShareType,
+  TxRequest,
+  Wallet,
+} from '@bitgo/sdk-core';
+import { EddsaMPSDsg, MPSComms, MPSTypes, MPSUtil } from '@bitgo/sdk-lib-mpc';
+import * as openpgp from 'openpgp';
+import nock = require('nock');
+import { TestableBG, TestBitGo } from '@bitgo/sdk-test';
+import {
+  EddsaMPCv2SignatureShareRound1Input,
+  EddsaMPCv2SignatureShareRound1Output,
+  EddsaMPCv2SignatureShareRound2Input,
+  EddsaMPCv2SignatureShareRound2Output,
+  EddsaMPCv2SignatureShareRound3Input,
+} from '@bitgo/public-types';
+import { BitGo } from '../../../../../../src';
+
+const MPCv2PartiesEnum = ECDSAUtils.MPCv2PartiesEnum;
+
+interface SignatureShareApiBody {
+  signatureShares: SignatureShareRecord[];
+  signerGpgPublicKey: string;
+}
+
+describe('signTxRequest:', function () {
+  let tssUtils: EDDSAUtils.EddsaMPCv2Utils;
+  let wallet: Wallet;
+  let bitgo: TestableBG & BitGo;
+  let baseCoin: BaseCoin;
+  let bitgoGpgKey: openpgp.SerializedKeyPair<string> & { revocationCertificate: string };
+  let bitgoPrvKeyObj: openpgp.PrivateKey;
+  const coinName = 'sol';
+
+  const reqId = new RequestTracer();
+  const txRequestId = 'randomTxReqId';
+  const signableHex =
+    '02010206c2d5b5f4fb9a9bcd8a2f303e4d06f78d8ded300713f456da2abff0b3ea0185aa051a34bc8acd438763976f96876115050f73828553566d111d7ac8bffebf587c4f5f5987bfe26aa66013efd96d36360f2b4336c91f993259fb56051305614d42f2ea13f8ff9d7958dbf269c6e36bfdf5cb5c43de4b4e1d3efb7dab3d5d028604000000000000000000000000000000000000000000000000000000000000000006a7d517192c568ee08a845f73d29788cf035c3145b21ab344d8062ea94000003a621f6d1cc4b8fb2a739aa08e4034da0fc588ece3bd857630de30f7edde45dd0204030205010404000000040200030c02000000f0a29a3b00000000';
+  const serializedTxHex = `02000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003bc9df0b397bec2ed3b6444a8c33f38267cc08b5fb2a7d37e26b6c487e26d15b7c07830eb78e26a88db5de4aa6986a327f09aed8c01533e5b972748ddf60b80f${signableHex}`;
+  const messageRaw = 'TOO MANY SECRETS';
+  const messageEncoded = Buffer.from(messageRaw).toString('hex');
+  const txParams = {
+    recipients: [
+      {
+        address: 'HMEgbR4S2hLKfst2VZUVpHVUu4FioFPyW5iUuJvZdMvs',
+        amount: '999990000',
+      },
+    ],
+  };
+  const txRequest: TxRequest = {
+    txRequestId,
+    enterpriseId: '4517abfb-f567-4b7a-9f91-407509d29403',
+    transactions: [
+      {
+        unsignedTx: {
+          serializedTxHex,
+          signableHex,
+          derivationPath: 'm/0', // Needs this when key derivation is supported
+        },
+        state: 'pendingSignature',
+        signatureShares: [],
+      },
+    ],
+    unsignedTxs: [],
+    date: new Date().toISOString(),
+    intent: { intentType: 'payment' },
+    latest: true,
+    state: 'pendingUserSignature',
+    walletType: 'hot',
+    walletId: 'walletId',
+    policiesChecked: true,
+    version: 1,
+    userId: 'userId',
+    apiVersion: 'full',
+  };
+
+  const txRequestForMessageSigning: TxRequest = {
+    txRequestId,
+    enterpriseId: '4517abfb-f567-4b7a-9f91-407509d29403',
+    messages: [
+      {
+        messageRaw,
+        messageEncoded,
+        derivationPath: 'm/0',
+        state: 'pendingSignature',
+        signatureShares: [],
+      },
+    ],
+    unsignedTxs: [],
+    date: new Date().toISOString(),
+    intent: { intentType: 'payment' },
+    latest: true,
+    state: 'pendingUserSignature',
+    walletType: 'hot',
+    walletId: 'walletId',
+    policiesChecked: true,
+    version: 1,
+    userId: 'userId',
+    apiVersion: 'full',
+  };
+
+  let userKeyShare: Buffer;
+  let backupKeyShare: Buffer;
+  let bitgoKeyShare: Buffer;
+
+  before(async () => {
+    bitgo = TestBitGo.decorate(BitGo, { env: 'mock' });
+    bitgo.initializeTestVars();
+    const bgUrl = common.Environments[bitgo.getEnv()].uri;
+    bitgoGpgKey = await openpgp.generateKey({
+      userIDs: [{ name: 'bitgo', email: 'bitgo@test.com' }],
+      curve: 'ed25519',
+      format: 'armored',
+    });
+    bitgoPrvKeyObj = await openpgp.readPrivateKey({ armoredKey: bitgoGpgKey.privateKey });
+    const constants = {
+      mpc: {
+        bitgoPublicKey: bitgoGpgKey.publicKey,
+        bitgoEddsaMpcv2PublicKey: bitgoGpgKey.publicKey,
+      },
+    };
+    nock(bgUrl).get('/api/v1/client/constants').times(20).reply(200, { ttl: 3600, constants });
+
+    const [userDkg, backupDkg, bitgoDkg] = await MPSUtil.generateEdDsaDKGKeyShares();
+    userKeyShare = userDkg.getKeyShare();
+    backupKeyShare = backupDkg.getKeyShare();
+    bitgoKeyShare = bitgoDkg.getKeyShare();
+
+    baseCoin = bitgo.coin(coinName);
+
+    const walletData = {
+      id: txRequest.walletId,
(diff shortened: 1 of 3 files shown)


zahin-mohammad (commented):
Reviewed against base #8687. Built locally — `tsc --noEmit` clean in `sdk-core`, `signTxRequest.ts` test passes (5/5).

The orchestration faithfully mirrors ECDSA's `EcdsaMPCv2Utils.signRequestBase` and the protocol structure looks right: WASM rounds 0/1/2 interleaved with API rounds 1/2/3, no client-side verification on round 3 (WP finalises). Nothing here freezes public API surface — the two new public methods (`signTxRequest`, `signTxRequestForMessage`) match the existing `ITSSUtils`/ECDSA shape, and `signRequestBase` is private. So everything below is **follow-up safe**.

Most significant finding is the missing `mpcv2PartyId` plumbing — backup signing is implicitly unsupported here even though the helpers from #8687 already accept the party id. ECDSA supports it; this is a behavioral gap vs the parallel path.

See inline comments.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -293,4 +305,196 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+      txOrMessageToSign = txRequest.messages[0].messageEncoded;
+      assert(txOrMessageToSign, 'Missing messageEncoded in messages[0]');
+      derivationPath = txRequest.messages[0].derivationPath || 'm/0';
+      bufferContent = Buffer.from(txOrMessageToSign, 'hex');
+    } else {
+      throw new Error('Invalid request type');
+    }
+
+    // ── WASM Round 0 ──────────────────────────────────────────────────────────
+    const userDsg = new EddsaMPSDsg.DSG(MPCv2PartiesEnum.USER);
```
zahin-mohammad: **Backup-signer support gap.** `params.mpcv2PartyId` (declared on `TSSParamsWithPrv` for exactly this purpose) is silently dropped. ECDSA's `signRequestBase` plumbs it through to both DSG init and the share helpers (see `ecdsaMPCv2.ts:784, 792, 827, 845, 876, 893`); here the user is hardcoded to `MPCv2PartiesEnum.USER`.

The helpers from #8687 already accept `partyId: 0 | 1`, so plumbing is mechanical:

```ts
const partyId = params.mpcv2PartyId ?? 0;
const userParty = partyId === 0 ? MPCv2PartiesEnum.USER : MPCv2PartiesEnum.BACKUP;
const userDsg = new EddsaMPSDsg.DSG(userParty);
userDsg.initDsg(userKeyShare, bufferContent, derivationPath, MPCv2PartiesEnum.BITGO);
// pass partyId through each getEddsaSignatureShareRoundN call
```

Follow-up safe — but worth filing as a Linear ticket so it doesn't get lost. Without it, EdDSA MPCv2 backup-signer flow can't go through this path.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -293,4 +305,196 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+
+  /**
+   * Full 3-round EdDSA MPCv2 (MPS) online signing orchestration.
+   *
+   * Protocol overview:
+   *   WASM Round 0  → getFirstMessage()      (produces userMsg1)
+   *   API  Round 1  → send userMsg1 / recv bitgoMsg1
+   *   WASM Round 1  → handleIncomingMessages([userMsg1, bitgoMsg1])  (produces userMsg2)
+   *   API  Round 2  → send userMsg2 / recv bitgoMsg2
+   *   WASM Round 2  → handleIncomingMessagexs([userMsg2, bitgoMsg2])  (produces userMsg3)
```
zahin-mohammad: Typo: `handleIncomingMessagexs` → `handleIncomingMessages`.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -293,4 +305,196 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+    }
+
+    if (requestType === RequestType.tx) {
+      assert(txRequest.transactions || txRequest.unsignedTxs, 'Unable to find transactions in txRequest');
+      const unsignedTx =
+        txRequest.apiVersion === 'full' ? txRequest.transactions![0].unsignedTx : txRequest.unsignedTxs![0];
+      txOrMessageToSign = unsignedTx.signableHex;
+      derivationPath = unsignedTx.derivationPath;
+      bufferContent = Buffer.from(txOrMessageToSign, 'hex');
+      assert(txOrMessageToSign, 'Missing signableHex in unsignedTx');
```
zahin-mohammad: Assert ordering — the assert runs *after* `Buffer.from(txOrMessageToSign, 'hex')`. If `signableHex` is `undefined`, `Buffer.from(undefined, 'hex')` throws a `TypeError` first and the helpful assertion message never surfaces. Same pattern at L388-389 for the message branch. Suggest moving the assert above the `Buffer.from` call:

```ts
txOrMessageToSign = unsignedTx.signableHex;
assert(txOrMessageToSign, 'Missing signableHex in unsignedTx');
bufferContent = Buffer.from(txOrMessageToSign, 'hex');
```

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -293,4 +305,196 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+      throw new Error('Unexpected signature share response. Unable to parse data.');
+    }
+
+    const bitgoDeserializedMsg1 = await verifyBitGoEddsaMessageRound1(
+      parsedBitGoToUserSigShareRoundOne,
+      bitgoGpgPubKey
+    );
+
+    // ── WASM Round 1 ──────────────────────────────────────────────────────────
+    const [userMsg2] = userDsg.handleIncomingMessages([userMsg1, bitgoDeserializedMsg1]);
```
zahin-mohammad, line 438: Defensive: `handleIncomingMessages` returns `DeserializedMessages` (an array). If WASM ever returns `[]` or an unexpected length, `userMsg2` becomes `undefined` and the next `getEddsaSignatureShareRound2(undefined, ...)` crashes with a confusing PGP error. A targeted `assert(userMsg2, 'WASM round 1 produced no message')` after the destructure would surface the real failure. Same applies to `userMsg3` at L478.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -293,4 +305,196 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+    assert(latestTxRequest.transactions || latestTxRequest.messages, 'Invalid txRequest object after round 2');
+
+    const txRequestSignatureShares =
+      requestType === RequestType.tx
+        ? latestTxRequest.transactions![0].signatureShares
+        : latestTxRequest.messages![0].signatureShares;
+
+    const parsedBitGoToUserSigShareRoundTwo = decodeWithCodec(
+      EddsaMPCv2SignatureShareRound2Output,
+      JSON.parse(txRequestSignatureShares[txRequestSignatureShares.length - 1].share),
```
zahin-mohammad: Fragile share lookup: `txRequestSignatureShares[length - 1]` assumes BitGo's response share is always last. ECDSA does the same thing so it's not a regression, but it would be more robust to filter explicitly:

```ts
const bitgoShare = txRequestSignatureShares.find(
  (s) => s.from === SignatureShareType.BITGO && s.to === SignatureShareType.USER
);
assert(bitgoShare, 'Missing BitGo round-2 share');
```

Follow-up safe; flagging here so the pattern doesn't propagate further.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -293,4 +305,196 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+   * @param {string} params.reqId - request id
+   * @returns {Promise<TxRequest>} fully signed TxRequest
+   */
+  async signTxRequest(params: TSSParamsWithPrv): Promise<TxRequest> {
+    this.bitgo.setRequestTracer(params.reqId);
+    return this.signRequestBase(params, RequestType.tx);
+  }
+
+  /**
+   * Signs the message associated with the transaction request.
```
zahin-mohammad, line 326: Note (not blocking, pre-existing pattern): `signTxRequestForMessage` accepts a `TSSParamsForMessageWithPrv` which includes `messageRaw: string` and `bufferToSign: Buffer`, but `signRequestBase` ignores both — it derives `bufferContent` from `txRequest.messages[0].messageEncoded`. ECDSA does the same thing so this PR isn't introducing the issue, but it's a quiet footgun: a caller who passes `bufferToSign` thinking that's what gets signed will be confused when actual signing uses the txRequest's messageEncoded. Worth filing as a separate cleanup ticket against the shared `TSSParamsForMessageWithPrv` contract.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/signTxRequest.ts
```diff
@@ -0,0 +1,455 @@
...
+        ],
+      });
+
+    const userPrvBase64 = Buffer.from(userKeyShare).toString('base64');
+    await tssUtils
+      .signTxRequest({ txRequest, prv: userPrvBase64, reqId, txParams })
+      .should.be.rejectedWith(/Unexpected signature share response/);
+  });
+
+  it('successfully signs a txRequest for an mps hot wallet after receiving multiple 429 errors', async function () {
```
zahin-mohammad: Coverage gap: rate-limit retries are tested for round 2 only. Rounds 1 and 3 use the same `sendSignatureShareV2` retry path but aren't exercised. A parameterised helper that injects 429s on each round in turn would be cheap to add.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/signTxRequest.ts
```diff
@@ -0,0 +1,455 @@
...
+      txRequest: txRequestForMessageSigning,
+      prv: userPrvBase64,
+      reqId,
+      messageRaw: txRequestForMessageSigning.messages![0].messageRaw,
+      bufferToSign: Buffer.from(messageEncoded, 'hex'),
+    });
+    nockPromises[0].isDone().should.be.true();
+    nockPromises[1].isDone().should.be.true();
+    nockPromises[2].isDone().should.be.true();
+    nockPromises[3].isDone().should.be.true();
```
zahin-mohammad, line 215: Coverage gaps that pair with the `mpcv2PartyId` finding on the impl side:
- Tests only cover party 0 (USER); BACKUP path is untested.
- No round-2 tampered-signature negative test (round 1 negative is at L198, but round 2's `verifyBitGoEddsaMessageRound2` could swallow tampering with no test catching it).
- No test for malformed/missing round-3 server response — currently any failure there surfaces as a confusing `sendTxRequest` error rather than something localised to round 3.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
Marzooqa: I'll look into this as a follow-up https://linear.app/bitgo/issue/WCI-362/fix-tssparamsformessage-buffertosign-footgun-validate-against

zahin-mohammad (dismissed):
LGTM — gated on #8687 landing first (per the `DO NOT MERGE` label).

All 8 prior comments are resolved in this revision:

- ✅ **Backup-signer plumbing**: `params.mpcv2PartyId` now flows through DSG init + all three `getSignatureShareRound{One,Two,Three}` calls; `signerShareType` derived from `partyId`
- ✅ Typo `handleIncomingMessagexs` → `handleIncomingMessages`
- ✅ Assert ordering moved before `Buffer.from(...)` in both tx and message branches
- ✅ Defensive `assert(userMsg2, ...)` / `assert(userMsg3, ...)` after each `handleIncomingMessages` destructure
- ✅ Fragile `[length - 1]` share lookup replaced with new `getBitgoSignatureShare(shares, signerShareType)` helper in `tss/common.ts` that filters by `from === BITGO && to === signerShareType`
- ✅ `messageRaw`/`bufferToSign` footgun tracked as follow-up [WCI-362](https://linear.app/bitgo/issue/WCI-362/fix-tssparamsformessage-buffertosign-footgun-validate-against)
- ✅ Rate-limit retry tests added for rounds 1 and 3 (round 2 was already covered)
- ✅ Test coverage: BACKUP path ("signs a txRequest with backup key") + malformed round-3 response added. Round-2 tampered-signature is covered at the helper level in #8687 (`verifyBitGoMessageRoundTwo should throw on a tampered message`), so the orchestration-level gap is acceptable

### Residual nits — non-blocking, address in a follow-up

1. **Inconsistent error style in `signRequestBase`.** The GPG-key check uses `if (!bitgoGpgPubKey) throw new Error(...)` while every other invariant in this method uses `assert(...)`. Unifying on `assert(bitgoGpgPubKey, 'Missing BitGo GPG key for MPCv2')` would keep the failure-mode style consistent.
2. **Unnamed boolean args** to `pickBitgoPubGpgKeyForSigning(true, params.reqId, txRequest.enterpriseId, true)`. This is a pre-existing pattern carried over from ECDSA, not introduced here, but the call site reads as flag soup. Worth a separate cleanup ticket against the helper signature (named-options object or named consts) so the pattern doesn't propagate further.
3. **Round-2 orchestration-level negative test** for a wrong-type response would mirror the existing `should throw if round 1 response has wrong type` and round the symmetry out. Not strictly needed since `verifyBitGoMessageRoundTwo` is unit-tested at the helper layer, but cheap to add.

None of these block — happy to approve as-is.

zahin-mohammad (approved):

PR #8700, opened 2026-05-06
fet(hbar): add AccountUpdate support to explainTransaction in BitGoJS

aditya-bitgo:
ticket: SI-511

### DIFF
--- a/modules/sdk-coin-hbar/src/hbar.ts
+++ b/modules/sdk-coin-hbar/src/hbar.ts
@@ -751,6 +751,15 @@ export class Hbar extends BaseCoin {
         });
         break;
 
+      case 'cryptoUpdateAccount':
+        const updateParams = txJson.instructionsData.params;
+        outputs.push({
+          address: updateParams.accountId,
+          amount: '0',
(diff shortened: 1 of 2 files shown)


Doddanna17 (commented):
Fix is correct and necessary. Without the `cryptoUpdateAccount` case, `verifyTransaction` throws for all staking ops since `explainTransaction` is called internally for every transaction type.

Two notes: (1) PR title typo — `fet(hbar)` → `feat(hbar)`. (2) Unstake test is missing `id`, `timestamp`, `memo` assertions that the stake test has — inconsistency could mask regressions.

LGTM.

Doddanna17 (approved):
Fix is correct and necessary. Without the `cryptoUpdateAccount` case, `verifyTransaction` throws for all staking ops since `explainTransaction` is called internally for every transaction type.

Two notes: (1) PR title typo — `fet(hbar)` → `feat(hbar)`. (2) Unstake test is missing `id`, `timestamp`, `memo` assertions that the stake test has — inconsistency could mask regressions.

LGTM.

PR #8703, opened 2026-05-06
fix(sdk-coin-xdc): add tokenApproval to verifyTssTransaction

ArunBala-Bitgo:
the bypass list in xdcToken.ts and xdc.ts was copied from a stale bscToken.ts and was missing tokenApproval and consolidate transaction types. This caused enableToken (TXCD:TMT) to fail with "missing txParams" because tokenApproval transactions have no recipients and were not in the bypass list.

Fixes CECHO-915.

TICKET: CECHO-915


MohammedRyaan786 (approved):
LGTM

PR #8715, opened 2026-05-07
fix(passkey-crypto): use base64url everywhere for the PRF salt

derranW26:
The attach and derive flows were feeding different bytes to the WebAuthn PRF extension for the same passkey, so the password derived at attach time could not decrypt the keychain blob written with that same password (ccm tag mismatch on every transaction approval).

Three encoding inconsistencies caused this:

- deriveEnterpriseSalt returned hex while the server stores base64url and the WebAuthn extension expects raw bytes — every consumer had to re-encode, and one of them did it wrong.
- attachPasskeyToWallet ran the hex output back through a hex-to-base64url conversion before handing it to provider.get, so the browser PRF saw the hex characters interpreted as base64 garbage.
- prfHelpers.buildEvalByCredential tried to convert the stored base64url salt to hex via Buffer.from(...).toString('hex'), which is a no-op under the browser Buffer polyfill and a real conversion in Node — same code, different bytes.

Standardise on base64url end-to-end: deriveEnterpriseSalt emits base64url, attachPasskeyToWallet passes that string straight through to the PRF eval, and prfHelpers reads device.prfSalt unchanged. The WebAuthn provider layer is the single point that decodes base64url to bytes for navigator.credentials.get.

Refs: WCN-410

TICKET: WCN-410

### DIFF
--- a/modules/passkey-crypto/src/deriveEnterpriseSalt.ts
+++ b/modules/passkey-crypto/src/deriveEnterpriseSalt.ts
@@ -1,4 +1,5 @@
 import { createHmac } from 'crypto';
+import { base64UrlToBuffer, toBase64Url } from './base64url';
 
 /**
  * Derives an enterprise-scoped PRF salt to prevent cross-enterprise key reuse.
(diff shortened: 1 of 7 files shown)


modules/passkey-crypto/src/deriveEnterpriseSalt.ts
```diff
@@ -6,11 +6,22 @@ import { createHmac } from 'crypto';
...
  */
 export function deriveEnterpriseSalt(baseSalt: string, enterpriseId: string): string {
   const keyBytes = Buffer.from(baseSalt.replace(/-/g, '+').replace(/_/g, '/'), 'base64');
-  return createHmac('sha256', keyBytes).update(enterpriseId).digest('hex');
+  return createHmac('sha256', keyBytes)
+    .update(enterpriseId)
+    .digest('base64')
+    .replace(/\+/g, '-')
+    .replace(/\//g, '_')
+    .replace(/=+$/, '');
```
mohammadalfaiyazbitgo: you use this pattern in quite a few places, could we make a utility function and reuse it?

mohammadalfaiyazbitgo (changes requested):

mohammadalfaiyazbitgo (approved):

PR #8737, opened 2026-05-11
fix: explicitly set isToken parameter in OFC token definitions

yashvanthbl137-crypto:
Ticket: CGD-1104


shubham-damkondwar (approved):
lgtm

PR #8741, opened 2026-05-12
feat: full TSS recovery flow for the Tempo chain (TIP-20 / EIP-7702

bhavesh-prasad:
## Summary

Implements the full TSS recovery flow for the Tempo chain (TIP-20 / EIP-7702).

### What was there before
- `buildUnsignedSweepTxnTSS` and `recoveryBlockchainExplorerQuery` were stubs returning empty/dummy objects
- No signed recovery path existed

### What this PR adds

**Recovery infrastructure (`tempo.ts`)**
- `recoveryBlockchainExplorerQuery` — routes Etherscan-style queries to Tempo's Alchemy JSON-RPC endpoint (configured via `environments.ts`)
- `queryTempoRpc` — translates `account/balance`, `account/tokenbalance`, and `account/txlist` queries into `eth_call` / `eth_getTransactionCount` JSON-RPC calls
- `buildTempoSweepTx` — shared helper that queries token balance, computes sweep amount, and builds an unsigned TIP-20 transfer transaction; used by both recovery paths
- `recoverTSS` — overrides base-class recovery with two paths:
  - **Unsigned sweep** (plain public key shares): delegates to `buildUnsignedSweepTxnTSS`
  - **Signed sweep** (encrypted keys + passphrase): builds a TIP-20 tx, hashes it, signs via MPC v2 ECDSA, attaches the signature, and returns the broadcast-ready hex
- `buildUnsignedSweepTxnTSS` — builds an unsigned TIP-20 sweep and returns the `UnsignedSweepTxMPCv2` structure for offline vault signing

**Dynamic gas margin**
- Removes hardcoded `GAS_MARGIN_UNITS = 20_000n` constant
- Gas margin is now computed from user-supplied params: `gasMarginUnits = gasLimit × maxFeePerGas / 10^12`
- Matches the `gasLimit × gasPrice` pattern used across all ETH-like coin recovery flows in `abstract-eth`; the `/10^12` factor converts from wei scale (18 decimals) to pathUSD's TIP-20 units (6 decimals)

**Environment config (`environments.ts`)**
- Adds Tempo `evm` config entry so `recoveryBlockchainExplorerQuery` can resolve the RPC URL and API token


modules/sdk-coin-tempo/src/tempo.ts
```diff
@@ -327,33 +341,236 @@ export class Tempo extends AbstractEthLikeNewCoins {
...
+      );
+    }
+
+    const sweepAmountHuman = tip20UnitsToAmount(sweepAmount);
+    const nonce = await this.getAddressNonce(walletAddress, params.apiKey);
+
+    const txBuilder = this.getTransactionBuilder() as unknown as Tip20TransactionBuilder;
+    txBuilder
+      .addOperation({ token: tokenAddress, to: params.recoveryDestination, amount: sweepAmountHuman })
+      .feeToken(PATH_USD_ADDRESS)
```
nvrakesh06: Let’s not use `PATH_USD_ADDRESS` as the default fee token.

This contract address is only for testnet and would fail on mainnet. Instead, we should use the same token being transferred as the fee token by default, unless the user explicitly mentions a different fee token in the WRW form.

modules/sdk-coin-tempo/src/tempo.ts
```diff
@@ -327,33 +341,236 @@ export class Tempo extends AbstractEthLikeNewCoins {
...
+   *
+   * - tokenContractAddress: the token to sweep (defaults to pathUSD)
+   * - feeToken: always pathUSD
+   * - If sweeping pathUSD itself, reserves gasLimit × maxFeePerGas / 10^12 pathUSD units for fees.
+   */
+  private async buildTempoSweepTx(
+    walletAddress: string,
+    params: RecoverOptions
+  ): Promise<{ tx: Tip20Transaction; nonce: number; sweepAmount: bigint; gasMarginUnits: bigint }> {
+    const tokenAddress = params.tokenContractAddress ?? PATH_USD_ADDRESS;
```
nvrakesh06: Let’s throw an error instead of defaulting to a token.

The current default token only exists on testnet, so using it could fail on mainnet and may also lead to accidental recovery with the wrong token. Throwing an error makes the fee token requirement explicit and safer.

nvrakesh06 (commented):
.

nvrakesh06 (changes requested):

nvrakesh06 (changes requested):
.

nvrakesh06 (commented):
## Fee estimation — two cases

**Case A — swept token == fee token** ✅ handled

When sweeping pathUSD (the fee token), the gas cost is correctly reserved before sweeping:

```ts
const sweepAmount = isPathUsd ? balance - gasMarginUnits : balance;
if (sweepAmount <= 0n) { throw ... }
```

The function throws early if the balance doesn't cover the fee, so the wallet is never left unable to pay.

---

**Case B — swept token != fee token** ❌ not handled

When sweeping a different TIP-20 (e.g. `tempo:usdc`), the full token balance is swept but the fee token balance is never checked. If the wallet has no pathUSD, the transaction will be submitted and fail on-chain silently.

The same guard from Case A needs to be applied to the fee token balance here:

```ts
if (feeTokenAddress.toLowerCase() !== tokenAddress.toLowerCase()) {
  const feeTokenRaw = await this.queryAddressTokenBalance(feeTokenAddress, walletAddress, params.apiKey);
  if (BigInt(feeTokenRaw.toString()) < gasMarginUnits) {
    throw new Error(
      `Insufficient fee token balance in ${feeTokenAddress}: ` +
      `have ${feeTokenRaw} units, need at least ${gasMarginUnits} units for gas`
    );
  }
}
```

modules/sdk-coin-tempo/src/tempo.ts
nayandas190: @bhavesh-prasad Regarding the feetoken, we can ask PM for the expected behaviour here.

bhavesh-prasad: I have started the thread but kept the flow as normal wallet transfers for now

nvrakesh06 (approved):

mohd-kashif (approved):

PR #8757, opened 2026-05-12
fix(root): bump protobufjs to 7.5.8 via yarn resolution

manojkumar138:
## Summary
- Add yarn resolution `"protobufjs": "^7.5.8"` to bump all protobufjs deps to patched version
- Fixes GHSA-66ff-xgx4-vchm (code injection via bytes field defaults) and GHSA-75px-5xx7-5xc7 (code generation gadget after prototype pollution)
- All protobufjs entries in yarn.lock now resolve to 7.5.8 (previously 6.11.4, 7.2.5, 7.5.4)

## Test plan
- [ ] CI audit step passes
- [ ] AppSec approval for dependency bump

CECHO-973

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,8 @@
     "sjcl": "npm:@bitgo/sjcl@1.0.1",
     "picomatch": ">=2.3.2",
     "fast-uri": "3.1.2",
-    "@babel/plugin-transform-modules-systemjs": "7.29.4"
+    "@babel/plugin-transform-modules-systemjs": "7.29.4",
+    "protobufjs": "7.5.6"
   },
   "workspaces": [
     "modules/*"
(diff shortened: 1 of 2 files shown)


bhargavirao24:
@manojkumar138  SafeChain is blocking 7.5.8 because it was published today and is still within the 7-day cooldown window. 7.5.6 already has the fix and has been out for around 2 weeks, so it’s past the cooldown. Can you update the resolution from 
"protobufjs": "^7.5.8"
"protobufjs": "7.5.6"
Exact pinning is nice for resolutions since it keeps the version reproducible and easier to audit. That should fix both advisories and unblock CI without needing any approval and bypass.

This should work but let me know if you run into any issues

mohd-kashif (approved):

PR #8759, opened 2026-05-12
chore: bump public types

alia-bitgo:
## Summary

- Bumps `@bitgo/public-types` from `5.97.0` to `6.9.0` in `sdk-core`, `abstract-lightning`, `sdk-coin-sol`, `sdk-coin-flrp`, `bitgo`, and `express`

## Problem

The `5.97.0` pin caused npm to hoist a separate `@bitgo/public-types@5.97.0` copy into `sdk/node_modules/` inside the `wallet-platform` workspace in bitgo-microservices. This shadowed the top-level `6.5.1` copy, so `listWebhooks.ts` (which lives in `packages/wallet-platform/sdk/`) compiled and ran against an `OrganizationWebhookType` that predates `idvStatus`.

At runtime, when sumsub-service called `POST /api/v2/internal/webhook/list` with `{ organizationId, type: "idvStatus" }`, wallet-platform's io-ts validation rejected `"idvStatus"` as an invalid `OrganizationWebhookType`, returning a 400 — even though `idvStatus` had been correctly added to the enum in `@bitgo/public-types@6.5.0`.

## Why it's safe

No exports were removed between `5.97.0` and `6.9.0`. The bump is purely additive — `6.9.0` is a superset of `5.97.0`'s public API.

## Resolutions                                                                                                                                               
                                                                                                                                                            
  Added two yarn resolutions entries to fix a hoisting regression introduced by this bump:                                                                  
                  
  "**/cliui/strip-ansi": "6.0.1",
  "**/cliui/string-width": "4.2.3",
  "**/yargs/cliui/string-width": "4.2.3"

  Upgrading 6 packages from @bitgo/public-types@5.97.0 to 6.5.0 changed yarn's hoisting strategy, causing strip-ansi@7.1.0 and string-width@5.1.2 (both ESM-only) to be promoted to the root node_modules/. cliui@8.0.1 and yargs require their CJS ^6/^4 counterparts — without a nested copy, require('strip-ansi') and require('string-width') returned ESM modules where mixin.stripAnsi and s.stringWidth are not functions, crashing lerna list and any yargs-based CLI. The resolutions force the CJS-compatible versions into cliui's dependency path without affecting other packages.

## Test plan

- [ ] Lock file resolves without a nested `@bitgo/public-types@5.97.0` copy in `wallet-platform/sdk/node_modules/`
- [ ] Existing unit tests pass across affected modules
- [ ] After wallet-platform picks up new `@bitgo-beta` builds from this, the sumsub-service `idvStatus` webhook flow succeeds end-to-end on staging

### DIFF
--- a/modules/abstract-lightning/package.json
+++ b/modules/abstract-lightning/package.json
@@ -39,7 +39,7 @@
     ]
   },
   "dependencies": {
-    "@bitgo/public-types": "5.97.0",
+    "@bitgo/public-types": "6.9.0",
     "@bitgo/sdk-core": "^36.44.0",
     "@bitgo/statics": "^58.39.0",
     "@bitgo/utxo-lib": "^11.22.1",
--- a/modules/bitgo/package.json
+++ b/modules/bitgo/package.json
@@ -139,7 +139,7 @@
     "superagent": "^9.0.1"
   },
   "devDependencies": {
-    "@bitgo/public-types": "5.97.0",
+    "@bitgo/public-types": "6.9.0",
     "@bitgo/sdk-opensslbytes": "^2.1.0",
     "@bitgo/sdk-test": "^9.1.42",
     "@openpgp/web-stream-tools": "0.0.14",
--- a/modules/express/package.json
+++ b/modules/express/package.json
@@ -60,7 +60,7 @@
     "superagent": "^9.0.1"
   },
   "devDependencies": {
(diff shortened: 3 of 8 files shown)


alia-bitgo:
dev-ex overrode breaking changes lint - was a false positive.

pranavjain97 (approved):

PR #8762, opened 2026-05-13
feat(sdk-lib-mpc): add executeTillRound handler util

vibhavgo:
## Description

Adds an executeTillRound orchestration utility for the EdDSA MPSv2 DSG protocol, mirroring DklsUtils.executeTillRound for ECDSA DKLS. Also exposes getPartyIdx and getOtherPartyIdx accessors on the DSG class to support index-free wiring inside the utility.

- Add executeTillRound to modules/sdk-lib-mpc/src/tss/eddsa-mps/util.ts; accepts two un-initialized DSG instances plus key shares, message, and derivation path, calls initDsg internally using getPartyIdx(), and drives all 3 interactive rounds (WaitMsg1 -> WaitMsg2 -> WaitMsg3 -> Complete)
- Return intermediate DeserializedMessages[] for rounds 1-2 and the final 64-byte Ed25519 signature Buffer for round 3
- Add getPartyIdx() and getOtherPartyIdx() accessors to DSG class
- Add unit tests in modules/sdk-lib-mpc/test/unit/tss/eddsa/eddsa-utils.ts covering executeTillRound: per-pair signatures across all three 2-of-3 party combinations (0+1, 0+2, 1+2), root-key signature verification via getCommonKeychain(), derived-path (m/0/0) key isolation check, intermediate round return-type assertions, and out-of-range round error handling
- Migrate existing dsg.ts callers from the local runEdDsaDSG test helper to MPSUtil.executeTillRound and remove runEdDsaDSG from test/unit/tss/eddsa/util.ts
- Add dsg.ts protocol test asserting wasm-mps rejects cross-message signing with the wrapped "round WaitMsg2: Protocol Error" failure

## Issue Number

WCI-386

## Type of change

## How Has This Been Tested?

Unit tests added in modules/sdk-lib-mpc/test/unit/tss/eddsa/eddsa-utils.ts under the executeTillRound describe block, plus a cross-message protocol test in dsg.ts. Tests cover:

- Per-pair signature production for all three 2-of-3 party combinations (0+1, 0+2, 1+2), each verifying under the DKG public key
- Round 3 returns a 64-byte signature that verifies under the root public key from getCommonKeychain()
- Round 3 with derivation path m/0/0 produces a signature that does not verify under the root public key, confirming key derivation occurred
- Rounds 1 and 2 return DeserializedMessages[][] and not a Buffer
- Round values 0 and 4 throw Invalid round number
- Cross-message signing fails at round WaitMsg2 with the wrapped "Protocol Error" from wasm-mps
- Existing dsg.ts protocol, derivation-path, and session-management tests migrated to executeTillRound and continue to pass

Run with:
yarn run unit-test --scope @bitgo/sdk-lib-mpc -- -- --grep "executeTillRound"

## Checklist

- [x] My commits follow Conventional Commits and I have properly described any BREAKING CHANGES
- [x] I have made corresponding changes to the documentation and on any new/updated functions and/or methods - jsdoc


modules/sdk-lib-mpc/test/unit/tss/eddsa/dsg.ts
```diff
@@ -203,6 +203,74 @@ describe('EdDSA MPS DSG', function () {
     });
   });
 
+  describe('executeTillRound', function () {
```
Marzooqa: These tests cover a utility function in `eddsa-mps/util.ts`, not the DSG class itself. They should live in `test/unit/tss/eddsa/eddsa-utils.ts` alongside the existing `concatBytes` and `generateEdDsaDKGKeyShares` tests, not here in `dsg.ts`.

modules/sdk-lib-mpc/test/unit/tss/eddsa/dsg.ts
```diff
@@ -203,6 +203,74 @@ describe('EdDSA MPS DSG', function () {
...
+    it('should return 64-byte signature verifying under root public key from getCommonKeychain()', function () {
+      const party1Dsg = new EddsaMPSDsg.DSG(0);
+      const party2Dsg = new EddsaMPSDsg.DSG(2);
```
Marzooqa: All five tests in this describe block hard-code partyIdx 0 and 2 (user + BitGo). The DKLS equivalent tests all three 2-of-3 combinations (0+1, 0+2, 1+2) with a vector table. Missing 0+1 and 1+2 leaves user+backup and backup+BitGo signing paths untested.

modules/sdk-lib-mpc/test/unit/tss/eddsa/dsg.ts
```diff
@@ -203,6 +203,74 @@ describe('EdDSA MPS DSG', function () {
...
+        MESSAGE,
+        'm/0/0'
+      ) as Buffer;
+      assert.strictEqual(sig.length, 64, 'Derived path signature must be 64 bytes');
+
+      // A derived path uses a different signing key, so the signature must NOT
+      // verify under the root public key.
+      const rootPubKey = Buffer.from(userDkg.getCommonKeychain().slice(0, 64), 'hex');
+      assert(
+        !ed25519.verify(sig, MESSAGE, rootPubKey),
```
Marzooqa: This assertion is negative-only: it confirms the signature does not verify under the root key but does not confirm it verifies under the derived key. If WASM derivation produced garbage, this test would still pass. Can be addressed as a follow-up once the derived public key extraction utility is available.

modules/sdk-lib-mpc/test/unit/tss/eddsa/dsg.ts
```diff
@@ -203,6 +203,74 @@ describe('EdDSA MPS DSG', function () {
...
+      assert.throws(
+        () => MPSUtil.executeTillRound(0, party1Dsg, party2Dsg, userKeyShare, bitgoKeyShare, MESSAGE, 'm'),
+        /Invalid round number/
+      );
+      assert.throws(
+        () => MPSUtil.executeTillRound(4, party1Dsg, party2Dsg, userKeyShare, bitgoKeyShare, MESSAGE, 'm'),
+        /Invalid round number/
+      );
+    });
+  });
```
Marzooqa: Missing test: two parties signing different messages should fail at the WASM level. The DKLS suite covers this (`dklsDsg.ts` line 243 — `should fail when signing two different messages`). Without it, the suite does not verify the protocol rejects cross-message signing attempts.

modules/sdk-lib-mpc/test/unit/tss/eddsa/dsg.ts
```diff
@@ -1,6 +1,6 @@
 import assert from 'assert';
 import { ed25519 } from '@noble/curves/ed25519';
-import { EddsaMPSDkg, EddsaMPSDsg, MPSTypes } from '../../../../src/tss/eddsa-mps';
+import { EddsaMPSDkg, EddsaMPSDsg, MPSTypes, MPSUtil } from '../../../../src/tss/eddsa-mps';
 import { generateEdDsaDKGKeyShares, runEdDsaDSG } from './util';
```
Marzooqa: `runEdDsaDSG` in `test/unit/tss/eddsa/util.ts` is now redundant given `executeTillRound`. Since `executeTillRound` mutates the DSG instances it receives in-place, callers that need the instances after the call (e.g. for `getState()`, `getSession()`, or calling `handleIncomingMessages` on a completed instance) still have access to them — so every existing use of `runEdDsaDSG` can be migrated.

This follows the same pattern as `generateEdDsaDKGKeyShares`, which was previously a local copy in `test/util.ts` and is now a re-export from the production source. `runEdDsaDSG` should be removed from the test utility and its callers updated to use `MPSUtil.executeTillRound` directly.

modules/sdk-lib-mpc/test/unit/tss/eddsa/dsg.ts
vibhavgo: Will defer to follow-up

modules/sdk-lib-mpc/test/unit/tss/eddsa/eddsa-utils.ts
```diff
@@ -51,4 +53,115 @@ describe('EdDSA Utility Functions', function () {
...
+        new EddsaMPSDsg.DSG(0),
+        new EddsaMPSDsg.DSG(2),
+        keySharesByIdx[0],
+        keySharesByIdx[2],
+        MESSAGE,
+        'm'
+      );
+      assert(!Buffer.isBuffer(result), 'round 2 should return message arrays, not a Buffer');
+      assert.strictEqual(result.length, 2, 'should contain message arrays for both parties');
+    });
```
Marzooqa, line 152: Follow-up: `dsg1` and `dsg2` are reused across both `assert.throws` calls. Safe today because the range check fires before `initDsg`, leaving both instances `Uninitialized`. If the range check ever moves below `initDsg`, the second call would silently receive already-initialised instances. Consider creating fresh instances per assertion.

modules/sdk-lib-mpc/test/unit/tss/eddsa/dsg.ts
```diff
@@ -121,27 +154,78 @@ describe('EdDSA MPS DSG', function () {
...
+
+      const r0_1 = dsg1.getFirstMessage();
+      const r0_2 = dsg2.getFirstMessage();
+
+      const [r1_1] = dsg1.handleIncomingMessages([r0_1, r0_2]);
+      const [r1_2] = dsg2.handleIncomingMessages([r0_1, r0_2]);
+
+      assert.throws(
+        () => dsg1.handleIncomingMessages([r1_1, r1_2]),
+        /Error while creating messages from party 0, round WaitMsg2: Protocol Error/
```
Marzooqa, line 200: Follow-up: `Protocol Error` is sourced from the WASM error string. If the WASM crate changes its wording this regex will stop matching. A pattern anchored to the deterministic prefix — e.g. `/Error while creating messages from party 0, round WaitMsg2/` — would be more resilient to WASM error message changes.

Marzooqa (approved):
Implementation is correct, well-tested, and consistent with the ECDSA DKLS pattern. Two minor test-only nits as follow-ups — no public API impact.

PR #8772, opened 2026-05-14
fix(abstract-eth): verify inner batch calldata recipients [CGD-1319]

kamleshmugdiya:
> Replaces #8771 — that PR auto-closed when its head branch was renamed; same commit `843b46332d`.

## Description

`verifyTransaction` for ETH batcher-contract sends only checked (a) the sum of `txParams.recipients` against `txPrebuild.recipients[0].amount` and (b) that the outer recipient was the network's `batcherContractAddress`. It never decoded the inner `batch(address[],uint256[])` calldata embedded in `txPrebuild.txHex`, and `signTransaction` then loaded that `txHex` verbatim.

A compromised BitGo platform could preserve both outer checks while swapping the inner `(address, amount)` pairs to attacker-controlled addresses, redirecting batched payouts. This PR decodes the inner calldata during verification and compares each pair to user intent.

## Issue Number

CGD-1319

## Type of change

## What changed

- `modules/abstract-eth/src/lib/walletUtil.ts` — exports `batchMethodId` (`0xc00c4e9e`), `batchMethodName`, `batchMethodTypes`.
- `modules/abstract-eth/src/lib/iface.ts` — adds `BatchTransferData` / `BatchTransferRecipient`.
- `modules/abstract-eth/src/lib/utils.ts` — adds `decodeBatchTransferData()`, modeled after the existing `decode*TransferData` helpers and reusing `getRawDecoded` / `getBufferedByteCode`.
- `modules/abstract-eth/src/abstractEthLikeNewCoins.ts` — adds private `verifyBatchInnerRecipients` which parses `txPrebuild.txHex`, decodes the outer `sendMultiSig`, decodes the inner `batch(address[],uint256[])`, and compares each `(address, amount)` pair to `txParams.recipients`. Wired into the existing batch path in `verifyTransaction` for native batches.

The verifier **fails closed** when:
- `txPrebuild.txHex` is missing
- the outer selector is not `sendMultiSig`
- the inner selector is not `batch(address[],uint256[])`
- inner recipient count, address, or amount differ from `txParams.recipients`

Token-batch (`txParams.tokenName`) and TSS paths are out of scope per the issue; the codebase has no `sendMultisigBatch`-style token wrapper, and the existing token branch already expects `txPrebuild.recipients[0].amount === 0`. `signTransaction` is unchanged — it now safely loads `txHex` because the verifier proves inner calldata matches user intent first.

ENS / non-hex `txParams` addresses skip the address comparison but still enforce the amount check, mirroring the normal-tx path at `abstractEthLikeNewCoins.ts:3340`.

# How Has This Been Tested?

- `modules/abstract-eth/test/unit/utils.ts` — 4 tests for `decodeBatchTransferData`: hardcoded selector matches runtime-computed value, round-trip encode/decode, wrong-selector rejection, mismatched-array-length rejection.
- `modules/sdk-coin-eth/test/unit/eth.ts` — adds a `buildBatchTxHex` helper and 5 new `verifyTransaction` tests: missing `txHex`, tampered inner address, tampered inner amount (with outer total preserved), wrong inner selector, mismatched inner recipient count. Updates the existing happy-path batch test to include a valid `txHex`.

Local runs:
- `yarn unit-test` in `modules/abstract-eth` — **77 passing**
- `yarn unit-test` in `modules/sdk-coin-eth` — **237 passing**
- `yarn lint` in both modules — clean

# Checklist:

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -23,6 +23,7 @@ import {
   MPCTx,
   MPCTxs,
   ParsedTransaction,
+  ITransactionRecipient,
   ParseTransactionOptions,
   PrebuildTransactionResult,
   PresignTransactionOptions as BasePresignTransactionOptions,
@@ -75,8 +76,11 @@ import secp256k1 from 'secp256k1';
 import { AbstractEthLikeCoin } from './abstractEthLikeCoin';
 import { EthLikeToken } from './ethLikeToken';
 import {
+  batchMethodId,
   calculateForwarderV1Address,
   coinFamiliesWithL1Fees,
+  decodeBatchTransferData,
+  decodeNativeTransferData,
   decodeTransferData,
   ERC1155TransferBuilder,
   ERC721TransferBuilder,
@@ -87,6 +91,7 @@ import {
   getRawDecoded,
   getToken,
   KeyPair as KeyPairLib,
+  sendMultisigMethodId,
   TransactionBuilder,
   TransferBuilder,
 } from './lib';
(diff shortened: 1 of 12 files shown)


kamleshmugdiya:
@claude

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -3325,6 +3411,13 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
           { address: txPrebuild.recipients[0].address, amount: txPrebuild.recipients[0].amount.toString() },
         ]);
       }
+
+      // Decode the inner batch(address[],uint256[]) calldata and verify each (address, amount) pair
+      // matches user intent. Without this, a compromised platform could swap inner recipients while
+      // preserving the outer total amount and batcher-address checks.
+      if (!txParams.tokenName) {
+        await this.verifyBatchInnerRecipients(txPrebuild, recipients, throwRecipientMismatch);
```
mullapudipruthvik, line 3616: shouldn't we do the same for `verifyTssTransaction` ? method

modules/sdk-coin-eth/test/unit/eth.ts
```diff
@@ -541,6 +593,182 @@ describe('ETH:', function () {
         .should.be.rejectedWith('recipient address of txPrebuild does not match batcher address');
     });
 
+    it('should reject a batch txPrebuild that omits txHex', async function () {
```
mullapudipruthvik, line 631: also add tests for other evm coins since its common method

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
kamleshmugdiya: Good catch — `verifyTssTransaction` previously had no batch verification at all (only single-recipient transfer + consolidation paths). I just pushed a commit that adds equivalent inner-recipient verification for the TSS path.

The TSS shape is slightly different from multi-sig:
- Multi-sig: outer `sendMultiSig(batcher, total, batchData, ...)` with batch calldata nested as the `data` field
- TSS: TSS wallets are EOAs, so the outer tx calls the batcher contract directly — `tx.to = batcherContractAddress`, `tx.value = total`, `tx.data = batch(addr[],amt[])` (no sendMultiSig wrapper)

I extracted the inner decode/compare into a shared `compareBatchCalldataAgainstRecipients` helper used by both paths. New TSS-specific tests cover: missing txHex, wrong outer `to`, wrong outer `value`, mismatched inner address, and wrong inner selector.

modules/sdk-coin-eth/test/unit/eth.ts
kamleshmugdiya: Added happy-path + one negative case in `sdk-coin-arbeth`, `sdk-coin-opeth`, and `sdk-coin-polygon` to confirm the abstract verifier picks up each coin's batcher contract address correctly. Each module now has a `should verify a batch txPrebuild whose inner calldata matches txParams.recipients` test plus a targeted rejection (mismatched address / amount / count) so coverage isn't only on Ethereum.

The verifier itself lives in `abstract-eth` and is shared by all EVM coins, so the comprehensive test matrix stays in `sdk-coin-eth`; the per-coin tests are smoke tests for the network-config wiring.

mullapudipruthvik (commented):
please resolve conflicts

PR #8779, opened 2026-05-14
chore: update audit exclusions for sanitize-html XSS

pranavjain97:
Remove stale exclusions GHSA-xq3m-2v4x-88gg and GHSA-rp42-5vxx-qpwr which are no longer reported by yarn audit. Add GHSA-rpr9-rxv7-x643 (sanitize-html xmp XSS, CVE-2026-44990, no upstream fix yet); our usage strips all tags with allowedTags: [] and outputs plain text in error messages, so there is no DOM rendering path.

### DIFF
--- a/.iyarc
+++ b/.iyarc
@@ -68,22 +68,10 @@ GHSA-9ppj-qmqm-q256
 GHSA-2w8x-224x-785m
 
 # Excluded because:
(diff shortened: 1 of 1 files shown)


gokulhost (approved):
approving per the PR description stating why the removed ones are stale

PR #7332, opened 2025-10-24
feat(statics): gating eth:anon for Dubai

aditya-bitgo:
ticket: WIN-6072

PR #8784, opened 2026-05-15
feat(hbar): skip amount validation for claim rewards self-transfer

Doddanna17:
## Summary
- Add `isHbarClaimRewards()` helper to `stakingWallet.ts` (follows existing `isBtcUndelegate()`, `isTrxStaking()`, `isStx()` pattern)
- Skip amount mismatch error in `validateBuiltStakingTransaction()` when HBAR claim rewards self-transfer is detected (sender address == recipient address)

## Problem
After merging the wire format fix (PR #8774), HBAR claim rewards fails post-signing:
```
Staking transaction validation failed before signing: Recipient 0.0.8933725 amount mismatch. Expected: 1, Got: 0
```

**Root cause**: HBAR claim rewards is a 1-tinybar self-transfer. The wire format correctly merges `[{acct, -1}, {acct, +1}]` into `[{acct, 0}]`, but `validateBuiltStakingTransaction()` compares `buildParams.recipients[0].amount` ("1") against `explainedTransaction.outputs[0].amount` ("0") and throws.

## Why BitGoJS (not multi-repo fix)
| Factor | This approach | Alternative (amount=0 end-to-end) |
|--------|--------------|----------------------------------|
| Files changed | 1 file | 3+ files across 2 repos |
| Base class risk | None | Modifies `BaseClaimRewardsRequestStrategy` (all coins) |
| Precedent | BTC/TRX/STX have coin-specific exemptions | No precedent |
| Semantic correctness | Both "1" and "0" are correct at their layer | Amount "0" is wrong -- we DO send 1 tinybar |

## Test plan
- [ ] Existing `stakingWalletNonTSS.ts` tests pass (no regression)
- [ ] HBAR claim rewards E2E on testnet passes post-signing validation
- [ ] Normal HBAR transfers still validated (address mismatch still caught)

Ticket: SI-605

### DIFF
--- a/modules/bitgo/test/v2/unit/staking/stakingWalletNonTSS.ts
+++ b/modules/bitgo/test/v2/unit/staking/stakingWalletNonTSS.ts
@@ -459,6 +459,136 @@ describe('non-TSS Staking Wallet', function () {
     });
   });
 
+  describe('HBAR Claim Rewards Validation', function () {
+    let thbarBaseCoin;
+    let thbarStakingWallet: StakingWallet;
+    let thbarStakingWalletData: WalletData;
+
+    before(function () {
+      thbarBaseCoin = bitgo.coin('thbar');
+      thbarStakingWalletData = {
+        approvalsRequired: 0,
+        balance: 0,
+        balanceString: '',
+        coinSpecific: {} as WalletCoinSpecific,
+        confirmedBalance: 0,
+        confirmedBalanceString: '',
+        keys: [],
+        label: '',
+        multisigType: 'onchain',
+        pendingApprovals: [],
+        spendableBalance: 0,
+        spendableBalanceString: '',
+        id: 'thbarStakingWalletId',
+        coin: 'thbar',
+        enterprise: enterprise.id,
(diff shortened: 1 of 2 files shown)


Doddanna17:
Putting this on hold — after analysis, this PR is likely **not needed**.

With PR #7315 (staking-service sends `amount="0"` in recipients) and PR #8774 (wire format merge fix already on master), both `buildParams.recipients[0].amount` and `explainedTransaction.outputs[0].amount` are `"0"` — so `validateBuiltStakingTransaction` passes without needing the skip logic here.

The remaining blocker was the WP `toBroadcastFormat` check, which is fixed in bitgo-microservices PR #58771.

Will revisit if E2E testing reveals any issues.

Doddanna17:
HBAR claim rewards feature has been dropped due to Hedera protocol limitations with self-transfers (memo-based addresses produce duplicate AccountIDs rejected by the network). All claim rewards code is being reverted across repos. See BitGoJS #8795, #8798, bitgo-microservices #58808, staking-service #7332.

PR #8789, opened 2026-05-18
fix: txHex needs to be set for the txn in case of TSS

bhuvanr159:
TICKET: CHALO-349

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -3526,6 +3526,22 @@ export class Wallet implements IWallet {
       delete prebuild.wallet;
       delete prebuild.buildParams;
 
+      // For TSS wallets the build endpoint returns only { txRequestId, stakingParams } — no txHex.
+      // Fetch the full txRequest to obtain serializedTxHex and populate txHex so that
+      // verifyTransaction (called inside prebuildAndSignTransaction) has the transaction bytes
+      // it needs. This mirrors what prebuildTransactionTxRequests does for other tx types.
+      if (this._wallet.multisigType === 'tss' && !prebuild.txHex && prebuild.txRequestId) {
+        const txRequest = await getTxRequest(this.bitgo, this.id(), prebuild.txRequestId, params.reqId);
+        const unsignedTx =
+          txRequest.apiVersion === 'full' ? txRequest.transactions?.[0]?.unsignedTx : txRequest.unsignedTxs?.[0];
+        if (!unsignedTx?.serializedTxHex) {
+          throw new Error(
+            `Expected serializedTxHex on TSS resource management prebuild for txRequestId ${prebuild.txRequestId}`
+          );
+        }
+        prebuild = _.extend({}, prebuild, { txHex: unsignedTx.serializedTxHex });
+      }
+
       prebuild = _.extend({}, prebuild, { walletId: this.id() });
       debug('final resource management transaction prebuild: %O', prebuild);
       prebuilds.push(prebuild);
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3526,6 +3526,19 @@ export class Wallet implements IWallet {
       delete prebuild.wallet;
       delete prebuild.buildParams;
 
+      // For TSS wallets the build endpoint returns only { txRequestId, stakingParams } — no txHex.
+      // Fetch the full txRequest to obtain serializedTxHex and populate txHex so that
+      // verifyTransaction (called inside prebuildAndSignTransaction) has the transaction bytes
+      // it needs. This mirrors what prebuildTransactionTxRequests does for other tx types.
+      if (!prebuild.txHex && (prebuild as any).txRequestId && this._wallet.multisigType === 'tss') {
```
bhavidhingra: let's move the tss wallet check in the front
```suggestion
      if (this._wallet.multisigType === 'tss' && !prebuild.txHex && (prebuild as any).txRequestId) {
```

bhavidhingra: can we avoid `as any`?
here and at line `3534`

pritam-gembali (changes requested):
unit tests?

bhavidhingra (dismissed):

pritam-gembali (changes requested):
The `if (unsignedTx?.serializedTxHex)` guard at `modules/sdk-core/src/bitgo/wallet/wallet.ts:3534-3536` is a silent fallback. If the txRequest comes back in an unexpected shape, the assignment is skipped, no error is raised, and the prebuild is returned with `txHex` still missing — re-introducing the very bug this PR is fixing, just with an extra HTTP round-trip.

### Concrete failure mode

Imagine the backend returns `apiVersion: 'full'` but `transactions` is `[]` (or its first element has no `unsignedTx` field — e.g. a future API rename):

1. Build returns `{ txRequestId: 'req-1', stakingParams: {...} }` with no `txHex`.
2. We enter the new branch, fetch the txRequest. ✓
3. `txRequest.transactions?.[0]?.unsignedTx` → `undefined`.
4. `if (unsignedTx?.serializedTxHex)` → false. **Block silently skipped.**
5. Prebuild is returned without `txHex`.
6. Caller hits `prebuildAndSignTransaction` → `verifyTransaction` at `wallet.ts:2189`, which sees `txPrebuild?.txHex` is undefined and either skips verification or throws something downstream that looks unrelated (e.g. a coin-specific verify failing on `undefined.length`).

The user sees the downstream error and has no breadcrumb pointing back to the real cause seven steps earlier. Debugging means re-running with a debugger.

### Suggested fix

```ts
if (!unsignedTx?.serializedTxHex) {
  throw new Error(
    `Expected serializedTxHex on TSS resource management prebuild for txRequestId ${prebuild.txRequestId}`
  );
}
prebuild = _.extend({}, prebuild, { txHex: unsignedTx.serializedTxHex });
```

This matches the assertive style already used at `wallet.ts:4193-4202` in `prebuildTransactionTxRequests`, where missing/malformed txRequest shapes throw explicitly.

### Test to update

The new test `should leave txHex undefined when txRequest has no serializedTxHex` in `modules/sdk-core/test/unit/bitgo/wallet/resourceManagement.ts` pins the silent-skip behavior as intended. Once that test merges, future reviewers will assume the silent fallback is by design — making it harder to harden later. Please flip it to expect a thrown error:

```ts
await wallet.buildResourceDelegations({ delegations: [delegations[0]] })
  .should.be.rejectedWith(/Expected serializedTxHex/);
```

Happy to approve once the throw is in place and the test is updated.

bhavidhingra (approved):

pritam-gembali (approved):

PR #8793, opened 2026-05-18
feat(sdk-core): add EdDSA MPCv2 offline signing helper utils

vibhavgo:
Centralizes shared MPCv2 helper logic in BaseTssUtils for reuse across ECDSA and EdDSA, and moves misplaced helper tests from the EdDSA keychain creation test to a dedicated BaseTssUtils test file in sdk-core.

- Move getSignableHexAndDerivationPath() into BaseTssUtils for shared
  ECDSA and EdDSA MPCv2 transaction extraction
- Move validateAdata() into BaseTssUtils to eliminate duplicated
  authenticated data validation
- Move getBitgoAndUserGpgKeys() into BaseTssUtils with v1 (SJCL)
  and v2 (Argon2id) envelope support; accepts a userGpgKeyDomainSeparator arg
- Remove redundant private getBitgoAndUserGpgKeys() from EcdsaMPCv2Utils
  and EddsaMPCv2Utils; delegate to base class
- Remove unused MPS_DSG_SIGNING_USER_GPG_KEY constant and isV2Envelope
  import from EddsaMPCv2Utils
- Remove misplaced External Signing Helpers tests from createKeychains.ts
- Add sdk-core/test/unit/bitgo/utils/tss/baseTSSUtils.ts with lean,
  strictly typed tests for all extracted helpers and ECDSA delegation

Ticket: WCI-386

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
@@ -1015,10 +1015,10 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
     let derivationPath: string;
     let serializedTxHex: string | undefined;
     if (requestType === RequestType.tx) {
-      assert(txRequest.transactions && txRequest.transactions.length === 1, 'Unable to find transactions in txRequest');
-      txToSign = txRequest.transactions[0].unsignedTx.signableHex;
-      derivationPath = txRequest.transactions[0].unsignedTx.derivationPath;
-      serializedTxHex = txRequest.transactions[0].unsignedTx.serializedTxHex;
+      const signableTx = this.getSignableHexAndDerivationPath(txRequest, 'Unable to find transactions in txRequest');
+      txToSign = signableTx.signableHex;
+      derivationPath = signableTx.derivationPath;
+      serializedTxHex = signableTx.serializedTxHex;
     } else if (requestType === RequestType.message) {
       // TODO(WP-2176): Add support for message signing
       throw new Error('MPCv2 message signing not supported yet.');
@@ -1049,67 +1049,6 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
     const hashBuffer = hash.update(Buffer.from(txToSign, 'hex')).digest();
 
     return { hashBuffer, derivationPath };
-  }
-
-  /**
-   * Gets the BitGo and user GPG keys from the BitGo public GPG key and the encrypted user GPG private key.
-   * @param {string} bitgoPublicGpgKey  - the BitGo public GPG key
-   * @param {string} encryptedUserGpgPrvKey  - the encrypted user GPG private key
-   * @param {string} walletPassphrase  - the wallet passphrase
-   * @param {string} adata - the additional data to validate the GPG keys
-   * @returns {Promise<{ bitgoGpgKey: pgp.Key; userGpgKey: pgp.SerializedKeyPair<string> }>} - the BitGo and user GPG keys
-   */
-  private async getBitgoAndUserGpgKeys(
-    bitgoPublicGpgKey: string,
-    encryptedUserGpgPrvKey: string,
-    walletPassphrase: string,
-    adata: string
-  ): Promise<{
-    bitgoGpgKey: pgp.Key;
-    userGpgKey: pgp.SerializedKeyPair<string>;
-  }> {
-    const bitgoGpgKey = await pgp.readKey({ armoredKey: bitgoPublicGpgKey });
-    let decryptedGpgPrvKey: string;
-    if (isV2Envelope(encryptedUserGpgPrvKey)) {
-      decryptedGpgPrvKey = await this.bitgo.decryptAsync({ input: encryptedUserGpgPrvKey, password: walletPassphrase });
-    } else {
-      decryptedGpgPrvKey = this.bitgo.decrypt({ input: encryptedUserGpgPrvKey, password: walletPassphrase });
-    }
-    if (adata) {
-      this.validateAdata(adata, encryptedUserGpgPrvKey, EcdsaMPCv2Utils.DKLS23_SIGNING_USER_GPG_KEY);
-    }
-    const userDecryptedKey = await pgp.readKey({ armoredKey: decryptedGpgPrvKey });
-    const userGpgKey: pgp.SerializedKeyPair<string> = {
-      privateKey: userDecryptedKey.armor(),
-      publicKey: userDecryptedKey.toPublic().armor(),
-    };
-    return {
-      bitgoGpgKey,
-      userGpgKey,
-    };
-  }
-
-  /**
-   * Validates the adata and cyphertext.
-   * @param adata string
-   * @param cyphertext string
-   * @returns void
-   * @throws {Error} if the adata or cyphertext is invalid
-   */
-  private validateAdata(adata: string, cyphertext: string, roundDomainSeparator: string): void {
-    let cypherJson;
-    try {
-      cypherJson = JSON.parse(cyphertext);
-    } catch (e) {
-      throw new Error('Failed to parse cyphertext to JSON, got: ' + cyphertext);
-    }
-    // using decodeURIComponent to handle special characters
-    if (
-      decodeURIComponent(cypherJson.adata) !== decodeURIComponent(`${roundDomainSeparator}:${adata}`) &&
-      decodeURIComponent(cypherJson.adata) !== decodeURIComponent(adata)
-    ) {
-      throw new Error('Adata does not match cyphertext adata');
-    }
   }
 
   // #endregion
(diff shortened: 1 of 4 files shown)


modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -515,4 +527,81 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+  // #region private utils
+  /**
+   * Get the signable hex and derivation path from the transaction request.
+   * @param {TxRequest} txRequest - the transaction request object
+   * @returns {{ signableHex: string; derivationPath: string }} - the signable hex and derivation path
+   */
+  private getSignableHexAndDerivationPath(txRequest: TxRequest): {
+    signableHex: string;
+    derivationPath: string;
+  } {
```
Marzooqa: Is it possible to extract this? Lets have like a common utils for ecdsa and eddsa if applicable

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -515,4 +527,81 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+
+  /**
+   * Validates the adata and cyphertext.
+   * @param adata string
+   * @param cyphertext string
+   * @param roundDomainSeparator string
+   * @returns void
+   * @throws {Error} if the adata or cyphertext is invalid
+   */
+  private validateAdata(adata: string, cyphertext: string, roundDomainSeparator: string): void {
```
Marzooqa: This as well?

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -515,4 +527,81 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+   */
+  private getSignableHexAndDerivationPath(txRequest: TxRequest): {
+    signableHex: string;
+    derivationPath: string;
+  } {
+    assert(
+      txRequest.transactions && txRequest.transactions.length === 1,
+      'createOfflineShare requires exactly one transaction in txRequest'
+    );
+    const unsignedTx = txRequest.transactions[0].unsignedTx;
```
Marzooqa: Missing null-check on `unsignedTx` before accessing its properties. If `transactions[0].unsignedTx` is `null`/`undefined`, this throws a generic `TypeError` instead of the descriptive assert message. Add `assert(unsignedTx, 'Missing unsignedTx in transaction')` before the two property asserts.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -515,4 +527,81 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+    const bitgoGpgKey = await pgp.readKey({ armoredKey: bitgoPublicGpgKey });
+
+    let decryptedGpgPrvKey: string;
+    if (isV2Envelope(encryptedUserGpgPrvKey)) {
+      decryptedGpgPrvKey = await this.bitgo.decryptAsync({ input: encryptedUserGpgPrvKey, password: walletPassphrase });
+    } else {
+      decryptedGpgPrvKey = this.bitgo.decrypt({ input: encryptedUserGpgPrvKey, password: walletPassphrase });
+    }
+
+    this.validateAdata(adata, encryptedUserGpgPrvKey, EddsaMPCv2Utils.MPS_DSG_SIGNING_USER_GPG_KEY);
```
Marzooqa: Bug + inconsistency with ECDSA: `EcdsaMPCv2Utils.getBitgoAndUserGpgKeys` wraps this call with `if (adata) { ... }`. Without that guard, v1 SJCL ciphertexts encrypted without `adata` will always throw `"Adata does not match cyphertext adata"` when a non-empty `adata` is passed in — which breaks the v1 test in the test file.

Please add:
```ts
if (adata) {
  this.validateAdata(adata, encryptedUserGpgPrvKey, EddsaMPCv2Utils.MPS_DSG_SIGNING_USER_GPG_KEY);
}
```

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
Marzooqa: This is byte-for-byte identical to `EcdsaMPCv2Utils.validateAdata`. Both classes share the same `baseTSSUtils<KeyShare>` ancestor, so this should be lifted there to eliminate the duplication. This is the most straightforward extraction — same signature, same body, no ECDSA/EdDSA differences.

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
```diff
@@ -298,6 +298,253 @@ describe('TSS EdDSA MPCv2 Utils:', async function () {
...
+    describe('getBitgoAndUserGpgKeys', function () {
+      it('should decrypt v1 SJCL envelope and return GPG keys', async function () {
+        const passphrase = 'test-password';
+        const adata = 'test-adata';
+
+        // Encrypt user GPG private key with v1 SJCL (no adata for simplicity in v1)
+        const encryptedUserGpgPrvKey = bitgo.encrypt({
+          input: userGpgKeyPair.privateKey,
+          password: passphrase,
+        });
```
Marzooqa: This test will fail as written. The ciphertext is encrypted without `adata`, but `adata = 'test-adata'` is passed to `getBitgoAndUserGpgKeys` on line 398. Since `validateAdata` is called unconditionally (no `if (adata)` guard), it will find `cypherJson.adata` is `undefined`, fail both comparisons, and throw `"Adata does not match cyphertext adata"`.

Either:
1. Add the `if (adata)` guard in the implementation (preferred, consistent with ECDSA), or
2. Pass an empty string for `adata` in this test to reflect the v1 no-adata case.

vibhavgo: Will add adata guard in impl

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
vibhavgo: Extracted common

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
Marzooqa: This test is still broken after the update. The ciphertext is encrypted without `adata`, but `adata = 'test-adata'` (truthy) is passed to `getBitgoAndUserGpgKeys` on line 398. The new `if (adata)` guard only skips validation when `adata` is falsy — `'test-adata'` is not, so `validateAdata` still runs, finds `cypherJson.adata` is `undefined`/empty, fails both comparisons, and throws.

Fix: either encrypt with matching adata in this test, or pass `''` (empty string) to `getBitgoAndUserGpgKeys` to reflect the no-adata v1 case:
```ts
const result = await (tssUtils as any).getBitgoAndUserGpgKeys(
  bitgoGpgKeyPair.publicKey,
  encryptedUserGpgPrvKey,
  passphrase,
  ''  // no adata for v1 SJCL
);
```

vibhavgo: Yes
Addressed with two separate checks

modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/createKeychains.ts
```diff
@@ -298,6 +298,275 @@ describe('TSS EdDSA MPCv2 Utils:', async function () {
     });
   });
 
+  describe('External Signing Helpers', function () {
```
Marzooqa: These utility helper tests (`getSignableHexAndDerivationPath`, `validateAdata`, `getBitgoAndUserGpgKeys`) do not belong in `createKeychains.ts` — this file is for testing the keychain creation flow.

More importantly, these methods now live on `baseTSSUtils` in `sdk-core`, so they should be tested there, not through an EdDSA subclass in the `bitgo` module. The right place is a new dedicated file: `modules/sdk-core/test/unit/bitgo/utils/tss/baseTSSUtils.ts`, consistent with how `addressVerification.ts`, `ecdsaMPCv2.ts`, and `eddsaMPCv2.ts` each have their own focused files in that directory.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -1014,9 +1014,9 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
     let txToSign: string;
     let derivationPath: string;
     if (requestType === RequestType.tx) {
-      assert(txRequest.transactions && txRequest.transactions.length === 1, 'Unable to find transactions in txRequest');
-      txToSign = txRequest.transactions[0].unsignedTx.signableHex;
-      derivationPath = txRequest.transactions[0].unsignedTx.derivationPath;
+      const signableTx = this.getSignableHexAndDerivationPath(txRequest, 'Unable to find transactions in txRequest');
+      txToSign = signableTx.signableHex;
+      derivationPath = signableTx.derivationPath;
```
Marzooqa, line 1020: This refactor (delegating to `this.getSignableHexAndDerivationPath`) is not directly tested. The existing ECDSA `signTxRequest` integration tests exercise the full flow and would catch a catastrophic failure, but there is no targeted test for:

1. That `getHashStringAndDerivationPath` still returns the correct `hashBuffer` + `derivationPath` after the delegation
2. The new `unsignedTx` null-check path (from the base class) being exercised on the ECDSA route

This should be covered in the new `baseTSSUtils.ts` test file alongside the other extracted helper tests.

vibhavgo: Added in baseTSSUtils.ts

Marzooqa (approved):
Reviewed the behavioral changes for `getBitgoAndUserGpgKeys` in ECDSA — the refactoring is safe. `readKey` → `readPrivateKey` is strictly stricter with no behavioral change for valid inputs. Domain separator, `validateAdata` logic, and `SerializedKeyPair` reconstruction are all equivalent. Test subclass pattern (`TestBaseTssUtils`) is consistent with existing conventions in this repo (e.g. `bitgoMpcGpgPubKeys.ts`). LGTM.

PR #8795, opened 2026-05-18
revert(sdk-coin-hbar): remove buildTransferData merge logic from PR #8774

Doddanna17:
## Summary

- Reverts the merge logic in `coinTransferBuilder.ts` introduced by PR #8774 (commit `c27392747f`)
- The merge was collapsing transfer entries with the same accountID, which broke normal HBAR transfers where sender and recipient share the same base address (memo-based addresses like `0.0.X?memoId=Y`)
- The merge was unnecessary for claim rewards -- that flow uses a different builder path. The self-transfer fallback in `transaction.ts` `getTransferData()` (from PR #8742) already handles claim rewards correctly

## Root Cause

`buildHederaAccountID()` strips memo from addresses, so `0.0.A` and `0.0.A?memoId=1` produce the same proto AccountID key. The merge collapsed `[{acct, -N}, {acct, +N}]` into `[{acct, 0}]`, causing `verifyTransaction()` to fail with "Tx outputs does not match with expected txParams recipients" because parsed amount `"0"` != expected amount `"N"`.

## Test plan

- [ ] Verify normal HBAR transfer with memo-based address works (sender != recipient base address)
- [ ] Verify normal HBAR self-transfer with memo (sender == recipient base address) no longer hits SDK validation error
- [ ] Verify claim rewards self-transfer still works (fallback in `getTransferData()` handles 1-tinybar positive entry)
- [ ] Existing unit tests pass

Ticket: SI-605

### DIFF
--- a/modules/sdk-coin-hbar/src/lib/coinTransferBuilder.ts
+++ b/modules/sdk-coin-hbar/src/lib/coinTransferBuilder.ts
@@ -41,22 +41,8 @@ export class CoinTransferBuilder extends TransferBuilder {
     });
     accountAmounts[0].amount = Long.fromString(totalSend.toString()).negate(); // update sender send amount
 
-    // Merge entries with the same accountID to prevent ACCOUNT_REPEATED_IN_ACCOUNT_AMOUNTS.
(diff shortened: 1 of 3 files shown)


aditya-bitgo (approved):
lgtm

PR #8803, opened 2026-05-19
feat(sdk-core): added OFC BitGo signing on wallet and coins object

alextse-bg:
allow wallet and coins object to sign using the BitGo key if the passphrase is not provided during signing

Ticket: WCN-217-2

### DIFF
--- a/modules/sdk-core/src/bitgo/trading/tradingAccount.ts
+++ b/modules/sdk-core/src/bitgo/trading/tradingAccount.ts
@@ -11,11 +11,16 @@ export class TradingAccount implements ITradingAccount {
   private readonly enterpriseId: string;
 
   public wallet: IWallet;
+  public readonly userKeySigningRequired: boolean;
 
   constructor(enterpriseId: string, wallet: IWallet, bitgo: BitGoBase) {
     this.enterpriseId = enterpriseId;
     this.wallet = wallet;
     this.bitgo = bitgo;
+
+    const walletData = this.wallet.toJSON();
+    this.userKeySigningRequired =
+      walletData.coinSpecific?.userKeySigningRequired ?? walletData.userKeySigningRequired ?? true;
   }
 
   get id(): string {
@@ -48,10 +53,9 @@ export class TradingAccount implements ITradingAccount {
     params: Omit<SignPayloadParameters, 'walletPassphrase' | 'prv'>
   ): Promise<string> {
     const walletData = this.wallet.toJSON();
-    const userKeySigningRequired = walletData.coinSpecific?.userKeySigningRequired ?? walletData.userKeySigningRequired;
-    if (userKeySigningRequired) {
+    if (this.userKeySigningRequired) {
       throw new Error(
-        'Wallet must use user key to sign ofc transaction, please provide the wallet passphrase or visit your wallet settings page to configure one.'
+        'Wallet must use user key to sign ofc transaction, please provide the wallet passphrase, the private key, or visit your wallet settings page to configure one.'
       );
     }
     if (walletData.keys.length < 2) {
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -2137,7 +2137,7 @@ export class Wallet implements IWallet {
    * @param params
    * - txPrebuild
    * - [keychain / key] (object) or prv (string)
-   * - walletPassphrase
+   * - walletPassphrase (optional ONLY for OFC wallets with userKeySigningRequired = false)
    * - verifyTxParams (optional) - when provided, the transaction will be verified before signing
    *   - txParams: transaction parameters used for verification
    *   - verification: optional verification options
@@ -2264,6 +2264,17 @@ export class Wallet implements IWallet {
       };
       return params.customSigningFunction(signTransactionParamsWithSeed);
     }
+
+    if (this.baseCoin.getFamily() === 'ofc') {
+      const userKeySigningRequired = this.toTradingAccount().userKeySigningRequired;
+      const prv = userKeySigningRequired ? await this.getUserPrvAsync(presign as GetUserPrvOptions) : undefined;
+      return this.baseCoin.signTransaction({
+        ...signTransactionParams,
+        prv,
+        wallet: this,
+      });
+    }
+
     return this.baseCoin.signTransaction({
       ...signTransactionParams,
       prv: await this.getUserPrvAsync(presign as GetUserPrvOptions),
--- a/modules/sdk-core/src/coins/ofcToken.ts
+++ b/modules/sdk-core/src/coins/ofcToken.ts
@@ -9,17 +9,25 @@ import {
   SignTransactionOptions as BaseSignTransactionOptions,
   SignedTransaction,
   ITransactionRecipient,
+  IWallet,
 } from '../';
 import { isBolt11Invoice } from '../lightning';
 
 import { Ofc } from './ofc';
 
-export interface SignTransactionOptions extends BaseSignTransactionOptions {
-  txPrebuild: {
-    payload: string;
-  };
+type OfcAllowRemoteSignOptions = BaseSignTransactionOptions & {
+  txPrebuild: { payload: string };
+  wallet: IWallet;
+  prv?: string; // optional: forwarded to signPayload if present
+};
+
+type OfcLocalSignOptions = BaseSignTransactionOptions & {
+  txPrebuild: { payload: string };
   prv: string;
-}
+  wallet?: never; // excludes wallet from this branch
+};
+
+export type SignTransactionOptions = OfcAllowRemoteSignOptions | OfcLocalSignOptions;
 
 export { OfcTokenConfig };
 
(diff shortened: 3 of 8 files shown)


modules/sdk-core/src/coins/ofcToken.ts
```diff
@@ -18,7 +19,8 @@ export interface SignTransactionOptions extends BaseSignTransactionOptions {
   txPrebuild: {
     payload: string;
   };
-  prv: string;
+  prv?: string;
+  wallet?: Wallet;
 }
```
mohammadalfaiyazbitgo: Nit / follow-up suggestion: consider renaming this local interface from `SignTransactionOptions` to `OfcSignTransactionOptions` and re-exporting it from the top-level sdk-core index.

Currently the type is accessible only via the `coins` namespace (`coins.SignTransactionOptions`), and the generic name makes it indistinguishable from the base type at a glance.

A named export at the top level would let consumers do:

```ts
import { OfcSignTransactionOptions } from '@bitgo/sdk-core';
```

modules/sdk-core/src/bitgo/trading/tradingAccount.ts
```diff
@@ -22,6 +22,11 @@ export class TradingAccount implements ITradingAccount {
     return this.wallet.id();
   }
 
+  get userKeySigningRequired(): boolean {
```
mohammadalfaiyazbitgo: Nit: since `wallet.toJSON()` is a synchronous in-memory read and `TradingAccount` is constructed fresh on every `toTradingAccount()` call, this could be a constructor-assigned `readonly` field instead of a getter:

```ts
public readonly userKeySigningRequired: boolean;

constructor(enterpriseId: string, wallet: IWallet, bitgo: BitGoBase) {
  ...
  const walletData = wallet.toJSON();
  this.userKeySigningRequired =
    walletData.coinSpecific?.userKeySigningRequired
      ?? walletData.userKeySigningRequired
      ?? true;
}
```

This also eliminates a redundant `toJSON()` call — `signPayloadByBitGoKey` currently calls it twice (once explicitly for `walletData.keys`, and once implicitly through this getter).

The one caveat: if the wallet's data were refreshed mid-lifetime (e.g. via `wallet.get()`), a constructor field would be stale. Given the transient nature of `TradingAccount` this seems acceptable, but worth a note if there are plans to make the object longer-lived.

modules/sdk-core/src/coins/ofcToken.ts
```diff
@@ -18,7 +19,8 @@ export interface SignTransactionOptions extends BaseSignTransactionOptions {
   txPrebuild: {
     payload: string;
   };
-  prv: string;
+  prv?: string;
+  wallet?: Wallet;
```
mohammadalfaiyazbitgo: Two suggestions here:

1. **Use `IWallet` over `Wallet`** — `IWallet` is the interface and is available from the same `'../'` import. Depending on the concrete class couples this interface to the implementation unnecessarily.

2. **Model `prv` / `wallet` as a discriminated union** — the current shape allows both to be absent (TypeScript won't catch it; the runtime throw is the only guard). A union makes the constraint explicit:

```ts
type OfcRemoteSignOptions = BaseSignTransactionOptions & {
  txPrebuild: { payload: string };
  wallet: IWallet;
  prv?: string; // optional: forwarded to signPayload if present
};

type OfcLocalSignOptions = BaseSignTransactionOptions & {
  txPrebuild: { payload: string };
  prv: string;
  wallet?: never; // excludes wallet from this branch
};

export type SignTransactionOptions = OfcRemoteSignOptions | OfcLocalSignOptions;
```

With this shape, passing neither `wallet` nor `prv` becomes a compile-time error instead of a runtime throw, and the `wallet?: never` on the local branch prevents accidentally constructing an ambiguous object.

modules/sdk-core/src/coins/ofcToken.ts
```diff
@@ -107,15 +109,25 @@ export class OfcToken extends Ofc {
...
    */
   async signTransaction(params: SignTransactionOptions): Promise<SignedTransaction> {
     const txPrebuild = params.txPrebuild;
     const payload = txPrebuild.payload;
-    const signatureBuffer = (await this.signMessage(params, payload)) as any;
-    const signature: string = signatureBuffer.toString('hex');
+
+    let signature: string;
+    if (params.wallet) {
+      signature = await params.wallet.toTradingAccount().signPayload({ payload, prv: params.prv });
```
mohammadalfaiyazbitgo: Bug: `params.prv` is optional and can be `undefined` here. When it is, `signPayload` sees `!walletPassphrase && !prv → true` and routes to `signPayloadByBitGoKey`, which then checks `userKeySigningRequired` and throws with a confusing deep error if the wallet requires user-key signing.

A wallet with `userKeySigningRequired=true` passed here without a `prv` would fail inside `signPayloadByBitGoKey` rather than being caught cleanly at this level.

Suggest an explicit guard:

```ts
if (params.wallet) {
  if (!params.prv && params.wallet.toTradingAccount().userKeySigningRequired) {
    throw new Error('prv is required for wallets with userKeySigningRequired=true');
  }
  signature = await params.wallet.toTradingAccount().signPayload({ payload, prv: params.prv });
}
```

This is also why the discriminated union suggested above helps — `OfcRemoteSignOptions` (wallet, no prv) would only be valid when remote signing is actually permitted.

modules/sdk-core/src/coins/ofcToken.ts
alextse-bg: good point, this should be compile time safe as well. I will revise on the TS types.

modules/sdk-core/src/coins/ofcToken.ts
alextse-bg: I can make this a followup change

modules/sdk-core/src/coins/ofcToken.ts
```diff
@@ -107,15 +109,25 @@ export class OfcToken extends Ofc {
...
+    } else if (params.prv) {
+      const signatureBuffer = (await this.signMessage({ prv: params.prv }, payload)) as any;
```
mohammadalfaiyazbitgo: Avoid the `as any` cast — both `signMessage` overloads on `Ofc` are typed to return `Promise<Buffer>`, so TypeScript can resolve this directly:

```ts
const signatureBuffer: Buffer = await this.signMessage({ prv: params.prv }, payload);
signature = signatureBuffer.toString('hex');
```

Or more concisely:

```ts
signature = (await this.signMessage({ prv: params.prv }, payload)).toString('hex');
```

alextse-bg: this was existing code, not sure why this case was there at the first place. Fixed.

modules/sdk-core/src/bitgo/trading/tradingAccount.ts
alextse-bg: addressed

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -2264,6 +2264,17 @@ export class Wallet implements IWallet {
       };
       return params.customSigningFunction(signTransactionParamsWithSeed);
     }
+
+    if (this.baseCoin.getFamily() === 'ofc') {
+      const userKeySigningRequired = this.toTradingAccount().userKeySigningRequired;
+      const prv = userKeySigningRequired ? await this.getUserPrvAsync(presign as GetUserPrvOptions) : undefined;
```
mohammadalfaiyazbitgo, line 2270: When `userKeySigningRequired = true`, this calls `getUserPrvAsync(presign)` — but `presign` will only contain a `keychain` if the hot-wallet gate earlier in this method populated it. That gate is guarded by `this.type() === 'hot' || this.type() === undefined`, and OFC wallets have `type = 'trading'`, so it is always skipped.

The result: calling `wallet.signTransaction({ txPrebuild, walletPassphrase: '...' })` on an OFC wallet with `userKeySigningRequired = true` always fails with `"keychain must be an object"`, regardless of whether the passphrase is correct. The only way this path currently succeeds is if the caller also provides a pre-fetched `keychain` explicitly — which the test does, but `prebuildAndSignTransaction({ walletPassphrase })` callers would not.

Fix: either extend the hot-wallet keychain-fetch gate to cover `trading`-type OFC wallets, or add a dedicated keychain fetch inside this block before calling `getUserPrvAsync`.

modules/sdk-core/src/coins/ofcToken.ts
```diff
@@ -107,15 +115,30 @@ export class OfcToken extends Ofc {
...
   async signTransaction(params: SignTransactionOptions): Promise<SignedTransaction> {
     const txPrebuild = params.txPrebuild;
     const payload = txPrebuild.payload;
-    const signatureBuffer = (await this.signMessage(params, payload)) as any;
-    const signature: string = signatureBuffer.toString('hex');
+
+    let signature: string;
+    if (params.wallet) {
+      const tradingAccount = params.wallet.toTradingAccount();
+      if (!params.prv && tradingAccount.userKeySigningRequired) {
```
mohammadalfaiyazbitgo, line 130: Nit: message omits `walletPassphrase` as a valid option. The equivalent guard in `tradingAccount.signPayloadByBitGoKey` says "please provide the wallet passphrase, the private key, or visit...". Both are new in this PR — worth keeping them consistent.

mohammadalfaiyazbitgo (changes requested):
Blocking on the walletPassphrase gap in the wallet.ts OFC path — see inline comment. Also a minor error message inconsistency in ofcToken.ts.

modules/sdk-core/src/bitgo/wallet/wallet.ts
alextse-bg: Didn't change the logic here. This is following existing behavior.

Originally `signTransaction` on OFC wallet will fall through until the very last line in line 2278, calling the following: `return this.<ofcToken>.signTransaction({ prv: await this.getUserPrvAsync(presign), ... })`. This already will fail if keychain is not being passed in.

All this code change does is to prevent an unnecessary API call to fetch the user key if we are not forcing them to use the user key. All existing code that call this method on an OFC wallet must be passing the keychain already. 

Though I agree with the analysis. Seems weird that hot wallets are allowed not to pass in keychains but OFC wallet aren't when the SDK can fetch it anyways. I can add it.

alextse-bg: also prebuildAndSignTransaction should fetch the keychains and decrypt based on the passphrase already.

Though now that I have taken a look at the code, that function might be passing in the bitgoKey as the backupKeychain incorrectly. Will address this as a followup.

mohammadalfaiyazbitgo (approved):

modules/sdk-core/src/coins/ofcToken.ts
alextse-bg: coin method didn't allowed passphrase to be passed in before, so I didn't add it for ofcToken here. Let me know if we want to include that, if so I can fix the error message.

PR #8826, opened 2026-05-21
feat(abstract-eth): expose getSignablePayload for ETH coin classes (CGD-1083)

0xPrabh:
## Description

Overrides `getSignablePayload(serializedTx)` on `AbstractEthLikeCoin` so that ETH and all ETH-like coins return the correct 32-byte keccak256 signing digest, rather than falling through to the base-class no-op that returns raw serialized bytes.

This is required to unblock AKM's external signing endpoint (`POST /sign`), which needs the correct hash to sign — not the raw transaction bytes.

## Changes

- **`AbstractEthLikeCoin.getSignablePayload`** — new override: deserializes via the transaction builder, casts to `EthTransaction`, and returns `tx.signablePayload`
- **`EthLikeTransactionData` interface** — adds `getSignablePayload(): Buffer`
- **`EthTransactionData`** — implements `getSignablePayload()` via `tx.getMessageToSign(true)` (keccak256 hash, same primitive used internally for TSS signing)
- **`Transaction.signablePayload` getter** — overrides `BaseTransaction`'s `NotImplementedError` default; throws `InvalidTransactionError` if no transaction data is set
- **`EthLikeToken.getSignablePayload`** — preserves pre-PR behavior (returns raw serialized bytes); tokens are out of scope for this ticket
- Inherited by **all non-token `AbstractEthLikeNewCoins` descendants** (Eth, Polygon, Arbeth, Opeth, BSC, AvaxC, EVM, etc.) and legacy `AbstractEthLikeCoin` descendants (Rbtc, Celo, Etc)

## Issue

CGD-1083 (unblocks WCN-397)

## Type of change

## How Has This Been Tested?

Unit tests added in:
- `modules/abstract-eth/test/unit/transaction.ts` — throws on empty tx; returns 32-byte hash for unsigned tx
- `modules/sdk-coin-eth/test/unit/transaction.ts` — hash pinned against `txData.id` fixture for both Legacy and EIP1559; distinct-payload check between tx types
- `modules/sdk-coin-eth/test/unit/eth.ts` — end-to-end `coin.getSignablePayload(serializedTx)` for Legacy and EIP1559 via `Teth`

Existing TSS signing flows are unaffected — they use `signableHex` from the unsigned tx object directly (`abstractEthLikeNewCoins.ts:2444`), not `getSignablePayload`.

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeCoin.ts
+++ b/modules/abstract-eth/src/abstractEthLikeCoin.ts
@@ -27,7 +27,7 @@ import {
 } from '@bitgo/sdk-core';
 import BigNumber from 'bignumber.js';
 
-import { isValidEthAddress, KeyPair as EthKeyPair, TransactionBuilder } from './lib';
+import { isValidEthAddress, KeyPair as EthKeyPair, Transaction as EthTransaction, TransactionBuilder } from './lib';
 import { VerifyEthAddressOptions } from './abstractEthLikeNewCoins';
 import { auditEcdsaPrivateKey } from '@bitgo/sdk-lib-mpc';
 
@@ -228,6 +228,14 @@ export abstract class AbstractEthLikeCoin extends BaseCoin {
       changeAmount: '0', // account base does not make change
       fee: params.feeInfo,
     };
+  }
+
+  /** @inheritDoc */
+  async getSignablePayload(serializedTx: string): Promise<Buffer> {
+    const txBuilder = this.getTransactionBuilder();
+    txBuilder.from(serializedTx);
+    const tx = (await txBuilder.build()) as EthTransaction;
+    return tx.signablePayload;
   }
 
   /**
--- a/modules/abstract-eth/src/ethLikeToken.ts
+++ b/modules/abstract-eth/src/ethLikeToken.ts
@@ -399,6 +399,11 @@ export class EthLikeToken extends AbstractEthLikeNewCoins {
     return txPrebuild.coin === this.tokenConfig.coin && txPrebuild.token === this.tokenConfig.type;
   }
 
+  /** @inheritDoc */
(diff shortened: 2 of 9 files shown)


venkateshv1266 (changes requested):
## 🔴 Token coins throw on `getSignablePayload` — contradicts PR description

`modules/abstract-eth/src/ethLikeToken.ts:406-408`: `EthLikeToken.getTransactionBuilder()` is stubbed to `throw new Error('Method not implemented')`. Because `EthLikeToken extends AbstractEthLikeNewCoins extends AbstractEthLikeCoin`, it inherits the new override at `abstractEthLikeCoin.ts:234-239`, whose first line is `this.getTransactionBuilder()`. Calling `<token>.getSignablePayload(serializedTx)` now throws `'Method not implemented'`.

**Why this blocks:** The PR description explicitly lists tokens as inheritors of this behavior ("Inherited by all `AbstractEthLikeNewCoins` descendants (Eth, Polygon, Arbeth, Opeth, BSC, AvaxC, EVM, **tokens**, etc.)"). Today the base-class default at least returned a Buffer (broken UTF-8, but non-throwing); now any ERC-20 path through AKM's `/sign` will hard-fail with an opaque error.

**Suggested fix:** Either
- (a) override `getSignablePayload` on `EthLikeToken` to construct a vanilla ETH-coin builder — the wire-level tx is just an ETH tx, the token semantics live in the calldata; or
- (b) explicitly throw a clear `NotImplementedError` so callers can fall back, and update the PR description to exclude tokens.

In either case, please add a token unit test mirroring the new `eth.ts` end-to-end tests (e.g. `tdai`) so this regression is caught by CI.

venkateshv1266 (approved):

PR #8833, opened 2026-05-22
feat(sdk-core): add externalSigner createOfflineRound1Share handler

vibhavgo:
Add the EdDSA MPCv2 offline round-1 handler for external signer flows. It stores encrypted carry-over state for the next signing round.

- Generate the round-1 EdDSA MPCv2 signature share from a fresh DSG session.
- Persist DSG session state and user message payload in the round-1 session.
- Encrypt session and ephemeral GPG private key data with signing-context adata.
- Cover SJCL encryption, v2 envelopes, payload shape, and transaction guards.

Ticket: WCI-378

### DIFF
--- a/modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
+++ b/modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
@@ -1,7 +1,9 @@
 import * as assert from 'assert';
 import * as sinon from 'sinon';
 import * as pgp from 'openpgp';
+import { randomBytes } from 'crypto';
 import { EddsaMPSDsg, MPSComms, MPSUtil } from '@bitgo/sdk-lib-mpc';
+import * as sjcl from '@bitgo/sjcl';
 import {
   EddsaMPCv2SignatureShareRound1Input,
   EddsaMPCv2SignatureShareRound1Output,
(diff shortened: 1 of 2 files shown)


modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -338,6 +340,158 @@ describe('EdDSA MPS DSG helper functions', async () => {
...
+          txRequest: { ...txRequest, transactions: undefined } as unknown as TxRequest,
+          prv: userKeyShare.toString('base64'),
+          walletPassphrase,
+        }),
+      /Unable to find transactions in txRequest/
+    );
+  });
+});
+
+function bytesToWord(bytes?: Uint8Array | number[]): number {
```
Marzooqa, line 487: nit: `bytesToWord` is defined between the two `describe` blocks. It works (function declarations are hoisted) but reads oddly. The ECDSA test file places the same helper at the very bottom after all `describe` blocks — consider doing the same here for consistency.

modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -338,6 +340,158 @@ describe('EdDSA MPS DSG helper functions', async () => {
...
+    assert.strictEqual(encryptedRound1Session.adata, `MPS_DSG_SIGNING_ROUND1_STATE:${expectedAdata}`);
+    assert.strictEqual(encryptedUserGpgPrvKey.v, 2);
+    assert.strictEqual(encryptedUserGpgPrvKey.adata, `MPS_DSG_SIGNING_USER_GPG_KEY:${expectedAdata}`);
+
+    const sessionPayload = JSON.parse(encryptedRound1Session.input);
+    assert.ok(sessionPayload.dsgSession, 'dsgSession should be persisted for round 2');
+    assert.ok(sessionPayload.userMsgPayload, 'userMsgPayload should be persisted for round 2');
+  });
+
+  it('should propagate the tx-only guard when transactions are missing', async () => {
```
Marzooqa, line 474: Missing test case: `encryptedPrv` provided but is a v1 (SJCL) blob, not a v2 envelope. Currently the only SJCL path test omits `encryptedPrv` entirely. The condition `encryptedPrv !== undefined && isV2Envelope(encryptedPrv)` is `false` when `encryptedPrv` is a v1 string, so it falls through to `bitgo.encrypt` — worth an explicit test to guard against a future regression where the v2 detection logic changes.

Marzooqa (approved):
Implementation matches the WCI-378 spec exactly and follows the ECDSA pattern faithfully. No bugs found. Two minor suggestions inline.

modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
vibhavgo: Will take up in following PR

modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
vibhavgo: Will take up in following PR

PR #8834, opened 2026-05-22
feat(sdk-core): add externalSigner createOfflineRound2Share handler

vibhavgo:
Add the EdDSA MPCv2 offline round-2 handler for external signer flows. It restores the encrypted round-1 DSG session and produces the round-2 share plus encrypted carry-over state for round 3.

- Restore user DSG session from encrypted round-1 state and verify BitGo round-1 output from txRequest signature shares.
- Run WASM round 1 and build the round-2 signature share with user GPG.
- Encrypt round-2 session state with MPS_DSG_SIGNING_ROUND2_STATE adata.
- Support SJCL and v2 envelope decryption paths for round-1 session input.
- Add unit tests with signBitgoEddsaRound1 helper, happy path, and guards.

Ticket: WCI-477

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
@@ -46,6 +46,7 @@ import { EddsaMPCv2KeyGenSendFn, KeyGenSenderForEnterprise } from './eddsaMPCv2K
 export class EddsaMPCv2Utils extends BaseEddsaUtils {
   private static readonly MPS_DSG_SIGNING_USER_GPG_KEY = 'MPS_DSG_SIGNING_USER_GPG_KEY';
   private static readonly MPS_DSG_SIGNING_ROUND1_STATE = 'MPS_DSG_SIGNING_ROUND1_STATE';
+  private static readonly MPS_DSG_SIGNING_ROUND2_STATE = 'MPS_DSG_SIGNING_ROUND2_STATE';
 
   /** @inheritdoc */
   async createKeychains(params: {
@@ -534,6 +535,7 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
 
   // #region external signer
 
+  // #region Round1Share
   async createOfflineRound1Share(params: {
     txRequest: TxRequest;
     prv: string;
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -597,6 +599,112 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+    const useV2 = isV2Envelope(encryptedRound1Session);
+
+    const { bitgoGpgKey, userGpgPrvKey } = await this.getBitgoAndUserGpgKeys(
+      bitgoPublicGpgKey,
+      encryptedUserGpgPrvKey,
+      walletPassphrase,
+      adata,
+      EddsaMPCv2Utils.MPS_DSG_SIGNING_USER_GPG_KEY
+    );
+
```
Marzooqa, line 632: nit: This `assert(txRequest.transactions, ...)` is unreachable dead code. `getSignableHexAndDerivationPath` (called 16 lines above) already throws with the same message when `transactions` is `undefined` or empty, so execution never reaches here with a falsy `transactions`. The only new guard that actually adds value is the `length === 1` assertion on the next line — consider removing this first assertion.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -597,6 +599,112 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+      decryptedRound1Session = await this.bitgo.decryptAsync({
+        input: encryptedRound1Session,
+        password: walletPassphrase,
+      });
+    } else {
+      decryptedRound1Session = this.bitgo.decrypt({
+        input: encryptedRound1Session,
+        password: walletPassphrase,
+      });
+    }
```
Marzooqa, line 664: nit: `validateAdata` is called **after** decryption, inconsistent with the ECDSA `createOfflineRound2Share` which calls it **before** decryption. Not a functional bug (adata is read directly from the serialized ciphertext blob, independent of plaintext), but the ordering is less defensive — you do the decryption work before rejecting a tampered envelope. Consider moving this call above the `if (useV2) { ... }` decrypt block to match the ECDSA pattern.

modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -597,6 +599,112 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
...
+    };
+
+    const userDsg = new EddsaMPSDsg.DSG(MPCv2PartiesEnum.USER);
+    userDsg.restoreSession(dsgSession);
+    const userMsg1: MPSTypes.DeserializedMessage = {
+      from: MPCv2PartiesEnum.USER,
+      payload: new Uint8Array(Buffer.from(userMsgPayload, 'base64')),
+    };
+
+    const [userMsg2] = userDsg.handleIncomingMessages([userMsg1, bitgoDeserializedMsg1]);
```
Marzooqa, line 678: nit: The error message `'WASM round 1 produced no message'` is misleading here — we are inside `createOfflineRound2Share` and `handleIncomingMessages` is the **second** WASM invocation (the first was `initDsg` + `getFirstMessage` in round 1). Consider `'DSG handleIncomingMessages produced no round-2 output'` or similar to avoid confusion.

Marzooqa (dismissed):
Implementation is correct and consistent with the ECDSA pattern — correct use of `verifyPeerMessageRoundOne`, `handleIncomingMessages`, `ed25519` GPG, and `userMsgPayload` carry-over in `sessionPayload`. Three minor nits inline; none blocking.

Marzooqa (approved):
Implementation is correct — session restore, `userMsgPayload` carry-over, `handleIncomingMessages` round-2 output, and re-encryption with `ROUND2_STATE` adata all look good. Previous nits addressed. One minor dead code note: the `assert(Array.isArray(transactions) && transactions.length === 1, ...)` after `getSignableHexAndDerivationPath` is unreachable since that method already makes the same check — safe to remove in a follow-up.

Marzooqa (approved):
Implementation is correct — session restore, `userMsgPayload` carry-over, `handleIncomingMessages` round-2 output, and re-encryption with `ROUND2_STATE` adata all look good. Previous nits addressed. One minor dead code note: the `assert(Array.isArray(transactions) && transactions.length === 1, ...)` after `getSignableHexAndDerivationPath` is unreachable since that method already makes the same check — safe to remove in a follow-up.

vibhavgo:
> Implementation is correct — session restore, `userMsgPayload` carry-over, `handleIncomingMessages` round-2 output, and re-encryption with `ROUND2_STATE` adata all look good. Previous nits addressed. One minor dead code note: the `assert(Array.isArray(transactions) && transactions.length === 1, ...)` after `getSignableHexAndDerivationPath` is unreachable since that method already makes the same check — safe to remove in a follow-up.

From TypeScript narrowing perspective, the local assert is still useful, because TS does not understand that a previous helper call proved txRequest.transactions is defined.
src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts:635:29 - error TS18048: 'transactions' is possibly 'undefined'.

635     const signatureShares = transactions[0].signatureShares;

PR #8845, opened 2026-05-25
feat: add eddsa MPCv2 SMC utils

kisslove-dewangan:
TICKET: WCI-242

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
@@ -64,8 +64,6 @@ export class MPCv2SMCUtils {
       coin: payload.coin,
       ovc: payload.ovc,
       platform: {
-        walletGpgPubKeySigs: (result as EddsaMPCv2KeyGenRound1Response & { walletGpgPubKeySigs: string })
-          .walletGpgPubKeySigs,
         sessionId: result.sessionId,
         bitgoMsg1: result.bitgoMsg1,
       },
@@ -115,6 +113,8 @@ export class MPCv2SMCUtils {
       coin: payload.coin,
       ovc: payload.ovc,
       platform: {
+        // sessionId carried over from payload.platform; safe because the assert above
+        // guarantees it equals result.sessionId.
         ...payload.platform,
         commonPublicKeychain: result.commonPublicKeychain,
         bitgoMsg2: result.bitgoMsg2,
--- a/modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
@@ -1,0 +1,163 @@
+import assert from 'assert';
+import {
+  EddsaBitgoToOVC1Round1Response,
+  EddsaBitgoToOVC1Round2Response,
+  EddsaKeyCreationMPCv2StateEnum,
+  EddsaMPCv2KeyGenRound1Response,
+  EddsaMPCv2KeyGenRound2Response,
+  EddsaOVC1ToBitgoRound1Payload,
+  EddsaOVC2ToBitgoRound2Payload,
+  OVCIndexEnum,
+} from '@bitgo/public-types';
+import { IBaseCoin } from '../../../../baseCoin';
+import { BitGoBase } from '../../../../bitgoBase';
+import { decodeOrElse, Keychain } from '../../../..';
+import { EddsaMPCv2Utils } from '../eddsaMPCv2';
+import { EddsaMPCv2KeyGenSendFn, KeyGenSenderForEnterprise } from '../eddsaMPCv2KeyGenSender';
+
+export class MPCv2SMCUtils {
+  private MPCv2Utils: EddsaMPCv2Utils;
+
+  constructor(private bitgo: BitGoBase, private baseCoin: IBaseCoin) {
+    this.MPCv2Utils = new EddsaMPCv2Utils(bitgo, baseCoin);
+  }
+
+  public async keyGenRound1(
+    enterprise: string,
+    payload: EddsaOVC1ToBitgoRound1Payload
+  ): Promise<EddsaBitgoToOVC1Round1Response> {
+    return this.keyGenRound1BySender(KeyGenSenderForEnterprise(this.bitgo, enterprise), payload);
+  }
+
+  public async keyGenRound2(
+    enterprise: string,
+    payload: EddsaOVC2ToBitgoRound2Payload
+  ): Promise<EddsaBitgoToOVC1Round2Response> {
+    return this.keyGenRound2BySender(KeyGenSenderForEnterprise(this.bitgo, enterprise), payload);
+  }
+
+  public async keyGenRound1BySender(
+    senderFn: EddsaMPCv2KeyGenSendFn<EddsaMPCv2KeyGenRound1Response>,
+    payload: EddsaOVC1ToBitgoRound1Payload
+  ): Promise<EddsaBitgoToOVC1Round1Response> {
+    assert(
+      payload.state === EddsaKeyCreationMPCv2StateEnum.WaitingForBitgoRound1Data,
+      `Invalid state for round 1, expected: ${EddsaKeyCreationMPCv2StateEnum.WaitingForBitgoRound1Data}, got: ${payload.state}`
+    );
+    decodeOrElse(EddsaOVC1ToBitgoRound1Payload.name, EddsaOVC1ToBitgoRound1Payload, payload, (errors) => {
+      throw new Error(`error(s) parsing payload: ${errors}`);
+    });
+
+    const ovc1 = payload.ovc[OVCIndexEnum.ONE];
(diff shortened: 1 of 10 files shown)


modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
```diff
@@ -0,0 +1,163 @@
...
+      userMsg1: ovc1.ovcMsg1,
+      backupMsg1: ovc2.ovcMsg1,
+    });
+
+    const response = {
+      state: EddsaKeyCreationMPCv2StateEnum.WaitingForOVC1Round2Data,
+      tssVersion: payload.tssVersion,
+      walletType: payload.walletType,
+      coin: payload.coin,
+      ovc: payload.ovc,
```
Marzooqa, line 65: **Bug: unsafe `as` cast hiding a type definition gap.**

`EddsaMPCv2KeyGenRound1Response` is typed as `{ sessionId, bitgoMsg1 }` only — `walletGpgPubKeySigs` is not declared on it. The cast `(result as EddsaMPCv2KeyGenRound1Response & { walletGpgPubKeySigs: string })` works around this gap.

For ECDSA, `MPCv2KeyGenRound1Response` properly declares `walletGpgPubKeySigs: NonEmptyString` and is accessed directly without any cast.

Two scenarios:
1. **If the EdDSA `/mpc/keygen/round1` API _does_ return `walletGpgPubKeySigs`** → `EddsaMPCv2KeyGenRound1Response` in `@bitgo/public-types` needs to be updated to include the field, and this cast should be removed.
2. **If the EdDSA API does _not_ return this field** → `walletGpgPubKeySigs` will be `undefined` at runtime. If `EddsaBitgoToOVC1Round1Response.platform.walletGpgPubKeySigs` is a required `NonEmptyString` (as it is for the ECDSA equivalent), `decodeOrElse` will throw `"error(s) parsing response"` on every call, making `keyGenRound1BySender` permanently broken.

Either fix the type in `@bitgo/public-types` or remove `walletGpgPubKeySigs` from the platform construction if EdDSA doesn't use it.

modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
```diff
@@ -0,0 +1,163 @@
...
+    const ovc1 = payload.ovc[OVCIndexEnum.ONE];
+    const ovc2 = payload.ovc[OVCIndexEnum.TWO];
+    const sessionId = payload.platform.sessionId;
+    const result = await this.MPCv2Utils.sendKeyGenerationRound2BySender(senderFn, {
+      sessionId,
+      userMsg2: ovc1.ovcMsg2,
+      backupMsg2: ovc2.ovcMsg2,
+    });
+
+    assert.equal(sessionId, result.sessionId, 'Round 1 and round 2 session IDs do not match');
```
Marzooqa, line 98: **Pattern inconsistency with ECDSA for `sessionId` in round 2.**

ECDSA `keyGenRound2BySender` explicitly overwrites the session ID from the API result:
```typescript
platform: {
  ...payload.platform,
  sessionId: result.sessionId,  // updated from result
  bitgoCommitment2: ...,
}
```

This PR asserts equality (`assert.equal(sessionId, result.sessionId, ...)`) and then keeps the original via `...payload.platform` — the response never includes `result.sessionId` directly.

These two choices are consistent with _each other_ (assert equality, then propagate original), but it's unclear if the assertion is correct. If the EdDSA round 2 API returns a different `sessionId` — as the ECDSA code implies is possible (since it overwrites it) — this assertion will throw unexpectedly.

Please add a comment documenting why EdDSA session IDs are stable across rounds (if that's the API contract), or switch to the ECDSA pattern and remove the assertion.

Marzooqa (dismissed):
Overall structure is correct — the 2-round key-gen flow (no round 3) is intentionally different from ECDSA because MPS EdDSA DKG only requires 2 BitGo rounds. `uploadClientKeys` is a faithful copy of the ECDSA version. Three issues below; one is a potential runtime failure on every call.

**No tests added.** The ECDSA `MPCv2SMCUtils` also has no unit tests, but this is still a gap — at minimum `keyGenRound1BySender` and `keyGenRound2BySender` should have state-assertion failure tests and a happy-path test using a stub sender.

modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
```diff
@@ -0,0 +1,163 @@
...
+    });
+
+    const response = {
+      state: EddsaKeyCreationMPCv2StateEnum.WaitingForOVC1Round2Data,
+      tssVersion: payload.tssVersion,
+      walletType: payload.walletType,
+      coin: payload.coin,
+      ovc: payload.ovc,
+      platform: {
+        walletGpgPubKeySigs: (result as EddsaMPCv2KeyGenRound1Response & { walletGpgPubKeySigs: string })
```
Marzooqa: **Type gap on `walletGpgPubKeySigs`.** `EddsaMPCv2KeyGenRound1Response` is typed as `{ sessionId, bitgoMsg1 }` only in `@bitgo/public-types` — `walletGpgPubKeySigs` is absent from the io-ts codec. ECDSA's `MPCv2KeyGenRound1Response` declares it as a required `NonEmptyString` and accesses it directly (`result.walletGpgPubKeySigs`, no cast needed).

If the EdDSA round-1 endpoint returns this field, add it to `EddsaMPCv2KeyGenRound1Response` in `@bitgo/public-types` and drop the cast. If it doesn't, `walletGpgPubKeySigs` will be `undefined` at runtime — and if `EddsaBitgoToOVC1Round1Response.platform.walletGpgPubKeySigs` is a required `NonEmptyString` (as it is in the ECDSA codec), `decodeOrElse` will throw `"error(s) parsing response"` on every call.

modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
```diff
@@ -0,0 +1,163 @@
...
+      source: 'backup',
+      keyType: 'tss',
+      commonKeychain: backupCommonKeychain,
+      isMPCv2: true,
+    });
+
+    const [userKeychain, backupKeychain] = await Promise.all([userKeychainPromise, backupKeychainPromise]);
+    return { userKeychain, backupKeychain, bitgoKeychain };
+  }
+}
```
Marzooqa, line 163: No unit tests added. At minimum: (1) a state-guard failure test for each of the two `assert` guards (`payload.state !== WaitingForBitgoRound1Data` and `!== WaitingForBitgoRound2Data`), and (2) a happy-path test for `keyGenRound1BySender` and `keyGenRound2BySender` using a stub `senderFn` that returns a mock API response. ECDSA `MPCv2SMCUtils` also lacks tests, but that's not a reason to skip them here. Follow-up safe.

Marzooqa (commented):
See inline comments.

modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
kisslove-dewangan: @claude can you please create a linear ticket under WCI for this test case task as mentioned above. assigned the ticket to me

modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
kisslove-dewangan: resolved, removed unnecessary field

modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
kisslove-dewangan: added comment

modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
kisslove-dewangan: resolved

modules/sdk-core/src/bitgo/utils/tss/eddsa/SMC/utils.ts
kisslove-dewangan: added

lokesh-bitgo:
@claude Please do a thorough review of this PR.

Focus especially on:

- any bugs or incorrect logic in the changes
- whether these changes could impact existing flows or cause regressions
- whether the implementation is aligned with the requirement/design
- whether the changes are correct and complete
- missing edge cases, validation gaps, and test coverage gaps

Please call out concrete issues clearly and separate them by severity if possible.

lokesh-bitgo:
@claude

kisslove-dewangan:
note: API spec breaking is intentional as I've confirmed from change owner, thread: https://bitgo.slack.com/archives/C057BHBRG4B/p1779958910469849

lokesh-bitgo (approved):
Overall looks good. Please add happy path test for the keyGenRound2 (enterprise) enterprise wrapper in follow up PR.

PR #8850, opened 2026-05-26
CSHLD-906: Add bridgingParams support for BTC-to-sBTC bridging transactions

abhi-bitgo:
## Summary

Add `bridgingParams` to the BTC transaction build pipeline to support BTC-to-sBTC cross-chain bridging. The parameter flows through all wallet types (hot, cold, custodial, TSS).

**Linear**: [CSHLD-906](https://linear.app/bitgo/issue/CSHLD-906/add-bridgingparams-support-for-btc-to-sbtc-bridging-transactions)

## Changes

- Add shared `SbtcBridgingParams` and `BridgingParams` interfaces in `iWallet.ts`
- Add `bridgingParams` to `PrebuildTransactionOptions` for SDK callers
- Whitelist `bridgingParams` in `BuildParams.ts` for non-TSS `/tx/build` path`
- Add `BridgingParamsCodec` io-ts codec in `sendmany.ts`, reused in `prebuildAndSignTransaction.ts`
- Add unit tests for BuildParams whitelist and non-TSS `/tx/build`

## Test Plan

- Unit tests for `BuildParams` codec whitelist (bridgingParams passes, unknown fields stripped)
- Unit test for non-TSS BTC wallet verifying `bridgingParams` reaches `/tx/build` endpoint

### DIFF
--- a/modules/express/src/typedRoutes/api/v2/sendmany.ts
+++ b/modules/express/src/typedRoutes/api/v2/sendmany.ts
@@ -106,6 +106,27 @@ export const TokenEnablement = t.intersection([
 ]);
 
 /**
+ * sBTC bridging parameters codec
+ */
+export const SbtcBridgingParamsCodec = t.type({
+  /** Amount in satoshis to bridge */
+  amount: t.union([t.number, t.string]),
+  /** Stacks recipient address */
+  stacksRecipient: t.string,
+  /** Maximum fee in satoshis */
+  maxFee: t.union([t.number, t.string]),
+  /** Lock time for the bridging transaction */
+  lockTime: t.number,
+});
+
+/**
+ * Bridging parameters codec for cross-chain bridging transactions (e.g., BTC to sBTC).
+ */
+export const BridgingParamsCodec = t.partial({
+  sbtc: SbtcBridgingParamsCodec,
+});
+
+/**
  * Request body for sending to multiple recipients (v2)
  *
  * This endpoint supports the full set of parameters available in the BitGo SDK
@@ -343,6 +364,9 @@ export const SendManyRequestBody = {
 
   /** Array of tokens to enable on the wallet */
   enableTokens: optional(t.array(TokenEnablement)),
+
+  /** Parameters for bridging transactions (e.g., BTC to sBTC). Used with type: 'bridging'. */
+  bridgingParams: optional(BridgingParamsCodec),
 
   /** Low fee transaction ID (for CPFP - Child Pays For Parent) */
   lowFeeTxid: optional(t.string),
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -2051,6 +2051,10 @@ export class Wallet implements IWallet {
       return this.prebuildTransactionTxRequests(params);
     }
 
+    if (params.type === 'bridging') {
+      assert(params.bridgingParams, "'bridgingParams' is required for bridging intent");
+    }
+
     // Whitelist params to build tx
     const whitelistedParams = this.baseCoin.preprocessBuildParams(_.pick(params, this.prebuildWhitelistedParams()));
     debug('prebuilding transaction: %O', whitelistedParams);
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -1710,6 +1710,55 @@ describe('V2 Wallet:', function () {
       postProcessStub.restore();
     });
 
+    it('should pass bridgingParams to tx/build for non-TSS BTC wallet', async function () {
+      const bridgingParams = {
+        sbtc: {
+          amount: 100000,
+          stacksRecipient: 'SP2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKNRV9EJ7',
+          maxFee: 5000,
+          lockTime: 144,
+        },
+      };
+      const expectedBuildParams = {
+        type: 'bridging',
+        txFormat: 'psbt',
+        recipients: [],
+        bridgingParams,
+        changeAddressType: ['p2trMusig2', 'p2wsh', 'p2shP2wsh', 'p2sh', 'p2tr'],
+      };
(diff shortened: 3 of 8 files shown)


mukeshsp (dismissed):

pritam-gembali (commented):
Nice, small, additive change. Posted inline comments on a couple of follow-ups worth addressing before merge — primarily an assert so a `bridging` request without `bridgingParams` fails loudly instead of silently no-op'ing, and a codec-level decode test for `BridgingParamsCodec` so the Express contract is enforced by tests, not just types.

The broader design (namespaced `BridgingParams.sbtc`) is good — leaves room for future bridge targets without breaking changes. Risk is low since no existing intent paths change behavior.

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -4162,6 +4162,21 @@ export class Wallet implements IWallet {
...
         break;
+      case 'bridging':
+        txRequest = await this.tssUtils!.prebuildTxWithIntent(
+          {
+            reqId,
+            intentType: 'bridging',
+            sequenceId: params.sequenceId,
+            comment: params.comment,
+            recipients: params.recipients || [],
+            bridgingParams: params.bridgingParams,
```
pritam-gembali: **Missing validation.** If a caller invokes `prebuildTransaction({ type: 'bridging' })` without `bridgingParams`, this happily forwards `bridgingParams: undefined` to `prebuildTxWithIntent`, which then gets silently dropped by the conditional spread in `mpcUtils.populateIntent` (the `&& params.bridgingParams` guard). Result: a `bridging` intent reaches the platform with no bridging params — likely a confusing downstream error rather than a clear one.

Recommend adding an assert here matching the style used elsewhere in `populateIntent`:

```ts
assert(params.bridgingParams, "'bridgingParams' is required for bridging intent");
```

Either here or in `mpcUtils.populateIntent` works — `populateIntent` is probably the better central spot since it would also cover the non-TSS path.

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
```diff
@@ -183,6 +183,7 @@ export abstract class MpcUtils {
         'transferOfferWithdrawn',
         'bridgeFunds',
         'cantonCommand',
+        'bridging',
```
pritam-gembali: Adding `bridging` to the recipients-exclusion list means recipients won't be required for this intent type. Tests and the new `case 'bridging'` in `wallet.ts` always pass `recipients: []`, which is consistent — but the PR description doesn't make it explicit whether bridging *never* carries a recipient (e.g., for a refund/return address) or *sometimes* does.

If it's truly always empty, worth a one-line comment here so future maintainers don't have to reverse-engineer it. If it can carry a recipient, this exclusion may be wrong.

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
```diff
@@ -222,6 +223,7 @@ export abstract class MpcUtils {
       recipients: intentRecipients,
       tokenName: params.tokenName,
       isTestTransaction: params.isTestTransaction,
+      ...(params.intentType === 'bridging' && params.bridgingParams ? { bridgingParams: params.bridgingParams } : {}),
```
pritam-gembali: The `&& params.bridgingParams` guard makes this a silent no-op when callers send `intentType: 'bridging'` without params (see related comment on `wallet.ts:4173`). Consider asserting `params.bridgingParams` is present when `intentType === 'bridging'` — putting it at the top of this function (alongside the existing recipients/cantonCommandParams asserts) would catch the gap on both TSS and non-TSS paths.

modules/express/src/typedRoutes/api/v2/sendmany.ts
```diff
@@ -105,6 +105,27 @@ export const TokenEnablement = t.intersection([
   }),
 ]);
 
+/**
+ * sBTC bridging parameters codec
+ */
+export const SbtcBridgingParamsCodec = t.type({
```
pritam-gembali, line 111: Two follow-ups on the new codec:

1. **No decode-validation tests.** The `BuildParams` test only verifies passthrough as `t.unknown` — it doesn't exercise the stricter shape this codec enforces. Worth adding a test that valid input round-trips and that missing/wrong-typed fields fail decode, so the Express contract is locked in by tests rather than just types.
2. **Codec / TS interface duplication.** `SbtcBridgingParams` in `iWallet.ts` and `SbtcBridgingParamsCodec` here define essentially the same shape twice. Consider deriving the TS type via `t.TypeOf<typeof BridgingParamsCodec>` and exporting from a shared module (the import from `./sendmany` in `prebuildAndSignTransaction.ts:12` already hints this should live somewhere more neutral). Not blocking — standard pattern in this repo — but reduces drift risk.
3. **Permissive numeric strings.** `amount`/`maxFee` accept arbitrary strings with no format check. Platform/WP presumably validates downstream, but at the Express boundary this is the only validation layer. Worth confirming WP rejects malformed values, or tightening to a numeric-string check.

modules/sdk-core/src/bitgo/wallet/wallet.ts
veetragjain: Done

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
veetragjain: Removed the TSS flow, it's only required for multisig BTC wallets

modules/sdk-core/src/bitgo/utils/mpcUtils.ts
veetragjain: Same here, TSS flow is removed

modules/express/src/typedRoutes/api/v2/sendmany.ts
veetragjain: Added decode validation tests and WP does reject the malformed values of amount and maxFee

pritam-gembali (approved):

mukeshsp (approved):

PR #8853, opened 2026-05-26
feat(sdk-core): add externalSigner createOfflineRound3Share handler

vibhavgo:
Add the EdDSA MPCv2 offline round-3 handler for external signer flows. It restores the encrypted round-2 DSG session and emits the final user signature share without persisting further signing state.

- Restore DSG state from the encrypted round-2 session payload.
- Verify BitGo round-2 output from txRequest signature shares before advancing DSG.
- Generate the final EdDSA MPCv2 round-3 signature share for BitGo.
- Cover happy path, v2 envelopes, adata tampering, missing BitGo share, and transaction guard cases in unit tests.

Ticket: WCI-478

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
@@ -707,6 +707,100 @@ export class EddsaMPCv2Utils extends BaseEddsaUtils {
   }
   // #endregion
 
+  // #region Round3Share
+  async createOfflineRound3Share(params: {
+    txRequest: TxRequest;
+    walletPassphrase: string;
+    bitgoPublicGpgKey: string;
+    encryptedUserGpgPrvKey: string;
+    encryptedRound2Session: string;
+  }): Promise<{
+    signatureShareRound3: SignatureShareRecord;
+  }> {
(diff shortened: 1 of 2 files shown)


Marzooqa (dismissed):
Implementation is correct — session restore, `userMsgPayload` carry-over, type-aware reverse scan for BitGo's round-2 share (necessary since both round-1 and round-2 BitGo shares are present by this point), and no session re-encryption on the final round all look good. Test coverage mirrors round 2 exactly. One minor nit carried over from #8834: the `assert(Array.isArray(transactions) && transactions.length === 1, ...)` after `getSignableHexAndDerivationPath` is dead code — safe to remove in a follow-up.

vibhavgo:
> One minor nit carried over from https://github.com/BitGo/BitGoJS/pull/8834: the assert(Array.isArray(transactions) && transactions.length === 1, ...) after getSignableHexAndDerivationPath is dead code — safe to remove in a follow-up.

From TypeScript narrowing perspective, the local assert is still useful, because TS does not understand that a previous helper call proved txRequest.transactions is defined.
src/bitgo/utils/tss/eddsa/eddsaMPCv2.ts:635:29 - error TS18048: 'transactions' is possibly 'undefined'.

635     const signatureShares = transactions[0].signatureShares;

Marzooqa (approved):

PR #8856, opened 2026-05-26
feat(statics): add XrpMptCoin class and token config for XRPL MPT

HimanshuS2212:
TICKET: CGD-1470


modules/statics/src/account.ts
```diff
@@ -607,6 +613,25 @@ export class XrpCoin extends AccountCoinToken {
...
+export class XrpMptCoin extends AccountCoinToken {
+  public readonly contractAddress: string; // MPTokenIssuanceID
+  public readonly canTransfer: boolean; // immutable — set at MPTokenIssuanceCreate
+  public readonly assetScale: number; // immutable — set at MPTokenIssuanceCreate
+
+  constructor(options: XrpMptCoinConstructorOptions) {
+    super({ ...options });
+    this.contractAddress = options.mptIssuanceId;
+    this.canTransfer = options.canTransfer;
+    this.assetScale = options.assetScale;
```
shivam-bitgo: this should be decimalPlaces?

HimanshuS2212: addressed

modules/sdk-coin-xrp/src/lib/utils.ts
```diff
@@ -223,13 +223,13 @@ class Utils implements BaseUtils {
...
    */
-  public isIssuedCurrencyAmount(amount: Amount): amount is IssuedCurrencyAmount {
+  public isIssuedCurrencyAmount(amount: Amount | MPTAmount): amount is IssuedCurrencyAmount {
     return (
       !!amount &&
       typeof amount === 'object' &&
-      typeof amount.currency === 'string' &&
-      typeof amount.issuer === 'string' &&
-      typeof amount.value === 'string'
+      typeof (amount as IssuedCurrencyAmount).currency === 'string' &&
```
mullapudipruthvik: this doesn't make sense 
amount is of type Amount | MPTAmount 
you should validate with this

modules/statics/src/account.ts
```diff
@@ -607,6 +613,22 @@ export class XrpCoin extends AccountCoinToken {
...
+ * Uses account_objects (not account_lines). No issuer::currency pattern.
+ * Named xrp:<token_name> — same pattern as trust-line tokens.
+ */
+export class XrpMptCoin extends AccountCoinToken {
+  public readonly contractAddress: string; // MPTokenIssuanceID
+  public readonly canTransfer: boolean; // immutable — set at MPTokenIssuanceCreate
+  constructor(options: XrpMptCoinConstructorOptions) {
+    super({ ...options });
+    this.contractAddress = options.mptIssuanceId;
+    this.canTransfer = options.canTransfer;
```
mullapudipruthvik, line 640: what about assetScale ?

modules/statics/src/index.ts
```diff
@@ -25,6 +25,9 @@ export {
   TaoCoin,
   PolyxCoin,
   XrpCoin,
+  XrpMptCoin,
+  xrpMptToken,
+  txrpMptToken,
```
mullapudipruthvik, line 30: follow the naming convention 
and also why do we have XrpMptCoin ? isn't this a token ?

mullapudipruthvik (commented):
can you add some tests

modules/statics/src/account.ts
HimanshuS2212: assetScale is passed as decimalPlaces in the factory (decimalPlaces: assetScale), so super() stores it as the inherited this.decimalPlaces. Storing it twice would be redundant use coin.decimalPlaces at runtime.

modules/statics/src/index.ts
HimanshuS2212: XrpMptCoin follows the existing convention  Erc20Coin, XrpCoin, SuiCoin are all tokens too. The separate class is needed so WP can route instanceof XrpMptCoin vs instanceof XrpCoin for MPT vs trust-line logic.

modules/sdk-coin-xrp/src/lib/utils.ts
HimanshuS2212: Fixed replaced the unsafe as IssuedCurrencyAmount cast with proper in narrowing:
`return typeof amount === 'object' && 'currency' in amount && 'issuer' in amount && 'value' in amount;`
Also added 5 unit tests covering all three cases  trust-line, native XRP string, and MPTAmount.

modules/statics/src/account.ts
```diff
@@ -607,6 +613,22 @@ export class XrpCoin extends AccountCoinToken {
...
+ * Identified by a 48-char hex MPTokenIssuanceID stored as contractAddress.
+ * Uses account_objects (not account_lines). No issuer::currency pattern.
+ * Named xrp:<token_name> — same pattern as trust-line tokens.
+ */
+export class XrpMptCoin extends AccountCoinToken {
+  public readonly contractAddress: string; // MPTokenIssuanceID
+  public readonly canTransfer: boolean; // immutable — set at MPTokenIssuanceCreate
+  constructor(options: XrpMptCoinConstructorOptions) {
+    super({ ...options });
+    this.contractAddress = options.mptIssuanceId;
```
shivam-bitgo, line 639: should we have any validation on the `mptIssuanceId`?

modules/statics/src/account.ts
```diff
@@ -3313,6 +3335,85 @@ export function txrpToken(
...
+      name,
+      fullName,
+      network,
+      mptIssuanceId,
+      canTransfer,
+      assetScale,
+      prefix,
+      suffix,
+      features,
+      decimalPlaces: assetScale, // assetScale IS the display decimal places — same concept as decimalPlaces elsewhere
```
shivam-bitgo, line 3389: dont we need any validation on assetScale?

modules/statics/src/account.ts
shivam-bitgo, line 640: will we check this canTrasfer before building the tx?

HimanshuS2212: Yes WP will reads coin.canTransfer from statics as a pre-flight before building the tx.

modules/statics/src/account.ts
```diff
@@ -607,6 +613,34 @@ export class XrpCoin extends AccountCoinToken {
...
+  public readonly canTransfer: boolean; // immutable — set at MPTokenIssuanceCreate
+
+  constructor(options: XrpMptCoinConstructorOptions) {
+    super({ ...options });
+
+    if (!/^[0-9a-fA-F]{48}$/.test(options.mptIssuanceId)) {
+      throw new InvalidContractAddressError(options.name, options.mptIssuanceId);
+    }
+
+    if (!Number.isInteger(options.assetScale) || options.assetScale < 0) {
```
shivam-bitgo: I think we should add upper bound as well here

HimanshuS2212: done

shivam-bitgo (approved):

mullapudipruthvik (approved):

mullapudipruthvik (approved):

modules/statics/src/account.ts
HimanshuS2212: added

modules/statics/src/account.ts
HimanshuS2212: added

PR #8874, opened 2026-05-27
feat(express): add qr param to sendmany, sendcoins, prebuildAndSign s…

davidkaplanbitgo:
…chemas

Allow the `qr` (quantum-resistant sweep) boolean parameter through Express typed route validation for all three transaction endpoints. Without this, Express strips the unknown field before forwarding to Wallet-Platform.

Ticket: T1-3417


pranishnepal (approved):
optional field

PR #8876, opened 2026-05-27
feat: add async counterparts for public encrypt/decrypt methods

bdesoky:
Ticket: WCN-284

This pull request adds async counterparts (of public-facing synchronous methods) that support v2 encrypt/decrypt as part of the Argon2 migration. It moves the bulk of the logic in these sync/async counterparts into helpers so as to minimize the opportunity for drift to occur between the sync/async versions (i.e. any future changes should go in the helpers so both sync/async methods inherit those changes) until the synchronous version are ready to be deprecated

**Key changes include:**

### Async/Await Refactoring and Test Modernization

* Refactored multiple test cases in `modules/bitgo/test/unit/bitgo.ts` to use async/await for secret sharing and authentication-related methods, replacing synchronous calls and improving error assertion clarity. Added a new test suite for async Shamir secret sharing methods. [[1]](diffhunk://#diff-d162df5ab23941770dfee7673cf2ae36a6eaae83ee2d3f5f2e44cbf54affeee2L240-R277) [[2]](diffhunk://#diff-d162df5ab23941770dfee7673cf2ae36a6eaae83ee2d3f5f2e44cbf54affeee2L284-R343) [[3]](diffhunk://#diff-d162df5ab23941770dfee7673cf2ae36a6eaae83ee2d3f5f2e44cbf54affeee2L439-R506)

* Updated wallet tests in `modules/bitgo/test/v2/unit/wallets.ts` to use async versions of `encryptPrvForUser` and added tests for new async methods `downloadKeycardAsync` and `encryptPrvForUserAsync`, including Node.js environment checks and output shape validation. [[1]](diffhunk://#diff-c022181f804241c92b5ab14fdce26284752681dc9731ddd0f1d6871e6054da83L4275-R4276) [[2]](diffhunk://#diff-c022181f804241c92b5ab14fdce26284752681dc9731ddd0f1d6871e6054da83R4331-R4404)

### Travel Rule Logic Improvements

* Refactored travel rule decryption and parameter preparation logic in `modules/sdk-api/src/v1/travelRule.ts`:
  - Introduced helper types and functions for validation and preparation.
  - Added async versions of `decryptReceivedTravelInfo` and `prepareParams`, supporting v2 encryption/decryption.
  - Improved modularity by extracting common logic and enhancing error handling. [[1]](diffhunk://#diff-f6b7c680bc6d1f7fb6504f1bbebb7daa1b9a380e46c663358bf80f1e5fdb5248L18-R29) [[2]](diffhunk://#diff-f6b7c680bc6d1f7fb6504f1bbebb7daa1b9a380e46c663358bf80f1e5fdb5248L104-R124) [[3]](diffhunk://#diff-f6b7c680bc6d1f7fb6504f1bbebb7daa1b9a380e46c663358bf80f1e5fdb5248L161-L164) [[4]](diffhunk://#diff-f6b7c680bc6d1f7fb6504f1bbebb7daa1b9a380e46c663358bf80f1e5fdb5248L180-R275)

### Improved Error Handling in Tests

* Simplified and modernized error assertions in `modules/sdk-api/test/unit/bitgoAPI.ts` to use concise `.should.throw` syntax, improving readability and reliability of test cases for key decryption.

---

These changes collectively improve code maintainability, test reliability, and readiness for future enhancements, especially around async cryptographic workflows.

### DIFF
--- a/modules/sdk-core/src/bitgo/internal/keycard.ts
+++ b/modules/sdk-core/src/bitgo/internal/keycard.ts
@@ -6,6 +6,7 @@
  */
 import { isUndefined } from 'lodash';
 import { Keychain } from '../keychain';
+import { EncryptFn, EncryptFnAsync } from '../../api';
 
 /**
  * Return the list of questions that will appear on the second page of the keycard
@@ -81,7 +82,7 @@ const generateQuestions = (coin: string) => {
 };
 
 interface GetKeyDataOptions {
-  encrypt: (params: { input: string; password: string }) => string;
+  encrypt: EncryptFn;
   userKeychain: Keychain;
   bitgoKeychain: Keychain;
   backupKeychain: Keychain;
@@ -92,40 +93,24 @@ interface GetKeyDataOptions {
   backupKeyID?: string;
 }
 
+type GetKeyDataAsyncOptions = Omit<GetKeyDataOptions, 'encrypt'> & {
+  encrypt: EncryptFnAsync;
+};
+
+interface BuildKeycardQrDataOptions {
+  userKeychain: Keychain;
+  bitgoKeychain: Keychain;
+  backupKeychain: Keychain;
+  coinShortName: string;
+  walletKeyID?: string;
+  backupKeyID?: string;
+}
+
 /**
- * Collect all data which will go onto the keycard
- * @param options
+ * Build QR code payload for each keycard box. Does not perform encryption.
  */
-function getKeyData(options: GetKeyDataOptions): any {
-  const {
-    encrypt,
-    userKeychain,
-    bitgoKeychain,
-    backupKeychain,
-    coinShortName,
-    passphrase,
-    passcodeEncryptionCode,
-    walletKeyID,
-    backupKeyID,
-  } = options;
-
-  // When using just 'generateWallet', we get back an unencrypted prv for the backup keychain
-  // If the user passes in their passphrase, we can encrypt it
-  if (backupKeychain.prv && passphrase) {
-    backupKeychain.encryptedPrv = encrypt({
-      input: backupKeychain.prv,
-      password: passphrase,
-    });
-  }
-
-  // If we have the passcode encryption code, create a box D with the encryptedWalletPasscode
-  let encryptedWalletPasscode;
-  if (passphrase && passcodeEncryptionCode) {
-    encryptedWalletPasscode = encrypt({
-      input: passphrase,
-      password: passcodeEncryptionCode,
-    });
-  }
+function buildKeycardQrData(options: BuildKeycardQrDataOptions, encryptedWalletPasscode: string | undefined): any {
+  const { userKeychain, bitgoKeychain, backupKeychain, coinShortName, walletKeyID, backupKeyID } = options;
 
   // PDF QR Code data
   const qrData: any = {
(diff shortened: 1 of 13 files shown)


modules/sdk-core/src/bitgo/internal/keycard.ts
```diff
@@ -382,3 +496,184 @@ export function drawKeycard(options: DrawKeycardOptions): any {
...
 }
+
+/**
+ * Async version of drawKeycard with support for v2 (Argon2id) encryption.
+ * Use this when the encrypt callback may return a Promise (e.g. encryptAsync).
+ *
+ * Draw a keycard into a new pdf document object
+ * @param options
+ */
+export async function drawKeycardAsync(options: DrawKeycardAsyncOptions): Promise<any> {
```
pranavjain97, line 452: have u tested what the v2 keycard looks like? Worth checking if it works for a recovery as well. (This can be done as part of seperate independent ticket too)

pranavjain97 (changes requested):
The ticket is for Phase 3 to deprecate the sync versions, but these changes are more so around creating async copy of all them, let's create a seperate ticket to track this. 

This PR duplicates 9 methods almost verbatim (sync -> async copy), with the only real difference being this.encrypt()/this.decrypt() vs await this.encryptAsync()/await this.decryptAsync(). These will need to be kept in sync until we fully deprecate the sync methods, which has no clear timeline right now.

A cleaner approach could be to parameterize the encrypt/decrypt dependency so the business logic lives in one place.

 For example, for splitSecret:

```
  // Single implementation accepting encrypt as a callback
  private async _splitSecretImpl(
    { seed, passwords, m }: SplitSecretOptions,
    encrypt: (params: { input: string; password: string }) => string | Promise<string>
  ): Promise<SplitSecret> {
    if (!Array.isArray(passwords)) throw new Error('passwords must be an array');
    if (!_.isInteger(m) || m < 2) throw new Error('m must be a positive integer greater than or equal to 2');
    if (passwords.length < m) throw new Error('passwords array length cannot be less than m');
    const n = passwords.length;
    const secrets = shamir.share(seed, n, m);
    const shards = await Promise.all(
      secrets.map((shard, i) => encrypt({ input: shard, password: passwords[i] }))
    );
    const node = bip32.fromSeed(Buffer.from(seed, 'hex'));
    return { xpub: node.neutered().toBase58(), m, n, seedShares: shards };
  }


  // Thin public wrappers
  splitSecret(opts: SplitSecretOptions): SplitSecret {
    // sync encrypt returns string, Promise.all resolves immediately
    return this._splitSecretImpl(opts, (p) => this.encrypt(p)) as unknown as SplitSecret;
  }

  async splitSecretAsync(opts: SplitSecretOptions): Promise<SplitSecret> {
    return this._splitSecretImpl(opts, (p) => this.encryptAsync(p));
  }
```

  Same pattern applies to reconstituteSecret, verifyShards, handleTokenIssuance, decryptKeys, encryptPrvForUser, etc. The keycard.ts functions (getKeyData, drawKeycard) already accept encrypt as a parameter -- they just need the type
  widened to () => string | Promise<string> and await added, rather than forking the entire function.

  This cuts the duplication significantly and eliminates the risk of the two copies diverging silently.

bdesoky:
hmm, the helper would need to be sync and then handle the async decrypt/encrypts with `.then()`, otherwise the synchronous functions would return a Promise

The issue with a single helper like this is that it introduces lots of fragile `instanceof` checks, and some logic is still duplicated (since we will need to use `.then()` and anything that relies on the `enryptAsync/decryptAsync` needs to be duplicated in the `.then()`)

Ive moved most of the logic in the sync/async methods into helpers. the sync/async methods are now just calling the helpers and `encrypt/encryptAsync` or `decrypt/decryptAsync`. This should help centralize the business logic while not introducing complicated async branching in a single internal function

modules/sdk-core/src/bitgo/internal/keycard.ts
bdesoky: Here is how a local pdf looks like, 
image

pranavjain97 (approved):

lcovar (approved):

davidkaplanbitgo (approved):

PR #8882, opened 2026-05-28
feat(express): add EdDSA MPCv2 external signer routing

vibhavgo:
Add Express routing for EdDSA MPCv2 external signer share generation. Wire MPCv2 wallet signing params to use the EdDSA round generators.

- Route EddsaMPCv2Round1/2/3 share requests to offline helpers
- Add EdDSA MPCv2 external signer generator factories
- Unify ECDSA and EdDSA MPCv2 response codecs into generic \`MPCv2Round{1,2,3}Response\` (both algorithms share identical round structures)
- Add EdDSA MPCv2 round response codec validation tests
- Cover route dispatch, generator requests, and signing-param wiring
- Avoid external signer test fixture rewrites during test imports

Ticket: WCI-380

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -8,6 +8,9 @@ import {
   CustomCommitmentGeneratingFunction,
   CustomGShareGeneratingFunction,
   CustomKShareGeneratingFunction,
+  CustomEddsaMPCv2SigningRound1GeneratingFunction,
+  CustomEddsaMPCv2SigningRound2GeneratingFunction,
+  CustomEddsaMPCv2SigningRound3GeneratingFunction,
   CustomMPCv2SigningRound1GeneratingFunction,
   CustomMPCv2SigningRound2GeneratingFunction,
   CustomMPCv2SigningRound3GeneratingFunction,
@@ -17,6 +20,7 @@ import {
   CustomSShareGeneratingFunction,
   EcdsaMPCv2Utils,
   EcdsaUtils,
+  EddsaMPCv2Utils,
   EddsaUtils,
   EncryptedSignerShareRecord,
   encryptRsaWithAesGcm,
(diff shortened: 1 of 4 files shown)


pranavjain97:
@claude

modules/express/src/typedRoutes/api/v2/generateShareTSS.ts
```diff
@@ -497,7 +498,33 @@ export const EcdsaMPCv2Round3Response = t.type({
...
+  /** Encrypted session state to continue to Round 3 */
+  encryptedRound2Session: t.string,
+});
+
+/** EdDSA MPCv2 Round 3 response with final signature share */
+export const EddsaMPCv2Round3Response = t.type({
+  /** Signature share for round 3 (final signature) */
+  signatureShareRound3: SignatureShareRecord,
+});
+
```
Marzooqa: Are these the same for ecdsa? If yes, we could make it generic for both curves and use it.

modules/express/src/clientRoutes.ts
```diff
@@ -455,18 +459,41 @@ export async function handleV2GenerateShareTSS(
...
+            return await eddsaMPCv2Utils.createOfflineRound1Share(req.body);
+          case ShareType.EddsaMPCv2Round2:
+            return await eddsaMPCv2Utils.createOfflineRound2Share(req.body);
+          case ShareType.EddsaMPCv2Round3:
+            return await eddsaMPCv2Utils.createOfflineRound3Share(req.body);
+          default:
+            throw new Error(
+              `Share type ${req.decoded.sharetype} not supported for EdDSA MPCv2, only EddsaMPCv2Round1, EddsaMPCv2Round2 and EddsaMPCv2Round3 is supported.`
+            );
+        }
```
Marzooqa, line 482: **`encryptedPrv` not forwarded — SJCL sessions only**

`CustomEddsaMPCv2SigningRound1GeneratingFunction` intentionally sends only `{ txRequest }` to the Express endpoint (passphrase stays server-side). As a result, `req.body.encryptedPrv` is never set, so `createOfflineRound1Share` always receives `encryptedPrv = undefined` → `useV2 = false` → SJCL session encryption regardless of the wallet's key format.

Once Argon2id migration completes, `encryptedPrv` will need to be added to the round-1 generator params and forwarded here. Follow-up safe for now.

vibhavgo: Created follow-up Linear ticket for this: https://linear.app/bitgo/issue/WCI-548/forward-encryptedprv-for-mpcv2-round-1-express-signing-after-argon2id

We will handle forwarding `encryptedPrv` after the Argon2id migration is complete.

modules/express/src/typedRoutes/api/v2/generateShareTSS.ts
vibhavgo: Yes. Will add generic type

vibhavgo: Will have to maintain backwards compatibility for preventing breaking changes.

Marzooqa (approved):
Routing logic, generator factories, and wallet version dispatch all look correct and consistent with ECDSA MPCv2. Tests cover all the key paths. Approving — inline notes are follow-up safe.

modules/express/src/typedRoutes/api/v2/generateShareTSS.ts
Marzooqa: Also noted: EcdsaMPCv2Round1/2/3Response are now aliases for MPCv2Round1/2/3Response but both appear in the GenerateShareTSSResponse union — the MPCv2Round* entries are unreachable since the deprecated aliases match first. Looks intentional for backward compat with existing consumers importing the old names. Worth a follow-up ticket to clean up the union once the deprecated aliases are removed.

Lets create a ticket to handle this

PR #8890, opened 2026-05-29
fix(sdk-core): preserve tss build params after presign

vibhavgo:
## Description

Preserve `txPrebuild.buildParams` across TSS presign flows in `sdk-core`.

Some coin level `presignTransaction` implementations can replace `txPrebuild` before TSS signing. For TSS wallets, `buildParams` carries SDK-local transaction intent metadata such as recipients and transaction type. If that metadata is dropped, downstream TSS signing and verification can receive empty or missing `txParams`, causing transaction verification failures.

This was observed in the `EdDSA MPCv2` Solana signing path. Solana’s presign flow can refresh the `txPrebuild` from the tx request and drop SDK-local `buildParams`. EdDSA MPCv2 then performs TSS signing with those missing params, which can cause recipient verification to fail before the transaction reaches the intended signing path.

This change restores the original `txPrebuild.buildParams` after presign when the presigned TSS prebuild omits them.

## Issue Number

WCI-505

## Type of change

### DIFF
--- a/modules/bitgo/test/v2/unit/wallet.ts
+++ b/modules/bitgo/test/v2/unit/wallet.ts
@@ -7,6 +7,7 @@ import * as sinon from 'sinon';
 import '../lib/asserts';
 import nock = require('nock');
 import * as _ from 'lodash';
+import * as assert from 'assert';
 
 import {
   BaseTssUtils,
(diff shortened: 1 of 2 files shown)


kisslove-dewangan (approved):
lgtm

PR #8891, opened 2026-05-29
feat(sdk-coin-xrp): add MPT transaction builders

HimanshuS2212:
TICKET: CGD-1470

### DIFF
--- a/modules/sdk-coin-xrp/src/lib/mpTokenAuthorizeBuilder.ts
+++ b/modules/sdk-coin-xrp/src/lib/mpTokenAuthorizeBuilder.ts
@@ -1,0 +1,85 @@
+import { BuildTransactionError, TransactionType } from '@bitgo/sdk-core';
+import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import { MPTokenAuthorize } from 'xrpl';
+import { XrpTransactionType } from './iface';
+import { Transaction } from './transaction';
+import { TransactionBuilder } from './transactionBuilder';
+import utils from './utils';
+
+export class MPTAuthorizeBuilder extends TransactionBuilder {
+  private _mptIssuanceId?: string;
+  private _mptHolder?: string;
+
+  constructor(_coinConfig: Readonly<CoinConfig>) {
+    super(_coinConfig);
+  }
+
+  protected get transactionType(): TransactionType {
+    return TransactionType.MPTokenAuthorize;
+  }
+
+  protected get xrpTransactionType(): XrpTransactionType.MPTokenAuthorize {
+    return XrpTransactionType.MPTokenAuthorize;
+  }
+
+  /**
+   * Set the MPTokenIssuanceID to authorize.
+   * @param {string} id - 48-character hex MPTokenIssuanceID
+   */
+  mptIssuanceId(id: string): this {
+    if (!/^[0-9a-fA-F]{48}$/.test(id)) {
+      throw new BuildTransactionError('MPTokenIssuanceID must be a 48-character hex string (192 bits)');
+    }
+    this._mptIssuanceId = id;
+    return this;
+  }
+
+  /**
+   * Set the Holder field for issuer-side authorization (Phase 2 only).
+   * Omit for standard holder self-authorization.
+   * @param {string} address - the holder account address (must be a valid XRP address)
+   */
+  mptHolder(address: string): this {
+    if (!utils.isValidAddress(address)) {
+      throw new BuildTransactionError('Invalid holder address: ' + address);
+    }
+    this._mptHolder = address;
+    return this;
+  }
+
+  initBuilder(tx: Transaction): void {
+    super.initBuilder(tx);
+    const { mptIssuanceId, mptHolder } = tx.toJson();
+    if (mptIssuanceId) {
+      this._mptIssuanceId = mptIssuanceId;
+    }
+    if (mptHolder) {
+      this._mptHolder = mptHolder;
+    }
+  }
+
+  /** @inheritdoc */
+  protected async buildImplementation(): Promise<Transaction> {
+    if (!this._sender) {
+      throw new BuildTransactionError('Sender must be set before building the transaction');
+    }
+    if (!this._mptIssuanceId) {
+      throw new BuildTransactionError('MPTokenIssuanceID must be set before building the transaction');
+    }
+
+    const authorizeFields: MPTokenAuthorize = {
+      TransactionType: this.xrpTransactionType,
+      Account: this._sender,
+      MPTokenIssuanceID: this._mptIssuanceId,
+    };
+
+    // Omit Holder for self-authorization — setting it causes XRPL rejection on holder self-auth.
+    if (this._mptHolder) {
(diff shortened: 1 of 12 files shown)


modules/sdk-coin-xrp/src/lib/utils.ts
```diff
@@ -221,13 +221,19 @@ class Utils implements BaseUtils {
...
   public isIssuedCurrencyAmount(amount: Amount | MPTAmount): amount is IssuedCurrencyAmount {
     return typeof amount === 'object' && 'currency' in amount && 'issuer' in amount && 'value' in amount;
   }
 
+  /**
+   * Determines if the provided `amount` is an MPT payment amount.
+   */
+  public isMPTAmount(amount: unknown): boolean {
+    return isMPTAmount(amount);
+  }
```
sachinBitgo, line 235: **`isMPTAmount` drops the type predicate**

The wrapper returns `boolean` instead of a type predicate, so TypeScript cannot narrow the type after the check. Every callsite in `transaction.ts` compensates with explicit `as MPTAmount` casts, which are unsafe — if the guard ever returns a false positive the cast silently proceeds.

The sibling method `isIssuedCurrencyAmount` is already typed correctly — `isMPTAmount` should match:

```typescript
// current
public isMPTAmount(amount: unknown): boolean {
  return isMPTAmount(amount);
}

// fix — also narrow the parameter type to match isIssuedCurrencyAmount
public isMPTAmount(amount: Amount | MPTAmount): amount is MPTAmount {
  return isMPTAmount(amount);
}
```

With the return type fixed, all the `as MPTAmount` casts in `transaction.ts` (lines 286, 368, 421) become unnecessary and can be removed.

sachinBitgo (dismissed):

modules/sdk-coin-xrp/src/lib/mpTokenAuthorizeBuilder.ts
```diff
@@ -0,0 +1,78 @@
+import { BuildTransactionError, TransactionType } from '@bitgo/sdk-core';
+import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import { MPTokenAuthorize } from 'xrpl';
+import { XrpTransactionType } from './iface';
+import { Transaction } from './transaction';
+import { TransactionBuilder } from './transactionBuilder';
+
+export class MPTokenAuthorizeBuilder extends TransactionBuilder {
```
sachinBitgo:
```suggestion
export class MPTAuthorizeBuilder extends TransactionBuilder {
```
or
```suggestion
export class MPTTokenAuthorizeBuilder extends TransactionBuilder {
```

0xPrabh:
@claude

modules/sdk-coin-xrp/src/lib/transaction.ts
```diff
@@ -83,18 +83,26 @@ export class Transaction extends BaseTransaction {
...
+          txData.mptAmount = paymentAmount;
         } else {
           throw new InvalidTransactionError('Invalid amount');
         }
         return txData;
+      }
+
+      case XrpTransactionType.MPTokenAuthorize: {
+        const mpTx = this._xrpTransaction as MPTokenAuthorize;
+        txData.mptIssuanceId = mpTx.MPTokenIssuanceID;
```
shivam-bitgo, line 103: should we also add `holder` field here?

modules/sdk-coin-xrp/src/lib/mpTokenAuthorizeBuilder.ts
```diff
@@ -0,0 +1,78 @@
...
+   * Set the Holder field for issuer-side authorization (Phase 2 only).
+   * Omit for standard holder self-authorization.
+   * @param {string} address - the holder account address
+   */
+  mptHolder(address: string): this {
+    this._mptHolder = address;
+    return this;
+  }
+
+  initBuilder(tx: Transaction): void {
```
shivam-bitgo, line 50: should we restore `holder` here from the tx

modules/sdk-coin-xrp/src/lib/mpTokenAuthorizeBuilder.ts
```diff
@@ -0,0 +1,78 @@
...
+    return this;
+  }
+
+  /**
+   * Set the Holder field for issuer-side authorization (Phase 2 only).
+   * Omit for standard holder self-authorization.
+   * @param {string} address - the holder account address
+   */
+  mptHolder(address: string): this {
+    this._mptHolder = address;
```
shivam-bitgo, line 46: any validation on `address`? like we have in `to()` method

modules/sdk-coin-xrp/src/lib/transactionBuilderFactory.ts
```diff
@@ -3,6 +3,8 @@ import { BaseCoin as CoinConfig } from '@bitgo/statics';
 import xrpl from 'xrpl';
 import { AccountDeleteBuilder } from './accountDeleteBuilder';
 import { AccountSetBuilder } from './accountSetBuilder';
+import { MPTokenAuthorizeBuilder } from './mpTokenAuthorizeBuilder';
```
shivam-bitgo: nit- naming inconsistency - `MPTokenAuthorizeBuilder` and `MptTransferBuilder`

modules/sdk-coin-xrp/src/lib/transaction.ts
```diff
@@ -152,7 +160,8 @@ export class Transaction extends BaseTransaction {
       const signablePayload = this.getSignablePayload();
 
       const xrpWallet = new xrpl.Wallet(pub, prv);
-      const signedTx = xrpWallet.sign(signablePayload, this._isMultiSig);
+      // Cast needed: MPTokenAuthorize is not in xrpl 4.6.0's Transaction union yet
```
shivam-bitgo: are we sure about this `MPTokenAuthorize` is not in xrpl 4.6.0? can we verify once

modules/sdk-coin-xrp/src/lib/transaction.ts
HimanshuS2212: done

modules/sdk-coin-xrp/src/lib/mpTokenAuthorizeBuilder.ts
HimanshuS2212: done

modules/sdk-coin-xrp/src/lib/mpTokenAuthorizeBuilder.ts
HimanshuS2212: done

modules/sdk-coin-xrp/src/lib/transaction.ts
HimanshuS2212: My bad MPTokenAuthorize is included in the Transaction union in XRPL 4.6.0. I removed the incorrect comment and simplified the cast from as unknown as xrpl.Transaction to as xrpl.Transaction.

shivam-bitgo (approved):

PR #8912, opened 2026-06-02
feat(sdk-core): add wallet.defi DeFi vault orchestration methods

hitansh-madan:
Add SDK-side orchestration for Galaxy × Morpho ERC-4626 vault deposits. Introduces wallet.defi.depositToVault() which sequences approve + deposit txRequests with fail-fast auto-cancel, plus resumeDeposit() for recovery, and getOperation()/listOperations() for status tracking.

Ticket: [CGD-1533](https://linear.app/bitgo/issue/CGD-1533)

### DIFF
--- a/modules/sdk-core/src/bitgo/defi/defiVault.ts
+++ b/modules/sdk-core/src/bitgo/defi/defiVault.ts
@@ -1,0 +1,233 @@
+/**
+ * @prettier
+ */
+import {
+  DefiOperation,
+  DefiOperationListResult,
+  DepositResult,
+  DepositToVaultOptions,
+  GetOperationOptions,
+  IDefiVault,
+  ListOperationsOptions,
+  ResumeDepositOptions,
+} from './iDefiVault';
+import { IWallet } from '../wallet';
+import { BitGoBase } from '../bitgoBase';
+
+/**
+ * Error thrown when a concurrent active deposit already exists for the (wallet, vault) pair.
+ */
+export class ActiveOperationExistsError extends Error {
+  public readonly operationId: string;
+
+  constructor(operationId: string) {
+    super(`An active deposit operation already exists: ${operationId}`);
+    this.name = 'ActiveOperationExistsError';
+    this.operationId = operationId;
+  }
+}
+
+/**
+ * Orchestrates ERC-4626 vault deposit and withdraw flows for a wallet.
+ *
+ * Exposed as `wallet.defi` on the Wallet class. See TDD §6.3.1 for the full
+ * design: the SDK sequences two sendMany calls (approve + deposit) and
+ * returns an operationId that the UI uses for status tracking and recovery.
+ *
+ * Uses wallet.sendMany() under the hood so that both custody wallets
+ * (txRequest creation only) and hot wallets (create + sign + broadcast)
+ * are handled by the existing infrastructure.
+ */
+export class DefiVault implements IDefiVault {
+  private readonly wallet: IWallet;
+  private readonly bitgo: BitGoBase;
+
+  constructor(wallet: IWallet) {
+    this.wallet = wallet;
+    this.bitgo = wallet.bitgo;
+  }
+
+  /**
+   * Deposit an amount of underlying asset into a vault.
+   *
+   * Internally issues two sendMany calls (approve + deposit) and returns the
+   * operationId that links them. If the deposit sendMany fails after
+   * the approve succeeds, the approve is auto-cancelled (fail-fast).
+   *
+   * @param params.vaultId - DeFi-service vault identifier
+   * @param params.amount - amount in base units of the underlying asset
+   * @param params.clientIdempotencyKey - optional client idempotency key
+   * @param params.walletPassphrase - required for hot wallets, omit for custody
+   */
+  async depositToVault(params: DepositToVaultOptions): Promise<DepositResult> {
+    if (!params.vaultId) {
+      throw new Error('vaultId is required');
+    }
+    if (!params.amount) {
+      throw new Error('amount is required');
+    }
+
+    // Layer-1 pre-flight: reject if an active deposit already exists for this (wallet, vault)
+    const activeOps: DefiOperationListResult = await this.bitgo
+      .get(this.bitgo.microservicesUrl(this.operationsUrl()))
+      .query({ vaultId: params.vaultId, state: 'active' })
+      .result();
+
+    if (activeOps.items && activeOps.items.length > 0) {
+      throw new ActiveOperationExistsError(activeOps.items[0].operationId);
+    }
+
+    // Step 1: Approve txRequest via sendMany
+    const approveResult = await this.wallet.sendMany({
+      type: 'defiApprove',
+      defiParams: {
+        vaultId: params.vaultId,
+        amount: params.amount,
+        ...(params.clientIdempotencyKey ? { clientIdempotencyKey: params.clientIdempotencyKey } : {}),
+      },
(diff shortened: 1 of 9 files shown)


modules/sdk-core/src/bitgo/defi/defiVault.ts
```diff
@@ -0,0 +1,231 @@
...
+      const depositTxRequest = await this.createTxRequest(depositIntent);
+      depositTxRequestId = depositTxRequest.txRequestId;
+    } catch (err) {
+      // Fail-fast: cancel the approve txRequest before throwing
+      try {
+        await this.cancelTxRequest(approveTxRequestId);
+      } catch {
+        // Best-effort cancel; the reconciler will clean up if this fails
+      }
+      throw err;
```
kamleshmugdiya: thoughts on this?
```suggestion
     await this.cancelTxRequest(approveTxRequestId);
```

modules/sdk-core/src/bitgo/defi/defiVault.ts
```diff
@@ -0,0 +1,231 @@
...
+  private async cancelTxRequest(txRequestId: string): Promise<void> {
+    await this.bitgo.del(this.bitgo.url('/wallet/' + this.wallet.id() + '/txrequests/' + txRequestId, 2)).result();
+  }
+
+  private operationsUrl(): string {
+    return `/api/defi-service/v1/wallets/${this.wallet.id()}/operations`;
+  }
+
+  private operationUrl(operationId: string): string {
+    return `/api/defi-service/v1/operations/${operationId}`;
```
kamleshmugdiya: I think this endpoint should also be be wallet.
without wallet scope, in the `resumeDeposit()` flow, it can potentially fetch operation from foreign wallet.

```suggestion
  private operationUrl(operationId: string): string {
    return `/api/defi-service/v1/wallets/${this.wallet.id()}/operations/${operationId}`;
```

modules/sdk-core/src/bitgo/defi/defiVault.ts
```diff
@@ -0,0 +1,231 @@
...
+      ...(params.clientIdempotencyKey ? { clientIdempotencyKey: params.clientIdempotencyKey } : {}),
+    };
+
+    const approveTxRequest = await this.createTxRequest(approveIntent);
+    const operationId = (approveTxRequest.intent as Record<string, unknown>)?.operationId as string;
+    const approveTxRequestId = approveTxRequest.txRequestId;
+
+    if (!operationId) {
+      throw new Error('operationId not found in approve txRequest response');
+    }
```
kamleshmugdiya, line 96: should we do cleanup if the operationId is not present in the response?

modules/sdk-core/src/bitgo/defi/defiVault.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  private async createTxRequest(intent: Record<string, unknown>): Promise<{ txRequestId: string; intent: unknown }> {
+    return await this.bitgo
+      .post(this.bitgo.url('/wallet/' + this.wallet.id() + '/txrequests', 2))
+      .send({ intent })
+      .result();
+  }
+
+  private async cancelTxRequest(txRequestId: string): Promise<void> {
+    await this.bitgo.del(this.bitgo.url('/wallet/' + this.wallet.id() + '/txrequests/' + txRequestId, 2)).result();
```
kamleshmugdiya: does this endpoint exist today in platform?
DELETE `/txrequests/{id}`

modules/sdk-core/src/bitgo/defi/defiVault.ts
```diff
@@ -0,0 +1,231 @@
...
+    if (params.state) query.state = params.state;
+    if (params.type) query.type = params.type;
+    if (params.limit) query.limit = params.limit;
+    if (params.cursor) query.cursor = params.cursor;
+
+    return await this.bitgo
+      .get(this.bitgo.microservicesUrl(this.vaultOperationsUrl(params.vaultId)))
+      .query(query)
+      .result();
+  }
```
kamleshmugdiya, line 199: `vaultId` redundant in the request
It's present in both URL path and query string.

modules/sdk-core/src/bitgo/defi/defiVault.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    // Step 1: Create approve txRequest
+    const approveIntent = {
+      intentType: 'defi-approve',
+      vaultId: params.vaultId,
+      amount: params.amount,
+      ...(params.clientIdempotencyKey ? { clientIdempotencyKey: params.clientIdempotencyKey } : {}),
+    };
+
+    const approveTxRequest = await this.createTxRequest(approveIntent);
```
kamleshmugdiya: instead of using `createTxRequest` can we use the `sendMany()` method?
tomorrow, if we need to extend the `depositToVault()` for hot wallets, it will be easier.

modules/sdk-core/src/bitgo/defi/defiVault.ts
hitansh-madan: updated. these calls now deviate from the TDD, but the original design had some redundancy and lacked the wallet scope as you mentioned, so I'll update the doc & defi side changes accordingly as well

modules/sdk-core/src/bitgo/defi/defiVault.ts
hitansh-madan: removed this as we discussed, the SDK shouldn't be responsible for clean-ups

modules/sdk-core/src/bitgo/defi/defiVault.ts
hitansh-madan: as discussed, the SDK shouldn't be responsible for clean-ups

modules/sdk-core/src/bitgo/defi/defiVault.ts
hitansh-madan: updated, we use the operations url now which only takes walletId as a path param

modules/sdk-core/src/bitgo/defi/defiVault.ts
hitansh-madan: removed, not needed anyways

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
```diff
@@ -280,6 +280,25 @@ interface IntentOptionsBase {
   custodianMessageId?: string;
 }
 
+/**
+ * DeFi intent fields for ERC-4626 vault operations.
+ * Extended by PopulatedIntent so these land at the top level of the payload.
+ */
+export interface DefiIntentFields {
+  vaultId?: string;
+  amount?: { value: string; symbol: string } | string;
```
kamleshmugdiya, line 289: why not?
```suggestion
  amount?: { value: string; symbol: string };
```

modules/sdk-core/src/bitgo/defi/defiVault.ts
hitansh-madan: using sendMany now as suggested

modules/sdk-core/src/bitgo/utils/tss/baseTypes.ts
hitansh-madan: the original intents define amount as a string and everywhere this value is consumed we deduce the symbol based on the vault. 
We can support symbol as input as well, but then we should also pass that along to all flows & validate that against the vault token in UI, WP & defi service, which requires all of these places to have context of the vault data. 
We can further discuss how we want to handle this

kamleshmugdiya (approved):

PR #8924, opened 2026-06-03
fix(sdk-core): enforce recipient verification in ECDSA TSS signing

mrdanish26:
## Summary

  Closes the ECDSA TSS default-path signing vulnerability (WCN-151 Phase 3)
  where a compromised BitGo API server could swap `signableHex` undetected.

## The Vulnerability
Both `EcdsaUtils` and `EcdsaMPCv2Utils` defaulted `txParams` when absent:

  ```ts
  // Before — vulnerable
  txParams: params.txParams || { recipients: [] }

  With no txParams (Classic UI pending approval, Express), verifyTransaction()
  gets an empty recipients list and skips all address/amount checks.

  Root cause of prior failures: stale hardcoded arrays in coin-level overrides
  were missing staking types, causing false throws on legitimate staking transactions.

  Loki is authoritative. Kibana misses wallet-platform-admin-nocluster entries.

  ┌────────┬────────────────┬──────┬──────────────────┐
  │  Coin  │    tx_type     │ Loki │      Status      │
  ├────────┼────────────────┼──────┼──────────────────┤
  │ canton │ enableToken    │  206 │ already exempt   │
  ├────────┼────────────────┼──────┼──────────────────┤
  │ sol    │ customTx       │  106 │ already exempt   │
  ├────────┼────────────────┼──────┼──────────────────┤
  │ canton │ createAccount  │   89 │ added in this PR │
  ├────────┼────────────────┼──────┼──────────────────┤
  │ sol    │ enableToken    │   22 │ already exempt   │
  ├────────┼────────────────┼──────┼──────────────────┤
  │ canton │ transferAccept │    8 │ added in this PR │
  ├────────┼────────────────┼──────┼──────────────────┤
  │ bsc    │ tokenApproval  │    6 │ already exempt   │
  ├────────┼────────────────┼──────┼──────────────────┤
  │ hash   │ contractCall   │    3 │ added in this PR │
  ├────────┼────────────────┼──────┼──────────────────┤
  │ ada    │ pledge         │    2 │ added in this PR │
  ├────────┼────────────────┼──────┼──────────────────┤
  │ apt    │ customTx       │    2 │ already exempt   │
  └────────┴────────────────┴──────┴──────────────────┘

  Changes

  Commit 1 — Enforce recipient verification

  - recipientUtils.ts (new) — 27-type NO_RECIPIENT_TX_TYPES set and
  resolveEffectiveTxParams(). Falls back to intent.recipients, resolves
  txType from intent.intentType, propagates stakingRequestId, throws
  InvalidTransactionError only when no recipients and no exemption applies.

  - ecdsa.ts, ecdsaMPCv2.ts — replace vulnerable default:
  txParams: resolveEffectiveTxParams(txRequest, params.txParams)

  - bsc.ts, bscToken.ts, xdc.ts, xdcToken.ts, evmCoin.ts,
  abstractEthLikeNewCoins.ts — replace stale hardcoded arrays with
  NO_RECIPIENT_TX_TYPES.has(txParams.type) + stakingRequestId bypass.

  - iBaseCoin.ts, baseTypes.ts, iWallet.ts — add
  stakingRequestId?: string to TransactionParams, PopulatedIntent,
  PrebuildTransactionResult.

  Test Plan

  - yarn test --filter=@bitgo/sdk-core passes
  - Normal ETH/BTC send — address + amount still verified
  - Post-deploy Loki: zero new "TSS signing with no intent recipients" hits
  - TypeScript compiles clean across all affected packages

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -46,6 +46,7 @@ import {
   isTssVerifyAddressOptions,
   DeriveAddressOptions,
   DeriveAddressResult,
+  NO_RECIPIENT_TX_TYPES,
 } from '@bitgo/sdk-core';
 import { getDerivationPath } from '@bitgo/sdk-lib-mpc';
 import { bip32 } from '@bitgo/secp256k1';
@@ -3149,18 +3150,9 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
       !(
         txParams.prebuildTx?.consolidateId ||
         txPrebuild?.consolidateId ||
-        (txParams.type &&
-          [
-            'acceleration',
-            'fillNonce',
-            'transferToken',
-            'tokenApproval',
-            'consolidate',
-            'bridgeFunds',
-            'enabletoken',
-            'defiApprove',
-            'defiDeposit',
-          ].includes(txParams.type))
+        txParams.stakingRequestId ||
+        txParams.prebuildTx?.stakingRequestId ||
+        (txParams.type && NO_RECIPIENT_TX_TYPES.has(txParams.type))
       )
     ) {
       throw new Error('missing txParams');
--- a/modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
@@ -1,0 +1,120 @@
+import { TransactionParams } from '../../baseCoin';
+import { InvalidTransactionError } from '../../errors';
+import { PopulatedIntent, TxRequest } from './baseTypes';
+
+/**
+ * Transaction types that legitimately carry no explicit recipients.
+ * These are the intentType strings as stored in TxRequest.intent.intentType by WP.
+ * verifyTransaction handles no-recipient validation for these internally.
+ * Mirrors the bypass list in abstractEthLikeNewCoins.ts verifyTssTransaction.
+ *
+ * ECDSA types: acceleration, fillNonce, transferToken, tokenApproval, consolidate,
+ *              bridgeFunds, enableToken, customTx, contractCall
+ * BSC/BNB delegation-based staking: delegate, undelegate, switchValidator
+ * CELO/ETH lock-based staking: stake, unstake, stakeWithCallData, unstakeWithCallData,
+ *              transferStake, increaseStake, goUnstake
+ * Claim rewards (BSC, CELO — TRX/SOL use EdDSA and are unaffected): claim, stakeClaimRewards
+ */
+export const NO_RECIPIENT_TX_TYPES = new Set([
+  // ECDSA types
+  'acceleration',
+  'fillNonce',
+  'transferToken',
+  'tokenApproval',
+  'consolidate',
+  'bridgeFunds',
+  'enableToken',
+  'enabletoken',
+  'disabletoken',
+  'customTx',
+  // DeFi vault operations — recipients/calldata built server-side from defiParams
+  'defiApprove',
+  'defiDeposit',
+  // Smart contract invocations with no explicit SDK-level recipients
+  'contractCall',
+
+  // BSC/BNB delegation-based staking — intentType strings from TxRequest.intent.intentType
+  'delegate',
+  'undelegate',
+  'switchValidator',
+
+  // CELO/ETH lock-based staking
+  'stake',
+  'unstake',
+  'stakeWithCallData',
+  'unstakeWithCallData',
+  'transferStake',
+  'increaseStake',
+  'goUnstake',
+
+  // Claim rewards — BSC and CELO (TRX/SOL/Cosmos use EdDSA, not affected by this guard)
+  'claim',
+  'stakeClaimRewards',
+
+  'createAccount',
+  'transferAccept',
+  'transferReject',
+  'transferOfferWithdrawn',
+  'cantonCommand',
+  'pledge',
+]);
+
+/**
+ * Resolves the effective txParams for TSS signing recipient verification.
+ *
+ * For smart contract interactions, recipients live in txRequest.intent.recipients
+ * (native amount = 0, so buildParams is empty). Falls back to intent recipients
+ * mapped to ITransactionRecipient shape when txParams.recipients is absent.
+ *
+ * Staking intents (BSC delegate/undelegate, CELO stake/unstake, etc.) are
+ * identified generically by the presence of `stakingRequestId` on the intent —
+ * a required field on BaseStakeIntent in @bitgo/public-types. These intents
+ * have no txParams recipients by design; validation is done at the coin layer.
+ *
+ * Throws InvalidTransactionError if no recipients can be resolved and the
+ * transaction is not a known no-recipient type.
+ */
+export function resolveEffectiveTxParams(
+  txRequest: TxRequest,
+  txParams: TransactionParams | undefined
+): TransactionParams {
+  const intentRecipients = (txRequest.intent as PopulatedIntent)?.recipients?.map((intentRecipient) => ({
+    address: intentRecipient.address.address,
+    amount: intentRecipient.amount.value,
+    data: intentRecipient.data,
+  }));
+
+  const effectiveTxParams: TransactionParams = {
+    ...txParams,
+    recipients: txParams?.recipients?.length ? txParams.recipients : intentRecipients,
+  };
+
+  // Fall back to intent.intentType when txParams.type is not explicitly set.
+  // Staking wallets call signTransaction without txParams, so the type lives only in the intent.
+  const txType = effectiveTxParams.type ?? (txRequest.intent as PopulatedIntent)?.intentType ?? '';
+
+  // Propagate the resolved type so downstream callers
+  if (!effectiveTxParams.type && txType) {
+    effectiveTxParams.type = txType;
+  }
+
+  // Propagate stakingRequestId from intent into effectiveTxParams so verifyTssTransaction
+  // overrides can bypass the no-recipient guard without needing access to txRequest directly.
+  const intentStakingRequestId = (txRequest.intent as PopulatedIntent)?.stakingRequestId;
+  if (intentStakingRequestId && !effectiveTxParams.stakingRequestId) {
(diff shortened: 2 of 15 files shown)


modules/abstract-eth/src/abstractEthLikeNewCoins.ts
```diff
@@ -3089,6 +3090,10 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
    * @throws {TxIntentMismatchRecipientError} if transaction recipients don't match user intent
    */
   async verifyTssTransaction(params: VerifyEthTransactionOptions): Promise<boolean> {
+    if (params.verification?.skipTssRecipientVerification) {
+      return true;
+    }
+
```
mohammadalfaiyazbitgo: i don't think you should skip verification entirely, just the ones that involve `recipients`

mohammadalfaiyazbitgo (changes requested):

modules/abstract-eth/src/abstractEthLikeNewCoins.ts
mrdanish26: Good catch! updated, the flag now only skips the recipient-specific checks

modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+ *              bridgeFunds, enableToken, customTx, contractCall
+ * BSC/BNB delegation-based staking: delegate, undelegate, switchValidator
+ * CELO/ETH lock-based staking: stake, unstake, stakeWithCallData, unstakeWithCallData,
+ *              transferStake, increaseStake, goUnstake
+ * Claim rewards (BSC, CELO — TRX/SOL use EdDSA and are unaffected): claim, stakeClaimRewards
+ * Dry-run confirmed (2026-05-20 investigation): createAccount, transferAccept, transferReject,
+ *              transferOfferWithdrawn, cantonCommand, pledge
+ * Dry-run confirmed (2026-06-03 investigation): contractCall (hash/Provenance — secp256k1,
+ *              smart contract invocation with no explicit recipients)
+ */
```
mohammadalfaiyazbitgo, line 17: please trim this comment to only report context for a generic reader./

modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+  'unstakeWithCallData',
+  'transferStake',
+  'increaseStake',
+  'goUnstake',
+
+  // Claim rewards — BSC and CELO (TRX/SOL/Cosmos use EdDSA, not affected by this guard)
+  'claim',
+  'stakeClaimRewards',
+
+  // Dry-run confirmed no-recipient types (13-day observation, 2026-05-20)
```
mohammadalfaiyazbitgo: redundant

modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+  // Claim rewards
+  'claim',
+  'stakeClaimRewards',
+
+  'createAccount',
+  'transferAccept',
+  'transferReject',
+  'transferOfferWithdrawn',
+  'cantonCommand',
+  'pledge',
```
mohammadalfaiyazbitgo, line 59: as a follow up, could we pull this from: https://github.com/BitGo/public-types/tree/master/src/schema/transactionRequest/intents

modules/sdk-core/src/bitgo/utils/tss/recipientUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+ * have no txParams recipients by design; validation is done at the coin layer.
+ *
+ * Throws InvalidTransactionError if no recipients can be resolved and the
+ * transaction is not a known no-recipient type.
+ */
+export function resolveEffectiveTxParams(
+  txRequest: TxRequest,
+  txParams: TransactionParams | undefined
+): TransactionParams {
+  const intentRecipients = (txRequest.intent as PopulatedIntent)?.recipients?.map((intentRecipient) => ({
```
mohammadalfaiyazbitgo, line 81: another follow up, let's not blindly cast, and instead use typescripts type guards to narrow the type.

mohammadalfaiyazbitgo (dismissed):

mohammadalfaiyazbitgo (dismissed):

mukeshsp (dismissed):

joeypinz (dismissed):

joeypinz (dismissed):
approved before conflicts resolved. code looks good after conflict resolved 👍

mukeshsp (dismissed):

mukeshsp (dismissed):

zahin-mohammad:
## Review summary

This routes both ECDSA TSS signing paths through a single `resolveEffectiveTxParams()` that resolves recipients (`txParams` → `intent.recipients`), exempts known recipient-less types via `NO_RECIPIENT_TX_TYPES`, and throws otherwise — replacing the `txParams: params.txParams || { recipients: [] }` default that disabled recipient verification when `txParams` was absent. Core logic looks sound and the test coverage is good. A few things worth resolving before merge, especially given the revert history below.

### Context: this is the third attempt
The same fix has merged and been reverted twice (#8462 → reverted by #8538; #8539 → reverted by #8665). Both reverts were caused by being too strict and throwing on legitimate recipient-less transactions:

- **#8462** required recipients unconditionally → broke token/staking/contract flows.
- **#8539** added intent fallback + an EVM-only exemption set, but resolved the type only from `txParams` (not `intent.intentType`), so flows that sign without `txParams` still threw.

This PR's `intent.intentType` resolution + expanded exemption set is the right fix for what broke #8539. The recurring failure mode (exemption set under-counts legitimate cases) is worth keeping in mind — the bar for completeness should be high here.

### Blocking concerns

**1. The `skipTssRecipientVerification` opt-out may undercut the fix on the main path.**
Per the description, a companion client-side PR sets this flag to `true`, which disables recipient verification for that path. But `resolveEffectiveTxParams` already falls back to `intent.recipients` for intent-only flows — so what case requires the blanket opt-out? As written, the most common signing path would skip the exact check this PR adds. Could you clarify why the intent fallback is insufficient there, or scope the opt-out more narrowly than a per-call boolean? The flag also reads as insurance against the exemption set being incomplete, which is worth being explicit about.

**2. Pending-approval recipient threading was dropped vs. the prior attempts.**
#8462 and #8539 both threaded the pending-approval recipients explicitly through `recreateTxRequest`. This PR removes that and relies entirely on `txRequest.intent.recipients` being populated after recreation. Can you confirm recreated pending-approval txRequests carry `intent.recipients`? If not, this could reintroduce a false-throw on that path.

**3. `enableToken` casing.**
`NO_RECIPIENT_TX_TYPES` uses camelCase `'enableToken'` (matching `intentType`), but the EVM coin layer sets `txParams.type` to lowercase `'enabletoken'` in some paths. The intent-sourced path is covered, but a caller passing the lowercase string with no recipients would now throw. Adding `'enabletoken'` to the set as well is a cheap safeguard.

### Lower-priority notes
- `skipTssRecipientVerification` is exposed on the Express route schema. Consistent with the threat model (the flag is set client-side), but it does broaden the bypass surface to every request — worth a confirming comment.
- The exemption set mixes intent types across coin families. Worth confirming each newly added non-EVM type is genuinely recipient-less and routes through this path, rather than being a value transfer.

### Strengths
- The `intent.intentType` resolution directly addresses the regression that reverted the previous attempt, and the `isPopulatedIntent()` type guard replaces the prior unsafe cast.
- Test coverage asserts value-transfer types (`payment`, `fanout`, `vote`) are **not** exempt, and the new coin-level suite covers the flag, hop, and missing-param ordering.
- Centralizing the previously divergent per-coin arrays into one set removes the drift that contributed to the earlier reverts.
- Reordering the wallet/hop checks ahead of the recipient guard is a genuine correctness improvement and is tested.

### Recommendation
Worth resolving (1) and (2) before merge — given two prior production reverts, the key questions are whether the opt-out leaves the main path protected, and whether the pending-approval path still resolves recipients. (3) is a one-line safeguard. The core logic is solid; the remaining risk is in completeness and in whether the opt-out hollows out the protection.

joeypinz (dismissed):

mrdanish26:
> ### Blocking concerns
> **1. The `skipTssRecipientVerification` opt-out may undercut the fix on the main path.** Per the description, a companion client-side PR sets this flag to `true`, which disables recipient verification for that path. But `resolveEffectiveTxParams` already falls back to `intent.recipients` for intent-only flows — so what case requires the blanket opt-out? As written, the most common signing path would skip the exact check this PR adds. Could you clarify why the intent fallback is insufficient there, or scope the opt-out more narrowly than a per-call boolean? The flag also reads as insurance against the exemption set being incomplete, which is worth being explicit about.
> 
The two layers do different things:
  - Layer 1 (resolveEffectiveTxParams): recipients exist — not bypassable by the flag
  - Layer 2 (verifyTssTransaction): recipients match the txPrebuild — currently bypassed by the flag

The flag was backup against the exemption set being incomplete (Layer 1 throws on unknown no-recipient types). 

> **2. Pending-approval recipient threading was dropped vs. the prior attempts.** #8462 and #8539 both threaded the pending-approval recipients explicitly through `recreateTxRequest`. This PR removes that and relies entirely on `txRequest.intent.recipients` being populated after recreation. Can you confirm recreated pending-approval txRequests carry `intent.recipients`? If not, this could reintroduce a false-throw on that path.
> 

The intent (including recipients) is persisted server-side at `txRequest` creation time and is not affected by signature deletion. I verified in production logs that all payment intents carry intent.recipients.

zahin-mohammad (commented):
per the huddle, needs a few more changes (removing the opt out bool)

zahin-mohammad (dismissed):
lgtm... but please keep a close eye on this for the beta release and the ui release

mohammadalfaiyazbitgo (dismissed):

mukeshsp (dismissed):

gokulhost (approved):

mohammadalfaiyazbitgo (approved):

mukeshsp (approved):

lokesh-bitgo (approved):

PR #8926, opened 2026-06-03
fix: reject duplicate contract address and NFT index keys in CoinMap

nvjsr:
**Description**

Fixes CGD-715 / BTGO-13: CoinMap.addCoin() enforced uniqueness for coin name, id, and alias, but silently overwrote entries in the contract-address and NFT-collection secondary indexes when two tokens shared the same lookup key. A later-registered token could shadow an existing one for coins.get(family:networkType:contractAddress) lookups, leading to wrong token metadata (e.g. decimals) during transaction construction.

**Changes:**

Throw DuplicateContractAddressDefinitionError / DuplicateNftCollectionIdDefinitionError on index collision (same pattern as name/id checks).
Include network.type in index keys (family:mainnet|testnet:address) so mainnet and testnet tokens on the same chain family no longer share one slot.
Assign unique placeholder contract addresses for eth:ctkn and eth:cusdt (both previously used 0x000…000).
Add unit tests for duplicate contract/NFT keys and update address/NFT lookup tests for the new key format.
Motivation: Sigma Prime audit finding — secondary indexes must not allow silent overwrite.

**Issue Number**
CGD-715 (BTGO-13)

**Type of change**
 Bug fix (non-breaking change which fixes an issue)
 New feature (non-breaking change which adds functionality)
 Breaking change (fix or feature that would cause existing functionality to not work as expected)

 This change requires a documentation update
Breaking change: Contract-address and NFT-collection lookups via CoinMap.get() must use the new composite key format, e.g. eth:mainnet:0x… instead of eth:0x…. Coin lookup by name is unchanged.

### DIFF
--- a/modules/statics/src/map.ts
+++ b/modules/statics/src/map.ts
@@ -1,5 +1,11 @@
 import { BaseCoin } from './base';
-import { DuplicateCoinDefinitionError, CoinNotDefinedError, DuplicateCoinIdDefinitionError } from './errors';
+import {
+  DuplicateCoinDefinitionError,
+  CoinNotDefinedError,
+  DuplicateCoinIdDefinitionError,
+  DuplicateContractAddressDefinitionError,
+  DuplicateNftCollectionIdDefinitionError,
+} from './errors';
 import { ContractAddressDefinedToken, NFTCollectionIdDefinedToken } from './account';
 import { EthereumNetwork } from './networks';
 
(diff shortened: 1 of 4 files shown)


modules/statics/src/map.ts
```diff
@@ -47,9 +61,19 @@ export class CoinMap {
 
     if (coin.isToken) {
       if (coin instanceof ContractAddressDefinedToken) {
-        this._coinByContractAddress.set(`${coin.family}:${coin.contractAddress}`, coin);
+        const contractAddressKey = CoinMap.contractAddressKey(coin);
+        const existingByContractAddress = this._coinByContractAddress.get(contractAddressKey);
+        if (existingByContractAddress) {
+          throw new DuplicateContractAddressDefinitionError(contractAddressKey, existingByContractAddress.name);
```
mullapudipruthvik, line 67: when does this run ?

mullapudipruthvik (approved):

modules/statics/src/map.ts
nvjsr: when a new token is added to the coin catalog

PR #8930, opened 2026-06-03
fix(sdk-core): pass round to getBitgoSignatureShare helper

vibhavgo:
EdDSA MPCv2 signing can receive multiple BitGo-to-signer shares in a txRequest.
Selecting only by sender/recipient can pick a stale round-1 output during round 2.
- Add shareType param to getBitgoSignatureShare to filter by round output type.
- Update EdDSA MPCv2 callers to request round-1 or round-2 output explicitly.
- Remove inline reverse().find() in offline round-3 path in favour of the helper.
- Update tests to expect round-specific missing share error messages.

Ticket: WCI-626

### DIFF
--- a/modules/sdk-core/src/bitgo/tss/common.ts
+++ b/modules/sdk-core/src/bitgo/tss/common.ts
@@ -1,6 +1,7 @@
 import assert from 'assert';
 import openpgp from 'openpgp';
 
+import { MPCv2SigningState } from '@bitgo/public-types';
 import { BitGoBase } from '../bitgoBase';
 import { TxRequestChallengeResponse } from './types';
 import {
@@ -20,12 +21,21 @@ const debug = require('debug')('bitgo:tss:common');
 
 export function getBitgoSignatureShare(
   signatureShares: SignatureShareRecord[],
-  signerShareType: SignatureShareType
+  signerShareType: SignatureShareType,
+  shareType: MPCv2SigningState
 ): SignatureShareRecord {
-  const bitgoShare = signatureShares.find(
-    (share) => share.from === SignatureShareType.BITGO && share.to === signerShareType
-  );
-  assert(bitgoShare, 'Missing BitGo signature share');
+  const bitgoShare = signatureShares.find((share) => {
+    if (share.from !== SignatureShareType.BITGO || share.to !== signerShareType) {
+      return false;
+    }
+
+    try {
+      return JSON.parse(share.share).type === shareType;
+    } catch {
+      return false;
+    }
+  });
+  assert(bitgoShare, `Missing BitGo ${shareType} signature share`);
   return bitgoShare;
 }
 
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/tss/common.ts
```diff
@@ -20,12 +20,16 @@ const debug = require('debug')('bitgo:tss:common');
 
 export function getBitgoSignatureShare(
   signatureShares: SignatureShareRecord[],
-  signerShareType: SignatureShareType
+  signerShareType: SignatureShareType,
+  shareType: 'round1Output' | 'round2Output' | 'round3Output'
```
Marzooqa: Do we have an enum for this?

Marzooqa:
@claude

modules/sdk-core/src/bitgo/tss/common.ts
vibhavgo: No existing enum

vibhavgo: Will use public-type

modules/sdk-core/src/bitgo/tss/common.ts
```diff
@@ -20,12 +21,16 @@ const debug = require('debug')('bitgo:tss:common');
...
-  signerShareType: SignatureShareType
+  signerShareType: SignatureShareType,
+  shareType: MPCv2SigningState
 ): SignatureShareRecord {
   const bitgoShare = signatureShares.find(
-    (share) => share.from === SignatureShareType.BITGO && share.to === signerShareType
+    (share) =>
+      share.from === SignatureShareType.BITGO &&
+      share.to === signerShareType &&
+      JSON.parse(share.share).type === shareType
```
Marzooqa: This `JSON.parse(share.share)` has no try/catch, which is a regression from the old inline implementation it replaces. In `eddsaMPCv2.ts` the equivalent code was wrapped in a `try/catch` that returned `false` on failure, so a malformed `share.share` field caused the entry to be skipped. Now a single malformed record will throw and crash the entire signing flow.

Suggested fix:
```ts
try {
  return JSON.parse(share.share).type === shareType;
} catch {
  return false;
}
```

modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/eddsaMPCv2.ts
```diff
@@ -799,7 +799,7 @@ describe('EddsaMPCv2Utils.createOfflineRound2Share', () => {
           encryptedUserGpgPrvKey: round1.encryptedUserGpgPrvKey,
           encryptedRound1Session: round1.encryptedRound1Session,
         }),
-      /Missing BitGo signature share/
+      /Missing BitGo round1Output signature share/
```
Marzooqa, line 803: The test changes here only update the error message strings. The core motivation for this fix — preventing the wrong round share from being selected when both a `round1Output` and a `round2Output` share are present in the list — is not covered. A test that populates `signatureShares` with both types and asserts the helper returns the correct one would directly validate the disambiguation logic.

Marzooqa (approved):

PR #8936, opened 2026-06-03
feat: default encryptAsync to v2

bdesoky:
Ticket: WCN-723

This pull request updates the encryption in `encryptAsync` to default to v2

**Encryption logic improvements:**

* Changed the default encryption version in `encryptAsync` to v2 (Argon2id + AES-256-GCM), so unless explicitly set to v1, new data will use the more secure encryption scheme.

**Testing enhancements:**

* Updated unit tests to verify that v2 is the default and explicit v1 encryption still works.

### DIFF
--- a/modules/sdk-api/src/encrypt.ts
+++ b/modules/sdk-api/src/encrypt.ts
@@ -44,17 +44,17 @@ export function encrypt(
 
 /**
  * Async encrypt that dispatches to v1 (SJCL) or v2 (Argon2id + AES-256-GCM)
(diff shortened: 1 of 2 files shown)


bdesoky:
Synced with @pranavjain97, as this change doesnt remove the `encrypt()` function and no external callers are using `encryptAsync`, this change is not breaking. `decryptAsync` can handle v1/2 decryption with no changes

pranishnepal (approved):

PR #8938, opened 2026-06-04
fix(abstract-eth): fix verifyTransaction for base multig address duri…

MohammedRyaan786:
TICKET: CHALO-530

This pull request adds support for ERC-7984 token enablement transactions on multisig base-address wallets, specifically handling cases where the transaction recipient is the Zama ACL contract instead of the wallet's base address. It also introduces new tests to verify this behavior.

**ERC-7984 Token Enablement Support:**

* Updated the recipient address verification logic in `AbstractEthLikeNewCoins` so that, for `enabletoken` transactions, if the transaction is targeting the Zama ACL contract (as configured for the network), the usual recipient address equality check is bypassed. This ensures that valid enabletoken transactions to the ACL contract are accepted.

**Testing Enhancements:**

* Added new unit tests in `eth.ts` to verify that `verifyTransaction` correctly allows enabletoken transactions where the `txParams` recipient is the wallet base address and the `txPrebuild` recipient is the Zama ACL contract, and that it rejects cases where the recipient is not the ACL contract.

### DIFF
--- a/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
+++ b/modules/abstract-eth/src/abstractEthLikeNewCoins.ts
@@ -3289,7 +3289,12 @@ export abstract class AbstractEthLikeNewCoins extends AbstractEthLikeCoin {
         `${this.getChain()} doesn't support sending to more than 1 destination address within a single transaction. Try again, using only a single recipient.`
       );
     }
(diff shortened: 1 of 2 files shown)


bhuvanr159 (approved):
lgtm

PR #8956, opened 2026-06-05
feat: add external signer callback support for onchain multisig wallet generation

danielpeng1:
Adding SDK external signer callback for onchain multisig wallet generation (for AKM).
- Added `generateWalletWithExternalSigner()` to `Wallets` in `sdk-core`; takes a `createKeychainCallback`, calls it for user and backup in parallel alongside `createBitGo`, uploads all three keychains, and posts the wallet
- Added `CreateKeychainCallback`, `CreateKeychainCallbackParams`, `CreateKeychainCallbackResult`, and `GenerateWalletWithExternalSignerOptions` to `iWallets.ts`
- `generateWallet()` redirects to to `generateWalletWithExternalSigner()` when `createKeychainCallback` is present. params that don't make sense for external signers (`passphrase`, `userKey`, `backupXpub`, `passcodeEncryptionCode`, `webauthnInfo`) throw explicitly rather than being silently ignored
- Note that the callback result's `type` field gets mapped to `keyType` internally when calling `keychains().add()` since I saw that `AddKeychainOptions` has a `type: string` field and a separate meaningful `keyType: KeyType` field, and AKM was already using `keyType` correctly, so we match that. The AKM callback factory for `WCN-684` (the associated AKM ticket) should just return `{ pub, type: key.type, source }` as-is and the SDK handles the mapping.
- Added unit tests in new `walletsExternalSigner.ts` covering the happy path (keys created, uploaded, wallet posted with correct params), source mismatch rejection, unsupported multisig types, and all the explicitly rejected param combinations (custodial type, passcodeEncryptionCode, webauthnInfo, isDistributedCustody edge cases, userKey conflict)

Ticket: WCN-685

### Type of change

### Checklist:

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallets.ts
@@ -70,8 +70,8 @@ export interface CreateKeychainCallbackParams {
 
 export interface CreateKeychainCallbackResult {
   pub: string;
-  type: string;
-  source: string;
+  type: 'independent';
+  source: 'user' | 'backup';
 }
 
 export type CreateKeychainCallback = (params: CreateKeychainCallbackParams) => Promise<CreateKeychainCallbackResult>;
@@ -116,6 +116,8 @@ export interface GenerateWalletWithExternalSignerOptions
   extends Omit<GenerateWalletOptions, 'passphrase' | 'userKey' | 'backupXpub' | 'backupXpubProvider'> {
   label: string;
   createKeychainCallback: CreateKeychainCallback;
+  /** Optional user-key signatures over backup/bitgo pubs. Omit when the external signer cannot produce them (equivalent to a cold wallet). */
+  keySignatures?: { backup: string; bitgo: string };
 }
 
 export const GenerateLightningWalletOptionsCodec = t.intersection(
--- a/modules/sdk-core/src/bitgo/wallet/iWallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallets.ts
@@ -63,6 +63,19 @@ export interface GenerateSMCMpcWalletOptions extends GenerateBaseMpcWalletOption
   coldDerivationSeed?: string;
 }
 
+export interface CreateKeychainCallbackParams {
+  source: 'user' | 'backup';
+  coin: string;
+}
+
+export interface CreateKeychainCallbackResult {
+  pub: string;
+  type: string;
+  source: string;
+}
+
+export type CreateKeychainCallback = (params: CreateKeychainCallbackParams) => Promise<CreateKeychainCallbackResult>;
+
 export interface GenerateWalletOptions {
   label?: string;
   passphrase?: string;
@@ -95,6 +108,14 @@ export interface GenerateWalletOptions {
   /** Optional WebAuthn PRF-based encryption info. When provided, the user private key is additionally encrypted with the PRF-derived passphrase so the server can store a WebAuthn-protected copy. */
   webauthnInfo?: WebauthnKeyEncryptionInfo;
   encryptionVersion?: EncryptionVersion;
+  /** Delegates user/backup key creation to an external signer (onchain multisig only). */
+  createKeychainCallback?: CreateKeychainCallback;
+}
+
+export interface GenerateWalletWithExternalSignerOptions
+  extends Omit<GenerateWalletOptions, 'passphrase' | 'userKey' | 'backupXpub' | 'backupXpubProvider'> {
+  label: string;
+  createKeychainCallback: CreateKeychainCallback;
 }
 
 export const GenerateLightningWalletOptionsCodec = t.intersection(
@@ -281,6 +302,7 @@ export interface IWallets {
   generateWallet(
     params?: GenerateWalletOptions
   ): Promise<WalletWithKeychains | LightningWalletWithKeychains | GoAccountWalletWithUserKeychain>;
+  generateWalletWithExternalSigner(params: GenerateWalletWithExternalSignerOptions): Promise<WalletWithKeychains>;
   listShares(params?: Record<string, unknown>): Promise<any>;
   getShare(params?: { walletShareId?: string }): Promise<any>;
   updateShare(params?: UpdateShareOptions): Promise<any>;
--- a/modules/sdk-core/src/bitgo/wallet/wallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallets.ts
@@ -12,7 +12,7 @@ import * as common from '../../common';
 import { IBaseCoin, KeychainsTriplet, SupplementGenerateWalletOptions } from '../baseCoin';
 import { BitGoBase } from '../bitgoBase';
 import { getSharedSecret } from '../ecdh';
-import { AddKeychainOptions, Keychain, KeyIndices } from '../keychain';
+import { AddKeychainOptions, Keychain, KeyIndices, KeyType } from '../keychain';
 import { decodeOrElse, promiseProps, RequestTracer } from '../utils';
 import {
   AcceptShareOptions,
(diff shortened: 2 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -63,6 +63,19 @@ export interface GenerateSMCMpcWalletOptions extends GenerateBaseMpcWalletOption
...
 }
 
+export interface CreateKeychainCallbackParams {
+  source: 'user' | 'backup';
+  coin: string;
+}
+
+export interface CreateKeychainCallbackResult {
+  pub: string;
+  type: string;
```
pranavjain97: `type: string` is too loose for a public callback. should be `KeyType` (or `'independent'` for onchain multisig).

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -718,6 +723,162 @@ export class Wallets implements IWallets {
...
+    const coin = this.baseCoin.getChain();
+
+    const createAndUploadKeychain = async (source: 'user' | 'backup'): Promise<Keychain> => {
+      const keychainFromCallback = await createKeychainCallback({ source, coin });
+      if (keychainFromCallback.source !== source) {
+        throw new Error(`createKeychainCallback returned source ${keychainFromCallback.source}, expected ${source}`);
+      }
+      return this.baseCoin.keychains().add({
+        pub: keychainFromCallback.pub,
+        keyType: keychainFromCallback.type as KeyType,
```
pranavjain97: `as KeyType` papers over the mismatch. test passes `'tbtc'` which isn't a valid `KeyType`. hardcode `'independent'` or validate.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -718,6 +723,162 @@ export class Wallets implements IWallets {
...
+
+    const { userKeychain, backupKeychain, bitgoKeychain }: KeychainsTriplet = await promiseProps({
+      userKeychain: createAndUploadKeychain('user'),
+      backupKeychain: createAndUploadKeychain('backup'),
+      bitgoKeychain: this.baseCoin
+        .keychains()
+        .createBitGo({ enterprise, reqId, isDistributedCustody: params.isDistributedCustody }),
+    });
+
+    walletParams.keys = [userKeychain.id, backupKeychain.id, bitgoKeychain.id];
```
pranavjain97, line 871: standard `generateWallet` computes `keySignatures.backup`/`keySignatures.bitgo` from the user prv. this path can't (prv lives in the callback) so those bindings get silently dropped. is AKM providing an equivalent attestation?

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -718,6 +723,162 @@ export class Wallets implements IWallets {
...
+    }
+
+    const reqId = new RequestTracer();
+    const coin = this.baseCoin.getChain();
+
+    const createAndUploadKeychain = async (source: 'user' | 'backup'): Promise<Keychain> => {
+      const keychainFromCallback = await createKeychainCallback({ source, coin });
+      if (keychainFromCallback.source !== source) {
+        throw new Error(`createKeychainCallback returned source ${keychainFromCallback.source}, expected ${source}`);
+      }
```
pranavjain97: we check `source` matches but not coin. a buggy callback could return a BTC pubkey when we asked for ETH and we'd upload it.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -718,6 +723,162 @@ export class Wallets implements IWallets {
...
+        pub: keychainFromCallback.pub,
+        keyType: keychainFromCallback.type as KeyType,
+        source: keychainFromCallback.source,
+        reqId,
+      });
+    };
+
+    const { userKeychain, backupKeychain, bitgoKeychain }: KeychainsTriplet = await promiseProps({
+      userKeychain: createAndUploadKeychain('user'),
+      backupKeychain: createAndUploadKeychain('backup'),
```
pranavjain97, line 865: callbacks run in parallel. if backup succeeds and user fails we've already uploaded a keychain server-side. also no test for callback throwing or returning bad data.

pranavjain97 (changes requested):
great progress, a few things to address before merging.

pranavjain97:
@claude

modules/sdk-core/src/bitgo/wallet/iWallets.ts
danielpeng1: Changed to `'independent'`, also tightened up `source` to be `'user' | 'backup'`

modules/sdk-core/src/bitgo/wallet/wallets.ts
danielpeng1: removed that cast, from the change above `type` is already `'independent'` so this is good to go now.

modules/sdk-core/src/bitgo/wallet/wallets.ts
danielpeng1: Makes sense, I just added optional `keySignatures?: { backup: string; bitgo: string }` on `GenerateWalletWithExternalSignerOptions` in `iWallets.ts` forwarded to `walletParams.keySignatures` if present, omitted otherwise. Added 2 tests covering presence and absence.

No equivalent attestation from AKM specifically right? some `prv` stays in the HSM so it can't be computed.

modules/sdk-core/src/bitgo/wallet/wallets.ts
danielpeng1: Ok I added `this.baseCoin.isValidPub(keychainFromCallback.pub)` in `wallets.ts` so a wrong-coin pub would be rejected before upload.

modules/sdk-core/src/bitgo/wallet/wallets.ts
danielpeng1: Right I changed this to use parallel execution via `promiseProps` now like how we do it in `generateWallet`. Error wrapping moved inside `createAndUploadKeychain` so context is preserved regardless of ordering. Added the associated tests for: callback throws, backup fails, invalid pub, non-Error rejection, and `keySignatures` passthrough.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -718,6 +723,180 @@ export class Wallets implements IWallets {
...
+    if (!_.isFunction(params.createKeychainCallback)) {
+      throw new Error('missing required function parameter createKeychainCallback');
+    }
+
+    const multisigType = params.multisigType ?? this.baseCoin.getDefaultMultisigType();
+    if (multisigType !== 'onchain') {
+      throw new Error('external signer wallet generation is only supported for onchain multisig wallets');
+    }
+
+    const conflictingParams = ['passphrase', 'userKey', 'backupXpub', 'backupXpubProvider'] as const;
```
pranishnepal: what is "conflicting" here? Readers of this won't have context on the conflict

danielpeng1: `generateWalletWithExternalSignerOptions` already omits those params at the type level so this check is a runtime guard. These params are all part of a passphrase-based key creation path thus mutually exclusive with `createKeychainCallback` which is the conflicting part. Maybe `passphrasePathParams` would be clearer here?

pranishnepal: that would be good; you can add the reasoning as a `// comment` too

danielpeng1: done

modules/sdk-core/src/bitgo/wallet/wallets.ts
pranavjain97: >  some prv stays in the HSM so it can't be computed

only for multisig yes, and in external signing mode

pranavjain97 (approved):
awesome

PR #8957, opened 2026-06-05
fix(statics): skip AMS tokens that reuse a static contract address or NFT collection id

mmcshinsky-bitgo:
Fixes **[CSHLD-976](https://linear.app/bitgo/issue/CSHLD-976)**. Supersedes the handoff reference **#8959** (closed by @zahin-mohammad — "will let the coins team take this forward"); this adopts that PR's `hasTokenAddressConflict` design.

## Problem

The AMS↔static merge in `createTokenMapUsingConfigDetails` (and the `…Trimmed…` variant that calls it) dedups via `isCoinPresentInCoinMap` (`coins.ts:434`), which only matches on **name / id / alias** — never the contract address:

```ts
export function isCoinPresentInCoinMap({ name, id, alias }) {
  return Boolean(coins.has(name) || (id && coins.has(id)) || (alias && coins.has(alias)));
}
```

But `CoinMap.fromCoins` → `addCoin` (`map.ts:62-68`) **also** dedups by **contract address** (`family:network.type:contractAddress`) and **NFT collection id**. An AMS-served token reusing a contract a static token already claims **under a different name** slips the guard and throws `DuplicateContractAddressDefinitionError`, aborting the **entire** token-map build for every consumer (bitgo-retail hydration, bitgo-admin, bitgo-microservices, indexerdb, prime, client-admin, `bitgo.ts`, …).

Surfaced by the static `eth:at` bot token (`0x0581ccdf…`, eth testnet, added in #8885 / `@bitgo/statics@58.43.0`) colliding with the live AMS token at the same address under a different name.

## Fix

Add `CoinMap.hasTokenAddressConflict(coin)` — reusing the map's own private `contractAddressKey` / `nftCollectionIdKey` derivation against its populated `_coinByContractAddress` / `_coinByNftCollectionID` indexes — and use it in the merge loop:

```ts
const token = createToken(tokenConfig);
if (token && !coins.hasTokenAddressConflict(token)) {
  BaseCoins.set(token.name, token);
}
```

Statics coins are seeded first, so **statics stays the source of truth** and the colliding *AMS* token is the one skipped. Checking the *built* coin against the map's own indexes means the key derivation/normalization always matches `addCoin` (no parallel bookkeeping, no public key-builder surface). Covers contract-address and NFT-collection-id collisions. Strict statics self-construction is unchanged (still throws on genuine duplicates at build/test time).

## Acceptance criteria (CSHLD-976)
- [x] AMS token sharing a contract address with a static token is skipped (no throw)
- [x] Same protection for NFT collection id collisions (`nftCollectionIdKey`)
- [x] Unit test reproducing the duplicate-contract / different-name case asserting no throw
- [x] UI no longer crashes (companion bitgo-retail #6934 + this root-cause fix)

## Scope note
Targets the **static↔AMS** collision (the incident). A hypothetical **AMS↔AMS** duplicate (two server tokens at the same address, neither in statics) is a separate, pre-existing AMS data-integrity concern — left to surface rather than be silently masked here; the bitgo-retail guard (#6934) keeps it from crashing the UI.

## Verification
- `npx mocha test/unit/coins.ts` → **28768 passing** (incl. new test)
- `tsc --build` clean, `prettier --check` clean

## Related
- Companion (UI defense-in-depth): **bitgo-retail #6934**
- Data follow-up (**AMS-side, not statics** — statics is canonical): reconcile the live AMS record sharing `eth:testnet:0x0581ccdf…` under a different name. (Earlier statics-removal PR #8960 was closed for this reason.)

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -476,7 +476,10 @@ export function createTokenMapUsingConfigDetails(tokenConfigMap: Record<string, 
     if (!isCoinPresentInCoinMap({ ...tokenConfig }) && !nftAndOtherTokens.has(tokenConfig.name)) {
       try {
         const token = createToken(tokenConfig);
-        if (token) {
+        // A token whose name is absent from the static map can still reuse a contract address (or
+        // NFT collection id) that a static token already claims. Skip it so the static coin remains
+        // the authoritative entry for that address.
+        if (token && !coins.hasTokenAddressConflict(token)) {
           BaseCoins.set(token.name, token);
         }
       } catch (e) {
--- a/modules/statics/src/map.ts
+++ b/modules/statics/src/map.ts
@@ -20,6 +20,25 @@ export class CoinMap {
     // Do not instantiate
   }
 
+
+  /**
+   * Whether a different token with the same contract address (or NFT collection id) is already
+   * registered. Token identity in the map is keyed by name/id/alias, but a token also claims a
+   * contract-address key (`family:contractAddress`) and, for NFTs, a collection-id key. Two tokens
+   * that share such a key but differ in name cannot coexist. Callers merging externally-sourced
+   * tokens use this to skip a colliding token rather than silently overwriting the existing entry.
+   */
+  public hasTokenAddressConflict(coin: Readonly<BaseCoin>): boolean {
+    if (coin instanceof ContractAddressDefinedToken) {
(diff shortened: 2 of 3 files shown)


mmcshinsky-bitgo:
Like https://github.com/BitGo/BitGoJS/pull/8959, will let coins team move forward with this.

PR #8959, opened 2026-06-05
fix(statics): skip AMS tokens that reuse a static contract address

zahin-mohammad:
## Summary
- Fixes an uncaught `DuplicateContractAddressDefinitionError` that crashes the BitGo UI when an AMS-served token reuses a contract address already claimed by a static token under a different name.
- Root cause: the AMS↔static merge in `createTokenMapUsingConfigDetails` dedups via `isCoinPresentInCoinMap`, which only matches on **name/id/alias** — never the contract address. The colliding token slipped through and `CoinMap.fromCoins` → `addCoin` then threw.
- Adds `CoinMap.hasTokenAddressConflict(coin)` (reusing the map's own `contractAddressKey` / `nftCollectionIdKey` derivation) and uses it in the merge loop to **skip** a colliding token instead of crashing. Checking the *built* coin means the key uses the same address normalization as static tokens, so it matches reliably. Covers both contract-address and NFT-collection-id collisions.

Surfaced by the `eth:at` bot token (`0x0581ccdf…aa`, eth testnet) added in #8885; the live AMS config serves a token at the same address under a different name. Observed crash path: `syncAmsCoinsToPresenter → createTokenMapUsingTrimmedConfigDetails → createTokenMapUsingConfigDetails → CoinMap.fromCoins → addCoin`.

Ships forward (no statics pin, no lost coins) — unlike the bitgo-retail revert (BitGo/bitgo-retail#6932), which stepped statics back only 3 beta builds and still shipped `eth:at`, so it would not have fixed the crash.

## Test plan
- [x] New TDD test in `modules/statics/test/unit/coins.ts` — fails before the fix (throws `DuplicateContractAddressDefinitionError`), passes after (token skipped, static token preserved).
- [x] Full statics unit suite: 28,769 passing, 0 failing.
- [x] `tsc --noEmit` clean; ESLint 0 errors on changed files.

### Scope note
Targets the **static-vs-AMS** collision (the incident). A hypothetical **AMS-vs-AMS** duplicate (two server tokens at the same address, neither in statics) would still throw at `fromCoins` — a separate, pre-existing AMS data-integrity concern, out of scope here.

Ticket: CSHLD-976

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -476,7 +476,10 @@ export function createTokenMapUsingConfigDetails(tokenConfigMap: Record<string, 
     if (!isCoinPresentInCoinMap({ ...tokenConfig }) && !nftAndOtherTokens.has(tokenConfig.name)) {
       try {
         const token = createToken(tokenConfig);
-        if (token) {
+        // A token whose name is absent from the static map can still reuse a contract address (or
+        // NFT collection id) that a static token already claims. Adding it would make the final
+        // CoinMap.fromCoins throw DuplicateContractAddressDefinitionError, so skip it instead.
+        if (token && !coins.hasTokenAddressConflict(token)) {
           BaseCoins.set(token.name, token);
         }
       } catch (e) {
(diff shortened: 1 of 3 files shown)


zahin-mohammad:
Will let the coins team take this forward.

PR #8962, opened 2026-06-08
feat(sdk-coin-canton): added message signing builders for canton

ravibitgo:
Ticket: [CHALO-545](https://linear.app/bitgo/issue/CHALO-545/add-message-signing-builder-for-canton)

### DIFF
--- a/modules/account-lib/src/index.ts
+++ b/modules/account-lib/src/index.ts
@@ -347,6 +347,8 @@ const coinMessageBuilderFactoryMap = {
   tada: Ada.MessageBuilderFactory,
   sol: Sol.MessageBuilderFactory,
   tsol: Sol.MessageBuilderFactory,
+  canton: Canton.MessageBuilderFactory,
+  tcanton: Canton.MessageBuilderFactory,
 };
 
 coins
--- a/modules/sdk-coin-canton/src/canton.ts
+++ b/modules/sdk-coin-canton/src/canton.ts
@@ -93,6 +93,11 @@ export class Canton extends BaseCoin {
     return true;
   }
 
+  /** @inheritDoc */
+  supportsMessageSigning(): boolean {
+    return true;
+  }
+
   /** inherited doc */
   getDefaultMultisigType(): MultisigType {
     return multisigTypes.tss;
--- a/modules/sdk-coin-canton/src/lib/clearSigning.ts
+++ b/modules/sdk-coin-canton/src/lib/clearSigning.ts
@@ -1,0 +1,93 @@
+import { MessageStandardType } from '@bitgo/sdk-core';
+import { PreparedTransaction } from '../../resources/proto/preparedTransaction.js';
+import utils from './utils';
+
+/**
+ * Clear signing and payload-type detection for Canton message signing requests.
+ *
+ * The Canton Signing Driver receives raw `tx` bytes and a `txHash` from the
+ * Canton Gateway's signTransaction() call. It does not know whether `tx` is a
+ * Daml prepared transaction or a topology transaction. This module owns that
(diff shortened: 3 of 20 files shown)


bhavidhingra (commented):
### Issues

**Medium — Missing error boundary on `decodePreparedTransaction`**

`clearSigning.ts` — no guard against calling this with topology bytes. `utils.extractCantonCommandInfo` will throw if proto parsing fails, and the caller gets an unhandled exception. The JSDoc says _"Only meaningful when `detectCantonSigningPayloadType()` returns `CANTON_SIGN_TRANSACTION`"_ but that is documentation, not enforcement. Should either:
- Throw a descriptive error: `throw new Error("Not a PreparedTransaction — call detectCantonSigningPayloadType first")`
- Or add a test for the invalid-input path so the failure mode is documented

**Medium — Duplicate `getSignablePayload` implementation**

`cantonSignTransactionMessage.ts` and `cantonSignTopologyMessage.ts` have byte-for-byte identical `getSignablePayload()` bodies:
```ts
this.signablePayload = Buffer.from(this.payload, "base64");
return this.signablePayload;
```
The only difference between the two classes is the `type` passed to `super()`. Consider extracting to a shared `CantonBaseMessage extends BaseMessage` to avoid future drift.

**Low — `@ts-ignore` → prefer `@ts-expect-error`**

`clearSigning.ts:3`:
```ts
// @ts-ignore – JS resource with hand-written .d.ts
```
`@ts-expect-error` will error if the suppression becomes unnecessary (e.g. when the `.d.ts` is replaced with proper types), making the tech debt visible.

**Low — Redundant test**

In `canton.ts`, the test `"tcanton should also return true"` fetches `tcanton` again, but `before()` already sets `basecoin = bitgo.coin("tcanton")`. Two tests asserting the same inherited method on the same coin instance.

---

### Positives

- Clean factory pattern — `getMessageBuilder` switch + clear error on unsupported type
- Payload-type detection via try/catch on proto parse is the right approach (nodes-length check is correct discriminator since proto is permissive)
- JSDoc on `clearSigning.ts` is genuinely useful — explains the HSM format split clearly
- Good edge case test coverage (empty bytes, random bytes, proto parse failure)

bhavidhingra (commented):
### Remaining Nits

**Nit — Double proto parse in `decodePreparedTransaction` happy path**

`clearSigning.ts` — the guard calls `detectCantonSigningPayloadType()` (which parses the proto via `PreparedTransaction.fromBinary`), then `utils.extractCantonCommandInfo()` parses the same bytes a second time. For invalid input the guard throws early so no waste, but valid input pays two full proto decodes. Not a correctness issue, but worth noting if Canton transactions are large. Could be addressed by combining into a single parse-and-decode function if perf ever matters.

**Nit — Unnecessary constructor in `CantonBaseMessage`**

`cantonBaseMessage.ts:15-17` — the constructor only calls `super(options)` with no other logic. TypeScript synthesizes this automatically; the explicit constructor is dead weight and can be removed.

---

### Summary

Both are nits, not blockers. The core logic is solid — error guard, base class extraction, and test coverage for invalid input are all well done. LGTM otherwise.

bhavidhingra (approved):
LGTM

Two minor observations (not blocking):

- **`MessageBuilderFactory` pass-through constructor** — same pattern that was cleaned up in `CantonBaseMessage`; `constructor(coinConfig) { super(coinConfig); }` can be removed since TypeScript synthesizes it.

- **Misleading test name** — `cantonSignTopologyMessage.ts`: `"should produce same byte structure as CANTON_SIGN_TRANSACTION for equivalent hash"` only instantiates `CantonSignTopologyMessage`; never creates a `CantonSignTransactionMessage` to compare against. The assertion is correct but the name implies a cross-type comparison that does not happen.

Approving.

PR #8966, opened 2026-06-08
fix(statics): reject duplicate contract address and NFT index keys in coinmap

nvjsr:
TICKET: CGD-715

This pull request introduces stricter checks and error handling for duplicate contract addresses and NFT collection IDs when registering tokens in the coin map. It ensures that two tokens cannot be registered with the same contract address or NFT collection ID for the same network, preventing subtle bugs and improving the reliability of token management. The changes also include new error types, updates to the core logic in `CoinMap`, and expanded test coverage.

**Error handling and validation improvements:**

* Added new error classes: `DuplicateContractAddressDefinitionError` and `DuplicateNftCollectionIdDefinitionError` to provide clear error messages when a duplicate contract address or NFT collection ID is detected during token registration.
* Updated the `CoinMap` class to check for and throw these errors in `addCoin` if a token with the same contract address or NFT collection ID already exists for the same network. [[1]](diffhunk://#diff-b420ea27333b84b67e6a5a33e1a4b47b4f59288995a8f8c6b780a6097a102501R29-R57) [[2]](diffhunk://#diff-b420ea27333b84b67e6a5a33e1a4b47b4f59288995a8f8c6b780a6097a102501L50-R103)
* Introduced helper methods in `CoinMap` (`contractAddressKey`, `nftCollectionIdKey`, and `hasTokenAddressConflict`) to centralize and simplify collision detection logic.

**Token registration and deduplication:**

* Modified `createTokenMapUsingConfigDetails` to skip adding tokens that would conflict with an existing contract address, preventing runtime errors during map creation.
* Refactored code to use the new helper methods for consistent key generation and deletion in `CoinMap`.

**Test coverage:**

* Added unit tests to verify that attempting to register tokens with duplicate contract addresses or NFT collection IDs correctly throws the new errors.
* Added tests to ensure that dynamically created tokens with conflicting contract addresses are skipped, and that the static token remains accessible.

**Other updates:**

* Updated contract addresses for test tokens in `erc7984Tokens.ts` to use unique values.
* Updated imports in test files to include new helper functions.

### DIFF
--- a/modules/statics/src/coins.ts
+++ b/modules/statics/src/coins.ts
@@ -468,6 +468,10 @@ export function createTokenMapUsingConfigDetails(tokenConfigMap: Record<string, 
     BaseCoins.set(coinName, coin);
   });
 
+  // Accumulates both static and already-accepted AMS tokens so AMS-vs-AMS contract address
+  // conflicts are caught in addition to static-vs-AMS conflicts.
+  const accumulatedMap = CoinMap.fromCoins(Array.from(BaseCoins.values()));
+
   // add the tokens not present in the static coin map
   for (const tokenConfigs of Object.values(tokenConfigMap)) {
     if (!tokenConfigs.length) continue;
@@ -476,8 +480,16 @@ export function createTokenMapUsingConfigDetails(tokenConfigMap: Record<string, 
     if (!isCoinPresentInCoinMap({ ...tokenConfig }) && !nftAndOtherTokens.has(tokenConfig.name)) {
       try {
         const token = createToken(tokenConfig);
-        if (token) {
+        // A token whose name is absent from the accumulated map can still reuse a contract address
+        // (or NFT collection id) already claimed by a static or previously-accepted AMS token.
+        // Adding it would make the final CoinMap.fromCoins throw, so skip it instead.
+        if (token && !accumulatedMap.hasTokenAddressConflict(token)) {
           BaseCoins.set(token.name, token);
+          accumulatedMap.addCoin(token);
+        } else if (token) {
+          console.warn(
+            `Skipping token with conflicting contract address or NFT collection id: name="${tokenConfig.name}" id="${tokenConfig.id}"`
+          );
         }
       } catch (e) {
         console.warn(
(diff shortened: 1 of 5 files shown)


modules/statics/src/coins.ts
```diff
@@ -476,7 +476,10 @@ export function createTokenMapUsingConfigDetails(tokenConfigMap: Record<string,
     if (!isCoinPresentInCoinMap({ ...tokenConfig }) && !nftAndOtherTokens.has(tokenConfig.name)) {
       try {
         const token = createToken(tokenConfig);
-        if (token) {
+        // A token whose name is absent from the static map can still reuse a contract address (or
+        // NFT collection id) that a static token already claims. Adding it would make the final
+        // CoinMap.fromCoins throw DuplicateContractAddressDefinitionError, so skip it instead.
+        if (token && !coins.hasTokenAddressConflict(token)) {
```
dpkjnr: [should-fix] `hasTokenAddressConflict` is checked against the static `coins` map only, not tokens already merged into `BaseCoins`. Two AMS-only tokens sharing the same contract address (or NFT collection id) will both pass this guard, get inserted into `BaseCoins`, and then `CoinMap.fromCoins` at line 495 throws — aborting the whole map build instead of skipping the later entry like the static-vs-AMS case. Consider checking against the accumulating set (e.g. a growing `CoinMap` or key set built from `BaseCoins.values()`).

modules/statics/src/coins.ts
```diff
@@ -476,7 +476,10 @@ export function createTokenMapUsingConfigDetails(tokenConfigMap: Record<string,
...
           BaseCoins.set(token.name, token);
```
dpkjnr, line 487: [nit] When a token is skipped due to address conflict there is no `console.warn`, unlike the malformed-token path at lines 486–490. A one-line warn (name, id, conflicting key) would make AMS ingestion failures easier to spot in logs.

mullapudipruthvik (approved):

dpkjnr (approved):

PR #8994, opened 2026-06-11
feat(statics): rename canton fullname in test & prod env

ravibitgo:
Ticket: [CHALO-581](https://linear.app/bitgo/issue/CHALO-581/rename-canton-fullname-in-prod-and-test-envs)


ananth99 (approved):
lgtm

PR #9000, opened 2026-06-11
feat(sdk-core): add deriveAddress primitive to BaseCoin

rajangarg047:
## Summary
Adds a `deriveAddress(params)` method to the `IBaseCoin` interface and a default `BaseCoin` implementation that throws `NotImplementedError`, so individual coins can opt in to **locally deriving a wallet receive address from a derivation path**.

This is the inverse of `isWalletAddress`: instead of checking a candidate address, it *produces* the address offline from public key material only — the user/backup/bitgo xpub triple for BIP32 multisig coins, or the `commonKeychain` for TSS/MPC coins. No private keys and no network access are required.

Introduces `DeriveAddressOptions` and `DeriveAddressResult`, mirroring the existing `VerifyAddressOptions` / `TssVerifyAddressOptions` shape.

## Context
Foundation for FR-465 (Bullish): expose a local address-derivation endpoint in BitGo Express so clients can independently reproduce and cross-check deposit addresses. This PR is the SDK primitive; the per-coin implementations (BTC/ETH/SOL) and the Express endpoint build on it.

- Linear: WCN-912 (child of WCN-864)
- This PR adds no behavior on its own (default throws); it's a non-breaking interface addition.

## Test plan
- [x] `tsc --noEmit` passes for `sdk-core`
- [ ] CI

### DIFF
--- a/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
+++ b/modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
@@ -47,6 +47,8 @@ import {
   AuditKeyParams,
   AuditDecryptedKeyParams,
   TssVerifyAddressOptions,
+  DeriveAddressOptions,
+  DeriveAddressResult,
 } from './iBaseCoin';
 import { IInscriptionBuilder } from '../inscriptionBuilder';
 import {
@@ -351,6 +353,22 @@ export abstract class BaseCoin implements IBaseCoin {
    * @return true iff address is a wallet address. Must return false if address is outside wallet.
    */
   abstract isWalletAddress(params: VerifyAddressOptions | TssVerifyAddressOptions): Promise<boolean>;
+
+  /**
+   * Locally derive and return a wallet receive address for the given derivation path,
+   * using public key material only (no private keys, no network access).
+   *
+   * This is the inverse of {@link isWalletAddress}: rather than checking a candidate
+   * address, it *produces* the address from the wallet's keychains and a chain/index.
+   *
+   * Coins opt in by overriding this method; the default implementation throws.
+   *
+   * @param {DeriveAddressOptions} params - parameters for address derivation (keychains, chain, index, …)
+   * @returns {Promise<DeriveAddressResult>} the derived address plus the path and coin-specific data
+   */
+  deriveAddress(params: DeriveAddressOptions): Promise<DeriveAddressResult> {
+    throw new NotImplementedError('deriveAddress is not supported for this coin');
+  }
 
   /**
    * convert address into desired address format.
--- a/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
+++ b/modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
@@ -210,6 +210,56 @@ export function isTssVerifyAddressOptions<T extends VerifyAddressOptions | TssVe
   );
 }
 
+/**
+ * Options for locally deriving a wallet receive address from a derivation path.
+ *
+ * Unlike {@link VerifyAddressOptions}, which checks a candidate address, these options are
+ * used to *produce* the address offline from public key material only:
+ * - the user/backup/bitgo xpub triple (`pub`) for BIP32 multisig coins, or
+ * - the `commonKeychain` for TSS/MPC coins.
+ *
+ * No private keys and no network access are required.
+ */
+export interface DeriveAddressOptions
+  extends Pick<VerifyAddressOptions, 'format' | 'derivedFromParentWithSeed' | 'multisigTypeVersion'> {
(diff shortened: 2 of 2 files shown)


rajangarg047:
@claude

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
```diff
@@ -210,6 +210,67 @@ export function isTssVerifyAddressOptions<T extends VerifyAddressOptions | TssVe
...
+  /** Optional address format override (e.g. UTXO script type). */
+  format?: CreateAddressFormat;
+  /**
+   * Public key material for derivation.
+   * - BIP32 multisig (UTXO, legacy EVM): the user/backup/bitgo xpub triple via `pub`.
+   * - TSS/MPC (SOL, EVM MPC, etc.): the `commonKeychain` (identical across keychains).
+   */
+  keychains?: {
+    pub?: string;
+    commonKeychain?: string;
```
pranavjain97: `pub` and `commonKeychain` should be a union type (`{ pub: string } | { commonKeychain: string }`) rather than two optional fields, so passing both is a type error.

pranavjain97 (dismissed):
lgtm, small comment

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
rajangarg047: Good call — done in 85e480479. Changed `keychains` to `({ pub: string } | { commonKeychain: string })[]` so a keychain with neither field is now a compile error. I kept it a non-exclusive union (rather than the `{ pub; commonKeychain?: never }` form) on purpose: TSS keychains commonly expose `pub` *alongside* `commonKeychain`, so forbidding both would reject valid inputs.

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
```diff
@@ -352,6 +354,20 @@ export abstract class BaseCoin implements IBaseCoin {
...
 
+  /**
+   * Locally derive and return a wallet receive address for the given derivation path,
+   * using public key material only (no private keys, no network access).
+   *
+   * This is the inverse of {@link isWalletAddress}: rather than checking a candidate
+   * address, it *produces* the address from the wallet's keychains and a chain/index.
+   *
+   * Coins opt in by overriding this method; the default implementation throws.
+   * @param params
```
mohammadalfaiyazbitgo:
```suggestion
    * @param {DeriveAddressOptions} params - The parameters for address derivation
 * @returns {Promise<DeriveAddressResult>} The derived address and related information
```

modules/sdk-core/src/bitgo/baseCoin/iBaseCoin.ts
```diff
@@ -210,6 +210,67 @@ export function isTssVerifyAddressOptions<T extends VerifyAddressOptions | TssVe
...
+   */
+  derivedFromParentWithSeed?: string;
+  /**
+   * Identifies the MPC signing protocol version of the wallet (e.g. 'MPCv2').
+   * Used to distinguish between MPCv1 and MPCv2 wallets.
+   */
+  multisigTypeVersion?: 'MPCv2';
+  /** Wallet version, used to disambiguate derivation strategy for some coin families. */
+  walletVersion?: number;
+}
```
mohammadalfaiyazbitgo: can any of these be derived from other common types in `sdk-core` ?

rajangarg047: Good point — yes. Updated in f5c2224b4:

- `DeriveAddressOptions` now `extends Pick<VerifyAddressOptions, 'format' | 'derivedFromParentWithSeed' | 'multisigTypeVersion'>` — those three were identical to the verify options, so they're now reused rather than redeclared (and stay in sync automatically).
- `DeriveAddressResult` now `extends AddressVerificationData` (`chain`, `index`, `coinSpecific`), adding only `address` and `derivationPath`.

`format` was already `CreateAddressFormat` and `coinSpecific` already `AddressCoinSpecific`. The remaining fields I kept explicit on purpose: `index` is required `number` here (it's optional / `number | string` elsewhere), and `keychains` is a deliberately stricter union (`{ pub } | { commonKeychain }`) than the verify types' shapes.

modules/sdk-core/src/bitgo/baseCoin/baseCoin.ts
rajangarg047: Done in f5c2224b4 — added the `@param` / `@returns` JSDoc to the default `deriveAddress`.

mohammadalfaiyazbitgo (approved):

PR #9001, opened 2026-06-11
refactor(sdk-core): extract deriveMPCWalletAddress from verifyMPCWalletAddress

rajangarg047:
## Summary
`verifyMPCWalletAddress` already computed the expected address internally and then compared it to a candidate. This PR extracts that derivation half into a reusable `deriveMPCWalletAddress()` that returns the derived address **and** the HD path used, and has `verifyMPCWalletAddress()` delegate to it before comparing.

This lets callers *produce* an MPC wallet address offline (public keys only) using the exact same code path as verification — so derive and verify can never diverge. **Behavior of `verifyMPCWalletAddress` is unchanged.**

## Context
Part of FR-465 (Bullish) local address derivation. The shared derive function is what the SOL (and other EdDSA/ECDSA TSS) `deriveAddress` implementations will call.

- Linear: WCN-913 (child of WCN-864)
- Pure refactor: no new imports, identical derivation logic, one new exported function.

## Test plan
- [x] `tsc --noEmit` passes for `sdk-core`
- [x] Existing `addressVerification` unit tests pass (run recursively; the 4 unrelated `EdDSA MPCv2 SMC Utils` failures pre-date this branch)
- [x] Added unit coverage for `deriveMPCWalletAddress`: ed25519 MPCv2, SMC prefix path, secp256k1, plus a **derive→verify round-trip** assertion
- [ ] CI

### DIFF
--- a/modules/sdk-core/src/bitgo/utils/tss/addressVerification.ts
+++ b/modules/sdk-core/src/bitgo/utils/tss/addressVerification.ts
@@ -49,28 +49,35 @@ export async function verifyEddsaTssWalletAddress(
 }
 
 /**
- * Verifies if an address belongs to a wallet using ECDSA TSS MPC derivation.
- * This is a common implementation for ECDSA-based MPC coins (ETH, BTC, etc.)
+ * Options for deriving an MPC wallet address. This is the subset of
+ * {@link TssVerifyAddressOptions} needed to *produce* an address (no `address` to check),
+ * plus the key curve.
+ */
+export type DeriveMPCWalletAddressOptions = Pick<
+  TssVerifyAddressOptions,
+  'keychains' | 'index' | 'derivedFromParentWithSeed' | 'multisigTypeVersion'
+> & {
+  keyCurve: 'secp256k1' | 'ed25519';
+};
+
+/**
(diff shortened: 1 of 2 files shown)


rajangarg047:
@claude

modules/sdk-core/src/bitgo/utils/tss/addressVerification.ts
```diff
@@ -49,28 +49,35 @@ export async function verifyEddsaTssWalletAddress(
...
+ * rather than comparing it against a candidate, so the two can never diverge.
+ *
+ * @param params - keychains (commonKeychain), derivation index, optional seed/version, and keyCurve
  * @param getAddressFromPublicKey - Coin-specific function to convert public key to address
- * @returns true if the address matches the derived address, false otherwise
- * @throws {InvalidAddressError} if the address is invalid
+ * @returns the derived address and the HD derivation path used to derive it
  * @throws {Error} if required parameters are missing or invalid
  */
-export async function verifyMPCWalletAddress(
```
mohammadalfaiyazbitgo, line 62: this is a breaking change.

mohammadalfaiyazbitgo (changes requested):

mohammadalfaiyazbitgo (approved):

PR #9013, opened 2026-06-12
feat(express): add POST /api/v2/:coin/address/derive endpoint

rajangarg047:
## Summary
Adds an **offline** BitGo Express endpoint that locally **derives and returns** a wallet receive address from a derivation path — the inverse of `iswalletaddress`.

```
POST /api/v2/{coin}/address/derive
```

The handler operates purely on the request body (`keychains` + `chain`/`index`) via `coin.deriveAddress(...)` — **no `wallets().get` lookup and no network access**, so it can run in an air-gapped Express. It's **stateless**: the caller supplies the index; the endpoint never allocates server-side. Pairs with `iswalletaddress` for a client **derive→verify** round-trip.

## Context
Builds on the merged `deriveAddress` primitive (WCN-912) and shared `deriveMPCWalletAddress` (WCN-913). This is the Phase-0 endpoint for FR-465 (Bullish local address derivation). Per-coin `deriveAddress` implementations (BTC/ETH/SOL) land in follow-up tickets; until a coin implements it, the base `deriveAddress` throws `NotImplementedError` (surfaced as a 500 here, covered by a test).

- Linear: WCN-914 (child of WCN-864)

## Changes
- `modules/express/src/typedRoutes/api/v2/deriveAddress.ts` — typed route schema (params/body/response). Keychain modelled as a `{ pub } | { commonKeychain }` union so a keychain with neither is rejected.
- `typedRoutes/api/index.ts` — registered the spec (import, apiSpec, type intersection, spread).
- `clientRoutes.ts` — `handleV2DeriveAddress` + route registration (uses existing `prepareBitGo` auth middleware; no extra access control).

## Test plan
- [x] `tsc --noEmit` clean for `modules/express` (the only errors are pre-existing `@iota/*` dependency `.d.ts` noise handled by the build's `skipLibCheck`; 0 in app source)
- [x] New codec + supertest integration tests (24): UTXO + TSS/MPC derive, `derivedFromParentWithSeed` passthrough, 400 on missing `index`/`keychains`, 500 when derivation throws. Success cases assert the offline path (mock coin exposes no `wallets()`).
- [ ] CI

### DIFF
--- a/modules/express/src/clientRoutes.ts
+++ b/modules/express/src/clientRoutes.ts
@@ -722,6 +722,20 @@ export async function handleV2IsWalletAddress(
 }
 
 /**
+ * handle v2 deriveAddress - locally derive and return a wallet receive address from a
+ * derivation path, using public key material only.
+ *
+ * Offline by design: operates purely on the request body (keychains + chain/index), with no
+ * `wallets().get` lookup and no network access. The inverse of {@link handleV2IsWalletAddress}.
+ * @param req
+ */
+export async function handleV2DeriveAddress(req: ExpressApiRouteRequest<'express.v2.address.derive', 'post'>) {
+  const bitgo = req.bitgo;
+  const coin = bitgo.coin(req.decoded.coin);
+  return await coin.deriveAddress(req.decoded as any);
+}
+
+/**
  * handle v2 approve transaction
  * @param req
  */
@@ -1963,6 +1977,7 @@ export function setupAPIRoutes(app: express.Application, config: Config): void {
     prepareBitGo(config),
     typedPromiseWrapper(handleV2IsWalletAddress),
   ]);
+  router.post('express.v2.address.derive', [prepareBitGo(config), typedPromiseWrapper(handleV2DeriveAddress)]);
 
   router.post('express.wallet.share', [prepareBitGo(config), typedPromiseWrapper(handleV2ShareWallet)]);
   app.post(
--- a/modules/express/src/typedRoutes/api/index.ts
+++ b/modules/express/src/typedRoutes/api/index.ts
@@ -58,6 +58,7 @@ import { PostWalletEnableTokens } from './v2/walletEnableTokens';
 import { PostWalletSweep } from './v2/walletSweep';
 import { PostWalletAccelerateTx } from './v2/walletAccelerateTx';
 import { PostIsWalletAddress } from './v2/isWalletAddress';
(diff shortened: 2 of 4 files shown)


rajangarg047:
@claude

rajangarg047:
Thanks for the review. Addressed in 88d811002:

**✅ Applied — non-negative integer validation.** Added a `NonNegativeInteger` refinement codec and applied it to `index` (required) and `chain` (optional). Derivation indices/chains are always non-negative integers, so negative/fractional values are now rejected at the codec layer (400). Added tests for negative index, non-integer index, and negative chain.

**⏭️ Skipped — wrapping `NotImplementedError` into a 400.** Kept for parity with the sibling `iswalletaddress` handler, which does no bespoke error wrapping. Matching on `error.message.includes('NotImplementedError')` is also fragile (it's a class, not a message substring). Once each coin implements `deriveAddress` (the follow-up Stage A/B/C tickets), supported coins return 200; unsupported ones surfacing a 500 is acceptable and consistent with existing endpoints.

**⏭️ Skipped — JSDoc usage examples.** The route description already documents the offline/derive→verify behavior; left as-is to avoid noise.

Test count is now 28 (all green); tsc + eslint clean.

rajangarg047:
Fixed the API-spec build failure in 3f286af39.

**Cause:** the `NonNegativeInteger` codec I'd added used `t.refinement`, which `@api-ts/openapi-generator` serializes as a `$ref` to `#/components/schemas/refinement` — a schema that doesn't exist, breaking spec generation for the `index`/`chain` fields.

**Fix:** reverted `index`/`chain` to plain `t.number`, matching the sibling endpoints (`createAddress`, `isWalletAddress`) which all use plain `t.number` for these fields. Dropped the 3 refinement-specific tests (negative/non-integer/negative-chain); the string-index rejection test stays (plain `t.number` still enforces it).

**Verified locally** by running the exact `audit-api-spec` CI command:
```
openapi-generator --codec-file modules/express/openapi-generator.rc.js modules/express/src/typedRoutes/api/index.ts
```
→ exits 0, no errors; `NonNegativeInteger` no longer in the spec; the `/api/v2/{coin}/address/derive` `index`/`chain` schemas are now plain `{"type":"number"}`.

Note: non-negative validation *could* be preserved spec-safely via `t.brand` with an inline predicate (as another module here does), but plain `t.number` keeps parity with the existing address endpoints, so I went with that.

pranishnepal (approved):
lgtm

PR #9014, opened 2026-06-12
fix(sdk-core): skip keychain fetch in createAddress for OFC wallets

zahin-mohammad:
## Summary
- `wallet.createAddress()` was unconditionally fetching all wallet keys before creating an address, including server-managed keys at index 1 that have no accessible keychain record in the OFC namespace
- OFC coins never consume those keychains — `isWalletAddress` always throws `MethodNotImplementedError` and verification is skipped — so the fetch was unnecessary
- Added guard: skip the keychain fetch entirely when `baseCoin.getFamily() === 'ofc'`

Ticket: WCN-942


pranishnepal (approved):
LGTM

alextse-bg (approved):

PR #9029, opened 2026-06-15
feat: add exclusions for new CVEs affecting dependencies and clarify usage context

mohd-kashif:
Ticket: CECHO-1353


bhargavirao24 (approved):
Approving this again since it is blocking the SDK release.

That said from a Security perspective, we would prefer not to keep these CVEs ignored in the exclude path long-term. I see there is already a follow-up ticket to explore revisiting the ignored SDK advisories which should help improve this process going forward:

https://linear.app/bitgo/issue/WCN-959/explore-using-night-agent-to-revisit-ignored-sdk-advisories

optimizedalgo (approved):
LGTM

PR #9064, opened 2026-06-18
feat(root): configure nightshift with iyarc-prune task

rashadjnizar:
Ticket: HSM-410

## Summary

Configures a Nightshift agent to automatically prune stale entries from `.iyarc`.
Adds three files:

- `.github/workflows/nightshift-scheduler.yaml` — weekday cron (06:00 UTC) that calls the reusable scheduler; 168h cooldown gates actual dispatch to ~weekly.
- `.github/workflows/nightshift-task.yaml` — thin wrapper calling the reusable task runner.
- `.nightshift.yaml` — the `iyarc-prune` custom task definition + prompt.

## How it works

Each cycle the agent walks every `.iyarc` exclusion, checks whether a compatible patched version exists, and if so bumps the dependency and removes the exclusion. Before opening a PR it runs the **release gates — `audit-high` + `check-deps` — plus a scoped build/test**, and abandons the change if any gate fails. Green runs open one non-draft PR. No-op cycles are expected and produce no PR.

## Testing

Validated the scheduler + config locally via the `nightshift-actions` CLI (offline dry-run):

```
Loaded 1 enabled tasks
--- Selected Tasks ---
  iyarc-prune (score: 18.00, priority: 8, staleness: 10.00, reason: never run)
Dry run — no tasks dispatched.
```

Confirms `.nightshift.yaml` parses and the task is selected correctly.

### DIFF
--- a/.nightshift.yaml
+++ b/.nightshift.yaml
@@ -1,0 +1,168 @@
+max_tasks_per_run: 1
+
+tasks:
+  enabled:
+    - iyarc-prune
+
+  priorities:
+    iyarc-prune: 8
+
+  cooldowns:
+    iyarc-prune: '168h'
+
+  custom:
+    - type: iyarc-prune
+      name: Audit Exclusion (.iyarc) Pruner
+      category: pr
+      cooldown: '168h'
+      priority: 8
+      prompt: |
+        You are a nightshift automated maintenance agent for the BitGoJS
+        monorepo. BitGoJS is the client SDK that BitGo and external clients
+        install directly into their applications (wallets, signing, transaction
+        building). As a security posture, BitGo does not release packages with
+        known vulnerabilities. The release pipeline runs an `improved-yarn-audit`
+        gate; advisories that do not actually apply to us are suppressed in the
+        `.iyarc` ignore file at the repo root, each with a justification comment.
+
+        Over time `.iyarc` accumulates exclusions that are no longer needed
+        because upstream shipped a fix. Nobody prunes them, so the suppressed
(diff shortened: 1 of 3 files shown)


vinhkhangtieu:
what does iyarc do? is this pr mainly for security audit of packages in that file?

vinhkhangtieu (dismissed):

rashadjnizar:
> what does iyarc do? is this pr mainly for security audit of packages in that file?

iyarc is the ignore/exclusion file for `improved-yarn-audit`. this pr adds a scheduled nightshift agent that walks every entry in .iyarc, checks whether a compatible patched version now exists, and if so bumps the dependency and removes the exclusion.

Marzooqa (dismissed):

zahin-mohammad (dismissed):
good job!

.nightshift.yaml
```diff
@@ -0,0 +1,168 @@
...
+    iyarc-prune: '168h'
+
+  custom:
+    - type: iyarc-prune
+      name: Audit Exclusion (.iyarc) Pruner
+      category: pr
+      cooldown: '168h'
+      priority: 8
+      prompt: |
+        You are a nightshift automated maintenance agent for the BitGoJS
```
zahin-mohammad, line 20: One suggestion: If this works well, we should create a claude skill for this and refer to the skill in this prompt. Makes it useable for both devs, agents and this nightshfit task.

zahin-mohammad (approved):

.nightshift.yaml
rashadjnizar: Sounds good.

PR #9067, opened 2026-06-19
fix(sdk-core): read defi vault operationId from coinSpecific

kamleshmugdiya:
## Summary

`DefiVault.depositToVault()` threw `Error: operationId not found in approve txRequest response` on every deposit. The approve `sendMany` succeeded and the WP minted an operationId, but the SDK couldn't locate it and aborted before issuing the deposit `sendMany`.

**Root cause:** a contract mismatch on *where* `operationId` lives in the approve txRequest response.

- **SDK** — `extractOperationId()` read `txRequest.intent.operationId`.
- **WP** — writes the defi-service-minted operationId into the built transaction's `coinSpecific` (next to `assignedNonce`), which for a `full` txRequest surfaces at `txRequest.transactions[0].unsignedTx.coinSpecific.operationId`. The `intent` block never carries it.

Both the SDK comment and the WP docstring *claimed* "operationId on the intent," and the SDK unit test mocked the response that way — so the bug shipped with green tests on both sides.

## Changes

- `extractOperationId()` now reads `transactions[0].unsignedTx.coinSpecific.operationId` (full apiVersion), falling back to `unsignedTxs[0].coinSpecific.operationId` (lite) and finally `intent.operationId` (forward-compat, in case WP later also populates the intent).
- Updated the misleading docstring in `defiVault.ts`.
- Fixed the stale unit-test mocks that put `operationId` on the `intent` and masked the regression — the approve mocks now use the real `coinSpecific` shape.
- Added regression tests covering the lite shape, the intent forward-compat fallback, and the not-found path.

## Test plan

- [x] `mocha test/unit/bitgo/defi/defiVault.ts` — 17 passing, 1 pending (CGD-1709 placeholder)
- [x] `eslint` — 0 errors
- [x] `prettier --check` — clean

Ticket: CGD-1829

### DIFF
--- a/modules/sdk-core/src/bitgo/defi/defiVault.ts
+++ b/modules/sdk-core/src/bitgo/defi/defiVault.ts
@@ -220,12 +220,39 @@ export class DefiVault implements IDefiVault {
   }
 
   /**
-   * Extract operationId from the intent of a sendMany result.
-   * The WP populates operationId in the intent of the approve txRequest.
+   * Extract operationId from a sendMany result.
+   *
+   * The WP writes the defi-service-minted operationId into the built
+   * transaction's `coinSpecific` (alongside `assignedNonce`), not into the
+   * intent. Read it from there: the `full` apiVersion surfaces it at
+   * `transactions[0].unsignedTx.coinSpecific.operationId`, the `lite` version
+   * at `unsignedTxs[0].coinSpecific.operationId`. Fall back to
+   * `intent.operationId` for forward-compat in case the WP later also
+   * populates the intent.
    */
   private extractOperationId(sendManyResult: Record<string, unknown>): string | undefined {
     const txRequest = sendManyResult.txRequest as Record<string, unknown> | undefined;
-    const intent = txRequest?.intent as Record<string, unknown> | undefined;
+    if (!txRequest) {
+      return undefined;
+    }
+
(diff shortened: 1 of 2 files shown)


modules/sdk-core/src/bitgo/defi/defiVault.ts
```diff
@@ -220,12 +220,39 @@ export class DefiVault implements IDefiVault {
...
    */
   private extractOperationId(sendManyResult: Record<string, unknown>): string | undefined {
     const txRequest = sendManyResult.txRequest as Record<string, unknown> | undefined;
-    const intent = txRequest?.intent as Record<string, unknown> | undefined;
+    if (!txRequest) {
+      return undefined;
+    }
+
+    // full apiVersion: transactions[0].unsignedTx.coinSpecific.operationId
+    const transactions = txRequest.transactions as Array<Record<string, unknown>> | undefined;
```
hafeezshaik933, line 240: nit: we can have types declared and re-use.

hitansh-madan (approved):

modules/sdk-core/src/bitgo/defi/defiVault.ts
kamleshmugdiya: Will do it in a follow-up.

PR #9082, opened 2026-06-22
fix(sdk-coin-*): pass isFirstSigner through transfer() in all EVM TransactionBuilders

mr-neptune:
## Description

All non-eth-family EVM coin modules overrode transfer(data?: string) without forwarding isFirstSigner to the TransferBuilder constructor, breaking the Trust HSM first-signer path. from(txHex, true) decoded first-signer ABI with second-signer offsets, producing 0x...00c0 as recipient instead of the real address, causing verifyExistingApproval 401 errors on TETC and tarbeth.

## Changes
Fixes 23 modules: arbeth, opeth, avaxc, celo, rbtc, bera, polygon, zketh, bsc, flr, sgb, apechain, coredao, xdc, mon, soneium, wemix, stt, oas, world, ethlike, evm, and etc.

## Tests
Adds regression tests on sdk-coin-etc for the first-signer round-trip.

TICKET: WCN-560
Past related issue: 
https://github.com/BitGo/bitgo-microservices/pull/59310
https://github.com/BitGo/bitgo-microservices/pull/59699
https://github.com/BitGo/bitgo-microservices/pull/59905

### DIFF
--- a/modules/sdk-coin-apechain/src/lib/transactionBuilder.ts
+++ b/modules/sdk-coin-apechain/src/lib/transactionBuilder.ts
@@ -15,12 +15,12 @@ export class TransactionBuilder extends AbstractTransactionBuilder {
   }
 
   /** @inheritdoc */
-  transfer(data?: string): TransferBuilder {
+  transfer(data?: string, isFirstSigner?: boolean): TransferBuilder {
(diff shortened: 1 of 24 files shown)


at31416 (approved):
lgtm

PR #9084, opened 2026-06-22
fix: use query params for GET address-book methods

seanchen0818:
## Summary
- `getConnections`, `getListingEntryContacts`, and `getListingEntryDirectory` were using `.send(params)` on GET requests, which puts params in the request body — ignored by the server
- Replaced with `.query(params)` so filters are correctly sent as URL query params
- Confirmed against address-book service SDK spec: all three routes define filters under `query:`, not `body:`

## Test plan
- [ ] Added unit tests for all three GET methods verifying `.query()` is called with the provided params
- [ ] All 4 tests pass locally

Fixes GN-2899

### DIFF
--- a/modules/sdk-core/src/bitgo/address-book/address-book.ts
+++ b/modules/sdk-core/src/bitgo/address-book/address-book.ts
@@ -50,7 +50,11 @@ export class AddressBook implements IAddressBook {
    */
   getConnections(params?: GetAddressBookConnectionsParams): Promise<GetAddressBookConnectionsResponse> {
     const url = this.bitgo.microservicesUrl('/api/address-book/v1/connections');
-    return this.bitgo.get(url).set('enterprise-id', this.enterpriseId).send(params).result();
+    return this.bitgo
+      .get(url)
+      .set('enterprise-id', this.enterpriseId)
+      .query(params ?? {})
+      .result();
   }
 
   /**
@@ -84,7 +88,6 @@ export class AddressBook implements IAddressBook {
     const response: GetAddressBookListingResponse = await this.bitgo
       .get(url)
       .set('enterprise-id', this.enterpriseId)
-      .send()
       .result();
     this._listing = response;
     return this.listing() as AddressBookListing;
@@ -125,7 +128,11 @@ export class AddressBook implements IAddressBook {
     params?: GetAddressBookListingEntryContactsParams
   ): Promise<GetAddressBookListingEntryContactsResponse> {
     const url = this.bitgo.microservicesUrl('/api/address-book/v1/listing/entry/contacts');
-    return this.bitgo.get(url).set('enterprise-id', this.enterpriseId).send(params).result();
+    return this.bitgo
+      .get(url)
+      .set('enterprise-id', this.enterpriseId)
+      .query(params ?? {})
+      .result();
   }
 
   /**
@@ -135,7 +142,11 @@ export class AddressBook implements IAddressBook {
     params?: GetAddressBookListingEntryDirectoryParams
   ): Promise<GetAddressBookListingEntryDirectoryResponse> {
     const url = this.bitgo.microservicesUrl('/api/address-book/v1/listing/entry/directory');
-    return this.bitgo.get(url).set('enterprise-id', this.enterpriseId).send(params).result();
+    return this.bitgo
+      .get(url)
+      .set('enterprise-id', this.enterpriseId)
+      .query(params ?? {})
+      .result();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


pranishnepal:
Tests failing

modules/sdk-core/src/bitgo/address-book/address-book.ts
lokesh-bitgo, line 1: **Nit (follow-up):** `getListing()` (line ~82) also calls `.send()` on a GET with no arguments — same anti-pattern as the methods fixed here. Less harmful since there are no params to drop, but it's inconsistent in the same class. Consider cleaning it up in a follow-up: just remove the `.send()` call entirely since `.get().set(...).result()` is sufficient for a parameterless GET.

modules/sdk-core/test/unit/bitgo/address-book/address-book.ts
```diff
@@ -0,0 +1,73 @@
...
+
+describe('AddressBook', function () {
+  let addressBook: AddressBook;
+  let mockBitGo: any;
+  const enterpriseId = 'test-enterprise-id';
+
+  function makeGetStub() {
+    const queryStub = sinon.stub().returns({ result: sinon.stub().resolves({}) });
+    const setStub = sinon.stub().returns({ query: queryStub });
+    mockBitGo.get.returns({ set: setStub });
```
lokesh-bitgo, line 14: **Low:** `mockBitGo.get` is stubbed to accept any URL — a typo in the endpoint path (e.g. `/connections` → `/connection`) would go undetected. Consider asserting the URL too:

```typescript
sinon.assert.calledWith(mockBitGo.get, 'https://app.bitgo.com/api/address-book/v1/connections');
```

modules/sdk-core/test/unit/bitgo/address-book/address-book.ts
```diff
@@ -0,0 +1,73 @@
...
+    it('should pass params as query string, not request body', async function () {
+      const { queryStub } = makeGetStub();
+      const params = { status: 'ACTIVE' as const, limit: 5 };
+
+      await addressBook.getListingEntryContacts(params);
+
+      sinon.assert.calledOnce(queryStub);
+      sinon.assert.calledWith(queryStub, params);
+    });
+  });
```
lokesh-bitgo, line 132: **Low:** `getConnections` has a no-params test verifying `undefined → {}` (the `?? {}` path), but `getListingEntryContacts` and `getListingEntryDirectory` only have one test each (with params). Consider adding a no-params case for both:

```typescript
it('should work with no params', async function () {
  const { queryStub } = makeGetStub();
  await addressBook.getListingEntryContacts();
  sinon.assert.calledWith(queryStub, {});
});
```

modules/sdk-core/src/bitgo/address-book/address-book.ts
```diff
@@ -50,7 +50,11 @@ export class AddressBook implements IAddressBook {
    */
   getConnections(params?: GetAddressBookConnectionsParams): Promise<GetAddressBookConnectionsResponse> {
     const url = this.bitgo.microservicesUrl('/api/address-book/v1/connections');
-    return this.bitgo.get(url).set('enterprise-id', this.enterpriseId).send(params).result();
+    return this.bitgo
+      .get(url)
+      .set('enterprise-id', this.enterpriseId)
+      .query(params ?? {})
```
lokesh-bitgo, line 56: **Nit:** `GetAddressBookConnectionsParams` includes array fields (`ownerWalletId?: string[]`, `targetWalletId?: string[]`). Superagent's default `.query()` serializes arrays as repeated keys — `?ownerWalletId=a&ownerWalletId=b`. Worth verifying the address-book service parses that format (vs bracket notation `ownerWalletId[]=a` or comma-separated). If the format doesn't match, these filters would silently not work.

lokesh-bitgo (commented):
The core fix is correct — `.send(params)` on GET was silently dropping all filter params since most servers ignore the request body on GET requests. `.query(params ?? {})` is the right approach. CI is green and the logic is sound.

Leaving a few non-blocking suggestions inline.

modules/sdk-core/src/bitgo/address-book/address-book.ts
seanchen0818: updated!

modules/sdk-core/test/unit/bitgo/address-book/address-book.ts
seanchen0818: updated!

modules/sdk-core/test/unit/bitgo/address-book/address-book.ts
seanchen0818: updated!

modules/sdk-core/src/bitgo/address-book/address-book.ts
seanchen0818: updated!

lokesh-bitgo (approved):
looks good

pranishnepal (approved):

PR #9086, opened 2026-06-23
fix(sdk-coin-flr): support FLR C->P atomic TSS signing

ArunBala-Bitgo:
FLR C-chain to P-chain export (and the matching P-chain import) fails during the MPC signing ceremony because:

1. sdk-coin-flr does not declare isSignablePreHashed, so the MPCv2 signing flow re-hashes the Avalanche atomic signableHex with keccak256. The signableHex is already SHA-256(txBody), so the user's signature share is computed over a different digest from BitGo's, surfacing as "Failed to combine signature shares" on the send-transaction endpoint.

2. The TSS recipient guard rejects intent types 'import' and 'importtoc' (Avalanche / Flare cross-chain atomic imports), even though those intents legitimately have no client-supplied recipients — the wallet imports its own UTXOs and the destination is the wallet itself. That surfaces as "Recipient details are required to verify this transaction before signing".

Add isSignablePreHashed to Flr mirroring the Flrp implementation, and add 'import' / 'importtoc' to NO_RECIPIENT_TX_TYPES so the pre-signing verifier no longer blocks the flow.

Ticket: CECHO-1425

### DIFF
--- a/modules/sdk-coin-flr/src/flr.ts
+++ b/modules/sdk-coin-flr/src/flr.ts
@@ -19,10 +19,12 @@ import {
   common,
   FeeEstimateOptions,
   IWallet,
+  isAvalancheAtomicTx,
   MPCAlgorithm,
   MultisigType,
   multisigTypes,
   Recipient,
+  SignableTransaction,
   TransactionExplanation,
   Entry,
 } from '@bitgo/sdk-core';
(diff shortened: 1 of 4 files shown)


mohd-kashif (approved):

ArunBala-Bitgo:
@claude review this PR as grug

PR #9089, opened 2026-06-23
fix(statics): correct hoodeth:usdg decimals from 18 to 6

yashvanthbl137-crypto:
Ticket: CGD-1081

update the decimal of HOODETH:USDG from 18 to 6


HimanshuS2212 (approved):
LGTM

PR #9091, opened 2026-06-23
fix(sdk-core): exempt FLRP import from TSS recipient guard 

ArunBala-Bitgo:
signTransactionTss passes buildParams.type as 'Import' while WP stores intentType as 'import', so the WCN-151 recipient guard still threw after adding lowercase import to NO_RECIPIENT_TX_TYPES. Also check intentType when resolving the exemption.

Co-authored-by: Cursor <cursoragent@cursor.com>

TICKET: CECHO-1425


mohd-kashif (dismissed):

manojkumar138 (approved):
LG

mohd-kashif (approved):

PR #9104, opened 2026-06-24
fix(sdk-coin-sol): support unsupported tokens in verifyTransaction

abhishekagrawal080:
TICKET: CSHLD-1111

  sol.ts — fixed the setter call and added fallback logic:                                                                                                                                                                                          
  - transaction.setUseTokenAddressTokenName(true) (line 537) — uses the correct setter                                                                                                                                                              
  - filteredRecipients pick includes tokenAddress and programId (line 574)
  - ATA derivation (lines 612–624) — falls back to recipientFromUser.tokenAddress / programId when the static map has no entry                                                                                                                      
  - Consolidation path (lines 648–663) — uses the mint address as-is for tokens not in the static map

  iBaseCoin.ts — ITransactionRecipient now has optional tokenAddress?: string and programId?: string

  test/unit/sol.ts — 4 new tests covering:
  1. Unsupported SPL token (TOKEN_PROGRAM_ID) with tokenAddress+programId → passes
  2. Unsupported SPL token without those fields → rejected
  3. Unsupported Token-2022 token with correct programId → passes
  4. Unsupported Token-2022 token with wrong programId → rejected

### DIFF
--- a/modules/sdk-coin-sol/src/sol.ts
+++ b/modules/sdk-coin-sol/src/sol.ts
@@ -533,6 +533,8 @@ export class Sol extends BaseCoin {
     }
 
     const transaction = new Transaction(coinConfig);
+    // Allow unsupported tokens (not in static map) to use their mint address as tokenName during parsing
+    transaction.setUseTokenAddressTokenName(true);
     const rawTx = txPrebuild.txBase64 || txPrebuild.txHex;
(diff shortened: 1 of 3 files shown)


alextse-bg (approved):
wallet changes lgtm

mukeshsp (approved):

PR #9126, opened 2026-06-26
feat(root): self-contained scheduled iyarc-prune workflow

rashadjnizar:
Ticket: HSM-429

## Summary

Configures a scheduled agent to automatically prune stale entries from .iyarc (the improved-yarn-audit exclusion list). This is a self-contained replacement for the earlier nightshift-actions approach, a public repo can't call the private nightshift-actions reusable workflow, so this runs Claude Code directly via Bedrock, reusing the same OIDC + inference role already proven in claude-pr.yml.

## Changes:

- Adds .github/workflows/iyarc-prune.yml: weekly cron (Mondays 06:00 UTC) + manual workflow_dispatch; OIDC → Bedrock auth, then runs claude-code-action with commit signing enabled.
- Adds .github/prompts/iyarc-prune.md: the prune task prompt.
- Removes the 3 old nightshift-actions files (nightshift-scheduler.yaml, nightshift-task.yaml, .nightshift.yaml).

## How it works

Each run the agent walks every .iyarc exclusion, checks whether a compatible patched version exists, and if so bumps the dependency and removes the exclusion. Before opening a PR it runs the release gates — audit-high + check-deps —  plus a scoped build/test, and abandons the change if any gate fails. Green runs open one assigned non-draft PR, with commits signed via use_commit_signing. No-op runs are expected and produce no PR.

## Testing

The nightshift scheduler dry-run no longer applies. Validated locally instead:
- `actionlint .github/workflows/iyarc-prune.yml` — valid, no errors
- `act -l` — job/trigger plan parses: `prune  iyarc-prune  schedule,workflow_dispatch`
- Release gates run against the current tree (the exact gates the agent must pass):
    - `yarn run audit-high` → 0 vulnerabilities
    - `yarn check-deps` → pass
    
Full agent run (OIDC → Bedrock → commit/PR) can't execute locally; it will be verified post-merge via a manual `workflow_dispatch`.


OttoAllmendinger (dismissed):

pranishnepal (dismissed):

.github/prompts/iyarc-prune.md
```diff
@@ -0,0 +1,141 @@
+You are an automated maintenance agent for the BitGoJS monorepo, run on a
```
zahin-mohammad: This seems better organized as an agent file 

> User subagents (~/.claude/agents/) are personal subagents available in all your projects.
Claude Code scans .claude/agents/ and ~/.claude/agents/ recursively, so you can organize definitions into subfolders such as agents/review/ or agents/research/. The subdirectory path doesn’t affect how a subagent is identified or invoked, because identity comes only from the name frontmatter field. Keep name values unique across the whole tree: if two files within one scope declare the same name, Claude Code keeps one and discards the other without warning.


https://code.claude.com/docs/en/sub-agents

this way its available to all devs running claude in bitgojs as well

.github/workflows/iyarc-prune.yml
```diff
@@ -0,0 +1,85 @@
...
+          {
+            echo "prompt<<IYARC_PRUNE_EOF"
+            cat .github/prompts/iyarc-prune.md
+            echo "IYARC_PRUNE_EOF"
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Run Claude Code (Bedrock)
+        uses: anthropics/claude-code-action@v1
+        with:
+          prompt: ${{ steps.read-prompt.outputs.prompt }}
```
zahin-mohammad: With a dedicated agent file, your prompt becomes verify simple


```suggestion
      - name: Run Claude Code (Bedrock)
        uses: anthropics/claude-code-action@v1
        with:
          prompt: "Use the <name> subagent to do XYZ"
```

or another option if we don't want to use a subagent: 

```suggestion
      - name: Run Claude Code (Bedrock)
        uses: anthropics/claude-code-action@v1
        with:
          prompt: ${{ steps.read-prompt.outputs.prompt }}
          claude_args: |
             --append-system-prompt "$(cat .claude/agents/<name>.md)"
```

.github/prompts/iyarc-prune.md
rashadjnizar: Done, moved it to a committed subagent at `.claude/agents/iyarc-prune.md`

.github/workflows/iyarc-prune.yml
rashadjnizar: Took the spirit of this, the workflow now loads the agent file. I didn't use either snippet verbatim since: 

- '--append-system-prompt "$(cat ...)"' would pass '$(cat ...)' literally in the Actions 'with:' block, and not evaluated like a shell command.
- The shorter 'prompt: "Use the <name> subagent..."' is documented for claude usage, but I wasn't confident relying on subagent delegation for a non interactive workflow.

Instead the workflow now reads the file, strips the frontmatter, and passes the agent body through the normal 'prompt:' input. This keeps the file available for devs and the workflow deterministic.

zahin-mohammad (approved):

PR #9128, opened 2026-06-26
feat: add external signer support for ECDSA and EdDSA TSS key generation

danielpeng1:
Add external-signer callback support for MPC/TSS wallet generation, in particular enabling `advanced-wallets` to delegate DKG orchestration to the SDK.

- ECDSA (MPCv2): New `createKeychainsWithExternalSigner` runs the full DKG in four steps; handles init, round 2, round 3, and finalize; the SDK talks to BitGo between each step (same flow as AWM today).
- EdDSA: New `createKeychainsWithExternalSigner` runs init, then finalize for user, then finalize for backup — user finalize must pass a counterparty share to backup finalize (same as AWM’s `v1.mpc.key.*` API).
- Wallet entry point: `generateWalletWithExternalSigner` now supports TSS; it picks ECDSA vs EdDSA from the coin and creates the wallet end-to-end (keychains + `/wallet/add`).
- Types exported from sdk-core so advanced-wallets can implement the callbacks (`EcdsaMPCv2KeyGenCallbacks`, `EddsaKeyGenCallbacks`, etc.).
- On- chain still works on TSS coins: If you pass `createKeychainCallback`, we treat it as on- chain even when the coin defaults to TSS; you can’t mix on- chain and MPC callbacks in one call.
- TSS path rejects passphrase/passcode/webauthn, blocks MPCv1 (wallet version 3), and keeps existing EVM wallet-version checks.
- Note: This is non-breaking / opt-in. Existing passphrase `generateWallet` and on-chain external signer are unchanged since the TSS path only runs when MPC callbacks are passed.

Tests
- `ecdsaMPCv2.ts`: callback ordering, round-response/state threading, common-keychain and session-ID mismatch rejection
- `eddsaExternalSigner.ts`: sequential finalize order, counterparty context, missing share and common-keychain mismatch rejection
- `walletsExternalSigner.ts`: TSS routing and validation (missing callbacks, enterprise, custodial, EVM version, MPCv1 v3), on -chain-on- TSS-coin routing, `type`/`walletVersion` forwarding, mutual-exclusion of callback sets

Ticket: WCN-682

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/iWallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallets.ts
@@ -150,7 +150,7 @@ export interface EddsaKeyGenInitializeResult {
   backupToBitgoKeyShare: ExternalSignerKeyShare;
   userState: ExternalSignerMpcState;
   backupState: ExternalSignerMpcState;
-  backupCounterPartyKeyShare: ExternalSignerKeyShare;
+  backupToUserCounterPartyKeyShare: ExternalSignerKeyShare;
 }
 
 export type EddsaKeyGenInitializeCallback = (params: {
@@ -160,6 +160,7 @@ export type EddsaKeyGenInitializeCallback = (params: {
 
 export type EddsaKeyGenFinalizeResult = {
   commonKeychain: string;
+  /** Required when the callback is invoked with source: 'user'; must be omitted for source: 'backup'. */
   counterpartyKeyShare?: ExternalSignerKeyShare;
 };
 
--- a/modules/sdk-core/src/bitgo/wallet/iWallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/iWallets.ts
@@ -1,8 +1,10 @@
 import * as t from 'io-ts';
+import { DklsTypes } from '@bitgo/sdk-lib-mpc';
+import { MPCv2BroadcastMessage, MPCv2P2PMessage } from '@bitgo/public-types';
 
 import { EncryptionVersion, IRequestTracer } from '../../api';
 import { KeychainsTriplet, LightningKeychainsTriplet } from '../baseCoin';
-import { Keychain, WebauthnInfo } from '../keychain';
+import { ApiKeyShare, Keychain, WebauthnInfo } from '../keychain';
 import { IWallet, PaginationOptions, WalletShare } from './iWallet';
 import { Wallet } from './wallet';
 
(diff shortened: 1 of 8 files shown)


danielpeng1:
@claude review this; verify that the `advanced-wallets` repo is in good shape to use this sdk callback as implemented.

modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -1714,6 +1727,117 @@ export class Wallets implements IWallets {
...
+    if (!enterprise) {
+      throw new Error('enterprise is required for TSS wallet generation with external signer');
+    }
+
+    if (
+      this.baseCoin.isEVM() &&
+      params.walletVersion !== undefined &&
+      !(params.walletVersion === 3 || params.walletVersion === 5 || params.walletVersion === 6)
+    ) {
+      throw new Error('EVM TSS wallets are only supported for wallet version 3, 5 and 6');
```
pranavjain97: the EVM guard allows version 3 but EVM coins are ECDSA, so any EVM wallet with version 3 always hits the rejection below. should the EVM guard be `walletVersion === 5 || walletVersion === 6` only?

modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -76,6 +78,105 @@ export interface CreateKeychainCallbackResult {
...
+}
+
+export interface EddsaKeyGenInitializeResult {
+  userGpgPublicKey: string;
+  backupGpgPublicKey: string;
+  userToBitgoKeyShare: ExternalSignerKeyShare;
+  backupToBitgoKeyShare: ExternalSignerKeyShare;
+  userState: ExternalSignerMpcState;
+  backupState: ExternalSignerMpcState;
+  backupCounterPartyKeyShare: ExternalSignerKeyShare;
```
pranavjain97: nit: reads as 'backup's counterparty share' rather than 'the share backup provides to user.' `backupToUserCounterPartyKeyShare` would clarify direction.

modules/sdk-core/src/bitgo/wallet/iWallets.ts
```diff
@@ -76,6 +78,105 @@ export interface CreateKeychainCallbackResult {
...
+}
+
+export type EddsaKeyGenInitializeCallback = (params: {
+  enterprise: string;
+  bitgoPublicGpgKey: string;
+}) => Promise<EddsaKeyGenInitializeResult>;
+
+export type EddsaKeyGenFinalizeResult = {
+  commonKeychain: string;
+  counterpartyKeyShare?: ExternalSignerKeyShare;
```
pranavjain97, line 164: nit: optional in the type but required when source is 'user'. no type-level indication of when it's required vs. omittable.

modules/sdk-core/test/unit/bitgo/wallet/walletsExternalSigner.ts
```diff
@@ -298,6 +331,479 @@ describe('Wallets - external signer onchain wallet generation', function () {
...
+        type: 'advanced',
+        multisigType: 'tss',
+        enterprise: 'enterprise-id',
+        ecdsaMPCv2Callbacks: callbacks,
+      });
+
+      send.firstCall.args[0].type.should.equal('advanced');
+    });
+
+    it('should default EVM walletVersion to 5 when TSS settings specify MPCv2', async function () {
```
pranavjain97: nit: the three integration-style tests here repeat the full callback stub setup. moving to `beforeEach` would cut ~80 lines.

modules/sdk-core/src/bitgo/wallet/wallets.ts
danielpeng1: i see, i've made the relevant changes for this in terms of the check and test adjustments.

also addressed and made changes for the nits below, thanks.

pranavjain97 (approved):

PR #9132, opened 2026-06-29
fix(sdk-lib-mpc): lazy-load wasm-mps to fix browser DSG

Marzooqa:
Static top-level import compiled to synchronous require() in CJS, causing fs.readFileSync at module init — unavailable in browser. DSG failed with: Cannot read properties of undefined (reading '__wbindgen_add_to_stack_pointer')

Similar to the lazy-load pattern from DKG fix:
- Browser: await import('@bitgo/wasm-mps/web') + webWasm.default()
- Node: await import('@bitgo/wasm-mps')
- initDsg() and restoreSession() made async
- eddsaMPCv2.ts call sites updated to await initDsg()
- Tests updated accordingly

WCI-866

### DIFF
--- a/modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
+++ b/modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
@@ -1,11 +1,10 @@
 import assert from 'assert';
-import {
-  ed25519_dsg_round0_process,
-  ed25519_dsg_round1_process,
-  ed25519_dsg_round2_process,
-  ed25519_dsg_round3_process,
-} from '@bitgo/wasm-mps';
+import type { MsgState } from '@bitgo/wasm-mps';
 import { DeserializedMessage, DeserializedMessages, DsgState } from './types';
+
+type NodeWasmer = typeof import('@bitgo/wasm-mps');
+type WebWasmer = typeof import('@bitgo/wasm-mps/web');
+type WasmMps = NodeWasmer | WebWasmer;
 
 /**
  * EdDSA Distributed Sign Generation (DSG) implementation using @bitgo/wasm-mps.
@@ -22,7 +21,7 @@ import { DeserializedMessage, DeserializedMessages, DsgState } from './types';
  * @example
  * ```typescript
  * const dsg = new DSG(0);  // partyIdx 0
- * dsg.initDsg(keyShare, message, 'm', 2);  // counterpart is party 2
+ * await dsg.initDsg(keyShare, message, 'm', 2);  // counterpart is party 2
  * const msg1 = dsg.getFirstMessage();
  * const msg2 = dsg.handleIncomingMessages([msg1, peerMsg1]);  // emits SignMsg2
  * const msg3 = dsg.handleIncomingMessages([msg2[0], peerMsg2]);  // emits SignMsg3
@@ -45,6 +44,8 @@ export class DSG {
   private dsgStateBytes: Buffer | null = null;
   /** Final 64-byte Ed25519 signature, available after WaitMsg3 -> Complete */
   private signature: Buffer | null = null;
+  /** Lazily loaded WASM module */
+  private wasmMps: WasmMps | null = null;
 
   protected dsgState: DsgState = DsgState.Uninitialized;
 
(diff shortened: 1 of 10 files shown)


modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/verifyMpsDerivation.ts
```diff
@@ -0,0 +1,90 @@
+/**
+ * Verifies that deriveUnhardenedMps (sdk-lib-mpc) produces the exact public key
```
vibhavgo: Not sure if this file should be in the scope of this issue

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
```diff
@@ -64,6 +65,33 @@ export class DSG {
     return this.otherPartyIdx;
   }
 
+  private async loadWasmMps(): Promise<void> {
+    if (!this.wasmMps) {
+      if (
+        typeof window !== 'undefined' &&
+        /* checks for electron processes */
+        !window.process &&
+        !window.process?.['type']
```
vibhavgo, line 74: `!window.process?.['type']` is redundant here
Since the previous condition already implies that `window.process` would be undefined

Marzooqa: It mirrors DKG, will have to clean both of it together as a follow-up

modules/sdk-core/test/unit/bitgo/utils/tss/eddsa/verifyMpsDerivation.ts
Marzooqa: Good catch, this shouldn't be added

vibhavgo (dismissed):

danielpeng1 (commented):
Looks good; only small comments

Also since asyncifying is technically a breaking change should we explicitly note that callers must now await these methods (via merge commit message, which updates `CHANGELOG.md`s for respective modules)?

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
```diff
@@ -74,7 +102,8 @@ export class DSG {
    * @param otherPartyIdx - Party index of the single counterpart in this signing session.
    *   Must differ from this party's own `partyIdx` and be in `[0, 2]`.
    */
-  initDsg(keyShare: Buffer, message: Buffer, derivationPath: string, otherPartyIdx: number): void {
+  async initDsg(keyShare: Buffer, message: Buffer, derivationPath: string, otherPartyIdx: number): Promise<void> {
```
danielpeng1, line 105: there are a couple other call sites that `await` can be added to, not sure if in scope

- http://github.com/BitGo/BitGoJS/blob/master/modules/bitgo/test/v2/unit/internal/tssUtils/eddsaMPCv2/signTxRequest.ts 2 here
- https://github.com/BitGo/BitGoJS/blob/master/modules/express/test/unit/clientRoutes/externalSign.ts 1 here

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
danielpeng1: ` * await dsg.initDsg(keyShare, message, 'm', 2);  // counterpart is party 2`
adding `await` for doc example

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
Marzooqa: Added, thanks for the catch!

danielpeng1 (approved):

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
```diff
@@ -64,6 +65,33 @@ export class DSG {
     return this.otherPartyIdx;
   }
 
+  private async loadWasmMps(): Promise<void> {
+    if (!this.wasmMps) {
+      if (
+        typeof window !== 'undefined' &&
+        /* checks for electron processes */
+        !window.process &&
```
pranavjain97, line 73: wouldn't an Electron renderer where `window.process` is set take the browser wasm path here? `!window.process` is `false` so the `type` check never runs.

pranavjain97 (approved):
lgtm, one small thing

ranga-r2 (approved):

modules/sdk-lib-mpc/src/tss/eddsa-mps/dsg.ts
Marzooqa: Yeah, good catch. I'll create a ticket to fix this and dkg.

Marzooqa: https://linear.app/bitgo/issue/WCI-884/fix-electronbrowser-wasm-path-detection-for-dkg-and-dsg

ArunBala-Bitgo (approved):

prithvishet2503 (approved):

nvjsr (approved):

PR #9133, opened 2026-06-29
fix(abstract-substrate): blake2_256 prehash signablePayload over 256 bytes

abhijit0943:
## Summary

Substrate signs the raw encoded `ExtrinsicPayload` only when it is at most 256 bytes; for payloads larger than 256 bytes it signs the `blake2_256` hash of those bytes instead (Polkadot.js `@polkadot/types/extrinsic/util`, mirrored in the HSM firmware).

The `signablePayload` getter in `@bitgo/abstract-substrate` always returned the raw payload. For large extrinsics — e.g. POLYX `switchValidator`/nominate with 6+ validators — the BitGoJS user signed the raw bytes while the HSM signed the hash, so the two parties signed different messages and TSS G-share combine failed with `InvalidSignature`.

This fix makes `signablePayload` return the bytes that are *actually* signed:
- `length <= 256` → raw bytes (unchanged)
- `length > 256` → `blake2_256(raw)` (32 bytes)

The decision is extracted into a shared `utils.getSubstrateSigningBytes(raw)` helper (with the threshold as the `MAX_RAW_SIGNING_PAYLOAD_BYTES` constant) so the rule lives in one documented place and gives the future DOT fix a copy-paste target. All downstream consumers (`sdk-coin-polyx`, recovery flows) then get the correct `signableHex` at the source, with no coin-specific logic needed in WP.

## Testing

- `sdk-coin-polyx` `nominateBuilder`: small nominate extrinsic returns raw bytes (≤256, not 32); 8-validator nominate returns a 32-byte hashed payload.
- `sdk-coin-polyx` `batchStakingBuilder`: `bond + nominate` with 1 validator stays raw (≤256); with 6 validators returns a 32-byte hash. The batch wraps `nominate` and crosses the threshold sooner, so it is covered explicitly.
- Boundary test through the helper: exactly 256 bytes stays raw, 255 stays raw, 257 hashes to 32 bytes (off-by-one guard).
- Existing POLYX MPC/TSS unit tests (including "should add TSS signature") pass with the updated semantics.

## Notes / follow-ups

- `sdk-coin-dot` has its own standalone `signablePayload` getter with the same raw-only bug, and the DOT wasm path (`wasmTx.signablePayload()`) needs separate verification. Both are out of scope here per the issue and tracked as a DOT follow-up; the new helper is the intended drop-in for that fix.
- This SDK change only fixes prod once WP bumps `@bitgo-beta/abstract-substrate`; the local blake2 workaround in WP `polyx.ts` should be reverted at that point to avoid duplicate logic.


abhijit0943:
@claude review

PR #9141, opened 2026-06-30
feat(abstract-substrate): expose rawExtrinsicPayload separate from signablePayload

abhijit0943:
## Summary

POLYX TSS follows Substrate signing rules: extrinsics over 256 bytes are signed as a `blake2_256` hash rather than the raw payload. `signablePayload` already applies this (correct for user MPC/combine), but nothing exposed the raw `ExtrinsicPayload` bytes that the HSM `polyx/signtx` path needs. Large nominate/switch-validator txs (e.g. 9 validators) therefore broke signing on prod.

This exposes the raw payload separately while keeping `signablePayload` as the correct MPC/combine signing bytes.

## Changes

- **`abstract-substrate`**: add a `rawExtrinsicPayload` getter (full `ExtrinsicPayload.toU8a({ method: true })`), and refactor `signablePayload` to derive from it via `utils.getSubstrateSigningBytes` — no duplicated 256-byte logic in `transaction.ts`.
- **`sdk-coin-polyx`**: surface the raw payload as `rawSignableHex` in the polyx `toJson` output, so consumers needing the raw bytes can read it alongside `signableHex`. The field is declared on the polyx-specific `TxData` (not the shared `abstract-substrate` one), so other Substrate coins (tao, dot, etc.) are unaffected.

## Testing

`yarn unit-test` for the nominate builder. New cases assert:
- ≤256B nominate: `signablePayload` equals the raw `rawExtrinsicPayload` bytes, and `toJson().rawSignableHex` matches the raw hex.
- &gt;256B nominate (9 validators): `signablePayload` is the 32-byte blake2_256 hash, `rawExtrinsicPayload` is the full raw payload (&gt;256B), and `toJson().rawSignableHex` matches the raw hex. (Each target adds 33 bytes over 79 bytes of fixed overhead, so the extrinsic crosses 256B from 6 targets up; 9 mirrors the SI-926 scenario.)

### DIFF
--- a/modules/abstract-substrate/src/lib/transaction.ts
+++ b/modules/abstract-substrate/src/lib/transaction.ts
@@ -533,6 +533,21 @@ export class Transaction extends BaseTransaction {
   }
 
   /**
+   * Returns the full raw encoded `ExtrinsicPayload` bytes (including the method) for this
+   * transaction, without applying the Substrate 256-byte blake2_256 signing rule.
+   *
+   * These are the bytes the HSM signs on the `polyx/signtx` path. They differ from
+   * {@link signablePayload} once the payload exceeds 256 bytes, where {@link signablePayload}
+   * is the blake2_256 hash instead of the raw bytes.
+   */
+  get rawExtrinsicPayload(): Uint8Array {
+    const extrinsicPayload = this._registry.createType('ExtrinsicPayload', this._substrateTransaction, {
+      version: EXTRINSIC_VERSION,
+    });
+    return extrinsicPayload.toU8a({ method: true });
+  }
+
+  /**
    * @inheritdoc
    *
    * Returns the bytes that are actually signed for this transaction. Substrate signs the raw
@@ -543,10 +558,7 @@ export class Transaction extends BaseTransaction {
    * messages, causing TSS signature combination to fail.
    */
   get signablePayload(): Buffer {
-    const extrinsicPayload = this._registry.createType('ExtrinsicPayload', this._substrateTransaction, {
-      version: EXTRINSIC_VERSION,
-    });
-    return utils.getSubstrateSigningBytes(extrinsicPayload.toU8a({ method: true }));
+    return utils.getSubstrateSigningBytes(this.rawExtrinsicPayload);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


abhijit0943:
@claude review this

abhijit0943:
@claude re-review

Doddanna17:
Three findings from reviewing this:

**1. `rawSignableHex` on the wrong interface**

Added to `abstract-substrate`'s shared `TxData`, but `modules/sdk-coin-polyx/src/lib/iface.ts` already has a coin-specific `TxData extends Interface.TxData` with POLYX-only fields (`assetId`, `fromDID`, etc). This field belongs there. As-is, TAO and DOT consumers see `rawSignableHex` in their `TxData` type with no guarantee it's ever populated — their `toJson()` never sets it.

**2. Missing `toJson().rawSignableHex` assertion in the ≤256B test**

The PR tests `toJson().rawSignableHex` only in the >256B case. The ≤256B test should also assert `tx.toJson().rawSignableHex === Buffer.from(tx.rawExtrinsicPayload).toString('hex')` to cover the code path where `rawSignableHex` is set at the top of `toJson()` (before the branch logic).

**3. Why 9 validators, not 8?**

Old comment said "6+" with 8 validators. PR bumps to 9 with comment "9 validators pushes over 256 bytes". What changed — is 8 no longer enough? Without an explanation (e.g. different validator address length in test data, or a payload field that grew), future maintainers can't reason about the boundary and might regress it.

nvrakesh06 (dismissed):

abhijit0943:
@claude re-review the changes are made

Doddanna17 (approved):
> `rawExtrinsicPayload` correctly exposes pre-hash bytes; `signablePayload` now delegates to it — no duplicated `createType` logic. Scope correctly limited to polyx `TxData`, other substrate coins unaffected.
>
> The `else` branch in polyx `toJson()` properly propagates `rawSignableHex` from the pre-computed result without re-invoking the getter.
>
> Tests cover both the sub-256B identity case and the >256B hash-vs-raw divergence, with 9-validator count mirroring SI-926.
>
> LGTM.

PR #9142, opened 2026-06-30
feat(statics): update receipt tokens coin features

hafeezshaik933:
- remove rebase token for few tokens which are not needed
- add a new coin feature to show approx price tooltip in ui

Ticket: CGD-1060

Why new coin feature?
The `receipt-token` feature is being used to block the withdrawals from UI and API. The same flag is being used to show approx. price tooltip in the UI. So, introducing a new flag to show price tooltip explicitly for all tokens which require the tooltip.


nvjsr (approved):
lgtm

mukeshsp (approved):

PR #9163, opened 2026-07-01
feat(sdk-core): check wallet type before accessing tssUtils

MohammedRyaan786:
TICKET: CHALO-726

This pull request introduces a small but important change to the `Wallet` class in `wallet.ts`. The code now ensures that TSS-specific logic for handling `txRequestId` is only executed when the wallet's `multisigType` is `'tss'`, preventing potential errors for non-TSS wallets.

* Restricted the check for `txRequestId` to only execute when `this._wallet.multisigType` is `'tss'`, ensuring TSS logic is not run for other wallet types.


bhuvanr159 (approved):
LGTM

PR #9165, opened 2026-07-02
fix: remove special char from symbol

prajwalu142:
Remove the special char(-) from hybond token
to avoid any issues in downstream services

Ticket: CSHLD-1127


abhishekagrawal080 (approved):
looks good

PR #9167, opened 2026-07-02
fix(xrp): route MPT token enablement to enableMpt type via coin config [CGD-2013]

0xPrabh:
## Summary

- Add `getEnableTokenType` to `TokenEnablementConfig` so coins can return the correct tx type per token
- XRP returns `'enableMpt'` for `XrpMptCoin` tokens and `'enabletoken'` for IOU — fixes the `txrp:feesec is not an XRP token` error caused by hardcoded `'enabletoken'` routing all enablements to the IOU/TrustSet path
- Add `xrpl` codec fallbacks in `ripple.ts` and `xrp.ts` for `MPTokenAuthorize` (unsupported by `ripple-binary-codec`)
- Simplify `handleV2EnableTokens` in express — remove offline MPT/IOU split, `sendTokenEnablements` handles both

## Root Cause

`wallet.buildTokenEnablements()` hardcoded `type = 'enabletoken'`, routing to wallet-platform's TrustSet path which calls `getXrpCurrencyFromTokenName` → `instanceof XrpCoin` → throws for `XrpMptCoin`.

### DIFF
--- a/modules/sdk-coin-xrp/src/ripple.ts
+++ b/modules/sdk-coin-xrp/src/ripple.ts
@@ -50,7 +50,7 @@ const signWithPrivateKey = function (txHex, privateKey, options) {
   } catch (e) {
     try {
       tx = JSON.parse(txHex);
-    } catch (e) {
+    } catch (e2) {
       throw new Error('txHex needs to be either hex or JSON string for XRP');
     }
   }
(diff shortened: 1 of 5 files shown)


0xPrabh:
@claude

modules/sdk-coin-xrp/src/ripple.ts
```diff
@@ -50,7 +50,7 @@ const signWithPrivateKey = function (txHex, privateKey, options) {
   } catch (e) {
     try {
       tx = JSON.parse(txHex);
-    } catch (e) {
+    } catch (e2) {
```
nvjsr, line 53: NIT: can we have better name like parsingError or simply error

nvjsr (approved):

PR #9169, opened 2026-07-02
fix(sdk-coin-eth): use txInfo to fetch recipients

MohammedRyaan786:
TICKET: CHALO-726


at31416 (approved):
pls create a follow-up ticket for adding tests

PR #9179, opened 2026-07-03
fix(sdk-core): migrate MPCv2 EdDSA address derivation to BIP32-Ed25519

Marzooqa:
Replace Silence Labs single-HMAC formula (deriveUnhardenedMps) with the standard BIP32-Ed25519 dual-HMAC formula (Eddsa.deriveUnhardened) for MPCv2 EdDSA address derivation in addressVerification.ts.

Both MPCv1 and MPCv2 ed25519 wallets now use the same derivation path through EDDSAMethods.getInitializedMpcInstance().

Ticket: WCI-644


ranga-r2 (approved):
LGTM

PR #9191, opened 2026-07-07
fix(sdk-coin-sui): keep token coin type on addr-balance rebuild

abhishekagrawal080:
TICKET: CSHLD-1146

Token transfers funded from the Sui address balance embed the coin type in the redeem_funds MoveCall type argument and the BalanceWithdrawal input. When such a transaction is deserialized and rebuilt to attach a signature before broadcast, buildSuiTransaction re-derived the coin type from the coin config. If the builder was constructed with the parent chain config (e.g. tsui) rather than the token config (tsui:deep), packageId/module/symbol are undefined and the coin type became "undefined::undefined::undefined". The rebuilt bytes then differed from the signed payload, so the Sui node rejected the transaction with "Invalid user signature: Fail to verify user sig". Recover the coin type from the deserialized BalanceWithdrawal TypeTag in initBuilder and prefer it over the config-derived value, so a rebuild produces
byte-identical output regardless of which config the builder was created with.


Ranjna-G (approved):
lgtm

akarath (approved):

PR #9204, opened 2026-07-08
feat: add script to reencrypt wallet with v2 encryption

bdesoky:
Ticket: WCN-174

This pull request introduces significant enhancements to the wallet encryption upgrade process and improves compatibility for PDF keycard generation in Node.js environments. The main changes include adding a new script to automate wallet keychain re-encryption and updating the keycard drawing utilities to support server-side (Node.js) usage.

**Key changes:**

### 1. New Wallet Encryption Upgrade Script

* Added `scripts/upgrade-wallet-encryption.ts`, a comprehensive CLI tool to upgrade wallet keychains from v1 (SJCL/PBKDF2-SHA256 + AES-256-CCM) to v2 (Argon2id + AES-256-GCM) encryption, with support for regenerating the keycard PDF and handling various edge cases (e.g., changed passphrases, legacy backups).

### 2. Node.js Compatibility for Keycard PDF Generation

* Updated `drawKeycard` and related functions in `modules/key-card/src/drawKeycard.ts` to detect Node.js environments and handle image and QR code rendering appropriately (e.g., using base64 data URLs instead of DOM elements). [[1]](diffhunk://#diff-4a27b411935ea3a641a387669372731bd15337eb90c15843b6ea44c13d84ba50R7-R8) [[2]](diffhunk://#diff-4a27b411935ea3a641a387669372731bd15337eb90c15843b6ea44c13d84ba50R136-R143) [[3]](diffhunk://#diff-4a27b411935ea3a641a387669372731bd15337eb90c15843b6ea44c13d84ba50L198-R218)
* Modified QR code image handling to allow both `HTMLCanvasElement` and base64 string inputs, ensuring PDF generation works in both browser and Node.js contexts. [[1]](diffhunk://#diff-4a27b411935ea3a641a387669372731bd15337eb90c15843b6ea44c13d84ba50L56-R58) [[2]](diffhunk://#diff-4a27b411935ea3a641a387669372731bd15337eb90c15843b6ea44c13d84ba50R73-R77)

These changes ensure that wallet upgrades and keycard regeneration can be performed reliably in automated or server-side workflows, improving maintainability and operational flexibility.

### DIFF
--- a/modules/key-card/src/drawKeycard.ts
+++ b/modules/key-card/src/drawKeycard.ts
@@ -4,6 +4,8 @@ import { IDrawKeyCard } from './types';
 import { splitKeys } from './utils';
 type jsPDFModule = typeof import('jspdf');
 
+const isNode = typeof window === 'undefined' || typeof window.document === 'undefined';
+
 async function loadJSPDF(): Promise<jsPDFModule> {
   let jsPDF: jsPDFModule;
 
@@ -60,7 +62,7 @@ function moveDown(y: number, ydelta: number): number {
 // continuation on a later page) and the y-offset just below the drawn QR column (so callers
 // can place content, e.g. a note, under the QR codes).
 function drawOnePageOfQrCodes(
-  qrImages: HTMLCanvasElement[],
+  qrImages: (HTMLCanvasElement | string)[],
   doc: jsPDF,
   y: number,
   qrSize: number,
@@ -75,7 +77,11 @@ function drawOnePageOfQrCodes(
       return { nextIndex: qrIndex, endY: y };
     }
 
-    doc.addImage(image, left(0), y, qrSize, qrSize);
+    if (typeof image === 'string') {
+      doc.addImage(image, 'PNG', left(0), y, qrSize, qrSize);
+    } else {
+      doc.addImage(image, left(0), y, qrSize, qrSize);
+    }
 
     if (qrImages.length === 1) {
       return { nextIndex: qrIndex + 1, endY: y + qrSize };
@@ -135,7 +141,13 @@ export async function drawKeycard({
 
   if (keyCardImage) {
     const [imgWidth, imgHeight] = computeKeyCardImageDimensions(keyCardImage);
-    doc.addImage(keyCardImage, left(0), y, imgWidth, imgHeight);
+    if (isNode) {
+      // In Node.js, jsPDF cannot extract pixels from an HTMLImageElement (no DOM/canvas).
+      // The script passes a duck-typed object whose .src is a base64 data URL — use it directly.
+      doc.addImage((keyCardImage as unknown as { src: string }).src, 'PNG', left(0), y, imgWidth, imgHeight);
+    } else {
+      doc.addImage(keyCardImage, left(0), y, imgWidth, imgHeight);
+    }
   }
 
   // Activation Code
@@ -204,10 +216,14 @@ export async function drawKeycard({
     const textLeft = left(qrSize + 15);
     let textHeight = 0;
 
-    const qrImages: HTMLCanvasElement[] = [];
+    const qrImages: (HTMLCanvasElement | string)[] = [];
     const keys = splitKeys(qr.data, QRBinaryMaxLength);
     for (const key of keys) {
-      qrImages.push(await QRCode.toCanvas(key, { errorCorrectionLevel: 'L' }));
+      if (isNode) {
+        qrImages.push(await QRCode.toDataURL(key, { errorCorrectionLevel: 'L' }));
+      } else {
+        qrImages.push(await QRCode.toCanvas(key, { errorCorrectionLevel: 'L' }));
+      }
     }
 
     const isMultiPart = qr?.data?.length > QRBinaryMaxLength;
--- a/modules/key-card/src/upgradeWalletEncryption.ts
+++ b/modules/key-card/src/upgradeWalletEncryption.ts
@@ -1,0 +1,89 @@
+/**
+ * PDF-generation helpers for the wallet encryption upgrade flow.
+ *
+ * Orchestration (unlocking the session, re-encrypting keychains, PUT-ing to the server) lives on
+ * `Wallet.upgradeEncryption` in `@bitgo/sdk-core`. This module contains only the pieces that
+ * depend on `jspdf`/`qrcode` — a factory that returns a `UpgradeEncryptionPdfGenerator` the
+ * caller can pass into that method, plus a Node-side coin-logo loader.
+ */
+
+import * as https from 'https';
+import { BaseCoin, coins } from '@bitgo/statics';
+import { UpgradeEncryptionPdfGenerator } from '@bitgo/sdk-core';
+import { generateQrData } from './generateQrData';
+import { generateFaq } from './faq';
+import { drawKeycard } from './drawKeycard';
+
+/**
+ * Fetch a coin logo image and return an HTMLImageElement-compatible object.
+ * jsPDF reads `.src` in Node.js; computeKeyCardImageDimensions reads `.width`/`.height`.
+ * Returns undefined on any failure — the keycard is generated without the logo in that case.
+ */
+export async function loadKeycardImage(
+  url: string,
+  httpGet: typeof https.get = https.get
+): Promise<HTMLImageElement | undefined> {
+  return new Promise((resolve) => {
+    httpGet(url, (res) => {
+      if (res.statusCode !== 200) {
+        console.warn(`Warning: coin logo not loaded (HTTP ${res.statusCode}) — keycard will be generated without it`);
+        resolve(undefined);
+        return;
+      }
+      const chunks: Buffer[] = [];
+      res.on('data', (chunk) => chunks.push(chunk));
+      res.on('end', () => {
+        const contentType = res.headers['content-type'] ?? 'image/png';
+        const dataUrl = `data:${contentType};base64,${Buffer.concat(chunks).toString('base64')}`;
+        const img = { src: dataUrl, width: 303, height: 40 } as unknown as HTMLImageElement;
+        resolve(img);
+      });
+      res.on('error', (err) => {
+        console.warn(`Warning: coin logo not loaded (${err.message}) — keycard will be generated without it`);
+        resolve(undefined);
+      });
+    }).on('error', (err) => {
+      console.warn(`Warning: coin logo not loaded (${err.message}) — keycard will be generated without it`);
+      resolve(undefined);
+    });
+  });
+}
+
+export interface KeycardPdfGeneratorOptions {
+  /** Base URL for loading the coin logo (e.g. Environments[env].uri). Omit to skip the logo. */
+  imageBaseUrl?: string;
+}
+
+/**
+ * Build a {@link UpgradeEncryptionPdfGenerator} that regenerates the keycard PDF with the
+ * canonical `generateQrData` + `drawKeycard` flow. Pass the returned function into
+ * `Wallet.upgradeEncryption({ generatePdf })`.
+ */
+export function createKeycardPdfGenerator(options: KeycardPdfGeneratorOptions = {}): UpgradeEncryptionPdfGenerator {
+  const { imageBaseUrl } = options;
+  return async ({
+    coinName,
+    userKeychain,
+    backupKeychain,
+    bitgoKeychain,
+    passphrase,
+    passcodeEncryptionCode,
+    walletLabel,
+  }) => {
+    const staticsCoin = coins.get(coinName) as BaseCoin;
+    const keyCardImage = imageBaseUrl
+      ? await loadKeycardImage(`${imageBaseUrl}/web/assets/keycards/${staticsCoin.family.toLowerCase()}.png`)
+      : undefined;
+    const qrData = await generateQrData({
+      coin: staticsCoin,
+      userKeychain,
+      backupKeychain,
+      bitgoKeychain,
+      passphrase,
+      passcodeEncryptionCode,
+      encryptionVersion: 2,
+    });
+    const questions = generateFaq(staticsCoin.fullName);
+    return drawKeycard({ qrData, questions, walletLabel, keyCardImage });
+  };
+}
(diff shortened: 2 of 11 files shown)


modules/key-card/src/drawKeycard.ts
OttoAllmendinger, line 1: this belongs in a different PR

vmccarty (dismissed):
The instructions read well to me 👍 I just have one question about the "dry-run", but it doesn't seem too high priority.

scripts/upgrade-wallet-encryption.ts
```diff
@@ -0,0 +1,355 @@
...
+  if (!dryRun) {
+    await bitgo.unlock({ otp: otp ?? '0000000', duration: 600 });
+    console.log('Session unlocked.');
+  }
+
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  console.log(`Wallet: ${wallet.label()} (${walletId})`);
+
+  // Fetch passcodeEncryptionCode — required to decrypt Box D and to generate Box D in the new keycard.
+  // Skipped in dry-run if not provided, since no keycard is produced anyway.
```
vmccarty: What's a dry run? Just like a test? Will this be something the client developers already have some frame of reference for, or does this need some kind of explanation somewhere?

bdesoky: I actually updated this so that it isnt skipped in dry-run when boxD is provided but forgot to push the change 😅 
A dry run will attempt the decrypt + reencrypt but wont update anything in the backend. Just a way to test that the passphrase/boxD work properly

scripts/upgrade-wallet-encryption.ts
```diff
@@ -0,0 +1,382 @@
...
+ * v2 = Argon2id + AES-256-GCM, v=2
+ */
+function getEncryptionVersion(ciphertext: string): 1 | 2 | undefined {
+  try {
+    const envelope = JSON.parse(ciphertext);
+    if (envelope.v === 2) return 2;
+    // SJCL envelopes have no "v" field or v=1
+    if (!envelope.v || envelope.v === 1) return 1;
+  } catch {
+    // not JSON
```
OttoAllmendinger: don't we want to about this condition?

scripts/upgrade-wallet-encryption.ts
```diff
@@ -0,0 +1,382 @@
...
+async function getPasscodeEncryptionCode(bitgo: BitGo, coin: string, walletId: string): Promise<string> {
+  const { recoveryInfo } = await bitgo
+    .post(bitgo.microservicesUrl(`/api/v2/${coin}/wallet/${walletId}/passcoderecovery`))
+    .result();
+  if (!recoveryInfo || !('passcodeEncryptionCode' in recoveryInfo)) {
+    throw new Error(
+      'passcoderecovery endpoint did not return a passcodeEncryptionCode — pass --passcodeEncryptionCode manually'
+    );
+  }
+  return recoveryInfo.passcodeEncryptionCode as string;
```
OttoAllmendinger: don't cast, validate

scripts/upgrade-wallet-encryption.ts
```diff
@@ -0,0 +1,382 @@
...
+}
+
+/**
+ * Fetch the coin logo image and return a duck-typed HTMLImageElement-compatible object.
+ * jsPDF reads `.src` in Node.js; computeKeyCardImageDimensions reads `.width`/`.height`.
+ * This replicates what the BitGo UI does before calling drawKeycard.
+ */
+async function loadKeycardImage(url: string): Promise<HTMLImageElement | undefined> {
```
OttoAllmendinger: why are you silently suppressing all errors?

scripts/upgrade-wallet-encryption.ts
```diff
@@ -0,0 +1,382 @@
...
+  const coin = get('--coin');
+  const walletId = get('--walletId');
+  const passphrase = get('--passphrase');
+  const accessToken = get('--accessToken');
+  const otp = get('--otp');
+  const boxD = get('--boxD');
+  const boxA = get('--boxA');
+  const boxB = get('--boxB');
+  const passcodeEncryptionCode = get('--passcodeEncryptionCode');
+  const dryRun = args.includes('--dry-run');
```
OttoAllmendinger: use a real option parser, I believe we typically use `yargs`

scripts/upgrade-wallet-encryption.ts
```diff
@@ -0,0 +1,382 @@
...
+    passcodeEncryptionCode,
+    dryRun,
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Main
+// ---------------------------------------------------------------------------
+
+async function main() {
```
OttoAllmendinger, line 69: this is long and complicated enough that it deserves to be in `modules/` properly, with a test

maybe even in `modules/key-card/`

OttoAllmendinger (dismissed):

scripts/upgrade-wallet-encryption.ts
bdesoky: updated to throw on unrecognized version

scripts/upgrade-wallet-encryption.ts
bdesoky: updated now to log a warning. the image isnt critical for the keycard so it wont throw

scripts/upgrade-wallet-encryption.ts
bdesoky: updated to use yargs

scripts/upgrade-wallet-encryption.ts
bdesoky: hmm, i think this script is supposed to only be a temporary measure for users to opt in to v2 wallets until the UI defaults to v2 and provides a way to reencrypt wallets. Not sure if we need to move this to the module, @pranavjain97 what do you think?

pranavjain97: we're going to need this script indefinitely for non-UI users with older wallets so we could potentially move it to the module. however, this can be done as a followup and shouldn't be blocking

scripts/upgrade-wallet-encryption.ts
```diff
@@ -0,0 +1,382 @@
...
+  const bitgo = new BitGo({ env });
+  bitgo.authenticateWithAccessToken({ accessToken });
+
+  if (!dryRun) {
+    await bitgo.unlock({ otp: otp ?? '0000000', duration: 600 });
+    console.log('Session unlocked.');
+  }
+
+  const wallet = await bitgo.coin(coin).wallets().get({ id: walletId });
+  console.log(`Wallet: ${wallet.label()} (${walletId})`);
```
pranavjain97: we should fail here if the wallet is MPCv2 and boxA/boxB values aren't provided, else it will give back the massive PDF

pranavjain97 (changes requested):

scripts/upgrade-wallet-encryption.ts
pranavjain97: fixed

pranavjain97 (dismissed):
mpcv2 guard added, was able to re-encrypt with reduced mpcv2 keycard

pranavjain97:
@claude give approval rating

OttoAllmendinger (commented):
not going to ACK without module and tests, flushing

OttoAllmendinger (commented):
see previous comment

pranavjain97:
> not going to ACK without module and tests, flushing

I've moved the business logic to the key card module with tests. still have the script to invoke/e2e sample for clients. @OttoAllmendinger can you review?

modules/key-card/src/upgradeWalletEncryption.ts
```diff
@@ -0,0 +1,420 @@
+/**
```
zahin-mohammad, line 1: Agree with Otto to a certain degree. This deserves a proper module + tests rather than living only as a one-off script — but I don't think `modules/key-card/` is the right home for it.

`@bitgo/key-card` already owns more than PDF drawing (`generateQrData` selects/assembles box payloads from keychains, including Box D encryption). That precedent covers "given keychains, produce QR data + PDF."

`runUpgradeWalletEncryption` goes well beyond that: session unlock, passcoderecovery, decrypt/reencrypt, PUT keychain updates, MPCv2 boxA/boxB recovery, dry-run orchestration. That's wallet/keychain lifecycle logic, not keycard presentation.

Here is a proposed split (take it with a grain of salt):
1. put the reencrypt primitives (`getEncryptionVersion`, v1→v2 reencrypt) on `Keychains` in sdk-core — it already has `updateSingleKeychainPassword` / `getEncryptionVersion`
2. keep wallet-scoped orchestration on Wallet (or a thin dedicated helper), with unit tests there
3. keep `@bitgo/key-card` as a dependency only for regenerating the PDF from the updated keychains

zahin-mohammad: Requesting changes here because removing this from this module after-the-fact will either lead to a breaking change or unnecessary tech-debt.

zahin-mohammad (changes requested):
In addition to moving this out, have we done an e2e test with the keycard with
- bitgo ui
- wallet recovery wizard

modules/key-card/src/upgradeWalletEncryption.ts
pranavjain97: Agreed, moved the primitives to wallet/keychain sdk-core

modules/sdk-core/src/bitgo/keychain/keychains.ts
```diff
@@ -209,6 +211,30 @@ export class Keychains implements IKeychains {
...
+    try {
+      prv = await this.bitgo.decrypt({ input: encryptedPrv, password: passphrase });
+    } catch {
+      if (originalPassphrase) {
+        prv = await this.bitgo.decrypt({ input: encryptedPrv, password: originalPassphrase });
+      } else {
+        throw new Error(
+          'Failed to decrypt with the provided passphrase. ' +
+            'If the wallet password was changed after creation, provide the original passphrase so the backup key can be decrypted.'
+        );
```
zahin-mohammad: Nothing about this method is specific to the backup key? I am confused by this error message...

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3382,6 +3384,221 @@ export class Wallet implements IWallet {
...
+
+    if (dryRun) console.log('[dry-run] No changes will be persisted.');
+
+    if (!dryRun) {
+      try {
+        await this.bitgo
+          .post(this.bitgo.microservicesUrl('/api/v1/user/unlock'))
+          .send({ otp: otp ?? '0000000', duration: 600 })
+          .result();
+        console.log('Session unlocked.');
```
zahin-mohammad, line 3432: Nit: don't add console logs into the sdk. If needed, accept a `logger` as a dependency, and let the caller pass in a logger (i.e. your script).

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3382,6 +3384,221 @@ export class Wallet implements IWallet {
...
+        console.log('Recovered original passphrase from Box D.');
+      }
+    }
+
+    const keyIds = this.keyIds();
+    const [userKeychain, backupKeychain, bitgoKeychain] = await Promise.all([
+      keychainsApi.get({ id: keyIds[0] }),
+      keychainsApi.get({ id: keyIds[1] }),
+      keychainsApi.get({ id: keyIds[2] }),
+    ]);
```
zahin-mohammad, line 3493: What about wallets that do not have 3 keys? (i.e. OFC)

zahin-mohammad (dismissed):
Logs can be moved in a follow up, along with OFC support. (OFC support can be done via key rotation instead of re-encryption)

scripts/upgrade-wallet-encryption.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import * as fs from 'fs';
+import * as path from 'path';
+import * as yargsLib from 'yargs';
+import { Environments, EnvironmentName } from '../modules/sdk-core/src/bitgo/environments';
+import { createKeycardPdfGenerator } from '../modules/key-card/src/upgradeWalletEncryption';
+import { BitGo } from '../modules/bitgo/dist/src/bitgo';
+
+function parseArgs() {
+  const argv = yargsLib
```
OttoAllmendinger, line 40: 👍

OttoAllmendinger (commented):
looks much better, agree with zahin

OttoAllmendinger (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
pranavjain97: will do in a followup

modules/sdk-core/src/bitgo/keychain/keychains.ts
pranavjain97: When passwords change, only user key is updated, and backup requires the original password. But I think there is scope to better handle this case + error messaging. will update

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3587,6 +3602,33 @@ export class Wallet implements IWallet {
     return { doc, walletLabel };
   }
 
+  /**
+   * Re-encrypt a ciphertext that was written at wallet creation time (backup key, boxA, boxB).
+   * These are not re-encrypted on password change, so the encryption passphrase may still be
+   * the original one from when the wallet was created. Try the current passphrase first; fall
+   * back to `originalPassphrase` if provided.
+   */
+  private async reencryptCreationTimeKey(
```
zahin-mohammad, line 3613: `reEncryptCreationTimeKey`

`reencryptCreationTimeKey` looks a little awkward.

zahin-mohammad (approved):
small nit

modules/sdk-core/src/bitgo/wallet/wallet.ts
pranavjain97: followup, out of scope for this pr

PR #9210, opened 2026-07-08
feat: include enterpriseId as AAD in passkeys encryption process

danielpeng1:
Adding `enterpriseId` as AES-GCM AAD to PRF-encrypted private keys
- Thus with ADD a ciphertext encrypted under a given enterprise has a cryptographic binding to that enterprise.
- Decryption is fully backward-compatible: decryptV2 reads adata from the envelope, old blobs (no adata) continue to decrypt normally, and no API signatures changed.

Note: what if no enterprise?
- `attachPasskeyToWallet`: safe, already throws before reaching encrypt if enterprise is missing

and
- `wallets.ts` wallet creation: `params.enterprise` is `enterprise?: string (optional)`. undefined is handled
- `wallets.ts` bulk share accept: `walletShare.enterprise` is also optional. undefined is handled

so enterprise membership should already be enforced by the server first.

Ticket: WCN-1074

### DIFF
--- a/modules/passkey-crypto/src/attachPasskeyToWallet.ts
+++ b/modules/passkey-crypto/src/attachPasskeyToWallet.ts
@@ -71,6 +71,8 @@ export async function attachPasskeyToWallet(params: {
     password: prfPassword,
     input: privateKey,
     encryptionVersion,
+    // Bind to this enterprise via AES-GCM AAD — moving this blob to another enterprise
+    // or tampering with the stored adata field invalidates the GCM tag.
     adata: enterpriseId,
   });
 
--- a/modules/passkey-crypto/src/attachPasskeyToWallet.ts
+++ b/modules/passkey-crypto/src/attachPasskeyToWallet.ts
@@ -67,7 +67,12 @@ export async function attachPasskeyToWallet(params: {
   }
 
   const prfPassword = derivePassword(authResult.prfResult);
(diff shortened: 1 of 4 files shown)


OttoAllmendinger (dismissed):

modules/passkey-crypto/src/attachPasskeyToWallet.ts
```diff
@@ -67,7 +67,12 @@ export async function attachPasskeyToWallet(params: {
   }
 
   const prfPassword = derivePassword(authResult.prfResult);
-  const encryptedPrv = await bitgo.encrypt({ password: prfPassword, input: privateKey, encryptionVersion });
+  const encryptedPrv = await bitgo.encrypt({
+    password: prfPassword,
+    input: privateKey,
+    encryptionVersion,
+    adata: enterpriseId,
```
OttoAllmendinger, line 76: add a comment explaining what this does

OttoAllmendinger (changes requested):
please add a test that shows that `adata: undefined` and omitted `adata` are equivalent

vibhavgo (commented):
Shouldn't we also add a test to ensure that a key encrypted in one enterprise context shouldn't be able to be decrypted in another enterprise context?

OttoAllmendinger (approved):

PR #9211, opened 2026-07-08
feat(sdk-core): depositToVault with getVaultConfig and Concrete BTC support

venkateshv1266:
## Summary
- Add `VaultProvider` enum (`Morpho`, `ConcreteBtccx`) and named result types `ConcreteDepositResult` / `MorphoDepositResult`; update `DepositResult` union to use them
- Add `getVaultConfig` to `DefiVault` and `IDefiVault`, proxying `GET /api/defi-service/v1/vaults/:id`
- Add protocol dispatch in `depositToVault`: routes `ConcreteBtccx` → `depositToConcreteVault`, `Morpho` → `depositToMorphoVault`, else throws unsupported provider error
- `depositToConcreteVault`: single `sendMany` with `type: 'defi-deposit'` and `defiParams`; no recipients (WP resolves escrow destination server-side)
- Add `defiParams: t.unknown` to `BuildParams` whitelist and optional `defiParams` to `SendManyOptions`; bump `@bitgo/public-types` to `^6.39.0`
- Unit tests: `getVaultConfig` happy path + error cases, Concrete dispatch (single sendMany, pendingApproval extraction, passphrase forwarding), Morpho regression guard

## Test plan
- [ ] `yarn unit-test -- --grep "DefiVault"` in `modules/sdk-core`: all passing
- [ ] `yarn build` in `modules/sdk-core`: no TypeScript errors
- [ ] Morpho existing tests are green (regression guard)
- [ ] Concrete path: `depositToVault` with a `concrete_btccx` vault returns `{ pendingApprovalId, state }`
- [ ] Unsupported provider throws descriptive error

CLOSES TICKET: DEFI-279

### DIFF
--- a/modules/sdk-core/src/bitgo/defi/defiVault.ts
+++ b/modules/sdk-core/src/bitgo/defi/defiVault.ts
@@ -1,18 +1,24 @@
 /**
  * @prettier
  */
+import * as t from 'io-ts';
+import { GetVaultResponse, VaultProtocol } from '@bitgo/public-types';
 import {
+  ConcreteDepositResult,
+  MorphoDepositResult,
   DefiOperation,
   DefiOperationListResult,
   DepositResult,
   DepositToVaultOptions,
   GetOperationOptions,
+  GetVaultConfigOptions,
   IDefiVault,
   ListOperationsOptions,
   ResumeDepositOptions,
 } from './iDefiVault';
 import { IWallet } from '../wallet';
 import { BitGoBase } from '../bitgoBase';
+import { decodeWithCodec } from '../utils';
 
 /**
  * Error thrown when a concurrent active deposit already exists for the (wallet, vault) pair.
@@ -48,12 +54,25 @@ export class DefiVault implements IDefiVault {
   }
 
   /**
+   * Fetch vault config from defi-service. Used internally to determine
+   * which deposit path to take (Concrete vs Morpho).
+   */
+  async getVaultConfig(params: GetVaultConfigOptions): Promise<GetVaultResponse> {
+    if (!params.vaultId) {
+      throw new Error('vaultId is required');
+    }
+    const raw = await this.bitgo
+      .get(this.bitgo.microservicesUrl(`/api/defi-service/v1/vaults/${params.vaultId}`))
+      .result();
+    return decodeWithCodec(GetVaultResponse, raw, 'getVaultConfig');
+  }
+
+  /**
    * Deposit an amount of underlying asset into a vault.
    *
-   * Internally issues two sendMany calls (approve + deposit) and returns the
-   * operationId that links them. If the deposit sendMany fails after
-   * the approve succeeds, the error propagates — the server-side reconciler
-   * handles orphaned approvals.
+   * Dispatches to the concrete or morpho path based on vault provider.
+   * The concrete path returns a pendingApproval (custodial wallet).
+   * The morpho path issues two sendMany calls (approve + deposit).
    *
    * @param params.vaultId - DeFi-service vault identifier
    * @param params.amount - amount in base units of the underlying asset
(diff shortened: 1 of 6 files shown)


modules/sdk-core/src/bitgo/defi/defiVault.ts
```diff
@@ -68,6 +83,41 @@ export class DefiVault implements IDefiVault {
...
+  }
+
+  /**
+   * Concrete BTC vault deposit path. The client BTC wallet is custodial, so
+   * sendMany returns a pendingApproval rather than a signed transfer.
+   * No recipients are sent — WP resolves the escrow destination server-side.
+   */
+  private async depositToConcreteVault(params: DepositToVaultOptions): Promise<ConcreteDepositResult> {
+    const sendManyResult = await this.wallet.sendMany({
+      type: 'defi-deposit',
```
nvjsr, line 108: can we have common intent type case as I can see most of intent type follow camel case but here we are using  Kebab Case ?

venkateshv1266: You can see the intentType being followed for other protocols in the same case: [https://github.com/BitGo/public-types/blob/cd468bc6db1cf4d47ce69a54100754ea5850c167/src/schema/transactionRequest/intents/defiDepositIntent.ts#L7](<https://github.com/BitGo/public-types/blob/cd468bc6db1cf4d47ce69a54100754ea5850c167/src/schema/transactionRequest/intents/defiDepositIntent.ts#L7>)

venkateshv1266: Same in BitGoJS sdk too, [https://github.com/BitGo/BitGoJS/blob/10e565144072382dcc4ab7caa5aabbfc9c7823c6/modules/sdk-core/src/bitgo/wallet/wallet.ts#L4440-L4469](<https://github.com/BitGo/BitGoJS/blob/10e565144072382dcc4ab7caa5aabbfc9c7823c6/modules/sdk-core/src/bitgo/wallet/wallet.ts#L4440-L4469>)

nvjsr: Since we are building a new service, we should align the intent type naming convention with what we follow in WP. The existing intent types use camelCase, so it would be better to keep the same convention here instead of introducing kebab-case.

nvjsr:
CI is failing

modules/sdk-core/src/bitgo/defi/defiVault.ts
venkateshv1266: In case of morpho, these changes are already done and its live in prod i guess cc: @kamleshmugdiya @hafeezshaik933

OttoAllmendinger (changes requested):
why are we still not using typed routes?

modules/sdk-core/src/bitgo/defi/defiVault.ts
```diff
@@ -256,6 +309,21 @@ export class DefiVault implements IDefiVault {
...
+   * Concrete BTC deposits return a `pendingApproval` instead of a txRequest —
+   * throws if `pendingApproval.id` is absent, indicating an unexpected shape.
+   */
+  private extractConcreteDepositResult(sendManyResult: Record<string, unknown>): ConcreteDepositResult {
+    const pendingApproval = sendManyResult.pendingApproval as Record<string, unknown> | undefined;
+    const id = pendingApproval?.id as string | undefined;
+    if (!id) {
+      throw new Error('Unexpected sendMany response for defi-deposit: no pendingApproval.id');
+    }
+    const state = (pendingApproval?.state as string | undefined) ?? 'awaitingSignature';
```
OttoAllmendinger: don't cast, use io-ts to validate (inline codec is fine)

OttoAllmendinger (changes requested):

modules/sdk-core/src/bitgo/defi/defiVault.ts
venkateshv1266: Done

OttoAllmendinger (dismissed):

OttoAllmendinger (dismissed):

vmccarty:
Breaking change linter override approve. Re:

Breaking change linter override approved. Re:

None of the 98 findings is the removal of a required, client-facing, in-production response field. So the premise ("each was intentionally broken elsewhere and got a DevEx override") doesn't map 1:1 — most aren't breaks at all. But every real change traces to a deliberate, reviewed, released public-types PR with a Linear ticket, and several were explicitly coordinated with the owning service:

Irys revert → #396 (SC-5765/6/7), owner-driven
DeFi clientIdempotencyKey/restructure → #389 (CGD-1400), [#414](https://github.com/BitGo/BitGoJS/issues/414) (CGD-2007, body: "Unblocks CGD-1821 (wallet-platform) and CGD-1822 (BitGoJS)")
Canton choiceArgument optional → v6.26.0
goStaking optionality → v6.29.0
The reason there's no override "in the public-types repo" is exactly the gap you identified: public-types has no breaking-change CI, and the consuming service (wallet-platform, which owns signtxtss/sweep) drove these via tickets rather than through a spec-diff gate.

Recommendation
Both PRs (9232 → 6.41.0; 9211 → ^6.39.0, same range, subset of the same findings) are safe to override and approve on client-impact grounds — the breaking-change gate is misfiring on oneOf reindexing, not detecting real contract breaks. I'd note in the DevEx override request: "All findings are positional oneOf artifacts in the signtxtss/sweep intent union from additive public-types changes 6.22→6.41 plus the intentional Irys revert ([#396](https://github.com/BitGo/BitGoJS/issues/396)); verified surviving fields still present; no required client-facing field removed."

Two things I can do next if useful:

Pull the literal override/approval paper trail from wallet-platform (owner of these endpoints) to attach as corroboration — I'd need to add that repo to the session.
For the CI/pipeline fix: the concrete bug is the positional-index diff on large unions. The durable fix is a name/discriminator-aware oneOf comparison (match variants by intentType literal, not array position) in BitGo/static-analysis's breaking-changes rule — I can scope that out.
Want me to grab the wallet-platform override records, or start on the static-analysis fix?

venkateshv1266:
Removed the bump commit (`chore(sdk-core): bump @bitgo/public-types to 6.40.1`) since a newer version was already merged to master, which was causing merge conflicts. Rebased the branch onto latest master — now only the feat commit remains.

Marzooqa (approved):

Marzooqa (commented):
Codeowners should be updated

PR #9213, opened 2026-07-08
fix(bitgo): ship npm-shrinkwrap.json for npm consumers

mihir-bitgo:
## Summary
- `overrides` (#9212) only pins versions when `bitgo` is the *root* of an `npm install`; consumers installing `bitgo` as a dependency still resolve unpinned transitive versions
- Add a `prepack` script on `modules/bitgo` (`scripts/generate-bitgo-shrinkwrap.ts`) that generates `npm-shrinkwrap.json` in an isolated temp copy (outside the workspace, since `npm shrinkwrap` isn't workspace-aware and `modules/bitgo/.npmrc` disables `package-lock`) and ships it in the published tarball
- Using `prepack` instead of a separate release-pipeline step means it fires after lerna bumps the version on disk but before packing, so the shrinkwrap's top-level version always matches what's actually published
- Add a post-publish workflow step that fails the release if the registry's `hasShrinkwrap` metadata isn't `true` — this was never confirmed on a real npm publish per the ticket (Verdaccio didn't set it)

## Test plan
- [x] Both files (`.github/workflows/npmjs-release.yml`, `modules/bitgo/package.json`) parse/lint cleanly
- [ ] **Canary publish required before merge**: publish `bitgo@0.0.0-canary.N` from a temp copy including a generated shrinkwrap and confirm `npm view bitgo@<canary> dist.hasShrinkwrap` is `true`
- [ ] Fresh-install the canary as a dependency in a throwaway project and confirm pinned versions land / `npm audit` is clean
- [ ] Dry-run the release workflow to confirm the new verification step doesn't false-fail

Depends on #9212 (root `overrides`) merging first so the generated shrinkwrap reflects the pinned tree.

TICKET: WCN-604

### DIFF
--- a/.github/workflows/npmjs-release.yml
+++ b/.github/workflows/npmjs-release.yml
@@ -325,6 +325,22 @@ jobs:
           fi
           echo "✅ All public package versions on rel/latest are present on npm."
 
+      # bitgo's `prepack` script generates npm-shrinkwrap.json at pack time (see
+      # scripts/generate-bitgo-shrinkwrap.ts). npm only honors a shipped shrinkwrap
+      # for installers when the registry's own metadata says hasShrinkwrap: true —
+      # confirm that here so a silent registry-side omission fails the release
+      # instead of shipping a package whose shrinkwrap consumers won't get.
+      - name: Verify bitgo package has shrinkwrap metadata
+        if: ${{ inputs.dry-run == false }}
+        run: |
+          version=$(jq -r '.version' modules/bitgo/package.json)
+          has_shrinkwrap=$(curl -sL "https://registry.npmjs.org/bitgo/${version}" | jq -r '.dist.hasShrinkwrap // false')
+          if [ "$has_shrinkwrap" != "true" ]; then
+            echo "::error::Published bitgo@${version} is missing hasShrinkwrap metadata on the npm registry."
+            exit 1
+          fi
+          echo "✅ bitgo@${version} published with hasShrinkwrap: true."
+
       - name: Generate version bump summary
         id: version-bump-summary
         if: inputs.dry-run == false
--- a/.github/workflows/npmjs-release.yml
+++ b/.github/workflows/npmjs-release.yml
@@ -290,6 +290,10 @@ jobs:
           yarn lerna publish --sign-git-tag --sign-git-commit --include-merged-tags --conventional-commits --conventional-graduate --yes
         env:
           NPM_CONFIG_PROVENANCE: true
+          # Tells bitgo's `prepack` script (scripts/generate-bitgo-shrinkwrap.ts) to
+          # actually generate npm-shrinkwrap.json — kept opt-in so a plain local
+          # `npm pack` doesn't force a network install of the full dependency tree.
+          BITGO_GENERATE_SHRINKWRAP: true
 
       - name: Publish missing versions (recovery)
         if: ${{ inputs.dry-run == false && inputs.recovery-mode }}
@@ -300,6 +304,7 @@ jobs:
           yarn lerna publish from-package --yes
         env:
           NPM_CONFIG_PROVENANCE: true
+          BITGO_GENERATE_SHRINKWRAP: true
 
       - name: Verify recovery published the missing versions
         if: ${{ inputs.dry-run == false && inputs.recovery-mode }}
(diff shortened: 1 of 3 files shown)


OttoAllmendinger:
## Review

### Overview
Adds a `prepack` script to `modules/bitgo` that generates `npm-shrinkwrap.json` in an isolated temp copy (stripping `devDependencies`/`scripts`, injecting root `overrides`, resolving against the real registry), ships it via the `files` array, and adds a post-publish workflow step that fails the release if npm's `dist.hasShrinkwrap` isn't `true`. The mechanism is sound — a dependency's `npm-shrinkwrap.json` *is* honored by npm to pin that dependency's own subtree, so this is the right lever for the goal.

### 🔴 Critical — sibling version resolution at pack time
`modules/bitgo` depends on **88 `@bitgo/*` siblings**, and the shrinkwrap is generated by running `npm install --package-lock-only` against the **public npm registry**. `prepack` fires inside `yarn lerna publish`, which bumps versions and packs. At the moment bitgo is packed, the newly-bumped sibling versions are almost certainly **not yet published/propagated** on registry.npmjs.org. Two failure modes, both bad:
1. `npm install` can't resolve `@bitgo/sdk-core@<new-version>` → pack fails → **every release blocked**.
2. It resolves the *previous* published versions → the shipped shrinkwrap **pins stale sibling versions**, arguably worse than no shrinkwrap.

The canary plan won't surface this — a standalone `bitgo@0.0.0-canary.N` publishes against an already-published sibling tree, hiding the exact ordering hazard of a coordinated monorepo release. Recommend validating against a real/dry-run lerna release, and consider resolving siblings from the local workspace tarballs rather than the registry.

### 🟠 Significant
- **`prepack` now requires network + a full `npm install` on every pack**, including local/dev/offline packing (slow 88+ dep resolution; hard failure offline). Consider gating on a CI/release env var so local packs aren't broken.
- **`hasShrinkwrap` verification may permanently false-fail.** The PR body admits `dist.hasShrinkwrap` was never confirmed on a real npm publish; if npm doesn't populate it, this step fails every release. Querying the registry immediately post-publish is also subject to metadata propagation lag — add retry/backoff (the workflow already uses that pattern elsewhere). Confirm the field is real before merge.
- **Undeclared dependency on root-hoisted `tsx`/`execa`.** Both are root `devDependencies`, not declared in `modules/bitgo`. Fine within the hoisted workspace publish, but the script silently relies on hoisting.

### 🟡 Minor / style
- `const { devDependencies, scripts, ...isolatedPackageJson }` — confirm eslint `no-unused-vars` has `ignoreRestSiblings: true`, else this lint-fails.
- Script otherwise matches existing `scripts/*.ts` conventions (execa default import, `main().catch(...)`, `process.exit(1)`). 👍
- `if (!rootPackageJson.overrides)` fail-fast is good; correctly depends on #9212.
- Consider logging which versions got pinned for release auditability, given the stale-pin risk.

### Test coverage
No automated test (reasonable for release tooling), but given the sibling-timing risk, the unchecked canary/dry-run boxes should be treated as merge blockers.

### Bottom line
Right mechanism, but the **pack-time sibling resolution against the public registry** is a real correctness/reliability hazard for a coordinated monorepo release, and the `hasShrinkwrap` gate could block releases if that metadata field isn't actually populated. Validate both on a real (dry-run) release before merge — the standalone canary alone is insufficient.

---
_Generated by [Claude Code](https://claude.ai/code)_

OttoAllmendinger (commented):
see comment

mihir-bitgo:
Thanks for the review — the critical finding is correct and I hadn't accounted for the sibling-publish race. Fixed:

- **Critical (sibling resolution race):** `@bitgo/*` deps are now excluded from the temp package.json before resolution, so `npm install` never touches the public registry for a just-bumped-but-not-yet-published sibling. Consumers resolve `@bitgo/*` siblings normally at install time (by then they're published); everything transitively pulled in beneath a sibling still gets pinned via `overrides`. The shrinkwrap's own `dependencies` field is patched back to the real sibling list after generation so what's shipped still matches package.json.
- **`prepack` forcing network installs on local/offline packs:** generation is now gated behind `BITGO_GENERATE_SHRINKWRAP=true`, set only on the release workflow's publish steps.
- **`hasShrinkwrap` propagation lag:** wrapped the registry check in the job's existing `retry` binary (`--up-to 5x --every 5s`) instead of a single curl.
- **Lint:** dropped the unused-var destructure (`no ignoreRestSiblings` in this repo's eslint config) in favor of explicit `delete`.
- Undeclared `tsx`/`execa` — left as-is, intentionally relying on root hoisting like other `scripts/*.ts` in this repo call out.

Agreed the canary-only validation was insufficient for the sibling-timing hazard specifically — that risk is now structurally eliminated rather than something to catch in testing. Canary/dry-run validation of the `hasShrinkwrap` metadata assumption is still outstanding before merge.

davidkaplanbitgo (approved):

PR #9216, opened 2026-07-08
feat(sdk-core): pass optional attestation through multisig send/initiate

mr-neptune:
## Summary
Add a local `AttestationPayload` codec and thread it through` wallet.sendMany'`s build/send/initiate path so a WebAuthn attestation survives to `POST /tx/send` and `/tx/initiate`: appended to `BuildParams` (clears the `sendMany` prebuild-whitelist pick), to `whitelistedSendParams`, and re-added via a local `t.intersection` partial since @bitgo/public-types' `TxSendBody` (t.exact) doesn't declare the field yet and would strip it on encode. Pure pass-through — no attestation validation in the SDK.

Scoped to what WCN-539 (UI) needs end-to-end: multisig only. The shared upstream codec change and the `/txrequests`, `/signatureshares`, and `/pendingapprovals` endpoints are left as TODO(WCN-541) markers for the dedicated SDK/Express pass-through ticket.

Ticket: WCN-539

### DIFF
--- a/modules/sdk-core/test/unit/bitgo/wallet/SendTransactionRequest.ts
+++ b/modules/sdk-core/test/unit/bitgo/wallet/SendTransactionRequest.ts
@@ -1,5 +1,7 @@
 import * as assert from 'assert';
+import * as t from 'io-ts';
 import { TxSendBody } from '@bitgo/public-types';
+import { AttestationPayload } from '../../../../src/bitgo/wallet/BuildParams';
 
 describe('SendTransactionRequest', function () {
   it('enforces codec', function () {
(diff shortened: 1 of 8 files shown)


Marzooqa:
@claude

modules/sdk-core/test/unit/bitgo/wallet/SendTransactionRequest.ts
```diff
@@ -24,4 +26,31 @@ describe('SendTransactionRequest', function () {
       addressType: 'p2sh',
     });
   });
+
+  it('TxSendBody alone drops attestation (upstream codec has no such field yet — TODO(WCN-541))', function () {
```
OttoAllmendinger: let's just fix the upstream codec first instead of hacking around it

OttoAllmendinger (changes requested):

modules/sdk-core/test/unit/bitgo/wallet/SendTransactionRequest.ts
mr-neptune: Addressed

OttoAllmendinger (changes requested):
CI conflicts

please do a separate initial PR for bumping the package

mr-neptune:
@OttoAllmendinger Addressed your comments, pls re-review at your convenience, thank you!

OttoAllmendinger (approved):

PR #9220, opened 2026-07-09
fix(sdk-coin-polyx): route v8 batchAll(bond,nominate) parse to v8 builder

abhijit0943:
## What

`TransactionBuilderFactory.getBuilder()` routed every `batchAll([bond, nominate])` to the v7 `BatchStakingBuilder` on the decode-success path. When the transaction is decoded against v8 material, the bond call carries no `controller` (v8 changed `bond(controller, value, payee)` → `bond(value, payee)`), so v7 `validateBondArgs` crashed with `Cannot read properties of undefined (reading 'id')` while accessing `args.controller.id`.

## Why

`tryGetV8Builder()` only runs when v7 `decode()` throws. Unlike v8 transfers (whose shifted call indices make v7 decode fail), v8 staking `batchAll` can decode without throwing, so the v8 fallback never fired and the tx was misrouted — breaking POLYX v8 stake dry-runs on staging.

## How

In the `batchAll([bond, nominate])` branch, inspect the decoded bond args: route to `V8BatchStakingBuilder` when no `controller` is present, and keep v7 bond+nominate on `BatchStakingBuilder`. Scoped to the factory only — no builder implementation changes.

## Testing

- New test reproduces the staging incident: with v8 material set, `factory.from()` on a v8 `batchAll([bond, nominate])` returns `V8BatchStakingBuilder` instead of crashing.
- New regression test confirms a v7 bond+nominate tx still routes to `BatchStakingBuilder`.
- `yarn tsc`, lint, and the polyx staking test files pass.

Linked to SI-981.

### DIFF
--- a/modules/sdk-coin-polyx/src/lib/transactionBuilderFactory.ts
+++ b/modules/sdk-coin-polyx/src/lib/transactionBuilderFactory.ts
@@ -220,6 +220,16 @@ export class TransactionBuilderFactory extends BaseTransactionBuilderFactory {
 
         // Check for batch staking pattern: bond + nominate
         if (firstCallMethod === 'bond' && secondCallMethod === 'nominate') {
+          // v8 dropped the controller leg of `staking.bond` (bond(controller, value, payee) →
+          // bond(value, payee)). When the transaction was decoded against v8 material the bond call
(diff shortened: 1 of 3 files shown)


abhijit0943:
@claude review

nvrakesh06 (approved):

PR #9209, opened 2026-07-08
feat: multi-root layout in keycard module for vaults

s84krish:
Ticket: WCN-1193

## Description

Keycard generation for wallet vault recovery. Combines 4 root keys into single boxes - upon decryption expect a JSON containing keys for each signing scheme (mock below):

```
{ 
"secp256k1Multisig":"xprv9s21ZrQH143K44CWrDtJ8fRrU94Yk7scQVBaUZce…",  
"ecdsaMpc": "uQAFaGJpZ1NMaXN0g5ghAxipGHYY…",                        
"eddsaMpc": "uQADaGtleVNoYXJlmIQCAwAY6…",                           
"ed25519Multisig": "xprv9s21ZrQH143K2aZdhLN92yi2Ei47MZ1JCEvjRSmxJ6…"      
}
```

**Vault keycard support and data handling:**

* Added new types (`VaultRootKeyType`, `VaultKeycardRoots`, `GenerateVaultQrDataParams`) and constants (`VAULT_ROOT_ORDER`) to represent and order vault root keys, and updated interfaces to support vault-specific data (`modules/key-card/src/types.ts`). [[1]](diffhunk://#diff-ca331280fee147ecf67051e855d143211b9e6120b9a6040a2687e02d2e8301f9R59-R83) [[2]](diffhunk://#diff-ca331280fee147ecf67051e855d143211b9e6120b9a6040a2687e02d2e8301f9R94-R95)
* Implemented `generateVaultQrData`, which serializes all four vault roots per role into JSON objects for each keycard box, and ensures correct selection of encrypted or public key material (`modules/key-card/src/generateQrData.ts`).
* Added `parseVaultKeycardBox` to parse and validate the vault box JSON data, ensuring all roots are present and correctly formatted (`modules/key-card/src/parseKeycard.ts`).

**Keycard rendering and layout improvements:**

[...]

PR #9225, opened 2026-07-09
chore: use io-ts to parse safe keycard box + parts encoding in QR headers for multipart QRs

s84krish:
Ticket: WCN-1193

## Description

This PR addresses followups from https://github.com/BitGo/BitGoJS/pull/9209#discussion_r3546011493. Also includes a new optional part-based encoding for multipart QRs. Mock below:

```
1/2|{"secp256k1Multisig":"{\"v\":2,\"m\":65536,\"t\":3,\"p\":4,\"salt\":\"ABleiJ/ySLZvoaYswbRzTg==\",\"iv\":\"VmwUpGXO2LJ1OSnC\",\"ct\":\"NBRX+x21OL7RkrH7yDjdX6hk9+f55zadWqNFX2/xvo9lapluoeZcOxifnplHH+YkKKTkCJYK74695iry0Ch2RMChYpjXlYWK2VfhWiAC24CRtBc8e8HEyNWnfnV7Q5UJlwbb8Wc6zBwlm+UC2gWQr0k04zKKJS40w2FK9ywOiQ==\"}","ecdsaMpc":"{\"v\":2,...(DKLS reduced share ciphertext)...","eddsaMpc":"{\"v\":2,... [truncated mid-value where the 1500-char split falls]
```

```
2/2|tainZcXm355bbrxTaoULbjqKJE5XiLVyQRjygnImY1ruxsgTXo=\"}","eddsaMpc":"{\"v\":2,...}","ed25519Multisig":"{\"v\":2,...\"ct\":\"Dtt5R7Q8...==\"}"}
```

Copy and formatting updates have been folded into: https://github.com/BitGo/BitGoJS/pull/9242

## Type of change

# How Has This Been Tested?

Generated PDFs locally for vault and wallet keycards using mock data. 

[safe-keycard.pdf](https://github.com/user-attachments/files/29984258/safe-keycard.pdf)

[wallet-keycard.pdf](https://github.com/user-attachments/files/29984260/wallet-keycard.pdf)

### DIFF
--- a/modules/key-card/src/drawKeycard.ts
+++ b/modules/key-card/src/drawKeycard.ts
@@ -58,6 +58,36 @@ function moveDown(y: number, ydelta: number): number {
   return y + ydelta;
 }
 
+/**
+ * Prefixes one fragment of a split key with a part header so a QR scanner can reassemble the
+ * fragments without relying on scan order.
+ *
+ * When a key box's payload exceeds {@link QRBinaryMaxLength}, {@link splitKeys} divides it into
+ * multiple QR codes. Each fragment's QR payload is encoded as a 1-based part header followed by
+ * the fragment:
+ *
+ *     "<index>/<total>|<fragment>"      e.g.  "1/3|<chunk>", "2/3|<chunk>", "3/3|<chunk>"
+ *
+ * A single-fragment payload is returned unchanged (no header).
+ *
+ * Reassembly contract for a consumer (e.g. a future recovery/scan tool):
+ *   1. Scan every QR code for a box.
+ *   2. Split each payload on the FIRST "|": the left side is "<index>/<total>", the right side
+ *      is the fragment.
+ *   3. Verify parts 1..total are all present (total is identical in every header).
+ *   4. Concatenate the fragments in ascending index order to recover the full box payload.
+ *   5. Parse/decrypt as usual (for a safe box: JSON.parse, then decrypt each root value).
+ *
+ * Notes:
+ *   - The "|" delimiter is safe: base64 ciphertext, JSON, and base58 xpubs never contain it.
+ *   - This header exists ONLY inside the QR image. The human-readable "Data:" text printed on
+ *     the card is the full, unheadered payload; the PDF-text parser reads that, so this header
+ *     does not affect PDF-based recovery.
+ */
+function encodeQrCodePart(fragment: string, index: number, total: number): string {
+  return total > 1 ? `${index + 1}/${total}|${fragment}` : fragment;
+}
+
 // Draws QR codes down the left column, returning the index of the next QR still to draw (for
 // continuation on a later page) and the y-offset just below the drawn QR column (so callers
 // can place content, e.g. a note, under the QR codes).
@@ -124,6 +154,7 @@ export async function drawKeycard({
   walletLabel,
   curve,
   pageBreakBeforeIndices = DEFAULT_PAGE_BREAK_INDICES,
+  useQrPartHeaders = false,
 }: IDrawKeyCard): Promise<jsPDF> {
   const jsPDFModule = await loadJSPDF();
 
@@ -218,11 +249,12 @@ export async function drawKeycard({
 
     const qrImages: (HTMLCanvasElement | string)[] = [];
     const keys = splitKeys(qr.data, QRBinaryMaxLength);
-    for (const key of keys) {
+    for (let i = 0; i < keys.length; i++) {
+      const payload = useQrPartHeaders ? encodeQrCodePart(keys[i], i, keys.length) : keys[i];
       if (isNode) {
-        qrImages.push(await QRCode.toDataURL(key, { errorCorrectionLevel: 'L' }));
+        qrImages.push(await QRCode.toDataURL(payload, { errorCorrectionLevel: 'L' }));
       } else {
-        qrImages.push(await QRCode.toCanvas(key, { errorCorrectionLevel: 'L' }));
+        qrImages.push(await QRCode.toCanvas(payload, { errorCorrectionLevel: 'L' }));
       }
     }
 
--- a/modules/key-card/src/parseKeycard.ts
+++ b/modules/key-card/src/parseKeycard.ts
@@ -1,4 +1,8 @@
-import { SafeKeycardRoots, SAFE_ROOT_ORDER } from './types';
+import * as t from 'io-ts';
+import { isLeft } from 'fp-ts/Either';
+import { PathReporter } from 'io-ts/lib/PathReporter';
+import { JsonFromString } from 'io-ts-types';
+import { SafeKeycardRoots } from './types';
 
 export type PDFTextNode = {
   text: string;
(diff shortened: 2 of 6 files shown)


s84krish:
@claude

modules/key-card/src/drawKeycard.ts
```diff
@@ -56,34 +56,67 @@ function moveDown(y: number, ydelta: number): number {
...
+ *   5. Parse/decrypt as usual (for a vault box: JSON.parse, then decrypt each root value).
+ *
+ * Notes:
+ *   - The "|" delimiter is safe: base64 ciphertext, JSON, and base58 xpubs never contain it.
+ *   - This header exists ONLY inside the QR image. The human-readable "Data:" text printed on
+ *     the card is the full, unheadered payload; the PDF-text parser reads that, so this header
+ *     does not affect PDF-based recovery.
+ */
+function encodeQrCodePart(fragment: string, index: number, total: number): string {
+  return total > 1 ? `${index + 1}/${total}|${fragment}` : fragment;
```
s84krish, line 88: Wondering what the best way to do this parts encoding would be - if it is even necessary.

modules/key-card/src/drawKeycard.ts
```diff
@@ -203,11 +237,26 @@ export async function drawKeycard({
...
     const keys = splitKeys(qr.data, QRBinaryMaxLength);
-    for (const key of keys) {
-      qrImages.push(await QRCode.toCanvas(key, { errorCorrectionLevel: 'L' }));
+    for (let i = 0; i < keys.length; i++) {
+      const payload = useQrPartHeaders ? encodeQrCodePart(keys[i], i, keys.length) : keys[i];
+      qrImages.push(await QRCode.toCanvas(payload, { errorCorrectionLevel: 'L' }));
     }
 
-    let nextQrIndex = drawOnePageOfQrCodes(qrImages, doc, y, qrSize, 0);
+    const isMultiPart = qr?.data?.length > QRBinaryMaxLength;
```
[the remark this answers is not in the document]
s84krish: negligible since it is intended to provide a way to keep track of parts and ordering

modules/key-card/src/drawKeycard.ts
```diff
@@ -56,34 +56,67 @@ function moveDown(y: number, ydelta: number): number {
...
+      return { nextIndex: qrIndex + 1, endY: y + qrSize };
     }
 
     y = moveDown(y, qrSize + textBuffer);
     doc.setFontSize(font.body).setTextColor(color.black);
     doc.text('Part ' + (qrIndex + 1).toString(), left(0), y);
     y = moveDown(y, 20);
   }
-  return qrIndex + 1;
+  return { nextIndex: qrIndex + 1, endY: y };
```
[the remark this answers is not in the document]
s84krish: Valid, updated

modules/key-card/src/parseKeycard.ts
```diff
@@ -13,29 +17,31 @@ export type KeycardEntry = {
...
+
 /**
  * Parses a vault keycard box value — the JSON packed by `generateVaultQrData`, e.g.
- * `{"secp256k1Multisig":"…","ecdsaMpc":"…",…}` — into its four roots. Validates that all four
- * roots are present. Recovery tooling calls this on the A/B/C box value returned by
- * {@link parseKeycardFromLines}, then decrypts each root value with the wallet password.
+ * `{"secp256k1Multisig":"…","ecdsaMpc":"…",…}` — into its four roots. Throws if the value is
+ * not valid JSON or any root is missing/non-string. Recovery tooling calls this on the A/B/C
+ * box value returned by {@link parseKeycardFromLines}, then decrypts each root value with the
+ * wallet password.
```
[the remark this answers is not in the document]
s84krish: updated

OttoAllmendinger (commented):
flush

modules/key-card/src/generateQrData.ts
```diff
@@ -132,16 +132,21 @@ function generateUserMasterPublicKeyQRData(publicKey: string): MasterPublicKeyQr
   };
 }
 
+// The product a keycard belongs to, used for user-facing wording (e.g. Box D copy).
+type KeycardEntity = 'wallet' | 'vault';
```
pranishnepal: should this go to the types? wdyt?

modules/key-card/src/drawKeycard.ts
```diff
@@ -273,9 +321,10 @@ export async function drawKeycard({
     }
 
     doc.setFont('helvetica');
-    // Move down the size of the QR code minus accumulated height on the right side, plus margin
-    // if we have a key that spans multiple pages, then exclude QR code size
-    const rowHeight = Math.max(qr.data.length > QRBinaryMaxLength ? qrSize + 20 : qrSize, textHeight);
+    // Move down past the taller of the two columns: the right-side text (textHeight) or the
+    // left-side QR column (plus the note printed under it for split keys), then add a margin.
```
pranishnepal, line 341: what with the +10? is there a reasoning for it? if there is -- maybe add a comment?

pranishnepal (commented):
added my commentary - will defer to folks involve in the project for a thorough review

modules/key-card/src/generateQrData.ts
s84krish: Agree, would be cleaner and could be used elsewhere. Moved!

modules/key-card/src/drawKeycard.ts
s84krish: Was a stylistic decision to add some padding between the the boxes, but I removed it and it is very negligible. Updated to remove it entirely!

zahin-mohammad (commented):
merge conflicts

zahin-mohammad (commented):
## Review

Verified locally: `tsc --noEmit` clean, 32/32 unit tests pass. Traced the layout math and QR-header logic by hand.

### Blocking: still uses "vault" — we renamed the product to "safe"

The whole module still says **vault** in both customer-facing copy and the public API, and this PR compounds it (adds `'vault'`/`'Vault'` copy paths). Needs a vault→safe pass:

- **Customer-facing copy** (highest priority — printed on the keycard):
  - `generateQrData.ts` — `titleNoun = ... ? 'Vault' : 'Wallet'` renders **"D: Encrypted Vault Password"**
  - `generateQrData.ts` — "…encrypted with your **vault** password." (x2) + the `'vault'` literal passed to `generatePasscodeQrData`
  - `parseKeycard.ts` — the `'encrypted vault password'` title matcher must change **in lockstep** with the rendered title, or PDF recovery of Box D breaks
- **Public API** (`generateVaultKeycard`, `generateVaultQrData`, `GenerateVaultQrDataParams`, `VaultKeycardRoots`, `VaultRootKeyType`, `VAULT_ROOT_ORDER`, `parseVaultKeycardBox`, `KeycardEntity: 'vault'`) plus internals (`buildVaultBoxData`, codecs) and the `vaultQrData.ts` test file.

Low-risk: these symbols have **zero consumers elsewhere in BitGoJS** and the feature is unreleased, so renaming breaks nothing. Note most of the "vault" naming came in the prior commit `15fdb259d5`, so a complete rename spans both commits. Leave the `faq.ts` "bank vault or home safe" line alone — that's a real-world example, not the product term.

### No other breaking changes
- Wallet keycard output is unchanged: `useQrPartHeaders` defaults `false` and `entityNoun` defaults `'wallet'`.
- New QR part-header format (`"<i>/<n>|<fragment>"`) has no existing consumer (reassembly is for a future recovery tool).
- Changed `parseVaultKeycardBox` error strings — grepped repo, no external matchers.
- `drawOnePageOfQrCodes` signature change is internal (not exported).

### Correctness — looks sound
Traced the refactored `rowHeight`, including multi-part keys overflowing to a second page (continuation resets `qrColumnBottom` to the new page; when all QRs already fit page 1 the page-2 call returns `endY == topY` so height falls back to `textHeight`). Single/non-multipart path pins `qrColumnHeight = qrSize`, matching prior behavior.

### Non-blocking notes
1. **Three new deps in a browser-bundled module** (`fp-ts` + `io-ts` + `io-ts-types`) replace a ~15-line dependency-free validator for a flat 4-key object. Intentional (io-ts is the repo standard, reviewer-requested) — flagging the bundle-weight trade-off.
2. **Dep version drift**: `fp-ts@^2.16.2` / `io-ts-types@^0.5.19` vs `^2.12.2` / `^0.5.16` in abstract-utxo/lightning. Consider aligning.
3. **`drawKeycard.ts` layout changes are untested** (~10% coverage) — verified only via the attached PDFs. This is the riskiest part of the diff; a regression would be silent. Consider a unit test for the multi-part `rowHeight`/note placement.
4. Header is added **after** splitting, so a QR fragment can slightly exceed `QRBinaryMaxLength` (1500 → ~1505). Harmless, well under the 2953 cap.

s84krish:
Will update name from vault -> safe and then rebase.

zahin-mohammad (approved):
Approving — clean, well-scoped change. io-ts refactor preserves all failure modes (verified against the tests) and the QR part-header path is opt-in so the wallet keycard output is unchanged. The Data:/QR asymmetry is correct: headers apply only to the QR image, so the PDF-text parser is unaffected.

One non-blocking follow-up: the new QR encoding is untested. `encodeQrCodePart` and the `useQrPartHeaders` branch in `drawKeycard` have no coverage — the suite only exercises `parseSafeKeycardBox`. Since the `<index>/<total>|<fragment>` format is a forward contract for a future scan/recovery tool, a small unit test locking it down (multi-part `1/3|…`/`3/3|…` plus single-fragment passthrough) would be worth adding before anything depends on the format.

PR #9232, opened 2026-07-10
feat: bump public types

davidkaplanbitgo:
WCN-1175

TICKET: WCN-1175

### DIFF
--- a/modules/abstract-lightning/package.json
+++ b/modules/abstract-lightning/package.json
@@ -39,7 +39,7 @@
     ]
   },
   "dependencies": {
-    "@bitgo/public-types": "6.22.0",
+    "@bitgo/public-types": "6.41.0",
     "@bitgo/sdk-core": "^38.1.0",
     "@bitgo/statics": "^58.55.0",
     "@bitgo/utxo-lib": "^11.24.0",
--- a/modules/bitgo/package.json
+++ b/modules/bitgo/package.json
@@ -141,7 +141,7 @@
     "superagent": "^9.0.1"
   },
   "devDependencies": {
-    "@bitgo/public-types": "6.22.0",
+    "@bitgo/public-types": "6.41.0",
     "@bitgo/sdk-opensslbytes": "^2.1.0",
     "@bitgo/sdk-test": "^9.1.58",
     "@openpgp/web-stream-tools": "0.0.14",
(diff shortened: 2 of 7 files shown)


zahin-mohammad (approved):

vmccarty:
Breaking change linter override approved. Re:

None of the 98 findings is the removal of a required, client-facing, in-production response field. So the premise ("each was intentionally broken elsewhere and got a DevEx override") doesn't map 1:1 — most aren't breaks at all. But every real change traces to a deliberate, reviewed, released public-types PR with a Linear ticket, and several were explicitly coordinated with the owning service:

Irys revert → #396 (SC-5765/6/7), owner-driven
DeFi clientIdempotencyKey/restructure → #389 (CGD-1400), [#414](https://github.com/BitGo/BitGoJS/issues/414) (CGD-2007, body: "Unblocks CGD-1821 (wallet-platform) and CGD-1822 (BitGoJS)")
Canton choiceArgument optional → v6.26.0
goStaking optionality → v6.29.0
The reason there's no override "in the public-types repo" is exactly the gap you identified: public-types has no breaking-change CI, and the consuming service (wallet-platform, which owns signtxtss/sweep) drove these via tickets rather than through a spec-diff gate.

Recommendation
Both PRs (9232 → 6.41.0; 9211 → ^6.39.0, same range, subset of the same findings) are safe to override and approve on client-impact grounds — the breaking-change gate is misfiring on oneOf reindexing, not detecting real contract breaks. I'd note in the DevEx override request: "All findings are positional oneOf artifacts in the signtxtss/sweep intent union from additive public-types changes 6.22→6.41 plus the intentional Irys revert ([#396](https://github.com/BitGo/BitGoJS/issues/396)); verified surviving fields still present; no required client-facing field removed."

Two things I can do next if useful:

Pull the literal override/approval paper trail from wallet-platform (owner of these endpoints) to attach as corroboration — I'd need to add that repo to the session.
For the CI/pipeline fix: the concrete bug is the positional-index diff on large unions. The durable fix is a name/discriminator-aware oneOf comparison (match variants by intentType literal, not array position) in BitGo/static-analysis's breaking-changes rule — I can scope that out.
Want me to grab the wallet-platform override records, or start on the static-analysis fix?

PR #9267, opened 2026-07-15
feat: implement safe createSafeKeys and finalize orchestrator

davidkaplanbitgo:
Implement Phase 2 (root key ceremonies) and Phase 3 (finalize) of safe creation in sdk-core, plus the generateSafe orchestrator chaining all three phases.

createSafeKeys runs the 4 root ceremonies tagged with safeId and returns the 12 key ids as ordered [user, backup, bitgo] triplets:
- 2.1 multisig roots (secp256k1Multisig, ed25519Multisig): local user/backup keypairs + BitGo key via the existing keychain flows.
- 2.2 MPC roots (ecdsaMpc DKLS, eddsaMpc EdDSA): the existing MPCv2 ceremonies via createMpc.

Thread an optional safeId through the shared key-generation paths: keychains.add/createMpc, the ecdsa/eddsa MPCv2 createKeychains ceremonies (round 1 generatekey envelope + participant registration), and their key-gen senders. Legacy MPCv1 paths accept and ignore safeId for signature compatibility.

finalizeSafe posts the rootKeys to the finalize route using the existing public-types codec; generateSafe chains init -> keys -> finalize.

WCN-1192

### DIFF
--- a/modules/sdk-core/src/bitgo/safe/safes.ts
+++ b/modules/sdk-core/src/bitgo/safe/safes.ts
@@ -23,16 +23,20 @@ import { Safe } from './safe';
  */
 const ROOT_COIN_BY_NETWORK: Record<'mainnet' | 'testnet', Record<RootKeyType, string>> = {
   mainnet: {
+    // multisig roots
     secp256k1Multisig: 'btc',
+    ed25519Multisig: 'xlm',
+    // MPC roots
     ecdsaMpc: 'eth',
     eddsaMpc: 'sol',
-    ed25519Multisig: 'algo',
   },
   testnet: {
+    // multisig roots
     secp256k1Multisig: 'tbtc',
+    ed25519Multisig: 'txlm',
+    // MPC roots
     ecdsaMpc: 'hteth',
     eddsaMpc: 'tsol',
-    ed25519Multisig: 'talgo',
   },
 };
 
@@ -103,15 +107,16 @@ export class Safes implements ISafes {
     const { safeId, passphrase } = params;
     const enterprise = this.enterpriseId;
 
-    // Slots are ordered to match the destructuring below; keep the two lists in sync.
-    const slots: RootKeyType[] = ['secp256k1Multisig', 'ecdsaMpc', 'eddsaMpc', 'ed25519Multisig'];
+    // `slots` MUST stay index-aligned with the Promise.allSettled array below. Ordered by scheme:
+    // the two multisig roots first, then the two MPC roots.
+    const slots: RootKeyType[] = ['secp256k1Multisig', 'ed25519Multisig', 'ecdsaMpc', 'eddsaMpc'];
     const results = await Promise.allSettled([
       // Phase 2.1 — multisig roots (①④): local user/backup keypairs + BitGo key, all safeId-tagged.
       this.createMultisigRoot('secp256k1Multisig', safeId, passphrase, enterprise),
+      this.createMultisigRoot('ed25519Multisig', safeId, passphrase, enterprise),
       // Phase 2.2 — MPC roots (②③): the existing DKLS (②) and EdDSA (③) ceremonies, safeId threaded.
       this.createMpcRoot('ecdsaMpc', safeId, passphrase, enterprise),
       this.createMpcRoot('eddsaMpc', safeId, passphrase, enterprise),
-      this.createMultisigRoot('ed25519Multisig', safeId, passphrase, enterprise),
     ]);
 
     // Single pass over the settled results: `status === 'fulfilled'` narrows `.value` to a
@@ -154,7 +159,7 @@ export class Safes implements ISafes {
    * POST /api/v2/enterprise/:eId/safes/:safeId/archive
    * @experimental
    */
-  private async archiveSafe(safeId: string): Promise<Safe> {
+  async archiveSafe(safeId: string): Promise<Safe> {
     const response = await this.bitgo
       .post(this.url(`/${safeId}/archive`))
       .send()
--- a/modules/sdk-core/src/bitgo/safe/safes.ts
+++ b/modules/sdk-core/src/bitgo/safe/safes.ts
@@ -4,13 +4,37 @@
  * @experimental The safe client surface is experimental and may change (including breaking
  * changes) before the public release.
  */
-import { InitializeSafeBody, SafeData } from '@bitgo/public-types';
+import { FinalizeSafeBody, InitializeSafeBody, RootKeyTriplet, RootKeyType, SafeData } from '@bitgo/public-types';
+import { Environments } from '../../common';
+import { IBaseCoin } from '../baseCoin';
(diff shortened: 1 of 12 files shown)


davidkaplanbitgo:
@claude

davidkaplanbitgo:
Addressed the type-safety point and clarified the network logic; leaving the rest as-is with rationale below.

**Addressed**
- **Type assertion (`safes.ts`)** — removed the `as PromiseFulfilledResult<...>` cast. The settled results are now consumed in a single pass where `status === 'fulfilled'` narrows `.value` to a `RootKeyTriplet`, so the `hot` triplets are built without any assertion.
- **`@experimental` markings** — added `@experimental` to every new safe symbol (options types, `ISafes`/`Safes` methods, `ROOT_COIN_BY_NETWORK`) and the `safeId` threading additions, since TypeDoc doesn't inherit the tag from the class/interface to members.

**Not worth changing (with reasons)**
- **Extract error-aggregation to a utility** — there's only one caller; extracting now is speculative (YAGNI). It's a 6-line local block. Happy to extract when a second use appears.
- **"Network detection may not scale" (`coinForRoot`)** — not a bug. `V1Network` is exactly `'bitcoin' | 'testnet'` (`types.ts`), and every `Environments` entry uses one of those two, so the ternary is exhaustive. Added a comment noting the invariant rather than restructuring.
- **Extract the ceremony count (`4`) into a constant** — the count isn't hardcoded anywhere; it falls out of `failures.length` / the results array. There's no magic number to name.
- **Debug logging for ceremony progress** — this module has no logging convention to match; adding ad-hoc logging here would be inconsistent. Can revisit if we add structured logging to the safe flows.
- **JSDoc for `ROOT_COIN_BY_NETWORK`** — it already has a doc comment explaining the curve-scoped coin selection.

davidkaplanbitgo:
@claude re-review and look at the comments to see if everything was either changed or addressed

modules/sdk-core/src/bitgo/safe/safes.ts
```diff
@@ -4,14 +4,38 @@
...
+    secp256k1Multisig: 'btc',
+    ecdsaMpc: 'eth',
+    eddsaMpc: 'sol',
+    ed25519Multisig: 'algo',
+  },
+  testnet: {
+    secp256k1Multisig: 'tbtc',
+    ecdsaMpc: 'hteth',
+    eddsaMpc: 'tsol',
+    ed25519Multisig: 'talgo',
```
zahin-mohammad: I think we typically use xlm as the representative coin when doing root keys. There might be some documents on confluence on this, lets discuss on slack.

modules/sdk-core/src/bitgo/safe/safes.ts
```diff
@@ -55,24 +87,165 @@ export class Safes implements ISafes {
...
+   * @experimental
    */
   async createSafeKeys(params: CreateSafeOptions & SafeCreationHandle): Promise<SafeKeys> {
-    throw new Error('Safes.createSafeKeys is not yet implemented (WCN-1192 Phase 2)');
+    const { safeId, passphrase } = params;
+    const enterprise = this.enterpriseId;
+
+    // Slots are ordered to match the destructuring below; keep the two lists in sync.
+    const slots: RootKeyType[] = ['secp256k1Multisig', 'ecdsaMpc', 'eddsaMpc', 'ed25519Multisig'];
+    const results = await Promise.allSettled([
```
zahin-mohammad, line 113: Nit: would be good to have the multisig roots beside each other then the mpc roots, or vice verca. Bothers me that ed25519Multisig is at the bottom LOL

modules/sdk-core/src/bitgo/safe/safes.ts
```diff
@@ -55,24 +87,165 @@ export class Safes implements ISafes {
...
+    return { rootKeys: { hot } };
+  }
+
+  /**
+   * Archive a safe. Also the abandonment path for a stuck `initializing` safe (used by
+   * `createSafeKeys` to clean up after a failed ceremony).
+   * POST /api/v2/enterprise/:eId/safes/:safeId/archive
+   * @experimental
+   */
+  private async archiveSafe(safeId: string): Promise<Safe> {
```
zahin-mohammad: I think we can make this public. Although the API has not been created just yet.

modules/sdk-core/src/bitgo/safe/safes.ts
```diff
@@ -55,24 +87,165 @@ export class Safes implements ISafes {
...
+  /**
+   * Coin used to route a given root's key ceremony (see ROOT_COIN_BY_NETWORK).
+   * @experimental
+   */
+  private coinForRoot(slot: RootKeyType): IBaseCoin {
+    // V1Network is exactly 'bitcoin' | 'testnet', so this branch is exhaustive over every
+    // environment: 'bitcoin' is mainnet and every other value is a testnet.
+    const network = Environments[this.bitgo.getEnv()].network === 'bitcoin' ? 'mainnet' : 'testnet';
+    return this.bitgo.coin(ROOT_COIN_BY_NETWORK[network][slot]);
+  }
```
zahin-mohammad, line 197: This means when using the modular sdk, developers that use `safes` must register the representative coins first. Thats fine, just something we need to keep in mind when creating the safe guides in dev portal.

modules/sdk-core/src/bitgo/safe/safes.ts
```diff
@@ -55,24 +87,165 @@ export class Safes implements ISafes {
...
+    });
+    const backupKeychainPromise = keychains.createBackup({ passphrase, safeId });
+    const bitgoKeychainPromise = keychains.createBitGo({ enterprise, safeId });
+
+    const [userKeychain, backupKeychain, bitgoKeychain] = await Promise.all([
+      userKeychainPromise,
+      backupKeychainPromise,
+      bitgoKeychainPromise,
+    ]);
+    return [userKeychain.id, backupKeychain.id, bitgoKeychain.id];
```
zahin-mohammad, line 229: I think we will have to likely change these return values to return the keychain for when we get into keycard generation. Future problem, just calling that out. Its important for MPC, as MPC generates a reducedPrv that does not exist server side.

modules/sdk-core/src/bitgo/utils/tss/ecdsa/ecdsaMPCv2.ts
```diff
@@ -642,8 +654,20 @@ export class EcdsaMPCv2Utils extends BaseEcdsaUtils {
...
+      MPCv2PartiesEnum.BITGO,
+      commonKeychain,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
```
zahin-mohammad, line 668: Lol

zahin-mohammad (commented):
lgtm, just need to check the multisig ed representative coin

zahin-mohammad (approved):

PR #9288, opened 2026-07-18
fix(abstract-substrate): skip recipient validation for token enablement txns

nvrakesh06:
## Summary

- `verifyTransaction` in `abstract-substrate` reads `txBuilder['_to']` to validate the recipient address against `txParams.recipients`. This works for transfer transactions (`TransferBuilder` sets `_to`), but `PreApproveAssetBuilder` / `V8PreApproveAssetBuilder` (used for token enablement / `preApproveAsset`) only has `_assetId` — so `_to` is `undefined`.
- Since `buildTokenEnablements` populates `buildParams.recipients` for TSS wallets, the recipient check fires and throws `TxIntentMismatchRecipientError` comparing the wallet's root address against `undefined`.
- Guard the recipients block on `txTo !== undefined` so non-transfer transaction types are correctly skipped.

**Root trigger:** The v8 transfer routing fix (CECHO-1471, `36d878be05`) made `factory.from(txHex)` correctly route v8-encoded `preApproveAsset` txns to `V8PreApproveAssetBuilder`, which surfaced this missing guard. Without the routing fix, v8 token enablement txns would fail at decode rather than reaching this check.

## Test plan

- [ ] New unit tests added in `sdk-coin-polyx/test/unit/polyx.ts` under `verifyTransaction > token enablement (preApproveAsset) transaction`
- [ ] All existing `verifyTransaction` tests continue to pass (`11 passing`)
- [ ] Manually reproduced on staging wallet `6a5b5835b08b0b019fe89a7313b3cbaa` — token enablement for `tpolyx:nvbitgot` was failing with `TxIntentMismatchRecipientError` before this fix

CECHO-1471

### DIFF
--- a/modules/abstract-substrate/src/abstractSubstrateCoin.ts
+++ b/modules/abstract-substrate/src/abstractSubstrateCoin.ts
@@ -157,6 +157,13 @@ export class SubstrateCoin extends BaseCoin {
     if (prebuildMaterial && typeof factory.material === 'function') {
       factory.material(prebuildMaterial);
     }
+    // Token enablement (preApproveAsset) transactions have no destination address — they are
+    // self-transactions signed by the wallet. Skip transfer recipient validation, consistent
+    // with how XRP and Algo handle enabletoken in their verifyTransaction implementations.
+    if (txParams.type === 'enabletoken') {
+      return true;
+    }
+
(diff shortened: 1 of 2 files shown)


nayandas190 (approved):
lgtm

PR #9295, opened 2026-07-20
feat(sdk-core): extend upgradeEncryption to support OFC wallets

pranavjain97:
- handle variable key count (OFC has 2 keys, no backup)
- skip PDF when backup/bitgo keychains absent
- include pub in user key PUT for OFC server validation

Ticket: WAL-1694

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -3486,11 +3486,10 @@ export class Wallet implements IWallet {
     }
 
     const keyIds = this.keyIds();
-    const [userKeychain, backupKeychain, bitgoKeychain] = await Promise.all([
-      keychainsApi.get({ id: keyIds[0] }),
-      keychainsApi.get({ id: keyIds[1] }),
-      keychainsApi.get({ id: keyIds[2] }),
-    ]);
+    const allKeychains = await Promise.all(keyIds.map((id) => keychainsApi.get({ id })));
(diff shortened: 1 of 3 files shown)


modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -3486,10 +3486,11 @@ export class Wallet implements IWallet {
...
     const keyIds = this.keyIds();
+    const fetchKeychain = (id: string | undefined) => (id ? keychainsApi.get({ id }) : Promise.resolve(undefined));
     const [userKeychain, backupKeychain, bitgoKeychain] = await Promise.all([
-      keychainsApi.get({ id: keyIds[0] }),
-      keychainsApi.get({ id: keyIds[1] }),
-      keychainsApi.get({ id: keyIds[2] }),
+      keychainsApi.get({ id: keyIds[KeyIndices.USER] }),
+      fetchKeychain(keyIds[KeyIndices.BACKUP]),
+      fetchKeychain(keyIds[KeyIndices.BITGO]),
     ]);
```
zahin-mohammad: Index 1 is not always a backup key. OFC keys have index 0 as user and index 1 as bitgo.

zahin-mohammad: Instead of relying on the index, use the source field on the keychain as the authority on which keys are user/backup/bitgo.

zahin-mohammad (approved):

PR #9324, opened 2026-07-22
feat(sdk-coin-sui): add fundsInAddressBalance support to StakingBuilder (AddStake)

abhijit0943:
## What

- Added `fundsInAddressBalance(amount: string)` method to `StakingBuilder` so wallets holding only address-balance SUI can stake
- When `fundsInAddressBalance > 0`, AddStake PTB inserts `redeem_funds(SUI)` + `mergeCoins(gas, [addrCoin])` before `splitCoins` + `request_add_stake`
- `StakingTransaction.toJson()` now emits `fundsInAddressBalance` (from `tx.expiration ?? { None: null }` and `tx.fundsInAddressBalance`); `getTxData()` handles `BalanceWithdrawal` inputs and uses `tx.expiration ?? { None: null }`
- `StakingBuilder.initBuilder()` restores `_fundsInAddressBalance` from deserialized `BalanceWithdrawal` input and restores `_expiration` if `ValidDuring`
- `StakingBuilder.buildSuiTransaction()` passes `expiration: this._expiration` through to the returned `SuiTransaction`
- Fixed `utils.getStakeRequests()` to only parse MoveCalls ending with `::sui_system::request_add_stake`, preventing crash on `redeem_funds` MoveCall (no `arguments[2]`)
- Fixed `utils.getWalrusStakeWithPoolRequests()` with same target filter (`::staking::stake_with_pool`) for defensive parity
- Fixed `WalrusStakingBuilder` to use `WalrusStakingTransaction` instead of `StakingTransaction` (fixes `loadInputsAndOutputs` calling the correct `getWalrusStakeWithPoolRequests`)
- Removed now-unused `StakingTransaction` import from `walrusStakingBuilder.ts`

## Why

- Wallets holding only address-balance SUI (no `Coin<SUI>` objects) cannot stake because `StakingBuilder` sourced the stake amount exclusively from gas coin objects. This mirrors the `fundsInAddressBalance` path in `TransferBuilder` Path 2c.
- `StakingTransaction.toJson()`/`getTxData()` previously hardcoded `expiration: { None: null }` — this would silently discard any `ValidDuring` expiration set by WP for Case 2 (addr-bal-only, empty `gasData.payment`). Fixed to use `tx.expiration ?? { None: null }`.
- The `getStakeRequests` bug would cause a runtime crash when parsing a staking transaction that includes a `redeem_funds` MoveCall (no `arguments[2]`).

## Test plan

- [x] New `fundsInAddressBalance` suite: mixed coin-objects + addr-bal command sequence, round-trip deserialization, multi-stake, default behavior unchanged
- [x] New Case 2 test: `fundsInAddressBalance > 0` + `gasData.payment = []` (addr-bal-only stake)
- [x] All 6 existing Sui Staking Builder tests still pass
- [x] All Walrus Staking Builder tests pass (fixed regression from `getStakeRequests` target filter)
- [x] 148 tests passing, 2 pre-existing failures unrelated to this change

Ticket: SI-1053

### DIFF
--- a/modules/sdk-coin-sui/src/lib/utils.ts
+++ b/modules/sdk-coin-sui/src/lib/utils.ts
@@ -383,7 +383,10 @@ export class Utils implements BaseUtils {
         const amountInputIdx = ((transaction as SplitCoinsTransaction).amounts[0] as TransactionBlockInput).index;
         amounts.push(utils.getAmount(tx.inputs[amountInputIdx] as TransactionBlockInput));
       }
-      if (transaction.kind === 'MoveCall') {
+      if (
+        transaction.kind === 'MoveCall' &&
+        (transaction as MoveCallTransaction).target.endsWith('::sui_system::request_add_stake')
+      ) {
         const validatorAddressInputIdx = ((transaction as MoveCallTransaction).arguments[2] as TransactionBlockInput)
           .index;
         const validatorAddress = utils.getAddress(tx.inputs[validatorAddressInputIdx] as TransactionBlockInput);
@@ -406,7 +409,10 @@ export class Utils implements BaseUtils {
         const amountInputIdx = ((transaction as SplitCoinsTransaction).amounts[0] as TransactionBlockInput).index;
         amounts.push(utils.getAmount(tx.inputs[amountInputIdx] as TransactionBlockInput));
       }
-      if (transaction.kind === 'MoveCall') {
+      if (
+        transaction.kind === 'MoveCall' &&
+        (transaction as MoveCallTransaction).target.endsWith('::staking::stake_with_pool')
+      ) {
         const validatorAddressInputIdx = ((transaction as MoveCallTransaction).arguments[2] as TransactionBlockInput)
           .index;
         const validatorAddress = utils.getAddress(tx.inputs[validatorAddressInputIdx] as TransactionBlockInput);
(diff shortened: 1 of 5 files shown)


modules/sdk-coin-sui/src/lib/walrusStakingBuilder.ts
```diff
@@ -30,7 +30,7 @@ export class WalrusStakingBuilder extends TransactionBuilder<WalrusStakingProgra
 
   constructor(_coinConfig: Readonly<CoinConfig>) {
     super(_coinConfig);
-    this._transaction = new StakingTransaction(_coinConfig);
+    this._transaction = new WalrusStakingTransaction(_coinConfig);
```
benkelcher-bitgo, line 33: good catch

modules/sdk-coin-sui/src/lib/utils.ts
```diff
@@ -406,7 +409,10 @@ export class Utils implements BaseUtils {
         const amountInputIdx = ((transaction as SplitCoinsTransaction).amounts[0] as TransactionBlockInput).index;
         amounts.push(utils.getAmount(tx.inputs[amountInputIdx] as TransactionBlockInput));
       }
-      if (transaction.kind === 'MoveCall') {
+      if (
+        transaction.kind === 'MoveCall' &&
+        (transaction as MoveCallTransaction).target.endsWith('::staking::stake_with_pool')
```
benkelcher-bitgo, line 414: is this check necessary yet given that we are not updating the WAL staking flow right now? (we will need it later anyway so not a big deal)

benkelcher-bitgo (approved):

modules/sdk-coin-sui/src/lib/utils.ts
abhijit0943: yup added it for WAL

Gautam2305 (approved):

PR #9333, opened 2026-07-22
ci: replace invalid osv-scanner severity flags

sachushaji:
osv-scanner v2.3.8 has no --severity flag, so #9331 broke the audit step with 'flag provided but not defined: -severity'. Replace it with a JSON output scan plus a composite-action gate that fails only on findings with CVSS >= 7.0 (HIGH/CRITICAL), which is what #9331 intended. Medium/low findings are logged as warnings and no longer block releases.

TICKET: VL-7134

TICKET: VL-7134

### DIFF
--- a/.github/actions/osv-severity-gate/action.yml
+++ b/.github/actions/osv-severity-gate/action.yml
@@ -1,0 +1,78 @@
+name: OSV severity gate
+description: >
+  Fails the build only when osv-scanner results contain a vulnerability at or
+  above the CVSS threshold (default 7.0, i.e. HIGH/CRITICAL). osv-scanner has
+  no severity filter of its own and exits non-zero on ANY finding, so the scan
+  step must run with `continue-on-error: true`, output JSON to a file, and let
+  this gate decide pass/fail.
+
+inputs:
+  scan-outcome:
+    description: Outcome of the osv-scanner step (`steps.<id>.outcome`)
+    required: true
+  results-file:
+    description: Path to the osv-scanner JSON results file
+    required: false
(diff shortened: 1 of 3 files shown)


davidkaplanbitgo (approved):

zahin-mohammad (approved):
Approve — correct, well-structured hotfix. #9331's `--severity=HIGH/CRITICAL` isn't a valid v2.3.8 `scan` flag, so the audit step has been erroring (no `continue-on-error`) and breaking every release/publish since it merged. This restores the gate.

Verified:
- `--severity` is not a v2.3.8 scan flag (only `--min-severity` on `fix`); `--format json` / `--output-file` are correct.
- `groups[].max_severity` exists (results.go:301) — jq path is right; CVSS ≥ 7.0 == HIGH+CRITICAL matches intent.
- The pinned action forwards `scan-args` verbatim, no injected format/output flags.
- Fail-closed on scan errors (invalid/missing JSON → exit 1); working tree cleaned via `rm -f`.

Non-blocking:
1. Fail-open on unscored advisories: by construction, any advisory with empty `max_severity` (no CVSS vector) can't block — warning only. Pre-#9331 blocked on any finding, so an unscored-but-severe advisory that used to block now passes. Intended direction, but worth a conscious sign-off.
2. Release/publish-only path is untestable in PR CI — recommend a `workflow_dispatch` dry-run before the next real release relies on it.
3. Trivial: hardcoded `WCN-1550` in the warning string will go stale; commit/PR body repeats `TICKET: VL-7134`.

PR #9335, opened 2026-07-22
feat(sdk-core,express): plumb userKeySigningRequired through generateWallet

zahin-mohammad:
## Summary
- Add optional `userKeySigningRequired?: boolean` to `GenerateWalletOptions` and `GenerateGoAccountWalletOptionsCodec` in `@bitgo/sdk-core`, forwarded into the `coinSpecific` payload of `POST /wallet/add` when creating OFC trading wallets via `generateGoAccountWallet`.
- Add the same field to the Express `GenerateWalletBody` io-ts codec so `POST /api/v2/{coin}/wallet/generate` accepts and forwards it to the SDK.
- CAAS clients create OFC wallets through Express or the SDK, not by hitting the raw wallet-platform endpoint directly. The `userKeySigningRequired` flag accepted at wallet creation time (WCN-1441) was unreachable via the public SDK/Express surface, leaving clients no way to opt out of user-key signing at creation.

## Test plan
- [ ] `yarn unit-test` in `modules/sdk-core` passes (new `walletsGoAccount.ts` suite: 3 tests — `false` forwarded, `true` forwarded, omitted → no `coinSpecific`)
- [ ] `yarn unit-test` in `modules/express` passes (new typed-route tests: happy path + codec rejection of non-boolean)
- [ ] Integration: SDK `generateWallet({ type: 'trading', userKeySigningRequired: false })` on an OFC coin → created wallet has `coinSpecific.ofc.userKeySigningRequired === false`
- [ ] Integration: `userKeySigningRequired` omitted → `coinSpecific` absent from request body (default preserved)

Ticket: WCN-1442

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallets.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallets.ts
@@ -234,7 +234,7 @@ export class Wallets implements IWallets {
     const reqId = new RequestTracer();
     this.bitgo.setRequestTracer(reqId);
 
-    const { label, passphrase, enterprise, passcodeEncryptionCode, encryptionVersion } = params;
+    const { label, passphrase, enterprise, passcodeEncryptionCode, encryptionVersion, userKeySigningRequired } = params;
 
     const keychain = this.baseCoin.keychains().create();
 
@@ -259,6 +259,9 @@ export class Wallets implements IWallets {
       type: 'trading',
       enterprise,
       keys: [userKeychain.id],
+      ...(userKeySigningRequired !== undefined && {
+        coinSpecific: { userKeySigningRequired },
+      }),
     };
 
     const newWallet = await this.bitgo.post(this.baseCoin.url('/wallet/add')).send(walletParams).result();
--- a/modules/express/src/typedRoutes/api/v2/generateWallet.ts
+++ b/modules/express/src/typedRoutes/api/v2/generateWallet.ts
@@ -55,6 +55,8 @@ export const GenerateWalletBody = {
       passphrase: t.string,
     })
   ),
+  /** OFC only. When false, BitGo signs OFC payloads on behalf of the user without requiring a user-key signature. Defaults to true (user must sign). */
+  userKeySigningRequired: optional(t.boolean),
 } as const;
 
 export const GenerateWalletResponse200 = t.union([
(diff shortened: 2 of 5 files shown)


modules/sdk-core/src/bitgo/wallet/wallets.ts
```diff
@@ -259,6 +259,9 @@ export class Wallets implements IWallets {
       type: 'trading',
       enterprise,
       keys: [userKeychain.id],
+      ...(userKeySigningRequired !== undefined && {
+        coinSpecific: { [this.baseCoin.getChain()]: { userKeySigningRequired } },
```
rishikeshdadam136: I think BitGoJS sends a different coinSpecific shape than what the BGMS reads
This line builds something like this:`{ "coinSpecific": { "ofc": { "userKeySigningRequired": false } } }`
BGMS ([PR](https://github.com/BitGo/bitgo-microservices/pull/61152)) does not read `coinSpecific.ofc.userKeySigningRequired` from the create-wallet request. It reads the flat path something like this:`body.coinSpecific.userKeySigningRequired`
Ref:https://github.com/BitGo/bitgo-microservices/blob/develop/packages/wallet-platform/app/controllers/api/v2/coins/ofc.ts#L549

With the nested body, that read is undefined, so BGMS treats the flag as omitted and CAAS still gets the default userKeySigningRequired = true.

rishikeshdadam136:
@claude can you please review this PR? Also, could you check this specific comment (https://github.com/BitGo/BitGoJS/pull/9335#discussion_r3637722627) and let me know if it is valid?

modules/sdk-core/src/bitgo/wallet/wallets.ts
zahin-mohammad: good catch

zahin-mohammad: Good catch — confirmed. BGMS reads the flat path `body.coinSpecific.userKeySigningRequired` (see `extractAdditionalCreateWalletData` in ofc.ts). Fixed in 0644990: `wallets.ts` now builds `coinSpecific: { userKeySigningRequired }` (dropped the chain-nesting), and the unit tests assert the flat shape.

zahin-mohammad:
@rishikeshdadam136 Thanks for the review. The flagged `coinSpecific` structure mismatch was valid and is now fixed (commit `805a095`):

- `wallets.ts` builds the flat shape `coinSpecific: { userKeySigningRequired }` — matching what BGMS reads in `extractAdditionalCreateWalletData` (`body.coinSpecific.userKeySigningRequired`). The chain-nesting is gone.
- Unit tests updated to assert the flat shape; all pass.
- Branch rebased onto latest master.

rishikeshdadam136 (approved):
Looks Good

rishikeshdadam136 (approved):
Looks Good

PR #9365, opened 2026-07-27
fix(sdk-core): return clear 401 for wrong wallet passphrase

shubham-damkondwar:
## Summary
- Return `IncorrectPasswordError` (HTTP 401, `wallet_passphrase_incorrect`) for wrong wallet passphrase instead of a generic 500 / Internal Server Error
- Stop string-wrapping TSS signing failures so original API/status/details are preserved
- Fix Express `retryPromise` and `decryptPrivKey` to keep actionable passphrase vs malformed-ciphertext errors

Ticket: [COINS-1257](https://linear.app/bitgo/issue/COINS-1257/robinhood-icp-withdrawal-api-not-working)

Note: Separate from BitGo Internal Agent PR #9344 (ICP broadcast certificate parsing), which addresses a different failure mode than this ticket’s remaining ask.

## Test plan
- [x] Express unit tests for wrong passphrase decrypt → 401 (`coinSign`, `generateShareTSS`, `ofcExtSignPayload`)
- [x] `retryPromise` preserves original error without wrapping
- [x] `signPayload` passphrase vs malformed ciphertext cases
- [ ] Manual: `POST /api/v2/icp/wallet/:id/signtxtss` with wrong `walletPassphrase` returns 401 + `code: wallet_passphrase_incorrect`

### DIFF
--- a/modules/sdk-core/src/bitgo/wallet/wallet.ts
+++ b/modules/sdk-core/src/bitgo/wallet/wallet.ts
@@ -2562,7 +2562,7 @@ export class Wallet implements IWallet {
       }
       userPrv = await decryptKeychainPrivateKey(this.bitgo, userKeychain, params.walletPassphrase);
       if (!userPrv) {
-        throw new Error('failed to decrypt user keychain');
+        throw new IncorrectPasswordError();
       }
     }
     return userPrv;
@@ -4813,16 +4813,16 @@ export class Wallet implements IWallet {
     await this.tssUtils.deleteSignatureShares(txRequestId, reqId);
 
     try {
-      const signedTxRequest = await this.tssUtils.signEddsaTssUsingExternalSigner(
+      return await this.tssUtils.signEddsaTssUsingExternalSigner(
         txRequestId,
         params.customCommitmentGeneratingFunction,
         params.customRShareGeneratingFunction,
         params.customGShareGeneratingFunction,
         reqId
       );
-      return signedTxRequest;
     } catch (e) {
-      throw new Error('failed to sign transaction ' + e);
+      debug('failed to sign transaction %O', e);
+      throw e;
     }
   }
 
@@ -4919,9 +4919,9 @@ export class Wallet implements IWallet {
       throw new Error('Generator function for S share required to sign transactions with External Signer.');
     }
 
+    assert(this.tssUtils, 'tssUtils must be defined');
     try {
-      assert(this.tssUtils, 'tssUtils must be defined');
-      const signedTxRequest = await this.tssUtils.signEcdsaTssUsingExternalSigner(
+      return await this.tssUtils.signEcdsaTssUsingExternalSigner(
         {
           txRequest: txRequestId,
           reqId: params.reqId || new RequestTracer(),
@@ -4932,9 +4932,9 @@ export class Wallet implements IWallet {
         params.customMuDeltaShareGeneratingFunction,
         params.customSShareGeneratingFunction
       );
-      return signedTxRequest;
     } catch (e) {
-      throw new Error('failed to sign transaction ' + e);
+      debug('failed to sign transaction %O', e);
+      throw e;
     }
   }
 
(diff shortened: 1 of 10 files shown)


shubham-damkondwar:
@claude

parasgarg-bitgo (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -4812,18 +4812,13 @@ export class Wallet implements IWallet {
...
-        params.customRShareGeneratingFunction,
-        params.customGShareGeneratingFunction,
-        reqId
-      );
-      return signedTxRequest;
-    } catch (e) {
-      throw new Error('failed to sign transaction ' + e);
-    }
+    return this.tssUtils.signEddsaTssUsingExternalSigner(
+      txRequestId,
```
kisslove-dewangan: this would be changing the error log for other cases as well

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -4919,23 +4914,18 @@ export class Wallet implements IWallet {
...
-        params.customKShareGeneratingFunction,
-        params.customMuDeltaShareGeneratingFunction,
-        params.customSShareGeneratingFunction
-      );
-      return signedTxRequest;
-    } catch (e) {
-      throw new Error('failed to sign transaction ' + e);
-    }
+    assert(this.tssUtils, 'tssUtils must be defined');
+    return this.tssUtils.signEcdsaTssUsingExternalSigner(
```
kisslove-dewangan: same

modules/sdk-core/src/bitgo/wallet/wallet.ts
```diff
@@ -4974,21 +4964,16 @@ export class Wallet implements IWallet {
...
-        params.customEddsaMPCv2SigningRound3GenerationFunction
-      );
-      return signedTxRequest;
-    } catch (e) {
-      throw new Error('failed to sign transaction ' + e);
-    }
+    assert(this.tssUtils, 'tssUtils must be defined');
+    return this.tssUtils.signEddsaMPCv2TssUsingExternalSigner(
+      {
+        txRequest: txRequestId,
```
kisslove-dewangan: same

davidkaplanbitgo (dismissed):

modules/sdk-core/src/bitgo/wallet/wallet.ts
shubham-damkondwar: Good catch — you're right that removing the `catch` outright changed the log surface for every failure in those methods, not just the wrong-passphrase case.

I couldn't simply restore the wrapper though, because it was the actual cause of this bug. The chain for a wrong passphrase in external-signer mode is:

1. External signer correctly returns **401** (`wallet_passphrase_incorrect`)
2. Comes back as a superagent error with `.status = 401`
3. `retryPromise` did `throw new Error(err)` — discards the status (fixed in this PR)
4. Reaches these `catch` blocks: `throw new Error('failed to sign transaction ' + e)` — **discards the status a second time**
5. Express sees no `.status`, falls back to its default → **500**

So `throw new Error(...)` in step 4 is load-bearing for the bug: it constructs a fresh error and drops `.status`/`.code` off the original.

Fixed by keeping the breadcrumb and rethrowing the original error unchanged:

```ts
try {
  return await this.tssUtils.signEddsaTssUsingExternalSigner(...);
} catch (e) {
  debug('failed to sign transaction %O', e);
  throw e;
}
```

`debug` was already imported and used in ~10 places in this file, so it matches existing style. Note `return` → `return await` is required — otherwise the promise escapes the `try` and the `catch` never fires.

Applied to all **5** sites with this pattern, including the two you didn't comment on (`signTransactionTssExternalSignerECDSAMPCv2` and `signTransactionTss`), since leaving those inconsistent would defeat the point.

Verified: sdk-core 529 passing, express 1913 passing, `bitgo` wallet suite 0 failures, tsc + eslint + prettier clean.

parasgarg-bitgo (approved):

vibhavgo (approved):
LGTM

OttoAllmendinger (approved):

PR #9373, opened 2026-07-28
fix: reject safe root creation on legacy MPCv1 ceremonies

davidkaplanbitgo:
MPCv1 silently ignores safeId, so keys never get tagged and WP's finalize step fails with a generic 400. Fail fast in createMpc instead.

Also propagate the server's Request-ID header through ApiResponseError so failed safe ceremonies can be correlated with server-side logs.

TICKET: WCN-1539

TICKET: WCN-1539


s84krish (approved):
lgtm

PR #9380, opened 2026-07-29
Onboard client requested tokens 

prajwalu142:
Ticket: CSHLD-1278


modules/statics/src/allCoinsAndTokens.ts
```diff
@@ -7627,6 +7706,14 @@ export const allCoinsAndTokens = [
...
   ),
+  zkethErc20(
+    'f684b83f-f2b7-4bff-9347-4510063b4809',
+    'zketh:zk',
+    'zkSync',
+    18,
+    '0x5a7d6b2f92c77fad6ccabd7ee0624e64907eaf3e',
+    UnderlyingAsset['zketh:zk']
+  ),
   tzkethErc20(
```
yashvanthbl137-crypto, line 7717: `CELO:CELO` is mentioned in ticket but not onboarded

prajwalu142: token with same contract address already onboarded as Celo gold, commenting the same in ticket

yashvanthbl137-crypto: we have same symbol for token and native coin?

yashvanthbl137-crypto (approved):
LGTM

modules/statics/src/allCoinsAndTokens.ts
prajwalu142: Yes

PR #9385, opened 2026-07-29
fix(sdk-coin-flrp): compute Flare C-chain import fee from real gas usage

ArunBala-Bitgo:
## Summary
- Flare (FLR/coston2) C-chain atomic imports were built with `evm.newImportTx(fee)`, treating the caller-supplied fee as a literal amount and never accounting for the ~10,000 gas `AtomicTxBaseCost` that flarejs's gas estimator applies internally.
- When a caller priced the import using an underestimated gas figure (e.g. ~5,700 gas instead of the real ~11,300), the network rejected the broadcast with insufficient funds — independent of transfer amount, matching the reported failure mode.
- Added an opt-in `baseFee()` builder method that instead calls `evm.newImportTxFromBaseFee(...)`, letting flarejs compute the real gas cost (including `AtomicTxBaseCost`) from the actual tx, and applies a 10% padding buffer on the supplied base fee to absorb volatility between signing and broadcast. This mirrors the pattern the export builder (`ExportInCTxBuilder`) already uses via `newExportTxFromBaseFee`.
- The existing literal `fee()` path is untouched for backward compatibility with existing (on-chain verified) test vectors.

## Test plan
- [x] `npx mocha --exclude test/unit/flrp.ts 'test/unit/**/*.ts'` in `modules/sdk-coin-flrp` — 275 passing
- [x] New tests confirm the base-fee-derived fee exceeds a naive gas*baseFee estimate and includes padding
- [x] `tsc --noEmit` and `eslint` clean (no new warnings)

Ticket: CECHO-1821

### DIFF
--- a/modules/sdk-coin-flrp/src/lib/ImportInCTxBuilder.ts
+++ b/modules/sdk-coin-flrp/src/lib/ImportInCTxBuilder.ts
@@ -14,9 +14,19 @@ import {
 import utils from './utils';
 import { DecodedUtxoObj, FlareTransactionType, SECP256K1_Transfer_Output, Tx } from './iface';
 
+/**
+ * Buffer applied to a caller-supplied base fee to absorb C-chain base-fee volatility
+ * between transaction signing and broadcast. Expressed in basis points (1000 = 10%).
+ */
+const BASE_FEE_PADDING_BPS = 1000n;
+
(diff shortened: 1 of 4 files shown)


ArunBala-Bitgo:
@claude review this PR

mohd-kashif (approved):

PR #9408, opened 2026-08-04
refactor(abstract-substrate): extract MPCv2 helpers into SubstrateCoin

vibhavgo:
## Summary

Extracts `isMpcv2SigningMaterial()` and `addSubstrateRecoverySignature()` as protected methods on `SubstrateCoin` so DOT (WCI-1227) and POLYX (WCI-1228) can call shared helpers rather than duplicating ~40 lines of identical MPCv2 signing logic.

**Linear**: [WCI-1239](https://linear.app/bitgo/issue/WCI-1239)

## Changes

- Add `protected isMpcv2SigningMaterial()` — detects MPCv2 vs MPCv1 keycard by inverting `isEddsaMpcV1SigningMaterial`
- Add `protected addSubstrateRecoverySignature()` — dispatches MPCv1/MPCv2 signing; MPCv2 path prepends `0x00` Ed25519 discriminant required by Substrate `MultiSignature` enum
- Refactor `recover()` to delegate to both helpers; MPCv1 path unchanged
- Add unit tests in `modules/abstract-substrate/test/unit/abstractSubstrateCoin.ts` using `Ttao` as the concrete `SubstrateCoin` subclass

## Test Plan

- `npx mocha` in `modules/abstract-substrate` — 6 passing
- `yarn unit-test` in `modules/sdk-coin-tao` — 91 passing, no regressions
- `yarn unit-test` in `modules/sdk-coin-polyx` — 252 passing, no regressions

### DIFF
--- a/modules/abstract-substrate/src/abstractSubstrateCoin.ts
+++ b/modules/abstract-substrate/src/abstractSubstrateCoin.ts
@@ -25,6 +25,8 @@ import {
   UnexpectedAddressError,
   verifyEddsaTssWalletAddress,
   VerifyTransactionOptions,
+  EDDSAUtils,
+  decryptKeychainPrivateKey,
 } from '@bitgo/sdk-core';
 import { CoinFamily, BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
 import { KeyPair as SubstrateKeyPair, Transaction } from './lib';
@@ -37,6 +39,12 @@ import BigNumber from 'bignumber.js';
 import { ApiPromise } from '@polkadot/api';
 
 export const DEFAULT_SCAN_FACTOR = 20;
+
+/**
+ * Discriminated union carrying keycard version and decrypted V1 user key (to avoid re-decryption).
+ * V1 keycards are JSON; V2 keycards are CBOR-encoded reduced key shares.
+ */
+type SubstrateSigningMaterial = { version: 'v1'; userPrv: string } | { version: 'v2'; encryptedUserKey: string };
 
 export class SubstrateCoin extends BaseCoin {
   protected readonly _staticsCoin: Readonly<StaticsBaseCoin>;
@@ -356,42 +364,17 @@ export class SubstrateCoin extends BaseCoin {
         throw new Error('missing wallet passphrase');
       }
 
-      const userKey = params.userKey.replace(/\s/g, '');
-      const backupKey = params.backupKey.replace(/\s/g, '');
-
-      // Decrypt private keys from KeyCard values
-      let userPrv;
-      try {
-        userPrv = await this.bitgo.decrypt({
-          input: userKey,
-          password: params.walletPassphrase,
-        });
-      } catch (e) {
-        throw new Error(`Error decrypting user keychain: ${e.message}`);
-      }
-      const userSigningMaterial = JSON.parse(userPrv) as EDDSAMethodTypes.UserSigningMaterial;
-
-      let backupPrv;
-      try {
-        backupPrv = await this.bitgo.decrypt({
-          input: backupKey,
-          password: params.walletPassphrase,
-        });
-      } catch (e) {
-        throw new Error(`Error decrypting backup keychain: ${e.message}`);
-      }
-      const backupSigningMaterial = JSON.parse(backupPrv) as EDDSAMethodTypes.BackupSigningMaterial;
-
-      // add signature
-      const signatureHex = await EDDSAMethods.getTSSSignature(
-        userSigningMaterial,
-        backupSigningMaterial,
+      const signingMaterial = await this.isMpcV2Keycard(params.userKey!, params.walletPassphrase!);
+      await this.addSubstrateRecoverySignature(
+        txBuilder,
+        signingMaterial,
+        params.backupKey!.replace(/\s/g, ''),
+        params.walletPassphrase!,
+        unsignedTransaction,
         currPath,
-        unsignedTransaction
+        bitgoKey,
+        accountId
       );
-
-      const substrateKeyPair = new SubstrateKeyPair({ pub: accountId });
-      txBuilder.addSignature({ pub: substrateKeyPair.getKeys().pub }, signatureHex);
       const signedTransaction = await txBuilder.build();
       serializedTx = signedTransaction.toBroadcastFormat();
     } else {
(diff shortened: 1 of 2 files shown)


at31416 (dismissed):
lgtm

modules/abstract-substrate/src/abstractSubstrateCoin.ts
```diff
@@ -356,42 +357,16 @@ export class SubstrateCoin extends BaseCoin {
...
-        throw new Error(`Error decrypting backup keychain: ${e.message}`);
-      }
-      const backupSigningMaterial = JSON.parse(backupPrv) as EDDSAMethodTypes.BackupSigningMaterial;
-
-      // add signature
-      const signatureHex = await EDDSAMethods.getTSSSignature(
-        userSigningMaterial,
-        backupSigningMaterial,
+      const isMpcV2 = await this.isMpcv2SigningMaterial(params.userKey, params.walletPassphrase);
+      await this.addSubstrateRecoverySignature(
```
nvjsr, line 368: isMpcv2SigningMaterial() decrypts the user key, and then addSubstrateRecoverySignature() decrypts it again for MPCv1.

A cleaner approach would be to detect the version while obtaining the decrypted material, or return richer information:
```

type SigningMaterialType =
  | { version: 'v1'; signingMaterial: UserSigningMaterial }
  | { version: 'v2' };
```

or simply have getSigningMaterialVersion(...) return enough metadata so the second decrypt isn't needed.

modules/abstract-substrate/src/abstractSubstrateCoin.ts
```diff
@@ -356,42 +357,16 @@ export class SubstrateCoin extends BaseCoin {
...
-      } catch (e) {
-        throw new Error(`Error decrypting backup keychain: ${e.message}`);
-      }
-      const backupSigningMaterial = JSON.parse(backupPrv) as EDDSAMethodTypes.BackupSigningMaterial;
-
-      // add signature
-      const signatureHex = await EDDSAMethods.getTSSSignature(
-        userSigningMaterial,
-        backupSigningMaterial,
+      const isMpcV2 = await this.isMpcv2SigningMaterial(params.userKey, params.walletPassphrase);
```
nvjsr: isMpcv2SigningMaterial() is slightly misleading.The function isn't checking signing material itself—it determines the key format/version.

`isMpcV2Keycard()`or`isMpcV2RecoveryMaterial()`

modules/abstract-substrate/src/abstractSubstrateCoin.ts
```diff
@@ -526,6 +501,87 @@ export class SubstrateCoin extends BaseCoin {
...
+        commonKeyChain
+      );
+      // 0x00 = Ed25519 discriminant in Substrate MultiSignature enum
+      const substrateSig = Buffer.concat([Buffer.from([0x00]), rawSig]);
+      txBuilder.addSignature({ pub: substrateKeyPair.getKeys().pub }, substrateSig);
+    } else {
+      let userPrv: string;
+      try {
+        userPrv = await this.bitgo.decrypt({ input: userKey, password: params.walletPassphrase! });
+      } catch (e) {
```
nvjsr: There are three nearly identical blocks:

```
try {
   ...
} catch (e) {
   throw new Error(...)
}
```

for user key, backup key and version detection

This could be centralized into something like `private async decryptKeychain(...)`

modules/abstract-substrate/src/abstractSubstrateCoin.ts
```diff
@@ -526,6 +501,87 @@ export class SubstrateCoin extends BaseCoin {
...
+    txBuilder: NativeTransferBuilder,
+    params: MPCRecoveryOptions,
+    unsignedTransaction: Transaction,
+    currPath: string,
+    bitgoKey: string,
+    accountId: string,
+    isMpcV2: boolean
+  ): Promise<void> {
+    const userKey = params.userKey!.replace(/\s/g, '');
+    const backupKey = params.backupKey!.replace(/\s/g, '');
```
nvjsr: There are
```
params.userKey!
params.backupKey!
params.walletPassphrase!
```

If these are required for recovery, it may be better to validate once at the start:

`assertRecoveryParams(params);`

Then the rest of the method can safely use non-optional values without repeated !.

modules/abstract-substrate/src/abstractSubstrateCoin.ts
```diff
@@ -526,6 +501,87 @@ export class SubstrateCoin extends BaseCoin {
...
+      }
+      const rawSig = await EDDSAUtils.signRecoveryEddsaMPCv2(
+        unsignedTransaction.signablePayload,
+        currPath,
+        userKeyShare,
+        backupKeyShare,
+        commonKeyChain
+      );
+      // 0x00 = Ed25519 discriminant in Substrate MultiSignature enum
+      const substrateSig = Buffer.concat([Buffer.from([0x00]), rawSig]);
```
nvjsr: can we do like this 
```
const ED25519_MULTI_SIGNATURE_PREFIX = 0x00;

Buffer.from([ED25519_MULTI_SIGNATURE_PREFIX])
```

nvjsr (changes requested):

modules/abstract-substrate/src/abstractSubstrateCoin.ts
vibhavgo: Added `SubstrateSigningMaterial`

nvjsr (dismissed):

modules/abstract-substrate/src/abstractSubstrateCoin.ts
```diff
@@ -526,6 +508,93 @@ export class SubstrateCoin extends BaseCoin {
...
+  protected async isMpcV2Keycard(userKey: string, walletPassphrase: string): Promise<SubstrateSigningMaterial> {
+    const normalized = userKey.replace(/\s/g, '');
+    const userPrv = await this.decryptKeychain(normalized, walletPassphrase, 'user');
+    let isV1: boolean;
+    try {
+      const m = JSON.parse(userPrv);
+      isV1 =
+        typeof m?.uShare?.seed === 'string' &&
+        typeof m?.bitgoYShare?.u === 'string' &&
+        (typeof m?.backupYShare?.u === 'string' || typeof m?.userYShare?.u === 'string');
```
Marzooqa: Can we make use of `EDDSAUtils.isEddsaMpcV1SigningMaterial` here?

modules/abstract-substrate/src/abstractSubstrateCoin.ts
```diff
@@ -526,6 +508,93 @@ export class SubstrateCoin extends BaseCoin {
...
+  /**
+   * Decrypts an encrypted keychain value, wrapping errors with a descriptive message.
+   */
+  private async decryptKeychain(encryptedKey: string, passphrase: string, label: string): Promise<string> {
+    try {
+      return await this.bitgo.decrypt({ input: encryptedKey, password: passphrase });
+    } catch (e) {
+      throw new Error(`Error decrypting ${label} keychain: ${e instanceof Error ? e.message : String(e)}`);
+    }
+  }
```
Marzooqa, line 521: I think we already have a utils we could use instead? `decryptKeychainPrivateKey` [here](https://github.com/BitGo/BitGoJS/blob/7e845c3041a08f8e7ade472eb43a5b997507d9ec/modules/sdk-core/src/bitgo/keychain/decryptKeychain.ts)

vibhavgo: Used

modules/abstract-substrate/src/abstractSubstrateCoin.ts
```diff
@@ -526,6 +509,89 @@ export class SubstrateCoin extends BaseCoin {
...
+    }
+    return { version: 'v2', encryptedUserKey: normalized };
+  }
+
+  /**
+   * Adds an MPCv1 or MPCv2 signature to a Substrate transaction builder.
+   * MPCv2 signatures are prefixed with ED25519_MULTI_SIGNATURE_PREFIX (Ed25519 discriminant
+   * in the Substrate MultiSignature enum).
+   */
+  protected async addSubstrateRecoverySignature(
```
shubham-damkondwar, line 570: This method has no unit tests. It's the critical signing path — it branches on MPCv1 vs MPCv2, prepends the `0x00` Ed25519 discriminant for MPCv2, and calls into `EDDSAUtils.signRecoveryEddsaMPCv2`. A bug here would silently produce an invalid signature at recovery time.

The existing tests only cover `isMpcV2Keycard()`. Would suggest adding at least:
- MPCv2 path: verify the `0x00` prefix is prepended to the raw signature
- MPCv1 path: verify `getTSSSignature` is called and the result passed to `addSignature`
- `commonKeyChain` mismatch: verify the guard throws

vibhavgo: Since this was a refactor to create common utility for children of the substrate class 
Earlier had decided to add tests in the child consumer PRs for MPCv2 path
But I agree, will also add tests here

Marzooqa (approved):
has a follow-up

hitansh-madan (approved):

PR #9412, opened 2026-08-04
fix: update vulnerable transitive dependencies

shubham-damkondwar:
WCN-1550

TICKET: CECHO-1286


zahin-mohammad:
Heads up — I hit the same beta-publish failure and opened #9414 for it before spotting this draft. Same four packages; two deltas worth pulling in either direction:

**`fast-uri`: 3.1.5 rather than 4.1.2.** GHSA-7p8r-x3mc-p8w7 is fixed on both lines (`3.x → 3.1.5`, `4.x → 4.1.2`). The only consumer is `ajv@^8`, which declares `fast-uri@^3.0.1`, so pinning 4.1.2 forces ajv across a major it never declared. 3.1.5 clears the advisory inside the declared range, and upstream still maintains 3.x (`three` dist-tag = 3.1.5, published 2026-07-31).

**`ip-address` 10.4.0 needs the socks compat stated.** It lands under `socks@2.7.3` (pinned in root `resolutions`), which declares `ip-address@^9.0.5` — a semver-major override in the proxy path. I verified it in the installed tree: `Address4`/`Address6` constructors, `Address6.fromByteArray().canonicalForm()`, and socks' `ipToBuffer` / `ipv4ToInt32` / `int32ToIpv4` / `validateSocksClientOptions` all behave identically on 10.4.0, and `new Address4('010.0.0.1')` now throws `AddressError` (the fix). So the bump is fine — it just shouldn't be implicit.

Also note #9355 pins `fast-uri` to 3.1.4, which no longer clears this advisory.

Happy for either PR to be the one that lands — whichever is easier, they just shouldn't both. Tracking on WCN-1798.

zahin-mohammad:
Follow-up on the `ip-address` point above — #9414 now upgrades `socks` (2.7.3 → 2.8.9) instead of forcing `ip-address` 10.x underneath the old pin. Upstream socks moved to `ip-address@^10.x` in 2.8.7, so after the bump the lock holds a single `ip-address@^10.1.1` spec resolved to 10.4.0 and no override crosses a declared major. Two extras that fell out of it: the 2.7.3 pin was itself a security floor (`57f2196f59`, WP-1413), so raising it preserves that intent, and it repairs a latent range violation — `socks-proxy-agent@^8` declares `socks@^2.8.3`, which 2.7.3 never satisfied.

Worth mirroring here if this PR is the one that lands.

PR #9414, opened 2026-08-04
WCN-1798: fix(deps): clear 4 HIGH osv advisories blocking beta publish

zahin-mohammad:
## Summary

- `Publish @bitgo-beta` has failed on every run since 2026-08-03 18:50 UTC (latest: [run 30904314719](https://github.com/BitGo/BitGoJS/actions/runs/30904314719)). The `osv-severity-gate` reports `4 of 41 advisory group(s) at or above CVSS 7.0`. `npmjs-release.yml` runs the same gate, so prod releases are blocked too.
- All four advisories were published 2026-08-03 and all have upstream fixes, so no `osv-scanner.toml` exclusion is warranted — this bumps the pins in both `resolutions` and `overrides`.

| Package | Before → after | Advisory | CVSS |
|---|---|---|---|
| `brace-expansion` | 5.0.8 → 5.0.9 | GHSA-rgw5-rvv9-x895 — DoS via unbounded intermediate arrays | 7.5 |
| `fast-uri` | 3.1.3 → 3.1.5 | GHSA-7p8r-x3mc-p8w7 — host confusion via backslash authority introducer | 7.5 |
| `socket.io-parser` | 4.2.6 → 4.2.7 | GHSA-2m8v-j782-fhvr — zero-attachment memory exhaustion | 7.5 |
| `ip-address` | 9.0.5 → 10.4.0 | GHSA-mwp4-54f8-5fhr — `Address4` decodes leading-zero octets as decimal while resolvers decode them as octal (SSRF) | 7.7 |
| `socks` | 2.7.3 → 2.8.9 | — (carries the `ip-address` bump) | — |

### Version choices

**`fast-uri` 3.1.5, not 4.1.2.** OSV lists fixes on both lines (`3.x → 3.1.5`, `4.x → 4.1.2`). The only consumer is `ajv@^8`, which declares `fast-uri@^3.0.1`; 3.1.5 clears the advisory inside the declared range instead of forcing ajv across a major it never declared. Upstream still maintains 3.x (`three` dist-tag = 3.1.5).

**`ip-address` needs a major bump — so `socks` is upgraded rather than overridden underneath.** There is no 9.x fix for GHSA-mwp4-54f8-5fhr. `ip-address`'s only consumer is `socks`, which was pinned to `2.7.3` (declaring `ip-address@^9.0.5`). Forcing 10.x under that pin would be a semver-major override in the proxy path, so instead `socks` moves to **2.8.9** — upstream socks itself switched to `ip-address@^10.x` in 2.8.7. After this change the lock holds a single `ip-address@^10.1.1` spec resolved to 10.4.0; `ip-address@^9.0.5` no longer appears anywhere and no override crosses a declared major.

Three supporting points:

- The `socks: 2.7.3` pin was itself a security **floor**, not a ceiling — `57f2196f59` "fix(root): lock socks version / locked socks version to fix vulnerability" (WP-1413). Raising it preserves that intent, and `socks@2.8.9` returns 0 advisories from the OSV API.
- It repairs a latent range violation: `socks-proxy-agent@^8` declares `socks@^2.8.3`, which `2.7.3` never satisfied. Both agents in the tree — 7.0.0 (`^2.6.2`) and 8.0.5 (`^2.8.3`) — resolve cleanly against 2.8.9.
- The explicit `ip-address: 10.4.0` pin is kept even though socks now declares `^10.1.1`, because that range still permits the vulnerable 10.1.x/10.2.x releases for npm consumers resolving from `overrides`.

## Test plan

- [x] `resolutions` and `overrides` updated together (npm consumers of the `bitgo` umbrella only see `overrides`)
- [x] `yarn.lock` regenerated; `brace-expansion@^1.1.7` and `@^2.0.1` still flatten onto 5.0.9, so the 1.x/2.x affected lines do not reappear
- [x] Lock holds a single `ip-address@^10.1.1` → 10.4.0; no `^9.0.5` spec remains
- [x] All chosen versions return 0 advisories from the OSV API (`api.osv.dev/v1/query`), including `socks@2.8.9`
- [x] `yarn check-deps` passes
- [x] Smoke-tested in the installed tree on `socks@2.8.9` + `ip-address@10.4.0`: `Address4`/`Address6` constructors, `Address6.fromByteArray().canonicalForm()`, socks helpers `ipToBuffer` (v4 → 4 bytes, v6 → 16) / `ipv4ToInt32` / `int32ToIpv4` / `validateSocksClientOptions` (IPv4 proxy + hostname destination, and IPv6 destination), `SocksClient.createConnection` present, `new SocksProxyAgent('socks5://127.0.0.1:1080')` constructs
- [x] Advisory fix observable in the same tree: `new Address4('010.0.0.1')` now throws `AddressError` instead of silently decoding as decimal
- [ ] `osv-scanner` reports 0 advisory groups at or above CVSS 7.0. This PR's own CI cannot show that: `ci.yml` deliberately skips the audit on pull requests ("We conciously do not audit dependencies as a PR step since errors are typically unrelated to the PR changes. This check is performed in `publish.yml`", `.github/workflows/ci.yml:180-181`), and `osv-scanner` is not installable locally. The gate only proves out on the beta publish run after this merges
- [ ] `Publish @bitgo-beta` goes green

## Overlap with in-flight PRs

- **#9412** (draft, `CECHO-1286`) bumps the same four packages. It pins `fast-uri 4.1.2` (forces ajv across a major) and forces `ip-address 10.4.0` under the old `socks@2.7.3` pin rather than upgrading socks. Suggest closing it in favour of this PR, or pulling these version choices into it — either is fine, but the two shouldn't both land.
- **#9355** bumps `fast-uri` to **3.1.4**, which no longer clears GHSA-7p8r-x3mc-p8w7 (needs 3.1.5). Its `tar 7.5.22` bump and 8 dropped exclusions are independent and still valuable, but its `fast-uri` pin needs raising or it will conflict with this change.

Ticket: WCN-1798


lokesh-bitgo (dismissed):

mohammadalfaiyazbitgo (approved):

pranavjain97 (approved):

bhargavirao24 (approved):
AppSec review is completed. 
brace-expansion, fast-uri, and ip-address are the only dependencies within the 7-day window. I checked each one against OSV, GHSA, Scorecard, and provenance. 

Everything looks clean and each update matches a reviewed CVE fix.

Good to merge from my side.

PR #9416, opened 2026-08-04
feat(sdk-core): always upgrade to v2 on password change

pranavjain97:
Flip `updateSingleKeychainPassword` from **preserving** the source envelope version to **always** re-encrypting as v2 (Argon2id + AES-256-GCM). A password change is a natural upgrade point — v1 (SJCL) keychains
  are transparently promoted; v2 stays v2.

  Impact

  - Keychains.updatePassword collects the results and the SDK POSTs them to /user/changepassword (via BitGoAPI.changePassword → sdk-api/src/bitgoAPI.ts:2192), so the new v2 blobs are persisted server-side.
  - Effectively a silent bulk v1→v2 migration for every user who changes their password.
  - No behavior change for v2 users. No breaking API change.

TICKET: WCN-1814


davidkaplanbitgo (approved):
CI failures but LGTM

